PR #1750, opened 2018-06-26
Add docker logo

tjamet:
Hi,

Awesome work there.
May I propose adding a docker badge?


RedSparr0w:
Thanks for submitting this PR,
But i think this logo may be a little too detailed for use in shields badges (it's also quite a large file size):
docker logo badge

tjamet:
yes agree, I have updated the PR with this one

tjamet:
with the white color, this is the rendering, rendered with https://svgtopng.com/

RedSparr0w:
Looks good to me:


Do you think we could remove the ® from down the bottom right?
As it's not legible due to the size anyway, and just adds extra data.

tjamet:
I can have a look at that
I kept it to be compliant with the logo terms of use, let me update the PR

RedSparr0w:
Thanks, Logo looks good,
Merged!
👍

Issue #1708, opened 2018-05-28
Deployment status in PR


platan:
The idea started in https://github.com/badges/shields/pull/1605#issuecomment-376595800. We want "to post a comment at the end of a PR thread, which indicates whether it's been deployed yet".

There are two things we need:
- Badge with Github commit merge status in branch (https://github.com/badges/shields/issues/1616)
- a [probot](https://probot.github.io/) app for creating a comment in closed PR

platan:
I think we have all required building blocks. We need to put them together. 
- Badge with Github commit merge status in branch (#1616) is merged to master #1709. We have to wait for a deployment. Meanwhile, we can use a temporary instance deployed at Now by me
https://shields-deployment.now.sh/github/commit-status/badges/shields/gh-pages/9d90bd1f9549b3c8b59ad8338a8acd92e1e36f3c.svg?label=deploy%20status (https://github.com/badges/shields/commit/9d90bd1f9549b3c8b59ad8338a8acd92e1e36f3c)
![](https://shields-deployment.now.sh/github/commit-status/badges/shields/gh-pages/9d90bd1f9549b3c8b59ad8338a8acd92e1e36f3c.svg?label=deploy%20status)
https://shields-deployment.now.sh/github/commit-status/badges/shields/gh-pages/fe4ac0bf1cbd0dbe04bf2cc439a2eee51f4c3907.svg?label=deploy%20status (https://github.com/badges/shields/commit/fe4ac0bf1cbd0dbe04bf2cc439a2eee51f4c3907)
![](https://shields-deployment.now.sh/github/commit-status/badges/shields/gh-pages/fe4ac0bf1cbd0dbe04bf2cc439a2eee51f4c3907.svg?label=deploy%20status)
- A https://github.com/bkeepers/workflow configuration. I've deployed an instance of the workflow app  https://github.com/apps/shields-deployment and used it in an example repo https://github.com/platan/shields-deployment-status-example, configuration https://github.com/platan/shields-deployment-status-example/blob/master/.github/probot.js, PR examples https://github.com/platan/shields-deployment-status-example/pull/4, https://github.com/platan/shields-deployment-status-example/pull/6. This example repo has branches similar to these in the shields repo (`master` + `gh-pages`). 

What's next?
[...]

PR #1754, opened 2018-06-29
Probot uses badges from img.shields.io

platan:
This is the last thing to be done in https://github.com/badges/shields/issues/1708


platan:
@PyvesB could you please look at this change? I would like to have 2 approves :-)

platan:
Thank you all! :-)

PR #1755, opened 2018-06-30
invalidJSON fixture documented

platan:
This pull request adds info about `invalidJSON` fixture to the service tests documentation. This should increase awareness about existence of this fixture. 

Not all tests which mocks an invalid JSON use this fixture. We can improve it in a separate PR :-)


platan:
@PyvesB could you please look at this change? I would like to have 2 approves :-)

platan:
Oh, @paulmelnikow thanks for improving this :-)

PR #1756, opened 2018-06-30
[cocoapods david lgtm] An invalid JSON is mocked using invalidJSON in all service tests

platan:
This PR standardize mocking a invalid JSON in service tests. This is a follow-up to https://github.com/badges/shields/pull/1755. I tried to find all custom mocks of the invalid JSON using grep, but I could miss something.


platan:
Thanks @chris48s! Fixed, all tests passed :-)

PR #1759, opened 2018-07-02
[scrutinizer] Fix #1739, add tests and refactor

### LINKED ISSUE #1739: fix scrutinizer error handling
At the moment we're doing this: 

screenshot at 2018-06-17 12 23 08

PyvesB:
This pull request covers the following:
* the NaN error reported in #1739 was fixed, as well as the example updated on the homepage.
* a proper test suite was added.
* the service code was refactored. We had three handler functions which were mostly copy-pasted code, which were merged into one. When we get round to modularising this service, we'll be able to break down the bulky _if_ statements into nice separate functions. I also modernised and rearranged the existing code.

I noticed that we had very similar `isBuildStatus` validators in several service tests, I'm planning on extracting them to _test-validators.js_ once this is merged. 😉 

As usual, looking forward to feedback! 👍

### DIFF
--- a/server.js
+++ b/server.js
@@ -2616,7 +2616,7 @@ cache(function(data, match, sendBadge, request) {
           badgeData.colorscheme = 'orange';
         } else if (status === 'unknown') {
           badgeData.colorscheme = 'gray';
-        } 
+        }
       } else {
         let score = parsedData.applications[branch].index._embedded
           .project.metric_values['scrutinizer.quality'];
--- a/server.js
+++ b/server.js
@@ -2598,7 +2598,7 @@ cache(function(data, match, sendBadge, request) {
         const percentage = parsedData.applications[branch].index._embedded
           .project.metric_values['scrutinizer.test_coverage'] * 100;
         if (isNaN(percentage)) {
-          badgeData.text[1] = 'inaccessible';
+          badgeData.text[1] = 'unknown';
           badgeData.colorscheme = 'gray';
         } else {
           badgeData.text[1] = percentage.toFixed(0) + '%';
--- a/server.js
+++ b/server.js
@@ -2569,20 +2569,12 @@ cache(function(data, match, sendBadge, request) {
 }));
 
 // Scrutinizer coverage integration.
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -2569,136 +2569,71 @@ cache(function(data, match, sendBadge, request) {
...
   // The repo may contain a branch, which would be unsuitable.
-  var repoParts = repo.split('/');
-  var branch = null;
-  // Normally, there are 2 slashes in `repo` when the branch isn't specified.
-  var slashesInRepo = 2;
-  if (repoParts[0] === 'gp') { slashesInRepo = 1; }
+  const repoParts = repo.split('/');
+  let branch = null;
+  // Normally, there are 2 slashes in `repo` when the branch is specified.
+  const slashesInRepo = repoParts[0] === 'gp' ? 1 : 2;
```
chris48s: Could this section here where we're working out if a branch has been passed in the URL be done in a slightly cleaner way by using a more sophisticated regex for the URL pattern? ContinuousPHP, Travis or Coveralls might be good badges to look at for an example of this.

server.js
```diff
@@ -2569,136 +2569,71 @@ cache(function(data, match, sendBadge, request) {
...
-        badgeData.colorscheme = 'yellow';
-      } else if (score > 4) {
-        badgeData.colorscheme = 'orange';
+      if (branch === null) { branch = parsedData.default_branch; }
+      if (type === 'coverage') {
+        const percentage = parsedData.applications[branch].index._embedded
+          .project.metric_values['scrutinizer.test_coverage'] * 100;
+        if (isNaN(percentage)) {
+          badgeData.text[1] = 'inaccessible';
+          badgeData.colorscheme = 'gray';
```
chris48s: Is 'inaccessible' the right error message here? Usually we use it when the service can't be contacted. Looks like if we got this far, then we got some kind of response and parsed it..

PyvesB:
@chris48s thanks for reviewing, I have addressed the comments above!

PR #1767, opened 2018-07-08
[appveyor bitbucket buildkite circleci codeship continuousphp docker readthedocs scrutinizer shippable travis] Extracted isBuildStatus test validator

PyvesB:
As mentioned in #1759.

### DIFF
--- a/services/test-validators.js
+++ b/services/test-validators.js
@@ -69,7 +69,9 @@ const isFormattedDate = Joi.alternatives().try(
 
 const isDependencyState = withRegex(/^(\d+ out of date|\d+ deprecated|up to date)$/);
 
-const isBuildStatus = withRegex(/^(building|cancelled|error|expired|fail(ed|ing)|no tests|not (built|run)|passing|pending|processing|queued|running|scheduled|skipped|stopped|success|timeout|unknown|unstable|waiting)$/);
+const isBuildStatus = Joi.equal('building', 'cancelled', 'error', 'expired', 'failed', 'failing', 'no tests',  
+                                'not built', 'not run', 'passing', 'pending', 'processing', 'queued', 'running',  
+                                'scheduled', 'skipped', 'stopped', 'success', 'timeout', 'unstable', 'waiting');
 
 module.exports = {
   isSemver,
--- a/services/test-validators.js
+++ b/services/test-validators.js
@@ -69,6 +69,8 @@ const isFormattedDate = Joi.alternatives().try(
 
 const isDependencyState = withRegex(/^(\d+ out of date|\d+ deprecated|up to date)$/);
 
+const isBuildStatus = withRegex(/^(building|cancelled|error|expired|fail(ed|ing)|no tests|not (built|run)|passing|pending|processing|queued|running|scheduled|skipped|stopped|success|timeout|unknown|unstable|waiting)$/);
+
 module.exports = {
   isSemver,
   isVPlusTripleDottedVersion,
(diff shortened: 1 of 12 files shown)


chris48s:
I have thought about doing this exact thing before, but I decided against it because I'm not convinced it represents a good tradeoff.

Although there are some common terms ('passing', 'failing', 'unknown' etc) that are common across lots of build/CI services, there can also be quite a lot of variance in terminology. Although we can save a bit of duplication here and there, the disadvantage of loading them all into a single regex is that we lose the connection of which terminology is specific to which service:

For example, this has the effect that 'waiting' becomes a valid value for every CI badge just because shippable uses it.

Conversely, if the docker service (for example) becomes deprecated, now the status 'building' is not valid for any badge, so we should remove 'building' as a valid status from the `isBuildStatus` regex. However because we've combined statuses from lots of different services, it is now very non-obvious that this is the case.

PyvesB:
In my opinion we are simply interested in knowing whether the badge returns some sort of build status or not. Some tests that are actually concerned about the exact label have matching checks such as `.expectJSON({name: 'docker build', value: 'passing'})`. The rest of the time, we simply want to check that the label matches the correct vocabulary family, which also allows for more flexibility if the service decides to tweak their returned status text a bit.

But I understand your point of view as well, I'm happy to close this if it doesn't get a consensus, or possibly make things more consistent by using `const` fields everywhere as some tests duplicates calls such `Joi.equal('failing', 'passing', 'unknown')` for the same service. 😃

chris48s:
> happy to close this if it doesn't get a consensus

Yep. Lets get a opinion from another reviewer.. Its a subjective call :)

I'm agreed that we don't need to declare `Joi.equal('failing', 'passing', 'unknown')` this many times and we can usefully reduce some duplication here.

RedSparr0w:
~~I don't see it being too much of  a problem having some redundant regex (like `waiting` if shippable changes/depreciated), But it may be worth adding a comment at the top of the service test stating which they use.~~
~~`// Build Status: passed, failed, building`~~
~~Then it seems it may also just be better off to put the custom regex there instead?~~

~~Or we could just use the isBuildStatus helper for the common options:~~
~~`const isBuildStatus = withRegex(/^(building|fail(ed|ing)|passing|unknown)$/);`~~

👍 from me, if the service is showing any of those options in the regex, chances are high that it is working as expected, it may lead to some redundant matches, but this won't affect the performance of the server and shouldn't really affect performance of the test.

PyvesB:
Let's sleep over this a few extra days and see whether anyone else has a strong opinion.

services/test-validators.js
```diff
@@ -69,6 +69,8 @@ const isFormattedDate = Joi.alternatives().try(
 
 const isDependencyState = withRegex(/^(\d+ out of date|\d+ deprecated|up to date)$/);
 
+const isBuildStatus = withRegex(/^(building|cancelled|error|expired|fail(ed|ing)|no tests|not (built|run)|passing|pending|processing|queued|running|scheduled|skipped|stopped|success|timeout|unknown|unstable|waiting)$/);
```
paulmelnikow: This doesn't need to be a regex and might be more legible with `Joi.equal()`.

paulmelnikow:
I don't feel strongly either way about this change.

Don't feel like having a few superfluous strings in there is a problem. However, I think the strings in here should be undeniably build statuses, not strings that signify the absence of a build status. In particular, let's remove **unknown** which sounds more like a nondescript error condition. If there's a good reason for a service to include it, it could extend it using `Joi.alternatives()`.

PR #1771, opened 2018-07-08
add tests for composer version compare function

chris48s:
There have been various discussions about whether we should change the behaviour of our code for comparing versions from composer. For the moment I'm not going to propose any functional changes to this, but one of the problems we have here is that that our code for comparing and ordering composer versions is non-trivial and has no tests. This makes it difficult to change.

This PR helps this by adding tests to formalise the 'happy path' behaviour of the `compare()` function (i.e: when processing valid version strings). Note that `compare()` exhibits some spurious behaviour when processing invalid versions. For example:

```js
> const { compare } = require("./lib/php-version");

> compare('1.0.0-aardvarks', '1.0.0-alpha');  // 1.0.0-aardvarks is not a valid version
0                                             // ..but we say it is equal to 1.0.0-alpha
```

.. but I'm going to abstract that issue for now. For the moment I just want to get this code under test so we can validate that proposed changes are not breaking correct behaviour.

### DIFF
--- a/lib/php-version.spec.js
+++ b/lib/php-version.spec.js
@@ -2,6 +2,7 @@
 
 const { test, given } = require('sazerac');
 const {
+  compare,
   minorVersion,
   versionReduction
 } = require('./php-version');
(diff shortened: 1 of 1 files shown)


paulmelnikow:
Thanks!

Issue #1458, opened 2018-01-17
Badges in footer have disappeared | Frontend not up to date


RedSparr0w:
Badges in the footer have dissapeared:
![image](https://user-images.githubusercontent.com/7288322/35024569-15127696-fba5-11e7-957c-86dc9c54db19.png)
How it should look:
![image](https://user-images.githubusercontent.com/7288322/35024542-f1494bfe-fba4-11e7-99fc-b61762aa3146.png)

Looks like they are trying to load the badges from `https://shields.io/` instead of `https://img.shields.io/`

_Sidenote: what is the `:wq` for?_

**Edit:**
Looks like the frontend hasn't been rebuilt from master?
As it is still using the gratipay badge instead of opencollective:
![image](https://user-images.githubusercontent.com/7288322/35024814-822b9982-fba6-11e7-8290-009201a0bf14.png)

CAD97:
> _Sidenote: what is the `:wq` for? this?_

How to exit vim: https://stackoverflow.com/q/11828270/3019990

paulmelnikow:
I mentioned this to @espadrine on the 13th and just sent him a follow up email.

All the maintainers have deploy access to the frontend, though I don't know exactly which commit is deployed to the servers.

paulmelnikow:
> Looks like they are trying to load the badges from `https://shields.io/` instead of `https://img.shields.io/`

This part is kind of puzzling. Since the old frontend is deployed, I'm wondering: did a change in the server cause these to stop working? ^^

espadrine:
The reason for that is likely to be because this link points nowhere, perhaps because the new build process with next requires adding a new folder, the build folder, in the git project when doing `make gh-pages`?

paulmelnikow:
Which links points nowhere?

Ah, right, if the folder doesn't exist, the build process creates it, however the deploy process doesn't ship it.

Would you like me to open a PR to fix the script, and test the (frontend) deploy in the process? If so, which commit?

espadrine:
I tried simply adding build in `git add -f build index.html` in the Makefile in deploy-gh-pages, but it turns out that when doing `npm run build`, the index.html file is not at the root of the repository.

Without it at the root, it won't be shown as the main page of https://badges.github.io/shields, which means it won't show up as shields.io (since shields.io maps to GitHub Pages through the CNAME trick, which lets us rely on GitHub for webpage hosting).

[...]

PR #1774, opened 2018-07-12
Fixes for github deploy

### LINKED ISSUE #1784: Missing favicon
It looks like we are missing `favicon.png` on the main website.

**Edit:** the staging server still has the favicon

paulmelnikow:
This addresses some long-standing comments in https://github.com/badges/shields/issues/1458#issuecomment-368270996

We should also adopt Gatsby, create-react-app, or something similar designed for static sites, to eliminate the unnecessary runtime dependency on Next.js while also letting someone else maintain our front-end tooling. :-D These alternative tools might work just fine in subdirectories without config, and we might be able to leave Jekyll turned on (though we don’t need it). However these git-related changes are orthogonal.

- Don’t check out master, making it possible to deploy the currently checked-out commit
- Disable Jekyll which we don’t need. This allows _next folders to be deployed, and the related URL rewriting to be removed.
- ~~Set NEXT_ASSET_PREFIX so the site works at both shields.io and badges.github.io/shields~~
    - This was possible in my github account, but doesn't seem to be possible in the badges account.
- Completely empty the deploy branch’s index before deployment. This prevents errors from broken symlinks, while preserving the commit history in the deploy branch.
- Do the deployment work in a git working tree. This requires Git 2.18 but makes it possible to do the above very safely.

Closes #1784

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ DEPLOY_TEMP=${TMPDIR}shields-deploy
 all: website favicon test
 
 favicon:
+	# This isn't working right now. See https://github.com/badges/shields/issues/1788
 	node lib/badge-cli.js '' '' '#bada55' .png > favicon.png
 
 website:
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ favicon:
 website: favicon
 	LONG_CACHE=false npm run build
 
-deploy: website deploy-s0 deploy-s1 deploy-s2 deploy-gh-pages deploy-gh-pages-clean
+deploy: deploy-s0 deploy-s1 deploy-s2 favicon deploy-gh-pages deploy-gh-pages-clean
 
 deploy-s0:
 	# Ship a copy of the front end to each server for debugging.
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 SHELL:=/bin/bash
 
+DEPLOY_TEMP=${TMPDIR}shields-deploy
+
 all: website favicon test
 
 favicon:
@@ -8,7 +10,7 @@ favicon:
 website:
 	LONG_CACHE=false npm run build
 
-deploy: website deploy-s0 deploy-s1 deploy-s2 deploy-gh-pages
+deploy: website deploy-s0 deploy-s1 deploy-s2 deploy-gh-pages deploy-gh-pages-clean
 
 deploy-s0:
 	# Ship a copy of the front end to each server for debugging.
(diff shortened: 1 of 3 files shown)


paulmelnikow:
Would like to get this addressed for #1783. Could I entreat a review?

RedSparr0w:
I'll give it a go, not sure how well this pc will handle it though.

paulmelnikow:
Thanks! Just reading it would be okay. If you do decide to run it, please do that on a fork so it doesn’t deploy something in the process.

Makefile
```diff
@@ -7,7 +7,7 @@ all: website favicon test
 favicon:
 	node lib/badge-cli.js '' '' '#bada55' .png > favicon.png
```
RedSparr0w: Is it just me, or does this no longer produce anything?
Also without the `.png` it produces this for me:

paulmelnikow: Huh, weird. `node lib/badge-cli.js '' '' '#bada55' > favicon.svg` produces the right svg for me. No quotes… wonder if that could be a windows thing? Is it better if `'' ''` is replaced with `"" ""` in the makefile?

However the png is not working for me either. It makes a valid PNG but it's all gray. I upgraded imagemagick, and it changed to a different gray (!) but still isn't working correctly.

paulmelnikow: Opened a new issue: #1788

Makefile
```diff
@@ -34,17 +38,26 @@ deploy-s2:
...
+	LONG_CACHE=true \
+		BASE_URL=https://img.shields.io \
+		NEXT_ASSET_PREFIX=https://shields.io \
+		npm run build
+	git worktree add -B gh-pages ${DEPLOY_TEMP}
+	git -C ${DEPLOY_TEMP} ls-files | xargs git -C ${DEPLOY_TEMP} rm
+	git -C ${DEPLOY_TEMP} commit -m '[DEPLOY] Completely clean the index'
+	cp -r build/* ${DEPLOY_TEMP}
+	cp favicon.png ${DEPLOY_TEMP}
+	echo shields.io > ${DEPLOY_TEMP}/CNAME
```
espadrine: The point of the CNAME file was to have GitHub redirect from badges.github.io/shields to shields.io.

I believe it no longer needs that, as it is now set up through the project settings: https://help.github.com/articles/adding-or-removing-a-custom-domain-for-your-github-pages-site/

paulmelnikow: Something in my testing suggested the file still has an effect. When I ran this script on my fork without the CNAME file in gh-pages, it turned off the custom domain.

This link suggests changes made in the UI are stored in the repo: https://help.github.com/articles/troubleshooting-custom-domains/#github-repository-setup-errors

Issue #1776, opened 2018-07-13
Jenkins tests always returns 'invalid'


HansBrende:
The following markup:

`[![Jenkins tests](https://img.shields.io/jenkins/t/https/builds.apache.org/Any23-trunk.svg?maxAge=86400)](https://builds.apache.org/job/Any23-trunk)`

produces:

[![Jenkins tests](https://img.shields.io/jenkins/t/https/builds.apache.org/Any23-trunk.svg?maxAge=86400)](https://builds.apache.org/job/Any23-trunk)

RedSparr0w:
Closing as fixed via #1777

PR #1777, opened 2018-07-13
fix for invalid [jenkins] tests

HansBrende:
This fixes issue #1776. The issue was caused because the following api call: 

https://builds.apache.org/job/Any23-trunk/lastBuild/api/json?tree=actions[failCount,skipCount,totalCount]

returns a 400 BAD REQUEST due to the "invalid characters" `[` and `]`.  

I simply URL-encoded these characters.


HansBrende:
@chris48s How's that?

PR #1743, opened 2018-06-19
[RFC] Add Joi-based request validation to BaseJsonService and rewrite [NPM] and [node] badges; also affects [apm appveyor cdnjs clojars gem]

paulmelnikow:
When JSON responses come back, they are sometimes not in the format expected by the API. As a result we have a lot of defensive coding (expressions like `(data || {}).someProperty`) to avoid exceptions being thrown in badge processing. Often we rely on the `try` blocks that wrap so much of the badge-processing code, which catch all JavaScript exceptions and return some error result, usually **invalid**. The problem with this is that these `try` blocks catch all sorts of programmer errors too, so when we see **invalid** we don't know whether the API returned something unexpected, or we've made a mistake. We also spend a lot of time writing defensive tests around malformed responses, and creating and maintaining the defensive coding.

A better solution is to validate the API responses using declarative contracts. Here the programmer says exactly what they expect from the API. That way, if the response isn't what we expect we can just say it's an **invalid json response**. And if our code then throws an exception, well that's our mistake; when we catch that we can call it a **shields internal error**. It's also less code and less error-prone. Over time we may be confident enough in the contracts that we won't need so many tests of malformed responses. The contract doesn't need to describe the entire response, only the part that's needed. Unknown keys can simply be dropped, preventing unvalidated parts of the response from creeping into the code. Checking what's in our response before calling values on it also makes our code more secure.

I used Joi here, since we're already using it for testing. There may be another contracts library that's a better fit, though I think we could look at that later.

Those changes are in base.js.

The rest is a rewrite of the remaining NPM badges, including the extraction of an NpmBase class. Inspired by @chris48s's work in #1740, this class splits the service concerns into fetching and plucking, validation, and rendering. There are two URL patterns, one which allows specifying a tag (used by e.g. the version badge `https://img.shields.io/npm/v/npm/next.svg`), and the other which does not accept a tag (e.g. the license badge `https://img.shields.io/npm/l/express.svg`). Subclasses like NpmLicense and NpmTypeDefinitions can specify the URL fragment, examples, the validation schema for the chunk of the package data they use, and a render function. The NpmVersion subclass uses a different endpoint, so it overrides the `handle` implementation from NpmBase.

Rewriting these was curiously difficult. There were a few factors, I think: a large paradigm shift, lots of code duplication to sift through, red herring abstractions (abstractions that already exist but are actually unhelpful in the new paradigm), and new problems to solve (namely validation). I leaned heavily on the thorough tests. Thank goodness for those!

[...]

PR #1790, opened 2018-07-20
Fix edge case for service detection in PR titles [cran] [crates]

paulmelnikow:
Service detection didn’t work correctly for PR titles with multiple bracketed entries.

e.g. https://circleci.com/gh/badges/shields/6797

### DIFF
--- a/lib/service-test-runner/pull-request-services-cli.js
+++ b/lib/service-test-runner/pull-request-services-cli.js
@@ -15,9 +15,9 @@
 
 'use strict';
 
-const difference = require('lodash.difference');
 const fetch = require('node-fetch');
 const { inferPullRequest } = require('./infer-pull-request');
+const servicesForTitle = require('./services-for-title');
 
 async function getTitle (owner, repo, pullRequest) {
   let uri = `https://api.github.com/repos/${owner}/${repo}/pulls/${pullRequest}`;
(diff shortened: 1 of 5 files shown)


chris48s:
Slightly difficult to tell if this has worked as expected because the Circle build says "This is not a pull request. Skipping." on the service tests. What would be the expected behaviour with the title "Fix edge case for service detection in PR titles [cran] [crates]" - should it run the tests for cran and crates, or only crates?

paulmelnikow:
Huh, okay. Well, if you look at this slightly earlier build, you can see it "working", except that one of the services I picked the first time has no tests: https://circleci.com/gh/badges/shields/6836

The expected behavior is that both sets of tests run.

The observed behavior is completely broken. The regex greedily grabs the entire string between the outermost brackets. On #1743 it would attempt to test a service called `rfc], add, joi-based, request, validation, to, basejsonservice, and, rewrite, [npm`.

You can see it here: https://circleci.com/gh/badges/shields/6797

chris48s:
Cool. That sounds sensible. Can we formalise that with a test for that case in `services-for-title.spec.js` as well then, so we're testing something like:

```js
given('make changes to [CRAN] and [CPAN]').expect([
  'cran',
  'cpan',
]);
```

paulmelnikow:
Sure, I can add one more test.

paulmelnikow:
Added! I was thinking this case was already covered by the test I had, though that's with the implementation in mind. From the outside it's a different case.

PR #1791, opened 2018-07-22
Fix time service

paulmelnikow:
To fix failing build: https://circleci.com/gh/badges/shields/6860?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link


chris48s:
Oh yeah, of course.. That makes sense: https://github.com/chris48s/shields/blob/issue1723/services/base.js#L15

PR #1793, opened 2018-07-22
Rewrite server deploy script

paulmelnikow:
~~I haven't run this yet but I will do that when I run the deploy later today. Would appreciate an "eyeballs" review first!~~

To run this requires renaming `private/secret.json` to `private/secret-production.json` in the working tree used for deployment.

Goals:

- Ensure production secrets are not used in development
- Avoid modifying the current working tree
- Avoid branch switching: make sure the current ref gets deployed
    - If something other than `master` is deployed, leave `HEAD` alone; don't reset to `master`
- Ensure the build runs before server deploy (#1941)

This makes use of Git working trees, which is a relatively new but stable feature in Git. I was initially reluctant to use git worktree, mostly because I don't like adding new tooling that isn't necessary. The other alternative I experimented with was copying or re-cloning to an entirely separate working copy. This was messier and more brittle than using `git worktree`.

cc @espadrine


Makefile
```diff
@@ -11,52 +12,59 @@ favicon:
...
 	LONG_CACHE=false npm run build
 
-# `website` is needed for the server deploys.
-deploy: website deploy-s0 deploy-s1 deploy-s2 deploy-gh-pages deploy-gh-pages-clean
+deploy: deploy-s0 deploy-s1 deploy-s2 clean-server-deploy deploy-gh-pages deploy-gh-pages-clean
 
-deploy-s0:
+deploy-s0: prepare-server-deploy push-s0
+deploy-s1: prepare-server-deploy push-s1
+deploy-s2: prepare-server-deploy push-s2
```
espadrine: All those are meant to be `.PHONY`, right?

Makefile
```diff
@@ -11,52 +12,59 @@ favicon:
...
+	rm -rf ${SERVER_TMP}
+	git worktree prune
+	git worktree add -B production ${SERVER_TMP}
+	cp -r build ${SERVER_TMP}
+	git -C ${SERVER_TMP} add build/
+	git -C ${SERVER_TMP} commit -m '[DEPLOY] Add frontend for debugging'
+	cp private/secret-production.json ${SERVER_TMP}/private/secret.json
+	cp Verdana.ttf ${SERVER_TMP}
+	git -C ${SERVER_TMP} add private/secret.json Verdana.ttf
+	git -C ${SERVER_TMP} commit -m '[DEPLOY] MUST NOT BE ON GITHUB'
```
espadrine: That is a significant change in behavior. It feels like it does more than just use secret-production.json instead of secret.json. Could you explain this change further please?

Makefile
```diff
@@ -11,52 +12,59 @@ favicon:
...
 	# Ship a copy of the front end to each server for debugging.
 	# https://github.com/badges/shields/issues/1220
-	git add -f Verdana.ttf private/secret.json build/
-	git commit -m'MUST NOT BE ON GITHUB'
-	git push -f s0 HEAD:master
-	git reset HEAD~1
-	git checkout master
+	rm -rf ${SERVER_TMP}
+	git worktree prune
+	git worktree add -B production ${SERVER_TMP}
```
espadrine: The use of git worktree feels a bit like tempting the devil. The use of `-B` in particular risks losing information without realizing.

 In general, I prefer having deployment procedures so dumb that it takes a lot for them to go wrong, and when they do, they are dumb enough that it is simple to fix them.

paulmelnikow:
> The use of git worktree feels a bit like tempting the devil. The use of `-B` in particular risks losing information without realizing.
>
> In general, I prefer having deployment procedures so dumb that it takes a lot for them to go wrong, and when they do, they are dumb enough that it is simple to fix them.

Yes, the deploy process should be safe, simple, and reliable. Using worktree makes it safer, by not doing the work inside a tree, and in a way throws away less because the deploy branch sticks around after the deploy is done. It also lets you push the same build to each machine rather than running a new build for each one.

Let's not rush this discussion; I'll circle back.

Makefile
paulmelnikow: Ah yes, thanks.

Makefile
paulmelnikow: Without a working tree, the solution as far as I can see is to:

1. Back up the existing `private/secret.json`, if it exists
2. Copy `private/secret-production.json` to `private/secret.json`
3. Check it in
4. Restore the original `private/secret.json` afterward

That has more moving parts than I'd like, seems brittle, and also seems like it could result in accidental data loss.

By moving the branch manipulation into a separate working tree, `private/secret.json` can be copied for deployment without needing to preserve the original.

The other change is how the branching is handled, which I'll respond to below.

paulmelnikow:
I can confirm that the current scripts, when aborted in the middle, leave a deploy commit hanging on the working copy. When I reset to last commit from master, Verdana.ttf, the server config, and the build are removed from the working copy.

Any thoughts about my comments?

Makefile
paulmelnikow: #1941 is another issue with the current scripts. `git worktree` has been working more reliably for `deploy-gh-pages` than the current deploy is for the servers.

paulmelnikow:
I continue to encounter issues where I accidentally run things using production credentials while deploying. I feel it's necessary to move forward with this. I will update it in light of #2178 and then it would be great if someone would please give a cursory 👍to merge it.

chris48s:
This shouldn't be a blocker for this PR, but for the front-end part of the build/deploy, have you ever looked at this package: https://github.com/tschaub/gh-pages ?

paulmelnikow:
> As an additional risk mitigation, do you think we should set up any branch names used in deployment with a branch protection rule in https://github.com/badges/shields/settings/branches to prevent us accidentally pushing it to GH?

It's not a bad idea. I just added branch protection for `server-deploy-working-branch` – see https://github.com/badges/shields/pull/1793#discussion_r204255558 – and I'll update this to use the same.

paulmelnikow:
Okay, this is updated.

I'll re-test it on a real deploy before merging.

PR #1794, opened 2018-07-22
List the named logos in the frontend

### LINKED ISSUE #1787: Named logo link not working
The [named logos](https://github.com/badges/shields/tree/gh-pages/logo) link recreated here from the [shields.io](https://shields.io/) website directs to a 404 error.

paulmelnikow:
This probably could be accomplished using a webpack config, however this approach is consistent with the way we handle badge examples and I like that it's very obvious what's happening.

I like @chris48s's idea of putting these on their own page, though we should resolve #1762 first. This seems like a good incremental improvement. 

Fix #1787

screen shot 2018-07-22 at 5 28 30 pm


chris48s:
Good call. I suspect #1762 may drag on for a bit, so this is good for for an interim solution.

One small thing is that the long list does push the content out very wide in a fullscreen window:

**before**:

screenshot at 2018-07-23 19 50 39

**after**:

screenshot at 2018-07-23 19 48 33

Do you think its worth putting this in a `<div>` with a `max-width` or something to keep it a bit more centred?

paulmelnikow:
Good call. I've updated in d7065ee and it'll be deployed here in a moment: https://shields-staging-pr-1794.herokuapp.com/

paulmelnikow:
Thanks!

PR #1796, opened 2018-07-23
Provide a descriptive message when json can’t be parsed [cdnjs clojars gem]

paulmelnikow:
This message **unparseable json response** is consistent with the error messaging in #1743. That's a fair amount of test, though that doesn't seem like a big problem since it's not something we expect users to see terribly often.

### DIFF
--- a/services/cdnjs/cdnjs.tester.js
+++ b/services/cdnjs/cdnjs.tester.js
@@ -3,7 +3,6 @@
 const Joi = require('joi');
 const ServiceTester = require('../service-tester');
 const { isVPlusTripleDottedVersion } = require('../test-validators');
-const { invalidJSON } = require('../response-fixtures');
 
 const t = new ServiceTester({ id: 'cdnjs', title: 'CDNJs' });
 module.exports = t;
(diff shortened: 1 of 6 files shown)


services/cdnjs/cdnjs.tester.js
```diff
@@ -31,7 +31,7 @@ t.create('cdnjs (unexpected response)')
     .get('/libraries/jquery?fields=version')
     .reply(invalidJSON)
   )
-  .expectJSON({name: 'cdnjs', value: 'invalid'});
+  .expectJSON({ name: 'cdnjs', value: 'unparseable json response' });
```
paulmelnikow: At this point, if we could get some coverage of this path in the BaseJsonService tests, I'd be comfortable getting rid of all these `invalidJSON` tests on new-style services. Thoughts?

chris48s: Good point. I reckon for badges/services that just use standardized error handling provided by the base classes we can dispense with testing the standard error cases in every service tester. Where we've done something custom we should still test that.

If we are going to do that, can we make sure those standard cases are covered in `base.spec.js`.

paulmelnikow: Have updated this, removing those test cases and adding a new test for BaseJsonService.

As `base.js` is getting long, I'm thinking we probably should break out BaseJsonService to its own module, so I put the tests in a new file.

PR #1797, opened 2018-07-23
Tweak layout of services to conveniently handle large service families [appveyor cdnjs clojars gem npm]

paulmelnikow:
As discussed here:
https://github.com/badges/shields/pull/1743#issue-195899374
https://github.com/badges/shields/pull/1743#issuecomment-398901078

Todo:
- [x] Adjust danger rules


services/index.js
```diff
@@ -2,14 +2,9 @@
...
 
-// Match modules with the same name as their containing directory.
-// e.g. services/appveyor/appveyor.js
-const serviceRegex = /\/services\/(.*)\/\1\.js$/;
-
 function loadServiceClasses() {
   // New-style services
-  const services = glob.sync(`${__dirname}/**/*.js`)
-    .filter(path => serviceRegex.test(path))
+  const services = glob.sync(`${__dirname}/**/*.service.js`)
```
chris48s: Should this still be filtered to only files under the `/services` directory? If we end up with a file called `something.service.js` in `/lib` or something I think we probably don't want to pull it in here

chris48s:
@RedSparr0w note that if this gets merged ahead of #1785 and/or #1802 they'll need rebasing to account for this, even though there won't be a merge conflict

or possibly the other way round if one or other of those goes in ahead of this.

services/index.js
paulmelnikow: Yea, true, though this is `services/index.js` so `__dirname` here is `/services`.

chris48s: Oh yes, good point. Looks like that call to `.filter()` was not doing anything useful :)

paulmelnikow: Before it was filtering out the testers, and any random helper modules that might have been stashed in `services/`. 😉

paulmelnikow:
Hey, could I get your 👍's on this?

chris48s:
> Hey, could I get your +1's on this?

The changes to the service layout itself are fine, but there's still an issue with the changes to `dangerfile.js`. If I run `danger` against this PR using the config on `master` it runs fine, but if I checkout this PR I get the same failure as show in https://github.com/badges/shields/pull/1797#issuecomment-406925979 locally.

paulmelnikow:
Should be fixed, thanks!

Issue #1359, opened 2017-12-08
Keep service tests green 💚


paulmelnikow:
We want our service tests to hit the real services. This makes some of them unreliable – out of necessity, not sloppiness. There's a long discussion of the test strategy in #927.

I’d like to use this issue as a place to identify and fix flaky service tests.

IcedFrisby/IcedFrisby#71 will allow us to set a per-test `timeout()` and per-test `retry()`, which should allow us to keep flaky tests green most of the time.

Of course we want the requests to be fast. Right now they fail after 2 seconds, which is a fairly long time to be waiting for a badge. Since readme images are served through GitHub’s proxy (camp) which has its own timeout, perhaps we should choose a threshold in relation to that. IcedFrisby has a separate method [`.expectMaxResponseTime()`](https://github.com/IcedFrisby/IcedFrisby/blob/master/API.md#expectmaxresponsetimems) which we could use to assert the response time. We could also generate a tidy report of slow badges and print it as part of our daily run. An often-red build is not the best way to accomplish this! :smile:

paulmelnikow:
This is a list of failures from a recent test run on master:

- [ ] bower
- [ ] codeclimate: let's track in #1329
- [ ] suggest (This is not even a service! Maybe it broke in #1282?)

```
  1) Bower
       licence

  [ GET http://localhost:1111/bower/l/bootstrap.json ]:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.


  2) Bower
       custom label for licence

  [ GET http://localhost:1111/bower/l/bootstrap.json?label="my licence" ]:

      AssertionError: expected { name: '"my licence"', value: 'MIT' } to deeply equal { name: 'my licence', value: 'MIT' }
      + expected - actual

       {
      -  "name": "\"my licence\""
      +  "name": "my licence"
         "value": "MIT"
       }

      at Object.pathMatch.matchJSON (node_modules/icedfrisby/lib/pathMatch.js:142:38)
      at node_modules/icedfrisby/lib/icedfrisby.js:760:10
      at IcedFrisbyNock._invokeExpects (node_modules/icedfrisby/lib/icedfrisby.js:1272:33)
      at node_modules/icedfrisby/lib/icedfrisby.js:1251:29
      at Request.runCallback [as _callback] (node_modules/icedfrisby/lib/icedfrisby.js:562:11)
      at Request.self.callback (node_modules/request/request.js:186:22)
[...]

PR #1798, opened 2018-07-23
Fix [bundlephobia] tests and rm errant inspectJSON

paulmelnikow:
I’m guessing something has changed with the way rounding happens in bundlephobia. I don’t expect this to happen again, but if it does we can change this to a regex.

For #1359


paulmelnikow:
Thanks for reviewing!

PR #1800, opened 2018-07-23
re-organise badge examples

chris48s:
Following on from https://github.com/badges/shields/pull/1762#issuecomment-406896626 I've extracted the changes to the examples. This re-shuffles the examples into categories:

- Build
- Chat
- Dependencies
- Downloads
- Funding
- Issue Tracking
- License
- Rating
- Social
- Version
- Other

#1762 can be rebased on to this later.

### DIFF
--- a/lib/all-badge-examples.js
+++ b/lib/all-badge-examples.js
@@ -207,16 +207,12 @@ const allBadgeExamples = [
         previewUri: '/jenkins/s/https/jenkins.qa.ubuntu.com/view/Precise/view/All%20Precise/job/precise-desktop-amd64_default.svg'
       },
       {
-        title: 'Jenkins Tests',
+        title: 'Jenkins tests',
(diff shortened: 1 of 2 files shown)


chris48s:
Yes- good suggestion. I wrote this little script to do that:

```js
const allBadgeExamples = require('./lib/all-badge-examples');

var examples = [];
allBadgeExamples.forEach(function(category) {
  examples = examples.concat(category.examples);
});
var previewUrls = examples.map(function(example) {
  return example.previewUri;
});
previewUrls.sort();
previewUrls.forEach(function(title) {
  console.log(title);
});
console.log(previewUrls.length);
```

Running that on master and my branch revealed there were a couple of things that got lost somewhere, so I've fixed those in 0076a6e .

Now I get the same result from running that script on my branch and latest `master`. You should be able to check that if you want.

chris48s:
For convenience, if you do want to test the way I did it was I ran that script, piped the output to a file on each branch and then diff the 2 output files.

paulmelnikow:
Thanks for doing that!

paulmelnikow:
This is live.

PR #1801, opened 2018-07-23
Fix colorB override not working for static badges

### LINKED ISSUE #1799: static badge colorB no longer overrides color in URL
Someone reported in Discord that this static badge used to be gray and is now red.

https://img.shields.io/badge/style-0-red.svg?style=for-the-badge&label=ERRORS&colorB=6A737D

There's a conflict there, where one color is specified in the path and another one in the query string. It looks like, after the change in #1635, `colorB` no longer is honored for static badges. The color from the path is used instead.

This is a really trivial case, though it does seem reasonable to me that `colorB` would take precedence, basically so that `colorB` behaves consistently. What do you think?

RedSparr0w:
Closes #1799 

`colorB` override should now have priority on the static badges

### DIFF
--- a/server.js
+++ b/server.js
@@ -7830,9 +7830,9 @@ function(data, match, end, ask) {
   // Badge creation.
   try {
     var badgeData = getBadgeData(subject, data);
-    if (data.label !== undefined) { badgeData.text[0] = '' + data.label; }
+    badgeData.text[0] = getLabel(subject, data);
     badgeData.text[1] = status;
-    setBadgeColor(badgeData, color);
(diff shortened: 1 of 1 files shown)


paulmelnikow:
http://localhost:8080/badge/style-0-red.svg?colorB=ccc

master:

screen shot 2018-07-23 at 6 32 35 pm

this PR:

screen shot 2018-07-23 at 6 32 39 pm

Looks like the testing is happening along with rewriting in #1802, so let's get this merged (and deployed!) for now.

PR #1802, opened 2018-07-23
move [static-badge] to own service & add test

RedSparr0w:
Move the static badge to its own service module,
Add tests.

I'm unsure how to add the custom headers back such as these though:
```javascript
ask.res.setHeader('Cache-Control', 'max-age=' + cacheDuration);
ask.res.setHeader('Last-Modified', serverStartTime.toGMTString());
```


RedSparr0w:
I can't set it so the label/message accepts being empty,
I've tried the following:
```javascript
    return {
      label: label || '',
      message: message || '',
      color
    };
```
```javascript
    return { label: '', message: '' };
  }
```

And those do not seem to help.

Any ideas?

paulmelnikow:
Phew, thanks for taking this on!

Re the custom headers and the caching:

Because the static badge is one-of-a-kind, I think we should deal with the differences in the static badge's service class rather than building a bunch of extra features into BaseService.

So I'd suggest we give StaticBadge its own implementation of `register()`. That method is responsible for invoking `camp.route`. When you invoke it you can provide your own callback like we do in `server.js` and then you'll have access to the `ask` parameter. Plus it'll allow us to maintain the same cache behavior. The `handleRequest` call in the second argument to `camp.route` is the same as `cache` in `server.js`. You'll notice there's no call to `cache` in for the static badge; another reason to override `register()`.

https://github.com/badges/shields/blob/f5fe85cda2b7ad478996ec7959c94b904dd31d92/services/base.js#L206

This code translates from named params + query params to serviceData:

https://github.com/badges/shields/blob/f5fe85cda2b7ad478996ec7959c94b904dd31d92/services/base.js#L209-L214

Since you don't need `sendAndCacheRequest` you could pass `{}` for `context`.

Re the empty message, I'd suggest we replace this code:

https://github.com/badges/shields/blob/f5fe85cda2b7ad478996ec7959c94b904dd31d92/services/base.js#L187-L190

with a function like this:

```js
function coalesce(...candidates) {
  return candidates.find(c => typeof c === 'string');
}

console.log(coalesce('', 'foo', ''));
console.log(coalesce('asdf', 'foo', ''));
console.log(coalesce(null, '', 'bar'));
console.log(coalesce(undefined, '', 'bar'));
```

I'm pretty sure accepting an empty string anywhere along that chain will do the right thing. We aren't in the habit of returning an empty string as a kind of null.

paulmelnikow:
@RedSparr0w Would you like me to pick up the changes to `register()`? I'm happy to hold off if you'd like to do it. 😁

RedSparr0w:
If you could that would be great! 👍

paulmelnikow:
Mind merging master into this branch when you have a moment? I wrote some steps here: https://github.com/badges/shields/pull/1866#issuecomment-411567524

paulmelnikow:
Reopened as #2284!

PR #1803, opened 2018-07-24
Bump icedfrisby version

paulmelnikow:
This isn't for a specific feature, just to use the latest. There's a feature I wanted to use, `inspectOnFailure`, but it's not working correctly so I'm opening a PR upstream.


paulmelnikow:
Thanks!

PR #1742, opened 2018-06-19
🚢 Deploy!

greysteil:
There are 82 commits on master waiting to be deployed.

If there's anything blocking this going out then I'd love to help :octocat:.

Issue #1723, opened 2018-06-06
Images have no-store, must-revalidate which absolutely kills performance


kornelski:
The responses from img.shields.io have

    cache-control: no-cache, no-store, must-revalidate

which is absolutely the worst case for HTTP requests. It means every client has to make a full request, all the way up to your origin server, with no help of the CDN, every. single. time. it displays the badge.

This seems unreasonably restrictive and clearly your origin server is struggling (I'm seeing responses taking over 10 seconds per badge). Even if you'd like the responses to be super fresh and update as soon as the data changes, allowing cache of even just 1 second would allow the CDN to take some traffic off your back-end servers without affecting freshness of the responses:

    cache-control: max-age=1

In case the `no-store, must-revalidate` is not added by your origin server, but the CDN, consider configuring the CDN to allow some caching (e.g. Page Rules can be used for this in Cloudflare).

chris48s:
Thanks for raising this.

We don't use cloudflare as a downstream cache. It has been discussed it in the `#maintainers` channel but there are some concerns about CloudFlare's use of tracking cookies: https://github.com/badges/shields/issues/608 and the conversation has somewhat fizzled out. Maybe this is a good opportunity to re-open that conversation though.

Aside from that, modifying the `cache-control` headers as suggested should give us benefit when using someone else's CDN (e.g: GitHub's camo proxy - a very common case). This seems like a good idea to me, but looking at the code, it does seem to be a very deliberate decision to avoid being cached by camo:

https://github.com/badges/shields/blob/bf53e612f5bbebd231f75e9899bdafe4a91aa098/lib/request-handler.js#L86-L88

also see https://github.com/badges/shields/issues/221

.. but I am not so clear on why this is such a bad thing. Is anyone more familiar with the history? Why this is something we need to avoid?

RedSparr0w:
I assume it was done to avoid outdated badges,
But at the current rate i feel like the badges should at least default to a ~1 hour cache (with option to set `?maxAge=0` for more important badges).

I've been tracking the badges uptime [here](http://shields.redsparr0w.com/1568/)
<details>
<summary>(note the uptime is based on the badge having a response time < 4 seconds)</summary>

1 second delay:
[![1 second](http://shields.redsparr0w.com/1568/1second.php)](http://shields.redsparr0w.com/1568/1second.php)
2 second delay:
[...]

PR #1806, opened 2018-07-25
Allow env vars to be checked in for the server deploy

paulmelnikow:
What do you think of this approach to handling config on the servers? This is intended as a way to turn on the work we did in #1725 on the VPS without SSH access.

Also ref: #1723 #1568 https://github.com/badges/shields/pull/1742#issuecomment-407446963

Todo:

- [ ] Test on staging
- [ ] Update deploy task
- [ ] Update gitignores?


chris48s:
I don't have strong opinions on this because It impacts on a process that you and @espadrine run, so I'll defer to @espadrine 's review.

PyvesB:
It's been over a week since this pull request was opened. In the meantime, could we not simply hard-code a non 0 value for `BADGE_MAX_AGE_SECONDS`? This would only be temporary and we can simply revert it once this pull request moves forward. 😉

paulmelnikow:
That seems reasonable to me.

Since we're so often having serious issues, maybe we should start with 60 seconds?

Nyholm:
Why not 300? or 600?

I rather see a badge with stale data then no badge at all.

espadrine:
We should federate env vars and secrets vars instead. That way, all env vars that we want to check in can be stored in secrets.json. No need for multiple files with poorly-defined differences, where people won't know which file contains which flag.

paulmelnikow:
This is coming up again because we need to turn on a feature flag for #2069.

Since we don't have _so_ much configuration, I'd tend to go in the opposite direction, and put everything in the environment. Environment vars are the most common way of handling this. For self-hosting users on Heroku or Now, configuring through env vars will be much easier because they can be set persistently from the command line or a UI, and don't require a deploy commit.

Occasionally for complex systems I've used JSON in environment variables, though I don't see any need for it here. We can use one env var per setting.

Though I'm not sure quite what the alternate proposal is, I don't think we should create a solution that requires the servers to talk to each other in order to configure themselves. Immutability isn't just trendy. A system with fewer moving parts is more predictable, and there's less code to write. I made a similar argument in https://github.com/badges/shields/issues/1848#issuecomment-434898047.

paulmelnikow:
Have labeled this a blocker because of #2069.

platan:
It looks good to me. This PR gives us another way to configure server, which is a good idea in my opinion. This looks similar to [externalized configuration in Spring Boot](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html), where you can configure application using several options. 
`dotenv` does not override existing environment variables ([doc](https://www.npmjs.com/package/dotenv#what-happens-to-environment-variables-that-were-already-set)). 
Conflicts must be resolved and currently `6.2.0` is the newest version of `dotenv`. 
@paulmelnikow how exactly do you want to set environment variables on our servers?

paulmelnikow:
I opened #2621 with a proposal to use node-config which would let us check in the non-secret part of the config in the main repo. We can use dotenv with a deploy commit to set `NODE_CONFIG_ENV` in the current / legacy production deploy, and handle the secrets via checked-in `local-shields-io-public.yml` or additional checked-in environment variables.

paulmelnikow:
Closing in favor of #2626.

PR #1808, opened 2018-07-25
Split front end into one page per category

chris48s:
Rising from the ashes of #1762 . I've had another crack at it using `react-router` this time.

I was hoping that pushing more commits to this branch would update #1762 but it seems like GH doesn't do that once the PR has been closed :(

Can someone deploy this to staging for me so I can test whether this works better than my last attempt (do I have access to do this myself?). Probably not worth spending time on review until I've checked it on staging.


paulmelnikow:
> do I have access to do this myself?

Heh, someone just brought up that this process wasn't documented. 😝

If you sign up for a Heroku account, and send me that email address in Discord I'll add you.

chris48s:
Cheers. This appears to be working as planned on staging, so good for review whenever.

pages/index.js
```diff
@@ -1,68 +1,67 @@
...
+      return this.preparedExamples.map(function(category, i) {
+        return (
+          <Link to={'/examples/' + category.category.id} key={i}>
+            <h3 id={category.category.id}>{ category.category.name }</h3>
+          </Link>
+        )
+      });
+    } else if (this.state.query.length === 1) {
+      return (
+        <div>Search term must have 2 or more characters</div>
```
paulmelnikow: Because this PR removes rows from the DOM rather than hiding them – by design of course! – it seems like we could be a little more conservative about when we initiate a search. If users want to browse, they can click the sections instead of searching. It seems like 3 would be enough… git, npm, etc. What do you think?

frontend/components/header.js
```diff
@@ -2,7 +2,9 @@ import React from 'react';
 
 export default () => (
   <section>
-    <img alt="Shields.io" src="static/logo.svg" />
+    <a href="/">
```
paulmelnikow: I think making this `/#/` will avoid the redirect that's occurring after clicking it.

pages/index.js
```diff
@@ -1,68 +1,67 @@
...
 
-const baseUri = process.env.BASE_URL;
-const longCache = envFlag(process.env.LONG_CACHE, false);
-
-export default class IndexPage extends React.Component {
-  state = {
-    query: null,
-    example: null
-  };
+class IndexPage extends ExamplesPage {
```
paulmelnikow: React is based on a one-way flow, where props and state are used to generate the component tree in `render()`. In particular, invoking `render()` should not result in a change to the state.

Inheritance between components is not a common pattern. Functions which generate components (called "higher-order components") are more common, though they probably are not needed here. I think you can accomplish the code and logic by factoring out components and helper functions, and using composition.

Here are some thoughts:

- You could create SharedHeader and SharedFooter components with the bits that are shared between the two pages. That way you can move the code from `getBody()` back into `render()` without duplicating anything.
- The SuggestionAndSearch component has a `queryChanged` prop. Passing your own function into that component is a clearer way for that component to emit its state changes back up the tree. You can pass that prop through SharedHeader if need be. In SuggestionAndSearch might be necessary to recompute `queryChangedDebounced`, perhaps in `componentWillReceiveProps`. Otherwise the notifications will keep being sent to the old function.
- Consider whether it might be possible to move some of the (shared) timeout logic into SuggestionAndSearch.

This is what I'm coming up with by _thinking_ about this problem, not by trying to make the changes… so take 'em with a hefty sprinkle of salt!

pages/index.js
RedSparr0w: I don't think having the minimum set to 2 should be too much of a problem,
The timeout between characters is set to 500ms and the average person types at ~300ms per character.
>[The average person types between 38 and 40 words per minute -- between 190 and 200 characters per minute.](https://www.google.co.nz/search?q=average+cpm+typing+speed)

So most of the time, there should be no difference between having the minimum as 2/3 if the user does type `>=3` characters.

chris48s:
Cheers for comments. I've had another go at refactoring this based on your ideas. I haven't gone with quite the same structure you suggested but I think I have arrived at something which is in a more idiomatic style for ReactJS than what I had previously. This should be ready for another look.

frontend/components/badge-examples.js
```diff
@@ -54,25 +58,32 @@ Badge.propTypes = {
...
-          ))
-        }
-      </tbody>
-    </table>
-  </div>
-);
+const Category = ({ category, examples, baseUri, longCache, onClick }) => {
+  if (examples.filter(example => example.shouldDisplay()).length === 0){
+    return null;
+  }
```
paulmelnikow: In a future pass it might be good to remove the `shouldDisplay` thing. That was kind of a hack to support hiding the rows in CSS, which we're no longer doing. Rendering the filtered examples would be cleaner.

frontend/components/examples-page.js
```diff
@@ -0,0 +1,90 @@
...
+    this.searchQueryChanged = this.searchQueryChanged.bind(this);
+  }
+
+  static propTypes = {
+    match: PropTypes.object.isRequired,
+  }
+
+  searchQueryChanged(query) {
+    this.setState({searchReady: false});
+    window.clearTimeout(this.searchTimeout);
```
paulmelnikow: Could you add a comment here explaining the behavior? If I understand correctly, it's to wait until after the user stops typing, and then run the search, and to show the "searching" message while they're waiting.

frontend/components/examples-page.js
```diff
@@ -0,0 +1,90 @@
...
+    this.searchTimeout = window.setTimeout(function() {
```
paulmelnikow: Is there any reason not to use an arrow function here in place of the `.bind`?

frontend/components/examples-page.js
```diff
@@ -0,0 +1,90 @@
...
+    if (this.state.searchReady) {
+      return (
+        <SearchResults
+          category={this.state.category}
+          query={this.state.query}
+          clickHandler={example => { this.setState({ example }); }} />
+      );
+    } else {
+      return 'searching...';
+    }
```
paulmelnikow: 👍

frontend/components/search-results.js
```diff
@@ -0,0 +1,63 @@
...
+import PropTypes from 'prop-types';
+import { BadgeExamples } from './badge-examples';
+import badgeExampleData from '../../badge-examples.json';
+import { prepareExamples, predicateFromQuery } from '../lib/prepare-examples';
+import { baseUri, longCache } from '../constants';
+
+
+export default class SearchResults extends React.Component {
+
+  constructor(props) {
```
paulmelnikow: The `React.Component` constructor sets `props` for you, so you don't need this constructor override.

frontend/components/search-results.js
```diff
@@ -0,0 +1,63 @@
...
+        onClick={this.props.clickHandler}
+        baseUri={baseUri}
+        longCache={longCache} />
+    );
+  }
+
+  renderCategories() {
+    return this.preparedExamples.map(function(category, i) {
+      return (
+        <Link to={'/examples/' + category.category.id} key={i}>
```
paulmelnikow: I've learned from a linter rule on another project that `key` is not supposed to be an array index, rather something that persistently identifies a piece of content. The purpose of `key` is so React can be smart about adding and removing DOM elements. If some of the child items go away, their keys should stay the same. Could we use the category id instead?

frontend/components/search-results.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  render() {
+    this.preparedExamples = this.prepareExamples(this.props.category);
+
+    if (this.props.category) {
+      return this.renderExamples();
+    } else if ((this.props.query == null) || (this.props.query.length === 0)) {
+      return this.renderCategories();
+    } else if (this.props.query.length === 1) {
```
paulmelnikow: This piece of logic – the message for the empty search term – feels like it belongs to `SuggestionAndSearch` or `ExamplesPage`. Would it be easy to move to one of those?

frontend/components/search-results.js
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      <BadgeExamples
+        categories={this.preparedExamples}
+        onClick={this.props.clickHandler}
+        baseUri={baseUri}
+        longCache={longCache} />
+    );
+  }
+
+  renderCategories() {
```
paulmelnikow: Is this always rendering empty categories? That seems fine, but maybe calling it `renderEmptyCategories` would make that more obvious.

frontend/components/suggestion-and-search.js
```diff
@@ -23,6 +23,7 @@ export default class SuggestionAndSearch extends React.Component {
     inProgress: false,
     projectUrl: null,
     suggestions: [],
+    searchReady: false,
```
paulmelnikow: Is this piece of state used anywhere?

pages/index.js
```diff
@@ -1,68 +1,23 @@
...
-        `}</style>
+        <Route path="/" exact component={ExamplesPage} />
+        <Route path="/examples/:id" component={ExamplesPage} />
       </div>
     );
+    if (typeof window !== 'undefined') {
+      return (<HashRouter>{ router }</HashRouter>);
+    } else {
+      const context = {};
+      return (<StaticRouter context={context}>{ router }</StaticRouter>);
```
paulmelnikow: Is this to support server-side rendering? Could you add a comment?

frontend/components/search-results.js
chris48s: How about `renderCategoryHeadings()`? That seems like quite a good description of what we are doing here.

frontend/components/suggestion-and-search.js
chris48s: It was.. but then I moved it :D

chris48s:
updated based on feedback

frontend/components/search-results.js
paulmelnikow: Works for me.

tooomm:
I've an issue with this PR.
Once you select a category, all others are gone and you can't switch unless you go back to the starting page to change the displayed category.

Is that intentional? @chris48s

chris48s:
> Is that intentional?

It is.. although note that you can still search all badges from the index page. Do you think it is now less obvious how to find stuff, or more cumbersome to browse? I'm genuinely quite interested in how people interact with this and how we can improve the UX, although I'm sure we'll get some feedback when we deploy this :D

Also what do you reckon might be an improvement? Do think it would be better if there were a menu of categories (I'm thinking a hamburger menu)? Or maybe if the categories expanded 'accordion' style?

tooomm:
Yes, imo it's more cumbersome to browse now.
I think many people browse through the site or search for general possibilities etc.
There are not only power users looking for a particular service by name for example.

I think if the other categories are shown on top that would be a good first, helpful step:
mockup
It could be pinned to the top and floating so that it's always visible when scrolling bigger categories for example.

I'm pretty sure there are other valid options as well!

chris48s:
@tooomm - lets continue in #1842

PR #1810, opened 2018-07-25
Add support for simple-icons, colored icons with ?logoColor

RedSparr0w:
Add support for [simple-icons](https://simpleicons.org/).

Support more icons without the overhead of reviewing them ourselves.
I also added support for colored icons.

It will use our icon > simple-icon if available.

### Examples:

params | badge
:-- | :--
`?logo=github&style=social` | 
`?logo=github&logoColor=blue` | 
`?logo=react` | 
`?logo=microsoft%20excel` | 
`?logo=microsoft%20excel&logoColor=blue` | 
`?logo=apple` | 
**NOTE:** | Some badges wont look good with `?logoColor`
`?logo=npm` | 
`?logo=npm&logoColor=blue` | 

_the npm & bitcoin logos (maybe others also) will be completely filled in when you change the logo color due to the background being solid_

### TODO:
- [x] Add `logoColor` information
- [x] Add `simple-icon` information
- [x] Add snapshot tests

image


lib/badge-data.js
```diff
@@ -39,8 +52,12 @@ function isHexColor (s = ''){
 function makeColor(color) {
   if (isHexColor(color)) {
     return '#' + color;
-  } else {
+  } else if (colorschemes[color] !== undefined){
+    return colorschemes[color].colorB;
+  } else if (isCSSColor(color)){
     return color;
+  } else {
+    return undefined;
```
paulmelnikow: Is this change related to simpleicons?

lib/badge-data.js
```diff
@@ -69,9 +86,19 @@ function makeLabel(defaultLabel, overrides) {
...
+  }
+  color = makeColor(color) || 'whitesmoke';
+  const svg = simpleIcons[icon].svg.replace('<svg', `<svg fill="${color}"`);
+  return 'data:image/svg+xml;base64,' + Buffer.from(svg).toString('base64');
+}
+
 function makeLogo(defaultNamedLogo, overrides) {
   if (overrides.logo === undefined){
-    return logos[defaultNamedLogo];
+    return logos[defaultNamedLogo] || getSimpleIcon(defaultNamedLogo, overrides.logoColor);
```
paulmelnikow: For parity, is it possible to support `logoColor` in our own logos?

lib/badge-data.js
RedSparr0w: Kind of, I wasn't sure whether to make a new function for the color or just modify this function as it seemed to still fit quite well, I think it makes more sense to only return a valid color rather than just assume anything that's not hex is a valid color/colorscheme, and the current test were expecting any non valid color to return undefined.

Also there wasn't any test for this function so i've added these too, which are very similar to `setBadgeColor()`.

The main differences between `setBadgeColor()` and `makeColor()` are:
### `setBadgeColor()`:
will set the either `badgeData.colorB` or `badgeData.colorscheme`_(which i think may be redundant now, but will look at that in another PR)_ to the desired color.
### `makeColor()`:
returns a valid color.

lib/badge-data.js
RedSparr0w: Should be able to, Do you know if there is a reason that we load the logos into memory on launch?
_(in `load-logos.js`)_
As that is the main reason i haven't tried yet, because it loads them and keeps them as base64.

RedSparr0w: Seems to be okay :+1:

paulmelnikow: It’s a performance optimization, to load and encode them once instead of per request. We could cache the unencoded logos instead, or keep both in memory, as a small optimization for the cases where logoColor is not specified.

RedSparr0w: Ah, yeah that makes sense, should I also do the same for the `simple-icons` and store them as their own default colors?

paulmelnikow: Sure, that couldn't hurt!

RedSparr0w:
- [x] use the base64 logos from memory (ours & simple-icons),
- [x] Updated descriptions for `?logoColor=` & `?logo=` _(see image in main comment)_
- [x] Simple-icons by default use their own color instead of `whitesmoke`:
```http
gitlab: E24329
react: 61DAFB
etc..
```
- [x] Updated main comment to show some examples.

If i could get another review please. @paulmelnikow 
**Note:** Unsure what the best way to show the logo options was, let me know if you think it should be changed back/adjusted at all.
:smile:

paulmelnikow:
I'd prefer to keep the list in the frontend so it stays in sync with the deployed code. Maybe the simplest thing would be to update `lib/export-supported-features-cli.js` so it includes both the simpleicons and internal icons.

What do you think about my suggestion to add a test that exercises the simple logo feature on a static badge? It could be a snapshot test. I don't want us to break this stuff by accident.

Otherwise this looks good to me!

RedSparr0w:
>I'd prefer to keep the list in the frontend so it stays in sync with the deployed code. Maybe the simplest thing would be to update lib/export-supported-features-cli.js so it includes both the simpleicons and internal icons.

I think it may end up making it look a little too busy if we included the simple-icons in the list as its already quite large just with the icons we have, what about if we just have a link to simple icons at the end of our list?

>What do you think about my suggestion to add a test that exercises the simple logo feature on a static badge? It could be a snapshot test. I don't want us to break this stuff by accident.

No worries, sounds like a good idea to me 👍

paulmelnikow:
> what about if we just have a link to simple icons at the end of our list?

Works for me!

RedSparr0w:
Added snapshot test for:
- shields GitHub logo
- shields GitHub logo w/custom color
- simple-icons javascript logo
- simple-icons javascript logo w/custom color

Should be ready for another review now @paulmelnikow 
🤞

paulmelnikow:
> `?logo=microsoft%20excel`

To make the URLs friendlier, what do you think about converting the spaces to dashes, similar to the way you've lowercased them?

RedSparr0w:
Sounds like a great idea, will definitely look cleaner.

RedSparr0w:
Now supports using a dash or space `?logo=microsoft%20excel` or `?logo=microsoft-excel`

JeanMertz:
Pretty sweet that all those icons are now available 👍. One question; is it correct that not all icon colours can be manipulated for every possible badge? For example, the `Linux` or `Arch Linux` icons don't seem to change color when using a URL like this:

https://img.shields.io/circleci/project/github/rusty-rockets/sm/master.svg?style=flat-square&logo=linux&logoColor=white



While using the static image does work:

https://img.shields.io/badge/style-flat--square-green.svg?style=flat-square&logo=linux&logoColor=white

PR #1815, opened 2018-07-29
Fix [circleci] badge

### LINKED ISSUE #1792: Our test badge incorrectly stays green
Our tests went red on master a couple days ago. There were failures in the main tests, not just the service tests. The badge has shown passing I think that entire time, and still is. I thought the problem was that Circle was showing results from the most recent tests on any branch instead of just master. However if I get the badge for master, I see the same thing.

https://img.shields.io/circleci/project/github/badges/shields/master.svg

https://github.com/badges/shields/commits/master

chris48s:
This changes the badge so it requires all jobs in the build to pass in order for the status to be 'passing'
Closes #1792

### DIFF
--- a/server.js
+++ b/server.js
@@ -5670,14 +5670,14 @@ cache(function(data, match, sendBadge, request) {
       let status;
       for (let i=0; i<data.length; i++) {
         status = data[i].status;
-        if (['success', 'fixed'].indexOf(status) !== -1) {
+        if (['success', 'fixed'].includes(status)) {
           passCount++;
         } else if (status === 'failed') {
           badgeData.colorscheme = 'red';
           badgeData.text[1] = 'failed';
           sendBadge(format, badgeData);
           return;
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -5665,28 +5665,33 @@ cache(function(data, match, sendBadge, request) {
...
+      for (let i=0; i<data.length; i++) {
+        status = data[i].status;
+        if (['success', 'fixed'].indexOf(status) !== -1) {
+          passCount++;
+        } else if (status === 'failed') {
+          badgeData.colorscheme = 'red';
+          badgeData.text[1] = 'failed';
+          sendBadge(format, badgeData);
+          return;
+        } else if (['no_tests', 'scheduled', 'not_run'].indexOf(status) !== -1) {
```
RedSparr0w: What do you think about using [`.includes()`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) here _(and above)_ instead?

Issue #1768, opened 2018-07-08
GitHub issues undefined


PyvesB:
When the count cannot be found for a given repository, our GitHub issues badges display labels similar to `undefined open`. See for instance the Readme of [this repository](https://github.com/PyvesB/asm-game-of-life) which I have renamed recently without changing the Shields badge URL.

We should probably deal with this in a cleaner way, possibly by displaying `not found` instead.

PR #1816, opened 2018-07-30
[GitHub] Fix undefined badges

PyvesB:
This pull request addresses issue #1768.

I noticed that many other GitHub badges were suffering from this same problem, but for now I've only fixed it for the "issues" badge and will wait for some initial feedback before tackling the other ones. 😉

### DIFF
--- a/lib/github-helpers.spec.js
+++ b/lib/github-helpers.spec.js
@@ -1,0 +1,16 @@
+'use strict';
+
+const chai = require('chai');
+const { assert } = chai;
+const { checkErrorResponse } = require('./github-helpers');
+
+chai.use(require('chai-as-promised'));
+
+describe('GitHub Error Handler', function() {
+  it('makes not found badge when 422 is returned', function() {
+    const badgeData = {'text': []};
+    assert.equal(true, checkErrorResponse(badgeData, null, {statusCode: 422}, 'repo not found'));
+    assert.equal('repo not found', badgeData.text[1]);
+    assert.equal('lightgrey', badgeData.colorscheme);
+  });
+});
--- a/lib/github-helpers.spec.js
+++ b/lib/github-helpers.spec.js
@@ -1,16 +1,13 @@
 'use strict';
 
-const chai = require('chai');
-const { assert } = chai;
+const { expect } = require('chai');
 const { checkErrorResponse } = require('./github-helpers');
-
-chai.use(require('chai-as-promised'));
 
 describe('GitHub Error Handler', function() {
   it('makes not found badge when 422 is returned', function() {
     const badgeData = {'text': []};
-    assert.equal(true, checkErrorResponse(badgeData, null, {statusCode: 422}, 'repo not found'));
-    assert.equal('repo not found', badgeData.text[1]);
-    assert.equal('lightgrey', badgeData.colorscheme);
+    expect(checkErrorResponse(badgeData, null, {statusCode: 422}, 'repo not found')).to.be.true;
+    expect(badgeData.text[1]).to.equal('repo not found');
+    expect(badgeData.colorscheme).to.equal('lightgrey');
   });
 });
--- a/services/github/github.tester.js
+++ b/services/github/github.tester.js
@@ -157,6 +157,13 @@ t.create('GitHub open issues by label (raw)')
     value: isMetric
   }));
 
+t.create('GitHub open issues repository not found')
+  .get('/issues-raw/PyvesB/ASMGameOfLife.json')
+  .expectJSON({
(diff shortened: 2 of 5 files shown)


RedSparr0w:
this is a rather strange way for GitHub to report when a repo is not found:
```json
{
  "message": "Validation Failed",
  "errors": [
    {
      "message": "The listed users and repositories cannot be searched either because the resources do not exist or you do not have permission to view them.",
      "resource": "Search",
      "field": "q",
      "code": "invalid"
    }
  ],
  "documentation_url": "https://developer.github.com/v3/search/"
}
```
Do you think it would be worth checking the `message` field matches and isn't trying to report a different error?

PyvesB:
I get your point and I did think about that alternative as well, but I felt slightly less enthusiastic about it:
* to avoid parsing the response body several times, we would probably have to parse it before we check the response codes and pass it as a parameter to `githubCheckErrorResponse`, which seems to me as the wrong ordering of operations. Also, would this lead us to more trouble when we migrate the GitHub services to the new refactored service structure?
* It seems quite clumsy, we would have to iterate through all the elements inside the `errors` array and look for a message that is more likely to change than a status code.

Does this make sense?

RedSparr0w:
No worries, I think it should be fine as is 👍

PyvesB:
I'll have a go at fixing the other badges in the coming days, we can think about it some more in the meantime and see how it looks once everything is done. 😉

services/github/github.tester.js
```diff
@@ -74,7 +74,7 @@ t.create('License - API rate limit exceeded')
       message: "API rate limit exceeded for 123.123.123.123. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
       documentation_url: "https://developer.github.com/v3/#rate-limiting"
   }))
-  .expectJSON({ name: 'license', value: 'inaccessible', colorB: colorsB.lightgrey });
+  .expectJSON({ name: 'license', value: 'invalid', colorB: colorsB.lightgrey });
```
PyvesB: This change is due to the default _error-helper.js_ handling. Wouldn't `inaccessible` be more appropriate for 403 responses everywhere?

RedSparr0w: It could be worth updating the error-helper to something similar to this:
```javascript
defaultErrorMessages = {
  404: 'not found',
  418: "I'm a teapot",
};

const checkErrorResponse = function(badgeData, err, res, errorMessage = 'not found') {
  if (typeof errorMessage === 'string')
    errorMessage = {...defaultErrorMessages, ...{404: errorMessage}}
  else
    errorMessage = {...defaultErrorMessages, ...errorMessage};
  if (err != null) {
    badgeData.text[1] = 'inaccessible';
    badgeData.colorscheme = 'red';
    return true;
  } else if (errorMessage[res.statusCode] !== undefined) {
    badgeData.text[1] = errorMessage[res.statusCode];
    badgeData.colorscheme = 'lightgrey';
    return true;
  } else if (res.statusCode !== 200) {
    badgeData.text[1] = 'invalid';
    badgeData.colorscheme = 'lightgrey';
    return true;
  } else {
    return false;
  }
};

//Usage:
checkErrorResponse(badgeData, err, res, {403: 'repo not found'}); // 403: repo not found
checkErrorResponse(badgeData, err, res, 'repo not found'); // 404: repo not found
checkErrorResponse(badgeData, err, res, {403: 'repo not found', 404: 'repo not found'}); // 403, 404: repo not found
```

That way if services return a strange error code, we can still set a relevant error message.

paulmelnikow: **inaccessible** would be a nice way to convey a permission problem. For our badges, thoguh, it usually means the service can't be reached.

Maybe **not authorized**?

PyvesB: Yes, "not authorized" may be good alternatives for 403 cases. "access denied" is a good alternative as well. I'm leaning towards a separate PR though, as this has not got much to do with "undefined" in GitHub badges.

paulmelnikow:
Is this waiting for another review? Let me know if you need me to take another look.

PyvesB:
Yes, I wouldn't mind someone taking a look at the final result! 👍

lib/github-helpers.spec.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const chai = require('chai');
+const { assert } = chai;
```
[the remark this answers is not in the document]
PyvesB: Done, we should probably revisit _error-helper.spec.js_ as weel at some point, and possibly some other specs. 😉

PR #1841, opened 2018-08-02
Support for custom names on maven-metadata badges

### LINKED ISSUE #1839: Gradle plugin badges can be served by maven-metadata but needs to have configurable naming
The gradle plugin repository can be referenced by maven-metadata:

https://img.shields.io/maven-metadata/v/https/plugins.gradle.org/m2/com/google/cloud/tools/jib/com.google.cloud.tools.jib.gradle.plugin/maven-metadata.xml.svg

However, this looks like this some text 

But it's really the gradle plugin version. Maybe a way to customize the text here would make it useful to gradle plugin developers.

loosebazooka:
- This adds support for the gradle plugin repository by leveraging
maven-metadata.xml in plugins.gradle.org/m2
- users can use custom names by specifying them before the
"maven-metadata" type. For example "gradle-maven-metadata" will be
handled by the "maven-metadata" handle and return a result with
"name = gradle"
- Also fixes link to testing page in tutorial

Fixes #1839, but I'm not a node developer, and I'm not sure what exactly the cost of this regular expression will be on the server. Would it better to limit this naming to just `gradle` instead of matching anything?


loosebazooka:
This appears to be unnecessary given the update on #1839

PR #1842, opened 2018-08-03
RFC add a menu to category pages

chris48s:
Based on @tooomm 's https://github.com/badges/shields/pull/1808#issuecomment-409732290 on the new front-end design, I have put together a rough implementation based on that screenshot. This isn't intended as ready-to-merge, but it gives me a way to deploy something to staging so we can test something and move the conversation on.

A horizontal menu is probably not great for mobile users, so I've hidden it at low-res for now. In general I'd like to move towards a more mobile-friendly front-end although I'm aware that right now we are rendering most of our content in tables.

Something like this is probably OK as long as we have a relatively small number of categories (we've just increased to 11), but as discussed in https://github.com/badges/shields/pull/1762#issuecomment-403583651 I can see a case for moving to a situation where we might have a much larger number of menu items. I wonder if anyone has any ideas about how we could present this, particularly if we had many more 'categories' or 'headings' (more than you'd want on a single row).

Feedback generally welcome. Also probably worth keeping en eye on the issues as more users see the new layout.

### DIFF
--- a/frontend/components/examples-page.js
+++ b/frontend/components/examples-page.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import Menu from './menu';
 import Meta from './meta';
 import Header from './header';
 import SuggestionAndSearch from './suggestion-and-search';
@@ -22,6 +23,10 @@ export default class ExamplesPage extends React.Component {
     this.searchTimeout = 0;
     this.renderSearchResults = this.renderSearchResults.bind(this);
     this.searchQueryChanged = this.searchQueryChanged.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({category: nextProps.match.params.id});
   }
 
   static propTypes = {
@@ -86,6 +91,7 @@ export default class ExamplesPage extends React.Component {
             href="https://opencollective.com/shields">
             donate
           </a>
+          { this.state.category && <Menu /> }
         </section>
         { this.renderSearchResults() }
         <Usage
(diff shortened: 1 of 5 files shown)


chris48s:
staging deploy at https://shields-staging-pr-1842.herokuapp.com

tooomm:
First observations:

- The new "menu headline" is not centered
- When you switch between categories the old batches stay visible until they get replaced with the new ones. they should be hidden/removed with the category switch.
This can be awkward because sometimes it takes a few seconds until they properly refresh

Both can be seen on this screenshot:
loading

paulmelnikow:
Mind merging master into this branch when you have a moment? I wrote some steps here: https://github.com/badges/shields/pull/1866#issuecomment-411567524

platan:
Good idea. Now I can easily move between categories. Before this I have to go back in browser to see all categories. 
My note. After page is loaded we have a vertical list of categories. After choosing one category vertical list changes to horizontal. I would like to have it consistent - we can display horizontal list form the beginning.

chris48s:
Given @mbrandau wants to work on menu as part of some wider (much needed) redesign work in #1899 I'm going to put this on hold for the moment as I think it fits there more naturally. If #1899 stalls, I'll revisit this.

tooomm:
Is it worth revisiting and giving this the last finishing (as a interim solution at least)? @chris48s 
Currently, it looks like the other approach hasn't been worked on since quite some time.

RedSparr0w:
With more categories being added such as `Activity` #2448, It may be worth pushing ahead with this.

paulmelnikow:
@chris48s Would you like me to pick this up after #2441 is merged? Or would you like to tackle it?

chris48s:
I'm not particularly attached to it. I'm aiming to try and have a look at #2441 tomorrow. If you want to have a go at this once that's in, feel free. Cheers

paulmelnikow:
Reopened as #2682.

Issue #1826, opened 2018-08-01
Disable danger for dependabot PRs


chris48s:
Danger messages are intended for humans. Bots talking to bots is not so helpful :) All the dependabot PRs are on a branch called `dependabot/something` so this should be reasonably easy to pick up in CI.

Refs #1822

paulmelnikow:
No danger message in #1860 but seeing one in #1861, so it looks like the fix worked!

paulmelnikow:
Yea, good callout. Let’s handle it in #1827.

PR #1843, opened 2018-08-03
don't run danger on dependabot PRs

chris48s:
refs #1826

This is hard to test, but should do what we want.


paulmelnikow:
Thanks for addressing this!

PR #1846, opened 2018-08-04
Increased default BADGE_MAX_AGE_SECONDS

PyvesB:
This is related to discussions in #1806. Until that pull request is merged and we have an easy way of configuring the environment variables on the server, we should have a non-0 value in the meantime.

@paulmelnikow suggested 60 seconds whereas @Nyholm was advocating larger values (300, 600), so I've opted for something intermediary, 120 seconds. We can easily change it if need-be and revert it once the other PR is merged.

### DIFF
--- a/frontend/components/usage.js
+++ b/frontend/components/usage.js
@@ -215,7 +215,7 @@ export default class Usage extends React.PureComponent {
               <td>
                 <code>?maxAge=3600</code>
               </td>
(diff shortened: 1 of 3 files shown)


paulmelnikow:
Under the circumstances, I'm fine with this. 😝 

Should we add a note to the website where we talk about `maxAge`, to indicate that stale responses might be cached for a couple minutes?

PyvesB:
@paulmelnikow I have included the current default inside the `maxAge` explanation on the website.

PR #1859, opened 2018-08-07
Show test status on master, not latest build

paulmelnikow:
This is a rather confusing behavior of CircleCI's API and our badge: it shows the status of the latest build, not the latest build on the default branch.

That seems like it's almost never what anyone would want. I wonder if we should require a branch going forward, and deprecate the option to show the latest build.


paulmelnikow:
> Nevertheless, I think we should leave the option to show the latest build, it may come in handy for some users/use cases.

I'm fine with that of course, though maybe we could require that the user be explicit about wanting this. It's not the nicest default IMO.

PR #1205, opened 2017-10-25
Rewrite and test Github auth logic, separating standard and search quota

paulmelnikow:
Summary:

- Create new priority queue-based token cycling logic which correctly handles simultaneous requests using the same token
- Correctly support search API rate limits, which are separate from core API rate limits (#1119)
- When tokens are exhausted, return an error instead of attempting fallback to our API key
- Use a static token in development, when one is configured (replacing #1118)
- Split up Github auth logic into separate modules which can be tested in isolation

Todo:

- [x] Test persistence
- [x] Integration test?
- [x] Incorporate recent changes to `lib/github-auth.js`
- [x] Catch persistence errors and send them to sentry
- [x] Rework TokenProvider / TokenPool abstraction which may not be so helpful. Github has two pools so things feel a little bit off.
- [x] Update token persistence for these changes to the api provider
- [x] Wait for merge of #2400

To handle in follow on work:

- [ ] Log / emit to sentry when tokens are exhausted
- [x] (maybe) Consolidate reserve logic, probably in `Token`

PR #1861, opened 2018-08-08
Refactor [github] initialization

paulmelnikow:
This creates a new convenience class which consolidates all the Github initialization. It supports dependency injection and facilitates refactoring the persistence along the lines of #1205.

### DIFF
--- a/lib/github-auth.js
+++ b/lib/github-auth.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const path = require('path');
 const crypto = require('crypto');
 const log = require('./log');
 const queryString = require('query-string');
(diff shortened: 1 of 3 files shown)


paulmelnikow:
> Should we add a few tests to _github-constellation.js_, to check what's going on in the `initialize` and `stop` functions?

I’m not sure it makes sense to add any. All they could really do is mock all the method calls, and if we do that we still don’t know the code actually works. IMO it’s better to test at a lower level, with specs for persistence, the token logic, and maybe the acceptor routes, and at a higher level, to depend on the service tests making sure we can hit github.

Issue #1848, opened 2018-08-05
Make shields runnable on a PaaS


paulmelnikow:
In https://github.com/badges/shields/pull/1742#issuecomment-407446963 I’ve proposed that our production hosting move to Zeit Now. Whether or not this happens, I’d like to make sure Shields is fully capable of running in a PaaS. We’re most of the way there already. Shields doesn't have a lot of state. Since last year we’ve run our staging server on Heroku, and @platan recently ran a copy of Shields on Zeit for #1708.

We have a few bits in Shields which expect a persistent file system. Our production server relies on deploy commits. Deploy commits are possible with a PaaS, though they violate one of the [12-factor principles](https://12factor.net/config), and are more difficult to manage. Now supports [secure, shared environment variables](https://zeit.co/docs/features/env-and-secrets#securing-env-variables-using-secrets) and I'd like to make it possible to use them.

There are three items that come to mind:

- [ ] ~~Analytics (already has redis support, but it needs tests because it's not used in production)~~
- [x] GitHub token persistence (work in #1205)
    - [x] Admin route #1813 #1886
    - [x] GithubApiProvider #1812
    - [x] Initialization code #1861 
    - [x] Refactor JSON persistence #1863 
    - [x] Extend to cloud DB or cloud storage #1906 #1939
- [x] Option to load server secrets from the environment #2626

I’d like to set up each of these in a way that doesn’t interfere with the current usage. That way things are in good shape, however we decide to proceed.

paulmelnikow:
[Compose](https://www.compose.com/pricing#redis), which I've used before, has Redis with persistence starting at $18.50/month. [RedisLabs](https://redislabs.com/pricing/cloud/) has it for $7/month, or if we want it highly available, $9/month. We have the budget for this.

Alternatively, given the nature of this data which could be infrequently read and written, we could use something like S3 or Google Cloud Storage, which would be much cheaper. We only need to write tokens when they change, and we could write stats once every 30 or 60 seconds.

paulmelnikow:
I discussed a bit with @chris48s offline about the possibility of basing some of this using a storage abstraction layer which self-hosters and the production servers could choose. The most widely adopted of these I know of for Node is [abstract-blob-store](https://github.com/maxogden/abstract-blob-store) which is not really an abstraction layer but an interface. Still, I think it could be a good choice for certain kinds of caching.

[...]

PR #1863, opened 2018-08-08
Refactor and test [github] token persistence

paulmelnikow:
After #1861 is merged I will get this into working shape.

Ref #1848


lib/text-formatters.spec.js
```diff
@@ -83,8 +83,12 @@ describe('Text formatters', function() {
   })
 
   context('in october', function() {
+    let clock
     beforeEach(function() {
-      sinon.useFakeTimers(new Date(2017, 9, 15).getTime())
+      clock = sinon.useFakeTimers(new Date(2017, 9, 15).getTime())
+    })
+    afterEach(function() {
+      clock.restore()
```
paulmelnikow: This was causing a timeout in the new tests.

lib/token-persistence.js
```diff
@@ -0,0 +1,43 @@
...
+
+// This is currently bound to the legacy github auth code. That will be
+// replaced with a dependency-injected token provider.
+class TokenPersistence {
+  constructor({ path }) {
+    this.path = path
+    this.save = undefined
+  }
+
+  async initialize() {
```
paulmelnikow: This code is a bit difficult to understand, in part because it's working against the interface of `json-autosave` which wants to own the data structure.

paulmelnikow:
This is ready for review!

lib/token-persistence.js
PyvesB: Not trivial indeed, but I think I get the gist of it. 👍

lib/token-persistence.spec.js
```diff
@@ -0,0 +1,93 @@
...
+  afterEach(function() {
+    clock.restore()
+  })
+
+  beforeEach(githubAuth.removeAllTokens)
+  afterEach(githubAuth.removeAllTokens)
+
+  let path, persistence
+  beforeEach(function() {
+    path = tmp.tmpNameSync()
```
PyvesB: I'm guessing tmp will automatically clean up on exit?

PyvesB: It does say the following in the [options](https://www.npmjs.com/package/tmp#options) section: "keep: signals that the temporary file or directory should not be deleted on exit, default is false". I'm generally not a fan of tests leaving droppings behind them, but we're probably fine here according to the documentation! 👍

paulmelnikow:
Before I merge this I want to be ready to deploy it. There's something wrong with the admin endpoint; so I can't back up the tokens. I'm going to fix that first.

paulmelnikow:
Okay, #1886 deployed and tokens backed up. Going to merge and deploy this now!

PR #1883, opened 2018-08-10
Validate response json in [apm appveyor cdnjs clojars gem] also affects [npm]

chris48s:
This fleshes out the stub validators added in #1743 into something a bit more robust. I've not looked at splitting into `fetch()` and `render()` here - just the validation. I'll save that for another PR.

I did find the first few quite took quite a long time, but once I'd done a few I got the hang of it. I think we can make this easier for new contributors with good documentation and examples, but it will need some explanation.

99fc836 and cf04796 aren't strictly related, but I spotted the examples were broken while I was working on that class and its a quick fix.


services/gem/gem-downloads.service.js
```diff
@@ -0,0 +1,149 @@
...
+  Joi.array()
+    .items(
+      Joi.object({
+        prerelease: Joi.boolean().required(),
+        number: Joi.string().required(),
+        downloads_count: positiveInteger,
+      })
+    )
+    .min(1)
+    .required()
```
chris48s: I hit an edge case here. Depending on which downloads badge we are building (`dt`/`dtv` or `dv`) we request a different endpoint on `rubygems.org`. This means there is more than one possible response schema. We can cover it by using `Joi.alternatives().try()` like I've done here, but I wonder if it could be better to think in terms of a URL pattern/schema pair here instead of a one-to-one badge/schema relationship? Any thoughts? Am I just over-complicating things? Are there other situations where we might hit a more complicated version of this?

paulmelnikow: Yea, the npm downloads badge had the same issue. There I handled it with a switch: https://github.com/badges/shields/blob/2ebb173a62bb6f291104920ed4ed8a9a38934217/services/npm/npm-downloads.service.js#L46

services/validators.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
+const Joi = require('joi')
+
+const positiveInteger = Joi.number()
```
paulmelnikow: `nonnegativeInteger`?

services/cdnjs/cdnjs.service.js
```diff
@@ -6,24 +6,27 @@ const { NotFound } = require('../errors')
 const { addv: versionText } = require('../../lib/text-formatters')
 const { version: versionColor } = require('../../lib/color-formatters')
 
+const cdnjsSchema = Joi.object({
+  version: Joi.string(),
```
paulmelnikow: I think you want `version: Joi.string().required()`.

services/gem/gem-owner.service.js
```diff
@@ -0,0 +1,49 @@
...
+
+const { BaseJsonService } = require('../base')
+const { floorCount: floorCountColor } = require('../../lib/color-formatters')
+
+module.exports = class GemOwner extends BaseJsonService {
+  async handle({ user }) {
+    const url = `https://rubygems.org/api/v1/owners/${user}/gems.json`
+    const json = await this._requestJson({
+      url,
+      schema: Joi.array(),
```
paulmelnikow: Looks like this one is pending ^^

services/apm/apm.service.js
```diff
@@ -32,6 +32,15 @@ class BaseAPMService extends BaseJsonService {
...
   }
+
+  static get examples() {
+    return [
+      {
+        previewUrl: 'vim-mode',
+        keywords: ['atom'],
+      },
+    ]
+  }
```
paulmelnikow: ❤️

services/gem/gem-downloads.service.js
chris48s: Yes - that's a much better way of doing it

services/cdnjs/cdnjs.service.js
chris48s: On both CDNJS and Clojars the version key is intentionally optional because `{}` should also pass validation.

services/gem/gem-owner.service.js
chris48s: I intentionally left this because in this badge we are not using any of the data in the array - we are just counting the number of elements in it. As long the response is an array we're fine. I've moved the declaration out of the class and made it `required()` in 410b329 though. I don't think we need to be any more specific

chris48s:
> It's hard to see if anything else has changed.

All the moving stuff around is in 8121a90 to try and make it easier to separate that from the commits that make functional changes

services/cdnjs/cdnjs.service.js
paulmelnikow: Could you add a comment in the schema explaining why?

services/gem/gem-rank.service.js
```diff
@@ -0,0 +1,91 @@
...
+const Joi = require('joi')
+
+const { BaseJsonService } = require('../base')
+const { floorCount: floorCountColor } = require('../../lib/color-formatters')
+const { ordinalNumber } = require('../../lib/text-formatters')
+const { nonNegativeInteger } = require('../validators.js')
+
+const rankSchema = Joi.array()
+  .items(
+    Joi.alternatives().try(
```
paulmelnikow: Do you want to do the same here?

PR #1885, opened 2018-08-10
Lengthen timeout on a flaky service test

paulmelnikow:
These two time out quite often

e.g. https://circleci.com/gh/badges/shields/9050


PyvesB:
This doesn't seem to have done the trick: I got the same timeout again : https://circleci.com/gh/badges/shields/9536

paulmelnikow:
Wow, it's taking longer than 8 seconds. That's absurdly long. We could increase it further, though I also wonder if anyone is using those badges…

PR #1886, opened 2018-08-10
Fix and test the github admin route

paulmelnikow:
Fix regressions in the github admin and token acceptor endpoints, introduced in #1813.

### DIFF
--- a/services/github/auth/admin.spec.js
+++ b/services/github/auth/admin.spec.js
@@ -24,6 +24,8 @@ describe('GitHub admin route', function() {
   let sandbox
   beforeEach(function() {
     sandbox = sinon.createSandbox()
+    // Make this work when there is no `shieldsSecret` defined.
+    serverSecrets.shieldsSecret = undefined
     sandbox
       .stub(serverSecrets, 'shieldsSecret')
       .value(validCredentials.password)
--- a/services/github/auth/admin.spec.js
+++ b/services/github/auth/admin.spec.js
@@ -1,0 +1,74 @@
+'use strict'
+
+const { expect } = require('chai')
+const sinon = require('sinon')
+const Camp = require('camp')
+const fetch = require('node-fetch')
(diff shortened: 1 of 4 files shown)


services/github/auth/admin.spec.js
```diff
@@ -0,0 +1,74 @@
...
+
+const { expect } = require('chai')
+const sinon = require('sinon')
+const Camp = require('camp')
+const fetch = require('node-fetch')
+const config = require('../../../lib/test-config')
+const serverSecrets = require('../../../lib/server-secrets')
+const { setRoutes } = require('./admin')
+
+function createAuthHeader({ username, password }) {
```
paulmelnikow: It's a bit annoying this must be done manually. We may want to consider a different request library like got or axios.

paulmelnikow:
Sentry issue: [SHIELDS-1B](https://sentry.io/shields/shields/issues/633736321/)

PyvesB:
@paulmelnikow I'll look at this in the morning if no one has done so in the meantime. 😉

PyvesB:
What was the regression that was introduced? Was it the usage of `crypto.timingSafeEqual`? The Sentry logs would probably answser my question, but I don't think I have access to them! 😄

paulmelnikow:
> What was the regression that was introduced? Was it the usage of `crypto.timingSafeEqual`?

Yes it was. That function accepts buffers, not strings, and does not accept `undefined`. There were actually two implementations of `constEq` in the codebase. Probably we could swap out the second for `crypto.timingSafeEqual` if we wrap some tests around `isValidToken`.

> The Sentry logs would probably answser my question, but I don't think I have access to them! 😄 

Indeed! It would be good to give access to more of the maintainers. I wonder if we were to write some guidelines, @espadrine would be open to adding more people.

paulmelnikow:
Thanks for reviewing!

PR #1888, opened 2018-08-10
[GitHub] Error message customisation for all status codes

PyvesB:
This pull request arises from [this discussion](https://github.com/badges/shields/pull/1816#discussion_r207731889) and is inspired by @RedSparr0w's suggestion in those comments.

The idea is to allow callers of the `checkErrorResponse` to provide a custom error message for all status codes, not just the 404 one.

I also converted the remaining `assert.equals` usages to `expect` in _error-helper.spec.js_.


paulmelnikow:
Hey, thanks for taking this on. I'll wait to leave comments 'till you're ready. There is a small conflict between this and #1890, so whichever goes in first will need a little bit of updating.

PyvesB:
Okay, I think I'm mostly done here, I'm happy to hear some feedback and thoughts at this point! 👍

lib/error-helper.spec.js
```diff
@@ -1,52 +1,49 @@
...
   it('makes inaccessible badge', function() {
     const badgeData = { text: [] }
-    assert.equal(
-      true,
-      checkErrorResponse(badgeData, 'something other than null', {})
-    )
-    assert.equal('inaccessible', badgeData.text[1])
-    assert.equal('red', badgeData.colorscheme)
+    expect(checkErrorResponse(badgeData, 'something other than null', {})).to.be
+      .true
```
RedSparr0w: Is this on a new line because of prettier/eslint?
I feel like it would look better on 1 line, but i'm fine with it either way.

lib/error-helper.spec.js
```diff
@@ -1,52 +1,49 @@
...
+    expect(badgeData.colorscheme).to.equal('lightgrey')
   })
 
   it('return false on 200 status', function() {
-    assert.equal(
-      false,
-      checkErrorResponse({ text: [] }, null, { statusCode: 200 })
-    )
+    expect(checkErrorResponse({ text: [] }, null, { statusCode: 200 })).to.be
+      .false
```
RedSparr0w: Is this on a new line because of prettier/eslint?
I feel like it would look better on 1 line, but i'm fine with it either way.

lib/error-helper.js
```diff
@@ -1,19 +1,18 @@
 'use strict'
 
 const { NotFound, InvalidResponse } = require('../services/errors')
+const defaultErrorMessages = {
+  404: 'not found',
```
RedSparr0w: Do you think it would be worth while adding a few more defaults?
```http
401: Unauthorized
403: Forbidden
408: Request Timeout
429: Too Many Requests
500: Internal Server Error
```
_Maybe some others too, not really sure which would be the more common ones we would possibly encounter_

lib/error-helper.spec.js
PyvesB: Yes, I believe this is due to Prettier.

lib/error-helper.js
RedSparr0w: I don't believe there is, but it would likely [fail due to timeouts](https://circleci.com/gh/badges/shields/9449).
Good point, i'm fine with just having 404 default for now, until we see a need for more.

PyvesB:
I have rebased as #1890 was causing conflicts. Would someone like to have one last look before I merge?

PyvesB:
It occurred to me that the new service base hadn't been updated, this should now be fixed.

lib/github-helpers.spec.js
```diff
@@ -4,12 +4,34 @@ const { expect } = require('chai')
...
+    expect(checkErrorResponse(badgeData, null, { statusCode: 422 })).to.be.true
+    expect(badgeData.text[1]).to.equal('repo not found')
+    expect(badgeData.colorscheme).to.equal('lightgrey')
+  })
+
+  it('makes user not found badge when 404 is returned', function() {
     const badgeData = { text: [] }
     expect(
-      checkErrorResponse(badgeData, null, { statusCode: 422 }, 'repo not found')
+      checkErrorResponse(badgeData, null, { statusCode: 422 }, 'user not found')
```
RedSparr0w: Should this statusCode be 404?

PyvesB: Yes indeed!

PR #1889, opened 2018-08-10
Split BaseService and BaseJsonService into separate modules; tested on [clojars gem npm appveyor cdnjs]

paulmelnikow:
There’s a lot of behavior here, and going to be even more, so I think it makes sense to split these up as I’ve done with the tests.

### DIFF
--- a/services/apm/apm.service.js
+++ b/services/apm/apm.service.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const Joi = require('joi')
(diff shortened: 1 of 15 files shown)


paulmelnikow:
Some PRs may need to be updated after this goes in, e.g. #1891.

RedSparr0w:
Definitely agree this is a big enough service that it warrants its own file.
👍

PR #1891, opened 2018-08-10
Rewrite [uptimerobot]

paulmelnikow:
This is based on #1890 and the diff will clean up once that goes in.


paulmelnikow:
Sounds good. Want to give that a second :+1: and then merge it?

services/uptimerobot/uptimerobot-base.js
```diff
@@ -0,0 +1,97 @@
...
+    monitors: Joi.array()
+      .length(1)
+      .items(monitorWithUptime)
+      .required(),
+  }).required()
+)
+
+module.exports = class UptimeRobotBase extends BaseJsonService {
+  static get category() {
+    return 'other'
```
chris48s: Can we just return 'monitoring' here and remove the `category()` functions from the child classes?

paulmelnikow: Ha! Yes. It's funny, I went looking for the category and couldn't find it. I was looking in the wrong place :)

PR #1894, opened 2018-08-10
Add some error messages for the developer when .service.js is malformed

paulmelnikow:
When loading `.service.js` files which don’t contain services, such as when the export is forgotten, print helpful error messages. This will only occur during development; the unit tests will catch these problems well before code reaches the server.


services/index.js
```diff
@@ -1,20 +1,31 @@
...
-    .map(path => require(path))
+  const servicePaths = glob.sync(`${__dirname}/**/*.service.js`)
 
   const serviceClasses = []
-  services.forEach(service => {
-    if (typeof service === 'function') {
-      serviceClasses.push(service)
+  servicePaths.forEach(path => {
+    const module = require(path)
+    if (!module) {
```
chris48s: I don't think this is working as you expect. A module that exports nothing gives `module` the value `{}` here but in that situation this condition is true because `{}` is truthy in javascript. Also if we explicitly `module.exports = []`, `[]` is also truthy, but we would want to follow this path in that situation too.

services/index.js
```diff
@@ -1,20 +1,31 @@
...
+      throw Error(
+        `Expected ${path} to export a service or a collection of services; got ${module}`
+      )
+    } else if (module.prototype instanceof BaseService) {
+      serviceClasses.push(module)
     } else {
-      for (const serviceClass in service) {
-        serviceClasses.push(service[serviceClass])
+      for (const key in module) {
```
chris48s: This assumes that if we've got something that isn't `instanceof BaseService` we've got an iterable, but if I declare a service file that starts

```js
module.exports = class MyService extends SomethingThatDoesntExtendBaseService...
```

we'll get to here and try to iterate `MyService`, which doesn't make much sense.

services/index.js
paulmelnikow: Ahh, yes, good point.

services/index.js
paulmelnikow: That's true, though is not extending BaseService a use pattern we need to support?

paulmelnikow:
This doesn't feel like a high priority for an automated test, though I see where you're coming from. If it seems important to you, I'll do it. Or if you feel like tackling the tests, go for it!

services/index.js
chris48s: No - we don't want to _support_ it. We want to throw an error. That's the point of testing against `(serviceClass.prototype instanceof BaseService)` - right?

but because `MyService` isn't iterable we never enter the loop so nothing ever gets tested against the condition `(serviceClass.prototype instanceof BaseService)` in this case.

chris48s: You'd still want to continue checking as you go, or also throw if

`num valid services != num total services`

to also cover the condition where >=1 services extend `BaseService` and >=1 don't, but yeah.. if we get to the end of that loop and have 0 valid services that's definitely an error.

paulmelnikow: We're right now supporting both arrays and objects, and if we keep doing that, it seems simpler to bail if we ever find a non-service, and check that each file has contained at least one before moving on to the next one.

Alternatively we could drop support for objects and require a single service or an array. Then we could use `count`, validate in `map`, and probably simplify this.

chris48s: In the relatively small subset of services we've refactored so far, we've already found valid uses for multiple export styles:

object:
https://github.com/badges/shields/blob/67193ce09b0c2bc69bb4f0175e6148e942a51e99/services/apm/apm.service.js#L122-L126

array:
https://github.com/badges/shields/blob/67193ce09b0c2bc69bb4f0175e6148e942a51e99/services/npm/npm-downloads.service.js#L94

class:
https://github.com/badges/shields/blob/67193ce09b0c2bc69bb4f0175e6148e942a51e99/services/clojars/clojars.service.js#L13

All of those are sensible module export styles and I think we should continue to support all of them so we keep things relatively flexible for future contributors and allow ourselves and contributors to write idiomatic code. Even if that complicates the service loading code a bit, I think that's worthwhile.

We can detect a module which exports a single class from one exporting an Array or Object using `instanceof Function`:

```
> class FooBar {
... }

> console.log(FooBar instanceof Function);
true

> console.log({} instanceof Function);
false

> console.log([] instanceof Function);
false
```

and special-case it.

paulmelnikow:
@chris48s What do you think about merging this? The existing files should prevent regression on the existing use cases, and we could manually test if this code ever gets touched again. It's kind of a pain to test and I feel like the time could be better spent elsewhere.

(This is having just spent a couple hours writing tests for #2021 because I want to get that merged…)

chris48s:
I don't mind merging it without tests if its a complete pain, but either way we should fix those two cases noted in the review where we're not throwing on malformed exports.

That said, having had another look I do have an idea for how to write tests for this. Let me have a quick go at it.. I'm happy to pick it up and push some commits to this branch (if I can write to it) or open a new PR to take it over the line..

paulmelnikow:
Awesome! You should be able to write to the branch.

chris48s:
The new commit I've pushed covers every case I can think of. Does that seem reasonable?

paulmelnikow:
Looks awesome! I can’t :+1: this since it’s my PR. Would you like to merge?

Issue #1845, opened 2018-08-04
Badge to show dependents of a npm package


sindresorhus:
Would be nice to have a badge to show off how many npm packages depend on my package. For example, the `1717 Dependents` number at https://www.npmjs.com/package/got.

You could either:
- Use the API. Example: https://github.com/chrisdickinson/npm-get-dependents
- Scrape. Example: https://github.com/astur/check-npm-dependents

paulmelnikow:
The API doesn't match the number on the npm site. See  chrisdickinson/npm-get-dependents#3. I confirmed that through my own experimentation with the endpoint it uses on the legacy npm registry mirror. (https://skimdb.npmjs.com/registry/_design/app/_view/dependedUpon)

Scraping npmjs is a nice solution, and more server capacity would enable that. See my proposal at https://github.com/badges/shields/pull/1742#issuecomment-407446963 and some actionable work in #1848.

I have an implementation using Libraries.io, which we've used elsewhere. (#1895) That API is on the slow side, but it works, and may be an acceptable option for the time being.

hiteshpr:
I can work on this.

paulmelnikow:
I just learned about [npms](https://npms.io/)! We could base this on their API.

ainthek:
> I just learned about [npms](https://npms.io/)! We could base this on their API.
I had a quick look at the API on npmjs, but is there any way to get dependents ? I did not find it. Thanx

calebcartwright:
> I had a quick look at the API on npmjs, but is there any way to get dependents ? I did not find it.

Though it's not explicitly called out in the [npms API docs](https://api-docs.npms.io/#api-Package-GetPackageInfo), the API response does include a value for dependents count in `collected.npm.dependentsCount`

https://api.npms.io/v2/package/got
```json
{ 
  "analyzedAt":"2019-10-19T11:32:46.782Z",
  "collected":{ 
    "metadata":{ .. },
    "npm":{ 
      "downloads":[ .. ],
      "dependentsCount":2349,
      "starsCount":124
    },
    "github":{ .. },
    "source":{ .. }
  },
  "evaluation":{ .. },
  "score":{ .. }
}
```

However, I think the same issue exists here relative to the counts being off. The registry shows (at the time of this post) 2,844 dependents for [got](https://www.npmjs.com/package/got) whereas the API response from npms shows 2,349

calebcartwright:
[...]

PR #1895, opened 2018-08-11
Add badge for dependents and dependent repos using Libraries.io [librariesio-dependents librariesio-dependent-repos]

paulmelnikow:
See #1845

### DIFF
--- a/services/librariesio/librariesio-base.js
+++ b/services/librariesio/librariesio-base.js
@@ -2,13 +2,23 @@
 
 const Joi = require('joi')
 const BaseJsonService = require('../base-json')
-const { nonNegativeInteger } = require('../validators.js')
+const { InvalidResponse } = require('../errors')
+const { nonNegativeInteger } = require('../validators')
 
 // API doc: https://libraries.io/api#project
-const schema = Joi.object({
+// We distinguishb between packages and repos based on the presence of the
+// `platform` key.
+const packageSchema = Joi.object({
+  platform: Joi.string().required(),
   dependents_count: nonNegativeInteger,
   dependent_repos_count: nonNegativeInteger,
 }).required()
+
+const repoSchema = Joi.object({
+  platform: Joi.any().forbidden(),
+}).required()
+
+const packageOrRepoSchema = Joi.alternatives(repoSchema, packageSchema)
 
 class LibrariesIoBase extends BaseJsonService {
   static buildUrl(base) {
@@ -19,12 +29,20 @@ class LibrariesIoBase extends BaseJsonService {
     }
   }
 
-  async fetch({ platform, packageName }) {
-    return this._requestJson({
-      schema,
+  async fetch({ platform, packageName }, { allowPackages, allowRepos } = {}) {
+    const json = await this._requestJson({
+      schema: packageOrRepoSchema,
       url: `https://libraries.io/api/${platform}/${packageName}`,
       notFoundMessage: 'package not found',
     })
+    const isPackage = Boolean(json.platform)
+    if (isPackage && !allowPackages) {
+      throw new InvalidResponse({ prettyMessage: 'not supported for packages' })
+    }
+    if (!isPackage && !allowRepos) {
+      throw new InvalidResponse({ prettyMessage: 'not supported for repos' })
+    }
+    return json
   }
 }
 
--- a/services/librariesio/librariesio-base.js
+++ b/services/librariesio/librariesio-base.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const Joi = require('joi')
-const { BaseJsonService } = require('../base')
+const BaseJsonService = require('../base-json')
 const { nonNegativeInteger } = require('../validators.js')
 
 // API doc: https://libraries.io/api#project
--- a/services/librariesio/librariesio-base.js
+++ b/services/librariesio/librariesio-base.js
@@ -1,0 +1,31 @@
+'use strict'
+
+const Joi = require('joi')
+const { BaseJsonService } = require('../base')
+const { nonNegativeInteger } = require('../validators.js')
+
+// API doc: https://libraries.io/api#project
+const schema = Joi.object({
+  dependents_count: nonNegativeInteger,
+  dependent_repos_count: nonNegativeInteger,
+}).required()
+
+class LibrariesIoBase extends BaseJsonService {
+  static buildUrl(base) {
+    return {
+      base,
+      format: '(\\w+)/(.+)',
+      capture: ['platform', 'packageName'],
+    }
+  }
+
+  async fetch({ platform, packageName }) {
+    return this._requestJson({
+      schema,
+      url: `https://libraries.io/api/${platform}/${packageName}`,
+      notFoundMessage: 'package not found',
+    })
+  }
+}
+
+module.exports = LibrariesIoBase
--- a/services/librariesio/librariesio-dependent-repos.service.js
+++ b/services/librariesio/librariesio-dependent-repos.service.js
@@ -6,7 +6,7 @@ const LibrariesIoBase = require('./librariesio-base')
 // https://libraries.io/api#project-dependent-repositories
 class LibrariesIoDependentRepos extends LibrariesIoBase {
   static get category() {
-    return 'downloads'
+    return 'other'
   }
 
   static get defaultBadgeData() {
--- a/services/librariesio/librariesio-dependent-repos.service.js
+++ b/services/librariesio/librariesio-dependent-repos.service.js
@@ -1,0 +1,47 @@
+'use strict'
+
+const { metric } = require('../../lib/text-formatters')
+const LibrariesIoBase = require('./librariesio-base')
+
+// https://libraries.io/api#project-dependent-repositories
+class LibrariesIoDependentRepos extends LibrariesIoBase {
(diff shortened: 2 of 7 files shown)


services/npm/npm-base.js
```diff
@@ -22,7 +22,7 @@ const schema = Joi.object({
 // Abstract class for NPM badges which display data about the latest version
 // of a package.
 module.exports = class NpmBase extends BaseJsonService {
-  static buildUrl(base, { withTag }) {
+  static buildUrl(base, { withTag } = {}) {
```
paulmelnikow: I copied this code along the way, and found this change improved it.

services/librariesio/librariesio-dependent-repos.service.js
```diff
@@ -0,0 +1,47 @@
+'use strict'
+
+const { metric } = require('../../lib/text-formatters')
+const LibrariesIoBase = require('./librariesio-base')
+
+// https://libraries.io/api#project-dependent-repositories
+class LibrariesIoDependentRepos extends LibrariesIoBase {
+  static get category() {
+    return 'downloads'
```
chris48s: Would these badges be better placed in the 'dependencies' category?

services/librariesio/librariesio-dependent-repos.service.js
```diff
@@ -0,0 +1,47 @@
...
+        title: 'Dependent repos (via libraries.io)',
+        previewUrl: 'npm/got',
+      },
+    ]
+  }
+
+  static render({ dependentReposCount }) {
+    return {
+      message: metric(dependentReposCount),
+      color: dependentReposCount === 0 ? 'orange' : 'brightgreen',
```
chris48s: Is this definitely a good/bad (green/orange), or would it be better to think of it as an info (blue)?

services/librariesio/librariesio-dependent-repos.tester.js
```diff
@@ -0,0 +1,28 @@
...
+      value: isMetric,
+    })
+  )
+
+t.create('dependent repo count (not a package)')
+  .get('/npm/foobar-is-not-package.json')
+  .expectJSON({
+    name: 'dependent repos',
+    value: 'package not found',
+  })
```
chris48s: I know we've moved the 'unparsable json response' up a level. I don't think we've done that yet with the "inaccessible" test but we probably should. Its nice to have less boilerplate in these test modules.

chris48s:
One of the slightly confusing things about libraries.io is that it deals with both packages and repositories so there is info about the got _package_ at https://libraries.io/npm/got but also the corresponding _repo_ at https://libraries.io/github/sindresorhus/got .

Do you think it is worth naming these endpoints/descriptions to make it clear that they relate to _packages_ and not _repositories_ ?
a) This would make it clearer what the badges do/reduce confusion about how to use them
b) If libraries.io were to add support for these statistics at repository level (as well as package level), we haven't pained ourselves into a corner with our endpoint names

Related: if I call `librariesio/dependent-repos/github/sindresorhus/got.svg` that will generate a `invalid json response` badge (which is not wrong - https://libraries.io/api/npm/got and https://libraries.io/api/github/sindresorhus/got are completely different objects, so the schema/validation works and handles the error for us: woo!), but do you reckon it we should try and handle this particular case in such a way that we provide a more userful error?

services/librariesio/librariesio-base.js
```diff
@@ -0,0 +1,31 @@
...
+const BaseJsonService = require('../base-json')
+const { nonNegativeInteger } = require('../validators.js')
+
+// API doc: https://libraries.io/api#project
+const schema = Joi.object({
+  dependents_count: nonNegativeInteger,
+  dependent_repos_count: nonNegativeInteger,
+}).required()
+
+class LibrariesIoBase extends BaseJsonService {
```
chris48s: Hmm.. one more thing: Thinking a step ahead, when we port over the [other libraries io badges](https://github.com/badges/shields/blob/db4bffb300a8d9c794fbd089a2f0f9586415e500/server.js#L6516-L6578) we'll also need to be able to pass another param (version/latest) which is not relevant here and we'll also call a different API endpoint, so probably a single `LibrariesIoBase` class which hard-codes this stuff isn't going to be the ideal abstraction.

Maybe we just cross that bridge when we come to it, but maybe worth considering now?

services/librariesio/librariesio-dependent-repos.tester.js
chris48s: Ignore this - I can see you've already done it in #1890

services/librariesio/librariesio-dependent-repos.service.js
paulmelnikow: I'm not sure. Dependencies is about conveying the status of the project and what it requires. Dependents is about conveying the popularity of the project and how widely used it is. It's more like downloads I think. Though obviously not a perfect fit.

We could consider renaming Downloads to Popularity. There's also some overlap between Social and Popularity. The Github badges, and maybe even the Twitter badges, could move into Popularity, too.

Thoughts?

services/librariesio/librariesio-dependent-repos.service.js
paulmelnikow: I based this on the download badges, which are green. I think it's more like that those than like a license badge.

paulmelnikow:
> Do you think it is worth naming these endpoints/descriptions to make it clear that they relate to _packages_ and not _repositories_ ?
> …
> Related: if I call `librariesio/dependent-repos/github/sindresorhus/got.svg` that will generate a `invalid json response` badge (which is not wrong - https://libraries.io/api/npm/got and https://libraries.io/api/github/sindresorhus/got are completely different objects, so the schema/validation works and handles the error for us: woo!), but do you reckon it we should try and handle this particular case in such a way that we provide a more userful error?

Is there an easy way to ID package vs repo in the response? Maybe we could detect which it is. That way the user doesn't have to specify. Then we can have a single endpoint, and until they have stats support on repos, we could return an error which says something like "only available for packages, not repos."

services/librariesio/librariesio-base.js
paulmelnikow: Yea! Initially I started writing that in here, but stopped because I didn't want to add code that had no use or a test. We could add a second parameter `{ withVersion }` like NpmBase has.

services/librariesio/librariesio-dependent-repos.service.js
chris48s: At the moment all of the badges under the 'downloads' category are actually counting a number of downloads and is already quite large. I think that makes it a sensible self-contained unit - it is one of the more clearly defined categories.

Maybe 'rating' and 'social' is better though of as 'popularity'?

chris48s:
> Is there an easy way to ID package vs repo in the response?

We could use presence/absence of a `platform` key in the response. Package managers are referred to as "platforms" on libraries.io: https://libraries.io/api/platforms whereas supported SCMs aren't "platforms". Example SCM responses:
https://libraries.io/api/github/sindresorhus/got
https://libraries.io/api/bitbucket/gutworth/six
https://libraries.io/api/gitlab/dautovri/TensorFlow-Tutorials

services/librariesio/librariesio-base.js
chris48s: > I didn't want to add code that had no use or a test

:+1:  Lets deal with that case when we actually refactor the service that needs this

services/librariesio/librariesio-dependent-repos.service.js
paulmelnikow: I could see download stats being well defined enough to leave alone. At the same time, I do think changing Rating to Reviews, and Downloads to Popularity or Adoption might be my preferred option. The downloads category is large, however download badges are very straightforward and heterogeneous, and I don't think adding star / fork counts, download rankings, and dependents is making things worse.

Rating is a mixed bag. Some of them are reviews, some are ranks, some are star counts, some are reputation. Gem rank is counting downloads.

Oh man, as I look through these groupings, I feel like we could review a lot of these. Several of them could be improved.

I just opened #1903 to fix a misclassified download badge.

This feels like it's own can of worms and I don't want to get hung up here. How about we put these in "other" for now, and open a new issue to think about the categories?

Tangentially related… I've been thinking we should consider splitting the legacy badge examples and legacy services into individual files in `services/`, even before the rewrite happens. It would become a bit harder to access git blame on the implementations, but could make these categorizing projects – and the rewrites! – a lot easier to deal with.

paulmelnikow:
> We could use presence/absence of a `platform` key in the response.

Sounds good to me!

services/librariesio/librariesio-dependent-repos.service.js
chris48s: > This feels like it's own can of worms and I don't want to get hung up here. How about we put these in "other" for now, and open a new issue to think about the categories?

Agreed. There are some related thoughts on this topic in https://github.com/badges/shields/pull/1762#issuecomment-403583651

paulmelnikow: Yea, I like what I'm seeing there.

I like the idea of having a separate page for each service, which offers a collection of related badges. I think that would be a nice addition to the category pages, which are more useful for browsing.

paulmelnikow:
This is ready for another look.

PR #1898, opened 2018-08-11
Add [librariesio-sourcerank]

### LINKED ISSUE #1532: Add Libraries.io SourceRank badge
Libraries.io provides a ScoreRank based on existense of metainfo (readme, license), number of dependants and subscribers etc.

Documentation: https://libraries.io/api#project-sourcerank

Note that the docs are outdated, the request actualy outputs detailed JSON, not just one number, they have to be summed up:

```
{"basic_info_present":1,"repository_present":1,"readme_present":1,"license_present":1,"versions_present":1,"follows_semver":0,"recent_release":1,"not_brand_new":1,"one_point_oh":1,"dependent_projects":10,"dependent_repositories":6,"stars":4,"contributors":1,"subscribers":1,"all_prereleases":0,"any_outdated_dependencies":-1,"is_deprecated":0,"is_unmaintained":0,"is_removed":0}
```

A badge like `sourcerank : sum` would be nice.

paulmelnikow:
Depends on #1895

It seems like 30+ is great; under 10 pretty bad. I picked some values for a color scheme.

Closes #1532


services/librariesio/librariesio-sourcerank.service.js
```diff
@@ -0,0 +1,48 @@
+'use strict'
+
+const { colorScale } = require('../../lib/color-formatters')
+const LibrariesIoBase = require('./librariesio-base')
+
+const sourceRankColor = colorScale([10, 15, 20, 25, 30])
+
+class LibrariesIoSourcerank extends LibrariesIoBase {
+  static get category() {
+    return 'other'
```
chris48s: This could also maybe sensibly live under 'dependencies'?

services/librariesio/librariesio-sourcerank.service.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  static get url() {
+    return this.buildUrl('librariesio/sourcerank')
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Libraries.io SourceRank',
```
chris48s: SourceRank 2.0 is not 'launched', but it is kind of in beta alongside SourceRank 1.0 (at least for some package ecosystems).

e.g:
SourceRank 1.0 breakdown: https://libraries.io/cargo/regex/sourcerank
SourceRank 2.0 breakdown: https://libraries.io/cargo/regex/score

can we clarify that this is a SourceRank 1.0 score here (maybe in the actual badge label too). I'm not sure if it should be in the endpoint. I'm slightly unclear on whether one day SR1 will just disappear and be replaced by SR2 or whether we might at some point need to maintain 2 versions of the badge.

services/librariesio/librariesio-sourcerank.service.js
paulmelnikow: I thought about that, though this is a different kind of badge. Dependencies is giving internal information about the project, rather conveying how widely used or well maintained it is.

Would it make sense to put it under Rating?

services/librariesio/librariesio-sourcerank.service.js
paulmelnikow: Well, we could take the lead from the SourceRank displayed on the libaries.io site. https://libraries.io/cargo/regex has SR 1.0 labeled SourceRank. What if we follow along with that?

It's possible they'll update the `rank` field in this endpoint when they officially launch.

services/librariesio/librariesio-sourcerank.service.js
chris48s: 'Rating' makes the most sense :+1:

services/librariesio/librariesio-sourcerank.service.js
chris48s: Yeah, that's a fair point - given they are still just referring to it as 'SourceRank', we should do the same for now. I'll keep an eye on progress of this anyway

PR #1900, opened 2018-08-11
Clean up our callback style and enforce no exclusive tests [gem gemnasium dynamic-json dynamic-xml]

paulmelnikow:
We use arrow functions in most places; this enforces it.

Passing arrow functions to Mocha is discouraged: https://mochajs.org/#arrow-functions

This was a mix of autofixes and hand adjustments.

### DIFF
--- a/services/base.spec.js
+++ b/services/base.spec.js
@@ -131,7 +131,7 @@ describe('BaseService', function() {
     })
   })
 
-  describe.only('Error handling', function() {
+  describe('Error handling', function() {
     it('Handles internal errors', async function() {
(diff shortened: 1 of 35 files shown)


services/base.spec.js
```diff
@@ -131,7 +131,7 @@ describe('BaseService', () => {
     })
   })
 
-  describe.only('Error handling', function() {
+  describe('Error handling', function() {
```
paulmelnikow: Yikes!

paulmelnikow: Feel like it's important to get this in ASAP. Could I get a second 👍?

PR #1740, opened 2018-06-17
WIP: [cdnjs] generate static examples without api call

chris48s:
Having chewed over the discussion in issue #1700 and drafted a few ideas, here is one possible solution:

If we use the service class internals to produce the example from an object and split the `handle()` function into 2:

* One function which fetches data
* One function which formats it as on object that can be consumed by `_makeBadgeData()`

then it becomes easier to format an example value consistently with the real badge behviour without making any API call. If we seperate concerns correctly, this should minimise the scope for our examples to 'drift' or require a lot of maintenance.

In most cases, what we're doing fits cleanly into that pattern. In the cases where we are doing something more complicated, we can write a more complex `handle()` function, elect not to return a `staticExample` object and just return a dynamic `previewUrl` as before. If we come up with a solution that addresses the common pattern, that allows us to make an improvement over where we are now.

The one downside is that the examples obviously don't reflect reality. This will probably cause some confusion:

gifrecord_2018-06-17_133911

.. but this is a problem if we implement the suggestion in issue #1700 in any way. It is not inherent to this implementation.

Thoughts? I've picked one simple example here to demonstrate, but if this approach seems sensible I can have a look at updating the examples for other badges we've already refactored, adding tests, etc..

PR #1901, opened 2018-08-11
use fetch/render pattern in [apm appveyor cdnjs clojars gem]

chris48s:
Once this is in, I will pick up #1740 again

### DIFF
--- a/services/apm/apm.service.js
+++ b/services/apm/apm.service.js
@@ -41,11 +41,13 @@ class BaseAPMService extends BaseJsonService {
 }
 
 class APMDownloads extends BaseAPMService {
+  static render(downloads) {
+    return { message: metric(downloads), color: 'green' }
+  }
+
   async handle({ repo }) {
     const json = await this.fetch(repo)
-
-    const downloads = json.downloads
-    return { message: metric(downloads), color: 'green' }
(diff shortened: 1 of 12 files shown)


services/apm/apm.service.js
```diff
@@ -41,11 +41,13 @@ class BaseAPMService extends BaseJsonService {
 }
 
 class APMDownloads extends BaseAPMService {
+  static render(downloads) {
```
paulmelnikow: Could we pass `{ downloads }` here instead? I prefer that pattern as it reads and grows well.

chris48s: Shall I switch all the `fetch()` and `render()` methods to use the object destructuring syntax, or just the `render()`s ? I don't have particularly strong opinions on this, I'm just slightly unsure what the tradeoff is in this situation..

paulmelnikow: I think named parameters are great. They make the calling code self-documenting, and allow functions to grow organically with minimal impact at the callsite. As the number of parameters grows past one, naming them reduces the risk of passing the wrong value to a function. Of course JS doesn't have named parameters, but destructuring provides basically the same thing.

So, I'd suggest doing both!

PR #1903, opened 2018-08-12
Create Monitoring category

paulmelnikow:
It's really not a downloads badge.

As we're considering our groupings, it might be nice to add an Uptime section for this + UptimeRobot. I'm open to doing that now if folks think we should.


paulmelnikow:
👍for Monitoring. Shall I update this PR to add that?

chris48s:
Yeah, why not..

PR #1907, opened 2018-08-13
Server start time badge

RedSparr0w:
- Moved `time.service` → `debug.service` or maybe `server.service`?
_thought it may be better to keep all this sort of thing in 1 folder/file rather than creating a bunch of extra folders etc._
- Added badge for server start time (to be used to check when server last restarted)
- Badges shown as UTC 0

not particularly sure how/what else needs to be done, so let me know if you have any suggestions.

**Example:**
starttime

**Note:**
this also changes `/servertime.svg` → `/debug/time.svg`:
time-blue.svg) → time


RedSparr0w:
Can't seem to get the test to run corectly.
```console


  1) Uncaught error outside test suite
  The server
Error: getaddrinfo ENOTFOUND lib/**/*.spec.js
    at errnoException (dns.js:50:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:92:26)
```

chris48s:
> Can't seem to get the test to run corectly.

I haven't checked the code out locally, but is the problem here a circular import?

You've got a service which imports from `server.js`

https://github.com/badges/shields/blob/3f95f81277152fd53a3b30e2e8774b6fc438fe0a/services/debug/debug.service.js#L4

but then in `server.js` we import all the services:

https://github.com/badges/shields/blob/9c64589944cb3edef3b0143b221fab9effbf4773/server.js#L190-L194

Maybe its trying to eat its own tail :snake:

RedSparr0w:
Thanks @chris48s, it sure was.
I've moved it to `/lib/server-config.js` although i'm not sure if that's the best place for it?

paulmelnikow:
How about in `lib/uptime.js` or something like that? I'm also trying to inject config rather than import it. It makes testing easier.

I backslid a bit in https://github.com/badges/shields/blob/3b9d2325b8203df6169a539564203082d946f3f6/services/base.js#L19-L26 but hopefully that is temporary.

server.js
```diff
@@ -7536,12 +7534,12 @@ function(data, match, end, ask) {
...
+  if (+(new Date(ask.req.headers['if-modified-since'])) >= +config.serverStartTime) {
     ask.res.statusCode = 304;
     ask.res.end();  // not modified.
     return;
   }
-  ask.res.setHeader('Last-Modified', serverStartTime.toGMTString());
+  ask.res.setHeader('Last-Modified', config.serverStartTime.toGMTString());
 
   // Badge creation.
   try {
```
niccokunzmann: This file is all fine.

lib/server-config.js
```diff
@@ -71,6 +72,7 @@ const config = {
   },
   rateLimit: envFlag(process.env.RATE_LIMIT, true),
   handleInternalErrors: envFlag(process.env.HANDLE_INTERNAL_ERRORS, true),
+  serverStartTime,
 }
```
niccokunzmann: this file is all fine.

services/debug/debug.service.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  static get url() {
+    return {
+      base: 'debug',
+      format: '(time|starttime)',
+      capture: ['detail'],
+    }
+  }
+}
```
niccokunzmann: looks good to me

paulmelnikow:
Reopened as #2631.

Issue #1358, opened 2017-12-08
The Great Service Rewrite 🏆


paulmelnikow:
Now that #963 is merged it’s time to start thinking about 🏆 **The Great Service Rewrite**! 🏆 

The rewrite has three goals:

- Fill in remaining service-test coverage
- Move services from `server.js` and `lib/all-badge-examples.js` into separate modules
- Rewrite service implementations at high quality and with consistent coding style (automatically checked)

The service rewrite necessitates upgrading to Node 8 on the servers. Today they're running Node 6. I’m not sure exactly how long it will take to do that. I don’t have access to upgrade Node, though I’ve reached out to @espadrine about it.

#963 which is now merged to the [node-8 branch](https://github.com/badges/shields/tree/node-8) included a new implementation of the AppVeyor badge. It’s about to fall out of date with the merge of #1321. I’m not concerned about AppVeyor. After all it's just one badge that we could re-port at any point.

However, we should decide on a short-term approach for how to handle badge changes.

Some ideas:

1. Make a list of **missing service tests** and start to work down the list, merging to master.
2. Open pull requests for **new badges** against **the `node-8` branch**, using the new service code. Put a temporary freeze on new badges in `master`.
3. Choose 5–10 badges to **port to the `node-8` branch**. Except for security fixes, freeze their implementations and tests in `master`.
4. Refactor helper functions in `master` and then merge into `node-8`.

We could do 1 + 2, or 1 + 3, or 1 + 2 + 3, or 1 + 2 + 3 + 4. None of these are mutually exclusive.

I created [a project to track this](https://github.com/badges/shields/projects/2), though it doesn't have anything useful in it yet.

Thoughts?

Let's get the ball rolling! 🎳

chris48s:
re: point 1, this is my rough assessment of where you are right now:

| service | has tests |
| ------- | --------- |
| amo | :heavy_check_mark: |
| ansible | :heavy_check_mark: |
| apm |  |
| appveyor | :heavy_check_mark: |
| aur |  |
| beerpay |  |
| bintray |  |
| bitbucket | :heavy_check_mark: |
| bithound |  |
| bitrise | :heavy_check_mark: |
| bountysource | :heavy_check_mark: |
| bower | :heavy_check_mark: |
| cauditor |  |
| cdnjs |  |
| chocolatey |  |
| chrome-web-store | :heavy_check_mark: |
| circleci |  |
| clojars |  |
| cocoapods |  |
| codacy |  |
[...]

PR #1897, opened 2018-08-11
Change format of [npm] type definitions badge

paulmelnikow:
This is based on feedback at chalk/chalk#286.

I'd like to add a second badge that pulls this information from a github repo, so that I can implement his suggestions to look for files within the project. There doesn't seem to be a way to get the file contents through the registry without downloading the tarball.

PR #1922, opened 2018-08-15
Rewrite [pypi]; affects [npm]

paulmelnikow:
When I worked on NPM and UptimeRobot, I created separate PRs for the refactors. Does that approach make sense here? Or would it be just as easy to review this in total?

Todo:

- [x] Fix broken tests
- [x] Contract
- [ ] Tests
    - [ ] Deprecated()
    - [ ] enforceDeprecation()
    - [ ] renderVersionBadge()
    - [x] getPackageFormats()

Ref: #1358


services/npm/npm-license.service.js
```diff
@@ -9,10 +9,6 @@ module.exports = class NpmLicense extends NpmBase {
     return 'license'
   }
 
-  static get defaultBadgeData() {
-    return { label: 'license' }
-  }
-
```
paulmelnikow: This already defaults to `this.category` which is `'license'`.

services/npm/npm.tester.js
```diff
@@ -213,7 +213,6 @@ t.create('license when network is off')
 // This tests error-handling functionality in NpmBase.
 t.create('when json is malformed for scoped package')
   .get('/l/@cycle%2Fcore.json')
-  .only()
```
chris48s: I haven't looked at the rest of this, but can we get this change in ahead of the rest of this PR. We've got other open PRS ( #1897 #1740 ) which should run against all the NPM tests.

Also, there's a couple of times a `.only()` has been missed in review. Maybe we should have a `danger` rule for this..

paulmelnikow: Done in #1926.

lib/deprecation-helpers.js
```diff
@@ -21,7 +22,14 @@ const getDeprecatedBadge = function(label, data) {
   return badgeData
 }
 
+function enforceDeprecation(effectiveDate) {
+  if (Date.now() >= effectiveDate.getTime()) {
```
RedSparr0w: `Date.now()`
Never knew this was a thing 👍

paulmelnikow:
Echoing my question from the top post…

> When I worked on NPM and UptimeRobot, I created separate PRs for the refactors. Does that approach make sense here? Or would it be just as easy to review this in total?

Any thoughts?

If this needs to be split up, I want to do that before/as I clean it up, not after I clean it up.

chris48s:
_In theory_ it is nicer to review stuff in smaller chunks, but given the original implementation defines a bunch of badges in one huge function I reckon trying to do that in this case will introduce more intermediate complexity than its worth.

paulmelnikow:
Okay. Agreed the pypi badges should just come in all at once. I was thinking I _could_ separate out the changes in the base functionality, and the little bit of refactoring in npm. Let me know if you think it’s worth it.

services/pypi/pypi-djversions.service.js
```diff
@@ -0,0 +1,53 @@
+'use strict'
+
+const PypiBase = require('./pypi-base')
+const { sortDjangoVersions, parseClassifiers } = require('./pypi-helpers')
+
+module.exports = class PypiDjangoVersions extends PypiBase {
+  static get category() {
+    return 'version'
```
PyvesB: This should be updated to 'platform-support' following #1948.

services/pypi/pypi-pyversions.service.js
```diff
@@ -0,0 +1,63 @@
+'use strict'
+
+const PypiBase = require('./pypi-base')
+const { parseClassifiers } = require('./pypi-helpers')
+
+module.exports = class PypiPythonVersions extends PypiBase {
+  static get category() {
+    return 'version'
```
PyvesB: This should be updated to 'platform-support' following #1948.

services/pypi/pypi-pyversions.service.js
```diff
@@ -0,0 +1,63 @@
...
+        keywords: ['python'],
+      },
+    ]
+  }
+
+  static render({ versions }) {
+    const versionSet = new Set(versions)
+    // We only show v2 if eg. v2.4 does not appear.
+    // See https://github.com/badges/shields/pull/489 for more.
+    ;['2', '3'].forEach(majorVersion => {
```
PyvesB: What is this semicolon doing here? 😄

services/pypi/pypi-helpers.js
```diff
@@ -50,8 +50,25 @@ const parseClassifiers = function(parsedData, pattern) {
   return results
 }
 
+function getPackageFormats(packageData) {
```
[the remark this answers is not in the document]
paulmelnikow: Good idea!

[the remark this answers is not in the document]
paulmelnikow: Updated.

services/pypi/pypi-pyversions.service.js
PyvesB: This makes much more sense now. Thanks for explaining!

paulmelnikow:
Thanks! I will get this merged and then circle back on the few remaining tests.

Issue #1905, opened 2018-08-12
Continuing the badge sort


paulmelnikow:
We had some discussion in https://github.com/badges/shields/pull/1895/files/f4e6e0c9c95b3c5530c98f614cba4efc65579426#r209434975 about the ratings, downloads, and social categories. There are some similar things in other, and we just added the dependents badges there, too.

Some initial proposals:

1. The Gem rank badges in http://shields.io/#/examples/rating are just rankings of downloads. Let's move them from **Rating** to **Downloads**.
2. Let's move the Docker Build badges from **Other** to **Build**.
3. Let's break out the code quality badges from **Build** to a new category. It could be called **Code Quality**. Or, we could lump in Libscore, SourceRank, Semver Stability, PyPi Status, CocoaPods docs %, and maybe the Maintenance badge, and call it **Project Quality**.
4. Create a **Size** category for the bundlephobia, github file, github repo, imagelayers, and microbadger size badges.

paulmelnikow:
I opened #1948 with a couple more proposals.

paulmelnikow:
Swinging by again, two proposals:

1. **Code Quality**
    - CocoaPods docs %
    - And many others
2. **Project Activity** or **Activity** category:
    - PyPi Status
    - Maintenance
    - And many others, like commits per year, latest release, etc.
3. Move a few into **Rating**:
    - Libscore
    - SourceRank
    - Semver Stability

paulmelnikow:
Opened #2483 which I think is the last of these ideas (that are still making sense to me).

PR #1927, opened 2018-08-16
refactor [circleci] integration

chris48s:
refs #1358


services/circleci/circleci.service.js
```diff
@@ -0,0 +1,106 @@
+'use strict'
+
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
+
+const circleSchema = Joi.array()
+  .items(Joi.object({ status: Joi.string() }))
+  .min(1)
```
chris48s: I _think_ this is a sensible rule. I can't see a case for treating `[]` as a valid response here..

services/circleci/circleci.service.js
```diff
@@ -0,0 +1,106 @@
...
+      options: { qs: query },
+      errorMessages: { 404: 'project not found' },
+    })
+  }
+
+  static render({ status, color }) {
+    return { message: status, color: color }
+  }
+
+  transform(data) {
```
chris48s: Is this supposed to be a `static` method? I can't remember what the rule is now. When stuff should/shouldn't be `static` is definitely a topic we need to cover when we write docs.

services/circleci/circleci.service.js
```diff
@@ -0,0 +1,106 @@
...
+      shieldsStatus = 'passing'
+    }
+
+    return { status: shieldsStatus, color: color }
+  }
+
+  async handle({ token, type, userRepo, branch }) {
+    const json = await this.fetch({ token, type, userRepo, branch })
+    const { status, color } = this.transform(json)
+    return this.constructor.render({ status, color })
```
chris48s: As I write and review more code using this `fetch()`/`render()` pattern, I think we're really close to a point where we should update the docs to formalise this pattern. I think if #1922 (another complex service 'family') goes in without any major change to the base classes (beyond adding the `Deprecated` exception), we should make docs the next priority.

services/circleci/circleci.service.js
```diff
@@ -0,0 +1,106 @@
+'use strict'
+
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
+
+const circleSchema = Joi.array()
+  .items(Joi.object({ status: Joi.string() }))
```
paulmelnikow: Should this be `status: Joi.string().required()`?

services/circleci/circleci.service.js
```diff
@@ -0,0 +1,106 @@
...
+    return this._requestJson({
+      url,
+      schema: circleSchema,
+      options: { qs: query },
+      errorMessages: { 404: 'project not found' },
+    })
+  }
+
+  static render({ status, color }) {
+    return { message: status, color: color }
```
paulmelnikow: We should turn on [the eslint rule for this](https://eslint.org/docs/rules/object-shorthand); `color: color` can just be `color`.

services/circleci/circleci.service.js
paulmelnikow: Yea, in general you can make anything static that doesn't use `this`. The exception is if you're overriding superclass method, in which case making it static would break the overriding.

services/circleci/circleci.service.js
```diff
@@ -0,0 +1,106 @@
...
+        color = 'red'
+        return { status: shieldsStatus, color: color }
+      } else if (['no_tests', 'scheduled', 'not_run'].includes(circleStatus)) {
+        color = 'yellow'
+        shieldsStatus = circleStatus.replace('_', ' ')
+        return { status: shieldsStatus, color: color }
+      } else {
+        color = 'lightgrey'
+        shieldsStatus = circleStatus.replace('_', ' ')
+        return { status: shieldsStatus, color: color }
```
paulmelnikow: So it seems the purpose of this function is to summarize the multiple statuses down to a single status. I'm fine with that, though it does seem possible that it's better thought of as a rendering concern. For example, we could have a badge option that showed these like **2 passed, 1 failed**.

Also, this looks buggy. What if the first test is scheduled and the second is failed? That should be reported as failed, but I think it will show as scheduled instead. Probably it would be better to check if any of the statuses are `'failed'`, then move down the list.

Would it be cleaner (and less buggy) to return an array of statuses, or extract an array of statuses in `handle()`, and delegate the summarizing and coloring to `render()`?

services/circleci/circleci.service.js
```diff
@@ -0,0 +1,106 @@
...
+  static get category() {
+    return 'build'
+  }
+
+  static get url() {
+    return {
+      base: 'circleci',
+      format:
+        '(?:token/(w+))?[+/]?project/(?:(github|bitbucket)/)?([^/]+/[^/]+)(?:/(.*))?',
+      capture: ['token', 'type', 'userRepo', 'branch'],
```
paulmelnikow: Re `type`: Circle calls this `vcsType` which seems a bit clearer.

services/circleci/circleci.service.js
chris48s: This is a very good point which I did not consider when I was looking at #1815 but now I think about this, there are many worms in this can..

It is clear that if *all* tests are passing, that should summarise to 'pass', and if *any* of the tests are failing, that should summarise to 'fail', but if we get back something like:

```
[
  { 'status': 'running' },
  { 'status': 'infrastructure_fail' },
  { 'status': 'scheduled' },
]
```

or

```
[
  { 'status': 'success' },
  { 'status': 'success' },
  { 'status': 'queued' },
  { 'status': 'no_tests' },
]
```

what is an appropriate summary of that?

Looking at [the docs](https://circleci.com/docs/api/v1-reference/), it seems that we could get any arbitrary combination of:

* retried
* canceled
* infrastructure_fail
* timedout
* not_run
* running
* failed
* queued
* scheduled
* not_running
* no_tests
* fixed
* success

paulmelnikow:
```
[
  { 'status': 'running' },
  { 'status': 'infrastructure_fail' },
  { 'status': 'scheduled' },
]
```

-> **infrastructure_fail**

```
[
  { 'status': 'success' },
  { 'status': 'success' },
  { 'status': 'queued' },
  { 'status': 'no_tests' },
]
```

Huh, is that actually possible? **no tests** seems to mean there would be no other results. If that ever happened I'd be okay displaying **queued** or **indeterminate**.

In the case of

```
[
  { 'status': 'success' },
  { 'status': 'success' },
  { 'status': 'no_tests' },
]
```

I think **indeterminate** is appropriate.

Maybe we could prioritize the results by severity, and display the highest severity result that appears? We'll have to make some judgement calls when we do that, though I think that's okay.

We could special-case **no tests** which really should only exist on its own.

chris48s: I think you might have taken those examples too literally - I just made up some arbitrary combinations :)

However, its very useful that this has prompted me to think about this more clearly and properly read the documentation because it turns out:

a) I've been trying to solve completely the wrong problem
b) It is only through luck that #1815 has not actually broken anything for users

What I _thought_ I've been doing here is combining the statuses of a list of jobs/workflows to make a build status, but what I've _actually_ been trying to do is combine >1 build statuses (which already summarise multipe jobs/workflows) to make.. a thing that makes no sense. By sheer good luck, we were passing `&limit=1` the whole time so in the real world it all made no difference anyway because we've only ever really been operating on an array of length 1 (despite having written some tests which assume an array of length >1).

So.. the upshot of all that is:
* #1815 was completely pointless and fixed nothing - I've put the code back how it was before I started in 90160bc
* I've also added validation and a test to clarify that we are expecting an array of length 1 and removed the pointless tests in 90160bc
* Whatever the issue was in #1792 is not fixed and we should re-open that issue. I was completely wrong about the problem.

:man_facepalming:

paulmelnikow: I agree with much of that, though I was just looking at https://circleci.com/docs/api/v1-reference/ and it seems to me that a build does _not_ combine multiple jobs. A build seems to refer to e.g. a single tile on https://circleci.com/gh/badges/shields = an individual job.

https://circleci.com/api/v1.1/project/github/badges/shields

Unless `outcome` and `status` are conveying different things, or there's something else I'm not seeing…

services/circleci/circleci.service.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  async handle({ token, vcsType, userRepo, branch }) {
+    const json = await this.fetch({ token, vcsType, userRepo, branch })
+    return this.constructor.render({ status: json[0].status })
+  }
+
+  // Metadata
+  static get defaultBadgeData() {
+    return { label: 'build' }
```
RedSparr0w: Do you think it would be worth leaving this out as it defaults to the `category` already?

services/circleci/circleci.service.js
RedSparr0w: I feel like `outcome` may actually be the one we are looking for:
```javascript
  "outcome" : "failed", // :canceled, :infrastructure_fail, :timedout, :failed, :no_tests or :success
  "status" : "failed", // :retried, :canceled, :infrastructure_fail, :timedout, :not_run, :running, :failed, :queued, :scheduled, :not_running, :no_tests, :fixed, :success
```

chris48s: Right, yep so we are only looking at one job when there may be N jobs in the build (although I think we are using using the wrong terminology there as the docs refer to one record in this array as a 'build'). I'm going to refer to 'build' and 'group' moving forwards.

`status` and `outcome` seem to both relate to a build. For example, if we look at the branch [dependabot/npm_and_yarn/simple-icons-1.8.4](https://circleci.com/api/v1.1/project/github/badges/shields/tree/dependabot/npm_and_yarn/simple-icons-1.8.4) `build_num`s 11388, 11387, 11386, 11385 and 11384 all group together, but if we look at the status and outcome values:

```
build_num    status    outcome
11388        fixed     success
11387        failed    failed
11386        success   success
11385        failed    failed
11384        failed    failed
```

Although 'outcome' is probably useful to reduce the number of possible values we have to deal with, if it was summarising the group, I'd expect them all to be 'failed'.

So we do need to aggregate multiple status (or outcome values). What I'm slightly struggling with now is how to connect builds in an array like [this](https://circleci.com/api/v1.1/project/github/badges/shields/tree/master) (which contains builds relating to multiple groups) as being part of the same group (which is what we're trying to summarise).

One thing I have thought of is we have the `vcs_revision` property, so we can identify builds in the array which were run against the same commit (on the same branch). Is that a solid way of doing this, or can we run the same build more than once on the same commit? I _think_ if we restart a build it re-uses the same number.

chris48s: Looking at an example, picking all of the objects with the same `vcs_revision` is going to double-count retried builds. For example:
https://circleci.com/gh/badges/shields/11764 and
https://circleci.com/gh/badges/shields/11769
are both `services-pr` on the same commit hash on the same branch.

paulmelnikow: Can we use `workflows.workflow_id`?

```js
workflows: {
  job_name: "frontend",
  job_id: "ef27df67-7aad-4022-a159-2c55718da667",
  workflow_id: "5b94f353-2060-4c80-8b35-941174632ce3",
  workspace_id: "5b94f353-2060-4c80-8b35-941174632ce3",
  upstream_job_ids: [
    "4fac3028-db01-4497-a3c5-97c75194eb0d"
  ],
  upstream_concurrency_map: { },
  workflow_name: "on-commit"
}
```

RedSparr0w: Yeah, i think that is going to be the only way, although we are also going to need to increase the `?limit=`

paulmelnikow:
Hey @chris48s this needs some conflict resolution. Since the changes here to `server.js` are simple, it should be easiest to grab the version from master, and then delete the circle implementation (and remove unused helper functions). I'm planning to validate the master branch and deploy tonight (UTC-7). It would be great to include this one, too!

chris48s:
> It would be great to include this one, too!

There are 2 bits of work going on in this PR:

* Refactoring the existing code (which has a bug)
* Fixing a bug

Do you mean you want to get the refactored (but still buggy) code into a state where you can merge it and then treat the bugfix as a different job?

If you want to just get all the service code out of `server.js` another option is to convert this into a `LegacyService` and I'll just rebase and deal with it.. I don't think it makes much difference.

paulmelnikow:
Ah, no, sorry, I forgot about the issue with multiple builds. I’m flexible. We can do this in whatever order makes the most sense to you.

services/circleci/circleci.service.js
chris48s: Having chewed this over, I think it is useful to keep labels explicit.

It is helpful that we can fall back to category to ensure we do have a label, but if you think about the re-organisation work going on in https://github.com/badges/shields/issues/1905#issue-349816846 I wouldn't want to be in a situation where we move badges around and all the labels change.

For example, if we renamed the 'Build' category to 'Continuous Integration' (not saying we should do this - just an example), we would probably still want this label to be 'build'. Lets not create a situation where doing something like that creates a domino effect causing unpredictable changes elsewhere.

chris48s:
Right, I've sorted out the merge conflicts and pushed that. What's here now is refactored into the new structure maintaining the existing functionality (only looking at the most recent one build).

I've had a look at some examples and I think using `workflows.workflow_id` to group related builds is the right thing to do so I'm going to look at that next.

If you want to merge this in the meantime, I can submit that as a new PR. If not, I'll append it to this one.

chris48s:
I spotted another issue with the way I've done the static examples, so this now needs another review.

I'm now quite close to being able to submit a fix for #1792 but this has spiralled into quite a complex job as I've found more and more edge cases. The fix will require adding a lot of additional code and tests. As such, I think once I've got another :+1: on this I'll merge this as it stands and we will deal with that as another PR to make the review easier to deal with because we won't be trying to deal with that and the refactoring all at once.

PR #1928, opened 2018-08-16
danger: highlight service in msg

tooomm:
I've no idea if that works as intended. :)
The idea was to somehow highlight the touched service from the rest of the text the bot places in PR's, e.g: https://github.com/badges/shields/pull/1740#issuecomment-397879413

I also thought about wrapping it in "`" to get a nice markdown highlight, but I wasn't sure about how to correctly escape it.
What do you think would be a suitable improvement here?


paulmelnikow:
Thanks for doing this @tooomm!

👍on @RedSparr0w's suggestion of using backticks.

tooomm:
Looking closer into `dangerfile.js`, there are several other usages of backticks already:
https://github.com/badges/shields/blob/8bbe2adfaf14099fa88b729217bcf617c6bdc87a/dangerfile.js#L47
Will try to find a uniform solution.

paulmelnikow:
Thanks @tooomm!

Issue #1923, opened 2018-08-15
Shortcut to run all tests for service family


chris48s:
In a case like [librariesio](https://github.com/badges/shields/tree/master/services/librariesio) where we have split tests out across multiple files, it would be helpful if  `npm run test:services -- --only=librariesio` ran all of the rests in `librariesio/*.tester.js`.

This would mean a PR that needs to run them all could just have `[librariesio]` in the title instead of  `[librariesio-dependent-repos librariesio-dependents librariesio-sourcerank]`

paulmelnikow:
This is addressed in #1933.

PR #1932, opened 2018-08-17
Reduce boilerplate for creating new testers

paulmelnikow:
This is a bit of sugar that reduces the boilerplate for creating testers in what I expect will become the standard case: a service in `foo/foo-thing.service.js` with its tests in `foo/foo-thing.tester.js`.

This makes a small stylistic change, which is to name the service by its CamelCase class name rather than an invented snake-case ID. This name only needs to be specified in once place: `class FooThing extends Base[Json]Service`.

We should address #1923 as well, though it's orthogonal to this.

~~If folks like this I will update the rest of the testers.~~ (Done!)


paulmelnikow:
Build is failing for some reason, so I'm going to try to reopen this.

PR #1944, opened 2018-08-18
Fix and test trace logging in BaseJsonService

paulmelnikow:
This fixes a regression introduced in #1929 which was caught by service tests post-merge.

### DIFF
--- a/services/base-json.spec.js
+++ b/services/base-json.spec.js
@@ -70,7 +70,6 @@ describe('BaseJsonService', function() {
       sandbox.restore()
     })
     beforeEach(function() {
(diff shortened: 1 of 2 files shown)


services/base-json.spec.js
```diff
@@ -36,10 +42,7 @@ describe('BaseJsonService', function() {
       { sendAndCacheRequest },
       { handleInternalErrors: false }
     )
-    const serviceData = await serviceInstance.invokeHandler(
-      { schema: Joi.any() },
```
paulmelnikow: That shouldn't have been there; looks like a copy-paste error ^^

PR #1948, opened 2018-08-19
Sort some more badges

paulmelnikow:
- Move the Docker Pulls badge from **Other** to **Downloads**.
- Create **Platform & Language Support** for the following:
    - django version support
    - python version support
    - node version support
    - type support
    - pypi - wheel (waiting to avoid conflict with #1922)
    - pipi - format (waiting to avoid conflict with #1922)
    - pipi - implementation (waiting to avoid conflict with #1922)
    - hhvm
    - cocoapods platform

Feedback is welcomed!

Ref: #1905


services/npm/npm-type-definitions.service.js
```diff
@@ -7,7 +7,7 @@ const NpmBase = require('./npm-base')
 // package.json.
 module.exports = class NpmTypeDefinitions extends NpmBase {
   static get category() {
-    return 'version'
+    return 'platform-support'
   }
```
paulmelnikow: The diffs in the new-style services are much more readable. ✔️

tooomm:
>Thanks for contributing to our documentation. We ❤️ our documentarians!

Where did you change documentation in this PR?! I'm confused...

paulmelnikow:
Ah, good call on the PHP versions and conda platform. I’ll update this.

Agreed on PyPI too. I’m waiting on #1922 to avoid conflicts.

paulmelnikow:
Right now it's called **Platform and Language Support** with the goal of trying to pull in the npm type defs badge, which is a horse of its own special color. Maybe **Platform and Version Support** is better, and we place the npm defs badge in **Other**?

paulmelnikow:
Updated!

paulmelnikow:
Yea, that's what I was thinking too: that we'd merge this without touching PyPI, and then adjust the PyPI categorization in #1922. Thanks!

Issue #1947, opened 2018-08-19
VersionEye has sunset


paulmelnikow:
The VersionEye badge should be deprecated, as the project has sunset.

https://blog.versioneye.com/2017/10/19/versioneye-sunset-process/

PR #1953, opened 2018-08-20
[VersionEye] Service deprecation

PyvesB:
Addresses #1947.


PyvesB:
Thanks to both of you for reviewing!

PR #1957, opened 2018-08-23
Reduce boilerplate in a few more service testers [cdnjs continuousphp node shippable wercker gem]

paulmelnikow:
Follow up to #1934.


chris48s:
Looks good :+1: 

It looks like NPM downloads and Gem downloads also got overlooked in #1934 for some reason:

https://github.com/badges/shields/blob/ae190c5f073e828ec2c0bc522f5dcff37d74b7cd/services/gem/gem-downloads.tester.js#L8-L13

https://github.com/badges/shields/blob/ae190c5f073e828ec2c0bc522f5dcff37d74b7cd/services/npm/npm-downloads.tester.js#L9-L14

was that oversight, or is there something we're waiting on for those? (like with APM)

paulmelnikow:
I thought I responded to this comment, though I don't see it now. 🤔 

I updated gem-downloads. npm-downloads can't be handled using this shortcut, because that module exposes four separate services.

This should be good to merge!

Issue #1968, opened 2018-08-23
Pull request message about tests should mention the documentation


niccokunzmann:
In PR #1965, I can see the message

> This PR modified the server but none of the service tests. That's okay so long as it's refactoring existing code.

I would like to add this to make more clear how to proceed.

> Otherwise, please consider adding tests to the service: https://github.com/badges/shields/blob/master/doc/service-tests.md#readme.

PyvesB:
Seem like a sensible thing to add in my opinion!

PR #1970, opened 2018-08-23
danger: help users to write server tests

niccokunzmann:
- fixes https://github.com/badges/shields/issues/1968
- adds the desired notice

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -67,7 +67,7 @@ if (server.modified && !serviceTests.createdOrModified) {
     'This PR modified the server but none of the service tests. ',
     "That's okay so long as it's refactoring existing code. ",
     "Otherwise, please consider adding tests to the service: ",
-    "https://github.com/badges/shields/blob/master/doc/service-tests.md#readme"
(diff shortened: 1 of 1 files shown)


dangerfile.js
```diff
@@ -65,7 +65,9 @@ if (packageJson.modified && !packageLock.modified) {
 if (server.modified && !serviceTests.createdOrModified) {
   warn([
     'This PR modified the server but none of the service tests. ',
-    "That's okay so long as it's refactoring existing code.",
+    "That's okay so long as it's refactoring existing code. ",
+    "Otherwise, please consider adding tests to the service: ",
+    "https://github.com/badges/shields/blob/master/doc/service-tests.md#readme",
```
platan: Do we need `#readme` at the of the url?

niccokunzmann: It skips directly to the things you like to read. The other documentation files are not interesting. I added it for ease of use. Why would you remove it?

platan: @niccokunzmann Thanks for explanation. I didn't notice the difference earlier. I agree with you, it's useful.

platan:
@niccokunzmann thanks for your contribution!

PR #1958, opened 2018-08-23
Move legacy services from server.js into services/

paulmelnikow:
This builds on the work of #1931 by moving the legacy services into `services/`. I haven't gotten through all of them yet, though I figure I should try to wrap them up in the next day or so and get this merged to avoid conflicts.

It would be great to get #1922 merged first, though if it can't be, it's not a huge deal.

I suggest we test this through the review app, and also by running all the service tests. Probably some of them will fail because they can be flaky, though if any of them crash, well then we'll know there's a problem that needs fixing. 😀 

There's a risk that this forklifting will break something, though the risk of catching it is relatively high because we're not moving any badge examples nor tests.

PR #1983, opened 2018-08-27
Obsolete old tutorial and future TODOs

niccokunzmann:
This obsoletes the implementation part of the tutorial.
As mentioned in https://github.com/badges/shields/pull/1958#pullrequestreview-149661148, 
old style badges should be avoided and new style badges are preferred.


niccokunzmann:
I would like to replace it with a link to a pull request of a service. Is there one?

paulmelnikow:
You could link to #1920 which is a good example of a simple rewrite, and/or #1940 which is more complex, as you see fit. I don't have a good example of a brand new badge that went in using the new format, because we've tended to tweak the service design as we go.

PR #1995, opened 2018-08-28
fix [circleci] badge for projects that use workflows

chris48s:
Well this escalated in complexity quite sharply...

In an attempt to solve #1792 I've had a go at grouping related builds based on `workflow_id`. There are a number of changes in this PR which are useful to discuss 'inline', so I will mark up the diff with specific line comments, but just to give a high-level summary this is how the new approach works:

* Attempt to work out if this project uses workflows
* If it doesn't use workflows, just grab the outcome of the latest build and return that
* If it does use workflows,
  * Find the most recent workflow where all the builds with that workflow_id have finished
  * Grab all the builds with that workflow_id
  * Summarise all the build outcomes into a single outcome for the workflow

In general, this seems to work pretty well. Unfortunately there are 3 edge cases (that I can think of) which could be problematic:

* If a workflow with multiple builds is currently in progress and some builds have finished and some have not yet been started/queued, we might summarise an incomplete build. I'm not sure exactly how likely it is that we would hit this as testing it requires some quite precise timing but even if this is possible I don't think there is any way we could fix or work around it using this API.
* We've hard-coded a limit on the number of builds, but if a project has a large build matrix there could potentially be a situation (e.g: if multiple workflows currently have builds in progress and we have to summarise the third or fourth workflow back in time), we might end up summarising only part of a build because some of the builds in the wrokflow are beyond our hard-coded limit. If this did happen it would fail by producing an incorrect/unexpected value and be very hard to detect. Again, I see no sensible way to fix or work around this using this API.
* If a project has a mix of recent builds that do use workflows and older builds that don't use workflows in its history (I actually have a good example of this: https://circleci.com/api/v1.1/project/github/DemocracyClub/ec2-tag-conditional/tree/master ), we will fail with an `InvalidResponse()` until every build in the build history specified by the limit param either does or doesn't use workflows. This case fails with an error rather than unexpected output (so is less serious). Also it may be possible to account for, but I've not attempted to do so.

### DIFF
--- a/services/circleci/circleci.helpers.js
+++ b/services/circleci/circleci.helpers.js
@@ -1,0 +1,121 @@
+'use strict'
+
+const Joi = require('joi')
+
+const getWorkflowId = function(build) {
+  return build.workflows.workflow_id
+}
+
+// find all the builds with the same workflow id
+const getBuildsForWorkflow = function(builds, workflowId) {
+  return builds.filter(build => getWorkflowId(build) === workflowId)
+}
+
+// find all the builds with the same workflow id which are also complete
+const getCompleteBuildsForWorkflow = function(builds, workflowId) {
+  return builds.filter(
+    build => getWorkflowId(build) === workflowId && build.outcome !== null
+  )
+}
+
+// Find the most recent workflow which contains only complete builds
+// and return all the builds with that workflow id
+const getBuildsForLatestCompleteWorkflow = function(builds) {
+  let allBuilds, completeBuilds
+  for (let i = 0; i < builds.length; i++) {
+    allBuilds = getBuildsForWorkflow(builds, getWorkflowId(builds[i]))
+    completeBuilds = getCompleteBuildsForWorkflow(
+      builds,
+      getWorkflowId(builds[i])
+    )
+    if (allBuilds.length === completeBuilds.length) {
+      return completeBuilds
+    }
+  }
+  throw new Error('No complete workflows found')
+}
+
+const countOutcomes = function(builds) {
+  let total = 0
+  const counts = {
+    canceled: 0,
+    infrastructure_fail: 0,
+    timedout: 0,
+    failed: 0,
+    no_tests: 0,
+    success: 0,
+  }
+  for (let i = 0; i < builds.length; i++) {
+    if (!(builds[i].outcome in counts)) {
+      throw new Error('Found unexpected outcome')
+    }
+    counts[builds[i].outcome]++
+    total++
+  }
+  return { total, counts }
+}
+
+// reduce the outcomes for an array of builds to a single status
+const summarizeBuilds = function(builds) {
+  const { total, counts } = countOutcomes(builds)
+
+  if (total === counts.success) {
+    return 'passing'
+  } else if (counts.no_tests >= 1) {
+    return 'no tests'
+  } else if (counts.infrastructure_fail >= 1) {
+    return 'infrastructure fail'
+  } else if (counts.canceled >= 1) {
+    return 'canceled'
+  } else if (counts.timedout >= 1) {
+    return 'timed out'
+  } else if (counts.failed >= 1) {
+    return 'failed'
+  }
+  throw new Error('Failed to summarize build status')
+}
+
+const summarizeBuildsForLatestCompleteWorkflow = function(builds) {
+  return summarizeBuilds(getBuildsForLatestCompleteWorkflow(builds))
+}
+
+// return the status of the latest complete build
+// we need this if a project doesn't use workflows
+const getLatestCompleteBuildOutcome = function(builds) {
+  for (let i = 0; i < builds.length; i++) {
+    if (builds[i].outcome != null) {
+      return summarizeBuilds([builds[i]])
+    }
+  }
+  throw new Error('No complete builds found')
+}
+
+const populatedArraySchema = Joi.array()
+  .items(
+    Joi.object({
+      // if we have >0 items in our array, every object must have an 'outcome' key
+      outcome: Joi.string()
+        .allow(null)
+        .required(),
+
+      // 'workflows' key is optional - not all projects have workflows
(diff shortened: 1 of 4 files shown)


services/circleci/circleci.helpers.spec.js
```diff
@@ -0,0 +1,177 @@
...
+        // this response doesn't have workflows
+        { outcome: 'failed' },
+        { outcome: 'success' },
+      ])
+    ).to.throw(Error, "Cannot read property 'workflow_id' of undefined")
+  })
+})
+
+describe('circleci: schema validation', function() {
+  const validate = function(data, schema) {
```
chris48s: This may not be the best place to get bogged down in it, but it seems like writing tests for non-trivial validators (like this one) is something we might want to do sometimes. I definitely found I needed to write tests to get this one correct. Lets have a think about a good pattern for this.

services/circleci/circleci.helpers.js
```diff
@@ -0,0 +1,121 @@
...
+      }),
+    }).required()
+  )
+  .min(1)
+  .max(100)
+  .required()
+const emptyArraySchema = Joi.array()
+  .min(0)
+  .max(0)
+  .required() // [] is also a valid response from Circle CI
```
chris48s: Turns out sometimes `[]` is a legit response example: https://circleci.com/api/v1.1/project/github/chris48s/shields/tree/issue1827

services/circleci/circleci.service.js
```diff
@@ -1,21 +1,20 @@
...
+} = require('./circleci.helpers.js')
 
 module.exports = class CircleCi extends BaseJsonService {
   async fetch({ token, vcsType, userRepo, branch }) {
     let url = `https://circleci.com/api/v1.1/project/${vcsType}/${userRepo}`
     if (branch != null) {
       url += `/tree/${branch}`
     }
-    const query = { filter: 'completed', limit: 1 }
+    const query = { limit: 50 }
```
chris48s: We can't filter on 'completed' any more here. We need to know about in-progress/queued builds so that if the first workflow we come across has some builds which are not complete we know to ignore that one and move on to the next workflow to try and find one where all the builds are complete. If we filter on complete here, sometimes we will summarise an incomplete workflow.

In terms of limit, I've picked 50 to balance 2 properties:
* We need a fairly large number of records so that on a project with a large build matrix where we can't take the first `workflow_id` (e.g: because some builds are still in progress) we've got enough records
* If we fetch _loads_ of records the badge takes a long time to render because we must wait a long time for the `fetch()` stage

services/circleci/circleci.helpers.js
```diff
@@ -0,0 +1,121 @@
...
+  } else if (counts.no_tests >= 1) {
+    return 'no tests'
+  } else if (counts.infrastructure_fail >= 1) {
+    return 'infrastructure fail'
+  } else if (counts.canceled >= 1) {
+    return 'canceled'
+  } else if (counts.timedout >= 1) {
+    return 'timed out'
+  } else if (counts.failed >= 1) {
+    return 'failed'
```
chris48s: I've taken a stab at a sensible order here, but I'm not committed to it. If you're reading this thinking "'canceled' obviously takes precedence over 'infrastructure fail'", feel free to make suggestions and I'll change it.

I've also switched to using 'outcome' here instead of status (outcome only applies to finished builds and we only want to summarise workflows where all the builds are finished). This also gives as smaller number of values to try and make sense of.

services/circleci/circleci.service.js
```diff
@@ -28,20 +27,34 @@ module.exports = class CircleCi extends BaseJsonService {
...
+    if (status === 'passing') {
+      color = 'brightgreen'
+    } else if (
+      ['failed', 'infrastructure fail', 'canceled', 'timed out'].includes(
+        status
+      )
+    ) {
+      color = 'red'
+    } else if (status === 'no tests') {
+      color = 'yellow'
```
chris48s: Same here. I've attempted to do something sensible-ish, but again if you're thinking "'canceled' should obviously be grey" or whatever, feel free to make suggestions and I'll change them. Once we've got this nailed I'll also add tests for the colour logic. There is relatively little consistency about this on other services. Prior art to maybe consider:
* https://github.com/amio/badgen-service/blob/master/libs/live-fns/circleci.js
* https://github.com/circleci/frontend/blob/c189f3546afe49b64c8ee86d92ff67ed9d2eda78/src-cljs/frontend/models/build.cljs#L174-L181
(although they are both looking at 'status' as opposed to 'outcome')

services/circleci/circleci.tester.js
```diff
@@ -38,25 +47,18 @@ t.create('circle ci (no response data)')
...
-      .reply(200, [{ status: 'success' }, { status: 'fixed' }])
+      .get('/api/v1.1/project/github/RedSparr0w/node-csgo-parser?limit=50')
+      .reply(200, [])
   )
-  .expectJSON({
-    name: 'build',
-    value: 'invalid json response',
-    colorB: '#9f9f9f',
-  })
+  .expectJSON({ name: 'build', value: 'could not summarize build status' })
```
chris48s: I've implemented a lot of the special-case testing for this badge logic in `circleci.helpers.spec.js` instead of using a lot of mocked responses here. I've kept the service tests pretty high-level

chris48s:
Based on chat in discord, I've updated this PR to use a more functional style in a couple more places in b6cda72 .

I think I'm going to leave `getBuildsForLatestCompleteWorkflow()` where it is. I find the mutable state approach easier to read here than the proposed `uniq(builds.map(getWorkflowId)).find(workflowId => builds.filter(…).every(…)))` alternative, although 'readable' is a matter of taste :)

Using `lodash.countby` in `countOutcomes()` is probably the big readabiliy gain, particularly as refactoring that function has also prompted me to move more validation logic into the schema - which is the correct place for it :+1:

PR #1996, opened 2018-08-28
Improved deployment bot message

PyvesB:
As discussed with @paulmelnikow on Discord. We still need to make the relevant channel public (or create a new one).


PyvesB:
Thanks for reviewing!

PyvesB:
Thanks for reviewing. We can already see the updated message in action above! 😄

PR #1998, opened 2018-08-29
Use render helpers in existing new-style services [apm cdnjs gemversion]

paulmelnikow:
This refactors the render methods in these services to use the new helper functions.


paulmelnikow:
Thanks for reviewing!

PR #1999, opened 2018-08-29
Enforce no extensions in require

paulmelnikow:
This removes remaining `.js` extensions from `require()` expressions and automatically enforces this rule.


paulmelnikow:
And another one! 😝

Issue #1309, opened 2017-12-01
Stars in badges broken (staging server misses font)


tooomm:
- on live page: `working`
![stars_live](https://user-images.githubusercontent.com/9874850/33507242-f81e3d84-d6f3-11e7-9302-6051c4bdcc56.png)

- on Heroku review app for these two staging links: `unreadable`
    - https://shields-staging.herokuapp.com/
    - https://shields-staging-pr-1305.herokuapp.com/
![stars_staging](https://user-images.githubusercontent.com/9874850/33507328-7041af4e-d6f4-11e7-8159-b42602333711.png)

paulmelnikow:
This is a problem with the staging server, which does not have the correct font installed. As a result the text width computation is off. The stars do look especially bad. If you look closely, nearly all the badges are a bit off.

Agree it would be nice to fix this.

tooomm:
>If you look closely, nearly all the badges are a bit off (on staging server).

This is hard to tell since all badges on live shields.io page look a bit off on Firefox anyway. :)

espadrine:
The right font is from the Mac's Verdana TTF. Unfortunately, for licensing issues, we cannot transmit it.

paulmelnikow:
I'm delighted to say this was fixed in #2311!

PR #2000, opened 2018-08-29
[AppVeyorTests] compact format and custom labels

paulmelnikow:
This picks up the work from #1321 on top of the work from #1940.

It allows the user to choose a compact tests format, or to override the labels with their own text or symbols.

(Will add some examples here once the review app is deployed.)

Oops, I forgot; the font width problem in staging (#1309) makes these look really bad.

Here's a screenshot instead:

screen shot 2018-08-29 at 12 26 55 pm


paulmelnikow:
Thanks for reviewing!

PR #1322, opened 2017-12-04
Fix and refactor front-end URL generation

paulmelnikow:
- Fix issue in Firefox 57 when run from static build
- Fix color parameter in dynamic badge maker
- Correctly apply maxAge in usage badges
- Follow the WHATWG lead and begin to standardize on URL, not URI (https://url.spec.whatwg.org/#goals)

Firefox issue reported at https://github.com/badges/shields/pull/1305#issuecomment-348643834 and mentioned in #1311.

PR #2006, opened 2018-08-29
Update uri -> url in the front end + examples

paulmelnikow:
This continues a consistency update we’ve been making to standardize on URL. This also helps with copying and pasting between all-badge-examples and new-style services, where it’s otherwise easy to make a mistake.


paulmelnikow:
Thanks for reviewing. Yea. There's also this recommendation from the WHATWG to standardize on "URL": https://url.spec.whatwg.org/#goals

Ref: #1322 #1341

PR #2018, opened 2018-08-29
Rewrite deprecated services and add tests [bithound cauditor dotnetstatus gemnasium gratipay magnumci snap-ci versioneye]

paulmelnikow:
I've rewritten the deprecated services using the `deprecatedService` helper from #1922. I added a test for `Deprecated`, and for `enforceDeprecation`, which isn't being used right now, but is there for future use.

This also makes it possible to write services using BaseService which do not have any named parameters (with a test).

Ref: #1358


paulmelnikow:
Ah, yes, I was running the tests locally but forgot to add them to the PR. 👍 

I'm of two minds about the tests. I modified the regexes because the legacy format is different and so I didn't have to name all the capture groups. The tests are a good way to make sure the patterns still work as intended. On the other hand, I agree it seems a little silly to maintain them.

We could write a contract for a deprecated badge that looks for **no longer available**. That would help keep the test code to a minimum.

chris48s:
> I modified the regexes because the legacy format is different and so I didn't have to name all the capture groups. The tests are a good way to make sure the patterns still work as intended

If the tests are performing a useful check, lets keep them in place.

> We could write a contract for a deprecated badge that looks for no longer available. That would help keep the test code to a minimum.

Thinking it through a bit, it is relatively uncommon that we deprecate. I think we have maybe 10 deprecated services. Lets not put a load of effort into it now, but if we find repeated test boilerplate for deprecated services becomes a chore in future we could revisit this.

paulmelnikow:
Thanks for the comments!

PR #2020, opened 2018-08-29
update logo submission docs

chris48s:
It is reasonably common for us to close a logo submission, advise a user to pass a base64-encoded logo instead and provide an example. Maybe this documentation change will help.


paulmelnikow:
I pushed a tweak; please take a look!

chris48s:
yep - those changes are helpful - thanks :+1:

PR #2021, opened 2018-08-29
Rework GitHub acceptor and move to its own module

paulmelnikow:
Continuing to merge the work from #1205.


paulmelnikow:
This really should have a test. I'm eager to get it in.

services/github/auth/acceptor.js
```diff
@@ -0,0 +1,133 @@
...
+          })
+        })
+    )
+  )
+}
+
+function setRoutes(server) {
+  const baseUrl = process.env.BASE_URL || 'https://img.shields.io'
+
+  server.route(/^\/github-auth$/, (data, match, end, ask) => {
```
paulmelnikow: Added a test.

services/github/auth/acceptor.js
```diff
@@ -0,0 +1,133 @@
...
+      redirect_uri: `${baseUrl}/github-auth/done`,
+    })
+    ask.res.setHeader(
+      'Location',
+      `https://github.com/login/oauth/authorize?${query}`
+    )
+    end('')
+  })
+
+  server.route(/^\/github-auth\/done$/, (data, match, end, ask) => {
```
paulmelnikow: Added a test.

paulmelnikow:
Thanks for reviewing! Yea, not sure why package-lock didn't get updated. I will do that now.

paulmelnikow:
Yea, it's a consistent problem with dependabot. The churn occurs every time a package-lock is written by the npm CLI after it's been written by dependabot, or vice versa. I'm not sure we've reported it. It sure would be nice to fix because it makes it basically impossible to review the changes in that file.

chris48s:
The lockfile optional thing is a reported issue on dependabot https://github.com/dependabot/feedback/issues/197 but required an upstream fix: https://github.com/npm/cli/pull/76

As I understand it, once we (and dependabot) are updated to a version on npm with that patch, this should stop happening.

Issue #2028, opened 2018-08-30
Consistent use of async/await


paulmelnikow:
At this point, our code is about 70% callbacks, 10% promises, and 20% async / await.

We're moving to async/await, though this has happened in phases:

1. As the legacy services are rewritten, they are being converted from callbacks (and promises here and there) to async/await. This is being tracked in #1358, though we should make it clear that the rewritten services should not use promises.
2. Scoutcamp runs on callbacks, and we should stick with callbacks for the code that interacts with it directly. (Scoutcamp has served us well, though it isn't being actively maintained, and, once the legacy services are migrated, we might want to evaluate other frameworks.)
3. Promise and callback-based code that _doesn't_ talk to Scoutcamp should be updated to use async/await.
4. We should document that async/await is the preferred paradigm. We'll have lots of good examples of async/await code on hand, and perhaps we should place comments in the legacy code, with an explanation or a link explaining that new code should not be written in that style.

The reason to invest in consistency is not because the old code is hard to read, but because having good examples in place makes us less likely to ask contributors to "please fix this" and "please update that." This leads to a more positive contributor experience. (Also mentioned in https://github.com/badges/shields/issues/2026#issuecomment-417495839.)

paulmelnikow:
@PyvesB Could you address this when you address this task?

> Move request method to super class and delete BaseHttp. I'm happy to tackle this one in the coming days as it's related to another piece of work I'm doing.

from https://github.com/badges/shields/pull/1965#pullrequestreview-151198384

PR #2029, opened 2018-08-31
Rework [suggest] code using async/await

paulmelnikow:
In #2028 I suggested that we update as much of the code as possible to make consistent use of async await. This snags a bunch of the utility code and attempts to do that.

Unit tests are passing; I'll take a pass through and identify a list of services to test along with these.


chris48s:
I've had a quick look over this. In general :+1: for moving towards a more readable style.

`svg-to-img` has its own tests and even though we don't have explicit tests for `getPhpReleases()` I think its fine that this is transitively covered by the travis and php-eye tests. That gives us a reasonable level of confidence in this.

I'm slightly nervous about how much 'suggestion and search' code is being refactored here which is not under test and I find it slightly difficult to assert that we're not introducing any regression here. I've made a staging deploy at https://shields-staging-pr-2029.herokuapp.com . The search code seems fine. The 'suggest' feature is throwing `An unexpected error has occurred.` for all queries, but then I looked at some other recent staging deploys and they seem to do the same thing. It may not be related to the changes in this PR, but this feature is working in production.

paulmelnikow:
Huh, there are some tests of suggest, which are written as service tests (not sure that’s the best, but that’s where they are). IIRC they have been failing for a while. Production was updated Saturday so I’m not sure why one is working and one not. Github auth perhaps? Let’s make sure this is working before merge.

If anyone wants to pick this up, you’re welcome to.

chris48s:
Aah right, nice. I hadn't realised that - I was looking for a `.spec.js`. Looking at some other recent staging deploys they seem to do the same thing so it may be a false positive rather than a consequence of refactoring (I must admit, the 'suggest' feature is something I rarely use). I'll try work out what is going on there and provide some more useful information on that.

Having actually run the tests now, it seems there's a failure with the php-eye code. I think the service tests 'passed' previously because CircleCI decided this pull request isn't a pull request and didn't actually run anything :(

paulmelnikow:
How about I cherry-pick the bits that are unrelated to suggest, and then we circle back to that?

paulmelnikow:
These changes do seem to be covered by the tests in https://github.com/badges/shields/blob/master/services/suggest/suggest.tester.js.

When I paste e.g. https://example.com/ or https://github.com/badges/shields into the suggest box at https://shields-staging-pr-2029.herokuapp.com/#/ and click **Suggest Badges**, I see what I expect:

screen shot 2018-10-30 at 11 05 17 pm

screen shot 2018-10-30 at 11 05 26 pm

So I think this might be ready after all. Do you want to take another look @chris48s?

chris48s:
I had another look at this and I'm still seeing the same thing on staging with firefox:

gifrecord_2018-10-31_191626

but I tried it in Chromium and it worked. This does work in firefox on prod. I'm seeing the same thing on other staging deploys though so I suspect its unrelated to this change. I think I've just stumbled upon this problem for the first time by coincidence because I've never attempted to use this feature on a staging deploy before.

It looks like it is `next.js` throwing the error which probably explains the discrepancy between staging and prod?

paulmelnikow:
Gotcha. Thanks! I opened #2245 to track that.

Issue #1985, opened 2018-08-27
Codacy badges broken


PyvesB:
Codacy have updated their own badges and we are no longer able to retrieve any relevant information using our `fetchFromSvg` function. As our result, Shields's badges are no longer working properly:

![](https://img.shields.io/codacy/grade/e27821fb6289410b8f58338c7e0bc686.svg)
![](https://img.shields.io/codacy/coverage/e27821fb6289410b8f58338c7e0bc686.svg)

They do have a proper API documented [here](https://api.codacy.com/swagger), but an API token is required. When none is provided, requests are rejected, even for public projects:

> {"error":"not found"}

We may need to contact support to get this sorted.

nemchik:
I have reached out to Codacy support to make them aware of the issue. I will report back with any relevant responses.

nemchik:
Received a response from Codacy support. They have added the information required back into the badges. It seems like previously they only included two text fields in the SVG but now they include 3, so the index of the field needed for shields has increased by one for both badges.

PyvesB:
This is great, thanks for keeping us updated!

The way forward is probably something along these lines:
* add tests to our Codacy service implementation. This will allow us to maintain it more easily in the future.
* extend `fetchFromSvg` to allow specifying a match index other than 1.
* make sure we keep an eye on #2031 to avoid any clashes with the ongoing work around some other SVG services.

I'm happy to look into this over the week-end, but if someone wants to help out and submit a pull request before that, please feel free to do so (simply make some noise here so that we don't both start working on the same thing). 😉

PyvesB:
Okay, a proposed fix has now been merged into the `master` branch and will likely be deployed in the coming days. 👍

PyvesB:
All good, we are live since yesterday:
![](https://img.shields.io/codacy/grade/e27821fb6289410b8f58338c7e0bc686.svg)
![](https://img.shields.io/codacy/coverage/59d607d0e311408885e418004068ea58.svg)

We're doing some more work to further improve the user experience with our Codacy badges, see #2102 for more information. 😉

PR #1565, opened 2018-03-12
Adopt async-await for most promise-based code

paulmelnikow:
A big advantage of Node 8 is support for async–await. Our service code will particularly benefit from this, but so does the rest of our async code.

It's remarkable how much clearer these look!

PR #2031, opened 2018-08-31
[WIP] Rewrite [vso] badges

paulmelnikow:
This tackles two of the services using `fetchFromSvg`. The others are in codacy, which are broken right now; see #1985.

There’s a design decision to make here about how to structure the shared code. One is with a subclass, and the other with a function.

After working on a few service families, I noticed a pattern. First I could code the first service. Then I would start to code the second, and  begin to separate out the shared code into an abstract superclass.

This involved a lot of typing:

- `require` statements would be moved into the module containing the base class
- new `require` statements would be added to the module with the subclass
- The `extends` line would change
- All the shared code would move

The current arrangement also produced base service code which couldn’t be tested without creating a subclass, which has meant that tests of `fetch` end up being written as service tests, though probably they would be more clearly expressed as unit tests with intercepted http calls.

I also dislike deep inheritance hierarchies because they make it harder for new developers to understand what is happening. They are also difficult to change, on account of all the implicit dependencies.

It would be inconceivable to create a base class for a single service, based on the possibility of writing more services someday. This means going from n=1 to 2 is services per family is (relatively) hard, while going from n >=2 to n+1 is easy.

This led me to want to experiment with composition; hence taking the “function” approach here.

Ref: #1358 #1565

### DIFF
--- a/services/test-validators.js
+++ b/services/test-validators.js
@@ -96,6 +96,7 @@ const isBuildStatus = Joi.equal(
   'failing',
   'no tests',
   'not built',
+  'never built',
   'not run',
   'passing',
   'pending',
--- a/lib/all-badge-examples.js
+++ b/lib/all-badge-examples.js
@@ -2,28 +2,6 @@
 
 const { loadServiceClasses } = require('../services')
 
-const visualStudioTeamServicesDoc = `
-<p>
-  To obtain your own badge, you will first need to enable badges for your
-  project:
-</p>
-<img
-  src="https://cloud.githubusercontent.com/assets/6189336/11894616/be744ab4-a578-11e5-9e44-0c32a7836b3b.png"
-  alt="Go to your builds, click General, then check Badge enabled." />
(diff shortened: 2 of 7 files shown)


services/test-validators.js
```diff
@@ -96,6 +96,7 @@ const isBuildStatus = Joi.equal(
   'failing',
   'no tests',
   'not built',
+  'never built',
```
paulmelnikow: This message is returned from vso.

From production:

https://img.shields.io/vso/build/larsbrinkhoff/953a34b9-5966-4923-a48a-c41874cfb5f5/1.svg

paulmelnikow:
In short: totally agreed this is the right solution to the wrong problem. I'll circle back after #2037 is merged, respond in detail, and update this.

paulmelnikow:
The VSO badge has changed, and the rest of this is in #2229, so I'm going to close this and start a new PR.

PR #1965, opened 2018-08-23
new badge: f-droid 

niccokunzmann:
I document the progress here as an imperfect PR and ask for review once I am confident it works.

- edit server.js
- first untested version
  - get the website
  - regex query as string
  - get version info

fixes #1963

PR #2037, opened 2018-09-01
Delete BaseHTTPService and implement new BaseXmlService (affects [eclipse-marketplace f-droid]; also testing on [uptimerobot circleci])

PyvesB:
There's quite a few things going on here:
- as a follow up to https://github.com/badges/shields/pull/1965#discussion_r213104483, _BaseHTTPService_ was removed by pushing up the `_request` function to the _BaseService_ class. Tests were added for this method.
- following a similar reasoning, `_validate` was also pushed up to _BaseService_, as this allowed to easily implement _BaseXmlService_ alongside the existing _BaseJsonService_ class with very little code duplication.
- `asJson` was pulled out of _error-helpers.js_ and now sits in _response-parsers.js_ with a new friend, `asXml`. These functions are now properly tested.
- the _eclipse-markteplace_ service was broken down and migrated to use the new _BaseXmlService_ class.
- calls were migrated to async/await were relevant, following https://github.com/badges/shields/issues/2028#issuecomment-417499726.

I'll launch a Heroku review app so we can play around with the migrated Eclipse badges and make sure we haven't missed anything! 👍


lib/response-parsers.js
```diff
@@ -0,0 +1,34 @@
...
+    throw new InvalidResponse({
+      prettyMessage: 'unparseable json response',
+      underlyingError: err,
+    })
+  }
+}
+
+async function asXml({ buffer, res }) {
+  let xml
+  xml2js.parseString(buffer, (err, parsedData) => {
```
PyvesB: This uses a callback and works fine because the underlying implementation is not async (see https://github.com/Leonidas-from-XIV/node-xml2js/issues/159#issuecomment-248599477), but there is probably a better way of doing this. Any ideas/guidance?

services/eclipse-marketplace/eclipse-marketplace-downloads.service.js
```diff
@@ -0,0 +1,93 @@
...
+    static render({ downloads }) {
+      return {
+        message: `${metric(downloads)}${messageSuffix}`,
+        color: downloadCountColor(downloads),
+      }
+    }
+
+    async handle({ name }) {
+      const { marketplace } = await this.fetch({ name, schema })
+      const downloads = base.endsWith('dt')
```
paulmelnikow: It might be slightly clearer to condition on `interval` here.

services/eclipse-marketplace/eclipse-marketplace-downloads.tester.js
```diff
@@ -0,0 +1,38 @@
+'use strict'
```
paulmelnikow: Were there any changes in the tests, or were they just moved?

services/base.js
```diff
@@ -306,6 +308,45 @@ class BaseService {
...
+      trace.logTrace(
+        'validate',
+        emojic.bathtub,
+        'Data after validation',
+        value,
+        { deep: true }
+      )
+      return value
+    }
+  }
```
paulmelnikow: I love the idea of moving `_validate` here. Nice work!

services/base.js
```diff
@@ -306,6 +308,45 @@ class BaseService {
...
+
+  async _request({ url, options = {}, errorMessages = {} }) {
+    const logTrace = (...args) => trace.logTrace('fetch', ...args)
+    logTrace(emojic.bowAndArrow, 'Request', url, '\n', options)
+    const { res, buffer } = await this._sendAndCacheRequest(url, options)
+    logTrace(emojic.dart, 'Response status code', res.statusCode)
+    return checkErrorResponse.asPromise(errorMessages)({ buffer, res })
```
paulmelnikow: AFAICT the only difference between json and xml is the parsing – is that right?

Another way we could consider handling the relationship between `_request` and `_validate` is to have `_request` delegate to a `_parse` function, and put the format-specific bits there. The service code would invoke `_request` which would be BaseService's implementation, and that would in turn invoke `_parse` which would be BaseXxService's.

This would force me to change slightly the way I handled #2031 because every service would then need a `_parse` implementation. It would also generally force everything to be a native object that can be checked by a schema. However I don't see a problem with that. We could make a BaseSvgBadge subclass easily enough. And encouraging schema validation is good.

I like this idea because the `_validate` function is in BaseService so it would feel natural to call it from its `_request` function. The logging code probably could be moved from BaseXxService to BaseService as well.

services/eclipse-marketplace/eclipse-marketplace-downloads.tester.js
PyvesB: The five existing tests we copy-pasted as is in the different tester classes. Additionally, I added one "not found" test case for each implementation.

services/base.js
PyvesB: The requests are slightly different as weel, in particular the `Accept` headers. If we go with your suggestion, we'll also have to add a `get defaultHeaders` function or something like that, which we would override in the implementations.

paulmelnikow: Ah, right, the request options, that's a good point. The headers could also be added in a `_request` override. Though the differences in the requests may mean it's clearer to handle this the way it is.

services/base.js
```diff
@@ -306,6 +308,45 @@ class BaseService {
...
+    })
+    if (error) {
+      trace.logTrace(
+        'validate',
+        emojic.womanShrugging,
+        'Response did not match schema',
+        error.message
+      )
+      throw new InvalidResponse({
+        prettyMessage: 'invalid data response',
```
paulmelnikow: How about **invalid response data**?

lib/response-parsers.js
PyvesB: I tied solution 1, switching to `fast-xml-parser`.
Good points:
* cleaner code in _response-parser.js_.
* `xml2js` has a tendency of introducing arrays everywhere even when not necessary. Switching has simplified the Joi schemas and the constructed JS objects.
Bad point:
* no helpful exception message if the XML is not parsable.

I ran `npm i fast-xml-parser` to update _package-lock.json_, is this the correct approach?

lib/response-parsers.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
+
+const { InvalidResponse } = require('../services/errors')
+const xml2js = require('xml2js')
```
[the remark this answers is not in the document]
PyvesB: With `fast-xml-parser`, the parser is now used in two places, so keeping the `require` up there. 😄

package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "^7.0.0"
+        "@babel/highlight": "7.0.0"
```
paulmelnikow: This package-lock diff is unusually large. Could you make sure you're on the latest npm, reset this file, and then run `npm install` again to see if that tightens it up?

lib/response-parsers.js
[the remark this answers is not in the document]
paulmelnikow: Ah, sorry, I meant the functions themselves, not the `require`s. Instead of having these in a separate module, could they just be inlined into the `_requestX` methods in the base services?

package-lock.json
PyvesB: Yep, updating to the latest version of `npm` seems to have helped a lot. 😉

PyvesB:
Indeed, I was getting confused with the return types, thanks for the help. This should now be ready to go! 👍

Issue #1984, opened 2018-08-27
Update tutorial for new service architecture


chris48s:
Tutorial should cover how to write a service that extends `BaseService`/`BaseJsonService`

paulmelnikow:
@niccokunzmann who wrote the original tutorial has offered to help with this!

Any thoughts on how the work could be divided?

chris48s:
OK. Having a quick scan over the docs, the testing docs also need some work. Maybe as a starting point, one of us can take the tutorial and the other one can do the service test docs? Code review provides an opportunity to check we're covering everything. @niccokunzmann - are you most keen to work on the tutorial? The perspective of someone who is a bit further removed might be really valuable here. If so I will take the test docs.

The most important first step is to get the tutorial to a stage where the docs reflect reality. Maybe we don't have to cover all of these for an opening gambit, but I do have a few topics in mind that we should cover in the docs:

* `fetch()`/`render()` pattern
* Contract by design, writing a validator
* When to use static/instance methods
* Available helper functions in `/lib` (maybe this is its own docs section to cover properly, but we should touch on this in the tutorial)
* Inheritance/service 'families'

do you have anything else in mind that should go on that list @paulmelnikow ?

I'll probably also take the opportunity to read over the rest of the docs and see if there's anything else that needs a 'refresh' :)

niccokunzmann:
From my perspective there is not "one" tutorial. We have these service families:
- JSON
- raw data
- in the future possibly also XML

So, the tutorial would best cover what is most needed and common to all of them first. Then, there are differences, each getting a sub-chapter. I like complete code examples that would work if you copy them in. Is there a service which would be good to take as an example from your point of view?

niccokunzmann:
> Just to be really clear on this: are you going to work on |TUTORIAL.md| ?

As long as I do not create a pull request, consider it a no. I can not
truthfully make a promise to you as I do not know you enough.
I am undecided at the moment and cannot tell for the coming days.
If you write something, I may still come in again, so do not wait for me.

chris48s:
OK cheers. I will aim to work on updating the tutorial myself in the next few days. Feel free to contribute your thoughts either in review or at a later date. Documentation can always be extended and improved over time.

paulmelnikow:
Two things come to mind:

- An overview of the design: in a few sentences, how does the service work? How does a service get registered, executed, tested? How is the frontend generated? Try to take the magic out of it.
- Advice on porting existing services, and on refactoring

[...]

PR #1582, opened 2018-03-19
Extract examples from new-style services

paulmelnikow:
Instead of centralizing examples, specify them from within a service.

PR #2042, opened 2018-09-01
Update tutorial to reflect new service layout

chris48s:
Refs #1984

To start off with, I've updated the tutorial to walk the user through writing a service class that extends `BaseJsonService` instead of defining a badge in `server.js` using the procedural style. This doesn't complete the documentation refresh - there are still other topics we need to cover/update - but I think this represents a chunk of work which is useful to review.

Next steps:

* Update/refresh the service testing docs
* Walk though defining a service family with an abstract parent class and multiple badge implementations extending it
* Add docs on how to port legacy services

(probably in that order)


doc/TUTORIAL.md
```diff
@@ -54,177 +46,197 @@ This information allows other humans to help and build on your work.
...
-
-Once you are done implementing your badge, you can add it to the collection on [shields.io](http://shields.io/).
-
-First, we make it visible locally at [http://[::1]:8080/][home]. Edit
-[lib/all-badge-examples.js][allbadges] in the right category (Build,
-Downloads, ...) and add your badge:
+Description of the code:
+1. As with the first example, we declare strict mode at the start of each file.
+2. Our badge will query a JSON API so we will extend `BaseJsonService` instead of `BaseService`. This contains some helpers to reduce the need for boilerplate when calling a JSON API.
+3. In this case we are making a version badge, which is a common patten. Instead of directly returning an object in this badge we will use a helper function to format our data consistently. There are a variety of helper functions to help with common tasks in `/lib`.
```
chris48s: I'd love to be able to link to some useful docs here, but we don't have any (yet). In lieu of a proper internal API reference, is there something more useful we can say here than "there's a bunch of stuff in `/lib` - good luck finding it ;)"

doc/TUTORIAL.md
```diff
@@ -54,177 +46,197 @@ This information allows other humans to help and build on your work.
...
+8. Our bage must implement the `async handle()` function. Because our URL pattern captures a variable called `gem`, our function signature is `async handle({ gem })`. We usually seperate the process of generating a badge into 2 stages or concerns: fetch and render. The `fetch()` function is responsible for calling an API endpoint to get data. The `render()` function formats the data for display. In a case where there is a lot of calculation or intermediate steps, this pattern may be thought of as fetch, transform, render and it might be necessary to define some helper functions to assist with the 'transform' step.
+9. The `async fetch()` method is responsible for calling an API endpoint to get data. Extending `BaseJsonService` gives us the helper function `_requestJson()`. Note here that we pass the schema we defined in step 4 as an argument. `_requestJson()` will deal with validating the response against the schema and throwing an error if necessary.
```
chris48s: I know we can also pass headers, query string etc here and I'd like to explain this, but there are so many layers of abstraction between here and [request](https://github.com/request/request) I struggled to work out how to clearly explain how we're wrapping it or what subset of the `request` docs are applicable. Any pointers on how to explain this well?

doc/TUTORIAL.md
```diff
@@ -54,177 +46,197 @@ This information allows other humans to help and build on your work.
...
+
+}
 ```
 
-Save, run `npm run build`, and you can see it [locally][home].
+1. The `category()` property defines which heading in the index our example will appear under.
+2. The examples property defines an array of examples. In this case the array will contain a single object, but in some cases it is helpful to provide multiple usage examples.
+3. Our example object should contain the following properties:
+    * `title`: Descriptive text that will be shown next to the badge
+    * `urlPattern`: Describe the variable part of the URL using `:param` syntax.
```
chris48s: Tangent:

For the moment, lets just document this but I wonder if we could actually find a way to generate this from the url regex and capture vars instead of requiring the user to manually define it. I am reminded of this conversation https://github.com/badges/shields/pull/1582#discussion_r176912366

doc/TUTORIAL.md
paulmelnikow: I'd suggest we survey the few functions that the existing new-style services are using, and list them. You mention renderVersionBadge but we could add e.g. renderLicenseBadge, metric, and whatever else the new-style badges are using now.

doc/TUTORIAL.md
paulmelnikow: It would be great to reduce the layers of abstraction. Though, meanwhile, we could say these things:

1. `_requestJson` automatically adds an `Accept` header, checks the status code, parses the response as JSON, and returns the parsed response.

2. Options can be passed to `request`, including method, query string, and headers. If headers are provided they will override the ones automatically set by `_requestJson`. There is no need to specify `json`, as the JSON parsing is handled by `_requestJson`. These are all the supported options: https://github.com/request/request#requestoptions-callback

3. Error messages corresponding to each status code can be returned by passing a dictionary of status codes -> messages in `errorMessages`.

doc/TUTORIAL.md
paulmelnikow: Yea, I think the solution may be to do this in the opposite direction: we define the `urlPattern` and automatically generate the regex using [url-pattern](https://github.com/snd/url-pattern).

paulmelnikow:
Chris, thanks for taking this on! I responded to your comments though I didn't read it through fully.

I'll be offline for a few days starting tomorrow, so feel free to merge this, and maybe I'll do a set of follow-up edits when I'm back online.

doc/TUTORIAL.md
```diff
@@ -3,48 +3,40 @@ Tutorial on how to add a badge for a service
...
-- [implementations and their commits][blame] in the [server.js][server] file.
-- [merged pull-requests][mergedpr].
+You can also read previous
+[merged pull-requests with the 'service-badge' label](https://github.com/badges/shields/pulls?utf8=%E2%9C%93&q=is%3Apr+label%3Aservice-badge+is%3Amerged)
+to see how other people implemented their badges.
 
 (2) Setup
 ---------
 
 I suppose you have [git](https://git-scm.com/) installed.
```
PyvesB: This is the last remaining bit using first person, could we change it?

chris48s: Tbh, there is probably a bit of a mashup of styles in the document at this point. The previous tutorial was mostly written in a more "instructive" tone e.g: "you should do this to your code". With the stuff I've added/edited I've mostly gone with a more "inclusive" tone e.g: "we will do this to our code". This is a quite a subtle language style point but it is probably useful to standardise one way or the other. Do you prefer the we/our style?

chris48s:
Thanks for the comments on this @paulmelnikow and @PyvesB .

As well as the comments provided here @niccokunzmann has submitted [this PR](https://github.com/chris48s/shields/pull/2) to my fork proposing some changes which..vary in scope and I haven't properly read yet.

I'll do another pass over this taking into account the suggestions made in both PRs, resolve the conflicts and update this.

doc/TUTORIAL.md
PyvesB: I don't mind whether it's "you" or "we", but I think we should avoid "I" as it sounds as if one authority is dictating the tutorial.

niccokunzmann:
Good idea. You can change it. We will avoid it.

chris48s:
I've pushed some additional edits addressing the various bits of feedback. Thanks, all. @PyvesB maybe you could give this a second look. Cheers

Issue #1570, opened 2018-03-13
Support Visual Studio Team Service Release Badge


iKaew:
VSTS just support release badge. It would be great if we can get it from here as same as build status. 
https://docs.microsoft.com/en-us/vsts/release-notes/2018/mar-05-vsts#share-deployment-status-using-a-badge

paulmelnikow:
Hi! Thanks for the suggestion. Would you be interested in working on this? We have [a great tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md).

Thilas:
I'm interested in such release badges and I have started to work on something.

paulmelnikow:
Thanks @Thilas! Seems like this can be closed now. If I've misunderstood, please post a comment and someone will reopen.

Thilas:
I confirm, everything seems good to me.

PR #2049, opened 2018-09-03
[vso] Add support for VSTS releases

Thilas:
Per #1570, here is the support for VSTS releases. I also added the support for named branches on VSTS builds.


Thilas:
Hi @PyvesB,

Thanks for your code review. Your remarks make totally sense. Actually I already was working on adding some tests here. Hope it is good to go now.

Let me know, I am eager to use this on my VSTS projects ;)

services/vso/vso.tester.js
```diff
@@ -0,0 +1,84 @@
...
+  .get('/build/foo/bar/foobar.json')
+  .networkOff()
+  .expectJSON({ name: 'build', value: 'inaccessible' })
+
+// Releases
+
+t.create('release status is succeeded')
+  .get('/release/devdiv/0bdbc590-a062-4c3f-b0f6-9383f67865ee/77/16.json')
+  .intercept(nock =>
+    nock('https://devdiv.vsrm.visualstudio.com')
```
PyvesB: Out of interest, why do we have to mock the responses for the release badges?

lib/all-badge-examples.js
```diff
@@ -22,6 +22,20 @@ const visualStudioTeamServicesDoc = `
...
 </p>
+<p>
+  Optionally, you can specify a named branch
+  <code>https://img.shields.io/vso/build/TEAM_NAME/PROJECT_ID/BUILD_DEFINITION_ID/NAMED_BRANCH</code>.
+</p>
+`
+
+const visualStudioTeamServicesReleaseDoc = `
+<p>
+  To obtain your own badge, you will first need to enable badges on you release definition.
```
PyvesB: Small typo: `your release definition`

services/vso/vso.tester.js
Thilas: The reason is quite simple: I didn't find any public VSTS release like the one I found for builds (from Microsoft .NET Core). Since I prefer not to use one from my company (since it is private, hence it can change without notice).

Thilas:
@PyvesB, done.

Issue #1997, opened 2018-08-29
Lowercase badge labels


paulmelnikow:
Let's review the badge labels and make sure they are all lowercase. We're very close, though there are a select few like AUR and OSS Lifecycle are not.

Let's also document this requirement in the [badge guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines).

Ref: https://github.com/badges/shields/pull/1965#discussion_r213098278

RedSparr0w:
Sounds like a plan,
labels + messages are both generally lowercase _(with the exception of the `for-the-badge` style)_

although i think some do justify not being lowercase:
![license](https://img.shields.io/:license-MIT-blue.svg)
_(not sure about the AUR and OSS badges)_

PyvesB:
I think messages are fine, we should probably only enforce lower-casing for the left hand-side of badges. 😉

PR #2058, opened 2018-09-06
[aur packagist php-eye travis vaadin-directory github david jitpack twitter] Enforced lowercase badge labels

PyvesB:
This pull request addresses issue #1997.

I implemented lower casing of the left-handside of badges in the `makeLabel` function, but only for `defaultLabel`. If the user specifies a label override, he can use upper case characters freely.

Note that when the function was called with hardcoded sting values (e.g. `getBadgeData('AUR', data)`), I lower cased the strings at the point of call (`getBadgeData('aur', data)`) as well to avoid any confusion in the code.

Query parameters specifying special styles are not affected (i.e. capitalisation for social badges, full upper casing for for-the-badge).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,7 +91,7 @@ There are three places to get help:
 Badge guidelines
 ----------------
 
-- The left-hand side of a badge should not advertise. It should be a *noun*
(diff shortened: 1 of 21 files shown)


PyvesB:
Thanks for reviewing!

Issue #2057, opened 2018-09-06
Logo color not changing for service badges


TobiasRoeddiger:
Hello. I realized that for services I am able to add a logo but when I try to change the color, the parameter is not applied.

Examples: 
(1) [https://img.shields.io/cdnjs/v/jquery.svg?logo=apple&logoColor=green](https://img.shields.io/cdnjs/v/jquery.svg?logo=apple&logoColor=green)
(2) [https://img.shields.io/apm/l/vim-mode.svg?logo=apple&logoColor=red](https://img.shields.io/apm/l/vim-mode.svg?logo=apple&logoColor=red)

chris48s:
Thanks for this report.

For info, if anyone is looking into this, it seems to be an issue only on badges which extend the new `BaseService` class. This does work as expected on badges which use the legacy architecture (extending `LegacyService`) e.g:

https://img.shields.io/travis/rust-lang/rust.svg?logo=apple&logoColor=red
![](https://img.shields.io/travis/rust-lang/rust.svg?logo=apple&logoColor=red)

PR #2059, opened 2018-09-06
fixed bug that prevented logo color from being changed

TobiasRoeddiger:
fixed the issue mentioned in #2057


chris48s:
Confirmed working on staging:

https://shields-staging-pr-2059.herokuapp.com/cdnjs/v/jquery.svg?logo=apple&logoColor=red


Thanks. I'll merge this now. It would also be nice if we had better test coverage for the overrides to prevent future regressions but I'll raise a new issue for that.

PR #2061, opened 2018-09-06
Made JSON snapshot test pass on Windows

PyvesB:
The test `should always produce the same JSON (unless we have changed something!)` was systematically failing on my laptop. If my understanding is correct, the snapshot data for the test was generated on a Unix platform and therefore uses Unix line endings (`\n`).

On Windows, when loading and preparing the badge templates, Windows line endings are used (`\r\n`). The comparison with the snapshot was therefore failing.

When run on Windows, the test suite will now use a different snapshot, with the proper line endings.

### DIFF
--- a/__snapshots__/make-badge.spec.js
+++ b/__snapshots__/make-badge.spec.js
@@ -9,6 +9,13 @@ exports['The badge generator JSON should always produce the same JSON (unless we
 }
 `
 
+exports['The badge generator JSON should always produce the same JSON on Windows (unless we have changed something!) 1'] = `
+{
+  "name": "cactus",
+  "value": "grown"
+}
+`
+
 exports['shields GitHub logo default color (#333333) 1'] = `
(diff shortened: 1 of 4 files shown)


chris48s:
I know at the moment we only have one test which is affected by this issue, but if we add more snapshot tests, it seems like it is going to become frustrating if we need to have 2 versions of every test. Also, having 2 different fixtures which are only different due to non-printing characters is quite confusing. It isn't visually obvious how those 2 are different.

Would a better way to do this be to use something like [eol](https://www.npmjs.com/package/eol) to generate the expected string in a cross-platform compatible way? e.g:

```js
exports['bla bla bla'] = eol.auto('mystring')
```

then we would only need one version of the test.

PyvesB:
Thanks for your feedback, it makes perfect sense. I've just committed an alernative approach based on your `eol` module suggestion.

PR #2065, opened 2018-09-08
Better support for VSTS build badges

Thilas:
Newly created build definitions on VSTS does not support old badge endpoint. Instead, a new one has been provided and supports all build definitions.


Thilas:
@PyvesB, thanks for the review. Will you also merge the PR? Thanks!

Issue #2054, opened 2018-09-05
Show popout style example with a logo on index page


TobiasRoeddiger:
First of all thank you very much for this awesome project. I figured there seems to be an issue with the logo when choosing the popout style:
This breaks the boundaries of the badge: https://img.shields.io/badge/style-popout-green.svg?logo=apple&style=popout
Whereas it is correct for e.g. the flat style: https://img.shields.io/badge/style-flat-green.svg?logo=apple&style=flat

PyvesB:
Hello @TobiasRoeddiger,

That's actually intended behaviour, the popout style is supposed to make the logo pop out of the boundaries of the badge. See #1478 for more information. 😉 

Cheers,

Pyves

TobiasRoeddiger:
Oh I missed something there - cheers!

chris48s:
I suspect this is confusing for users because the example on the front page shows the popout style example without a logo. The main thing this style does is display the logo differently. I think we could make it more obvious what this style is for by showing an example of a flat badge with a logo and a popout badge with a logo.

chris48s:
closed in #2084

PR #2084, opened 2018-09-16
show examples with logos on index page

chris48s:
Refs #2054

I settled on this layout as the one I like best:

screenshot at 2018-09-16 15-26-41

I tried a few others though. These are a couple of the rejects:

screenshot at 2018-09-16 15-27-04

screenshot at 2018-09-16 15-29-47

Does the one I've gone with seem like the right solution? Any other thoughts?


PyvesB:
I'm in two minds here:
* on the one hand, this makes things more informative for users, which is great!
* on the other hand, I feel that it overloads that part of the homepage. On my mobile device, the URLs are now broken down into two separate lines.

Yet another option could be to only have the badges with the logos, and get rid of the plain examples. To make things extra clear, in addition to the description "The following styles are available (flat is the default as of Feb 1st 2015):", we could add a sentence similar to "The following badges are rendered with an example logo to show how the style affects image rendering, but it is of course optional.".

RedSparr0w:
I like either of the first 2,
Although i think the 2nd option would be better for mobile with the current front page layout.
Or @PyvesB suggestion would also work nicely.

chris48s:
I've gone with @PyvesB 's suggestion - I think I was over-complicating things :)

PR #2086, opened 2018-09-16
refactor arch user repositories [aur] service

chris48s:
Refs #1358

This PR ports the Arch User Repositories integration to the new service architecture.


PyvesB:
We seem to usually break down different badges implementations into different files, shouldn't we have version, votes and license separated?

chris48s:
We've got a mix. For some services we've got the base class in one file and each service in its own file, but for example APM also defines 3 badge classes in one file: https://github.com/badges/shields/blob/264f5e39dc4ed626dbc107a364972724394e9901/services/apm/apm.service.js#L157-L161

I definitely wouldn't want to define a large service family like NPM or PyPI all in one file, but I reckon when we're dealing with a couple hundred lines of code its fine. I guess its a matter of taste though.

PyvesB:
Fair enough. I'm a bit short on time at the moment, but will aim at reviewing this more carefully later during the week. 😉

services/aur/aur.service.js
```diff
@@ -1,67 +1,174 @@
...
-          }
-        })
-      })
-    )
+const aurSchema = Joi.object({
+  resultcount: nonNegativeInteger,
+  results: Joi.alternatives(
+    Joi.array()
+      .min(0)
+      .max(0)
```
PyvesB: I believe you could simply use `length(0)` here.

services/aur/aur.service.js
```diff
@@ -1,67 +1,174 @@
...
+      // body = {"version":1,"type":"info","resultcount":0,"results":[]}
+      throw new NotFound()
+    }
+    return super._validate(data, schema)
+  }
+}
+
+class AurLicense extends BaseAurService {
+  static render({ license }) {
+    return { message: license, color: 'blue' }
```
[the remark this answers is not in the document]
PyvesB: Fair enough.

services/aur/aur.tester.js
```diff
@@ -16,7 +15,7 @@ t.create('version (valid)')
   .get('/version/yaourt.json?style=_shields_test')
   .expectJSONTypes(
     Joi.object().keys({
-      name: 'aur',
+      name: 'version',
```
PyvesB: I'm somewhat confused here. Why has the label changed to `version`?

chris48s: Good question. Looking over the examples we usually use the platform name as the label on version badges so I shouldn't have changed this.

chris48s:
cheers :)

PR #2095, opened 2018-09-18
correct GitHub capitalization

coliff:


chris48s:
Thanks for having a look at this. Just scanning over the index page, there are actually several other titles where we're inconsistently using 'Github' instead of 'GitHub'. Would you be willing to update those as well?

coliff:
sorry for delayed reply - if I find some time I'll hunt down any others and make a new PR :-)

chris48s:
No probs. I've already followed up on the others in #2134
Thanks

PR #2098, opened 2018-09-18
Support NuGet packages using SemVer 2

### LINKED ISSUE #1044: NuGet badges don't work for SemVer 2 packages
Might be related to use of the v3 endpoint or not (or possibly just needing to add the right semverlevel=2.0.0 query parameter), but badges for SemVer 2 packages aren't currently working.

https://github.com/badges/shields/issues/655

https://img.shields.io/nuget/vpre/RSAKeyVaultProvider.svg
https://www.nuget.org/packages/RSAKeyVaultProvider

See https://github.com/NuGet/Home/wiki/Semver-2.0.0-Protocol for the protocol support update. Needs an additional query parameter to return semVer 2.0 packages.

sharwell:
Fixes #1044

### DIFF
--- a/services/nuget/nuget.service.js
+++ b/services/nuget/nuget.service.js
@@ -182,7 +182,8 @@ function mapNugetFeed({ camp, cache }, pattern, offset, getInfo) {
             searchQueryResources[randomEndpointIdx]['@id'] +
             '?q=packageid:' +
             encodeURIComponent(id.toLowerCase()) + // NuGet package id (lowercase)
-            '&prerelease=true' // Include prerelease versions?
+            '&prerelease=true' + // Include prerelease versions?
+            '&semVerLevel=2' // Include packages with SemVer 2 version numbers
 
           request(reqUrl, (err, res, buffer) => {
             if (err != null) {


chris48s:
Hi. Thanks for having a look at this. I've edited your PR title so that the existing myget/nuget test suites are run against your changes. Having a quick scan over the result, I suspect the only thing that needs to happen is that the existing nuget/myget tests which intercept a request to mock the response need to be updated so that the request URL being intercepted includes `&semVerLevel=2`. Those tests are in

https://github.com/badges/shields/blob/master/services/nuget/nuget.tester.js
https://github.com/badges/shields/blob/master/services/myget/myget.tester.js

sharwell:
@chris48s another question: I tested this manually against the NuGet API endpoint linked above. It sounds like the same code might apply to other services. Do you have a list of all services we'd want to test to ensure semVerLevel is allowed (possibly ignored) by each service? For example, MyGet and Chocolatey seem like candidates, but I don't know how to test this hypothesis.

sharwell:
@chris48s Tests in those files have been updated, thanks!

chris48s:
There's 2 different functions for working with NuGet-based feeds: `mapNugetFeed()` and `mapNugetFeedv2()`. This PR modifies `mapNugetFeed()` which is used by the MyGet and NuGet badges only.

Chocolatey, ReSharper Gallery and Powershell Gallery use `mapNugetFeedv2()` (I think those services are based on a different version of the platform) so this change won't affect those badges.

There are now some minor code style failures but they can be auto-fixed with `npm run prettier`

sharwell:
> There are now some minor code style failures but they can be auto-fixed with `npm run prettier`

I don't have a way to do this, and it doesn't say what the problems are

PR #2102, opened 2018-09-19
Fixed [Codacy] coverage examples and improved unknown coverage handling

PyvesB:
The project used on the homepage doesn't have any relevant coverage data: 

I've switched it to one that does: 

As you can see above the unknown coverage case is handled pretty poorly, with the exclamation mark and the bright green colour. From now on, such a badge would render as:

### DIFF
--- a/services/codacy/codacy-coverage.service.js
+++ b/services/codacy/codacy-coverage.service.js
@@ -39,8 +39,13 @@ module.exports = class CodacyCoverage extends LegacyService {
               return
             }
             try {
-              badgeData.text[1] = res
-              badgeData.colorscheme = coveragePercentageColor(parseInt(res))
+              const coverage = parseInt(res)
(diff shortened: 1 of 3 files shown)


services/codacy/codacy-coverage.service.js
```diff
@@ -39,8 +39,13 @@ module.exports = class CodacyCoverage extends LegacyService {
               return
             }
             try {
-              badgeData.text[1] = res
-              badgeData.colorscheme = coveragePercentageColor(parseInt(res))
+              const coverage = parseInt(res)
+              if (isNaN(coverage)) {
```
chris48s: This might be a case where it is better to use [Number.isNaN()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) in preference to `isNaN()`

Issue #2076, opened 2018-09-13
Is Wercker badge reporting the correct status?


chris48s:
While working on #2075 I noticed that the API endpoint we're using for Wercker doesn't seem to be reporting the latest build status

e.g:
https://app.wercker.com/wercker/wercker/
https://app.wercker.com/api/v3/applications/wercker/wercker/builds

https://app.wercker.com/wercker/go-wercker-api/
https://app.wercker.com/api/v3/applications/wercker/go-wercker-api/builds

It looks to me like what we should be doing is making a call to the /runs endpoint
https://devcenter.wercker.com/development/api/endpoints/runs/

so for

https://app.wercker.com/wercker/wercker/
we should actually be calling
https://app.wercker.com/api/v3/runs?applicationId=56a9f09b4239abda60001670

and for

https://app.wercker.com/wercker/go-wercker-api/
we should actually be calling
https://app.wercker.com/api/v3/runs?applicationId=559e33c8e982fc615500b357

..but I'm slightly unclear what the difference between a [build](https://devcenter.wercker.com/development/api/endpoints/applications/#list-builds) and a [run](https://devcenter.wercker.com/development/api/endpoints/runs/) is here as I'm not really familiar with the service, so I might be making the wrong assumption here.

Does anyone have any better insight on this? @pyvesb - you were the last person to refactor this, so I nominate you as the wercker expert :wink:

chris48s:
OK. Thanks for looking to that. Given our current badge endpoint is `/ci`, it seems like we should switch our `/ci` badge to report the status of the upstream `/run` endpoint.

We could also add another endpoint for `/build`, given they are different things.

Does that seem right to you? Given it links in with stuff I'm working on in #2075 anyway I'm ahppy to pick that up.

PyvesB:
Some thoughts. Our current Wercker badges support two different URL styles, one with the project ID (`/wercker/ci/projectId`) and the other with the application name (`/wercker/ci/applicationName`, with a different regex). The latter is probably the most used as it's the one that has long been advertised on the homepage.

[...]

PR #2075, opened 2018-09-13
update service tests tutorial, affects [wercker] also [dependabot docker dynamic-json dynamic-xml dynamic-yaml npm website]

chris48s:
I've gone over the service tests tutorial and updated it so that we work through writing tests for a service which extends `BaseJsonService`, reflecting our current testing practice.

Wercker seemed like a good example to use - it allowed me to cover the things I wanted to cover without being too complicated. The tests I've added for wercker aren't massively useful but since I put them in the tutorial we might as well add them to the test suite :)

refs #1984

PR #456, opened 2015-05-29
Add wercker v3 integration

hatchan:
This allows users to specify a owner + application name, instead of a application ID. 

Made the regex for the old endpoint a bit more strict so it wouldn't conflict with the new one.

It also uses the public, supported API.

PR #2103, opened 2018-09-19
fixes to [wercker] badge

chris48s:
Started looking at #2076 and as I dug into it found some more issues. I've made several changes to the wercker integration in this PR:

1. If a application or branch does exist but doesn't have any builds or CI runs, a response of `[]` is returned. Example: https://app.wercker.com/api/v3/applications/wercker/stern/builds I've modified the validation and added a case for this.

2. Our current/previous regex was not working correctly when using a project ID:

    If I call https://img.shields.io/wercker/ci/559e33c8e982fc615500b357.svg that generates  based on (correctly) calling https://app.wercker.com/getbuilds/559e33c8e982fc615500b357?limit=1

    If I call https://img.shields.io/wercker/ci/559e33c8e982fc615500b357/master.svg that generates  because "559e33c8e982fc615500b357/master" is matching the regex pattern `([^/]+/[^/]+)` so we are (incorrectly) calling https://app.wercker.com/api/v3/applications/559e33c8e982fc615500b357/master/builds?limit=1 instead of (correctly) calling https://app.wercker.com/getbuilds/559e33c8e982fc615500b357?limit=1&branch=master .

    I've updated the regex so we try to specifically match `([a-fA-F0-9]{24})` first then fall back to `([^/]+/[^/]+)` which fixes that (unless, of course, your github username is `fb1ed340ffed75c22dc301c3`, in which case I broke it :smile:  )

3. Following on from #2076 I've changed this so that:

    * `ci/{projectId}` reports the result of a CI run (totally sensible)
    * `build/{projectId}` reports the result of a build (totally sensible)
    * `build/{applicationName}` reports the result of a build (totally sensible)
    * `ci/{applicationName}` reports the result of a _build_ but isn't documented anywhere (I'm still in 2 minds about whether this is actually the right thing to do here)

I've also updated/added tests accordingly.

Annoyingly, this also has the side effect that it probably makes this a bad example to base documentation on, so I probably now want to go back and change #2075 but lets get this right first and then I'll worry about what I'm going to do with that.


PyvesB:
As far as 2. is concerned, when I added branch support in #1920 I actually didn't intend on supporting branches for our old ID based URL, as it had no longer been advertised on the homepage for years. I simply carried over the backwards compatibility for old badges that may still be lying around, so I wouldn't actually say that it "was not working correctly". 😄

This PR looks overall good to me, but I don't think we should add the new `build/{projectId}` URL. It maps to `https://app.wercker.com/getbuilds` underneath the hood, which I believe is the old undocumented and deprecated API, which we started phasing out on our side in #456. It seems wrong to add new Shields badge based on it.

In my opinion we should only have `build/{applicationName}` and `ci/{projectId}`, and keep `ci/{applicationName}` for backwards compatibility but not actually advertise it on the homepage.

Nevertheless, ti would be nice for users to use `ci/{applicationName}` without having to figure out their project ID, maybe we raise a ticket with Wercker so that support using names for the /runs endpoint? Our badges would then be consistent with `build/{applicationName}` and `ci/{applicationName}` as main URLs, and we could hide `ci/{projectId}` again and only keep it for backwards compatibility.

Does this make sense?

chris48s:
> ..so I wouldn't actually say that it "was not working correctly"

OK, well since we _can_ make it work, we might as well while I'm at it..

> In my opinion we should only have `build/{applicationName}` and `ci/{projectId}`, and keep `ci/{applicationName}`

Done

Issue #2112, opened 2018-09-24
Subject in custom badge forces lowercase


garetmckinley:
Not sure if this is a bug or a stylistic choice, but I noticed that custom badges only respect the case of the status, not the subject.

Example: `https://img.shields.io/badge/TEST-CASE-green.svg` ![](https://img.shields.io/badge/TEST-CASE-green.svg)

I swear I've used badges with uppercase in the subject line, but maybe I'm remembering incorrectly. I first noticed the issue on http://makeapullrequest.com/ the URL Kent uses has `PRs` in it, but the badge looks odd rendering at `prs`. Which makes me think it's a bug that was introduced over time

mabbas64:
The same has happend to all my badges. They were nicely formated like  "MyTitle | 1234"  and now they're all lowercased during last week or two somehow.  and read "mytitle | 1234" 

e.g.: 

```
<img src="https://img.shields.io/badge/EQ BankCard 8 - 735 | 1-6b8b08.svg">
```

chris48s:
Hi thanks for reporting this. It looks like this was introduced in #2058 although it is not the intended behaviour. I've proposed a fix in #2116

PR #2116, opened 2018-09-24
treat label as override in static badge

chris48s:
This fixes the issue reported in issue #2112 by passing the label to `makeLabel()` as an override instead of a `defaultLabel`.

This does also change the behaviour of a call like https://img.shields.io/badge/foo-bar-green.svg?label=baz which will now do this:

instead of this:

but that's probably a good thing :D

Does this badge have any service tests, or is it covered in the main test suite?


PyvesB:
Thanks for fixing this! Aren't the tests for these generate badges in [server.spec.js](https://github.com/badges/shields/blob/master/server.spec.js)?

chris48s:
> Aren't the tests for these generate badges in server.spec.js?

that's the badger :+1:

PR #2125, opened 2018-09-27
Add [JenkinsPluginInstalls] service

svenschoenung:
This adds a badge that shows the number of installations of a particular Jenkins plugin. You can show the total number of installations across all versions or pick a particular version.

The data is retrieved from the [official statistics](https://stats.jenkins.io/) that the Jenkins project provides. For example, the data for the [View Job Filters Plugin](https://plugins.jenkins.io/view-job-filters) can be found here:

https://stats.jenkins.io/plugin-installation-trend/view-job-filters.stats.json

Note that these are actual installs, not downloads (i.e if somebody uninstalls a plugin, the number goes down). I've still added the badge to the downloads category because that's the closest category there is.

### DIFF
--- a/services/jenkins/jenkins-plugin-installs.service.js
+++ b/services/jenkins/jenkins-plugin-installs.service.js
@@ -110,8 +110,7 @@ class JenkinsPluginInstalls extends BaseJsonService {
         staticExample: this.render({
           label: this._getLabel(),
           installs: 10247,
-        }),
-        keywords: ['jenkins'],
+        })
       },
       {
         title: 'Jenkins Plugin installs',
@@ -120,8 +119,7 @@ class JenkinsPluginInstalls extends BaseJsonService {
         staticExample: this.render({
           label: this._getLabel('1.26'),
           installs: 955,
-        }),
-        keywords: ['jenkins'],
+        })
       },
     ]
   }
--- a/services/jenkins/jenkins-plugin-installs.service.js
+++ b/services/jenkins/jenkins-plugin-installs.service.js
@@ -3,7 +3,7 @@
 const Joi = require('joi')
 
 const BaseJsonService = require('../base-json')
-const { NotFound, InvalidParameter } = require('../errors')
+const { NotFound } = require('../errors')
 const {
   downloadCount: downloadCountColor,
 } = require('../../lib/color-formatters')
--- a/services/jenkins/jenkins-plugin-installs.service.js
+++ b/services/jenkins/jenkins-plugin-installs.service.js
@@ -59,27 +59,23 @@ class JenkinsPluginInstalls extends BaseJsonService {
     }
   }
 
-  async handle({ info, plugin, version }) {
+  async handle({ plugin, version }) {
     const label = this.constructor._getLabel(version)
     const json = await this.fetch({ plugin, version })
 
     let installs
-    if (info === 'iv') {
+    if (version) {
       installs = json.installationsPerVersion[version]
       if (!installs) {
         throw new NotFound({
           underlyingError: new Error('non-existent version'),
         })
       }
-    } else if (info === 'i') {
+    } else {
       const latestDate = Object.keys(json.installations)
         .sort()
         .slice(-1)[0]
       installs = json.installations[latestDate]
-    } else {
-      throw new InvalidParameter({
-        underlyingError: new Error('invalid info'),
-      })
     }
 
     return this.constructor.render({ label, installs })
@@ -95,9 +91,9 @@ class JenkinsPluginInstalls extends BaseJsonService {
 
   static get url() {
     return {
-      base: 'jenkins/plugin',
-      format: '(i|iv)/([^/]+)/?([^/]+)?',
-      capture: ['info', 'plugin', 'version'],
+      base: 'jenkins/plugin/i',
+      format: '([^/]+)/?([^/]+)?',
+      capture: ['plugin', 'version'],
     }
   }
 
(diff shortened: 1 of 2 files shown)


services/jenkins/jenkins-plugin-installs.service.js
```diff
@@ -0,0 +1,130 @@
...
+          underlyingError: new Error('non-existent version'),
+        })
+      }
+    } else if (info === 'i') {
+      const latestDate = Object.keys(json.installations)
+        .sort()
+        .slice(-1)[0]
+      installs = json.installations[latestDate]
+    } else {
+      throw new InvalidParameter({
```
PyvesB: Can this case actually ever be reached? Shouldn't the regex guarantee that we can only have either `iv` or `i`?

services/jenkins/jenkins-plugin-installs.service.js
```diff
@@ -0,0 +1,130 @@
...
+
+  static get url() {
+    return {
+      base: 'jenkins/plugin',
+      format: '(i|iv)/([^/]+)/?([^/]+)?',
+      capture: ['info', 'plugin', 'version'],
+    }
+  }
+
+  static get examples() {
```
PyvesB: I'm wondering about the URLs here. Wouldn't simply having `'i/view-job-filters` and `i/view-job-filters/1.26` be enough to indicate whether we're targeting the total or a specific version? In other words, do we really need to make the distinction with two path variants, `iv` and `i`?

Similarly, for a lot of build badges, we optionally support branches, and branches are simply indicated by appending something like `/master` to the standard URL.

In addition to being slightly more consistent with those badges, this will simplify the URL regex

services/jenkins/jenkins-plugin-installs.service.js
```diff
@@ -0,0 +1,130 @@
...
+  async handle({ info, plugin, version }) {
+    const label = this.constructor._getLabel(version)
+    const json = await this.fetch({ plugin, version })
+
+    let installs
+    if (info === 'iv') {
+      installs = json.installationsPerVersion[version]
+      if (!installs) {
+        throw new NotFound({
+          underlyingError: new Error('non-existent version'),
```
PyvesB: By specifying the `prettyMessage` property, I believe we could be a bit more specific in the badge error message and display "version not found" for instance.

For the total installs, we could specify the 404 response code message by adding a parameter similar to the following to requestJSON:
`````
errorMessages: {
    404: 'plugin not found',
},

File: services/jenkins/jenkins-plugin-installs.service.js [hunk 4]
```diff
@@ -0,0 +1,130 @@
...
+    return [
+      {
+        title: 'Jenkins Plugin installs',
+        exampleUrl: 'i/view-job-filters',
+        urlPattern: 'i/:plugin',
+        staticExample: this.render({
+          label: this._getLabel(),
+          installs: 10247,
+        }),
+        keywords: ['jenkins'],
```
PyvesB: The "jenkins" keyword is actually already inferred by the service name. 😉

services/jenkins/jenkins-plugin-installs.service.js
svenschoenung: I was wondering why `non-existent version` from the `underlyingError` didn't show up, but didn't bother to look into it since that's how the `GemDownloads` service [is doing it as well](https://github.com/badges/shields/blob/454575c4cbc8a736a597f878b9afbb2fd8122a75/services/gem/gem-downloads.service.js#L91).

svenschoenung:
Thanks for the feedback, I think I incorporated it all. Have another look when you find the time.

services/jenkins/jenkins-plugin-installs.tester.js
```diff
@@ -0,0 +1,107 @@
...
+      .reply(200, { name: 'view-job-filters', installationsPerVersion: {} })
+  )
+  .expectJSON({ name: 'installs', value: 'invalid response data' })
+
+t.create('version installs | not found: non-existent plugin')
+  .get('/not-a-plugin/1.26.json')
+  .expectJSON({ name: 'installs', value: 'plugin not found' })
+
+t.create('version installs | not found: non-existent version')
+  .get('/view-job-filters/1.99.json')
```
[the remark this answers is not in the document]
svenschoenung: I actually picked 1.99 because it's never going to exist. I released 2.0 of this plugin recently, it's just that the stats that Jenkins provides haven't caught up yet.

But you're right, there's no way to see this from the test alone, it's much better to make the non-existence of the version explicit.

services/jenkins/jenkins-plugin-installs.tester.js
```diff
@@ -0,0 +1,107 @@
...
+      name: 'installs@1.26',
+      value: isMetric,
+    })
+  )
+
+t.create('version installs | valid: alpha-numeric version')
+  .get('/view-job-filters/1.27-DRE1.00.json')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'installs@1.27-DRE1.00',
```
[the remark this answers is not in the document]
svenschoenung: I prefer the Gem one personally. 

The way I see it, the left side of the badge is the query, the right side is the result. So left side says "show me the number of downloads for version 4.1.0" or "show me the build status of the testing branch" or something like that. The right side then displays the result: "50k" or "failing" etc.

But that's up to you guys to decide. I'll stick to whatever convention you come up with.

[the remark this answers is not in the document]
PyvesB: I agree with @svenschoenung reasoning.

platan:
I've merged master to this branch.

PR #2130, opened 2018-09-30
Add [Steam] Service badges WIP

### LINKED ISSUE #2121: Steam workshop item stats to shield
Any chance for getting Steam workshop item subscribers and other info on a shield? That info I believe can be freely retrieved so I assume it could be used for a shield

TagnumElite:
This adds multiple badges which comes from what data can be fetched by the Steam WEB API

The API documentation can be found [here](https://partner.steamgames.com/doc/webapi/ISteamRemoteStorage) with the import bits being:

1. [GetCollectionDetails](https://partner.steamgames.com/doc/webapi/ISteamRemoteStorage#GetCollectionDetails)
2. [GetPublishedFileDetails](https://partner.steamgames.com/doc/webapi/ISteamRemoteStorage#GetPublishedFileDetails)

Since this is my first request into this repo, I would like for people to tell me where I can improve if possible.

I have run the tests on my local server but I may have not added all the required tests if so do tell, please!

This pull request resolves #2121

### DIFF
--- a/services/steam/steam.service.js
+++ b/services/steam/steam.service.js
@@ -1,0 +1,357 @@
+'use strict'
+
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
+const { metric, formatDate } = require('../../lib/text-formatters')
+const { age: ageColor } = require('../../lib/color-formatters')
+const prettyBytes = require('pretty-bytes')
+
+const steamCollectionSchema = Joi.object({
+  response: Joi.object().keys({
+    result: Joi.number().integer().min(1).max(1).required(),
+    resultcount: Joi.number().integer().min(0).max(1).required(),
+    collectiondetails: Joi.array().items(Joi.object({
+      publishedfileid: Joi.string(),
+      result: Joi.number().integer(),
+      children: Joi.array()
+    })).required()
+  }).required(),
+}).required()
+
+const steamFileSchema = Joi.object({
+  response: Joi.object().keys({
+    result: Joi.number().integer().min(1).required(),
+    resultcount: Joi.number().integer().min(0).required(),
+    publishedfiledetails: Joi.array().items(Joi.object({
+      publishedfileid: Joi.string().required(),
+      result: Joi.number().integer().required(),
+      file_size: Joi.number().integer(),
+      time_created: Joi.number().integer(),
+      subscriptions: Joi.number().integer(),
+      favorited: Joi.number().integer(),
+      lifetime_subscriptions: Joi.number().integer(),
+      lifetime_favorited: Joi.number().integer(),
+      views: Joi.number().integer(),
+    }))
+  }).required(),
+}).required()
+
+class SteamCollectionCount extends BaseJsonService {
+  async fetch({collectionId}) {
+    const url = 'https://api.steampowered.com/ISteamRemoteStorage/GetCollectionDetails/v1/?format=json'
+    return this._requestJson({
+      url,
+      schema: steamCollectionSchema,
+      errorMessages: {
+        400: 'bad request'
+      },
+      options: {
+        method: 'POST',
+        form: {
+          collectioncount: '1',
+          'publishedfileids[0]': collectionId
+        }
+      }
+    })
+  }
+
+  static render({ size }) {
+    return { message: metric(size), color: 'green' }
+  }
+
+  async handle({ collectionId }) {
+    const json = await this.fetch({ collectionId })
+    return this.constructor.render({ size: json.response.collectiondetails[0].children.length})
+  }
+
+  static get category() {
+    return 'other'
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'files' }
+  }
+
+  static get url() {
+    return {
+      base: 'steam/collection-size',
+      format: '(.+)',
+      capture: ['collectionId'],
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Steam Collection Size',
+        exampleUrl: '16524',
+        urlPattern: ":collection_id",
+        staticExample: this.render({ size: 20 }),
+        keywords: ['steam']
+      }
+    ]
+  }
+}
+
+class SteamFileService extends BaseJsonService {
+
+  async fetch({fileId}) {
+    const url = 'https://api.steampowered.com/ISteamRemoteStorage/GetPublishedFileDetails/v1/?format=json'
+    return this._requestJson({
+      url,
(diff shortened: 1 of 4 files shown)


PyvesB:
Hello @TagnumElite, thank you for your first contribution to Shields! I haven't gone through your implementation in details yet, but one of the tests seems to be failing when run on the CI. 😉

chris48s:
> I haven't gone through your implementation in details yet

I've already started looking over this one. It is quite a large PR adding 7 badges so will probably be useful for a couple of us to look over at some point, but we'll end up duplicating the same points if we both look over it now (again). Maybe wait until I've done a first pass so we don't waste time picking up the same issues.

There is a review app at https://shields-staging-pr-2130.herokuapp.com/

services/steam/steam.service.js
```diff
@@ -0,0 +1,395 @@
...
+        .integer()
+        .min(0)
+        .max(1)
+        .required(),
+      collectiondetails: Joi.array()
+        .items(
+          Joi.object({
+            publishedfileid: Joi.string(),
+            result: Joi.number().integer(),
+            children: Joi.array(),
```
chris48s: `result` and `children` should be `.required()` here. Any property we're going to reference without a `try`/`catch` in the `handle()` should be `.required()` in the validator.
Does `publishedfileid` actually get used? If not, we can exclude it from the validator.

services/steam/steam.service.js
```diff
@@ -0,0 +1,395 @@
...
+          result: Joi.number()
+            .integer()
+            .required(),
+          file_size: Joi.number().integer(),
+          time_created: Joi.number().integer(),
+          subscriptions: Joi.number().integer(),
+          favorited: Joi.number().integer(),
+          lifetime_subscriptions: Joi.number().integer(),
+          lifetime_favorited: Joi.number().integer(),
+          views: Joi.number().integer(),
```
chris48s: Again, I think all of these properties should be `.required()`. We need to specify `.required()` in order to throw an error if it isn't present in the response.

services/steam/steam.service.js
```diff
@@ -0,0 +1,395 @@
...
+  }
+
+  async handle({ collectionId }) {
+    const json = await this.fetch({ collectionId })
+    if (json.response.collectiondetails[0].result === 1) {
+      return this.constructor.render({
+        size: json.response.collectiondetails[0].children.length,
+      })
+    } else {
+      return { message: 'collection not found', color: 'red' }
```
chris48s: Instead of manually constructing error reponses, we can throw an exception, for example:

https://github.com/badges/shields/blob/302c8606ff4c975b6310fe2cdfe58f92cede290b/services/npm/npm-version.service.js#L95-L97

This will ensure consistent behaviour and formatting with other errors. Also in the cases where we are failing with "file not found", "collection not found" etc can you add test cases to the test suite covering those cases. You can use `nock` to provide a mock response recreating these conditions. There are some tests using mocked responses in the NPM licence tests that you can use as examples: https://github.com/badges/shields/blob/master/services/npm/npm-license.tester.js

services/steam/steam.service.js
```diff
@@ -0,0 +1,395 @@
...
+      capture: ['collectionId'],
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Steam Collection Files',
+        exampleUrl: '180077636',
+        urlPattern: ':collection_id',
```
chris48s: How does a user get or work out their collection ID or file ID? Is this really clea/obvious to users in this ecosystem or do we need to document how to obtain these IDs?

services/steam/steam.service.js
```diff
@@ -0,0 +1,395 @@
...
+      capture: ['fileId'],
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Steam File Size',
+        exampleUrl: '100',
+        urlPattern: ':file_id',
```
chris48s: Minor style point: I think we're mostly using `camelCase` for the url params in preference to `snake_case`

services/steam/steam.service.js
```diff
@@ -0,0 +1,395 @@
...
+      base: 'steam/favorites',
+      format: '(.+)',
+      capture: ['fileId'],
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Steam Favourites',
```
chris48s: I note you've gone with the UK spelling "favourite" in the description but the American "favorite" on the badge and everywhere else. Can we go with the American style consistently.

chris48s:
Thanks for submitting your first contribution. This looks like a great start. I've left some comments inline but I don't think any of them are major. I also have a few points which don't really fit inline in the diff:

---

I've updated the PR title so that it includes "[steam]". This tells the CI build to run the service tests you've defined in `services/steam/steam.tester.js`. As @pyvesb notes it looks like there is a failing test. Can you have a look: https://circleci.com/gh/badges/shields/15783

Also If I call https://shields-staging-pr-2130.herokuapp.com/steam/favorites/100.svg it renders `[favorites | undefined]`. I think that's the same issue.

---

I'm not a big fan of `lime` as a colourscheme e.g:


We tend to use brightgreen in this context. Its a bit easier on the eye :)


---

How about we move the 'subscriptions' and 'favorites' badges into the 'rating' category?

services/steam/steam.service.js
TagnumElite: The id is in the URL of the workshop item.
Example https://steamcommunity.com/sharedfiles/filedetails/?id=1525218777 => `?id=1525218777`

How to tell the other users of this I don't know how to do that.

services/steam/steam.service.js
TagnumElite: If the item is not found is results in a JSON format looking like this:
```json
{"response":{"result":1,"resultcount":1,"publishedfiledetails":[{"publishedfileid":"-1","result":9}]}}
```

With the request being successful `response.result == 1` but the item not being found `response.publishedfiledetails.result == 9`. 

If I say these things need to be required the schema fails right? Are there examples of how to use this to catch errors?

TagnumElite:
Hmmm, duplicate commits???

TagnumElite:
I could also be a bad person and should burn if I do do it, but I am sure I can just, delete the commits or revert to a previous commit. I know this is heavily frowned upon though.

I suspect that making a new pull request would be better and sections so I will be doing that while implementing the changes you asked for unless asked otherwise.

TagnumElite:
I will be creating a new pull request without the mess above.

Issue #2106, opened 2018-09-22
Cobertura badges broken


PyvesB:
I believe that our Jenkins Cobertura badge was broken in #1593. When the Jacoco badge was added, some of the common code was extracted, but we ended applying Jacoco parsing to the Cobertura API responses. There was also some confusion about the tests and we don't actually have any Cobertura tests even if some test descriptions say so.

The examples on the homepage should also be fixed as the Jacoco one is no longer showing valid data and the Jenkins build badge example seem unnecessarily complicated.

PR #2154, opened 2018-10-07
[Jenkins] coverage badges rewrite

PyvesB:
This pull request addresses #2106.

Additionally both JaCoCo and Cobertura coverage badges have been migrated to the new service architecture.


services/jenkins/jenkins-coverage.service.js
```diff
@@ -1,80 +1,166 @@
...
+      schema: jacocoCoverageSchema,
+    })
+    return this.constructor.render({
+      coverage: json.instructionCoverage.percentage,
+    })
+  }
+
+  static get url() {
+    return {
+      base: 'jenkins/j',
```
PyvesB: The old badges supported an optional `-ci` in the URL: `/^\/jenkins(?:-ci)?\/(c|j)\/(http(?:s)?)\/([^/]+)\/(.+)\.(svg|png|gif|jpg|json)$/`. 

This was designed from the start in 6f616fc58d2c7aa6ea25292f45268839157bf9c7, but was never advertised in the homepage examples. Unless I'm missing something it's most likely that no one is using it, unless they've dug into the code and gone "cool, I can make my badges URL longer with an optional useless bit". Therefore I think we should drop that bit for consistency with the other Jenkins badges and also because I'm unsure how we would translate this possibility in the new architecture (would it entail making the base empty and moving it to the format parameter?).

services/jenkins/jenkins-coverage.service.js
```diff
@@ -1,80 +1,165 @@
...
+      schema,
+      errorMessages: {
+        404: 'job or coverage not found',
+      },
+    })
+  }
+
+  static render({ coverage }) {
+    return {
+      message: coverage.toFixed(0) + '%',
```
chris48s: You can use any js expression in a template string, so this could be `${coverage.toFixed(0)}%` if you wanted.

services/jenkins/jenkins-coverage.service.js
```diff
@@ -1,80 +1,165 @@
...
+  static get defaultBadgeData() {
+    return { label: 'coverage' }
+  }
+
+  static get category() {
+    return 'build'
+  }
+
+  static buildUrl(scheme, host, job, plugin) {
+    return job.indexOf('/') > -1
```
chris48s: I know this is copied from the legacy code, but what is the intent of `job.indexOf('/') > -1`? Is this actually an indirect check that is supposed to be working out if (the variable) `job` already contains (the string) `"/job"`? If so, shall we check for that explicitly, or catch it in the regex? If you test for it here you might also use `.includes()` instead of `indexOf()`.

services/jenkins/jenkins-coverage.service.js
chris48s: I did a code search on GitHub.

* `"https://img.shields.io/jenkins"`: 837 results
* `"https://img.shields.io/jenkins-ci"`: 0 results

I reckon we can probably bin it :wastebasket:

PyvesB:
@chris48s thanks for reviewing, I believe that I've addressed your two suggestions!

Issue #2117, opened 2018-09-25
Github tag not showing the correct value


Sopor-:
I have noticed that this
```https://img.shields.io/github/tag/osmc/osmc.svg?label=OSMC (Vero 4K)```
have been the same for a long time and when i check the project `https://github.com/osmc/osmc/tags` it is now at tag `2018.08-1`. It seems that it can't handle the `-`. That is the only reason i can find out...

I also found out that this is also showing the wrong tag, 3.51b:
```https://img.shields.io/github/tag/airdcpp/airgit.svg?label=AirDC```
The correct tag should be `3.51`

If i use the release it seems to show 3.51:
```https://img.shields.io/github/release/airdcpp/airgit.svg?label=AirDC```
But not all projects use `Releases`, only `Tags`.

chris48s:
I think both of these things are fundamentally because the badge sorts tags as semantic versions rather than chronologically.

* On the first one, the - is causing those tags to be treated as a pre-releases. If you call https://img.shields.io/github/tag-pre/osmc/osmc.svg it does render ![](https://img.shields.io/github/tag-pre/osmc/osmc.svg) because it includes pre-releases although I'm not really sure what the meaning of - is on a project which uses calendar versioning.
* With the second one, I think `3.51b` is sorting unexpectedly because `3.51b` isn't a valid semantic version.

More generally, there's several strange things going on here..

* We seem to be making the assumption that all GitHub tags have to be semantic versions, which is not true.
* If we receive a version that isn't a valid semantic version (which definitely will happen sometimes, given the above) we're attempting to sort it anyway, producing unexpected results.
* We don't document that in the help for the badge. FWIW, my initial expectation was that this would just show the chronologically most recent tag until I read the source code.
* `tag-pre` exists but there isn't an example for it on the home page.
* To include pre-releases we're inconsistently using `/tag-pre` for GH tags and `/releases/all` for GH releases.

There are probably a lot of users depending on the existing functionality (which assumes semver), but there are probably also some things we can do to make it clearer what is going on here or make things a bit more consistent. @RedSparr0w I'm conscious you did some work on this fairly recently to solve several other problems but I can't remember what they were. Do you have any thoughts on this one?

RedSparr0w:
>@RedSparr0w I'm conscious you did some work on this fairly recently to solve several other problems but I can't remember what they were

#1628 was recently merged which worked on the latest version function.

>Do you have any thoughts on this one?

I'm really not quite sure what the best way to handle this would be.
[...]

PR #1589, opened 2018-03-20
BaseService: Query params

paulmelnikow:
This is a fairly straightforward improvement to `BaseService` which provides support for query parameters, in parity with the functionality in old-style services

This functionality is necessary to port many of the existing services, and arose while I was working on this WIP branch: https://github.com/badges/shields/compare/master...paulmelnikow:rewrite-npm-typedefs

PR #2157, opened 2018-10-07
Allow [GitHub] tag badge to sort by date

chris48s:
Following on from https://github.com/badges/shields/issues/2117#issuecomment-425558353

* I've modified the `/release` badge so we can use `/release-pre` and changed the home pages example. The legacy `/all` suffix will still work (for backwards-compatibility) but its not documented.
* I haven't changed anything else on `/release`. Looking into it, we're using the `prerelease` key there rather than assuming semver so we don't have the same issue there that we do with tags.
* I've added a `tag-date` endpoint. I changed my mind and decided not to go with `sort=semver/sort=date` for 2 reasons:
    * It allows `/tag-pre/username/repo.svg?sort=date` which is a bit odd/unclear
    * We don't really have a good way to get the querystring in `BaseJsonService`. This is going to be a problem with a few other services which we need to deal with when we port them, but on reflection I'd prefer not make this problem bigger.

This PR won't change the behaviour for anyone already using these badges but improves consistency and provides a way for projects that don't use SemVer (for whatever reason) to use the `/tag` badge with sensible expected behaviour.

Does this seem reasonable?

### DIFF
--- a/services/github/github-release.service.js
+++ b/services/github/github-release.service.js
@@ -13,13 +13,13 @@ const {
 module.exports = class GithubRelease extends LegacyService {
   static registerLegacyRouteHandler({ camp, cache, githubApiProvider }) {
     camp.route(
-      /^\/github\/release\/([^/]+\/[^/]+)(?:\/(all))?\.(svg|png|gif|jpg|json)$/,
+      /^\/github\/release(-?pre)?\/([^/]+\/[^/]+)(?:\/(all))?\.(svg|png|gif|jpg|json)$/,
       cache((data, match, sendBadge, request) => {
-        const userRepo = match[1] // eg, qubyte/rubidium
-        const allReleases = match[2]
-        const format = match[3]
+        const includePre = Boolean(match[1]) || match[3] === 'all'
+        const userRepo = match[2] // eg, qubyte/rubidium
+        const format = match[4]
         let apiUrl = `/repos/${userRepo}/releases`
-        if (allReleases === undefined) {
+        if (!includePre) {
           apiUrl = apiUrl + '/latest'
         }
         const badgeData = getBadgeData('release', data)
@@ -33,7 +33,7 @@ module.exports = class GithubRelease extends LegacyService {
           }
           try {
             let data = JSON.parse(buffer)
-            if (allReleases === 'all') {
+            if (includePre) {
               data = data[0]
             }
             const version = data.tag_name
--- a/services/github/github-release.service.js
+++ b/services/github/github-release.service.js
@@ -13,7 +13,7 @@ const {
 module.exports = class GithubRelease extends LegacyService {
   static registerLegacyRouteHandler({ camp, cache, githubApiProvider }) {
     camp.route(
-      /^\/github\/release(-?pre)?\/([^/]+\/[^/]+)(?:\/(all))?\.(svg|png|gif|jpg|json)$/,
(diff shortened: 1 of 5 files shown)


services/github/github-tag.service.js
```diff
@@ -15,9 +15,10 @@ const {
 module.exports = class GithubTag extends LegacyService {
   static registerLegacyRouteHandler({ camp, cache, githubApiProvider }) {
     camp.route(
-      /^\/github\/tag(-?pre)?\/([^/]+)\/([^/]+)\.(svg|png|gif|jpg|json)$/,
+      /^\/github\/(tag-?pre|tag-date|tag)\/([^/]+)\/([^/]+)\.(svg|png|gif|jpg|json)$/,
```
RedSparr0w: Do you think it would be better to make the `-` required for `tag-pre`?
```diff
-/^\/github\/(tag-?pre|tag-date|tag)\/([^/]+)\/([^/]+)\.(svg|png|gif|jpg|json)$/
+/^\/github\/(tag-pre|tag-date|tag)\/([^/]+)\/([^/]+)\.(svg|png|gif|jpg|json)$/
```
or an alternative:
```regexp
/^\/github\/tag(?:-(pre|date))?\/([^/]+)\/([^/]+)\.(svg|png|gif|jpg|json)$/ (- required)
/^\/github\/tag(?:-?(pre|date))?\/([^/]+)\/([^/]+)\.(svg|png|gif|jpg|json)$/ (- optional)
```

services/github/github-release.service.js
```diff
@@ -13,13 +13,13 @@ const {
 module.exports = class GithubRelease extends LegacyService {
   static registerLegacyRouteHandler({ camp, cache, githubApiProvider }) {
     camp.route(
-      /^\/github\/release\/([^/]+\/[^/]+)(?:\/(all))?\.(svg|png|gif|jpg|json)$/,
+      /^\/github\/release(-?pre)?\/([^/]+\/[^/]+)(?:\/(all))?\.(svg|png|gif|jpg|json)$/,
```
RedSparr0w: Do you think it would be better to make the `-` required for `tag-pre`?
```diff
-/^\/github\/release(-?pre)?\/([^/]+\/[^/]+)(?:\/(all))?\.(svg|png|gif|jpg|json)$/,
+/^\/github\/release(-pre)?\/([^/]+\/[^/]+)(?:\/(all))?\.(svg|png|gif|jpg|json)$/,
```

services/github/github-tag.service.js
chris48s: I was keeping this to avoid making a breaking change:
https://github.com/badges/shields/blob/7ea540d3c67b2f98a97fd9651296c6fb717d528b/services/github/github-tag.service.js#L18

..but now you mention it, one of the issues here was we've never put an example of this on the home page so I doubt there is much usage in the wild. It was only added quite recently. Lets change it now :+1:

paulmelnikow:
Catching up on some reading!

> We don't really have a good way to get the querystring in `BaseJsonService`. This is going to be a problem with a few other services which we need to deal with when we port them, but on reflection I'd prefer not make this problem bigger.

Could you clarify what you mean? I added some query param support in #1589. The handling of examples isn't completely correct as we've seen, and it's not stripping the default parameters, though the core of the query params are working in the the (new-style) NPM badges.

chris48s:
Sorry - turns out I'm mistaken on this. I hadn't seen an example where we've used the `queryParams` arg to `handle()` before, but its good to know its there. That's one less thing to worry about :)

PR #2171, opened 2018-10-13
Fix: colorB override not working for [Github] Stars

### LINKED ISSUE #2170: colorB override not working for Github Stars
Are you experiencing this issue on shields.io or your own instance?
 - [x] shields.io
 - [ ] My instance

:beetle: **Description**

I'm trying to change color B for Github starts badge, and something went wrong.

:link: **Link to the badge**

https://img.shields.io/github/stars/badges/shields.svg: 
https://img.shields.io/github/stars/badges/shields.svg?colorB=brightgreen&colorA=red: 

:bulb: **Possible Solution**

https://github.com/badges/shields/blob/bedba47d777cb3b27467fa8b020ac4c5a0105fdd/services/github/github-stars.service.js#L35-L40

Lentil1016:
Fix #2170 
Hi. I believe this is a Bug fix. So I'm not sure whether I should add a test, but added it anyway.

### DIFF
--- a/services/github/github-stars.service.js
+++ b/services/github/github-stars.service.js
@@ -34,8 +34,7 @@ module.exports = class GithubStars extends LegacyService {
           }
           try {
             badgeData.text[1] = metric(JSON.parse(buffer).stargazers_count)
-            badgeData.colorscheme = null
-            badgeData.colorB = '#4183C4'
+            badgeData.colorscheme = 'blue'
             sendBadge(format, badgeData)
           } catch (e) {
             badgeData.text[1] = 'invalid'
--- a/services/github/github.tester.js
+++ b/services/github/github.tester.js
@@ -300,6 +300,26 @@ t.create('Stars (repo not found)')
     value: 'repo not found',
   })
 
+t.create('Stars (named color override)')
+  .get('/stars/badges/shields.json?colorB=yellow&style=_shields_test')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'stars',
+      value: Joi.string().regex(/^\w+$/),
+      colorB: Joi.equal(colorsB.yellow).required(),
+    })
+  )
+
+t.create('Stars (hex color override)')
+  .get('/stars/badges/shields.json?colorB=abcdef&style=_shields_test')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'stars',
+      value: Joi.string().regex(/^\w+$/),
+      colorB: Joi.equal('#abcdef').required(),
+    })
+  )
+
 t.create('Forks')
   .get('/forks/badges/shields.json')
   .expectJSONTypes(


Lentil1016:
@platan Sorry for bothering. I have handled my careless mistake in this PR, but I hit some wired phenomena when running test on CircleCI. Every time I rerun the test, there is a great chance that multiple timeout failure will be reported in `services-pr@node-latest` and `services-pr`. The most recently failed test case is:

- Github (open pull requests by label (raw)) `[ GET /issues-pr-raw/badges/shields/service-badge.json ]`
- Github (hit counter) `[ GET /search/torvalds/linux/goto.json ]`
- Github (hit counter for nonexistent repo) `[ GET /search/torvalds/not-linux/goto.json ]`

And:

1. There is a chance that some failure goes away at every rerunning.
2. These failure seems not quite relative to the fix I've done.
3. These failure case works fine on my machine with `curl`, not likely to exceed the time limit.
4. I've doubled the timeout values, but failure remain the same on CircleCI. Then I changed it back.
5. I checked the [build history](https://circleci.com/gh/badges/shields/tree/pull%2F2157) of #2157. He meet the same phenomena, but not as many times as I did.

I guess this is probably caused by my ignorance or mistake. Luckily it passed the last time.

platan:
@Lentil1016 thanks for this PR, for the latest improvements and for thorough description of problems with running tests on CI. 
There is a room for improvement for this issue. Currently we have 22 GitHub services (`.service.js` files in `services/github`) but only one file with service tests and all of them (near 120) have to pass. It is something we should improve.

chris48s:
Is there a reason to implement this manually in this badge? If we set `.colorscheme`, wouldn't this be handled elsewhere?

Lentil1016:
@chris48s Hi. Sorry for I can't understand that clearly, I'm not quite familiar with both JS and this project. It's that mean I should set `.colorscheme` instead of `.colorB` there? I've checked how others use '.colorscheme', and still have no clue how to use it there.

platan:
@chris48s Could you please explain more precisely how would you like to solve this? `BaseService` has `defaultBadgeData` which is used in `_makeBadgeData`. But `github-stars` is a legacy service which do not use `_makeBadgeData`. I think we have to handle this that way.

services/github/github-stars.service.js
```diff
@@ -35,7 +36,7 @@ module.exports = class GithubStars extends LegacyService {
           try {
             badgeData.text[1] = metric(JSON.parse(buffer).stargazers_count)
             badgeData.colorscheme = null
```
platan: This can be removed, `setBadgeColor` will unset `colorscheme`. 
```suggestion
```

Lentil1016:
Thanks. I think I've got the point. It's very kind of you.
Considering the previous commit I do set default value to `.colorB`, which makes `.colorscheme` default value trivial. I'll redo my commits.

platan:
@chris48s thanks for explanation. I just thought we don't want to change default color value (#4183C4), but is OK to change it to `blue` and it simplifies code.

services/github/github-stars.service.js
```diff
@@ -34,8 +35,8 @@ module.exports = class GithubStars extends LegacyService {
           }
           try {
             badgeData.text[1] = metric(JSON.parse(buffer).stargazers_count)
-            badgeData.colorscheme = null
-            badgeData.colorB = '#4183C4'
+            badgeData.colorscheme = 'blue'
+            setBadgeColor(badgeData, data.colorB)
```
platan: Please remove this line (`setBadgeColor(badgeData, data.colorB)`). It causes a problem when you do not override `colorB`. In this case `data.colorB` is not defined and `setBadgeColor` sets color to `red` 
https://github.com/badges/shields/blob/master/lib/badge-data.js#L64

Lentil1016: > Please remove this line (`setBadgeColor(badgeData, data.colorB)`). 

I'm not sure that's the way to fix this. Maybe I shouldn't set `.colorscheme` directly because `setBadgeColor()` will actually do that for me. I think the first version might be a better way. And I think to simplified the code I should remove `badgeData.colorscheme = null`.
```diff
-            badgeData.colorscheme = null
-            badgeData.colorB = '#4183C4'
+            setBadgeColor(badgeData, data.colorB || 'blue')
```

Lentil1016: 😓 I just realize I've misunderstood something. It's ashamed. Sorry.

platan:
@Lentil1016 now it looks good to me :-)
@chris48s could you please look at this small change again?

PR #2175, opened 2018-10-13
[bintray] test

Jowitam:
Hi! I've added basic tests for the bintray service.


Jowitam:
@chris48s thanks for your review. I added new test.

PR #2178, opened 2018-10-14
Precommit hook with Prettier and ESLint

platan:
This PR adds precommit hook which runs ESLint and Prettier against staged git files. Why is it a good idea: https://www.npmjs.com/package/lint-staged#why? It uses [lint-staged](https://www.npmjs.com/package/lint-staged) and [husky](https://www.npmjs.com/package/husky). I've added the autofix (`--fix`) option to ESLint, because a lot of problems can be fixed automatically (https://eslint.org/docs/rules/). 

What do you think about this?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,9 +121,6 @@ This project formats its source code using Prettier. The most enjoyable way to
 use Prettier is to let is format code for you when you save. You can [integrate
 it into your editor][integrate prettier].
 
-If for whatever reason you don't want to do that, you can run
-`npm run prettier` to format the code from the command line.
-
 [integrate prettier]: https://prettier.io/docs/en/editors.html
 
 ### Tests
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -121,9 +121,6 @@ This project formats its source code using Prettier. The most enjoyable way to
 use Prettier is to let is format code for you when you save. You can [integrate
 it into your editor][integrate prettier].
 
-If for whatever reason you don't want to do that, you can run
-`npm run prettier` to format the code from the command line.
-
```
chris48s: I think we should replace this with a sentence explaining that we run lint/prettier in a pre commit hook.

platan: Good idea. I've added a note. 
And thanks for the `git commit --no-verify`, I didn't know it :-)

platan:
I've resolved conflict in `package-lock.json`.

PR #2187, opened 2018-10-17
Add support for private [npm] packages

tjaneczko:
This PR adds support for private npm packages via an `npm_token` secret.  This allows anyone running their own shields.io service to create badges for their private packages.


chris48s:
Nice. Thanks for the contribution. Could you add a sentence to the self-hosting docs:
https://github.com/badges/shields/blob/master/doc/self-hosting.md#server-secrets
just so that there is something explaining this feature exists to self-hosting users :)

tjaneczko:
Will do!

PR #2194, opened 2018-10-19
[ctan] refactor service

pnowaczyk:
Refactor ctan service, fixing single licenses along the way (ie. https://img.shields.io/ctan/l/tex.svg from unknown -> knuth)


pnowaczyk:
sure thing! sorry I've missed this in first place

Issue #927, opened 2017-03-31
Service testing strategy


paulmelnikow:
I'd like to suggest a goal for the project: **automated tests of the vendor badges**.

As Thaddée mentioned in #411:

> Testing is of course imperfect (as in, manual). The CLI and backend engine have tests, but the server does not. What would be helpful however isn't tests here; it would be automated error reporting. We know that a list of badges _should_ work. Having a program that goes through them and reports what the vendor server sent back if the badge failed would be great. However, it should absolutely not complexify the existing logic of the server.

I agree that tests like these would be valuable for end users. They would detect API problems when they occur, rather than hoping/waiting someone will eventually show up and report them.

They'd be relatively simple to write. They would help badge developers , and push more of the testing responsibility from the reviewer to the contributor.

In PR reviews, I spend a significant amount of time manually testing various combinations of inputs. After requesting changes I really _should_ start all over again. Tests would make this faster and much more reliable.

I'm skilled enough to spot many common problems in this type of code, however unlike Thaddée I'm not an expert in this codebase. I don't have the eye to recognize all the problematic patterns that he would. I can't tell the code is obviously right, or obviously wrong code, just by looking at it.

Given how easy it is to write vendor badges, and what's sure to be endless growth of available repositories and tools, _PR review will remain a bottleneck_. With automatic tests, testing responsibility can be scaled out to a wide net of contributors. Writing tests for _existing vendors_ is another easy way for contributors to put love into the project, even those who are content with the existing feature set.

Some downsides:

1. The tests would be slow, due to slow external servers. We'd want to easily run a subset.
2. The tests would be somewhat flaky due to flaky external servers. When running the tests, we'd want to include some amount of retrying.
3. Since assertions depend on external servers and external data, they would need to be somewhat flexible. A valid regex for semver, e.g., or for a build which may be passing, failing, or errored. We risk making the assertions so complex that they allow false negatives: assertions which pass when they should fail.
4. They would cover something pretty fundamental that can’t be tested any other way: _changes to a vendor's API_.

There are a few important attributes of tests like these:

- Easy to read and write
- Simple assertions – no false negatives
- Retriable – so it’s possible to run automatically
- Non-invasive – they should not add complexity to the server

We could call these **integration tests**.

[...]

PR #2195, opened 2018-10-19
Run service tests on a given (remote) instance; test on [jetbrains]

platan:
This PR gives an option (`--url`) to run service tests on a given instance. It can be used to check if given instance is working correctly, e.g. after deployment. 
Moreover I had to add an option (`--skip-intercepted`) to skip tests which intercepts requests (by `intercept` or by `networkOff`). 
An example command:
```bash
 npm run test:services -- --only=jetbrains --url=https://shields-metrics1.now.sh --skip-intercepted
```
Is it useful? What do you think about it?


paulmelnikow:
I think this makes sense. It's one of the use cases I had in mind when I worked on #927. It could be a nicer way to run our nightly tests on staging.

> To check that an instance is working correctly after a deployment, wouldn't treating it as a black box by hitting it with a few badge requests be sufficient to assert that it is working properly?

Hehe, maybe not, though it's a good start, and it's basically what we do now. 🙃I might not wait for the entire suite to run, but hitting a few of the service tests in a reliable and automated way would make a nice post-deploy smoke tests.

Running the full suite exercises a lot of code and provides a lot of confidence things are working.

lib/test-config.js
```diff
@@ -1,5 +1,12 @@
 'use strict'
 
 module.exports = {
+  tested: {
+    // protocol from https://nodejs.org/docs/latest-v8.x/api/url.html#url_urlobject_protocol
+    // is followed by a colon
+    protocol: 'http:',
+    hostname: 'localhost',
+    port: 1111,
```
paulmelnikow: I see that this is trying to make things more DRY, though I wonder if it would be clearer to put the URL here, and also more consistent with the other URLs we configure. I also think this could be loaded from the environment, as it's a pretty unusual case.

Something like:

```js
liveServerUrl: process.env.LIVE_SERVER_URL
```

services/service-tester.js
```diff
@@ -87,14 +93,16 @@ class ServiceTester {
...
-  toss() {
+  toss(options) {
     const specs = this.specs
 
     const fn = this._only ? describe.only : describe
     // eslint-disable-next-line mocha/prefer-arrow-callback
     fn(this.title, function() {
       specs.forEach(spec => {
-        spec.toss()
+        if (!options.skipIntercepted || !spec.intercepted) {
```
paulmelnikow: It's a bit difficult to follow how `skipIntercepted` is getting passed around here. I'm not sure I have a better solution, though.

lib/service-test-runner/runner.js
```diff
@@ -16,6 +20,9 @@ class Runner {
    * Prepare the runner by loading up all the ServiceTester objects.
    */
   prepare() {
+    if (this.options.testedServerUrl) {
+      testConfig.tested = this.options.testedServerUrl
```
paulmelnikow: It would be better not to mutate the test config; perhaps better to set it from the environment inside `test-config`, as I mention below.

paulmelnikow:
Hey @platan, are you up for picking this up?

platan:
@paulmelnikow I will be able to look at this next week.

lib/test-config.js
platan: Good idea. Done.

lib/service-test-runner/runner.js
platan: I agree with you, done :-)

services/service-tester.js
platan: I used an environment variable to set this option via test-config and code looks better now in opinion. It's a separate commit and revert it if required https://github.com/badges/shields/pull/2195/commits/0974be49dfc5a3ae07053983849404f9d3febe4e

platan:
@paulmelnikow Thank you for the review. I've changed the code, diff is smaller and it should look better now.

PR #2196, opened 2018-10-19
use most permissive license color

pnowaczyk:
Hello, I'd like to propose a solution for a TODO left in `lib/licenses.js` 😄 
I think that since something is available under many licenses, it can be obtained under the most permissive one, thus we can set the badge color to that specific one.
What do you think? I'm open for discussion 😃

### DIFF
--- a/lib/licenses.js
+++ b/lib/licenses.js
@@ -62,8 +62,14 @@ Object.keys(licenseTypes).forEach(licenseType => {
   })
 })
 const defaultLicenseColor = 'lightgrey'
-const licenseToColor = spdxIds => {
-  const licenseType = spdxIds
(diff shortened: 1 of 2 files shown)


platan:
I closed and reopened this PR because WIP check hung. But it did not help at all.

pnowaczyk:
Any way I can help with that?

platan:
I changed PR title and it helped :-)
@chris48s @PyvesB have you noticed any problems with the WIP check before?

chris48s:
@platan  I've seen this problem a few times. This PR for example: https://github.com/badges/shields/pull/2200 (unless we've changed something before you read this). I had a couple of dependabot PRs incorrectly stuck on WIP as well.

Issue #1388, opened 2017-12-23
Release gh-badges to npm


felixfbecker:
The library readme on npm looks incredibly outdated and the examples in there don't work:
```ts
var badge = require('gh-badges');
badge({ text: [ "build", "passed" ], colorscheme: "green" },
  function(svg) {
    // svg is a String… of your badge.
  });
```
throws an error because `template` is not set (which is mentioned in the README on master).

paulmelnikow:
Yea, agreed, it would be a good idea to do a release. There are a few things that need doing:

- [ ] Update the documentation in the readme (and document the API in `doc/` if needed)
- [ ] Make sure the public interface has tests
- [ ] Make sure we're publishing the files we need
- [ ] Publish a 2.0 RC

felixfbecker:
What are the breaking changes for 2.0?

paulmelnikow:
1. `makeBadge` has a new interface which is synchronous.
2. There's a new text measurer that needs to be passed in (see #1390) which is also synchronous.
3. For consumers of the library, we probably should add a convenience function which memoizes the measurer and avoids boilerplate in code which consumes it.
4. The CLI should be the same.

felixfbecker:
How far away are we from this? Maybe we could get incremental improvements instead of blocking on all these changes if it delays the release so long? I.e. if there is another breaking change, just cut 3.0, version numbers are cheap.

stclairdaniel:
I'm curious about this too - 1.3.0 has a lot of security vulnerabilities and I'd like to upgrade.

chris48s:
Looking over the tasks, I don't mind having a look at this job. I assume when you say "make sure we're publishing the files we need", you mean ensuring this is right: https://github.com/badges/shields/blob/9b6ba776f004afff76ac0cc651e5b3c77a9c84e2/package.json#L101-L111

@paulmelnikow - if I can spend a bit of time to get the docs into shape, check the tests and test a package, do you have access to `npm publish` it?

chris48s:
I've submitted PR #2200 with some changes supporting this. It isn't an aspect of the project I've given any attention to before, but having done a bit of work on this, I think there are several other jobs we should do to improve the NPM package.

There is one issue I've raised in #2200 which I think we should consider before publishing a 2.0.0 release (should we force the user to make a choice of text measurer in order to generate a badge?). Additionally, these shouldn't be blockers to a `2.0.0` but we should think about them for future:

[...]

PR #2200, opened 2018-10-20
Give the NPM package some love

chris48s:
Following on from the discussion in #1388 I've done a bit of work to whip the NPM package into shape so we can do a `2.0.0-beta1` release. There is one additional issue I think we should consider before publishing a 2.0.0 release:

Is it useful to force the user to make a choice of text measurer in order to generate a badge? Would it be better to be opinionated on this and provide a simpler public interface using either `PDFKitTextMeasurer` or `QuickTextMeasurer` (at least as a default)? As a point of reference, the CLI uses `PDFKitTextMeasurer` only.


chris48s:
> I agree that we should hide `PDFKitTextMeasurer` and `QuickTextMeasurer` from the public interface.

OK - cheers. Given this is the only outstanding thing and will change the public interface and docs, lets tack that on to this PR.

Do you have a feel for whether it is sensible for the default (or possibly only) measurer to be `QuickTextMeasurer` or `PDFKitTextMeasurer`? We use `QuickTextMeasurer` on shields.io and `PDFKitTextMeasurer` in the CLI but I'm unclear on whether they are different for a good reason or because we changed it in one place and forgot to change it in the other.

Any thoughts?

PyvesB:
After looking into this some more, as far as I can tell, these are both custom classes that we've written in `text-measurer.js`. `QuickTextMeasurer` uses `PDFKitTextMeasurer` but adds an additional layer of caching. Therefore I don't think that saying one is fast and the other accurate is correct.

The CLI targets one-off generations, therefore it doesn't make sense to add the caching capabilities, it would only slow things down.

If users use this as part of a node module, they may or may not benefit from having caching depending on their use-case. However I think that they are most likely to write something that will call `makeBadge` several times and even if they don't, the performance impact of having to initialise and populate the cache will probably be negligible. Therefore I would go for `QuickTextMeasurer` as a default in the NPM module. We could also provide an alternative method or an optional `withoutCache` flag if users really want to override the default. 😉

chris48s:
> After looking into this some more, as far as I can tell, these are both custom classes that we've written in _text-measurer.js_. `QuickTextMeasurer` uses `PDFKitTextMeasurer` but adds an additional layer of kerning/caching. Therefore I don't think that saying one is fast and the other accurate is correct.

Aah yes I see you've used the classic "work out what's going on by actually reading the code" technique - cunning :)

My [initial thought on this](https://github.com/badges/shields/issues/1388#issuecomment-431593799) was that we should probably decouple the LRU cache from the badge generation (which would imply just using `PDFKitTextMeasurer`) but maybe that instinct is wrong.

> We could also provide an alternative method or an optional `withoutCache` flag

Where possible I prefer the boolean condition to be positive. i.e: `true` should enable the thing and `false` should diable it rather than using a `disableThing=true` param) so I'd tend to think of it the other way round: call the param `cache` and default it to `true`..  but yeah, agreed.

I'll update this tomorrow. Cheers.

chris48s:
How does this feel as a new public surface for the package?

README.md
```diff
@@ -88,28 +88,24 @@ badge build passed :green .png > mybadge.png
...
   text: ['build', 'passed'],
   colorscheme: 'green',
   template: 'flat',
 }
 
-badge.loadFont('/path/to/Verdana.ttf', err => {
-  badge(format, (svg, err) => {
-    // svg is a string containing your badge
-  })})
+const svg = bf.create(format)
```
felixfbecker: So this is sync now?

lib/gh-badges.js
```diff
@@ -0,0 +1,20 @@
...
+const { PDFKitTextMeasurer, QuickTextMeasurer } = require('./text-measurer')
+
+class BadgeFactory {
+  constructor({ fontPath, fallbackFontPath, cache = true }) {
+    this.measurer = cache
+      ? new QuickTextMeasurer(fontPath, fallbackFontPath)
+      : new PDFKitTextMeasurer(fontPath, fallbackFontPath)
+  }
+
+  create(format) {
```
felixfbecker: It would be nice to have a docblock here with `@param` and `@return` types

chris48s: Good suggestion. Added in cb94899

README.md
chris48s: Yep. As noted in https://github.com/badges/shields/issues/1388#issuecomment-354929354 its now synchronous

paulmelnikow:
Using `PDFKitTextMeasurer` will be faster for one-shot badge generation, since it only measures the two strings that are being used. On the other hand, `QuickTextMeasurer` builds the entire kerning table on startup, so it's slower. It would only be faster if you're generating hundreds of badges inside a single Node process, which I expect most people are not. (We should get rid of both of these things and just publish the lookup table, but that's a project for another day.)

lib/gh-badges.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+const { makeBadge } = require('./make-badge')
+const { PDFKitTextMeasurer, QuickTextMeasurer } = require('./text-measurer')
+
+class BadgeFactory {
+  constructor({ fontPath, fallbackFontPath, cache = true }) {
```
paulmelnikow: `QuickTextMeasurer` doesn't cache as much as it memoizes the most common inputs. Something like `precomputeWidths` would be more accurate.

lib/gh-badges.js
```diff
@@ -0,0 +1,33 @@
...
+   * Create a badge
+   *
+   * @param {object} format - Object specifying badge data
+   * @param {string[]} format.text
+   * @param {string} format.colorscheme
+   * @param {string} format.colorA
+   * @param {string} format.colorB
+   * @param {string} format.format
+   * @param {string} format.template
+   * @return {string} Badge in SVG or JSON format
```
paulmelnikow: I'd love to switch this over to the schema we're using for new-style services, though that could definitely happen another day.

lib/gh-badges.js
chris48s: Changed. Do you think this should default to `true` or `false`?

lib/gh-badges.js
chris48s: The DocBlock is helpful, because it allows editors/IDEs to show inline help, but we could definitely improve the validation. Lets save that for a minor release though

lib/gh-badges.js
paulmelnikow: `false`, since I’d wager the most common use will be generating a single badge.

Issue #1518, opened 2018-02-21
Some of Code Climate badges are not working


ghost:
Hi, guys! It seems to be an issue again/still.
Badges with percentage are not working.
Any info on that?

PyvesB:
Hello @voplex95!

I hopefully fixed all the badges in  #1387. Even though the pull request has been merged, it has not yet been pushed to production. 😉

kachkaev:
@PyvesB thanks for the fix! Is it in production now? It could be me, but I can't make my coverage badge working.

repo: https://codeclimate.com/github/gicentre/prettier-plugin-elm
badge: ![coverage](https://img.shields.io/codeclimate/coverage/github/gicentre/prettier-plugin-elm.svg?style=flat-square)

Example badges on [shields.io](https://shields.io) are also grey.

PyvesB:
@kachkaev unfortunately not. You can check out the fixed version on our [staging environment](https://shields-staging.herokuapp.com/), but we'll still have to wait a bit before getting it pushed to production. ^^

kachkaev:
Thanks for this info @PyvesB! When about are you planning to push the fix into production? Just curious 🙂 

Meanwhile, will it be OK if use a staging badge at `https://shields-staging.herokuapp.com/codeclimate/coverage/gicentre/prettier-plugin-elm.svg?style=flat-square`?

PyvesB:
I simply help maintaining the project, I'm afraid I'm not going to able to provide a more helpful answer. Let's ping @paulmelnikow and @espadrine , they may be able to shed some light on the next deployment plans. 😉

paulmelnikow:
See #1538!

I'd request that the staging server be restricted to testing. Since we're on the free plan, Heroku limits the number of hours per month the server can be used. If we exceed those with production usage, the staging server becomes unusable for testing, which is why it's there. That said, you're welcome to host your own server on Heroku!

PyvesB:
As far as I can tell, the Code Climate badges _should_ be working as expected. This is what I see when I visit the Shields homepage:

![codeclimate](https://user-images.githubusercontent.com/10694593/38572346-c78e92b0-3cea-11e8-8bc7-0574c80c0166.png)

I did submit a yet undeployed pull request to make our Code Climate code more reliable (#1613) and we are facing more general issues due to lack of capacity (see #1568), but I don't believe much else can be done right now.

DannyBen:
This is what I see (at the time of this comment):
[...]

PR #2201, opened 2018-10-20
Static Code Climate examples

PyvesB:
We've hit a wall in #1518, our Code Climate badges have been unreliable for months now.

This pull request aims at sparing some of our limited API quota by switching to static examples on the homepage.

I'm not willing to migrate the badges to the new service architecture at this point as we've already put in quite a lot of investigation and development effort in these badges. The least we can say is that @CodeClimate have not been very cooperative and have been dragging their feet since one of our users first contacted them back in April.

### DIFF
--- a/services/codeclimate/codeclimate.service.js
+++ b/services/codeclimate/codeclimate.service.js
@@ -140,7 +140,7 @@ module.exports = class Codeclimate extends LegacyService {
       base: 'codeclimate',
     }
   }
-  
+
   static get category() {
     return 'build'
   }
--- a/services/codeclimate/codeclimate.service.js
+++ b/services/codeclimate/codeclimate.service.js
@@ -140,6 +140,10 @@ module.exports = class Codeclimate extends LegacyService {
       base: 'codeclimate',
     }
   }
+  
+  static get category() {
+    return 'build'
+  }
 
   static get examples() {
     return [
--- a/services/codeclimate/codeclimate.service.js
+++ b/services/codeclimate/codeclimate.service.js
@@ -134,4 +134,63 @@ module.exports = class Codeclimate extends LegacyService {
       })
     )
   }
+
+  static get url() {
+    return {
+      base: 'codeclimate',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Code Climate issues',
+        exampleUrl: 'issues/twbs/bootstrap.svg',
+        urlPattern: 'issues/:userRepo.svg',
+        staticExample: { label: 'issues', message: '89', color: 'red' },
+      },
+      {
+        title: 'Code Climate maintainability',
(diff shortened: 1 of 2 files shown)


PyvesB:
I Have deployed a [review app](https://shields-staging-pr-2201.herokuapp.com), everything seems fine apart from the static examples not shwoing up. Am I being thick? 😄

chris48s:
I think if you want to use `staticExample` in `all-badge-examples.js`, you need to define it as a static badge URL instead of just a javascript object because its not going though any of this:

https://github.com/badges/shields/blob/95f1b50194a58f4fa11c235822442573005729e0/services/base.js#L90-L106

I've not checked if this works, but given `LegacyService` extends `BaseService`, you can probably move the examples out of `all-badge-examples.js` and into the service class without having to do a full refactor of the service class. You might also need to define `url()` in there so you've got a base..

PyvesB:
@chris48s thank you for your help! This piece of work should now be ready for review.

services/codeclimate/codeclimate.service.js
```diff
@@ -134,4 +134,67 @@ module.exports = class Codeclimate extends LegacyService {
...
+        },
+      },
+      {
+        title: 'Code Climate coverage',
+        exampleUrl: 'coverage/jekyll/jekyll',
+        urlPattern: 'coverage/:userRepo',
+        staticExample: {
+          label: 'coverage',
+          message: '95%',
+          color: 'yellowgreen',
```
chris48s: This should be green (or a lower value)

services/codeclimate/codeclimate.service.js
```diff
@@ -134,4 +134,67 @@ module.exports = class Codeclimate extends LegacyService {
...
+        },
+      },
+      {
+        title: 'Code Climate coverage (letter)',
+        exampleUrl: 'coverage-letter/jekyll/jekyll',
+        urlPattern: 'coverage-letter/:userRepo',
+        staticExample: { label: 'coverage', message: 'A', color: 'green' },
```
chris48s: this should be brightgreen (or a lower value)

services/codeclimate/codeclimate.service.js
```diff
@@ -134,4 +134,67 @@ module.exports = class Codeclimate extends LegacyService {
...
+      },
+      {
+        title: 'Code Climate technical debt',
+        exampleUrl: 'tech-debt/jekyll/jekyll',
+        urlPattern: 'tech-debt/:userRepo',
+        staticExample: {
+          label: 'technical debt',
+          message: '3%',
+          color: 'green',
```
chris48s: this should also be brightgreen (or a lower value)

PyvesB:
You're right, I got confused with the shades of green!

RedSparr0w:
Nice work, hopefully this reduces the API hits a fair amount.

PR #2204, opened 2018-10-21
refactor [depfu] service

chris48s:


services/depfu/depfu.tester.js
```diff
@@ -24,7 +24,7 @@ t.create('depfu dependencies (valid)')
 
 t.create('depfu dependencies (repo not found)')
   .get('/pyvesb/emptyrepo.json')
-  .expectJSON({ name: 'dependencies', value: 'invalid' })
+  .expectJSON({ name: 'dependencies', value: 'not found' })
 
 t.create('depfu dependencies (connection error)')
```
PyvesB: We usually remove this test from refactored services services if I'm not mistaken, as it's already tested elsewhere.

PR #2205, opened 2018-10-21
allow badge maxAge to be set by category; affects [discord]

chris48s:
This PR adds the ability for us to override `BADGE_MAX_AGE_SECONDS` with a different value for a category or a specific badge. This will allow us to set a long cache on badges that change infrequently (like license) and a shorter cache on badges that change more frequently (like build status). I've set some values in this PR:

license: 1 hour
version: 5 minutes
build: 30 seconds

as a starting point, but I'm not wedded to them. I'm more interested in feedback on the approach for now.

This only applies to the services we've already refactored, but as we refactor more of the codebase, this becomes an increasingly helpful feature :)

I've also shown an example in fe3552a of how we can use this feature to override the default in a specific badge. 2 mins cache is fine for most of the 'chat' badges, but the discord "how many users are online now" badge shows something more immediate so we might want to set a lower cache time on it.

One downside of this is that it becomes tricky to clearly explain the behaviour of the `maxAge` param:

https://github.com/badges/shields/blob/9df1da137c19d93e0d4e0a0a2526e807cddfbed4/frontend/components/usage.js#L267-L275

as potentially every badge could define its own default (lets not actually do that). That said, as we become better at caching on our side, I wonder if exposing that for users is still a valuable feature? :thinking:

### DIFF
--- a/services/base.js
+++ b/services/base.js
@@ -171,9 +171,7 @@ class BaseService {
       license: 3600,
       version: 300,
     }
-    return cacheLengths.hasOwnProperty(this.category)
-      ? cacheLengths[this.category]
-      : undefined
+    return cacheLengths[this.category]
   }
 
   static _namedParamsForMatch(match) {
--- a/services/base.js
+++ b/services/base.js
@@ -165,6 +165,19 @@ class BaseService {
     return new RegExp(fullRegex)
   }
 
+  static get _cacheLength() {
+    if (this.category === 'build') {
+      return 30
+    }
+    if (this.category === 'license') {
+      return 3600
+    }
+    if (this.category === 'version') {
+      return 300
+    }
+    return undefined
+  }
+
   static _namedParamsForMatch(match) {
     const names = this.url.capture || []
 
@@ -315,6 +328,7 @@ class BaseService {
           const format = match.slice(-1)[0]
           sendBadge(format, badgeData)
         },
+        cacheLength: this._cacheLength,
       })
     )
   }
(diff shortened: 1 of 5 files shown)


services/base.js
```diff
@@ -165,6 +165,19 @@ class BaseService {
     return new RegExp(fullRegex)
   }
 
+  static get _cacheLength() {
+    if (this.category === 'build') {
```
PyvesB: Would it be more readable to have a map category -> cache length? Keeping on adding `if` statements as we support more categories may become unpractical. `get _cacheLength()` would simply do a lookup in that map.

RedSparr0w:
Looks good to me! 👍

I think @PyvesB suggestion will definitely help future proof it though.

chris48s:
Cheers - good suggestions, both.

I reckon with the `maxAge` param, lets leave it there for now with @PyvesB 's suggested wording, but once we've refactored a larger chunk of the codebase we can probably just quietly remove that param. It is primarily there to allow users to manually work around a problem which we're now taking a more sophisticated approach to solving internally.

services/base.js
```diff
@@ -165,6 +165,17 @@ class BaseService {
     return new RegExp(fullRegex)
   }
 
+  static get _cacheLength() {
+    const cacheLengths = {
+      build: 30,
+      license: 3600,
+      version: 300,
+    }
+    return cacheLengths.hasOwnProperty(this.category)
```
PyvesB: Do we actually need this extra check? Wouldn't calling `cacheLengths[this.category]` for an unsupported category return `undefined` anyway?

Issue #2082, opened 2018-09-15
VSTS badge instructions no longer work for Azure DevOps


MichaelMauderer:
When trying to create a badge for Azure DevOps I was trying to follow the instructions for VSTS, but it seems that the ULR scheme has changed and the project ID and is no longer exposed (or a different name-based identifier is used instead).

Example for the current badge url scheme: 
```
https://colour-science.visualstudio.com/Colour/_apis/build/status/colour-science.colour?branchName=develop
                                                                  |--------------------|
                                                                  |----Project ID?-----|
```

chris48s:
Thanks for raising this. I've never used Visual Studio Team Services, so this is a bit hard for me to debug/advise. Working backwards from our example, it does seem to be valid to call a URL like:

https://larsbrinkhoff.visualstudio.com/953a34b9-5966-4923-a48a-c41874cfb5f5/_apis/build/status/1

@Thilas - you've recently made a useful contribution to the Visual Studio integrations which uses a similar URL pattern to extract data from visualstudio.com . Are you able to provide advice on how to find the correct 'project id' and 'build definition id' to make this badge?

Thilas:
To my understanding, the project guid is expected instead of its name when getting a badge from the "public" API (i.e. that does not require any authentication).

@MichaelMauderer, the url you mentioned is using the "private" API therefore accept the project name. But for the shield.io badge, you have to find the guid of your project which is the only one accepted here. You can get it using VSTS api from a browser where you already are authenticated:

````
https://colour-science.visualstudio.com/_apis/projects/Colour
```

Look at the `id` property.

FYI:
```
https://colour-science.visualstudio.com/Colour/_apis/build/status/colour-science.colour?branchName=develop
```

- "colour-science" = team name
- "Colour" = project name
- "colour-science.colour" = build name, useless here

MichaelMauderer:
Ah yes, that api call showed the correct id/giud. Thanks!

chris48s:
Thanks for helping out here @Thilas  :+1: 

Do you think we need to change the documentation?

This is the current help we provide: https://github.com/badges/shields/blob/f75cbf04c848d3488679e7fdc41d65384c7d86b3/lib/all-badge-examples.js#L5-L29

Thilas:
[...]

PR #2206, opened 2018-10-21
Rename VSTS into Azure DevOps

Thilas:
Per #2082, this PR aims at renaming VSTS into Azure DevOps. I also updated the service documentation so that it is easier to create badges for both builds and releases.

### DIFF
--- a/services/vso/vso.tester.js
+++ b/services/vso/vso.tester.js
@@ -2,7 +2,8 @@
 
 const Joi = require('joi')
 const ServiceTester = require('../service-tester')
-const { isBuildStatus } = require('../test-validators')
+
+const isValidStatus = Joi.equal('passing', 'failing')
 
 const t = new ServiceTester({
   id: 'vso',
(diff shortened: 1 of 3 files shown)


Thilas:
@chris48s, could you review this?

services/vso/vso.tester.js
```diff
@@ -42,37 +45,13 @@ t.create('build status with connection error')
 // Releases
 
 t.create('release status is succeeded')
-  .get('/release/devdiv/0bdbc590-a062-4c3f-b0f6-9383f67865ee/77/16.json')
```
chris48s: Is there a good reason for removing the mocked tests here instead of switching `nock` to mock the request to https://vsrm.dev.azure.com instead e.g:

```js
t.create('release status is succeeded')
  .get('/release/devdiv/0bdbc590-a062-4c3f-b0f6-9383f67865ee/77/16.json')
  .intercept(nock =>
    nock('https://vsrm.dev.azure.com')
      .get(
        '/devdiv/_apis/public/Release/badge/0bdbc590-a062-4c3f-b0f6-9383f67865ee/77/16'
      )
      .reply(200, '<svg><g><text>succeeded</text></g></svg>')
  )
  .expectJSON({ name: 'deployment', value: 'passing' })
```

and so on

(I've not tested that code, but hopefully you get the idea)

Thilas:
You get it right: the new endpoint is backward compatible with old VSO/VSTS projects. I made sure of this testing it on private projects (from my company).

services/vso/vso.tester.js
Thilas: Actually, when I introduced these "release" tests at first, someone asked me why I mocked them instead of doing the same thing as for "build" tests. So what I did here is to build an actual Azure DevOps project (see line 14: https://dev.azure.com/totodem/Shields.io) with a build pipeline and a release one in order to run integration tests.

It seems that's how tests are usually done here, doesn't it?

chris48s: Right, yep. Hadn't clocked there was no 'live' test for `/release`. Seems reasonable :)

PR #2215, opened 2018-10-24
Refactor [Swagger] Service

TagnumElite:
Also created `swagger.tester.js`

### DIFF
--- a/services/swagger/swagger.service.js
+++ b/services/swagger/swagger.service.js
@@ -1,67 +1,69 @@
 'use strict'
 
-const LegacyService = require('../legacy-service')
-const { makeBadgeData: getBadgeData } = require('../../lib/badge-data')
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
 
-// For a Swagger Validator.
-module.exports = class Swagger extends LegacyService {
-  static registerLegacyRouteHandler({ camp, cache }) {
-    camp.route(
-      /^\/swagger\/(valid)\/(2\.0)\/(https?)\/(.+)\.(svg|png|gif|jpg|json)$/,
-      cache((data, match, sendBadge, request) => {
-        // match[1] is not used                 // e.g. `valid` for validate
-        // match[2] is reserved for future use  // e.g. `2.0` for OpenAPI 2.0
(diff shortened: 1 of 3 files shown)


services/swagger/swagger.service.js
```diff
@@ -1,67 +1,69 @@
...
-              badgeData.colorscheme = 'red'
+    if (!valMessages || valMessages.length === 0) {
+      return this.constructor.render({ message: 'valid', clr: 'brightgreen' })
+    } else {
+      return this.constructor.render({ message: 'invalid', clr: 'red' })
+    }
+  }
+
+  async fetch({ protocol, urlF }) {
+    const url = `http://online.swagger.io/validator/debug?url=${protocol}://${urlF}`
```
PyvesB: Could we improve on the old implementation and use URL parameters here? Here's an [example](https://github.com/badges/shields/blob/99ec8fa239f517a0a6383c7d310ef145d953fad9/services/wercker/wercker.service.js#L31) (`qs` option).

services/swagger/swagger.service.js
```diff
@@ -1,67 +1,69 @@
...
-    )
+  static get examples() {
+    return [
+      {
+        title: 'Swagger Validator',
+        urlPattern: 'https/:url',
+        staticExample: this.render({ message: 'valid', clr: 'brightgreen' }),
+        exampleUrl:
+          'https/raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v2.0/json/petstore-expanded.json',
+        keywords: ['swagger'],
```
PyvesB: This keyword is actually redundant.

services/swagger/swagger.service.js
```diff
@@ -1,67 +1,69 @@
...
-            sendBadge(format, badgeData)
-          }
-        })
-      })
-    )
+  static get examples() {
+    return [
+      {
+        title: 'Swagger Validator',
+        urlPattern: 'https/:url',
```
PyvesB: In our Jenkins coverage integration, we refer to http/https as "scheme". Would it be a good idea to make things consistent?

PR #2222, opened 2018-10-28
Migrate babel 6.x to babel 7.1

piekar294:
Hi, I have upgrade babel to newest version.
Babel have moved their packages to new destination, so dependabot probably will not be able to bump this dependencies.

After switching to babel 7 the ```"next": "^5.0.0"``` was not compatible with new babel, because it requires babel-core in 6.23 version, which caused application start failure. I tested new dependencies locally and application looks running just fine.

I have also observed that after running ```npm i``` package-lock.json change a little more that I would expect. All ```"optional": true``` lines are removed from requirements. I use
```
npm: 6.4.1
node: v8.12.0
```
Have you encountered similar problem?

In the near future, I would like to add some test for frontend react component. I needed https://www.npmjs.com/package/babel-preset-react so I thought it will be more convenient, to migrate to new babel first, instead of adding old dependency. (First component test can be found in my other branch https://github.com/piekar294/shields/blob/enzyme/frontend/components/markup-modal.spec.js).

Let me know what you think about this change :)


paulmelnikow:
> Have you encountered similar problem?

Yes, I have encountered this, though I haven't determined which node / npm versions trigger it. I'm using npm 6.4.0 and Node 10.7.0.

chris48s:
As far as I can tell, the `"optional": true` lockfile churn happens when dependabot was the last user to commit `package-lock.json`. If the last user that touched it was a human, it doesn't seem to happen. I'm also not sure exactly why - I guess dependabot is running a different version of NPM or running it with different flags or something. I'm not deeply concerned though

paulmelnikow:
Thanks!

piekar294:
Great you found this change useful :)

PR #2229, opened 2018-10-29
Add BaseSvgScrapingService and rewrite [ReadTheDocs]; also affects [codacy vso]

paulmelnikow:
Based on discussion in #2031, this adds an abstract service for SVG badges. I started with Readthedocs and the other services can be done as a follow-on.

I called it **BaseSvgScrapingService** rather than **BaseSvgService** to clarify that it's for badges from svg source data – not svg badges, which is all the badges.

Since I don't expect the svg parsing function to be used anywhere else once the services are refactored, I moved it into the class. I added a default value for `valueMatcher`, which works on Shields-style badges and seems to be used more than once.

The tests are based on XmlBaseService. I added one for valueMatcher, and also moved the SVG parsing badge here. Testing on codacy + vso should ensure the old `fetchFromSvg` is still working.


services/base-svg-scraping.js
```diff
@@ -0,0 +1,42 @@
...
+        underlyingError: Error(`Can't get value from SVG:\n${svg}`),
+      })
+    }
+  }
+
+  async _requestSvg({ valueMatcher, url, options = {}, errorMessages = {} }) {
+    const logTrace = (...args) => trace.logTrace('fetch', ...args)
+    const { buffer } = await this._request({
+      url,
+      options,
```
chris48s: Do you think we should pass an `Accept: image/svg+xml` header here by default?

services/readthedocs/readthedocs.service.js
```diff
@@ -1,49 +1,68 @@
...
+      format: '([^/]+)(?:/(.+))?',
+      capture: ['project', 'version'],
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Read the Docs',
+        previewUrl: 'pip',
```
chris48s: It would be helpful to also define static examples for the index page here.

services/readthedocs/readthedocs.service.js
```diff
@@ -1,49 +1,68 @@
...
+      color,
+    }
+  }
+
+  async handle({ project, version }) {
+    const { message: status } = await this._requestSvg({
+      url: `https://readthedocs.org/projects/${encodeURIComponent(
+        project
+      )}/badge/`,
+    })
```
chris48s: The old badge code was passing also a version param here:

```js
if (version != null) {
  url += '?version=' + encodeURIComponent(version)
}
```

We seem to have lost that logic here. You could also move this out to a `fetch()` method rather than making the call directly in `handle()`

paulmelnikow: Good catch, thank you!

PR #2231, opened 2018-10-29
bump simple-icons version

### LINKED ISSUE #2223: Not all logos from simple-icons will work?
It is possible to use logos from different projects and even from [simple-icons](https://simpleicons.org). I have tested a lot of them on simple-icons and they work, but not [Lua](https://img.shields.io/badge/style-flat-green.svg?logo=lua) and that is the icon i want to use. I noticed that `SparkFun` is not working either.
Shouldn't all of them work?

RedSparr0w:
This PR attempts to fix #2223 
>The production server runs `npm install` every deploy, though it doesn’t respect package-lock.

I've changed the required version to `1.9.9` (removed the `^` char) as i assume dependabot will update the required version in `package.json` if it doesn't match the current pattern?

Another option could be `latest`?
But this could cause problems if/when `simple-icons` moves to `2.0.0`.

**Note:**
I was going to change it to use `~1.9.9`, but figured we would still end up behind every patch, and only end up updating every minor version, so removed the tilde aswell.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "redis": "~2.8.0",
     "request": "~2.88.0",
     "semver": "~5.6.0",
-    "simple-icons": "^1.7.1",
+    "simple-icons": "~1.9.9",
     "svgo": "~1.1.1",
     "xml2js": "~0.4.16",
     "xmldom": "~0.1.27",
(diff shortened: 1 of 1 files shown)


RedSparr0w:
**Note:**

> Warnings
> ⚠️	
> This PR modified `package.json`, but not `package-lock.json` - Perhaps you need to run `npm install`?

Already updated via dependabot #2216

paulmelnikow:
Think this is a good solution. Pinning specific versions in package.json may be a partial workaround for production not respecting package-lock.

PR #2232, opened 2018-10-29
Refactor [Jetbrains] service; affects [eclipse-marketplace]

platan:
Jetbrains service gets data from XML. Downloads value is stored in an attribute. [fast-xml-parser](https://github.com/NaturalIntelligence/fast-xml-parser#readme) ignores attributes by default. Version value is stored in an element, but all elements are parsed by default (`Parse the value of text node to float, integer, or boolean.`) - this causes a problem if version value is `1.0` - it's parsed to `1`. I had to change default parser options for this service. I've added an option `parserOptions` in `_requestXml` to pass parser options. 
I've left service tests in one file - it should be easier to review changes (in a separate PR I can split them). 
Do you have an idea how to test `parserOptions` in unit test (`base-xml.spec.js`)?


PyvesB:
@TagnumElite you've recently submitted a few service refactoring pull requests. Would you like to tackle a review and have a look at the work done by @platan? No pressure though, just ping me if you're not comfortable doing this just yet or if you're short on time at the moment! 😉

TagnumElite:
I would like to, however, my exams have just started so right now till 20th
November is a bad time for anything. I will only have available time over
the weekend. I do also live in South Africa, so my late responses may come
from differentiating timezones.

PyvesB:
No worries, please focus on your exams and good luck with them! I'll tackle this review myself in the coming days. 👍

services/jetbrains/jetbrains-base.js
```diff
@@ -0,0 +1,33 @@
...
+      ignoreAttributes: false,
+    }
+    const pluginData = await this._requestXml({
+      schema: Joi.object({
+        'plugin-repository': Joi.any().required(),
+      }),
+      url: `https://plugins.jetbrains.com/plugins/list?pluginId=${pluginId}`,
+      parserOptions,
+    })
+    if (!pluginData['plugin-repository']) {
```
PyvesB: Why is this being checked manually? Is the response code returned by JetBrains unhelpful? If that's the case, I would suggest using an approach similar to the _aur_ service, i.e. override `_validate`, check the data for not found, then call the parent's `_validate` method. This will make things clearer and will avoid having different Joi schemas being used in different parts of the implementation.

services/jetbrains/jetbrains.tester.js
```diff
@@ -121,7 +142,20 @@ t.create('empty XML')
...
+      nock('https://plugins.jetbrains.com')
+        .get('/plugins/list?pluginId=9435')
+        .reply(200, '<?xml version="1.0" encoding="UTF-8"?><plugin />'),
+    {
+      'Content-Type': 'text/xml;charset=UTF-8',
+    }
+  )
+  .expectJSON({ name: 'downloads', value: 'invalid response data' })
 
 t.create('404 status code')
```
PyvesB: This test (and probably others) are no longer relevant as they're now covered by the parent classes.

services/base-xml.js
```diff
@@ -26,7 +32,7 @@ class BaseXmlService extends BaseService {
         underlyingError: validateResult.err,
       })
     }
-    const xml = fastXmlParser.parse(buffer)
+    const xml = fastXmlParser.parse(buffer, parserOptions)
```
PyvesB: As _base-xml.spec.js_ uses real XML fixtures, it should be fairly easy to assert that the parser options are correctly provided to _fastXmlParser_ by writing a test that would fail with its default options. 😉

platan: Good idea :-) I've added a test.

services/jetbrains/jetbrains.tester.js
platan: I agree that the parent class is responsible for checking response data against a schema and this is tested somewhere. But with all tests with `value: 'invalid response data'` I want to verify that schema is defined properly, e.g. `required()` set on required elements ;-) 
If I remove one of the `required()` in the schema this test will fail with `TypeError: Cannot read property 'category' of undefined`
I would like to test schema. Of course it can be done in another way. What do you think about this?

PyvesB:
Can this be merged @platan ? 😉

platan:
@PyvesB I was waiting for a second review :-) Is it OK to merge it with one review only?

services/jetbrains/jetbrains-downloads.service.js
```diff
@@ -0,0 +1,60 @@
...
+  'plugin-repository': Joi.object({
+    category: Joi.object({
+      'idea-plugin': Joi.array()
+        .min(1)
+        .items(
+          Joi.object({
+            '@_downloads': nonNegativeInteger,
+          })
+        )
+        .single()
```
paulmelnikow: Is the single() needed here?

platan: When plugin has several versions (releases) `idea-plugin` is an array. If it has only one version - `idea-plugin` is an object. I use `single()` https://github.com/hapijs/joi/blob/v14.0.4/API.md#arraysingleenabled to allow single values and convert them to an array. We use `single()` in https://github.com/badges/shields/blob/d0fe97d136bbe981593c4821af44fa809eb14a86/services/ctan/ctan.service.js#L11 as well. In that case `license` is returned as a string or as an array.

services/hackage/1.json
```diff
@@ -0,0 +1 @@
+{ "plugin-repository": "" }
```
paulmelnikow: Oops, I didn't notice this file sneak in!

PR #2233, opened 2018-10-29
Readme examples

PyvesB:
* removed build status as the real badge equivalent is displayed just above in the Readme.
* replaced Code Climate by Codacy as the badges for the former service are unreliable (#1518). Also, as Code Climate are not being very cooperative there is no point in giving them free advertising.
* replaced Gratipay by Liberapay, its maintained fork.
* added three new examples which showcase different usages of our badges.

I'm wondering though, why do we not insert real badge examples here? With a high `maxAge` parameter and static badges, it would probably put little extra load on the server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,16 +42,16 @@ specification.
 Examples
 --------
 
-* code coverage percentage: `coverage | 80%`
-* stable release version: `version | 1.2.3`
-* package manager release: `gem | 1.2.3`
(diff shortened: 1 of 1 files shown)


chris48s:
I'm :+1: on replacing them with static badge examples

PyvesB:
The result can be viewed [here](https://github.com/PyvesB/shields/tree/readme-examples).

PR #2234, opened 2018-10-30
Move build badge examples into services/

paulmelnikow:
all-badge-examples is a common cause of merge conflicts. It’s difficult to adjust the badge categorization in that file – or to understand the diff – because it requires moving a block from one point to another. It’s much easier to edit a badge’s category in one place.

This starts the process of breaking up what’s left of that file, following up on the work from #1931. New-style services can only be in one category, which means legacy service examples have to be split along category lines. I split out separate legacy service classes where I could do so easily, leaving behind the ones which require more work, for one reason or another.

I did this manually, which wasn't as bad as I thought. Doing it one category at a time made it easier. Also someone must have done some work to harmonize the keys. 😁(That wasn't I, was it?)

Anyway, these were done manually, so I'll check them on the deploy app. If someone else wants to do that too, an extra set of eyes wouldn't hurt!


services/docker/docker.service.js
```diff
@@ -80,7 +80,36 @@ module.exports = class Docker extends LegacyService {
         })
       })
     )
+  }
+}
```
paulmelnikow: I split this up because new-style services can only be in one category, which means legacy service examples have to be split along category lines.

chris48s:
> the other part is putting the .svg before the query string

Aah yeah - that'll be it. We probably need to be able to pass a `queryString` in the example object to get round that. In general we don't really have good handling for badges that use queryString params in the new base class architecture - there will need to be some extra base class work before we can port those. If those ones are the only things broken, we could move those examples back to `all-badge-examples` for now and deal with that as another issue.

paulmelnikow:
I don't think moving them back to all-badge-examples will solve the problem. As far as I can tell, dockbit was broken before the move.

I'm working on a little refactor of `prepareExamples` toward solving this problem.

We'll be able to change the schema more easily once we empty out all-badge-examples – looking forward to that!

paulmelnikow:
Ah, after running the automated test I see dockbit and bitrise got _more_ broken by the move. I'll put them back.

paulmelnikow:
I used your script to compare the previewUrls in master and in the branch. This is the diff:

```
124,125d123
< /bundlephobia/min/react.svg
< /bundlephobia/minzip/react.svg
175c173
< /david/babel/babel.svg?path=packages/babel-core
---
> /david/babel/babel.svg?path=packages%2Fbabel-core
367c365
< /twitter/follow/espadrine.svg?style=social&label=Follow
---
> /twitter/follow/espadrine.svg?label=Follow&style=social
383c381
< 382
---
> 380
```

- bundlephobia I removed on purpose, because the examples had been duplicated under size and version.
- david is arguably more correct now
- twitter the order doesn't really matter

I repeated the test for `urlPattern` and `exampleUrl` and everything lines up.

This should be good to go!

Issue #2027, opened 2018-08-30
Consolidate URL encoding functions


PyvesB:
We have a number of functions encoding URLs, we should probably refactor things to have one that covers all cases and that is properly tested.

See https://github.com/badges/shields/pull/1965#discussion_r213100696.

PR #2240, opened 2018-10-31
Remove some duplicated URL generation code

paulmelnikow:
I went down a rabbit hole while trying to untangle the bug in the dockbit and bitrise examples https://github.com/badges/shields/pull/2234#pullrequestreview-169997546.

The URL generation code is spaghetti-like, with functions, many of which I wrote, with opaque names, doing similar but not identical things, and making slightly incompatible assumptions about the way query strings are handled.

I got a bit lost and need to take a step back.

Meanwhile, this is a small piece of work I did that’s worth keeping. It doesn’t scratch the surface of the tangle, but it does remove a bit of duplication.

It also makes a minor stylistic ES6 change in the handling of default arguments.

Ref: #2027


frontend/lib/badge-url.js
```diff
@@ -1,7 +1,11 @@
 import resolveUrl from './resolve-url'
+import { staticBadgeUrl as makeStaticBadgeUrl } from '../../lib/make-badge-url'
 
-export default function resolveBadgeUrl(url, baseUrl, options) {
-  const { longCache, style, queryParams: inQueryParams } = options || {}
+export default function resolveBadgeUrl(
+  url,
+  baseUrl,
+  { longCache, style, queryParams: inQueryParams } = {}
+) {
```
paulmelnikow: ^^ Use default arguments

frontend/lib/badge-url.js
```diff
@@ -12,13 +16,9 @@ export default function resolveBadgeUrl(url, baseUrl, options) {
   return resolveUrl(url, baseUrl, outQueryParams)
 }
 
-export function encodeField(s) {
-  return encodeURIComponent(s.replace(/-/g, '--').replace(/_/g, '__'))
-}
-
-export function staticBadgeUrl(baseUrl, subject, status, color, options) {
-  const path = [subject, status, color].map(encodeField).join('-')
-  return resolveUrl(`/badge/${path}.svg`, baseUrl, options)
```
paulmelnikow: Move this logic from `frontend/lib/` to `lib/` ^^

frontend/lib/badge-url.js
```diff
@@ -12,13 +16,9 @@ export default function resolveBadgeUrl(url, baseUrl, options) {
...
+export function staticBadgeUrl(baseUrl, label, message, color, options) {
+  const path = makeStaticBadgeUrl({ label, message, color })
+  return resolveUrl(path, baseUrl, options)
```
paulmelnikow: And delegate to the new function, while keeping the old options handling for now.

frontend/lib/badge-url.js
```diff
@@ -28,10 +28,8 @@ export function dynamicBadgeUrl(
   label,
   dataUrl,
   query,
-  options = {}
+  { prefix, suffix, color, queryParams = {}, ...rest } = {}
 ) {
-  const { prefix, suffix, color, queryParams = {}, ...rest } = options
```
paulmelnikow: ^^ Use default arguments

frontend/lib/badge-url.spec.js
```diff
@@ -27,14 +23,6 @@ describe('Badge URL functions', function() {
...
   })
 
-  test(encodeField, () => {
-    given('foo').expect('foo')
-    given('').expect('')
-    given('happy go lucky').expect('happy%20go%20lucky')
-    given('do-right').expect('do--right')
-    given('it_is_a_snake').expect('it__is__a__snake')
-  })
-
```
paulmelnikow: Moved ^^

Parity with before vv

lib/make-badge-url.js
```diff
@@ -0,0 +1,30 @@
...
+}) {
+  if (!label || !message) {
+    throw Error('label and message are required')
+  }
+  const path = [label, message, color].map(encodeField).join('-')
+  const outQueryString = queryString.stringify({
+    style,
+  })
+  const suffix = outQueryString ? `?${outQueryString}` : ''
+  return `/badge/${path}.svg${suffix}`
```
chris48s: We could make this slightly more general by accepting a format param (which also accepts `.gif`, `.png` etc) with a default of `.svg` if you wanted.

paulmelnikow: Sounds good.

PR #2241, opened 2018-10-31
Enforce using async-await [f-droid]

paulmelnikow:
Re: #2028 

Note: Lint will fail until #2029 is merged.


niccokunzmann:
Interesting. I learned something from this: "await"

paulmelnikow:
This will conflict with #2242. Let's merge that in first.

PR #2242, opened 2018-10-31
Enforce use of template literals

paulmelnikow:
This is consistent with what we're pretty much already doing, and saves us from making the request during code review.

These were all autofixed and most of them seem easier to read. Some in the legacy services should be rewritten in more legible forms during refactor (ie using intermediate variables, or using request’s qs option). There are some in helper functions and elsewhere that should get rewritten separately. I don't want to change them in this PR because the changes will get lost in this diff, though we could identify them here and fix them before or just after.

One surprising but common conversion is from

```
'' + foo
```

to

```
`${foo}`
```

🤔


paulmelnikow:
> `'' + foo` is a probably a quick n dirty cast to string?

Yea, for sure! The `${foo}` construction I'd seen once before, but didn't realize it was doing the same thing.

Issue #749, opened 2016-07-18
Request: 'Time passed since'  shield


majkinetor:
It would be awesome to have a way to get how much time is passed since argument, for instance

```
[age | 1 hour]  
[age | 3 years]

or

[published before | 1 minute]
[last build before | 1 day]
```

URL Example:

```
https://img.shields.io/badge/age-2016_07_18_T_15_05_30-red.svg
```

GammaNu:
i'm looking for similar behaviors.
Github badge for : 
- time since last commit (how alive/dynamic is the project)
- time since first commit (how old/mature is the project)

Ang-YC:
@RedSparr0w @chris48s I would like to ask if this issue need any PR and my suggestion for the endpoint for this issue is
`https://img.shields.io/badge/date/<SUBJECT>-<TIMESTAMP>-<COLOR>.svg`

Example
`https://img.shields.io/badge/date/christmas-1545696000-red.svg`
![](https://img.shields.io/badge/christmas-in%2055%20days-red.svg)

I will be using `moment` which is already a dependency.
If the timestamp is 2 minutes before now, return `2 minutes ago`.
If the timestamp is 2 minutes after now, return `in 2 minutes`.

Ang-YC:
Tagging @paulmelnikow as well

paulmelnikow:
Hi, yes, and thanks for taking this on! What you’re suggesting sounds good.

We’ve been moving away from APIs with dashes like the static badge, so let’s model on the dynamic badge instead:

https://img.shields.io/date/1545696000.svg?label=christmas&colorB=red

PR #2244, opened 2018-10-31
[Date] Relative date badge

Ang-YC:
This pull request is related to issue #749

| | |
|---|---|
| **Endpoint** | `/date/{timestamp}.svg` |
| **Regex** | `/^\/date\/([0-9]+)\.(svg\|png\|gif\|jpg\|json)$/` |
| **Example** | `/date/1545696000000.svg?label=christmas&colorB=red` |
| **Result** | revision |


services/date/date.service.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Relative date',
+        urlPattern: ':timestamp',
+        staticExample: this.render({ relativeDateString: '2 days ago' }),
+        exampleUrl: '1540814400000',
```
chris48s: I found it confusing that we're using a 13 digit (no of milliseconds) timestamp here instead of the more usual 10 digit unix format (no of seconds). We probably don't need millisecond precision for this application, so it might be worth going with the 10 digit format. What do you think?
Either way, we should probably pass a `'documentation'` key here which explains which format we use and maybe link to a useful converter.

chris48s: Sorry.. just realised that comment about a documentation key might not have been clear. There's an example of this here - we define a `docs` var here:

https://github.com/badges/shields/blob/52d642cf91d8587a3d664c960c10c173486ab9a5/services/steam/steam-workshop.service.js#L10-L22

and then pass it in a key called `'documentation'` when we define the examples:

https://github.com/badges/shields/blob/52d642cf91d8587a3d664c960c10c173486ab9a5/services/steam/steam-workshop.service.js#L181-L188

that renders it in the modal popup from the index page :)

RedSparr0w: I feel like `ms` may make more sense as we are parsing it straight to `moment.js` rather than using `time * 1000` or `time + '000'`,
Although the end user wouldn't necessarily know we are using javascript.

paulmelnikow:
@Ang-YC It would be great to get this feature shipped! Would you be up for addressing these comments? If not, no problem, I or one of the other maintainers can pick it up.

Ang-YC:
Hey @paulmelnikow, thanks for reminding me about this. Sure I will update it to follow their comments. However, should I use 10 or 13 digits timestamp as mentioned by @chris48s? Thanks!

services/date/date.service.js
chris48s: Yeah. We should definitely support the 10-digit format. If you wanted, we could do both. If we declare this route:

```js
static get route() {
  return {
    base: 'date',
    format: '([0-9]{13}|[0-9]{10})',
    capture: ['timestamp'],
  }
}
```

we'll capture either a 10 or 13-digit timestamp. Then we can process the timestamp appropriately based on length and we'll reject invalid formats. Everyone's happy.

paulmelnikow:
Hi @Ang-YC did you see the response above? Let's support both, so a user can provide either 10 or 13. Would be great to get this merged!

paulmelnikow:
https://shields-staging-pr-2244.herokuapp.com/date/0.svg?label=epoch

PR #2252, opened 2018-11-01
Rewrite vso, rename to [AzureDevops], validate SVG [readthedocs]

paulmelnikow:
This PR does a couple things:

1. Add validation to BaseSvgScrapingService and update readthedocs accordingly.
2. Rewrite vso and add more tests. Rename it internally to azure-devops. URLs are still `/vso` for now. Should we make a way to let a service register multiple URL patterns?
3. Handle shared code using a functional pattern instead of inheritance. This comes from a discussion https://github.com/badges/shields/pull/2031#issuecomment-417893819. I like the functional approach because it's more direct, nimble, and easy to reason about; plus it allows services to grow from a family of one to two more easily.


services/azure-devops/azure-devops-build.tester.js
```diff
@@ -0,0 +1,48 @@
...
+    Joi.object().keys({
+      name: 'build',
+      value: isBuildStatus,
+    })
+  )
+
+// t.create('default branch (alternate project)')
+//   .get('/larsbrinkhoff/953a34b9-5966-4923-a48a-c41874cfb5f5/1.json')
+//   .expectJSONTypes(
+//     Joi.object().keys({
```
chris48s: Is this supposed to be commented out

chris48s:
> Should we make a way to let a service register multiple URL patterns?

The way we've handled this on other badges is to stick the alternatives in a non-capturing group

e.g: https://github.com/badges/shields/blob/4a298cbcb0dc5497aeee23a75b21ac9f237320d7/services/wordpress/wordpress-rating.service.js#L81-L87

so you could use `(?:vso|azure-devops)` to 'rebrand' the URLs `/azure-devops` but still allow `/vso` for backwards compatibility.

> Handle shared code using a functional pattern instead of inheritance.

Seems reasonable :+1: Its a useful pattern to add to the toolbelt. I suspect its probably going to be most useful for larger service families where there is commonality but also variance as opposed to a rigid parent/child structure. I wouldn't particulary want to mandate style here though. Lets leave composition vs inheritance up to the implementer.

services/azure-devops/azure-devops-build.tester.js
paulmelnikow: I guess it's not really necessary. I'll remove it. This other project (larsbrinkoff/953a…) is used below in some of the other tests. It doesn't work to have a backup project on hand. 😉

paulmelnikow:
> I suspect its probably going to be most useful for larger service families where there is commonality but also variance as opposed to a rigid parent/child structure.

Could be true. I could also see it being valuable for single-service families which grow larger over time.

services/azure-devops/azure-devops-build.service.js
```diff
@@ -0,0 +1,76 @@
...
+      capture: ['organization', 'projectId', 'definitionId', 'branch'],
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Azure DevOps builds',
+        previewUrl:
+          'azure-devops/build/totodem/8cf3ec0e-d0c2-4fcd-8206-ad204f254a96/2',
```
chris48s: I didn't notice the first time, but it would also be useful to replace the examples with static badges as part of this refactor

paulmelnikow:
Thanks for reviewing!

PR #2253, opened 2018-11-01
Start on the Github rewrite, with [GithubPullRequestCheckState]

paulmelnikow:
The GitHub service family is the largest, and as yet untouched by our service rewrite. I thought I would start the process by tackling one service.

This pull request has a few things going on:

1. Rename pull-request-status to pull-request-check-state. We have another badge called pull request status. It seems like the checks are called one thing in the UI and another thing in the API, which is unfortunate. If other folks have strong feelings about the name, I’ll defer.
2. Move its tests and tighten up the syntax.
3. Move its badge examples including the doc string.
4. Add a new helper `errorMessagesFor` to use in the new services in place of `githubCheckErrorResponse`. It seems like we didn’t really use the `errorMessages` parameter to `githubCheckErrorResponse`, so I pared this down. I’m not sure if this is the function we’ll ultimately want, but it seems like a good place to start.
5. Pull fetch code I _know_ we use in other places into `github-common-fetch`. As in the PR I just opened for azure-devops, this takes a functional approach to the shared code, which is more direct, nimble, and easy to reason about than inheritance.
6. Create `GithubAuthService` which functions identically to BaseJsonService, except for one thing, which is that it uses the token pool. I accomplished this by adding a `_requestFetcher` property to BaseService, which is initialized to `sendAndCacheRequest` in the constructor, and can be overridden in subclasses. Since we weren’t using `_sendAndCacheRequest` directly except in BaseService and tests, I removed that property. I like this approach to patching in the GitHub auth because it’s very simple and creates no new API exposure. However, the way we’re doing the dependency injection feels a bit odd. Maybe the eventual refactor of request-handler would be a godo time to revisit this.

The GitHub requests go through many, many layers of indirection at this point. Later on it would be good to shave some of these off, perhaps once the legacy GitHub services have been converted, or when all the services are done and we can take another look at the base service hierarchy. The work in #2021 and #1205 is also related.

### DIFF
--- a/services/github/github-pull-request-check-state.service.js
+++ b/services/github/github-pull-request-check-state.service.js
@@ -22,11 +22,10 @@ module.exports = class GithubPullRequestCheckState extends GithubAuthService {
     return 'other'
   }
 
-  static get url() {
+  static get route() {
     return {
       base: 'github/status',
-      format: '(s|contexts)/pulls/([^/]+)/([^/]+)/(\\d+)',
-      capture: ['which', 'user', 'repo', 'number'],
+      pattern: ':which(s|contexts)/pulls/:user/:repo/:number(\\d+)',
     }
   }
 


services/github/github-auth-service.js
```diff
@@ -0,0 +1,18 @@
...
+const BaseJsonService = require('../base-json')
+
+module.exports = class GithubAuthService extends BaseJsonService {
+  constructor(context, config) {
+    super(context, config)
+
+    const { sendAndCacheRequestWithCallbacks, githubApiProvider } = context
+
+    this._requestFetcher = async (url, query) =>
+      githubApiProvider.requestAsPromise(
```
chris48s: :+1: nice

services/github/github-pull-request-check-state.service.js
```diff
@@ -0,0 +1,90 @@
...
+    return [
+      {
+        title: 'GitHub pull request check state',
+        previewUrl: 's/pulls/badges/shields/1110',
+        keywords: ['GitHub', 'pullrequest', 'detail', 'check'],
+        documentation,
+      },
+      {
+        title: 'GitHub pull request check contexts',
+        previewUrl: 'contexts/pulls/badges/shields/1110',
```
chris48s: Lets convert these to static examples

services/github/github-pull-request-check-state.tester.js
```diff
@@ -0,0 +1,16 @@
...
+  .get('/s/pulls/badges/shields/1110.json')
+  .expectJSON({ name: 'checks', value: 'failure' })
+
+t.create('github pull request check state (pull request not found)')
+  .get('/s/pulls/badges/shields/5110.json')
+  .expectJSON({ name: 'checks', value: 'pull request or repo not found' })
+
+t.create('github pull request check contexts')
+  .get('/contexts/pulls/badges/shields/1110.json')
+  .expectJSON({ name: 'checks', value: '1 failure' })
```
chris48s: Could we also add a (mocked) test for the 'commit not found' case.
https://github.com/badges/shields/blob/7a664ca3e845c840fa299fa1ee1ee0e787593038/services/codeclimate/codeclimate.tester.js#L127-L138
is probably a good example to draw on.

services/github/github-pull-request-status.service.js
```diff
@@ -1,85 +0,0 @@
...
-module.exports = class GithubPullRequestStatus extends LegacyService {
-  static registerLegacyRouteHandler({ camp, cache, githubApiProvider }) {
-    camp.route(
-      /^\/github\/status\/(s|contexts)\/pulls\/([^/]+)\/([^/]+)\/(\d+)\.(svg|png|gif|jpg|json)$/,
-      cache((queryParams, match, sendBadge, request) => {
-        const [, which, owner, repo, number, format] = match
-        const issueUri = `/repos/${owner}/${repo}/pulls/${number}`
-        const badgeData = getBadgeData('checks', queryParams)
-        if (badgeData.template === 'social') {
-          badgeData.logo = getLogo('github', queryParams)
```
chris48s: We've lost some functionality here:
 - https://shields-staging-pr-2253.herokuapp.com/github/status/s/pulls/badges/shields/1110.svg?style=social
 - https://img.shields.io/github/status/s/pulls/badges/shields/1110.svg?style=social

This seems like a good thing to implement in a base class which the GH service badges inherit from

services/github/github-pull-request-check-state.service.js
paulmelnikow: All set. Added a transform function too.

services/github/github-pull-request-check-state.tester.js
paulmelnikow: Added!

services/github/github-pull-request-status.service.js
paulmelnikow: Ah, yes, thanks for catching that!

This "default social logo" functionality belongs in BaseService. I think I'd probably call it `logo` or `socialLogo` and put it in `defaultBadgeData`. An option like this could also facilitate adding a `?logo=yes`-type option in the future.

The base Github service doesn't seem like a bad place to specify it, though many subclasses will need to override it. It might be less error-prone to include it in each badge class. That way it'll never be forgotten if a one of the subclasses gets a default badge data property at a later date.

paulmelnikow: Ha! Don't you just love realizing features like these are already implemented?

https://github.com/badges/shields/blob/bc4bd79e90058426278a973e7f742761b90d3b9c/services/base.js#L314-L317

services/github/github-pull-request-check-state.service.js
```diff
@@ -0,0 +1,105 @@
...
+        exampleUrl: 's/pulls/badges/shields/1110',
+        keywords: ['GitHub', 'pullrequest', 'detail', 'check'],
+        documentation,
+      },
+      {
+        title: 'GitHub pull request check contexts',
+        urlPattern: 'contexts/pulls/:user/:repo/:number',
+        staticExample: this.render({
+          which: 'contexts',
+          stateCounts: { passed: 5, pending: 1 },
```
chris48s: this example should also be passing the summary state param as well as the raw state counts

paulmelnikow: Do you mean, adding in `state` even though it's unused? I'm okay with that.

chris48s:
this is.. really strange. The change in 3173c2e looks like it should work but it seems to have not worked and somehow broken the other example:

screenshot at 2018-11-09 16-30-19

I am very confused by this

paulmelnikow:
Ah… that was a regression in #2240. It's unrelated. I'll open a new PR with a fix.

Issue #2078, opened 2018-09-13
PowerShellGallery downloads show invalid


thedavecarroll:
After today's release of the updated PowerShellGallery, the download data shows invalid, even on the shields.io site.

thedavecarroll:
The version badge is also showing invalid.

chris48s:
Hi. thanks for reporting this.

It looks like the issue here is that since the recent changes to powershell gallery (related reading: https://blogs.msdn.microsoft.com/powershell/2018/09/12/new-look-and-features-for-powershell-gallery/ ), the API no longer serves JSON when requested. For example

```
$ curl -H "Accept: application/atom+json,application/json" "https://www.powershellgallery.com/api/v2/Packages()?$filter=Id%20eq%20'ACMESharp'%20and%20IsLatestVersion%20eq%20true"
```

now returns an XML response.

This does not affect resharper or chocolatey which share the same NuGet v2 API parsing code.

I've raised an issue over on the powershell gallery support repo - https://github.com/PowerShell/PowerShellGallery/issues/21 Follow there for updates/progress.

If we're not able to get the data as JSON, the badge code would need to be rewritten to parse the XML feed instead, but hopefully there is a quicker solution.

thedavecarroll:
Thank you so much for responding. It's given me a lot of insight into parsing data from the gallery. So much easier than screen scraping. lol

cormacpayne:
+1 on behalf of [Azure PowerShell](https://github.com/Azure/azure-powershell) 😀

iricigor:
As per comment on https://github.com/PowerShell/PowerShellGallery/issues/21 there is a link available that gives JSON back. How complicated is to change  `www.powershellgallery.com` to `msconfiggallery.cloudapp.net`? As I see the only definition is [here](https://github.com/badges/shields/blob/ca6909938920c55872e792656c5d903699346527/services/nuget/nuget.service.js#L347).

`curl -H "Accept: application/atom+json,application/json" "https://msconfiggallery.cloudapp.net/api/v2/Packages()?$filter=Id%20eq%20'ACMESharp'%20and%20IsLatestVersion%20eq%20true"`

chris48s:
Yeah that issue seems to have gone quiet and the "temporary workaround" is still available. I've opened PR #2189 proposing the change.

jzabroski:
The PowerShell team is famous for not testing their software, in my experience...

paulmelnikow:
[...]

PR #2257, opened 2018-11-03
Rewrite [NuGet] badges including [myget chocolatey resharper powershellgallery]

paulmelnikow:
The NuGet badge examples are straggling in all-badge-examples. Rather than move them as is, I thought it made more sense to refactor the services and see if they could be generated. I didn't take that on here; this is a straight rewrite of the badges. The old implementations were fairly difficult to follow. The new implementations are complicated too, though I hope much more readable.

Though the NuGet behaviors could be consolidated into a single flag, I split `withTenant` and `withFeed` into separate flags, thinking naming the behaviors makes the implementations easier to understand. I defaulted these to true, thinking that really this is really a MyGet implementation which is generalized to NuGet. Though maybe it makes more sense to have the MyGet style as the default. Probably it doesn't matter much either way.

I added a helper class ServiceUrlBuilder to construct the Shields service URL. It's useful in this complex case where the URL must be built up conditionally. This might be useful in a couple other places.

This isn't completely finished, though thought it's ready to be read, so I though I'd open it for comments.


services/myget/myget.tester.js
```diff
@@ -32,21 +32,21 @@ t.create('total downloads (valid)')
 
 t.create('total downloads (not found)')
   .get('/mongodb/dt/not-a-real-package.json')
-  .expectJSON({ name: 'downloads', value: 'not found' })
+  .expectJSON({ name: 'downloads', value: 'package not found' })
```
paulmelnikow: The next two tests are failing because of the way we're propagating errors through `regularUpdate`. It might just be a problem with `nock`. I have to think it through a bit more.

services/myget/myget.tester.js
```diff
@@ -58,11 +58,12 @@ t.create('total downloads (unexpected second response)')
...
     nock('https://api-v2v3search-0.nuget.org')
       .get(
-        '/query?q=packageid:mongodb.driver.core&prerelease=true&semVerLevel=2'
+        '/query?q=packageid%3Amongodb.driver.core&prerelease=true&semVerLevel=2'
       )
       .reply(invalidJSON)
   )
-  .expectJSON({ name: 'downloads', value: 'invalid' })
+  // .expectJSON({ name: 'myget', value: 'unparseable json response' })
+  .expectJSON({ name: 'downloads', value: 'unparseable json response' })
```
paulmelnikow: I'm not sure I feel strongly about the label on some of these error badges. They're a bit inconsistent, some showing the service name and some showing "downloads."

services/myget/myget.tester.js
```diff
@@ -163,11 +164,11 @@ t.create('version (unexpected second response)')
   .intercept(nock =>
     nock('https://api-v2v3search-0.nuget.org')
       .get(
-        '/query?q=packageid:mongodb.driver.core&prerelease=true&semVerLevel=2'
+        '/query?q=packageid%3Amongodb.driver.core&prerelease=true&semVerLevel=2'
       )
       .reply(invalidJSON)
   )
-  .expectJSON({ name: 'mongodb', value: 'invalid' })
+  .expectJSON({ name: 'myget', value: 'unparseable json response' })
```
paulmelnikow: The error messages now get the label from `defaultBadgeData`, which is why the MyGet error badges no longer show the feed name. I think this is fine.

services/nuget/nuget-v2-service-family.js
```diff
@@ -0,0 +1,120 @@
...
+async function fetch(
+  serviceInstance,
+  { baseUrl, repo, includePrereleases = false }
+) {
+  const releaseTypeFilter = includePrereleases
+    ? 'IsAbsoluteLatestVersion eq true'
+    : 'IsLatestVersion eq true'
+  const filter = `Id eq '${repo}' and ${releaseTypeFilter}`
+  const data = await serviceInstance._requestJson({
+    schema: Joi.any(),
```
paulmelnikow: ^^ Needs to be fixed.

services/nuget/nuget-v3-service-family.js
```diff
@@ -0,0 +1,238 @@
...
+      json.resources.filter(
+        resource => resource['@type'] === 'SearchQueryService'
+      ),
+  })
+  return randomElementFrom(searchQueryServices)['@id']
+}
+
+const schema = Joi.object({
+  data: Joi.array()
+    .items(Joi.any())
```
paulmelnikow: ^^ Needs to be fixed.

paulmelnikow:
Hmmm… I'm not sure whether these PowerShellGallery timeouts might relate to these changes or not.

paulmelnikow:
I wrote a new service to handle the Powershell badges. They've diverged a little bit from the Nuget v2. There's a bit of shared code which I factored out.

If the XML Nuget APIs are more reliable, we could consider switching everything else over to them, though for now I would like to get this merged and get #2078 fixed.

services/myget/myget.tester.js
chris48s: I don't think this is an issue with `nock`. If you set `apiBaseUrl` to something that won't resolve and test it through the front end, we throw an `UnhandledPromiseRejectionWarning` rather than rendering an "inaccessable" badge like we do with other badges.

services/chocolatey/chocolatey.service.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+const { createServiceFamily } = require('../nuget/nuget-v2-service-family')
+
+module.exports = createServiceFamily({
```
chris48s: I know you've just left the examples in `all-badge-examples` for now, but is the plan here to override `examples()` in these files so we do something like:

```js
{ Downloads, Version } = createServiceFamily({ /*...*/ })

Downloads.examples = function() { /*...*/ }

Version.examples = function() { /*...*/ }

module.exports = { Downloads, Version }
```

(not tested, but you get the idea)

services/powershellgallery/powershellgallery.tester.js
```diff
@@ -36,170 +27,28 @@ t.create('total downloads (not found)')
...
-    nock('https://msconfiggallery.cloudapp.net')
-      .get(
-        '/api/v2/Packages()?$filter=Id%20eq%20%27ACMESharp%27%20and%20IsLatestVersion%20eq%20true'
-      )
-      .reply(200, nuGetV2VersionJsonFirstCharNotZero)
-  )
-  .expectJSON({
-    name: 'powershellgallery',
-    value: 'v1.2.7',
-    colorB: colorscheme.blue.colorB,
```
chris48s: We should probably replace all these deleted tests with versions that use XML fixtures.

chris48s:
Also there are some valid calls that are throwing an error on staging but work in production:

myget downloads:
* https://shields-staging-pr-2257.herokuapp.com/dotnet.myget/dotnet-coreclr/dt/Microsoft.DotNet.CoreCLR.svg
* https://img.shields.io/dotnet.myget/dotnet-coreclr/dt/Microsoft.DotNet.CoreCLR.svg

myget version:
* https://shields-staging-pr-2257.herokuapp.com/dotnet.myget/dotnet-coreclr/v/Microsoft.DotNet.CoreCLR.svg
* https://img.shields.io/dotnet.myget/dotnet-coreclr/v/Microsoft.DotNet.CoreCLR.svg

services/powershellgallery/powershellgallery.tester.js
paulmelnikow: I’m not sure they’re needed. It seems that the only difference between these and the main tests is the coverage of the color logic. All the plumbing of those values is the same.

A much simpler way to test the color logic would be a unit test on the render function, and meanwhile it’s covered by tests on other nuget services.

services/chocolatey/chocolatey.service.js
paulmelnikow: Yea, that's a possibility. The other possibility is that we can modify the examples to use the same example across all the badges for a given family. That way they'll be easier to generate programmatically. That approach probably won't work for myget, though, which has a lot of variants.

services/powershellgallery/powershellgallery.tester.js
chris48s: Fair point - previously it was all such a big ball o' mud that we couldn't test that logic independently

PR #2261, opened 2018-11-03
refactor [bitbucket] service

chris48s:


services/bitbucket/bitbucket.tester.js
```diff
@@ -147,10 +106,6 @@ t.create('branch build result (valid)')
     })
   )
 
-t.create('branch build result (not found)')
-  .get('/pipelines/atlassian/not-a-repo/some-branch.json')
-  .expectJSON({ name: 'build', value: 'not found' })
```
paulmelnikow: I know we've decided to generally drop the "invalid JSON" tests. I've generally found the "not found" tests to be valuable, in making sure we're showing the message we expect. What do you think?

PyvesB: I've put a contradictory review comment in #2232, as I've seen us dropping the "not found" ones as well in other places. Unless we're overriding the "not found" message with something more specific, in which case it's valuable to check that the override is done correctly, I'm unsure there's much value in repeating a test that is already covered by the parent class.

chris48s: Yeah I think if we're relying on the default behaviour of the base class, that behaviour is already under test

paulmelnikow: True. Though what we aren't testing anymore is that the service returns a 404 status.

chris48s: This is a fair point - we use the tests to tell us about changes to the behaviour of the upstream API as well. I've added the ones relevant to this service back in here, but I'll have a look at the other tests we removed and updating the testing docs as another PR.

PR #2277, opened 2018-11-06
Fix suggest on staging in Firefox

### LINKED ISSUE #2245: Suggest doesn't work on staging in Firefox
 - [ ] My own instance
 - [x] [Staging server](https://shields-staging.herokuapp.com/)

:beetle: **Description**

The suggest feature doesn't work on staging in Firefox. It displays a Next.js error, "An unexpected error has occurred." It _does_ work in Chromium, and in production in Firefox.

See full report by @chris48s: https://github.com/badges/shields/pull/2029#issuecomment-434818044

paulmelnikow:
To fix #2245.

The suggest request on Firefox is returning `{"err":"Disallowed"}` which suggests the problem is CORS-related.

https://github.com/badges/shields/blob/88402dd7a8ed64953937157de866d9a62dab8c48/lib/suggest.js#L112-L120

On staging, unlike production, the hostname is the same, which means browsers might not send CORS headers by default. Maybe there's a difference between how Chrome and Firefox handle that.

Anyway, here's a fix based on that hunch.

**Added:** Took some experimentation, but I tracked down the problem.


Procfile
```diff
@@ -0,0 +1 @@
+web: node server
```
paulmelnikow: We've been running the development server on Heroku. This fixes that. The frontend is built during the build process, and then served through the regular server.

frontend/components/suggestion-and-search.js
```diff
@@ -49,7 +49,7 @@ export default class SuggestionAndSearch extends React.Component {
       let suggestions
       try {
         const json = await res.json()
-        suggestions = json.badges
+        suggestions = json.badges || []
```
paulmelnikow: This fixes the crash in the frontend when a response comes back without a `badges` array.

lib/suggest.js
```diff
@@ -107,16 +107,31 @@ function setRoutes(allowedOrigin, githubApiProvider, server) {
...
+
+      if (host !== ask.req.headers.host) {
+        if (allowedOrigin.includes(origin)) {
+          ask.res.setHeader('Access-Control-Allow-Origin', origin)
+        } else {
+          ask.res.setHeader('Access-Control-Allow-Origin', 'null')
+          end({ err: 'Disallowed' })
+          return
+        }
+      }
```
paulmelnikow: This fixes the issue in Firefox, which is sending an Origin header.

package.json
```diff
@@ -194,8 +194,8 @@
     "wait-promise": "^0.4.1"
   },
   "engines": {
-    "node": ">= 8.x",
-    "npm": "5.x"
+    "node": ">= 8",
+    "npm": ">= 5"
```
paulmelnikow: This prevents an `npm` downgrade during the Heroku build.

PR #2278, opened 2018-11-06
BaseService terminology: Rename `url` to `route`

paulmelnikow:
The term “url” is overloaded in services, to refer to the Shields route and also the API URL. Calling the Shields URL a “route” makes it clearer and more obvious which one of these we’re talking about. It’s a small thing, though seems like an improvement.

We have a few functions called `buildUrl`. I’ve renamed them to `buildRoute` when they refer to routes, and left them as `buildUrl` when they refer to API URLs.

I included a minor style tweak and some formatting cleanup in `TUTORIAL.md`.


paulmelnikow:
Thanks for reading and testing. I think basically the entire thing was broken because of de613ba, and those issues are all fixed now. When I pulled in the changes from #2279 I was updating a bunch of PRs and must not have waited for the builds to complete.

Issue #2050, opened 2018-09-04
Reduce duplication in badge regex/url patterns


chris48s:
Can we use something like [url-pattern](https://github.com/snd/url-pattern) or [path-to-regexp](https://github.com/pillarjs/path-to-regexp) to define routes in a nicer way so we don't have to define this twice?

relevant discussions:
https://github.com/badges/shields/pull/2042#discussion_r214523205
https://github.com/badges/shields/pull/1582#discussion_r176912366

paulmelnikow:
I played with `path-to-regexp` a bit back in March. This is based on what I did then:

```js
const pathToRegexp = require('path-to-regexp')

const keys = []
const re = pathToRegexp('/appveyor/ci/:user/:repo/:branch?.:ext(png|gif)', keys, {
  delimiter: '.',
  strict: true,
  sensitive: true,
})

console.log(re)

const result = {}
re.exec('/appveyor/ci/foo/bar/baz.png').slice(1).forEach((match, i) => {
  result[keys[i].name] = match
})
console.log(result)
```

Output:

```
/^\/appveyor\/ci\/([^\/]+?)\/([^\/]+?)\/([^\/]+?)?\.(png|gif)$/
{ user: 'foo', repo: 'bar', branch: 'baz', ext: 'png' }
```

Cool stuff!

The `compile()` method would be useful in the frontend to build a badge URL from user input. I ran into some issues with that, IIRC. It'd be good to make sure that works before we settle on this library.

paulmelnikow:
`compile()` seems to be working now!

```js
const toPath = pathToRegexp.compile('/appveyor/ci/:user/:repo/:branch?.:ext(png|gif)', {
  delimiter: '.',
  strict: true,
  sensitive: true,
})

console.log(toPath({ user: 'foo', repo: 'bar', branch: 'baz', ext: 'png' }))
```

What do you think? Should we move forward with this approach?

paulmelnikow:
#2279 is about ready for merge!

The next step would be to reduce duplication in defining examples, and maybe do this with an eye toward #701 and #776.

PR #2279, opened 2018-11-06
Reduce duplication in badge regex/url patterns; affects [ansible apm gem wercker]

paulmelnikow:
This reduces duplication in badge regex/url patterns, and reduces the need to understand regexes in order to create badges.

The badge examples could be harmonized to reduce the duplication even further, though that could be handled in a follow-on PR.

Ref: #2050


services/base.spec.js
```diff
@@ -320,7 +368,7 @@ describe('BaseService', function() {
   })
 
   describe('ScoutCamp integration', function() {
-    const expectedRouteRegex = /^\/foo\/([^/]+).(svg|png|gif|jpg|json)$/
+    const expectedRouteRegex = /^\/foo\/([^/]+?)\.(svg|png|gif|jpg|json)$/
```
paulmelnikow: 1. Today I learned that `+?` is a [lazy qualifier](https://javascript.info/regexp-greedy-and-lazy), which causes the segment to match as little as possible. It doesn't make this any less correct, and I imagine makes it easier to build pattern regexes more reliably.
2. `.` to `\.` fixes a minor bug. Before this change, we accept any character as the extension separator.

doc/TUTORIAL.md
```diff
@@ -124,9 +123,9 @@ Description of the code:
 3. Our module must export a class which extends `BaseService`
 4. `url()` declares a route. We declare getters as `static`.
     * `base` defines the static part of the route.
-    * `format` is a [regular expression](https://www.w3schools.com/jsref/jsref_obj_regexp.asp) defining the variable part of the route.
-    * We can use `capture` to extract matched regex clauses into one or more named variables. Here we are declaring that we want to store the string matched by `([^/]+)` in a variable called `text`.
-    This declaration adds the route `/^\/test\/([^\/]+)\.(svg|png|gif|jpg|json)$/` to our application.
+    * `pattern` defines the variable part of the route. It can include any
+      number of named parameters. These are converted into
+      regular expressions by [`path-to-regexp`][path-to-regexp].
```
chris48s: Do you think its worth mentioning the format/capture args in the docs too, or is it too much detail?

doc/TUTORIAL.md
```diff
@@ -165,8 +166,7 @@ module.exports = class GemVersion extends BaseJsonService {     // (5)
   static get url() {                                            // (6)
     return {
       base: 'gem/v',
-      format: '(.+)',
-      capture: ['gem'],
+      format: ':gem',
```
chris48s: Can we also apply this to the actual gem service so that the example in the docs reflects real life

services/base.js
```diff
@@ -175,13 +176,41 @@ class BaseService {
...
-    // Regular expressions treat "/" specially, so we need to escape them
-    const escapedPath = this.url.format.replace(/\//g, '\\/')
-    const fullRegex = `^${this._makeFullUrl(
-      escapedPath
-    )}.(svg|png|gif|jpg|json)$`
-    return new RegExp(fullRegex)
+    const { pattern, format, capture } = this.url
+    if (pattern) {
+      if (format || capture) {
+        throw Error(
```
chris48s: I know we didn't have this check before, but can we also add check here that if we _haven't_ got a `pattern` key we should have both a `format` and `pattern` and throw a nice error if we haven't. At the moment we'll get something like `TypeError: Cannot read property 'replace' of undefined` or `Service declares incorrect number of capture groups`

doc/TUTORIAL.md
paulmelnikow: Hmmmm… well I guess my hope is that we won't need to use them. Also yea, it seems like it's a bit detailed to include in such a simple example in a tutorial.

I'm curious: are there services that you think will require a hand-crafted regex?

doc/TUTORIAL.md
paulmelnikow: Haha, sure.

services/base.js
paulmelnikow: I think you meant `capture`, not `pattern`, though `capture` isn't required (and there's a test that it works with no `capture`).

But yea, I believe `format` is. I'll add a check for that.

doc/TUTORIAL.md
chris48s: > I'm curious: are there services that you think will require a hand-crafted regex?

This is an example I'd pick of one that I suspect would be quite hard:

https://github.com/badges/shields/blob/5019d81642716895df4784f411cc001a5a5c772a/services/wercker/wercker.service.js#L81

although I've not tried to port it (there is _history_ to that one). Happily, these examples are in a minority.

In general, routes that use a non-capturing group might be difficult but I may wrong on that? If so, I suppose you could just capture stuff and not use it..

services/base.js
chris48s: Yeah, sorry.. wordsing is hard. What you said :)

doc/TUTORIAL.md
paulmelnikow: Wow, that is a monster! I can't say I can follow what it's doing, even from the tests. I'm guessing it's handling some cases which aren't covered by the tests.

I imagine it's too complex to deal with using path-to-regexp with a single pattern. Though I'd wager it could be expressed clearly using two or three, two of which I imagine are deprecated.

There is support for [unnamed parameters](https://github.com/pillarjs/path-to-regexp#unnamed-parameters), which could be a way to handle non-capturing groups. We'd have to build a little support around those.

Though I wonder if it would be better to allow services to register multiple patterns. To eventually build the URLs from fields in the UI, it would be helpful to separate the deprecated patterns from the current pattern.

It could also make matching more efficient, down the line, because it could allow pushing things from the regex into a prefix, which could be indexed in a trie.

chris48s: > Though I'd wager it could be expressed clearly using two or three, two of which I imagine are deprecated

> Though I wonder if it would be better to allow services to register multiple patterns.

This is a good point. We're getting into the weeds a bit on an unrelated issue here, but in this specific example there is a reason why we've defined this as a regex rather than several classes:

Using the regex allows us to specify the order of matching (i.e: try to match this first, then if that fails, try to match that), whereas if we had defined one route for  `ci/([a-fA-F0-9]{24})(?:/(.+))?` and one for `ci/([^/]+/[^/]+)(?:/(.+))?`, it matters what order we register the routes in (or what order we try to match in) because if we end up testing against `ci/([^/]+/[^/]+)(?:/(.+))?` first `ci/559e33c8e982fc615500b357/master` will match that, but we don't want it to.

The upshot of that is if we want to allow a single badge to define multiple routes, we'd want to be able to also specify which order we'll try to match those routes in, if you see what I mean.

All of that said, lets not get too bogged down in this one case for now. As it stands, this will allow us to improve 99% of cases and it we can continue to use a regex in other situations. We can worry about these odd edge cases antoher day.

paulmelnikow: Yea, that makes sense.

I'm thinking it would be good to identify canonical forms of these routes, and use a 301 redirect. There's a small performance penalty, but it significantly increases the chance that the URLs will be updated because if they're pasted into an address bar, they will be replaced by the canonical URL. There are a couple services where we've renamed things, and there's an "old static badge" example in the bottom of `server.js`. (Honestly perhaps that last one could be deleted. I wonder if it's getting any hits.)

It could be useful to set a priority for services so they can be configured to register earlier or later.

Not the highest priorities, but probably worth tackling someday.

paulmelnikow:
For a small change this is pretty high impact. While I'd love to give the other maintainers a chance to weigh in, I'm also eager to land it so we can start using it, and also so we can push forward on the next step which would be reducing some duplication between the route and the examples.

Hopefully everyone will be cool with this change – though if not I'm certainly happy to continue the discussion! 😉 

@RedSparr0w @platan @PyvesB

PyvesB:
The new approach is fine by me. 👍

RedSparr0w:
Nice work! :+1: 
Looks good to me :smile:

PR #2282, opened 2018-11-06
fix [hexpm] validation

### LINKED ISSUE #2272: Fix hexpm validation
 - [ ] My own instance

:beetle: **Description**
Some HexPM badges throwing `invalid response data`

:link: **Link to the badge**
[Hex pm](https://hex.pm/packages/straw_hat_map)

:bulb: **Possible Solution**

The HexPM validation expects the `downloads` object to have keys `all`, `week` and `day`:

https://github.com/badges/shields/blob/07b282fa1fe2cb9816819eefd416173f3b39f11f/services/hexpm/hexpm.service.js#L16-L28

but sometimes some of them aren't there. For example this is a valid response:

```json
  "downloads": {
    "all": 130,
    "recent": 36
  },
```

I suspect this happens for packages that don't get much traffic, but because we use the same schema for all the hex badges it affects version etc as well as downloads.

We need to adjust the validation and badge code to handle if those keys don't exist.

chris48s:
closes #2272

### DIFF
--- a/services/hexpm/hexpm.service.js
+++ b/services/hexpm/hexpm.service.js
@@ -11,13 +11,13 @@ const {
   downloadCount: downloadCountColor,
   version: versionColor,
 } = require('../../lib/color-formatters')
-const { nonNegativeInteger } = require('../validators')
 
 const hexSchema = Joi.object({
   downloads: Joi.object({
-    all: nonNegativeInteger,
-    week: nonNegativeInteger,
-    day: nonNegativeInteger,
+    // these keys may or may not exist
+    all: Joi.number().integer(),
+    week: Joi.number().integer(),
+    day: Joi.number().integer(),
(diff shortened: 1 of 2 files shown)


services/hexpm/hexpm.service.js
```diff
@@ -149,7 +149,9 @@ function DownloadsForInterval(interval) {
 
     async handle({ pkg }) {
       const json = await this.fetch({ pkg })
-      return this.constructor.render({ downloads: json.downloads[interval] })
+      return this.constructor.render({
+        downloads: json.downloads[interval] || 0,
```
paulmelnikow: I've a slight preference to apply the defaults in the Joi object instead. i.e. `Joi.number().integer().default(0)` We're using that in a few places with arrays.

What do you think?

PR #2284, opened 2018-11-07
Move [StaticBadge] to own service & add test; also affects [gitter]

paulmelnikow:
This picks up @RedSparr0w's work in #1802.

1. The handler for the static badge is moved into its own service with a synchronous handler. Avoiding an async call may make the static badges slightly faster, though it may be worth profiling this if it turns out we want asynchronous static badges in the future. If it doesn't make a performance difference we could make this handler `async` like the others.
2. Most of the custom static-badge logic is in a BaseStaticBadge base class.
3. Rewrite the static Gitter badge to use BaseStaticBadge.
4. A bit of minor cleanup in related functions.


lib/request-handler.js
```diff
@@ -61,13 +61,13 @@ function getBadgeMaxAge(handlerOptions, queryParams) {
     ? parseInt(process.env.BADGE_MAX_AGE_SECONDS)
     : 120
   if (handlerOptions.cacheLength) {
-    // if we've set a more specific cache length for this badge (or category),
-    // use that instead of env.BADGE_MAX_AGE_SECONDS
-    maxAge = parseInt(handlerOptions.cacheLength)
+    // If we've set a more specific cache length for this badge (or category),
+    // use that instead of env.BADGE_MAX_AGE_SECONDS.
+    maxAge = handlerOptions.cacheLength
```
paulmelnikow: Technically unrelated, as in the current implementation the static badge does not call this function.

There's no need for `parseInt` since we're passing numbers, not strings.

services/base-static.js
```diff
@@ -0,0 +1,57 @@
+'use strict'
+
+const { makeMakeBadgeFn } = require('../lib/make-badge')
+const { makeSend } = require('../lib/result-sender')
+const BaseService = require('./base')
+
+const serverStartTime = new Date(new Date().toGMTString())
```
paulmelnikow: There's no point in passing this down from `server.js`.

services/base.js
```diff
@@ -214,6 +219,52 @@ class BaseService {
     return result
   }
 
+  _handleError(error) {
```
paulmelnikow: Extract into a method so `BaseStaticService` can invoke it from overridden `register()`.

services/base.js
```diff
@@ -299,8 +308,8 @@ class BaseService {
 
     const badgeData = {
       text: [
-        overrideLabel || serviceLabel || defaultLabel || this.category,
-        serviceMessage || 'n/a',
+        coalesce(overrideLabel, serviceLabel, defaultLabel, this.category),
+        coalesce(serviceMessage, 'n/a'),
```
paulmelnikow: This is covered by the static badge tests @RedSparr0w wrote.

services/base.js
```diff
@@ -318,28 +327,26 @@ class BaseService {
   }
 
   static register({ camp, handleRequest, githubApiProvider }, serviceConfig) {
-    const ServiceClass = this // In a static context, "this" is the class.
```
paulmelnikow: Some references use `this` and some use `ServiceClass`. Don't really care which we use, though consistent = better.

services/base-static.js
```diff
@@ -0,0 +1,57 @@
...
+  static register({ camp, measurer }, serviceConfig) {
+    const makeBadge = makeMakeBadgeFn(measurer)
+
+    camp.route(this._regex, (queryParams, match, end, ask) => {
+      if (+new Date(ask.req.headers['if-modified-since']) >= +serverStartTime) {
+        // Send Not Modified.
+        ask.res.statusCode = 304
+        ask.res.end()
+        return
+      }
```
paulmelnikow: Currently this behavior is not tested.

services/base-static.js
```diff
@@ -0,0 +1,57 @@
...
+        console.time('makeBadge total')
+      }
+      const svg = makeBadge(badgeData)
+      if (serviceConfig.profiling.makeBadge) {
+        console.timeEnd('makeBadge total')
+      }
+
+      const cacheDuration = 3600 * 24 * 1 // 1 day.
+      ask.res.setHeader('Cache-Control', `max-age=${cacheDuration}`)
+      ask.res.setHeader('Last-Modified', serverStartTime.toGMTString())
```
paulmelnikow: Currently this behavior is not tested.

chris48s:
I just realised this breaks the "no label" case e.g:

 - https://img.shields.io/badge/all%20one%20color-red.svg
 - https://shields-staging-pr-2284.herokuapp.com/badge/all%20one%20color-red.svg

PR #2292, opened 2018-11-08
Make a few [github] tests more reliable

paulmelnikow:
The version test is failing because the shields repo version is not a dotted version.

Ref: #1359


paulmelnikow:
On the first run, it timed out at 10 seconds. I wonder if we should remove these hit counter badges; the underlying API really doesn't work well enough. They only work if the badges are cached for a long time, like a day or a week.

PR #2296, opened 2018-11-09
2.0.0 🚢

chris48s:
Seems like there were no showstoppers on the beta (or nobody tested it).
Lets publish 2.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gh-badges",
-  "version": "2.0.0-beta1",
+  "version": "2.0.0",
   "description": "Official Shields.io badge library.",
   "keywords": [
     "GitHub",


felixfbecker:
Would be nice to have some GitHub release notes highlighting breaking changes

paulmelnikow:
https://github.com/badges/shields/blob/master/gh-badges/CHANGELOG.md

felixfbecker:
I know, but could it be put into or linked from the GitHub release notes? Renovate wasn't able to detect this changelog file (I assume because it's in a subdir, #2823 could help with that)

chris48s:
related thread: https://github.com/renovatebot/renovate/issues/2926

Slight side issue: One of the other things we've discussed is doing more concrete 'releases' for the website/server for self-hosting users. If we're going to look at our tags/releases it is worth moving to expressing our releases/tags as gh-badges@2.1.0 or similar to accommodate that.

paulmelnikow:
Ah, yes, and also: https://github.com/renovatebot/renovate/issues/2162#issuecomment-399689029

re `gh-badges@2.1.0` that makes sense. I know lerna has a convention for independent tags. Maybe we could follow that.

PR #1192, opened 2017-10-20
added clojars downloads count

flyboarder:
Add Clojars Downloads Count Badge

Closes #1152 

Does not include tests, can someone help with that?

PR #2305, opened 2018-11-12
[Clojars] Add downloads badge

### LINKED ISSUE #1152: [Request] Add clojars downloads badge
Would be awesome to get clojars downloads.

emlyn:
Since #1192 seems to have been abandoned, I had a go at creating a Clojars downloads badge myself.
For now just total downloads, although in the json response there are download counts for some recent versions.

Closes #1152.

### DIFF
--- a/services/clojars/clojars-downloads.service.js
+++ b/services/clojars/clojars-downloads.service.js
@@ -1,0 +1,67 @@
+'use strict'
+
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
+const { metric } = require('../../lib/text-formatters')
+const { NotFound } = require('../errors')
+const { nonNegativeInteger } = require('../validators')
+const { downloadCount: downloadsColor } = require('../../lib/color-formatters')
+
+const clojarsSchema = Joi.object({
+  downloads: nonNegativeInteger,
+}).required()
+
+module.exports = class Clojars extends BaseJsonService {
+  async fetch({ clojar }) {
+    const url = `https://clojars.org/api/artifacts/${clojar}`
+    return this._requestJson({
+      url,
+      schema: clojarsSchema,
+    })
+  }
+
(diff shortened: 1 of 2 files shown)


services/clojars/clojars-downloads.service.js
```diff
@@ -0,0 +1,67 @@
...
+      color: downloadsColor(downloads),
+    }
+  }
+
+  async handle({ clojar }) {
+    const json = await this.fetch({ clojar })
+
+    if (json.downloads === undefined) {
+      throw new NotFound()
+    }
```
chris48s: This check can be removed. The typical "not found" case is a 404 (which gets handled by `_requestJson()` )

```sh
$ curl -I "https://clojars.org/api/artifacts/not-a-package"
HTTP/1.1 404 Not Found
```

If we've got a `200 OK` response and `downloads` is `undefined`, it will be rejected by the schema because `nonNegativeInteger` is `required()`: https://github.com/badges/shields/blob/510491f37644432740548afd8cdbac7e1978b59f/services/validators.js#L6-L9

services/clojars/clojars-downloads.tester.js
```diff
@@ -0,0 +1,29 @@
...
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'downloads',
+      value: /^[0-9]+$/,
+    })
+  )
+
+t.create('clojars downloads (lots)')
+  .get('/prismic.json')
+  .expectJSONTypes(
```
chris48s: I'm not clear on the difference between these two tests - it seems like there should only be one. I think you can just call `/prismic.json` and use the `isMetric` test validator to check the output:

https://github.com/badges/shields/blob/510491f37644432740548afd8cdbac7e1978b59f/services/test-validators.js#L60

(require it from `test-validators` )

Issue #701, opened 2016-04-30
Add Username and Repo Name to Shield Creation fields


oscarmorrison:
As a new user to Shields.io badges it is not obvious where in the URL 
should be the username and the repository name. 
It would be helpful to add fields such as 

**Github User Name**
**Repo Name**

to Shield Creation form such as below:
(obviously this would be different field names per badge)

![Screenshot](http://i.imgur.com/JIAQ3Fd.png)

oscarmorrison:
CC @espadrine

nnadeau:
+1

paulmelnikow:
Related to #466.

tooomm:
See https://github.com/badges/shields/pull/1331#issuecomment-349474676 and following:
**Request to add a `direct link to generated/shown custom image` and a `custom parameter` field.**

paulmelnikow:
Let's track this in #2496.

PR #2308, opened 2018-11-12
Declare static examples using namedParams

### LINKED ISSUE #2050: Reduce duplication in badge regex/url patterns

paulmelnikow:
This continues the work from #2279, by allowing example badges to be specified using `namedParams`. Using an object makes it possible for us to display these in form fields down the line. (#701)

I've called this the "preferred" way, and labeled the other ways deprecated. I've also added some doc to the `examples` property in BaseService. Then I realized we had some doc in the tutorial, though I think it's fine to have a short version in the tutorial, and the gory detail in BaseService.

I've also added a `pattern` keyword, and made `urlPattern` an alias.

Closes #2050.


paulmelnikow:
Have labeled this a blocker as it blocks future work on the frontend including a branch where I've started refactoring toward #701, and future work on BaseService.

services/base.js
```diff
@@ -116,28 +154,37 @@ class BaseService {
...
             )
           }
-          if (!exampleUrl) {
+          if (namedParams && exampleUrl) {
+            throw new Error(
+              `Static example for ${
+                this.name
+              } at index ${index} declares both namedParams and exampleUrl`
+            )
+          } else if (!namedParams && !exampleUrl) {
```
chris48s: Do we also want a check here that if you've declared `namedParams` we also need a `pattern` (or at least the URL shouldn't be defined using a regex)? Also `prepareExamples()` seems to have become quite large and unwieldy. I think we could usefully split all these validation rules out into their own function. In general we are now supporting quite a lot of different ways of doing the same thing (for valid reasons). It would be nice to be able to simplify this at some point in future, but I realise there is more refactoring work that must be done before we can simplify this.

services/base.spec.js
```diff
@@ -42,6 +42,18 @@ class DummyService extends BaseService {
...
+        exampleUrl: 'World',
+        staticExample: this.render({ namedParamA: 'foo', queryParamA: 'bar' }),
+        keywords: ['hello'],
+      },
+      {
+        pattern: ':world',
+        namedParams: { world: 'World' },
+        staticExample: this.render({ namedParamA: 'foo', queryParamA: 'bar' }),
+        keywords: ['hello'],
+      },
```
chris48s: As the validation logic grows more complicated its all got a bit confusing it would be helpful to formalise the behaviour with some tests in here ensuring we're doing the right thing with various combinations of `previewUrl` / `exampleUrl` / `namedParams` etc. Moving the validation rules to their own function will also help with this.

services/base.js
```diff
@@ -116,28 +154,37 @@ class BaseService {
...
           previewUrl,
+          pattern,
           urlPattern,
           staticExample,
           documentation,
           keywords,
         },
         index
       ) => {
+        pattern = pattern || urlPattern || this.route.pattern
```
chris48s: I'm confused about why we want both `pattern` and `urlPattern` and what situation we use each one in. Can you explain the difference.

paulmelnikow: We wouldn't want both. It's a temporary alias to `pattern` to make this diff smaller. I added it to the docstring. Does that make sense?

services/base.js
paulmelnikow: We can't get here if there's no pattern since we throw an error a few lines up. The only case we don't check is previewUrl + namedParams + !staticExample. At some point this stuff just needs to be rewritten using a Joi schema, though I expect the error messages will be a bit less friendly. Totally agreed it would be nice to consolidate ways of doing things. That said, this work is a lot more manageable if it's done in backward-compatible ways. Hopefully we can keep pushing the rewrite forward and move toward things being consistent.

I started a refactor of the badge preparation, which is basically keeping the example declarations as close as possible to what the frontend gets. In that branch there's only one transform to do here, and the only other responsibility of this function is validation. I'd be happy to pick up the validation refactor as part of that follow-on work. I tried to keep these changes incremental.

services/base.spec.js
paulmelnikow: Do you mean tests of invalid examples? Yea, agreed.

services/base.js
chris48s: > We can't get here if there's no pattern since we throw an error a few lines up. 

:+1:


> That said, this work is a lot more manageable if it's done in backward-compatible ways

Yes agreed. As I say, it is complicated for valid reasons. It will just be nice when they stop being valid :)

Regardless of future plans, what do you think of splitting the `if` statement/exception throwing block out into a function (maybe `validateExample()`? ) just so that `prepareExamples()` is doing less stuff. Even if we change from a bunch of `if` statements to a `Joi` schema at a later date, that's still a sensible seperation of concerns.

services/base.js
chris48s: > Does that make sense?

Maaaybe.. Am I right that essentially what you really want to do is rename `urlPattern` to `pattern`, but we're supporting both so that we don't have to change every single badge that declares `urlPattern` to `pattern` in this PR.. but then we want to do that at a later date and then remove `urlPattern` when we've done that?

services/base.js
paulmelnikow: > Yes agreed. As I say, it is complicated for valid reasons. It will just be nice when they stop being valid :)

😁 

> Regardless of future plans, what do you think of splitting the `if` statement/exception throwing block out into a function (maybe `validateExample()`? )

Okay. You wore me down. 😉

paulmelnikow:
Looks good. Thanks for the tests!

PR #2309, opened 2018-11-12
Move github examples into services/github

paulmelnikow:


paulmelnikow:
Have labeled this a blocker, as it will lead to merge conflicts in all-badge-examples.js and is necessary for #2320.

services/github/github-contributors.service.js
```diff
@@ -7,10 +7,32 @@ const {
...
+    return {
+      base: 'github',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'GitHub contributors',
+        previewUrl: 'contributors/cdnjs/cdnjs',
```
chris48s: Should this one be

```
base: 'github/contributors',

previewUrl: 'cdnjs/cdnjs',
```

paulmelnikow: Thinking ahead: that won't work when this is rewritten, because there's `contributors-anon` in the pattern, and BaseService wants a slash between `base` and `pattern`.

(There are a bunch of others like this.)

services/github/github-commits-since.service.js
```diff
@@ -6,8 +6,36 @@ const {
   makeLabel: getLabel,
   makeLogo: getLogo,
 } = require('../../lib/badge-data')
+const { documentation } = require('./github-helpers')
 
 module.exports = class GithubCommitsSince extends LegacyService {
+  static get category() {
+    return 'size'
```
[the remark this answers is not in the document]
paulmelnikow: Nope, good catch, that's copy-pasta. Fixed.

PR #1305, opened 2017-12-01
webpage: add missing titles + more badge previews

tooomm:
`Node 9 support` (#1290)
over
`Github rate limits cause transient service test failures in CI` (#979)

Was messing around with the layout due to the very wide column needed:
untitled

~~Follow-up commit will fix tests (editing from web).~~

PR #2311, opened 2018-11-13
POC for precomputing the text-width lookup table [WIP]

paulmelnikow:
This is a proof of concept for doing text measuring QuickTextMeasurer style. It's different from QuickTextMeasurer because the widths are computed in advance and serialized, which entirely removes the need for PDFKit at runtime. This has the advantage of fixing #1305 – and in general producing the same result everywhere – without having to deploy a copy of Verdana.

The lifting is delegated to these three libraries:

- https://github.com/metabolize/char-width-table-builder
- https://github.com/metabolize/char-width-table-consumer
- https://github.com/metabolize/anafanafo

I'd be happy to move these into the badges org if folks want to collaborate on maintaining them.

QuickTextMeasurer took kerning pairs into account, whereas this implementation does not. I was thinking kerning would be a necessary refinement, though maybe this will work well enough?

I dropped in a binary-search package to traverse the data structure, in part to conserve space. We should make sure this isn't slower than the existing Map-based implementation. I didn't benchmark it yet.

Question: Do we want the npm package to support fonts other than Verdana, perhaps with pdfkit as a peer dependency?

If this works it will obviously need heavier cleanup. It will also need rebasing after #2300 is merged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "url": "https://github.com/badges/shields"
   },
   "dependencies": {
+    "anafanafo": "^0.1.0",
     "camp": "~17.2.2",
     "chalk": "^2.4.1",
     "check-node-version": "^3.1.0",
--- a/lib/text-measurer.js
+++ b/lib/text-measurer.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const PDFDocument = require('pdfkit')
+const anafanafo = require('anafanafo')
 
 class PDFKitTextMeasurer {
   constructor(fontPath, fallbackFontPath) {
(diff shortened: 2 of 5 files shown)


paulmelnikow:
After thinking about this some more, I realized we've never really supported fonts other than Verdana since all the templates refer to Verdana. Passing a font other than Verdana wouldn't produce a badge in that font; it would produce a Verdana badge with an incorrect width. I'll push this forward based on that.

chris48s:
sorry. This needs a giant rebase now :(

In general I like this approach and it probably solves several issues although I haven't really looked at the new packages you've farmed out the font width computation to. I do like the number of lines added/lines removed ratio :)

I think given how much merge conflict/rebase action there is outstanding its probably best to defer detailed comments as any line diff comments are likely to change.

paulmelnikow:
Have labeled this a blocker, as it will cause conflicts in many files and affects future work on the npm package.

package.json
```diff
@@ -22,6 +22,7 @@
     "url": "https://github.com/badges/shields"
   },
   "dependencies": {
+    "anafanafo": "^0.1.0",
```
chris48s: This requirement should be in `gh-badges/package.json` instead of here

gh-badges/lib/index.js
```diff
@@ -1,15 +1,8 @@
 'use strict'
 
-const { makeBadge } = require('./make-badge')
-const { PDFKitTextMeasurer, QuickTextMeasurer } = require('./text-measurer')
+const makeBadge = require('./make-badge')
 
 class BadgeFactory {
-  constructor({ fontPath, fallbackFontPath, precomputeWidths = false }) {
```
chris48s: I know this isn't a backwards-incompatible change as such in that if we pass `fontPath`, `fallbackFontPath` or `precomputeWidths` nothing bad will happen - we will just ignore the values.

Do you reckon its worth also logging a deprecation message or something if we see one of those params passed here?

lib/request-handler.js
```diff
@@ -6,6 +6,7 @@ const request = require('request')
 const { makeBadgeData: getBadgeData } = require('./badge-data')
 const log = require('./log')
 const LruCache = require('../gh-badges/lib/lru-cache')
```
chris48s: Am I right that this is the only place we're using `lru-cache` now? If so we could move it out of the gh-badges package and back into `/lib`

paulmelnikow:
As I mentioned above, I made a tradeoff which is using a space-optimized data structure to hold the char width table.

In performance tests, using `./benchmark_performance.sh` with 1000 iterations, it's apparent the new code is slower, by roughly 28 ms per badge.

master: 0.052393, 0.05744, 0.052076
this branch: 0.089798, 0.083445, 0.086067, 0.080138

We had two optimizations in place which played well to the benchmark:

1. The LRU cache for the left-hand side, which means the benchmark only computes the left-hand width once. It's hard to know how much benefit we get from the LRU cache in production. It's probably faster, though there's no way to know. The cache certainly plays very well to the benchmark: it cuts the measuring work in half.
2. A fast data structure for the most commonly used (ASCII) characters. Since that is likely to be matched in production, I'll add that optimization to char-width-table-consumer and see how much it helps. It should play well to the benchmark, too.

gh-badges/lib/index.js
paulmelnikow: Yea, adding a deprecation warning makes sense.

package.json
paulmelnikow: Oops, thanks. That was a merge error.

lib/request-handler.js
paulmelnikow: Yep, makes sense. Though based on https://github.com/badges/shields/pull/2311#issuecomment-439182704 it may need to come back.

chris48s: OK. Happy to leave it where it is for now if we're not sure yet

paulmelnikow:
I'm a little nervous about merging this, in part because it's a little hard to revert. That said, the two optimizations I'm proposing will be easy changes. I'm game to give it a shot.  🚀🤞

PR #1838, opened 2018-08-02
Add [gitlab] badges

LVMBDV:
I have implemented the following badges for GitLab, mainly focused on `gitlab.com` but most of it should work with self-hosted instances as well. Hopefully, this resolves #541.

- star count
- fork count
- contributor count
- open/closed issue count
- open/closed merge request count
- last build status
- top language
- language count
- last activity date

Issue #541, opened 2015-09-27
Badge request: Gitlab CI and Gitlab


tankerkiller125:
It would be awesome if Shields supported Gitlab and Gitlab CI. They have a Decent API and it should be 100% feasible.

espadrine:
Thanks for the suggestion! I believe they detail their API [here](http://doc.gitlab.com/ce/api/), although I haven't looked into what kind of badges we can provide.

tankerkiller125:
Unfortantly based on the documentation a authentication key is required no matter what. Would it be possible to get a key that Shields can use? Say for example a Shields account on gitlab just for this purpose?

espadrine:
We use `secrets.json` for authentication (for instance, we use it with GitHub). As long as it isn't a per-user authentication, it works.

tankerkiller125:
They only seem to be user specific for private projects (not a suggested thing to badge) internal projects and public projects can be viewed by anyone so long as they have an API key.

ishitatsuyuki:
+1: any plans on this?

moeabdol:
:+1:

LestaD:
Gitlab now supports CI badge!

LestaD:
![image](https://cloud.githubusercontent.com/assets/5620073/14941636/f1a3b93c-0faa-11e6-8f0d-17a28c208888.png)

LestaD:
Maybe add Gitlab issues count?

gluons:
@LestaD But I still prefer Shields badge style. 😅

paulmelnikow:
PR #841 has gone stale and needs work. If anyone would like to adopt it, please do!

sum01:
The v4 API doesn't require authentication for most things. Unsure if that's what's stalling this or not.

LVMBDV:
I wanted to poke around the API on <https://gitlab.com> before I began working on this and I must say it's weird. The documentation is lacking and in some cases, misleading (wrong?).

The docs for [issue](https://docs.gitlab.com/ee/api/issues.html) and [merge request](https://docs.gitlab.com/ee/api/merge_requests.html) state that *every* API call to these functions must be authenticated but I can access them without any authentication on public projects. Try it yourselves:

```
curl -v 'https://gitlab.com/api/v4/projects/coldnight%2fci-test/issues'
curl -v 'https://gitlab.com/api/v4/projects/coldnight%2fci-test/merge_requests'
```

This is good for making badges but still, weird documentation.

[...]

PR #2325, opened 2018-11-15
Add [GitlabPipeline] badge

paulmelnikow:
There's a lot of demand for the Gitlab badges (#541) and the PR has been lingering, so I thought I'd start off one of the simple ones as a new-style service. This one is SVG-based, so it shouldn't require the API-token logic which could use some more testing and will require us to create an app and configure it on our server.

We don't have any validation in place for `queryParams`. Probably this should be added to BaseService, though for the time being I extracted a helper function.

Thanks to @LVMBDV for getting this work started in #1838!


paulmelnikow:
Have labeled this a blocker, as I won't prioritize the work of continuing to rewrite #1838 until this is merged.

services/gitlab/gitlab-pipeline-status.service.js
```diff
@@ -0,0 +1,108 @@
...
+const badgeSchema = Joi.object({
+  message: Joi.alternatives()
+    .try([isPipelineStatus, Joi.equal('unknown')])
+    .required(),
+}).required()
+
+const queryParamsSchema = Joi.object({
+  // TODO This accepts URLs with query strings and fragments, which should be
+  // rejected.
+  gitlab_url: Joi.string().uri({ scheme: ['https'] }),
```
chris48s: In this case, maybe it would be easier to validate and then work out if we've got a queryString (or strip it away) by parsing into a `URLObject` instead of using Joi:

```js
> const { URL } = require('url')
> const url = new URL('http://foo.bar/baz?foo=bar')
> url
URL {
  href: 'http://foo.bar/baz?foo=bar',
  origin: 'http://foo.bar',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'foo.bar',
  hostname: 'foo.bar',
  port: '',
  pathname: '/baz',
  search: '?foo=bar',
  searchParams: URLSearchParams { 'foo' => 'bar' },
  hash: '' }
```

services/gitlab/gitlab-pipeline-status.service.js
```diff
@@ -0,0 +1,108 @@
...
+      // pattern: ':user/:repo/:branch?',
+      queryParams: ['gitlab_url'],
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Gitlab pipeline status',
+        urlPattern: ':user/:repo',
```
chris48s: Now that #2308 is in, lets use `pattern` in these 3 examples (I realise you submitted this PR before we merged that)

services/gitlab/gitlab-pipeline-status.service.js
```diff
@@ -0,0 +1,108 @@
...
+        exampleUrl: 'gitlab-org/gitlab-ce',
+        staticExample: this.render({ status: 'passed' }),
+      },
+      {
+        title: 'Gitlab pipeline status (branch)',
+        urlPattern: ':user/:repo/:branch',
+        exampleUrl: 'gitlab-org/gitlab-ce/master',
+        staticExample: this.render({ status: 'passed' }),
+      },
+      // TODO Find an example of a self-hosted gitlab.
```
chris48s: https://gitlab.gnome.org/ has lots of public repos and probably isn't going to go away in the near future. It might also be a good idea to have a service test that hits a self-hosted instance.

services/gitlab/gitlab-pipeline-status.service.js
paulmelnikow: That could work… though it's a somewhat messy thing to replicate across other services that might use this. Adding some flags to the Joi function seems like a better way to handle it.

Checking for valid URLs is a good start. It doesn't feel important to get this perfect.

services/gitlab/gitlab-pipeline-status.service.js
paulmelnikow: Updated to used `pattern`, and `namedParams`, too.

services/gitlab/gitlab-pipeline-status.service.js
```diff
@@ -56,10 +56,8 @@ module.exports = class GitlabPipelineStatus extends BaseSvgScrapingService {
       {
         title: 'Gitlab pipeline status (self-hosted)',
         pattern: ':user/:repo',
-        namedParams: { user: 'gitlab-org', repo: 'gitlab-ce' },
-        // It would be good to find an example of a publicly accessible, self-
-        // hosted gitlab.
-        query: { gitlab_url: 'https://gitlab.com' },
+        namedParams: { user: 'GNOME', repo: 'pango' },
+        query: { gitlab_url: 'https://gitlab.gnome.org' },
```
chris48s: It might not be directly related to the changes in this PR (I suspect it was introduced in #2308 ), but I noticed that the query gets passed through to the placeholder example but not the one with the params filled in:

screenshot at 2018-11-18 12-46-46

paulmelnikow: Indeed. I noticed that too: #2339

chris48s: one step ahead :athletic_shoe:

paulmelnikow: Since this is approved why don't we merge this. It'll make #2339 easier to review. I probably should have just added that change here.

paulmelnikow:
Yea… they did run, though in #2326 they got new names so I changed the branch protection config, and now GitHub is waiting for the new names. Merging should fix it.

screen shot 2018-11-18 at 9 59 56 am

Issue #2237, opened 2018-10-30
Badge Request: Azure DevOps Code Coverage


calebcartwright:
:clipboard: **Description**

Azure DevOps (formerly known as VSTS) now has publicly accessible APIs that allow for the retrieval of code coverage on builds, so would be great to have a Code Coverage badge for Azure DevOps builds just like the ones that already exist for Jenkins, etc. 

I've got a good feel for the APIs involved to make this happen, so I'd love the opportunity to try to add this functionality myself!

- Which service is this badge for e.g: Azure DevOps Pipelines
- What sort of information should this badge show? Code Coverage

:link: **Data**
- Is there a public API? Yes
- Does the API require an API key? No (for Public Projects), Yes for Private Projects
- https://docs.microsoft.com/en-us/rest/api/vsts/test/code%20coverage/get%20build%20code%20coverage?view=vsts-rest-5.0

chris48s:
> I'd love the opportunity to try to add this functionality myself

Excellent. Sounds like all the data you need is available. We welcome PRs from new contributors. We've got a tutorial on adding new badges that you can follow: https://github.com/badges/shields/blob/master/doc/TUTORIAL.md

We do have a couple of existing badges using the Azure DevOps APIs in https://github.com/badges/shields/blob/master/services/vso/vso.service.js but they use the legacy badge format which we're trying to migrate away from. If you're adding new badges for this service, can you add them using the newer format described in the tutorial. Cheers

calebcartwright:
Perfect, thanks @chris48s . I'll start working on this once I've had a chance to review the provided docs

calebcartwright:
Okay I think I've got the gist of this. A few clarification questions:

1. Looks like someone has since renamed the `vso` services into `azure-devops` services under https://github.com/badges/shields/tree/master/services/azure-devops and updated them to use the new badge format already, correct?
2. Retrieving the code coverage results for a private project will require an account-specific API token. Based on what I've seen in other services, this would only be supported for a self-hosted instance of shields where the user would add his/her specific token to their `private/secret.json` file, correct?
3. Based on the tutorial info, the code for this coverage badge should be placed into its own file within the [azure-devops](https://github.com/badges/shields/tree/master/services/azure-devops) directory. I'm assuming this file should be called `azure-devops-coverage.service.js` given the file names used for other services (i.e. Jenkins), correct?

paulmelnikow:
[...]

PR #2327, opened 2018-11-16
[AzureDevOpsCoverage] Adds Coverage Badge for Azure DevOps

calebcartwright:
Details for new badge described in #2237

The sample project (https://dev.azure.com/totodem/Shields.io/) used as the example for Build & Release Badges didn't have any code coverage so I used some of my own pipelines for the examples.

I also tested this against a few private projects as well to validate


services/azure-devops/azure-devops-coverage.service.js
```diff
@@ -0,0 +1,168 @@
...
+        keywords: ['vso', 'vsts', 'azure-devops'],
+        documentation,
+      },
+    ]
+  }
+
+  static get route() {
+    return {
+      base: '',
+      format: '(?:azure-devops|vso)/coverage/([^/]+)/([^/]+)/([^/]+)(?:/(.+))?',
```
paulmelnikow: Since this is a new badge, let's not support the `vso/` alias. That way you can move `azure-devops` to `base` and remove references to it elsewhere.

services/azure-devops/azure-devops-coverage.service.js
```diff
@@ -0,0 +1,168 @@
...
+      options,
+      schema,
+      errorMessages: {
+        404: 'build pipeline or coverage not found',
+      },
+    })
+  }
+
+  async getLatestBuildId(organization, project, definitionId, branch, options) {
+    let url = `https://dev.azure.com/${organization}/${project}/_apis/build/builds?definitions=${definitionId}&$top=1&api-version=5.0-preview.4`
```
paulmelnikow: Could you pass this to `request` through a `options.qs` object instead? That way the parameters will be correctly url-encoded.

I realize `options` is being generated through a helper. Since the options will then depend on the request, perhaps it would be clearer for the helper to return either the `headers` object or the auth header. Then this could be e.g.:

```js
options: {
  qs: { ... }
  headers: { Authorization: createAuthHeader() }
}
```

services/azure-devops/azure-devops-coverage.tester.js
```diff
@@ -0,0 +1,215 @@
...
+const firstLinesCoverage = `${(
+  (firstLinesCovStat.covered / firstLinesCovStat.total) *
+  100
+).toFixed(0)}%`
+const multiLinesTotal = firstLinesCovStat.total + secondLinesCovStat.total
+const multiLinesCovered = firstLinesCovStat.covered + secondLinesCovStat.covered
+const multiLinesCoverage = `${(
+  (multiLinesCovered / multiLinesTotal) *
+  100
+).toFixed(0)}%`
```
paulmelnikow: Would it be possible to simplify the above and make it more direct by using literal or explicit fixtures? While it's nice to use symbolic names to DRY up assertions, as I'm not familiar with this API, it's difficult to follow what's happening.

services/azure-devops/azure-devops-coverage.service.js
```diff
@@ -0,0 +1,168 @@
...
+</p>
+<p>
+  Optionally, you can specify a named branch:
+  <code>https://img.shields.io/azure-devops/coverage/ORGANIZATION/PROJECT/DEFINITION_ID/NAMED_BRANCH.svg</code>.
+</p>
+`
+const {
+  coveragePercentage: coveragePercentageColor,
+} = require('../../lib/color-formatters')
+
```
paulmelnikow: Could you add a comment with a link to the API doc?

services/azure-devops/azure-devops-coverage.service.js
```diff
@@ -0,0 +1,168 @@
...
+        documentation,
+      },
+      {
+        title: 'Azure DevOps coverage (branch)',
+        urlPattern:
+          'azure-devops/coverage/:organization/:project/:definitionId/:branch',
+        staticExample: this.render({ coverage: 100 }),
+        exampleUrl: 'azure-devops/coverage/swellaby/opensource/25/master',
+        keywords: ['vso', 'vsts', 'azure-devops'],
+        documentation,
```
paulmelnikow: 👍

services/azure-devops/azure-devops-coverage.tester.js
```diff
@@ -0,0 +1,215 @@
...
+const macDefinitionId = 26
+const windowsDefinitionId = 24
+const nonExistentDefinitionId = 234421
+const buildId = 946
+const uriPrefix = `/${service}/${noun}/${org}/${project}`
+const azureDevOpsApiBaseUri = `https://dev.azure.com/${org}/${project}/_apis`
+const mockBadgeUriPath = `${uriPrefix}/${macDefinitionId}.json`
+const mockLatestBuildApiUriPath = `/build/builds?definitions=${macDefinitionId}&$top=1&api-version=5.0-preview.4`
+const mockCodeCoverageApiUriPath = `/test/codecoverage?buildId=${buildId}&api-version=5.0-preview.1`
+const percentageRegex = /^[1-9][0-9]?%|^100%|^0%$/
```
paulmelnikow: This looks like a better version of `isIntegerPercentage` from `test-validators`. 😁Could you use + update that regex instead?

services/azure-devops/azure-devops-coverage.service.js
calebcartwright: Makes sense, will do

services/azure-devops/azure-devops-coverage.service.js
calebcartwright: Good idea

services/azure-devops/azure-devops-coverage.tester.js
calebcartwright: Yup, will do. My initial motivation was as much about trying to avoid doing the math myself as anything else but making this more readable as you suggest will be better

services/azure-devops/azure-devops-coverage.tester.js
calebcartwright: Sigh.. I checked test-validators at least 4 times last night and somehow managed to miss `isIntegerPercentage`  🤦‍♂️ Will make the updates

services/azure-devops/azure-devops-coverage.service.js
```diff
@@ -100,27 +100,36 @@ module.exports = class AzureDevOpsCoverage extends BaseJsonService {
...
       errorMessages: {
         404: 'build pipeline or coverage not found',
       },
     })
   }
 
-  async getLatestBuildId(organization, project, definitionId, branch, options) {
-    let url = `https://dev.azure.com/${organization}/${project}/_apis/build/builds?definitions=${definitionId}&$top=1&api-version=5.0-preview.4`
+  async getLatestBuildId(organization, project, definitionId, branch, headers) {
+    // Microsoft documentation: https://docs.microsoft.com/en-us/rest/api/azure/devops/build/builds/list?view=azure-devops-rest-5.0
```
paulmelnikow: 👍

services/azure-devops/azure-devops-coverage.service.js
```diff
@@ -100,27 +100,36 @@ module.exports = class AzureDevOpsCoverage extends BaseJsonService {
...
+    const url = `https://dev.azure.com/${organization}/${project}/_apis/build/builds`
+    const options = {
+      qs: {
+        definitions: definitionId,
+        $top: 1,
+        'api-version': '5.0-preview.4',
+      },
```
paulmelnikow: 👍

services/azure-devops/azure-devops-coverage.tester.js
calebcartwright: I did some condensing here but was wondering if you're asking instead to use literals in each test? Happy to make that change if so. 

I definitely learned what the Tutorial means when it says that nock is "fussy" so once I got those paths just right I wanted to ensure it would be easy to update them as the paths/query params/etc. change as the Previews of the the underlying APIs progress

calebcartwright:
Oops. I forgot to update the example URIs after changing the base path

paulmelnikow:
Thanks so much!

PR #2333, opened 2018-11-17
Deprecate old [StaticBadge] using a redirect

paulmelnikow:
Now that the static badge has been moved in #2284, next in line for cleaning out `server.js` is this “static badge, old format.” I imagine this route is _very, very old_. (I wouldn’t be surprised if it’s not used at all. I’d be curious to see some stats on that endpoint. If it's not regularly getting requests I could see dropping it.)

In the case of URLs which have permanently changed, an approach I’d like to try is issuing a 301 Redirect.

The benefit is that if a user pastes the URL into the address bar while they are previewing or editing it, the browser will replace the address with the corrected URL when it loads. I figure this will cause some people to update their URLs with no effort, simply because they previewed the badge in their browser, and others to change over, if they notice it.

We incur a slight cost, which is a second request. However many browsers cache the 301’s indefinitely, and we can set an effectively infinite cache duration so the CDN and most other downstream caches will keep them a long time. And handling the redirect is extremely cheap.

This is a nice way to preserve backward compatibility of old routes without having to complicate the new route, such as in the case of vso -> azure-devops. For maintenance purposes, the route that redirects can effectively be treated separately.

It’s also a nice, gentle, and confidence-inspiring way to signal that users should update their URLs.

We could generalize this code, though I think this is a good place to start. This route is tricky because it needs to be loaded last, complicating a reusable solution.


paulmelnikow:
This pull request **introduces 1 alert** when merging 99781bc85440fbb38993d8b5bd595c628a4f3891 into 065dd570ad1ee4fc5c8436049a1d8b90f987ff49 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-ff7cf44de6123b91008f2a4c7098285d511ed9a5)

**new alerts:**

---

PR #1525, opened 2018-02-25
[badge-json] new "badge-json" service

bkdotcom:
fixes #1519 
closes #1752 

route: /^\/json\/(https?.+)\.(svg|png|gif|jpg|json)$/
added documentation to usage.js (with json-badge-maker)

a few tweaks to documentation 
* better separate "static", "dynamic", and "json"
* added "style" parameter to parameter list & moved rendered styles from above the list to below
* removed the ginormous margin-left: 25% style given to UL -> now applies to ul.options   (there was only one UL in in documentation)
* renamed duplicate "miscellaneous" id to "miscellaneous2"

Screenshot of updated docs:
shield-doc

PR #2336, opened 2018-11-17
replace calls to makeBadge() with BadgeFactory.create()

chris48s:
This PR replaces all the calls to `makeBadge()` with `BadgeFactory.create()` and converts the relative imports to package imports so we are depending on the public interface of our package rather than its internals. PR #2311 has also been a big help with this.

There are still some relative imports left for
* `gh-badges/lib/colorscheme.json`
* `gh-badges/lib/lru-cache` and
* `gh-badges/lib/svg-to-img`

### DIFF
--- a/lib/request-handler.js
+++ b/lib/request-handler.js
@@ -6,11 +6,13 @@ const request = require('request')
 const { makeBadgeData: getBadgeData } = require('./badge-data')
 const log = require('./log')
 const LruCache = require('../gh-badges/lib/lru-cache')
-const makeBadge = require('../gh-badges/lib/make-badge')
+const { BadgeFactory } = require('gh-badges')
 const analytics = require('./analytics')
 const { makeSend } = require('./result-sender')
 const queryString = require('query-string')
 const { Inaccessible } = require('../services/errors')
+
+const bf = new BadgeFactory()
 
 // We avoid calling the vendor's server for computation of the information in a
 // number of badges.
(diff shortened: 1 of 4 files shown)


paulmelnikow:
Is this moving in the wrong direction, I wonder? The BadgeFactory isn't really needed; it's providing backward compatibility. Would it be better to export `makeBadge` to provide a path toward deprecation?

chris48s:
Yeah good question. Now that we've removed the LRU cache we don't need to store state between `create()`s, so at the moment we've got a constructor that does nothing and a method, which is basically.. a function. So we've got:

```js
class BadgeFactory {
  constructor(options) {
    //bla
  }

  create(format) {
    return makeBadge(format)
  }
}
```

Just trying to think another step or 2 ahead, if we implement your suggestion in https://github.com/badges/shields/pull/2200#discussion_r228720000 of having a schema to validate the input object, do you think it is useful to have this in a class as a container if we want to be able to change this to something like:

```js
const schema = Joi.stuff()

class BadgeFactory {
  constructor(options) {
    //bla
  }

  validate(format) {
    //bla
  }

  create(format) {
    return makeBadge(this.validate(format))
  }
}
```

?

paulmelnikow:
It seems like `makeBadge` should validate its inputs directly. The only reason I can think of to keep the factory is that it lets us change the interface to `makeBadge` without breaking changes. For that to help, though, we'd need to keep `makeBadge` out of the public interface.

Maybe it would be better to add a schema version parameter to `makeBadge`'s argument. `{ schemaVersion: 0 }` could be the one everyone is using now; `{ schemaVersion: 'next' }` would be semver unstable (a.k.a. whatever we feel like), and when we settle on something that would become `{ schemaVersion: 1 }`.

For something like #1525, schema versions would be pretty darn useful, so we'd get some extra mileage for the effort.

chris48s:
OK. Lets put this one on hold and leave it as a relative import for now

PR #2339, opened 2018-11-17
Examples: Fix queryParams with namedParams

paulmelnikow:
These tests clearly should be refactored, though I’d like to get this fix in, in advance of a more involved refactor that shifts most of the responsibility away from this function. Maybe we can even eliminate at least one of these cases in the meantime.

Deploy blocker.


services/base.js
```diff
@@ -175,21 +175,24 @@ class BaseService {
       let outPreviewUrl
       let outPattern
       if (namedParams) {
-        outExampleUrl = this._makeFullUrlFromParams(pattern, namedParams)
         outPreviewUrl = this._makeStaticExampleUrl(staticExample)
         outPattern = `${this._dotSvg(this._makeFullUrl(pattern))}${suffix}`
+        outExampleUrl = `${this._makeFullUrlFromParams(
+          pattern,
+          namedParams
+        )}${suffix}`
```
paulmelnikow: This is the operative change: `${suffix}` was missing.

The rest is changing the order of the lines to make this more "obviously correct."

paulmelnikow: Yea, that's a nice idea. I like the idea of putting `?:branch=:branch`. Though it's a new idea, and will take a little effort to implement.

This is a fix for the bug you saw in #2325, which is obviously a bug, and a blocker. I think merging that will clarify it. In retrospect given the dependency I should have made this change in that PR.

chris48s: Agreed. We can fix this now and make the other improvement later :+1:

PR #2350, opened 2018-11-17
Friendlier test failures when services won’t register

paulmelnikow:
Avoids crashes in `server.spec.js` accompanied by this cryptic message:

```
Error: listen EADDRINUSE :::1111
  at Object._errnoException (util.js:1022:11)
  at _exceptionWithHostPort (util.js:1044:20)
  at Camp.setupListenHandle [as _listen2] (net.js:1367:14)
  at listenInCluster (net.js:1408:12)
  at doListen (net.js:1517:7)
  at _combinedTickCallback (internal/process/next_tick.js:141:11)
  at process._tickDomainCallback (internal/process/next_tick.js:218:9)
```

e.g. https://circleci.com/gh/badges/shields/24094


server.spec.js
```diff
@@ -9,9 +9,43 @@ const isSvg = require('is-svg')
...
+    //   at Object._errnoException (util.js:1022:11)
+    //   at _exceptionWithHostPort (util.js:1044:20)
+    //   at Camp.setupListenHandle [as _listen2] (net.js:1367:14)
+    //   at listenInCluster (net.js:1408:12)
+    //   at doListen (net.js:1517:7)
+    //   at _combinedTickCallback (internal/process/next_tick.js:141:11)
+    //   at process._tickDomainCallback (internal/process/next_tick.js:218:9)
+    //
+    // Putting it here prevents the tests in this `describe()` from running at
+    // all.
```
chris48s: Change looks great, but I found this comment slightly confusing. Can we clarify that it prevents the tests in this `describe()` from running _if a service fails to register_

paulmelnikow: Does that look better?

paulmelnikow:
Thanks for pointing out what needed clarification!

PR #2353, opened 2018-11-18
refactor [conda] service

chris48s:


services/conda/conda-downloads.service.js
```diff
@@ -1,27 +1,46 @@
...
-module.exports = class CondaDownloads extends LegacyService {
+module.exports = class CondaDownloads extends BaseCondaService {
   static get category() {
     return 'downloads'
   }
 
   static get route() {
     return {
       base: 'conda',
+      pattern: ':which(d|dn)/:channel/:pkg',
```
chris48s: I'm not sure if we actually need the `(d|dn)` or `(p|pn)` params. The behaviour of both the `p`/`d` and `pn`/`dn` versions is tested, but there are no examples in the front-end for `d` or `p` :thinking:

services/conda/conda.tester.js
```diff
@@ -87,27 +86,13 @@ t.create('downloads (skip prefix, relabel)')
 
 t.create('unknown package')
   .get('/d/conda-forge/some-bogus-package-that-never-exists.json')
-  .expectJSON({ name: 'conda|downloads', value: 'invalid' })
+  .expectJSON({ name: 'conda', value: 'not found' })
 
 t.create('unknown channel')
   .get('/d/some-bogus-channel-that-never-exists/zlib.json')
-  .expectJSON({ name: 'conda|downloads', value: 'invalid' })
+  .expectJSON({ name: 'conda', value: 'not found' })
```
chris48s: We use the label from `defaultBadgeData()` here now. As with some other recent badges, I think we can live with losing the customisation in the error case :man_shrugging:

services/conda/conda-platform.service.js
```diff
@@ -1,27 +1,43 @@
...
   static get examples() {
     return [
       {
         title: 'Conda',
-        previewUrl: 'pn/conda-forge/python',
+        namedParams: { channel: 'conda-forge', package: 'python' },
+        pattern: 'pn/:channel/:package',
+        staticExample: this.render({
+          which: 'pn',
+          platforms: 'linux-64 | win-32 | osx-64 | win-64',
```
paulmelnikow: The `|` concatenation feels like a rendering concern. What do you think about passing the array into `render()`?

services/conda/conda.tester.js
```diff
@@ -87,27 +86,13 @@ t.create('downloads (skip prefix, relabel)')
...
 
 t.create('unknown info')
```
paulmelnikow: This test seems unnecessary. This was enforced by the regex / pattern, and still is.

services/conda/conda-downloads.service.js
paulmelnikow: It seems the only difference between `xn` and `x` is in the label, and the developer picks one or the other based on whether they're publishing to the default channel. (I wonder if we can't get that information from the API.) It's honestly a bit confusing to me; wouldn't mind getting clarification from some developers who use it. Not necessary for this PR though.

PR #2360, opened 2018-11-18
Refactor cache-header handling and config, create NonMemoryCachingBaseService, rewrite [flip]

paulmelnikow:
There's a lot going on in this PR, though it's all interdependent, so the only way I can see to break it up into smaller pieces would be serially.

1. I completely refactored the functions for managing cache headers. These have been added to `services/cache-headers.js`, and in some ways set the stage for the rest of this PR.

    - ~~As~~ There are ample higher-level test of the functionality via `request-handler`~~, there are few unit tests of these functions~~. Refactoring these tests ~~and writing new tests~~ was deferred. Cache headers were previously dealt with in three places:
        - `request-handler.js`, for the dynamic badges. This function now calls `setCacheHeaders`.
        - `base-static.js`, for the static badges. This method now calls the wordy `serverHasBeenUpSinceResourceCached` and `setCacheHeadersForStaticResource`.
        - The bitFlip badge in `server.js`. 👈 This is what set all this in motion. This badge has been refactored to a new-style service based on a new `NoncachingBaseService` which does not use the Shields in-memory cache that the dynamic badges user.
    - I'm open to clearer names for `NoncachingBaseService`, which is kind of terrible. Absent alternatives, I wrote a short essay of clarification in the docstring. 😝 

2. In the process of writing `NoncachingBaseService`, I discovered it takes several lines of code to instantiate and invoke a service. These would be duplicated in three or four places in production code, and in lots and lots of tests. I kept the line that goes from regex to namedParams (for reasons) and moved the rest into a static method called `invoke()`, which instantiates and invokes the service. This _replaced_ the instance method `invokeHandler`.
    - I gently reworked the unit tests to use `invoke` instead of `invokeHandler`– generally for the better.
    - I made a small change to `BaseStatic`. Now it invokes `handle()` async as the dynamic badges do. This way it could use `BaseService.invoke()`.

3. There was logic in `request-handler` for processing environment variables, validating them, and setting defaults. This could have been lifted whole-hog to `services/cache-headers.js`, though I didn't do that. Instead I moved it to `server-config.js`. Ideally `server-config` is the only module that should access `process.env`. This puts the defaults and config validation in one place, decouples the config schema from the entire rest of the application, and significantly simplifies our ability to test different configs, particularly on small units of code. (We were doing this well enough before in `request-handler.spec`, though it required mutating the environment, which was kludgy.) Some of the `request-handler` tests could be rewritten at a higher level, with lower-level data-driven tests directly against `cache-headers`.


lib/request-handler.js
```diff
@@ -11,6 +11,7 @@ const analytics = require('./analytics')
 const { makeSend } = require('./result-sender')
 const queryString = require('query-string')
 const { Inaccessible } = require('../services/errors')
+const { setCacheHeaders } = require('../services/cache-headers')
```
paulmelnikow: The functionality removed from this file has been moved here ^^

lib/request-handler.js
```diff
@@ -89,31 +73,23 @@ function getBadgeMaxAge(handlerOptions, queryParams) {
 // (undesirable and hard to debug).
 //
 // Pass just the handler function as shorthand.
-function handleRequest(handlerOptions) {
+function handleRequest(cacheHeaderConfig, handlerOptions) {
```
paulmelnikow: Instead of reading from the environment, inject the config (via `BaseService` and `LegacyService`, in turn via `server.js`).

lib/request-handler.spec.js
```diff
@@ -32,8 +30,6 @@ describe('The request handler', function() {
   before(analytics.load)
 
   let camp
-  const initialBadgeMaxAge = process.env.BADGE_MAX_AGE_SECONDS
```
paulmelnikow: These tests now invoke `handleRequest` with different config instead of a modified environment.

lib/request-handler.spec.js
```diff
@@ -82,38 +79,50 @@ describe('The request handler', function() {
...
+          handleRequest(
+            cacheHeaderConfig,
+            (queryParams, match, sendBadge, request) => {
+              ++handlerCallCount
+              fakeHandler(queryParams, match, sendBadge, request)
+            }
+          )
         )
-      })
+      }
```
paulmelnikow: The next few tests are unchanged. Merely the indent level changed.

lib/request-handler.spec.js
```diff
@@ -143,25 +152,18 @@ describe('The request handler', function() {
...
-        process.env.BADGE_MAX_AGE_SECONDS = 0
+        register({ cacheHeaderConfig: { defaultCacheLengthSeconds: 0 } })
         const res = await fetch(`${baseUri}/testing/123.json`)
         expect(res.headers.get('expires')).to.equal(res.headers.get('date'))
         expect(res.headers.get('cache-control')).to.equal(
           'no-cache, no-store, must-revalidate'
         )
       })
 
-      it('should set the expires header to current time + 120 if BADGE_MAX_AGE_SECONDS not set', async function() {
```
paulmelnikow: This functionality was removed. Responsibility for setting the default is now with `server-config`.

lib/request-handler.spec.js
```diff
@@ -205,14 +207,4 @@ describe('The request handler', function() {
       })
     })
   })
-
-  describe('getBadgeMaxAge function', function() {
```
paulmelnikow: This test was moved to `cache-headers.spec`.

lib/server-config.js
```diff
@@ -63,14 +63,19 @@ const config = {
       baseUri: process.env.GITHUB_URL || 'https://api.github.com',
       debug: {
         enabled: envFlag(process.env.GITHUB_DEBUG_ENABLED, false),
-        intervalSeconds: process.env.GITHUB_DEBUG_INTERVAL_SECONDS || 300,
+        intervalSeconds:
+          parseInt(process.env.GITHUB_DEBUG_INTERVAL_SECONDS) || 300,
```
paulmelnikow: Wrong, so changed while I was at it.

services/base-json.spec.js
```diff
@@ -35,44 +35,44 @@ class DummyJsonService extends BaseJsonService {
...
-        { handleInternalErrors: false }
-      )
     })
 
     it('invokes _sendAndCacheRequest', async function() {
-      await serviceInstance.invokeHandler({}, {})
+      await DummyJsonService.invoke(
+        { sendAndCacheRequest },
+        { handleInternalErrors: false }
+      )
```
paulmelnikow: The above is an example of the changed API. `new DummyJsonService` followed by `serviceInstance.invokeHandler` are replaced by a single call to `DummyJsonService.invoke`.

services/base-json.spec.js
```diff
@@ -35,44 +35,44 @@ class DummyJsonService extends BaseJsonService {
...
-        {
-          headers: { Accept: 'application/json' },
-        }
+        { headers: { Accept: 'application/json' } }
       )
     })
 
     it('forwards options to _sendAndCacheRequest', async function() {
-      Object.assign(serviceInstance, {
+      class WithOptions extends DummyJsonService {
```
paulmelnikow: The only loss due to the above is that we have to build a new class; we can't modify the service instance.

services/base-json.spec.js
```diff
@@ -91,12 +91,12 @@ describe('BaseJsonService', function() {
...
-        { handleInternalErrors: false }
-      )
-      const serviceData = await serviceInstance.invokeHandler({}, {})
-      expect(serviceData).to.deep.equal({
+      expect(
+        await DummyJsonService.invoke(
+          { sendAndCacheRequest },
+          { handleInternalErrors: false }
+        )
+      ).to.deep.equal({
```
paulmelnikow: This and the tests below are the same as before.

services/base-noncaching.js
```diff
@@ -0,0 +1,50 @@
...
+// `lib/request-handler.js` and `BaseService.prototype.register()`.
+//
+// In contrast, services deriving from `BaseNoncachedService` are _not_ cached
+// on the server. This means that each request that hits the server triggers
+// another call to the handler. This makes them more useful for server
+// diagnostics.
+module.exports = class NoncachingBaseService extends BaseService {
+  static register({ camp }, serviceConfig) {
+    const { cacheHeaders: cacheHeaderConfig } = serviceConfig
+    const { _cacheLength: serviceCacheLengthSeconds } = this
```
paulmelnikow: This is effectively generalizing the logic that came from the bit flip handler in `server.js`. I set `debug` services default `_cacheLength` to 60 seconds.

services/base-static.js
```diff
@@ -4,53 +4,49 @@ const makeBadge = require('../gh-badges/lib/make-badge')
...
 const analytics = require('../lib/analytics')
 const BaseService = require('./base')
-
-const serverStartTime = new Date(new Date().toGMTString())
+const {
+  serverHasBeenUpSinceResourceCached,
+  setCacheHeadersForStaticResource,
+} = require('./cache-headers')
 
 module.exports = class BaseStaticService extends BaseService {
```
paulmelnikow: There are no changes in behavior here. It's a straight refactor of functionality.

services/base-static.js
```diff
@@ -4,53 +4,49 @@ const makeBadge = require('../gh-badges/lib/make-badge')
...
-  // Note: Since this is a static service, it is not `async`.
-  handle(namedParams, queryParams) {
-    throw new Error(`Handler not implemented for ${this.constructor.name}`)
-  }
-
   static register({ camp }, serviceConfig) {
-    camp.route(this._regex, (queryParams, match, end, ask) => {
+    const {
+      profiling: { makeBadge: shouldProfileMakeBadge },
+    } = serviceConfig
```
paulmelnikow: This is a teensy optimization. It avoids dotting into `serviceConfig.profiling.makeBadge` – twice! – every time this is called.

services/base-svg-scraping.spec.js
```diff
@@ -56,33 +56,30 @@ describe('BaseSvgScrapingService', function() {
...
-        { handleInternalErrors: false }
-      )
     })
 
     it('invokes _sendAndCacheRequest with the expected header', async function() {
-      await serviceInstance.invokeHandler({}, {})
+      await DummySvgScrapingService.invoke(
+        { sendAndCacheRequest },
+        { handleInternalErrors: false }
+      )
```
paulmelnikow: Same basic idea as the changes in `request-handler`.

services/base.js
```diff
@@ -11,6 +11,7 @@ const {
   InvalidParameter,
   Deprecated,
 } = require('./errors')
+const coalesce = require('../lib/coalesce')
```
paulmelnikow: (I ran out of time, but I'll add a few comments to this file and the ones below when I can get back to it.)

services/cache-headers.js
```diff
@@ -0,0 +1,92 @@
...
+
+const assert = require('assert')
+const coalesce = require('../lib/coalesce')
+
+const serverStartTimeGMTString = new Date().toGMTString()
+const serverStartTimestamp = Date.now()
+
+const numRegex = /^[0-9]+$/
+function isInt(number) {
+  return number !== undefined && numRegex.test(number)
```
chris48s: Why can't we just use `Number.isInteger` here?

services/base.js
```diff
@@ -321,20 +319,28 @@ class BaseService {
...
-      emojic.womanCook,
-      'Service class',
-      this.constructor.name
-    )
+  static async invoke(
+    context = {},
+    config = {},
+    namedParams = {},
+    queryParams = {}
+  ) {
```
chris48s: This is probably a slightly subjective point, but I'd be inclined to leave most of this method in an `invokeHandler()` method so it is easier to write something like:

```js
const service = new Service()
service.property = 'value'
service.invokeHandler()
```

If we retain `invokeHandler()` as a method and keep `invoke()` thin so its implementation is like:


```js
static async invoke(
  context = {},
  config = {},
  namedParams = {},
  queryParams = {}
) {
  const serviceInstance = new this(context, config)
  return await serviceInstance.invokeHandler(namedParams, queryParams)
}
```

That way we retain that flexibility if we need it, but you can still call `await Service.invoke()` in the majority of cases where you don't need to do that. I can see in the cases where we need to mock something you've handled it by extending the class to override a method, but I think keeping `invoke()` thin is probably a useful property. What do you think?

paulmelnikow: Yea… agreed, subjective.

I'm not sure where outside of tests on BaseService subclasses we'd need to set these properties on classes, and there the subclass approach is working fine. After all, these are base classes, intended for subclassing.

I was trying to avoid "lasagne code." If spaghetti code is code where the wires are crossed all over the place and you can't follow what's happening, then lasagne code has lots of layers you have to peel through, none of which do much at all.

There was also the matter of naming: `invoke` and `invokeHandler` are basically two names that mean the same thing. I suppose we could name them _both_ invoke, since they are in different namespaces. Or name the class method something like `run` and leave the instance method as `invokeHandler`.

I certainly don't feel strongly about this.

paulmelnikow:
I like `NonMemoryCachingBaseService`. It's wordy but hopefully gets the point across better.

Yea, I like `_downstreamCacheLength`! That makes things really clear. Let's defer changing that until after this PR.

services/cache-headers.js
chris48s: Oh right. We're looking for a string that looks like an int, not an int. Also `Number.isInteger` will accept negative numbers, whereas this will only accept >=0. Perhaps this function is actually testing the right thing but `isInt()` isn't the right name for it.

services/base.js
chris48s: > I'm not sure where outside of tests on BaseService subclasses we'd need to set these properties

I'm mainly thinking of tests. I guess we can always extract this logic at a later date if we need to.

paulmelnikow:
Thanks for reading @chris48s!

@PyvesB This is a pretty big PR. Do you think you'd have time to give it a third set of  👀?

services/base-static.js
PyvesB: What are the advantages over simply doing `const shouldProfileMakeBadge  = serviceConfig.profiling.makeBadge`?

services/cache-headers.js
```diff
@@ -0,0 +1,99 @@
...
+  // The config returns a number so this would only fail if we break the
+  // wiring. Better to fail obviously than silently.
+  assert(defaultCacheLengthSeconds !== undefined)
+
+  const ourCacheLength = coalesce(
+    serviceCacheLengthSeconds,
+    defaultCacheLengthSeconds
+  )
+
+  const { value: { maxAge: overrideCacheLength } = {}, error } = Joi.validate(
```
PyvesB: I find this hard to read. What is the meaning of the additional `= {}`?

services/legacy-service.js
```diff
@@ -8,13 +8,11 @@ class LegacyService extends BaseService {
...
-  static register(
-    { camp, handleRequest: cache, githubApiProvider },
-    serviceConfig
-  ) {
+  static register({ camp, handleRequest, githubApiProvider }, serviceConfig) {
+    const { cache: cacheHeaderConfig } = serviceConfig
     this.registerLegacyRouteHandler({
       camp,
-      cache,
+      cache: (...args) => handleRequest(cacheHeaderConfig, ...args),
```
PyvesB: This is hard to read, what is happening on this line?

services/base-static.js
paulmelnikow: The difference is purely stylistic.

services/cache-headers.js
paulmelnikow: When there's an error, Joi will return something like `{ error: ... }`. You can't destructure undefined, which would result in `TypeError: Cannot destructure property `bar` of 'undefined' or 'null'`. The `= {}` is a default value for `value` that avoids the crash.

This is how destructuring works:

```jsconsole
> var { foo } = undefined
TypeError: Cannot destructure property `foo` of 'undefined' or 'null'.
> var { foo } = {}
undefined
> foo
undefined
```

And how a default value `{ foo } = {}` figures in:

```jsconsole
> var { bar: { foo } } = {}
TypeError: Cannot destructure property `foo` of 'undefined' or 'null'.
> var { bar: { foo } = {} } = {}
undefined
> foo
undefined
```

services/legacy-service.js
paulmelnikow: At a high level, it's partially applying the `cacheHeaderConfig` parameter to the `handleRequest` function, returning a new function. The new function can be called with any number of parameters, and they will all be forwarded to `handleRequest`, though `cacheHeaderConfig` will be inserted at the beginning.

This is pretty ugly, though I've done it here because this is legacy code stapled on top of other legacy code providing a consistent interface to even more legacy code… so it's both temporary, and not really expected to be be pretty.

I just searched for a good explanation of partial application but almost everything I'm finding is filled with functional-programming jargon that makes the concept seem very complicated.

This article gives a pretty good explanation, and a simple example that motivates why you'd want it:

https://scotch.io/tutorials/javascript-functional-programming-explained-partial-application-and-currying

He's using Ramda:

```js
const buildHttpsUri = R.partial(buildUri, ['https'])
```

Though that requires even _more_ context to understand than the native ES6 equivalent:

```js
const buildHttpsUri = (...args) => buildUri('https', ...args)
```

Using [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters#Description) and [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax), it works regardless of the number of arguments. Though if you knew the number of arguments and their names, you could write it more explicitly:

```js
const buildHttpsUri = (foo, bar, baz) => buildUri('https', foo, bar, baz)
```

PyvesB:
Thanks for the additional explanations, those bits of code make more sense now. Let's get this merged! 👍

paulmelnikow:
Thank you both for taking the time to review. 🤞🚀

Issue #2314, opened 2018-11-14
CI doesn't always run the service tests


paulmelnikow:
CircleCI builds often don't correctly detect that they are on a PR. Currently the only way to know whether or not the service tests ran is to inspect the build logs.

That's because there is no way to reliably distinguish branch and PR builds in CircleCI.

Somtimes the cause is having CircleCI set up on the fork where the PR originates, triggering a branch build ahead of time, though we've also had cases where this happened on forks _without_ CircleCI. Plus Dependabot branches are never on a fork.

As it turns out, there is a workaround [suggested by Danger](https://danger.systems/js/guides/getting_started.html): configure CircleCI to run _only on PRs_.

That would necessitate either:

1. A separate CI service – or simply a separate repo – to run the daily tests
2. A PR-triggered bot that triggers the service-test builds on Circle programmatically
3. A cron-triggered bot that triggers the nightly Circle build programmatically

I experimented in #1936 with a radically different alternative, which is to specify the affected services in the commit history. This has an advantage of making development a bit easier, but it didn't really pan out. Maybe that work could be harvested for a dev-only tool.

paulmelnikow:
I went ahead and set up [another repo](https://github.com/badges/daily-tests) with Circle, and got [Coveralls](https://coveralls.io/builds/20137843) working as well. (Thanks for the reminder about that!)

The reason I chose Circle is because #979 was a long-standing issue solved by Circle, and I didn't want to solve _that_ problem again.

I turned on "Only build PRs" for the main repo. I _think_ the service test runner should start working reliably now, without other changes. Though let's see!

paulmelnikow:
This is looking good. Coveralls is looking nice, too!

chris48s:
Just thinking about this is slightly more detail, was the `CI_PULL_REQUEST` variable the thing that we couldn't rely on?

If so, now that we only run on PR, should we also remove the checks on that e.g:

https://github.com/badges/shields/blob/b922fca982fbd19f90f5d034a8bb9c506c650696/.circleci/config.yml#L176

PR #2371, opened 2018-11-19
Circle: Remove obsolete check

paulmelnikow:
As discussed at #2314.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,21 +172,11 @@ jobs:
 
       - run:
           name: Identify services tagged in the PR title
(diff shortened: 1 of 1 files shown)


paulmelnikow:
I think the previous change did. Once there are no branch builds, Circle stops responding to push hooks. Then all the builds are triggered by PR hooks, so reliably are connected to their PRs.

paulmelnikow:
Ah, hrmmm. It might be. But we can fix it, by preventing those tasks from running on master via config (rather than “dynamically.”)

Issue #1064, opened 2017-08-30
CircleCI : use only Status permission for the API calls


Horgix:
**tl;dr : I think the CircleCI badge with token should be able to work with an API token having the "Status" scope only**

When generating an API token for CircleCI (used for private projects badges), it is possible to assign a scope to it. They currently support 3 scopes : "Status", "Build artifacts" and "all".

"Status" scope should be enough to get informations for a badge, and it is the scope used for the "official" CircleCI tokens. Yet, shields.io requires a "all" scope, and will display a "permission denied" badge otherwise. I think that it could and should be limited to API calls allowed by the "Status" permission since it's made for it; and I am a bit concerned about using "all" scope for a badge since it gives full access to the API on a private project...

Horgix:
Any opinion on that?

If you tell me that a PR fixing this would get merged I might implement it 🙂

paulmelnikow:
Hi! This sounds like an excellent idea. If you can work up a pull request and add some tests I would be really happy to review and merge it.

Horgix:
It looks like there is no endpoint on CircleCI that is able to make use of this restricted scope for now 🙁 
I opened a Feature Request to CircleCI here : ["CircleCI / Get branch status with status-only scoped API token "](https://discuss.circleci.com/t/get-branch-status-with-status-only-scoped-api-token/18535)

I'll follow up here when I get some news!

RedSparr0w:
There was a little bit of talk about this awhile ago [here](https://github.com/badges/shields/pull/798#issuecomment-255618281), not sure if anything has changed on CircleCI's end since then though.

Horgix:
Thanks for the input @RedSparr0w , much appreciated 👍 
I can confirm that it works with the "Build Artifacts" scope; however I still think that it would be best to be able to use only a "Status" scoped token, do you agree?

I think I'll let this open until I get some news from CircleCI about the request of a new endpoint being able to work with a status-only scoped token

paulmelnikow:
Huh. How come doesn't the status token doesn't work for our badge?

RedSparr0w:
The status token doesn't seem to do anything from what i can tell, always get a `permission denied` message from the API with the `status` token.
```json
{
  "message" : "Permission denied"
}
```
Haven't found any endpoint which accepts it.
I hope its just an oversight on CircleCI's end but not sure.

chris48s:
[...]

PR #2373, opened 2018-11-19
Fix CircleCI token example

haythem:
The CircleCI token example contains a redundant fragment `/circleci/circleci/:token`.

In addition, I think that you should allow a comment or a hint to be associated with the example or the provider at least to guide the user especially with private repositories.

For instance, in CircleCI, there are 3 token scopes :
1. `status` used for build status
2. `view-builds` for downloading artifacts
3. `all` to grant full read access to the project

Although it might seem that the first type is the one that should be used with the badge, the badge will not render unless an `all` scope token is specified.


paulmelnikow:
Fabulous, thank you!

Agreed the permissions issue should be documented until it's fixed. (#1064)

Would you like to add the documentation? If you put a `documentation` key in the badge examples containing HTML it'll render on the badge popup. Here's an example: https://github.com/badges/shields/blob/master/services/appveyor/appveyor-tests.service.js

haythem:
Thanks @paulmelnikow for merging the PR.

> Would you like to add the documentation ?

Yeah, sure.

Issue #2083, opened 2018-09-16
F-Droid badge not working?


Catfriend1:
Hi,
I'm trying to add a badge for my github project published on f-droid.
URL is: https://img.shields.io/f-droid/v/com.github.catfriend1.syncthingandroid.svg
So far I'm just seeing "app not found", but that's not true.
See here, it's on the repository:
https://f-droid.org/packages/com.github.catfriend1.syncthingandroid/
Please help me on this.
Thanks in advance
Catfriend1

PyvesB:
Hello @Catfriend1,

Our current implementation uses _.txt_ metadata files (see [this code line](https://github.com/badges/shields/blob/ee2ce3747f57a3b04cb49f472233a9d419df3c55/services/f-droid/f-droid.service.js#L10)). Unfortunately your app seems to be one of the rare ones having a _.yml_ metadata file, therefore the badge is not working properly.

Pinging @niccokunzmann as he may be able to assist you further or give us some additional information. 😉 

Cheers,

Pyves

niccokunzmann:
@Catfriend1 I should be easy to add the badge for the .yml files, too. yml can be converted to JSON and from there, we can get all the necessary information.

This line constructs the URL:
https://github.com/badges/shields/blob/master/services/f-droid/f-droid.service.js#L11
```
const url = `https://gitlab.com/fdroid/fdroiddata/raw/master/metadata/${appId}.txt`
```
You can see what to change...

These lines parse the URL:

      // we assume the layout as provided here:
      // https://gitlab.com/fdroid/fdroiddata/raw/master/metadata/axp.tool.apkextractor.txt
      const positionOfCurrentVersionAtEndOfTheFile = metadata.lastIndexOf(
        'Current Version:'
      ) // credits: https://stackoverflow.com/a/11134049
      const lastVersion = metadata.substring(
        positionOfCurrentVersionAtEndOfTheFile
      )
      const match = lastVersion.match(/^Current Version:\s*(.*?)\s*$/m)


I see two ways to go forward which I ask @PyvesB to give input about:

1. Query .txt and if not present, query .yml - this would leave the badge svg url intact
2. create one badge for txt and one for .yml

Catfriend1:
Thanks @niccokunzmann  for your suggestion, looks good to me. I hope it can get implemented.

niccokunzmann:
By the way: We have the tutorial in the docs folder on setup and
implementation of a badge. If you find anything you like to change
[...]

PR #2377, opened 2018-11-20
Improve [F-droid] support to .yml metadata format

vxcamiloxv:
Improve F-droid service adding .yml metadata format, currently, the default metadata format in f-droid is YAML, but I keep the .txt as default to backward compatibility

closes: https://github.com/badges/shields/issues/2083


vxcamiloxv:
@PyvesB currently the standard format is `.yml`, the txt is the old format but used for many apps, the people don't go and dig out the format of the metadata hidden in the GitLab repository, because is not hidden, is public, and who wants to use this badged is a developer or app maintainer that knows in what type of format sending the app, also the `format` name can be reword to `metadata_format`

paulmelnikow:
Thanks @vxcamiloxv for picking this up!

When it's reasonable to do so, we should avoid the developer needing to specify things we can infer. Based on that I agree with @PyvesB with this implementation suggestion (discussed in #2083):

> Query .txt and if not present, query .yml - this would leave the badge svg url intact

Ordinarily I'd hate to do something like this because of the overhead of two requests, however, hopefully this is only a medium-term solution and we'll eventually get F-Droid to update this.

If you'd like to harmonize these two options into one badge – fetching first the yml and then the text if the yml doesn't load – let us know! If not, one of the maintainers can pick it up.

vxcamiloxv:
@paulmelnikow I thought that option but can be impact load time when use the badge, But is fine to you I can do it

paulmelnikow:
It's not an ideal solution, though I think it is better than the alternatives.

One optimization we could offer: try yml then txt by default; and when `metadata_format=txt`, try txt then yml.

PyvesB:
I'm wondering whether we should not try txt then yml rather than the other way around. Even though YAML is the new preferred format, there are currently only 112 of them compared to 2829 in TXT. 😉

vxcamiloxv:
@PyvesB @paulmelnikow I think if `metadata_format` is defined by the user the double request is unnecessary due the user knows the option exists and can change manually between yml and txt

vxcamiloxv:
also, I think is the double request is the option, the `metadata_format` is unnecessary

paulmelnikow:
With double request, we could use `metadata_format` to let the developer hint which one we should fetch first, resulting in a faster badge. By checking both before giving up, that parameter would be optional, and this would only be a hint, so if their format switched, the badge would keep working.

Maybe not worth it; probably just as simple to fetch txt then yml as @PyvesB suggested.

vxcamiloxv:
@paulmelnikow  @PyvesB I added a fallback request when `txt` is not found, and when `metadata_format` is defined only makes a single request to that format

services/f-droid/f-droid.service.js
```diff
@@ -61,14 +94,15 @@ module.exports = class FDroid extends BaseService {
...
 
   static get examples() {
     return [
       {
         title: 'F-Droid',
-        exampleUrl: 'org.thosp.yourlocalweather',
+        exampleUrl: 'org.thosp.yourlocalweather?metadata_format=yml',
         pattern: ':appId',
         staticExample: this.render({ version: '1.0' }),
         keywords: ['fdroid', 'android', 'app'],
```
paulmelnikow: Since `metadata_format` is optional, could you add a second example called something like **F-Droid (explicit metadata format)**, and restore the first one? If you use `query: { metadat_format: 'yml' }` it should render correctly in the frontend.

services/f-droid/f-droid.service.js
```diff
@@ -42,9 +54,30 @@ module.exports = class FDroid extends BaseService {
...
+      {
+        ErrorClass: InvalidParameter,
+        prettyErrorMessage:
+          'invalid parameter, valid metadata_format=yml or txt',
+        traceErrorMessage: 'Query params did not match schema',
+        traceSuccessMessage: 'Query params after validation',
+      },
+      queryParams,
+      queryParamsSchema
+    )
```
paulmelnikow: 👍

services/f-droid/f-droid.service.js
```diff
@@ -1,31 +1,43 @@
...
+const validate = require('../../lib/validate')
 const { addv: versionText } = require('../../lib/text-formatters')
 const { version: versionColor } = require('../../lib/color-formatters')
-const { InvalidResponse } = require('../errors')
+const { InvalidResponse, InvalidParameter } = require('../errors')
 
 module.exports = class FDroid extends BaseService {
-  async fetch({ appId }) {
+  async fetch({ appId, format }) {
     // currently, we only use the txt format. There are few apps using the yml format.
```
paulmelnikow: Could you update this comment to describe the new behavior? It should make the rest of this easier to follow for an unfamiliar reader.

services/f-droid/f-droid.service.js
```diff
@@ -1,31 +1,43 @@
...
     const metadata = buffer.toString()
     // we assume the layout as provided here:
     // https://gitlab.com/fdroid/fdroiddata/raw/master/metadata/axp.tool.apkextractor.txt
-    const positionOfCurrentVersionAtEndOfTheFile = metadata.lastIndexOf(
-      'Current Version:'
-    ) // credits: https://stackoverflow.com/a/11134049
+    const value = (metadata.match(/CurrentVersion:|Current Version:/) || [
+      'Current Version:',
+    ])[0]
+    const positionOfCurrentVersionAtEndOfTheFile = metadata.lastIndexOf(value) // credits: https://stackoverflow.com/a/11134049
```
paulmelnikow: Parsing files manually tends to create extra work and introduce risk. When fetching the yaml version, could you use the yaml parser instead?

It might be easier to split that into separate `fetchText` and `fetchYaml` functions, and to move the retry logic into `handle()`.

services/f-droid/f-droid.tester.js
```diff
@@ -57,36 +57,125 @@ const testString =
...
+  .intercept(nock =>
+    nock(base)
+      .get(`${path}.txt`)
+      .reply(404)
+  )
+  .intercept(nock =>
+    nock(base)
+      .get(`${path}.yml`)
+      .reply(200, testYmlString)
+  )
```
paulmelnikow: Nice!

services/f-droid/f-droid.service.js
```diff
@@ -42,9 +54,30 @@ module.exports = class FDroid extends BaseService {
...
+
+    return constructor.render(result)
+  }
+
+  static validateParams(queryParams) {
+    const queryParamsSchema = Joi.object({
+      metadata_format: Joi.string().valid(['yml', 'txt']),
+    }).required()
+
+    return validate(
```
paulmelnikow: Just a heads up, particularly to the other maintainers, that I've moved this function into `BaseService` in #2399. Presumably this PR will land first and the duplication can be removed in #2399, or later. The changes don't conflict.

paulmelnikow:
Hey @vxcamiloxv it would be great to get this in! Do you think you'd have some time to address the review comments? If not, no problem, I or one of the maintainers can pick it up.

vxcamiloxv:
I'll do it this weekend

services/f-droid/f-droid.service.js
```diff
@@ -1,31 +1,92 @@
...
+
+  async fetchText(url, options) {
+    const { buffer } = await this._request({
+      url: `${url}.txt`,
+      ...options,
     })
     const metadata = buffer.toString()
-    // we assume the layout as provided here:
+    // we assume the txt structure as provided here:
     // https://gitlab.com/fdroid/fdroiddata/raw/master/metadata/axp.tool.apkextractor.txt
```
paulmelnikow: Can the changes below now be reset?

paulmelnikow:
I'll click the button to merge master into this branch. That may cause a couple test failures for things that have changed in BaseService in the meantime.

services/f-droid/f-droid.service.js
vxcamiloxv: done https://github.com/badges/shields/pull/2377/commits/ed53088f3a3e31341fafac5051b2514cc832a6e4

services/f-droid/f-droid.service.js
vxcamiloxv: done https://github.com/badges/shields/pull/2377/commits/5156c821e0669b8c1918a5293069349734eafb8d

vxcamiloxv:
@paulmelnikow  looks like is by `query` example

services/f-droid/f-droid.service.js
paulmelnikow: Sorry, I might not have been clear. I wasn't referring to the comments above, but to the changes in the code below which extract the version from the .txt metadata file. I think those changes were meant to accommodate the .yml format and are no longer needed. Is that right?

vxcamiloxv: yes, I was restore before to see this comment :D https://github.com/badges/shields/pull/2377/commits/3a745b694fd06c7e7fbba77fae9f80e0dff49e81

paulmelnikow:
So it looks like the example with explicit `metadata_format` is not working:

http://shields-staging-pr-2377.herokuapp.com/f-droid/v/org.thosp.yourlocalweather.svg?metadata_format=yml


Is that a txt metadata app?

vxcamiloxv:
yes, is `txt` https://gitlab.com/fdroid/fdroiddata/raw/master/metadata/org.thosp.yourlocalweather.txt

paulmelnikow:
Does it make more sense to pick a different app, or to switch the parameter to `metadata_format=txt`?

vxcamiloxv:
when `metadata_format` is specified only made a single request to speed up the response, but when a format not is specified try txt and then yml

https://shields-staging-pr-2377.herokuapp.com/f-droid/v/org.dystopia.email.svg?metadata_format=yml

Issue #2190, opened 2018-10-19
Packagist version error


Yurunsoft:
URL: https://img.shields.io/packagist/v/yurunsoft/yurun-http.svg
Packagist: https://packagist.org/packages/yurunsoft/yurun-http

My newest version is `v3.0.2`, but the img show `V1.3.5`

chris48s:
Hi. The reason this happens is because we assume versions from packagist follow the packagist version schema defined at https://getcomposer.org/doc/04-schema.md#version but packagist doesn't enforce its own schema - it is possible to publish a release which doesn't follow that pattern. If we get a package which has a non-compliant version number, our sort process will produce a spurious result. I think in your case the issue is that some of your version numbers start with an uppercase `V` (which isn't strictly compliant with the packagist schema) and some start with a lowercase `v` and they're getting sorted differently. Broadly this falls into the same bucket as these issues: https://github.com/badges/shields/issues/1586 , https://github.com/badges/shields/issues/885 - they're symptoms of the same thing.

Issue #1586, opened 2018-03-19
Packagist version badges not being updated


Majkl578:
We have some libraries using Shilds.io badges. The latest version badges are stuck and not being updated.

Package [kdyby/datetime-provider](https://packagist.org/packages/kdyby/datetime-provider):
Version stuck at 1.0.0-BETA1 while current is 1.0.0: [![Latest stable](https://img.shields.io/packagist/v/kdyby/datetime-provider.svg)](https://packagist.org/packages/kdyby/datetime-provider)

Package [kdyby/datetime-provider-bundle](https://packagist.org/packages/kdyby/datetime-provider-bundle):
Version stuck at 1.0.0-BETA2 while current is 1.0.0: [![Latest stable](https://img.shields.io/packagist/v/kdyby/datetime-provider-bundle.svg)](https://packagist.org/packages/kdyby/datetime-provider-bundle)

Both releases are over a month old, yet not updated.

paulmelnikow:
Hi there! Thanks for reporting this.

![](https://img.shields.io/packagist/v/kdyby/datetime-provider.svg)
https://img.shields.io/packagist/v/kdyby/datetime-provider.svg

![](https://img.shields.io/packagist/v/kdyby/datetime-provider-bundle.svg)
https://img.shields.io/packagist/v/kdyby/datetime-provider-bundle.svg

Hmm, **beta** is supposed to be lowercase.

Is this the same issue as #885? It would be great if packagist enforced the [composer versioning scheme](https://getcomposer.org/doc/04-schema.md#version)…

Majkl578:
Imho there's difference between composer version and tag name. Composer parses the tags and creates versions from those. Also upper-case BETA is used in examples directly in Composer's documentation: https://getcomposer.org/doc/articles/versions.md#vcs-tags-and-branches.
Semver also only talks about lexicographic order, not mentioning upper/lower case explicitly: https://semver.org/spec/v2.0.0.html#spec-item-11
Finally the Composer Semver version parser is case-insensitive: https://github.com/composer/semver/blob/1.4.2/src/VersionParser.php#L59

So this may be similar/same issue as #885, but this seems to be an issue in your code (as per the comment there):
> It looks like our code recognizes RC1 but not rc1.

chris48s:
[...]

Issue #885, opened 2017-03-15
Packagist badge does not handle lowercase "rc", only uppercase


kocsismate:
Hi,

I am a happy user of shields.io but I unfortunately noticed a rather annoying bug (?) in the recent days:

Although I released v2.0 of my lib many days ago (here is its read me: https://github.com/woohoolabs/yin/blob/master/README.md), the Packagist final version badge still shows v2.0.0-rc1 (https://img.shields.io/packagist/v/woohoolabs/yin.svg).

Could you tell me why it happens? Is it a bug or I messed up something?

paulmelnikow:
Hi, thanks for reporting this.

Shields has some pretty elaborate logic for sorting the Packagist versions. See https://github.com/badges/shields/issues/319#issuecomment-74417866 when it was discussed.

It looks like [our code recognizes `RC1` but not `rc1`](https://github.com/badges/shields/blob/1ffd5d949c2dada61ff6f049eeb0684ea9a8ea41/server.js#L6468-L6470). Should it?

kocsismate:
Thank you for insights! I would be happy if the check could be made case-insensitive:

```
var firstLetter = modifier.toLowerCase().charCodeAt(0);
// ...
} else if (firstLetter === 114) { // r
// ...
```

Hopefully, this change would make the decision more foolproof.

paulmelnikow:
I just checked [the spec](https://getcomposer.org/doc/04-schema.md#version) and it looks like the RC is supposed to be capitalized. Shields could interpret it leniently, though if we do, we would introduce skew. The composer command line would interpret the values one way, Packagist another, Shields still another, making it more difficult for these tools to interoperate.

Here's a good example of the principle. Most JSON parsers are strict. They croak on inputs which don't meet the spec. Even when the error is completely unambiguous, like a dangling comma, they bail. Counterintuitively, this is a reason JSON has been successful. Because every parser interprets the spec strictly, they can all interoperate with each other.

After laying out the JSON spec in five short paragraphs, Douglas Crockford wrote (emphasis mine):

> That's it. That is all of JSON. JSON's design goals were to be minimal, portable, textual, and a subset of JavaScript. _The less we need to agree on in order to interoperate, the more easily we can interoperate._

Certainly reasonable people could disagree. It may be inconvenient for your package which is already published.

kocsismate:
OK, I completely understand your opinion. However, I think Postel's Law can also be a good guideline:
> Be conservative in what you send, be liberal in what you accept.

[...]

PR #2391, opened 2018-11-25
behave consistently in [packagist] badges when called with no version arg

### LINKED ISSUE #2356: Invalid PHP version on Packagist
**Description:**

PHP version badge on Packagist returns `invalid` for some packages. It works normally for others.

**Working badges:**

PHP from Packagist - https://packagist.org/packages/filips123/monolog-phpmailer/

PHP from Packagist - https://packagist.org/packages/symfony/symfony/

**Invalid badges:**

PHP from Packagist - https://packagist.org/packages/filips123/config-writer

PHP from Packagist - https://packagist.org/packages/slim/slim/

chris48s:
Closes #2356 (although it does change behaviour by assuming we want a stable release by default instead of defaulting to an unstable branch alias).

Making this change doesn't do anything to address #2190, #1586 or #885 - all 3 badges still have the same behaviour in the case where a version like `1.0.0-ALPHA` or `1.0.0-rc` is used, but

`/packagist/l/namespace/package`
`/packagist/v/namespace/package`
`/packagist/php-v/namespace/package`

now all work out latest version in the same way instead of doing 3 different things.


services/packagist/packagist-helpers.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
+
+const {
+  compare: phpVersionCompare,
+  latest: phpLatestVersion,
+  isStable: phpStableVersion,
+} = require('../../lib/php-version')
+
+const getLatestVersion = function(versionsData, stable = true) {
```
paulmelnikow: Could you add some sazerac tests of this function? It's dense with logic so which makes it a little difficult to understand what its behavior is supposed to be.

services/packagist/packagist-helpers.js
```diff
@@ -0,0 +1,49 @@
...
+
+  // Map aliases (eg, dev-master).
+  const aliasesMap = {}
+  versions.forEach(version => {
+    const versionData = versionsData[version]
+    if (
+      versionData.extra &&
+      versionData.extra['branch-alias'] &&
+      versionData.extra['branch-alias'][version]
+    ) {
```
paulmelnikow: This is fine, though you could write the above even more functionally like this:

```js
const aliasesMap = {}
Object.entries(versionsData)
  .filter(
    (version, versionData) =>
      versionData.extra &&
      versionData.extra['branch-alias'] &&
      versionData.extra['branch-alias'][version]
  )
  .forEach((version, versionData) => {
    ...
```

paulmelnikow:
@chris48s Would you be up for addressing these comments?

chris48s:
I think I'm going to bail on this particular one :door: :running_man:

PR #1478, opened 2018-02-01
Add logo style badges

RedSparr0w:
This PR adds 2 new style badges which emphasize the logo on the badge:
- `popout`
- `popout-square`

These allow for a query parameter called `logoPosition`:
_(By default the logo is centered)_
negative numbers moves the badge down/logo up:
`-5`:

positive numbers move the badge up/logo down:
`5`:

If a logo is not specified it will automatically fallback to serve the `flat` or `flat-square` style badge

### Examples:

`popout` | `popout-square`
:--: | :--:
image | image
image | image
image | image
image | image
image | image
image | image
image | image
image | image
image | image
image | image
image | image

PR #2394, opened 2018-11-26
Add a "pill" style

metaa:
I had a look at the theme selection of [cms.js](https://github.com/chrisdiana/cms.js) earlier. When I saw  the author pill in [the middle one in the top row](https://github.com/chrisdiana/cms.js/blob/d20f1ca80065a2855af0b2a033afd5eef868a753/img/themes.png) I thought to myself "hmmm, couldn't we do that with `gh-pages`?"

Since there currently is no such thing as a `pill` style, I added it!

### What does it look like?
The styles section of the front page would yield this:

style-pill-green

I added two pixels of padding (as variable `containerPadding`*) to the left and right so that the more intense rounding doesn't affect the readability of the badge or makes the logo or text seem too close to the borders.

*the logic here may be debatable as it should probably also be applied for other styles(?)

### Why not make the corner radius in `flat` configurable?
I considered this as a solution better than adding a whole new svg template. Maybe it actually makes more sense, but I think that would be something that should be discussed before.

### DIFF
--- a/gh-badges/lib/make-badge.js
+++ b/gh-badges/lib/make-badge.js
@@ -169,8 +169,9 @@ function makeBadge({
     logoPadding = logo ? 3 : 0
   }
 
+  let containerPadding = 0
   if (template === 'pill') {
-    logoPadding += 2
+    containerPadding = 2
   }
 
   const context = {
@@ -185,6 +186,7 @@ function makeBadge({
     colorA,
     colorB,
     escapeXml,
+    containerPadding,
   }
 
   const templateFn = templates[`${template}-${format}`]
--- a/gh-badges/lib/make-badge.js
+++ b/gh-badges/lib/make-badge.js
@@ -169,6 +169,10 @@ function makeBadge({
     logoPadding = logo ? 3 : 0
   }
 
+  if (template === 'pill') {
+    logoPadding += 2
+  }
+
   const context = {
     text: [left, right],
     escapedText: text.map(escapeXml),
--- a/gh-badges/templates/pill-template.svg
+++ b/gh-badges/templates/pill-template.svg
@@ -1,0 +1,39 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="{{=(it.widths[0] -= it.text[0].length ? 0 : (it.logo ? (it.colorA ? 0 : 7) : 11))+it.widths[1]}}" height="20">
+  <linearGradient id="smooth" x2="0" y2="100%">
+    <stop offset="0" stop-color="#bbb" stop-opacity=".1"/>
+    <stop offset="1" stop-opacity=".1"/>
+  </linearGradient>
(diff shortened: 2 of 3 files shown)


chris48s:
Thanks for contributing. I like the style.

Just for completeness, here's an example with no logo:

style-pill-green-nologo

..and one with a hard-corner square logo:

style-pill-green-squarelogo

IMO, in terms of the internals, if we can make this a variation of `flat` by modifying that template and feeding in different params for padding and corner radius, I think that is preferable to having 2 slightly different versions of the same template.

From the user's perspective, it should be `?style=pill` not `?style=flat&cornerRadius=10` though, if you see what I mean.

If you modify the `flat` template you may need to update the snapshot tests slightly.

Do you have any thoughts on this @RedSparr0w ?

metaa:
> From the user's perspective, it should be `?style=pill` not `?style=flat&cornerRadius=10` though, if you see what I mean.

This is where I think that making the corner radius configurable would make the buttons less unique because everyone could then just mix badges with whatever corner radius, but not only having 100% round (`pill`), slightly round-ish (`flat`) and `flat-squared` may also be a good thing.

paulmelnikow:
Thanks for taking this on!

Some tangential thoughts on rewriting the generator…

Since we have template literals, I imagine we could rewrite the templates in plain JavaScript. This would make it easy to assign symbolic names to various calculations, which should improve readability a lot, and also let us DRY up the ones we repeat. It should also simplify loading a bit, drop DoT, and maybe even improve performance.

I agree with @chris48s that the user-facing options should be flat-square and pill. Perhaps a middle path between adding another template and rewriting the whole thing is passing the corner radius as another value in the template context and aliasing pill to flat-square.

RedSparr0w:
Thanks for the PR, I could see this style badge getting some use. :+1: 

I agree with @chris48s points `?style=pill` with an optional `&radius=x` makes the most sense, this would be better to be added to the flat template rather than creating an new template.

>Since we have template literals, I imagine we could rewrite the templates in plain JavaScript. This would make it easy to assign symbolic names to various calculations, which should improve readability a lot, and also let us DRY up the ones we repeat. It should also simplify loading a bit, drop DoT, and maybe even improve performance.

:+1:  Definitely think this could help in the future.

chris48s:
I'm in favour of just `?style=pill` and `?style=flat`. I don't think we need to make the radius user-configurable

chris48s:
topical discussion: https://github.com/badges/shields/pull/1478#issuecomment-446623428

metaa:
I'm a fan of adding things that provoke a cleanup to the mess first, but maybe adding this as a variation of the flat template (or merging and adding options) after #2428 sounds better in this case.

paulmelnikow:
Good by me.

Any interest in taking on #2428 @metaa?

metaa:
@paulmelnikow:
> Any interest in taking on #2428 @metaa?

Sure, I can look into it.

Issue #2345, opened 2018-11-17
Update GitHub manifest + dynamic JSON tests


paulmelnikow:
Two of our GitHub tests depend on the contents of this repo's `package.json`, which has changed.

```
  10) Github
       Package name
         
	[ GET /package-json/n/badges/shields.json ]:

      AssertionError: expected { Object (name, value) } to deeply equal { Object (name, value) }
      + expected - actual

       {
         "name": "package name"
      -  "value": "shields.io"
      +  "value": "gh-badges"
       }
      
      at Object.pathMatch.matchJSON (node_modules/icedfrisby/lib/pathMatch.js:138:38)
      at _expect (node_modules/icedfrisby/lib/icedfrisby.js:583:10)
      at IcedFrisbyNock._invokeExpects (node_modules/icedfrisby/lib/icedfrisby.js:1261:26)
      at start (node_modules/icedfrisby/lib/icedfrisby.js:1244:12)
      at Request.runCallback [as _callback] (node_modules/icedfrisby/lib/icedfrisby.js:1131:16)
      at Request.self.callback (node_modules/request/request.js:185:22)
      at Request.<anonymous> (node_modules/request/request.js:1161:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:1083:12)
      at endReadableNT (_stream_readable.js:1064:12)
      at _combinedTickCallback (internal/process/next_tick.js:139:11)
      at process._tickDomainCallback (internal/process/next_tick.js:219:9)

  11) Github
       Package name - Custom label
         
	[ GET /package-json/name/badges/shields.json?label=Dev Name ]:

      AssertionError: expected { Object (name, value) } to deeply equal { Object (name, value) }
      + expected - actual

       {
         "name": "Dev Name"
      -  "value": "shields.io"
      +  "value": "gh-badges"
       }
      
      at Object.pathMatch.matchJSON (node_modules/icedfrisby/lib/pathMatch.js:138:38)
      at _expect (node_modules/icedfrisby/lib/icedfrisby.js:583:10)
      at IcedFrisbyNock._invokeExpects (node_modules/icedfrisby/lib/icedfrisby.js:1261:26)
      at start (node_modules/icedfrisby/lib/icedfrisby.js:1244:12)
      at Request.runCallback [as _callback] (node_modules/icedfrisby/lib/icedfrisby.js:1131:16)
      at Request.self.callback (node_modules/request/request.js:185:22)
[...]

Issue #1446, opened 2018-01-15
User color overrides Error color


RedSparr0w:
When `colorB` is specified this color will override any of the badges error colorscheme colors.

badge | actual | expected
:-- | :--: | :--:
`/travis/rust-lang/notarepo.svg?colorB=10ADED` | ![](https://img.shields.io/:build-unknown-10ADED.svg) | ![](https://img.shields.io/:build-unknown-9F9F9F.svg)
`/notabadge.svg?colorB=10ADED` | ![](https://img.shields.io/:404-badge%20not%20found-10ADED.svg) | ![](https://img.shields.io/:404-badge%20not%20found-e05d44.svg)
`/notabadge.svg?label=Label&colorB=10ADED` | ![](https://img.shields.io/:Label-badge%20not%20found-10ADED.svg) | ![](https://img.shields.io/:404-badge%20not%20found-e05d44.svg)

paulmelnikow:
Hmm, interesting. Yes, I suppose I agree this is a bug, and that the user's specified color should only  apply when a badge is presented successfully.

RedSparr0w:
I don't think it's too much of a problem, but should probably be fixed in the `node-8` branch along the way.

paulmelnikow:
For whoever may pick this up, there's some useful discussion here: https://github.com/badges/shields/pull/2399#issuecomment-445024154

PR #1721, opened 2018-06-04
[github] update package/manifest for private repos

CynicalBusiness:
This edit allows for the `/github/(package|manifest)-json/` badges to work with private repositories.

I've updated the `raw.githubusercontent.com` URL to use GitHub's own API instead, and this allows it to work for private repositories. The old tests for this badge are still passing.

EDIT: The checks are failing on an unrelated badge, I think? I only ran tests for this one badge and everything looked good there. The failure seems to be for `hit-counter`.

PR #2399, opened 2018-11-26
Rewrite the [DynamicJson] badge; affects [gitlab dynamic-xml dynamic-yaml]

paulmelnikow:
This starts the rewrite of the dynamic badges. I've pulled into BaseService an initial version of the query param validation from #2325.

I've extended from BaseJsonService to avoid duplicating the deserialization logic, though it means there is a bit of duplicated code among the three dynamic services. The way to unravel this would be to move the logic from `_requestJson` and friends from the base classes into functions so DynamicJson can inherit from BaseDynamic. Would that be worth it?

Marked blocker since it affects other code that may validate query parameters (e.g. #2377), the other dynamic badges, and further cleanup of `server.js`.

This introduces a regression of #1446 for this badge.

Ref: #2345 (partial fix)


paulmelnikow:
> The way to unravel this would be to move the logic from `_requestJson` and friends from the base classes into functions so DynamicJson can inherit from BaseDynamic. Would that be worth it?

I've found another case where the coupling of the request logic to the base class is causing problems. #1721 is about using github auth on a particular request, but only when static auth is present (i.e. for self-hosting). So I want the logic from `BaseGithubAuthService` if I'm using static auth, and otherwise what's in `BaseJsonService`.

If these functions were extracted to separate modules, it might leave open more options open for handling this unusual case.

services/dynamic/dynamic-json.tester.js
```diff
@@ -57,8 +52,8 @@ t.create('JSON from url')
   )
   .expectJSON({
     name: 'custom badge',
-    value: 'gh-badges',
-    colorB: colorsB.brightgreen,
+    value: 'shields.io',
+    colorB: 'lightgreen',
```
chris48s: I think the previous value was right. This PR changes the default to 

 - https://shields-staging-pr-2399.herokuapp.com/badge/dynamic/json.svg?url=https%3A%2F%2Fraw.githubusercontent.com%2Fbadges%2Fshields%2Fmaster%2Fpackage.json&label=package&query=$.name

but I think we still want it to be

 - https://img.shields.io/badge/dynamic/json.svg?url=https%3A%2F%2Fraw.githubusercontent.com%2Fbadges%2Fshields%2Fmaster%2Fpackage.json&label=package&query=$.name

chris48s:
> This introduces a regression of #1446 for this badge.

so that is what causes:

 - https://shields-staging-pr-2399.herokuapp.com/badge/dynamic/json.svg?url=https%3A%2F%2Fraw.githubusercontent.com%2Fbadges%2Fshields%2Fmaster%2Fpackage.jso&label=package&query=$.name&colorB=blue

instead of 

 - https://shields-staging-pr-2399.herokuapp.com/badge/dynamic/json.svg?url=https%3A%2F%2Fraw.githubusercontent.com%2Fbadges%2Fshields%2Fmaster%2Fpackage.jso&label=package&query=$.name

right? Have you have any thoughts on how we might solve that? (if its large, it doesn't necessarily have to go into this PR)

services/dynamic/dynamic-json.tester.js
paulmelnikow: Oops! Thanks for catching that.

paulmelnikow:
Yes, exactly right.

> Have you have any thoughts on how we might solve that? (if its large, it doesn't necessarily have to go into this PR)

I suppose we'll need to add an `'isError'` field to `serviceData`. It can be set automatically in `_handleError()`, and manually in `handle()` for rare case like Crates. Then add logic in `_makeBadgeData()` to handle it.

https://github.com/badges/shields/blob/8070f0bef9f9774c7ab75c04e5f8a2d8420ac5d3/services/base.js#L392-L437

I'd like to fix this, though it's a bit fiddly. There are possible merge conflicts to contend with in `base.js` in #2441. It's a minor enough regression that I think we can do it in a follow on. If this bug were important it wouldn't have been open almost a year. 😉

The test I disabled here really belongs in `base.spec.js`, and should get moved as part of #1446.

paulmelnikow:
Gitlab failure seems transient.

chris48s:
> for rare case like Crates

Hrm. On reflection, I think with crates what I did in PR #2462 was not the best way to do it. Perhaps that case would be better handled by throwing a custom `ShieldsRuntimeError` more like this situation:

https://github.com/badges/shields/blob/b2d5d3eccaea0b6f77b3b5868e4115fa2b8dc8d7/services/cdnjs/cdnjs.service.js#L29-L33

If we say we should always be throwing an exception for cases like this, would that allow us to handle this issue one layer further out (where we catch the exceptions and wrap them in a badge)?

> There are possible merge conflicts to contend with
> It's a minor enough regression that I think we can do it in a follow on

Agreed - its just useful to think about how to do that, given its a bit fiddly

paulmelnikow:
> If we say we should always be throwing an exception for cases like this, would that allow us to handle this issue one layer further out (where we catch the exceptions and wrap them in a badge)?

We apply overrides further out than handling errors, which probably makes sense… since on error we should still respect e.g. `?style=flat-square`. Probably there are places even further up, like cache handling, where we may want to consider whether or not it was an error.

Custom `ShieldsRuntimeError` seems nice; it could let us attach the error to what gets bubbled up to `handleRequest`, which could let us adjust the cache settings appropriately for probably recoverable vs probably permanent errors.

Probably we could catch `ShieldsRuntimeError` here…
https://github.com/badges/shields/blob/b2d5d3eccaea0b6f77b3b5868e4115fa2b8dc8d7/services/base.js#L330-L332

PR #2400, opened 2018-11-26
TokenPool: Add debug logging helper

paulmelnikow:
Slice another sliver from #1205.

### DIFF
--- a/lib/token-pool.js
+++ b/lib/token-pool.js
@@ -1,6 +1,19 @@
 'use strict'
 
+const crypto = require('crypto')
 const PriorityQueue = require('priorityqueuejs')
+
+// Compute a one-way hash of the input string.
+function sanitizeToken(id) {
+  return crypto
+    .createHash('sha256')
+    .update(id, 'utf-8')
+    .digest('hex')
+}
+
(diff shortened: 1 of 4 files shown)


lib/token-pool.js
```diff
@@ -160,7 +189,10 @@ class TokenPool {
       }
     }
 
-    while ((next = this.priorityQueue.peek())) {
+    while (
+      !this.priorityQueue.isEmpty() &&
+      (next = this.priorityQueue.peek())
+    ) {
```
paulmelnikow: This is the bug fix ^^

lib/token-pool.spec.js
```diff
@@ -122,4 +174,11 @@ describe('The token pool', function() {
       expectPoolToBeExhausted(tokenPool)
     })
   })
+
+  context('when empty', function() {
+    it('next() should return the expected error', function() {
+      const tokenPool = new TokenPool()
+      expect(() => tokenPool.next()).to.throw('Token pool is exhausted')
```
paulmelnikow: Previously the error was **Priority queue is empty**, which is incorrect.

Issue #1961, opened 2018-08-23
Convert more badges to use static previews


chris48s:
We show example badges on our home page but they don't need to show real data. For performance reasons, it is useful to render a badge using some fake example data instead of making a call to a real API. It is also useful to show the URL format in a parametrised form.

We can do this by switching from defining a `previewUrl` to defining:

* ~~`urlPattern`~~ `pattern` (the URL format in a parameterised form)
* ~~`exampleUrl` (a real badge URL the user can call if they want to)~~ `namedParams` (mappings of pattern placeholders to values for a real, working example)
* `staticExample` (a JSON object or call to the service's `render()` function which generates a fake badge)

You can find examples of service badges which use a `previewUrl` using this search:

https://github.com/search?utf8=%E2%9C%93&q=previewUrl+user%3Abadges+repo%3Ashields+path%3A%2Fservices+language%3AJavaScript&type=Code&ref=advsearch&l=JavaScript&l=

An example pull request will probably look like this:

https://github.com/badges/shields/pull/2230/files

paulmelnikow:
@chris48s I tagged this "good first issue." Would you be up for updating your top post with a few sentences about how new contributors could go about helping with this?

chris48s:
no probs. I've updated the top post

PR #2405, opened 2018-11-27
Tweak [AUR] and add static examples

paulmelnikow:
Ref #1358 #1961

### DIFF
--- a/services/aur/aur.service.js
+++ b/services/aur/aur.service.js
@@ -2,7 +2,7 @@
 
 const Joi = require('joi')
 const { floorCount: floorCountColor } = require('../../lib/color-formatters')
-const { addv: versionText } = require('../../lib/text-formatters')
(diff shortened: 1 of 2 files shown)


services/aur/aur.service.js
```diff
@@ -66,17 +67,15 @@ class AurLicense extends BaseAurService {
   static get route() {
     return {
       base: 'aur/license',
-      format: '(.+)',
-      capture: ['pkg'],
+      pattern: ':packageName',
```
paulmelnikow: `packageName` for consistency with other badges (and because `package` is reserved).

paulmelnikow: Makes sense. This one used `package` in the examples and `pkg` in `capture`. It'll be simpler to use one or the other, and when we're building a form in the frontend, I suppose **package name** seems nicer than **pkg** which is why I went with that.

PR #2409, opened 2018-11-27
Add static examples for [Node]

paulmelnikow:
Ref #1961


services/node/node.service.js
```diff
@@ -20,33 +22,66 @@ module.exports = class NodeVersion extends NPMBase {
...
+          tag: 'latest',
+        }),
+        keywords,
       },
     ]
   }
 
+  static renderStaticExample({ tag, nodeVersionRange }) {
+    // This should match the behavior of `async render()`, which is enforced
+    // with a unit test.
```
chris48s: It took me a while to work out that the reason why we have to do this non-standard thing is because `render()` has to call an async function but we need this to be a synchronous call. Can we make this extra clear in the comment here.

paulmelnikow: I updated the comment. See if that's clearer?

Issue #2378, opened 2018-11-20
Last Discussed on Stack Overflow Badge Proposal


bui1:
:clipboard: **Description**
This badge is for repositories hosted on Github to demonstrate if the API/Service/Library etc. has 
*recent community support*. Often times I'm picking a library that is outdated or
suggests that many contributors are not following anymore. 
A dynamic badge to show the last date of the library has been talked about on a popular Q&A site, Stack Overflow would be helpful!
An example would be as below,
![Last Discussed on Stack Overflow badge](https://user-images.githubusercontent.com/20286010/48801780-7bd67000-ecca-11e8-85bf-db7f6b8b82f3.png)

:link: **Data**
We can get the data via the StackExchange API using the StackOverflow parameters.
This does require an application key that can be thrown in together with the API url
for example, https://api.stackexchange.com/2.2/questions?order=desc&sort=creation&site=stackoverflow&tagged=SOMELIBNAME&key=apikey

Documentation: https://api.stackexchange.com/docs

I am also happy to try implementing this badge and make a PR to Shields!

paulmelnikow:
Nice idea!

Other possibilities: questions in the last year, or questions in the last month. I think those figures would convey a better sense of the amount of activity, because the metric would be less sensitive to a single outlier question. Let's discuss and pin down the design.

Would be great to have your help to implement this! Our stackexchange badge is written using our legacy API and needs to be ported to a BaseJsonService subclass. The service also needs tests; it's one of relatively few at this point which doesn't have them. If you're up for it, what I'd suggest is doing a port of the existing functionality first, and adding the new badge in round 2.

bui1:
@paulmelnikow Those are also good suggestions especially with your point regarding the outliers.
Would something like this design be more suitable?
![image](https://user-images.githubusercontent.com/20286010/48876826-7efe5880-edbd-11e8-99aa-7e02cf888ff3.png)

In terms of figuring out what to do next to make sure i'm understanding correctly
* Port existing Stack Exchange badge into the new BaseJsonService subclass
* Write new tests for said badge above
* Add new proposed badge?

paulmelnikow:
Or a bit more concisely:

![](https://img.shields.io/badge/stackoverflow%20questions-30/month-blue.svg)

Yea! That's a good course of action. The best place to start is the tests – you can write tests for the old badge first, and we can even get them reviewed in a separate PR. It gives you something to test the rewrite against. And then, yea! You can add the new proposed badge in the last step.

[...]

PR #2414, opened 2018-11-27
Add tests for [StackExchange]

bui1:
Per part of issue #2378, tests for StackExchange have been implemented
to help with porting the existing badge with the new BaseJsonService class.
If merged, the next steps would be to port the badge, then add a new
StackOverflow badge part of #2378

### DIFF
--- a/services/stackexchange/stackexchange.tester.js
+++ b/services/stackexchange/stackexchange.tester.js
@@ -10,21 +10,11 @@ module.exports = t
 
 t.create('Invalid t parameter')
   .get('/stackoverflow/t/invalidimage.json')
-  .expectJSONTypes(
-    Joi.object().keys({
-      name: 'stackoverflow',
-      value: 'invalid',
-    })
-  )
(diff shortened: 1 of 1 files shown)


services/stackexchange/stackexchange.tester.js
```diff
@@ -0,0 +1,59 @@
...
+
+module.exports = t
+
+t.create('Invalid t parameter')
+  .get('/stackoverflow/t/invalidimage.json')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'stackoverflow',
+      value: 'invalid',
+    })
```
paulmelnikow: When everything is literal, these can be written a little more simply (as you've done with the networkOff test):

```js
  .expectJSON({ name: 'stackoverflow', value: 'invalid' })
```

bui1: Resolved with newest commit!

paulmelnikow:
Thanks so much!

PR #2418, opened 2018-11-28
Port [StackExchange] from legacy to BaseJsonService

bui1:
- Per part of issue #2378, convert StackExchange service to the newer BaseJson class. 
- Refactor also to seperate StackExchange reputation and questions tag info into new StackExchange sub service files as the legacy class contained both which may be difficult to keep track of in the future as the service continues to grow.
- More tests have been added to make sure other StackExchange sites besides
StackOverflow will still work with the new badge setup.

If merged, next step would be to add a new StackOverflow badge part of #2378.


services/stackexchange/stackexchange-reputation.service.js
```diff
@@ -0,0 +1,82 @@
...
+        503: 'inaccessible',
+      },
+    })
+
+    const numValue = parsedData.items[0].reputation
+
+    return this.constructor.render({
+      label,
+      numValue,
+      color: floorCountColor(1000, 10000, 20000),
```
paulmelnikow: Ha, this is a bug in the old code! The reputation count needs to be passed to `floorCountColor`.

services/stackexchange/stackexchange-reputation.service.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'stackoverflow' }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'StackOverflow',
```
paulmelnikow: It would be better to call these **Stack Exchange reputation** / **Stack Exchange questions**.

services/stackexchange/stackexchange-reputation.tester.js
```diff
@@ -0,0 +1,33 @@
...
+t.create('Reputation for Tex user 22656')
+  .get('/tex/226.json')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'tex reputation',
+      value: isMetric,
+    })
+  )
+
+t.create('No connection')
```
paulmelnikow: Since this behavior is covered by the BaseService tests and not by code in the badge, it's okay to remove it. BaseService gets its own tests.

services/stackexchange/stackexchange-taginfo.service.js
```diff
@@ -0,0 +1,81 @@
...
+      Joi.object({
+        count: Joi.number()
+          .min(0)
+          .required(),
+      })
+    )
+    .required(),
+}).required()
+
+module.exports = class StackExchangeTagInfo extends BaseJsonService {
```
paulmelnikow: StackExchangeQuestions seems like a clearer name for this class + file.

services/stackexchange/stackexchange.tester.js
```diff
@@ -1,49 +0,0 @@
...
-  .get('/stackoverflow/r/22656.json')
-  .networkOff()
-  .expectJSON({ name: 'stackoverflow', value: 'inaccessible' })
-
-t.create('Hit IP Rate Limit')
-  .get('/stackoverflow/r/22656.json')
-  .intercept(nock =>
-    nock('https://api.stackexchange.com/2.2/')
-      .get('/users/22656?site=stackoverflow')
-      .reply(429, [{ error_name: 'throttle_violation' }])
```
paulmelnikow: Is it worth including a custom `errorMessage` for status code 429? Probably **rate limit exceeded** would make it really clear.

services/stackexchange/stackexchange-reputation.service.js
```diff
@@ -0,0 +1,82 @@
+'use strict'
+
+const BaseJsonService = require('../base-json')
+const { metric } = require('../../lib/text-formatters')
+const { floorCount: floorCountColor } = require('../../lib/color-formatters')
+const Joi = require('joi')
+
+const reputationSchema = Joi.object({
+  items: Joi.array()
+    .min(1)
```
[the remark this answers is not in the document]
bui1: For this particular API call, there's always a single value returned unless the query provided by the client doesn't exist in StackExchange and therefore return an empty array.

I guess I can make an error case for if I get an empty array back from StackExchange,
I output a badge with "0" as its value? Then change the schema so that its min(0), max(1)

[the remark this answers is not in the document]
paulmelnikow: Okay. Yea, you can use `.length(1)`, then. I don't think another test case is needed; when the schema is broken we can trust the Joi will do the right thing.

services/stackexchange/stackexchange-reputation.service.js
```diff
@@ -37,46 +37,44 @@ module.exports = class StackExchangeReputation extends BaseJsonService {
...
 
-  static render({ label, numValue, color }) {
+  static render({ stackexchangesite, numValue }) {
+    const label = `${stackexchangesite} reputation`
+
     return {
       label,
       message: metric(numValue),
-      color,
+      color: floorCountColor(numValue),
```
paulmelnikow: To keep the intended color scale, this should be `floorCountColor(numValue, 1000, 10000, 20000)`.

bui1: Makes sense thanks for catching that

paulmelnikow:
I deployed a review app: http://shields-staging-pr-2418.herokuapp.com/

Something's not quite right with the examples:

screen shot 2018-11-28 at 8 16 48 pm

paulmelnikow:
Nice work! 🏆

And please keep it up! There's so much to do in this project and I and the other maintainers are happy to help you find more work that interests you.

paulmelnikow:
This is deployed to s0 and it looks like it's not working quite right:

https://s0.shields-server.com/stackexchange/stackoverflow/r/893113.svg
 

This is the old code:

https://s1.shields-server.com/stackexchange/stackoverflow/r/893113.svg


Do you mind taking a look?

paulmelnikow:
I'm not exactly sure what you mean. However, we should make sure the URL pattern that worked in the past also work going forward.

paulmelnikow:
Thanks! My bad for not noticing it during the review! The tests moved from one file to another and I didn't realize the routes changed.

paulmelnikow:
@bui1 Is this something you could do in the next day? It would be great to get other work shipped and this is a deploy blocker. If not, zero pressure, I could take in on.

PR #2422, opened 2018-11-29
TUTORIAL.md: Service handles requests to external APIs

techtonik:
Correct me if I wrong.

If there are any kind of incoming requests that the service can or should answer, I would like to touch it with my browser right away.


paulmelnikow:
The services _do_ handle incoming requests. When a request comes in that matches the route for a badge, the server invokes the service to handle it. Maybe I can clarify the bullet point in another way.

paulmelnikow:
@techtonik What do you think about my changes?

techtonik:
Seems good after following the tutorial. :) Now it is clear that the code parses parameters from incoming requests and generates the badge.

PR #2425, opened 2018-11-29
TUTORIAL.md: Mention URL mapping upfront and clarify static meaning

techtonik:


techtonik:
Maybe I am making it too obvious...

doc/TUTORIAL.md
```diff
@@ -121,12 +122,12 @@ Description of the code:
 1. We declare strict mode at the start of each file. This prevents certain classes of error such as undeclared variables.
 2. Our service badge class will extend `BaseService` so we need to require it. We declare variables with `const` and `let` in preference to `var`.
 3. Our module must export a class which extends `BaseService`
-4. `route()` declares a route. We declare getters as `static`.
-    * `base` defines the static part of the route.
-    * `pattern` defines the variable part of the route. It can include any
+4. `route()` maps URL paths to handler parameters. `static` getters don't have access to object properties, so they contain everything they need inside.
```
paulmelnikow: I like the intent of this comment on `static`, though I think this is going a bit in the wrong direction.

Instead of just saying it's `static`, it's nice to explain what that means. However we have routes that invoke static helper methods and outside helper functions, so it's not about containing what they need inside, so what's written here seems misleading. I'll try another edit.

paulmelnikow:
Thoughts?

techtonik:
New explanation about why `static` is needed makes it a much better docs. Looks good.

PR #2429, opened 2018-11-29
Add tips for rewriting legacy services

paulmelnikow:
Here are some tips I threw together. Not comprehensive, though perhaps a good start!

If you have friendly edits, please just go ahead and make them. If you're not a collaborator on this repo you'll have to open a new PR, but I'm okay with that.

### DIFF
--- a/doc/rewriting-services.md
+++ b/doc/rewriting-services.md
@@ -1,0 +1,204 @@
+# Tips for rewriting legacy services
+
+## First, write some tests
+
+If service tests don’t exist for the legacy service, stop and write them first.
+It’s recommended to PR these separately. If there’s some test coverage, it’s
+probably fine to move right ahead and add more in the process. Make sure the
+tests are passing, though.
+
+## Organization
+
+1. When there’s a single legacy service that handles lots of different things
+(e.g. version, license, and downloads), it should be split into three separate
+service classes and placed in three separate files.
+`myservice-version.service.js`, `myservice-license.service.js`,
+`myservice-downloads.service.js`.
+
+2. When a service has different variants of basically the same thing, it’s okay
+to put them in the same service. For example, daily/weekly/monthly/total
+downloads can go in one badge, and star rating vs point rating vs rating count
+can go in one badge, and same with various kinds of detail about a pull request.
+The hard limit (as of now anyway) is *one category per service*.
+
+3. If the tests haven’t been split up, split them up too and make sure they
+still pass.
+
+## Get the route working
+
+1. Disable the legacy service by adding a `return` at the top of
+`registerLegacyRouteHandler()`.
+
+2. Set up the route for one of the badges. First determine if you can express
+the route using a `pattern`. A `pattern` is the simplest way to declare the
+route, also the most readable, and will be useful for displaying a badge builder
+with fields in the front end and generating badge URLs programmatically. One
+limitation to keep in mind is that, at present, the trailing parameter of a
+pattern can't be optional. If the last part of a route is optional, like a
+branch, you will need to use a `format` regex.
+
+3. When creating the initial route, you can stub out the service. A minimal
+service extends BaseJsonService (or BaseService, or one of the others), and
+defines `route()` and `handle()`. `defaultBadgeData` is optional but suggested:
+
+```js
+const BaseJsonService = require('../base-json')
+
+class ExampleDownloads extends BaseJsonService {
+  static get route() {
+    return {
+       base: 'example/d',
+       pattern: ':param1/:param2’:',
+     }
+   }
+
+   static defaultBadgeData() {
+     return { label: 'downloads' } // or whatever
+   }
+
+   async handle({ param1, param2 }) {
+     return { message: 'hello' }
+   }
+}
+```
+
+4. We don’t have really good tools for debugging matches, so the best you can do
+is run a subset of your tests. To run a single service test, add `.only()`
+somewhere in the chain, and run `npm run test:services:trace — —only=example`.
+
+```js
+t.create('build status')
+  .get('/pip.json')
+  .only() // Prevent this ServiceTester from running its other tests.
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'docs',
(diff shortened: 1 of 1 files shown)


doc/rewriting-services.md
```diff
@@ -0,0 +1,204 @@
...
+(e.g. version, license, and downloads), it should be split into three separate
+service classes and placed in three separate files.
+`myservice-version.service.js`, `myservice-license.service.js`,
+`myservice-downloads.service.js`.
+
+2. When a service has different variants of basically the same thing, it’s okay
+to put them in the same service. For example, daily/weekly/monthly/total
+downloads can go in one badge, and star rating vs point rating vs rating count
+can go in one badge, and same with various kinds of detail about a pull request.
+The hard limit (as of now anyway) is *one category per service*.
```
chris48s: Is there a clearer word we can use than 'service' here? (maybe 'badge or 'class'?) We refer to `myservice-version.service.js` etc in the previous paragraph. Can we make it clearer that `myservice-version` (and so on) must belong to one category but `myservice` can contain badges belonging to multiple categories. In general we tend to use the terms 'badge' and 'service' a bit ambiguously, but it is important to be clear in materials aimed at new contributors.

doc/rewriting-services.md
```diff
@@ -0,0 +1,204 @@
...
+1. Disable the legacy service by adding a `return` at the top of
+`registerLegacyRouteHandler()`.
+
+2. Set up the route for one of the badges. First determine if you can express
+the route using a `pattern`. A `pattern` is the simplest way to declare the
+route, also the most readable, and will be useful for displaying a badge builder
+with fields in the front end and generating badge URLs programmatically. One
+limitation to keep in mind is that, at present, the trailing parameter of a
+pattern can't be optional. If the last part of a route is optional, like a
+branch, you will need to use a `format` regex.
```
chris48s: It would be nice to add a bit more explanation of the pattern/format thing here. Perhaps the minimum is just an example. e.g:

> A `pattern` (e.g: `pattern: ':param1/:param2',`) is the simplest way...

> ...you will need to use a format regex. (e.g: `format: '([^/]+/[^/]+)'` )

but maybe this whole topic of defining routes needs its own more detailed "chapter" of docs we can just link to here.

doc/rewriting-services.md
```diff
@@ -0,0 +1,204 @@
...
+defines `route()` and `handle()`. `defaultBadgeData` is optional but suggested:
+
+```js
+const BaseJsonService = require('../base-json')
+
+class ExampleDownloads extends BaseJsonService {
+  static get route() {
+    return {
+       base: 'example/d',
+       pattern: ':param1/:param2’:',
```
chris48s:
```suggestion
       pattern: ':param1/:param2',
```

doc/rewriting-services.md
```diff
@@ -0,0 +1,204 @@
...
+
+6. If the legacy service had a base URL and you've changed it, you’ll need to
+update the tests _and_ the examples. Take care to do that.
+
+## Implement `render()` and `handle()`
+
+1. Once the route is working, fill out `render()` and `handle()`. If there’s a
+single service, you can implement fetch as a method or a function at the top of
+the file. If there are more than one service which share fetching code, you can
+put the fetch function in `example-common.js` like this one for github:
```
chris48s: Lets try and keep the terminology consistent. Earlier we were using `myservice-version.service.js` but now we're using `example-common.js`. Lets use `example` or `myservice` as a prefix everywhere. Given the comment about overloading the term 'service', maybe using `example` is better?

doc/rewriting-services.md
```diff
@@ -0,0 +1,204 @@
...
+  releases: Joi.object()
+    .pattern(
+      Joi.string(),
+      Joi.array()
+        .items(
+          Joi.object({
+            packagetype: Joi.string().required(),
+          })
+        )
+        .required()
```
chris48s: This schema object is incomplete. Also the majority of this code snippet is the schema definition but it isn't really the main thing we're trying to demonstrate. Can we pick an example that uses a less verbose schema so the example is clearer.

doc/rewriting-services.md
```diff
@@ -0,0 +1,204 @@
...
+1. Convert all the examples to `pattern`, `namedParams`, and `staticExample`. In some cases you can use the `pattern` inherited from `route`, though in other cases you may need to specify a pattern in the example. For example, when showing download badges for several periods, you may want to render the example with an explicit `dt` instead of `:which`. You will also need to specify a pattern for badges that use a `format` regex in the route.
+
+2. Open the frontend and check that the static preview badges look good.
+Remember, none of them are live.
+
+3. Open up the prepared example URLs in their own tabs, and make sure they work correctly.
+
+## Validation
+
+Add the schemas toward the end.
```
chris48s: If we're going to suggest adding the schema at the end, we should clarify that you can use `Joi.any()` to start with. We usually need to pass _some_ schema at an earlier stage than this.

Also from the perspective of refactoring, we should explain something about converting manual validation rules to schema. A typical legacy service will have some `if (myvar !== undefined)...` or `const somevar = myvar || 'default'` etc that we want to convert to schema validation.

chris48s:
There is one more additional topic here that we should probably cover here which is (removal of) error handling. A typical legacy service is going to have some stuff like `checkErrorResponse()`, `badgeData.text[1] = 'invalid'`. Mostly we want to replace this with custom errors in `fetch()` or schema validation etc. Can we add a few sentences about this somewhere?

doc/rewriting-services.md
paulmelnikow: Yea, one thing that has occurred to me while writing this is that once the rewrite is done, it could serve as the basis of a developer guide to services – something that goes into more depth than the tutorial, and also talks about many of the various edge cases and contingencies.

doc/rewriting-services.md
paulmelnikow: Is there another good one to show? I dug around a bit before settling on this one. I'll snip the schema for now, though feel free to replace it with a better example (or suggest one, and I'll update).

paulmelnikow:
I've addressed the rest of your comments and this is ready for another round of review!

PR #2432, opened 2018-11-30
Add [StackExchange] monthly questions badge service (#2378)

bui1:
- Badge shows number of questions for a given library for the previous month
(not current month because its in progress)
- Service works for not just the StackOverflow site but for other StackExchange sites also

Example: 
screen shot 2018-11-29 at 10 09 45 pm

If merged, closes issue #2378


services/stackexchange/stackexchange-monthlyquestions.service.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  static render({ stackexchangesite, numValue }) {
+    const label = `${stackexchangesite} questions`
+
+    return {
+      label,
+      message: `${metric(numValue)}/month`,
+      color: floorCountColor(numValue, 1000, 10000, 20000),
```
paulmelnikow: For consistency with the "total questions" badge, could you include the query in the label?

It seems like this and StackExchangeQuestions could then share a `renderQuestionsBadge` helper function. If the labels are made the same, the only difference is the suffix, which could be another parameter. Then you could call it from both services like this:

```js
static render(props) {
  return renderDownloadBadge({
    interval: 'month',
    ...props
  })
}
```

services/stackexchange/stackexchange-monthlyquestions.service.js
```diff
@@ -0,0 +1,88 @@
...
+    const parsedData = await this._requestJson({
+      schema: tagSchema,
+      options: {
+        gzip: true,
+        qs: {
+          site: stackexchangesite,
+          fromdate: prevMonthStart,
+          todate: prevMonthEnd,
+          filter: 'total',
+          tagged: query,
```
paulmelnikow: 👍

services/stackexchange/stackexchange-monthlyquestions.tester.js
```diff
@@ -0,0 +1,25 @@
+'use strict'
+
+const Joi = require('joi')
+const withRegex = re => Joi.string().regex(re)
+const isMonthMetric = withRegex(/^[1-9][0-9]*[kMGTPEZY]?\/(month)?$/)
```
paulmelnikow: To avoid the need for another regex, you could use `isMetricOverTimePeriod` even though it's slightly more lenient.

services/stackexchange/stackexchange-monthlyquestions.tester.js
```diff
@@ -0,0 +1,25 @@
...
+      value: isMonthMetric,
+    })
+  )
+
+t.create('Monthly Questions for Tex Spacing')
+  .get('/tex/monthlyquestions/spacing.json')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'tex questions',
+      value: isMonthMetric,
```
paulmelnikow: 👍

services/stackexchange/stackexchange-monthlyquestions.service.js
```diff
@@ -0,0 +1,88 @@
...
+        }),
+        keywords: ['stackexchange', 'stackoverflow'],
+      },
+    ]
+  }
+
+  static get route() {
+    return {
+      base: 'stackexchange',
+      pattern: ':stackexchangesite/monthlyquestions/:query',
```
paulmelnikow: I feel like this pattern and the one for total questions should be brought into alignment. However total questions uses just `t`. Probably that was for **tag**, or maybe **total**, but I doubt it.

I guess I'd suggest we make this one `stackexchange/stackexchangesite/qm/:query`, where `qm` means **monthly questions**. Then we can think of `t` as standing for **total questions**.

PR #2438, opened 2018-12-01
show dynamic badge examples with a base

chris48s:
this makes them a bit nicer to copy & paste

### DIFF
--- a/frontend/components/usage.js
+++ b/frontend/components/usage.js
@@ -138,6 +138,7 @@ export default class Usage extends React.PureComponent {
 
         <p>
           <code>
+            {baseUrl}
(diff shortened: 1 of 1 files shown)


paulmelnikow:
I guess this isn't showing up in the review app because there is no `baseUrl` configured.

The idea of the badge maker is that it'll give you a full URL you can copy and paste. These paths feel like examples which give you the general idea. Even with the prefix, they are not copy-and-paste friendly.

I prefer this without the prefix, though don't want to block this if you feel strongly about it.

paulmelnikow:
(Not that the badge maker is as friendly as it could be.)

paulmelnikow:
Yea, I hear you. I'm fine with that. Though we should definitely make the builder better!

I code on a small laptop, and that region of the page looks especially cluttered in a small browser window.

Issue #2427, opened 2018-11-29
namedParams are not interpolated in examples


techtonik:
Are you experiencing an issue with...
 - [x] [shields.io](https://shields.io/#/)

:beetle: **Description**
Go https://shields.io/ enter `gem` watch this.

![image](https://user-images.githubusercontent.com/515889/49246186-5281b680-f425-11e8-9203-857205c52c50.png)

Click and watch this.

![image](https://user-images.githubusercontent.com/515889/49246211-60cfd280-f425-11e8-8979-22e349dcc9aa.png)

`formatador` string that is passed as `namedParams` here - https://github.com/badges/shields/blob/97139648129eaed8293307a0d175160f98786aba/services/gem/gem-version.service.js#L51 - is only interpolated in Example field, but not in Image field, making it broken.

paulmelnikow:
Not lovely I agree, though that's the expected behavior. The idea is to show the developer what part they need to fill in. The CircleCI badge long appeared this way with `USER` and `REPO` in that field, and it's now the preferred standard for all badges.

A minor tweak, to make this a little nicer, could be to hide the live preview until all the colons have been filled in.

The real fix is #701, and there is a lot of progress. Examples will need to be switched over to patterns to work this way, but that's underway in #1961. And there's a PR to generate a new file format, with enough information for the frontend, in #2397. (That's the blocker right now.)

paulmelnikow:
> A minor tweak, to make this a little nicer, could be to hide the live preview until all the colons have been filled in.

This has been implemented in #2441.

paulmelnikow:
Let's track this in #2496.

Issue #1899, opened 2018-08-11
Frontend redesign


mbrandau:
As mentioned in #1070, this will cover a complete redesign of the frontend.
The current frontend is rather useful than good looking. The goal is to make it both easy to use and giving it a modern look on both desktop and mobile devices.

**TODOs:**

- Designs
  - [ ] style form elements (*input*, *select*, *button*)
  - [ ] create visual designs for mobile and desktop viewports
- Adopt styled-components
  - [ ] Container
  - [ ] form elements
- UX errors
  - [ ] categories links aren't displayed on example pages
  - [ ] emphasize search box
  - [ ] "donate" link is nearly invisible
  - [ ] consistent naming of the left side of the badge (*label* -> *subject*)

I also thought about adding an about section to the index page with some information about the project.

mbrandau:
This would be a perfect opportunity to address #843 too!

chris48s:
Big :+1: on this and thanks for picking it up again. As I've noted elsewhere, I'd love to see us adopt a more mobile-friendly style.

TODO list looks great but I think the best way to work on this would be to split this into more than 1 PR and work on a subset of functionality as a first step. I'd prioritise the the layout changes, but leave out stuff like an 'about' section, #843, maybe even initially the menu. That will make it easier to review and make progress.

mbrandau:
@paulmelnikow I think it would be better if we maintain a list of design ideas in this issue. I don't know too much about styled-components, but I'm ready to take a quick dive into that topic.

@chris48s I would definitely not do it in just one PR. I agree with you on leaving out the 'about' section, but the missing menu is a real UX problem. #843 doesn't seem too hard to implement, especially with React, so I would like to give this a shot too.

rozaxe:
I would be happy to help you integrate the tool from #843 into the new interface if I can get some guidance about React because I never used it.

paulmelnikow:
Cool! Shall we continue that discussion at #843?

paulmelnikow:
Cool, I'll work on writing up my thoughts on the UX.

@mbrandau It sounds like you're down to work on both design and implementation, which is great! Would you like to work with @rozaxe on #843? I'd suggest starting the implementation work by breaking out the "Your Badge" section – with the static + dynamic badge builders – onto their own page. Does that jive with your thoughts about the design?

mbrandau:
@paulmelnikow Splitting the page up sounds good! I'm gonna get the basics set up so @rozaxe can work with it.

[...]

Issue #1272, opened 2017-11-12
Add Flow or TypeScript typing to repo


Daniel15:
It would be nice to have either Flow or TypeScript in the codebase, to better understand return types for some of the more complex functions.

paulmelnikow:
I don't have much experience with either, though I'm happy to learn. I'll find an area in this project or another, to do a proof of concept. Any suggestions? The frontend perhaps?

It's really important that this project have a wide base of developers. There are lots and lots of package repositories and CI services for all sorts of different developer communities, and it's better to empower and enable all those developers to build out the features they use. This means lots of non-JS devs are hacking on this code.

So, I think we should decide whether to do this based on _how likely it makes people to contribute_. If we think we'll get more contributors by doing this, let's do it. If we think we'll get fewer, let's look at options involving ES6/ES7/ES8.

Another possibility is that we use a type system in the frontend and not in the server. React devs –
who are our target for frontend contributions – may be more likely knowledgeable of one of these type systems than a non-JS dev who is adding a service for their language.

Daniel15:
I was thinking it'd be good for the server, because there's several data structures in use (for example, the `badgeData` passed to `sendBadge`). Maybe a good starting point is to add types to a few of the smaller files in the `lib` directory, to get a taste of it. 

With Flow, you opt-in per-file by adding an `@flow` annotation to a docblock at the top of the file, so you don't need to convert the entire app at once ad can instead just gradually add types :)

paulmelnikow:
Cool, I'll take that on once #963 is merged.

styfle:
I vote for TypeScript! I could give some pointers or review a PR.

See my repo [styfle/react-server-example-tsx](https://github.com/styfle/react-server-example-tsx) for an example implementation.

Daniel15:
Yeah I think either one is fine. I'm more familiar with Flow (as it's what we use at Facebook) and it tends to handle React a bit better, but they're both good 😃

paulmelnikow:
I'm working on a PR for [Danger](http://danger.systems/js/), which uses TypeScript. So I'm getting some exposure!

paulmelnikow:
After working on that PR for Danger, my overall impression is that the types make working with code faster, and avoid common problems. It makes many things self-documenting, and forces you to think about how to model your data, which is awesome. It took me longer to make some changes, probably that is in part because I'm still learning the type system. The tooling is different and was initially unfamiliar, and that took me a good bit of time to adapt.

[...]

PR #2441, opened 2018-12-03
Refactor frontend main page and badge-example code

paulmelnikow:
- The goal of this PR is:
    - Consume the new service-definition format
    - Make the frontend more readable.
- Behavior changes:
    - I changed the **Image** field in the markup modal to show only the path.
    - I added another click-to-select field below that shows the complete URL.
    - This made it easier to suppress the live badge preview while it contains placeholders like `:user` or `:gem`, a minor tweak discussed at https://github.com/badges/shields/issues/2427#issuecomment-442972100.
- I did not deliberately address performance, though I ripped out a bunch of anonymous functions and avoided re-filtering all the examples by category on every render, which I expect will not hurt. I haven't really tested this on a mobile connection and it'd be worth doing that.
- It would be great to have some tests of the components, though getting started with that seemed like a big project and I did not want to make this any larger than it already is.

It's a medium-sized refactor:

1. Replace `BadgeExamples`, `Category` and `Badge` component with a completely rewritten `BadgeExamples` component which renders a table of badges, and `CategoryHeading` and `CategoryHeadings` components.
2. Refactor `ExamplesPage` and `SearchResults` components into a new `Main` component.
3. Rewrite the data flow for `MarkupModal`. Rather than rely on unmounting and remounting the component to copy the badge URL into state, employ the `getDerivedStateFromProps` lifecycle method.
4. Remove `prepareExamples` and `all-badge-examples`.
5. Rewrite the `$suggest` schema to harmonize with the service definition format. It's not backward-compatible which means at deploy time there probably will be 10–20 minutes of downtime on that feature, between the first server deploy and the final gh-pages deploy.  🤷‍♂️ (We could leave the old version in place if it seems worth it.)
6. Added two new functions in `make-badge-url` with tests. I removed _most_ of the uses of the old functions, but there are some in parts of the frontend I didn't touch like the static and dynamic badge generators, and again I didn't want to make this any larger than it already is.
7. Fix a couple bugs in the service-definition export.


paulmelnikow:
This pull request **introduces 1 alert** when merging debfa4667d846c26b47653952aee7ffba739fd53 into 0a6afb447877e9b9d973146be65cb1b23fe7e588 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-1fe6121f2157fefc47875813be21ae99e5ddab64)

**new alerts:**

* 1 for Unused or undefined state property

---

paulmelnikow:
This pull request **introduces 2 alerts** when merging b475529238a804ea04845ecad45507c69226cd0e into 0a6afb447877e9b9d973146be65cb1b23fe7e588 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-eb3a2d753465ff7ac12dd2b9b0d72b8c92ae0552)

**new alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused or undefined state property

---

paulmelnikow:
AFAICT the LGTM warning is a false positive.

screen shot 2018-12-02 at 10 39 03 pm

screen shot 2018-12-02 at 10 40 22 pm

paulmelnikow:
This pull request **introduces 2 alerts** when merging c45fc6e7da8bc5d42d74e897180736a7b451d879 into ec07f23942373930a855ad13c8e1b5f6e1b02052 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-2edc63459b3df11193c277182a248e7c458f3834)

**new alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused or undefined state property

---

frontend/components/main.js
```diff
@@ -0,0 +1,174 @@
...
+      isSearchInProgress: false,
+      isQueryTooShort,
+      searchResults,
+    })
+  }
+
+  searchQueryChanged(query) {
+    /*
+    Add a small delay before showing search results
+    so that we wait until the user has stipped typing
```
chris48s:
```suggestion
    so that we wait until the user has stopped typing
```

this was probably originally my typo, but I just spotted it :)

frontend/lib/service-definitions/service-definition-set-helper.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  notDeprecated() {
+    return ServiceDefinitionSetHelper.create(
+      this.definitionData.filter(({ isDeprecated }) => !isDeprecated)
+    )
+  }
+
+  asNative() {
+    return this.definitionData
```
chris48s: I find this method name a bit confusing

frontend/components/markup-modal.js
```diff
@@ -2,86 +2,163 @@ import React from 'react'
...
   static propTypes = {
-    example: PropTypes.shape({
-      title: PropTypes.string.isRequired,
-      exampleUrl: PropTypes.string,
-      previewUrl: PropTypes.string,
-      urlPattern: PropTypes.string,
-      documentation: PropTypes.string,
-      link: PropTypes.string,
-    }),
+    example: PropTypes.object,
```
chris48s: Are we losing some useful validation here?

frontend/components/markup-modal.js
```diff
@@ -160,29 +237,8 @@ export default class MarkupModal extends React.Component {
...
-          </p>
-          <p>
-            AsciiDoc&nbsp;
-            <ClickToSelect>
-              <input className="code clickable" readOnly value={asciiDoc} />
-            </ClickToSelect>
-          </p>
-          {this.renderDocumentation()}
+          {this.isOpen && this.renderMarkup()}
+          {this.isOpen && this.renderDocumentation()}
```
chris48s: It looks like you've added code to render the documentation but if I look at some examples on https://shields-staging-pr-2441.herokuapp.com/ which include a documentation key, it isn't rendering

frontend/lib/service-definitions/service-definition-set-helper.js
paulmelnikow: Would `toArray()` be better?

frontend/components/markup-modal.js
paulmelnikow: Part of what makes prop types useful is that they gives a single point of reference on the props schema. Since the service definition data is passed straight through to the components – this is one of the `example` keys from it – the service definition schema can be used as a reference.

It seems like a lot of work to maintain this schema in two places, in two incompatible formats.

Prop types also so not prevent components from incorrectly referencing props, or referencing props that don't exist. Related to that… in #1272 @Daniel15 suggested adding typing to the repo. This changeset took many hours to produce, and most of the reason is that the types were changing drastically, from the top all the way down. Type checking seems pretty essential for frontends that can be refactored with ease. I suspect adding TypeScript might make the frontend easier for the average potential frontend contributor, despite the added learning curve. 

I continue to think this is not as pressing an issue for the server code and definitely think it will deter some contributors. JavaScript, even with ES6, is a least-common-denominator kind of language. It's pretty easy for anyone to read, and even easier to write.

I'm wanting to make significant improvements to the frontend, and I'm thinking we may need to recruit some frontend contributors in a targeted way in order to do that. It's on my mind to create a front end dev experience that's as good as the back end dev experience. Of course, we also need to spend some time on the design (#1899) so that can happen in an organized way.

But that's all an aside. TypeScript is hardly pressing.

tldr; I think the validation is less important than the documentation. Maybe adding a link to the schema documentation would be sufficient.

frontend/components/markup-modal.js
paulmelnikow: Fixed in 1a8feaf. Ironically enough, had I specced out the input as you suggested in https://github.com/badges/shields/pull/2441#discussion_r240007528, that would have surfaced the problem.

paulmelnikow:
This pull request **introduces 2 alerts** when merging 44095635f667cf6c23b5114085132e900de69a3b into 8a8311d931b9949ae9fdff72c96abb60019825da - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-0c76a9d20b1d3264fab6bfbf55b2c5d177025552)

**new alerts:**

* 1 for Useless assignment to local variable
* 1 for Unused or undefined state property

---

frontend/lib/service-definitions/service-definition-set-helper.js
chris48s: yeah, that seems more descriptive

frontend/components/markup-modal.js
chris48s: There's quite a lot of stuff there, but I think the key point is

> Prop types also so not prevent components from incorrectly referencing props, or referencing props that don't exist

:+1:

paulmelnikow:
This pull request **introduces 1 alert** when merging 74741f70621a53614b5db927869c064ada8d7ff0 into 8a8311d931b9949ae9fdff72c96abb60019825da - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-5c952a564b8057592d3a6196f9138bfeb48e1842)

**new alerts:**

* 1 for Unused or undefined state property

---

frontend/components/markup-modal.js
Daniel15: > Part of what makes prop types useful is that they gives a single point of reference on the props schema. Since the service definition data is passed straight through to the components – this is one of the example keys from it – the service definition schema can be used as a reference.

You can reuse prop types; you don't have to redefine them everywhere. You could make a module that exports the prop types
```
module.exports = {
  Example: PropTypes.shape({
    title: PropTypes.string.isRequired,
    exampleUrl: PropTypes.string,
    previewUrl: PropTypes.string,
    urlPattern: PropTypes.string,
    documentation: PropTypes.string,
    link: PropTypes.string,
  }),
};
```

Then you could do `example: MyShapes.Example` or whatever here.

That's pretty similar to how you'd do this with Flow or TypeScript - Define the type in some reusable place, and reuse it everywhere.

paulmelnikow: Hi! Right, thanks for that suggestion. I have done that in some places. The issue here is different: [the schema is defined using Joi](https://github.com/badges/shields/blob/master/services/service-definitions.js).

That's the validation framework we're using for the JSON responses we get back, which we adopted for runtime use without a ton of consideration (beyond it being integrated into our API testing framework).

That said, last I looked, I did not find a library that could do runtime validation of plain JS objects using TypeScript types.

There do seem to be some attempts to use Joi with React. I didn't dig too far into [the most promising example](https://www.npmjs.com/package/react-joi-validation), though I also wouldn't want to ship Joi to the frontend.

I wouldn't mind using it in dev, though it seems like it would take some effort to configure webpack to accomplish that.

PR #2443, opened 2018-12-03
Examples: Drop deprecated aliases query, urlPattern

paulmelnikow:


chris48s:
Do the references to `urlPattern` and `query` in `validate-example`:

https://github.com/badges/shields/blob/e8c411b6f97b9b73ec2149b2bf138a8bcd3a54c4/services/validate-example.js#L3-L23

also want to be edited here?

paulmelnikow:
After this change, using the old keys will result in an error regardless. That code's up for deletion in #2441 so I don't think it's worth maintaining it.

PR #2446, opened 2018-12-03
Link together some of the documentation

paulmelnikow:


doc/rewriting-services.md
```diff
@@ -1,5 +1,50 @@
...
+adapter called [LegacyService][] which provides a place to put the
+`camp.route()` invocation. The wrapper extends from [BaseService][], so it
+supports badge examples via `category`, `examples`, and `route`. Setting `route`
+also enables `create-service-tester.js` to infer a service's base path, reducing
+boilerplate for [creating the tester][creating a tester].
+
+Legacy services look like:
+
+```js
+camp.route(
```
chris48s: should we show this wrapped in a `module.exports = class ExampleService extends LegacyService {}` as that's what almost every example "in the wild" now looks like

doc/rewriting-services.md
```diff
@@ -1,5 +1,50 @@
...
+    badgeData.colorB = '#008bb8'
+    sendBadge(format, badgeData)
+  })
+)
+```
+
+References:
+
+- [Old tutorial on legacy services][old tutorial]
+- Current documentation
```
chris48s: This is a really minor point and I might be over-thinking it slightly, but shall we put the link to the old docs further down the list? From the perspective of a newcomer, the first link in this document is to a tutorial that reflects bad practice. I can see why its a useful link, but it seems like a good idea to make the docs that reflect good practice the top thing in the list.

PR #2447, opened 2018-12-04
Modernize some more examples

paulmelnikow:
Ref #1961 

No base URLs or server code was changed here, so we don't need to run service tests.


services/steam/steam-workshop.service.js
```diff
@@ -179,11 +179,9 @@ class SteamCollectionSize extends BaseSteamAPI {
     return [
       {
         title: 'Steam Collection Files',
-        exampleUrl: '180077636',
-        pattern: ':id',
```
RedSparr0w: I'm assuming we don't require a pattern if there is only 1 parameter,
Do you think it could still be worth keeping it though?

paulmelnikow: Ah, it's that by default, an example inherits its pattern from the route. Since I've brought the names into alignment, it doesn't need to be specified again.

RedSparr0w: Ah yeah just saw that :laughing:

Thanks :+1:

PR #2456, opened 2018-12-04
couples of fixes to docs on porting legacy services

chris48s:
- use the right issue no and link to it
- fix the trace command example

### DIFF
--- a/doc/rewriting-services.md
+++ b/doc/rewriting-services.md
@@ -3,7 +3,8 @@
 ## Background
 
 The services are in the process of being rewritten to use our new service
-framework (#1584). Meanwhile, the legacy services extend from an abstract
+framework ([#1358](https://github.com/badges/shields/issues/1358)).
+Meanwhile, the legacy services extend from an abstract
 adapter called [LegacyService][] which provides a place to put the
 `camp.route()` invocation. The wrapper extends from [BaseService][], so it
 supports badge examples via `category`, `examples`, and `route`. Setting `route`
@@ -120,7 +121,7 @@ class ExampleDownloads extends BaseJsonService {
 
 4. We don’t have really good tools for debugging matches, so the best you can do
 is run a subset of your tests. To run a single service test, add `.only()`
-somewhere in the chain, and run `npm run test:services:trace — —only=example`.
+somewhere in the chain, and run `npm run test:services:trace -- --only=example`.
 
 ```js
 t.create('build status')


paulmelnikow:
For what it's worth, I'm finding the PR overhead for documentation improvements discourages me from writing and updating them. I wonder if we should consider an alternative, like keeping a branch running with doc changes to which all maintainers can push edits, and running it through PR once a week.

chris48s:
Yeah our branch protection rules should work for us, not against us.

With code, the checks/review requirement is there for good reasons.
With docs, I can see that its less important to review/test docs: We're unlikely to _break_ anything as a result of a mistake in the docs in the same way that a bug in code can do damage. On the other hand, if a docs change spends a couple of days in review, its also generally not an urgent issue either.

Batching up docs into bigger PRs is also less problematic. Reviewing say three-hundred lines of English definitely feels like less cognitive overhead than three-hundred lines of javascript.

paulmelnikow:
> On the other hand, if a docs change spends a couple of days in review, its also generally not an urgent issue either.

For me it's more about reducing friction in something that we'd like to encourage. Waiting or PR comments + responding to PR comments + waiting for a new :+1: + merging means that even small improvements take a lot of active time. It means I'm more likely to put off writing docs. 😁

As you've correctly observed, there is basically infinite possible work in this project, and I think reducing the overhead writing might be a way to steer a bit more of the effort in that direction.

RedSparr0w:
Do you think it would be worth opening the wiki to allow more people to make quick changes? It may also make it easier for people to find the information they are looking for.

paulmelnikow:
In GitHub it's a bit strange, because as far as I know you can't propose edits. Either you have write access or you don't. So we could make them only editable by maintainers, but then no one else can propose changes. Or we could make them editable by anyone, but then we'd have to keep an eye on them.

Have you used them on any other projects?

RedSparr0w:
I've used them awhile ago.
From memory, you can't propose edits, but you can review changes and reverse them easy enough, most of the time everyone is pretty sensible about it. And I think there is an option for contributors only, which would limit who could edit it without being too strict.
Will look into it a bit more and see what other options it has.

chris48s:
Another option we could consider is having a separate docs rep under the @badges namespace with slightly more liberal branch protection rules. We'd have to ensure we link to them in plenty of places to ensure people can find them. If possible, its nicer to have the docs in the same repo as the code for discoverability its worth considering.

Thinking ahead, if we move to a situation where the internal API in `/lib` has proper API documentation, it opens up having a compiled GH pages site or something as those API docs grow.

paulmelnikow:
A good reason for docs to live alongside code is that it's much more likely they'll be kept up to date. Also agreed we should generate compiled docs for the base services and the functions they call.

They may end in `services/`, which is becoming the place where most of the service helpers live.

For higher level doc, like guides and tutorials, that's not so critical… especially as these APIs start to simmer down. A separate repo could work nicely.

Another possibility may be to use [code owners](https://help.github.com/articles/about-codeowners/) to require approval for everything else, while allowing doc changes to be merged without code owner approval.

PR #2458, opened 2018-12-04
Add Leanpub Book Summary Badges [LeanpubBookSummary]

### LINKED ISSUE #2270: LeanPub Badges
:clipboard: **Description**
Badges for [LeanPub](https://leanpub.com) books
- [ ] "pages | 226" (= `page_count` of Book Summary Information `json` response)
- [ ] "readers | 2691" (= `total_copies_sold` of Book Summary Information `json` response)
- [ ] "revenue | $ 985.05" (= `total_revenue` of Book Summary Information `json` response)

* All above badges should by default link to the `url` of their Book Summary Information `json`

:link: **Data**
* API Documentation: https://leanpub.com/help/api
* Book Summary Information: `https://leanpub.com/<book>.json` for example: https://leanpub.com/juice-shop.json (no API Key required)

calebcartwright:
Adds new service to provide badges for [Leanpub](https://leanpub.com/) book metrics which closes #2270 

Comments:
* I wasn't sure what category this would fall under, so I chose `other`
* ~~One of the badges (`revenue`) will only be available on self-hosted Shields instances. I added a note to the docs for that badge to indicate this, but not sure if there are other conventions/standards followed for self-hosted only badges~~
* ~~The Leanpub API docs all reference providing the API Key (which would only be done for self-hosted Shields instances) via a query param. I'd prefer that be done via auth headers, but figured I'd just go with what their docs say, especially since I don't personally use the service~~


chris48s:
Thanks for submitting this.

I'm not sure what to do about the revenue badge. I don't think its particularly useful to add/document a badge on shields.io that nobody can use. Does anyone know of another example where we've got a badge in core that only works on a self-hosted instance?

paulmelnikow:
> I'm not sure what to do about the revenue badge. I don't think its particularly useful to add/document a badge on shields.io that nobody can use. Does anyone know of another example where we've got a badge in core that only works on a self-hosted instance?

I'm not aware of any.

In theory I wouldn't mind maintaining a badge that targets self-hosing users, though I'm not sure if there is anyone who wants the revenue badge enough to self-host.

Given that, maybe we could leave off the revenue badge for now?

calebcartwright:
I think that's fair, and also aligns with what the original requestor [said](https://github.com/badges/shields/issues/2270#issuecomment-443950803) when we found it out would only be available for self-hosted.

I'll remove the revenue badge. Any other changes needed?

services/leanpub/leanpub-book-summary.service.js
```diff
@@ -0,0 +1,88 @@
...
+    })
+
+    let label, value
+
+    if (metric === 'pages') {
+      label = 'pages'
+      value = json.page_count_published
+    } else {
+      label = 'sold'
+      value = json.total_copies_sold
```
chris48s: Instead of doing

```js
if (metric === 'pages') {
  label = 'pages'
} else if (metric === 'sold') {
  label = 'sold'
}
```

you could just use metric as the label directly :)

```js
return this.constructor.render({ label: metric, message: value })
```

services/leanpub/leanpub-book-summary.service.js
```diff
@@ -0,0 +1,88 @@
...
+        staticExample: this.render({ label: 'sold', message: 2691 }),
+        keywords,
+      },
+    ]
+  }
+
+  static get route() {
+    return {
+      base: 'leanpub/book/summary',
+      format: '([^/]+)/(pages|sold)',
```
chris48s: Usually our URL schema is like

`npm/version/:package`
`github/tag/:user/:repo` etc

Seems like it would be more consistent to have

`leanpub/pages/:book` or
`leanpub/book/pages/:book` (if we might have badges for leanpub stuff that isn't books)

Also, you could define this using a `pattern` here instead of a regex and then you don't need to define it in the examples. So this could be


```js
static get route() {
  return {
    base: 'leanpub/book',
    pattern: ':book(pages|sold)/:metric',
  }
}
```

and then you can dispense with the `pattern` key in the examples.

chris48s:
cheers. I've left a couple of other minor comments to follow up on

services/leanpub/leanpub-book-summary.service.js
calebcartwright: Good call. I was originally adding a different label but switched back, making that unnecessary.

services/leanpub/leanpub-book-summary.service.js
calebcartwright: disclaimer: I've never used Leanpub myself :) In looking at their API docs it does seem like there would be the potential (however remote) for other Leanpub badges in the future not related to book metrics. 

I think I understand the suggestion and will update accordingly

calebcartwright: So I made the changes and it is functional, including removing the  `pattern` key from the examples:
```js
{
  title: 'Leanpub Book Page Count',
  namedParams: {
    metric: 'pages',
    book: 'juice-shop',
  },
  staticExample: this.render({ label: 'pages', message: 226 }),
  keywords,
},
```
But this results in an odd looking url showing on the examples page: 

`http://[::]:8080/leanpub/book/:metric(pages%7Csold)/:book.svg`

Is there another key I should add to the examples so that it will pull in the right metric? I.e. I'd assume the urls should look like: 

`http://[::]:8080/leanpub/book/pages/:book.svg` and
`http://[::]:8080/leanpub/book/sold/:book.svg`

chris48s: aah - I hadn't considered that. We probably need to improve that in the front-end, but lets not shave that yak in this PR. We can leave it as a regex for now, so we can define the route as:

```js
    base: 'leanpub/book',
    format: '(pages|sold)/([^/]+)',
    capture: ['metric', 'book']
```

and set a pattern in the examples

calebcartwright:
Also a semi-random question. I noticed in the code coverage results that the `examples` method is the only one that doesn't get hit (the other static methods like `category` and `defaultBadgeData` do get hit) resulting in coverage < 100%. Am I missing something in my tests or is this the expected behavior?

services/leanpub/leanpub-book-summary.service.js
calebcartwright: I already pushed the change to use `pattern` in the route, and I added a modified `pattern` key back to the examples so they'd display correctly:

```js
{
  title: 'Leanpub Book Page Count',
  namedParams: {
    metric: 'pages',
    book: 'juice-shop',
  },
  pattern: 'pages/:book',
  staticExample: this.render({ label: 'pages', message: 226 }),
  keywords,
},
```

Will that work or would you prefer I switch back to the `format/capture` approach in the route?

chris48s: aah yeah - that works too. Even better, in fact :)

paulmelnikow: > I added a modified `pattern` key back to the examples so they'd display correctly:

For what it's worth, that's the approach I've mostly been using.

I'm not sure how best to handle this in the long run. It's possible we could get to a point that we show one pattern per service class in the listing, and present the examples on the popup. The biggest blocker right now is pillarjs/path-to-regexp#173, though there are a few other cases as you've pointed out, where the regex is too complicated.

chris48s:
Good question. I think if you run the full test suite (as opposed to just the service tests) with `npm test` there is a test in there somewhere which builds the full front-end data structure. That has the effect of touching all the service examples which will cause it to fail if something in there throws an exception, but we don't really _test_ the examples beyond that.

Issue #2259, opened 2018-11-03
dependency version badges for applications, from package.json


guylepage3:
**Description**
I would love to be able to add `dependency` version badges from my `package.json` and have those badges change to let developers know that a dependency is outdated. I am aware of david-dm.org for example...

[![peerDependencies Status](https://david-dm.org/guylepage3/lepage.cc.svg?colorA=212121&colorB=007BFF)](https://david-dm.org/guylepage3/lepage.cc)

But it would be amazing to have a badge for each of my dependencies..

guylepage3:
Hey @paulmelnikow, I was thinking more along the lines of...

Up to date version is blue or green...
[![React Router DOM version](https://img.shields.io/badge/React_Router_DOM-v4.3.1-color.svg?colorA=212121&colorB=007BFF)](https://www.npmjs.com/package/react-router-dom)

[![React Router DOM version](https://img.shields.io/badge/React_Router_DOM-v4.2.1-color.svg?colorA=212121&colorB=FF0000)](https://www.npmjs.com/package/react-router-dom)

nottrobin:
Any movement on this? Did anyone ever find a way to do this? I would also like to be able to display the exact version of a specific dependency that's used by the project.

paulmelnikow:
Yea, we chatted on Discord about using the dynamic badge. @guylepage3 could you point us to what you ended up with?

guylepage3:
Yeah. I was only able to execute what @chris48s is suggesting above. I would really love to remove the tilde `~` or `^` and replace it with `v`. But yeah. I didn't have anymore time to put into it. Sorry.

nottrobin:
Works perfectly, thanks @chris48s, @guylepage3. I'm not sure replacing the `~` or `^` would be the right thing to do, as someone installing dependencies from `package.json` might not get that exact version. If the script could instead read from `package-lock.json` or `yarn.lock`, then you could confidently communicate the exact version.

Does anyone know if there's a way to get a similar badge for specific Python dependencies from `requirements.txt`?

guylepage3:
For my purposes I was using the badge as a quick visual to let the viewer know which version of the dependency was currently in use. So for me it made sense. Much like how shields.io has set up my npm badge.
[![npm version](https://img.shields.io/npm/v/@universelabs/meta.svg?colorA=%23212121&colorB=%23007BFF)](https://www.npmjs.com/package/@universelabs/meta)

paulmelnikow:
[...]

PR #2461, opened 2018-12-05
Add [npm] badges for collaborator count and dependency version

paulmelnikow:
This adds a badge for collaborator count. When evaluating a library, it can be useful to know that there's not a single-contributor bottleneck for publishing. Having more than one collaborator is a sign of library maturity.

It adds another badge for dependency version of published dependencies, which solves a similar problem as the node-version badge. I will find this useful for making sure dependencies are up to date in a library.

There are lots of possible refinements here:

- Explicit support for popular frameworks. These could automatically pick the right dependencies, show status colors (since we can easily cache the latest version of popular frameworks), and possibly provide options to say which versions are supported and which are used for testing.
- package.json support, for applications, which I'll add next (#2259)

It also updates the base service to use a pattern instead of a regex, in the case where that's possible. Hopefully the other one will be possible after the fix in pillarjs/path-to-regexp#173.


services/npm/npm-base.js
```diff
@@ -33,15 +41,15 @@ module.exports = class NpmBase extends BaseJsonService {
     if (withTag) {
       return {
         base,
-        format: '(?:@([^/]+))?/?([^/]*)/?([^/]*)',
+        // The trailing optional means this has to be a regex.
+        format: '(?:(@[^/]+)/)?([^/]*)/?([^/]*)',
```
paulmelnikow: Have moved the `@` inside the capture group to harmonize with the pattern.

paulmelnikow:
Hey @calebcartwright, how would you feel about reviewing this PR? I've been writing a lot of code so there's a bit of a backlog to review. It'd be great to have this kind of help from you! (Also, no pressure!)

calebcartwright:
@paulmelnikow happy to help, will do!

paulmelnikow:
Awesome 😄

services/npm/npm-base.js
```diff
@@ -20,6 +24,10 @@ const schema = Joi.object({
       Joi.alternatives(Joi.string(), deprecatedLicenseObjectSchema)
     )
   ),
+  maintainers: Joi.array()
```
calebcartwright: First thought I had here is about the usage of `maintainers` vs. `collaborators` key. I've seen both used before although I'm honestly not sure what difference they have to npm. However, if we're going to use maintainers here would it make sense to use `maintainers` instead of `collaborators` for the service and path name?

services/npm/npm-collaborators.service.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+const { colorScale } = require('../../lib/color-formatters')
+const NpmBase = require('./npm-base')
+
+const keywords = ['node']
+
+const collaboratorColor = colorScale([2, 3, 4, 6])
+
+module.exports = class NpmCollaborators extends NpmBase {
```
calebcartwright: Same comment as above in `base` service about consistency between `maintainers` + `collaborators`

services/npm/npm-base.js
paulmelnikow: The API uses `maintainers`, so in the schema we don't have a choice.

The npm website uses Collaborators:

screen shot 2018-12-08 at 6 41 07 pm

I figure they've updated their terminology but did not modify the API, which is why I went with **collaborators**.

calebcartwright: Gotcha that makes sense

services/npm/npm-collaborators.service.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+const { colorScale } = require('../../lib/color-formatters')
+const NpmBase = require('./npm-base')
+
+const keywords = ['node']
+
+const collaboratorColor = colorScale([2, 3, 4, 6])
```
chris48s: Low bus factor is good, but I'm not totally convinced on trying to devise an absolute scale of what number of collaborators is 'good' or 'bad' and applying that to all projects regardless of the size of their codebase or community.

I think its reasonable to say only one maintainer is bad (and so is zero) for all projects, but maybe just report the number in blue for everything >1 ?

paulmelnikow: Part of my use case is discerning the size of the community, so it’s a little chicken and egg. I can see your point that we’re not in a position to say that 4 is too few or 6 is enough, and that the expectation of what’s good would change for widely used tools.

1 is bad. Should 2 be caution? (Maybe orange or yellow.) Keeping in mind a badge this administrative is less likely to be used on individual readmes than on organizational badgeboards or project comparison scorecards tracking a large number of projects.

chris48s:
btw, cheers for reviewing @calebcartwright

Issue #2431, opened 2018-11-29
Automatically show light or dark logos depending on left color


paulmelnikow:
As was variously suggested in this thread…

https://github.com/badges/shields/issues/1535#issuecomment-371082250
https://github.com/badges/shields/issues/1535#issuecomment-371568668
https://github.com/badges/shields/issues/1535#issuecomment-372479642

it would be nice to support light and dark variants of logos. Perhaps these could be chosen automatically based on the template's left color.

The light and dark versions could be curated or determined programmatically.

It's already possible to specify an explicit color thanks to @RedSparr0w's work in #1810; this is a refinement that would make it automatic.

RedSparr0w:
How would we go about current logos with colors/multi colors?
Will we just change them all to be automatic.

Also would we want to add an option like `?logoColor=default` or `?logoColor=brand` to allow for the default simple icons colors?

paulmelnikow:
Since we only support two colors for the left side, whatever the social template uses (white?) and `gray`, and are likely to want to customize some of these anyway, it seems simplest to just explicitly override the ones that need overriding.

We could build a diagnostic page that makes it easy to see all the logo colors. That would help with diagnosing this.

> Also would we want to add an option like `?logoColor=default` or `?logoColor=brand` to allow for the default simple icons colors?

At the moment it doesn't seem necessary. I think we should pick good default colors for the two background colors we support in the templates. If someone is customizing the background color, they are way out on a limb, and can inspect and specify the brand color easily enough.

By the way I started refactoring the logo code as a sub-task of #2473 but it depends on the color code so I had to refactor that first. After that's merged (#2742) I'll take a pass at the existing logo functionality. I won't implement this, but I'll keep it in mind.

chris48s:
So my thought was that we want to avoid a situation where we've got light logo on light background or dark logo on dark background, preferring the brand colour if possible. I'd also like to avoid having to individually store our preferred colour for each icon in the simple-icons set (they hold over 500 icons so far). We can work out if a colour is 'light' or 'dark' using something like https://awik.io/determine-color-bright-dark-using-javascript/ (or there might be a npm package for it). Then my thought was we do something like (pseudo-code):

```
if we're using our own icon, assume its multi-coloured and just use it as-is

if we're using simpleicons {
    if isLight(backgroundColor) and isLight(brandColor) {
        iconColor = black
    }
    if isDark(backgroundColor) and isDark(brandColor) {
        iconColor = white
    } else {
[...]

PR #2463, opened 2018-12-05
Add Azure DevOps logo

aalmada:


aalmada:
I think it now follows the guidance for logo submissions. Azure DevOps is one of the supported CI platforms and the logo was minimised using svgomg.

chris48s:
Nice one. Thanks for minifying. Do you think it would be better to make the default logo white so it pops a bit more on the default background:




White seems to be an approved colour for the logo: https://twitter.com/azuredevops?lang=en

aalmada:
I can change it to white but please note that most logos in that folder are in their original colors. The color can be changed using the _logoColor_ parameter. 
For example, the GitHub logo is black at https://github.com/badges/shields/blob/master/logo/github.svg but I'm using it in light gray at https://github.com/NetFabric/NetFabric.DoubleLinkedList/blob/master/README.md

chris48s:
@RedSparr0w - do you have an opinion on this one?

paulmelnikow:
It's hard to imagine anyone wanting blue on gray; it doesn't seem like the best default. For what it's worth, the GitHub logo is dark because it was originally designed for the static badge, which has a light background. If white is an approved color, my vote is for that, at least until we add something like #2431.

calebcartwright:
FWIW - the default logo color is white for the Azure Pipelines badge (although it is a different logo) 

https://github.com/Microsoft/vscode#visual-studio-code---open-source
image

RedSparr0w:
I think we should try to defer all icons to [simple-icons](https://github.com/simple-icons/simple-icons).
And if they do not accept it or it doesn't fit their spec for whatever reason then open one here.
What are your thoughts? @paulmelnikow @chris48s 

Otherwise if we are going to be merging this PR, i agree the white works better.

paulmelnikow:
> I think we should try to defer all icons to [simple-icons](https://github.com/simple-icons/simple-icons).

I agree, I don't see a real advantage to keeping our own collection of icons.

So that we don't string along @aalmada, I might suggest:

1. Merge this with white.
2. Modify our logo guidelines to say we are no longer accepting new logos
3. Match up our existing logo collection with simple-icons, and get our existing logos matched in there. Maybe in the process, one of us could ask to join as a co-maintainer.
4. Remove our existing logos, leaving aliases in place for if needed for backward compatibility.

aalmada:
White it is then... :)
Commits are all squashed to a single one.

Issue #1752, opened 2018-06-27
Add functionality for change color dynamically


chandankumar4:
- Add functionality to change color dynamically for badge just like their data while creating the dynamic badge

RedSparr0w:
#1525 Adds this functionality 😄

paulmelnikow:
It could be nice to support in the standard json badge as well. I can't remember if the code in #1525 is doing that.

paulmelnikow:
I think the endpoint badge, when it's merged, will be a good way of handling this. Let's track in #2473.

chandankumar4:
Thanks @paulmelnikow for adding this feature

Issue #1534, opened 2018-02-28
Actual full gem downloads number?


AlphaDaniel:
Is it possible to show the full number of downloads instead of rounding. I came across [ruby-gem-downloads-badge](https://github.com/bogdanRada/ruby-gem-downloads-badge) which was built off of this repo and it displays full 

![](http://ruby-gem-downloads-badge.herokuapp.com/ruby_doc?type=total)

but it seems to have caching issues where after I guess the initial request following requests do not get to the application and the badge never changes. 

Would be awesome if had an option with shields.io, added to the other gem downloads options, where it can display exact number.

RedSparr0w:
related - #999

AlphaDaniel:
Oh ok so that was for stars... is the request for gem the same? 

Just seen it was already done here and looks perfect ![](http://ruby-gem-downloads-badge.herokuapp.com/ruby_doc?type=total) would it not be possible to add this option to gem downloads?

My original request was for a simple additional extra option for gem downloads badges not all so in theory

![11](https://user-images.githubusercontent.com/33204668/36775208-20b79120-1c2f-11e8-8c40-44d25e7fa3e8.png)
Gem:         
 ![](http://ruby-gem-downloads-badge.herokuapp.com/rails?type=total)  
 `https://img.shields.io/gem/dt/rails.svg?type=total`

RedSparr0w:
Hey, sorry about the short reply (was on mobile and haven't had much time on this project lately), Just thought i would quickly link some relevant post in case something happened in those which affected this or vice versa.

As this would be quite a big change it would have to be approved by @paulmelnikow and or @espadrine.

As per [this comment](https://github.com/badges/shields/issues/999#issuecomment-341309430):
> My question is still this:
>> This request is for https://github.com/mingrammer/go-web-framework-stars. If the Shields badge designs were integrated into that page, how would it look?
>
> If anyone would like to mock something up (using static badges, say), I'd be happy to consider it.

I think creating a few mockups would be the best place to start.

AlphaDaniel:
Hey thanks for replying and hmm so weird didn't get an email would've replied sooner. As far as the mock up goes, I'd see it looking exactly as [bogdanRada](https://github.com/bogdanRada/ruby-gem-downloads-badge) did it. See below

![](http://ruby-gem-downloads-badge.herokuapp.com/ruby_doc?type=total)

bogdanRada:
[...]

Issue #1519, opened 2018-02-21
Status based on current month


swapagarwal:
Is it possible to have a badge that stays green in October and red in other months?

bkdotcom:
A badge that displays the current month?
Always displays month x, but color is based on whether x is current?

what's your use case?

swapagarwal:
I want to add a badge here in Hacktoberfest: https://github.com/swapagarwal/swag-for-dev
The badge should show `status=active` in October and `status=inactive` in other months. 😅

RedSparr0w:
Unfortunately we do not have a badge which does that,
Although you could setup a tiny script to return the image you were wanting:

[Example in php:
![](https://shields.redsparr0w.com/1519)](https://shields.redsparr0w.com/1519)
```php
<?php
header('Content-type: image/svg+xml');

$isOctober = Date('F') === 'October';

$label = 'status';
$message = $isOctober ? 'active' : 'inactive';
$color = $isOctober ? 'brightgreen' : 'red';
echo file_get_contents("https://img.shields.io/:{$label}-{$message}-{$color}.svg");
?>
```

bkdotcom:
ya.. at first I'd say.. just dynamically generate ...
ie https://img.shields.io/badge/Hacktoberfest-inactive-red.svg
but if you're wanting other's to easily use the badge...

*Idea* 💡
a json-url service
something like
https://img.shields.io/json/http%3A%2F%2Fmyurl.com%2Fbadge.json

where http://myurl.com/badge.json returns
```json
{
  "label": "Hactoberfest",
  "status": "inactive",
  "colorA": null, // or omit for default
  "colorB": "red",
  "logo": null,  // or omit
  "logoWidth": null, // or omit
}
```

RedSparr0w:
```php
<?php
header('Content-type: image/svg+xml');
[...]

PR #2473, opened 2018-12-07
[Endpoint] badge

### LINKED ISSUE #204: NuGet badges for multiple packages
Some projects/repos, e.g. https://github.com/scriptcs/scriptcs, produce a number of NuGet packages, e.g. https://www.nuget.org/profiles/scriptcs/.

It would be nice to be have NuGet badges which took all packages into account, certainly for download count.

Version would also work if the versions of all the packages are the same. If they differ, perhaps the badge could display the latest version, a version range, or just '4 packages'.

I think the name of each package would have to be specified in the URL, e.g. `http://img.shields.io/nuget/multi/dt/shields.svg?ScriptCs.Contracts&ScriptCs.Core&ScriptCs.Engine.Roslyn&ScriptCs.Hosting`

Specifying the packages using a filter is difficult since the only options are search, author and tags. The only one under definitive control of the publisher is author, but even that will only be narrowly useful since most projects don't create an author specifically for their package publishing. Even if they did, it would be fairly common that they would publish packages from many repos which means the badge wouldn't be useful for puttting on a repo specific readme. Still, that could also provided as an option, e.g. `http://img.shields.io/nuget/profile/dt/scriptcs.svg`.

I guess another badge could give package count, e.g. `http://img.shields.io/nuget/multi/pc/shields.svg?ScriptCs.Contracts&ScriptCs.Core&ScriptCs.Engine.Roslyn&ScriptCs.Hosting` and `http://img.shields.io/nuget/profile/pc/scriptcs.svg`.

paulmelnikow:
This reimplements the idea @bkdotcom came up with in #1519, and took a stab at in #1525. It’s a really powerful way to add all sorts of custom badges, particularly considering [tools like RunKit endpoints and Jupyter Kernel Gateway](https://github.com/badges/shields/issues/2259#issuecomment-444186589), not to mention all the other ways cloud functions can be deployed these days.

I gave this a name which did not have JSON in it, so as not to confuse it with the dynamic JSON badges.

There are a few things that need doing before this can be shipped, including supporting several of the options I included in the schema.

- [x] labelColor (#2740)
- [x] isError (#2693)
- [ ] ~~link (I honestly don't know much about our link support; should we have leftLink and rightLink?)~~ (Removed for this iteration)
- [x] namedLogo, logoSvg (#2796)
    - [x] test of logoSvg
- [x] logoColor, logoWidth, logoPosition (#2796)
    - [x] test of logoWidth, ~~logoPosition~~
- [x] style
- [x] cacheSeconds (#2755)
    - [x] Service test for cacheSeconds

Alternatively, they could be removed from the schema and added later on.

Curious for feedback! And thought I would open this up for comments before I got too far along.

Ref #1534 #2259. Closes #204 #358 #742 #999 #1519 #1752 #2625.


calebcartwright:
I don't really have any useful feedback on this one yet. React is wayyyy too knew for me and I'm still trying to get a feel for that endpoint service. 

Will try to get back to this one hopefully soon once I've read up a bit

paulmelnikow:
> React is wayyyy too new for me and I'm still trying to get a feel for that endpoint service.

Probably understanding the React code isn't that important here as it's just documentation. Looking at the review app would probably be sufficient.

calebcartwright:
Okay I think I've found it now 😄 I was looking for an `endpoint` badge via the search function as well as perusing the page but couldn't find it. 

I was able to get it via direct uri at https://shields-staging-pr-2473.herokuapp.com/#/endpoint

Can I point this thing to any endpoint that returns any JSON response?

paulmelnikow:
> Can I point this thing to any endpoint that returns any JSON response?

Yea, that's exactly the idea! 😁Or rather, a JSON response in the specified format.

This is one I created for #2259: https://runkit.com/melnikow/version-from-requirements-txt

(Oops, that formatting will need some work.)

calebcartwright:
Gotcha. So this is conceptually similar to the dynamic badges, except that it offers the users the ability to define their own processing functionality

calebcartwright:
Do we have any idea how we might define guidelines recommendations for when a user should use this type of badge vs. implementing their own service in the Shields codebase and submitting that as a PR?

paulmelnikow:
> Gotcha. So this is conceptually similar to the dynamic badges, except that it offers the users the ability to define their own processing functionality

Right.

It also allows a service to publish a prerendered badge which can be proxied through shields.io to provide formatting flexibility.


> Do we have any idea how we might define guidelines recommendations for when a user should use this type of badge vs. implementing their own service in the Shields codebase and submitting that as a PR?

It's a great way to get something going quickly and to quickly iterate and remix.

Beyond that I'd say if folks have something that's working well and they think it's generally useful, they should open an issue to suggest it. I don't have clear guidelines of what's too specific or too generic to include.


We probably wouldn't merge "is such-and-such the current month or not" (#1519) or "how many days are left in the year," however I could see building something along those lines that's specific to hacktoberfest.

This is also a good candidate for things we don't want to do, like exact star count or download count (#742), or nuget aggregate download counts (#204).

calebcartwright:
> It also allows a service to publish a prerendered badge which can be proxied through shields.io to provide formatting flexibility. 

Nice! I was thinking along the same lines of use cases. I was envisioning a scenario where someone may use this to get something going quickly, but then once they get it working well they just use it as is and don't not bother to share it back with the community 😄 

Obviously that's not something that can be prevented, but maybe just a note in docs down the line that includes the exact sentiment you articulated:

> Beyond that I'd say if folks have something that's working well and they think it's generally useful, they should open an issue to suggest it

paulmelnikow:
> I was envisioning a scenario where someone may use this to get something going quickly, but then once they get it working well they just use it as is and not bother to share it back with the community 😄

Right, that's a possibility. Though the advantage of pulling the implementation into the shields server is that it will take load off their server endpoint, and probably will perform better and be more reliable.

calebcartwright:
One odd thing I've noticed is that the first time I hit the endpoint page I see what I showed above in https://github.com/badges/shields/pull/2473#issuecomment-451310460, but after I reload the page I see this:

image

paulmelnikow:
Another guideline of when you might want to use an endpoint badge: for a service which is too slow or unreliable for Shields to advertise it.

paulmelnikow:
I'm thinking about deferring link support in the endpoint badge in order to give more thought to what Shields should really offer, and how to best design the API in accordance.

paulmelnikow:
This is getting really close!

paulmelnikow:
Other than the frontend formatting issue, I think this is ready to go.

I'm running into an interop issue between styled-components and Next.js SSR. I've spent a fair amount of time trying to fix it and don't think I should spend any more given #1943.

We could either block on the formatting fix, or open a bug ticket and fix the formatting in a follow-on.

The formatting doesn't look great, but it's functional, and probably people who are interested in the functionality would forgive us for a week of bad formatting. In the interest of keeping things moving, and avoiding creating conflicts in this PR as we implement #1943, I'd lean toward merging and opening a bug ticket for the fix.

lib/validate.js
```diff
@@ -11,17 +11,21 @@ function validate(
     includeKeys = false,
     traceErrorMessage = 'Data did not match schema',
     traceSuccessMessage = 'Data after validation',
+    allowAndStripUnknownKeys = true,
```
calebcartwright: i have no clue what all this is used for, but my first thought here was whether any consumers of this function would ever need different values for these two keys (allowUnknown false and stripUnknown true)?

services/cache-headers.spec.js
```diff
@@ -35,6 +35,11 @@ describe('Cache header functions', function() {
         serviceDefaultCacheLengthSeconds: 900,
         queryParams: { maxAge: 1000 },
       }).expect(1000)
+      given({
+        cacheHeaderConfig,
+        serviceDefaultCacheLengthSeconds: 900,
```
calebcartwright: Would it be feasible here to update the naming conventions used here (and in cache-headers.js`) to match with the new naming conventions used in the base service (basically dropping `Length` I believe)

services/endpoint/endpoint.service.js
```diff
@@ -0,0 +1,129 @@
...
+
+const blockedDomains = ['github.com', 'shields.io']
+
+const queryParamSchema = Joi.object({
+  url: optionalUrl.required(),
+}).required()
+
+const anySchema = Joi.any()
+
+const optionalStringWhenNamedLogoPrsent = Joi.alternatives().when('namedLogo', {
```
calebcartwright: nitpick here on typo optionalStringWhenNamedLogoPrsent -> optionalStringWhenNamedLogoPr**e**sent

services/endpoint/endpoint.service.js
```diff
@@ -0,0 +1,129 @@
...
+
+  async handle(namedParams, queryParams) {
+    const { url } = this.constructor._validateQueryParams(
+      queryParams,
+      queryParamSchema
+    )
+
+    const { protocol, hostname } = new URL(url)
+    if (protocol !== 'https:') {
+      throw new InvalidParameter({ prettyMessage: 'please use https' })
```
calebcartwright: 👍

calebcartwright:
I went through the changes and left feedback/questions where I could, but have to defer on the frontend content to someone with a better eye for that sort of thing. Bravo on another big add 🍻! 

> I'd lean toward merging and opening a bug ticket for the fix. 

That makes sense to me. I feel like there's been a lot of issues opened lately that will be resolved by this so I definitely see value in getting the functionality out even if there's some short term formatting items.

services/cache-headers.spec.js
paulmelnikow: Yea, I dropped `Length` which doesn't seem to add anything (cache seconds == cache length seconds). Agreed that cleaning up the rest of them is a nice idea, though I'd rather deal with it in a follow-on. `BaseService` has a property called `_cacheLength` that probably should be renamed to `cacheSeconds` and that's bound to affect a handful of files and code that's spread about a few places.

lib/validate.js
paulmelnikow: It seems unsafe to `allowUnknown` and not `stripUnknown` so if we did that, I'd tend to think we weren't doing a good job of validating.

Originally I'd passed `options` directly through which would have supported any combination, though I'm not sure we would ever want to do that.

paulmelnikow:
I'm going to add a **Beta** label to this feature, which will give us a chance to change the API if we decide we want to.

services/cache-headers.spec.js
calebcartwright: Gotcha and agreed 👍 When I first saw this I thought I was still up in the BaseService file, took me a second to remember I'd moved on to another file

frontend/components/endpoint-page.js
```diff
@@ -0,0 +1,217 @@
...
+    <Header />
+    <H3 id="static-badge">Endpoint (Beta)</H3>
+    <Snippet snippet={`${baseUrl}/badge/endpoint.svg?url=...&style=...`} />
+    <p>Endpoint response:</p>
+    <JsonExample
+      data={{
+        schemaVersion: 1,
+        label: 'hello',
+        message: 'sweet world',
+        color: 'orange',
```
calebcartwright: 😆 For some reason my brain read this as `hello sweet orange`. wonder if i am hungry...

paulmelnikow:
Do you mean you don't have much to offer on the way it's coded, or about what it's saying to the user?

paulmelnikow:
Opened #2837 for the formatting issues. I'm gonna ship this!!!

paulmelnikow:
And opened #2838 for further discussions.

PR #2474, opened 2018-12-07
Modernize chrome web store examples; PR check state -> build

paulmelnikow:


services/chrome-web-store/chrome-web-store.service.js
```diff
@@ -102,43 +109,27 @@ class ChromeWebStoreRating extends LegacyService {
...
-        previewUrl: 'rating-count/ogffaloegjglncjfehdfplabnoondfjo',
-        keywords: ['chrome'],
-      },
-    ]
-  }
-
-  static registerLegacyRouteHandler() {}
-}
-
-class ChromeWebStoreUsers extends LegacyService {
```
paulmelnikow: `/users` and `/d` are the same so I dropped `/d` and updated `/users` above.

PR #2470, opened 2018-12-06
Rewrite [GithubManifest] and [GithubPackageJson] badges; also affects [GithubPullRequestCheckState DynamicJson]

paulmelnikow:
Pave the way for #2259 and rewrite #1721 along the way.

Ref: #2320 

Marked blocker because of the `BaseJsonService` changes.

Issue #2347, opened 2018-11-17
Investigate failing nexus badge tests


paulmelnikow:
```
  18) Nexus
       search release version
         
	[ GET /r/https/repository.jboss.org/nexus/jboss/jboss-client.json ]:
     ValidationError: child "value" fails because ["value" with value "invalid" fails to match the required pattern: /^v4(\.\d+)+$/]
      at Object.exports.process (node_modules/joi/lib/errors.js:201:19)
      at internals.Object._validateWithOptions (node_modules/joi/lib/types/any/index.js:751:31)
      at module.exports.internals.Any.root.validate (node_modules/joi/lib/index.js:146:23)
      at Object.pathMatch.matchJSONTypes (node_modules/icedfrisby/lib/pathMatch.js:303:9)
      at _expect (node_modules/icedfrisby/lib/icedfrisby.js:563:10)
      at IcedFrisbyNock._invokeExpects (node_modules/icedfrisby/lib/icedfrisby.js:1261:26)
      at start (node_modules/icedfrisby/lib/icedfrisby.js:1244:12)
      at Request.runCallback [as _callback] (node_modules/icedfrisby/lib/icedfrisby.js:1131:16)
      at Request.self.callback (node_modules/request/request.js:185:22)
      at Request.<anonymous> (node_modules/request/request.js:1161:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:1083:12)
      at endReadableNT (_stream_readable.js:1064:12)
      at _combinedTickCallback (internal/process/next_tick.js:139:11)
      at process._tickDomainCallback (internal/process/next_tick.js:219:9)

  19) Nexus
       search release version of an inexistent artifact
         
	[ GET /r/https/repository.jboss.org/nexus/jboss/inexistent-artifact-id.json ]:

      AssertionError: expected { name: 'nexus', value: 'invalid' } to deeply equal { name: 'nexus', value: 'no-artifact' }
      + expected - actual

       {
         "name": "nexus"
      -  "value": "invalid"
      +  "value": "no-artifact"
       }
      
      at Object.pathMatch.matchJSON (node_modules/icedfrisby/lib/pathMatch.js:138:38)
      at _expect (node_modules/icedfrisby/lib/icedfrisby.js:583:10)
      at IcedFrisbyNock._invokeExpects (node_modules/icedfrisby/lib/icedfrisby.js:1261:26)
      at start (node_modules/icedfrisby/lib/icedfrisby.js:1244:12)
[...]

PR #2477, opened 2018-12-08
Add new service for CII Best Practices badges [CIIBestPractices]

### LINKED ISSUE #1530: Add Core Infrastructure Initiative (CII) Best Practices badge
https://bestpractices.coreinfrastructure.org/ provides a "checklist" for OSS projects, guiding in project management, CI, security etc. It supports its own badge, but no options for it. Example:

`[Linux Kernel Badge](https://bestpractices.coreinfrastructure.org/projects/34)`
[Linux Kernel Badge](https://bestpractices.coreinfrastructure.org/projects/34)

REST API URL: https://bestpractices.coreinfrastructure.org/projects/34/badge.json (34 is project id)

calebcartwright:
Adds support for badges from [CII Best Practices](https://bestpractices.coreinfrastructure.org) which closes #1530 

List of projects can be found at https://bestpractices.coreinfrastructure.org/en/projects for reference

I've never used the service before so some comments/info based on what I've seen: 
* The CII service provides a rating for a project which is represented with a value of 0-300 in a tiered scale (value is represented as `tiered_percentage` in the API response)
* More details of how CII derives that score can be found [here](https://github.com/coreinfrastructure/best-practices-badge) and [here](https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md), but the important point is that CII natively provides a badge (albeit a non-configurable one) that is ultimately defined by that score
* The API response for each service has a field for that raw score value, as well as a grade/level that is defined based on that score:
  * level is `in_progress` for score of 0-99 (inclusive)
  * level is `passing` for score of 100-199 (inclusive)
  * level is `silver` for score of 200-299 (inclusive)
  * level is `gold` for a perfect score of 300
* The message of the default badge provided by the service varies both in content and format depending on the value of the level/score (when the score is less than `100` the badge message includes the score percentage with a wide array of colors, otherwise the badge message is simply the level)

As such, I decided to add support for 3 different types of badges for CII Best Practices Data that will give end users a lot more flexibility over the native CII badge:
1.  `level` - uses the CII level for the badge message with a simple color scheme
2. `percentage` uses the CII score/percentage with the standard Shields percentage color scheme
3. `default` - uses the same badge message & color scale as the native CII badge provides, but with all the configuration and goodness that Shields provides! (I'm not crazy about `default` as the path name/label but I wasn't sure what else to call this one)


PyvesB:
I'll review this shortly. In the meantime, here's the [link ](https://shields-staging-pr-2477.herokuapp.com) to a Heroku review app with the contents of this PR deployed.

calebcartwright:
Awesome, thanks @PyvesB!

services/cii-best-practices/cii-best-practices.tester.js
```diff
@@ -0,0 +1,252 @@
...
+const Joi = require('joi')
+const t = require('../create-service-tester')()
+module.exports = t
+
+t.create('live: level known project')
+  .get(`/level/1.json`)
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'cii',
+      value: Joi.string(),
```
PyvesB: The three first live tests would benefit from stronger validation on the `value` parameter. Currently, if the tested projects were deleted or if the API changed and we were no longer able to parse the response, the tests would still pass as the value would still be a string. You can have a look at [test-validators.js](https://github.com/badges/shields/blob/master/services/test-validators.js) for some inspiration. 😉

services/cii-best-practices/cii-best-practices.tester.js
```diff
@@ -0,0 +1,252 @@
...
+    })
+  )
+
+t.create('live: unknown project')
+  .get(`/level/abc.json`)
+  .expectJSON({ name: 'cii', value: 'project not found' })
+
+t.create('level: unknown project')
```
PyvesB: I think the `live: unknown project` test is enough as far as unknown projects are concerned. The bit of code we're testing is the `404: 'project not found'`, having three additional mocked tests will not improve our coverage nor give us more confidence that the badge is wired up correctly.

services/cii-best-practices/cii-best-practices.service.js
```diff
@@ -0,0 +1,162 @@
...
+    return [
+      {
+        title: 'CII Best Practices Level',
+        namedParams: {
+          metric: 'level',
+          projectId: '1',
+        },
+        pattern: 'level/:projectId',
+        staticExample: this.render({
+          label: 'cii',
```
PyvesB: I don't think you have to repeat `cii` here, nor have the `label` parameter in the render function. This is specified as part of the default badge data.

services/cii-best-practices/cii-best-practices.service.js
```diff
@@ -0,0 +1,162 @@
...
+        }),
+        keywords,
+      },
+    ]
+  }
+
+  static get route() {
+    return {
+      base: 'cii',
+      pattern: ':metric(level|percentage|default)/:projectId',
```
PyvesB: Nice! 👍

services/cii-best-practices/cii-best-practices.service.js
```diff
@@ -0,0 +1,162 @@
...
+        pattern: 'percentage/:projectId',
+        staticExample: this.render({
+          label: 'cii',
+          message: '107%',
+          color: 'brightgreen',
+        }),
+        keywords,
+      },
+      {
+        title: 'CII Best Practices Default',
```
PyvesB: Maybe we could make the meaning of "default" clearer by adding some documentation to the example? Pretty much taken from your PR's description: "This badge uses the same message and color scale as the native CII one, but with all the configuration and goodness that Shields provides!". You can refer to our [Matrix](https://github.com/badges/shields/blob/master/services/matrix/matrix.service.js) badge for an example of how to add documentation.

paulmelnikow: > 1. `level` - uses the CII level for the badge message with a simple color scheme
> 2. `percentage` uses the CII score/percentage with the standard Shields percentage color scheme
> 3. `default` - uses the same badge message & color scale as the native CII badge provides, but with all the configuration and goodness that Shields provides! (I'm not crazy about `default` as the path name/label but I wasn't sure what else to call this one)

`status` or `summary` might be a more descriptive name that emphasizes the content that’s being shown.

services/cii-best-practices/cii-best-practices.tester.js
calebcartwright: 🤦‍♂️ - good catch!

services/cii-best-practices/cii-best-practices.service.js
calebcartwright: Yeah I was having similar thoughts myself, but just wasn't sure what to put in there for the documentation. I like the idea though!

@paulmelnikow - are you suggesting `status` or `summary` as a replacement for `default` (both of which are far better than `default`) and/or in place of the other two (`level` or `percentage`)

services/cii-best-practices/cii-best-practices.service.js
```diff
@@ -0,0 +1,162 @@
...
+      color,
+    }
+  }
+
+  static get defaultBadgeData() {
+    return { color: 'blue', label: 'cii' }
+  }
+
+  static get category() {
+    return 'other'
```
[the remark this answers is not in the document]
calebcartwright: I had no clue on this one so definitely defer to whatever you all think is best. I spent _just_ enough time reviewing CII to be able to write the badge service for it but I still know next to nothing about it :smile: 

It does seem like CII is an amalgamation of things similar to other badges under the `build` category so that seems logical to me FWIW

calebcartwright:
Thanks again for the feedback! Think I addressed everything with the exception of one question to clarify whether some of the tests should be removed. 

Also the [circleci:danger](https://circleci.com/gh/badges/shields/28415?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) job is failing on `npm install` with an interesting conflict error

services/cii-best-practices/cii-best-practices.service.js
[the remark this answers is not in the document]
calebcartwright: I switched the category to `build`

services/cii-best-practices/cii-best-practices.service.js
calebcartwright: i changed `default` to `summary`

services/cii-best-practices/cii-best-practices.tester.js
PyvesB: Yes, my suggestion was initially to remove all the project not found tests apart from the live one. You're right by saying that we might need additional ones in the future if the implementation changes, however in the meantime they (slightly) slow down our test suite and add code to maintain. ^^

PyvesB:
Thanks for addressing my review comments! I've clarified the one regarding the project not found tests. I've also kicked the CI server again, that spurious error seems to have gone away. 😉

calebcartwright:
Alrighty then, tests removed as requested

calebcartwright:
Excellent, much appreciated. Now I need to see which issues I can work on next!

PyvesB:
@calebcartwright I noticed that you had started a discussion in #2347, maybe it's a good one to tackle? If you're more interested in reviewing something at this point, #2470 is probably a good choice as well. 😉

calebcartwright:
Thanks @PyvesB and yeah the Nexus service one is pretty high on my list (especially since I haven't refactored a legacy service yet), but happy to take a look at the PR as well if it's ready for review. 

@paulmelnikow mentioned in another thread that there was plenty of work to do here, so I'm just happy to contribute any way I can. If there's anything in particular I can to do to help please let me know, otherwise I'm just perusing the open issues for ones I think I can work on.

PR #2478, opened 2018-12-08
More example urls and static examples [apm waffle]

paulmelnikow:


services/github/github-followers.service.js
```diff
@@ -25,7 +25,15 @@ module.exports = class GithubFollowers extends LegacyService {
...
         title: 'GitHub followers',
-        previewUrl: 'espadrine',
+        pattern: ':user',
+        namedParams: {
+          user: 'espadrine',
+        },
+        staticPreview: {
+          label: 'Follow',
+          message: '150',
+          // style: 'social',
```
paulmelnikow: ^^ This probably needs to be fixed before this is merged. Or else this one could be reverted.

services/jira/jira-issue.service.js
```diff
@@ -19,7 +19,17 @@ module.exports = class JiraIssue extends LegacyService {
     return [
       {
         title: 'JIRA issue',
-        previewUrl: 'https/issues.apache.org/jira/KAFKA-2896',
+        pattern: ':protocol/:hostAndPath*/:issueKey',
```
PyvesB: This probably doesn't matter much as it's the example pattern, however wouldn't we be expecting `:hostAndPath+` rather than `:hostAndPath*`?

PR #2487, opened 2018-12-08
Rewrite [DynamicXml] and [DynamicYaml] badges

paulmelnikow:
- A little related cleanup in `server.js`: remove a couple import renames.

Blocking next steps in `server.js`.


services/dynamic/dynamic-yaml.service.js
```diff
@@ -0,0 +1,78 @@
...
+    return createRoute('yaml')
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'custom badge',
+    }
+  }
+
+  parseYml(buffer) {
```
chris48s: I think it would be useful to extract this as a generic helper function so we can use it in the F-Droid service (and in a `BaseYamlService` in future)

Its a bit scope creep-ey though, so we could merge this PR and raise it as an issue for future so that we don't have to fiddle with the F-Droid badge in this PR, if you want.

paulmelnikow: Makes sense. That came up in https://github.com/badges/shields/pull/2399#issuecomment-444970457, as well.

PR #2492, opened 2018-12-09
refactor [bountysource] service

chris48s:


services/bountysource/bountysource.service.js
```diff
@@ -1,64 +1,57 @@
...
-              badgeData.text[1] = 'invalid'
-            }
-            sendBadge(format, badgeData)
-          }
-        })
-      })
-    )
+  static render({ total }) {
+    return {
+      message: total,
```
paulmelnikow: I've been meaning to change this to use `metric`. What do you think?

Issue #2495, opened 2018-12-09
Path in modal should work with URL base


chris48s:
Are you experiencing an issue with...
 - [x] [shields.io](https://shields.io/#/)
 - [ ] My own instance
 - [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

Since deploying #2441 to production, the 'path' field expects the content _not_ to have a URL base set, but we use one everywhere else:

this doesn't work:

![screenshot at 2018-12-09 16-36-37](https://user-images.githubusercontent.com/6025893/49700099-35da4100-fbd2-11e8-8650-b16274a93d85.png)

you have to do this to get a preview to render:

![screenshot at 2018-12-09 16-37-10](https://user-images.githubusercontent.com/6025893/49700102-3ecb1280-fbd2-11e8-831b-f2e297c70f4b.png)

This was difficult to pick up on staging because there is no base set.

paulmelnikow:
Ah, hmm. I mentioned this change in the PR. Local dev looks like production. I didn't realize staging looked so different.

To be honest, I feel like the long URLs everywhere have to go. They make the page look incredibly busy:

<img width="1169" alt="screen shot 2018-12-09 at 8 18 44 pm" src="https://user-images.githubusercontent.com/1487036/49705853-ab083f00-fbef-11e8-9662-aa9515e42963.png">

I suppose you copied and pasted the example URL into the Path box and expected it to render. Is that right?

If so, I wonder if this can be made clearer. I have an idea about a possible solution.

Reverting the behavior is a possibility too.

paulmelnikow:
After discussing offline we decided to hold off on changing this given the changes in #2496 will wipe out this part of the workflow.

PR #2499, opened 2018-12-10
Fix for confusing path behavior in modal

paulmelnikow:
Proposed quick and dirty fix for #2495. This isn't perfect, though I think it addresses the worst of the confusing experience.

I tried to fix this in a better way but iterating on the styling is very slow. I want to address that developer experience before pushing too much further on that. Also I want to keep focused on #2495, which would change this modal in more significant ways.

### DIFF
--- a/frontend/components/markup-modal.js
+++ b/frontend/components/markup-modal.js
@@ -85,6 +85,27 @@ export default class MarkupModal extends React.Component {
     }
   }
 
+  handleResetClicked = event => {
+    const { badgeUrlForProps } = this.state
+    this.setState({ badgeUrl: badgeUrlForProps })
+  }
+
+  handleExampleClicked = event => {
+    const {
+      example: {
+        example: { pattern, namedParams, queryParams },
+      },
+    } = this.props
+    const exampleUrlPath = badgeUrlFromPattern({
+      pattern,
+      namedParams,
+      queryParams,
+    })
+    this.setState({
+      badgeUrl: exampleUrlPath,
+    })
+  }
+
   generateBuiltBadgeUrl() {
     const { baseUrl } = this.props
     const { badgeUrl, style } = this.state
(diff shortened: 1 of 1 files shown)


paulmelnikow:
This pull request **fixes 1 alert** when merging 301a8fe24256b7356f1ba9699b9cbfa63793ea14 into 1649a42a4a4865d03cb17640df8865c860f18c58 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-4f035670ff67db676663fa5c447a8c97e630b5a9)

**fixed alerts:**

* 1 for Unused or undefined state property

---

chris48s:
Thanks for having a look at this. This seems like an improvement, but it also seems like we're really going to a lot of effort to avoid using `baseUrl` in the Path `<input>` but I'm not sure why its important to do that given we're using it everywhere else in the modal. What am I missing?

paulmelnikow:
What makes the base url inconvenient is editing. The places we’re showing it aren’t editable, I don’t believe.

chris48s:
Hmm. I wonder if we're trying to solve 2 different problems? Sometimes wordsing is hard :)


> The places we’re showing it aren’t editable, I don’t believe

That's completely true, but its also not a new constraint.

For example, if I roll back `master` to 8a8311d (the commit before we merged #2441 ), there is only one editable field and I can either use an absolute or relative URL in it:

At current `HEAD` there is there is still one editable field and I can only use a relative URL in it.

Is the thing that has actually changed that we're now supressing the preview based on whether the contents of that box contains the `:` character, which means a string like `https://` (containing a colon) is causing the preview to be suppressed?

https://github.com/badges/shields/blob/43faea45613574ea5755b82c4e0d68796327f8d7/frontend/components/markup-modal.js#L102

If so, could we solve this by detecting if placeholders are present in a slightly more sophisticated way (e.g: by testing against a regex)?

Issue #2489, opened 2018-12-09
One microbadger test times out


paulmelnikow:
This specific Microbadger test has failed for at least the last three days, and maybe farther back.

```
  3) MicroBadger
       layers with a specified tag
         
	[ GET /layers/_/httpd/alpine.json ]:
     Error: Request timed out after 5000 ms
      at Timeout.handleRequestTimeout [as _onTimeout] (node_modules/icedfrisby/lib/icedfrisby.js:1098:35)
```

Refs:
- https://circleci.com/gh/badges/daily-tests/29
- https://circleci.com/gh/badges/daily-tests/28
- https://circleci.com/gh/badges/daily-tests/27

Issue #2497, opened 2018-12-10
Route `pattern` doesn't support trailing optional


paulmelnikow:
As mentioned in the [rewriting services doc](https://github.com/badges/shields/blob/master/doc/rewriting-services.md#get-the-route-working):

> At present, the trailing parameter of a pattern can't be optional. If the last part of a route is optional, like a branch, you will need to use a format regex string (e.g. `format: '([^/]+/[^/]+)')`.

Have got an issue open upstream in pillarjs/path-to-regexp#173, though it's not clear how that will be resolved and it's worth having a place to track this here.

Issue #2496, opened 2018-12-10
Better modal popup


paulmelnikow:
I've been working on a major improvement to the modal popup which is basically an epic. I've been referencing it as #701, however #777, #1146, and #2427 are also included.

I'm opening this to track the feature in a way that gives a more transparency to the goalpost and the process.

Here's the behavior I'd like to implement:

1. Show individual fields on the modal popup for each part of the URL path.
2. Same for the query string.
3. Pre-populate the fields with the example.
4. Provide a **Clear** button and a **Reset** button.
5. Add UI elements for the remaining badge options, e.g. logo.
6. Provide a non-editable box (like the **URL** box now) which contains the full URL as it's built.
7. Show the live preview whenever all the fields are filled in.

These are some of the technical milestones:

- [x] Define the routes in a way that enables building a URL from individual fields (#2050)
- [ ] Convert `exampleUrl` to `pattern`, `namedParams`, and `staticPreview` (#1961)
    - [ ] Examples: #2407 #2409 #2444 #2447 #2478
    - [ ] Needed features: #2479
    - [x] Nice to have: #2497
- [x] Extract the service definitions in a way that enables the front end to know the fields (#2397)
- [x] Consume that format in the frontend and drop the old format (#2441)
- [x] Set up dev experience
    - [x] Add basic frontend tests #2490
    - [x] Adopt [styled-components](https://www.styled-components.com/) #2517
- [x] Add fields for the named params and the query params
- [ ] Add UI elements for the remaining badge options

Superceding: #701 #777 #1146 #2427
Also ref: #1311

paulmelnikow:
Let's track the remaining UI elements in #2867 and the remaining static previews in #1961.

PR #2512, opened 2018-12-11
Improved [MicroBadger] tests speed

PyvesB:
Addresses #2489. Selected images that are less "heavy" and seemed to give faster response times from the API. I also aligned the example `previewUrls` to match what was used in the test.


paulmelnikow:
It's not essential, though it'd be great if you could update the `previewUrl`s to `namedParams` to help with #2496.

PyvesB:
No tests have timed out this time round and they execute much faster with the new images, below one second.

PyvesB:
@paulmelnikow can these examples really be converted at this point? As the tag is optional, wouldn't we be affected by #2497?

paulmelnikow:
In examples I haven't been using optionals, so #2497 doesn't apply. Instead there's usually a different pattern for each example, one with and without the optional part. That way each example renders the way you'd expect:

screen shot 2018-12-11 at 5 08 48 pm

PR #2517, opened 2018-12-12
Rework styling using styled-components

paulmelnikow:
The CSS in the project is relatively difficult to change. While it is very DRY, it relies heavily on inheritance. It's difficult to make changes in the markup modal without it also affecting styles elsewhere.
 
[styled-components](https://www.styled-components.com/) is one of the leading CSS-in-JS libraries. By reducing dependency on global state and CSS inheritance, styles become explicit and are easier to inspect and change. It's also convenient that styles can be embedded with the components they modify.

At runtime, the library creates CSS classes, so it's pretty efficient.

We were using a little bit of [styled-jsx](https://github.com/zeit/styled-jsx) before, which ships with Next.js, though styled-components is more widely used and I've had good experiences with it all around.

In a few cases I've duplicated styles where it feels more natural to do that: for example, `text-align: center` is duplicated in `Main` and `MarkupModal`.

Much of this is a refactor, though there are a few visual changes, particularly in the markup modal and the style examples:

screen shot 2018-12-12 at 4 46 15 pm

screen shot 2018-12-12 at 4 46 48 pm

Blocking further work on #2496.


paulmelnikow:
Thanks for digging in. Some of this looks different than what I'm seeing. Which browser are you using?

> Truncating this text is unhelpful - it means we can't copy & paste it properly:

The truncation happens in CSS so copy and paste should work…

frontend/components/usage.js
```diff
@@ -42,98 +110,109 @@ export default class Usage extends React.PureComponent {
   renderStyleExamples() {
     const { baseUrl } = this.props
     return (
-      <table className="badge-img">
+      <QueryParamTable>
```
chris48s: Do you think this one would also look better with `textAlign = 'left'` ?

chris48s:
> The truncation happens in CSS so copy and paste should work…

Yep. With you.. I still think this one is a bit of an odd design decision.

Now that we're moving away from showing the full URL in the edit box (and in a future PR we'll move away from having an edit box at all), it would be useful to show the full URL of the badge we've generated somewhere in un-tuncated form. i.e: it would be nice to be able to see all of the thing I'm generating without having to copy and paste it into another window.

Also,

This content is really wide, but we're not truncating it:

screenshot at 2018-12-17 20-39-03

This content is really narrow, but we are truncating it:
screenshot at 2018-12-17 20-39-25


If we want to keep the width constrained, perhaps it would be better to wrap it, or allow it to be expanded somehow?

paulmelnikow:
I set the markup back to its original width, and turned off `truncate` for the built URL.

chris48s:
There is one more strange bit of alignment I've noticed:

screenshot at 2018-12-18 19-34-31

screenshot at 2018-12-18 19-32-52

paulmelnikow:
Huh, strange. None of them are supposed to be over there. Probably the width change caused that.

paulmelnikow:
Huh, I can't reproduce that.

screen shot 2018-12-18 at 3 55 14 pm

chris48s:
How strange. I've just taken these 2 screenshots on the staging deploy at https://shields-staging-pr-2517.herokuapp.com

Firefox:
screenshot at 2018-12-18 21-06-16


Chromium Stable:
screenshot at 2018-12-18 21-06-52

Is the difference that `<WeeSnippet truncate>` has some left padding that `<WeeSnippet>` doesn't?

paulmelnikow:
I think it only changes the font size. I've done some monkeying around with the `display` property (inline vs block vs inline-block), though it's hard to know exactly what's causing this.

Does disabling the cache help? If you run locally can you isolate what's different?

paulmelnikow:
Ahhh actually might it be a window-size issue?

Issue #2340, opened 2018-11-17
Redirect aliased badges to the canonical form


paulmelnikow:
In #2333 I replaced an old route with a redirect.

As I wrote there:

> In the case of URLs which have permanently changed, an approach I’d like to try is issuing a 301 Redirect.
> 
> The benefit is that if a user pastes the URL into the address bar while they are previewing or editing it, the browser will replace the address with the corrected URL when it loads. I figure this will cause some people to update their URLs with no effort, simply because they previewed the badge in their browser, and others to change over, if they notice it.
> 
> We incur a slight cost, which is a second request. However many browsers cache the 301’s indefinitely, and we can set an effectively infinite cache duration so the CDN and most other downstream caches will keep them a long time. And handling the redirect is extremely cheap.
> 
> This is a nice way to preserve backward compatibility of old routes without having to complicate the new route, such as in the case of vso -> azure-devops. For maintenance purposes, the route that redirects can effectively be treated separately.
> 
> It’s also a nice, gentle, and confidence-inspiring way to signal that users should update their URLs.

Some possibilities:

- [ ] vso
- [ ] Wordpress stars
- [ ] snap-ci
- [ ] codacy?

Others?

Issue #1455, opened 2018-01-16
Tests don't exit on Mac OS


paulmelnikow:
Starting in 0da212b7deb95c0239e8c24899c846a9a2d9c038 the test process doesn't exit when tests finish. This is happening on my machine, though not in CI which is curious. Maybe it's OS-related?

The cause is that something is continuing to run in the node process, like a `setTimeout`. It might be related to the camp update, though probably more likely, one of the many other libraries that got updated in that change.

Steps to reproduce: run `npm run test:js:server` (or the service tests). Observed behavior is that the process stays running after the tests finish. Expected behavior is that it exits.

This could be worked around with a flag to Mocha, though I'd rather track down the cause of the problem.

PR #1574, opened 2018-03-15
Do not rate limit in dev

paulmelnikow:
I ran into a rate limit while running the service tests, which seems not ideal!

I wonder if it would be better to set RATE_LIMIT=true on the server and default to false.

PR #2519, opened 2018-12-12
Overhaul initialization pattern for server + server tests [githubpullrequestcheckstate githublicense pypi]

paulmelnikow:
Because `server.js` was long a monolith, there are a bunch of shims in place to facilitate unit testing. A few of the test suites share port 1111 which means if one of them fails to set up, the port won't be freed and other unrelated tests will fail. Some of the tests which trigger server setup include timeouts which were added to give setup code time to run. In one the test suites, we actually modify `process.argv`, which seems completely gross.

This implements a few changes which improve this:

1. Separate the server from the server startup script, splitting out `lib/server.js`.
2. Inject config into the server and validate the config schema.
3. Inject config into the service test runner.
4. Use `portfinder`, a popular utility for grabbing open ports during testing.
5. Switch more of the setup code from callbacks to async-await.

Overall it leaves everything acting more reliably and looking rather cleaner, if in a few places more verbose.

It also fixes the root cause of #1455, a `setTimeout` in `rate-limit`. Off and on during development of this changeset, Mocha would decide not to exit, and that turned out to be the culprit.

In the interest of keeping this a bit smaller, this defers some work:

- Handling config defaults using Joi.
- Wrapping the monitor (`lib/sys/monitor.js`) in a class.

Marking blocker because it rewrites code that would be affected by #2340.


lib/in-process-server-test-helpers.js
```diff
@@ -1,69 +1,27 @@
...
- */
-function start() {
-  if (startCalled) {
-    throw Error(
-      'Because of the way Shields works, you can only use this ' +
-        'once per node process. Once you call stop(), the game is over.'
-    )
+const Server = require('./server')
+
+function createTestServer({ port }) {
```
paulmelnikow: I deferred handling defaults using Joi, though when that is implemented this can be done more cleanly, and likely the helper function can be removed.

lib/request-handler.spec.js
```diff
@@ -29,9 +27,15 @@ function fakeHandler(queryParams, match, sendBadge, request) {
 describe('The request handler', function() {
   before(analytics.load)
 
+  let port, baseUrl
```
paulmelnikow: Choose a port dynamically to keep this suite independent from other suites.

lib/server-config.js
```diff
@@ -27,11 +27,7 @@ const baseUri = url.format({
 
 // The base URI provides a suitable value for development. Production should
 // configure this.
-const allowedOrigin = envArray(
-  process.env.ALLOWED_ORIGIN,
-  baseUri.replace(/\/$/, ''),
-  ','
-)
+const allowedOrigin = envArray(process.env.ALLOWED_ORIGIN, [address], ',')
```
paulmelnikow: As I read this change, I'm sorta questioning it. I think the bug this is trying to fix may be elsewhere.

lib/server.js
```diff
@@ -0,0 +1,222 @@
...
+const sysMonitor = require('./sys/monitor')
+const PrometheusMetrics = require('./sys/prometheus-metrics')
+const { makeSend } = require('./result-sender')
+const { handleRequest, clearRequestCache } = require('./request-handler')
+const { clearRegularUpdateCache } = require('./regular-update')
+
+const optionalUrl = Joi.string().uri({ scheme: ['http', 'https'] })
+const requiredUrl = optionalUrl.required()
+
+const configSchema = Joi.object({
```
paulmelnikow: New config schema based on `server-config.js`.

lib/server.js
```diff
@@ -0,0 +1,222 @@
...
+    makeBadge: Joi.boolean().required(),
+  },
+  cacheHeaders: {
+    defaultCacheLengthSeconds: Joi.number().integer(),
+  },
+  rateLimit: Joi.boolean().required(),
+  handleInternalErrors: Joi.boolean().required(),
+}).required()
+
+module.exports = class Server {
```
paulmelnikow: Basically this is a rewrite of `server.js` as an object which as much as possible encapsulates its own state.

lib/server.spec.js
```diff
@@ -1,22 +1,27 @@
...
+const portfinder = require('portfinder')
+const svg2img = require('../gh-badges/lib/svg-to-img')
+const { loadServiceClasses } = require('../services')
+const { createTestServer } = require('./in-process-server-test-helpers')
+const { handleRequest } = require('./request-handler')
+const serverConfig = require('./server-config')
 
 describe('The server', function() {
-  let dummyCamp = Camp.start({ port: config.port, hostname: '::' })
+  let dummyCamp
```
paulmelnikow: Use dynamic ports.

lib/service-test-runner/cli.js
```diff
@@ -52,26 +52,39 @@
...
-  this.timeout(5000)
-  server = serverHelpers.start()
-})
-after('Shut down the server', async function() {
-  await serverHelpers.stop(server)
-})
+let baseUrl, server
+if (process.env.TESTED_SERVER_URL) {
+  baseUrl = process.env.TESTED_SERVER_URL
+} else {
```
paulmelnikow: When `TESTED_SERVER_URL` is set, the server should not be run in process.

lib/sys/monitor.js
```diff
@@ -17,7 +16,7 @@ function secretInvalid(req, res) {
   return false
 }
 
-function setRoutes(server) {
+function setRoutes({ rateLimit }, server) {
```
paulmelnikow: Inject config, eliminating one of the few places `server-config` is still imported outside of the server start script.

lib/sys/monitor.js
```diff
@@ -86,6 +85,12 @@ function setRoutes(server) {
       referer: refererRateLimit.toJSON(),
     })
   })
+
+  return function() {
```
paulmelnikow: Calling this function during cleanup fixes #2340.

lib/sys/prometheus-metrics.spec.js
```diff
@@ -2,32 +2,31 @@
 
 const { expect } = require('chai')
 const Camp = require('camp')
+const portfinder = require('portfinder')
 const fetch = require('node-fetch')
-const config = require('../test-config')
 const Metrics = require('./prometheus-metrics')
 
 describe('Prometheus metrics route', function() {
-  const baseUrl = `http://127.0.0.1:${config.port}`
```
paulmelnikow: Dynamic port.

lib/sys/rate-limit.js
```diff
@@ -11,7 +11,12 @@ module.exports = class RateLimit {
     this.banned = new Set()
     this.bannedUrls = new Set()
     this.whitelist = options.whitelist || /(?!)/ // Matches nothing by default.
-    setInterval(this.resetHits.bind(this), this.period * 1000)
+    this.interval = setInterval(this.resetHits.bind(this), this.period * 1000)
+  }
+
+  stop() {
+    clearInterval(this.interval)
+    this.interval = undefined
```
paulmelnikow: The other half of the fix for #2340.

services/github/auth/acceptor.spec.js
```diff
@@ -25,15 +24,21 @@ describe('Github token acceptor', function() {
     delete serverSecrets.shieldsIps
   })
 
+  let port, baseUrl
```
paulmelnikow: Dynamic port.

services/github/auth/admin.spec.js
```diff
@@ -34,16 +34,20 @@ describe('GitHub admin route', function() {
     sandbox.restore()
   })
 
-  const baseUrl = `http://127.0.0.1:${config.port}`
```
paulmelnikow: Dynamic port.

services/service-tester.js
```diff
@@ -92,14 +90,16 @@ class ServiceTester {
   /**
    * Register the tests with Mocha.
    */
-  toss() {
-    const specs = this.specs
+  toss({ baseUrl, skipIntercepted }) {
```
paulmelnikow: Inject config. This happens in `toss()` because that gets invoked by the tester, whereas the constructor is invoked directly by the tester modules.

calebcartwright:
I'll take a look at this, although it will probably take me longer than most as I'm very unfamiliar with this part of the codebase

paulmelnikow:
If you have any questions about it, feel free to hit me up here or on Discord!

lib/server.js
```diff
@@ -0,0 +1,234 @@
...
+
+    await new Promise(resolve => camp.on('listening', () => resolve()))
+  }
+
+  static resetGlobalState() {
+    clearRequestCache()
+    clearRegularUpdateCache()
+  }
+
+  reset() {
```
chris48s: Why do we need both `resetGlobalState()` and `reset()` here?

chris48s:
In general this is a big improvement in terms of structure and readability. Having played with it locally, there are some badges which seem to have been broken by this.

All (or most of) the Github badges are throwing:

```
Vendor hook error: TypeError: Cannot read property 'request' of undefined [server] at Object.camp.route.cache [as handler]
```

Dynamic XML and Yaml badges are both throwing:

```
Uncaught TypeError: renderDynamicBadge is not a function
```

After noticing this, I tried doing a full service test run but it couldn't complete due to `Exceeded limit 500 requests per 200 seconds`. Not sure if that is new in this PR as I rarely do a full service test run locally, but if its new, presumably this will also affect the daily tests so we'll need to be able to disable the rate limit in that context.

paulmelnikow:
Thank you both for reviewing!

> All (or most of) the Github badges are throwing:
>
> ```
> Vendor hook error: TypeError: Cannot read property 'request' of undefined [server] at Object.camp.route.cache [as handler]
> ```

Fixed.

> Dynamic XML and Yaml badges are both throwing:
>
> ```
> Uncaught TypeError: renderDynamicBadge is not a function
> ```

Yikes! Not related, but fixed in #2578.

> After noticing this, I tried doing a full service test run but it couldn't complete due to `Exceeded limit 500 requests per 200 seconds`. Not sure if that is new in this PR as I rarely do a full service test run locally, but if its new, presumably this will also affect the daily tests so we'll need to be able to disable the rate limit in that context.

This sounds like a recurrence of #1574, though the cause might be different. I suppose we could reset the global rate limit state in `reset()`, which seems correct and should fix the issue.

lib/server.js
paulmelnikow: Over time we should transition this stuff from global state to instance state.

calebcartwright:
Sorry I never got a chance to go through this in detail, but glad to see it merged!

PR #2526, opened 2018-12-13
Set up CircleCI test summary [npmcollaborators]

paulmelnikow:
See https://circleci.com/docs/2.0/collect-test-data/

This provides a list of failed tests in the Circle UI and a summary of which tests have failed. It should make interpreting test results easier

It also provides information about how frequently individual tests fail:

screen shot 2018-12-13 at 12 30 02 pm

These seem _particularly_ helpful in the daily service tests, which will also need a couple tweaks made: https://circleci.com/gh/badges/daily-tests/33

I'm not sure if it's possible to access these totals through the API, though if it is, it would be cool to show how many service tests have failed. That would make it a little more fun to keep on top of them. 😉 

Example failures:

https://circleci.com/gh/badges/shields/29091
screen shot 2018-12-13 at 12 34 34 pm

https://circleci.com/gh/badges/shields/29100
screen shot 2018-12-13 at 12 35 34 pm


calebcartwright:
```I'm not sure if it's possible to access these totals through the API, though if it is, it would be cool to show how many service tests have failed. That would make it a little more fun to keep on top of them. 😉
```
https://circleci.com/docs/api/#test-metadata

I remember looking for this API a while back and found nothing so I think this may be _relatively_ new. Seems like it may now finally be possible to have a Test Results badge for Circle (akin to Appveyor, Jenkins, Azure DevOps, etc.)!

package.json
```diff
@@ -157,6 +157,8 @@
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
     "mocha": "^5.0.0",
+    "mocha-env-reporter": "^3.0.0",
+    "mocha-junit-reporter": "^1.18.0",
```
calebcartwright: Also FWIW, Circle can parse the "xunit" format that is emitted by mocha's [native/bundled xunit reporter](https://mochajs.org/#xunit) which could save having to add another dev dependency

package.json
```diff
@@ -157,6 +157,8 @@
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
     "mocha": "^5.0.0",
+    "mocha-env-reporter": "^3.0.0",
```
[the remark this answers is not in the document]
paulmelnikow: `mocha-env-reporter` should use the default reporter outside of CI, though if there’s some other benefit to showing both I could change it.

package.json
paulmelnikow: Oh, nice! I’ll give that a shot.

package.json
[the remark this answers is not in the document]
calebcartwright: Nice! I remember looking at that a while back but I see now where I may have gotten confused by the list of CI servers in the readme

package.json
paulmelnikow: It looks like with `xunit` there's not a way to configure the output file through the environment, whereas `mocha-junit-reporter` supports this.

paulmelnikow:
I appreciate the reviews, from both of you!

package.json
calebcartwright: Makes sense, not worth fighting xunit here IMO :) My guess is that they will point you to one of the community reporters (like [xunit-file](https://www.npmjs.com/package/xunit-file)) that provide support for env var defined output file, at which point we're right back to what the junit reporter already gives us

paulmelnikow: Heh, possibly. I have a soft spot for Mocha, especially after some recent unpleasant experiences with Jest doing unexpected things with setup. However I do find the way Mocha handles reporters and configuration to be frustrating, fiddly, and to require more effort than it should. I _was_ happy with some upcoming improvements to the config which they mentioned in mochajs/mocha#3612. 🤞

calebcartwright: I'm a big fan of Mocha too. I've only worked with Jest in one project, but I do hear a lot of good things about it. What sort of setup issues have you seen with Jest?

paulmelnikow: I hit these issues:

- facebook/jest#2713 (it won't be fixed in Jasmine so it's waiting for Jest Circus)
- jasmine/jasmine#577
- jasmine/jasmine#529

Issue #2528, opened 2018-12-14
Document badge deprecation


paulmelnikow:
It would be helpful to mention in our contributing guidelines how to handle deprecated badges. Sometimes people come around and helpfully remove them and then [we have to point them in the right direction](https://github.com/badges/shields/pull/2410#issuecomment-442103338). And sometimes people just ask! Since the deprecated badges don't currently appear in the UI, that we handle them specially may be unexpected.

Here's a recent example: #2352.

chris48s:
closed in #2529

Issue #2511, opened 2018-12-11
Investigate failing NSP tests


paulmelnikow:
[NSP shut down](https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting); maybe the API is gone?

```
  5) Node Security Platform
       Format '/nsp/npm/:package/:version.:format' where it doesn't exist
         
	[ GET /npm/some-no-exist/1.0.0.json ]:

      AssertionError: expected { name: 'nsp', value: 'invalid' } to deeply equal { Object (name, value) }
      + expected - actual

       {
         "name": "nsp"
      -  "value": "invalid"
      +  "value": "no known vulnerabilities"
       }
      
      at Object.pathMatch.matchJSON (node_modules/icedfrisby/lib/pathMatch.js:138:38)
      at _expect (node_modules/icedfrisby/lib/icedfrisby.js:583:10)
      at IcedFrisbyNock._invokeExpects (node_modules/icedfrisby/lib/icedfrisby.js:1261:26)
      at start (node_modules/icedfrisby/lib/icedfrisby.js:1244:12)
      at Request.runCallback [as _callback] (node_modules/icedfrisby/lib/icedfrisby.js:1131:16)
      at Request.self.callback (node_modules/request/request.js:185:22)
      at Request.<anonymous> (node_modules/request/request.js:1161:10)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:1083:12)
      at endReadableNT (_stream_readable.js:1064:12)
      at _combinedTickCallback (internal/process/next_tick.js:139:11)
      at process._tickDomainCallback (internal/process/next_tick.js:219:9)

  6) Node Security Platform
       Format '/nsp/npm/:package/:version.:format' without vulnerabilities
         
	[ GET /npm/bronze/1.4.0.json ]:

      AssertionError: expected { name: 'nsp', value: 'invalid' } to deeply equal { Object (name, value) }
      + expected - actual

       {
         "name": "nsp"
      -  "value": "invalid"
      +  "value": "no known vulnerabilities"
       }
      
      at Object.pathMatch.matchJSON (node_modules/icedfrisby/lib/pathMatch.js:138:38)
      at _expect (node_modules/icedfrisby/lib/icedfrisby.js:583:10)
      at IcedFrisbyNock._invokeExpects (node_modules/icedfrisby/lib/icedfrisby.js:1261:26)
      at start (node_modules/icedfrisby/lib/icedfrisby.js:1244:12)
[...]

PR #2529, opened 2018-12-14
Deprecate [nsp] service

calebcartwright:
The nsp service was [shutdown](https://blog.npmjs.org/post/175511531085/the-node-security-platform-service-is-shutting) so deprecating those badges which addresses #2511  

Also adds a doc/tutorial for deprecating badges as suggested in #2528


doc/deprecating-badges.md
```diff
@@ -0,0 +1,116 @@
...
+  cauditor: new Date('2018-02-15'),
+  dotnetstatus: new Date('2018-04-01'),
+  magnumci: new Date('2018-07-08'),
+  bithound: new Date('2018-07-08'),
+  versioneye: new Date('2018-08-20'),
+  issuestats: new Date('2018-09-01'),
+  libscore: new Date('2018-09-22'),
+  imagelayers: new Date('2018-11-18'),
+  nsp: new Date('2018-12-13'),
+}
```
chris48s: Instead of reproducing all this in the docs (it will just drift out-of-sync), lets just link to the file: https://github.com/badges/shields/blob/master/lib/deprecated-services.js

calebcartwright: Sounds good

Issue #2510, opened 2018-12-11
Delegate logos to simple-icons


paulmelnikow:
As discussed at https://github.com/badges/shields/pull/2463#issuecomment-445571253:

> I think we should try to defer all icons to [simple-icons](https://github.com/simple-icons/simple-icons).

I agree, I don't see a real advantage to keeping our own collection of icons.

Tasks:

- [ ] Modify our logo guidelines to say we are no longer accepting new logos
- [ ] Match up our existing logo collection with simple-icons, and get our existing logos matched in there. Maybe in the process, one of us could ask to join as a co-maintainer.
- [ ] Remove our existing logos, leaving aliases in place for if needed for backward compatibility.

chris48s:
I've had a quick pass over our logos dir. These ones are already in simpleicons:

* bitcoin.svg
* circleci.svg
* discord.svg
* docker.svg
* github.svg
* gitlab.svg
* gitter-white.svg
* npm.svg
* paypal.svg
* serverfault.svg
* slack.svg
* stackexchange.svg
* stackoverflow.svg
* superuser.svg
* telegram.svg
* travis.svg
* twitter.svg
* windows.svg

The simpleicons version isn't the _exact_ same logo in all cases, but I'm in favour of just standardizing on the simpleicons version.

These ones don't exist in simpleicons:

* appveyor.svg
* azuredevops.svg
* codeship.svg
* dependabot.svg
* eclipse.svg
* lgtm.svg
* liberapay.svg
* matrix.svg
* postgresql.svg
* scrutinizer.svg
* sourcegraph.svg
* tfs.svg

I'm not sure if every single one of these will definitely meet all of their [guidelines](https://github.com/simple-icons/simple-icons/blob/develop/CONTRIBUTING.md#requesting-an-icon), but its worth a bash submitting some PRs and seeing where we get to.

We do also carry a couple of logos relating to deprecated services:

* bithound.svg https://github.com/badges/shields/blob/master/services/bithound/bithound.service.js
* gratipay.svg https://github.com/badges/shields/blob/master/services/gratipay/gratipay.service.js

which we should take the opportunity to remove.

RedSparr0w:
Thanks for that list @chris48s,
I've done a quick mock up showing the differences between each of the logos:

Icon | Shields | Simple-Icons
:-- | :--: | :--:
bitcoin | ![](https://svgshare.com/i/9vB.svg) | ![](https://svgshare.com/i/9wJ.svg)
[...]

PR #2539, opened 2018-12-16
remove logos for deprecated services

chris48s:
refs #2510


paulmelnikow:
Thanks for taking care of this.

Issue #2518, opened 2018-12-12
Extract fetchYaml and maybe other fetch functions


paulmelnikow:
See discussions:

- https://github.com/badges/shields/pull/2487#discussion_r241169902
- https://github.com/badges/shields/pull/2399#issuecomment-444970457

paulmelnikow:
Should we keep this open?

The thing we could still do is delegate from `_requestJson`, `_requestYaml`, etc. to functions which can be invoked on any instance of `BaseService`.

chris48s:
Aah yes. I was unsure on this. My initial thought was we could extract them to stand-alone functions. The reason I didn't do this was because inside the `_requestFormat()` functions we're calling `this._request()` and `this.constructor._validate()` so `_requestFormat()` doesn't extract so nicely outside of the class.

We could do it so that we pass in a `requester` function and a `validator` function as params, or so that we can pass an instance of `BaseService` as a param. Another option might be to try and do it as a [mixin](https://javascript.info/mixins). I'm not really sure which of those makes the most/least sense.

Given the F-Droid badge could just extend `BaseYamlService` I decided to just leave it and deal with it once we've got an example where we actually need this abstraction.

paulmelnikow:
> pass an instance of `BaseService` as a param

That's the approach I've taken before:

https://github.com/badges/shields/blob/b4b29ab0f787069111db4bf332810493adf363c5/services/azure-devops/azure-devops-helpers.js#L18

paulmelnikow:
Gonna close this for now, as it is not seeming pressing.

PR #2540, opened 2018-12-16
[dynamicyaml, f-droid] add yaml base service class

chris48s:
refs  #2518


services/f-droid/f-droid.service.js
```diff
@@ -45,28 +44,17 @@ module.exports = class FDroid extends BaseService {
   }
 
   async fetchYaml(url, options) {
-    const { buffer } = await this._request({
+    const schema = Joi.object({
```
paulmelnikow: Could you move the schema up top so it's not recompiled on each request?

PR #2541, opened 2018-12-16
Migrate [Jira] services to new service model

calebcartwright:
Migrates the Jira badge services to the new service model, and also adds tests for those badges. 

I did make one minor changes to the badge labels/messages when the issue or sprint cannot be found. For example: 

With the LegacyService implementation, when an issue was not found the badge would display as 'issue key | inaccessible' such as `KAFKA-2869 | inaccessible` but this changes that behavior to `jira | issue not found`. I personally find this to be a bit more informative of a message, but want to call it out as I know how important backwards compatibility is. However, at first glance I'm also not sure how we'd replicate the old behavior of using the issue key (provided by user as a param) as the badge label when an issue is not found because the API response has a status code of 404 and a different schema than what Joi is expecting/validating. 

Similarly for the sprints, the old format for a sprint not found was `completion | inaccessible` it's now `jira | sprint not found`


calebcartwright:
Also is there a pattern when writing service tests to cover lines that are using auth info from `server-secrets`?

services/jira/jira-issue.tester.js
```diff
@@ -0,0 +1,71 @@
...
+        fields: {
+          status: {
+            name: 'done',
+          },
+        },
+      })
+  )
+  .expectJSON({ name: 'bar-345', value: 'done' })
+
+t.create('endpoint with port and no path')
```
calebcartwright: let me know if some of these tests are unnecessarily duplicative. for whatever reason I was struggling with the pattern/regex so i added different tests to cover the pattern variants for sanity checks

paulmelnikow:
> Also is there a pattern when writing service tests to cover lines that are using auth info from `server-secrets`?

No… though it would be possible!

You could use an IcedFrisby `before()` hook in a service test, like the one @PyvesB just added here:

https://github.com/badges/shields/blob/b4b29ab0f787069111db4bf332810493adf363c5/services/wheelmap/wheelmap.tester.js#L18

The cleanup would go in an IcedFrisby `.finally()` hook. Probably most reliable using `sinon`.

You could generalize the code here into a `mockShieldsSecrets()` function that creates and returns a `sinon` sandbox.

https://github.com/badges/shields/blob/b4b29ab0f787069111db4bf332810493adf363c5/services/github/auth/admin.spec.js#L24-L35

services/jira/jira-base.js
```diff
@@ -0,0 +1,25 @@
...
+
+const BaseJsonService = require('../base-json')
+const serverSecrets = require('../../lib/server-secrets')
+
+module.exports = class BaseJiraService extends BaseJsonService {
+  static get category() {
+    return 'issue-tracking'
+  }
+
+  async fetch({ url, options = {}, schema, errorMessages }) {
```
paulmelnikow: It might be better to accept `qs` here rather than the full `options` object. It conveys the intent better.

services/jira/jira-issue.service.js
```diff
@@ -1,100 +1,92 @@
...
+  }
+
+  static get defaultBadgeData() {
+    return { color: 'lightgrey', label: 'jira' }
   }
 
   static get route() {
     return {
       base: 'jira/issue',
+      pattern: ':protocol(https?)/:host/:path?/:issueKey',
```
paulmelnikow: - The regex in this pattern confused me for a moment. `:protocol(http|https)` seems clearer.
- I think you need `:path*` to allow multiple path segments.

services/jira/jira-issue.service.js
```diff
@@ -1,100 +1,92 @@
...
     }
   }
 
   static get examples() {
     return [
       {
         title: 'JIRA issue',
-        pattern: ':protocol/:hostAndPath+/:issueKey',
+        pattern: ':protocol/:host/:issueKey',
```
paulmelnikow: Since `host` and `path` are broken out above, probably it makes sense to do the same here.

services/jira/jira-issue.service.js
```diff
@@ -1,100 +1,92 @@
...
+      schema,
+      options: {},
+      errorMessages: {
+        404: 'issue not found',
+      },
+    })
+    const issueStatus = json.fields.status
+    const statusName = issueStatus.name
+    let color = 'lightgrey'
+    if (issueStatus.statusCategory) {
```
paulmelnikow: Could you move the color-selection logic into `render()`?

services/jira/jira-issue.tester.js
paulmelnikow: They look good to me! It's possible we could solve them with a smaller bracket unit test, though this is more than fine. 😄

services/jira/jira-sprint.service.js
```diff
@@ -33,74 +56,57 @@ module.exports = class JiraSprint extends LegacyService {
...
-            }
-            sendBadge(format, badgeData)
-          } catch (e) {
-            badgeData.text[1] = 'invalid'
-            sendBadge(format, badgeData)
-          }
-        })
-      })
-    )
+    const percentComplete = numTotalIssues
```
paulmelnikow: From this point down you've moved on to rendering. Could you move this into `render()`?

services/jira/jira-base.js
calebcartwright: Will do. I can't see any use cases when a service class would need to define anything other than `qs` anyway

services/jira/jira-issue.service.js
calebcartwright: 👍 on protocol

`path` is where I kept hitting a a mental wall with the pattern. I was trying to use nested non-capture groups but pattern was having none of it 😆 Let me see if I can clean that up a bit

services/jira/jira-sprint.service.js
```diff
@@ -33,74 +56,57 @@ module.exports = class JiraSprint extends LegacyService {
...
+    if (path) {
+      url += `/${path}`
+    }
+    // Atlassian Documentation: https://developer.atlassian.com/cloud/jira/platform/rest/v2/#api-group-Search
+    // There are other sprint-specific APIs but those require authentication. The search API
+    // allows us to get the needed data without being forced to authenticate.
+    url += '/rest/api/2/search'
+    const options = {
+      qs: {
+        jql: `sprint=${sprintId} AND type IN (Bug,Improvement,Story,"Technical task")`,
```
calebcartwright: I added the above comment on `url` to indicate the usage of `search` over `sprint` API, and I totally get why it is implemented that way. However, I could envision some potential issues due the issue type clause that we have to put into the jql query. Jira is used by some teams at my day job, and we've created many custom issue types (like 12387123897612387123 of them 😄)  that wouldn't be accurately reflected by this query.

IMO we shouldn't try to solve that in this PR, but it may be worth revisiting at some point

calebcartwright:
Thanks for the feedback! I think I've addressed everything, and also added a test helpers file to help facilitate the setup/restore to validate auth scenarios

services/jira/jira-issue.service.js
paulmelnikow: I'm fine with the change you made, though I think `hostAndPath` would better describe the field for the user (this is with #2496 in mind).

services/jira/jira-test-helpers.js
```diff
@@ -0,0 +1,25 @@
...
+
+function mockJiraCreds() {
+  serverSecrets['jira_username'] = undefined
+  serverSecrets['jira_password'] = undefined
+  sinon.stub(serverSecrets, 'jira_username').value(user)
+  sinon.stub(serverSecrets, 'jira_password').value(pass)
+}
+
+function restore() {
+  sinon.restore()
```
paulmelnikow: It looks like this function is not intended to be used publicly: https://github.com/sinonjs/sinon/issues/1390

services/jira/jira-issue.tester.js
```diff
@@ -0,0 +1,93 @@
...
+  )
+  .expectJSON({ name: 'test-001', value: 'in progress' })
+
+t.create('with auth')
+  .before(jiraTestHelpers.mockJiraCreds)
+  .get('/https/myprivatejira.com/secure-234.json')
+  .intercept(nock =>
+    nock('https://myprivatejira.com/rest/api/2/issue')
+      .get(`/${encodeURIComponent('secure-234')}`)
+      .basicAuth({
```
paulmelnikow: Since this pattern is a bit unfamiliar, you could comment here that by including this here, it ensures the credentials are actually being sent out with the HTTP request. Otherwise the request wouldn't match and the test would fail.

paulmelnikow: Nice work on this test, by the way!

services/jira/jira-test-helpers.js
calebcartwright: Since sinon@5.x the default object is itself a sandbox, and most of the documentation says to use that directly for the majority of use cases, https://sinonjs.org/releases/v7.2.2/sandbox/. 

They show this pattern, including the `restore` function on the docs for migrating to sinon@5.x https://sinonjs.org/guides/migrating-to-5.0.html. 

In that issue I see references to sinon@2.x/4.x so maybe that's changed?

services/jira/jira-issue.service.js
calebcartwright: yeah I like that better too (and that's what was there before so +1 for consistency!)

calebcartwright: i'm probably being overly pedantic here, but just noticed the example pattern before was actually `hostAndPath+` does the `+` suffix have any significance in the example patterns/is it worth including?

paulmelnikow:
btw @calebcartwright I wanted to reach out and see about setting up a time to say hi. Could you shoot a quick message? The address on my profile works.

calebcartwright:
will do!

PR #2548, opened 2018-12-17
Migrate [Codetally] to new service model

calebcartwright:


services/codetally/codetally.tester.js
```diff
@@ -37,3 +32,13 @@ t.create('Empty')
       })
   )
   .expectJSON({ name: 'codetally', value: '$0.00' })
+
+t.create('Non existent')
+  .get('/not/real.json')
+  .intercept(nock =>
```
paulmelnikow: Can you do a live test for this case? That way we'll know it's still working if something changes on their end. Also even though it's a 503, it's probably better to show a custom error message.

calebcartwright: Yup, good idea! I was debating whether to add a custom error message for `503`. It feels a little odd to me that repo not found is giving 500 errors 🤔 I'll probably add a `repo not found` error on 404 and 503 to cover our bases

paulmelnikow: Sounds fine to me. Yea, it's odd. The formalist in me is like 😏and the pragmatist is like 🤷‍♂️.

PR #2549, opened 2018-12-17
Refactor [AzureDevOpsCoverage AzureDevOpsTests] to only report on completed builds; test all [azuredevops]

DeanPDX:
This is in response to [this conversation](https://github.com/badges/shields/issues/2237#issuecomment-446812796). The change turned out being even easier than I had anticipated because the [API supports a buildStatus filter](https://docs.microsoft.com/en-us/rest/api/azure/devops/build/builds/list?view=azure-devops-rest-5.0#buildstatus). I did quite a bit of local testing to make sure this is solid. I also ran tests and checked coverage.

One thing I thought might be slightly odd is the [name of this class](https://github.com/badges/shields/blob/master/services/azure-devops/azure-devops-base.js#L18) doesn't match the file name. Not an issue and I'm not sure if it's intentional, but figured it was worth noting.

### DIFF
--- a/services/azure-devops/azure-devops-base.js
+++ b/services/azure-devops/azure-devops-base.js
@@ -25,7 +25,7 @@ module.exports = class BaseAzureDevOpsService extends BaseJsonService {
     })
   }
 
-  async getLatestBuildId(
+  async getLatestCompletedBuildId(
     organization,
     project,
     definitionId,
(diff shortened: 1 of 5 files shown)


calebcartwright:
> One thing I thought might be slightly odd is the [name of this class](https://github.com/badges/shields/blob/master/services/azure-devops/azure-devops-base.js#L18) doesn't match the file name. Not an issue and I'm not sure if it's intentional, but figured it was worth noting.

Good catch, feel free to change that to `AzureDevOpsBase`

DeanPDX:
OK - I'll update the base class name real quick. I wasn't sure if I should update the name of the class or the filename because it seemed as if `BaseAzureDevOpsService` was attempting to be consistent with `BaseJsonService`, etc.

paulmelnikow:
Thanks so much!

PR #585, opened 2015-12-02
Consistently use "passing".

larsbrinkhoff:
Consistently use "passing" (previously 4 vendors) instead of "passed" (4), "success" (1), or "shippable" (1).  I arbitrarily picked "passing", but I'm not married to that particular choice.

Affected vendors are: Shippable, Wercker (also V3), TeamCity, Codeship, Scrutinizer, and Snap CI.

Fixes part of #575.

PR #2550, opened 2018-12-18
Migrate [Coverity] to new service model

calebcartwright:
* Migrate Coverity Scan to new service model (and adds tests) refs #1358 
* Deprecate Coverity On Demand badges (and adds tests)

Everything I've seen indicates that Coverity On Demand has been decommissioned for a while, although I haven't seen any _official_ announcements from Coverity/Snyopsys. If anyone knows otherwise please let me know, but for now I believe deprecation is the best course of action.

All of the on-demand endpoints are down/inaccessible:
* https://ondemand.coverity.com/ 
* http://ondemand.coverity.com/
* https://api.ondemand.coverity.com
* https://api.ondemand.coverity.com/streams/44b25sjc9l3ntc2ngfi29tngro/badge
* https://api.ondemand.coverity.com/jobs/p4tmm8031t4i971r0im4s7lckk/badge 

And the few online references I have found for Coverity On Demand speak about it in the past tense, like this from the Coverity Wikipedia page:
> Coverity Code Advisor on Demand **_was_** a cloud hosted version of Coverity Code Advisor.

### DIFF
--- a/services/coverity/coverity-scan.service.js
+++ b/services/coverity/coverity-scan.service.js
@@ -1,16 +1,49 @@
 'use strict'
 
-const LegacyService = require('../legacy-service')
-const { makeBadgeData: getBadgeData } = require('../../lib/badge-data')
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
 
-module.exports = class CoverityScan extends LegacyService {
+const messageRegex = /passed|passed .* new defects|pending|failed/
+const schema = Joi.object({
+  message: Joi.string()
+    .regex(messageRegex)
+    .required(),
+}).required()
+
+module.exports = class CoverityScan extends BaseJsonService {
+  static render({ message }) {
+    let color
+    if (message === 'passed') {
+      color = 'brightgreen'
+      message = 'passing'
+    } else if (/^passed .* new defects$/.test(message)) {
+      color = 'yellow'
+    } else if (message === 'pending') {
+      color = 'orange'
+    } else {
+      color = 'red'
+    }
+
+    return {
+      message,
+      color,
+    }
+  }
+
   static get category() {
     return 'quality'
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'coverity',
+    }
   }
 
   static get route() {
     return {
       base: 'coverity/scan',
+      pattern: ':projectId',
     }
   }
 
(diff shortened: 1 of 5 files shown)


services/coverity/coverity-scan.service.js
```diff
@@ -1,63 +1,79 @@
...
         title: 'Coverity Scan',
-        previewUrl: '3997',
+        pattern: ':projectId',
+        namedParams: {
+          projectId: '3997',
+        },
+        staticPreview: this.render({
+          message: 'passed',
+        }),
+        keywords: ['coverity', 'scan'],
```
chris48s: we don't need to specify either of these as keywords because they both appear in the title

services/coverity/coverity-scan.tester.js
```diff
@@ -0,0 +1,71 @@
...
+t.create('passed')
+  .get('/2.json')
+  .intercept(nock =>
+    nock('https://scan.coverity.com/projects/2')
+      .get('/badge.json')
+      .reply(200, {
+        message: 'passed',
+      })
+  )
+  .expectJSON({ name: 'coverity', value: 'passing' })
```
chris48s: If you want to mock a response for each case, we could usefully make the calls with the `?style=_shields_test` param here and test the colouring logic too. There's an example of this in the docs https://github.com/badges/shields/blob/master/doc/service-tests.md#5-mocking-responses

services/coverity/coverity-scan.tester.js
```diff
@@ -0,0 +1,71 @@
...
+
+const Joi = require('joi')
+const t = (module.exports = require('../create-service-tester')())
+
+t.create('live: known project id')
+  .get('/3997.json')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'coverity',
+      value: Joi.string().regex(/passing|passed .* new defects|pending|failed/),
```
chris48s: Lets move `messageRegex` out into a helper module so we can use the same regex in the test here and the validation schema.

services/coverity/coverity-scan.service.js
calebcartwright: 👍 I didn't realize that, good to know!

services/coverity/coverity-scan.tester.js
calebcartwright: Excellent! I was wondering about color validation. Will add this (I should probably update some other service tests to include this as well)

services/coverity/coverity-scan.tester.js
calebcartwright: I almost did that but there is a _very_ subtle difference in the regex. Coverity returns `passed` (and variances of passed), but the LegacyService implementation was converting that to `passing`. I'm not sure why `passed` was converted to `passing` but the other status (`passed 5 new defects`) kept `passed` in the message.

services/coverity/coverity-scan.tester.js
paulmelnikow: Another approach which can be taken here is a Sazerac-based test of `render` in a `.spec.js` file.

calebcartwright: Is there a preferred approach or is either good?

services/coverity/coverity-scan.tester.js
chris48s: Well spotted. Looks like that got done here: https://github.com/badges/shields/pull/585

I don't have strong opinions on it. Happy for you to either leave it as-is. I only mentioned it because I didn't notice that the 2 regexes aren't exactly the same.

Issue #2320, opened 2018-11-15
Rewrite GitHub services


paulmelnikow:
In #2253 I got the ball rolling with rewriting the GitHub services. The changes to the service in that PR are a good model for rewriting any of the remaining services.

You can ignore the changes to the base services and to the GithubAuthService.

~~There have been some small enhancements to BaseService since then which should also be taken into account: #2278 #2279 (and #2308 once it's merged)~~

If folks are comfortable taking any of the remaining `LegacyService`s in `services/github/` and rewriting them along the lines of #2253, please give it a go!

~~At the moment there's also a pull request open (#2309) to move the rest of the GitHub badge examples into `services/github/`, though that hopefully will be merged soon.~~

Ref #1358

paulmelnikow:
If anyone is interested in helping with this, we've now got some tips!

https://github.com/badges/shields/blob/master/doc/rewriting-services.md

paulmelnikow:
Seems like we may as well close this out in favor of #2863. Is there any more info we should add there related to the GitHub services? Maybe pointers to some which are done?

calebcartwright:
Once I split out the GH services in our spreadsheet, I can add the remaining list of GH services needing refactor in here for reference, and links to the classes that have already been refactored. 

Given the number of GH services and some of the complexity I think it may also make sense to add a pointer over to here from #2863 

Thoughts?

calebcartwright:
Sure, will do

calebcartwright:
I've finished splitting out the GH services in the spreadsheet. On second thought I don't want to add them inline here as it will inevitably slide out of date, but I'll pick out one to refactor and update this accordingly

paulmelnikow:
These have gotten easier!

#3095 #3098 #3099

paulmelnikow:
Just three of these left! Let's track at https://github.com/badges/shields/issues/2863#issuecomment-482874488.

PR #2560, opened 2018-12-19
Split up [github] testers

paulmelnikow:
- Update github contributors badge for `create-service-tester`.

Conflicts nominally with #2470. Let's merge that ASAP and this right afterward so as to unblock the rest of #2320.


services/github/github-contributors.service.js
```diff
@@ -42,7 +42,6 @@ module.exports = class GithubContributors extends LegacyService {
           message: '397',
           color: 'blue',
         },
-        keywords: ['GitHub', 'contributor'],
```
calebcartwright: More of a question as I'm not familiar with search, but does it do partial word matches? I.e. will `contributor` still be matched in search since the title contains `contributors`

paulmelnikow: Yea. I think it matches on leading substrings:

screen shot 2018-12-19 at 4 51 45 pm

calebcartwright: Nice!

services/github/github-search.tester.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+const Joi = require('joi')
+const { isMetric } = require('../test-validators')
+
+const t = (module.exports = require('../create-service-tester')())
+
+t.create('hit counter')
+  .get('/torvalds/linux/goto.json')
```
calebcartwright: This isn't related to the refactored tests in this PR but the GH Search hit counter badge isn't rendering for me in the staging deploy env.  You mentioned in https://github.com/badges/shields/issues/1358#issuecomment-448743020 that perhaps it is just that particular repo that is slow, so would it be worth using a different repo for this test?

paulmelnikow: Yea, I think it would. Perhaps we should also update the example, which is what renders in the staging deploy.

One of the open issues PRs also tends to time out. I think it's the undefined label test. We could try it on a smaller repo or give it a longer timeout.

I try to avoid changing things while they're being moved because it's hard to compare what's changed, so let's handle this in a follow-on.

PR #2563, opened 2018-12-19
Fix frontend coverage and tweak ignores

paulmelnikow:
Since we use Babel for the frontend but not the server, I split the coverage config into two separate files, one for the frontend and the other for the server.

Refs:

- https://github.com/istanbuljs/nyc#nyc
- https://github.com/istanbuljs/babel-plugin-istanbul#mocha-on-nodejs-through-nyc

Close #2552

PR #2569, opened 2018-12-20
Fix frontend coverage in CI, update dev scripts, adopt npm-run-all

paulmelnikow:
Ref https://github.com/badges/shields/pull/2563#issuecomment-449079252


package.json
```diff
@@ -63,7 +63,7 @@
     "coverage:test:package": "nyc node_modules/mocha/bin/_mocha --opts mocha.opts \"gh-badges/**/*.spec.js\"",
     "coverage:test:integration": "nyc node_modules/mocha/bin/_mocha --opts mocha.opts \"lib/**/*.integration.js\" \"services/**/*.integration.js\"",
     "coverage:test:services": "nyc node_modules/mocha/bin/_mocha --opts mocha.opts --delay lib/service-test-runner/cli.js",
-    "coverage:test": "rimraf .nyc_output coverage; npm run coverage:test:server; npm run coverage:test:package; npm run coverage:test:frontend; npm run coverage:test:integration; npm run coverage:test:services",
+    "coverage:test": "rimraf .nyc_output coverage; npm run defs; npm run coverage:test:server; npm run coverage:test:package; npm run coverage:test:frontend; npm run coverage:test:integration; npm run coverage:test:services",
```
calebcartwright: will `npm run defs` be needed on the `test:js:frontend` script too or is that something coverage specific? 

Probably out of scope here but any thoughts/opinions on utilizing the `pre`/`post` npm scripts (like `precoverage:test`) as well as chaining scripts together to avoid some of the repetition? 

For example if we did something like:
```json
  "coverage:test:frontend": "BABEL_ENV=test nyc --nycrc-path .nycrc-frontend.json npm run test:js:frontend"
```

paulmelnikow: Yea, it is needed. It does get run by `test`. I'm open to revisiting that, though there's a tradeoff in running it which is that running extra stuff slows down the tests between iterations.

I realized I forgot `features`, which is the task we actually need here, along with `defs`. 😝 

Does that chaining work with `nyc`? If it does, that would be _excellent_. Maintaining these test invocations can be a bit exhausting.

calebcartwright: > Does that chaining work with `nyc`? If it does, that would be _excellent_. Maintaining these test invocations can be a bit exhausting.

It's always worked for me (example [here](https://github.com/swellaby/vsts-bump/blob/master/package.json#L25)) so it _should_ work here too (knock on wood!)

paulmelnikow: Ah, nice. I like how you've moved things into multiple mocha.opts files. We should look at that after they finish overhauling their config.

paulmelnikow:
Feel free to pick this up if you'd like! If not I'll cycle back to it probably tomorrow.

calebcartwright:
I'm fighting a refactor of the SensioLabs badge right now but I'll take a look if I win that battle 😁

paulmelnikow:
I made an update using the approach from https://github.com/badges/shields/pull/2569/files#r243427382.

I also made a few other changes:

1. Adopt `npm-run-all` which provides cleaner output (easy to access via `--silent`) and terser syntax (`run-s`).
2. Drop the `js` from `test:js:server` and the other tasks with that pattern. (It's extra typing for little benefit, and I forget it's there… especially when I've been away from the project for a while.) Alias the old names.
3. Use `pre` scripts to avoid some chaining.
4. Change up the coverage scripts, so `coverage:report` now builds and opens the report.
5. Pull the Heroku pre-start + start tasks into a single entry point to match the `now-start` script.

package.json
```diff
@@ -76,7 +76,7 @@
     "test:js:frontend": "echo \"Deprecated; run `npm run test:frontend` instead.\" && npm run test:frontend",
     "test:js:package": "echo \"Deprecated; run `npm run test:package` instead.\" && npm run test:package",
     "test:frontend": "BABEL_ENV=test mocha --opts mocha.opts --require @babel/polyfill --require @babel/register --require mocha-yaml-loader \"frontend/**/*.spec.js\"",
-    "test:server": "HANDLE_INTERNAL_ERRORS=false mocha --opts mocha.opts \"*.spec.js\" \"lib/**/*.spec.js\" \"services/**/*.spec.js\"",
```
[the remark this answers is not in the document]
paulmelnikow: Correct. There are no longer any specs in the root of the project.

PR #2486, opened 2018-12-08
[Wheelmap] service rewrite and tests

PyvesB:
Fixes issue #2469 and benefits from [new functionnality in icedfrisby-nock](https://github.com/paulmelnikow/icedfrisby-nock/pull/25)!

Before we can deploy this, we have to coordinate on adding the API token to the production server, and optionally on Circle CI.

Looking forward to feedback! 👍

PR #2572, opened 2018-12-21
[WIP] Migrate [Sensiolabs] to new service model

calebcartwright:
Migrate Sensiolabs/Symfony Insight to new service model. Refs #1358 

There were no tests for this service before, but this adds tests with 100% coverage (all mocked tests). Live tests for this service are a little tricky as the API **always** requires auth (it's one of the few services that have tokens plugged in https://github.com/badges/shields/blob/master/doc/production-hosting.md#server-secrets) 

I've got some ideas on how I can add some live tests that we can run selectively (like in the daily service job) without impacting other users/services. Let me know if there's already a pattern in use for this type of scenario though.

### DIFF
--- a/services/sensiolabs/sensiolabs.service.js
+++ b/services/sensiolabs/sensiolabs.service.js
@@ -1,23 +1,66 @@
 'use strict'
 
-const LegacyService = require('../legacy-service')
-const { makeBadgeData: getBadgeData } = require('../../lib/badge-data')
+const Joi = require('joi')
+const BaseXmlService = require('../base-xml')
 const serverSecrets = require('../../lib/server-secrets')
+const { statusRegex } = require('./sensiolabs-helpers')
 
-// This legacy service should be rewritten to use e.g. BaseJsonService.
-//
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class Sensiolabs extends LegacyService {
+const schema = Joi.object({
+  project: Joi.object({
+    'last-analysis': Joi.object({
+      status: Joi.string()
+        .regex(statusRegex)
+        .required(),
+      grade: Joi.string()
+        .regex(/platinum|gold|silver|bronze|none/)
+        .allow('', null),
+    }),
+  }).required(),
+}).required()
+
+module.exports = class Sensiolabs extends BaseXmlService {
+  static render({ status, grade }) {
+    if (status !== 'finished') {
+      return {
+        message: 'pending',
+        color: 'grey',
+      }
+    }
+    let color,
+      message = grade
+    if (grade === 'platinum') {
+      color = 'brightgreen'
+    } else if (grade === 'gold') {
+      color = 'yellow'
+    } else if (grade === 'silver') {
+      color = 'lightgrey'
+    } else if (grade === 'bronze') {
+      color = 'orange'
+    } else {
+      message = 'no medal'
+      color = 'red'
+    }
+
+    return {
+      message,
+      color,
+    }
+  }
+
   static get category() {
     return 'build'
   }
 
   static get route() {
     return {
-      base: 'sensiolabs',
+      base: 'sensiolabs/i',
+      pattern: ':projectUuid',
+    }
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'check',
     }
   }
 
@@ -25,86 +68,52 @@ module.exports = class Sensiolabs extends LegacyService {
     return [
       {
         title: 'SensioLabs Insight',
-        previewUrl: 'i/45afb680-d4e6-4e66-93ea-bcfa79eb8a87',
+        pattern: ':projectUuid',
+        namedParams: {
+          projectUuid: '45afb680-d4e6-4e66-93ea-bcfa79eb8a87',
+        },
+        staticPreview: this.render({
+          status: 'finished',
+          grade: 'bronze',
+        }),
       },
     ]
   }
 
-  static registerLegacyRouteHandler({ camp, cache }) {
-    camp.route(
-      /^\/sensiolabs\/i\/([^/]+)\.(svg|png|gif|jpg|json)$/,
-      cache((data, match, sendBadge, request) => {
-        const projectUuid = match[1]
-        const format = match[2]
-        const options = {
-          method: 'GET',
-          uri: `https://insight.sensiolabs.com/api/projects/${projectUuid}`,
-          headers: {
-            Accept: 'application/vnd.com.sensiolabs.insight+xml',
-          },
-        }
+  async fetch({ projectUuid }) {
+    const url = `https://insight.sensiolabs.com/api/projects/${projectUuid}`
+    const options = {
+      headers: {
+        Accept: 'application/vnd.com.sensiolabs.insight+xml',
+      },
+    }
 
-        if (serverSecrets && serverSecrets.sl_insight_userUuid) {
-          options.auth = {
-            user: serverSecrets.sl_insight_userUuid,
-            pass: serverSecrets.sl_insight_apiToken,
-          }
-        }
+    if (serverSecrets && serverSecrets.sl_insight_userUuid) {
+      options.auth = {
+        user: serverSecrets.sl_insight_userUuid,
+        pass: serverSecrets.sl_insight_apiToken,
+      }
+    }
 
-        const badgeData = getBadgeData('check', data)
+    return this._requestXml({
+      url,
+      options,
+      schema,
+      errorMessages: {
+        401: 'not authorized to access project',
+        404: 'project not found',
+      },
+    })
+  }
 
-        request(options, (err, res, body) => {
-          if (err != null || res.statusCode !== 200) {
-            badgeData.text[1] = 'inaccessible'
-            sendBadge(format, badgeData)
-            return
-          }
+  transform({ data }) {
+    const lastAnalysis = data.project['last-analysis']
+    return { status: lastAnalysis.status, grade: lastAnalysis.grade }
+  }
 
-          const matchStatus = body.match(
-            /<status><!\[CDATA\[([a-z]+)\]\]><\/status>/im
-          )
-          const matchGrade = body.match(
-            /<grade><!\[CDATA\[([a-z]+)\]\]><\/grade>/im
-          )
-
-          if (matchStatus === null) {
-            badgeData.text[1] = 'inaccessible'
-            sendBadge(format, badgeData)
-            return
-          } else if (matchStatus[1] !== 'finished') {
-            badgeData.text[1] = 'pending'
-            sendBadge(format, badgeData)
-            return
-          } else if (matchGrade === null) {
-            badgeData.text[1] = 'invalid'
-            sendBadge(format, badgeData)
-            return
-          }
-
-          if (matchGrade[1] === 'platinum') {
-            badgeData.text[1] = 'platinum'
-            badgeData.colorscheme = 'brightgreen'
-          } else if (matchGrade[1] === 'gold') {
-            badgeData.text[1] = 'gold'
-            badgeData.colorscheme = 'yellow'
-          } else if (matchGrade[1] === 'silver') {
-            badgeData.text[1] = 'silver'
-            badgeData.colorscheme = 'lightgrey'
-          } else if (matchGrade[1] === 'bronze') {
-            badgeData.text[1] = 'bronze'
-            badgeData.colorscheme = 'orange'
-          } else if (matchGrade[1] === 'none') {
-            badgeData.text[1] = 'no medal'
-            badgeData.colorscheme = 'red'
-          } else {
-            badgeData.text[1] = 'invalid'
-            sendBadge(format, badgeData)
-            return
-          }
-
-          sendBadge(format, badgeData)
-        })
-      })
-    )
+  async handle({ projectUuid }) {
+    const data = await this.fetch({ projectUuid })
+    const { status, grade } = this.transform({ data })
+    return this.constructor.render({ status, grade })
   }
 }
--- a/services/sensiolabs/sensiolabs.tester.js
+++ b/services/sensiolabs/sensiolabs.tester.js
@@ -1,0 +1,151 @@
+'use strict'
+
+// const Joi = require('joi')
+const sinon = require('sinon')
+const { colorScheme } = require('../test-helpers')
+const t = (module.exports = require('../create-service-tester')())
+const serverSecrets = require('../../lib/server-secrets')
+const {
+  // statusRegex,
+  runningMockResponse,
+  platinumMockResponse,
+  goldMockResponse,
+  silverMockResponse,
+  bronzeMockResponse,
+  noMedalMockResponse,
+} = require('./sensiolabs-helpers')
+
+// const messageRegex = /platinum|gold|silver|bronze|no medal/
+const sampleProjectUuid = '45afb680-d4e6-4e66-93ea-bcfa79eb8a87'
+const mockSLUser = 'admin'
+const mockSLApiToken = 'password'
+
+function mockSensiolabsCreds() {
+  serverSecrets['sl_insight_userUuid'] = undefined
+  serverSecrets['sl_insight_apiToken'] = undefined
+  sinon.stub(serverSecrets, 'sl_insight_userUuid').value(mockSLUser)
+  sinon.stub(serverSecrets, 'sl_insight_apiToken').value(mockSLApiToken)
+}
+
+t.create('404 project not found')
+  .get(`/${sampleProjectUuid}.json`)
+  .intercept(nock =>
+    nock('https://insight.sensiolabs.com/api/projects')
+      .get(`/${sampleProjectUuid}`)
+      .reply(404)
+  )
+  .expectJSON({
+    name: 'check',
+    value: 'project not found',
+  })
+
+t.create('401 not authorized')
+  .get(`/${sampleProjectUuid}.json`)
+  .intercept(nock =>
+    nock('https://insight.sensiolabs.com/api/projects')
+      .get(`/${sampleProjectUuid}`)
+      .reply(401)
+  )
+  .expectJSON({
+    name: 'check',
+    value: 'not authorized to access project',
+  })
+
+t.create('pending project')
+  .get(`/${sampleProjectUuid}.json?style=_shields_test`)
+  .intercept(nock =>
+    nock('https://insight.sensiolabs.com/api/projects')
+      .get(`/${sampleProjectUuid}`)
+      .reply(200, runningMockResponse)
+  )
+  .expectJSON({
+    name: 'check',
+    value: 'pending',
+    colorB: colorScheme.grey,
+  })
+
+t.create('platinum medal project')
+  .get(`/${sampleProjectUuid}.json?style=_shields_test`)
+  .intercept(nock =>
+    nock('https://insight.sensiolabs.com/api/projects')
+      .get(`/${sampleProjectUuid}`)
+      .reply(200, platinumMockResponse)
+  )
+  .expectJSON({
+    name: 'check',
+    value: 'platinum',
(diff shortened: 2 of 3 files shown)


services/sensiolabs/sensiolabs.service.js
```diff
@@ -1,110 +1,119 @@
...
+    }),
+  }).required(),
+}).required()
+
+module.exports = class Sensiolabs extends BaseXmlService {
+  static render({ status, grade }) {
+    if (status !== 'finished') {
+      return {
+        message: 'pending',
+        color: 'grey',
```
calebcartwright: I wasn't sure what color to use here (when an analysis is still running) since orange is already used to indicate a different status (completed analysis, bronze medal)

services/sensiolabs/sensiolabs.service.js
```diff
@@ -1,110 +1,119 @@
...
 
-// This legacy service should be rewritten to use e.g. BaseJsonService.
-//
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class Sensiolabs extends LegacyService {
   static get category() {
     return 'build'
```
calebcartwright: IMO `quality` is a better fitting category, although I wasn't sure if changing the category as part of a refactor was okay?

services/sensiolabs/sensiolabs.tester.js
```diff
@@ -0,0 +1,151 @@
+'use strict'
+
+// const Joi = require('joi')
```
calebcartwright: Planning to use this (plus the regexes on lines 9 and 18) as part of the live tests which still need to be written

paulmelnikow:
Thanks for picking this up :) As far as testing strategy, Wheelmap is a similar case that @PyvesB recently did some work on in #2486. It's a combination of mocked tests and live tests, with live tests that run when a token is available. We created accounts with tokens for prod, CI, and staging.

I forgot that document existed; just opened a PR to bring a couple things up to date.

services/sensiolabs/sensiolabs.service.js
paulmelnikow: Yes, I think it makes sense to fix things like that as services are refactored.

services/sensiolabs/sensiolabs.service.js
paulmelnikow: Is there a reference that describes what the grades mean?

calebcartwright: Not that I found (although I didn't spent a ton of time looking)

paulmelnikow:
They are not, though we probably should add one. All the CI tokens are referenced in the CircleCI config. Probably we should also collect these secrets somewhere.

services/sensiolabs/sensiolabs.service.js
paulmelnikow: https://insight.symfony.com/what-we-analyse

screen shot 2018-12-21 at 1 46 11 pm

For the grade badge, I'd probably go with the flow of the design, and assign a color scheme that matches the words. (Their colors aren't bad except the platinum which is pretty green.)

If we can get the raw counts, I'd suggest we offer a second format, which is a bit more in shields-style, showing how many failed critical, major, minor, and info checks there are. In that case, we can use these colors:

screen shot 2018-12-21 at 1 49 44 pm

One or more failed critical is red, one or more failed major is orange, minor yellow, info yellowgreen or green, and all good in bright green.

Something like:

- 
- 
- 

It looks like this product is called SymfonyInsight; maybe we should update our badge name.

calebcartwright:
Nice I like that! I'm traveling most of the next couple days but will definitely add those changes
________________________________

services/sensiolabs/sensiolabs.service.js
calebcartwright: I like that a lot!

> It looks like this product is called SymfonyInsight; maybe we should update our badge name.

I had the same thought. Is it possible to configure the route base to support both `sensiolabs` and `symfony` to maintain backwards compatibility for existing badges? I.e. could we do something like:

```js
static get route() {
    return {
      base: '(sensiolabs|symfony)/i',
      pattern: ':projectUuid',
    }
  }
```

If so should the examples all focus on the "new" path (`symfony`) and not mention the legacy path (`sensiolabs`)?

services/sensiolabs/sensiolabs.tester.js
```diff
@@ -0,0 +1,176 @@
...
+  logTokenWarning,
+} = require('./sensiolabs-test-helpers')
+
+const gradeRegex = /platinum|gold|silver|bronze|no medal/
+const sampleProjectUuid = '45afb680-d4e6-4e66-93ea-bcfa79eb8a87'
+
+t.create('live: valid project')
+  .before(logTokenWarning)
+  .get(`/${sampleProjectUuid}.json`)
+  .timeout(10000)
```
calebcartwright: I've been running some tests locally (using my own user/api token that I created) to validate this test in "live" mode. More often than not it is timing out, even with this 10s timeout period. Hopefully it's just the airport wifi being slow though

services/sensiolabs/sensiolabs.service.js
calebcartwright: My local testing seems to indicate that won't work 😄

services/sensiolabs/sensiolabs.service.js
```diff
@@ -1,110 +1,258 @@
...
+  project: Joi.object({
+    'last-analysis': Joi.object({
+      status: Joi.string()
+        .regex(/ordered|running|measured|analyzed|finished/)
+        .required(),
+      grade: Joi.string()
+        .regex(/platinum|gold|silver|bronze|none/)
+        .allow('', null),
+      violations: Joi.object({
+        violation: Joi.array()
```
calebcartwright: I'm not crazy about this structure but this was the only thing I could get to work. I tried several different parser options based on the `fast-xml-parser` docs but to no avail. If anyone has any tips let me know!

services/sensiolabs/sensiolabs.service.js
```diff
@@ -1,110 +1,258 @@
...
+          .optional()
+          .items(
+            Joi.object({
+              severity: Joi.string()
+                .regex(/info|minor|major|critical/)
+                .required(),
+            })
+          ),
+      })
+        .optional()
```
calebcartwright: Still trying to validate what the xml response structure looks like during various conditions (analysis in progress, analysis completed but no violations found, etc.). This covers all the possibilities based on the xml response node structure, but I may be able to tighten it up more

services/sensiolabs/sensiolabs.service.js
calebcartwright: I added support for including the raw results, as well as providing the corresponding badge as suggested above. I still need to update the colors for the grade badge though

services/sensiolabs/sensiolabs.service.js
```diff
@@ -1,110 +1,258 @@
...
+
   static get category() {
-    return 'build'
+    return 'quality'
   }
 
   static get route() {
     return {
-      base: 'sensiolabs',
+      // base: '(symfony|sensiolabs)/i',
```
calebcartwright: I could not get any sort of pattern/regex to work in `base`, so if there's a way to do so please let me know! In the absence of a dynamic `base` pattern, I had two ideas for approaches that would allow us to update the badge accordingly to reflect the product rebrand to SymfonyInsight without breaking existing users of the `sensiolabs` badge:

1. Stick with a single service class (which I still need to rename, etc.) and ensure the `pattern` would support the old path `sensiolabs` as well as the new path `symfony`
2. Have two service classes, one for `symfony` that would contain the core implementation, and then another for `sensiolabs` that could extend the `symfony` class to reuse that logic

services/sensiolabs/sensiolabs.tester.js
calebcartwright: This is working for me better now (even on hotel wifi) so I'm going to chalk it up to the airport network. Will keep an eye on this test once merged though to see if timeouts are a problem

services/sensiolabs/sensiolabs.service.js
paulmelnikow: I'm not sure about the rest of the pattern, though if you try setting `base` to `''` you can move the dynamic prefix into the format / pattern. We do that with azure devops.

Btw, in the medium term I'd like to solve the aliasing problem using a redirect (#2340).

calebcartwright: I ended up doing just that, adding it to the pattern:

```js
  pattern: '(symfony|sensiolabs)/i/:metric(grade|violations)?/:projectUuid'
```

I also made the metric optional (and default to `grade` when not provided) as the old path was simply `sensiolabs/i/:projectUuid`

What's the pattern for redirects with the new service model?

paulmelnikow: (Totally misread that the first time. Here's take 2!)

Yea, you can't use a pattern in `base`. It doesn't really make sense for it to be dynamic because of the way we use it in testers and examples.

At some point I'd like to solve the aliasing problem using a redirect (#2340). If we can get #2519 merged we could unblock that, though I'm not sure it's a higher priority than finishing up the rewrite. So I'd probably suggest option 1 for now, which is also what we did in azure devops, and then revisiting when we get some redirect capability in core.

paulmelnikow: I wonder if we could do something like this (or its regex equivalent) to avoid providing new API surface area:

```js
pattern: '(sensiolabs/i|symfony/i/:metric(grade|violations))/:projectUuid'
```

If we never start supporting `sensiolabs/i/violations` it's less that needs to be redirected later.

calebcartwright: gotcha, that's a good idea. will do

paulmelnikow:
From the review app:

screen shot 2019-01-01 at 9 45 23 pm

Are these previews wrong? I feel like the LHS on the first one probably should be **grade**.

Colors look good!

These example links are both showing **not authorized to access project**:

- https://shields-staging-pr-2572.herokuapp.com/symfony/i/grade/45afb680-d4e6-4e66-93ea-bcfa79eb8a87.svg
- https://shields-staging-pr-2572.herokuapp.com/symfony/i/violations/45afb680-d4e6-4e66-93ea-bcfa79eb8a87.svg

services/symfony/symfony-insight.service.js
```diff
@@ -0,0 +1,277 @@
...
+  severity: Joi.string()
+    .regex(/info|minor|major|critical/)
+    .required(),
+}).required()
+
+const schema = Joi.object({
+  project: Joi.object({
+    'last-analysis': Joi.object({
+      status: Joi.string()
+        .regex(/ordered|running|measured|analyzed|finished/)
```
paulmelnikow: This can use `Joi.equal`.

services/symfony/symfony-insight.service.js
```diff
@@ -0,0 +1,277 @@
...
+        .required(),
+      grade: Joi.string().regex(/platinum|gold|silver|bronze|none/),
+      violations: Joi.object({
+        // RE: https://github.com/NaturalIntelligence/fast-xml-parser/issues/68
+        // The BaseXmlService uses the fast-xml-parser which doesn't support forcing
+        // the xml nodes to always be parsed as an array. Currently, if the response
+        // only contains a single violation then it will be parsed as an object,
+        // otherwise it will be parsed as an array.
```
paulmelnikow: Does [`Joi.array().single()`](https://github.com/hapijs/joi/blob/master/API.md#arraysingleenabled) help with this?

services/symfony/symfony-insight.service.js
```diff
@@ -0,0 +1,277 @@
...
+    numViolations,
+    numCriticalViolations,
+    numMajorViolations,
+    numMinorViolations,
+    numInfoViolations,
+  }) {
+    if (status !== 'finished') {
+      return {
+        message: 'pending',
+        color: 'grey',
```
paulmelnikow: How about using `lightgrey` for this? I think that's more consistent with other pending states.

services/symfony/symfony-insight.service.js
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  static get route() {
+    return {
+      base: '',
+      // The SymfonyInsight service was previously branded as SensioLabs, and
+      // accordingly the badge path used to be /sensiolabs/i/projectUuid'.
+      // This is used to provide backward compatibility for the old path as well as
+      // supporting the new/current path.
+      format: '(?:sensiolabs/i|symfony/i/(grade|violations))/([^/]+)',
```
paulmelnikow: `violations` is the new path, right? After looking at the example badge, I'm wondering if it should be labeled `checks` instead.

calebcartwright:
> Are these previews wrong? 

Good catch, the violations example was using the wrong render function

> I feel like the LHS on the first one probably should be **grade**.

Agreed. I seriously considered changing the label to `grade` or `medal` but ended up leaving `checks` as that was the label used in original. `grade` certainly seems like a more meaningful message to me so I'm all for changing it. Does the label need to be conditional based on the path? I.e. should `symfony/i/violations/projectid` use `grade` for the label but badges on the legacy path `sensiolabs/i/projectId` keep `checks`? (i'd vote for changing it across the board)
 
> These example links are both showing **not authorized to access project**:
> 
> * https://shields-staging-pr-2572.herokuapp.com/symfony/i/grade/45afb680-d4e6-4e66-93ea-bcfa79eb8a87.svg
> * https://shields-staging-pr-2572.herokuapp.com/symfony/i/violations/45afb680-d4e6-4e66-93ea-bcfa79eb8a87.svg  

Do the staging apps have the SI API and User token secrets set?

paulmelnikow:
> (i'd vote for changing it across the board)

Agreed!

> Do the staging apps have the SI API and User token secrets set?

They don't. Though I wouldn't want to expose the production token there. We could sign up for a separate account though, to get a token to use in staging + tests.

services/symfony/symfony-insight.service.js
calebcartwright: It looks like it does!

services/symfony/symfony-insight.service.js
calebcartwright: I actually think it should be `violations` or `alerts`. From what I've gathered from the Symfony docs they refer to `checks` as the things their analysis performs (the sorts of conditions things they scan/check for, etc.), so their "checks" are fairly synonymous to "rules" in my mind.

paulmelnikow: Okay. So the URL will stay `violations`. Should we change the label from **checks**  to **violations**, then?

calebcartwright: that's what I've done locally 😄, getting ready to push

paulmelnikow:
Do you think we should change the error message when a token is not configured?

calebcartwright:
Interesting idea. What exactly were you envisioning? Throwing some kind of error when the tokens are missing, or making the request anyway and adding a new/different error messages? 

I could see that being helpful for us when looking at staging/prod, and also for any potential self-hosting Shields users that are trying to create badges from their private SymfonyInsight analyses. I definitely think it is worth making a distinction between "oops, you have no tokens" vs. "well you have a token but that token isn't authorized to access project foo"

paulmelnikow:
Yea, I'm thinking throwing an error when there are no tokens, perhaps a new subclass ImproperlyConfigured.

calebcartwright:
I really like that idea. Are you good with updating this PR with a throw, and doing the subclass work in a separate stream?

calebcartwright:
Alright the service will now throw an error (I chose `Inaccessible` as it seemed the best fit with current error options) when the tokens are missing. This had an interesting impact on the tests. There's one test that validates the error is thrown when the tokens are missing, but all the other tests need to have auth present. 

I've added a review comment on the relevant portions of the test file on that topic

services/symfony/symfony-insight.tester.js
```diff
@@ -0,0 +1,316 @@
...
+  majorViolation,
+  minorViolation,
+  infoViolation,
+  multipleViolations,
+} = require('./symfony-test-helpers')
+
+const sampleProjectUuid = '45afb680-d4e6-4e66-93ea-bcfa79eb8a87'
+
+beforeEach(function() {
+  mockSymfonyInsightCreds()
```
calebcartwright: I was hoping to be able to leverage mocha's `describe/context` to separate the different types of tests with their own respective setup methods, but it seems that doesn't work (I could only get top level `before`/`beforeEach` to work).

As such I figured I'd either have to add `.before(..)` and `.finally(..)` calls on each individual test, or use this approach with a singular `beforeEach` and `afterEach` function. This did require a tweak on the checks used in the live tests to ensure that if real tokens existed they would be used, noted below

paulmelnikow: Ah, gotcha. In IcedFrisby you can wrap calls to `frisby.create().toss()` inside an `it()` inside a `describe()` block, and IIRC Mocha will end up running that `describe`'s before hooks. It won't work here because `toss()` doesn't get invoked until well after all the `t.create()` calls have run, which is too late. If you invoke `toss()` directly, the rest of the service test runner won't get a chance to do its thing (e.g. limit the runner to running what we want in a particular PR). (The `beforeEach` hook you've added here works because it's global: it applies to _every_ service test, not just the ones defined in this file. )

You could work around it like this:

```js
function createWithMockCreds(title) {
  return t.create(title)
    .before(mockSymfonyInsightCreds)
    .finally(...)
}
```

or

```js
function create(title, { withMockCreds = false } = {}) {
  const result = t.create(title)
  if (withMockCreds) {
    result.before(...)
    result.finally(...)
  }
  return result
}

Comment (calebcartwright) [reply services/symfony/symfony-insight.tester.js]: > The beforeEach hook you've added here works because it's global: it applies to every service test, not just the ones defined in this file.

Yikes!

Comment (calebcartwright) [reply services/symfony/symfony-insight.tester.js]: I am borderline asleep ha so just to make sure I am following correctly

I can add one of those functions, and then update the test calls to use that function instead of `t.create(...`) correct? For example

```js
t.create('platinum grade')
  .get(`/symfony/i/grade/${sampleProjectUuid}.json?style=_shields_test`)
  .intercept(nock =>
    nock('https://insight.symfony.com/api/projects')
      .get(`/${sampleProjectUuid}`)
      .reply(200, platinumMockResponse)
  )
  .expectJSON({
    name: 'grade',
    value: 'platinum',
    colorB: '#E5E4E2',
  })
```

would become: 


```js
create('platinum grade', { withMockCreds: true })
  .get(`/symfony/i/grade/${sampleProjectUuid}.json?style=_shields_test`)
  .intercept(nock =>
    nock('https://insight.symfony.com/api/projects')
      .get(`/${sampleProjectUuid}`)
      .reply(200, platinumMockResponse)
  )
  .expectJSON({
    name: 'grade',
    value: 'platinum',
    colorB: '#E5E4E2',
  })
```

paulmelnikow: Yup, exactly!

services/symfony/symfony-insight.service.js
```diff
@@ -189,11 +190,15 @@ module.exports = class SymfonyInsight extends BaseXmlService {
       },
     }
 
-    if (serverSecrets && serverSecrets.sl_insight_userUuid) {
-      options.auth = {
-        user: serverSecrets.sl_insight_userUuid,
-        pass: serverSecrets.sl_insight_apiToken,
-      }
+    if (!serverSecrets.sl_insight_userUuid) {
```
paulmelnikow: Could you check both keys instead of just the one?

calebcartwright:
I believe this is ready to merge (barring any final comments/suggestions), but there's now a merge conflict that GH is telling me I don't have access to resolve 🤔 

I deleted the `sensiolabs.service.js` file as part of the refactor & rename, and the service content was moved to `symfony` related files.

paulmelnikow:
Have you tried resolving it in the CLI? If you merge master into this branch you should be able to.

calebcartwright:
Thanks!

Issue #923, opened 2017-03-31
Document private/secret.json


niccokunzmann:
`private/secret.json` contains all information necessary to use APIs with authentication.

Please add the following to the documentation:

- In the [INSTALL.md][s]
  - [ ] a JSON example that shows the form but is not usable
  - [ ] What the [output](https://github.com/badges/shields/blob/2fb5c72ec663ad34a53c2a28059e53d30a3af6bd/server.js#L27) means and that the server works without
  - for each attribute
    - `bintray_apikey` and `bintray_user`
      - [ ] What is it for and what can you do with it.
      - [ ] How to generate it
      - [ ] Examples that describe the form
      - [ ] How to check if it works
    - `gh_client_id` and `gh_client_secret`
      - [ ] What is it for and what can you do with it.
      - [ ] How to generate it
      - [ ] Examples that describe the form
      - [ ] How to check if it works
    - `shieldsIps` and `shieldsSecret`
      - [ ] What is it for and what can you do with it.
      - [ ] How to generate it
      - [ ] Examples that describe the form
      - [ ] How to check if it works
    - `sl_insight_apiToken` and `sl_insight_userUuid`
      - [ ] What is it for and what can you do with it.
      - [ ] How to generate it
      - [ ] Examples that describe the form
      - [ ] How to check if it works
  - [ ] add a description of what the documentation requires if you add a new attribute
  - [ ] Link to [tutorial][t]
- [ ] Fill the tutorial section about [authentication][t]
  - [ ] How to add a new attribut
  - [ ] How to check if it is present
  - [ ] Link to [secrets section][s]

Further Reading:

- Related issues: #593 #747
- Related Code:
  - https://github.com/badges/shields/blob/master/server.js#L36
  - Search for [serverSecrets](https://github.com/badges/shields/search?utf8=%E2%9C%93&q=serverSecrets+&type=)
- [Tutorial][t]
- [secrets section][s]

[s]: https://github.com/badges/shields/blob/master/INSTALL.md#secretjson
[t]: https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#42-querying-an-api-with-authentication

paulmelnikow:
There was some discussion of this in #2573. Let's tackle it in some form.

paulmelnikow:
Thanks for sorting this!

PR #2573, opened 2018-12-21
Update list of prod secrets

paulmelnikow:


chris48s:
None of this necessarily needs to block this change, but reading this document for the first time, I have a couple of thoughts..

1. This documentation is for the use of maintainers, rather than end-users, right? i.e: this document describes the setup of shields.io only. It would be helpful for us to clarify this. Maybe we should even store our own internal documentation for shields.io separate from documentation that is applicable to contributors or self-hosting users.
2. It seems like there are some other services which can consume tokens or credentials which aren't listed here (e.g: JIRA, Jenkins) which are relevant for self-hosting users but which aren't in use on shields.io. As far as I can see we don't have this list fully documented in either https://github.com/badges/shields/blob/master/secret.tpl.json or https://github.com/badges/shields/blob/master/doc/self-hosting.md#server-secrets or here. This seems like something we should improve.

calebcartwright:
> 2. It seems like there are some other services which can consume tokens or credentials which aren't listed here (e.g: JIRA, Jenkins) which are relevant for self-hosting users but which aren't in use on shields.io. As far as I can see we don't have this list fully documented in either https://github.com/badges/shields/blob/master/secret.tpl.json or https://github.com/badges/shields/blob/master/doc/self-hosting.md#server-secrets or here. This seems like something we should improve.

Agreed! The self-hosting doc is excellent, but one area I found it lacking while running my own instance was which services supported tokens, what those tokens were, etc. I think #923 is requesting something along the same lines

PR #2576, opened 2018-12-22
refactor [shippable] service

chris48s:


services/shippable/shippable.service.js
```diff
@@ -1,23 +1,52 @@
...
+  70: { color: '#6BAFBD', label: 'cancelled' },
+  80: { color: '#DC5F59', label: 'failed' },
+  90: { color: '#555555', label: 'stopped' },
+}
+
+const schema = Joi.array()
+  .items(
+    Joi.object({
+      branchName: Joi.string().required(),
+      statusCode: nonNegativeInteger,
```
calebcartwright: Can status code be any non-negative integer, or will it always be one of the values from the above set (0, 10, 20, ... 90)?

chris48s:
I've switched the validation to the schema instead of throwing an exception :+1:

PR #2579, opened 2018-12-23
Use beta release of nock

paulmelnikow:
I've been working on nock these past couple weeks. Since that work has including dropping support for Node 6, it's been shipping to the `beta` tag instead of the `latest` tag.

It'd be great to run Shields off the beta release so I can continue to validate it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1374,7 +1374,7 @@
       "dependencies": {
         "chalk": {
           "version": "1.1.3",
-          "resolved": "http://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
+          "resolved": "https://registry.npmjs.org/chalk/-/chalk-1.1.3.tgz",
           "integrity": "sha1-qBFcVeSnAv5NFQq9OHKCKn4J/Jg=",
           "dev": true,
           "requires": {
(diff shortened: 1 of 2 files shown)


paulmelnikow:
It’s already mentioned in the docs. Postinstall throws an error :)

calebcartwright:
My bad, I see it in the main Readme now (I was looking at the tutorial earlier)

calebcartwright:
On Ubuntu 16.x what version of node is in `nodejs-legacy`? For some reason I thought it was 6.x but it's been a while

(This was the part of the tutorial I was looking at earlier https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#2-setup)

paulmelnikow:
I’ve no idea. I wonder if that’s even the recommended way to use node. On MacOS I recommend nvm.

calebcartwright:
Folks will see the main, highly visible Dev Instructions on the main page before getting to the repo and know they will require 8.x. They'll know the best way to do that on their platform so no worries

paulmelnikow:
I wonder if we should remove some of those tutorial instructions, or link out to some. Given the nature of the project, we do get first time Node devs contributing.

calebcartwright:
So maybe pointing them to something like https://nodejs.org/en/ or https://nodejs.org/en/download/? I think that's fair (the tutorial does the same thing for git) 

We could also do the same for imagemagick (https://www.imagemagick.org/script/download.php)

paulmelnikow:
Is that mentioned in the tutorial? Probably it doesn’t need to be. It’s only needed for raster badges, which is a pretty rare use these days.

PR #2588, opened 2018-12-24
Update coverage script

calebcartwright:
I noticed that our coverage went _way_ down (~49%) in the latest daily build and I think this is why

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "coverage:test:services": "nyc npm run test:services",
     "coverage:clean": "rimraf .nyc_output coverage",
     "precoverage:test": "run-s --silent coverage:clean defs features",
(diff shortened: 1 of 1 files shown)


calebcartwright:
So revert this change and update the config for the daily service tests? If so, can do!

If so then it sounds like the intent of the `coverage:test` script is to run tests w/coverage for everything _except_ the services correct? If so I see the value in that (especially having looked through the daily test logs), but it's not immediately clear to me that was the intent/purposed based on the script name (I assumed that that particular script was to run all tests w/coverage for the entire codebase)

calebcartwright:
Do I have this correct? 

Add new npm script in the daily-service repo (https://github.com/badges/daily-tests/blob/master/package.json#L12) like `coverage:services` and then update the `test` script in that repo to add that new script?

paulmelnikow:
Yes. And we should make sure the scripts that run `coverage:test` and `coverage:services` are set to `when: always`.

Issue #28, opened 2013-04-10
Badges for Semaphore


vlasar:
Hello,

I'm working on [Semaphore](https://semaphoreapp.com/) and would like to get it "shielded".

We uses badges for the following **build statuses**:
- passing
- pending
- failing
- unknown

Build badges from here:
![Shields Badges](https://raw.github.com/olivierlacan/shields/master/shields_white.png)

suite our color taste but we need a badge for the "unknown" status instead of the "error" status.

Could you create those badges for us or what is the standard workflow?

stevelacey:
Would be nice if these were updated to match the new badges:

![screen shot 2015-02-17 at 12 03 25](https://cloud.githubusercontent.com/assets/289531/6227887/010a5c9a-b69d-11e4-9ae5-0c106a5a1b37.png)

espadrine:
@stevelacey Are the Semaphore badges not defaulting to the right style? Do they rely on `?style=flat`?

stevelacey:
@espadrine https://semaphoreapp.com/api/v1/projects/eb289e04-3982-4de1-9e66-8436a67cf607/262663/shields_badge.svg?style=flat doesn't seem to do anything – they self-hosted – so need an update? Dunno – there's a few services with outdated badges though, see https://github.com/crysalead/kahlan

espadrine:
Indeed. @vlasar do you have information on that? Do you need help?

vlasar:
I'm no longer working on Semaphore, but I am trying to get in touch with them. cc @markoa

markoa:
Hi folks, it looks like it is now expected that shields.io is proxying the badge URLs to the CI service, while on Semaphore we're still hosting the images ourselves. What is the procedure to do it the proxy way?

espadrine:
@markoa To avoid having to deal with image updates manually, the simplest way to do is to request the badge from shields.io at `https://img.shields.io/badge/build-passing-brightgreen.svg`. If possible, supporting the `?style=` modifier would be great (eg, `https://img.shields.io/badge/build-passing-brightgreen.svg?style=flat-square`.)

If you wish to continue to self-host those badges, you can of course download them from shields.io, and update them periodically.

markoa:
@espadrine Thanks for the explanation. I think we'll start requesting the badge from shields.io. Would it be useful to post here when we implement it?

espadrine:
@markoa Yes! Definitely.

stevelacey:
@markos any progress on this?

[...]

PR #2591, opened 2018-12-24
Added [bStats] badges

Prouser123:
Added bStats badges.

The main thing here is that I had to make a 'bridge' of sorts as the bStats API for players/servers helpfully returns `[[1545685200000,57224]]` [[time,amount]] and so isn't formatted properly. The 'bridge' formats it and returns as `{"pluginid": 1, "servers": 57224}` so that Joi can validate it.

It is hosted with Glitch (which usually turns off envs after 15 mins) but Uptime Robot will be pinging it every 5 mins (which Glitch allows) to keep it online, but this isn't really a permanent solution.

Bridge: 

[bStats Issue #28- Add shields](https://github.com/Bastian/bStats/issues/28)


services/bstats/bstats-players.service.js
```diff
@@ -0,0 +1,68 @@
+'use strict' // (1)
```
calebcartwright: Can you remove these inline numbered comments (`// (1)`, etc.)

Prouser123: Forgot about that, give me a moment.

calebcartwright: no worries!

calebcartwright:
Can you also include a link to the repo where your bridge code lives? I want to get a feel for what the networking calls look like from beginning to end

Prouser123:
@calebcartwright https://glitch.com/edit/#!/bstats-api-format

services/bstats/bstats-players.service.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+const BaseJsonService = require('../base-json')
+
+const Joi = require('joi')
+const schema = Joi.object({
```
calebcartwright: I think we could use a different schema here that would allow for interfacing with the bStats API directly and avoid the need for the intermediate bridge hop:

```js
const schema = Joi.array()
  .items(Joi.array().items([Joi.number().required(), Joi.number().required()]))
  .required()
```

services/bstats/bstats-servers.service.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+const BaseJsonService = require('../base-json')
+
+const Joi = require('joi')
+const schema = Joi.object({
```
calebcartwright: Same comment as the players service:

```js
const schema = Joi.array()
  .items(Joi.array().items([Joi.number().required(), Joi.number().required()]))
  .required()
```

services/bstats/bstats-servers.service.js
```diff
@@ -0,0 +1,61 @@
...
+  async fetch({ pluginid }) {
+    return this._requestJson({
+      schema,
+      url: `https://bstats-api-format.glitch.me/servers/${pluginid}`,
+    })
+  }
+
+  static render({ servers }) {
+    return {
+      label: 'servers',
```
calebcartwright: You can exclude `label` here (as the value will always be `servers` and it is defined in `defaultBadgeData`) and you can also move `color` to `defaultBadgeData` as well

services/bstats/bstats-servers.service.js
```diff
@@ -0,0 +1,61 @@
...
+      message: servers,
+      color: 'blue',
+    }
+  }
+
+  // Front page
```
calebcartwright: Let's get rid of this one too :smile:

services/bstats/bstats-servers.service.js
```diff
@@ -0,0 +1,61 @@
...
+
+  static get category() {
+    return 'other'
+  }
+  static get examples() {
+    return [
+      {
+        title: 'bStats Servers',
+        exampleUrl: '1',
```
calebcartwright: We want to use `namedParams` here. For example replace `exampleUrl` with:

```js
namedParams: {
  pluginId: 1,
}
```

services/bstats/bstats-servers.service.js
```diff
@@ -0,0 +1,61 @@
...
+        pattern: ':pluginid',
+        staticPreview: {
```
calebcartwright: You can also change the value of `staticPreview` to call your render function:

`staticPreview: this.render({ servers: 57479 })`

services/bstats/bstats.tester.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
```
calebcartwright: We should have a `tester` for each defined `service` so if you're going to have different services for `servers` and `players` then let's split this into two different tester files (`bstats-servers.tester.js` and `bstats-players.tester.js`)

services/bstats/bstats.tester.js
```diff
@@ -0,0 +1,33 @@
...
+
+const Joi = require('joi')
+
+const ServiceTester = require('../service-tester')
```
calebcartwright: We're trying to use a new shorthand for tester instantiation, so would you replace lines 5-9 with this:

`const t = (module.exports = require('../create-service-tester')())`

services/bstats/bstats.tester.js
```diff
@@ -0,0 +1,33 @@
...
+const t = (module.exports = new ServiceTester({
+  id: 'bstats',
+  title: 'bStats badges',
+}))
+
+const withRegex = re => Joi.string().regex(re)
+
```
calebcartwright: For future reference, you'll find a lot of these sorts of validators defined in the test-validators file (`services/test-validators.js`) so you'll generally want to check there first to see of something already exists (`withRegex` already does for example)

services/bstats/bstats.tester.js
```diff
@@ -0,0 +1,33 @@
...
+const isNumber = withRegex(/^[-+]?\d+$/)
```
calebcartwright: Does the underlying response from the bStats API include an optional `-` or `+` prefix? Or is it just a number? From what I've read it (albeit briefly) it looks to just be integers, in which case you could just use `Joi.number()` for the value checks on lines 21 and 31 right?

services/bstats/bstats-players.service.js
```diff
@@ -0,0 +1,61 @@
...
+  static get category() {
+    return 'other'
+  }
+  static get examples() {
+    return [
+      {
+        title: 'bStats Players',
+        exampleUrl: '1',
+        pattern: ':pluginid',
+        staticPreview: {
```
calebcartwright: You can also change the value of `staticPreview` to call your render function:

`staticPreview: this.render({ players: 74299 })`

services/bstats/bstats-players.service.js
```diff
@@ -0,0 +1,61 @@
...
+  async fetch({ pluginid }) {
+    return this._requestJson({
+      schema,
+      url: `https://bstats-api-format.glitch.me/players/${pluginid}`,
+    })
+  }
+
+  static render({ players }) {
+    return {
+      label: 'players',
```
calebcartwright: You can exclude `label` here (as the value will always be `players` and it is defined in `defaultBadgeData`) and you can also move `color` to `defaultBadgeData` as well

services/bstats/bstats-players.service.js
```diff
@@ -0,0 +1,61 @@
...
+  // Front page
+
+  static get category() {
+    return 'other'
+  }
+  static get examples() {
+    return [
+      {
+        title: 'bStats Players',
+        exampleUrl: '1',
```
calebcartwright: We want to use `namedParams` here. For example replace `exampleUrl` with:

```js
namedParams: {
  pluginId: 1,
}
```

paulmelnikow:
@calebcartwright Care to give this another pass?

calebcartwright:
@paulmelnikow  - @Prouser123 and I have been chatting and discord and there's a couple other pending changes. Once those come through I'll happily review

paulmelnikow:
@Prouser123 It would be great to get this merged!

calebcartwright:
Alrighty @Prouser123 I've got this working. I've included the relevant snippets below:

### Players Service
Replace your schema with:
```js
const schema = Joi.array()
  .items(Joi.array().items([Joi.number().required(), Joi.number().required()]))
  .required()
```

Replace your `handle` and `fetch` functions with (note the new `transform` function):
```js
async handle({ pluginid }) {
  const json = await this.fetch({ pluginid })
  const { players } = this.transform({ json })
  return this.constructor.render({ players })
}

transform({ json }) {
  const players = json[0][1]
  return { players }
}

async fetch({ pluginid }) {
  const url = `https://bstats.org/api/v1/plugins/${pluginid}/charts/players/data`

  return this._requestJson({
    schema,
    options: {
      qs: {
        maxElements: 1,
      },
    },
    url,
  })
}
```

paulmelnikow:
@calebcartwright Merge when ready! 😀

PR #2581, opened 2018-12-23
Visual Studio Code gitignore + recommended extensions

Prouser123:
- Added `.vscode` (workspace settings directory) to the gitignore.

PR #2598, opened 2018-12-27
[Bitbucket]: Pull Requests: Support Bitbucket Server

### LINKED ISSUE #1757: Is it possible to connect to a private repo in bitbucket?
HI, I have a private bitbucket repo . all my servers are local |
how can I use shields  ? 
Thanks for helping

nlowe:
Add support for counting pull requests from a bitbucket server instance as requested in #1757. I didn't do build status since that can be handled by other badges.

Fixes #1757


calebcartwright:
Thanks for your contribution to Shields! FYI I modified the name of the PR to put the service into brackets so the corresponding service tests would be run as part of the cheks

secret.tpl.json
```diff
@@ -1,8 +1,10 @@
...
-  "wheelmap_token": "${WHEELMAP_TOKEN}"
-}
+{
+  "gh_client_id": "${GH_CLIENT_ID}",
+  "gh_client_secret": "${GH_CLIENT_SECRET}",
+  "shieldsIps": [ "${SHIELDS_IP}" ],
+  "gh_token": "${GH_TOKEN}",
+  "npm_token": "${NPM_TOKEN}",
+  "wheelmap_token": "${WHEELMAP_TOKEN}",
+  "bitbucket_server_username": "${BITBUCKET_SERVER_USERNAME}",
```
calebcartwright: I think we typically only add tokens here that are used by the main Shields.io service, and not self-hosted instances (see the [Jira service](https://github.com/badges/shields/blob/master/services/jira/jira-base.js#L14) for example. @paulmelnikow is that right?

nlowe:
Yeah that makes sense. Even though Bitbucket Server "versions" its API, it looks like they never change the version (5.16.0, released a month ago, still lists `1.0` as the newest according to [their documentation](https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html#idm46229602363312)).

As for unifying the service URL for both versions, I'm not familiar enough with the project to know how to add an optional parameter (thinking `${proto}://${host}` here). Would that be best added as a query parameter?

calebcartwright:
> As for unifying the service URL for both versions, I'm not familiar enough with the project to know how to add an optional parameter (thinking ${proto}://${host} here). Would that be best added as a query parameter?

So the first thing we want to make sure we do here is ensure we don't break anyone using the existing PR badges for BB. Currently those folks are using the paths:

`/bitbucket/pr/:user/:repo` (open PRs)
`/bitbucket/pr-raw/:user/:repo` (all PRs)

Therefore the route would need to support that pattern as well as the new type of pattern we want to add to support BBS. I'd _think_ we'd want to have something like:

`/bitbucket/pr/:user/:repo` for PR Counts from BB and then support something like `/bitbucket/pr/:protocol/:hostAndPath/:project/:repo` for BBS

Obviously that regex starts to get a little tricky 😄 In similar instances with other services I've had to use `format` and `capture` (instead of `pattern`) in the route to support these sorts of scenarios. I haven't actually tested this regex, but we should be able to do something like the below for the route (or whatever the valid regex is for the format):

```js
static get route() {
  return {
    base: `bitbucket/${routePrefix}`,
    format: '(?:(http|https)/(.+)/)?([^/]+)/([^/]+)',
    capture: [ 'protocol', 'hostAndPath', 'user', 'repo'],
  }
}
```

Note that the `hostAndPath` will support BBS deploys like: `https://myinternal.bitbucketserver.com:8000/foo`. `User` would be the same as `project` as far as the regex is concerned so I think using a single capture group & name there would be best (the examples shown to end users can use `:user` and `:project` respectively as needed regardless)

With that kind of route we could update the `handle` signature to take in the parameters and based on those parameters we could determine if we're targeting BBS or BB and act accordingly : 
```js
async handle({ protocol, hostAndPath, user, repo }) {
  let usingBitbucketServer = false
  let url = `https://bitbucket.org/api/2.0/repositories/${user}/${repo}/pullrequests/`
  if (hostAndPath) {
    usingBitbucketServer = true
    url = `${protocol}://${hostAndPath}/rest/api/1.0/projects/${user}/repos/${repo}/pull-requests`
  }
  ...
}

That's obviously a little raw and in need of some cleanup, but I think it conveys the idea well enough

File: services/bitbucket-server/bitbucket-server.service.js [hunk 2]
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  async fetch({ proto, host, project, repo }) {
+    let authenticatedHost = host
+    if (
+      serverSecrets &&
+      serverSecrets.bitbucket_server_username &&
+      serverSecrets.bitbucket_server_password
+    ) {
+      authenticatedHost = `${serverSecrets.bitbucket_server_username}:${
```
calebcartwright: Is it possible to use auth headers for this instead? See the [jira auth](https://github.com/badges/shields/blob/master/services/jira/jira-base.js#L14) for an example of a service that use basic auth and there's other services using tokens, etc. 

The creds can be added to an `options` object which can be past to `request` via the `options` param to the `_requestJson()` function

services/bitbucket-server/bitbucket-server.service.js
```diff
@@ -0,0 +1,77 @@
...
+        serverSecrets.bitbucket_server_password
+      }@${host}`
+    }
+
+    return this._requestJson({
+      schema,
+      url: `${proto}://${authenticatedHost}/rest/api/1.0/projects/${project}/repos/${repo}/pull-requests?limit=100&state=OPEN&wthProperties=false&withAttributes=false`,
```
calebcartwright: Similar to the authentication, let's use the query string object on the `options` param to the `_requestJson()` function. For example:

```js
const options = {
  qs: {
    limit: 100,
    state: OPEN
  ...
  }
}
```

calebcartwright: Also are there are any typical HTTP response codes we could see from BBS (404, etc.)? For example when the specified `repo` doesn't exist in the specified `project`, specified `project` doesn't exist, not authorized to access specified repo and/or project, BBS instance isn't reachable, etc. 

If there then we can utilize the `errorMessages` param to the `_requestJson()` function to provide a more meaningful error message for the user. For example:

```js
errorMessage: {
  404: 'project or repo not found`,
  403: 'not authorized to access project or repo',
}
```

services/bitbucket-server/bitbucket-server.service.js
nlowe: This should definitely be possible.

services/bitbucket-server/bitbucket-server.service.js
nlowe: I'll have to try some examples and/or dig through their documentation but there are most likely some codes we could watch out for.

nlowe:
@calebcartwright Thanks for all the help. I finally had some time to circle back to this and was porting it from a legacy service I had created earlier in the year. I'll take a look at your suggestions tomorrow if I have some time.

services/bitbucket-server/bitbucket-server.service.js
calebcartwright: No worries if there aren't. I've used BBS before but never the APIs directly. I personally use the Advanced Rest Client frequently when working on a service as I find it helps me check the responses (codes and headers in particular) very easily

secret.tpl.json
paulmelnikow: It's the opposite: this file is used in the Docker deployment, not in production.

nlowe:
@calebcartwright I was able combine the new service with the existing one. I also followed your suggestions and used the object-style query strings, basic auth, and error code handlers. PTAL when you get a chance.

calebcartwright:
👍 thanks, will review shortly

package.json
```diff
@@ -96,6 +96,7 @@
     "heroku-postbuild": "run-s --silent build",
     "analyze": "cross-env ANALYZE=true LONG_CACHE=false BASE_URL=https://img.shields.io npm run build",
     "start:server": "cross-env HANDLE_INTERNAL_ERRORS=false RATE_LIMIT=false node server 8080 ::",
+    "start:server:debug": "cross-env ALLOWED_ORIGIN=http://localhost:3000 HANDLE_INTERNAL_ERRORS=false RATE_LIMIT=false node --nolazy --inspect=9229 server 8080 ::",
```
calebcartwright: was this an intentional add? I'm wondering if `9229` is a process id and how dynamic that value will be

secret.tpl.json
calebcartwright: Gotcha, just wondering for the sake of consistency whether the BB auth info should be here, and/or if this should also be updated to include the auth info for all the other services (and if so if we should include this step in the docs as part of #2599)

nlowe: Definitely, I'd love to see the other options added there as well, it makes on-prem deployment a lot easier.

package.json
nlowe: Yes. `9229` is a port. I was following the [VS Code Debugging instructions](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_launch-configuration-support-for-npm-and-other-tools), I figure others would find it useful. I can revert this if you'd like.

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -24,15 +25,57 @@ function pullRequestClassGenerator(raw) {
...
+        errorMessages: {
+          401: 'Authentication Required',
+          403: 'Private Repo',
+          404: 'Repo Not Found',
+        },
+      })
+    }
+
+    async fetch({ proto, hostAndPath, user, repo }) {
+      if (proto !== undefined && hostAndPath !== undefined) {
```
calebcartwright: This doesn't necessarily need to be changed, but with the regex defined in the route format we don't need to check both of these (if one has a value then the other will always be guaranteed to have a value as well)

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -24,15 +25,57 @@ function pullRequestClassGenerator(raw) {
...
+      const options = {
+        qs: {
+          state: 'OPEN',
+          limit: 100,
+          withProperties: false,
+          withAttributes: false,
+        },
+      }
+
+      if (
```
calebcartwright: The one _potential_ limitation I see with only including auth when hitting a BBS instance is that it means it won't be possible to access private BB repos via a self-hosted shields instance (the second scenario from a previous comment). 

I know your primary motivation here is to get support for BBS, but do we know if the same basic auth can work with BB (or can we determine quickly if so)?

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -60,6 +103,12 @@ function pullRequestClassGenerator(raw) {
           pattern: ':user/:repo',
           staticExample: this.render({ prs: 22 }),
         },
+        {
+          title: 'Bitbucket Server open pull requests',
+          exampleUrl: 'https/bitbucket.mydomain.net/foo/bar',
```
calebcartwright: I know `exampleUrl` usage was already in here, but would you mind updating this (and the above) to use `namedParams` instead? It's a change we're slowly making across the board in all services for performance, etc. improvements. Basically just swap out the `exampleUrl` key and replace it with:

```js
namedParams: {
  proto: 'https',
  hostAndPath: 'bitbucket.mydomain.net',
  user: 'foo',
  repo: 'bar',
}
```

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -24,15 +25,57 @@ function pullRequestClassGenerator(raw) {
...
+          user: serverSecrets.bitbucket_server_username,
+          pass: serverSecrets.bitbucket_server_password,
+        }
+      }
+
+      return this._requestJson({
+        url,
+        schema: bitbucketPullRequestsSchema,
+        options,
+        errorMessages: {
```
calebcartwright: 👍

package.json
calebcartwright: I was just wondering whether this would be user/machine specific. If it isn't and everyone can run it as-is on his or her machine then 👍, but if not then we should prob remove

services/bitbucket/bitbucket-pull-request.service.js
nlowe: Makes sense

services/bitbucket/bitbucket-pull-request.service.js
nlowe: I think I only have a bitbucket cloud account through github, which I don't think will work with basic auth. I'll create an account to test with.

calebcartwright: Much appreciated! Don't spend too much time on this though. No one's asked about private BB repos so far and I don't want to hold up BBS support for that. If this is something we can cover quickly great but if not then we can roll with this as-is and cross the private BB repo bridge if/when it comes up

package.json
nlowe: Unless you're debugging another app on that same port it should work OOTB, you just have to add the debug profile in VSCode's `launch.json`:

```json
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
      {
        "type": "node",
        "request": "launch",
        "name": "Launch via NPM",
        "cwd": "${workspaceFolder}",
        "runtimeExecutable": "npm",
        "runtimeArgs": [
          "run-script", "start:server:debug"
        ],
        "port": 9229
      }
    ]
}
```

services/bitbucket/bitbucket-pull-request.service.js
nlowe: Actually, now that I think about it more, I don't think this will work. If we want to support private repos on cloud we will have to add a separate secret, otherwise we won't be able to mix cloud and server requests on one instance, as the credentials for the BBS account probably don't work for BB. Thoughts?

calebcartwright: That's correct. We'd have two sets of creds, one for BB and another for BBS. I was thinking that could be handled by creating the `options` object in the main `fetch` method, doing the auth checks there, and passing the options object along to the respective fetch functions, or just adding a similar auth check to the `fetchCloud` function

nlowe:
@calebcartwright Thanks for all the help, I added extra tests for the other status codes. Is there an easy way to mock out the credentials so I can test the presence / absence of the basic auth header? It's the only test cases I think I'm missing.

calebcartwright:
> @calebcartwright Thanks for all the help, I added extra tests for the other status codes. Is there an easy way to mock out the credentials so I can test the presence / absence of the basic auth header? It's the only test cases I think I'm missing.

Great question! We do have a pattern that allows for stubbing the serverSecret values as part of the setup/teardown process of the test to cover those scenarios. One example is [here](https://github.com/badges/shields/blob/master/services/jira/jira-issue.tester.js#L166) in the Jira service

nlowe:
Do you guys want this squashed?

calebcartwright:
> Do you guys want this squashed?

No don't worry about it (we'll squash on merge)

package.json
paulmelnikow: Could this be documented somewhere in the contributing guidelines? And if it's vscode-specific, could that be reflected in the script name? Perhaps `start:server:vscode-debug` or something like that.

secret.tpl.json
paulmelnikow: Makes sense. I posted a note about that here: https://github.com/badges/shields/pull/2599#issuecomment-450378504. I'd like to fold the secrets and the environment into a single configuration method that covers both. Hopefully that could subsume this file and provide a consistent way that works for on-prem and PaaS without a deploy commit.

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -47,19 +98,33 @@ function pullRequestClassGenerator(raw) {
...
+            repo: 'python-bitbucket',
+          },
           pattern: ':user/:repo',
           staticExample: this.render({ prs: 22 }),
         },
+        {
+          title: 'Bitbucket Server open pull requests',
+          namedParams: {
+            proto: 'https',
+            hostAndPath: 'bitbucket.mydomain.net',
```
paulmelnikow: It sounds like there was some consideration about using a query parameter. Re https://github.com/badges/shields/pull/2598#issuecomment-450236192, I'd like to move away from complicated patterns since they are difficult to express in the UI.

I like the interface that you're using here, though `?server_url=...` would be easier to express in the UI. I wonder if there's any difference in ease of use, from the user's perspective.

I don't have a strong feeling either way, though happy to continue to discuss.

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -47,19 +98,33 @@ function pullRequestClassGenerator(raw) {
     static get route() {
       return {
         base: `bitbucket/${routePrefix}`,
-        format: '([^/]+)/([^/]+)',
-        capture: ['user', 'repo'],
+        format: `(?:(http|https)/(.+)/)?([^/]+)/([^/]+)`,
```
paulmelnikow: What do you think about limiting the Bitbucket Server support to `https`?

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -14,14 +15,64 @@ function pullRequestClassGenerator(raw) {
...
+        args.options.auth = {
+          user: serverSecrets.bitbucket_username,
+          pass: serverSecrets.bitbucket_password,
+        }
+      }
+
+      return this._requestJson(args)
+    }
+
+    async fetchServer({ args, proto, hostAndPath, user, repo }) {
```
paulmelnikow: Could you add a link here to the API doc?

services/bitbucket/bitbucket.tester.js
```diff
@@ -85,6 +92,119 @@ t.create('pr (private repo)')
   .get('/pr/chris48s/example-private-repo.json')
   .expectJSON({ name: 'pull requests', value: 'private repo' })
 
+t.create('pr (server)')
```
paulmelnikow: Can we find any publicly facing projects that would support a live test?

package.json
calebcartwright: We're also in the process of adding a vs code `extensions.json` file to provided recommended extensions in #2581 so perhaps we could do all of the above :

* update the script name to include vscode as suggested
* include the `.vscode/launch.json` file in the repo (update the gitignore with a negated pattern for this or change the excluded path)
* update contributor docs with relevant info (this could be done in future work) on the script/launch settings and extensions

services/bitbucket/bitbucket-pull-request.service.js
calebcartwright: I'm also good with query params, just haven't worked with any services like that so I suggested the pattern

paulmelnikow: As of now the param validation has to happen in the handler, though there's no reason it couldn't be added to BaseService, down the line.

secret.tpl.json
chris48s: Thinking about the conversation in https://github.com/badges/shields/pull/2599#discussion_r244350388
could we make these params `_user` and `_pass` as opposed to `_username` and `_password`.

services/bitbucket/bitbucket-pull-request.service.js
chris48s: Given `hostAndPath` is optional, I think it makes more sense for it to be a GET param in this case instead of being part of the route (especially given its optional and in the middle of the route). It makes more sense to think of this case as being analogous to the NPM badges where we're mostly going to be using NPM itself but we can optionally specify a custom registry, as opposed to something like JIRA or Jenkins CI where we always need to manually specify an instance.

calebcartwright: I'm 👍  on using query params here too, but just a side note/question:

Isn't `hostAndPath` optional here though because the current/existing implementation assumes Bitbucket Cloud as the default instance? The Jira instance could've assumed Jira Cloud as the default instance as well (and then made `hostAndPath` optional) it just didn't make that assumption

secret.tpl.json
nlowe: Sounds good, I'll rename them.

package.json
nlowe: @paulmelnikow Do you want the documentation / task in this PR or a different PR?

services/bitbucket/bitbucket-pull-request.service.js
nlowe: My only hesitation on using a query param is people who are manually writing badge paths have to URL-Encode the param. I'm not sure how often people would run into that but I think it's still something to consider.

services/bitbucket/bitbucket-pull-request.service.js
nlowe: While I agree everyone ***should*** be using `https` in this day and age I don't know that that's always going to be the case. I know my company still has **tons** of internal apps on `http` only. Do other on-prem services have this limitation?

services/bitbucket/bitbucket.tester.js
nlowe: I'll look but unless it's a well known company there's no guarantee that it'll stick around.

services/bitbucket/bitbucket-pull-request.service.js
paulmelnikow: All good points. I'd suggest a middle path which makes us part of the solution: if there's a documented limitation or pattern of running _this_ service without https then let's support it. If not we could at least wait until someone complains…

nlowe: The default documentation for running BBS via [docker](https://hub.docker.com/r/atlassian/bitbucket-server/) and on a [linux server](https://confluence.atlassian.com/bitbucketserver/install-bitbucket-server-on-linux-from-an-archive-file-868977010.html)  both use `http` as the default scheme, with extra links for setting up `https`. While this is certainly not requiring the use of `http`, I feel we should still allow users to specify the scheme if other services do so. If we want to start locking down services to only support `https`, I feel that should be a separate issue.

services/bitbucket/bitbucket.tester.js
nlowe: Yeah I haven't found any. Going to mark this as resolved unless someone can find one we can use.

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -14,14 +15,65 @@ function pullRequestClassGenerator(raw) {
...
-    async fetch({ user, repo }) {
-      const url = `https://bitbucket.org/api/2.0/repositories/${user}/${repo}/pullrequests/`
-      return this._requestJson({
-        url,
+    async fetchCloud({ args, user, repo }) {
+      args.url = `https://bitbucket.org/api/2.0/repositories/${user}/${repo}/pullrequests/`
+      args.options = { qs: { state: 'OPEN', limit: 0 } }
+
+      if (
+        serverSecrets &&
```
calebcartwright: Let's remove the truthy check on `serverSecrets` here (see #2606 and #2634 for more info)

services/bitbucket/bitbucket-pull-request.service.js
paulmelnikow: Okay, I'm fine with that.

services/bitbucket/bitbucket-pull-request.service.js
paulmelnikow: Think converting this to a query parameter is the main thing pending on this.

nlowe:
I think that's everything, although I need to read through #2626 to see if I have to migrate over any of the secrets we added.

nlowe:
Did the format of examples change? CI is failing:

```bash
$ #!/bin/bash -eo pipefail
npm run defs && npm run features

> shields.io@0.0.0 defs /home/circleci/project
> node scripts/export-service-definitions-cli.js > service-definitions.yml

/home/circleci/project/node_modules/joi/lib/index.js:193
                throw error;
                ^

ValidationError: Example for BitbucketPullRequests at index 1 {
  "title": "Bitbucket Server open pull requests",
  "namedParams": {
    "user": "foo",
    "repo": "bar"
  },
  "queryParams": {
    "server": "https://bitbucket.mydomain.net"
  },
  "pattern": ":user/:repo",
  "staticExample" [1]: {
    "message": "42",
    "color": "yellow"
  }
}

[1] "staticExample" is not allowed
    at Object.exports.process (/home/circleci/project/node_modules/joi/lib/errors.js:203:19)
    at internals.Object._validateWithOptions (/home/circleci/project/node_modules/joi/lib/types/any/index.js:764:31)
    at module.exports.internals.Any.root.validate (/home/circleci/project/node_modules/joi/lib/index.js:147:23)
    at module.exports.internals.Any.root.attempt (/home/circleci/project/node_modules/joi/lib/index.js:177:29)
    at validateExample (/home/circleci/project/services/transform-example.js:39:22)
    at transformExample (/home/circleci/project/services/transform-example.js:91:7)
    at examples.examples.map (/home/circleci/project/services/base.js:141:7)
    at Array.map (<anonymous>)
    at Function.getDefinition (/home/circleci/project/services/base.js:140:36)
    at loadServiceClasses.map.ServiceClass (/home/circleci/project/services/index.js:57:39)
    at Array.map (<anonymous>)
    at collectDefinitions (/home/circleci/project/services/index.js:57:6)
    at Object.<anonymous> (/home/circleci/project/scripts/export-service-definitions-cli.js:7:21)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Function.Module.runMain (module.js:694:10)
    at startup (bootstrap_node.js:204:16)
    at bootstrap_node.js:625:3
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! shields.io@0.0.0 defs: `node scripts/export-service-definitions-cli.js > service-definitions.yml`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the shields.io@0.0.0 defs script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/circleci/.npm/_logs/2019-01-07T17_48_56_386Z-debug.log
Exited with code 1
```

nlowe:
@paulmelnikow @calebcartwright Let me know if there's anything else that needs tweaked, I think this is ready for merging.

calebcartwright:
Thanks @nlowe! giving this a final look over now

.gitignore
```diff
@@ -1,4 +1,5 @@
 # Project-specific files
+/.vscode
```
calebcartwright: Can you remove this line? VS Code info was added to the gitignore in https://github.com/badges/shields/pull/2581/files to ensure we'd exclude settings files, but allow for the inclusion of other files (like extensions currently and potentially launch, etc. in the future)

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -14,14 +15,63 @@ function pullRequestClassGenerator(raw) {
   const badgeSuffix = raw ? '' : ' open'
 
   return class BitbucketPullRequests extends BaseJsonService {
```
calebcartwright: One final nitpick 😄 Can we make the class name here consistent with the file name? I didn't realize the discrepancy (services class name ends with `requests` whereas file name ends with `request`) til trying to run the tests

PR #919, opened 2017-03-29
Fetch Bower data using libraries.io

paulmelnikow:
- Avoid downloading bower packages via git
- Service is well resourced: https://libraries.io/team
- Same service used by https://bower.io/search/

A standard API key with https://libraries.io/ covers 60 requests per minute. Will that be enough for the public server?

Do you see any reason not to use `let`, `const`, template strings, and arrow functions in new code?

PR #2599, opened 2018-12-27
Document private/secret.json; affects [bower jira]

### LINKED ISSUE #923: Document private/secret.json

chris48s:
Closes #923 plus a couple of tidy-ups to bower and JIRA services which I noticed while writing the docs.

Having a json file and some docs for it is an improvement, but I suspect this will drift out-of-sync over time. One of the issues with json is you can't put comments in it. Maybe another future improvement here would be to convert `secret.json` to a YAML or JS file. Then we could maintain a template file and move the docs to inline comments where they are more likely to be found/updated?


doc/server-secrets.md
```diff
@@ -0,0 +1,106 @@
...
+will have access to your private repositories.
+
+When a `gh_token` is specified, it is used in place of the Shields token
+rotation logic.
+
+[github rate limit]: https://developer.github.com/v3/#rate-limiting
+[personal access tokens]: https://github.com/settings/tokens
+
+* `gh_client_id`
+* `gh_client_secret`
```
chris48s: Am I right that these 2 keys are only relevant to the rotating token pool? If so, I think we can probably omit them from these docs (along with `shieldsIps`) as they aren't relevant to self-hosting users. If they are relevant to a wider audience, what do they do in that context?

doc/server-secrets.md
```diff
@@ -0,0 +1,106 @@
...
+The SymfonyInsight API requires authentication. To obtain a token,
+Create an account, sign in and obtain a uuid and token from your
+[account page](https://insight.sensiolabs.com/account).
+
+## SonarQube
+
+* `sonarqube_token`
+
+Provide a token to give your self-hosted Shields installation access to a
+private SonarQube instance.
```
chris48s: I've just kind of guessed here. I'm not really sure. Whereas with the other services, I was able to work out the purpose of the credentials from the code or by going back over old issues/PRs, this one was more difficult. I'm not really clear on whether this allows us to access a private instance, or private projects on a public instance, or a bit of both. Does anyone know? Also, it appears there are a number of different permissions we can assign to a token: https://docs.sonarqube.org/display/DEV/Web+API#WebAPI-Noteaboutapi/permissions Do we know which are needed to generate a token for this purpose?

services/jira/jira-issue.tester.js
```diff
@@ -162,26 +168,27 @@ t.create('blue-gray status color')
...
-      // Without this the request wouldn't match and the test would fail.
-      .basicAuth({
-        user,
-        pass,
-      })
-      .reply(200, {
-        fields: {
-          status: {
-            name: 'in progress',
+;[mockJiraCreds, mockLegacyJiraCreds].map(mockCreds => {
```
chris48s: Prettier doesn't want to let us have a blank line here :(

doc/server-secrets.md
calebcartwright: I don't know definitively but I've hooked a self-hosted Shields instance up to a SonarQube server using this token so here's what I've seen:

My experience with creating tokens for Sonar (both for self-hosted SonarQube servers and SonarCloud) has always been the same, I just create a token and then use it elsewhere. The process described at https://docs.sonarqube.org/latest/user-guide/user-token/ has always been my experience. Sonar doesn't ask what permissions/scopes to apply and I've never even seen anywhere I can change it.

The SonarQube server I connected had all of our projects pubic, but it did require first authenticating. Therefore my guess is that the token is used for both scenarios you described above (both for authentication in general, and by sonar for authorization when requesting info from a private project)

doc/server-secrets.md
```diff
@@ -0,0 +1,106 @@
...
+* `npm_token`
+
+[Generate an npm token][npm token] to give your self-hosted Shields
+installation access to private npm packages
+
+[npm token]: https://docs.npmjs.com/getting-started/working_with_tokens
+
+## SymfonyInsight (formerly Sensiolabs)
+
+* `sl_insight_userUuid`
```
calebcartwright: I was trying to decide whether to change these as part of #2572 but maybe this would be a good time. Thoughts? We could do something similar to the Jira cred tokens to support both if necessary

paulmelnikow:
> Maybe another future improvement here would be to convert `secret.json` to a YAML or JS file. Then we could maintain a template file and move the docs to inline comments where they are more likely to be found/updated?

👍

services/jira/jira-issue.tester.js
paulmelnikow: Is it this issue? prettier/prettier#1510

doc/server-secrets.md
paulmelnikow: They could be relevant to _large_ self-hosting installations which require more than 5k badges/min, though that is a fringe case. Probably it's fine to leave a little pointer here to the production docs..

services/bower/bower-license.service.js
```diff
@@ -45,11 +44,7 @@ module.exports = class BowerLicense extends LegacyService {
           json: true,
           uri: `https://libraries.io/api/bower/${repo}`,
         }
-        if (serverSecrets && serverSecrets.libraries_io_api_key) {
-          options.qs = {
-            api_key: serverSecrets.libraries_io_api_key,
-          }
-        }
```
paulmelnikow: This was added in #919 to avoid rate limits in production. What's the reasoning for removing it?

services/jira/jira-base.js
```diff
@@ -11,10 +11,19 @@ module.exports = class JiraBase extends BaseJsonService {
...
-    if (serverSecrets && serverSecrets.jira_username) {
+    if (
+      serverSecrets &&
+      (serverSecrets.jira_user || serverSecrets.jira_username)
+    ) {
+      /*
+      for legacy reasons we still allow
+      jira_username, jira_password
+      but prefer
+      jira_user, jira_pass
```
paulmelnikow: Is this change to bring these more in line with the others?

services/jira/jira-issue.tester.js
chris48s: yeah - looks like it

doc/server-secrets.md
chris48s: What do you think they should be changed to?

services/bower/bower-license.service.js
chris48s: This is explained in the commit message for 31c253f :

> The libraries.io docs claim you need to be authenticated to make any API request: https://libraries.io/api#authentication

> In practice we can call https://libraries.io/api/bower/jquery just fine with no token and based on chucking a load of requests at it and examining the `x-ratelimit-remaining` headers you actually seem to get a better limit with no authentication.

> All of our libraries.io badges in `services/librariesio` seem to have been running fine with no token for some time.

I'm not exactly sure how quickly its resetting the limit, but with no auth you seem to get 30 new requests every second (or possibly some interval less than a second) whereas with a token you're restricted to only 60 requests per minute (the documented limit). I know the docs say you must authenticate, but you can check it with `curl -I`.

services/jira/jira-base.js
chris48s: Yep - as noted in the commit message for 523ba3f :


> All the other services use servicename_user, servicename_pass

> This switches JIRA to use that convention by preference but supports _username and _password for legacy users.


While I was looking at all of them in one place, it seemed like a good idea to bring JIRA into line, but I'm not married to that decision. If we think this is more hassle than its worth, I can back out that change and we can just document that the keys are `bintray_user`, `jenkins_user`, `nexus_user` but `jira_username`.

chris48s:
> Could danger be configured to help with doc drift?

Good idea - done in c195c4c

doc/server-secrets.md
calebcartwright: I'd suggest `symfony_insight_userUUid` and `symfony_insight_password`

doc/server-secrets.md
```diff
@@ -0,0 +1,111 @@
...
+
+[github rate limit]: https://developer.github.com/v3/#rate-limiting
+[personal access tokens]: https://github.com/settings/tokens
+
+* `gh_client_id`
+* `gh_client_secret`
+
+These settings are used by shields.io for GitHub OAuth app authorization
+but will not be necessary for most self-hosted installations. See
+[production-hosting.md](./production-hosting.md).
```
paulmelnikow: 👍

doc/server-secrets.md
chris48s: oh yes. Definitely makes sense to replace the `sl` prefix. Is it definitely a _password_ not a _token_ ?

chris48s:
Good point on sentry. I've added a note here but I think it still makes sense to primarily cover it in the self-hosting docs. There are some oddities around exactly what config goes in `secret.json` and what is an env var etc but lets track that in #2621

Issue #2420, opened 2018-11-29
Parse GitHub URL to create Travis badge


techtonik:
This is the current form for Travis CI badge.

![image](https://user-images.githubusercontent.com/515889/49199310-38ee5980-f3a8-11e8-9a15-f5dbf0c3b2ec.png)

And editing it to point to Travis CI is cumbersome. It will be more intuitive if the interface could parse the link automatically. I am willing to do this, but I need some guidance as I don't know React at all, and a pointer to the place of the code where I can change this behavior.

paulmelnikow:
Hi! If I understand this request correctly, it's a refinement of #701. Rather than individual fields, there would be a place to paste the GitHub URL. Nice idea!

Also worth doing, and easier to tackle quickly, is to add Travis to the badge suggestions. That would be done in the [suggestion API](https://github.com/badges/shields/blob/master/lib/suggest.js). The way a user would access it is by pasting a github URL into the search/suggest box and clicking **Suggest Badges**. This is momentarily blocked on #2441 which is modifying that endpoint, though you could grab that branch and start playing with the suggest feature if you'd like.


There's a bunch of work underway for the general case (#701) which I think needs to land first. Then there probably needs to be a flag in the service definition for services whose URLs can be inferred from a GitHub URL, and new some code in the front end to do that.

I'd suggest reading #2441 as a good place to start. Feel free to post comments and questions.

techtonik:
Looks like Suggest Badges is broken on FF. If I paste https://github.com/badges/shields/ no suggestions are proposed.

Still trying to get the time for React.

techtonik:
I still have no idea how

![image](https://user-images.githubusercontent.com/515889/64257423-0290cc80-cf2e-11e9-8129-a46c19a3e006.png)

maps to https://github.com/badges/shields/blob/master/services/travis/travis-build.service.js

Inspecting element doesn't help, because it is generated HTML.

paulmelnikow:
Would be nice to include a link to the source code! You could open a new issue for that if you like.

PyvesB:
I don't think this is relevant anymore nowadays, we've got a simple interface to create Travis.com badges:
<img width="681" height="387" alt="Image" src="https://github.com/user-attachments/assets/c55ad233-8354-41da-b503-8c79688c17c5" />

Will go ahead and close.

PR #2604, opened 2018-12-28
Fix [suggest]

paulmelnikow:
Ref https://github.com/badges/shields/issues/2420#issuecomment-450400904

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -195,7 +195,7 @@ module.exports = class Server {
     githubConstellation.initialize(camp)
     metrics.initialize(camp)
 
-    const { githubApiProvider } = this.githubConstellation
+    const { apiProvider: githubApiProvider } = this.githubConstellation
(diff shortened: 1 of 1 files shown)


techtonik:
Is it possible to have a smoke test for this functionality?

paulmelnikow:
It gets tested as part of the service test run. The test was failing, but hadn’t been picked up on yet.

techtonik:
I don't see any failures in https://github.com/badges/shields/commits/master - is service test run is a manual process?

techtonik:
> We can’t run all the tests on every commit

Why? It is only 4 minutes in the daily log.

paulmelnikow:
Our PR build runs in about 20 seconds; 4 minutes is much too slow. The other issue is that the service tests rely on external services. #927 is the discussion about that. We used to run the full suite on every commit to master, but stopped because they often fail for reasons outside our control. The thing we could use more help with is keeping up with fixing the failures identified tests (i.e. #1359).

PR #2612, opened 2019-01-01
Add timeout for server test setup/teardown

### LINKED ISSUE #2611: Investigate consistent server test failures in nightly job
I noticed that there's been an error in the daily job for a while now related to the service tests. Specifically it seems like the suite setup hook is exceeding the 2 second timeout. Link to the job [here](https://circleci.com/gh/badges/daily-tests/52) and log output below  

```sh
  1) The server
       "before all" hook: Start the server:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/project/shields/lib/server.spec.js)
```sh

calebcartwright:
Fixes #2611. I added a 10 second to the `before` and `teardown` functions directly so only those functions will be impacted (timeout period for the tests remains the default 2 seconds)

### DIFF
--- a/lib/server.spec.js
+++ b/lib/server.spec.js
@@ -14,6 +14,7 @@ const { createTestServer } = require('./in-process-server-test-helpers')
 describe('The server', function() {
   let server, baseUrl
   before('Start the server', async function() {
+    // Fixes https://github.com/badges/shields/issues/2611
     this.timeout(10000)
     const port = await portfinder.getPortPromise()
     server = createTestServer({ port })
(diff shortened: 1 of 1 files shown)


paulmelnikow:
I've noticed this too. Thank you for picking it up!

I suspect the cause of the timeout is the time it takes to parse and load all the services. What do you think about adding just the one timer to the setup?

Also could you add a comment linking to the issue?

calebcartwright:
> What do you think about adding just the one timer to the setup?

I'm good either way. Given the consistent slowness I figured it would be worth playing it safe on setup and teardown. I'll remove it for now and I suppose we can add the teardown back if it becomes an issue
 
> Also could you add a comment linking to the issue?

👍

PR #2617, opened 2019-01-02
redirectUrl: Update terminology and add a test

paulmelnikow:


paulmelnikow:
We've been standardizing on "url", so yea, `baseUri` should get changed too.

Changing the environment variable names is fine too, though I want to get #2621 resolved before messing with them too much.

PR #2620, opened 2019-01-02
Add [Ansible] quality score badge

### LINKED ISSUE #2602: Badge Request: Ansible Galaxy Quality
:clipboard: **Description**

Ansible Galaxy is now scanning roles being uploaded for quality.

Grabbing a random example: 
Project: https://galaxy.ansible.com/geerlingguy/php
API: https://galaxy.ansible.com/api/v1/content/432/

| Site Title     | Site Score | API Query      | API Result |
|----------------|------------|----------------|------------|
| Quality Score  | 4.1        | quality_score  | 4.125      |
| Syntax Score   | 3.3        | content_score  | 3.25       |
| Metadata Score | 5          | metadata_score | 5.0        |

:link: **Data**

https://galaxy.ansible.com/api/v1/content/432/

Antoine38660:
Adds quality score badge from Ansible, closes #2602


calebcartwright:
Thanks so much for your contribution to shields! It looks like two of the tests are failing, can you take a look? 
https://circleci.com/gh/badges/shields/31712?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

services/ansible/ansible-quality.service.js
```diff
@@ -0,0 +1,63 @@
...
+const ansibleContentSchema = Joi.object({
+  quality_score: Joi.number().required(),
+}).required()
+
+class AnsibleGalaxyContent extends BaseJsonService {
+  async fetch({ projectId }) {
+    const url = `https://galaxy.ansible.com/api/v1/content/${projectId}/`
+    return this._requestJson({
+      url,
+      schema: ansibleContentSchema,
```
calebcartwright: Do we know what sort of HTTP error response codes we could get from this API? For example what happens on unknown/invalid project id, etc. If we know we'll get back certain responses (404 for example) in certain situations then we can add that to the request call:

```js
schema: ansibleContentSchema,
errorMessages: {
  404: 'project not found',
  403: 'foo',
}
```

services/ansible/ansible.tester.js
```diff
@@ -26,3 +28,13 @@ t.create('role downloads (valid)')
 t.create('role downloads (not found)')
   .get('/role/d/does-not-exist.json')
   .expectJSON({ name: 'role downloads', value: 'not found' })
+
+t.create('quality score (valid)')
```
calebcartwright: Since you've split the original service file into `roles` and `quality` files, can you do the same for the tester files (at least for `quality` via a `ansible-quality.tester.js`?

services/ansible/ansible-quality.service.js
```diff
@@ -0,0 +1,63 @@
...
+    const json = await this.fetch({ projectId })
+    return this.constructor.render({ qualityScore: json.quality_score })
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'quality' }
+  }
+
+  static get category() {
+    return 'other'
```
calebcartwright: Also couldn't this be placed in the `quality` category? That seems like a more natural fit

services/ansible/ansible-quality.service.js
chris48s: There are no custom error messages defined on the existing ansible badges: https://github.com/badges/shields/blob/d96d8ae193213ff6f6166071fde1a0638bed2633/services/ansible/ansible.service.js#L19-L27

Unless there's some specific condition we're expecting to handle on this endpoint, its OK to stick with the default error handling.

services/ansible/ansible-quality.service.js
```diff
@@ -0,0 +1,63 @@
...
+      pattern: ':projectId',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: `Ansible Quality Score`,
+        pattern: ':projectId',
+        exampleUrl: ':projectId',
```
chris48s: `pattern` should be a placeholder, but we use `exampleUrl` to give an example of a value that will render a real badge. Can we replace this placeholder with an example (e.g: `29226`)

services/ansible/ansible.tester.js
```diff
@@ -4,6 +4,8 @@ const Joi = require('joi')
 const ServiceTester = require('../service-tester')
 const { isMetric } = require('../test-validators')
 
+const isQualityScore = Joi.string().regex(/^[0-9]+\.[0-9]/)
```
chris48s: Any reason we're validating the input with `number()` but then using a regex in the test?

services/ansible/ansible-quality.service.js
calebcartwright: Works for me. I mentioned it because the corresponding badge(s) `quality | not found` wasn't really clear to me, but I'm not a user of the service

services/ansible/ansible-quality.service.js
```diff
@@ -0,0 +1,63 @@
+'use strict'
+
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
+const { floorCount } = require('../../lib/color-formatters')
+
+const ansibleContentSchema = Joi.object({
+  quality_score: Joi.number().required(),
```
chris48s: It seems like some roles have a `quality_score` value of `null` e.g: https://galaxy.ansible.com/api/v1/content/24559/
At the moment we'll error with "invalid response data" in that case.. do we know what a `null` score indicates? If so, would it be better to allow null here and then throw an exception to provide a friendlier error message in that case? I'm genuinely undecided on this.. maybe "invalid response data" is fine?

services/ansible/ansible.tester.js
Antoine38660: Thanks for your feedbacks @calebcartwright. I started to split the tester files, but there are two services in the `ansible-roles.service.js` (`AnsibleGalaxyRoleDownloads` and `AnsibleGalaxyRoleName`). Should I also split this service file in two files? And of course split testers.

I ask this question because I tried to use the 
`const t = (module.exports = require('../create-service-tester')())` declaration and compiler throw me an error.

services/ansible/ansible-quality.service.js
Antoine38660: Maybe @calvinbui you can help us to clarify the best message as you open the issue?

Antoine38660:
Thanks a lot for your feedbacks and goodwill @calebcartwright & @chris48s 😄

services/ansible/ansible-quality.service.js
calvinbui: `No Score Available` like this one here: https://galaxy.ansible.com/s2504s/datadog

calvinbui:
Thanks for taking this on @Antoine38660 :grin:

services/ansible/ansible.tester.js
calebcartwright: Happy to help!

> compiler throw me an error

I'm not sure what you mean there. Are you getting an error when you try to launch the tests, or are the tests running but failing with an error message?

chris48s: The `create-service-tester` shortcut assumes a file exports only one service. If your file exports more than one service then you need to create a new `ServiceTester` object for it.

One way to do it is have `ansible-roles.service.js` export 2 services and we create the tester object using  `const t = (module.exports = new ServiceTester({ id: 'ansible', title: 'Ansible Galaxy Role' }))`

The other way would be to move the shared code to a helper file and split into 2 service files each with a corresponding tester then make the tester object using `const t = (module.exports = require('../create-service-tester')())` and naming convention.

I don't have a strong preference either way really.

services/ansible/ansible-quality.service.js
chris48s: Seems like `No Score Available` is the sensible error here. Lets accept `null` in the schema. If its `null`, throw an `InvalidResponse` with that message https://github.com/badges/shields/blob/5233a25ddcee9cc14eb8b67329aa019af613325e/services/errors.js#L40

paulmelnikow:
@calebcartwright It looks like this is ready for another review?

paulmelnikow:
@Antoine38660 Could you merge master into your branch and resolve the conflicts?

calebcartwright:
@paulmelnikow - I believe there's still a few pending changes (schema updates to cover null score scenario, splitting out tests, etc.)

services/ansible/ansible.tester.js
```diff
@@ -26,3 +26,11 @@ t.create('role downloads (valid)')
 t.create('role downloads (not found)')
   .get('/role/d/does-not-exist.json')
   .expectJSON({ name: 'role downloads', value: 'not found' })
+
+t.create('quality score (valid)')
+  .get('/score/quality/432.json')
+  .expectJSONTypes(Joi.object().keys({ name: 'quality', value: Joi.number() }))
+
+t.create('quality score (not found)')
```
calebcartwright: I thought we were changing the message here to be `No score available`?

services/ansible/ansible-quality.service.js
```diff
@@ -0,0 +1,76 @@
...
+      color: floorCount(qualityScore, 2, 3, 4),
+    }
+  }
+
+  async handle({ projectId }) {
+    const json = await this.fetch({ projectId })
+
+    const score = json.quality_score
+    console.log(score)
+    if (!score)
```
calebcartwright: Since this is a falsy check, we will throw the error if the score exists but has a value of 0. I'm not sure what the score scale is, but Is that the desired behavior? is a score of `0` a valid/possibile score? Or does a score of 0 mean that there is no score?

paulmelnikow:
Hey @Antoine38660 would you have a moment to respond to respond to the last comments? It would be great to get this shipped! 🚀

PR #2626, opened 2019-01-04
PaaS, CI, and production-friendly config; affects [jira]

### LINKED ISSUE #2621: Consistent way of managing config including secrets
Currently we have three ways of loading secrets:

1. In Heroku, other PaaS, and CI, from dynamic `private/secret.json` generated in a npm task at build time or runtime
2. In Docker, from dynamic `private/secret.json` via `secret.tpl.json` generated by the Dockerfile at build time
3. In the current / legacy production deploy, from checked-in `private/secret.json` generated in a deploy comment via the Makefile

I'd like to fold these three methods into a single method that covers on-premise, PaaS, and the current / legacy production deployment environment. To help solve #1848, this should not require a deploy commit.

See discussions: https://github.com/badges/shields/pull/2599#issuecomment-450378504 https://github.com/badges/shields/pull/2598#discussion_r244448146

Now that there is a config schema in `lib/server.js`, the server secrets could be folded into that. (However I want to make sure we do not accidentally leak credentials into log messages and error messages. So let's be careful that our config method doesn't do that.)

[node-config](https://github.com/lorenwest/node-config) is a great tool I've used in the past for cascading config. It lets you check in default configs which can be overridden by other checked-in configs for development, staging, production, or other environments (like self-hosting). You can create local config files which override those, and override anything else through environment variables. Here's a bit of explanation about [how the cascade works](https://github.com/lorenwest/node-config/wiki/Configuration-Files).

Cascading config would replace the logic in `lib/server-config.js`, exposing a config object which could then be validated by Joi.

We could set `NODE_CONFIG_ENV` to values like `shields-io-staging` and `shields-io-public` which would then cause them to draw on additional config files which can be checked in and managed in this repo.

For the current / legacy production deployment, we still need a way to set `NODE_CONFIG_ENV` (#1806). The non-secret parts of the config can be checked into the main repo. For the secrets there are two options: (1) a checked-in `local-shields-io-public.yml` at deploy time, and (2) using additional environment variables.

There's a related proposal at https://github.com/badges/shields/issues/2577#issuecomment-449675068 about fetching the secrets from a secure token store for the legacy production deploy. However that proposal is adjacent to this one. That's about where the secrets are loaded at deploy time; this is about how they are loaded at runtime.

nconf looks like an alternative to node-config.

paulmelnikow:
This implements the configuration mechanism I described in #2621. The heavy lifting is delegated to [node-config](https://github.com/lorenwest/node-config) with a minor assist from [dotenv](https://github.com/motdotla/dotenv).

One open is how we deal with Docker. In the old version, the secrets are baked into the image. Now that they can be read from the environment at runtime, it probably would be better to do that. Alternatively we could pull them from the build-time environment into the image. Or both! @calebcartwright perhaps you have thoughts about how we could deal with that.

Some tasks pre-merge:

- [x] Set `GH_TOKEN` in staging and CI (it was `GITHUB_TOKEN`)
- [x] Sort out docker
- [x] Update production deploy task, adding `.env` setting `NODE_CONFIG_ENV`

Some housekeeping post-merge:

- [ ] Prepare the production config file and do the first deploy
- [ ] Start a changelog for self hosting
- [ ] Inject secrets via the server
- [ ] Inject config into `trace`

Close #1806 #2621


doc/server-secrets.md
```diff
@@ -4,26 +4,33 @@ It is possible to provide a token or credentials for a number of external
 services. These may be used to lift a rate limit or provide access to
 private resources from a self-hosted instance.
 
-Secrets can be set in  `private/secret.json`. For example:
+There are two ways of setting secrets:
```
calebcartwright: I missed this in #2599 but the Azure DevOps token should be added to this list (and corresponding config locations).

calebcartwright: The token name is `azure_devops_token`

calebcartwright: I can update this doc with the relevant info/docs if you'd like, but not sure where all else the token would need to be added in the other changed files.

paulmelnikow: That'd be great! Besides the doc, it needs to be added to the config schema in `lib/server.js` and have an entry added with the environment variable mapping in `custom-environment-variables.yml`.

Probably _that_ ^^ should be documented somewhere too.

paulmelnikow: Definitely! Thanks

calebcartwright: Done! I feel pretty confident the Azure DevOps content in `server-secrets.md` will make sense to Azure DevOps users, hope it's clear enough for new users too

config/local.template.yml
```diff
@@ -0,0 +1,5 @@
+# Copy this file to `config/local.yml` and fill it in! It will be gitignored.
```
chris48s: lets link to doc/server-secrets.md here

doc/self-hosting.md
```diff
@@ -123,7 +123,7 @@ Redis installation.
 Server secrets
 --------------
 
-You can add your own server secrets in `private/secret.json`.
+You can add your own server secrets in environment variables or `config/local.yml`.
```
chris48s: One comment I'd have here is that if you're a self-hosting user and you've got all your secrets set up in `private/secret.json`, upgrading to this revision or later is going to break your install :wrench:

I don't really have a clear idea how self-hosting users are managing their installs, but is there a way we can communicate this change well to a self-hosting user? As you've already flagged, one tool available is documentation, although maintaining a changelog when we don't really do releases/tags is a challenge. Maybe we should be doing releases too :thinking: . Perhaps we could also do something like in the bootstrap process could we check for the existence of `private/secret.json` and if it exists throw an exception with a message saying "you need to move your config to env vars/yaml"?

Also, if everyone has to re-write their config file anyway, maybe we should just drop this https://github.com/badges/shields/blob/5233a25ddcee9cc14eb8b67329aa019af613325e/services/jira/jira-base.js#L14-L25 in this PR as there isn't really any backwards-compatibility to preserve at this point.

paulmelnikow: > I don't really have a clear idea how self-hosting users are managing their installs, but is there a way we can communicate this change well to a self-hosting user? As you've already flagged, one tool available is documentation, although maintaining a changelog when we don't really do releases/tags is a challenge. Maybe we should be doing releases too 🤔 . Perhaps we could also do something like in the bootstrap process could we check for the existence of `private/secret.json` and if it exists throw an exception with a message saying "you need to move your config to env vars/yaml"?

Yea, I'm a big fan of failing loudly. That's what's happening here: https://github.com/badges/shields/pull/2626/files#diff-e2e5bd472244fbfb4a40d38300d42534R6

This PR should make it possible for PaaS users to use the shields repo as-is, without forking or checking in any files. If we can solve that for Docker users, even better. Maybe we can publish an official docker image. Regardless, tagging releases probably would be a good idea.

Besides failing loudly and linking to a changelog, I'm not sure we have any other way to communicate this. The changelog will provide this answer going forward, though: it gives a single place to look when you're updating your self-hosted install.

I'd like to provide a second path for self-hosting someday, which is for advanced users who want to do something custom. We could ship an npm package with the server logic. Then the user can create their own project and use `shields-server` as a dependency. They can add their own services, and maybe even pick and choose which shields services should be enabled. This server package would be an ordinary library, and could follow semver. (Obviously this is not going to be tackled here.)

chris48s: > Yea, I'm a big fan of failing loudly. That's what's happening here: https://github.com/badges/shields/pull/2626/files#diff-e2e5bd472244fbfb4a40d38300d42534R6

oh right yes sorry - you've already done that exact thing and I missed it :man_facepalming:

calebcartwright:
> Or both! @calebcartwright perhaps you have thoughts about how we could deal with that 

Sorry I just noticed this 😄 At my day job we're running a self-hosted Shields instance (using the Docker image) within our network to connect it to some private tool instances. It's been several weeks since I've re-deployed, but I've noted some things below that we were doing (not speaking to whether or not they are needed/helpful/optimal/etc., just what we have been doing):

* The first part of our deployment process for our self-hosted Shields instance would grab the code and then:
  * inject some things into the Docker image (proxy stuff so the npm install would work)
  * overwrite the `lib/server-secrets.js` with a tweaked version that would pull the tokens we use from env variables instead of from `private/secret.json`
  * build and publish the Docker image to our internal Docker registry
* Next we'd deploy to an internal host by pulling the image and starting the container, which included pulling our token/cred values from the token store and pumping those token values into the environment variables for the container

Needless to say, the changes in this PR will allow us to get the same result much more easily! 💯 

I elected to inject the secrets at run time and not bake our secrets into the image itself, as doing so meant that anyone else who pulled that Docker image from our registry would be able to grab those secrets/creds (which we did not want). It may be preferable in other circumstances though to bake them into the image though so flexibility is always good when feasible IMO

paulmelnikow:
> I elected to inject the secrets at run time and not bake our secrets into the image itself, because if we had then anyone else who pulled that Docker image from our registry would be able to grab those secrets/creds (which we did not want). It may be preferable in other circumstances though to bake them into the image though so flexibility is always good when feasible IMO

It seems way more secure to deal with them at runtime, plus it's more consistent with the way they'd be handled in PaaS or other environments. Maybe we can leave good enough alone and see if someone comes by needing that.

calebcartwright:
Agreed. Runtime injection offers a lot more portability IMO too (especially if you want or need to use different tokens/creds for staging, prod, ....)

app.json
```diff
@@ -18,6 +13,10 @@
     "WHEELMAP_TOKEN": {
```
calebcartwright: What all is this list supposed to include? Since Wheelmap is in here I'm wondering if this enumerates tokens that are required for certain services and if Symfony/Sensiolabs Insight should be included?

doc/server-secrets.md
```diff
@@ -40,25 +62,25 @@ rotation logic.
...
 
 Provide a username and password to give your self-hosted Shields installation
 access to a private Jenkins CI instance.
 
 ## JIRA
 
-* `jira_user`
-* `jira_pass`
+* `JIRA_USER` (yml: `jira_user`)
+* `JIRA_PASS` (yml: `jira_pass`)
```
calebcartwright: I think I've confused myself as to whether the legacy format `jira_username` is still supported/if we're planning to support it 🤔

app.json
paulmelnikow: The only thing I know the placeholder entries are needed for is review apps. The env vars mentioned here get copied from the parent app (`shields-staging`) when a review app is deployed. In that sense they're only needed if you're self-hosting review apps.

Once you have a Heroku app you can always configure more settings for it.

lib/server-secrets.js
```diff
@@ -1,22 +1,15 @@
...
-    module.exports = require(secretsPath)
-  } catch (e) {
-    console.error(`Error loading secret data: ${e.message}`)
-    process.exit(1)
-  }
-} else {
-  console.warn(
-    `No secret data found at ${secretsPath} (see lib/server-secrets.js)`
+const legacySecretsPath = path.join(__dirname, '..', 'private', 'secret.json')
+if (fs.existsSync(legacySecretsPath)) {
```
calebcartwright: Is there a reasonable possibility that a self-hoster could have moved to the new model (added env vars and/or yml config file) already but left `private/secret.json` behind? If so, maybe the message could include something about detecting `private/secret.json` and that the user needs to delete it?

lib/server.js
```diff
@@ -20,34 +21,29 @@ const { clearRegularUpdateCache } = require('./regular-update')
...
+      Joi.string()
+        .hostname()
+        .required()
+    ),
   },
   metrics: {
     prometheus: {
       enabled: Joi.boolean().required(),
-      allowedIps: Joi.string(),
+      allowedIps: Joi.array().required(),
```
calebcartwright: Is it worth adding the `Joi.string().ip()` validation on the individual array items?

lib/server-secrets.js
paulmelnikow: Want to go ahead and edit it?

calebcartwright: Will do

lib/server.js
calebcartwright: Done

app.json
calebcartwright: 👍

doc/server-secrets.md
chris48s: If we're going to drop it, now is a good time since we're breaking BC anyway

paulmelnikow:
@chris48s Do you want to have another look at this?

chris48s:
Looks great. The only outstanding point from me is lets remove the block that references `jira_username` and `jira_password`. I think `privateConfigSchema` is filtering them out anyway now

paulmelnikow:
This pull request **introduces 1 alert** when merging aa3c0fdddde2344ee928f8b577523a8a30b0ff05 into 566b58acbad0e4ba721bd77beeebe20cb6858cb9 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-5c071f6e2f1d662a3f335f1f12d3bf040ca9d6f4)

**new alerts:**

---

Issue #2357, opened 2018-11-18
Improve handling of query params in examples


paulmelnikow:
In https://github.com/badges/shields/pull/2339#discussion_r234446985 @chris48s suggested a nice improvement to the way we handle query params in example URLs:

> pattern: `service/v/:package?branch=:branch`
> example: `service/v/mypackage?branch=master`

Constructing URLs with the colons requires a little bit of work. `queryString` will URL-encode the colons, so that substitution may have to be made after the fact. Or we can build the query string ourselves. (My least favorite kind of code!)

Also, we'll need to at least clarify the semantics of `query` in examples. Will the keys of the `query` object determine which querystring placeholders are included in the composed pattern?

Semi-related: `examples` use `query` while `route` uses `queryParams`. To avoid programmer errors we should pick one and make them the same.

chris48s:
We have just launched a completely redesigned frontend https://shields.io/
Closing this issue as it relates to the previous frontend

PR #431, opened 2015-04-25
Display the clojar name in clojure fashion

dotemacs:
Instead of the current:

v + clojar-version number: v1.1.2

e.g.: [pdfboxing](http://clojars.org/pdfboxing)

use full name and version (which can be cut & pasted and included in a project): [clojar-name "1.1.2"]

e.g. ["clojars"](https://clojars.org/pdfboxing) 

As per discussion: https://github.com/ato/clojars-web/pull/315#issuecomment-94726248

PR #2633, opened 2019-01-04
Convert a bunch of `exampleUrl`s and formats [ansible bitrise circleci clojars codacy codeclimate codecov depfu discord]

paulmelnikow:
In addition to the service tests, these should be verified in the UI.


services/bitrise/bitrise.service.js
```diff
@@ -24,9 +24,14 @@ module.exports = class Bitrise extends LegacyService {
     return [
       {
         title: 'Bitrise',
-        previewUrl: 'cde737473028420d/master?token=GCIdEzacE4GW32jLVrZb7A',
-        pattern: ':app-id/:branch?token=:app-status-badge-token',
-        exampleUrl: 'cde737473028420d/master?token=GCIdEzacE4GW32jLVrZb7A',
+        pattern: ':appId/:branch',
+        namedParams: { appId: 'cde737473028420d', branch: 'master' },
+        queryParams: { token: 'GCIdEzacE4GW32jLVrZb7A' },
```
calebcartwright: I noticed in the review app that the path is displaying as:

`/bitrise/:appId/:branch.svg?token=GCIdEzacE4GW32jLVrZb7A` 

With the token query param value plugged in directly. Is that the expected behavior for query params? Is it possible to still have that display as `/bitrise/:appId/:branch.svg?token=:token` ?

services/clojars/clojars-downloads.service.js
```diff
@@ -44,16 +44,15 @@ module.exports = class Clojars extends BaseJsonService {
   static get route() {
     return {
       base: 'clojars/dt',
-      format: '(.+)',
-      capture: ['clojar'],
+      pattern: ':clojar',
```
calebcartwright: Should the pattern be `:clojar(.+)` here to support [group/artifact targets](https://github.com/clojars/clojars-web/wiki/Data#artifact-httpsclojarsorgapiartifactsgroup_nameartifact_name)?

I believe it's possible that folks are using the badge with something like `/clojars/dt/cljsjs/react.svg`:

https://img.shields.io/clojars/dt/cljsjs/react.svg
 

Which wouldn't work with this pattern:
https://shields-staging-pr-2633.herokuapp.com/clojars/dt/cljsjs/react.svg

services/clojars/clojars-version.service.js
```diff
@@ -50,16 +50,14 @@ module.exports = class Clojars extends BaseJsonService {
   static get route() {
     return {
       base: 'clojars/v',
-      format: '(.+)',
-      capture: ['clojar'],
+      pattern: ':clojar',
```
calebcartwright: Same comment/question here as clojars-downloads service

services/codecov/codecov.service.js
```diff
@@ -28,17 +28,35 @@ module.exports = class Codecov extends LegacyService {
...
         title: 'Codecov private',
-        previewUrl: 'github/codecov/example-python',
-        pattern: 'token/:token/github/codecov/example-python',
-        exampleUrl: 'token/My0A8VL917/github/codecov/example-python',
+        pattern: 'token/:token/:vcsName/:user/:repo/:branch',
+        namedParams: {
+          token: 'My0A8VL917',
+          vcsName: 'github',
+          user: 'codecov',
+          repo: 'example-python',
```
calebcartwright: should add `branch: 'master'` here i think since the pattern for this example includes `:branch`

services/bitrise/bitrise.service.js
paulmelnikow: Yea, it is for now. Not yet: see #2357.

services/clojars/clojars-downloads.service.js
paulmelnikow: Gotcha. Yea, I think actually `':clojar+'` is the succinct way of saying there can be a slash.

calebcartwright:
Just as an aside on the clojars-version service.. 

The behavior here is the same as before, but is the badge message here what we'd expect for a version badge?



I would've expected the message to be `1.1` vs. `[ prismic "1.1" ]`

paulmelnikow:
> The behavior here is the same as before, but is the badge message here what we'd expect for a version badge?

That badge has an unusual format. I don't use clojure though I suspect the badge message is in a format that can be pasted into a CLI or config file.

services/codecov/codecov.service.js
paulmelnikow: Good catch. I updated this to match the old pattern instead (no `branch`).

chris48s:
Clojars does have a legitimately unusual version format - see https://github.com/badges/shields/pull/431

PR #2644, opened 2019-01-06
Integrate new path-to-regexp with trailing optionals [appveyor azuredevops npm gitlab date githubmanifest githubpackage jenkins matrix readthedocs requires shippable snyk wordpress]

### LINKED ISSUE #2497: Route `pattern` doesn't support trailing optional

paulmelnikow:
Fix #2497

Ref pillarjs/path-to-regexp#176


calebcartwright:
> How do folks feel about merging this with the github link in place while we wait for a published version? 

Generally speaking I avoid doing so, but agree that if this is temporary (seems like that should be merged and published soon) then it is probably okay, especially considering the benefits of the trailing optional

doc/rewriting-services.md
```diff
@@ -89,10 +89,7 @@ still pass.
 the route using a `pattern`. A `pattern` (e.g. `pattern: ':param1/:param2'`) is
 the simplest way to declare the route, also the most readable, and will be
 useful for displaying a badge builder with fields in the front end and
-generating badge URLs programmatically. One limitation to keep in mind is that,
-at present, the trailing parameter of a pattern can't be optional. If the last
-part of a route is optional, like a branch, you will need to use a `format`
-regex string (e.g. `format: '([^/]+/[^/]+)'`).
+generating badge URLs programmatically.
```
calebcartwright: Do we have any guidance/tips on using `pattern` anywhere else? I've learned about various shortcuts/patterns that path-to-rexep` supports from places like comments folks have left on PRs that I didn't know about. 

If not maybe we could link out to https://github.com/pillarjs/path-to-regexp?

services/appveyor/appveyor-base.js
```diff
@@ -37,8 +37,7 @@ module.exports = class AppVeyorBase extends BaseJsonService {
   static buildRoute(base) {
     return {
       base,
-      format: '([^/]+/[^/]+)(?:/(.+))?',
-      capture: ['repo', 'branch'],
+      pattern: ':user/:repo/:branch?',
```
calebcartwright: 👍  so much more readable!

services/github/github-manifest.service.js
```diff
@@ -25,8 +25,7 @@ class GithubManifestVersion extends ConditionalGithubAuthService {
   static get route() {
     return {
       base: 'github/manifest-json/v',
-      format: '([^/]+)/([^/]+)/?([^/]+)?',
-      capture: ['user', 'repo', 'branch'],
+      pattern: ':user/:repo/:branch?',
```
calebcartwright: Sidenote, I'm seeing an `invalid response data` badge message for the example badge used here. Seeing this in prod too so not related to this PR.

https://img.shields.io/github/manifest-json/v/RedSparr0w/IndieGala-Helper.svg

services/github/github-package-json.service.js
```diff
@@ -23,8 +23,7 @@ class GithubPackageJsonVersion extends ConditionalGithubAuthService {
   static get route() {
     return {
       base: 'github/package-json/v',
-      format: '([^/]+)/([^/]+)/?([^/]+)?',
-      capture: ['user', 'repo', 'branch'],
+      pattern: ':user/:repo/:branch?',
```
calebcartwright: Ditto on example error here

services/jenkins/jenkins-plugin-installs.service.js
```diff
@@ -98,8 +98,7 @@ class JenkinsPluginInstalls extends BaseJsonService {
   static get route() {
     return {
       base: 'jenkins/plugin/i',
-      format: '([^/]+)/?([^/]+)?',
-      capture: ['plugin', 'version'],
+      pattern: ':plugin/:version?',
```
calebcartwright: Another tangential thought but could we update the second example title to include something like `version`?

services/wordpress/wordpress-downloads.service.js
```diff
@@ -89,16 +89,14 @@ function InstallsForExtensionType(extensionType) {
     static get route() {
       return {
         base: `wordpress/${extensionType}/installs`,
-        format: '(.+)',
-        capture: ['slug'],
+        pattern: ':slug',
```
calebcartwright: Will using the default pattern cause any issues with slugs that contain a `/`? I'm not sure what slugs look like/are allowed to contain but just wanted to double check given the previous expression

services/wordpress/wordpress-platform.service.js
```diff
@@ -114,8 +114,7 @@ class WordpressPluginTestedVersion extends BaseWordpressPlatform {
   static get route() {
     return {
       base: `wordpress/plugin/tested`,
-      format: '(.+)',
-      capture: ['slug'],
+      pattern: ':slug',
```
calebcartwright: Same question as wordpress-downloads service regarding pattern

services/wordpress/wordpress-downloads.service.js
paulmelnikow: I looked on the wordpress site and it doesn't look like they have slashes in them. It's possible digging further could have revealed some. If we go too restrictive, we'll find out and can add a test.

It occurred to me that _git branches can have slashes in them_. There are probably a lot of those which need updating in here…

doc/rewriting-services.md
paulmelnikow: I think that link is in the tutorial. What are the tips you'd give? Maybe we can start creating a new section.

paulmelnikow: One that comes to mind: you can't use non-capturing groups in a `pattern`.

services/github/github-package-json.service.js
paulmelnikow: Did you mean to add something like **(branch)** to the title? I've just gone ahead and done that.

doc/rewriting-services.md
calebcartwright: That's a good example of one. I'd tried to fight it to make non-capturing groups work (obviously unsuccessfully 😆 ), but I wasn't sure if it wouldn't work because it wasn't possible or if I just didn't have the syntax/escapes right. 

The other thing I just learned the other day are some of the shorthands like `*` and `+` can be added directly to the param (I think one example is `:foo*` works as the shorthand for `:foo(.*)` .

services/wordpress/wordpress-downloads.service.js
calebcartwright: Oh wow. Good catch, and yeah I think there will be a _lot_ of those patterns.

services/github/github-package-json.service.js
calebcartwright: Sorry I should've been more clear. I meant that with this badge is showing an `invalid response data` error with the example too (also occurs in prod so not related to these changes)

doc/rewriting-services.md
calebcartwright: I'm still thinking that some pattern tips _may_ be useful somewhere, but doesn't need to be here nor part of this PR IMO

services/wordpress/wordpress-downloads.service.js
calebcartwright: It looks like you got all of these already right?

paulmelnikow: Yes, I think so.

paulmelnikow:
Could you re-up this @calebcartwright? Every time a merge conflict has to be resolved manually the review gets dismissed.

PR #2646, opened 2019-01-06
refactor [bower] service

chris48s:


services/bower/bower.tester.js
```diff
@@ -56,21 +65,30 @@ t.create('custom label for pre version') // e.g. pre version|v0.2.5-alpha-rc-pre
...
-  .expectJSON({ name: 'license', value: 'invalid' })
+  .expectJSON({ name: 'license', value: 'not found' })
 
-t.create('Version label should be `no releases` if no official version')
+t.create('Version label should be `no releases` if no stable version')
   .get('/v/bootstrap.json')
   .intercept(nock =>
     nock('https://libraries.io')
       .get('/api/bower/bootstrap')
-      .reply(200, { latest_stable_release: { name: null } })
```
chris48s: I'm not 100% convinced `{ latest_stable_release: { name: null } }` is a real condition. I think you actually get `{ latest_stable_release: null }` . I suspect this was reverse-engineered from the code when the tests were written rather than based on a real API response. I've said in the schema that if `latest_stable_release` is an object, `name` is required. We'll render `invalid response data` if this does actually happen.

PR #2656, opened 2019-01-06
Set redirectUrl for production; rely on NODE_CONFIG_ENV being set earlier

paulmelnikow:


Makefile
```diff
@@ -28,9 +28,8 @@ prepare-server-deploy: website
 	cp -r build ${SERVER_TMP}
 	git -C ${SERVER_TMP} add -f build/
 	git -C ${SERVER_TMP} commit --no-verify -m '[DEPLOY] Add frontend for debugging'
-	echo 'NODE_CONFIG_ENV=shields-io-production' > ${SERVER_TMP}/.env
```
paulmelnikow: This setting moved to badges/ServerScript#3.

paulmelnikow:
> Looks good! As mentioned in the ServerScript changes, we should probably remove the need for the _environment_ environment variable in the future.

As I mentioned in chat, I am not sure how to reconcile that with the idea of an automatic cascade.

Perhaps there is another library which takes a different approach, maybe one that uses an explicit cascade instead of so much convention.

I do like a couple other things about `config`:

1. I can create a `local.yml` which stays on my machine.
2. I can combine production config between a file in the main repo (`shields-io-production.yml`) and a file in a deploy commit (`local-shields-io-production.yml`).

espadrine:
> I am not sure how to reconcile that with the idea of an automatic cascade.

That doesn't seem incompatible to me.

We can have s1.yml overriding shields-io.yml overriding default.yml, without relying on environment environment variables (nor environment environment files).

paulmelnikow:
How would it know which files to load?

PR #2660, opened 2019-01-06
[Travis-Build] service rewrite, run [travis-php-version]

PyvesB:
In addition to `failing`, the badge also now shows red for `error` status.


services/travis/travis-build.service.js
```diff
@@ -1,26 +1,25 @@
...
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class TravisBuild extends LegacyService {
+const BaseSvgScrapingService = require('../base-svg-scraping')
+
+const schema = Joi.object({
+  message: Joi.string()
+    .regex(/^[a-z]+$/)
```
calebcartwright: is there a smaller set of values we could test against? the tester is using the `isBuildStatus` from `test-validators.js`

calebcartwright: I mentioned it as I remembered seeing a similar pattern for the Azure DevOps builds (which is a badge scraping implementation) https://github.com/badges/shields/blob/master/services/azure-devops/azure-devops-helpers.js#L9 where there is validation.

I suppose it's a little philosophical, but IMO I'm not sure why we'd have prod validation for some services but not others. I think we're following the same pattern in other (JSON API based) services where the returned value is the source of truth, but we still have schema validation. 

For example if Travis were to change their status message from `passing` to `passed` tomorrow, what would happen to the badges in prod (in particular the color scheme)? Would the daily test run catch this to let us know?

I don't feel strongly about it either way, but if we are going to establish a practice of not doing validation in certain scenarios (for example badge scraping services) then we should document that somewhere.

PyvesB: To be honest I was not aware that we had cases where we do perform very strict validation in production, generally we don't seem to do that with build badges (Wercker, CircleCi, AppVeyor, BitbucketPipelines to name just a few).

If Travis were to change their status message from `passing` to `passed`, I would much rather have users reporting that the colour is wrong rather than having to do emergency fixing because the badge is completely broken and displays `inaccessible` or something like that. ^^

But you're right that we should try to be consistent, let's wait for someone else's opinion. 😉

paulmelnikow: I'm a fan of failing loudly because it means issues are more likely to be reported and fixed. I'm also a fan of adapting to specific circumstances, like starting with something strict and loosening it if need be. If we find we're deploying emergency patches for schema validations more than once, then we could switch tacks, leaving it loose, and basically hoping the colors are right.

paulmelnikow: That sounds like a good plan. I did not realize we were handling it so inconsistently.

PyvesB: I'll raise an issue once we're happy with this as it will help narrow down the work that has to be done.

I've provided stronger validation and tried to extract things as much as possible so that can easily be reused in different places. Let me know what you think!

calebcartwright: > I would much rather have users reporting that the colour is wrong rather than having to do emergency fixing because the badge is completely broken and displays inaccessible or something like that

That's a fair point too. Ultimately I suppose it just boils down to the tradeoffs between more/less strict validation. I'm good with whatever we decide 👍 but agree consistency should be our target. 

If I'm following correctly it sounds like we want to lean towards more strict but also reasonably support exceptions, for example if the validation becomes to brittle for a particular service?

lib/build-status.js
```diff
@@ -0,0 +1,45 @@
...
+  'stopped',
+  'timeout',
+  'waiting',
+]
+
+const isBuildStatus = Joi.equal(
+  happyStatuses.concat(unhappyStatuses).concat(otherStatuses)
+)
+
+function renderBuildStatusBadge({ label, status }) {
```
paulmelnikow: I like this idea!

In terms of reusability, should reformat `happyStatuses` to use our standard text? I know in a few places we do that kind of translation, i.e. display *passing* instead of *passed*. We could do the same thing for **failing** -> **failed**.

PyvesB: Done!

PR #2661, opened 2019-01-06
Document production hosting

paulmelnikow:
Please feel free to make edits! If there are things that should be covered here let me know and I'll try to add them.

Marked blocker: to avoid duplicated work, probably I will not pick up self-hosting changelog until after this is merged.


doc/production-hosting.md
```diff
@@ -1,40 +1,231 @@
...
+    `local-shields-io-production.yml` from the server (or obtain the current
+    version of that file some other way). Save it in `config/`.
+2.  Check out the commit you want to deploy.
+3.  Run `npm ci`. **This is super important for the frontend build!**
+4.  Run `make deploy-s0` to make a canary deploy.
+5.  Check the canary deploy:
+    - [Visit the server][s0]. Don't forget that most of the preview badges
+      are static!
+    - Look for errors in [Sentry][].
+    - Keep an eye on the [status page][].
```
calebcartwright: status page link is not showing for me (and I'm not sure what it should be)

doc/production-hosting.md
```diff
@@ -1,40 +1,231 @@
...
+
+## Logs
+
+Logs are available on the individual servers via SSH. @espadrine has access.
+
+## Error reporting
+
+[Error reporting][sentry] is one of the most useful tools we have for monitoring
+the server. It's generously donated by [Sentry][sentry home]. We bundle
+[`raven`][] into the application, and the Sentry DSN is configured via `local- shields-io-production.yml` (see [documentation][sentry configuration]).
```
calebcartwright: raven link not working here (think it is because of the `` )

doc/production-hosting.md
```diff
@@ -1,40 +1,231 @@
...
+
+## Frontend
+
+The frontend is served by [GitHub Pages][] via the [gh-pages branch][gh-pages]. SSL is enforced.
+
+`shields.io` resolves to the GitHub Pages hosts. It is not proxied through
+Cloudflare.
+
+Any maintainer can push to `gh-pages` though in practice this needs to be kept
+in sync with the production deploy.
```
calebcartwright: are there any rollback related steps with the frontend, especially as it relates to keeping in sync with a server/backend rollback?

doc/production-hosting.md
paulmelnikow: Fixed!

doc/production-hosting.md
paulmelnikow: Fixed!

doc/production-hosting.md
paulmelnikow: A rollback is the same as a regular deploy, so the process is exactly the same.

The most likely way a rollback happens is after a bad canary deploy.

Also worth mentioning: you can, though normally should not, deploy commits made to a local branch.

Want to update where needed to clarify that?

calebcartwright: > though in practice this needs to be kept
in sync with the production deploy

This is the part I was thinking through. My first time reading through from top to bottom I didn't have context on this yet. What if we swapped the order of the `Deployment` and the `Frontend` sections? 

Although I suppose it doesn't really matter, this doc has a very specific, target audience (not the masses) 😄

paulmelnikow: Another possibility is to consolidate all the "who has access to what" into a table at the end. We could qualify the point there. The reason to mention it is that while _technically_ anyone with push access to the repo can deploy the frontend, in practice they can't.

calebcartwright: I'm def 👍 💯 on that!

paulmelnikow: Do you think it makes sense to keep the order of sections where it is? It's tricky because you can't explain the deploy without explaining the infrastructure.

calebcartwright: Yeah the order makes sense now. I like the verbiage you added in the table too. What if we reused the sentiment of the text you added in the table down in the `Frontend` section? 

in the Frontend section, replace:
 `Any maintainer can push to gh-pages though in practice this needs to be kept in sync with the production deploy` 
With:
`Technically any maintainer can push to gh-pages, but in practice this must be deployed with the badge server via the deployment process described below`

paulmelnikow:
I'm sure there will be more things that need to be added to this doc, though it seems like a good start, and worth moving on to the next piece. Thanks for reviewing! It's a big help.

calebcartwright:
Is this ready for final approval? I'm 👍 on approving this whenever ready

PR #2672, opened 2019-01-07
Replace open collective with shields rendered badges

### LINKED ISSUE #2498: Remove third-party images from homepage
Shields has a no-tracking promise. I interpret that to mean we should not load third-party images from https://shields.io/.

At some point OpenCollective images snuck in:

screen shot 2018-12-09 at 10 16 04 pm

I like having them there. Maybe we could replace them with Shields-rendered badges.

mbrandau:
old
new

Closes #2498


mbrandau:
I certainly messed my fork up...

paulmelnikow:
We can squash. In the future probably best to work off branches in your fork :)

PR #2681, opened 2019-01-08
Fix frontend start script

paulmelnikow:
Without this change none of the badges load. Probably happened in #2626.


paulmelnikow:
Ah, maybe a Windows thing?

calebcartwright:
Could be, I haven't used my Ubuntu machine in a bit but I seem to remember it working on that one now that you mention it

PR #2682, opened 2019-01-08
Add category nav menu

paulmelnikow:
Adapted from @chris48s’ work from #1842.

We have a lot more categories than we did in #1842 so I changed the formatting a bit.


tooomm:
Jep, looks good and should be a big improvement 👍

PR #2698, opened 2019-01-08
[do not merge] Reorganize all the library and helper code

paulmelnikow:
As I’ve started writing documentation I realize I have an idea that “core” means the server and the base service framework. However it’s difficult to communicate that beacuse the code isn’t organized that way. There are still several chunks of legacy code which either need to stick around until after the refactor is done or be shimmed in the meantime. When a contributor is working on a new service, there’s not an easy place to look for functions they should use. We left the base service implementations close to the service, which makes them easy to import and inspect see, but they’ve grown and all their details have become so sprawling that the root of `services/` is probably not useful to first time contributors.

This tries to address that.

- “Good” and “pretty good” helper functions for services have been moved from `lib/` to `services/`
- “Bad” helper functions which haven’t been refactored have been moved from `lib/` to `core/legacy/`
- Everything related to the implementation of BaseService has moved to `core/base-service/` (except for a few common legacy bits which are in `/core/legacy`).
- The stuff the server needs is in `core/server`.
- The service tester and service test runner are in `/core/service-test-runner`.
- The token pooling code is in `core/token-pooling`.
- Shields badge URL functions are in `core/badge-urls`.
- `lib/` is gone.

This leaves a clear distinction between `core` and `services`. To keep the distinction clear, we should minimize the imports which cross this line.

Old:
```
const { floorCount: floorCountColor } = require('../../lib/color-formatters')
const { addv, metric } = require('../../lib/text-formatters')
const BaseJsonService = require('../base-json')
const { NotFound } = require('../errors')
```

New:
```js
const { floorCount: floorCountColor } = require('../color-formatters')
const { addv, metric } = require('../text-formatters')
```

Old:
```js
const t = (module.exports = require('../create-service-tester')())
```

New:
```js
const t = (module.exports = require('..').createServiceTester())
```

Looking forward to comments as always! If this seems like a good way to go I will wrap it up.


services/index.js
```diff
@@ -1,76 +1,11 @@
 'use strict'
 
-const glob = require('glob')
```
paulmelnikow: This was moved to `core/base-service/loader.js`.

core/base-service/loader.js
```diff
@@ -0,0 +1,78 @@
+'use strict'
```
paulmelnikow: This was moved from `services/index.js`.

paulmelnikow:
Any opinions on this arrangement? It wasn't 100% done though I changed enough of it that the idea can be seen.

Because this requires moving files and updating `require` statements it looks like it's going to slide out of date quickly. So we'll need to work together to agree on the change, and then make the edits and merge them.

PyvesB:
The new organisation makes sense to me. I think we should move forward with this idea. 😉

paulmelnikow:
@PyvesB Sounds good. Maybe we can use this PR as a reference and make the changes in a few waves.

chris48s:
> Because this requires moving files and updating require statements it looks like it's going to slide out of date quickly

Yeah, agreed. Also one PR that touches >100 files has so many moving parts it is very difficult to review. A PR that is hard to review will stay open for a bit and pick up a lot of merge conflicts.

> Maybe we can use this PR as a reference and make the changes in a few waves.

How about attacking this by setting up the structure, moving some files to new locations, updating the imports and then doing some follow-up PRs until `/lib` is empty. For example maybe each of `server`, `token-pooling`, `service-test-runner` etc is a self-contained PR?

paulmelnikow:
Yea, I think that's a good approach. I'd like to avoid touching hundreds of files multiple times, though, so we may need structure things so the code that touches many, many services is all moving at once.

> For example maybe each of `server`, `token-pooling`, `service-test-runner` etc is a self-contained PR?

That sounds good, though a few hours ago I did one that was slightly larger than that, which moved `server` and `service-test-runner`. See if that looks okay? If not I'll split it up.

paulmelnikow:
Do you think we should bother to move all the legacy stuff into `core/legacy`, or should we just leave it in `lib/`?

Kind of depends how quickly we think the refactor will be done 🤔

chris48s:
I think it makes it more obvious to new contributors that they shouldn't be adding new stuff to `/legacy`. `/lib` sounds like a legitimate place to be adding new code. We're making good progress, but I can see us still having legacy services in a few months time

calebcartwright:
This makes sense to me as well. I remember being a bit puzzled myself when I first started working on a service and I feel like this will be easier to grok. 

I assume that `core/legacy` will go away at some point right?

paulmelnikow:
> I assume that `core/legacy` will go away at some point right?

Let's hope! 😉

paulmelnikow:
Opened #2831 for the balance of this.

PR #2707, opened 2019-01-08
Update [opencollective] error handling

### LINKED ISSUE #2699: Update opencollective error handling
The error handling @mbrandau asked for in opencollective/opencollective#1609 was implemented!

We should update the tests and tweak the implementation if need be.

@mbrandau would you like to take care of it?

mbrandau:
Backers by tier badge now shows "unknown tier" as a label when the collective was not found
Closes #2699

### DIFF
--- a/services/opencollective/opencollective-by-tier.service.js
+++ b/services/opencollective/opencollective-by-tier.service.js
@@ -18,6 +18,12 @@ module.exports = class OpencollectiveByTier extends OpencollectiveBase {
     ]
   }
 
+  static get defaultBadgeData() {
+    return {
+      label: 'unknown tier',
+    }
+  }
+
   static get route() {
     return this.buildRoute('tier', true)
   }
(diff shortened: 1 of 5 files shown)


services/opencollective/opencollective-by-tier.service.js
```diff
@@ -18,6 +18,12 @@ module.exports = class OpencollectiveByTier extends OpencollectiveBase {
     ]
   }
 
+  static get defaultBadgeData() {
+    return {
+      label: 'unknown tier',
```
paulmelnikow: It's a bit confusing to default to an error message. Probably `'open collective'` would be better.

mbrandau: My intention was to show "unknown tier: collective not found" when the API returns a 404. If you prefer "open collective" to "unknown tier", that's fine with me.

paulmelnikow: Ah, I gotcha. Yea, I think that would be better. Thanks!

Issue #2722, opened 2019-01-10
Coverity Scan Service is Down


calebcartwright:
The Coverity Scan service has been offline for emergency maintenance for a couple days now. This outage means our Coverity badges are unavailable and it has also been causing our [daily builds to fail](https://circleci.com/gh/badges/daily-tests). 

There's not anything actionable we can do on our end, but opening this for awareness.

https://twitter.com/CoverityScan/status/1082446527493066752

paulmelnikow:
https://community.synopsys.com/s/article/Coverity-Scan-Update

Why don't we deprecate this, at least temporarily. Perhaps we could use a custom message like **coverity | extended downtime** to help clarify that the problem is not with the project.

calebcartwright:
Yeah I think that's probably a good idea considering. Should this be possible with the standard `deprecatedService` function (including with the `extended downtime` message) or will we need to do something different?

calebcartwright:
Sounds good, I'll take a look at doing this here shortly

calebcartwright:
I think Coverity is back online (at least in some capacity). I'll poke around and if it seems usable again I'll open a PR to un-deprecate the service

PR #2723, opened 2019-01-10
Split “quality” into “code coverage” and “analysis” [codeclimate scrutinizer sonar]

paulmelnikow:
With the menu in place I think having more categories is helping that process because it's grouping more similar things together. Given #2722, improving our discoverability in the analysis area may be particularly useful to developers right now.

Related: We may want to tighten up the menu on the home page so it doesn't flow off the screen, though I don't think that is blocking.


services/sonarqube/sonarqube.service.js
```diff
@@ -8,15 +8,9 @@ const {
...
-//
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class Sonarqube extends LegacyService {
+class SonarqubeCoverage extends LegacyService {
   static get category() {
-    return 'quality'
+    return 'coverate'
```
calebcartwright: typo here, should be `coverage`

paulmelnikow: Huh, I'm confused why the unit tests did not detect this.

PR #2726, opened 2019-01-10
limit the size of response we will accept

chris48s:
This allows us to set a limit on the size of the request body we'll accept, preventing dynamic badge users from using a excessive of resources by submitting huge files to us. The limit will also apply to our own badges, but this probably isn't a bad thing.

I think 1MB is a fairly sensible limit to set, but I'm open to feedback on that. I reckon the largest responses we routinely parse are probably going to be the NuGet XML responses e.g: https://www.powershellgallery.com/api/v2/Packages()?$filter=Id%20eq%20'ACMESharp'%20and%20IsLatestVersion%20eq%20true which are a few hundred kB but if anyone else has any ideas of other large responses we should check, suggestions are welcome. Hopefully we don't have any badges which require us to parse >1MB of data to report one value.

I had a look at writing a test for this in `request-handler.spec.js`, but I'm not sure how to do it without either making the test depend on an external resource (which is ok for `services`, but I'd rather avoid it with `lib`) or doing major refactoring. Any thoughts?


config/default.yml
```diff
@@ -34,4 +34,7 @@ public:
 
   handleInternalErrors: true
 
+  # 1,048,576 bytes = 1MB
+  maxResponseSizeBytes: 1048576
```
chris48s: I had a look to see if I can find a library that allows us to declare a nice string here like `maxResponseSize: "512kB"` or `maxResponseSize: "100MB"` and parse it into a number of bytes, but I couldn't find one. They all seem to go the other way (i.e: from bytes to human-readable notation). Any ideas?

paulmelnikow: Found one!

https://github.com/visionmedia/bytes.js#bytesparsestringnumber-value-numbernull

config/custom-environment-variables.yml
```diff
@@ -40,6 +40,8 @@ public:
 
   rateLimit: 'RATE_LIMIT'
 
+  maxResponseSizeBytes: 'MAX_RESPONSE_SIZE_BYTES'
```
paulmelnikow: After reading this code the effect of this setting was really clear, but before reading the code it wasn't.

How about calling it `fetchLimitBytes`?

lib/request-handler.js
```diff
@@ -181,6 +187,18 @@ function handleRequest(cacheHeaderConfig, handlerOptions) {
         }
         callback(err, res, body)
       })
+      r.on('data', chunk => {
+        buffer += chunk
+        if (buffer.length > config.public.maxResponseSizeBytes) {
```
paulmelnikow: Since AFAICT we don't care about the actual value of the `buffer`, could you accumulate the length instead of the contents to save memory?

lib/request-handler.js
```diff
@@ -189,9 +207,13 @@ function handleRequest(cacheHeaderConfig, handlerOptions) {
...
-            // Wrap the error in an Inaccessible so it can be identified
-            // by the BaseService handler.
-            reject(new Inaccessible({ underlyingError: err }))
+            if (err instanceof ShieldsRuntimeError) {
+              reject(err)
+            } else {
+              // Wrap the error in an Inaccessible so it can be identified
+              // by the BaseService handler.
+              reject(new Inaccessible({ underlyingError: err }))
+            }
```
paulmelnikow: 👍

paulmelnikow:
> I had a look at writing a test for this in `request-handler.spec.js`, but I'm not sure how to do it without either making the test depend on an external resource (which is ok for `services`, but I'd rather avoid it with `lib`) or doing major refactoring. Any thoughts?

I think the way to write a test is by using `nock` directly to provide a mock response. There's one example in https://github.com/badges/shields/blob/da12f00d87dad9f8fe1b17c10f7c399e34c09363/services/github/auth/acceptor.spec.js#L80-L86 though you can do something more simple here, [invoking `.reply()` with the string](https://github.com/nock/nock#specifying-replies).

Something like:

```js
const bigResponse = 'x'.repeat(10e7)

...

.reply(200, bigResponse)
```

paulmelnikow:
> I think 1MB is a fairly sensible limit to set, but I'm open to feedback on that. I reckon the largest responses we routinely parse are probably going to be the NuGet XML responses e.g: https://www.powershellgallery.com/api/v2/Packages()?$filter=Id%20eq%20'ACMESharp'%20and%20IsLatestVersion%20eq%20true which are a few hundred kB but if anyone else has any ideas of other large responses we should check, suggestions are welcome. Hopefully we don't have any badges which require us to parse >1MB of data to report one value.

Sad to say, I would guess we have a lot of badges which are upwards of 1 MB. I wouldn't be surprised if the npm endpoints were often larger, for example. I'm certain some of the `regularUpdate`s are also larger.

I do like the idea of a global option like this, though we may need a way to override it for those requests.

One approach to determine an appropriate number and identify exceptions could be an experiment that runs a full service test run, which logs the URLs and their response lengths.

chris48s:
Will have a look at these comments, but they all sound reasonable.

> Sad to say, I would guess we have a lot of badges which are upwards of 1 MB. I wouldn't be surprised if the npm endpoints were often larger, for example

Fair enough. I'll have a look at those. Another approach is just to set the limit at like 20MB or something - enough that any vaguely sensible response should be under it, but low enough that we don't try and parse a response which is so big its malicious/abusive.

paulmelnikow:
> Fair enough. I'll have a look at those. Another approach is just to set the limit at like 20MB or something - enough that any vaguely sensible response should be under it, but low enough that we don't try and parse a response which is so big its malicious/abusive.

I think 10 MB is probably safe, though if the service tests pass on 5 MB (doubtful) we could go with that. That seems like a fine way to get started with this. Let's not let the perfect be the enemy of the good.

lib/request-handler.spec.js
```diff
@@ -78,6 +93,46 @@ describe('The request handler', function() {
...
+        .get('/foo/bar')
+        .once()
+        .reply(200, 'x'.repeat(101))
+      const res = await fetch(`${baseUrl}/testing/123.json`)
+      expect(res.ok).to.be.true
+      expect(await res.json()).to.deep.equal({
+        name: 'testing',
+        value: 'Maximum response size exceeded',
+      })
+      nock.cleanAll()
```
paulmelnikow: These look good.

This line won't be reached if the `expect`s fail. It's probably safest to put the `nock.cleanAll()` call in an `afterEach` hook.

services/base.js
```diff
@@ -395,6 +397,7 @@ class BaseService {
           sendBadge(format, badgeData)
         },
         cacheLength: this._cacheLength,
+        fetchLimitBytes: bytes(config.public.fetchLimit),
```
paulmelnikow: This could be passed through `register()` in `lib/server.js`:

https://github.com/badges/shields/blob/afcc0e3920aa7f8f6f80f1ca40b757315f1c02ce/lib/server.js#L181-L185

chris48s:
Well, nothing failed with `"value" with value "Maximum response size exceeded" fails to match the required pattern` so I'm claiming victory :rocket:

PR #2732, opened 2019-01-11
refactor [VSMarketplace] to new service model

### LINKED ISSUE #2689: VS Code Marketplace Downloads Badge using wrong data point

- [ ] My own instance

:beetle: **Description**
The VS Marketplace tracks and differentiates between "installs" and "downloads". It defines `installs` as the total number of unique users that have installed an extension, but the `install` count does not include extension updates. `Downloads` on the other hand equals the number of unique installs + the number of extension updates downloaded by users who have installed the extension. 

For example if I publish extension`my.extension-foo@1.0.0`, and User A and User B both install my extension, then the Marketplace will show `installs = 2` and `downloads = 2`. If I then publish a new version (@2.0.0), and user B grabs the new version, the marketplace will show `installs = 2`, `downloads = 3` 

Our VS Code Marketplace badge uses the term `downloads`, but the badge is actually using the API field that the Marketplace refers to `installs`, which leads to a mismatch (which is especially visible on the Marketplace). For example:

image

:link: **Link to the badge**
https://img.shields.io/vscode-marketplace/d/:repo.svg 
 https://img.shields.io/vscode-marketplace/d/swellaby.rust-pack.svg (badge shown in screenshot)

:bulb: **Possible Solution**
I propose two steps, the first one should resolve this discrepancy and the second involves adding a new badge 

1) Update the existing badge to properly reflect the `downloads` count. The required data is already available in the API response (snippet shown below). We're currently grabbing the value of `install` statistic and we just need to take that and add the value of `updateCount` statistic
```json
"statistics": [
  {
    "statisticName": "install",
    "value": 159,
  },
  {
    "statisticName": "averagerating",
    "value": 5,
  },
  {
    "statisticName": "ratingcount",
    "value": 1,
  },
  {
    "statisticName": "trendingdaily",
    "value": 0.6329113924050633,
  },
  {
    "statisticName": "trendingmonthly",
    "value": 106.9620253164557,
  },
  {
    "statisticName": "trendingweekly",
    "value": 63.92405063291139,
  },
  {
    "statisticName": "updateCount",
    "value": 13,
  },
  {
    "statisticName": "weightedRating",
    "value": 4.462021091977039,
  },
]
```

2) Add a new badge called `installs` that reflects the install count the Marketplace tracks (this would be the exact same as our current implementation)

calebcartwright:
* Refactor VS Marketplace services to new service model
  * Also renames directory, file, and service classnames to use VS Marketplace for consistency with example text and the [branding of the marketplace](https://docs.microsoft.com/en-us/azure/devops/marketplace/overview?view=vsts) (VS Marketplace is the overall name as it includes extensions for Visual Studio IDE and Azure DevOps in addition to VS Code)
  * Updated the route accordingly to support `vs-marketplace` in addition to maintaining support for legacy path of `vscode-marketplace`
* Fixes #2689 by adding new badge for Installs and fixing value used for Downloads badge


services/vs-marketplace/vs-marketplace-base.js
```diff
@@ -0,0 +1,108 @@
...
+
+  transformExtension({ json }) {
+    const extensions = json.results[0].extensions
+    if (extensions.length === 0) {
+      throw new NotFound({ prettyMessage: 'extension not found' })
+    }
+    return { extension: extensions[0] }
+  }
+
+  transformStatistics({ json }) {
```
calebcartwright: I broke out these transform functions this way to make it easier for the services to grab the specific data points they need without having to redefine transform logic in each service

services/vs-marketplace/vs-marketplace-base.js
```diff
@@ -0,0 +1,108 @@
...
+    const { extension } = this.transformExtension({ json })
+    return { statistics: extension.statistics }
+  }
+
+  transformStatisticValue({ statistics, statisticName }) {
```
paulmelnikow: What about calling this `findStatistic` or `getStatistic`?

services/vs-marketplace/vs-marketplace-rating.service.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  static render({ format, averageRating, ratingCount }) {
+    const message =
+      format === 'r'
+        ? `${averageRating}/5 (${ratingCount})`
+        : starRating(averageRating)
+    return {
+      label: 'rating',
```
paulmelnikow: The label could be moved to `defaultBadgeData`.

services/vs-marketplace/vs-marketplace-version.service.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  static get route() {
+    return {
+      base: '',
+      pattern: '(vscode-marketplace|vs-marketplace)/v/:extensionId',
+    }
+  }
+
+  static render({ version }) {
```
paulmelnikow: You could use `renderVersionBadge` here.

services/vs-marketplace/vs-marketplace-base.js
calebcartwright: 👍

services/vs-marketplace/vs-marketplace-rating.service.js
calebcartwright: I was wondering if that was a possibility (meant to test but forgot to). I wasn't sure if the fact that the base class defined `defaultBadgeData` would cause problems?

services/vs-marketplace/vs-marketplace-version.service.js
calebcartwright: 👍

services/vs-marketplace/vs-marketplace-rating.service.js
paulmelnikow: This one would take precedence.

I didn't realize we were using it in the base class. It's one of the reason I prefer functional composition to inheritance in the services; it's easy to miss stuff like that.

Adding one here would change the label in the error case, though I think that is okay.

calebcartwright: Agreed! Updated accordingly

PR #2735, opened 2019-01-11
update tutorial to use localhost instead of others

Prouser123:
Seeing as npm start now uses localhost instead of instead of 127.0.0.1 or [::], just updating it in the tutorial.

_Edit: updated the title as checks are failing because of [::]_


paulmelnikow:
Thanks!

PR #2737, opened 2019-01-11
Add a script used for backing up the production GitHub tokens

paulmelnikow:
Seems better to have this checked in here than sitting on my dev machine.


calebcartwright:
Actually I have a semi-related question: what all belongs in the `scripts` directory in this repo vs. what belongs over in https://github.com/badges/ServerScript?

paulmelnikow:
I'm inclined to keep ServerScript to bits related to the legacy deploy itself, and bring the other operational bits in here, for more visibility.

calebcartwright:
That makes sense, thanks!

PR #2740, opened 2019-01-11
Validate service data + support `labelColor` for the endpoint badge

paulmelnikow:
Ref #2473


paulmelnikow:
Failures seem unrelated. GemRank fixed in #2743.

Issue #1584, opened 2018-03-19
Code coverage service for shields


platan:
What do you think about adding a code coverage service (like https://codecov.io/#features or https://coveralls.io/features, an example results https://codecov.io/gh/babel/babel) to this project? I know that our service tests are not 100% reliable, but in my opinion such service would be helpful in finding untested parts of the code in PR. What are your experiences with this kind of services?

paulmelnikow:
Yes, I'd love it! I haven't used Codecov, though with IcedFrisby I've used Coveralls and it works well. A scoreboard might be a big motivator – both to get coverage up in general, and to complete the rewrite (#1358).

Once we choose, we might need to ask Thaddée to set it up for us. I don't have "owner" access to the Shields organization so there are certain apps I'm not able to set up.

RedSparr0w:
:+1: Definitely agree it would be useful to have coverage checked on PR's

chris48s:
I think for something like coveralls it is a problem that we don't run all of our tests each merge/PR. i.e: unless we prefix every PR with "[amo, ansible, appveyor, bitbucket....." or just run all the service tests on every PR (which would be slow among other things ), I think something like coveralls will just generate a meaningless number that goes up and down, rather than telling us anything useful. Maybe codecov is more configurable in that respect.

If not, maybe we could set up a config that defines a subset of the codebase where we are measuring coverage which excludes the services?

paulmelnikow:
> I think something like coveralls will just generate a meaningless number that goes up and down, rather than telling us anything useful. Maybe codecov is more configurable in that respect.

Hmm, that's a great point!

> If not, maybe we could set up a config that defines a subset of the codebase where we are measuring coverage which excludes the services?

This seems like a good way to go.

When you're writing tests for a service, it would also be good to see coverage of the service code. When I look at https://coveralls.io/github/IcedFrisby/IcedFrisby, I see it does a nice job of showing coverage across different files and/or parts of the tree. Though it wouldn't be as useful for the kind of "single number" coverage we'd want to track over time for master, it might be useful enough.

Another ~~hack~~ option would be to use codecov for core coverage, and coveralls for services coverage, or vice versa. 😁 The core number should be reliable, and trending upward, while for the services the meaningful information is at the file level.

I did set up some coverage tooling using istanbul/nyc which similarly breaks out the core code from the service tests. I'm not sure if it's working right now, though. I haven't run it in a while.

platan:
[...]

PR #2742, opened 2019-01-12
Refactor badge color functions

paulmelnikow:
- Replace the idea of color schemes with the idea of named colors (since none of our colorschemes have used `colorA`)
- Pass through the normalized color to `_shields_test` to harmonize with BaseService and simplify testing
    - Update service tests
- Move responsibility for color generation into the npm package
- Remove several color helper functions and their tests
- Update gh-badge public API to accept `color` and `labelColor`

This is a precursor to refactoring some of the logo code for #2473.


chris48s:
I've had a go at reviewing this. I'll leave a few comments but it probably needs another pass.

One of the issue I'm hitting trying to test is that editing stuff in the modal popups in the front end is throwing:

```
TypeError: Cannot read property 'toLowerCase' of null
at isHexColor (shields/gh-badges/lib/color.js:22:31)
```

which is making it difficult to test overrides, logos etc

gh-badges/lib/badge-cli.js
```diff
@@ -4,15 +4,15 @@
 
 const makeBadge = require('./make-badge')
 const svg2img = require('./svg-to-img')
-const colorscheme = require('./colorscheme.json')
+const { namedColors } = require('./color')
 
 if (process.argv.length < 4) {
   console.log('Usage: badge subject status [:colorscheme] [.output] [@style]')
```
chris48s: Does `[:colorscheme]` want updating here?

services/base.js
```diff
@@ -339,6 +334,14 @@ class BaseService {
       label: defaultLabel,
     } = this.defaultBadgeData
 
+    let color
+    if (isError) {
+      // Disregard the override color.
+      color = coalesce(serviceColor, defaultColor, 'lightgrey')
+    } else {
+      color = coalesce(overrideColor, serviceColor, defaultColor, 'lightgrey')
+    }
```
chris48s: This section will want updating to account for #2740

services/dub/dub.tester.js
```diff
@@ -3,19 +3,18 @@
...
-  colorScheme.red,
-  colorScheme.yellow,
-  colorScheme.yellowgreen,
-  colorScheme.green,
-  colorScheme.brightgreen
+  'red',
+  'yellow',
+  'yellowgreen',
+  'green',
+  'brightgreen'
```
chris48s: This will need updating to account for #2751 - I moved/renamed some stuff

chris48s:
It seems to be a bit inconsistent exactly which keypresses throw it, but here's some screen capture from me filling some text into form fields - You will need to fullscreen the GIF to see the details:

gifrecord_2019-01-13_184256

Does that help? (This is with `HEAD` of my branch at b60b8fb )

paulmelnikow:
Gotcha. Yes! Thanks a bunch. Looks like a backend issue. I'll see if I can't track that down.

gh-badges/lib/color.js
```diff
@@ -0,0 +1,49 @@
...
+  orange: '#fe7d37',
+  red: '#e05d44',
+  blue: '#007ec6',
+  grey: '#555',
+  gray: '#555',
+  lightgrey: '#9f9f9f',
+  lightgray: '#9f9f9f',
+}
+
+const hexColorRegex = /^([\da-f]{3}){1,2}$/i
```
calebcartwright: do we need to add an optional `#` to this?

calebcartwright: after looking through the tests I'm guessing the answer is no since `normalizeColor` handles this?

paulmelnikow: I thought the same thing. However `isCSSColor('#ccc')` is true, so we don't need to accept it here.

calebcartwright: Ahhh, thanks for that. I was scratching my head on that one

calebcartwright:
Should we also run these changes against all service tests to validate there's no unexpected color change behavior?

paulmelnikow:
> Should we also run these changes against all service tests to validate there's no unexpected color change behavior?

Yea, I did the full service test run in a number of the earlier builds, and cleaned up all the failures that seemed to be related. It's possible one or two have crept in since then, though at this point I think we can catch them more easily just after merging.

paulmelnikow:
> Gotcha. Yes! Thanks a bunch. Looks like a backend issue. I'll see if I can't track that down.

@chris48s This issue is fixed, by the way.

calebcartwright:
> Yea, I did the full service test run in a number of the earlier builds 

Ah sorry, I missed that

calebcartwright:
Tangential question, but have we ever evaluated the feasibility of having Coveralls chime in on PRs with info about coverage changes? I imagine that could get tricky given the different test runs we have (and I know those coverage bots can get a little spammy on some PRs), but I think that could be valuable in some cases, especially for the core code

paulmelnikow:
There was a little discussion in #1584 and I wrote a bit the other day in https://github.com/badges/shields/pull/2756#discussion_r247362497.

I think ideally we have two suites:

1. Core
    - Only runs the core tests
    - Is expected to climb toward 100%
    - Is reported in PRs
2. Services
     - Runs the service tests specified in PRs
     - Gets reported in PRs
     - Ideally, only includes code belonging to the services being run. In some cases it may be difficult to detect which service code should be included. If we run `[pypidownloads]` then which files in `pypi` should be covered? I suppose if we run a subset of a service's tests we could err on the side of under-reporting.

1 seems easier to build. We could start there. Ideally if we can find a way to hack this with Coveralls, I think we should do that… it'd involve somehow keeping the two coverage runs separate. It might be possible if we switch `daily-tests` to think it's reporting on a branch, though I'm not totally sure if that would work.

calebcartwright:
I agree with targetting/emphasizing the 1. Core option 

I suppose we could leverage multiple .nycrc files, and perhaps another option would be to use Coveralls for one run and maybe a different service like Codecov for another? I'd venture to say others have encountered a similar scenario before (for example monorepos) and wanted coverage on different parts so I'd be curious to see what others have done.

calebcartwright:
Also as I read through #1584 I realize I am just repeating things you've all already thought of 😆

calebcartwright:
While searching for Coveralls docs I found this from codecov: https://docs.codecov.io/docs/flags. I'll see what I can find on Coveralls and will open another issue if need be to avoid taking this PR any more off topic!

paulmelnikow:
Ooh, possibly useful: https://stackoverflow.com/a/49907438/893113

paulmelnikow:
Fabulous! Thanks for reviewing. I know this was kind of a bear. 🐻

Merge conflict sorted.

PR #2745, opened 2019-01-12
Added various [Spiget] badges

Prouser123:
Had some fun with base classes, interesting learning experience. 👍 

Added:
- Spiget Downloads badge
- Spiget Download size badge
- Spiget Latest version badge
- Spiget rating badge
- Spiget star rating badge 

- Spiget base class
- SpigetRating base class

### DIFF
--- a/services/spiget/spiget-base.js
+++ b/services/spiget/spiget-base.js
@@ -10,6 +10,11 @@ const schema = Joi.object({
     size: Joi.number().required(),
     sizeUnit: Joi.string().required(),
   }).required(),
+
+  rating: Joi.object({
+    count: Joi.number().required(),
+    average: Joi.number().required(),
+  }).required(),
 }).required()
 
 const versionSchema = Joi.object({
--- a/services/spiget/spiget-base.js
+++ b/services/spiget/spiget-base.js
@@ -11,6 +11,8 @@ const schema = Joi.object({
     sizeUnit: Joi.string().required(),
   }).required(),
 
+  testedVersions: Joi.array(),
+
   rating: Joi.object({
     count: Joi.number().required(),
     average: Joi.number().required(),
--- a/services/spiget/spiget-base.js
+++ b/services/spiget/spiget-base.js
@@ -1,0 +1,52 @@
+'use strict'
+
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
+
+const schema = Joi.object({
+  downloads: Joi.number().required(),
+
+  file: Joi.object({
+    size: Joi.number().required(),
+    sizeUnit: Joi.string().required(),
+  }).required(),
+}).required()
+
+const versionSchema = Joi.object({
+  downloads: Joi.number().required(),
+  name: Joi.string().required(),
+}).required()
+
+const documentation = `
+<p>You can find your resource ID in the url for your resource page.</p>
+<p>Example: <code>https://www.spigotmc.org/resources/essentialsx.9089/</code> - Here the Resource ID is 9089.</p>`
+
+class BaseSpigetService extends BaseJsonService {
+  async fetch({ resourceid }) {
+    const url = `https://api.spiget.org/v2/resources/${resourceid}`
+
+    return this._requestJson({
+      schema,
+      url,
+    })
+  }
+
+  async fetchLatestVersion({ resourceid }) {
+    const url = `https://api.spiget.org/v2/resources/${resourceid}/versions/latest`
+
+    return this._requestJson({
+      schema: versionSchema,
+      url,
+    })
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'spiget' }
+  }
+
+  static get category() {
+    return 'other'
+  }
+}
+
+module.exports = { documentation, BaseSpigetService }
--- a/services/spiget/spiget-download-size.service.js
+++ b/services/spiget/spiget-download-size.service.js
@@ -1,0 +1,65 @@
+'use strict'
+
+const BaseJsonService = require('../base-json')
+
+const Joi = require('joi')
+
+const schema = Joi.object({
+  file: Joi.object({
+    size: Joi.number().required(),
+    sizeUnit: Joi.string().required(),
+  }).required(),
+}).required()
+
+//const schema = Joi.any()
+
+module.exports = class SpigetDownloadSize extends BaseJsonService {
+  static get route() {
+    return {
+      base: 'spiget/download-size',
+      pattern: ':resourceid',
+    }
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'size',
+      color: 'brightgreen',
+    }
+  }
+
+  async handle({ resourceid }) {
+    const { file } = await this.fetch({ resourceid })
+    return this.constructor.render({ size: file.size, unit: file.sizeUnit })
+  }
+
+  async fetch({ resourceid }) {
+    const url = `https://api.spiget.org/v2/resources/${resourceid}`
+
+    return this._requestJson({
+      schema,
+      url,
+    })
+  }
+
+  static render({ size, unit }) {
+    return {
+      message: `${size} ${unit}`,
+    }
+  }
+
+  static get category() {
+    return 'size'
+  }
+  static get examples() {
+    return [
+      {
+        title: 'Spiget Download Size',
+        namedParams: {
+          resourceid: '9089',
+        },
+        staticPreview: this.render({ size: 2.5, unit: 'MB' }),
+      },
+    ]
+  }
+}
--- a/services/spiget/spiget-download-size.service.js
+++ b/services/spiget/spiget-download-size.service.js
@@ -1,19 +1,8 @@
 'use strict'
 
-const BaseJsonService = require('../base-json')
+const { BaseSpigetService, documentation } = require('./spiget-base')
 
-const Joi = require('joi')
-
-const schema = Joi.object({
-  file: Joi.object({
-    size: Joi.number().required(),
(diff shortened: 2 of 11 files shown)


calebcartwright:
Looks like linting is failing, can you review and address?

Prouser123:
Done.

calebcartwright:
Perfect, thanks! I'll take a look at this one in a little bit

services/spiget/spiget-base.js
```diff
@@ -0,0 +1,59 @@
...
+  async fetch({ resourceid }) {
+    const url = `https://api.spiget.org/v2/resources/${resourceid}`
+
+    return this._requestJson({
+      schema,
+      url,
+    })
+  }
+
+  async fetchLatestVersion({ resourceid }) {
```
calebcartwright: This function is only called from a single service (`spiget-latest-version`) so it appears to be tailored to a very specific use case. Let's try refactoring your `fetch` function a bit to take in additional parameters for `schema` and `url` (you can give them default values of `schema` and `https://api.spiget.org/v2/resources/${resourceid}`). 

You can then remove this function from the base class, move the `versionSchema` and `url` to the `spiget-latest-version` service, and update that service to call `fetch`, and pass the `url` (https://api.spiget.org/v2/resources/${resourceid}/versions/latest) and `versionSchema` as arguments

services/spiget/spiget-versions.service.js
```diff
@@ -0,0 +1,56 @@
...
+  static render({ versions }) {
+    return {
+      message: versions,
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Spiget Downloads',
```
calebcartwright: Should be something like `Spiget Tested Versions`?

services/spiget/spiget-versions.service.js
```diff
@@ -0,0 +1,56 @@
+'use strict'
+
+const { BaseSpigetService, documentation } = require('./spiget-base')
+
+module.exports = class SpigetTestedVersions extends BaseSpigetService {
+  static get route() {
+    return {
+      base: 'spiget/versions',
```
calebcartwright: Let's update the route (and file name) to both be consistent and a bit more explicit that this is for `tested-versions`

services/spiget/spiget-latest-version.service.js
```diff
@@ -0,0 +1,46 @@
...
+  static get route() {
+    return {
+      base: 'spiget/version',
+      pattern: ':resourceid',
+    }
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'version',
```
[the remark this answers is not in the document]
Prouser123: well, SpiGet is the API for getting resource information from [SpigotMC](https://www.spigotmc.org/), which is the actual platform.

Personally I think a label of `version` would be the most meaningful, or we could change it to `spigot` or similar.

services/spiget/spiget-rating.service.js
```diff
@@ -0,0 +1,85 @@
+'use strict'
```
calebcartwright: You can simplify this to just have a single class definition (`SpigetRatings`). The only conditional part is really the value of the `message` being set in the `render` function. You could either leverage a class exporter function, or update the route to handle the two formats. The latter approach would have a route function like this:

```js
static get route() {
  return {
    base: 'spiget',
    pattern: ':format(rating|stars)/:resourceid',
  }
}
```

and then update the `handl` function to include the format parameter and pass it along as an argument to the render function:

```js
async handle({ format, resourceid }) {
  const { rating } = await this.fetch({ resourceid })
    format,
    total: rating.count,
    average: rating.average,
  })
}
```

and then finally move the conditional rendering logic to your render function with something like:

```js
static render({ format, total, average }) {
  const message = format === 'stars'
    ? starRating(average)
    : `${average}/5 (${metric(total)})`
  return {
    message,
    color: floorCount(average, 2, 3, 4),
  }
}
```

services/spiget/spiget-base.js
```diff
@@ -0,0 +1,59 @@
+'use strict'
+
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
+
+const schema = Joi.object({
+  downloads: Joi.number().required(),
+
```
calebcartwright: minor formatting: can you remove the line breaks in the schema properties?

services/spiget/spiget-latest-version.service.js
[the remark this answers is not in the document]
calebcartwright: In the base class it looks like you've set the default label to `spiget` but I'll defer to you here as to whatever makes the most sense for users 👍

services/spiget/spiget-download-size.service.js
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      base: 'spiget/download-size',
+      pattern: ':resourceid',
+    }
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'size',
+      color: 'brightgreen',
```
calebcartwright: the `brightgreen` color is typically used for badges to indicate some kind of success (and red for failures, etc.) for informational badges like this I think we should probably use something like `blue`

services/spiget/spiget-download-size.tester.js
```diff
@@ -0,0 +1,14 @@
...
+
+const t = (module.exports = require('../create-service-tester')())
+
+t.create('EssentialsX (id 9089)')
+  .get('/9089.json')
+  .expectJSONTypes(Joi.object().keys({ name: 'size', value: isFileSize }))
+
+t.create('Invalid Resource (id 1)')
+  .get('/1.json')
+  .expectJSONTypes(Joi.object().keys({ name: 'size', value: 'not found' }))
```
calebcartwright: When we know **exactly** what the response will be, we can and should use the `.expectJSON` function:

```js
.expectJSON({
  name: 'size',
  value: 'not found',
})
```

services/spiget/spiget-latest-version.tester.js
```diff
@@ -0,0 +1,26 @@
...
+
+// Note that Spigot versions can be anything (including just a string), so we'll make sure it's not returning 'not found'
+
+t.create('EssentialsX (id 9089)')
+  .get('/9089.json')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'version',
+      value: withRegex(/^(?!not found$)/),
+    })
```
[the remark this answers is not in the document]
calebcartwright: Well that's odd! Not sure we can do anything differently here in that case, 👍

services/spiget/spiget-base.js
```diff
@@ -0,0 +1,44 @@
...
+  rating: Joi.object({
+    count: Joi.number().required(),
+    average: Joi.number().required(),
+  }).required(),
+}).required()
+
+const documentation = `
+<p>You can find your resource ID in the url for your resource page.</p>
+<p>Example: <code>https://www.spigotmc.org/resources/essentialsx.9089/</code> - Here the Resource ID is 9089.</p>`
+
```
calebcartwright: You mentioned before that `spiget` is the API platform for the service `spigot` or `spigotmc` right? If so what do you think about also adding `spigot` and/or `spigotmc` as keywords (can do it here in the base class the same way you added documentation) so that the badges will also show up on search results for `spigot` etc.?

services/spiget/spiget-rating.tester.js
```diff
@@ -0,0 +1,28 @@
...
+const ServiceTester = require('../service-tester')
+
+const { isStarRating } = require('../test-validators')
+
+const t = new ServiceTester({
+  id: 'spiget',
+  title: 'SpigetStars',
+})
+module.exports = t
+
```
calebcartwright: can you also add test cases for the `/rating/` formatted badges?

services/spiget/spiget-tested-versions.service.js
```diff
@@ -0,0 +1,56 @@
...
+  transform({ testedVersions }) {
+    const earliest = testedVersions[0]
+    const latest = testedVersions.slice(-1)[0]
+    let versions = ''
+    if (earliest === latest) {
+      versions = earliest
+    } else {
+      versions = `${earliest}-${latest}`
+    }
+    return { versions }
```
calebcartwright: We should add some additional test cases (using mock responses) to validate all of this logic

paulmelnikow:
Hey @Prouser123 would you have a moment to address these comments? It would be great to get this shipped! 🚀

services/spiget/spiget-downloads.service.js
```diff
@@ -0,0 +1,46 @@
...
+const { downloadCount } = require('../../lib/color-formatters')
+
+module.exports = class SpigetDownloads extends BaseSpigetService {
+  static get route() {
+    return {
+      base: 'spiget/downloads',
+      pattern: ':resourceid',
+    }
+  }
+
```
calebcartwright: Sorry I missed this last time but this should have the category set to `downloads` (it's currently showing up under other because of the base class category)

services/spiget/spiget-rating.tester.js
```diff
@@ -0,0 +1,41 @@
...
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'rating',
+      value: withRegex(/^(\d*\.\d+)(\/5 \()(\d+)(\))$/),
+    })
+  )
+
+t.create('Rating - Invalid Resource (id 1)')
+  .get('/rating/1.json')
+  .expectJSONTypes(
```
calebcartwright: Let's use `expectJSON(` here (same as the one starting on line 20)

calebcartwright:
Thanks @Prouser123! Two final requests and then I'll merge this

PR #2748, opened 2019-01-12
Update [VisualStudioMarketplace] download calculation for Azure DevOps Extensions

calebcartwright:
Related to #2689. The fix applied in #2738 resolved the `download` count values for Marketplace extensions for VS Code and Visual Studio, but I just realized that extensions for Azure DevOps utilize a different statistic name than the other two.

This tweaks the Marketplace Downloads service to resolve that so the badges will display the correct download count for all extension types on the Visual Studio Marketplace


services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
```diff
@@ -13,7 +13,7 @@ module.exports = class VisualStudioMarketplaceDownloads extends VisualStudioMark
     return {
       base: '',
       pattern:
-        '(visual-studio-marketplace|vscode-marketplace)/:measure(d|i)/:extensionId',
+        '(visual-studio-marketplace|vscode-marketplace)/:platform(ado)?/:measure(d|i)/:extensionId',
```
calebcartwright: I elected to use the shorthand `ado` here instead of the lengthier `azure-devops`. However, given the convenience of the new modal window around populating badge URLs, there may not be much value in the shorthand. 

Let me know if you think the longer version would be better

paulmelnikow:
Would it possible for us to detect this case rather than have the user specify it?

calebcartwright:
> Would it possible for us to detect this case rather than have the user specify it?

That was my preferred approach as well but I talked myself into believing it wasn't possible 😄 Let me take another look at it and see

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
```diff
@@ -42,27 +42,53 @@ module.exports = class VisualStudioMarketplaceDownloads extends VisualStudioMark
...
+    // get the correct number of downloads in all cases.
+    if (updateCount > 0) {
+      updates = updateCount
+    } else {
+      const { value: onpremDownloads } = this.getStatistic({
+        statistics,
+        statisticName: 'onpremDownloads',
+      })
+      updates = onpremDownloads
+    }
```
calebcartwright: I added the above (lengthy) comment to hopefully clarify the what & why of what's happening here. But basically: although it's not possible to definitively know what type of extension we're dealing with, we _can_ still handle the Azure DevOps extension scenario behind the scenes without requiring the user to specify that explicitly in the route path

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
```diff
@@ -42,27 +42,53 @@ module.exports = class VisualStudioMarketplaceDownloads extends VisualStudioMark
         namedParams: { extensionId: 'ritwickdey.LiveServer' },
         staticPreview: this.render({ measure: 'd', count: 1239 }),
         keywords: this.keywords,
+        documentation: `For Azure DevOps extensions, this includes total downloads for both on-prem Azure DevOps Server and Azure DevOps Services`,
```
paulmelnikow: As someone not familiar with these tools, the second half of this isn't super clear. What does it mean "for both on-prep Azure DevOps Server and Azure DevOps Services?"

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
```diff
@@ -42,27 +42,53 @@ module.exports = class VisualStudioMarketplaceDownloads extends VisualStudioMark
...
     })
+
+    // updateCount will only be greater than 0 if the extension is for VS or VS Code.
+    // If the value of updateCount is zero then the extension is either:
+    // (A) For VS or VS Code, but has no published updates that have been downloaded
+    // (B) For Azure DevOps
+    // It is not possible to definitively know whether A or B is true, but in either case we should
+    // check the value of the onpremDownloads statistic tracked for Azure DevOps extensions.
+    // If the extension is for VS or VS Code then onpremDownloads will be 0 so we'll still
+    // get the correct number of downloads in all cases.
```
paulmelnikow: Sounds good. Could you link to the issue as well?

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
calebcartwright: As far as Azure DevOps Server & Azure DevOps Services in this context it is the same concept as Bitbucket Server & Bitbucket Cloud, GitHub Enterprise & GitHub, Shields.io vs. self-hosted Shields 😄 etc. Services refers to the SaaS/cloud offering, Server is the one you download, install, and run yourself.

It's a concept that will definitely be familiar to publishers and consumers: it's the labeling used on extensions in the marketplace, every extension publisher has to specify whether their extension is supported for Azure DevOps Services and/or Server, and consumers acquire the extension specifically for one of those two targets. 

image

I'm definitely open to any suggested changes on the wording, but I do think the description is important.

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
calebcartwright: Which issue, #2689 or this PR?

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
paulmelnikow: I could try to suggest an improvement. What does it mean to "download" these extensions?

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
paulmelnikow: Whichever you think is more helpful!

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
calebcartwright: When a user wants to acquire an extension for Azure DevOps, they start from the marketplace but the way they go about actually getting that extension varies noteably between the SaaS and self-hosted platforms. 

For SaaS, acquiring an extension basically happens via a click/confirmation process by the user and the extension is installed in that users account automagically by the platform.

For self-hosted, the user "acquires" the extension by downloading the extension package from the marketplace (a `.vsix` package file if I recall correctly) to their local machine and from there they add that extension to their self-hosted Azure DevOps server.

calebcartwright: Honestly it may make more sense to handle the badges for extensions targeting Azure DevOps differently although that would require the user specifying their target was for Azure DevOps (which we were initially trying to avoid) 

In theory there's 3 data points for Marketplace extensions that target Azure DevOps which we could provide badges for:

1) Number of times the extension has been acquired for SaaS target
2) Number of times the extension has been acquired for Self-hosted target
3) Total of 1 + 2

paulmelnikow: Heh. It's still kind of confusing, isn't it.

For Server, is the act of downloading the `.vsix` what gets counted?

For Services, is it the install that gets counted?

paulmelnikow: > * Number of times the extension has been acquired for SaaS target
> * Number of times the extension has been acquired for Self-hosted target

I was going to suggest a second badge, though we do already have the Visual Studio Marketplace Installs badge. Is downloads vs installs making a similar distinction?

calebcartwright: > Heh. It's still kind of confusing, isn't it.

Yeah it is. Even as someone who has published extensions on that marketplace.. I still have questions about some of the metrics/labels (there's a deeper layer viewable in the publisher portal that is even more confusing). 


> For Server, is the act of downloading the `.vsix` what gets counted?
> 
> For Services, is it the install that gets counted?

Yup. The way I've been trying to look at is what info matters to an extension publisher (I presume the main target audience for badges), and match that with the data available to Shields via the marketplace API. 

The API response has two relevant numerical data points for an extension, Which I find to be a straightforward distinction:

`install` - the number of times the extension was added to SaaS instance
`onPremDownloads` - the number of times the extension was added to a self-hosted 

The part that makes it all the more confusing is that the Marketplace UI lists `installs` as the total of those two values (SaaS + Self-hosted) so that it is consistent with what it shows for VS Code and VS Extensions

Am I making this worse? 🤔

paulmelnikow:
@calebcartwright It would be great to close this out, one way or another. We discussed this offline for a while and you were going to think about it some more after that. 😁What did you decide?

calebcartwright:
IMO the best approach is still to move to more explicit badge route paths where the user specifies if the extension is for Azure DevOps.

I know it would be ideal if we could auto-magically handle everything behind the scenes but as it stands today, the data we have available via the API is simply too ambiguous for us to be able to dynamically handle all the various scenarios we'll encounter the way we want to. 

If we _really_ want to avoid having users specify 'azure-devops' in the route, then the current changes here will achieve that functional result, albeit with the caveats/edge cases I listed offline. 

My preferred approach:
1) Create a new service to handle azure devops extensions on the marketplace
2) the route can be something like `visual-studio-marketplace/azure-devops/metric(total|onprem|services)`
3) that'll add the 3 badges we discussed offline (total installs, # "installs" for the on-prem version, # "installs" for the cloud/saas version)
4) we can leave the current/existing route in place, as is for vscode/vs extensions, and any existing badges using that route for an azure devops extension would continue to work as well (albeit with the off-numbers just like today0

paulmelnikow:
I'm good with that.

calebcartwright:
I think this now includes the desired functionality via approach I outlined above. Welcome any feedback! 

I was debating whether some example documentation would be helpful but decided to refrain (at least for now)

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
```diff
@@ -46,23 +46,30 @@ module.exports = class VisualStudioMarketplaceDownloads extends VisualStudioMark
...
 
-    const { value: updates } = this.getStatistic({
+    const { value: updateCount } = this.getStatistic({
       statistics,
       statisticName: 'updateCount',
     })
-    const downloads = +installs + +updates
-    return this.constructor.render({ measure, count: downloads })
+
+    return { count: updateCount + installs }
```
paulmelnikow: I think the version of this I suggested is clearer because it makes the logic very clear. It's true this version saves a function call in a few cases, however that function is extraordinarily fast.

It could probably be even clearer if `transformStatistics` turned this into a normal object that could be destructured. Then the helper method wouldn't even be needed:

```js
const result = {}

extension.statistics.forEach(({ statisticName, value }) => {
  result[statisticName] = value
})

return { statistics: result }
```

paulmelnikow:
Only thought on documentation: should we mention in the "regular" Visual Studio Marketplace badges that we have a separate badge to show the correct stats for Azure Devops Extensions?

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
calebcartwright: This is a pretty minor to change to what I had added in https://github.com/badges/shields/pull/2732 as part of the original refactor of this service, but certainly worth discussion/refactor (especially since some of the original complexity has been removed with Azure DevOps moved to a separate service)

The other downside with this current approach is that it involves repeated iterating through the statistic objects returned in that array 😉  That felt.. "dirty" to me so I started down a different path when I first started on the refactor, but ultimately I decided to stick with this implementation that was used in the LegacyService. I went that route since the Marketplace API in use is completely undocumented (as far as I can tell) and I actually felt that approach would be easier to grok for someone who wasn't intimately familiar with that API response and how Marketplace conditions change that response. 

The objects that exist in that statistic array can vary pretty wildly based on a variety of factors. From what I've seen, albeit anecdotally with a few extensions, there's not a ton of objects in that array (10-20 range max), but I have no idea what (if any) the upper limit is. However, when an extension has been installed at least once the `install` statistic has always been the first element in the array and the other two objects needed for the other badges are within the first ten elements. 

The other odd thing is that statistic objects aren't returned if a stat has a value of zero. To handle that scenario, all of the service implementations would need to have their own conditional assignment check:

`const installs = statistic.install ? statistic.install : 0;`

or in that base `transformStatistics` function we'd need to check all 3 of the potentially used values and ensure all three existed with a numeric value. 

What you're proposing (with the minor accompanying changes I described above) makes perfect sense to me, but I've done quite a bit of work with the API and have both published several extensions to the Marketplace. Do we think that approach (and _why_ behind some of those items) will be as clear to someone without that domain knowledge looking at this code a year+ from now? I could add some inline info although it would probably end up needing to be a short story (much like this comment 😆 )

calebcartwright:
> Only thought on documentation: should we mention in the "regular" Visual Studio Marketplace badges that we have a separate badge to show the correct stats for Azure Devops Extensions?

That's exactly what I was pondering as well. I figured if we did it on one, then we'd need to do it on the other (for the Azure DevOps Extension path we'd need to say something like `use the other badge for Visual Studio/Visual Studio Code Extensions` and on the other path for VS/VSCode `use the other one for Azure DevOps Extensions`

paulmelnikow:
> That's exactly what I was pondering as well. I figured if we did it on one, then we'd need to do it on the other (for the Azure DevOps Extension path we'd need to say something like `use the other badge for Visual Studio/Visual Studio Code Extensions` and on the other path for VS/VSCode `use the other one for Azure DevOps Extensions`

Linking both seems fine, though it seems more important to direct Azure DevOps usage to the correct badge, as it's less likely someone would mistakenly use the `azure-devops` badge on a non-Azure DevOps extension.

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
paulmelnikow: I really appreciate your thoroughness and thoughtfulness about this stuff! I did not realize there was special handling of 0; thanks for explaining that.

There are two ways I might suggest handling the issue you mention. One is destructuring with defaults:

```js
const { install = 0, update = 0 } = this.transform(...)
```

That would replace the missing values with zero.

Alternatively since the services are using a relatively short list of keys, the `transform` function could include a hard-coded safe list of keys, and return only those keys from `transform`, with zeros in place of missing values.

Part of what makes a good `transform` function is that it takes a sort of messy input and tidies it up into its useful form. Hopefully this is done with an implementation that is very safe and very obvious. That would steer me toward the har-coded safe list of keys, because it encapsulates the complexity. However if the list of keys is changing frequently, the first might be better because it just works, even if it breaks encapsulation slightly.

An advantage of the declarative Joi schema is that it documents quite a bit of the magic. Even if the schema changes later, at least there's one place you can look to see what kind of assumptions the code was making.

Another option is unit testing. You could make the transform method static, and add a couple `sazerac`-based tests. A couple sets of sample inputs and outputs – or honestly, even a single one – will do a nice job of documenting its function to someone else who might not know the API.

So paragraphs of documentation are fine, though a unit test beats a paragraph. 😄 

The transform code is fast enough that I wouldn't worry about the degrees of extra processing we're discussing. There probably is more overhead in the Joi validation, for example. I'd rather have easy to read and test badge code than opaque, prematurely optimized code.

None of this is blocking, really. Fine to consider this in a follow-on.

calebcartwright: > None of this is blocking, really. Fine to consider this in a follow-on.

I don't think this PR has anything urgent so I think I'd rather knock this out in this PR too (I'm ready to be done with the VS Marketplace badges 😄) 

>  Alternatively since the services are using a relatively short list of keys, the transform function could include a hard-coded safe list of keys, and return only those keys from transform, with zeros in place of missing values.

I'm thinking this will be the best approach too. It would only be three lines and a relatively short inline comment should be able to explain _why_ that's being done. I'll make this change tonight/tomorrow

> An advantage of the declarative Joi schema is that it documents quite a bit of the magic. Even if the schema changes later, at least there's one place you can look to see what kind of assumptions the code was making

I really wish the API response would always contain the core statistic key/value pair objects, with the value of zero when warranted. If that were the case this would've been a pretty basic schema. I'm not sure how feasible (or even possible) it would be to have Joi create an object in an array with a default value if that array doesn't contain an object with a certain field value... but I learn something new about Joi nearly every day ;)

paulmelnikow: > I really wish the API response would always contain the core statistic key/value pair objects, with the value of zero when warranted. If that were the case this would've been a pretty basic schema. I'm not sure how feasible (or even possible) it would be to have Joi create an object in an array with a default value if that array doesn't contain an object with a certain field value... but I learn something new about Joi nearly every day 😉

You _could_ use Joi to filter the _post-transform_ object. i.e. turn the array into an object, then run it through a Joi schema that picks out the whitelisted values and sets the defaults. 😁

calebcartwright: Bingo! I added this to the `transformStatistics` function to ensure the defaults would be set.

calebcartwright:
Alright I think this may finally be good to go. The `transformStatistics` function has been updated with the solution discussed above (and secondary Joi validation to set defaults) and docs added to examples for VS/VSCode extensions to use the Azure DevOps Installs Badge for Azure DevOps extensions

services/visual-studio-marketplace/visual-studio-marketplace-base.js
```diff
@@ -95,16 +103,15 @@ module.exports = class VisualStudioMarketplaceBase extends BaseJsonService {
...
-      if (statistic.statisticName === statisticName) {
-        return { value: statistic.value }
-      }
-    }
+    extension.statistics.forEach(({ statisticName, value }) => {
+      statistics[statisticName] = value
+    })
 
-    return { value: 0 }
+    const { value } = Joi.validate(statistics, statisticSchema, {
```
paulmelnikow: It seems a schema failure at this point is a programmer error. Probably good to check for `error` and throw it.

Also I think `stripUnknown: true` is probably a good idea here; that way you will never end up with keys that don't have defaults declared in the schema. If keys aren't in the schema they simply won't appear, rather than appearing but not being initialized.

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
```diff
@@ -4,6 +4,15 @@ const VisualStudioMarketplaceBase = require('./visual-studio-marketplace-base')
 const { metric } = require('../../lib/text-formatters')
 const { downloadCount } = require('../../lib/color-formatters')
 
+const documentation = `
+<p>
+  This is for Visual Studio and Visual Studio Code Extensions.
+</p>
+<p>
+  Use the Azure DevOps Installs Badge for Azure DevOps Extensions.
+</p>
```
paulmelnikow: 👍

services/visual-studio-marketplace/visual-studio-marketplace-rating.service.js
```diff
@@ -63,20 +63,10 @@ module.exports = class VisualStudioMarketplaceRating extends VisualStudioMarketp
...
-    if (format === 'stars') {
-      return this.constructor.render({ format, averageRating })
-    }
-
-    const { value: ratingCount } = this.getStatistic({
-      statistics,
-      statisticName: 'ratingcount',
-    })
-
-    return this.constructor.render({ format, averageRating, ratingCount })
```
paulmelnikow: This is definitely clearer!

services/visual-studio-marketplace/visual-studio-marketplace-base.js
calebcartwright: I was wondering about that specifically under what circumstances could possibly result in the Joi validation failing? 

Would `stripUnknown` be in addition to, or in place of, `allowUnknown`?

services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js
calebcartwright: Yeah that's better!

services/visual-studio-marketplace/visual-studio-marketplace-rating.service.js
calebcartwright: +12937123!

This flow definitely looks better (in all services), glad we ended up with this approach!

services/visual-studio-marketplace/visual-studio-marketplace-base.js
paulmelnikow: Right, it sounds like it would only fail if there's been a mistake in one of the Joi schema or the other. In other words, if a programmer has made a mistake. In that case it's probably the most helpful to fail loudly here, rather than cryptically later. You can let it throw its own error, which is not one of the handled error types, so will be reported as **shields internal error**.

> Would `stripUnknown` be in addition to, or in place of, `allowUnknown`?

I've assumed you need to specify both, though I'm not completely sure!

calebcartwright: Lol me either 😆 I will play with it to see if I can get away with just `stripUnknown`

services/visual-studio-marketplace/visual-studio-marketplace-base.js
```diff
@@ -95,16 +103,23 @@ module.exports = class VisualStudioMarketplaceBase extends BaseJsonService {
...
+
+    extension.statistics.forEach(({ statisticName, value }) => {
+      statistics[statisticName] = value
+    })
+
+    // The Visual Studio Marketplace API response will only include statistic key/value pairs
+    // for statistics with values greater than 0. This ensures that the key statistics used by
+    // badges will all exist with a numerical value defaulted to 0.
+    const { value, error } = Joi.validate(statistics, statisticSchema, {
+      stripUnknown: true,
```
calebcartwright: This worked on its own and is what I _actually_ wanted the first time around 👍

services/visual-studio-marketplace/visual-studio-marketplace-base.js
```diff
@@ -95,16 +103,23 @@ module.exports = class VisualStudioMarketplaceBase extends BaseJsonService {
...
+    })
 
-  getStatistic({ statistics, statisticName }) {
-    for (const statistic of statistics) {
-      if (statistic.statisticName === statisticName) {
-        return { value: statistic.value }
-      }
+    if (error) {
+      throw error
```
calebcartwright: Is there a better way to handle this? One minor nuisance is that I don't see how I could test this block with service tests (so coverage drops from 100% to 90+%). Thinking I would need a more typical unit test to cover this block but that seems like an unnecessary effort

paulmelnikow: `Joi.attempt` does exactly what you want, though it doesn't accept `stripUnknown`.

Maybe a little hacky, though you could invoke `validate` from `core/base-service/validate.js`. Though I'm not sure what would happen if you pass `AssertionError` to that function.

calebcartwright: I'd vote we defer that edge test case (for catching a programmer error) for a later date in that case

paulmelnikow: I think we should get this merged :) If you want to leave a todo comment I'm fine with that.

calebcartwright: 👍 done

paulmelnikow: Oh! I thought you meant removing the `throw`. I think leaving it in is better. It may not be possible to cover it with a test, though it can be refactored away.

calebcartwright: So leave the throw in place, future work to refactor it out?

paulmelnikow: If you want to change it I'll 👍 again. Think this is fine, though.

calebcartwright: What's the alternative? I'm on the verge of falling asleep so may not be following 😄

paulmelnikow: I think `validate` is probably the way to go. It's not in `BaseService` but it's a helper function alongside: https://github.com/badges/shields/blob/master/core/base-service/validate.js

services/visual-studio-marketplace/visual-studio-marketplace-base.js
```diff
@@ -34,6 +35,16 @@ const extensionQuerySchema = Joi.object({
     .required(),
 }).required()
 
+const statisticSchema = Joi.object().keys({
+  install: Joi.number().default(0),
+  updateCount: Joi.number().default(0),
+  onpremDownloads: Joi.number().default(0),
+  averagerating: Joi.number()
+    .default(0)
+    .precision(2),
```
calebcartwright: Kinda lucked into this, as one of the live tests started failing today due to some rating changes in the marketplace. My current understanding is that this will result in the value for `averagerating` having a numeric value with a precision of 2 post the Joi validation (as opposed to failing validation if the value has more than 2 decimal places)

services/visual-studio-marketplace/visual-studio-marketplace-base.js
```diff
@@ -95,16 +106,14 @@ module.exports = class VisualStudioMarketplaceBase extends BaseJsonService {
...
-    for (const statistic of statistics) {
-      if (statistic.statisticName === statisticName) {
-        return { value: statistic.value }
-      }
-    }
+    extension.statistics.forEach(({ statisticName, value }) => {
+      statistics[statisticName] = value
+    })
+
+    const value = validate({ ErrorClass: Error }, statistics, statisticSchema)
```
calebcartwright: Wasn't sure if the `ErrorClass` specification was required here, but it didn't look like `validate` was setting a default for this

services/visual-studio-marketplace/visual-studio-marketplace-base.js
paulmelnikow: Yea, I just confirmed that.

Is this for formatting? Would it be better to handle in `render()`?

calebcartwright: Yeah I suppose it is, good call. So this should be flipped back, and then in the services that use this field, they should format it as needed (in their `render` functions)?

services/visual-studio-marketplace/visual-studio-marketplace-rating.tester.js
```diff
@@ -4,7 +4,7 @@ const Joi = require('joi')
 const t = (module.exports = require('..').createServiceTester())
 const { withRegex, isStarRating } = require('../test-validators')
 
-const isVscodeRating = withRegex(/[0-5].[0-9]{2}\/5?\s*\([0-9]*\)$/)
+const isVscodeRating = withRegex(/[0-5]\.[0-9]{1}\/5?\s*\([0-9]*\)$/)
```
calebcartwright: The Marketplace actually only uses 1 decimal place so I thought this (plus the format in the `render` function) would make our badges consistent

PR #2749, opened 2019-01-12
remove asterisks from examples; affects [AzureDevops GitHubDownloads GitHubManifest GitHubPackageJson GitLab]

chris48s:
Usually we don't do this, but there are small handful of example where we're exposing `*` for optional params:

screenshot at 2019-01-12 19-05-26

This is an implementation detail of path-to-regexp, not something we should expose to users. Also if we're giving multiple examples (one with a branch, one without a branch), we're covering the case where the param is absent and where it is present - its not necessary to characterise the param as "optional".


paulmelnikow:
I think it's fine to change these, though something else we could consider doing is hiding the asterisks and format strings at render time. That way, going forward, we don't have to override the pattern in the example just for that purpose.

chris48s:
I think we still want the ability to define the route as `/:user/:repo/:branch*` but then show `/:user/:repo` or `/:user/:repo/:branch` as appropriate in the examples. Were you thinking we would just use less overrides (i.e: we'd default to `/:user/:repo/:branch` but override in the `/:user/:repo` case), or were you thinking we should infer the example `pattern` based on which `namedParams` were passed?

PR #2752, opened 2019-01-12
Refactor the rest of [PyPI] to use static previews

Prouser123:
See #1961.


paulmelnikow:
Thanks for your help working through these!

PR #2753, opened 2019-01-13
Refactor [lgtm] to new service model and static previews

Prouser123:
Refactor LGTM to the new service model and to use static previews.


services/lgtm/lgtm-alerts.service.js
```diff
@@ -1,71 +1,45 @@
...
+    }
   }
 
-  static get route() {
+  async handle({ user, repo }) {
+    const { alerts } = await this.fetch({ user, repo })
+    let color = 'yellow'
+    if (alerts === 0) {
+      color = 'brightgreen'
+    }
```
chris48s: Assigning colours is a rendering concern. Can we reduce `handle()` to just

```js
const { alerts } = await this.fetch({ user, repo })
return this.constructor.render({ alerts })
```

and move the colours block to render()

services/lgtm/lgtm-base.js
```diff
@@ -0,0 +1,29 @@
...
+    return { label: 'lgtm' }
+  }
+
+  async fetch({ user, repo }) {
+    const url = `https://lgtm.com/api/v0.1/project/g/${user}/${repo}/details`
+
+    return this._requestJson({
+      schema,
+      url,
+    })
```
chris48s: I think prettier should let us do this on one line: `return this._requestJson({ schema, url })` ?

services/lgtm/lgtm-grade.service.js
```diff
@@ -1,99 +1,88 @@
...
+          color = 'green'
+        } else if (languageData.grade === 'B') {
+          color = 'yellowgreen'
+        } else if (languageData.grade === 'C') {
+          color = 'yellow'
+        } else if (languageData.grade === 'D') {
+          color = 'orange'
+        }
+      }
+    }
```
chris48s: Can we split these 2 blocks for generating the colour and label out into helper functions. Again, calling from `render()`.

services/lgtm/lgtm-base.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
+
+const Joi = require('joi')
+const BaseJsonService = require('../base-json')
+
+const schema = Joi.object({
+  alerts: Joi.number(),
+
+  languages: Joi.array(),
```
chris48s: Can we make both of these required. Also, if we've got an array of objects, we should validate the contents of that array, so if we're expecting the elements to have keys `lang` and `grade`, that should be validated in the schema

services/lgtm/lgtm-grade.service.js
```diff
@@ -1,99 +1,88 @@
...
+  }
+
+  async handle({ language, user, repo }) {
+    const data = await this.fetch({ user, repo })
+
+    if (!('languages' in data)) {
+      return this.constructor.render({
+        grade: 'invalid response data',
+        color: 'red',
+      })
```
chris48s: If you make `languages` required in the schema, we can get rid of this block - `validate()` will throw the error for us

services/lgtm/lgtm.tester.js
```diff
@@ -67,7 +67,7 @@ t.create('alerts: invalid json')
       .get('/api/v0.1/project/g/apache/cloudstack/details')
       .reply(invalidJSON)
```
chris48s: You can get rid of any tests that use `.reply(invalidJSON)` - once we migrate away from the legacy format, this stuff is handled by the base classes and has its own tests

services/lgtm/lgtm.tester.js
```diff
@@ -79,14 +79,14 @@ t.create('alerts: lgtm inaccessible')
...
   .expectJSON({
-    name: 'code quality: java',
-    value: 'project not found',
+    name: 'lgtm',
+    value: 'not found',
   })
 
 t.create('grade: lgtm inaccessible')
   .get('/grade/java/g/apache/cloudstack.json')
   .networkOff()
```
chris48s: Same with tests that use `.networkOff()`. We can remove them because this case is handled by the base classes and has its own tests in core

services/lgtm/lgtm.tester.js
```diff
@@ -21,7 +21,7 @@ t.create('alerts: missing project')
   .get('/alerts/g/some-org/this-project-doesnt-exist.json')
   .expectJSON({
     name: 'lgtm',
-    value: 'project not found',
+    value: 'not found',
```
paulmelnikow: One request from me: could you pass `errorMessages` when you make the request, so that the more detailed message is preserved?

paulmelnikow:
Hey @Prouser123, it would be great to get this merged! Would you have a few minutes to address the review comments?

services/lgtm/lgtm-base.js
paulmelnikow: Thanks so much for updating this!

I think this is the only thing still outstanding. You can use something like this should work:

```js
Joi.array()
  .items(Joi.object({
    lang: Joi.string().requred(),
    grade: Joi.equal('A', 'B', 'C', ...).required()
  }))
  .required()
```

I think `.items().required()` implies `.min(1)`, oddly enough.

Prouser123: We could do that, but Joi will only return the first language as each language is a separate object, so we need a way for Joi to validate more than one object without knowing how many there will be.

chris48s: if its wrapped in `.array().items()`, Joi will pass multiple objects:

```
$ node

> const Joi = require('joi')
undefined

> const schema = Joi.array().items(Joi.object({
...   lang: Joi.string().required(),
...   grade: Joi.string().required()
... })).required()
undefined

> const data = [
...   { lang: 'javascript', grade: 'A' },
...   { lang: 'python', grade: 'A' },
...   { lang: 'php', grade: 'A' },
... ]
undefined

> Joi.validate(data, schema, { stripUnknown: true })
{ error: null,
  value: 
   [ { lang: 'javascript', grade: 'A' },
     { lang: 'python', grade: 'A' },
     { lang: 'php', grade: 'A' } ],
  then: [Function: then],
  catch: [Function: catch] }
```

paulmelnikow: You could also experiment here: https://runkit.com/melnikow/joi-schema-testing

paulmelnikow:
@Prouser123 Would you be up for addressing https://github.com/badges/shields/pull/2753#discussion_r251216375?

We're about ready to wrap up #1961 and the LGTM `previewUrl`s are the only thing left! They've been converted here though we need to get this landed :)

services/lgtm/lgtm-base.js
paulmelnikow: Ah, gotcha, yea, no need to make `grade` required.

paulmelnikow:
Awesome. Thanks so much!!

PR #2756, opened 2019-01-13
Temporarily deprecate [CoverityScan] 

calebcartwright:
* temporarily deprecates Coverity Scan badges due to extended outage of the service, discussion in #2722 
  * https://community.synopsys.com/s/article/Coverity-Scan-Update
* updates `deprecatedService` function to enable custom deprecation badge messages
* adds unit tests for `DeprecatedService` definition

I did **not** add Coverity Scan to the deprecated services lists given that this is only temporary, but let me know if it needs to be added anyway

### DIFF
--- a/services/deprecated-service.js
+++ b/services/deprecated-service.js
@@ -4,7 +4,7 @@ const BaseService = require('./base')
 const { Deprecated } = require('./errors')
 
 // Only `url` is required.
-function deprecatedService({ url, label, category, examples = [] }) {
+function deprecatedService({ url, label, category, examples = [], message }) {
   return class DeprecatedService extends BaseService {
     static get category() {
       return category
@@ -27,7 +27,7 @@ function deprecatedService({ url, label, category, examples = [] }) {
     }
 
     async handle() {
-      throw new Deprecated()
+      throw new Deprecated({ prettyMessage: message })
     }
   }
 }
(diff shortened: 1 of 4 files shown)


services/deprecated-service.js
```diff
@@ -4,7 +4,7 @@ const BaseService = require('./base')
 const { Deprecated } = require('./errors')
 
 // Only `url` is required.
-function deprecatedService({ url, label, category, examples = [] }) {
+function deprecatedService({ url, label, category, examples = [], message }) {
```
paulmelnikow: I feel like this change ought to have a test in `deprecated-service.spec.js`, though we don't yet seem to have that file yet. We're testing it all indirectly through the services, so it seems more like a nice to have than a blocker.

I'd like to get to a point where we can make a test suite that hits 100% coverage on the core code without running the service tests. It should let us refactor the core code with more confidence and speed, certain that nothing has broken. As things are now a lot of functionality has to be tested through the services which is fairly slow, and makes it especially harder for newer contributors to change things in core. I was surprised to realize the other day that this project is 40k lines of code, and it's just unrealistic to expect people to have read most of it, to know how all the services may be using various parts of the core. But that's a bigger project for another day.

calebcartwright: > I was surprised to realize the other day that this project is 40k lines of code

WOAH! did not know that

calebcartwright: Let me take a look to see how easily tests could be added for the deprecated service def.

calebcartwright:
Tests added!

Issue #2706, opened 2019-01-08
Extend usage of build-status.js


PyvesB:
A number of services display information about the current build status (for example: ![](https://img.shields.io/appveyor/ci/gruntjs/grunt.svg)), however, we do not always do things consistently:
* in terms of Joi schema validation, some have very strict validation (e.g. AzureDevOpsBuild), others have rather loose validation (e.g. Wercker). The `isBuildStatus` validator in _build-status.js_ provides a good compromise, with rather strict validation but flexibility thanks to the wide range of accepted keywords.
* the displayed text and colours are not necessarily consistent either.

See discussions in #2660 for more information. We should go through our refactored services and use the functions in `build-status.js` where appropriate.

The `isBuildStatus` validator in _test-validators.js_ is now duplicated, shall we remove it and use the one from _build-status.js_ inside our tests?

paulmelnikow:
> The `isBuildStatus` validator in _test-validators.js_ is now duplicated, shall we remove it and use the one from _build-status.js_ inside our tests?

It's the same list? Then I say definitely! 👍

PR #2763, opened 2019-01-14
Extended usage of build-status.js to [AppveyorCi AzureDevOps Bitbucket CircleCi Gitlab Readthedocs Shippable Wercker]

PyvesB:
Takes care of the bulk of #2706.

Quite a lot of similar code was removed and most services now have stricter schema validation. I've introduced a third colour in `build-status.js`, orange, however, we could still add more granularity. In particular, some services were previously using yellow for some of the statuses.

Also mentioned the helper class in the tutorial! 👍


services/bitbucket/bitbucket-pipelines.service.js
```diff
@@ -39,18 +46,7 @@ module.exports = class BitbucketPipelines extends BaseJsonService {
   }
 
   static render({ status }) {
-    const responses = {
-      SUCCESSFUL: { message: 'passing', color: 'brightgreen' },
-      FAILED: { message: 'failing', color: 'red' },
-      ERROR: { message: 'error', color: 'red' },
-      STOPPED: { message: 'stopped', color: 'yellow' },
-      EXPIRED: { message: 'expired', color: 'yellow' },
-      'never built': { message: 'never built', color: 'lightgrey' },
```
calebcartwright: at first I thought we missed the `never built` status in the status enumerations in `build-status.js`, but I think i'm wrong. This status will be handled exactly the same way since the `renderBuildStatusBadge` function will apply this label (since this doesn't exist in happy, nor orange, nor other lists) and the default `lightgrey` color right?

services/wercker/wercker.service.js
```diff
@@ -65,8 +61,8 @@ module.exports = class Wercker extends BaseJsonService {
         result: 'no builds',
       })
     }
-    const { status, result } = json[0]
```
calebcartwright: Since we're removing `status` from the schema, render fn, etc. can we also remove the the `status` argument passed in the examples?

services/bitbucket/bitbucket-pipelines.service.js
PyvesB: Yes, exactly. 😃

services/wercker/wercker.service.js
PyvesB: Good spot, I missed some of them!

PR #2765, opened 2019-01-15
Set static previews for [jenkins jenkinsplugin] and tweak tests

paulmelnikow:


services/jenkins/jenkins-plugin-installs.tester.js
```diff
@@ -16,42 +16,10 @@ t.create('total installs | valid')
     })
   )
 
-t.create('total installs | invalid: no "installations" property')
```
paulmelnikow: These tests are basically testing Joi and BaseService and don't seem necessary. We can trust that Joi and BaseService are rejecting objects which do not match the schema without adding more tests for them.

PR #2769, opened 2019-01-15
Provide better dev feedback by validating services when they are loaded

paulmelnikow:
It's easy to push services that don't validate, because much of the tooling, including the service test runner and the service definition generator, do not validate all the services. This leads to errors that manifest in CI. It would be more helpful to see these errors sooner.

This moves the `validateDefinition()` check to `loadServiceClasses()`, where the services are first loaded, and fixes related validation errors.


services/base.js
```diff
@@ -393,8 +393,6 @@ class BaseService {
   }
 
   static register({ camp, handleRequest, githubApiProvider }, serviceConfig) {
-    this.validateDefinition()
```
paulmelnikow: This was moved to `loadServiceClasses()`.

services/amo/amo.service.js
```diff
@@ -157,6 +157,10 @@ class AmoUsers extends LegacyService {
 //
 // Do not base new services on this code.
 class Amo extends LegacyService {
+  static get category() {
+    return 'other'
+  }
```
paulmelnikow: These are meaningless categories; however these are all legacy services and the goal is to split them up.

PR #2775, opened 2019-01-15
Fix partially succeeded builds on Azure DevOps

Thilas:
There is a small typo preventing partially succeeded builds from being correctly displayed (i.e. "passing" in orange)


calebcartwright:
I think this will have been resolved in #2763 but if you're still seeing this once 2763 is deployed let us know!

Thilas:
Right after creating this PR, I just noticed #2763. And actually I found an issue (also regarding the partially succeeded status). I'm just creating another PR then: #2776.

PR #2777, opened 2019-01-16
Fix [AzureDevOps] branch specific badges

### LINKED ISSUE #2774: Azure Devops Test Badge always refers to latest build

- [ ] My own instance

:beetle: **Description**
The Azure DevOps badge always shows test results of latest build, even if this build not refers to the configured branch.

:link: **Link to the badge**

https://img.shields.io/azure-devops/tests/jopohl/urh/2/master.svg

Note, badge says 2176 tests passed while this should truly be 2720. The number 2176 comes from a build which was made on another branch. Overview of builds is here: https://dev.azure.com/jopohl/urh/_build?definitionId=2

:bulb: **Possible Solution**
Configured branch should be evaluated to get test results from latest build on branch and not latest build overall.

calebcartwright:
Fixes #2774 (the underlying API changed the query param name from `branch` to `branchName`)


paulmelnikow:
Do you think it's worth mentioning in the doc that `/refs/heads/whatever` can be specified as `whatever`?

paulmelnikow:
Are you sure we should bother to support both ways? If we don't support "with" `/refs/heads/` we can always add it if someone complains.

calebcartwright:
Doesn't matter to me either way so I'll change it. Definitely more important to get the fix out so the branch filter will actually work than supporting fancy branch naming patterns!

paulmelnikow:
I was a little preoccupied with the color bugs the last day or so. There was a lot of code moving! 😝

calebcartwright:
> I was a little preoccupied with the color bugs the last day or so. There was a lot of code moving! 😝

I was thinking more along the lines of... why did I add that in the first place? 🤣

PR #2780, opened 2019-01-16
Fix numeric colorB [static githublanguages]

paulmelnikow:
Numeric colors weren't properly being handled by `makeBadge` after #2742.

Since this function really does not need to be accepting colors as strings, rather than make the function more lenient to work with Scoutcamp, I coerced the types of the colors on the way in.

Two tests cover the functionality in the modern service. I don't feel strongly that the legacy version needs coverage at this point, though I've added one for the moment on the github languages badge where this manifested.

Fix #2778

Issue #2687, opened 2019-01-08
Add a helper for service test expectations


paulmelnikow:
Our service tests could be made more concise and consistent by adding an `expectBadge` helper to the Frisby objects we create.

It could look like this:

```js
  .expectBadge({ label: 'something', message: 'something-else', color: 'blue' })
```

It could translate this into multiple assertions on the response, and handle literals and Joi schemas interchangeably. It could also magically translate color aliases to their actual color, so that we don't have to import colorscheme.

I'd rather not specify `?format=_shields_test` everywhere, so maybe we can do this in conjunction with  #2686.

paulmelnikow:
Some unexpected behavior related to our current setup: https://github.com/badges/shields/pull/2780#issuecomment-454879950

PyvesB:
I'm happy to look into this. However, I suggest we tackle the field renaming and color addition parts in #2686 alone, I'm expecting the diff to be quite big already.

What I have in mind for now is to implement a helper that translates calls similar to the following:
```
.expectJSON({ name: 'something', value: 'something-else' })
.expectJSONTypes(Joi.object().keys({ name: 'build', value: isBuildStatus }))
.expectJSONTypes(Joi.object().keys({ name: 'build', value: isBuildStatus, color: 'brightgreen' })
```
to;
```
.expectBadge({ name: 'something', value: 'something-else' })
.expectBadge({ name: 'build', value: isBuildStatus })
.expectBadge({ name: 'build', value: isBuildStatus, color: 'brightgreen' })
```

paulmelnikow:
Thanks for picking this up.

Hmm, I'm not quite sure I'm following how you're proposing to split up the work. Could you clarify what order of operations you're suggesting?

PyvesB:
I'm suggesting this ordering:
* leave the JSON response schema unchanged but introduce a test helper.
* modify the JSON response schema in #2686 and tweak the test helper to operate on the new field names and an always present color field.

Does this make sense?

paulmelnikow:
For the most part, yes!

The thing I'm not clear about is whether you want the calls to use the modern key names:

```js
  .expectBadge({ label: 'something', message: 'something-else', color: 'blue' })
```

or the legacy key names:

```js
  .expectBadge({ name: 'something', value: 'something-else', color: 'blue' })
```

PyvesB:
The old ones, as I don't intend on changing the JSON response.

paulmelnikow:
[...]

PR #2782, opened 2019-01-16
Fix color again, for legacy badges

paulmelnikow:
Ref: https://github.com/badges/shields/pull/2780#issuecomment-454873296

### DIFF
--- a/gh-badges/templates/_shields_test-template.json
+++ b/gh-badges/templates/_shields_test-template.json
@@ -1,9 +1,7 @@
 {
-{{?it.color}}
-  "color": {{=JSON.stringify(it.color)}},
-{{?}}
+  "color": {{=JSON.stringify(it.color || null)}},
 {{?it.labelColor}}
-  "labelColor": {{=JSON.stringify(it.labelColor)}}
+  "labelColor": {{=JSON.stringify(it.labelColor)}},
 {{?}}
   "name": {{=JSON.stringify(it.text[0])}},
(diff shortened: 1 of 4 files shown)


lib/badge-data.js
```diff
@@ -107,8 +107,14 @@ function makeBadgeData(defaultLabel, overrides) {
...
-    colorA: `${overrides.colorA}`,
-    colorB: `${overrides.colorB}`,
+    colorA:
+      typeof overrides.colorA === 'number'
+        ? `${overrides.colorA}`
+        : overrides.colorA,
+    colorB:
+      typeof overrides.colorB === 'number'
+        ? `${overrides.colorB}`
+        : overrides.colorB,
```
paulmelnikow: This was the real issue. It was coercing `undefined` to `'undefined'` which was then taking precedence over any value of `colorscheme`.

gh-badges/templates/_shields_test-template.json
```diff
@@ -1,9 +1,7 @@
 {
-{{?it.color}}
-  "color": {{=JSON.stringify(it.color)}},
-{{?}}
+  "color": {{=JSON.stringify(it.color || null)}},
```
paulmelnikow: This restores `color` always be present in this object, for parity with the old behavior.

If we implement #2687 we can revisit this format. The behavior I mentioned in https://github.com/badges/shields/pull/2780#issuecomment-454879950 was very unexpected and it would be nice if we could depend on this more.

PR #2792, opened 2019-01-17
Reorganize token pooling code

paulmelnikow:
Ref #2698


calebcartwright:
I re-ran the failed services jobs to see if those failures were just a fluke, but the two GH Search tests and two additional ones failed this time.

Not sure to what degree those failures are related to the reorg, if at all.

paulmelnikow:
FWIW I don't think they are related!

calebcartwright:
Yeah I didn't see anything in this PR that jumped out as being the cause of those failures (although I don't know why they are failing)

I'm still 👍 on this change

paulmelnikow:
It's pretty common that some of the GitHub tests will fail. In each test run I looked at, the failures looked different. A lot of them say "invalid." It'll be interesting to see how the messages evolve when the services are refactored. I'm curious what kind of data we're getting back and the new error messages give a stronger idea.

Issue #2479, opened 2018-12-08
Support static preview with social style


paulmelnikow:
The social badges have typically included `?style=social`. However `staticPreview` returns `serviceData` which does not support the `style` parameter. This makes it impossible to render a static preview for the social badges.

e.g. 

https://github.com/badges/shields/blob/f8b0472b4c26a7bafcf4c91a527414054d5ac922/services/github/github-followers.service.js#L32-L36

This also came up in #2473 where I was adding an Endpoint badge, and thinking we should make it possible for a developer to make a badge which defaults to a particular style.

I'm not totally convinced we should do that. In some ways it violates a user's expectation, which is that Shields will always default to flat. Should we also let the developer specify `flat-square` or `for-the-badge`? At the same time, these are _custom_ badges, and it's useful to be able to include especially logos by default. Maybe it's okay, for custom badges.

chris48s:
Given we've only got 6 social badges in the home page examples (out of some-hundred in total), I reckon static preview for social badges would be nice to have but its quite low priority.

paulmelnikow:
It's true it's not so important to render it statically, though it would be nice to drop support for `exampleUrl` which adds complexity in BaseService and the frontend.

PR #2796, opened 2019-01-17
Add logo support for the endpoint badge and refactor logo functions

paulmelnikow:
It was a bit difficult to add logo support for the endpoint badge because of the way the logo overrides were being processed. The overriding logic is common-sensical, though relatively complex, and this decouples the three main parts: extracting an svg logo from the query string, processing an svg logo, and locating and preparing a named logo. There's a bit more code in BaseService than there was before, though outside of legacy code, that function is now the only place the overrides are happening.

It may soon be a good idea to move `_makeBadgeData` into its own file. The function isn't necessarily complex, though it is 100 lines long.

There may be more refactoring to do with the logos. In particular I think it may make sense to move some more of the logo processing into gh-badges, as I did with the color processing. It may also make sense to bundle the Shields logos and simple-icons with gh-badges. (@chris48s as I recall we discussed this a bit but decided not to do it at the time.) That seems cleanest, as the memoization we're doing makes the simple-icons and shields logos want to live alongside the code that colorizes them. It seems useful to have the logos available in the package, and also seems desirable that the endpoint schema and the `makeBadge()` interface be as similar as possible. It might make #2479 simpler to implement, too, since the names of named logos would be passed through, to be resolved later.


services/base.spec.js
```diff
@@ -414,30 +415,77 @@ describe('BaseService', function() {
...
-        const expLogo =
-          'data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSI0MCIgaGVpZ2h0PSI0MCIgdmlld0JveD0iMTIgMTIgNDAgNDAiPgo8cGF0aCBmaWxsPSIjMzMzMzMzIiBkPSJNMzIsMTMuNGMtMTAuNSwwLTE5LDguNS0xOSwxOWMwLDguNCw1LjUsMTUuNSwxMywxOGMxLDAuMiwxLjMtMC40LDEuMy0wLjljMC0wLjUsMC0xLjcsMC0zLjIgYy01LjMsMS4xLTYuNC0yLjYtNi40LTIuNkMyMCw0MS42LDE4LjgsNDEsMTguOCw0MWMtMS43LTEuMiwwLjEtMS4xLDAuMS0xLjFjMS45LDAuMSwyLjksMiwyLjksMmMxLjcsMi45LDQuNSwyLjEsNS41LDEuNiBjMC4yLTEuMiwwLjctMi4xLDEuMi0yLjZjLTQuMi0wLjUtOC43LTIuMS04LjctOS40YzAtMi4xLDAuNy0zLjcsMi01LjFjLTAuMi0wLjUtMC44LTIuNCwwLjItNWMwLDAsMS42LTAuNSw1LjIsMiBjMS41LTAuNCwzLjEtMC43LDQuOC0wLjdjMS42LDAsMy4zLDAuMiw0LjcsMC43YzMuNi0yLjQsNS4yLTIsNS4yLTJjMSwyLjYsMC40LDQuNiwwLjIsNWMxLjIsMS4zLDIsMywyLDUuMWMwLDcuMy00LjUsOC45LTguNyw5LjQgYzAuNywwLjYsMS4zLDEuNywxLjMsMy41YzAsMi42LDAsNC42LDAsNS4yYzAsMC41LDAuNCwxLjEsMS4zLDAuOWM3LjUtMi42LDEzLTkuNywxMy0xOC4xQzUxLDIxLjksNDIuNSwxMy40LDMyLDEzLjR6Ii8+Cjwvc3ZnPgo='
         const badgeData = DummyService._makeBadgeData(
-          { logo: 'github', style: 'social' },
-          {}
+          { logo: 'github' },
+          { namedLogo: 'appveyor' }
```
chris48s: OK, so `logo` is more important than `namedLogo` here for...reasons.

I guess my feedback on this PR is broadly the same as the cache PR. Everything seems to work in staging, but we now have this non-trivial relationship between various ways we could inject a logo in different forms at different layers in the application. i.e: there is some relationship beween: `namedLogo`, `logo`, `defaultLogo`, `overrideLogo`, `logoSvg` etc and some of those override each other in various situations.

I haven't fully worked out what those relationships are from the code, but it should at least be explained somewhere so we can answer the question "if I pass `namedLogo`, `defaultLogo` and `logoSvg` all at the same time, which one is supposed to 'win'?".

Are you thinking of this as a permanent solution, or is it like the cache PR where as we refactor we can remove some of these layers?

chris48s:
> There may be more refactoring to do with the logos. In particular I think it may make sense to move some more of the logo processing into gh-badges, as I did with the color processing. It may also make sense to bundle the Shields logos and simple-icons with gh-badges.

One of the things we should consider with moving the logos code into the NPM package is that simple-icons is a ~2Mb dependency. As I understand it, our reason for wanting to do this is our own internal code structure as opposed to user need. (AFAIK no users of the NPM module have raised an issue requesting logos in the NPM module - link me to an issue if I'm wrong on that). I think we should consider that as we think about that approach. Perhaps simple-icons should be an optional dependency if we do go down that route?

paulmelnikow:
> AFAIK no users of the NPM module have raised an issue requesting logos in the NPM module - link me to an issue if I'm wrong on that

I believe that is correct. There are very few issues about the NPM module at all – I've been surprised how many people actually use it!

The main motive for moving logo resolution there is to make a clean interface for `makeBadge`. Ideally that should be something like the Endpoint schema, which is the schema that most conveniently describes the full contents of a badge. We should be able to fully inspect a badge's contents using the input to `makeBadge` which probably is what our JSON endpoint should return once the legacy services go away. (Running JSON badges through templates is pretty bonkers.

I don't mind making `simple-icons` optional; that seems like a nice way to handle it.

services/base.spec.js
paulmelnikow: > I guess my feedback on this PR is broadly the same as the cache PR. Everything seems to work in staging, but we now have this non-trivial relationship between various ways we could inject a logo in different forms at different layers in the application. i.e: there is some relationship beween: `namedLogo`, `logo`, `defaultLogo`, `overrideLogo`, `logoSvg` etc and some of those override each other in various situations.
> 
> I haven't fully worked out what those relationships are from the code, but it should at least be explained somewhere so we can answer the question "if I pass `namedLogo`, `defaultLogo` and `logoSvg` all at the same time, which one is supposed to 'win'?".

Yea, that makes sense. I'll add a paragraph here.

I think a partial solution is to split out `_makeBadgeData` into its own file and to reorganize its tests so all the logo tests are together, all the label tests are together, all the color tests are together, etc., instead of now, where they're grouped by the kind of test they are. Between a descriptive paragraph, and reading a collection of tests all about logo, it should be clear to work out the desired behavior and reason about the correctness of the implementation.

> Are you thinking of this as a permanent solution, or is it like the cache PR where as we refactor we can remove some of these layers?

There are two things left that I think we can get rid of.

1. If we move named logo resolution into `makeBadge` then that part will be more contained. In particularly it'll be clear where the tests belong. The resolution and preparation logic will be tested around `makeBadge`, and only the overriding logic will need to be tested here. There won't be any less overriding logic, though it should be clearer whose job it is to deal with it and to ensure it's correct.
2. When the GitHub, Twitter, and Liberapay services are refactored, the legacy logo overriding pathway (`makeLogo`) will become an implementation detail of `makeBadgeData`. When the rest of the legacy services are gone, we'll be able to get rid of it entirely.

paulmelnikow: I've added some docs about how the logo options are resolved.

PR #2798, opened 2019-01-17
test:server -> test:core

paulmelnikow:
This renames `npm run test:server` to `npm run test:core` to go along with the reorganization of #2698. `server` has been a bit imprecise, since there's a lot of stuff under test besides the server, and most of the tests don't hit the server itself.

Probably files like `luarocks.spec.js` ought to be run as part of a different target, especially if we build out a separate code-coverage metric for core. Though I'm not in a huge rush to sort that out.


paulmelnikow:
Good call, updated. My search didn't turn up the references because they missed the _last_ update!

PR #2803, opened 2019-01-17
refactor [packagecontrol] service

chris48s:


services/packagecontrol/packagecontrol.service.js
```diff
@@ -1,148 +1,125 @@
...
-          message: '10k/month',
-          color: 'brightgreen',
-        },
-        keywords,
+function DownloadsForInterval(interval) {
+  const { base, messageSuffix, counter } = {
+    day: {
+      base: 'packagecontrol/dd',
+      messageSuffix: '/day',
+      counter: function(resp) {
```
paulmelnikow: What do you think about calling this `transform`? It could use an arrow function.

paulmelnikow: No technical advantage, just stylistic.

I thought it's what we were using in most places for anonymous functions, though perhaps that is not the case. I'll open a new issue.

paulmelnikow: I tried to find a linter rule to deal with this, but couldn't find one. It seems like a fairly unusual case: a function expression which is a property, and not being used as a callback.

We do have a few function expressions among mostly function declarations. I'll open a PR to tidy up those.

Issue #1937, opened 2018-08-17
CircleCI cache on forks


paulmelnikow:
CircleCI caching stopped working on PRs from forks, which is to say, most of our PRs. This is resulting in failed builds on branches which change `package-lock.json`.

The change is a feature, which is a security change that prevents caching on forks. The change makes sense, really, though it's a bit inconvenient for us.

More details on the issue here:
https://discuss.circleci.com/t/saving-cache-stopped-working-warning-skipping-this-step-disabled-in-configuration/24423

I'll open a PR to turn it off for now…

chris48s:
Should we close this issue now that you've disabled the cache in #1938 or is there some other way forward?

paulmelnikow:
I’m hoping they’ll provide a way to fix this. One option is to use workspaces, as folks have noted on that thread. How about we keep it open so we remember to track that thread, and if they don’t fix it within a month, close it out?

PyvesB:
We've moved away from CircleCI for PRs. Gonna go ahead and close this issue as it's no longer relevant.

PR #2808, opened 2019-01-18
Temporarily remove useless npm-install task in CI

paulmelnikow:
It still seems worth using workspace caching to properly tackle #1937, though in the meantime we're wasting time with a useless build. This should cut our total build latency roughly by half.


paulmelnikow:
> LGTM, for my own sake: each individual job is already running it's own npm install (and cache restore) step, so the requires removed here were duplicative right?

Yea, the idea of the npm-install, when it worked, was to seed the cache, which each of the jobs then picked up.

Though, now that you mention it… I think we need to add `save_cache` to one of these.

paulmelnikow:
Okay, I added a `save_cache` step  to `main`. I also switched the checksum over to the lockfile, which is in closer correspondence with the contents of the cache.

PR #2809, opened 2019-01-18
Create shortcut for BaseService-related imports

paulmelnikow:
Continue to implement #2698:

- Add `core/base-service/index.js` (but hold off on moving the things it imports)
- Add shortcuts in `services/index.js` for Base*Service, errors, and deprecatedService. This file will be streamlined later to avoid cluttering it with rarely used bits.
- Apply consistent ordering of imports and use of `module.exports` in testers.
- Remove some renaming of imports.
- Remove obsolete tests here and there.


paulmelnikow:
Fix for failing test is in #2810.

calebcartwright:
I changed the title to include the new token for all services 5 `*` although now I'm wondering if the intent was to run all service tests here? 

My apologies if not!

paulmelnikow:
Service test failures seem unrelated.

paulmelnikow:
This is drifting out of date. It would be great to get this reviewed soon!

calebcartwright:
Sorry I'll try to look at it today. I keep starting on it but there's just so many files that I don't make it very far 😄

doc/TUTORIAL.md
```diff
@@ -170,8 +170,8 @@ This example is based on the [Ruby Gems version](https://github.com/badges/shiel
 ```js
 'use strict'                                                    // (1)
 
-const BaseJsonService = require('../base-json')                 // (2)
-const { renderVersionBadge } = require('../../lib/version')     // (3)
+const { renderVersionBadge } = require('../../lib/version')     // (2)
+const BaseJsonService = require('..')                           // (3)
```
calebcartwright: Should this be `const { BaseJsonService } = require('..')`?

doc/service-tests.md
```diff
@@ -37,11 +37,19 @@ We'll start by adding some boilerplate to our file:
...
 
-const t = (module.exports = require('../create-service-tester')())
+const t = (module.exports = require('..').createServiceTester())
 ```
 
-1. Import [Joi][] We'll use this to make assertions. This is the same library we use to define schema for validation in the main badge class.
-2. If our `.service.js` module exports a single class, we can `require('../create-service-tester')` and use convention to create a `ServiceTester` object. Calling this inside `services/wercker/wercker.tester.js` will create a `ServiceTester` object configured for the service exported in `services/wercker/wercker.service.js`. We will add our tests to this `ServiceTester` object `t`, which is exported from the module.
+1. Import [Joi][] We'll use this to make assertions. This is the same library
+   we use to define schema for validation in the main badge class.
+
```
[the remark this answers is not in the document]
paulmelnikow: No, I shortened the lines since I was editing it.

doc/TUTORIAL.md
```diff
@@ -253,7 +253,16 @@ Specifically `BaseJsonService` will handle the following errors for us:
 * API returns a response which can't be parsed as JSON
 * API returns a response which doesn't validate against our schema
 
-Sometimes it may be necessary to manually throw an exception to deal with a non-standard error condition. If so, standard exceptions can be imported from [errors.js](https://github.com/badges/shields/blob/master/services/errors.js) and thrown.
+Sometimes it may be necessary to manually throw an exception to deal with a
+non-standard error condition. If so, standard exceptions can be imported from
+[errors.js](https://github.com/badges/shields/blob/master/services/errors.js)
+and thrown:
```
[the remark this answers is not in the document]
paulmelnikow: Yea, that's a good idea. Feel free to edit!

doc/TUTORIAL.md
paulmelnikow: Gooood catch!

services/dub/dub-download.service.js
```diff
@@ -1,11 +1,9 @@
 'use strict'
 
 const Joi = require('joi')
-const BaseJsonService = require('../base-json')
 const { metric } = require('../../lib/text-formatters')
-const {
-  downloadCount: downloadCountColor,
```
calebcartwright: Sidebar comment: I know it's been mentioned in other threads we want to remove the import renames like this, which makes sense, but there's one formatter (I think it's `version`?) where the function name is identical to the most frequently used parameter name. What do folks think about tweaking some of the format function names to more explicitly reference the formatting aspect? 

Presumably folks were renaming those imports for a similar reason?

paulmelnikow: Yes, agreed.

The history sort of goes the other way. When those helper functions were pulled out into modules, some of them were given new names (and then the renames were placed at the top of 7k-line `server.js`). When I split up `server.js` I moved the route handlers intact so the renames got copied along with them.

A lot of the new names were good standalone, though a few of them were styled in a way that made them difficult to understand without the context of their module. I suppose scoping them under their module is a reasonable stylistic choice, though it seems hard to maintain in a project like this.

calebcartwright:
😌 made it through! I bet the changes in this PR were a bit tedious to make 😉 

I noticed that service validators were not included in the new shortcut. Were those left out intentionally and if so why?

services/dub/dub-download.service.js
calebcartwright: Gotcha. It's the `version` color formatter function that always makes me feel guilty since I almost always have to rename the import.

paulmelnikow:
> I noticed that service validators were not included in the new shortcut. Were those left out intentionally and if so why?

The shortcut namespace obscures where the helper code actually lives, and on the flip side, makes it harder to track down where a helper function is used. I think the negative of the obscuring is worth it for these imports which we use, and tack on and off so frequently, though not great for huge namespaces with dozens of things. I think it's better if we import the validators and helper functions in an ordinary way, and reserve the shortcut import for the testers, the base classes, and the errors.

calebcartwright:
I made some minor text updates per above discussion on errors in the tutorial and have approved. I'll re-approve those once the merge conflicts are resolved as well.

Not sure what to do about danger job failing (presumably due to github rate limit issues) 🤔

paulmelnikow:
> Not sure what to do about danger job failing (presumably due to github rate limit issues) 🤔

The failure seems limited to this PR. I wonder if Danger is making too many requests (in parallel?) because there are a lot of files in flight in this particular PR.

doc/deprecating-badges.md
```diff
@@ -40,10 +40,10 @@ module.exports = deprecatedService({
 
 Locate the test file(s) for the service, which can be found in `*.tester.js` files located in the service directory (`./services/:service-name/`), such as `./services/imagelayers/imagelayers.tester.js`.
 
-With `DeprecatedService` classes we cannot use the utility functions from `create-service-tester.js` so you will need to create the `ServiceTester` class directly. For example:
+With `DeprecatedService` classes we cannot use `createServiceTester()` so you will need to create the `ServiceTester` class directly. For example:
 
 ```js
-const ServiceTester = require('../service-tester')
+const { ServiceTester } = require('..')
```
calebcartwright: Good catch!

calebcartwright:
`token pool exhausted` error on the integration tests. 

Do we have a way of rehydrating the pool in the CI jobs or will we just have to wait for the rate limit throttle/block to end?

paulmelnikow:
> Do we have a way of rehydrating the pool in the CI jobs or will we just have to wait for the rate limit throttle/block to end?

We could reconfigure the variable but I don't think it's worth it. Alternatively we could temporarily make the check optional but again probably just as easy to wait a few minutes.

It seems like GitHub is returning an abuse-specific error. If they are, we could try to improve our error message in this case. It's a bit different than using up all of one's quota.

paulmelnikow:
We could also pull the GitHub integration tests into their own Circle job and make it optional in GitHub. These tests do fail more often than the others.

PR #2810, opened 2019-01-18
Rewrite test of analytics endpoint

paulmelnikow:
This test is being weirdly flaky in #2809. The problem seems to be in the test helper code, so I rewrote this using Joi.

I imagine the change has to do with a change to the test ordering. It's a bit puzzling.

However, the new test seems fine (and the endpoint is rarely used; not critical to begin with).

### DIFF
--- a/core/server/server.spec.js
+++ b/core/server/server.spec.js
@@ -137,6 +137,7 @@ describe('The server', function() {
             .required()
         )
         .length(36)
+        .required()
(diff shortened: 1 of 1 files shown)


paulmelnikow:
An unrelated test is failing due to what looks like triggering GitHub abuse limits. Hopefully in time this will clear itself up.

PR #2813, opened 2019-01-19
Avoid func declarations using `function` keyword; affects [packagist]

paulmelnikow:
We had only a few function expressions declared with the function keyword; almost everything is using function declarations.

This came up after this discussion: https://github.com/badges/shields/pull/2803#discussion_r249011621 though is actually not related ot that example.

What’s being removed is a third option, which is assigning to a variable a function expression using the `function` keyword. There’s still room for the programmer to choose between arrow function expressions and function declarations.


services/response-fixtures.js
```diff
@@ -1,9 +1,9 @@
 'use strict'
 
-const invalidJSON = function() {
```
RedSparr0w: Do you think it would be better to keep `function invalidJSON() {` here rather than in `module.exports`?

paulmelnikow: They seem equally clear to me. I moved it inline since it's more of a fixture than a real function.

RedSparr0w: No worries :+1:

PR #2817, opened 2019-01-20
fix logo imports in [twitter liberapay]

chris48s:
this was moved in #2796 but we missed updating some of the imports


chris48s:
Hmm. I'm not sure why those 2 tests are failing on `GithubSearch`. They both pass if I run the test suite locally

paulmelnikow:
Those badges are often a bit flaky, and I think our GitHub token in CI is sometimes getting abuse limited. I don't think it's related.

PR #2823, opened 2019-01-21
Add directory field to package.json

felixfbecker:
https://github.com/npm/rfcs/blob/latest/accepted/0010-monorepo-subdirectory-declaration.md

This helps tools find this package in the repository.


chris48s:
Have we... run out of GitHub? :open_mouth:

paulmelnikow:
We're getting abuse rate limited. We use GitHub in a lot of ways so there are a lot of possibilities, though this has only happened twice and both times was while working on https://github.com/badges/shields/pull/2809. That makes me suspect that Danger is triggering more requests when there are lot of files changed. #2825 could also help, or perhaps dialing down the number of requests made in the integration tests.

Aha! https://twitter.com/orta/status/1087439065321033735

chris48s:
OK. Lets move the conversation over to #2826 instead of hijacking this unrelated PR

PR #2825, opened 2019-01-21
Split integration tests into their own stage

paulmelnikow:
These tests fail more often than others, and sometimes we run into abuse rate limits. It's inconvenient to hold up merging when the changes are unrelated.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,14 +58,6 @@ jobs:
           command: npm run test:core
 
       - run:
-          name: Integration tests
-          when: always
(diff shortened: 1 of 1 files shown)


paulmelnikow:
Heh, we still can't merge this until `services` passes, though I still think it's a good idea and should make it easier to getting flaky tests through CI.

paulmelnikow:
@calebcartwright Shall we merge this before I update #2809 and find out whether the change from #2827 prevents us from being rate-limited again?

PR #2856, opened 2019-01-23
pre-compute the 3 most common icon styles on server init

chris48s:
refs https://github.com/badges/shields/pull/2833#discussion_r249597638

was this what you had in mind?


RedSparr0w:
Nice work!

PR #2857, opened 2019-01-23
Delete some of our logos (part 1); affects [endpoint]

chris48s:
Refs #2510

I'm going to delete or change some more logos in a further PR or two, but lets start off with the (hopefully) non-controversial ones. I think in all of these cases it is fairly clear-cut that we are not losing anything by removing our icon in favour of simple-icons now that we apply a sensible colour by default.

| Before | After |
|---|---|
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  `*` |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |

`*` Note this one is out-of-date now, but there's a PR open to update to the new slack logo: https://github.com/simple-icons/simple-icons/pull/1172 This is one of the other nice things about not maintaining our own logos.

There are a few more to go, but I'll be quite glad when I don't need to make any more of these tables! :)


chris48s:
Gah - I'd forgotten that there are some tests and services which directly reference logos - thanks, Continuous Integration! I'll come back to this and get all the tests passing tomorrow. The visuals shouldn't change though.

RedSparr0w:
Nice work on this, looking really good :+1:

paulmelnikow:
Nice! The tables really are so much work, it's cool this is moving ahead.

Re two of these:
- I think the shape of our telegram logo looks better. They seem mix 3d-looking logos with flat-looking ones (like the simple-icons version). I suppose since their favicon looks a lot like the simpleicons logo we may as well use the flat version.
- I think our multicolor bitcoin looks better and it's traditionally orange with white. Can we keep this one around with the multicolor logos?

chris48s:
mostly non-controversial ;)

Issue #2838, opened 2019-01-22
Announcing the Endpoint badge (beta)


paulmelnikow:
Hi all, Shields has launched a beta features: the [Endpoint badge][].

**What is the Endpoint badge?**

The endpoint badge is a new kind of dynamic badge endpoint which renders a badge from a JSON endpoint.

**What is it useful for?**

1. Rapidly prototype new ideas and see the results on live badges before those features make it into Shields.
2. Handle use cases deemed too specific to include in Shields.
3. Implement things that Shields won't implement (e.g. for performance or stylistic reasons).

**How does it work?**

- Your endpoint returns static or dynamic JSON data in the Endpoint schema.
- Your user can override formatting via the query string.
- Shields take care of formatting, rendering, and caching.

**Where can I host the JSON endpoint?**

Literally anywhere that supports HTTPS.

- [Tools like RunKit and Jupyter Kernel Gateway][tools]
- Any kind of JSON-returning cloud function
- Any server of your choosing
- Any static JSON file

**Feedback**

Your feedback on the new feature is much appreciated! Please feel free to comment in this issue with feedback or questions about how to use it. The #support room in Discord is also a great place to discuss it.

[Endpoint badge]: https://shields.io/#/endpoint
[tools]: https://github.com/badges/shields/issues/2259#issuecomment-444186589

**Known issues**

- [x] #2837
- [x] #2998

chris48s:
Good work on shipping this :+1: I think this really opens the door to more of a "plugin" style architecture - users will be able to enjoy a wider range of more esoteric integrations and use-cases without us necessarily having to host a service integration in core for every single programming based service under the sun.

Sorry I didn't get a chance to raise this at the review stage, but one thing that I think is likely to be a point of confusion is the cache key. I think some users will try to set it to 0 and get unexpected results. It would be useful to more clearly communicate that there is a floor on the cache length (but as I remember it, its a bit fiddly to surface the value because the variable isn't set when we build the front-end). Is there a way we can more clearly communicate this in the docs?

paulmelnikow:
> Sorry I didn't get a chance to raise this at the review stage, but one thing that I think is likely to be a point of confusion is the cache key. I think some users will try to set it to 0 and get unexpected results. It would be useful to more clearly communicate that there is a floor on the cache length (but as I remember it, its a bit fiddly to surface the value because the variable isn't set when we build the front-end). Is there a way we can more clearly communicate this in the docs?

Yea, good shout.

[...]

PR #2858, opened 2019-01-23
Provide friendlier error messages for [endpoint]

paulmelnikow:
Ref https://github.com/badges/shields/issues/2838#issuecomment-456594803

### DIFF
--- a/core/base-service/validate.spec.js
+++ b/core/base-service/validate.spec.js
@@ -89,7 +89,10 @@ describe('validate', function() {
         try {
           validate(
             { ...options, includeKeys: true },
-            { requiredString: ['this', "shouldn't", 'work'] },
+            {
+              requiredString: ['this', "shouldn't", 'work'],
+              requiredNumber: 'neither this',
+            },
             schema
           )
           expect.fail('Expected to throw')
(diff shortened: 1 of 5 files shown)


core/base-service/validate.spec.js
```diff
@@ -107,12 +110,19 @@ describe('validate', function() {
...
       )
-    ).to.throw(InvalidParameter, '"extra" is not allowed')
+      expect.fail('Expected to throw')
+    } catch (e) {
+      expect(e).to.be.an.instanceof(InvalidParameter)
+      expect(e.message).to.equal(
+        'Invalid Parameter: "extra" is not allowed. "more" is not allowed'
+      )
+      expect(e.prettyMessage).to.equal(`${prettyErrorMessage}: extra, more`)
+    }
```
paulmelnikow: These are new tests. The functionality is basically the same except that I've inserted a space after the comma.

services/endpoint/endpoint.tester.js
```diff
@@ -146,7 +146,10 @@ t.create('Invalid schema (mocked)')
         schemaVersion: -1,
       })
   )
-  .expectJSON({ name: 'custom badge', value: 'invalid response data' })
+  .expectJSON({
+    name: 'custom badge',
+    value: 'invalid properties: schemaVersion',
```
calebcartwright: I'm a little confused in terminology differences (`properties` vs `parameters` vs `keys`)? Are they all synonymous and if not which is which in this context?

paulmelnikow: "Keys" is the name that's internal to the code. **properties** seems like the friendliest way to explain to the endpoint provider that the object they're returning doesn't have the right stuff in it. I think it's a little clearer than keys.

I think we use the word "parameters" for query parameters, which makes sense.

calebcartwright: Ahh okay, i forgot this was the reverse end. it would be invalid/unexpected properties in the response object the endoint was returning that would trigger that right?

paulmelnikow: Yea, correct. In this case `schemaVersion` is set to an invalid value.

calebcartwright: Thanks, I'm up to speed now 👍

PR #2859, opened 2019-01-23
Refactor _makeBadgeData -> coalesceBadge; test on [static dynamic endpoint]

paulmelnikow:
`base.js` is pretty long and `_makeBadgeData` is one of the most complex
functions in it.

It's a pure function so it's easy to test in isolation. This moves the function into its own module and reorganizes the tests in a way that makes it evaluate what it's doing, and easier to test what is and isn't covered.

Ref https://github.com/badges/shields/pull/2796#discussion_r249251008


core/base-service/coalesce-badge.spec.js
```diff
@@ -0,0 +1,257 @@
...
+    it('uses the category as a default label', function() {
+      expect(coalesceBadge({}, {}, {}, { category: 'cat' }).text).to.deep.equal(
+        ['cat', 'n/a']
+      )
+    })
+
+    it('preserves an empty label', function() {
+      expect(
+        coalesceBadge({}, { label: '', message: '10k' }, {}).text
+      ).to.deep.equal(['', '10k'])
```
calebcartwright: Just an arbitrary comment that I'm surprised prettier aligned these three `expect` blocks in these three different ways 😄

core/base-service/coalesce-badge.spec.js
```diff
@@ -0,0 +1,257 @@
...
+describe('coalesceBadge', function() {
+  describe('Label', function() {
+    it('uses the default label', function() {
+      expect(coalesceBadge({}, {}, { label: 'heyo' }).text).to.deep.equal([
+        'heyo',
+        'n/a',
+      ])
+    })
+
+    // This behavior isn't great and we might want to remove it.
```
calebcartwright: What's the thinking on removing it? I actually just noticed this not too long ago when I forgot to set a default badge label and noticed the category being used in my local env. If we removed this, would there be an error thrown (programmer error) or what would the label show as alternatively?

core/base-service/coalesce-badge.spec.js
```diff
@@ -0,0 +1,257 @@
...
+
+    it('applies the service label color', function() {
+      expect(coalesceBadge({}, { labelColor: 'red' }, {}).labelColor).to.equal(
+        'red'
+      )
+    })
+
+    it('overrides the label color', function() {
+      expect(
+        coalesceBadge({ colorA: '42f483' }, { color: 'green' }, {}).labelColor
```
calebcartwright: Just trying to get my bearings around `colorA` vs `labelColor` and `colorB` vs `color`. Are the two on the left (A & B) still valid, but deprecated aliases for the labelColor and color or are A & B still undeprecated/promoted?

core/base-service/coalesce-badge.spec.js
```diff
@@ -0,0 +1,257 @@
...
+    it('when a social badge, uses the default named logo', function() {
+      // .not.be.empty for confidence that nothing has changed with `getShieldsIcon()`.
+      expect(
+        coalesceBadge({ style: 'social' }, {}, { namedLogo: 'appveyor' }).logo
+      ).to.equal(getShieldsIcon({ name: 'appveyor' })).and.not.be.empty
+    })
+
+    it('applies the named logo', function() {
+      expect(coalesceBadge({}, { namedLogo: 'github' }, {}).logo).to.equal(
+        getShieldsIcon({ name: 'github' })
```
calebcartwright: this will need to be switched to `getSimpleIcon` once #2857 is merged right?

core/base-service/base-static.js
```diff
@@ -33,8 +34,12 @@ module.exports = class BaseStaticService extends BaseService {
...
       )
 
-      const badgeData = this._makeBadgeData(queryParams, serviceData)
-      // The final capture group is the extension.
+      const badgeData = coalesceBadge(
+        queryParams,
+        serviceData,
+        this.defaultBadgeData,
+        this
+      ) // The final capture group is the extension.
```
calebcartwright: The comment applies to the line below right (`const format = match.slice(-1)[0]`)? Might be better to have the comment on its own line instead of on the closing parenthesis.

core/base-service/coalesce-badge.spec.js
paulmelnikow: Throwing an error is a good idea.

paulmelnikow: I think @chris48s was the first person who articulate this was a bad/unexpected use of the category.

core/base-service/coalesce-badge.spec.js
paulmelnikow: As of now, still undeprecated.

core/base-service/coalesce-badge.spec.js
paulmelnikow: Heh yep. Or to `npm`.

core/base-service/base-static.js
paulmelnikow: Yes, go for it… though tbh I doubt I put that there so it probably was Prettier. If it doesn’t stay where you want you could try adding a blank line.

paulmelnikow:
Could if you’d like! I think the coverage is there in “core” though I’ve zero objections if you’d like to run more tests.

core/base-service/coalesce-badge.spec.js
calebcartwright: Gotcha, thanks

PR #2860, opened 2019-01-24
Refactor route functions in BaseService

paulmelnikow:
The route helper functions are fairly well isolated from the rest of BaseService, with a few convenient entry points. They are easier to test in isolation.

The way the code was written before, `pathToRegexp` was invoked once for every request, which seems inefficient.

`route` was validated when it was used, though it seems more helpful to validate it up front.

This breaks out `_makeFullUrl`, `_regex`, `_regexFromPath` into new helper functions `makeFullUrl`, `assertValidRoute`, `prepareRoute`, and `namedParamsForMatch`.

It adds validation to route, and updates the services without patterns to include one, in order to pass the new validation rules.


core/base-service/route.js
```diff
@@ -0,0 +1,75 @@
...
+
+function makeFullUrl(base, partialUrl) {
+  return `/${[base, partialUrl].filter(Boolean).join('/')}`
+}
+
+const routeSchema = Joi.object({
+  base: Joi.string().allow(''),
+  pattern: Joi.string().allow(''),
+  format: Joi.string(),
+  capture: Joi.alternatives().when('format', {
```
calebcartwright: and today the thing I learned about `Joi` was ^^

core/base-service/route.js
```diff
@@ -0,0 +1,75 @@
...
+  Joi.assert(route, routeSchema, message)
+}
+
+function prepareRoute({ base, pattern, format, capture }) {
+  let regex, captureNames
+  if (pattern === undefined) {
+    regex = new RegExp(
+      `^${makeFullUrl(base, format)}\\.(svg|png|gif|jpg|json)$`
+    )
+    captureNames = capture || []
```
calebcartwright: Should the schema validation ensure `capture` is set by the time we get here?

paulmelnikow:
Yea, that’s a good question.

IIRC the dynamic, endpoint, and debug badges all look like that.

core/base-service/route.js
paulmelnikow: There’s no `.required()` on the array, so I think the answer is no. However if it’s easy to make that happen I’d be down to make the change.

calebcartwright: Yeah this made more sense to me as I went through the rest of the code. Even if there were some Joi magic there's no guaranteed it would have been called. 

I get it now 👍

paulmelnikow:
No, it’s orthogonal to the refactor. The legacy badges entirely sidestep `route` (apart from its involvement in the examples).

The problem is with badges, modern or needing-to-be-refactored legacy, which depend on things like negative lookahead, arbitrary non-capturing groups, or “or” patterns.

I think registering multiple patterns per service could help with some of them. Redirecting obsolete patterns to a canonical pattern could help with others. Finally, registering patterns with priority order could help.

I really like the limitations of `path-to-regexp`, because it keeps us from doing complicated things that are harder to express. However we’ve a bit of ground to cover before it’s not needed.

calebcartwright:
> arbitrary non-capturing groups,

As soon as you said I remembered facing that one myself haha 🤦‍♂️ Fortunately though that forced me to rethink what I was trying to do with that regex and I was able to figure out a simpler way to get the same result (with `pattern`) 

Noted though, thanks for the explanation. `format` shall live on for the forseeable future

PR #2869, opened 2019-01-25
Add semantic color keywords and improve formatting of color examples

### LINKED ISSUE #1522: semantic colorschemes & color keywords
thoughts on adding a few new colorschemes?

semantic | color
---|---
error |  red
notice | orange
success | brightgreen
info | blue
default | lightgrey

currently, setting the color is completely non-semaantic

Also: is there a reason [color keywords](https://www.w3.org/TR/SVG/types.html#ColorKeywords) don't work?

[doc/gh-badges.md](https://github.com/badges/shields/blob/master/doc/gh-badges.md#defaults) ("CSS/SVG Color") links to  http://www.w3.org/TR/SVG/types.html#DataTypeColor
which in turn references the color keywords doc:  https://www.w3.org/TR/SVG/types.html#ColorKeywords

paulmelnikow:
This is based on the list I proposed at #1522 (comment). As I started documenting `default` I realized it didn't feel quite right. It's not semantic in relation to the content the way the others are, and it's also not the default left color. I changed it to `disabled` which isn't perfect, but seems better. I'm open to other suggestions.

I updated `gh-badges/readme.md` but the colors won't render correctly until this is deployed

Close #1522


core/badge-urls/make-badge-url.js
```diff
@@ -61,8 +61,8 @@ function staticBadgeUrl({
   style,
   format = 'svg',
 }) {
-  if (!label || !message) {
-    throw Error('label and message are required')
+  if (!message) {
```
calebcartwright: What are the implications of `label` no longer being required?

paulmelnikow: It lets you use this function to make a badge without a label, nothing beyond that.

calebcartwright: 👍

calebcartwright:
I'm still working my way through `color.js`, but on the alias for `default` it looks like you went with `inactive` instead right? `inactive` or `disabled` both make sense to me. The only other word I mentally associate with that color is `inaccessible` but not sure that makes sense in this context

gh-badges/lib/color.js
```diff
@@ -29,6 +42,8 @@ function normalizeColor(color) {
     return undefined
   } else if (color in namedColors) {
     return color
+  } else if (color in aliases) {
+    return aliases[color]
```
calebcartwright: Could we use `resolvedAliases` here instead? It's been a long day and my brain is struggling sorry 😄

paulmelnikow: The idea of this function is to take the input color and return something that is normalized, yet still human readable and pleasant to test. For named colors it passes through their names; they don't translate to their hex equivalents. So, the appropriate response for the aliases is the sort of normalized / canonical color name (e.g. `red` -> `red`; `critical` -> `red`).

paulmelnikow:
> I'm still working my way through `color.js`, but on the alias for `default` it looks like you went with `inactive` instead right? `inactive` or `disabled` both make sense to me. The only other word I mentally associate with that color is `inaccessible` but not sure that makes sense in this context

Aha yes, `inactive`. I like that better than `disabled` which is what I had mistakenly written in the top comment. If it were an action button it would be clearer, but as an informational button that's not displaying any information… what do you call that? Inactive is the best name I've thought of for that situation.

While `unknown` could work semantically, as a color specification, it seems really confusing. We do use this color when information is inaccessible, though `inaccessible` again feels like an odd way to specify it.

We could be really explicit and call it something like `information-missing` or `no-content`.

gh-badges/lib/color.js
calebcartwright: Ah okay, thanks for the explanation. I was looking at it from the perspective of being called from within `toSvgColor` but totally missed that these can (and I believe are) called independently

paulmelnikow:
@RedSparr0w @chris48s Any thoughts / objections?

paulmelnikow:
If there are no objections I'll merge this at the end of the weekend.

paulmelnikow:
@calebcartwright want to merge this?

chris48s:
Didn't get a chance to look at this one before it was merged. Broadly I'm on board with it. I find "inactive" a confusing label, but I'm not sure I have a better suggestion :man_shrugging:

paulmelnikow:
Hasn't deployed yet so it's still a good time to tweak it. Among the alternatives, `no-content` is my favorite, though not sure it's better.

PR #2870, opened 2019-01-25
Refactor [GemDownloads] for readability; improve error messages

### LINKED ISSUE #1960: Gem downloads: generate label text in render function
see https://github.com/badges/shields/commit/1a5d94ac59c8882e00ece563740742c169744093#commitcomment-30270656

paulmelnikow:
Close #1960


services/gem/gem-downloads.service.js
```diff
@@ -159,7 +169,7 @@ module.exports = class GemDownloads extends BaseJsonService {
         pattern: 'dt/:gem',
```
calebcartwright: Could we add some title descriptors/differentiators? May just be me since I've never done much with Ruby but it wasn't immediately clear to me which example was which. I know those weren't changed in this PR, just a general thought

paulmelnikow: Ha, I just tried to do this and I'm actually not sure what the differences are. `dt` is total downloads, and I think `dtv` is downloads for latest version. But then I'm not sure what the difference is between that and `dv` with `stable`.

paulmelnikow: @chris48s Do you know off hand what these three routes represent?

Either way, let's tackle this in a follow-on.

PR #2871, opened 2019-01-25
 Social badge support for static previews

### LINKED ISSUE #2479: Support static preview with social style

paulmelnikow:
The static previews don't support the social badges. Adding that lets us remove support for `exampleUrl`. Close #2479.

This includes `style` and `namedLogo` in the service-definition export and updates the frontend to use it. To accomplish this, it passes `namedLogo` through `coalesceBadge`. After logo resolution is moved to `makeBadge` this duplication can be removed, as `logo` will no longer be needed in the result of `coalesceBadge`.


frontend/components/markup-modal/builder-common.js
```diff
@@ -32,11 +32,12 @@ BuilderContainer.propTypes = {
 const labelFont = `
   font-family: system-ui;
   font-size: 11px;
-  text-transform: lowercase;
 `
 
 const BuilderLabel = styled.label`
   ${labelFont}
+
+  text-transform: lowercase;
```
paulmelnikow: Don't lowercase capital letters in example text (e.g. **e.g. Follow**).

frontend/components/markup-modal/query-string-builder.js
```diff
@@ -48,8 +48,9 @@ export default class QueryStringBuilder extends React.Component {
     })
 
     const badgeOptions = {}
-    supportedBadgeOptions.forEach(({ name, defaultValue = '' }) => {
-      badgeOptions[name] = defaultValue
+    const defaults = { style: defaultStyle }
```
paulmelnikow: Allow the caller to dynamically set the default style.

calebcartwright:
I haven't looked at most of these files before, but the changes here look pretty straightforward to me 👍

core/base-service/transform-example.js
```diff
@@ -143,13 +144,21 @@ function transformExample(inExample, index, ServiceClass) {
...
       staticPreview,
       ServiceClass.defaultBadgeData,
       ServiceClass
     )
-    preview = { label, message: `${message}`, color }
+    preview = {
+      label,
+      message: `${message}`,
+      color,
+      style: style === 'flat' ? undefined : style,
```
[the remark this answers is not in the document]
paulmelnikow: `flat` is the default style so it's what gets used for examples even if it's not specified.

PR #2872, opened 2019-01-25
Delete more icons

chris48s:
These are more subjective, but I think switching to simple-icons for these 2 gives us something which is clearer by default on both main badge styles.

## Scrutinizer
**before:**
* 
* 

**after:**
* 
* 

## Superuser
**before:**
* 
* 

**after:**
* 
*


paulmelnikow:
Again, appreciate your working through these!

Scrutinizer looks great 👍

The proportions on Superuser are way better post-change, however I like our two-color version. Do you have thoughts on under what circumstances we would keep a two-color logo?

chris48s:
Its probably a bit of an unfair comparison. If I apply the same process as #2873 to the 2-colour version so it properly fills the viewbox, you get this:

* 
* 

My reason for proposing the switch is that I think the black on dark grey doesn't provide enough contrast, but I don't have deeply held views on it. As I say, these ones are pretty subjective. Happy to just update the 2-colour version one with this version if you prefer.

paulmelnikow:
It’s subjective for sure. The light one looks great, obvs. The dark one is borderline though I like it and think works okay.

Alternatively, we could make it display all blue on the dark background.

chris48s:
Lets go with the 2-colour one but scaled correctly.

> we could make it display all blue on the dark background.

Looking at what logos are left, there are a few multi-colour logos where applying `logoColor` does something sensible-ish (this would be one of them), but mostly it doesn't. I think what we should probably do is try and make sure any multi-colour logos we hold in our own icon set look good on both of our default backgrounds, allow the `logoColor` param only if the logo defines a single `fill=""` attribute in the `<svg ` tag (e.g: dependabot, discord) but just ignore `logoColor` if the logo uses >1 colours. That's another PR though..

PR #2873, opened 2019-01-25
optimise logos

chris48s:
In this PR I've applied a number of minor technical optimisations to some of the multi-colour logos we want to keep in preference to simple-icons:

* I've put al of the icons in a square `viewbox`, and scaled them to fit the viewbox, preserving the icon's original proportions so that the icon is touching at least two sides of the viewbox. Some of our icons had space around them.
* I've centered the icons horizontally and vertically. Some of our icons were off-centre.

Those 2 things help the icons to display in a more consistent location within the badge.

* I've removed any extraneous attributes/groups/etc (this makes the SVGs slightly smaller).

Visually this results in only small changes to the icon size/placement:

| Before | After |
|---|---|
|  |  |
|  |  |
|  |  |
|  |  |
|  |  |


RedSparr0w:
Nice work 👍 
Those logos look quite a bit nicer!

PR #2879, opened 2019-01-27
refactor [liberapay] service

chris48s:


services/liberapay/liberapay-goal.service.js
```diff
@@ -0,0 +1,41 @@
...
+        namedParams: { entity: 'Changaco' },
+        staticPreview: this.render({ percentAcheived: 33 }),
+      },
+    ]
+  }
+
+  static render({ percentAcheived }) {
+    return {
+      label: 'goal progress',
+      message: `${percentAcheived}%`,
```
calebcartwright: Would decimal precision ever be a consideration here? Not sure whether the underlying data points could ever allow for an ugly decimal here or not

services/liberapay/liberapay-goal.tester.js
```diff
@@ -0,0 +1,31 @@
...
+
+const Joi = require('joi')
+const t = (module.exports = require('..').createServiceTester())
+
+t.create('Goal Progress (valid)')
+  .get('/Liberapay.json')
+  .expectJSONTypes(
+    Joi.object().keys({
+      name: 'goal progress',
+      value: Joi.string().regex(/^[0-9]+%/),
```
calebcartwright: Could we use `isPercentage` from test-validators here?

services/liberapay/liberapay-goal.service.js
chris48s: We're already rounding it to the nearest int in `handle()`

calebcartwright: My bad! Missed that

PR #2882, opened 2019-01-28
Drop-downs for multiple choice in patterns

paulmelnikow:
~~This isn't quite done, though I'm curious to get some input on the idea.~~

You can see it in effect on the PyPI downloads and the David badges.

screen shot 2019-01-28 at 5 42 37 pm

screen shot 2019-01-28 at 5 42 46 pm


paulmelnikow:
I wager this is ready for review 👍

calebcartwright:
Haven't looked at it live or even a line of code yet, but very intriguing idea!

services/david/david.service.js
```diff
@@ -25,30 +25,10 @@ module.exports = class David extends LegacyService {
     return [
       {
         title: 'David',
-        pattern: ':user/:repo',
+        pattern: ':kind(dev|optional|peer)?:user/:repo',
         namedParams: { user: 'expressjs', repo: 'express' },
```
calebcartwright: The examples were one of the things I was wondering about. Will we need to have multiple entry for each option, or only in certain cases, or something else?

calebcartwright:
This will only be enabled on badges where the examples have the conditional pattern right? The work we did recently on the Visual Studio Marketplace badges for Azure DevOps includes the same sort of options, so I wanted to check to see how it looked in the UI as the options were longer there (`total|services|onprem`)

calebcartwright:
Another thought/question I had was around any potential variances on example documentation and/or keywords that sometimes vary between examples (although admittedly I'm not sure if that ever happens on any of the multi-choice/`which` types of parameters anyway)

calebcartwright:
It's an awesome user experience! It's nice to be able to view/switch right there in the context of the modal without having to close to go back, then click to open the next, etc.

services/david/david.service.js
paulmelnikow: I think we can start with all the cases where we have patterns which are identical, other than a multiple-choice option. The ones that come to mind are the download badges, the `v`/`vpre` badges, and a few with parameters called `:which`, like GitHub pull request detail.

It may turn out that we want to define the terms in the UI, though to start, that could be done in documentation.

paulmelnikow:
> This will only be enabled on badges where the examples have the conditional pattern right?

Exactly right.

> The work we did recently on the Visual Studio Marketplace badges for Azure DevOps includes the same sort of options, so I wanted to check to see how it looked in the UI as the options were longer there (`total|services|onprem`)

I bet they'd look good. Feel free to push a couple more to this branch if you'd like.

> Another thought/question I had was around any potential variances on example documentation and/or keywords that sometimes vary between examples (although admittedly I'm not sure if that ever happens on any of the multi-choice/`which` types of parameters anyway)

I don't think I've seen a _service class_ where the documentation or keywords varied.

One though on a general trajectory here: it may be that we move toward may be a listing where there's an entry per service, with one or more examples provided in the popup. The general idea is to reveal a bit less up front, and then more detail as you dial your way in. I don't have a specific design in mind, though.

services/pypi/pypi-downloads.service.js
```diff
@@ -77,31 +77,14 @@ module.exports = class PypiDownloads extends BaseJsonService {
     return [
       {
         title: 'PyPI - Downloads',
-        pattern: 'dd/:packageName',
+        pattern: ':period(dd|dw|dm)/:packageName',
```
chris48s: If you're going to just expose the guts of the route here, is there any advantage to aliasing it in the example rather than just using the route pattern directly?

services/david/david.service.js
chris48s: Is that question mark supposed to be a `/`?

paulmelnikow: Ah, the `?` is supposed to be there but it's supposed to have a `/` after it.

services/pypi/pypi-downloads.service.js
paulmelnikow: Updated!

calebcartwright:
> it may be that we move toward may be a listing where there's an entry per service, with one or more examples provided in the popup  

Does that mean one listing on the page per service (service being npm, GitHub, CircleCI, PyPI, etc. in this context)? If so I could see that getting tricky for the services we offer multiple badge types for (GitHub being the most obvious example, but also others where we have Build Status, Code Coverage, etc.)

> There had been one external constraint: a single service has to reside within one category

That constraint still exists currently correct? I can't yet add a service classes to multiple categories can I?

paulmelnikow:
>> it may be that we move toward may be a listing where there's an entry per service, with one or more examples provided in the popup
>
> Does that mean one listing on the page per service (service being npm, GitHub, CircleCI, PyPI, etc. in this context)? If so I could see that getting tricky for the services we offer multiple badge types for (GitHub being the most obvious example, but also others where we have Build Status, Code Coverage, etc.)

No, I mean more like for a service like npm version, instead of putting five entries in the listing, we'd put one entry in the listing, and show the five examples (somehow) within the popup.

screen shot 2019-01-30 at 6 18 56 pm

paulmelnikow:
That makes sense. Maybe we could add a little icon or count (a badge, if you will) that invites discoverability.

calebcartwright:
So where do we stand on this PR? Is it ready for a 👍 or more discussion needed?

chris48s:
Cheers for bumping this so it doesn't get completely lost.

I'm just having another look over this with fresh eyes and I do have one other thought.. 

There are some cases where this is really clear and intuitive, for example:

screenshot at 2019-02-14 21-19-56

Simultaneously, the way that we currently do the examples (explicitly showing an example for each variation) does allow us to "explain" short params a bit. For example:

screenshot at 2019-02-14 21-15-51

if we were to 'collapse' all these examples (so this one becomes `(v|vn)` for example) with a drop-down in the modal, it becomes more difficult to explain or document the non-obvious difference between `v` and `vn`.

Do you have any thoughts on this? Is it your thought that if we merge this, every route with a `(this|that)` becomes a drop-down, or should there be some cases where we just leave continue to expose multiple examples like this?

paulmelnikow:
I think it’d be up to the maintainers to evaluate these case by case.

IMO those two Conda examples should probably be kept separate because the title is explaining the difference.

I also think adding translations of what the parameter values mean could be useful down the line, eg “monthly”, “weekly”, and maybe even things like “channel only.”

calebcartwright:
> I think it’d be up to the maintainers to evaluate these case by case.

So we'll have the ability to control this on a case by case (I assume example by example) basis?

paulmelnikow:
Yea, any example pattern that uses multiple choice will get one of these. If the route pattern uses multiple choice but the example does not, it’ll show up with the literal value, as before.

calebcartwright:
Excellent. Sorry I hadn't picked up on that distinction yet. I think the fact that we'll have the ability to opt in to this on a per-param, per-example basis IMO gives us the flexibility we'll need to handle the various scenarios that have been discussed here. 

I'm 👍 on this feature

calebcartwright:
Also I think that once we get a feel for the patterns where leveraging this makes sense, and where it doesn't, we should update our service badge docs accordingly (definitely as a later, follow-up item)

Issue #2845, opened 2019-01-22
Services -> integrations?


paulmelnikow:
Here's a quiz. Is a "service":

- A badge endpoint?
- A subclass of `BaseService?
- A service in the colloquial sense, e.g. PyPI, npm, GitHub?
- All of the above?

The maintainers have discussed offline that this nomenclature is ambiguous.

The word "badge" has a similar problem: is the badge the result, or the code that generates it? I think my favorite of the proposals was @PyvesB' suggestion to adopt the word "integration".

Here are a few related proposals:

1. Rename Base*Service to Base*Integration.
4. Rename all the `.service.js` files to `.integration.js`.
3. In code and documentation, refer to service tests as integration tests. Refer to the existing integration test suite (GitHub and Redis) as _core integration tests_.
4. Keep the name `services/` for the code tree? Rename the **service-badge** issue label to **services**.
5. Or rename `services/` to `integrations/` and rename **service-badge** to **integrations**.
6. In documentation:
    1. Refer to each BaseIntegration (neé BaseService) as an "integration"
    2. Adopt the colloquial meaning of "service": GitHub is a service; Travis is a service; npm is a service. We've sometimes called these "service families" in the past. Now they will be called "services."

Thoughts?

chris48s:
:+1: I think last time we discussed this, we agreed the terminology we should use is:

* NPM is a **service**
* `class NpmVersion` is an **integration**
* ![](https://img.shields.io/npm/v/npm.svg) is a **badge**

I think what you've suggested there broadly fits with that. The only thing that is slightly unfortunate about it is that calling the tests for our integrations "integration tests" is slightly confusing because "integration testing" is a thing. :thinking:

calebcartwright:
I don't have any insight into the historical discussions/perspectives on this topic, but I personally didn't find the notion of Service classes (that provide the functionality of integrating with a 3rd party service) confusing nor ambiguous. I've found the usage of Service classes that encapsulate the functionality of integration pretty common.

paulmelnikow:
> The only thing that is slightly unfortunate about it is that calling the tests for our integrations "integration tests" is slightly confusing because "integration testing" is a thing. 🤔

On the plus side, given our testing strategy, the purpose of the tests that use ServiceTester is substantially the same as integration tests: to exercise our code as it interacts with the live upstream service. So that may run in our favor as we're trying to explain these tests to people (and will help them understand why they are often flaky).

[...]

PR #2886, opened 2019-01-28
 Move suggest code and rewrite tests

paulmelnikow:
The suggest code was an exception to our usual organization pattern. There was a service test, but it's not a service. The code would sometimes regress because it wasn't being tested all the time.

This makes them no longer run as service tests, which is good because they run as part of every build. Some of them are smaller-bracket tests which is good too, because it will make them easier to test, especially as this code grows.

I'd have liked to keep using frisby for the ones that make requests to the server, though I ran into some issues with sequencing of setup that I think will require upstream changes.


services/suggest.integration.js
```diff
@@ -0,0 +1,145 @@
+'use strict'
```
calebcartwright: I'm not sure where we landed with #2845, and I know we use this file naming pattern for other integration tests (under `core` for example) but if we end up renaming `services` (files, classes, and/or otherwise) to something `integration` related as part of #2845, I could see things under `services` getting confusing. 

This is probably good as is, and we can rename/refactor as needed as part of the #2845 work, but just wanted to jot this down

PR #2887, opened 2019-01-28
Convert some preview URLs [packagist php-eye puppetforge scrutinizer sonarqube website]

paulmelnikow:


core/badge-urls/make-badge-url.js
```diff
@@ -62,9 +62,6 @@ function staticBadgeUrl({
   namedLogo,
   format = 'svg',
 }) {
-  if (!label || !message) {
```
calebcartwright: Apologies as I'm _pretty_ sure I asked this question before recently (having serious deja vu if not), but for my own education why do we no longer need to require one of these (I'm assuming they were needed before)?

paulmelnikow: 😁It was here: https://github.com/badges/shields/pull/2869#discussion_r251168779

I had to change it again here because this badge has _no message_:

screen shot 2019-01-29 at 4 45 16 pm

calebcartwright: Thanks for helping me on this (twice 😉). Committed to memory this time!

services/continuousphp/continuousphp.service.js
```diff
@@ -17,15 +17,26 @@ module.exports = class ContinuousPhp extends LegacyService {
...
   }
 
   static get examples() {
     return [
       {
         title: 'continuousphp',
-        previewUrl: 'git-hub/doctrine/dbal/master',
+        pattern: ':provider/:user/:repo/:branch',
+        namedParams: {
+          provider: 'git-hub',
```
calebcartwright: Sidenote, I've never seen GitHub split like this before

services/website/website.service.js
```diff
@@ -95,9 +95,17 @@ module.exports = class Website extends LegacyService {
   static get examples() {
     return [
       {
-        previewUrl: 'website-up-down-green-red/https/shields.io',
-        queryParams: { label: 'my-website' },
-        keywords: ['website'],
+        title: 'Website',
+        pattern: 'website-up-down-green-red/:protocol/:hostAndPath',
```
calebcartwright: Should we use `:protocol(https|http)` here too for consistency? I noticed the above Twitter service had it that way

calebcartwright:
Code changes look good. Have to step away for a minute, but will take a quick look at these in the staging app and then approve when I get back

services/scrutinizer/scrutinizer.service.js
```diff
@@ -14,16 +14,25 @@ class ScrutinizerBuild extends LegacyService {
...
   }
 
   static get examples() {
     return [
       {
-        title: 'Scrutinizer Build',
-        previewUrl: 'build/g/filp/whoops',
+        title: 'Scrutinizer build',
+        namedParams: {
+          vcsType: 'g',
```
calebcartwright: No changes needed, but just an observation that since the first parameter has the subtext prefix `e.g. ` in the modal window, this one kinda looks like `egg` at first glance!

services/website/website.service.js
calebcartwright: Also the `up-down-green-red` part is optional on this badge right? I'm not sure what the pattern would need to look like to make that part configurable in the modal (maybe `website-:(options)/:protocol/...` with `up-down-green-red` as the value for the named `options` param?), but is there value in allowing the user to enter their desired options in the model to build the badge urls?

paulmelnikow: Yea… I tried messing with the pattern for a little while and it seems like supporting this format is going to require further changes to the frontend. I'd prefer to limit the kinds of patterns we use, and move options like that to the query string.

e.g. `?up_message=&down_message=&up_color=&down_color=`

Besides reducing the amount of custom code needed to build the URLs, it has an added benefit of keeping our badge APIs relatively uniform.

services/scrutinizer/scrutinizer.service.js
paulmelnikow: Hehe, fortunately not something that comes up too much.

services/website/website.service.js
paulmelnikow: Updated!

services/continuousphp/continuousphp.service.js
paulmelnikow: Yea, it's really weird.

services/website/website.service.js
calebcartwright: I like the sound of that

paulmelnikow:
All failures are intermittent and unrelated except one in sonar, which seems connected to this changeset.

calebcartwright:
Yeah I feel like quite a few of our service tests are starting to flake out!

paulmelnikow:
I wonder if there's something systemic going on.

calebcartwright:
> I wonder if there's something systemic going on, like in a cloud provider.

Yeah that'd make sense. These days if the likes of AWS, Microsoft, or Google have any platform issues then they take like 80% of the internet down with them!

PR #2890, opened 2019-01-28
Add a diagnostic page for testing logos

paulmelnikow:
It can be helpful to have some diagnostic pages for development and quality control. I added one here for the logos, which renders all the named logos in ?style=flat and ?style=social.


chris48s:
On staging this hit `Exceeded limit 500 requests per 200 seconds` halfway through loading the page :D

Are you thinking this is something to expose to end users, or something for local development use?

paulmelnikow:
No, not end users, but it's useful in development. It would be nice if we didn't run into rate limits like this on staging!

calebcartwright:
Another observation (not sure if this is expected or not), but some of the logos on the social badges at the top of the page are missing for me:

image

paulmelnikow:
The page is doing its job!

> Another observation (not sure if this is expected or not), but some of the logos on the social badges at the top of the page are missing for me:

Confirmed. I guess there is some pending work to recolor those logos so they are visible on the social badge.

chris48s:
Yes - there are definitely some light logo on light background issues with our bespoke logo set

It would be quite nice to make this view (and potentially other 'debug' views we might want to declare) available in dev, but not in production. I've slightly lost track of this in the settings re-organisation, but at the point we build the front-end, do we have an env var available that indicates prod/dev? i.e: can we declare the route conditionally on `proceess.env.ENVIRONMENT == 'dev'` or something so it exists in local dev, but not when we do the optimised production build?

paulmelnikow:
> i.e: can we declare the route conditionally on proceess.env.ENVIRONMENT == 'dev' or something so it exists in local dev, but not when we do the optimised production build?

That sounds good. Let's get #2906 merged and then revisit this.

paulmelnikow:
We still may need to solve the rate limiting, though even with it, I think this page remains more useful than not.

PR #2895, opened 2019-01-29
Consolidate service definition schema, removing support for previewUrl

paulmelnikow:
The diff should clean up after #2887 is merged.

This removes `LONG_CACHE` and its descendants, which was a feature that added `?maxAge` to the live preview badges in the frontend. Since they are all static that is no longer needed, as the static badges all have longer cache timeouts regardless.


core/base-service/transform-example.js
```diff
@@ -46,66 +43,45 @@ function validateExample(example, index, ServiceClass) {
...
-          ServiceClass.name
-        } at index ${index} does not declare a pattern`
-      )
-    } else if (!namedParams) {
-      throw new Error(
-        `Static preview for ${
-          ServiceClass.name
-        } at index ${index} does not declare namedParams`
-      )
-    }
```
paulmelnikow: `pattern` and `namedParams` are now required by the schema.

core/base-service/transform-example.js
```diff
@@ -46,66 +43,45 @@ function validateExample(example, index, ServiceClass) {
...
-    if (previewUrl) {
-      throw new Error(
-        `Static preview for ${
-          ServiceClass.name
-        } at index ${index} also declares a dynamic previewUrl, which is not allowed`
-      )
-    }
```
paulmelnikow: Now rejected by the schema.

core/base-service/transform-example.js
```diff
@@ -46,66 +43,45 @@ function validateExample(example, index, ServiceClass) {
...
+  keys = keys.map(({ name }) => name)
+  const extraKeys = Object.keys(namedParams).filter(k => !keys.includes(k))
+  if (extraKeys.length) {
+    throw Error(
+      `In example for ${
         ServiceClass.name
-      } at index ${index} is missing required previewUrl or staticPreview`
+      } at index ${index}, namedParams contains unknown keys: ${extraKeys.join(
+        ', '
+      )}`
```
paulmelnikow: This diff looks strange. It's a removal of the `else` block accompanied by a change in indentation level.

core/base-service/transform-example.js
```diff
@@ -121,59 +97,39 @@ function transformExample(inExample, index, ServiceClass) {
     queryParams,
     pattern,
     staticPreview,
-    previewUrl,
     keywords,
     documentation,
   } = validateExample(inExample, index, ServiceClass)
 
-  let example
```
paulmelnikow: This removes two now-unnecessary `if` statements.

README.md
```diff
@@ -19,9 +19,6 @@
     <a href="https://docs.google.com/spreadsheets/d/1cHIUSVaiKrIFw3KIu0yt-EMNlMkIfU5alE7YKZ4PeOE/edit#gid=0">
         <img src="https://img.shields.io/github/search/badges/shields/extends%20LegacyService.svg?label=legacy%20services%20needing%20refactor"
             alt="legacy services needing refactor"></a>
-    <a href="https://github.com/search?utf8=%E2%9C%93&q=previewUrl+repo%3Abadges%2Fshields+path%3A%2Fservices&type=Code&ref=advsearch&l=&l=">
-        <img src="https://img.shields.io/github/search/badges/shields/previewUrl.svg?label=`previewUrl`s%20needing%20refactor"
-            alt="`previewUrl`s needing refactor"></a>
```
calebcartwright: ✨ 💥 👏

core/base-service/transform-example.js
```diff
@@ -10,29 +10,26 @@ const optionalObjectOfKeyValues = Joi.object().pattern(
...
-  color: Joi.string(),
-  style: Joi.string(),
-})
-
 const schema = Joi.object({
   // This should be:
   // title: Joi.string().required(),
   title: Joi.string(),
-  namedParams: optionalObjectOfKeyValues,
+  namedParams: optionalObjectOfKeyValues.required(),
```
calebcartwright: I'm a little confused here, I think just because the verbiage starts with `optional` and ends with `required`. Is it just that `namedParams` is required on all examples now that `previewUrl` is removed?

core/base-service/transform-example.js
```diff
@@ -46,66 +43,45 @@ function validateExample(example, index, ServiceClass) {
     `Example for ${ServiceClass.name} at index ${index}`
```
calebcartwright: This validation will fail if there's any extra keys, such as `previewUrl`, which is what now ensures there's no previewUrl right?

frontend/components/badge-examples.js
```diff
@@ -30,47 +30,30 @@ export default class BadgeExamples extends React.Component {
...
-        color,
-        style,
-        namedLogo,
-      })
-    } else {
-      const { path, queryParams } = preview
-      previewUrl = badgeUrlFromPath({ baseUrl, path, queryParams, longCache })
-    }
+    const { label, message, color, style, namedLogo } = preview
+    const previewUrl = staticBadgeUrl({
```
calebcartwright: QQ for myself, where/how is this used with static examples and the modal window? (or perhaps I should ask the question on the `exampleUrl` below as I'm not sure what the difference is anymore

core/base-service/transform-example.js
paulmelnikow: The `required()` modifies `optionalObjectOfKeyValues` and makes it required.

The translation is that `namedParams` is required, whereas `queryParams` can be omitted and in that case is interpreted as `{}`.

calebcartwright: I missed this schema being used below on `queryParams` (which is why we wouldn't just make it required in the schema itself). 

Gotcha 👍

core/base-service/transform-example.js
paulmelnikow: Correct, exactly.

frontend/components/badge-examples.js
paulmelnikow: This function renders a row in the badge listings. The `handleClick` function defined in here opens the modal. The row contains three columns.

The title:

screen shot 2019-01-30 at 5 59 50 pm

`<ClickableTh onClick={handleClick}>{title}:</ClickableTh>`

The preview badge:

screen shot 2019-01-30 at 5 59 52 pm

```js
<td>
  <Badge clickable onClick={handleClick} src={previewUrl} />
</td>
```

The example pattern:

screen shot 2019-01-30 at 5 59 54 pm

```js
<td>
  <ClickableCode onClick={handleClick}>{exampleUrl}</ClickableCode>
</td>
```

Does that help?

paulmelnikow:
I just remembered another big pile of now dead code in `frontend/components/markup-modal/index.js`. Do you want me to push it here or save it for another PR?

frontend/components/badge-examples.js
calebcartwright: Yeah, a lot actually thanks! I swear one of these days I'm going to find and read a React for dummies 😄 That `ClickableCode` element sounds pretty sweet

calebcartwright:
I'm ready to approve this as is, but happy to wait and take a look at the updates to `index` too. If that cleanup is related to the previewUrl removal/service definitione/etc. then might as well stick it in here IMO

frontend/components/markup-modal/index.js
```diff
@@ -2,28 +2,13 @@ import React from 'react'
...
-import { Snippet } from '../snippet'
-import { BaseFont, H3, Badge, BlockInput } from '../common'
+import { BaseFont } from '../common'
 import MarkupModalContent from './markup-modal-content'
 
 const ContentContainer = styled(BaseFont)`
   text-align: center;
 `
 
-const WeeSnippet = ({ snippet, truncate = false }) => (
```
calebcartwright: I have no idea what I'm looking at here 😄

frontend/components/markup-modal/index.js
```diff
@@ -2,28 +2,13 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import Modal from 'react-modal'
 import styled from 'styled-components'
-import {
```
paulmelnikow: Probably you remember that this code was substantially duplicated into https://github.com/badges/shields/pull/2554/files#diff-559d49006536e9e410b609d2668583ec

I didn't bother to DRY it up because I hoped it wouldn't be around for long: https://github.com/badges/shields/pull/2554/files#r245429994

frontend/components/markup-modal/index.js
paulmelnikow: In a nutshell: React is built around a single abstraction called components. Virtually everything in React is a component. The component's `render()` method is the most important function inside it. It gets called every time something changes in the component, and transforms the immutable attributes passed in from the parent – called `props` – and the state that lives in the component – appropriately enough called `state` – into a tree of child components. The leaves of the component tree correspond to DOM elements. After rendering finishes, React intelligently reconciles the component tree with the DOM.

What you're looking at _right_ here is a [function component](https://reactjs.org/docs/components-and-props.html#function-and-class-components); it's basically shorthand where instead of making a class, you just write the render function (it gets a single parameter, the `props`).

In `<Snippet truncate={truncate} fontSize="10pt" snippet={snippet} />`, a `Snippet` component is being created, with three props: `truncate`, `fontSize`, and `snippet`.

These are probably useful starting points for reading React code:

1. https://reactjs.org/docs/rendering-elements.html
2. https://reactjs.org/docs/components-and-props.html
3. https://reactjs.org/docs/state-and-lifecycle.html

Issue #2891, opened 2019-01-29
endpoint.svg requires https for localhost


niccokunzmann:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

When I want to test my service to provide and [endpoint](https://shields.io/#/endpoint), 
I see a badge requiring https.
I can not offer HTTPS on localhost but I would like to test the setup with a running shields version.

:link: **Link to the badge**

:bulb: **Possible Solution**

The code could test for localhost before requiring https:
[code](https://github.com/badges/shields/blob/4e9763b4c3d36593cacca6176d509b7b880dc06d/services/endpoint/endpoint.service.js#L112)

calebcartwright:
Hi @niccokunzmann, thanks for reaching out and trying the new endpoint service!  

The endpoint service is still in beta, and we're tracking discussions/feedback/etc. on it over in #2838. Would you mind sharing your suggestion over there?

paulmelnikow:
I think adding a check for localhost would be fine. #2838 is a good place to discuss this kind of stuff, but we may as well finish this discussion while we're here!

niccokunzmann:
I wonder if it makes sense at all to require HTTPS. e.g. if I just want to share a badge which is a [free Heroku Dyno](https://www.heroku.com/pricing#dyno-comparison), I do not have HTTPS. While it is preferable, it limits experimentation. Why require this?

niccokunzmann:
Very well. Now, I understand what is written there: "Free SSL on custom domains." ^^ SSL on their domains is then always available. Okay. Point taken! So, this is only http on localhost as a problem, then, as the issue states.

calebcartwright:
I also see a potential benefit for supporting http in certain self-hosted scenarios too (especially as we've been eyeballing this internally). 

Totally agreed on SSL as best practice, but agree that requiring SSL could hinder the inner-dev loop, and I can also hear some coworkers complaining about having to implement SSL for a private endpoint on our corporate network only access by our internal self-hosted shields instance. 

I remember the SSL conversation we recently had relative to Bitbucket Server support and it got me wondering if we could potentially have an SSL toggle that would require SSL in all cases. That way we could require SSL on Shields.io on all endpoints/services, but self-hosters would have the ability to turn that off if they _really_ needed to connect to an endpoint/service over http

[...]

PR #2903, opened 2019-02-02
allow localhost as domain for http

niccokunzmann:
- fixes https://github.com/badges/shields/issues/2891
This is a proposal.
This may allow portscans on localhost for servers connected to the Internet. I do not think this is wanted.

Questions:
- https://github.com/niccokunzmann/shields/pull/new/endpoint-allow-http-for-localhost proposes to make this configurable.
  I still would need to look into how a test can be run with a specific configuration.
- Should it be forbidden in general to connect to localhost?

### DIFF
--- a/services/endpoint/endpoint.service.js
+++ b/services/endpoint/endpoint.service.js
@@ -109,7 +109,7 @@ module.exports = class Endpoint extends BaseJsonService {
     )
 
     const { protocol, hostname } = new URL(url)
-    if (protocol !== 'https:') {
+    if (protocol !== 'https:' && hostname !== 'localhost') {
       throw new InvalidParameter({ prettyMessage: 'please use https' })
     }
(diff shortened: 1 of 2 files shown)


paulmelnikow:
Hi! Thanks for picking this up.

Since this feature is targeting development and self-hosting, instead of adding code to specifically allow certain hosts, it would be better to add a config option, something like `allowUnsecuredRequests: true` which turns off the safety check: https://github.com/badges/shields/issues/2891#issuecomment-458739686

As far as testing, what I'd suggest is to set `allowUnsecuredRequests: false` in the test config, and include service tests that establish that `https` works and `http` does not. Then break out the conditional requesting  / error throwing into a function that can have its own `.spec.js` tests.

paulmelnikow:
@niccokunzmann As discussed above this needs a bit of reworking. Feel free to open a new PR when you're ready!

niccokunzmann:
Sure!

Issue #2876, opened 2019-01-26
Docker issue: production code imports test dependencies


calebcartwright:
I just did a deploy of the internal Shields instance we host at my day job and ran into a couple dev vs. prod dependency issues. I had to move the below dependencies from dev to prod deps in order to get the container to run:

* nock 
* icedfrisby
* icedfrisby-nock
* caller

Everything worked fine afterwards. I haven't had a chance to dig into the underlying cause yet, but I wouldn't have expected those to be runtime/prod dependencies

paulmelnikow:
Ah… `caller` is a good clue, since we only import it in one place, which is `create-service-tester`. After #2809 we import `create-service-tester` from `services/index.js`, and that's imported by all the services. Methinks we're going to need to revisit that.

Calls for building some Docker validation sooner rather than later… would have been nice to have caught that earlier in the process (before doing so much work).

calebcartwright:
That makes sense. Should we close this and track it under #2816 then?

paulmelnikow:
It’s separate from fixing the immediate issue. We could track them separately.

calebcartwright:
Fair enough. I'm personally working around this by running an `npm uninstall --save-dev` and than `npm i --save` on those 4 dependencies in my build process prior to building the Docker image. 

I'm assuming we want reconsider whether `create-service-tester` gets imported which sounds like a bigger fix vs. doing a short term fix by adding those 4 deps as prod dependencies? 

If anyone else is having problems running the image we could point them here to run those two npm commands prior to the `docker build ...` command.

paulmelnikow:
Another workaround is to use the commit before that one, temporarily.

I don’t think we should make these prod dependencies. I don’t think it needs to take a long time to fix the problem.

paulmelnikow:
Thoughts on how to fix this?

I’m reluctant to add more magic here, like conditional imports or try/catch imports. How about moving the test shortcut imports from `services/index.js` to `services/tester.js`, and updating all the tester imports to reference that instead?

calebcartwright:
I like that idea. I'll take a shot at it later this week

calebcartwright:
Just adding a note that `ServiceTester` which is also imported in `services/index.js` also imports some dev deps (`icedfrisby` and `icedfrisby-nock`) so I'm going to move both `ServiceTester` and `create-service-tester` from `services/index.js` to `services/tester.js`

PR #2906, opened 2019-02-02
Adopt Gatsby

### LINKED ISSUE #1943: Adopt CRA or Gatsby
We should also adopt Gatsby, create-react-app, or something similar designed for static sites, to eliminate the unnecessary runtime dependency on Next.js while also letting someone else maintain our front-end tooling.

This has been said many times before, though there was no issue for it.

Hopefully this will take care of #1941.

### LINKED ISSUE #2616: Some frontend files still missing from coverage
I noticed that https://coveralls.io/github/badges/shields is still missing a couple of the frontend files:

screen shot 2019-01-01 at 10 52 46 pm

screen shot 2019-01-01 at 10 51 34 pm

### LINKED ISSUE #2837: Fix formatting issues with the endpoint badge doc page
The page for the new (beta) Endpoint badge has some serious issues. This seems to be caused by an interop issue between styled-components and Next.js SSR. I've spent a fair amount of time trying to fix it and don't think I should spend any more given #1943.

To reproduce the problem, go to the home page, then click the link for the Endpoint badge.

After refresh, the page looks a little better, but it still not applying all the styles that are being defined. It's just a different subset of them that happens to look better.

I think the way to fix this in Next.js is to collect the missing stylesheets in https://github.com/badges/shields/blob/master/pages/_document.js and I'm not opposed to doing that if someone knows how.

However, I'd lean toward resolving this in tandem with tackling #1943.

paulmelnikow:
While Next.js can handle static sites, we've had a few issues with it, notably a performance hit at runtime and some bugginess around routing and SSR. Gatsby being fully intended for high-performance static sites makes it a great technical fit for the Shields frontend. The `createPages()` API should be a really nice way to add a page for each service family, for example.

This migrates the frontend from Next.js to Gatsby. Gatsby is a powerful tool, which has a bit of downside as there's a lot to dig through. Overall I found configuration easier than Next.js. There are a lot of plugins and for the most part they worked out of the box. The documentation is good.
 Links are cleaner now: there is no #. This will break old links though perhaps we could add some redirection to help with that. The only one I’m really concerned about `/#/endpoint`. I’m not sure if folks are deep-linking to the category pages.

There are a lot of enhancements we could add, in order to speed up the site even more. In particular we could think about inlining the SVGs rather than making separate requests for each one.

While Gatsby recommends GraphQL, it's not required. To keep things simple and reduce the learning curve, I did not use it here.

Close #1943 
Fix #2837 Fix #2616

----

To do before merging:

- [x] Get frontend tests working again
- [x] Fix coverage
- [x] Fix danger
- [x] See if there are other dependencies which can be removed
- [x] Test/fix Heroku deploy
- [ ] ~~Test/fix Now deploy~~ (not working, but let's not block)
- [x] Test/fix self hosting – Not working because of #2876
- [x] Fix or remove former use of NavLink
- [x] Redirect old # links on index page


paulmelnikow:
@mbrandau Are you up for giving this a try?

paulmelnikow:
When changing categories, seems to respond faster than the current site:

http://shields-staging-pr-2906.herokuapp.com/

vs.

https://shields-staging.herokuapp.com/

calebcartwright:
> When changing categories, seems to respond faster than the current site:
> 
> http://shields-staging-pr-2906.herokuapp.com/
> 
> vs.
> 
> https://shields-staging.herokuapp.com/

Yeah I think I'm observing the same behavior, although in all honesty both are pretty snappy 😄 

Two observations:

1) The Shields logo in the browser tab is black and white in this PR staging app
image

2) In the current site, after quickly hopping around between a few categories, several of the categories started to take a noticeably long time to render the badges, but I've been able to reproduce that in this PR staging app 👍

calebcartwright:
> Could you clarify this?

Sure! I was switching between categories fairly quickly in both apps in an attempt to do an eyeball performance comparison. After some time of doing that in the current site (https://shields-staging.herokuapp.com/) the listings on some of the category pages appeared to not be rendering (they eventually did render, but took 30+ seconds)

paulmelnikow:
This PR has the same problem, or no?

calebcartwright:
This PR does **not** have the same problem. Could've just been a fluke, but was only happening on https://shields-staging.herokuapp.com/

package.json
```diff
@@ -186,6 +193,8 @@
     "prop-types": "^15.6.0",
     "react": "^16.7.0",
     "react-dom": "^16.7.0",
+    "react-error-overlay": "^5.1.2",
+    "react-helmet": "^5.2.0",
     "react-modal": "^3.8.1",
     "react-pose": "^4.0.6",
     "react-router-dom": "^4.3.1",
```
chris48s: It seems like this PR removes all the references to `react-router-dom` - is this dependency still needed?

paulmelnikow: Nope! I just removed it.

paulmelnikow:
> Can you give a bit more info on how will the routing work once we're deployed on GH pages? It seems like in order to do away with the client-side # style links, we need to add an extra build step to the deploy process to build static pages for each URL or something?

`gatsby build` writes out all those files:

```
public
├── 0-67fa5a5264bb654b8669.js
├── 0-67fa5a5264bb654b8669.js.map
├── 0-ad913e07e45d935c6172.js
├── 0-ad913e07e45d935c6172.js.map
├── 2-660cc00aea5bf7e39cfd.js
├── 2-660cc00aea5bf7e39cfd.js.map
├── 2-cb227f2fc74ce66d55d8.js
├── 2-cb227f2fc74ce66d55d8.js.map
├── app-14d7ce459d4ac392693f.js
├── app-14d7ce459d4ac392693f.js.map
├── app-9e1789bcf880397fdde9.js
├── app-9e1789bcf880397fdde9.js.map
├── category
│   ├── activity
│   │   └── index.html
│   ├── analysis
│   │   └── index.html
│   ├── build
│   │   └── index.html
│   ├── chat
│   │   └── index.html
│   ├── coverage
│   │   └── index.html
│   ├── dependencies
│   │   └── index.html
│   ├── downloads
│   │   └── index.html
│   ├── funding
│   │   └── index.html
│   ├── issue-tracking
│   │   └── index.html
│   ├── license
│   │   └── index.html
│   ├── monitoring
│   │   └── index.html
│   ├── other
│   │   └── index.html
│   ├── platform-support
│   │   └── index.html
│   ├── rating
│   │   └── index.html
│   ├── size
│   │   └── index.html
│   ├── social
│   │   └── index.html
│   └── version
│       └── index.html
├── chunk-map.json
├── component---frontend-components-main-js-438efc3da2edda687363.js
├── component---frontend-components-main-js-438efc3da2edda687363.js.map
├── component---frontend-components-main-js-ea8ec189fb5a125c3404.js
├── component---frontend-components-main-js-ea8ec189fb5a125c3404.js.map
├── component---frontend-pages-endpoint-js-21b461837339c365660b.js
├── component---frontend-pages-endpoint-js-21b461837339c365660b.js.map
├── component---frontend-pages-endpoint-js-7191d3f0fc948a41bbf1.js
├── component---frontend-pages-endpoint-js-7191d3f0fc948a41bbf1.js.map
├── component---frontend-pages-index-js-c3db22348ada5bbbd9da.js
├── component---frontend-pages-index-js-c3db22348ada5bbbd9da.js.map
├── component---frontend-pages-index-js-d5d233367cb92aa4606e.js
├── component---frontend-pages-index-js-d5d233367cb92aa4606e.js.map
├── endpoint
│   └── index.html
├── index.html
├── pages-manifest-3aebc6c37146c164f6ea.js
├── pages-manifest-3aebc6c37146c164f6ea.js.map
├── pages-manifest-d8d2c9f92f905bee0160.js
├── pages-manifest-d8d2c9f92f905bee0160.js.map
├── pages-manifest-efd19252f95bb7a71410.js
├── pages-manifest-efd19252f95bb7a71410.js.map
├── static
│   └── d
│       ├── 117
│       │   └── path---category-rating-046-2e3-KdA7G0YPKQTjYGAyoqXB4cvtk.json
│       ├── 132
│       │   └── path---category-social-06-c-ae9-4VZRdurTidfJEQBJzS8VIf9PQk.json
│       ├── 173
│       │   └── path---index-6a9-NZuapzHg3X9TaN1iIixfv1W23E.json
│       ├── 253
│       │   └── path---category-version-cf-1-a18-gXHyDTzYNgar0wLLw0WZixpAj5c.json
│       ├── 284
│       │   └── path---category-downloads-ac-6-ad2-i1qVAXGTJHbgxNgVU8X7VTyJMg.json
│       ├── 383
│       │   └── path---category-size-7-c-5-f3c-IA7jm7CtmkAoZxX7H1UoYKKqU.json
│       ├── 418
│       │   └── path---category-funding-c-9-c-3d6-fXFX1kL0ODtgGpsFPmDeGpBViw.json
│       ├── 420
│       │   └── path---category-monitoring-1-e-4-713-bEZ1Ow7ciXXVUOK28ye6ImBRmeM.json
│       ├── 428
│       │   ├── path---category-platform-support-54-b-1cb-Z5tIFulGwp1r5jnYKKaRp0ozYg0.json
│       │   └── path---endpoint-a-3-a-f9f-NZuapzHg3X9TaN1iIixfv1W23E.json
│       ├── 485
│       │   └── path---category-issue-tracking-e-8-e-5e4-2rlEZhPS9BHCJs3cEHkP5Hn8WQA.json
│       ├── 600
│       │   └── path---category-dependencies-3-f-4-f6d-Ab6kijLqMLCJfo5PPjulSt60K8.json
│       ├── 678
│       │   └── path---category-other-c-96-41b-RndZH3c3mZfk6hjfU6H8nEoka3I.json
│       ├── 69
│       │   └── path---category-chat-9-aa-fda-Ao51OaMIHXU8mgnpMIOxT7x2kVc.json
│       ├── 792
│       │   └── path---category-activity-571-37c-tsoCfBAcVbrQK2D7rSGCYN11YQ.json
│       ├── 88
│       │   └── path---category-analysis-36-e-d1e-JbegFHdHBYgMoxjX2zbzP05xw.json
│       ├── 955
│       │   └── path---category-license-77-a-1b3-2Fwgu2pA9WVHICxqtSBKuQVYwP8.json
│       ├── 972
│       │   └── path---category-coverage-ffc-771-getZDHJLqCIMepXXS3Ctk7Uqbso.json
│       └── 976
│           └── path---category-build-94-b-71c-AvOamVOIE8vKzzf7fxl1YlaO4.json
├── webpack-runtime-45ab0c5fe180a44c9a6e.js
├── webpack-runtime-45ab0c5fe180a44c9a6e.js.map
├── webpack-runtime-76e0ac706c7f4f18b05a.js
├── webpack-runtime-76e0ac706c7f4f18b05a.js.map
├── webpack-runtime-b3639ec9365f37fed072.js
├── webpack-runtime-b3639ec9365f37fed072.js.map
└── webpack.stats.json
```

Whichever link you load first, that's the `index.html` that gets loaded. Then, when you switch routes, the route change gets intercepted on the frontend, some magic happens to load whatever new needs to be loaded, and the page re-renders, without actually loading the other `index.html` page. It's pretty cool.

paulmelnikow:
I can't make heads or tails of what's going on with the Now deploy. I've experimented with a bunch of different values in `.nowignore`.

I opened a discussion thread: https://spectrum.chat/zeit/general/troubleshooting-source-file-selection-in-v1~8dc07b08-99ac-4a02-bfc6-6d40d5e44792

paulmelnikow:
Feel like we could review and merge this. I’d like to get the Now build working again quickly, though I don’t think that needs to block. I sometimes get a little weary of Now. It can be a bit unstable, hard to troubleshoot, and is poorly documented.

I feel like the product is improving and changing a lot so I’m not writing it off, and I do use it a lot. Though I find products frustrating when I can’t solve my own problems 🤔

Makefile
```diff
@@ -48,7 +48,7 @@ push-s2:
 deploy-gh-pages:
 	rm -rf ${FRONTEND_TMP}
 	git worktree prune
-	BASE_URL=https://img.shields.io \
+	GATSBY_BASE_URL=https://img.shields.io \
 		NEXT_ASSET_PREFIX=https://shields.io \
```
chris48s: is `NEXT_ASSET_PREFIX` still relevant?

paulmelnikow:
> * There is a bit of next-specific code left in the dangerfile: `_document.js is used by convention by Next`

Fixed.

paulmelnikow:
> * I'm getting `TypeError: category is undefined` trying to use search or suggest

Thanks for finding that. Should be fixed.

calebcartwright:
I'm going to be out of my depth on most of the changes in this PR, but going to take a look anyway (if for no reason other than getting myself more familiarized with this part of the codebase) 

Semi-related question: Do we have docs anywhere that cover where/how we're using Next.js (and now Gatsby), Now, etc.?

.nowignore
```diff
@@ -0,0 +1,30 @@
+*
```
calebcartwright: This file looks a little intense at first glance. I'm sure it's necessary but it's not clear to me why without having any familiarity with Now. 

Are there any circumstances going forward where we'd need to make updates here (and if so what are those so I can try to commit them to memory 😉), or is this pretty much set for the forseeable future?

.nycrc-frontend.json
```diff
@@ -1,9 +1,10 @@
 {
   "reporter": ["lcov"],
-  "all": true,
+  "all": false,
```
calebcartwright: What's the thinking behind disabling this option? If I recall correctly this setting controls whether to count source files even if they aren't required by tests (which I figure would give us a more accurate coverage reading).

paulmelnikow: Right, so it doesn't work with `instrument: false`. See https://github.com/badges/shields/pull/2906/files#diff-a232581bf91df05dba97d6a2b92ebb3d which patches around the situation and fixes #2616.

The limitation is poorly documented and I started to track down the loose ends and contribute documentation upstream, but I don't understand it quite well enough to articulate it clearly.

.nowignore
paulmelnikow: Now ignores seem to be totally broken; I'm deferring it for now: https://github.com/badges/shields/pull/2906#issuecomment-460457391

frontend/components/header.js
```diff
@@ -10,7 +11,7 @@ const Highlights = styled.p`
 export default () => (
   <section>
     <Link to="/">
-      <img alt="Shields.io" src="/static/logo.svg" />
+      <Logo />
```
calebcartwright: Do we lose the alt text with this, or is there some Gatsby magic (or is it baked into the svg already)?

paulmelnikow:
> Semi-related question: Do we have docs anywhere that cover where/how we're using Next.js (and now Gatsby), Now, etc.?

No, not really. Two good follow-on tasks would be:

1. Write a few sentences, probably in the code walkthrough, about how we use Gatsby to build the frontend, and link out to a tutorial and to this PR which includes the basic wiring. We could also mention styled-components and the entirely separate transpilation pipelines for unit tests and linting.
2. Add a little more context at the top of the self-hosting doc that mentions the three deployment targets we support: Docker, Heroku, and Zeit Now. Probably that should include a short explanation of the build process for each one, and how to validate it.

.nowignore
calebcartwright: Oops! I didn't put two and two together there

.nycrc-frontend.json
calebcartwright: I thought I remembered an issue with frontend coverage. That articulation is good enough to cover the why 👍

frontend/components/main.js
```diff
@@ -48,11 +48,12 @@ export default class Main extends React.Component {
   }
 
   static propTypes = {
-    match: PropTypes.object.isRequired,
-  }
-
-  get category() {
-    return this.props.match.params.category
+    pageContext: {
```
calebcartwright: `pageContext` is a Gatsby thing or a React thing?

frontend/components/header.js
paulmelnikow: I'm not sure! You could check the staging app. This is getting handled by `babel-plugin-inline-react-svg`, which I imagine has a way to provide alt text.

frontend/components/main.js
paulmelnikow: Gatsby thing. Usually it gets piped through a GraphQL query, but GraphQL adds unnecessary complexity to what we're doing so I used this approach instead:

https://www.gatsbyjs.org/docs/using-gatsby-without-graphql/#the-approach-fetch-data-and-use-gatsbys-createpages-api

frontend/components/main.js
```diff
@@ -139,12 +140,6 @@ export default class Main extends React.Component {
           {this.renderCategory(category, definitions)}
         </div>
       )
-    } else if (categoryId) {
```
calebcartwright: What sort of circumstances (user action) would have gotten us here with the old implementation? I think this happened if I added a random/non-existent category to the url in the browser. If I do that now (say http://shields-staging-pr-2906.herokuapp.com/category/foo) I'm getting the 404 page. Is that what we want to do now vs. the unknown category we used to show (https://shields-staging.herokuapp.com/#/examples/foo)

paulmelnikow: The old route would match anything that was typed in to this page. You can see the code that did that here: https://github.com/badges/shields/pull/2906/files#diff-e14ec8fd2b038fbccea5f8090d26ace4L11

That means it was possible for this component to be loaded with an invalid category.

Now the routes are established by the `createPage()` calls in `gatsby-node.js` so we don't reach this point if the category doesn't exist.

If we want a custom 404 page that's probably possible, though it would no longer be the responsibility of this component.

frontend/components/header.js
calebcartwright: I _think_ it is, but I don't think it matters all that much

frontend/components/main.js
calebcartwright: I think the new behavior is good (and I'm a fan of the existing 404 page anyway). If I manually type in a path in the url that doesn't exist, i deserve a 404 IMO 👍

frontend/components/main.js
paulmelnikow: I added a comment about this.

calebcartwright:
I've finished going through the files (my inline questions, etc. have already been addressed), and am just poking around in the PR staging app UI now. 

I just noticed one minor thing, probably off topic since this occurs in prod too, but when configuring a static badge after I select a color, part of the text of that color gets cut off if I pick a longer color (like bright green): 

image

paulmelnikow:
Yea, that doesn't look great. That whole component could be improved. Maybe it could build off some of the work in the new builders.

paulmelnikow:
Great. Yea, that happened when I fixed the favicon. Will look into it…

paulmelnikow:
Looks like that fix worked.

paulmelnikow:
It's live!

Issue #2904, opened 2019-02-02
Allow embedding endpoint badges into the frontend


niccokunzmann:
## Motivation

The endpoint badge #2838 allows implementing the api-to-badge conversion logic in different locations without need to modify or add shields.io services. This leads to a possible diversification and emergence of new badges, which should be faster than badges added directly to shields.io source.

## Context

In order to provide an endpoint badge, one "only" needs to create an API as specified in https://shields.io/#/endpoint.
However, to make this badge usable by other people, one as developer would like to provide the same template functionality as shields.io does, such as `https://img.shields.io/appveyor/ci/:user/:repo.svg`
Example: http://transifex.quelltext.eu/

## Problem

Each new endpoint wants to provide configurable badges.
Currently, the endpoint badge only allows using your endpoint.
Configuration needs to be done manually by users or as crappy as on https://transifex.quelltext.eu

## Possible Solution

A possible solution to this would be

- [ ] A way of including endpoint badges into the shields.io frontend.
- [ ] A documentation of how to achieve this
  - [ ] located in https://shields.io/#/endpoint or /docs?

Related:
- #2838

paulmelnikow:
Hi! That's a nice suggestion. I think we'd handle this like any other badge proposal. Since the upstream API uses the Shields.io format, the implementation is easier and could reuse existing code.

1. Would you like us to add one for Transifex? There's been interest: #497. I'm happy to do a PR if you'd like.
2. We could add on the endpoint page some documentation describing the process of getting a badge included, and maybe point to an example PR.
3. It could also be cool to have an embeddable front-end component that allows a service provider to provide a badge popup-like interface in their website, which allows users to set customization options and get a badge URL.

niccokunzmann:
> (1) Would you like us to add one for Transifex? There's been interest: #497. I'm happy to do a PR if you'd like.

We can add one for Transifex, yes. If you like, you can do it. I can do it, too. I think, it would be located in the endpoint directory, shouldn't it? Or not? Should it be a different service directory? Probably yes, because the sorting of the services is based on the topic and not on the implementation - although the API is probably the same if the topic ("Docker", "Transifex") is the same.

(2) Yes!  
(3) I think this would be a new issue and not the scope of this one.

paulmelnikow:
There's a bunch of helper code that needs to be moved / extracted, so I'm happy to take this one.

[...]

Issue #497, opened 2015-07-22
Add Transifex service


nelson6e65:
Should be nice have badges from Transifex (https://www.transifex.com), like translation progress, translation progress by language, number of resources/languages/translators... 

http://docs.transifex.com

espadrine:
Haha, I learned about transifex literally two weeks ago. Great suggestion!

Plus they have an API: http://docs.transifex.com/api/

I'm not sure if we can do without the password requirement.

nelson6e65:
Yes... I was wondering about that... "limitation". :/

There is some Web Hook integration... but... I should learn more about it. Maybe can be useful.
http://docs.transifex.com/integrations/webhook/

nelson6e65:
Sheck this from widgets:

[![Resource example](https://www.transifex.com/projects/p/transifex/resource/core/chart/image_png)](https://www.transifex.com/projects/p/transifex/widgets/)

`https://www.transifex.com/projects/p/transifex/resource/core/chart/image_png`

https://www.transifex.com/projects/p/transifex/widgets/

jens-maus:
See here (https://github.com/transifex/transifex/issues/319) for more requests for proper github-type badges for transifex services. Would be really great to have these transifex type badges hopefully soon!

teleivo:
Yes please :1

UdoKifferbrehl:
+1

greatislander:
Not a perfect solution, but I made [this microservice](https://github.com/greatislander/tenpercent) which uses the Transifex API to get a resource's translation status. You can [use it](https://github.com/greatislander/tenpercent#usage) in a [dynamic badge](http://shields.io/#dynamic-badge).

nelson6e65:
Interesting, @greatislander. I'll check this later.

niccokunzmann:
There has be progress on Transifex badges since the creation of this issue. You can check it out as it is linked in the issue.

niccokunzmann:
I added a service for badges here: https://transifex-open-api.herokuapp.com/

mwithi:
> I added a service for badges here: https://transifex-open-api.herokuapp.com/

Why not a **summary** badge with all languages, expanded names, in a table layout?

I had to do like [this](https://github.com/informatici/openhospital-gui/pull/203/files)

niccokunzmann:
@mwithi a pull request is welcome if you like to add this functionality. There might be something like this provided by Transifex already.

PR #2909, opened 2019-02-02
Refactor endpoint functions [dynamic endpoint]

paulmelnikow:
Modularize the endpoint badge functions, making it easier to create a service which pipes an upstream service's endpoint badge data out to a Shields user. It could be useful for the Transifex badge being discussed in #2904 and #497.

Some of this, like moving `errorMessages` from `dynamic-helper` to `dynamic-common` is useful regardless. I think same for the fetch and validate, which we seem likely to use.

Though let's hold off briefly on merging this in case we go in a different direction with formatting / rendering the Transifex badge.

### DIFF
--- a/services/endpoint-common.js
+++ b/services/endpoint-common.js
@@ -1,0 +1,120 @@
+'use strict'
+
+const Joi = require('joi')
+const validate = require('../core/base-service/validate')
+const { InvalidResponse } = require('.')
+
+const optionalStringWhenNamedLogoPresent = Joi.alternatives().when(
+  'namedLogo',
+  {
+    is: Joi.string().required(),
+    then: Joi.string(),
+  }
+)
+
+const optionalNumberWhenAnyLogoPresent = Joi.alternatives()
+  .when('namedLogo', { is: Joi.string().required(), then: Joi.number() })
+  .when('logoSvg', { is: Joi.string().required(), then: Joi.number() })
+
+const endpointSchema = Joi.object({
+  schemaVersion: 1,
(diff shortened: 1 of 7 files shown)


niccokunzmann:
I had a look at it but I do not know how to support this pull request.

services/endpoint-common.js
```diff
@@ -0,0 +1,120 @@
...
+    url,
+    errorMessages,
+  })
+  return validateEndpointData(json, {
+    prettyErrorMessage: validationPrettyErrorMessage,
+    includeKeys,
+  })
+}
+
+function renderEndpointBadge(
```
calebcartwright: Do we envision this being used anywhere outside the Endpoint service?

services/endpoint/endpoint.service.js
```diff
@@ -72,34 +36,8 @@ module.exports = class Endpoint extends BaseJsonService {
...
-      labelColor,
-      namedLogo,
-      logoSvg,
-      logoColor,
-      logoWidth,
-      logoPosition,
-      style,
-      cacheSeconds,
-    }
+  static render(props) {
```
calebcartwright: Would it be more clear to maintain the original parameter definition here? I know there's a lot of properties but wondering if it would be helpful to be able to see what all is involved with the rendering from within the class

services/endpoint-common.js
paulmelnikow: My initial thought is "no," and I wonder about putting the code back where it was.

PR #2910, opened 2019-02-03
[NodePing] support for NodePing uptime monitoring service

fileformat:


NodePing:
The PR author isn't associated with NodePing (other than a happy customer) but we're willing to support this effort however we can.  If the project would find a free ongoing NodePing account useful, we can provide that.

services/nodeping/nodeping-status.service.js
```diff
@@ -0,0 +1,114 @@
...
+      upMessage,
+      downMessage,
+      upColor,
+      downColor,
+    })
+  }
+
+  static render({ status, label, upMessage, downMessage, upColor, downColor }) {
+    return status
+      ? { label, message: upMessage || 'up', color: upColor || 'brightgreen' }
```
calebcartwright: We should also have some tests that validate this logic. If you add the query param (`?style=_shields_test`) to the test `get` call then you can check the `color` on the json response

services/nodeping/nodeping-uptime.service.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  static formatPercentage(uptime) {
+    if (uptime === 100.0) {
+      return '100%'
+    }
+    return `${uptime.toFixed(3)}%`
+  }
+
+  static render({ uptime, label }) {
```
calebcartwright: I think we can get rid of `label` here

services/nodeping/nodeping-uptime.service.js
```diff
@@ -0,0 +1,122 @@
...
+    return {
+      label: 'Uptime',
+    }
+  }
+
+  static get route() {
+    return {
+      base: 'nodeping/uptime',
+      pattern: ':checkUuid',
+      queryParams: ['label'],
```
calebcartwright: You don't need to re-define the `label` query param here, this is default option available on all badges

services/nodeping/nodeping-uptime.service.js
```diff
@@ -0,0 +1,122 @@
...
+      new Date().getTime() - 30 * 24 * 60 * 60 * 1000
+    )
+      .toISOString()
+      .slice(0, 10)
+
+    const rows = await this._requestJson({
+      schema,
+      url: `https://nodeping.com/reports/uptime/${checkUuid}`,
+      options: {
+        method: 'GET',
```
calebcartwright: When using the `_requestJson` helper you don't need to explicitly specify the verb unless you need something other than `GET`

services/nodeping/nodeping-status.service.js
```diff
@@ -0,0 +1,114 @@
...
+      label: 'Status',
+    }
+  }
+
+  static get route() {
+    return {
+      base: 'nodeping/status',
+      pattern: ':checkUuid',
+      queryParams: [
+        'label',
```
calebcartwright: No need to re-define `label` as a query param here, it's already available for all badges

services/nodeping/nodeping-status.service.js
fileformat: I tried to run coverage tests, but the command in the tutorial ("npm run coverage:test:services -- --only=nodeping") does not like the only=nodeping ('npm WARN invalid config only="nodeping"'), so it tries to run all the tests, and some (before nodeping) fail.

calebcartwright: Yeah it needs a second `--` now, we'll need to update the tutorial accordingly. 

Try `npm run coverage:test:services -- -- --only=nodeping && npm run coverage:report`

NodePing:
@paulmelnikow - please sign up over at nodeping.com and then drop me a note at support@nodeping.com with the email address you used to sign up and I'll covert it to a free account.  Happy to support the project.

services/nodeping/nodeping-status.service.js
fileformat: That worked.  But I don't see where the uncovered branches are...

services/nodeping/nodeping-uptime.service.js
fileformat: Done

NodePing:
@fileformat, sorry, we don't have the ability to create a 'fake' check like that.

services/nodeping/nodeping-status.service.js
paulmelnikow: You might avoid the problem by using default parameters:

```js
static render({ status, upMessage = 'up', downMessage = 'down', …
```

services/nodeping/nodeping-uptime.service.js
```diff
@@ -0,0 +1,122 @@
...
+    })
+    return { uptime: rows[rows.length - 1].uptime }
+  }
+
+  async handle({ checkUuid }) {
+    const { uptime } = await this.fetch({ checkUuid })
+    return this.constructor.render({ uptime })
+  }
+
+  static calcColor(uptime) {
```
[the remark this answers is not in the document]
paulmelnikow: Ha! The ones in UptimeRobot are a bit silly, aren't they.

paulmelnikow:
We don't want a fake check, we want a real check :)

Take a look in `services/test-validators.js`; you could use or adapt `isPercentage`, for example.

NodePing:
@paulmelnikow, if you need a check that "always returns a specific uptime value", that won't be a real check.  Uptime will change for a real check.

services/nodeping/nodeping-status.service.js
calebcartwright: > That worked. But I don't see where the uncovered branches are...

What we should strive for are tests that validate all expected behaviors, which was the intent of my comment. It's not just about code coverage.

We've got default values being defined for the message and the color, with user definable values that will override those defaults. We'd want tests that validate those default values are applied as expected, and that the user-defined values are indeed used

calebcartwright:
Hi folks, just to clarify, we don't need to have a test that checks for a specific uptime *value*. This is covered in more detail in our [tutorial for writing service tests](https://github.com/badges/shields/blob/master/doc/service-tests.md), but in short the purpose of these tests is to validate that Shields can still integrate with a service and generate the respective badges correctly. 

For NodePing this would be a test that used a real checkUuid, and then verified the label and message (that the message is a percentage). For example:

```
t.create('some test title')
  .get(''/jkiwn052-ntpp-4lbb-8d45-ihew6d9ucoei.json'')
  .expectJSONTypes(
    Joi.object().keys({
      name: 'Uptime',
      value: isDecimalPercentage,
    })
  )
```

calebcartwright:
No worries! It does indeed look fixed now. I'll give this a final look over tonight but I think it's just about ready to be merged

services/nodeping/nodeping-status.tester.js
```diff
@@ -0,0 +1,79 @@
...
+  .expectJSON({
+    name: 'Status',
+    value: 'up',
+  })
+
+t.create('NodePing status - down (live)')
+  .get('/wf47q49c-zdmk-4ngx-88g4-wa7mzgluhhaf.json')
+  .expectJSON({
+    name: 'Status',
+    value: 'down',
```
calebcartwright: How confident are we that this will _always_ be down and that jkiwn052-ntpp-4lbb-8d45-ihew6d9ucoei will _always_ be up? 

I'm thinking it might be better to do a `withRegex` to ensure the value is `up` or `down` on these two tests to avoid any future test failures if the status for either of those changes briefly

fileformat:
The "up" test is on @NodePing, so if it is down, every other NodePing test will also be down.  They have a pretty good track record: if you look at the [human-readable version](https://nodeping.com/reports/uptime/jkiwn052-ntpp-4lbb-8d45-ihew6d9ucoei) for this checkUuid, you can see that they've only had 10 minutes of downtime in the past 2 years.

The "down" report is pinging a server that doesn't exist.  I can't imagine how it could start working.  It is in my personal account, but I'm a very satisfied NodePing customer (they are great, especially if you are cost-sensitive and have a zillion microservices and domains).  I could just delete this test, since the mocked results should still get us 100% code coverage.  Let me know if you want me to.

services/nodeping/nodeping-status.tester.js
calebcartwright: > The "up" test is on @NodePing, so if it is down, every other NodePing test will also be down. They have a pretty good track record: if you look at the [human-readable version](https://nodeping.com/reports/uptime/jkiwn052-ntpp-4lbb-8d45-ihew6d9ucoei) for this checkUuid, you can see that they've only had 10 minutes of downtime in the past 2 years.

> The "down" report is pinging a server that doesn't exist. I can't imagine how it could start working. It is in my personal account, but I'm a very satisfied NodePing customer (they are great, especially if you are cost-sensitive and have a zillion microservices and domains). I could just delete this test, since the mocked results should still get us 100% code coverage. Let me know if you want me to.

You can delete one of those tests (remember the live tests are more like integration/functional tests, not unit tests for code coverage purposes), but please change the validation to the set of allowed values vs. an explicit value to be consistent with the other tests. 

The easiest way to do this is probably via `Joi` by importing it:

```
const Joi = require('joi')
```

And then update the expect to:
```
.expectJSONTypes(
  Joi.object().keys({
    name: 'Status',
    value: Joi.equal('up', 'down').required(),
  })
)
```

calebcartwright:
I also noticed that if I supply a checkUuid that's invalid, the resulting badge renders `invalid response data` which is what happens when the schema validation fails.

It looks like NodePing returns an HTTP 200 response with an html page that includes a message `That report is not available` in that scenario (which is why the schema validation fails), which has me wondering what sort of error scenarios we may encounter and how the NodePing API responses look in those scenarios. 

Ideally, we want the badge to display a message that is meaningful for the user (as often times it happens due to user error, such as a typo in the badge parameters). For example with HTTP 404 scenarios our badge messages are typically something like `X not found` (`repository not found`, `package not found`, etc)

There may not be anything else we can do, but worth thinking through those scenarios

fileformat:
Changed the tests per your instructions.  Still 100% test coverage.

I don't want to try to parse the HTML that they return when there is an error:  seems iffy, and even if I got it to work, there are no stability guarantees for it.  The documented API will have to do.

calebcartwright:
> I don't want to try to parse the HTML that they return when there is an error: seems iffy, and even if I got it to work, there are no stability guarantees for it. The documented API will have to do.

Not proposing parsing html, we never do that 😄 Ideally, the API would return a different status code in that scenario but as I said there's not much we can do there. I was however wondering if there are any other error scenarios/status code we could receive, or whether the API will always return with a 200

fileformat:
I have only seen 200 responses.  The [docs](https://nodeping.com/reporting.html) don't say anything about status codes.

Note: this is NodePing's public API.  They have a more detailed private API, but that requires a security token and has write-access for the whole the account: not usable for shields.

PR #2919, opened 2019-02-04
delete LGTM logo

chris48s:
Before:

* 
* 

After:

* 
*


paulmelnikow:
Hey @samlanning,

For context, this is part of a laborious process in which @chris48s has contributed most of the Shields-maintained logos to a project called https://simpleicons.org/. That project is both widely available, and is kept up to date by people other than us 😄 See #2510 for that discussion.

This has already been helpful, e.g. our Slack logo is getting a bump for free.

The logos from simple-icons are monochrome, so we use the brand color when it works on the background, and replace it with a basic light or dark color when it provides too little contrast. (#2833)

The result is the lgtm logo is now monochrome. It renders a little differently on the regular badges, and is visible on the social badge (whereas it was completely invisible before).

Are you cool with this? Would be great to hear from you before this goes live.

chris48s:
If you need to revert this change to deploy, go for it

paulmelnikow:
Let's not block on it. We can revisit and quickly redeploy if needed.

samlanning:
Hey @paulmelnikow thanks for looping me in on this.

In general I'm pretty happy with this change, monochrome logo looks good to me.

The one subtle thing with our logo is that when the colours are inverted, the tarsier's head looks a little ghostly / skull like (and its eyes look spooky). So as part of our branding guidelines, we generally have 2 logo alternatives for light and dark backgrounds. But given that at least for shields.io the white is what we want, and it looks good, this LGTM. I'll consider working with simpleicons re: light and dark alternatives for icons.

PR #2930, opened 2019-02-05
Add CLI for debugging badges

paulmelnikow:
Often when responding to bug reports it would be helpful to easily run an example failing badge URL. It takes a while to do that, because you have to copy and paste just the right part of the badge URL. This works on `img.shields.io` links as well as partial paths and should make this really easy. 💨


chris48s:
Amazing - this is just obviously incredibly useful as soon as I started playing with it. How did it take this long to think of this?

paulmelnikow:
Yea, I was dreaming up more awesome ways to do this using a UI, which I think would be amazing and super-helpful.

But I realized the simple thing would make a big difference on its own :)

If we tweak the logging for service tests to print the base URL along with the sub-path, that would make it easy to use this to debug failing tests as well.

PR #2931, opened 2019-02-05
Update Dependabot logo

feelepxyz:
Hey, we've just rolled out an updated logo on Dependabot and would like to update it here too 🤖

## Preview with background colour
screen shot 2019-02-05 at 17 27 01


feelepxyz:
@PyvesB ah cool didn't know about simple-icons, will update that one too.

I would rather use the current one in this PR for Shields (without the hexagon) - it will look much better in the badge than the full version in simple-icons.

PyvesB:
@chris48s what's your take on this?

chris48s:
Dependabot was one of the services where we kept a custom logo instead of using the simple-icons one. The custom version is clearer at small size and is the same one dependabot use on their own badge:

simple-icons:
https://svgshare.com/s/AWi
https://svgshare.com/s/AVi

shields:
https://svgshare.com/s/AWP

I actually used the dependabot logo as an example in the logo tests, which is why changing the logo has cause the the build to fail :smile:  so I'm going to push a couple of fixup commits - one to minify the logo and one to fix the tests, but thanks for updating us with the new design :+1:

chris48s:
oh, and the new logo: 
looks almost the same at this size, but it is subtly rounder :)

before: https://svgshare.com/s/AWP
after: http://svgshare.com/s/B2N

feelepxyz:
@chris48s amazing! Thanks for fixing it up!

PR #2934, opened 2019-02-05
Refactor the NuGet v2 badges; switch Resharper to XML; patch for missing download counts [resharper powershell chocolatey]

### LINKED ISSUE #2921: Resharper Tests Failing
Service tests started failing ~5 days ago

:camera: **Live badge**

I don't think so (getting unparseable json response)

:link: **CircleCI link**
https://circleci.com/gh/badges/daily-tests/131

:beetle: **Stack trace**
6 tests failing with varying degrees of stack trace output detail, best example is probably:
```
ValidationError: child "value" fails because ["value" with value "unparseable json response" fails to match the required pattern: /^[1-9][0-9]*[kMGTPEZY]?$/]
```

:bulb: **Possible solution**
🤷‍♂️

paulmelnikow:
This closes #2921 by switching ReSharper to the XML API used by Powershell, and refactors the powershell code back into the common nuget v2 service class. It also removes mocked tests of the color logic, replacing them with smaller-bracket tests that accomplish the same thing more concisely.

The ReSharper download counts are returning -1 for now; see https://github.com/badges/shields/issues/2921#issuecomment-461078513; so I've added a temporary message.

The test will fail when they start working again, and we can remove this code at the same time.


paulmelnikow:
Let's merge this and then keep #2921 open to deal with the still-failing download badge.

calebcartwright:
Self-requesting a review to remind me to look at this one tonight 😄

services/nuget/nuget-helpers.spec.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
+
+const { renderVersionBadge, odataToObject } = require('./nuget-helpers')
+const { test, given } = require('sazerac')
+
+describe('NuGet helpers', function() {
+  test(renderVersionBadge, () => {
+    given({ version: '1.2-beta' }).expect({
+      label: undefined,
```
calebcartwright: This confused me at first, but it's just because we're not specifying a `feed` argument right?

services/nuget/nuget-v2-service-family.js
```diff
@@ -27,25 +36,55 @@ const schema = Joi.object({
...
+  }
 
   if (packageData) {
     return packageData
   } else if (!includePrereleases) {
     return fetch(serviceInstance, {
+      odataFormat,
       baseUrl,
       packageName,
       includePrereleases: true,
```
calebcartwright: This is my first time looking at this service family so trying to get a feel for it. Is this `else if` block kinda like a retry? If we didn't get any package data, and prereleases were excluded, then try again?

I'm wondering why the clause is `!includePrereleases` but the arg is set to `true` here

services/nuget/nuget-v2-service-family.js
```diff
@@ -155,10 +205,14 @@ function createServiceFamily({
 
     async handle({ packageName }) {
       const packageData = await fetch(this, {
+        odataFormat,
         baseUrl: apiBaseUrl,
         packageName,
       })
       const { DownloadCount: downloads } = packageData
+      if (downloads === -1) {
+        throw new Inaccessible({ prettyMessage: 'temporarily unavailable' })
```
calebcartwright: 👍

services/nuget/nuget-helpers.spec.js
paulmelnikow: Right. `const { foo, bar } = { foo: 123 }` will give `bar` a value of `undefined`.

services/nuget/nuget-v2-service-family.js
paulmelnikow: Probably deserves a comment. IIRC the idea is that if there are _only_ prereleases, we show the latest of them.

PR #2939, opened 2019-02-06
Redirect from alias to canonical URL [azuredevops vso]

### LINKED ISSUE #2340: Redirect aliased badges to the canonical form

paulmelnikow:
As described in #2340, this provides a way to replace an old alias with a redirect. This makes it easier to migrate our URLs over time without making our matching patterns more complicated.

The 301 redirect is sent back to the requester. If a user pastes the aliased URL into the address bar, it'll be replaced in the browser with the new URL, which gently encourages them to migrate.

We can do some follow-ons to migrate a few more of these.

Close #2340


calebcartwright:
Haven't gone through this yet, but I'm pumped about the feature 💯

calebcartwright:
Self-requesting a review to remind me to look at this one tonight 😄

core/base-service/redirector.js
```diff
@@ -0,0 +1,66 @@
...
+    static get category() {
+      return category
+    }
+
+    static get route() {
+      return route
+    }
+
+    static get isDeprecated() {
+      return true
```
calebcartwright: Will we need to register all the redirected services in the deprecated services list?

core/base-service/redirector.spec.js
```diff
@@ -0,0 +1,95 @@
...
+          followRedirect: false,
+        }
+      )
+
+      expect(statusCode).to.equal(301)
+      expect(headers.location).to.equal(
+        '/new/service/hello-world.svg?colorB=123&style=flat-square'
+      )
+    })
+  })
```
calebcartwright: Is it feasible (or even helpful for that matter) to test the cached/304 scenario?

services/azure-devops/azure-devops-build.service.js
```diff
@@ -34,17 +34,16 @@ module.exports = class AzureDevOpsBuild extends BaseSvgScrapingService {
 
   static get route() {
     return {
-      base: '',
-      format: '(?:azure-devops|vso)/build/([^/]+)/([^/]+)/([^/]+)(?:/(.+))?',
-      capture: ['organization', 'projectId', 'definitionId', 'branch'],
+      base: 'azure-devops/build',
+      pattern: ':organization/:projectId/:definitionId/:branch*',
```
calebcartwright: so much better 👍

services/azure-devops/vso-redirect.service.js
```diff
@@ -0,0 +1,30 @@
...
+
+const { redirector } = require('..')
+
+module.exports = [
+  redirector({
+    category: 'build',
+    route: {
+      base: 'vso/build',
+      format: '([^/]+)/([^/]+)/([^/]+)(?:/(.+))?',
+      capture: ['organization', 'projectId', 'definitionId', 'branch'],
```
calebcartwright: Will we need to use `format` and `capture` in redirects or can we use `pattern` in redirect services?

services/azure-devops/vso-redirect.service.js
```diff
@@ -0,0 +1,30 @@
...
+    },
+    target: ({ organization, projectId, definitionId, branch }) => {
+      let path = `/azure-devops/build/${organization}/${projectId}/${definitionId}`
```
calebcartwright: It would be cool if we could grab this value from the route defined in the corresponding service (azure-devops-build service in this case) to having these defined in multiple places. 

If that's not possible/feasible, is there anything we could do to help avoid route updates being made in a service but forgotten here? Maybe another danger check?

calebcartwright:
As a follow up item we should probably add documentation for redirects somewhere

services/azure-devops/vso-redirect.service.js
paulmelnikow: I started to think that maybe this should use path-to-regexp compile. After thinking about it for a while, I concluded it wasn't worth it for two reasons:

1. With few exceptions, we don't make breaking changes in routes. When we change routes, we are careful to keep the old ones working. So if the target route changes, it's likely we will have kept the old target route working, too.
2. There are not going to be a massive amount many of these redirects, so this feels less important to get perfect compared to other parts of the codebase like the core services and badge helper functions (both of which could use attention). It doesn't seem essential to make this totally DRY.

services/azure-devops/vso-redirect.service.js
paulmelnikow: It's exactly the same as BaseService so it can be `pattern`. I thought I updated one of them but I guess I missed this one? Feel free to edit it if you'd like.

core/base-service/redirector.spec.js
paulmelnikow: The underlying functions are tested. 🤷‍♂️Seems like a test would be nice to have, not essential (mostly in comparison to other parts of the roadmap). Also I'd really like to replace scoutcamp with something more modern and anything that touches a request would likely need to be rewritten. I guess that's not an argument against testing, more contributing to an instinct to put a light touch on this part of the code. This is all adapted from `base-static.js` by the way.

core/base-service/redirector.js
paulmelnikow: That list doesn't do anything. Those dates are useful to have readable in code, because it could let us write tests that remind us to downgrade or remove them. Though I think the dates could be placed in the call to `deprecatedService` as easily.

calebcartwright: Would it be worth writing down redirected services somewhere in a similar fashion to keep track of them?

core/base-service/redirector.spec.js
calebcartwright: works for me

services/azure-devops/vso-redirect.service.js
calebcartwright: > I concluded it wasn't worth it for two reasons

I'm good with that 

> 1. When we change routes, we are careful to keep the old ones working 

That's the exact use case I was thinking through, and wanting to make sure that when/if that happens, that route updates would have to be made here to prevent breaking users leveraging the old route. 

> 2. There are not going to be a massive amount many of these redirects, so this feels less important to get perfect compared to other parts of the codebase...

Also agreed, doesn't need to be a blocker. I've had to do a few of these pattern-based old/new routes while refactoring services so just wanted to pose the question to avoid any potential future route changes being missed on redirects

paulmelnikow: > > 1. When we change routes, we are careful to keep the old ones working
>
> That's the exact use case I was thinking through, and wanting to make sure that when/if that happens, that route updates would have to be made here to prevent breaking users leveraging the old route.

If we keep the old one working, updating this is optional.

1. /vso/foo/bar redirects to /azure-devops/foo/bar
2. /azure-devops/foo/bar is changed to /azure-devops/build/foo/bar.
3. To provide backward compatibility with badge URLs in the wild, a new redirect is added, redirecting /azure-devops/foo/bar to /azure-devops/build/foo/bar
4. /vso/foo/bar could be updated too. That would eliminate an extra redirect. However if it isn't, it will still work because of 3.

core/base-service/redirector.js
paulmelnikow: I suppose we could stop supporting the old routes after a good while. In that case having the date we started redirecting would be helpful.

I'd suggest we do that inline by throwing another parameter in the function call.

services/azure-devops/vso-redirect.service.js
calebcartwright: Done

core/base-service/redirector.js
calebcartwright: I'm good to approve this, can wait if we wanna do that though

services/azure-devops/vso-redirect.service.js
calebcartwright: Step 3 is something the user would have to add to the redirect service as part of updating the route (step 1) right?

paulmelnikow: Hmm, no I don't think so. Step 3 is something that would happen in conjunction with Step 2, by the Shields developer who decides to change the route again.

calebcartwright: Oops, bad wording on my part. I meant user in the context of a developer sorry.

core/base-service/redirector.js
paulmelnikow: Have just added dates.

PR #2943, opened 2019-02-06
refactor [jitpack] service

chris48s:


services/jitpack/jitpack.service.js
```diff
@@ -31,58 +31,26 @@ module.exports = class Jitpack extends LegacyService {
...
-            badgeData.text[1] = 'not found'
-            sendBadge(format, badgeData)
-            return
-          }
-          try {
-            const data = JSON.parse(buffer)
-            const status = data['status']
-            let color = versionColor(data['version'])
-            let version = versionText(data['version'])
-            if (status !== 'ok') {
```
chris48s: I don't have a clear example of when this happens, but instead of explicitly handling this, I've moved `status == 'ok'` to being enforced in the schema so we'll throw invalid response there instead. Does that seem reasonable?

PR #2952, opened 2019-02-07
Update from RawGit to local path

pxgamer:
As per RawGit's index, the service is being shut down in October 2019. They mention:

>If you're currently using RawGit, please stop using it as soon as you can.

They recommend using jsDelivr or a few other CDNs going forward. jsDelivr has an easy [transfer page](https://www.jsdelivr.com/rawgit) available

See [rawgit.com](https://rawgit.com) for details on the sunset phase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-    <img src="https://cdn.jsdelivr.net/gh/badges/shields@master/frontend/images/logo.svg"
+    <img src="./frontend/images/logo.svg"
         height="130">
 </p>
 <p align="center">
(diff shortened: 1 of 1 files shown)


paulmelnikow:
Huh, doesn't GitHub provide something we can use instead?

pxgamer:
Not that I know of (I guess it could use GitHub pages). But using pure GitHub returns a plaintext mimetype.
https://github.com/badges/shields/raw/master/frontend/images/logo.svg

I assumed this was the reason for using RawGit (3rd party service) in the first place.

paulmelnikow:
> If this is only used in the README, I guess it could just point to `./frontend/images/logo.svg` 🤔

I feel like that would be fine.

pxgamer:
Ah, ok. 👍 I'll change to that 😋

PR #2956, opened 2019-02-07
Example keywords validation

### LINKED ISSUE #2551: Validate keywords in validateExample
_Originally posted by @chris48s in https://github.com/badges/shields/pull/2550#discussion_r242696044_

> we don't need to specify either of these as keywords because they both appear in the title

Indeed!

This could be checked in `validateExample`.

PyvesB:
This pull request closes #2551: making sure that the keywords don't already appear in the example's title.

I also added validation that checks that they are at least two characters long, as this is enforced by the homepage when type your search.


core/base-service/transform-example.js
```diff
@@ -85,6 +85,34 @@ function validateExample(example, index, ServiceClass) {
...
+        } at index ${index}, keywords contains words that are less than two characters long: ${tinyKeywords.join(
+          ', '
+        )}`
+      )
+    }
+    // Make sure none of the keywords are already included in the title.
+    const title = (example.title || ServiceClass.name).toLowerCase()
+    const redundantKeywords = example.keywords.filter(k =>
+      title.includes(k.toLowerCase())
+    )
```
paulmelnikow: This _seems_ equivalent to the code that's in https://github.com/badges/shields/blob/de916536e42f90c204ca56830b47dcef041152c4/frontend/lib/service-definitions/service-definition-set-helper.js#L3-L14. It's hard to tell for sure that this is equivalent. Also if that code changed, I could see this slipping out of skew

It would be good to refactor that code and this code so it's the same code logic invoked in either case. I think that could be handled in a follow-on.

services/cran/cran.service.js
```diff
@@ -75,7 +74,6 @@ class CranVersion extends BaseCranService {
         title: 'CRAN/METACRAN',
         namedParams: { packageName: 'devtools' },
         staticPreview: this.render({ version: '2.0.1' }),
-        keywords: ['R'],
```
paulmelnikow: 🎉

core/base-service/transform-example.js
PyvesB: I don't think these two bits of code serve the same purpose. In my case we want to check whether any of the keywords are a subset of the title or service name, and fail validation by listing all redundant keywords. In the bit of code you linked, we know that all keywords are valid and we group them together with the title to form one big string we try to match the user input with. I don't see much opportunity for extraction or refactoring. 😕

PyvesB: To be honest, I'm not really seeing your point here. You're suggesting to replace a single line `title.includes(k.toLowerCase())` statement by four nested function calls (titleMatchesQuery -> exampleMatchesQuery -> predicateFromQuery -> exampleMatchesRegex) amounting to roughly ten times the amount of code in that call stack. The code in question is non trivial, probably has poorer performance (regexing + many transformations), and we're relying on its internal behaviour and the fact that `[title].concat(undefined)` doesn't blow up when faking our example as part of the `{ title } -> example` transformation when calling `exampleMatchesQuery`.

I do understand your rational for wanting to extract common logic and am all for it, but when it's one-liners as simple as "does include ignore case" for bits of code that belong in two different realms (frontend vs. backend), in my humble opinion I don't think we should get ahead of ourselves. I believe we would get more value from trying to simplify what's in _service-definition-set-helper.js_ and possibly in the future extract functionality if the search logic becomes more fancy or changes too often.

Maybe we could set up a call on Discord/a live pair programming session if we can't reach an agreement here? 😄

paulmelnikow: My point is it's difficult to know that these two expressions of logic are equivalent.

I think refactoring service-definition-set-helper to make it more readable, while ensuring it stays fast, is a good idea, though I don't think that's blocking this work. I also don't think this is crucial; there are a lot of opportunities to improve things in this codebase, and plenty of ideas about which ones are the most important.

I'm happy to discuss this more if you want, though I mentioned I thought this was optional so by all means merge it if you'd like.

paulmelnikow: I'll open another PR adding some comments.

PyvesB:
Thanks for merging, I was somewhat lacking the bandwidth to finalise this in the past couple of days!

PR #2962, opened 2019-02-08
Add some more perfunctory coverage of the frontend

paulmelnikow:


frontend/components/common.spec.js
```diff
@@ -0,0 +1,67 @@
...
+
+  describe('<H2 />', function() {
+    it('renders', function() {
+      shallow(<common.H2 />)
+    })
+  })
+
+  describe('<H3 />', function() {
+    it('renders', function() {
+      shallow(<common.H3 />)
```
chris48s: Might be just being slightly ignorant here, but what are these tests actually checking? What would I need to change/break to cause this test to fail?

paulmelnikow: Not much. That the code doesn’t crash. Otherwise you have to play with the whole site to figure that out.

PR #2937, opened 2019-02-06
Remove dev dep imports in production code

calebcartwright:
Fixes #2876 with @paulmelnikow's suggestion 

Moved imports of `ServiceTester` and `createServiceTester` to a separate file so that dev dependencies are not imported by service classes

PR #2966, opened 2019-02-09
Update coverage ignores

paulmelnikow:


.nycrc.json
```diff
@@ -10,11 +10,9 @@
     "**/*-test-helpers.js",
     "dangerfile.js",
     "gatsby-*.js",
+    "core/service-test-runner",
     "services/**/*.tester.js",
```
calebcartwright: I just realized I missed adding `services/tester.js` in #2937

PR #2969, opened 2019-02-10
Add badges for [Keybase]

SkypLabs:
Hi,

I've implemented different badges for the [Keybase](https://keybase.io/) service:

* PGP address
* Bitcoin address
* Zcash address
* Stellar address

image

I look forward to hearing your feedback.


Prouser123:
can you rename your PR to something like `Add badges for [Keybase]`?

The square brackets `[]` are used by CircleCI to run the correct tests.

Prouser123:
👍 , next can you merge in the latest changes from the main master branch (`upstream/master`)? That should also run the tests.

SkypLabs:
The latest changes in `upstream/master` are behind mine in `keybase`. This is nothing to merge.

Prouser123:
You are 16 commits ahead (your changes), but 5 commits behind (changes merged since you forked).

services/keybase/keybase-btc.tester.js
```diff
@@ -0,0 +1,23 @@
...
+const Joi = require('joi')
+
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('existing bitcoin address')
+  .get('/skyplabs.json')
+  .expectJSONTypes(
+    Joi.object({
+      name: 'btc',
+      value: Joi.string(),
```
Prouser123: `Joi.string()` will allow 'not found' to get through.

You can use `withRegex(/^(?!not found$)/)` instead to match everything except `not found`.

Alternatively you can use/create something like`withRegex(/^[13][a-km-zA-HJ-NP-Z1-9]{25,34}$/)` which will only match a valid bitcoin address.

Add `withRegex` with `const { withRegex } = require('../test-validators')` at the top of the file.

SkypLabs:
> You are 16 commits ahead (your changes), but 5 commits behind (changes merged since you forked).

Indeed, I didn't see them.

> You can use `git fetch upstream` if you are using the command-line. You can then checkout your local branch and run `git merge upstream/master`

Thanks but no worries, I know how to do it :+1:

services/keybase/keybase-xlm.tester.js
```diff
@@ -0,0 +1,23 @@
...
+const Joi = require('joi')
+
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('existing stellar address')
+  .get('/skyplabs.json')
+  .expectJSONTypes(
+    Joi.object({
+      name: 'xlm',
+      value: Joi.string(),
```
Prouser123: Same as before, `Joi.string()` on its own will allow `not found`.

services/keybase/keybase-zec.tester.js
```diff
@@ -0,0 +1,23 @@
...
+const Joi = require('joi')
+
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('existing zcash address')
+  .get('/skyplabs.json')
+  .expectJSONTypes(
+    Joi.object({
+      name: 'zec',
+      value: Joi.string(),
```
Prouser123: Again, `Joi.string()` will allow `not found`.

services/keybase/keybase-btc.tester.js
SkypLabs: Thanks. It's indeed better now.

Done for BTC.

services/keybase/keybase-xlm.tester.js
SkypLabs: I used the `withRegex(/^(?!not found$)/)` regex as I haven't found a good one for Stellar addresses. I will read the Stellar documentation to figure it out when I get a chance.

services/keybase/keybase-zec.tester.js
SkypLabs: Same as for Stellar addresses, I used the `withRegex(/^(?!not found$)/)` regex for now.

services/keybase/keybase-btc.service.js
```diff
@@ -0,0 +1,98 @@
...
+
+  static get examples() {
+    return [
+      {
+        title: 'Keybase BTC',
+        namedParams: { username: 'Keybase username' },
+        staticPreview: this.render({
+          address: '12ufRLmbEmgjsdGzhUUFY4pcfiQZyRPV9J',
+        }),
+        keywords: ['keybase', 'btc', 'bitcoin'],
```
calebcartwright: the values in `title` are automatically included as searchable keywords, so we can remove `keybase` and `btc` here

services/keybase/keybase-btc.service.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'btc' }
+  }
+
+  async handle({ username }) {
+    const options = {
+      method: 'GET',
```
calebcartwright: it's not necessary to include the method here unless you need to use a verb other than `get`

services/keybase/keybase-btc.service.js
```diff
@@ -0,0 +1,98 @@
...
+        message: 'not found',
+        color: 'inactive',
+      }
+    }
+  }
+
+  static render({ address }) {
+    return {
+      message: address,
+      color: 'informational',
```
calebcartwright: since the color will always be static let's move that into the `defaultBadgeData` function

services/keybase/keybase-profile.js
```diff
@@ -0,0 +1,20 @@
...
+const { BaseJsonService } = require('..')
+
+module.exports = class KeybaseProfile extends BaseJsonService {
+  static get apiVersion() {
+    throw new Error(`apiVersion() is not implemented for ${this.name}`)
+  }
+
+  async fetch({ schema, options }) {
+    const apiVersion = this.constructor.apiVersion
+    const url = `https://keybase.io/_/api/${apiVersion}/user/lookup.json`
```
calebcartwright: Can you add an inline comment here that provides a link to the API documentation?

services/keybase/keybase-btc.service.js
SkypLabs: Done :+1:

services/keybase/keybase-btc.service.js
```diff
@@ -0,0 +1,98 @@
...
+    const data = await this.fetch({
+      schema: bitcoinAddressFoundOrNotFound,
+      options,
+    })
+
+    try {
+      const address = data.them[0].cryptocurrency_addresses.bitcoin[0].address
+      return this.constructor.render({ address })
+    } catch (err) {
+      return {
```
calebcartwright: How common a scenario will the `bitcoin` array be empty? If it's something that could occur relatively frequently then it's better to handle it just like you have here (allows for providing the user a better error message), but if there really should be an entry then this could be enforced in the schema validation, and we could remove the `try/catch` block

services/keybase/keybase-profile.js
SkypLabs: Done :+1:

services/keybase/keybase-btc.service.js
```diff
@@ -0,0 +1,98 @@
...
+        message: 'not found',
```
calebcartwright: What sort of error responses could we see from the API? For example if a user specifies an invalid value for the `username` param, will the API return a 404 error? The default error message that will be displayed on 404 responses is `not found` so I'm thinking we may be able to provide more detailed/context specific error messages (for example something like `bitcoin address not found` here vs. `username not found` if the API returns a 404)

services/keybase/keybase-btc.service.js
SkypLabs: Indeed. Done.

services/keybase/keybase-btc.service.js
SkypLabs: Good to know. Thanks.

services/keybase/keybase-btc.service.js
SkypLabs: It's common actually. Lots of Keybase profiles don't have any bitcoin addresses.

services/keybase/keybase-btc.service.js
SkypLabs: The API returns 200 even if the profile doesn't exist.

See it by yourself:

* A valid profile (mine): https://keybase.io/_/api/1.0/user/lookup.json?usernames=skyplabs
* A non-existing profile: https://keybase.io/_/api/1.0/user/lookup.json?usernames=skyplabssssss

calebcartwright: Gotcha thanks. In that case, what if we did something like:

```
if (data.them[0] === null) {
  return {
    message: 'profile not found',
    color: 'critical'
  }
}

const bitcoinAddresses = data.them[0].cryptocurrency_addresses.bitcoin
if (bitcoinAddresses.length === 0) {
  return {
    message: 'no bitcoin addresses',
    color: 'inactive',
  }
}

return this.constructor.render({ address: bitcoinAddresses[0].address })

File: services/keybase/keybase-btc.service.js (hunk 8 above)
Comment (calebcartwright) [reply services/keybase/keybase-btc.service.js]: 👍

**Comment by calebcartwright** [comment]
Viewable @ https://shields-staging-pr-2969.herokuapp.com/

**Comment by Prouser123** [comment]
image

Do you think we should change the category? Every badge in Social is a social badge.

**Comment by calebcartwright** [comment]
> Do you think we should change the category?

I'd say yes. I get the original idea but I'm not sure Keybase fits in the Social category of Shields badges. I'd say `other` is probably the best category

File: services/keybase/keybase-btc.service.js (hunk 9 above)
Comment (SkypLabs) [reply services/keybase/keybase-btc.service.js]: I like this idea. However, I encountered some unexpected side effects while implementing it.

First, when I use the dashboard (the one on the screenshots) to generate a bitcoin badge, a request is sent for each letter of my username I type. My username being `skyplabs`, a first request is sent for `s`. The keybase API seems to return an HTTP 100 response in this case: https://keybase.io/_/api/1.0/user/lookup.json?usernames=s&fields=cryptocurrency_addresses.

I added this error to the `errorMessages` argument of `_requestJSON`:

```
  async fetch({ schema, options }) {
    const apiVersion = this.constructor.apiVersion
    // See https://keybase.io/docs/api/1.0/call/user/lookup.
    const url = `https://keybase.io/_/api/${apiVersion}/user/lookup.json`

    return this._requestJson({
      url,
      schema,
      options,
      errorMessages: {
        100: 'bad list value',
      }
    })
  }
}
```

But it doesn't seem to catch it:

```
[server] (node:10480) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'length' of undefined                                                                                                     
[server]     at KeybaseBTC.handle (/home/skyper/Documents/shields/services/keybase/keybase-btc.service.js:70:19)                                                                                                  
[server]     at <anonymous>
[server]     at process._tickDomainCallback (internal/process/next_tick.js:228:7)
```

The `property '[...] length' of undefined` error (line 70) comes from this piece of code:

```
 65     const data = await this.fetch({                                                                                                                                                                            
 66       schema: bitcoinAddressFoundOrNotFound,
 67       options,
 68     })
 69 
 70     if (data.them.length === 0) {
 71       return {
 72         message: 'profile not found',
 73         color: 'critical',
 74       }
 75     }
```

However, knowing that the API response doesn't contain the field `them` at all, it shouldn't match the two schemas I defined and consequently, shouldn't execute this chunk of code.

I also have a similar error with `const bitcoinAddresses = data.them[0].cryptocurrency_addresses.bitcoin`. Some profiles don't have the `cryptocurrency_addresses` field at all and it is the case of `sky`: https://keybase.io/_/api/1.0/user/lookup.json?usernames=sky&fields=cryptocurrency_addresses. This request is sent while typing `skyplabs` and similarly to the other error described earlier, my schemas let it pass.

Could you please have a look at my schemas? I have no experience with `Joi` and all of my attempts to fix these issues failed. Thank you in advance.

Comment (calebcartwright) [reply services/keybase/keybase-btc.service.js]: > Could you please have a look at my schemas?

Sure! If you haven't already done so, can you push your latest changes? I'll take a look later today

Comment (SkypLabs) [reply services/keybase/keybase-btc.service.js]: > Sure! If you haven't already done so, can you push your latest changes? I'll take a look later today

Thanks! I just pushed my changed (only for Keybase BTC for the moment).

> The keybase API seems to return an HTTP 100 response in this case

My bad. Even in this case, the API returns an HTTP 200 response containing an arbitrary status code 100 in the response body. The layout of my JSON viewer plug-in misled me.

**Comment by SkypLabs** [comment]
Keybase is a social platform, hence my choice to put it in the social category. If you think that all social badges should have a logo, there is a Keybase one on SimpleIcons: https://simpleicons.org/?q=keybae.

Being saying that, I go by your choice.

File: services/keybase/keybase-btc.service.js (hunk 9 above)
Comment (calebcartwright) [reply services/keybase/keybase-btc.service.js]: The problem is that the API response on a missing profile returns an array for `them` with the first item being a `null` (that's why the original check I proposed was `if (data.them[0] === null) {`) . If there's other scenarios where `them` is an empty array then we can update the first check accordingly to account for both scenarios

```js
if (data.them.length === 0 || !data.them[0]) {
  return {
    message: 'profile not found',
    color: 'critical',
  }
}
```

The tests are passing in my local env with the above:
```sh
  KeybaseBTC
    existing bitcoin address
      ✓
        [ GET /skyplabs.json ] (389ms)
    unknown username
      ✓
        [ GET /skyplabsssssss.json ] (409ms)


  2 passing (1s)

```

calebcartwright: > First, when I use the dashboard (the one on the screenshots) to generate a bitcoin badge, a request is sent for each letter of my username I type. My username being `skyplabs`, a first request is sent for `s`

I assume you're talking about the modal badge-creation window that helps users build the correct badge url? (below screenshot from my local env after making the changes mentioned above)

image

I'm pretty sure the request is only fired after a timeout period after the last keystroke, or after focus is removed from the parameter text boxes. Either way, I wouldn't worry about that too much.

calebcartwright: And lastly, regarding the schemas, I think what will work best is a single schema (shown below), and remove the secondary schema and alternatives. 

```js
const bitcoinAddressSchema = Joi.object({
  them: Joi.array()
    .items(
      Joi.object({
        cryptocurrency_addresses: Joi.object({
          bitcoin: Joi.array()
            .items(
              Joi.object({
                address: Joi.string().required(),
              }).required()
            )
            .required(),
        }).required(),
      })
        .required()
        .allow(null)
    )
    .min(0)
    .max(1)
    .required(),
}).required()
```

services/keybase/keybase-btc.service.js
```diff
@@ -0,0 +1,107 @@
...
+    return {
+      message: address,
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Keybase BTC',
+        namedParams: { username: 'Keybase username' },
```
calebcartwright: For the example params, we want to list a real value so that a user could plug that in and get a real badge. Let's put a real username here (I'd suggest putting your `skyplabs` if you don't mind 😄)

SkypLabs: Ahah done :+1:

services/keybase/keybase-btc.service.js
SkypLabs: > I'm pretty sure the request is only fired after a timeout period after the last keystroke, or after focus is removed from the parameter text boxes. Either way, I wouldn't worry about that too much.

Actually, I didn't mean that it was a problem. I just mentioned it because, having used it during my tests, it sent requests while I was typing my username that allowed my to identify potential issues, such as a simple `s` as username which returned an error message not documented. It is a good thing in the end.

> And lastly, regarding the schemas, I think what will work best is a single schema (shown below), and remove the secondary schema and alternatives.

Indeed, with only one schema and using the `allow(null)` method, I managed to make it work. Thanks!

calebcartwright: Awesome! I was browsing the API docs and came across this: https://keybase.io/docs/api/1.0/responses which details the response object, including the status code of 100 which is something I can reproduce with certain username values. It's not immediately clear to me why sometimes we receive that 100 error response, but other times there's a `them` array with a single `null` entry. 

I think we could make one final tweak to the schema to catch this and provide a better error message to the user (`profile not found` instead `invalid response data`)

You should be able to utilize our `nonNegativeInteger` schema helper from `validators`:
```js
```

Then update the schema to include the status and code (and remove the `required` on the `them` array):
```js
const bitcoinAddressSchema = Joi.object({
  status: Joi.object({
    code: nonNegativeInteger.required(),
  }).required(),
  them: Joi.array()
    .items(
      Joi.object({
        cryptocurrency_addresses: Joi.object({
          bitcoin: Joi.array().items(
            Joi.object({
              address: Joi.string().required(),
            }).required()
          ),
        })
          .required()
          .allow(null),
      })
        .required()
        .allow(null)
    )
    .min(0)
    .max(1),
}).required()
```

And then finally update the first if condition to check the value of that status code (the schema `min` check on the `them` will ensure there's a value so we don't have to include the 0 length check in the code):
```js
if (data.status.code !== 0 || !data.them[0]) {
  return {
    message: 'profile not found',
    color: 'critical',
  }
}    
```

SkypLabs: > I was browsing the API docs and came across this: https://keybase.io/docs/api/1.0/responses which details the response object, including the status code of 100 which is something I can reproduce with certain username values.

Oh I didn't pay much attention to the number of the error code described on this page when I started my work. So, the status code 100 is well referenced in their documentation unlike what I said. These errors are pretty rare so it didn't ring a bell.

> I think we could make one final tweak to the schema to catch this and provide a better error message to the user (`profile not found` instead `invalid response data`)

Agreed. However, knowing that a status code 100 is semantically different from a status code 0 with an empty array, I would suggest returning `invalid username` instead of `profile not found` in such cases.

SkypLabs: Implemented.

SkypLabs:
Regarding the story about the category, would it be worth adding the Keybase logo by default and keep it in `Social`?

services/keybase/keybase-btc.service.js
calebcartwright: Awesome, thanks!

calebcartwright:
> Regarding the story about the category, would it be worth adding the Keybase logo by default and keep it in `social`?

A couple items there. We recently moved away from managing our own logos (with a handful of exceptions) and relying on [Simple Icons](https://github.com/simple-icons/simple-icons) for logos. It looks like they already have the Keybase logo over there (https://simpleicons.org/?q=keybase) 

For example: 

Logos can be used with any badge (a badge in any category). The social badges in Shields have a unique badge style/intent, and even though Keybase is a social platform, I still don't think the Shields social badge style is the best fit for it

calebcartwright:
Also let's leave the keybase badges in the `social` category for now, but keep the default style as-is (we don't want to apply the social style by default)

SkypLabs:
> It looks like they already have the Keybase logo over there (https://simpleicons.org/?q=keybase)

Yes I've seen it. I think you missed my comment about it: https://github.com/badges/shields/pull/2969#issuecomment-462534507

> Also let's leave the keybase badges in the `social` category for now, but keep the default style as-is (we don't want to apply the social style by default)

Ok :+1:

calebcartwright:
> I think you missed my comment about it: [#2969 (comment)](https://github.com/badges/shields/pull/2969#issuecomment-462534507)

Yup! I missed the `by default` part

SkypLabs:
Is it ready to be merged so? I'm just checking if you need more actions from me.

calebcartwright:
> Is it ready to be merged so? I'm just checking if you need more actions from me.

I'm going to take a closer look at it in a bit, but yes I think it's close

services/keybase/keybase-btc.service.js
```diff
@@ -0,0 +1,111 @@
...
+    .min(0)
+    .max(1),
+}).required()
+
+module.exports = class KeybaseBTC extends KeybaseProfile {
+  static get apiVersion() {
+    return '1.0'
+  }
+
+  static get category() {
```
calebcartwright: Since we're going to use the same category, we can move this function to the base class

services/keybase/keybase-pgp.service.js
```diff
@@ -0,0 +1,106 @@
...
+      options,
+    })
+
+    if (data.status.code !== 0) {
+      return {
+        message: 'invalid username',
+        color: 'critical',
+      }
+    }
+
```
calebcartwright: We've got these same two `if` blocks repeated in every service class. Let's refactor these into a `transform` method in the base class (I'll add more details later on)

services/keybase/keybase-profile.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+const { BaseJsonService } = require('..')
+
```
calebcartwright: Add an import for our `NotFound` error:
```
const { NotFound } = require('..')
```

services/keybase/keybase-profile.js
```diff
@@ -0,0 +1,21 @@
...
+    // See https://keybase.io/docs/api/1.0/call/user/lookup.
+    const url = `https://keybase.io/_/api/${apiVersion}/user/lookup.json`
+
+    return this._requestJson({
+      url,
+      schema,
+      options,
+    })
+  }
+}
```
calebcartwright: Add a `transform` function to the base class to encapsulate the invalid/unknown username checks:
```js
  transform({ data }) {
    if (data.status.code !== 0) {
      throw new NotFound({ prettyMessage: 'invalid username' })
    }

    if (data.them.length === 0 || !data.them[0]) {
      throw new NotFound({ prettyMessage: 'profile not found' })
    }

    return { user: data.them[0] }
  }
```

services/keybase/keybase-pgp.service.js
```diff
@@ -0,0 +1,106 @@
...
+    if (data.them.length === 0 || !data.them[0]) {
+      return {
+        message: 'profile not found',
+        color: 'critical',
+      }
+    }
+
+    const primaryKey = data.them[0].public_keys.primary
```
calebcartwright: with the above change to leverage the base `transform` function, this line can be changed to:
```js
const { user } = this.transform({ data })
const primaryKey = user.public_keys.primary
```

calebcartwright:
This is looking great! I've got one final ask around refactoring the username checks into a single function in the base class. I added details inline, but basically let's add a `transform` function to the base class:

```js
  transform({ data }) {
    if (data.status.code !== 0) {
      throw new NotFound({ prettyMessage: 'invalid username' })
    }

    if (data.them.length === 0 || !data.them[0]) {
      throw new NotFound({ prettyMessage: 'profile not found' })
    }

    return { user: data.them[0] }
  }
```

and then update the various service classes accordingly (remove the `if` blocks checking the username/profile and call the `transform` function). for example for btc:

```js
    const { user } = this.transform({ data })
    const bitcoinAddresses = user.cryptocurrency_addresses.bitcoin
```

services/keybase/keybase-pgp.service.js
SkypLabs: Good idea!

services/keybase/keybase-btc.service.js
SkypLabs: Make sense. Now you say it, I don't think we will one day have new Keybase badges in another category.

SkypLabs:
Looks way better now. Thanks!

SkypLabs:
Awesome! Thanks for your detailed review.

Issue #2862, opened 2019-01-24
Badge request: JSDelivr


douglasg14b:
:clipboard: **Description**

Add a badge for the JSDeliver Service for versioning.

The version information for a package can be retrieved from their API https://github.com/jsdelivr/data.jsdelivr.com#list-package-versions

calebcartwright:
Thanks for the suggestion! We've got a great [tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) for writing new services if anyone would like to work on this!

paulmelnikow:
They also have [download stats](https://www.jsdelivr.com/blog/jsdelivr-reloaded-2017/)!

This would be a high impact badge to implement.

PR #2970, opened 2019-02-10
[jsDelivr] badges

Prouser123:
Requested in #2862.

Currently NPM + GitHub hit badges.

I don't personally feel that a versioning badge is necessary (jsDelivr being a CDN) but feel free to prove me wrong and I'll happily add it in.


services/jsdelivr/jsdelivr-hits-github.service.js
```diff
@@ -0,0 +1,78 @@
...
+    const { total } = await this.fetch({ period, user, repo })
+    return this.constructor.render({ period, hits: total })
+  }
+
+  async fetch({ period, user, repo }) {
+    const url = `https://data.jsdelivr.com/v1/package/gh/${user}/${repo}/stats/date/${
+      periodMap[period]
+    }`
+
+    return this._requestJson({
```
MartinKolarik: Please add a `User-Agent` header [as recommended](https://github.com/jsdelivr/data.jsdelivr.com#let-us-know-how-you-use-this-api).

services/jsdelivr/jsdelivr-base.js
```diff
@@ -0,0 +1,40 @@
...
+  hw: 'week',
+  hm: 'month',
+  hy: 'year',
+}
+
+class BaseJsDelivrService extends BaseJsonService {
+  static render({ period, hits }) {
+    return {
+      message: `${metric(hits)} hits/${periodMap[period]}`,
+      //color: downloadCount(hits),
```
chris48s: Can we get rid of this line if we're not using it.

services/jsdelivr/jsdelivr-base.js
```diff
@@ -0,0 +1,40 @@
...
+    }
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'jsDelivr',
+      color: 'orange',
```
chris48s: If we want to just make everything one colour, can we go with blue (which we use for "information" as opposed to good/bad/etc).

services/jsdelivr/jsdelivr-base.js
```diff
@@ -0,0 +1,40 @@
...
+  static render({ period, hits }) {
+    return {
+      message: `${metric(hits)} hits/${periodMap[period]}`,
+      //color: downloadCount(hits),
+    }
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'jsDelivr',
```
chris48s: The left-hand side of the badge should be all lowercase: https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines (even if the service usually styles their name in mixed-case)

services/jsdelivr/jsdelivr-base.js
Prouser123: I used orange as that was the color scheme used on their website and their badges.

PR #2972, opened 2019-02-10
refactor [amo] service

chris48s:


services/amo/amo-users.service.js
```diff
@@ -0,0 +1,46 @@
...
+  static get defaultBadgeData() {
+    return {
+      label: 'users',
+    }
+  }
+
+  static render({ users }) {
+    return {
+      message: metric(users),
+      color: 'brightgreen',
```
calebcartwright: I know this is consistent with the legacy service, but if the color will always be the same, would informational/blue make more sense than success/brightgreen? Or perhaps a scale?

PR #2977, opened 2019-02-11
refactor [cocoapods] service

chris48s:


services/cocoapods/cocoapods-base.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = class BaseCocoaPodsService extends BaseJsonService {
+  async fetch({ spec }) {
+    return this._requestJson({
+      schema,
+      url: `https://trunk.cocoapods.org/api/v1/pods/${spec}/specs/latest`,
+    })
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'cocoapods' }
```
paulmelnikow: Is this used anywhere? It seems like all the badges override it.

services/cocoapods/cocoapods-platform.service.js
```diff
@@ -19,15 +19,26 @@ module.exports = class CocoapodsPlatform extends LegacyService {
...
+    return {
+      message: platforms,
+      // https://github.com/badges/shields/pull/184
+      color: '#989898',
+    }
+  }
+
+  async handle({ spec }) {
+    const data = await this.fetch({ spec })
+    const platforms = Object.keys(data.platforms).join(' | ')
```
paulmelnikow: Nit: The `.join(' | ')` probably belongs in `render`. Passing `Object.keys(data.platforms)` seems slightly better.

PR #2995, opened 2019-02-14
Add ability to support version badge from azure devops release pipeline

Teesit-Draycir:


knight1219:
Any idea when this is getting released? My org has dozens of builds each at least 4 deployments and knowing what version is in environment is would be super useful.

calebcartwright:
Hi @knight1219, given that this PR was closed immediately after opening, I think it would be better if someone would open a new issue requesting the feature (which I agree would be useful, if it is indeed possible)

PR #2997, opened 2019-02-15
TypeError: s.toLowerCase is not a function

### LINKED ISSUE #2854: TypeError: s.toLowerCase is not a function
https://sentry.io/shields/shields/issues/855518408/?referrer=github_plugin

```
TypeError: s.toLowerCase is not a function
  File "/home/m/shields/gh-badges/lib/color.js", line 24, in isHexColor
    return hexColorRegex.test(s.toLowerCase())
  File "/home/m/shields/gh-badges/lib/color.js", line 32, in normalizeColor
    } else if (isHexColor(color)) {
  File "/home/m/shields/gh-badges/lib/color.js", line 42, in toSvgColor
    const normalized = normalizeColor(color)
  File "/home/m/shields/lib/logos.js", line 65, in getSimpleIcon
    const svgColor = toSvgColor(color)
  File "/home/m/shields/lib/logos.js", line 78, in prepareNamedLogo
    return getShieldsIcon({ name, color }) || getSimpleIcon({ name, color })
...
(6 additional frame(s) were not displayed)

TypeError: s.toLowerCase is not a function
```

RedSparr0w:
This should Fix #2854.

Unless i'm missing something, I believe we don't need to convert `s` to lowercase as the regex is already case insensitive.


paulmelnikow:
Have you been able to reproduce the crash? Makes sense about toLowerCase not being necessary, though the problem occurs when the parameter is not a string. I don’t want to replace one crash with a different one.

RedSparr0w:
I beleive the error occurs when you do something like this:
`https://img.shields.io/:Heroku-Logo-7673C0.svg?logo=heroku&logoColor=red&logoColor=blue`

Doesn't seem to throw at all with `.test()`
```javascript
/\d+/.test(undefined) // false
/\d+/.test(null) // false
/\d+/.test([]) // false
/\d+/.test({}) // false
/\d+/.test(123) // true
/\d+/.test('123') // true
```

PyvesB:
Could we add a small unit test to prove the issue with non-strings is indeed fixed and not reintroduced by mistake in the future?

PR #3004, opened 2019-02-16
refactor [Packagist] service

Prouser123:


paulmelnikow:
This pull request **introduces 1 alert** when merging cd17b48bcac3c65b16391e6a670e5f1792788e49 into 6bfa9b1b41915ac07a4b6dfa304436a97197d115 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-0a848ed4d95f7785203e08f9088e150c85641968)

**new alerts:**

* 1 for Useless assignment to local variable

---

services/packagist/packagist-downloads.service.js
```diff
@@ -1,123 +1,86 @@
...
+  async handle({ type, user, repo }) {
+    const {
+      package: { downloads },
+    } = await this.fetch({ user, repo })
+
+    switch (type) {
+      case 'dm':
+        return this.constructor.render({
+          downloads: downloads.monthly,
+          ending: '/month',
```
paulmelnikow: Instead of computing `ending` here, could you pass the interval to `render` and handle the suffix there?

services/packagist/packagist-base.js
```diff
@@ -0,0 +1,59 @@
...
+  async fetch({ user, repo, schema = latestVersionSchema }) {
+    const url = `https://packagist.org/packages/${user}/${repo}.json`
+
+    return this._requestJson({
+      schema,
+      url,
+    })
+  }
+
+  static get defaultBadgeData() {
```
paulmelnikow: Can this be moved into the service that needs it? When these parameters cascade into superclasses they are harder to reason about.

services/packagist/packagist-license.service.js
```diff
@@ -1,90 +1,44 @@
...
-                if (isUnstable(version) && !isUnstable(current)) {
-                  version = current
-                } else if (
-                  version.version_normalized < current.version_normalized
-                ) {
-                  version = current
-                }
-              } else {
-                version = current
-              }
```
paulmelnikow: Why was this logic removed?

services/packagist/packagist-php-version.service.js
```diff
@@ -1,77 +1,69 @@
...
+      user,
+      repo,
+      schema: allVersionsSchema,
+    })
+    try {
+      return this.constructor.render({
+        php: allData.package.versions[version].require.php,
+      })
+    } catch (e) {
+      throw new NotFound({ prettyMessage: 'invalid version' })
```
paulmelnikow: What is this `try`/`catch` catching? Could this be done using Joi instead? Let's use Joi for validation whenever we can, and avoid wrapping `render()` in a `try`/`catch`.

paulmelnikow:
Hey @Prouser123, would you have a moment to respond to these comments?

Prouser123:
I'm a little caught up with other (school) work right now, but when I have some free time I'll gladly take a look! 😃

paulmelnikow:
Hi! Do you have a moment to wrap this up? If not, I or one of the other maintainers could do it. 😁

services/packagist/packagist-license.service.js
Prouser123: I made it just use the `dev-master` (latest) version and get the license from there.

Prouser123: bump @paulmelnikow

Prouser123:
Interesting error to note before force-push:
https://circleci.com/gh/badges/shields/46763

calebcartwright:
> Interesting error to note before force-push:
> https://circleci.com/gh/badges/shields/46763

Spurious error, no worries!

services/packagist/packagist-downloads.service.js
```diff
@@ -1,28 +1,58 @@
...
+    } = await this.fetch({ user, repo })
+
+    return this.constructor.render({ downloads, interval })
+  }
+
+  static render({ downloads, interval }) {
+    let amount
+    let ending = ''
+
+    switch (interval) {
```
calebcartwright: This feels a bit like the render function is also performing the transform/data parsing. 

A common approach used with other badges that leverage is intervals is an object map (such as [pypi downloads](https://github.com/badges/shields/blob/master/services/pypi/pypi-downloads.service.js#L19)).

With something like that the render function could simply take in the download number, the responsibility to extract of the correct download number from the response could shift back to the handle function, the examples would simplify, etc.

services/packagist/packagist-php-version.service.js
calebcartwright: Agreed

services/packagist/packagist.tester.js
```diff
@@ -34,7 +34,7 @@ t.create('gets the package version of symfony 2.8')
```
calebcartwright: Could we split this into multiple `*.tester.js` files to match the respective `*.service.js` files?

calebcartwright:
https://shields-staging-pr-3004.herokuapp.com/

services/packagist/packagist-php-version.service.js
Prouser123: Joi will match and return all valid versions, but it will throw a TypeError if you try to use an invalid version. Do you think we should do something like this?

```
if (!allData.package.versions.hasOwnProperty(version)) {
    throw new NotFound({ prettyMessage: 'invalid version' })
}
```

Prouser123: pushed.

paulmelnikow:
@calebcartwright Good to go?

calebcartwright:
@paulmelnikow yup - 🚀

PR #3005, opened 2019-02-16
refactor [maintenance] service

Prouser123:


services/maintenance/maintenance.service.js
```diff
@@ -1,27 +1,56 @@
...
-    return 'other'
-  }
+const { BaseService } = require('..')
 
+module.exports = class Maintenance extends BaseService {
   static get route() {
     return {
       base: 'maintenance',
-      pattern: ':maintained(yes|no)/:year(\\d{4})',
+      pattern: ':maintained/:year(\\d{4})',
```
paulmelnikow: What's the reason for removing `(yes|no)` from the pattern? Is there a reason to accept other values?

services/maintenance/maintenance.service.js
```diff
@@ -1,27 +1,56 @@
...
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class Maintenance extends LegacyService {
-  static get category() {
-    return 'other'
-  }
+const { BaseService } = require('..')
 
+module.exports = class Maintenance extends BaseService {
```
paulmelnikow: Probably this should use `NonMemoryCachingBaseService` as there's nothing being fetched here.

services/maintenance/maintenance.service.js
paulmelnikow: Oh, I see, there's a test that hits `/maintenance/until end of 2019/2019.svg`



Kind of strange that we support that!

PR #3007, opened 2019-02-16
improve docs for simple-icons usage

chris48s:


paulmelnikow:
Could we pick one of these to officially support, maybe the dasherized version, and leave the others as undocumented / not guaranteed to work?

chris48s:
If we have to pick only one, I'd be inclined to pick only the urlencoded format: The way the icons are indexed in the actual simple-icons library is with the spaces, so that's more likely to be intuitive to someone who already knows the library. If you copy names from the simple icons website, its going to have spaces.
A dash is a bit more readable than `%20` in plain text, but its something we've invented, if you see what I mean.

paulmelnikow:
Yea, all makes sense. Agree that the one we invented is a bit nicer.

I guess a couple ideas before we settle this:

1. Did we invent the dasherized slug, or is it already buried within what simpleicons ships? If it is part of what they ship, could we change their website so it shows the slug. Maybe it could even use a copyable button. I feel like other applications using simpleicons would benefit from using and surfacing the standard slug.
2. Would having a searchable picker with all the icons on our badge modal solve the problem? Might be even friendlier than sending people to the simple-icons site.

chris48s:
The version with dashes is nothing to do with the simple-icons library - its added here: https://github.com/badges/shields/blob/6e36e520847b3ca3de3dc299427dab08e17f0f51/lib/load-simple-icons.js#L8

In general, it would be a nice enhancement if "named logo" in the modal was a searchable drop-down. That's a good stretch goal. Adding a sentence to clarify the docs is a useful interim improvement.

paulmelnikow:
Are you open to guiding people to the dasherized version (replace spaces with dashes), with a callout that the picker in the UI is a stretch goal?

PR #3010, opened 2019-02-17
refactor [discourse] service

chris48s:


services/discourse/discourse.service.js
```diff
@@ -1,176 +1,110 @@
...
-      })
-    )
+  static render() {
+    return {
+      message: 'online',
+      color: 'brightgreen',
+    }
+  }
+
+  async handle({ scheme, host }) {
```
paulmelnikow: I guess what’s happening here is that if the fetch fails, we are offline. It’s a little magical; could you add a comment explaining that?

PR #3011, opened 2019-02-17
omit style=flat if default style selected

chris48s:


frontend/components/customizer/query-string-builder.js
```diff
@@ -53,10 +53,10 @@ export default class QueryStringBuilder extends React.Component {
       badgeOptions[name] = defaults[name] || ''
     })
 
-    this.state = { queryParams, badgeOptions }
+    this.state = { queryParams, badgeOptions, defaultStyle }
```
paulmelnikow: Since this is effectively fixed for an instance of the component, it's unnecessary to copy it from `props` to `state`. It also requires passing it around through these function calls. Unless it complicates testing for some reason, what I'd suggest is to change `getQueryString` from a static method to an instance method, and then pull `defaultStyle` out of props within `getQueryString`. That should avoid confusion around why that value is in the state, and also reduce the amount of code needed to copy the value around. Does that make sense?

chris48s: Yeah - that sounds sensible. I'll have a look at re-organising it later or tomorrow

PyvesB:
I had a look at the code in _query-string-builder.js_. Are we really on the right track here?

As far as I can tell, we've already got [a check](https://github.com/badges/shields/blob/21612dd0033962d71f00d8908cc202bbabcd2195/frontend/components/customizer/query-string-builder.js#L83) that ignores the query string parameter if it value is equal to the default.

If I understand correctly, the bug is on the [line just above](https://github.com/badges/shields/blob/21612dd0033962d71f00d8908cc202bbabcd2195/frontend/components/customizer/query-string-builder.js#L82): we're trying to get a property called `defaultValue` from an object which only contains properties `name` and `label`. Therefore, isn't the conditional check always true?

paulmelnikow:
Also… as I thought about this some more, I wondered why `defaultStyle` was being passed down instead of hard coded. I think the reason is the social examples use a different style by default. In that case we need to keep `social` in there.

To do the right thing, I think the check needs to ignore `defaultStyle` and be an explicit comparison to `’flat’` (or to a constant like `shieldsDefaultStyle` defined to `’flat’`).

paulmelnikow:
@PyvesB I’ll dig in further to your comment when I’m on a keyboard (on mobile now).

paulmelnikow:
I tried this PR in the review app and it seems it's broken the query param generation.

> If I understand correctly, the bug is on the [line just above](https://github.com/badges/shields/blob/21612dd0033962d71f00d8908cc202bbabcd2195/frontend/components/customizer/query-string-builder.js#L82): we're trying to get a property called `defaultValue` from an object which only contains properties `name` and `label`. Therefore, isn't the conditional check always true?

Agreed this is the problem. I've added a bunch of comments, and renamed `defaultStyle` to `initialStyle` to try to make this clearer. Will open a new PR.

chris48s:
closing in favour of #3025

PR #3014, opened 2019-02-17
Fall back to grabbing version from "X :: Only" PyPI classifiers

scop:
If e.g. `3 :: Only` alone is specified in PyPI classifiers, use the `3` from it. But keep ignoring `* :: Only` when it occurs with other specified versions.


services/pypi/pypi-pyversions.service.js
```diff
@@ -58,6 +58,15 @@ module.exports = class PypiPythonVersions extends PypiBase {
       packageData,
       /^Programming Language :: Python :: ([\d.]+)$/
     )
+    // If no versions are found yet, check "X :: Only" as a fallback.
+    if (versions.length === 0) {
+      versions.push(
+        ...parseClassifiers(
+          packageData,
+          /^Programming Language :: Python :: (\d+) :: Only$/
```
paulmelnikow: Is something gained by this conditional? Would it be simpler to remove the conditional, and optionally match `:: Only` in a single pass?

services/pypi/pypi.tester.js
```diff
@@ -215,6 +215,14 @@ t.create('python versions (valid, no package version specified)')
     })
   )
 
+t.create('python versions ("Only" and others)')
+  .get('/pyversions/uvloop/0.12.1.json')
+  .expectJSON({ name: 'python', value: '3.5 | 3.6 | 3.7' })
+
+t.create('python versions ("Only" only)')
+  .get('/pyversions/hashpipe/0.9.1.json')
+  .expectJSON({ name: 'python', value: '3' })
```
paulmelnikow: I've a slight preference to extract this functionality to a helper function so it can be tested through unit tests, as we've done with `getLicenses` and `getPackageFormats` in `pypi-helpers.js`.

Package metadata can change, so even if these classifiers are present on these packages today, these tests don't prevent regressions and aren't self-documenting.

services/pypi/pypi-pyversions.service.js
scop: The conditional is in my opinion desirable. For example, given `3 :: Only`, `3.5`, `3.6`, and `3.7`, I think the "correct" output would be `3.5 | 3.6 | 3.7`.  But for `3`, `3.5`, `3.6`, and `3.7`, it would be `3 | 3.5 | 3.6 | 3.7`. So the `X :: Only` version would not be taken to include all `X.*` versions if more specific version info is available.

services/pypi/pypi.tester.js
scop: I tend to agree, but I'm merely following the existing practice established in pypi.tester.js here; bigger structural changes belong in separate commits/PR's in my opinion.

Then again, FWIW, metadata on versioned releases is not directly changeable by authors, because PyPI doesn't allow a single version to be uploaded more than once.

services/pypi/pypi-pyversions.service.js
paulmelnikow: That would be a good example to add a test for. It really clarifies the intent of the code.

services/pypi/pypi.tester.js
paulmelnikow: I see; these are tagged versions.

I'll open a new issue for the refactor and link to this example: https://github.com/badges/shields/pull/3014#discussion_r

services/pypi/pypi-pyversions.service.js
scop: Yep, the `"Only" and others` test case for uvloop 0.12.1 was already included in PR for this reason, it mostly covers it.

PR #3012, opened 2019-02-17
Rename colorB -> color; colorA -> labelColor; test on [endpoint dynamic static]

mbarkhau:
This makes it possible to override the parameters `color` and `labelColor`.

ref #2673

PR #3024, opened 2019-02-18
[QueryStringStatic] service

### LINKED ISSUE #2673: Reimagining the static badge URLs
The static badge URL is well established, yet quirky. It uses a nonstandard escaping practice which only exists in Shields. It's inconvenient to generate programmatically for just that reason.

@RedSparr0w made an effort a while back to create a new one using the query string: #1395. I think it's a nice idea to do that and that we _should_ do that using the schema for the endpoint badge, once that goes live #2495. That will support well the cases where badge URLs need to be generated in code.

Someone on Discord today was asking about a url like https://img.shields.io/badge/PyCalVer-v201901.0019--beta-blue.svg where the version sometimes has a suffix and sometimes doesn't, and thus sometimes but not always requires escaping. I'm checking if that is a programmatic case or a manual case.

We could also consider adding a version that uses the more traditional `/` separator, e.g. `/badge/:label?/:message/:color?` My only concern is about how to include slashes in the label and message. I'm not sure how to solve that. Maybe that case could fall back to either the query-string version or the legacy version.

Obviously we'll need to support the legacy version as well.

mbarkhau:
Ref #2673 

This is WIP (no updated documentation yet) and is based on PR #3012


core/badge-urls/make-badge-url.js
```diff
@@ -71,9 +71,38 @@ function staticBadgeUrl({
   return `${baseUrl}/badge/${path}.${format}${suffix}`
 }
 
+function queryStringStaticBadgeUrl({
+  baseUrl = '',
+  label,
+  message,
+  color = 'lightgray',
```
paulmelnikow: How about defaulting to no color, since that's allowed in the query string badge? In the path-based static badge I think the color needs to be included regardless.

services/static-badge/query-string-static.service.js
```diff
@@ -0,0 +1,27 @@
...
+    return {
+      format: 'static/v([0-9])',
+      capture: ['schemaVersion'],
+      // All but one of the parameters are parsed via coalesceBadge. This
+      // reuses what is the override behaviour for other badges.
+      queryParams: ['message'],
+    }
+  }
+
+  handle(namedParams, queryParams) {
```
paulmelnikow: It would be good to validate the `queryParams` using Joi, as we've done in e.g. the dynamic badge. It will ensure that `message` is a string or numeric string, and not e.g. an array. (same issue as #2854)

services/static-badge/query-string-static.tester.js
```diff
@@ -0,0 +1,69 @@
...
+  .get(
+    '/static/v1.svg?label=&messageall%20one%20color&color=red&style=_shields_test'
+  )
+  .expectJSON({ name: '', value: 'all one color', color: 'red' })
+
+t.create('Not a valid color')
+  .get(
+    '/static/v1.svg?label=label&message=message&color=notacolor&style=_shields_test'
+  )
+  .expectJSON({ name: 'label', value: 'message', color: null })
```
paulmelnikow: These seem like they are testing functionality in `coalesceBadge` + BaseService. There's already a "set color" test at the bottom, so how about removing all of these?

If it seems important to add more tests of that code I would prefer we find a way to do it with a unit test instead of a service test, as the unit tests get run more frequently.

services/static-badge/query-string-static.tester.js
```diff
@@ -0,0 +1,69 @@
...
+
+t.create('Case is preserved')
+  .get('/static/v1.svg?label=LiCeNsE-mIt-blue&style=_shields_test')
+  .expectJSON({ name: 'LiCeNsE', value: 'mIt', color: 'blue' })
+
+t.create('"Shields-encoded" dash')
+  .get(
+    '/static/v1.svg?label=best-license&message=Apache-2.0&color=blue&style=_shields_test'
+  )
+  .expectJSON({ name: 'best-license', value: 'Apache-2.0', color: 'blue' })
```
paulmelnikow: The "Shields-encoded dash" is the unusual escaping used in the url-based static badge. It doesn't seem relevant here and could be removed.

core/badge-urls/make-badge-url.js
mbarkhau: :+1:

services/static-badge/query-string-static.tester.js
mbarkhau: Good point!

paulmelnikow:
I updated the PR title so the service tests will run starting with the next build.

services/static-badge/query-string-static.tester.js
mbarkhau: Fair point.

services/static-badge/query-string-static.service.js
mbarkhau: I added validation as best as I could figure out how. I'm sure there's a better way.

mbarkhau:
So I guess the next step here is to update the shields.io page and add documentation for this. Should it be more like the endpoint badge, on a separate page, or should it be a widget similar to the current static badge, or something else?

Maybe the most simple thing would be to add another output url on the current static widget.

image

paulmelnikow:
> Maybe the most simple thing would be to add another output url on the current static widget.

Yea, seems fine to do the simple thing for now.

That text is in flight in #3032 so perhaps best to get that merged first.

core/badge-urls/make-badge-url.js
```diff
@@ -71,9 +71,38 @@ function staticBadgeUrl({
...
+  color,
+  labelColor,
+  style,
+  namedLogo,
+  logoColor,
+  logoWidth,
+  logoPosition,
+  format = 'svg',
+  schemaVersion = '1',
+}) {
```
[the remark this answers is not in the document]
mbarkhau: Yeah, it's just a footgun at this point, I'll remove the parameter.

mbarkhau:
3aca1ef adds a snippet on the front page, which looks like this

image

I'm not sure how you want to treat the new static urls vs the old ones. With the commit f9b5c20 I've updated the fronted to use the /static/v1.svg urls everywhere, which may be a bit much, idk.

In order to keep the history linear I usually like to do `git rebase` and assuming nobody else is working on this branch I then do `git push --force`. Since this branch has merge conflicts with master again, how would you like that to be resolved? Should I merge master to this branch and give up on the linear history?

paulmelnikow:
> In order to keep the history linear I usually like to do `git rebase` and assuming nobody else is working on this branch I then do `git push --force`. Since this branch has merge conflicts with master again, how would you like that to be resolved? Should I merge master to this branch and give up on the linear history?

We keep PRs to single units of work and use a squash merge workflow, and as a result can optimize the history in PRs for people who are following the PRs. So please merge master into your branch, and we'll squash it down when it's done.

paulmelnikow:
Thanks for carrying this forward!

> [3aca1ef](https://github.com/badges/shields/commit/3aca1ef03ad1e5dd1aa03e2e14eb6846b5c20072) adds a snippet on the front page, which looks like this
> 
> image

Looks good!

I'd like to ship a second version of the URL path version that is a bit friendlier. But this feature is intended for (1) easy programmatic generation and (2) complicated text that is otherwise hard to escape. So, could you put the new text below instead of above?

Also I don't think there's a need to label it beta. We've been thinking about this for a long time and I'm feeling confident in the changes you're making. 😁 

> I'm not sure how you want to treat the new static urls vs the old ones. With the commit [f9b5c20](https://github.com/badges/shields/commit/f9b5c2061f149003b63695b677847d0128dc4e51) I've updated the fronted to use the /static/v1.svg urls everywhere, which may be a bit much, idk.

Again some of the code you're changing is in flight in #3032; so apologies, but I think we need to land that first. I'd lean toward using the old badge everywhere else because the URLs are more readable than query strings. If we get a more streamlined URL-path version down the line we could switch them over.

mbarkhau:
> I'd lean toward using the old badge everywhere else because the URLs are more readable than query strings.

Sounds reasonable. I'll revert f9b5c20.

mbarkhau:
I swaped the ordering of the urls and removed the "(beta)" text. Anything else this needs before a merge?

paulmelnikow:
It looks good! Did you merge in the changes from #3032? That should make it quick to merge this after that lands.

The query param changes landed in #3042; I'll leave a comment about that.

services/static-badge/query-string-static.service.js
```diff
@@ -0,0 +1,39 @@
...
+    if (namedParams.schemaVersion !== '1') {
+      throw new InvalidParameter({ prettyMessage: 'Invalid schemaVersion' })
+    }
+    const result = Joi.validate(
+      { message: queryParams.message },
+      queryParamSchema
+    )
+    if (result.error) {
+      throw new InvalidParameter({ prettyMessage: 'Missing message' })
+    }
```
paulmelnikow: In place of this code, you can now swap in `queryParamSchema` for `queryParams` in the route and it'll be validated for you. `BaseService` should do the right thing. It'll change the error message slightly but otherwise should work the same as what you've got here.

mbarkhau: Done.

mbarkhau:
This PR is now pending on the merge of #3032 since I merged `consolidate-badge-gen`.

paulmelnikow:
Thanks for the updates! Will approve and merge when #3032 is in.

PR #3030, opened 2019-02-18
Improve handling of some edge cases, rm unused fn, and get 100% coverage in generate-image-markup

paulmelnikow:


frontend/lib/generate-image-markup.js
```diff
@@ -33,13 +34,11 @@ export function reStructuredText(badgeUrl, link, title) {
 }
 
 function quoteAsciiDocAttribute(attr) {
-  if (typeof attr === 'string') {
```
calebcartwright: Was this an unnecessary check before, or are now confident that only strings will be passed to this function?

paulmelnikow: Yea, as far as I can tell, the only things passed in here will be strings (either empty or nonempty) that come from `this.state` in the badge customizer.

PR #3032, opened 2019-02-18
Consolidate badge URL generation functions

### LINKED ISSUE #2027: Consolidate URL encoding functions

paulmelnikow:
This consolidates all the badge URL generation functions into a single place, and updates the rest of the calling code to use it. It renames some things and updates labels to bring the names into alignment with the current API.

I think this should resolve #2027. ~~It has a small conflict with #3012.~~


paulmelnikow:
This is blocking #3024. It would be great to merge it today if possible.

calebcartwright:
> This is blocking #3024. It would be great to merge it today if possible.

I will review this evening (US Time) if no one else is able to get to it beforehand

core/badge-urls/make-badge-url.js
```diff
@@ -71,9 +71,43 @@ function staticBadgeUrl({
...
+  baseUrl,
+  datatype,
+  label,
+  dataUrl,
+  query,
+  prefix,
+  suffix,
+  color,
+  style,
+  format = 'svg',
```
calebcartwright: I noticed the previous implementation allowed `queryParams` to be passed in as an object along with `rest`, so were those unused/no longer necessary? (I see `style` can now be passed in directly and is then added to the queryParams object)

frontend/components/customizer/customizer.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import clipboardCopy from 'clipboard-copy'
-import { staticBadgeUrl } from '../../lib/badge-url'
+import { staticBadgeUrl } from '../../../core/badge-urls/make-badge-url'
```
calebcartwright: If I'm tracing through things correctly, it seems like this results in eliminating the call to the function in `frontend/lib/resolve-url.js`. I was wondering if this would eliminate the need for `resolve-url` altogether, but see that it's still used in two components (`footer` and `suggestion-and-search`).

Is `resolve-url` something that may eventually change/be consolidated? Or will it continue live on since it's not badge specific?

frontend/components/usage.js
```diff
@@ -69,7 +69,7 @@ const ColorExamples = ({ baseUrl, colors }) => (
       <Badge
         alt={color}
         key={color}
-        src={staticBadgeUrl(baseUrl, '', color, color)}
+        src={staticBadgeUrl({ baseUrl, label: '', message: color, color })}
```
calebcartwright: 👍 these changes help with readability, especially for someone like me looking at for the first-ish time

core/badge-urls/make-badge-url.js
paulmelnikow: Yea, that's right. We didn't seem to be using that anywhere so I thought it was better to make a more direct API that calls out out the params we are using.

frontend/components/customizer/customizer.js
paulmelnikow: Ugh, I kind of forgot it was still there. May as well take this moment to excavate it. I'd really like to get rid of it.

paulmelnikow: Okay, they're removed.

Suggestions are still broken (see #3033) but I confirmed locally in the debugger that the correct request was being fired and coming back which is what this change affected.

calebcartwright: Lol, I didn't mean to add more work here (was curious when one should use `resolve-url`) but this certainly makes it simpler!

paulmelnikow: Didn't mean to come across as complaining! This is not a pile of cruft I've been happy with and it's a relief to have it gone. I was hoping I'd found the bottom of it 😁

Part of what makes this hard is that there are a lot of object schemas floating around in the frontend which can make for some tricky debugging. (I also had forgotten about #3033.)

calebcartwright: > Didn't mean to come across as complaining!

no worries, didn't come across that way at all 👍

PR #3034, opened 2019-02-18
npm install -> npm ci

paulmelnikow:
[npm ci](https://medium.com/better-things-digital/npm-ci-a-command-you-should-know-d8d67847884c) provides clean installs of packages, using the exact resolution in the lockfile. It's faster, and reliably produces results matching the lockfile.

The only time you really need `npm install` is when you're actually changing the dependencies, which doesn't apply to CI (where Dependabot writes the lockfile), or to any of the cases documented here.

For a while I thought `npm ci` was for **continuous improvement**, as in what you want to use in your CI server, though it seems that it actually means **clean install**.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,6 +2,7 @@ package.json
 package-lock.json
 /__snapshots__
 /.next
+/.cache
 /build
 /public
 /coverage
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.9.4-alpine
+FROM node:8-alpine
 
 RUN apk add --no-cache gettext imagemagick librsvg git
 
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.9.4-alpine
+FROM node:8-alpine
```
paulmelnikow: Bumping this, as `npm ci` requires npm 5.7.

.prettierignore
```diff
@@ -2,6 +2,7 @@ package.json
 package-lock.json
 /__snapshots__
 /.next
+/.cache
```
calebcartwright: 👍 I just noticed prettier was looking at this directory, glad it's getting fixed!

Issue #2719, opened 2019-01-09
GitHub package.json badge croaks on dependencies with github links


AdityaTD:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**
I'm getting `Invalid Response Data` while trying to use the badge for `package.json` version.

:link: **Link to the badge**
https://img.shields.io/github/package-json/v/AdityaTD/PenguBot.svg?style=for-the-badge

:bulb: **Possible Solution**
Not really, I'm not sure why this is happening.

paulmelnikow:
Ah, this is a bug. Shields' schema validation for `package.json` is failing because of the github dependencies. Thanks for reporting!

PyvesB:
Fixed in linked pull request, closing. 😉

PR #3035, opened 2019-02-18
Fix [GithubManifest] and [GithubPackageJson] when static auth is not configured

PyvesB:
This started as investigation work for #2719. I noticed that none of our Manifest or Package.json badges seemed to be working in production, not just @AdityaTD's specific example:
* https://img.shields.io/github/manifest-json/v/RedSparr0w/IndieGala-Helper.svg 
* https://img.shields.io/github/package-json/v/badges/shields.svg 

When we retrieve the file contents via the API, we get a JSON payload as described in the [documentation](https://developer.github.com/v3/repos/contents/#get-contents). The content is embedded in one of the JSON fields, encoded in Base64. Our Joi schema expects to find the `content` and `encoding` fields in the API response.

The issue is that we try to apply similar schema validation and parsing when static auth is not configured, i.e. when we retrieve the file's contents via the "raw" endpoint. In this case, I suggest we just parse the JSON as plaintext.


PyvesB:
The same two badges as above on the review app:
* 
*

PR #3036, opened 2019-02-18
Use drop-down for [jsdelivr] and make badges more consistent with other download badges

paulmelnikow:
- Move to Downloads
- Use brightgreen
- Remove “hits” from the message
- "package name" for better appearance in the UI
- Simplify validation


services/jsdelivr/jsdelivr-base.js
```diff
@@ -18,19 +18,19 @@ const periodMap = {
...
-      message: `${metric(hits)} hits/${periodMap[period]}`,
+      message: `${metric(hits)}/${periodMap[period]}`,
     }
   }
 
   static get defaultBadgeData() {
     return {
       label: 'jsdelivr',
-      color: 'orange',
+      color: 'brightgreen',
```
calebcartwright: Thoughts on using `informational` here if the color will always be the same (I don't think the services are using the dynamic `downloadCount` from `color-formatters.js`?

services/jsdelivr/jsdelivr-hits-github.service.js
```diff
@@ -27,40 +27,13 @@ module.exports = class jsDelivrHitsGitHub extends BaseJsDelivrService {
...
-      },
-      {
-        title: 'jsDelivr Hits (GitHub)',
-        pattern: 'hy/:user/:repo',
-        namedParams: {
-          user: 'jquery',
-          repo: 'jquery',
-        },
-        staticPreview: this.render({ period: 'hy', hits: 95317723 }),
+        staticPreview: this.render({ period: 'hd', hits: 9809876 }),
```
calebcartwright: `period` should probably be `hm` to match named param (or vice versa), I'll update this one

services/jsdelivr/jsdelivr-base.js
paulmelnikow: Yea, probably they should use `downloadCount`, though I think we ought to move `renderDownloadBadge` out of `nuget-helpers` and use it more widely. I didn't really want to tackle that here…

I don't feel strongly, though it seems short of using `downloadCount`, making these brightgreen is a step toward more consistentcy with the other download badges.

paulmelnikow: Eh, it's not that hard to call `downloadCount`. I'll update.

PR #3037, opened 2019-02-18
Combine some badge examples using drop-downs, tweak some titles

paulmelnikow:
This updates several badges to take advantage of the change from #2882. These all seem improved by this change, though I'm happy to revisit if others disagree.

As discussed in that issue, there are several download badges which exist as separate services per time interval, which will need to be refactoried into a single service to work with this change.

I wonder if there are any common elements to these services which could be factored out at the same time. For example, we have a `renderDownloadBadge` function but it's not widely used.


services/github/github-commit-activity.service.js
```diff
@@ -23,16 +23,17 @@ module.exports = class GithubCommitActivity extends LegacyService {
...
-        title: 'GitHub commit activity the past year',
-        pattern: 'y/:user/:repo',
-        namedParams: { user: 'eslint', repo: 'eslint' },
+        title: 'GitHub commit activity',
+        // Override the pattern to omit the deprecated interval "4w".
+        pattern: ':interval(y|m|w)/:user/:repo',
+        namedParams: { interval: 'm', user: 'eslint', repo: 'eslint' },
         staticPreview: {
           label: 'commit activity',
           message: '457/year',
```
calebcartwright: With the new dropdown support, should we make the previews consistent with the selection (for example here using `457/m` to be consistent with the named `interval`) or make them different (as they are here) to indicate multiple options?

services/visual-studio-marketplace/visual-studio-marketplace-azure-devops-installs.service.js
```diff
@@ -35,29 +42,14 @@ module.exports = class VisualStudioMarketplaceAzureDevOpsInstalls extends Visual
   static get examples() {
     return [
       {
-        title:
-          'Visual Studio Marketplace - Azure DevOps Extension (Total Installs)',
-        pattern: 'total/:extensionId',
-        namedParams: { extensionId: 'swellaby.mirror-git-repository' },
+        title: 'Visual Studio Marketplace Installs - Azure DevOps Extension',
+        namedParams: {
+          measure: 'total',
```
calebcartwright: I was initially a little skeptical about using dropdowns here, but I think it's growing on me.

services/mozilla-observatory/mozilla-observatory.service.js
```diff
@@ -63,8 +63,7 @@ module.exports = class MozillaObservatory extends BaseJsonService {
     return [
       {
         title: 'Mozilla HTTP Observatory Grade',
-        pattern: 'grade/:host',
-        namedParams: { host: 'github.com' },
+        namedParams: { which: 'grade', host: 'github.com' },
```
calebcartwright: I'm on the fence about leveraging the dropdowns for this service. My initial thinking when we recently merged this PR was that since the `grade-score` option includes an entirely additional/new data point (the score), that it warrants its own listing on the main page. 

The other services in this PR make sense to me as the dropdown is for a param that's essentially for different filters on the same data set (download period, install targets) or an attribute of the same data entity in the case of the github-issue-detail-service.

Here on the Mozilla Observatory service, it's the same data (the analysis result) but with or without the extra data point (the score)

services/github/github-commit-activity.service.js
paulmelnikow: Oops! I tried to keep them the same. Feel free to fix!

services/mozilla-observatory/mozilla-observatory.service.js
paulmelnikow: A similar set of these are the rating badges, where there's rating + score. I think it's arguable that it's the same badge with a different format option, and that showing both forms doesn't provide a lot of extra information in the listing that wouldn't be discovered by anyone who needs it, after clicking through. Though I don't feel strongly! I'm happy to remove this one for now.

calebcartwright: I guess I was only looking at this service in comparison with the others in the PR. Do you know off hand of some of the other rating + score badges? I can think of a few that are one or the other but not combos off the top of my head

Also I don't feel strongly either 😄 I keep going back and forth internally

services/github/github-commit-activity.service.js
calebcartwright: Done!

services/mozilla-observatory/mozilla-observatory.service.js
paulmelnikow: > Do you know off hand of some of the other (configurable) rating + score badges

Mozilla add-on, VS Code Marketplace, Redmine plugin…

None of them show combined info. But there are two different ways of showing the same metric.

calebcartwright: Yeah this one makes more sense to me if I look at it more as the `short` vs. `long` form of the analysis. 

I've approved as I'm good with dropdowns here.

PR #3041, opened 2019-02-19
Test transformExample directly

paulmelnikow:
These tests in `base.spec.js` are difficult to maintain.


core/base-service/base.spec.js
```diff
@@ -343,59 +331,7 @@ describe('BaseService', function() {
...
-          color: 'lightgrey',
-          label: 'cat',
-          message: 'Hello namedParamA: foo with queryParamA: bar',
-          namedLogo: undefined,
-          style: undefined,
-        },
-        keywords: ['hello'],
-        documentation: undefined,
-      })
+      expect(examples).to.have.lengthOf(1)
```
calebcartwright: I was thinking it might be helpful to have a comment here indicating that the meat of the examples testing can be found over in `transform-example.spec.js`. Thoughts?

paulmelnikow: Yes, that's a good idea.

calebcartwright: Alrighty I've added the comment and approved. Will hold off on merging this though in case you want to tweak the comment text

paulmelnikow: Think that works for now!

PR #3042, opened 2019-02-19
Validate query params in BaseService [npm nodeping mozillaobservatory matrix gitlab f-droid endpoint dynamic bitbucket appveyor]

### LINKED ISSUE #2676: Perform query param validation within BaseService
We've a few badges which validate query params by invoking `_validateQueryParams` on BaseService.

If we provided the route with a Joi schema instead of an array of names, we could validate these _within_ BaseService instead – i.e. prior to invoking `handle()`.

If #2675 hasn't been completed yet, we could use a placeholder `Joi.object().required()` for those services in the meantime.

paulmelnikow:
This is a mid-sized PR that adds query param validation to BaseService and updates most of the services which use query param validation to use it. There are a couple minor tweaks I made along the way.

Fix #2676


services/appveyor/appveyor-tests.tester.js
```diff
@@ -56,14 +57,14 @@ t.create('Test status with custom labels')
 
 t.create('Test status with compact message and custom labels')
   .timeout(10000)
-  .get('/NZSmartie/coap-net-iu0to.json', {
-    qs: {
+  .get(
+    `/NZSmartie/coap-net-iu0to.json?${queryString.stringify({
```
paulmelnikow: The old version produced `?compact_message=`; the new one produces `?compact_message` which is what we are supposed to support.

services/mozilla-observatory/mozilla-observatory.service.js
```diff
@@ -26,6 +26,10 @@ const schema = Joi.object({
     .required(),
 }).required()
 
+const queryParamSchema = Joi.object({
+  publish: Joi.equal(''),
```
paulmelnikow: As above, this is how `?publish` is represented which is what the badge customizer creates for this badge. Have update the documentation to match.

services/nodeping/nodeping-status.service.js
```diff
@@ -38,24 +43,24 @@ class NodePingStatus extends BaseJsonService {
...
+          up_message: 'online',
           up_color: 'blue',
-          down_message: 'Offline',
+          down_message: 'offline',
           down_color: 'lightgrey',
         },
         staticPreview: this.render({
           status: true,
-          upMessage: 'Online',
+          upMessage: 'online',
```
paulmelnikow: Use our usual caps conventions.

core/base-service/route.js
```diff
@@ -67,9 +69,19 @@ function namedParamsForMatch(captureNames = [], match, ServiceClass) {
   return result
 }
 
+function getQueryParamNames({ queryParams = [], queryParamSchema }) {
+  if (queryParamSchema) {
+    const { children, renames = [] } = Joi.describe(queryParamSchema)
```
calebcartwright: I didn't know `Joi.describe` was a thing. Just another chapter in my "today the thing I learned about Joi was..." book

services/appveyor/appveyor-tests.service.js
```diff
@@ -19,16 +20,18 @@ const documentation = `
 </p>
 `
 
+const queryParamSchema = Joi.object({
+  compact_message: Joi.equal(''),
```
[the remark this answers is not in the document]
calebcartwright: Odd indeed 🤔 I've no idea why that's happening, so feel inclined to ask how common an occurrence this would be? Would this only be necessary on Test related badges that had compact (I believe the Azure DevOps Test and Jenkins Test services have the compact message too) or could it happen with other params?

services/dynamic/dynamic-helpers.js
```diff
@@ -20,6 +12,14 @@ const queryParamSchema = Joi.object({
...
+  return {
+    base: `badge/dynamic/${which}`,
+    pattern: '',
+    queryParamSchema,
+  }
+}
+
 module.exports = {
   createRoute,
   queryParamSchema,
```
calebcartwright: With the updates to `createRoute`, does the `queryParamSchema` still need to be exported?

services/mozilla-observatory/mozilla-observatory.service.js
calebcartwright: I'm more confused now around why/when this is necessary. Is it when we're expecting query params representing a boolean value?

calebcartwright: Also if anyone is using this badge and had explicitly added a query param with false (i suppose `?publish=false`) will the param maintain the false value/get dropped entirely in the validation? Just want to make sure that wouldn't somehow get converted to `?publish` and have the reverse behavior

services/dynamic/dynamic-helpers.js
paulmelnikow: Nope! Updated.

services/mozilla-observatory/mozilla-observatory.service.js
paulmelnikow: > I'm more confused now around why/when this is necessary.

It's super confusing. `camp`, `querystring` and basically every other querystring library have different ideas about how to represent these values.

> Is it when we're expecting query params representing a boolean value?

I'm actually not sure whether Scoutcamp passes through the string `'false'` or converts it to a boolean. Since the schema here isn't converting anything, I think it would reject `?publish=false`.

services/appveyor/appveyor-tests.service.js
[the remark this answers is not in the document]
paulmelnikow: Ah, as you surmised elsewhere, it's for any of the boolean params.

core/base-service/base.js
```diff
@@ -174,7 +175,8 @@ module.exports = class BaseService {
 
     let base, format, pattern, queryParams
     try {
-      ;({ base, format, pattern, query: queryParams = [] } = this.route)
+      ;({ base, format, pattern } = this.route)
+      queryParams = getQueryParamNames(this.route)
```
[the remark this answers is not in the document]
calebcartwright: 👍

core/base-service/base.js
```diff
@@ -286,11 +320,12 @@ module.exports = class BaseService {
   static register({ camp, handleRequest, githubApiProvider }, serviceConfig) {
     const { cacheHeaders: cacheHeaderConfig, fetchLimitBytes } = serviceConfig
     const { regex, captureNames } = prepareRoute(this.route)
+    const queryParams = getQueryParamNames(this.route)
```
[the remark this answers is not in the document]
calebcartwright: I forget what the sequence of function calls looks like in here at runtime, but it seems like we're making this same call twice (here and on line 179 in `getDefinition`. Is it possible to make this call to get the query params once, store and then reuse them?

core/base-service/route.js
```diff
@@ -15,9 +15,11 @@ const routeSchema = Joi.object({
     is: Joi.string().required(),
     then: Joi.array().items(Joi.string().required()),
   }),
+  queryParamSchema: Joi.object().schema(),
   queryParams: Joi.array().items(Joi.string().required()),
 })
   .xor('pattern', 'format')
+  .oxor('queryParamSchema', 'queryParams')
```
calebcartwright: Is this just because some services are still specifying query params via the `queryParams` array, or will the `queryParams` continue to live on long term?

core/base-service/base.js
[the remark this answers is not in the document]
paulmelnikow: `getDefinition()` is used at build time and `register()` is run at runtime. (In any case, `register()` is called only once at runtime.)

In the tests we do call both, but without profiling I'm not convinced it's worth adding any complexity to memoize this (it's that thing about premature optimization).

core/base-service/route.js
paulmelnikow: Originally I'd thought this was needed for legacy services, though I now realize that's not true. So it could be removed once all the services are all cleaned up.

core/base-service/route.js
```diff
@@ -67,9 +69,19 @@ function namedParamsForMatch(captureNames = [], match, ServiceClass) {
   return result
 }
 
+function getQueryParamNames({ queryParams = [], queryParamSchema }) {
```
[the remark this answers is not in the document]
calebcartwright: What sort of circumstances warrant using Joi `rename` in the schema for query params?

core/base-service/base.js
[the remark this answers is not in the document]
calebcartwright: Gotcha thanks for the explanation

core/base-service/route.js
calebcartwright: 👍

services/mozilla-observatory/mozilla-observatory.service.js
calebcartwright: I've finished going through all the changes and ready to approve. I'd like run a quick test in my local env on this particular one just to triple check if that's cool with you

paulmelnikow: Yea, of course!

PR #3052, opened 2019-02-19
Require constructors to be called with new

### LINKED ISSUE #3051: throw or throw new?
I noticed whilst debugging  #3035 that we were incorrectly using `throw InvalidResponse`. This results in the following error:
`Uncaught TypeError: Class constructor InvalidResponse cannot be invoked without 'new'`

I didn't fix it as part of that PR as I thought there may be other similar occurrences and was planning on fixing them all in one pass, but it turns out that this is the only problematic call.

However, we are quite inconsistent in the way we throw exceptions, as the standard ones accept both syntaxes (i.e. both `throw Error` and `throw new Error` are valid and should be equivalent according to the [specification](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-error-constructor)). We have 45 occurrences of `throw X` and 93 occurrences of `throw new X` in our current codebase.

Should we:
* enforce the use of `throw new X` using [this ESLint plugin](https://github.com/Trott/eslint-plugin-new-with-error) or similar? We can specify custom exception names which would prevent that one bug from being introduced again.
* modify our custom exceptions to add a factory method so that they can be called without `new` and consequently drop the `new` keyword everywhere?
* not worry too much and just fix that one bug we currently have?
* do something else? After all, as we improve our code coverage we will reduce the risk of these mistakes slipping through.

I'm happy to tackle this once we've reached an agreement. 👍

PyvesB:
Closes #3051.

I removed support for the "newless" LRU constructor, as the rule was alerting on the corresponding test and I believe we weren't using it anywhere else anyway.


core/base-service/base-svg-scraping.js
```diff
@@ -12,7 +12,7 @@ const leadingWhitespace = /(?:\r\n\s*|\r\s*|\n\s*)/g
 class BaseSvgScrapingService extends BaseService {
   static valueFromSvgBadge(svg, valueMatcher = defaultValueMatcher) {
     if (typeof svg !== 'string') {
-      throw TypeError('Parameter should be a string')
+      throw new TypeError('Parameter should be a string')
```
PyvesB: This was not actually a bug, but `TypeError` is not part of [CAPS_ALLOWED](https://github.com/eslint/eslint/blob/b9aabe34311f6189b87c9d8a1aa40f3513fed773/lib/rules/new-cap.js#L16).

Issue #2701, opened 2019-01-08
Auto-alphabetize imports


paulmelnikow:
@Daniel15 suggested at https://github.com/badges/shields/pull/2554#pullrequestreview-190125958:

> In general, does this project have rules about alphabetization? At Facebook we alphabetize `require()` statements and JSX props (the latter enforced using [react/jsx-sort-props](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-sort-props.md) lint rule).

If we can automate this, seems easy enough and will help keep things tidy. I do tend to spend a bit of time putting these in order when I move things around.

Daniel15:
I actually suggested two things: Alphabetize imports, and alphabetize props 😃

PR #3055, opened 2019-02-20
Alphabetize JSX props in the frontend

paulmelnikow:
I'm game to give this a shot.

Ref #2701


frontend/components/common.spec.js
```diff
@@ -37,7 +37,7 @@ describe('Common modules', function() {
 
     it('contains a link to the image', function() {
       const wrapper = render(<common.Badge src="/badge/foo-bar-blue.svg" />)
-      expect(wrapper.html()).to.contain('<img src="/badge/foo-bar-blue.svg"')
```
calebcartwright: Was the inclusion of the `img` tag unnecessary before, or if not, what's the thinking behind removing it (asking mostly for my own education on testing in React)

paulmelnikow: Before I think it was `` and now I  think it's flipped around and turned into `` so the old version won't match anymore. This is now a really confusing assertion; why don't I figure out the full source of the `` tag and use that.

paulmelnikow:
> Is this something we can get prettier to automatically update/enforce for us?

Appears not (prettier/prettier#323), though it is handled here 99% automatically by eslint. I think all but one of these was handled by `npm run lint -- --fix`.

.eslintrc-frontend.yml
```diff
@@ -22,3 +22,4 @@ rules:
   no-console: 'error'
 
   import/extensions: ['error', 'never', { 'json': 'always', 'yml': 'always' }]
+  react/jsx-sort-props: 'error'
```
paulmelnikow: This is the addition of the linter rule.

frontend/components/common.spec.js
calebcartwright: 👍

PR #3056, opened 2019-02-20
Sort imports and requires

### LINKED ISSUE #2701: Auto-alphabetize imports

paulmelnikow:
This will definitely save time, and ensure more uniformity.

It moves the `createServiceTester()` calls to a different place from where I'd like them, though I'm happy to have them checked by the linter.

~~Ref~~ Closes #2701


calebcartwright:
Apologies if we've already got this documented somewhere, but as far the conventions on imports we want to use/follow: 

First: external packages (alphabetized by package name)
Second: relative paths (alphabetized by file name) 

With no new lines in between the external/3rd party modules and our own code. Is that correct, and will this new rule will enforce that?

paulmelnikow:
All there is now is a pattern I've tried to maintain. It's not documented or even something I comment on.

After this change it's enforced by the linter, along these lines:

This comes from [the `import/order` rule](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#importorder-enforce-a-convention-in-module-import-order):

```js
// 1. node "builtin" modules
import fs from 'fs';
import path from 'path';
// 2. "external" modules
import _ from 'lodash';
import chalk from 'chalk';
// 3. "internal" modules
// (if you have configured your path or webpack to handle your internal paths differently)
import foo from 'src/foo';
// 4. modules from a "parent" directory
import foo from '../foo';
import qux from '../../foo/qux';
// 5. "sibling" modules from the same or a sibling's directory
import bar from './bar';
import baz from './bar/baz';
// 6. "index" of the current directory
import main from './';
```

That's ESM syntax, but you get the idea.

There is a better module that handles ESM beautifully [eslint-plugin-simple-import-sort](https://github.com/lydell/eslint-plugin-simple-import-sort) and fully autofixable, but it doesn't work on `require`.

calebcartwright:
Oops I forgot the builtins but yeah that's the order I follow (or at least attempt to follow) in my own projects.

👍

core/service-test-runner/service-tester.js
```diff
@@ -1,10 +1,11 @@
 'use strict'
 
 const emojic = require('emojic')
+const trace = require('../base-service/trace')
 const frisby = require('./icedfrisby-no-nock')(
+  // eslint-disable-next-line import/order
   require('icedfrisby-nock')(require('icedfrisby'))
```
calebcartwright: Interesting. I figured eslint would've recognized this automatically and not barked

frontend/components/category-headings.spec.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
 import { shallow } from 'enzyme'
 import { expect } from 'chai'
-import { CategoryHeading, CategoryHeadings } from './category-headings'
 import { H3 } from './common'
+import { CategoryHeading, CategoryHeadings } from './category-headings'
```
calebcartwright: I would've expected the previous order (`./cat..` before `./com...`) but guess I should re-read the rule (I assume eslint made all these changes)

core/service-test-runner/service-tester.js
paulmelnikow: Yea, the rule probably could be improved. IMO if there are multiple requires on a line it should only consider the outermost one. 🤷‍♂️

frontend/components/category-headings.spec.js
paulmelnikow: I think this rule doesn't alphabetize, unfortunately.

calebcartwright: ahhh that makes so much more sense. i was seriously questioning my mental sorting abilities

paulmelnikow:
Me too. I asked at lydell/eslint-plugin-simple-import-sort#6 but we'll just have to wait until we can use `import` or someone comes along to make this for `require`.

Daniel15:
Nice work! This is great.

PR #3057, opened 2019-02-20
Move flip cache debug service into [debug]

### LINKED ISSUE #2636: Move `flip` cache debug service into the new debug service
See https://github.com/badges/shields/pull/2631#pullrequestreview-189535208.

paulmelnikow:
Close #2636


services/debug/debug.service.js
```diff
@@ -35,6 +36,22 @@ module.exports = class Debug extends NonMemoryCachingBaseService {
           label: 'start time',
           message: new Date(serverStartTime).toUTCString(),
         }
+      // For production cache debugging.
+      case 'flip':
```
calebcartwright: I didn't even know these existed! I assume based on the inline context that these are solely for our usage/consumption, and not something an end user would have a need to consume, nor would this service by referenced by a contributor?

paulmelnikow: Yup, exactly.

calebcartwright: In that case it LGTM. I was going to ask about the lack of a `render` function but it makes more sense now given the purpose 👍

Issue #2068, opened 2018-09-10
Monitoring with Prometheus


platan:
I would like to propose a monitoring solution for Shields based on [Prometheus](https://prometheus.io/) and [Grafana](https://grafana.com/). 

Prometheus is an open-source monitoring and alerting system ([overview](https://prometheus.io/docs/introduction/overview/)). [This page](https://prometheus.io/docs/introduction/comparison/) compares Promethues with alternatives. 

How Prometheus works? An application exposes it's metrics as a plain text via HTTP. Then Prometheus periodically pulls these metrics and allows to display graphs based on these metrics. So it works different comparing to Graphite, where application push data to Graphite. 

Grafana can be configured to use Prometheus as a data source for graphs. 

https://github.com/siimon/prom-client is a good client for Node.js we can use. It collects [recommended default metrics](https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors) and can be used to prepare custom metrics. 

What do we have now?
- Shields version with default Prometheus metrics (https://github.com/badges/shields/pull/2069)
- Two instances of Shields with metrics https://shields-metrics1.now.sh/metrics, https://shields-metrics2.now.sh/metrics
- A Prometheus instance pulling metrics from these 2 Shields instances
- A Grafana instance with a dashboard with default metrics https://metrics.shields.platan.space available for all (address and availability can change). I've prepared this dashboard myself and there is probably a place for improvements in it. 

What do we have to do?
- discuss this idea!
- decide whether we want `/metrics` resource be available for all or limited to some IP/users
- add custom metrics (I've already started with some basic custom metrics)
- decide where Prometheus/Grafana should be hosted (it's ok for me to host it at  https://metrics.shields.platan.space for some time)
- decide who should access to Grafana (available roles: Viewer, Editor, Admin)
- decide whether we want to use Prometheus/Grafana as an alerting system
- document how to set Prometheus/Grafana instance from scratch

\* I'm not a monitoring systems expert. I just have some experience with Grafana, Graphite and Prometheus :-). 

![screencapture-metrics-shields-platan-space-d-g_1b7zhik-prom-client-default-metrics-2018-09-10-21_44_00](https://user-images.githubusercontent.com/1526000/45320358-b55c6380-b542-11e8-95d0-62b9aa1eec2b.png)

chris48s:
[...]

PR #3074, opened 2019-02-22
Refactor [Codecov]

calebcartwright:
This is ready for review now 

Couple notes (more detailed versions below):
- Token (for private repos) moved from route param to query param via new redirector functionality
- The coverage value is now rounded to the nearest whole number to be consistent with our other coverage badges (the legacy service always rounded down before, as does the native Codecov badge)

~~Still a work in progress, but opening this as a draft PR to start a dialog on a few items~~

- **the `token` parameter**
  The legacy service supported coverage badges for private repos via a user-supplied token in a route param. Typical/public repo pattern was `codecov/c/:vcs/:user/:repo`, for private repos the user adds `/token/:token` before the vcs param: `codecov/c/token/:token/:vcs/:user/:repo`. I was thinking it would be easier if the token was supplied via a query param (`codecov/c/:vcs/:user/:repo?token=`) vs. trying to keep the optional token in the route and having to use`format/capture`. I figured I could then create a redirect service for the token route to maintain backwards compatibility. Thoughts?
- **codecov api**
  The legacy service was using Codecov's native badge (https://codecov.io/gh/codecov/example-python/graphs/badge.svg) with plain text (https://codecov.io/gh/codecov/example-python/graphs/badge.txt). Codecov has a documented REST API for the data we need (https://docs.codecov.io/reference#section-get-a-single-repository) that returns a friendly JSON object, so I've switched to that. Any concerns there?
- **coverage rounding**
     The Codecov UI displays the exact coverage value, but their badges display coverage rounded down to the integer value (90.98% coverage shows as 90%), and that integer value is what our legacy service implementation saw and rendered. The Codecov API returns the precise coverage value (i.e. the 90.98%) so we've now got the option to either continue the Math.floor approach and be consistent with the native Codecov badges, or we could apply the same rounding that we do on other Shields coverage badges (90.98% would show as 91%) *I went with the latter in the refactored service, but let me know if anyone thinks we should switch back to rounding down*

~~Note: I've intentionally left the legacy service content in place (renamed) while this PR is in draft mode for easier side-by-side comparisons, but of course will remove it~~


calebcartwright:
I'm having a little difficulty with the redirect around inserting the token query param _after_ the extension. 

I.e. we want to redirect the old path:
`/codecov/c/token/abc123def456/gh/codecov/private-example.svg`

to:
`/codecov/c/gh/codecov/private-example.svg?token=abc123def456` 

but right now my redirect is currently being sent to: 
`/codecov/c/gh/codecov/private-example?token=abc123def456.svg` 

Is there a way in the redirect service definition target to specify query params?

paulmelnikow:
It definitely doesn't do that yet! 😁

I'd probably suggest adding a second function like `targetQueryParams`. Or maybe `target` should be renamed to `transformPath`, and `transformQueryParams` added.

I have a couple changes to `redirector.js` in flight right now for some work I'm doing related to #2068. Let me get to a pausing point and put a small bit of that into PR.

calebcartwright:
> I'd probably suggest adding a second function like `targetQueryParams`. Or maybe `target` should be renamed to `transformPath`, and `transformQueryParams` added.

That's exactly what I was thinking too! I had started with renaming `target` to `targetUrl` and added `addTargetQueryParams` but I like what you've described here much better. There's nothing urgent about the Codecov refactor here whatsoever so happy to hold off on this one. 

Let me know if I can be of any help with the `redirector` stuff, otherwise I'll just keep an eye out and come back to this one whenever we're ready 👍

paulmelnikow:
So… that PR got big (#3093). 😝

If you want I could pick out the redirector validation bits, which probably could be reviewed separately, much quicker.

services/codecov/codecov.spec.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
+
+const { expect } = require('chai')
```
calebcartwright: I'm going to look into leveraging `sazerac` here instead

calebcartwright:
> If you want I could pick out the redirector validation bits, which probably could be reviewed separately, much quicker

I'd say don't worry about. Happy to wait til #3093 gets merged

services/codecov/codecov.service.js
```diff
@@ -46,63 +76,58 @@ module.exports = class Codecov extends LegacyService {
           repo: 'example-python',
           branch: 'master',
         },
-        staticPreview: { label: 'coverage', message: '90%', color: 'green' },
+        staticPreview: this.render({ coverage: 90 }),
       },
       {
-        title: 'Codecov private',
-        pattern: 'token/:token/:vcsName/:user/:repo',
+        title: 'Codecov private repository',
```
calebcartwright: This doesn't actually have to be a separate example anymore does it? I believe I can/should add the token query param to the two above examples instead

paulmelnikow: Yup, that makes sense to me. Some of the other ones like this, like the npm custom registry examples, probably should be converted in the same way.

calebcartwright:
I'm going to take a stab at updating the redirector function to support transforming query params

services/codecov/codecov.service.js
```diff
@@ -1,108 +1,139 @@
...
+  }
+
   static get route() {
     return {
       base: 'codecov/c',
-      pattern: '',
+      // https://docs.codecov.io/docs#section-common-questions
+      // Github, BitBucket, and GitLab are the only supported options (long or short form)
+      pattern:
+        ':vcsName(github|gh|bitbucket|bb|gl|gitlab)/:user/:repo/:branch*',
```
calebcartwright: I was torn on whether to include both the long and shorthand value for the `vcsName`. The legacy service regex would let either through, so I felt that we'd need to support both long/short forms to avoid breaking any existing users. 

If we prefer one form vs. the other, then I could probably set up a few redirectors to maintain backwards compatibility with the other form. However, I elected to stick with the broader route param support here since there's already a redirector transforming the legacy token route parts to a query param.

Let me know what folks think

paulmelnikow: I'm inclined to standardize on the longer forms at some point, though I don't think it's important to create more redirectors right now.

services/codecov/codecov.service.js
```diff
@@ -1,108 +1,139 @@
...
+  token: Joi.string(),
+}).required()
+
+const documentation = `
+  <p>
+    You may specify a Codecov token to get coverage for a private repository.
+  </p>
+  <p>
+    See the <a href="https://docs.codecov.io/reference#authorization">Codecov Docs</a>
+    for more information about creating a token.
```
paulmelnikow: As an aside, do you know if these are read-only / status-only tokens? I hate to remove wanted functionality, though at some point we really should lock down the remaining endpoints that accept read-write tokens. For the purpose of this PR preserving the behavior is 👍

services/codecov/codecov.service.js
calebcartwright: 👍

services/codecov/codecov.spec.js
```diff
@@ -0,0 +1,20 @@
...
+const { test, given } = require('sazerac')
+const Codecov = require('./codecov.service')
+
+const unknownCoverage = { coverage: 'unknown' }
+
+describe('Codecov', function() {
+  test(Codecov.prototype.transform, () => {
+    given({ json: {} }).expect(unknownCoverage)
+    given({ json: { commit: {} } }).expect(unknownCoverage)
+  })
```
paulmelnikow: `sazerac` has a cool thing called `forCases` you can use here:

```js
test(Codecov.prototype.transform, () => {
  forCases([
    given({ json: {} }),
    given({ json: { commit: {} } }),
  ]).expect({ coverage: 'unknown' })
}
```

services/codecov/codecov.tester.js
```diff
@@ -1,23 +1,59 @@
...
+  .get('/gh/codecov/private-example-python.json')
+  .intercept(nock =>
+    nock('https://codecov.io/api')
+      .get('/gh/codecov/private-example-python')
+      .reply(401)
+  )
+  .expectBadge({
+    label: 'coverage',
+    message: 'not authorized to access repository',
+  })
```
paulmelnikow: This test is arguably helpful to ensure that things are wired up correctly, though it feels like it's mostly a test of `errorMessages`.

Could it use a real private project instead?

calebcartwright: Fair point! I don't actually know of any private projects that have Codecov enabled, but if anyone has such a repo and is willing to share I'd be happy to plug it in.

services/codecov/codecov.spec.js
calebcartwright: 👍

services/codecov/codecov.service.js
calebcartwright: I don't know, and it wasn't clear to me from the docs whether there were different permission scopes. My only experience with Codecov tokens has been repository-specific tokens that have permissions to publish coverage results

services/codecov/codecov-redirect.service.js
```diff
@@ -0,0 +1,18 @@
...
+    category: 'coverage',
+    route: {
+      base: 'codecov/c',
+      pattern:
+        'token/:token/:vcsName(github|gh|bitbucket|bb|gl|gitlab)/:user/:repo/:branch*',
+    },
+    transformPath: ({ vcsName, user, repo, branch }) =>
+      `/codecov/c/${vcsName}/${user}/${repo}${branch ? `/${branch}` : ''}`,
+    transformQueryParams: ({ token }) => ({ token }),
+    dateAdded: new Date('2019-02-23'),
```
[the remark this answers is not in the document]
calebcartwright: 😄 Good catch. I plugged that in before I realized the need for the redirector to be able to transform query params

services/codecov/codecov-redirect.service.js
```diff
@@ -0,0 +1,18 @@
...
+module.exports = [
+  redirector({
+    category: 'coverage',
+    route: {
+      base: 'codecov/c',
+      pattern:
+        'token/:token/:vcsName(github|gh|bitbucket|bb|gl|gitlab)/:user/:repo/:branch*',
+    },
+    transformPath: ({ vcsName, user, repo, branch }) =>
+      `/codecov/c/${vcsName}/${user}/${repo}${branch ? `/${branch}` : ''}`,
```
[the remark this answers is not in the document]
calebcartwright: Long form?

services/codecov/codecov.service.js
```diff
@@ -1,108 +1,139 @@
...
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class Codecov extends LegacyService {
+// https://docs.codecov.io/reference#totals
+// A new repository that's been added but never had any coverage reports
+// uploaded will not have a `commit` object in the response and some times
+// the `totals` object can also be missing for the latest commit.
+// Keeping these as optional in the schema and handling it in the transform
+// function to maintain consistent badge behavior with the legacy service implementation.
```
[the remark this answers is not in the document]
calebcartwright: Working on updating this. Initially this was another one of those "because that's the way it was done before in the legacy service" things 😄 

However, I think the underlying scenarios that the legacy service implementation was trying to support was valid (non error) user cases when the coverage was not yet known. 

I can try to dig into it a bit more

services/codecov/codecov.tester.js
paulmelnikow: I don't know one… Maybe we could just leave a comment.

services/codecov/codecov.service.js
calebcartwright: I don't think their anchors are working, but:

https://codecov.io/site/security#how-does-codecov-authenticate-access-to-a-repository
https://codecov.io/site/security#what-scope-is-used-when-signing-up-using-github 

TL;DR if I'm reading correctly (at least for GH), scope carries through to Codecov perms, based on the token-owners permissions on the repo. So a user with read-only access to a private GH repo would create a token that would only be able to be used to see/read Codecov coverage info for that repo. A user with read/write access to that private GH repo would create a token that could both see/read coverage info, and publish/upload coverage results for that repo in Codecov.

services/codecov/codecov.tester.js
calebcartwright: Works for me!

services/codecov/codecov.service.js
paulmelnikow: Sounds like it is _possible_ that folks might create separate account that enable this to be done securely, though it's also possible they won't have…

paulmelnikow:
@calebcartwright Let me know if you want anything else from me before merging!

calebcartwright:
@paulmelnikow sorry I should've left a comment. I want to spend a bit more time challenging the schema in relation to the API response as I thought of a couple other scenarios. Will also add a couple of the code comments we've discussed in this thread

Will give you a shout when I'm done

calebcartwright:
Alrighty, I created some sample apps and ran through some various scenarios and concluded the schema is best as-is 👍 

@paulmelnikow - I'm all set here now

PR #3083, opened 2019-02-23
fix [amo] badges

### LINKED ISSUE #3079: Mozilla Addon API URL Broken
I noticed the Firefox Addon badges were broken on [my project's README.md](https://github.com/mitchmoser/sputnik/blob/master/README.md) so I did some digging.

On line `6083` of [` shields/server.js`](https://github.com/badges/shields/blob/c1a8166af26d79ec2b966a56b0d0b55920df6b8d/server.js#L6083) I noticed:
```
 var url = 'https://services.addons.mozilla.org/api/1.5/addon/' + addonId;
```

According to [Mozilla Add-ons Manager API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/AMO/Add-ons_manager_API):

> This API is deprecated and we do not recommend projects use this API

It looks like all of the examples included on that page are broken as well.

There were no other posts tagged `AMO API` on the [MDN webdocs](https://developer.mozilla.org/en-US/docs/tag/AMO%20API)

Is there another solution that would make these Mozilla badges functional again?

chris48s:
This switches to the new mozilla add-ons API.

Its not a completely like-for-like replacement. I was able to replicate the version, users and rating badges but we no longer have total download stats. We've got weekly downloads instead, so I've deprecated `/d` and replaced it with a new `/dw`

closes #3079


paulmelnikow:
What do you think about redirecting the download badge to the weekly download badge instead of deprecating it? We could leave some documentation behind which clarifies while we're doing that, and I imagine people won't really mind. That way, most people won't need to immediately change their badges.

services/amo/amo-base.js
```diff
@@ -1,25 +1,27 @@
...
     version: Joi.string().required(),
   }).required(),
+  ratings: Joi.object({
+    average: Joi.number().required(),
+  }).required(),
+  weekly_downloads: nonNegativeInteger,
 }).required()
 
-class BaseAmoService extends BaseXmlService {
+class BaseAmoService extends BaseJsonService {
```
paulmelnikow: Yay json!

paulmelnikow:
I'd like to show it in the docs on the website. I'm happy to make the change. I'll push a commit in a second.

paulmelnikow:
Feel free to tweak or merge.

PR #3090, opened 2019-02-23
Rename maxAge to cacheSeconds

### LINKED ISSUE #3069: Harmonize cache param name: maxAge -> cacheSeconds
The cache control parameter has long been `?maxAge`. I'd like to change this to `?cacheSeconds` for consistency with the Endpoint badge.

The reason I used "cache seconds" on the endpoint badge is that it's more colloquial and understandable. We also set a few cache directives, not just `Cache-Control: max-age` so I think it's better not to use that name.

This could look similar to the change @mbarkhau just implemented for the color params: #3012

paulmelnikow:
Close #3069

### DIFF
--- a/core/base-service/cache-headers.js
+++ b/core/base-service/cache-headers.js
@@ -7,23 +7,25 @@ const coalesce = require('./coalesce')
 const serverStartTimeGMTString = new Date().toGMTString()
 const serverStartTimestamp = Date.now()
 
+const isOptionalNonNegativeInteger = Joi.number()
+  .integer()
+  .min(0)
+
 const queryParamSchema = Joi.object({
-  // Not using nonNegativeInteger because it's not required.
-  maxAge: Joi.number()
-    .integer()
-    .min(0),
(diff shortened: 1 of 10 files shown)


core/base-service/cache-headers.js
```diff
@@ -7,23 +7,25 @@ const coalesce = require('./coalesce')
...
-      queryParams,
-      queryParamSchema,
-      { allowUnknown: true }
-    )
-    return overrideCacheLength
+    const {
+      cacheSeconds: overrideCacheLength,
+      maxAge: legacyOverrideCacheLength,
+    } = Joi.attempt(queryParams, queryParamSchema)
+    return coalesce(overrideCacheLength, legacyOverrideCacheLength)
```
calebcartwright: This is how this avoids breaking existing usage of `?maxAge` correct?

paulmelnikow: Yes, exactly.

PR #766, opened 2016-08-16
LuaRocks module version badge

un-def:
luarocks/v/USER/MODULE.FORMAT
or
luarocks/v/USER/MODULE/VERSION.FORMAT

PR #3094, opened 2019-02-24
Refactor [luarocks]

paulmelnikow:


paulmelnikow:
> To my eye there's some non-trivial, domain specific content we have here (the version calcs, the version prefixes used for coloring in `render`, etc.). It's all well tested, but if there's any supporting docs/content we could reference with links in comments I think that'd be helpful for future observers

Yea, it's true. That code came in as part of a year-plus-long saga: #766

Hi @un-def 👋, would you happen to have any references we could link to?

un-def:
@paulmelnikow 

### Version parser
Version parser has been inspired by `luarocks` CLI util:
https://github.com/luarocks/luarocks/blob/405ee29ba8444d97646f62e72effeaff2bfe3f79/src/luarocks/search.lua#L182
https://github.com/luarocks/luarocks/blob/405ee29ba8444d97646f62e72effeaff2bfe3f79/src/luarocks/core/vers.lua#L83

Some specific things about LuaRocks package versioning:

1. A version string consists of two parts: the package version (version of the package _code_) and the revision of `rockspec` file https://github.com/luarocks/luarocks/wiki/Rockspec-format
    > **version** (string, mandatory field) - the version of the package, plus a suffix indicating the revision of the rockspec. Example: "2.0.1-1"
2. The package version (the first part of the LuaRocks version, the part before dash) can be anything, SemVer is not mandatory and is not used very often. Non-digit versions are not rare: `scm`, `dev`, `alpha`, etc.
3. `scm` (or `cvs`) is used to indicate dev version built from VCS (e.g., `"scm-1"`). `dev` version string is also used sometimes, and I don't what the difference.
4. LuaRocks API does not tell you which version is the “latest”, and the package maintainer cannot mark a specific version as the latest. You should figure out it yourself.

### Coloring

The badge colors are following the heuristic rule where `scm < dev < stable` (e.g., `scm-1` < `dev-1` < `0.1.0-1`). It is not something standardized.

paulmelnikow:
Thank you!!

PR #3095, opened 2019-02-24
Rewrite [GitHubLicense]

paulmelnikow:


services/github/github-license.tester.js
```diff
@@ -68,23 +70,5 @@ t.create('License for unknown repo')
   .expectJSON({
     name: 'license',
     value: 'repo not found',
-    color: 'lightgrey',
-  })
-
-t.create('License - API rate limit exceeded')
```
paulmelnikow: It's not completely clear what should happen with this test and/or functioonality. The mock response is what you get when you don't have an API token, though "access denied" and "rate limit exceeded" are different things. I believe in practice our GitHub auth code will throw a different error when there is not enough rate limit. In general it (probably?) wouldn't fail out in a request.

We have a lot of contributed quota, so I don't see this being a big concern.

PR #3096, opened 2019-02-24
First adoption of React hooks

paulmelnikow:
I did this as a warm-up to using [React hooks](https://reactjs.org/docs/hooks-intro.html), which provide a better way to accomplish stateful things and side effects using functional components. This allows all components to be written in the same style, improves testability, facilitates code reuse, etc.

There's [a intro here](https://reactjs.org/docs/hooks-intro.html) which links to [Dan's talk at React Conf](https://www.youtube.com/watch?time_continue=3599&v=dpw9EHDh2bM) which does a really good job of explaining why hooks are a good way to write components. He describes hooks as being the electrons and neutrons to components which are atoms. Low-level functionality which was always there in React, though not as accessibly or visibly.

This adds a lint rule that enforces "the rule of hooks" which says they have to be declared at the top level in the functional component.

I don't think this changeset does a fabulous job of showing off the improvements hooks allows, though I think it is still a good direction for this code.


calebcartwright:
These look pretty helpful to my untrained-React eye. I don't have any objections to anything in here, but figure I'll defer to the more React-fluent folks

PyvesB:
Hmm, I don't think we have many React-fluent folks in the team. 😄

paulmelnikow:
If no one objects to these changes, let's push them through.

As an aside, I think recruiting more front-end devs is a good idea for this project. I think continuing to get the app in good shape to be worked on by experienced front-end devs and thus making it fun is the best way to make it an attractive project for them. Also I think pushing through a redesign will help; that's been on my mind as a major project of the next few months. We might add TypeScript; though it's a bit of extra tooling it makes the experience faster and more pleasant and I think will encourage more people to contribute.

PR #3097, opened 2019-02-24
New expectBadge helper, affects [Wercker]

PyvesB:
This is a start on #2687.

* implemented a simple `expectBadge` helper in _icedfrisby-no-nock.js_. I chose to add it to this existing file rather than create a new one as we've already got quite a few nested factories [here](https://github.com/badges/shields/blob/c7d707363c0bcebade7284fc973b1f5d2dd74d5c/core/service-test-runner/service-tester.js#L5).
* renamed _icedfrisby-no-nock_ to _icedfrisby-shields_ as it no longer only contains functionality related to mocking requests.
* migrated Wercker to use the new helper.
* updated the service test documentation.

 Once we agree on this first step, it should be much easier to migrate other services in one go. 👍


PyvesB:
@chris48s thanks for reviewing! To be honest, I found that a little strange as well, hence the slight confusion in #2687. But hopefully we'll be able to tackle everything in a short timeframe, so it shouldn't have too much of an impact on new contributors.

Will take care of converting all the examples in the coming days.

PR #3114, opened 2019-02-27
refactor [chromewebstore] service

chris48s:


services/chrome-web-store/chrome-web-store-price.tester.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const Joi = require('joi')
+const { ServiceTester } = require('../tester')
+
+const t = (module.exports = new ServiceTester({
```
paulmelnikow: Could these use `createServiceTester()`?

services/chrome-web-store/chrome-web-store-base.js
```diff
@@ -0,0 +1,15 @@
...
+const { BaseService } = require('..')
+const chromeWebStore = require('chrome-web-store-item-property')
+const { checkErrorResponse } = require('../../lib/error-helper')
+
+module.exports = class BaseChromeWebStoreService extends BaseService {
+  async fetch({ storeId }) {
+    try {
+      return await chromeWebStore(storeId)
+    } catch (e) {
+      return checkErrorResponse.asPromise({})({ buffer: '', res: e })
```
paulmelnikow: Other than here we only use `checkErrorResponse.asPromise` in one place and I was thinking we could inline it. This is actually using it to re-render a chrome web store error object, yea? It seems a little opaque though fine for now.

Could you keep the `networkOff()` connection test since effectively we're not using `BaseService._request`?

chris48s: Yep. The `HTTPError` from `chrome-web-store-item-property` has a `statusCode` property, which is the only thing we're looking at inside `checkErrorResponse` so it seemed reasonable to use that helper here given the object "quacks like a duck", as it were. I've added a comment here to explain it.

services/chrome-web-store/chrome-web-store-price.tester.js
chris48s: Yeah good question. I thought Version and Price should be able to use that because both of those only export a single integration, but if I switch Version and Price to use `const t = (module.exports = require('../tester').createServiceTester())` then if I run `npm run test:services -- --only=chrome-web-store` it only runs the tests for Rating and Downloads (although I can run `npm run test:services -- --only=chromewebstoreversion` and `npm run test:services -- --only=chromewebstoreprice` to individually run the version/price tests).

I figured it is better to retain the property that `npm run test:services -- --only=chrome-web-store` runs the whole test suite so I left them all manually creating a `ServiceTester`. I haven't looked into it in any more detail though. Any thoughts on this?

services/chrome-web-store/chrome-web-store-base.js
chris48s: Just looking at the `networkOff()` case, but I'm now really confused about where we usually catch this. How is this case normally handled?

services/chrome-web-store/chrome-web-store-price.tester.js
paulmelnikow: If you use`{ id: 'ChromeWebStorePrice', pathPrefix: '/chrome-web-store/p' }` I think that should give the result you're looking for. That way, everything aligns with the `createServiceTester()` naming convention and the way to run them all becomes `--only=chromewebstore` or `--only=ChromeWebStore`.

We should look at the ServiceTester interface and try to align it a bit better with BaseService, maybe by changing `pathPrefix` to `base`, removing its leading `/`, not defaulting to the value of `id`, and also renaming `id` to `name`.

services/chrome-web-store/chrome-web-store-base.js
paulmelnikow: `BaseService._request` delegates to `checkErrorResponse.asPromise` which emits `Inaccessible` for 5xx errors. Is that what you're asking? 😀

chris48s: Ah no. In the 5XX error case, that's fine. We can pass the exception object to `checkErrorResponse` because we've got a `statusCode` but in the case where we just get no response at all  (simulated with `networkOff()` in the tests), there is a different error which doesn't have a `statusCode` because we didn't get a response. That's the bit of handling I'm struggling to track down.

paulmelnikow: `.networkOff()` causes a request to immediately trigger an ENETUNREACH error.

I believe that gets passed as the `err` parameter to the `request` callback – that callback gets `(error, response, body)` – which eventually is re-emitted here:

https://github.com/badges/shields/blob/0941b6c39ca0021378abfe825b34f428ef553d86/core/base-service/legacy-request-handler.js#L245

This is the code in nock that generates it; as you say there is no status code:

https://github.com/nock/nock/blob/8493b80656a1b4c6952fa74ba814f1ee56e9404a/lib/intercept.js#L32-L34

chris48s: I was looking for <pre><strong>throw</strong> new Inaccessible</pre> which is why I wasn't matching this. Thought I was going out of my mind there.

paulmelnikow: More reasons to get rid of legacy code!

PR #3115, opened 2019-02-27
Remove Heroku buildpack config; allow devDependencies to be pruned

paulmelnikow:
The buildpack config doesn't seem to be needed. This would affect review apps, staging, and self-hosting.

Also, devDependencies are installed by default during the build: https://devcenter.heroku.com/changelog-items/1376

By removing `NPM_CONFIG_PRODUCTION=false` we allow these to be pruned so they're kept out of the slug. This also has the advantage of creating a test environment in PRs where missing production dependencies will cause things to break. That'd be a good thing!


calebcartwright:
I suppose the working Review App here is as good a proof as any that the packs are indeed unused

> This also has the advantage of creating a test environment in PRs where missing production dependencies will cause things to break. 

👍

paulmelnikow:
I just recreated the app because from the build log it was clear the env variable change didn't take effect. Makes sense, since the env vars are copied to the app when it's created.

paulmelnikow:
It's in effect now! [The slug](https://devcenter.heroku.com/articles/slug-compiler#slug-size) went from 76.2 MB to 31.6 MB, so that's a plus, too!

calebcartwright:
Nice! The dev dependency change will be a big help in catching would-be issues for the Docker build much earlier

PR #3117, opened 2019-02-28
Refactor [bitrise]

paulmelnikow:


services/bitrise/bitrise.service.js
```diff
@@ -1,23 +1,27 @@
...
-// This legacy service should be rewritten to use e.g. BaseJsonService.
-//
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class Bitrise extends LegacyService {
+// https://devcenter.bitrise.io/api/app-status-badge/
+const schema = Joi.object({
+  status: Joi.equal('success', 'error', 'unknown'),
```
calebcartwright: My very first thought here was "could we use the `build-status` helpers", but then realized that probably wouldn't make sense here since Bitrise isn't just a build service, we need to handle `unknown` values with a specific message, etc. 

I get it now 👍

calebcartwright: Although I see we're now using build status in the tests 😄

paulmelnikow: The old tests defined a new schema, which I'd be okay with doing, though using isBuildStatus seems marginally easier to maintain. And yea, I think the schema on the request should be tailored to the request. Much of the point of these schemas is to make the expectations of the response really clear so if things change in the future you can read the contract and understand the assumptions made by the contributor who came before you.

calebcartwright: If Bitrise changed the value they returned from `success` to a different value, say `succeeded`, the badges would break (`invalid json response`) but our daily service tests would still be passing since `succeeded` is in `buildStatus` right?

calebcartwright: Nevermind!

Issue #2686, opened 2019-01-08
Modernize field names in JSON-format badge responses


paulmelnikow:
In our badges we're using **label** and **message** which are semantically really good names, however our JSON endpoint still uses **name** and **value**… and so do all our service tests.

I'd suggest we deprecate **name** and **value** in the JSON response (maybe keep them around for a year or two?) and add **label** and **message** as synonyms. At the same time, we can add **color**.

Most services do not consider it a breaking change to add another item to an object. However there is a slight chance that adding unexpected keys could break someone's integration.

The service tests will need to be modified to accept extra keys. However we could take the opportunity to make those tests more concise (#2687).

chris48s:
I would be really interested to understand what JSON format badges are being used for, outside of our own test suite.

paulmelnikow:
Good question 🤔 

One thought if we do #2674 is that JSON could be handled in a query parameter and/or request header. That way we could keep the old `.json` format around for a while, while using the Endpoint schema (#2473) for the new one

paulmelnikow:
> But really I can't actually see it being use outside of our testing, do we have any metrics for format usage?

No, we don't, but it would be a good idea to include that when we do #966.

Issue #3089, opened 2019-02-23
Use fixtures for tests of dynamic badges


paulmelnikow:
The dynamic badge tests depend on live data, and it's one more thing that can break: https://github.com/badges/shields/pull/3049#issuecomment-466599836

There's no benefit to having the dynamic badges depend on data that must be fetched live. It'd be simpler and more self-documenting to make 100% of the tests use fixtures and `intercept()`.

PR #3122, opened 2019-02-28
Migrated most service tests to use new expectBadge

PyvesB:
Follow up to #3097 and related to #2687. This pull request converts most service tests to use the new `expectBadge` helper, apart from:
* Wercker's tests which was completed as part of #3097.
* Bitrise's tests which was completed as part of #3117.
* the static-badge `Not a valid color` test. This checks for a `null` color, which is not currently supported by the `expectBadge` helper. I will take care of this is a subsequent pull request.
* two tests in `endpoint.tester.js`, which are checking for unusual fields such as `labelColor` or `logoWidth` in the JSON response. I was hoping that we could remove the `_shields_test` style by adding the `color` field to the JSON response when tackling #2686 and therefore have `expectBadge` only ever check for `label`, `message` and `color`, but these two special tests will prevent us from doing that. Thoughts?

Don't be too frightened by the size of the diff! Most changes are pretty straightforward and I've left a few explanations for the less obvious ones. Also, to make things easier for the reviewer I've only strictly modified the `*.tester.js` files in `services`. No changes to the service or helper code. 😉


services/nodeping/nodeping-status.tester.js
```diff
@@ -3,13 +3,12 @@
 const Joi = require('joi')
 const t = (module.exports = require('../tester').createServiceTester())
 
-t.create('NodePing status - live').get(
-  '/jkiwn052-ntpp-4lbb-8d45-ihew6d9ucoei.json'
-)
-Joi.object().keys({
```
PyvesB: This was incorrectly set up.

services/opencollective/opencollective-all.tester.js
```diff
@@ -20,43 +19,35 @@ t.create('renders correctly')
...
-    })
-  )
+  .expectBadge({
+    label: 'backers and sponsors',
+    message: nonNegativeInteger,
+  })
 
 t.create('renders not found correctly')
-  .get('/nonexistent-collective.json?style=_shield_test')
+  .get('/nonexistent-collective.json?style=_shields_test')
```
PyvesB: Several of the OpenCollective tests were missing the `s` in `_shields_test`, the new helper highlighted this mistake.

services/opencollective/opencollective-by-tier.tester.js
```diff
@@ -1,11 +1,10 @@
 'use strict'
 
-const Joi = require('joi')
 const { nonNegativeInteger } = require('../validators')
 const t = (module.exports = require('../tester').createServiceTester())
 
 t.create('renders correctly')
-  .get('/shields/2988.json')
+  .get('/shields/2988.json?style=_shields_test')
```
PyvesB: Several of the OpenCollective tests were missing the `_shields_test` style entirely but were silently ignoring the color check, the new helper highlighted this mistake.

services/dependabot/dependabot.tester.js
```diff
@@ -1,31 +1,26 @@
...
 const { isIntegerPercentage } = require('../test-validators')
 const t = (module.exports = require('../tester').createServiceTester())
 
 t.create('semver stability (valid)')
   .get('/bundler/puma.json')
-  .expectJSONTypes(
-    Joi.object().keys({
-      name: 'semver stability',
-      value: isIntegerPercentage,
-      link:
```
PyvesB: This is not returned even with `_shields_test` specified.

PyvesB:
Thanks for taking a look @chris48s. Things also look good on my side when I run the tests, most failures seem related to #3089. Merging!

paulmelnikow:
Thanks for taking this on! Nice work!

PR #3123, opened 2019-02-28
Update instructions for self-hosting Shields with Heroku

henrypoydar:
The Heroku setup instructions were outdated. There's no need for a build pack.


calebcartwright:
Thanks @henrypoydar (and good timing! --> https://github.com/badges/shields/pull/3115)

PR #3125, opened 2019-03-01
Add ability to transform query params in redirct service, run [Endpoint GitHubReleaseRedirect SensioLabs VSO]

### LINKED ISSUE #3118: Add ability to transform query params on redirector services
Dialog in #3074, in particular https://github.com/badges/shields/pull/3074#issuecomment-466714842

Use case/benefit is being able to simplify route patterns by moving certain optional params (for example tokens, urls, etc.)  from route to query params

calebcartwright:
Close #3118,  refs https://github.com/badges/shields/pull/3074#issuecomment-466714842

Adds a new function (`transformQueryParams`) that can (optionally) be leveraged for redirect services to support transforming route/path params to query params, accordingly renames the `target` function to `transformPath`

Adding blocker label as this change is required to complete #3074


core/base-service/redirector.js
```diff
@@ -15,18 +16,19 @@ const trace = require('./trace')
...
-  target: Joi.func()
+  transformPath: Joi.func()
     .maxArity(1)
     .required()
     .error(
       () =>
-        '"target" must be a function that transforms named params to a new path'
+        '"transformPath" must be a function that transforms named params to a new path'
     ),
+  transformQueryParams: Joi.func().arity(1),
```
calebcartwright: Given the sole purpose of this function (transforming a route param to a query param, https://github.com/badges/shields/issues/3118#issuecomment-468503034) I believe it makes sense to require the argument be passed to the function

core/base-service/redirector.js
```diff
@@ -76,12 +78,21 @@ module.exports = function redirector(attrs) {
         trace.logTrace('inbound', emojic.ticket, 'Named params', namedParams)
         trace.logTrace('inbound', emojic.crayon, 'Query params', queryParams)
 
-        const targetUrl = target(namedParams)
+        const targetUrl = transformPath(namedParams)
```
paulmelnikow:
```suggestion
        const targetPath = transformPath(namedParams)
```

core/base-service/redirector.js
```diff
@@ -76,12 +78,21 @@ module.exports = function redirector(attrs) {
...
         trace.logTrace('validate', emojic.dart, 'Target', targetUrl)
 
+        let urlSuffix = ask.uri.search || ''
+
+        if (transformQueryParams) {
+          const transformedParams = transformQueryParams(namedParams)
+          const outQueryString = queryString.stringify(transformedParams)
+          const separator = urlSuffix ? '&' : '?'
+          urlSuffix = `${urlSuffix}${separator}${outQueryString}`
```
paulmelnikow: I wonder if we should let the parsed query parameters override the values that are produced here, rather than combining them. It may not be worth the effort to handle that case, though the concatenation of two query strings made me think of it.

calebcartwright: I'm not sure I'm following so here's my thinking/summary, please correct me/update as needed:

There's the potential for some query params to already be in the ask/search (`label=foo&style=...`) which isn't (at least not currently) made available to the `transformQueryParams` function. That's mostly because I'm looking at the function's purpose to solely provide the ability to transform a path/route param to a query param (it'd be a bit verbose, but if that's the intent should the signature be something more explicit like `transformPathParamsToQueryParams` )?

> I wonder if we should let the parsed query parameters override the values that are produced here

So by `override` are you thinking that we'd want to pass the named route params _and_ the existing query params to the `transformQueryParams` function and have that function return all query params? And/or are you thinking that there'd be a case where a redirector would want to transform a route/path param to a query param that may already be specified in the original (to be redirected) url? For example:

original route: `/very/old/service/token/:token/:otherParam` that wanted to move the token param to a query paramater, would be called with a query param such as `/very/old/service/token/abc/foo.svg?token=123` 

and that to avoid the redirector sending to something like.. `/very/old/service/token/abc/foo.svg?token=123&token=abc` (which is a technical possibility with this current implementation) that within the redirector implementation we should handle that duplication via merge/override (in which case I'd wonder which one would "win")?

core/base-service/redirector.js
calebcartwright: 👍 I've still yet to figure out how to  suggest changes via this new GH feature (def. a PEBCAK issue though 😆)

core/base-service/redirector.js
```diff
@@ -76,12 +78,21 @@ module.exports = function redirector(attrs) {
...
+        }
+
         // The final capture group is the extension.
         const format = match.slice(-1)[0]
-        const redirectUrl = `${targetUrl}.${format}${ask.uri.search || ''}`
+        const redirectUrl = `${targetUrl}.${format}${urlSuffix}`
```
calebcartwright:
```
const redirectUrl = `${targetPath}.${format}${urlSuffix}`
```

core/base-service/redirector.js
```diff
@@ -76,12 +78,21 @@ module.exports = function redirector(attrs) {
         trace.logTrace('inbound', emojic.ticket, 'Named params', namedParams)
         trace.logTrace('inbound', emojic.crayon, 'Query params', queryParams)
 
-        const targetUrl = target(namedParams)
+        const targetUrl = transformPath(namedParams)
         trace.logTrace('validate', emojic.dart, 'Target', targetUrl)
```
calebcartwright:
```suggestion
        trace.logTrace('validate', emojic.dart, 'Target', targetPath)
```

core/base-service/redirector.js
paulmelnikow: Yes, preventing `/very/old/service/token/abc/foo.svg?token=123&token=abc` is the case I’m thinking of.

I don’t think there’s a clear answer about which should win. TBH maybe keeping both is just fine. It generates a URL that will fail query parameter validation, which maybe is better than producing a valid URL that has ignored part of the original URL.

calebcartwright: Gotcha. When I first started I briefly considered trying to merge them, but it opened up too many questions so I punted 😄 

As I've been thinking on this some more though, I think we can probably handle this case by merging the two sets of query params relatively simply. I'd propose we pick a default priority to use, but then we can also add a new (optional) boolean parameter on the redirector function to allow an individual redirect service to flip that priority/win.

Will try to update this today

calebcartwright: Alright, I've taken a pass at adding this. Let me know what you think!

paulmelnikow: Okay. I think I could have said what I was saying more clearly! I was thinking that returning the garbage URL was better – since it would fail in an obvious way – though what you've done here is fine too! And this is arguably better behaved.

calebcartwright: Ah my bad. I'm good either way, my hope is that this a pretty uncommon edge case. 

Happy to let this sit for another day to simmer

calebcartwright:
The `Amo` tests are not failing due to this PR, but started failing because the addon being used in the live tests (`IndieGala-Helper) currently has 0 downloads which is rejected by the `isMetricOverTimePeriod` validator function.

paulmelnikow:
Aha, hmm, maybe we could find a more popular addon.

calebcartwright:
> Aha, hmm, maybe we could find a more popular addon.

I was just getting caught up on some of the backstory on this in Discord

PR #3126, opened 2019-03-01
Refactor [buildkite]

calebcartwright:


services/build-status.js
```diff
@@ -31,6 +31,7 @@ const otherStatuses = [
   'skipped',
   'starting',
   'stopped',
+  'unknown',
```
paulmelnikow: We've had some discussions in the past about what belonged in here, and my feeling is that `'unknown'` does not. It is used by several services as an error condition, not an actual build status.

calebcartwright: 👍 will change it. Had a lot of inner debate with myself on this one solely in the context of buildkite, so makes sense!

calebcartwright: Have reverted this back, and updated buildkite's schema to allow for it's unusual handling `unknown`

PyvesB: By the way, the discussion that @paulmelnikow is referring to took place [here](https://github.com/badges/shields/pull/1767#issuecomment-404379521).

PR #3128, opened 2019-03-01
update [clojars] endpoint

flyboarder:
Updates clojars badge to use api endpoints

Added pre-release endpoint which returns latest snapshot if available.

Updates version endpoint which will always return latest stable version or snapshot if no stable versions are available.


calebcartwright:
Hi @flyboarder, thanks for the PR! 

I updated the PR title with our naming convention for service related changes so that the tests for the clojars services would be executed. As you can see this PR is causing some of those tests to fail, could you please take a look?

services/clojars/clojars-release.service.js
```diff
@@ -0,0 +1,71 @@
+'use strict'
+
+const Joi = require('joi')
+const { version: versionColor } = require('../color-formatters')
+const { BaseJsonService, NotFound } = require('..')
+const { InvalidResponse } = require('../errors');
```
calebcartwright: `InvalidResponse` can be grabbed from index too, just like `BaseJsonService` and `NotFound`

flyboarder: already on it!

services/clojars/clojars-release.service.js
```diff
@@ -0,0 +1,70 @@
...
+      color: versionColor(version),
+    }
+  }
+
+  async handle({ clojar }) {
+    const json = await this.fetch({ clojar })
+
+    if (!json.latest_release) {
+
+      throw new InvalidResponse({ underlyingError: new Error('version is invalid or null') });
```
calebcartwright: Two suggestions here:

1) This is a minor nit but can you remove the linebreaks above and below this line?
2) 
```suggestion
      throw new InvalidResponse({ prettyMessage: 'version is invalid or null' })
```

services/clojars/clojars-release.service.js
```diff
@@ -0,0 +1,70 @@
...
+    if (Object.keys(json).length === 0) {
+      /* Note the 'not found' response from clojars is:
+          status code = 200, body = {} */
+      throw new NotFound()
+    }
+
+    return this.constructor.render({ clojar, version: json.latest_release })
+  }
+
+  // Metadata
```
calebcartwright: Can delete this line

services/clojars/clojars-release.service.js
calebcartwright: 👍

services/clojars/clojars-release.service.js
flyboarder: `// Metadata`?

services/clojars/clojars-version.service.js
```diff
@@ -28,13 +28,19 @@ module.exports = class ClojarsVersion extends BaseJsonService {
...
+      throw new InvalidResponse({
+        underlyingError: new Error('version is invalid or null'),
+      })
+    }
+
     if (Object.keys(json).length === 0) {
       /* Note the 'not found' response from clojars is:
           status code = 200, body = {} */
       throw new NotFound()
     }
```
calebcartwright: A common pattern we like to leverage in service classes is defining a function (usually called `transform`) that handles the responsibility of extracting the desired data point(s) out of the json response. That would be a good update here, especially since this block of code is duplicated in both service classes. 

Let's try to define the `transform` function in a single place (either in a separate/helpers file like `clojars-helpers.js` or create a base class and add it there) and then leverage that function in both classes. 

That would result in the handle function becoming something like:
```js
  async handle({ clojar }) {
    const json = await this.fetch({ clojar })
    const { version } = transform(json)
    return this.constructor.render({ clojar, version })
  }
```

and you can have a transform function somewhere like:
```js
transform(json) {
  if (!json.latest_version) {
    throw new InvalidResponse({
      underlyingError: new Error('version is invalid or null'),
    })
  }

  if (Object.keys(json).length === 0) {
    /* Note the 'not found' response from clojars is:
      status code = 200, body = {} */
    throw new NotFound()
  }

  return { version: json.latest_version }
}
```

services/clojars/clojars-release.service.js
calebcartwright: 👍

flyboarder:
@calebcartwright I will refactor to include a clojars base service class.

flyboarder:
@calebcartwright Should be good now!

services/clojars/clojars-version.service.js
flyboarder: done!

services/clojars/clojars-release.service.js
flyboarder: removed!

services/clojars/clojars-release.service.js
flyboarder: fixed!

flyboarder:
@calebcartwright we should be good now!

calebcartwright:
https://shields-staging-pr-3128.herokuapp.com/

services/clojars/clojars-base.js
```diff
@@ -0,0 +1,65 @@
...
+}).required()
+
+class BaseClojarsService extends BaseJsonService {
+  async fetch({ clojar }) {
+    const url = `https://clojars.org/api/artifacts/${clojar}`
+    return this._requestJson({
+      url,
+      schema: clojarsSchema,
+      errorMessages: {
+        404: 'not found',
```
calebcartwright: `not found` is the default error message displayed automatically. You can use this if you want to customize that message (for example other services do things like `repo not found`, `project not found`, etc.), but if you're good with the default `not found` message then you don't need this

services/clojars/clojars-base.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  static transform(json, version) {
+    if (version === 'release') {
+      return { version: json.latest_release }
+    } else if (version === 'version') {
+      return { version: json.latest_version }
+    } else {
+      throw new InvalidResponse({
+        prettyMessage: new Error('version is invalid or null'),
```
calebcartwright: you can pass your desired message as a string (not an error) via `prettyMessage`. Please change to:

```suggestion
        prettyMessage: 'version is invalid or null',
```

services/clojars/clojars-base.js
```diff
@@ -0,0 +1,65 @@
...
+      url,
+      schema: clojarsSchema,
+      errorMessages: {
+        404: 'not found',
+      },
+    })
+  }
+}
+
+class ClojarsVersionService extends BaseClojarsService {
```
calebcartwright: If this class is also intended to be a base class, then let's include `Base` somewhere in the service name. It was a little confusing at first when I saw this line `module.exports = class ClojarsVersion extends ClojarsVersionService {` in the `clojars-version.service.js` file

services/clojars/clojars-base.js
```diff
@@ -0,0 +1,65 @@
...
+
+const clojarsSchema = Joi.object({
+  downloads: nonNegativeInteger,
+  latest_version: Joi.string().required(),
+  latest_release: Joi.string().required(),
+}).required()
+
+class BaseClojarsService extends BaseJsonService {
+  async fetch({ clojar }) {
+    const url = `https://clojars.org/api/artifacts/${clojar}`
```
calebcartwright: If there's API documentation, can you add a comment with a link? For example:

`// Clojars Doc: https://....`

services/clojars/clojars-base.js
flyboarder: removed `errorMessages`

services/clojars/clojars-base.js
flyboarder: converted `prettyMessage` to string

services/clojars/clojars-base.js
flyboarder: renamed `ClojarsVersionService` to `BaseClojarsVersionService`

services/clojars/clojars-base.js
flyboarder: added comment to Clojars API reference on wiki

calebcartwright:
Thanks for the updates! I'll take another pass through this shortly

flyboarder:
Demo with changes:

ClojarsVersion: ClojarsVersion
ClojarsVersionSnapshot: ClojarsVersion

calebcartwright:
@flyboarder - the second badge in your example (that was using `vpre`) is no longer working as you've subsequently removed that route. 

Do you think there's a benefit/desire for users to be able to differentiate between release/prerelease? If that data is available via the API it may be worth considering but we didn't have that before so I'll defer to you whether you think that's worth adding. 

Also thanks for your continued work on this PR! The new API looks great (especially the returning of 404 for errors vs. the old endpoint 😄 )

services/clojars/clojars-base.js
```diff
@@ -0,0 +1,63 @@
...
+class BaseClojarsVersionService extends BaseClojarsService {
+  static render({ clojar, version }) {
+    return {
+      message: `[${clojar} "${version}"]`,
+      color: versionColor(version),
+    }
+  }
+
+  static transform(json, version) {
+    if (version === 'release') {
```
calebcartwright: the transition to the new API has eliminated a lot of the repetitive code that existed before. It seems that with the new API now, the snapshot badge is grabbing the `latest_version` field, and the release version badge is grabbing the `latest_release` field (no sharing/duplication).

It might be easier to just delete of this function now, and then updating the two service classes to grab their desired field in their respective `handle` functions. For example the release version service could become:

```js
  async handle({ clojar }) {
    const json = await this.fetch({ clojar })
    return this.constructor.render({ clojar, version: json.latest_release })
  }
```

Sorry, didn't mean to flip flop on `transform` 😄  but the code required for the transform process was dramatically reduced 👍 With the current service implementations the `else` block here can't/won't be hit otherwise

flyboarder: removed transform as it's now not needed.

PR #3131, opened 2019-03-02
Refactor [Codeship]

calebcartwright:


services/build-status.js
```diff
@@ -31,6 +32,7 @@ const otherStatuses = [
   'skipped',
   'starting',
   'stopped',
+  'testing',
```
calebcartwright: These three additional status (`infrastructure_failure` for a red status, and `initiated` and `testing` for other status) seemed like reasonable additions to the build status lists that wouldn't conflict with any other service (not that I could find, nor that I could envision). 

Let me know if anyone disagrees and/or if there's a reason we think these shouldn't be added

services/codeship/codeship.service.js
```diff
@@ -1,19 +1,44 @@
...
+const statusMap = {
+  testing: pendingStatus,
+  waiting: pendingStatus,
+  initiated: pendingStatus,
+  stopped: notBuiltStatus,
+  ignored: notBuiltStatus,
+  blocked: notBuiltStatus,
+}
+
+module.exports = class Codeship extends BaseSvgScrapingService {
```
calebcartwright: There is a JSON based API. However, it's rate limited (60 requests per minute) so I figured it would make sense to continue using the SVG.

paulmelnikow: Agreed. The other alternative would be to see if they'd provide us a mega-token with a higher quota.

services/codeship/codeship.service.js
```diff
@@ -1,19 +1,45 @@
...
+
+const statusMap = {
+  testing: pendingStatus,
+  waiting: pendingStatus,
+  initiated: pendingStatus,
+  stopped: notBuiltStatus,
+  ignored: notBuiltStatus,
+  blocked: notBuiltStatus,
+  infrastructure_failure: 'failed',
+}
```
paulmelnikow: This mapping seems unusual, though no one has complained about it, so maybe we ought to leave it alone for now.

services/codeship/codeship.service.js
calebcartwright: 👍 Once we starting get a count of how frequently this is badge is used we'll have a better idea. If it's only a few per minute then that rate limit may not be even an issue

services/codeship/codeship.service.js
calebcartwright: Yeah I know what you mean. This was one of those things that just felt easier to keep consistent with the legacy service experience, at least for the time being

PR #3132, opened 2019-03-02
Refactor [bundlephobia]

calebcartwright:


services/bundlephobia/bundlephobia.service.js
```diff
@@ -27,78 +42,31 @@ module.exports = class Bundlephobia extends LegacyService {
       {
         title: 'npm bundle size',
         pattern: ':format(min|minzip)/:packageName',
```
chris48s: Looking at this on the staging app, we're missing some of the params in the front-end:

screenshot at 2019-03-02 21-58-31

It might just be a matter of removing this line, or maybe we need to provide several examples in order to cover the scope/version cases

calebcartwright: That's the experience we currently have in prod today, but agreed on adding/updating examples 👍

calebcartwright: I tried the single example with no pattern overriding, but the scope param kept crashing things. Added a few different examples accordingly:

image

PR #3133, opened 2019-03-02
Refactor [waffle]

calebcartwright:


services/waffle/waffle-label.service.js
```diff
@@ -0,0 +1,104 @@
...
+      label: Joi.object({
+        name: Joi.string().required(),
+        color: Joi.string()
+          .hex()
+          .required(),
+      }),
+      count: nonNegativeInteger,
+    })
+  )
+  .required()
```
chris48s: It looks like this schema is rejecting some valid cases. Here are a couple of valid API responses which are generating "invalid response data":

* https://api.waffle.io/waffleio/waffle.io/columns
* https://api.waffle.io/mysociety/fixmystreet/columns

I think we need to be more lenient on the inputs but maybe we also need to do some extra checks in `handle()` or `transform()`

services/waffle/waffle-label.service.js
```diff
@@ -0,0 +1,104 @@
...
+          count: col.count,
+          color: col.label.color,
+        }
+      }
+    }
+
+    return { count: 0 }
+  }
+
+  async handle({ user, repo, label = 'ready' }) {
```
chris48s: I can see this is a default that we already had before refactoring, but I'm not sure this is a 'well-known' default as such. I don't think waffle boards necessarily have to have a 'ready' label. e.g: https://waffle.io/waffleio/waffle.io doesn't have one. If we're going to use this as a default for BC, can we document it in the example help.

services/waffle/waffle-label.service.js
calebcartwright: Good catch! It looks like there can apparently be labels that do not have a color, so I think a simple tweak to the schema should resolve that. 

I believe that means in the current/legacy implementation that the color would have been `undefined` in those use cases, so I wonder if we should keep that here or use the default color (78bdf2 apparently)?

services/waffle/waffle-label.service.js
calebcartwright: I've never used the service myself so I wasn't really sure what to do here either. I seriously considered removing that default (as it wasn't at all clear to me _why_ that was the default), but was more concerned about breaking existing users/badges that didn't explicitly specify a label. 

I can certainly add docs on the example stating that we will assume `ready` if no label is specified, but the only rationale I have for making that assumption is "because that's the way the legacy service implementation did it" 😄

calebcartwright: I suppose another option would be to make `label` a required route parameter in this service, and remove the default value of `ready`. I could add a redirector service that would catch the no label paths (`/waffle/label/user/repo`) and forward that along with `ready` to maintain backwards compatibility. 

We still get the same functional result, but at least the UI/net users will all see `label` as a required param

chris48s: Yeah that's a pretty nice solution. Good thinking :+1:

services/waffle/waffle-label.service.js
calebcartwright: I've updated the schema accordingly, and the two above examples are working in the staging app now 👍 

image


image

chris48s:
I've not looked at it in a lot of detail, but based on the error messages, I expect this needs to be updated to account for https://github.com/badges/shields/pull/3125 in order to get the tests to pass

calebcartwright:
Yup! Will update this later today

PR #3134, opened 2019-03-02
Refactor [continuousphp]

calebcartwright:


services/continuousphp/continuousphp.service.js
```diff
@@ -32,75 +49,25 @@ module.exports = class ContinuousPhp extends LegacyService {
           repo: 'dbal',
           branch: 'master',
```
chris48s: Can we do 2 examples here - one with the branch param and one without.

Giving only the one essentially makes `branch` required in the front-end:

screenshot at 2019-03-02 22-07-52

calebcartwright: That's the experience we currently have in prod today, but agreed on adding some more examples 👍 

I forget which badge it was, but I saw one in the model where it was like `branch (optional)`. Not sure how to do that though

calebcartwright: Made an update, thoughts on this?  There's only a single example listing, but this makes it much more clear that the branch is optional

image

chris48s: I think this would make this the only badge where we're exposing `:branch*` in the front-end instead of doing 2 examples. Maybe now that we have a better editor in the front-end we could review how we present optional params, but that's probably a larger chunk of work. Shall we just keep this one consistent for now

calebcartwright: Works for me! I've seen a handful of other badges (refactored ones), like Bitrise for example, where we are exposing `:branch*` in the frontend so I just assumed that's what we were striving for. 

The vast majority do still have multiple examples so I'll update this accordingly to be consistent 👍

PR #3139, opened 2019-03-02
feat: [DockerCloud] and automated.

### LINKED ISSUE #1400: Problem with Docker build badge: repo not found
=> works fine
 => not found

Same repo. Am I missing something?

ldez:
A "new" automated build:
```console
$ curl -s 'https://hub.docker.com/api/build/v1/source/?image=ldez%2Fgha-mjolnir'  -H 'Accept: application/json' --compressed | jq
{
  "meta": {
    "limit": 25,
    "next": null,
    "offset": 0,
    "previous": null,
    "total_count": 1
  },
  "objects": [
    {
      "autotests": "OFF",
      "build_in_farm": true,
      "build_settings": [
        "/api/build/v1/setting/e3db21aa-8847-4789-9f2e-dc136ea432ae/",
        "/api/build/v1/setting/6fe17d9d-a946-47f0-a7ae-b20dba0bfb72/"
      ],
      "channel": "Stable",
      "image": "ldez/gha-mjolnir",
      "owner": "ldez",
      "provider": "Github",
      "repo_links": false,
      "repository": "gha-mjolnir",
      "resource_uri": "/api/build/v1/source/9a37d4ca-aad6-4820-932c-5d8bd94fc8d7/",
      "state": "Success",
      "uuid": "9a37d4ca-aad6-4820-932c-5d8bd94fc8d7"
    }
  ]
}
```

`build_settings`: contains the build triggers (each entry is a different trigger). Can be see as the old "automated" flag.

`state`: the build state.

Fixes #1400


services/docker/docker-cloud-build.service.js
```diff
@@ -0,0 +1,61 @@
...
+const { dockerBlue, buildDockerUrl } = require('./docker-helpers')
+
+const cloudBuildSchema = Joi.object({
+  objects: Joi.array()
+    .items(Joi.object({ state: Joi.string() }).required())
+    .required(),
+}).required()
+
+module.exports = class DockerCloudBuild extends BaseJsonService {
+  async fetch({ user, repo }) {
```
calebcartwright: This `fetch` function can be extracted out into a single location (perhaps a DockerCloud base class) and reused instead of having it redefined in both classes

services/docker/docker-cloud-automated.service.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'docker build' }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Docker Automated build (new hub)',
```
calebcartwright: Is there a different title that could be used? The current (old) Docker build badge example title is `Docker Build Status`, so perhaps `Docker Cloud Automated build` here?

calebcartwright:
Reviewable @ https://shields-staging-pr-3139.herokuapp.com/

paulmelnikow:
Thanks for picking this up @Idez!

services/docker/docker-cloud-automated.service.js
paulmelnikow: I like @calebcartwright's suggestion.

It could also be useful to add some documentation that clarifies when each of these badges should be used. If you add a `documentation` key to the `examples` containing some HTML, it'll render on the badge popup.

services/docker/docker-cloud-build.service.js
paulmelnikow: If you don't want to create a base class, another option is to make a function:

https://github.com/badges/shields/blob/879581466cd94cbad706b92fdc5449080774fd12/services/github/github-common-fetch.js#L7-L19

One of the reasons I like using functions instead of classes to share code among services is that it makes this particularly common refactor easier to execute.

services/docker/docker-cloud-common-fetch.js
```diff
@@ -0,0 +1,26 @@
...
+        build_settings: Joi.array(),
+      }).required()
+    )
+    .required(),
+}).required()
+
+async function fetchBuild(serviceInstance, { user, repo }) {
+  return serviceInstance._requestJson({
+    schema: cloudBuildSchema,
+    url: `https://cloud.docker.com/api/build/v1/source/?image=${user}/${repo}`,
```
calebcartwright: let's use `qs` here just to ensure everything gets handled correctly, 

```suggestion
    url: `https://cloud.docker.com/api/build/v1/source`,
    options: { qs: { image: `${user}/${repo}` } },
```

services/docker/docker-cloud-automated.tester.js
```diff
@@ -0,0 +1,45 @@
...
+  .expectBadge({
+    label: 'docker build',
+    message: isAutomatedBuildStatus,
+  })
+
+t.create('docker cloud automated build (not found)')
+  .get('/badges/not-a-real-repo.json')
+  .intercept(nock =>
+    nock('https://cloud.docker.com/')
+      .get('/api/build/v1/source/?image=badges/not-a-real-repo')
```
calebcartwright: I think nock is probably barking on `...source/?image` vs `...source?image...` you may also want to use a template string and do a `${encodeURIComponent(...)}` on that part if it keeps failing

services/docker/docker-cloud-common-fetch.js
ldez: it breaks tests :cry:

calebcartwright: yeah I noticed that too, I left an inline comment on one of the testers as to what I think may be the cause

calebcartwright: 👍 working now

PR #1536, opened 2018-02-28
[microbadger] Handle no DownloadSize being returned

PyvesB:
I noticed that we had one failing MicroBadger test as well as an error badge on our homepage.

Turns out that some API responses now come back with no top-level `DownloadSize` field (for instance [this query](https://api.microbadger.com/v1/images/library/centos)). Our parsing currently throws an exception resulting in the error MicroBadger badge.

This pull request handles things more cleanly with a new "unknown" badge, and fixes the test and homepage example.

@goozler as you submitted the MicroBadger implementation a few months ago, could you please give this a look and see whether we're on the right track here? 😉

PR #3143, opened 2019-03-03
refactor [microbadger] service

chris48s:
@PyvesB - I've left a couple of comments on the diff for this one. I know you've done quite a bit of work on this service in the past so maybe you have some ideas on this.


services/microbadger/microbadger-size.tester.js
```diff
@@ -0,0 +1,48 @@
...
+  })
+
+t.create('missing download size')
+  .get('/puppet/puppetserver.json')
+  .intercept(nock =>
+    nock('https://api.microbadger.com')
+      .get('/v1/images/puppet/puppetserver')
+      .reply(200, {})
+  )
+  .expectBadge({ label: 'image size', message: 'invalid response data' })
```
chris48s: I've switched this so we'll just fail validation in this case rather than having a special error case for it. Does this seem reasonable? Do we often see a case where the size is missing?

services/microbadger/microbadger-size.tester.js
```diff
@@ -0,0 +1,48 @@
...
+
+t.create('specified tag when repository has only one')
+  .get('/_/alpine/wrong-tag.json')
+  .expectBadge({ label: 'image size', message: 'not found' })
+
+t.create('nonexistent repository')
+  .get('/_/not-a-real-repo.json')
```
chris48s: I've switched the 'not found' tests to hit the live service, but previously we were mocking the 404. Were we mocking the 404 for a reason?

services/microbadger/microbadger-size.tester.js
PyvesB: It's quite a common case. If you're accessing an image (or the new version of an image) for the first time through Microbadger, the metadata will not be available and will be populated asynchronously. I believe it may take a few seconds for the information to be available through the API, you would basically need to query again after that period.

I think it would be ideal to have a special case saying "information not yet available" or something like that.

services/microbadger/microbadger-size.tester.js
PyvesB: Maybe for speed? The Microbadger tests have always been rather slow, we've had quite a few timeout issues in the past. But if the 404 response comes back quickly, I'm also happy for us to hit the live service. 👍

PyvesB:
I've left my thoughts on your two questions. Give me a shout when you want me to review this in more details. 😉

services/microbadger/microbadger-size.tester.js
chris48s: Right - makes sense. Looking back over the code added https://github.com/badges/shields/pull/1536 it looks like this is a problem for size, but not layers and it will manifest by `DownloadSize` just not being present in the response at all (as opposed to `DownloadSize: null`), so I've tweaked the schema to handle that case.

services/microbadger/microbadger-size.tester.js
chris48s: I've not seen it time out, but we'll see how it goes..

chris48s:
Thanks - updated. Should be good to go

PR #3157, opened 2019-03-05
refactor [GitHubFollowers]

calebcartwright:


services/github/github-followers.service.js
```diff
@@ -26,57 +22,42 @@ module.exports = class GithubFollowers extends LegacyService {
...
   static get examples() {
     return [
       {
         title: 'GitHub followers',
         namedParams: {
           user: 'espadrine',
         },
-        staticPreview: {
+        staticPreview: Object.assign(this.render({ followers: 150 }), {
           label: 'Follow',
```
calebcartwright: Wasn't sure how badly we wanted to the example listing to keep the `Follow` label, but this was the only way I could get it to work

paulmelnikow: Yea… I think I left a literal in one of the other GitHub services I refactored recently. This seems better.

calebcartwright:
Crap! The example needs to have `style: 'social'` explicitly set, probably alongside the label setting. I was messing around with the examples and forgot to put that back in

PR #3160, opened 2019-03-05
Refactor [librariesio]

paulmelnikow:
This takes another pass over the modern services to remove unused code. I switched the shared code to use a function instead of a class and removed the indirection in the route params which led to skew between the route and examples and wasn't making things any clearer.


services/librariesio/librariesio-dependents.tester.js
```diff
@@ -17,10 +17,3 @@ t.create('dependent count (nonexistent package)')
     label: 'dependents',
     message: 'package not found',
   })
-
-t.create('dependent count (repo)')
```
paulmelnikow: It's not clear when this last could have passed 🤔

services/librariesio/librariesio-dependencies.service.js
```diff
@@ -1,134 +1,137 @@
...
-// Do not base new services on this code.
-module.exports = class LibrariesioDependencies extends LegacyService {
+function transform({ dependencies }) {
+  return {
+    deprecatedCount: dependencies.filter(dep => dep.deprecated).length,
+    outdatedCount: dependencies.filter(dep => dep.outdated).length,
+  }
+}
+
+function renderDependenciesBadge({ deprecatedCount, outdatedCount }) {
```
calebcartwright: I know that all these paths weren't fully tested in the legacy service implementation, but would it be worth adding a test/tests to validate this?

These two functions (`transform` and `renderDependenciesBadge`) are ones I've been testing with `sazerac` and friends lately, although it looks like testing with that approach may not be straightforward with this current implementation

paulmelnikow: I could add some sazerac tests of the render function. I could add some of `transform` too, though I'm not sure they're as meaningful as it's so trivial.

services/librariesio/librariesio-dependencies.service.js
```diff
@@ -1,134 +1,137 @@
...
+  } else {
+    return {
+      message: 'up to date',
+      color: 'brightgreen',
+    }
+  }
+}
+
+const schema = Joi.object({
+  dependencies: Joi.array().default([]),
```
paulmelnikow: Oops, forgot to update this!

services/librariesio/librariesio-dependencies.service.js
calebcartwright: It's not a blocker for me, just noticed that the branch coverage for the librariesio family is only 66.67%

services/librariesio/librariesio-dependencies.service.js
calebcartwright: I'm good to approve this PR once you're set with the schema 👍

services/librariesio/librariesio-dependencies.service.js
paulmelnikow: Have added some tests.

services/librariesio/librariesio-dependencies.service.js
paulmelnikow: Added!

PR #3164, opened 2019-03-05
Add query param validation to remaining new-style services [azuredevops appveyor npm]

### LINKED ISSUE #2675: Add query param validation to new-style services which don't have it yet
We've a few services which validate query params by ~~invoking `_validateQueryParams` on BaseService~~ returning a `queryParamSchema` from `route`.

The ones that don't yet should have it added!

paulmelnikow:
Remove now-obsolete code.

Close #2675


core/base-service/base.js
```diff
@@ -106,9 +106,7 @@ module.exports = class BaseService {
    *  - capture: Array of names for the capture groups in the regular
    *             expression. The handler will be passed an object containing
    *             the matches.
-   *  - queryParams: Array of names for query parameters which will the service
-   *                 uses. For cache safety, only the whitelisted query
-   *                 parameters will be passed to the handler.
+   *  - queryParamSchema: Joi schema for valid query params.
```
paulmelnikow: This certainly could be improved. There basically are two types, boolean params and string params, and we should explain each one. If anyone wants to take a stab at that, please do!

services/test-results.js
```diff
@@ -0,0 +1,91 @@
+'use strict'
+
```
calebcartwright: 👍 on the encapsulating all the helpers in here

core/base-service/base.js
calebcartwright: As part of this PR or a follow up? My takeaway has been that for booleans we need use `Joi.equal('')` although I'm still not sure how to best articulate _why_ that's the case.

IMO it'd be sufficient to just mention that fact for booleans in the comment doc (and probably the tutorial at some point too). Or do you think explanation/details accompany it in the comments?

paulmelnikow: I opened #3170 which adds a usage comment. It partly answers why… see if that seems complete enough an explanation?

I don't really like how the query params are handled by the badge builder; seems like it could be time for another redesign of the builder. I'll open an issue with some proposals.

PR #3169, opened 2019-03-07
Show missing query param for observatory; clean up duplication in azure devops

paulmelnikow:


services/mozilla-observatory/mozilla-observatory.service.js
```diff
@@ -74,6 +74,7 @@ module.exports = class MozillaObservatory extends BaseJsonService {
           grade: 'A+',
           score: 115,
         }),
+        queryParams: { publish: null },
```
calebcartwright: Is this something we should start doing for all examples that have query params?

PR #3170, opened 2019-03-07
Document queryParams + rename example module

paulmelnikow:
This file does more than transform, and `examples` seems is a bit more consistent with e.g. `trace.js` and `route.js`.


paulmelnikow:
In the process of documenting this I found a lot of behaviors that seem less than ideal, with the way query params are rendered in the front end. I'll open a new issue with proposals.

core/base-service/base.js
```diff
@@ -106,7 +106,22 @@ module.exports = class BaseService {
...
    *             expression. The handler will be passed an object containing
    *             the matches.
-   *  - queryParamSchema: Joi schema for valid query params.
+   *  - queryParamSchema: (Optional) A Joi schema (`Joi.object({ ... }).required()`)
+   *                      for the query param object. If you know a parameter
+   *                      will never receive a numeric string, you can use
+   *                      `Joi.string()`. Because of quirks in Scoutcamp and Joi,
+   *                      alphanumeric strings should be declared using
+   *                      `Joi.alternatives().try(Joi.string(), Joi.number())`,
+   *                      otherwise a value like `?success_color=999` will fail.
```
calebcartwright: To clarify, we need to use `Joi.alternatives().try(Joi.string(), Joi.number())` because `?success_color=999` would fail on `Joi.number()` ?

calebcartwright: Never mind, based on the below content in the docs I believe it's the latter case 👍

Issue #3165, opened 2019-03-05
docker run does hangs forever


cthulu:
:beetle: **Description**

I'm following the instructions in `doc/self-hosting.md` to first build the Docker image (which works fine) and then run the container using `docker run --rm -p 8080:80 --name shields shields`.

It looks like the container hangs forever:

```
$ docker run --rm -p 8080:80 --name shields shields
raven@2.6.4 alert: no DSN provided, error reporting disabled
Configuration:
{ bind: { address: 'localhost' },
  metrics: { prometheus: { enabled: false } },
  ssl: { isSecure: false },
  cors: { allowedOrigin: [ 'http://localhost:3000' ] },
  persistence: { dir: './private' },
  services:
   { github:
      { baseUri: 'https://api.github.com/',
        debug: { enabled: false, intervalSeconds: 200 } },
     trace: false },
  profiling: { makeBadge: false },
  cacheHeaders: { defaultCacheLengthSeconds: 120 },
  rateLimit: false,
  handleInternalErrors: false,
  fetchLimit: '10MB' }
0305092655 FsTokenPersistence configured with /usr/src/app/private/github-user-tokens.json
0305092655 Server is starting up: http://localhost/
```

I think it expects a response from Gatsby or something (I'm not a Node expert).

If I run the app manually using `npm start` it works, but I'd like to containerize it.

paulmelnikow:
What is the expected behavior?

calebcartwright:
I haven't checked today, but my scheduled job successfully built the image and then ran it on Friday March 1st. That was with commit https://github.com/badges/shields/commit/56e71d7c765fa8ae2953126e2cbe512ef7f23a2d

cthulu:
@paulmelnikow the expected behavior is that http://localhost:8080 will be exposing port :80 of the container and having shields running. However I get connection refused as the container starts but the application startup never finishes - I even have to manually kill the container `docker kill shields`.

I have no issues with running any other containers (docker engine 18.0.9.2, Mac).

When running the `npm start` (not in Docker, just locally), the startup continues:
```
[server] 0306104015 Server is starting up: http://localhost:8080/
success open and validate gatsby-configs — 0.008 s
```

In container I'm stuck on the line `0306104216 Server is starting up: http://localhost/`

cthulu:
I've tried also 56e71d7c765fa8ae2953126e2cbe512ef7f23a2d with the same result.

[...]

PR #3173, opened 2019-03-07
Tweak Docker initialization

paulmelnikow:
1. Set NODE_ENV=production in Docker.
2. When NODE_ENV is production, bind to all interfaces. This seems like a
   sensible default.
3. Exclude Dockerfile from container to improve layer cacheability when
   modifying the dockerfile.

Ref #3165


paulmelnikow:
> I was setting the bind address env var as part of my docker run command, so this seems like a good move 👍

Excellent! That's great validation to have.

Dockerfile
```diff
@@ -6,20 +6,23 @@ RUN mkdir -p /usr/src/app
...
 
 COPY . /usr/src/app
 RUN npm run build
 RUN npm prune --production
 RUN npm cache clean --force
 
+# Run the server using production configs.
+ENV NODE_ENV production
+
 # Do we need to list the environment variables here?
```
paulmelnikow: @calebcartwright Does that mean we don't need to explicitly list any env vars here?

calebcartwright: Just reviewing all the env vars I'm setting on my `docker run` command, other than the tokens/secrets for various badges, and proxy config that's specific to where I'm running, I believe the only relevant one is the `-e BIND_ADDRESS=0.0.0.0` 

I'm also still adding `-e PORT=80 -e BASE_URL=http://localhost:8080` but I don't think either of those are necessary

paulmelnikow: `-e BASE_URL=http://localhost:8080` is also likely to not work when you run the container in your production environment, whereas there's a decent chance nothing will work. 😁

calebcartwright: Oddly enough, the app runs fine in the container, even with that arg, but I probably need to clean some of this up. 

I got it working a few months ago and was happy to leave it untouched as long as it kept running smoothly

calebcartwright:
I'm re-running my Docker build and deploy process, and have removed all the runtime env vars to validate

calebcartwright:
Still working for me post these changes (and removing those three env vars I mentioned above) 👍

Dockerfile
paulmelnikow: Including the frontend?

I'd imagine if the frontend were used on another machine, it would generate badge URLs that include `localhost`.

calebcartwright: Yup! I don't we actually use that env var anywhere do we?

paulmelnikow: Oh! Ha! Yes we do, but only at build time!

PR #3176, opened 2019-03-07
Fix console error before applying a frontend route redirect

paulmelnikow:
The error was generated in the `onRouteUpdate` handler, which is invoked before the component. This moves redirecting responsibility to the handler, which seems like a better place for it anyway.


calebcartwright:
Code changes LGTM, is there anything in particular worth looking over in the review app? I did a bit of poking around and things looked fine there

paulmelnikow:
Only thing to do would be to confirm it eliminates the console error!

PR #3177, opened 2019-03-07
Rewrite a bunch of components as functions, with consistent func style

paulmelnikow:


frontend/components/badge-examples.js
```diff
@@ -27,66 +30,70 @@ const ClickableCode = styled(StyledCode)`
...
-    )
-  }
+  return (
+    <ExampleTable>
+      <tbody>
+        {flattened.map(exampleData => (
+          <Example
+            baseUrl={baseUrl}
+            exampleData={exampleData}
+            key={`${exampleData.title} ${exampleData.example.pattern}`}
```
calebcartwright: This all looks good, including in the review app. I'm not sure what's going on with a few things, like `key` for example. but that stems from my lack of familiarity with React.

frontend/components/dynamic-badge-maker.js
```diff
@@ -62,6 +64,7 @@ function DynamicBadgeMaker({ baseUrl = document.location.href }) {
       </select>{' '}
       {inputs.map(({ name, placeholder = name }) => (
         <InlineInput
+          key={name}
```
calebcartwright: Just a question out of curiosity, why is the `key` needed now?

paulmelnikow: Yea, good question. I believe it was a regression in the previous PR: #3096

frontend/components/badge-examples.js
paulmelnikow: Identical adjacent elements need a unique key or React issues a warning. (It helps it reconcile the component tree with the DOM.) The key was generated before inside `renderExample` which was a helper function that returned a rendered component. The helper function has been replaced with a component, which means the key has to be moved here. Otherwise, what we'd have here would be a bunch of adjacent identical <Example> elements (regardless of what may be different inside them).

PR #3184, opened 2019-03-08
Rewrite [website]; also affects [nodeping]

paulmelnikow:


services/website/website.tester.js
```diff
@@ -3,35 +3,37 @@
 const t = (module.exports = require('../tester').createServiceTester())
 
 t.create('status of http://shields.io')
-  .get('/website/http/shields.io.json?style=_shields_test')
-  .expectBadge({ label: 'website', message: 'online', color: 'brightgreen' })
+  .get('/http/shields.io.json?style=_shields_test')
+  .expectBadge({ label: 'website', message: 'up', color: 'brightgreen' })
```
paulmelnikow: I changed the defaults here to match what we did for nodeping. They seem a bit nicer, though I don't feel strongly about it.

services/website/website-redirect.service.js
```diff
@@ -0,0 +1,85 @@
...
+      Nothing:&nbsp;
+      <code>…/website/…</code>
+    </li>
+    <li>
+      Online and offline text:&nbsp;
+      <code>…/website-up-down/…</code>
+    </li>
+    <li>
+      Online and offline text, then online and offline colors:&nbsp;
+      <code>…/website-up-down-green-orange/…</code></li>
```
paulmelnikow: I created redirects for the old URL pattern. The new one matches what we did for Nodeping, and can be fully customized in the badge modal.

services/website/website-redirect.service.js
```diff
@@ -0,0 +1,85 @@
...
+      'unused7',
+      'downColor',
+      'unused8',
+      'protocol',
+      'hostAndPath',
+    ],
+  },
+  transformPath: ({ protocol, hostAndPath }) =>
+    `/website/${protocol}/${hostAndPath}`,
+  transformQueryParams: ({ upMessage, downMessage, upColor, downColor }) => ({
```
paulmelnikow: This is coming in handy @calebcartwright!

services/website/website.service.js
```diff
@@ -38,118 +30,76 @@ const documentation = `
...
+      down_message: downMessage,
+      up_color: upColor,
+      down_color: downColor,
+    }
+  ) {
+    let isUp
+    try {
+      const {
+        res: { statusCode },
+      } = await this._request({
```
paulmelnikow: `this._requestFetcher` throws http errors, so this try block is needed regardless.

services/website/website-redirect.service.js
calebcartwright: 👍 nice!

services/website-status.spec.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+const { test, given } = require('sazerac')
+const { renderWebsiteStatus } = require('./website-status')
+
+describe('Website status helpers', function() {
+  const customOptions = {
+    upMessage: 'groovy',
+    upColor: 'papayawhip',
```
calebcartwright: Woah haha. Did not know there was such a color

services/website/website.service.js
```diff
@@ -38,118 +30,76 @@ const documentation = `
...
   static get category() {
     return 'monitoring'
   }
 
   static get route() {
     return {
-      base: '',
-      pattern: '',
+      base: 'website',
+      pattern: ':protocol(https|http)/:hostAndPath',
```
calebcartwright: We're able to use the standard pattern format here for `hostAndPath` because the badge requires end users to do the url encoding correct? 

Would it be possible to treat this like the other badge patterns that accept a `hostAndPath` and then do any required encoding?

services/website/website.tester.js
calebcartwright: Works for me

services/website/website.service.js
paulmelnikow: Hmm, the docs do say that, though I don't think it has to be encoded in order to work. The examples in the test do not encode the slashes, for example.

calebcartwright: One of the examples in the docs used paths, i.e. `http/www.website.com/path/to/page.html` which I think would have to be encoded or the route param would need to be `:hostAndPath+`

Is that a scenario we need to support?

paulmelnikow: Ah, yes, it should be `hostAndPath+`. Sorry, I misunderstood when you first said that.

PR #3186, opened 2019-03-09
Rewrite GitHub commit status

paulmelnikow:
This decorates response errors with the response and response buffer, to make it easy for a service to parse an error response when it needs to. Since this is needed so rarely, I didn't put the error-parsing functionality in BaseJsonService, to save resources in the vast majority of cases.

### DIFF
--- a/services/github/github-commit-status.service.js
+++ b/services/github/github-commit-status.service.js
@@ -5,6 +5,11 @@ const { NotFound, InvalidParameter } = require('..')
 const { GithubAuthService } = require('./github-auth-service')
 const { documentation, errorMessagesFor } = require('./github-helpers')
 
+const schema = Joi.object({
+  // https://stackoverflow.com/a/23969867/893113
+  status: Joi.equal('identical', 'ahead', 'behind', 'diverged'),
+}).required()
+
 module.exports = class GithubCommitStatus extends GithubAuthService {
   static get category() {
     return 'issue-tracking'
@@ -64,7 +69,7 @@ module.exports = class GithubCommitStatus extends GithubAuthService {
       ;({ status } = await this._requestJson({
         url: `/repos/${user}/${repo}/compare/${branch}...${commit}`,
         errorMessages: errorMessagesFor('commit or branch not found'),
-        schema: Joi.object().required(),
+        schema,
       }))
     } catch (e) {
       if (e instanceof NotFound) {
--- a/services/github/github-commit-status.service.js
+++ b/services/github/github-commit-status.service.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const Joi = require('joi')
+const { NotFound, InvalidParameter } = require('..')
 const { GithubAuthService } = require('./github-auth-service')
 const { documentation, errorMessagesFor } = require('./github-helpers')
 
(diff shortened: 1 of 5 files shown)


services/github/github-commit-status.service.js
```diff
@@ -35,71 +27,56 @@ module.exports = class GithubCommitStatus extends LegacyService {
...
+    }
+  }
+
+  async handle({ user, repo, branch, commit }) {
+    let status
+    try {
+      ;({ status } = await this._requestJson({
+        url: `/repos/${user}/${repo}/compare/${branch}...${commit}`,
+        errorMessages: errorMessagesFor('commit or branch not found'),
+        schema: Joi.object().required(),
```
chris48s: If the schema is vague here so that we pass the success response and error response, could we do something like `Joi.alternatives(errorSchema, validSchema)`?

services/github/github-commit-status.tester.js
```diff
@@ -147,6 +87,6 @@ t.create('commit status - 404 with invalid JSON form github')
   )
```
chris48s: Minor point, but AFAIK, this isn't a realistic test case. The 404 response body for this endpoint should be:

```
{
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3/repos/commits/#compare-two-commits"
}
```

Would it be more realistic to test this response as a 5XX status code?

services/github/github-commit-status.service.js
paulmelnikow: Oops, I meant to fill that in.

Declaring the error schema doesn't do any good because _requestJson won't even parse the response on a 404.

services/github/github-commit-status.tester.js
paulmelnikow: We have live tests above of unknown branch and unknown commit. Should we add one for unknown repo?

This test is to make sure the code in the service that parses the error response works as desired, and doesn't flip out if it gets a 404 that doesn't contain valid JSON.

A 5xx error would get rejected normally (and re-thrown by the service). We could add a test for that if you think it seems helpful.

chris48s: So we're not actually expecting the GH to return a non-JSON response for some 404 case, we just want to use some status code that will mean we attempt to parse an invalid body (and 404 is a number which happens to satisfy that need)?

paulmelnikow: The API provides two kinds of 404 response. I think they all provide JSON responses, though normally we don't parse the JSON from 404 responses.

In this case we need to parse it. That way we can determine whether it's the "no common ancestor…" case, which we report as **no common ancestor**, or any other case, which we report as **repo not found**. There are tests above of each of those cases.

Something we're not expecting per se, but which _could_ happen is that GitHub could send some garbage back with a 404 response. We also don't want to crash when that happens. For the most part, we've been removing `invalidJSON` tests from services, not because we don't care about that case anymore, but because that behavior is provided by BaseJsonService, and that code has its own tests.

Here, I'd argue that, this service is providing its own behavior, so it ought to have it's own test.

We need to use a 404 and not a 5xx, because it's only in the case of a 404 that the service receives a NotFound error and follows the response-parsing code path.

https://github.com/badges/shields/pull/3186/files#diff-84ca388a6016d105807cc195db748b5bR75

chris48s: right, yeah I see what you mean now

PR #3187, opened 2019-03-10
fix [clojars] snapshot fallback

flyboarder:
Fix clojars fallback to snapshot when no stable version is available.


calebcartwright:
Thanks @flyboarder. You mentioned in Discord that you were seeing some "invalid response data" errors. Can you share the badge links? 

That particular error message means that the API response does not match the schema (typically due to a field required in the schema missing in the response), so in order to fix that the schema will have to be updated

flyboarder:
@calebcartwright could you look at the schema and let me know what else I need to fix?

calebcartwright:
Yup, will take a look in a few hours

services/clojars/clojars-base.js
```diff
@@ -7,9 +7,11 @@ const { BaseJsonService } = require('..')
...
 const clojarsSchema = Joi.object({
   downloads: nonNegativeInteger,
-  latest_version: Joi.string().required(),
-  latest_release: Joi.string().required(),
-}).required()
+  latest_version: Joi.string(),
+  latest_release: Joi.string(),
+})
+  .required()
+  .xor('latest_release', 'latest_version')
```
calebcartwright: I don't think we can/should use any exclusive clauses here, as it seems like a very common scenario for both of these fields to exist (this is why all the live tests are failing)

```json
{
  "jar_name": "prismic",
  "group_name": "prismic",
  "homepage": "http://prismic.io",
  "description": "Clojure kit for prismic.io API",
  "user": "erwan",
  "latest_version": "1.1",
  "latest_release": "1.1",
  "recent_versions": [...],
  "downloads": 126
}
```

calebcartwright: My assumption is that `latest_version` will always exist in the response for every valid jar (non-existent jars would return a 404 error) so it may just be a matter of removing that `xor` chain, keeping `latest_release` as optional, and making `latest_version` required

athos:
Hi, there. I've also run into this issue yesterday.

Here is an problematic response example:

- https://clojars.org/api/artifacts/kitchen-async

As you can see, the `latest_release` field may be `null` in a certain situation, so I think the schema for it should be `Joi.string().allow(null)` or something like that.

services/clojars/clojars-base.js
paulmelnikow: There is also [`.or()`](https://github.com/hapijs/joi/blob/master/API.md#objectorpeers) which would make one or the other required, but both valid.

calebcartwright:
Not sure what's going on with Circle ATM, random npm install failures

calebcartwright:
Deployed to review app: https://shields-staging-pr-3187.herokuapp.com/

calebcartwright:
Everything's looking good now in the review app:

re-find - https://shields-staging-pr-3187.herokuapp.com/clojars/v/re-find.svg


kitchen-async - https://shields-staging-pr-3187.herokuapp.com/clojars/v/kitchen-async.svg


The last thing I'm wondering about is whether the "release" labeling is still the best, since it's really more of a latest version, regardless of whether that's pre or not. Is there a scenario where folks would want to have a badge that showed only the latest _release_ version?

flyboarder:
@calebcartwright I think the terms used by Clojars are misleading.

Latest Version - could have snapshots if they are available.
Latest Release - only has stable versions.

> The last thing I'm wondering about is whether the "release" labeling is still the best. With this change, this badge really becomes more of a latest version, regardless of whether that's pre or not. Is there a scenario where folks would want to have a badge that showed only the latest _release_ version?

I think it makes sense for a developer to only want to show stable release versions. In clojure, generally snapshot versions are not recommended for use as the package could change without notice. So I think having to explicitly opt into showing them makes sense.

calebcartwright:
> With this change, the "Release" version will only show a snapshot if a stable version does not exist 

Understood, and that's what's got me wondering if there's a desire to have a badge that will only ever show the latest release version. 

> maintains behavioural compatibility with the legacy endpoint which was used for badges prior to the change

Agreed. I can go ahead and merge this if that's what folks would prefer? If there's a desire for a latest release-only badge, that can be handled in a follow up

Issue #3174, opened 2019-03-07
Improvements to badge builder and listing


paulmelnikow:
I'd like to propose some design changes to the badge builder and listing. There are two goals:

1. Make the badge listing easier to read:
      1. The entries should hint at what is inside, at the expense of documenting every possible option.
      2. Overall, the listing should avoid visual noise and reduce redundancy, at the expense of not showing full URLs.
2. Provide a richer experience on click:
      1. Include multiple examples.
      2. Allow any combination of options.

To achieve this, the idea is to aim to consolidate on a **single pattern per service class with multiple examples**.

- Continuing the direction of #2882, this implies a bit more rigidity on what is combined into one service, and what is kept separate. Some services may need to be combined or split up. (This can be done incrementally.)
- Some routes may need new redirectors so the main service can use a pattern.
- It's possible a few edge cases will require redirectors that have some degree of route prioritization. (I forget which these are. codeclimate?)

**Changes to the builder**

1. Optional parameter called out as such (already implemented but not turned on)
2. Show "optional" next to optional params.
3. Allow boolean params to default to either true or false.
4. Supports any combination of options.
5. Show a list of one or more filled-in examples.

**Changes to the listing**

1. Suppress the base URL.
2. Suppress the extension (partly in anticipation of #2674, but that would be a separate effort).
3. Suppress any optional params.
4. Suppress multiple-choice patterns. `:interval(dt|dw)` becomes `:interval`.

**Builder, combined into one:**

<img width="823" alt="screen shot 2019-03-06 at 10 28 40 pm" src="https://user-images.githubusercontent.com/1487036/53930428-382c2a00-405f-11e9-8071-dc649601f717.png">

**List of examples:**

The bottom of the model would show a bulleted list of examples, with full paths:

- `/azure-devops/tests/azuredevops-powershell/azuredevops-powershell/1.svg`
- `/azure-devops/tests/azuredevops-powershell/azuredevops-powershell/1/master.svg`
- `/azure-devops/tests/azuredevops-powershell/azuredevops-powershell/1.svg?compact_message`
- `/azure-devops/tests/azuredevops-powershell/azuredevops-powershell/1.svg?failed_label=bad&passed_label=good&skipped_label=n%2Fa`

As a "stretch goal," each of the URLs could have a **Preview** button next to it that updates the fields in the path and query string, and repopulates the live preview.

**Listing:**

[...]

PR #3189, opened 2019-03-11
Refactor [Sonar]

### LINKED ISSUE #1075: Badge Request: sonarcloud.io
[SonarCloud](https://sonarcloud.io) is online service base on SonarQube, Do you have any plan for support it?

calebcartwright:
Close #1075

Things to note:
- Our badges will work with the SaaS [SonarCloud.io](https://sonarcloud.io/projects) and any other SonarQube deployment that can be reached
  - SonarCloud provides their own badges, but still good to have Shields support for consistent look and feel, customization/enhancement (many of the native SonarCloud badges use a dark grey color regardless of value), etc.
  - Added badge documentation to clearly explain that we work with both
- The legacy service had the optional `version` param as a route param prefix. I moved `version` from route param to query param, with a redirector service to prevent breaking existing badges
  - The Sonar API contract changed dramatically with version 5.4. For prior versions, we need to use the legacy API contract, and any user still using Sonar 5.3 or earlier can get badges working by specifying a value of <= `5.3` for the `version` param so we know to use the legacy api. The latest version of Sonar is now 7.6 so I believe that the `version` parameter (only needed for Sonar 5.3 and below) falls under the "not typically used" category, making it a better fit for a query param.
  - Added badge documentation to explain this
- There was a good chunk of untested code in the legacy service implementation, which I've attempted to address here
  - The Fortify Security Rating badge is using mocked tests, as I don't know of any live endpoint that can be used for test (not available on SonarCloud.io and no known SonarQube deployments with the Fortify plugin installed and in use). I added an inline comment in that tester for future awareness
- ~~Added support for additional Sonar metrics (our legacy service recognized a few, but for the others would just display the raw value with brightgreen color)~~
  - ~~Examples of "new" badge support: Quality Gate, Bugs, Code Smells, Duplicated Lines %, Lines of Code, Maintainability, Reliability, Security, Vulnerabilities, and Issues~~
  - The refactor alone will result in a decent sized diff, so I've decided to push adding these new metrics to future follow ups, and have opened https://github.com/badges/shields/issues/3236 to track that follow up work
  - I went ahead and implemented new support QualityGate badges to serve as a reference


calebcartwright:
https://docs.sonarqube.org/latest/user-guide/metric-definitions/

calebcartwright:
Alright I think I've figured out a way in which I can split up refactoring the existing Sonar badges and handling all the others in a follow up. Will open an issue to track that latter one shortly and then finish off the last couple pieces with the refactor here

calebcartwright:
I know this has been running for a long time (it seems to keep spawning off new threads of work 😄) but I _think_ it's finally just about done. 

All that's left is to add the redirector service (to convert `version` to a query param to maintain support for old SonarQube versions) and update some the badge examples.

I'll mark the PR ready for review once I'm done with those, but it's probably ready now for some new eyes to review if anyone is up for looking at the rest of it.

services/sonar/sonar-generic.service.js
```diff
@@ -0,0 +1,152 @@
...
+// The legacy implementation simply rendered a brightgreen badge with the value,
+// regardless of what the value actually was.
+//
+// See https://github.com/badges/shields/issues/3236 for more information.
+//
+// This service should gradually be replaced by services that handle
+// their respective metrics by providing badges with more context
+// (i.e. a red/error badge when there are multiple security issues)
+//
+// https://docs.sonarqube.org/latest/user-guide/metric-definitions
```
calebcartwright: I hope this provides sufficient info behind why this service exists. Definitely open to any suggestions here though, as I'm likely too close to this

services/sonar/sonar-generic.service.js
```diff
@@ -0,0 +1,152 @@
...
+  complexityMetricNames,
+  duplicationMetricNames,
+  issuesMetricNames,
+  maintainabilityMetricNames,
+  reliabilityMetricNames,
+  securityMetricNames,
+  sizeMetricNames,
+  testsMetricNames
+)
+const metricNameRouteParam = metricNames.join('|')
```
calebcartwright: I elected to enumerate the various metrics in groups based on how Sonar docs typically group those metrics. I've still not had a chance to review _each_ metric in detail, but I think that when implementing #3236 a reasonable starting point would be a service class for each group (I expect there will be a lot of similarity). Some of those metrics (like the coverage ones) could probably be tacked into one of the existing services

services/sonar/sonar-generic.service.js
```diff
@@ -0,0 +1,152 @@
...
+
+  static get defaultBadgeData() {
+    return { label: 'sonar' }
+  }
+
+  static render({ metricName, metricValue }) {
+    return {
+      label: getLabel({ metric: metricName }),
+      message: metric(metricValue),
+      color: 'brightgreen',
```
calebcartwright: This is certainly unusual, but it is consistent with the legacy service behavior. I'm inclined to change it to `informational` but curious what others think

services/sonar/sonar-generic.tester.js
```diff
@@ -0,0 +1,11 @@
...
+
+const { isMetric } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Security Rating')
+  .get('/https/sonarcloud.io/com.luckybox:luckybox/security_rating.json')
+  .expectBadge({
+    label: 'security rating',
+    message: isMetric,
+  })
```
calebcartwright: Debated whether I should add generic tests like this one for the other 60+ metrics. I decided against it adding them here as part of this PR since it was really just supposed to be a refactor. 

Sonar code coverage was < 50% beforehand, but is now 100% with tests for all of the metrics we provide first class badges for (the ones in the other service classes) so I think we're going to be in a much better place Sonar validation wise. 

I'd rather add tests for the other metrics when they are enhanced as part of #3236

services/sonar/sonar-quality-gate.service.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'quality gate' }
+  }
+
+  static render({ qualityState }) {
+    if (qualityState === 'OK') {
+      return {
+        message: 'passed',
```
calebcartwright: I almost used our build status messages (`passing` and `failing`) here, but decided to go with the wording Sonar uses for the quality gate metric since this isn't a build status badge

services/sonar/sonar-quality-gate.service.js
```diff
@@ -0,0 +1,47 @@
...
+    return {
+      message: 'failed',
+      color: 'critical',
+    }
+  }
+
+  static get route() {
+    return {
+      base: 'sonar',
+      pattern: `${patternBase}/:metric(quality_gate|alert_status)`,
```
calebcartwright: The metric sonar uses under the hood is `alert_status` but in all the UI/reporting this is presented to users as Quality Gate. I added support for both in the route as I think `quality_gate` will make more sense to the average user, but also kept `alert_status` since some users may use that instead

calebcartwright: The metric sonar uses under the hood is `alert_status` but in all the UI/reporting this is presented to users as Quality Gate. I added support for both in the route as I think `quality_gate` will make more sense to the average user, but also kept `alert_status` since some users may use that instead

services/sonar/sonar-tech-debt.service.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  async handle({ protocol, host, component, metric }, { version }) {
+    const json = await this.fetch({
+      version,
+      protocol,
+      host,
+      component,
+      //special condition for backwards compatibility
+      metricName: 'sqale_debt_ratio',
```
calebcartwright: This was replicated from the legacy service implementation, but I'm not 100% convinced it is needed in all cases (at least not for newer versions of Sonar). Need to do some more reading on this

services/sonar/sonar-violations.service.js
```diff
@@ -0,0 +1,146 @@
...
+module.exports = class SonarViolations extends SonarBase {
+  static get category() {
+    return 'analysis'
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'violations' }
+  }
+
+  static renderLongViolationsBadge(violations) {
```
calebcartwright: This is a new badge, but one I think was worth adding as it provides an option for users that we offer in many other similar badges

services/sonar/sonar-base.spec.js
```diff
@@ -0,0 +1,4 @@
+'use strict'
+
```
calebcartwright: At some point I thought I was going to need some traditional unit tests but no longer appears to be the case 🤔 

Will probably delete this as part of some final cleanup

services/sonar/sonar-helpers.js
```diff
@@ -0,0 +1,56 @@
...
+const patternBase = ':protocol(http|https)/:host(.+)/:component(.+)'
+const ratingPercentageScaleSteps = [10, 20, 50, 100]
+const ratingScaleColors = [
+  'brightgreen',
+  'yellowgreen',
+  'yellow',
+  'orange',
+  'red',
+]
+const badMetricColorScale = colorScale(
```
calebcartwright: there's probably better names for these two color scales but these sufficed my needs 😆

services/sonar/sonar.legacyservice.js
```diff
@@ -0,0 +1,267 @@
+// 'use strict'
+
```
calebcartwright: Kept this file around for easier side-by-side comparisons in my local env, but will be deleting it (and the tester) shortly

services/sonar/sonar-base.js
```diff
@@ -0,0 +1,91 @@
...
+        user: serverSecrets.sonarqube_token,
+      }
+    }
+
+    return this._requestJson({
+      schema: useLegacyApi ? legacyApiSchema : schema,
+      url,
+      options,
+      errorMessages: {
+        404: 'component or metric not found, or legacy API not supported',
```
calebcartwright: This makes for a pretty verbose error, but unfortunately there's no good way to distinguish between them; Sonar gives us a 404 in each case

services/sonar/sonar-tech-debt.service.js
calebcartwright: Leaving this as-is for now

services/sonar/sonar-base.js
```diff
@@ -0,0 +1,91 @@
...
+const serverSecrets = require('../../lib/server-secrets')
+const { BaseJsonService } = require('..')
+const { isLegacyVersion } = require('./sonar-helpers')
+
+const schema = Joi.object({
+  component: Joi.object({
+    measures: Joi.array()
+      .items(
+        Joi.object({
+          metric: Joi.string(),
```
chris48s: Should this be `.required()`

services/sonar/sonar-base.js
```diff
@@ -0,0 +1,91 @@
...
+  }).required(),
+}).required()
+
+const legacyApiSchema = Joi.array()
+  .items(
+    Joi.object({
+      msr: Joi.array()
+        .items(
+          Joi.object({
+            key: Joi.string(),
```
chris48s: Also here. Should it be `.required()` ?

services/sonar/sonar-coverage.tester.js
```diff
@@ -0,0 +1,22 @@
...
+  })
+
+t.create('Coverage (legacy API supported)')
+  .get(
+    '/http/sonar.petalslink.com/org.ow2.petals%3Apetals-se-ase/coverage.json?version=4.2'
+  )
+  .expectBadge({
+    label: 'coverage',
+    message: isIntegerPercentage,
+  })
```
chris48s: Quite a few of these test files don't include a test for the "not found" case. Given we're supporting lots of versions of an API that may be self- hosted, I think the role of the 'not found' test as a canary for an upstream API changes is less clear, but do you reckon its worth adding one test just on the base class that hits `sonarcloud.io`?

services/sonar/sonar-generic.service.js
```diff
@@ -0,0 +1,152 @@
...
+  'skipped_tests',
+  'uncovered_conditions',
+  'new_uncovered_conditions',
+  'uncovered_lines',
+  'new_uncovered_lines',
+  'tests',
+  'test_execution_time',
+  'test_errors',
+  'test_failures',
+  'test_success_density',
```
chris48s: Do we actually need to explicitly whitelist every possible metric value here?

If the _only_ thing we're gaining here is a more specific regexion the URL, I'd be happy to just accept any string in the URL schema and throw a `NotFound()` or something if the key gives us an error

services/sonar/sonar-generic.service.js
```diff
@@ -0,0 +1,152 @@
...
+]
+const metricNames = [].concat(
+  complexityMetricNames,
```
chris48s: Minor style point, but if we keep this (see above comment) you could write this statement more nicely using the spread operator :)

services/sonar/sonar-generic.service.js
chris48s: Definitely agree. Lets change it to blue/informational

services/sonar/sonar-coverage.tester.js
```diff
@@ -0,0 +1,22 @@
...
+    '/http/sonar.petalslink.com/org.ow2.petals%3Apetals-se-ase/coverage.json'
+  )
+  .expectBadge({
+    label: 'coverage',
+    message: isIntegerPercentage,
+  })
+
+t.create('Coverage (legacy API supported)')
+  .get(
+    '/http/sonar.petalslink.com/org.ow2.petals%3Apetals-se-ase/coverage.json?version=4.2'
```
chris48s: Not sure quite where to leave a comment for it, but "version" is quite a loaded word. Do you think there is any ambiguity about what the `version` param means in this context?

i.e: `version` here refers to the version of SonarQube server I'm running as opposed to the version of my software (in the sense of a tag on a repository). Do you think it is worth changing the param name to make that more clear or noting that in the docs, or is that not an ambiguity we need to consider?

Happy to defer to your view on this as I'm not deeply familiar with SonarQube

services/sonar/sonar-helpers.js
chris48s: My suggestion would be `positiveMetric` / `negativeMetric`. Do you think that is any better?

services/sonar/sonar-documented-api-density.service.js
```diff
@@ -0,0 +1,48 @@
...
+const {
+  patternBase,
+  queryParamSchema,
+  getLabel,
+  goodMetricColorScale,
+} = require('./sonar-helpers')
+
+const metric = 'public_documented_api_density'
+
+module.exports = class SonarDocumentedApiDensity extends SonarBase {
```
chris48s: You haven't defined any examples on this class so it won't appear on the index page

services/sonar/sonar-fortify-rating.service.js
```diff
@@ -0,0 +1,51 @@
...
+const colorMap = {
+  0: 'red',
+  1: 'orange',
+  2: 'yellow',
+  3: 'yellowgreen',
+  4: 'green',
+  5: 'brightgreen',
+}
+
+module.exports = class SonarFortifyRating extends SonarBase {
```
chris48s: This one is also missing examples

services/sonar/sonar-generic.service.js
```diff
@@ -0,0 +1,152 @@
...
+  issuesMetricNames,
+  maintainabilityMetricNames,
+  reliabilityMetricNames,
+  securityMetricNames,
+  sizeMetricNames,
+  testsMetricNames
+)
+const metricNameRouteParam = metricNames.join('|')
+
+module.exports = class SonarGeneric extends SonarBase {
```
chris48s: Again, there are no examples for this one either, although given its for legacy compatibility maybe it was intentional in this case?

services/sonar/sonar-coverage.tester.js
chris48s: Also, the `version` param doesn't seem to be working at all on staging for any badges. If I call for example

https://shields-staging-pr-3189.herokuapp.com/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/coverage.svg?version=4.2 or https://shields-staging-pr-3189.herokuapp.com/sonar/https/sonarcloud.io/swellaby:azdo-shellcheck/quality_gate.svg?version=4.2

I get "invalid query parameter: version"

services/sonar/sonar-violations.service.js
```diff
@@ -0,0 +1,198 @@
...
+          metric: 'violations',
+        },
+        queryParams: {
+          format: 'short',
+          version: '4.2',
+        },
+        staticPreview: this.render({
+          violations: 0,
+          metricName: 'violations',
+          format: 'short',
```
chris48s: Not sure what the best solution is for this - it might be a problem in the base class. Your examples work fine unless I try to change the params, so the "short" example works and the "long" example works, but if I click the "short" example and try to change it to "long" in the editor, it still shows the short version:

Screenshot at 2019-04-13 19-38-08

chris48s: Also, these examples generate a really really long example URL:

Screenshot at 2019-04-13 19-41-58

Do you think it is worth aliasing the URL to just `:metric` in the examples given that all the many different options are discoverable via the drop-down in the editor?

calebcartwright:
Thanks @chris48s. This one really turned into a beast of a dif so I really appreciate the feedback!

services/sonar/sonar-helpers.js
calebcartwright: yes indeed 👍 will update them accordingly

services/sonar/sonar-base.js
calebcartwright: I want to say I ran into an issue with the API response occasionally had the measures array that would have one or two items that did not have a `metric` property, resulting in `invalid response data`. It's been a while though so I'll double check

services/sonar/sonar-coverage.tester.js
calebcartwright: My personal preference would have been to include a `not found` case in each service tester, but in some previous PR reviews there was pushback on those cases from other maintainers when the services all used the same `fetch` functions in the same base class. 

As such I only added one live test case validating the `not found` scenario on sonarcloud.io that is in the `fortify` [tester](https://github.com/badges/shields/blob/256c0d07664e332bf7fb360803f4de2a2b45af1d/services/sonar/sonar-fortify-rating.tester.js#L83) file

services/sonar/sonar-documented-api-density.service.js
calebcartwright: Understood, and that was done intentionally based on what was included/excluded from the Sonar example listings originally. 

I meant to pose a question here though on examples, as I'm not sure what the best approach is. Should we have an example for every single one? Should we maintain existing behavior of only listing a handful of examples for the more core/well-known metrics?

services/sonar/sonar-violations.service.js
calebcartwright: Lol, that's a pretty ugly url. Happy to change it.

services/sonar/sonar-coverage.tester.js
calebcartwright: 🤔 not sure off hand but looking into it. it was working at one point (albeit admittedly another classic case of "works on my machine")

calebcartwright: > Do you think there is any ambiguity about what the version param means in this context?

I see what you mean. Honestly this is really more of a boolean determination anyway to answer the question: "is the SonarQube instance old (5.3 or earlier)". My personal preference would have been to have this as a boolean query param like `useLegacyApi`. 

However, I'm hesitant to change it this point unless we can find a clean way to maintain backwards compatibility without requiring a redirect every time (this could technically be done with a catch all redirector with a `transformQueryParam` value but I'm not keen on that approach)

services/sonar/sonar-generic.service.js
calebcartwright: This may very well have been an unrelated user error on my part, but I was having some routing issues locally and I thought it may have been due to having a conflict between the services that explicitly declare their metric(s) (like `coverage`) and an identical route here but with a catch all metric. 

i.e.:
`sonar/:protocol(http|https)/:host(.+)/:component(.+)/coverage`

and 

`sonar/:protocol(http|https)/:host(.+)/:component(.+)/:metric`

since we should only hit the generic catch all if the specific metric is not defined explicitly in another service class' route

services/sonar/sonar-violations.service.js
calebcartwright: > Not sure what the best solution is for this - it might be a problem in the base class. Your examples work fine unless I try to change the params, so the "short" example works and the "long" example works, but if I click the "short" example and try to change it to "long" in the editor, it still shows the short version

I think that's because the specified component has zero violations. Both the long and short format render the same for zero violations, but if you use a different component that has violations you can toggle between them:

https://shields-staging-pr-3189.herokuapp.com/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/violations.svg?format=long


https://shields-staging-pr-3189.herokuapp.com/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/violations.svg?format=short

services/sonar/sonar-coverage.tester.js
calebcartwright: I believe this should be set now. I've renamed the query param to `sonarVersion` to be more explicit, and I fixed the previous issue with the query param (was related to the query param schema validation):

https://shields-staging-pr-3189.herokuapp.com/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/coverage.svg?sonarVersion=4.2


https://shields-staging-pr-3189.herokuapp.com/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/quality_gate.svg?sonarVersion=5.4


https://shields-staging-pr-3189.herokuapp.com/sonar/https/sonarcloud.io/swellaby:azdo-shellcheck/quality_gate.svg?version=7.7

services/sonar/sonar-generic.service.js
chris48s: Right yeah I see the problem. We want to avoid making a breaking change and need to avoid defining two conflicting or ambiguous routes :/

services/sonar/sonar-coverage.tester.js
chris48s: Nice - this is looking good now. With regard to treating it as a boolean for current/legacy, I think you're right to explicitly pass a version number. Thinking ahead, a future versions might introduce more API changes so we could in principle need to support more than 2 formats.

services/sonar/sonar-violations.service.js
chris48s: You're right about the long/short param - I was just looking at a bad example :)

You're also right about shortening the URL example. The popup will use the route we define in the example definiteion. Lets leave it how it is and punt the job of thinking about how to deal with displaying (long) URLs to https://github.com/badges/shields/issues/3174

services/sonar/sonar-documented-api-density.service.js
chris48s: Yeah I'm not really sure. For the `SonarGeneric` badge, I wouldn't want to explicitly list an individual example for every metric in the index, or show the URL in the index with

`:metric(new_violations|new_blocker_violations|new_critical_violations|new_major_violations|new_minor_violations.....)` or whatever in the URL pattern.

I think for that one we should probably list it with just `:metric` in the pattern and then explain in the documentation/help text how to find the valid values for `:metric` in the Sonar docs. Does that sound reasonable?

I'm less clear about `SonarDocumentedApiDensity` / `SonarFortifyRating`. If we're supplying a custom badge for it then I think we should probably list an example.

services/sonar/sonar-base.js
calebcartwright: Have made this required, and no issues 👍

services/sonar/sonar-base.js
calebcartwright: Have made this required, and no issues 👍

services/sonar/sonar-documented-api-density.service.js
calebcartwright: Will do

services/sonar/sonar-generic.service.js
calebcartwright: Exactly. I think my personal preference would be to remove this generic/catch-all support, even if it does potentially break anyone that's using one of those metrics. 

I'm struggling to see/understand the use case where this would be used, since it's basically just a pass through and ends up with badges that don't make a ton of sense:



If we want to maintain BC support for those badges, then this looks approach seems likely to be the "least bad" solution from my perspective. I suppose the silver lining is that this generic service is just a temporary/tactical solution until we can implement more traditional service classes for the other metrics

services/sonar/sonar-fortify-rating.service.js
calebcartwright: Example added

services/sonar/sonar-documented-api-density.service.js
calebcartwright: Example added

services/sonar/sonar-generic.service.js
calebcartwright: (quoted from above comment)
> I think for that one we should probably list it with just :metric in the pattern and then explain in the documentation/help text how to find the valid values for :metric in the Sonar docs. Does that sound reasonable?

I think I'd still be inclined to not add an example listing here. I don't want to promote/advertise these badges given their nature (just pass throughs) and since this exists for BC only. If we put this out there with an example listing and editor window then I would expect new users to be confused/surprised when they get badges like:

 

I suppose we could add some detailed docs on that example to attempt to explain the nature of those badges, but I'd prefer to add examples for the metrics once more traditional first class support is added: add example listings for Security Rating once we've implemented support to provide a badge that truly handles the security rating data.

I can add an example if we want, but wanted to articulate my concerns

services/sonar/sonar-generic.service.js
calebcartwright: Done

services/sonar/sonar-generic.service.js
calebcartwright: I'm second guessing myself now 😄 Changing these badge color to `blue`/`informational` does make them a lot more tolerable IMO. 

Let me add an example with documentation to see how it looks

calebcartwright: Okay example added with docs, and I like the way it looks

services/sonar/sonar-generic.service.js
chris48s: I think what we've done here is fine, but just reading your ~crossed out~ comments and merging in the conversation in https://github.com/badges/shields/pull/3189#discussion_r275127813 , I agree it would be nice not to have to explicitly take on support for the entire kitchen sink of Sonar functionality. Here's another idea to consider before we merge:

Although we theoretically allow _every sonar metric ever_ , in reality there is probably a relatively small subset of them in use in the real world. If we don't really want to support all of them, we could merge this but not include any examples for `SonarGeneric`. Once we convert the badge to the new service class style, we should be able to start picking up metrics for it (legacy badge traffic doesn't get picked up by prometheus). When we do that, we'll be able to understand which metrics are actually getting any real-word usage and then we could constrain the `SonarGeneric` to only support the metrics which get any traffic before we add it to the index page.

services/sonar/sonar-fortify-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Sonar Fortify Security Rating',
+        namedParams: {
+          protocol: 'http',
+          host: 'sonar.petalslink.com',
+          component: 'org.ow2.petals:petals-se-ase',
```
chris48s: this  `namedParams` object doesn't seem to generate a valid badge:

https://shields-staging-pr-3189.herokuapp.com/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/fortify-security-rating.svg

calebcartwright: Yeah, we don't know of any live Sonar endpoints that have the Fortify plugin installed as noted above in the PR summary.

The SaaS service sonarcloud.io doesn't have nor does http://sonar.petalslink.com. We don't have a live endpoint that we can use for the example, nor for the tests (that's why they're all mocked)

services/sonar/sonar-generic.service.js
calebcartwright: I'm good with that too

services/sonar/sonar-fortify-rating.service.js
chris48s: Aah yes - you did mention this elsewhere in the PR :+1:

calebcartwright: Should I remove the example listing? I'm a fan of the editor experience it provides, but my instinct says that there's not a ton of folks using this badge today.

Alternatively, I could update the documentation. It currently calls out that it def. won't work on SonarCloud, but I could update that to note that the Fortify stuff comes from a Sonar plugin that's not available OOTB

services/sonar/sonar-generic.service.js
chris48s: OK. I'll defer to your judgement to either keep the examples in or remove them with the intention that we can look at the (Prometheus) metrics and decide later.

Beyond that, I think everything else is resolved and this is the last thing outstanding, so I'll aim to come back to this and give it one final read over tomorrow evening.

calebcartwright: Sorry I'm multi tasking but should have been more explicit.

I like your proposal and will make that change shortly. I think your hypothesis is correct as well, and I'd guess there's only a handful of "other" sonar metrics being used... although I'm regularly surprised by what the Shields metrics show 😄

services/sonar/sonar-fortify-rating.service.js
chris48s: I think its ok. If we ever find a public instance with the plugin, it would be nice to update it.
We can review if we find there are hoards of confused Sonar users opening issues.

calebcartwright:
> I think this is good to merge

Thanks for all the feedback!

PR #3198, opened 2019-03-11
refactor [githubsize]

chris48s:


services/github/github-size.service.js
```diff
@@ -1,21 +1,20 @@
...
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class GithubSize extends LegacyService {
+const schema = Joi.alternatives(
+  Joi.object({
+    size: nonNegativeInteger,
+  }).required(),
+  Joi.array().required()
```
calebcartwright: This is so we can provide a more user friendly error message (`not a regular file`) vs. rejecting it at the schema level with the `invalid response data` error right?

chris48s: yep - exactly

PR #3199, opened 2019-03-11
refactor [githublastcommit]

chris48s:


services/github/github-last-commit.service.js
```diff
@@ -57,50 +54,40 @@ module.exports = class GithubLastCommit extends LegacyService {
...
+    }
+  }
+
+  async fetch({ user, repo, branch }) {
+    const options = { qs: {} }
+    if (branch) {
+      options.qs.sha = branch
+    }
+    return this._requestJson({
+      url: `/repos/${user}/${repo}/commits`,
```
paulmelnikow: You can write `options: { qs: { sha: branch } }` without a conditional and the result should be the same.

Issue #3220, opened 2019-03-19
Jitpack Downloads Badge


erayerdin:
:clipboard: **Description**

This badge is for Jitpack's downloads data. Shields already have a Jitpack badge, yet it doesn't
have download count. Jitpack apparently provides [downloads endpoint](https://jitpack.io/docs/API/#downloads) and it can be used to provide a badge.

This badge should be similar to other downloads badges, such as:

 - **Monthly Badge / Great Amount of Downloads Example:** ![](https://img.shields.io/badge/downloads-10k%2Fmonth-brightgreen.svg)
 - **Weekly Badge / No Downloads Example:** ![](https://img.shields.io/badge/downloads-0%2Fweek-red.svg)

:link: **Data**

The API is public and does not require authentication except it is a private repository (Jitpack version badge also only supports public repositories only). You can see the documentation [here](https://jitpack.io/docs/API/#downloads).

:microphone: **Motivation**

Any other packaging platforms have download badges but Jitpack does not.

calebcartwright:
Added `needs-upstream-help` label as the underlying Jitpack API needs to stabilize before this can be completed

erayerdin:
Seems #3291 has been merged. I think it is appropriate to close this issue @calebcartwright.

PR #3221, opened 2019-03-19
Added [Jitpack] Downloads Counter

erayerdin:
See #3220 for example badges.

### DIFF
--- a/services/jitpack/jitpack-downloads.service.js
+++ b/services/jitpack/jitpack-downloads.service.js
@@ -22,18 +22,16 @@ const periodMap = {
   },
 }
 
-const groupIdMatcher = /^[a-z]+\..+\..+$/
-
 module.exports = class JitpackDownloads extends BaseJsonService {
   static get examples() {
     return [
       {
         title: 'JitPack - Downloads',
         namedParams: {
-          groupId:
-            'jitpack (github username) or com.github.jitpack (full groupID)',
-          artifactId: 'maven-simple',
           period: 'dm',
+          vcs: 'github',
+          user: 'jitpack',
+          repo: 'maven-simple',
         },
         staticPreview: JitpackDownloads.render({
           downloads: 14000,
@@ -55,30 +53,22 @@ module.exports = class JitpackDownloads extends BaseJsonService {
   static get route() {
     return {
       base: 'jitpack',
-      pattern: ':period(dw|dm)/:groupId/:artifactId',
+      pattern: ':period(dw|dm)/:vcs(github|bitbucket|gitlab)/:user/:repo',
     }
   }
 
-  async handle({ period, groupId, artifactId }) {
-    const json = await this.fetch({ groupId, artifactId })
+  async handle({ period, vcs, user, repo }) {
+    const json = await this.fetch({ vcs, user, repo })
     return this.constructor.render({
       downloads: json[periodMap[period].api_field],
       period,
     })
   }
 
-  async fetch({ groupId, artifactId }) {
-    let url = null
-
-    if (groupId.match(groupIdMatcher)) {
-      url = `https://jitpack.io/api/stats/${groupId}/${artifactId}`
-    } else {
-      url = `https://jitpack.io/api/stats/com.github.${groupId}/${artifactId}`
-    }
-
+  async fetch({ vcs, user, repo }) {
     return this._requestJson({
       schema,
-      url,
+      url: `https://jitpack.io/api/stats/com.${vcs}.${user}/${repo}`,
       errorMessages: { 401: 'project not found or private' },
     })
   }
--- a/services/jitpack/jitpack-downloads.service.js
+++ b/services/jitpack/jitpack-downloads.service.js
@@ -30,7 +30,8 @@ module.exports = class JitpackDownloads extends BaseJsonService {
       {
         title: 'JitPack - Downloads',
         namedParams: {
-          groupId: 'jitpack',
+          groupId:
+            'jitpack (github username) or com.github.jitpack (full groupID)',
           artifactId: 'maven-simple',
           period: 'dm',
         },
--- a/services/jitpack/jitpack-downloads.service.js
+++ b/services/jitpack/jitpack-downloads.service.js
@@ -1,0 +1,81 @@
+'use strict'
+
+const Joi = require('joi')
+const { nonNegativeInteger } = require('../validators')
+const { BaseJsonService } = require('..')
+const { downloadCount } = require('../color-formatters')
+const { metric } = require('../text-formatters')
+
+const schema = Joi.object({
+  week: nonNegativeInteger,
+  month: nonNegativeInteger,
+}).required()
+
+const periodMap = {
+  dw: {
+    api_field: 'week',
+    suffix: '/week',
+  },
+  dm: {
+    api_field: 'month',
+    suffix: '/month',
+  },
+}
+
+module.exports = class JitpackDownloads extends BaseJsonService {
+  static get examples() {
+    return [
+      {
+        title: 'JitPack - Downloads',
+        namedParams: {
+          groupId: 'jitpack',
+          artifactId: 'maven-simple',
+          period: 'dm',
+        },
+        staticPreview: JitpackDownloads.render({
+          downloads: 14000,
+          period: 'dm',
+        }),
+        keywords: ['java', 'maven'],
+      },
+    ]
+  }
+
+  static get category() {
+    return 'downloads'
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'downloads' }
+  }
+
+  static get route() {
+    return {
+      base: 'jitpack/d',
+      pattern: ':groupId/:artifactId/:period(dw|dm)',
+    }
+  }
+
+  async handle({ groupId, artifactId, period }) {
+    const json = await this.fetch({ groupId, artifactId })
+    return this.constructor.render({
+      downloads: json[periodMap[period].api_field],
+      period,
+    })
+  }
+
+  async fetch({ groupId, artifactId }) {
+    return this._requestJson({
+      schema,
+      url: `https://jitpack.io/api/stats/com.github.${groupId}/${artifactId}`,
+      errorMessages: { 401: 'project not found or private' },
+    })
+  }
+
+  static render({ downloads, period }) {
+    return {
+      message: `${metric(downloads)}${periodMap[period].suffix}`,
+      color: downloadCount(downloads),
+    }
+  }
+}
--- a/services/jitpack/jitpack-downloads.service.js
+++ b/services/jitpack/jitpack-downloads.service.js
@@ -22,6 +22,8 @@ const periodMap = {
   },
 }
 
+const groupIdMatcher = /^[a-z]+\..+\..+$/
+
 module.exports = class JitpackDownloads extends BaseJsonService {
   static get examples() {
     return [
@@ -65,9 +67,17 @@ module.exports = class JitpackDownloads extends BaseJsonService {
   }
 
   async fetch({ groupId, artifactId }) {
+    let url = null
+
+    if (groupId.match(groupIdMatcher)) {
+      url = `https://jitpack.io/api/stats/${groupId}/${artifactId}`
+    } else {
+      url = `https://jitpack.io/api/stats/com.github.${groupId}/${artifactId}`
+    }
+
     return this._requestJson({
       schema,
-      url: `https://jitpack.io/api/stats/com.github.${groupId}/${artifactId}`,
+      url,
       errorMessages: { 401: 'project not found or private' },
     })
   }
--- a/services/jitpack/jitpack-downloads.service.js
+++ b/services/jitpack/jitpack-downloads.service.js
@@ -51,12 +51,12 @@ module.exports = class JitpackDownloads extends BaseJsonService {
 
   static get route() {
     return {
-      base: 'jitpack/d',
-      pattern: ':groupId/:artifactId/:period(dw|dm)',
+      base: 'jitpack',
+      pattern: ':period(dw|dm)/:groupId/:artifactId',
     }
   }
 
-  async handle({ groupId, artifactId, period }) {
+  async handle({ period, groupId, artifactId }) {
     const json = await this.fetch({ groupId, artifactId })
     return this.constructor.render({
       downloads: json[periodMap[period].api_field],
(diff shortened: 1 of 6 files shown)


erayerdin:
Sorry @calebcartwright, I just forgot about those square brackets even though I've read about them in the documentation somewhere. :stuck_out_tongue_closed_eyes:

calebcartwright:
No worries!

services/jitpack/jitpack-downloads.service.js
```diff
@@ -0,0 +1,81 @@
...
+  static get category() {
+    return 'downloads'
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'downloads' }
+  }
+
+  static get route() {
+    return {
```
calebcartwright: I couldn't get the inline suggestion to work for multiple lines, but in order to be consistent with other download badges we should drop the `/d` from the base and move `period` up as the first parameter, i.e.
```js
    return {
      base: 'jitpack',
      pattern: ':period(dw|dm)/:groupId/:artifactId',
    }
```

services/jitpack/jitpack-downloads.service.js
```diff
@@ -0,0 +1,81 @@
...
+    return this.constructor.render({
+      downloads: json[periodMap[period].api_field],
+      period,
+    })
+  }
+
+  async fetch({ groupId, artifactId }) {
+    return this._requestJson({
+      schema,
+      url: `https://jitpack.io/api/stats/com.github.${groupId}/${artifactId}`,
```
calebcartwright: I'm not familiar with jitpack myself, but is the `com.github.` prefix something we should always inject into the url, or was that supposed to be part of a groupId?

services/jitpack/jitpack-downloads.service.js
erayerdin: Strangely, when I set `jitpack/v` to `jitpack` only, it apparently fails all tests (including the former Jitpack version implementation) probably because of [this line](https://github.com/erayerdin/shields/blob/master/services/jitpack/jitpack.service.js#L21).

That is why I've set it as `jitpack/d`, to be more consistent with the former version implementation.

Let me have a further look into this. Maybe I'm missing something?

erayerdin: All tests fail due to the errors similar to below:

```
AssertionError: expected { Object (name, value) } to deeply equal { Object (name, value) }                                                          
      + expected - actual

       {
         "name": "jitpack"
      -  "value": "inaccessible"
      +  "value": "project not found or private"
       }
```

Should I commit the failing codebase for you to have a look?

services/jitpack/jitpack-downloads.service.js
erayerdin: Okay, okay, I think I've done it. It was my stupidity. I was just passing `period` to `render` method without its name. I'll push the new commits regarding to the changes that you've requested.

erayerdin: I'm sorry. I meant it for the other review. Maybe I should sleep more. Whatever.

Back to the point, the former version implementation had the same structure. See [this](https://github.com/erayerdin/shields/blob/master/services/jitpack/jitpack.service.js#L47).

So I thought I should be consistent with it. Well, you are right at a point, what if a user deploys to Jitpack from another VCS host like Gitlab or Bitbucket. The problem is, if we change that behavior, people's Jitpack version badges might break. That wouldn't be good for them I think?

services/jitpack/jitpack-downloads.service.js
erayerdin: Okay, okay, I think I've done it. It was my stupidity. I was just passing `period` to `render` method without its name. I'll push the new commits regarding to the changes that you've requested.

services/jitpack/jitpack-downloads.service.js
calebcartwright: >  On the other hand, if we change that behavior, people's Jitpack version badges might break. That wouldn't be good for them I think

While it is definitely true that we never want to break any existing users, updating the url here in the downloads service has no impact on any other services, including the version badge. 

Currently, the existing version badge presents the prompts to the user as `group id` and `artifact id`, but I think what it's really expecting is a GitHub user and repo. This existing behavior wasn't exactly clear to me from the UI/badge path. 

If I enter the group id myself (`com.github.jitpack`) then I will get an error, and similarly if I was using Jitpack with Gitlab/Bitbucket and tried to create a badge (`com.gitlab.foo`) I would get errors and in my opinion the _real_ reason behind the error would not be clear to me as a user:

image

Given that Jitpack works with multiple VCS providers and that the Group Id does reflect the VCS (https://jitpack.io/docs/#other-git-hosts) I'd suggest three possible options:

1) Remove the manual `com.github` prefix we are adding to the new downloads badge. That way users could enter their respective group id's `com.github.jitpack`, `com.gitlab.foo`, `com.bitbucket.bar`, etc.
2) Update the route parameter names to more explicitly indicate that the user should enter a GitHub user name and repo (i.e. `:user/:repo`)
3) Add documentation to the examples on both the new Jitpack downloads badge and the existing Jitpack version badge to clearly convey that the badges will only work with GitHub 

I think I'd prefer option 1. Yes, going with option 1 would mean that in the short term users would have to enter the full group id for the downloads badge (`com.github.jitpack`) but just the GH user name for the version badge (`jitpack`). However, we could update the Jitpack version badge route in a follow up, including adding a redirect service to keep the existing badges working.

Curious what others think though

calebcartwright:
Deployed to review app @ https://shields-staging-pr-3221.herokuapp.com/

services/jitpack/jitpack-downloads.service.js
erayerdin: I considered option 1 makes sense but, yet, I was again thinking about the existing users. So I came up with an idea, I don't know what you'd think about it.

Check this [regex](https://regex101.com/r/DXEuGq/1). At first, we will check if this expression maches the given value. If so, then we will render the badge with the given value.

If it does not match, though, we will prepend `com.github.` prefix to the value.

With this method, it will not break the existing users' badges and will work with both at the same time. If you want, I can implement this method to both `JitPackVersion` and `JitPackDownloads` classes. What do you think?

calebcartwright: Thanks @erayerdin 

Will review in more detail in a little while, but remember the JitPackDownloads badge is brand-new. There are no existing users of the downloads badge/it is not possible to break any existing users. 

The JitPack Version badge is untouched and not impacted by the downloads badge

erayerdin: I have implemented it, it waits for your review.

erayerdin:
I will do a couple more changes, wait for it.

erayerdin:
I think that's all the changes I can do. Btw, latest fails happened because Jitpack API today have some issues I think. There were times that I couldn't reach it. Is it possible to start the circleci tests again?

calebcartwright:
The existing Shields Jitpack Version badge is going to have to be updated for the reasons we discussed above. As such, I don't want to replicate those problematic pieces and solutions in this new Jitpack Downloads badge (nor any potential other future Jitpack badges). We should build this new badge the way we want, and update the *version badge* to match it accordingly.

For the moment, let's set aside the existing version badge/pretend like it doesn't exist. If this Jitpack Downloads badge was the very first Jitpack badge, what do you think is the best badge-creating experience for Jitpack users @erayerdin? (don't worry about the code yet, just the user experience). 

Would it be something like this, where the user specifies his/her VCS type, then username, then repo?
image

Or would it be the user specifying his/her group id and artifact id?
image

Or something else entirely?

erayerdin:
> Would it be something like this, where the user specifies his/her VCS type, then username, then repo?

I thought it is a great idea, yet I somehow wondered if some usernames are *rendered differently*. Like, usernames like below:

 > some-bogus-user
 > someBogusUser
 > _someBogus-user123
 > 123Some-Bogususer

...are valid usernames but some of these are not valid `groupID`s and some are not convenient. So I've just sent an email to Jitpack about how they render such usernames and waiting for their reply. Let's decide on this after I get the reply from Jitpack.

paulmelnikow:
I like the idea of making it work with user / repo, and encoding that understanding into the service. Combined with a way to auto-detect jitpack, it could make it easier to suggest relevant badges given a repo, which is something I'd like to do as part of an improved frontend.

We could also support `/jitpack/dm/group/com.github.jitpack/maven-simple`, to handle difficult-to-encode usernames. "group" is not really a "vcs" though it doesn't seem the end of the world to shoehorn that in.

erayerdin:
Sorry, I've been busy for a week and seems like I've found a little while that I can continue implementing this feature.

I'm quoting the reply I've got:

 > > Is there any case that username in "groupID" and username in Github/Bitbucket/Gitlab are different?
 >
 > They will always be the same. We don't change the groupID even if the username is  invalid, as you say.

So, I think the implementation below would be great for user experience:

image

Now, is it OK if I implement this to both `JitpackVersion` and `JitpackDownloads`? Or should I just implement this to only `JitpackDownloads`?

calebcartwright:
@erayerdin 
> Sorry, I've been busy for a week and seems like I've found a little while that I can continue implementing this feature.

> I'm quoting the reply I've got:

No worries, thanks for confirming and sharing!

> Now, is it OK if I implement this to both JitpackVersion and JitpackDownloads
Yes, please go ahead and update both service classes accordingly

calebcartwright:
> calebcartwright closed this just now

Oops! hit the wrong button apologies

calebcartwright:
For the redirector service to maintain compatibility with the old version badge paths:

```js
'use strict'

module.exports = [
  redirector({
    category: 'version',
    route: {
      base: 'jitpack/v',
      pattern: ':groupId/:artifactId',
    },
    transformPath: ({ groupId, artifactId }) => `/jitpack/v/github/${groupId}/${artifactId}`,
    dateAdded: new Date('2019-03-31'),
  }),
]
```

erayerdin:
Okay, I did some changes but there're some things to say.

# About Download Endpoint of JitPack's API

JitPack's download endpoint seems to be quite problematic especially the last couple of days. I have many problems accessing it. So, if you see some CI processes fail, that might be because JitPack does not respond in time. It is highly likely that they deal with issues on weekdays. So, if CI fails, try Monday afternoon again.

# New Endpoints

These are patterns:

 - `/jitpack/:vcs(github|bitbucket|gitlab)/:user/:repo` for version.
 - `/jitpack/:period(dw|dm)/:vcs(github|bitbucket|gitlab)/:user/:repo` for downloads

# About Redirecting

I am really amazed and happy to see this feature in the codebase, really useful. Now, the pattern below:

 - `/jitpack/v/:groupId/:artifactId`

redirects to the new pattern.

While the old badge was clearly mentioning `groupId`, it was not a *real* `groupId` in terms of Java. You can check out what I'm currently using [here](https://github.com/erayerdin/kappdirs/blob/master/README.md).

Normally, `com.github.erayerdin` is a *real* `groupId`, but in shields, only username was used. In short, I just made some minor changes regarding to what you have suggested [here](https://github.com/badges/shields/pull/3221#issuecomment-478360478).

calebcartwright:
Excellent! I'll take another pass through this tomorrow, but just want to make sure the new path for the Version badge will include the `v` for version correct? 

`/jitpack/v/:vcs(github|bitbucket|gitlab)/:user/:repo`

erayerdin:
No it doesn't. I thought we would agree on base URL to be /jitpack/ for
both. I can change it though to /jitpack/v and /jitpack/d if you think that
would be more proper.

erayerdin:
Let me have a look at the Pypi stats (mostly because I'm more familiar with Python rather than Java) and have a decision.

erayerdin:
I have done everything I can and tests seem to pass, but I still, still get this badge:

image

`/api/stats/com.github.USER/REPO` does not respond and fails with timeout (yet the tests somehow magically passed). I have tried to contact @jitpack with every possible way I can, via email, Twitter, Gitter and so on, but they seem to be unresponsive.

So, my suggestion is to let this PR wait because `/api/stats` of JitPack API is, from what I can see, far from stable. And by *wait*, I mean for a long time, maybe a month, until I consider the `/api/stats` is stable.

During this time, I can consider whatever you (@calebcartwright) and the community suggest and apply them.

calebcartwright:
Both the pypi version badge, and npm version badge include the `v` prefix. Our badges follow a standard route format that includes the service name (be it pypi, npm, jitpack, etc.) along with the type of badge that is (downloads, versions, license, etc.) 

`pypi version`: https://img.shields.io/pypi/v/pytest.svg (note /pypi/**v**/...)
`npm version`: https://img.shields.io/npm/v/mocha.svg (note /npm/**v**/...)

When a service makes download counts available for different time periods, the badge identifier uses a `d*` prefix (like `dt` for total downloads, `dw` for weekly, etc.)

For example with npm downloads badge:
https://img.shields.io/npm/dt/mocha.svg (note /npm/**dt**/...)

And when services don't provide downloads in different time intervals, the badge routes will simply use a `d` or sometimes the more verbose `downloads`.

For example with the VS Marketplace downloads badge:
https://img.shields.io/visual-studio-marketplace/d/swellaby.rust-pack.svg (note /visual-studio-marketplace/**d**/...)

calebcartwright:
> /api/stats/com.github.USER/REPO does not respond and fails with timeout

Agreed that we need to wait on this. It does not make sense to deploy this new badge given the current state of the API

erayerdin:
> Both the pypi version badge, and npm version badge include the v prefix.

Sorry, I didn't see that. Then I will make regarding changes to `JitpackVersion`.

calebcartwright:
> Sorry, I didn't see that. Then I will make regarding changes to JitpackVersion.

No worries!

> And by wait, I mean for a long time, maybe a month, until I consider the /api/stats is stable.

Given the instability in that, what do you think about moving all the changes for JitpackVersion into a separate PR that we can go ahead and merge? We can then leave this PR to handle Downloads and come back to it once the API it uses has stabilized

erayerdin:
> what do you think about moving all the changes for JitpackVersion into a separate PR

Well, that would be pretty great. The thing is, I don't know if there's a way to separate `jitpack.service.js` and related other JS files to another branch.

calebcartwright:
@erayerdin - I'm going to close out this PR for now. Once the API has stabilized we can re-open this.

erayerdin:
I've been testing the endpoint for a long while now and it seems it is stabilized enough. I will rebase my `fix/jitpack-version` branch and will push some changes.

calebcartwright:
@erayerdin awesome! Feel free to reopen this PR or open a new one (whichever's more convenient for you)  and I'll be happy to take a look

erayerdin:
I think I will do this with a seperate branch and PR since it seems I messed up my merging process. Dammit. :laughing:

PR #3225, opened 2019-03-24
rewrite Github Languages, split into [githubcodesize githublanguagecount githubtoplanguage]

chris48s:
also more tests


services/github/github-languages-base.js
```diff
@@ -0,0 +1,28 @@
...
+const Joi = require('joi')
+const { GithubAuthService } = require('./github-auth-service')
+const { errorMessagesFor } = require('./github-helpers')
+
+/*
+Only validate the response is an object because we're expecting a response like
+{ "Python": 39624, "Shell": 104 }
+The keys could be anything and {} is a valid response (e.g: for an empty repo)
+*/
+const schema = Joi.object().required()
```
paulmelnikow: I think this could use something like `Joi.object().pattern(/./, Joi.number().integer().required())`.

services/github/github-languages-base.js
```diff
@@ -0,0 +1,28 @@
...
+  async fetch({ user, repo }) {
+    return this._requestJson({
+      url: `/repos/${user}/${repo}/languages`,
+      schema,
+      errorMessages: errorMessagesFor(),
+    })
+  }
+
+  getTotalSize(data) {
+    return Object.keys(data).reduce((acc, language) => acc + data[language], 0)
```
paulmelnikow: It's slightly simpler this way:

```suggestion
    return Object.values(data).reduce((acc, size) => acc + size, 0)
```

PR #3239, opened 2019-03-29
Fix [packagist] badges

### LINKED ISSUE #3238: Packagist badges: invalid response data

- \[x] [shields.io](https://img.shields.io/packagist/l/cse/helpers-json.svg?style=flat-square)
- [ ] My own instance

:beetle: **Description**

MIT Packagist license is not set. Display information: "invalid response data".

:link: **Link to the badge**
https://img.shields.io/packagist/l/cse/helpers-json.svg
[Packagist](https://github.com/cs-eliseev/helpers-json/blob/master/LICENSE.md)

calebcartwright:
Fix #3238 

- Moved definitions of schemas from base service class to their respective service classes (and simplified them accordingly), save for the one schema used in multiple places
- Updated version badge to re-add the logic used to properly select correct latest/pre version that existed in the legacy service implementation. I believe that when the legacy services were refactored that this was attempted to be handled within the schema, but that won't be able to cover all cases (as evident by the reported bugs)
- Added an additional live/integration test with packagist for a vpre badge

Note: some additional tests should be added in the future to cover some of the re-added logic in the transform function of the version badge service. However, I'd prefer to go ahead and merge this given the prod issues and add those new tests in a follow on

### DIFF
--- a/services/packagist/packagist-version.spec.js
+++ b/services/packagist/packagist-version.spec.js
@@ -1,20 +1,6 @@
 'use strict'
 
-const { expect } = require('chai')
-const { InvalidResponse } = require('..')
-const PackagistVersion = require('./packagist-version.service')
+// const { expect } = require('chai')
+// const PackagistVersion = require('./packagist-version.service')
 
-describe('PackagistVersion', function() {
-  it('throws InvalidResponse error when extra key is missing for vpre', function() {
-    try {
-      PackagistVersion.prototype.transform({
-        type: 'vpre',
-        json: { package: { versions: { 'dev-master': {} } } },
-      })
-      expect.fail('should have thrown')
-    } catch (e) {
-      expect(e).to.be.an.instanceof(InvalidResponse)
-      expect(e.prettyMessage).to.equal('prerelease version data not available')
-    }
-  })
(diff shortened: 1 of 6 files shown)


services/packagist/packagist-version.spec.js
```diff
@@ -0,0 +1,6 @@
+'use strict'
+
+// const { expect } = require('chai')
+// const PackagistVersion = require('./packagist-version.service')
+
+describe('PackagistVersion', function() {})
```
calebcartwright: This is where I was starting to add tests to cover some of the transform logic, but decided to hold off on that for now (see my comment in PR description).

kylekatarnls:
Thanks a lot for this fix!

But it seems still not deployed 😢

@calebcartwright @chris48s @paulmelnikow is something broken in the deployment process?

calebcartwright:
@kylekatarnls - no, nothing broken, just hasn't been deployed yet. The above [comment](https://github.com/badges/shields/pull/3239#issuecomment-478348161) will be updated once the changes have been deployed.

I don't have an exact timeline, but suspect it will be deployed fairly soon

PR #3240, opened 2019-03-30
[Drone] Add Drone build badge

### LINKED ISSUE #294: Badge request: drone.io
Is it possible to provide drone.io status shields?
There was issue #150, drone now serves shields that are more compatible to the normal style shields but still no svg.

byCedric:
Hi! I recently "discovered" drone, absolutely loved it, and I noticed it's still a feature request here (#294). This PR adds a basic SVG scraping badge to Shields. It's my first PR here, might have done some _weird sh*t(TM)_. 

Please, do tell me if I'm doing it wrong! I'm here to help.

**Why use the SVG scraper?**
Unfortunately, as [mentioned in the comments here](https://github.com/badges/shields/issues/294#issuecomment-155906875) Drone doesn't have a "good" API for fetching last build status based on branches. The one which is linked there works ([the beta-endpoint](http://beta.drone.io/api/repos/drone/drone/builds/latest?branch=0.4.0)), but not everyone has access to the beta. And in the latest stable version (`1.0.0`), this doesn't work as expected for specifying branches [as you can see here](https://cloud.drone.io/api/repos/drone/drone/builds/latest?branch=0.4.0).

**What did you do then?**
- I basically ripped off the Travis CI scraper and modified some code. The main difference is the API endpoints (of course) and returning `none` for errored, or `unknown` builds. 
- The `static get route()` part is copied from Bitbucket Pipelines, it looked more "understandable" then some regex I don't fully understand 😅.
- And for the last part, the branches specification, I used Drone's settings page to create a badge URL to a specific branch. They use this pattern `/refs/heads/{branch}` as `ref` query parameter. I feel that this is the best way to implement it.

Again, I tried to be as compliant as possible, keeping it nice and clean. The lint passed, ran prettier and tested everything. Unfortunately, on my machine, the tests aren't 100% completed successfully, but that's the code I cloned from `master`.

Hope it helps ❤️ 

Oh and for GitHub:
fixes #294


byCedric:
Sorry for the failed service tests and force pushed. I tried to get everything working but can't get the last error out. The error message is a bit confusing to me. If anyone knows where or what I should be fixing, I'm all ears 😄

services/drone/drone-build.tester.js
```diff
@@ -0,0 +1,40 @@
...
+  .expectBadge({
+    label: 'build',
+    message: Joi.alternatives().try(isBuildStatus, Joi.equal('none')),
+  })
+
+t.create('unknown repo')
+  .get('/build/this-repo/does-not-exist.json')
+  .expectBadge({ label: 'build', message: 'none' })
+
+t.create('invalid svg response')
```
calebcartwright: You can just remove this test (it is the one failing right now). It's not validating anything specific to the drone service class and the internals of the BaseSvgScraping service are validated elsewhere via other tests

services/drone/drone-build.tester.js
```diff
@@ -0,0 +1,40 @@
...
+  .get('/build/foo/bar.json')
+  .intercept(nock =>
+    nock('https://cloud.drone.io/api')
+      .get('/foo/bar.svg')
```
calebcartwright: FWIW though, I think the reason this is failing is because it's missing the `badges` part of the url path here, so nock isn't matching the actual request (`https://cloud.drone.io/api/badges/foo/bar.svg`

services/drone/drone-build.tester.js
```diff
@@ -0,0 +1,40 @@
+'use strict'
+
+const Joi = require('joi')
+const { isBuildStatus } = require('../build-status')
+const { ServiceTester } = require('../tester')
+
+const t = (module.exports = new ServiceTester({
```
calebcartwright: Should be able to replace this with the `createServiceTester` helper  function, see https://github.com/badges/shields/blob/master/doc/service-tests.md#1-boilerplate for more info

services/drone/drone-build.tester.js
byCedric: Perfect, thanks for the quick reply/help! Overlooked that one. I'll keep it in and fix it now!

calebcartwright:
Thanks @byCedric!

Left a few comments inline, also deployed a staging app with the changes @ https://shields-staging-pr-3240.herokuapp.com/

services/drone/drone-build.tester.js
byCedric: Done! Thanks, I briefly looked at Wercker but didn't notice this.

byCedric:
You guys are fast! Thanks a lot for the tips and staging deployment. The success and none badges are working great. 

Unfortunately, Drone is a bit different for returning the failed status in their badge. Instead of `failed` (or anything other listed in `build-status.js`), it returns `failure`. Is this something I should cover in the service or add to the `build-status.js`?

You can see it the staging examples here! 😄 

status | badeg 
---    | ---   
none   | [None Badge](https://shields-staging-pr-3240.herokuapp.com/drone/build/byCedric/GitHub-Website/not-existing.svg)
success | [Success Badge](https://shields-staging-pr-3240.herokuapp.com/drone/build/drone/drone.svg)
error  | [Failure Badge](https://shields-staging-pr-3240.herokuapp.com/drone/build/byCedric/GitHub-Website.svg)

services/drone/drone-build.tester.js
calebcartwright: I'd still suggest removing it. I know the Travis badge tester had this, but it's not a crucial test and not one typically added (see the testers for Codeship, Codacy, Beerpay, Gitlab Pipeline, Readthedocs, Snyk, etc.)

byCedric: Hmm, the actual URL being sent out is:

```
https://cloud.drone.io/api/badges/foo/bar/status.svg
```

I changed both the `api/badges` and `foo/bar/status.svg` part. But still, the test is failing. Do you have any additional tips/tricks?

calebcartwright:
>  Is this something I should cover in the service or add to the build-status.js

Makes sense to add `failure` as a red status in `build-status.js` to me 👍

services/drone/drone-build.tester.js
byCedric: Sorry, didn't see your previous comment. I'll remove it!

services/build-status.js
```diff
@@ -13,7 +13,13 @@ const greenStatuses = [
 
 const orangeStatuses = ['partially succeeded', 'unstable', 'timeout']
 
-const redStatuses = ['error', 'failed', 'failing', 'infrastructure_failure']
+const redStatuses = [
+  'error',
+  'failed',
+  'failing',
+  'failure',
```
calebcartwright: Thanks! Could you also update the corresponding test in `build-status.spec.js`?

byCedric: My bad! Will do it now!

byCedric: Done! 🚀

byCedric:
Sorry for the spam! I hope I wasn't too annoying. Thanks for the help, really nice to see open source work like this! ❤️ 🚀

bradrydzewski:
> Unfortunately, as mentioned in the comments here Drone doesn't have a "good" API for fetching last build status based on branches.

this is not longer the case. you can use the following endpoint:
`https://cloud.drone.io/api/repos/{owner}/{name}/builds/latest?ref=refs/heads/{branch}`

example:
`https://cloud.drone.io/api/repos/drone/hello-world/builds/latest?ref=refs/heads/test-on-arm`

byCedric:
Ah I'm sorry, I tried the `ref=refs/heads/{branch}` pattern on `https://cloud.drone.io/api/repos/{owner}/{name}/builds` only. Couldn't find the `/latest` endpoint, my bad! I might have some time tomorrow evening (CEST) to refactor it!

calebcartwright:
> I would also recommend including the protocol and hostname in the badge, similar to TeamCity, since there are many self-hosted Drone instances

Does anyone know what, if any, authentication is required to connect with a self-hosted Drone instance?

calebcartwright:
> Sorry for the spam! I hope I wasn't too annoying. 

Not at all! Thanks for the PR

> Thanks for the help, really nice to see open source work like this! ❤️ 🚀

🎉 💯

byCedric:
I think you have to authenticate for self-hosted instances, at least, I don't think it's publicly accessible. And there is this section in the docs: https://docs.drone.io/administration/user/admins/.

I'm in Poland now for a conference, will be back next week. Will make some time available to finish this 😄

calebcartwright:
hey @byCedric - just want to confirm the plan is still to refactor this to switch to the API instead of using the BaseSvgScraping service, correct?

byCedric:
Hi @calebcartwright, I'm sorry for not doing this earlier! Yes, that's still the plan. I think I can free up some time this evening to make a first version with the API service.

bradrydzewski:
Also just to confirm the implementation will be adjusted to allow a custom url, similar to Team City or Jenkins, since drone is primarily a self-hosted CI system and has many more self-hosted users.

Example:
`https://img.shields.io/jenkins/s/:scheme/:host/:job.svg`

Thanks again for this 🙂

byCedric:
@bradrydzewski Thank you for the support! Yes, I'll first refactor it to use the json service. After that I was thinking of supporting the following urls:

**For Drone Cloud builds**
- `https://img.shields.io/drone/cloud/build/:user/:repo`
- `https://img.shields.io/drone/cloud/build/:user/:repo/:branch`

**For custom hosted Drone**
- `https://img.shields.io/drone/:scheme/:host/build/:user/:repo`
- `https://img.shields.io/drone/:scheme/:host/build/:user/:repo/:branch`

Regarding the (possible?) user authentication, is this going to be a problem for unauthenticated requests from shields? Or are we good if we implement it like this?

calebcartwright:
@bradrydzewski sure. I do still have a question of how viable it will be for the main Shields.io service to be able to provide badges for self-hosted Drone instances (and I've admittedly not had a chance to read through documentation in detail yet)

If the Done API we'll need to use always requires authentication, and if each self-hosted Drone instance requires instance specific auth creds, then folks would have to deploy their own self-hosted Shields server (with their respective Drone auth details) to be able to get badges for their self-hosted Drone instance.

If the API doesn't need authentication, or if there's a generic auth that can be used (TeamCity supports a "guest" auth on the API used by Shields) then Shields.io should be able to work fine with any internet-accessible Drone instance.

bradrydzewski:
> After that I was thinking of supporting the following urls:

looks great!

> Regarding the (possible?) user authentication, is this going to be a problem for unauthenticated requests from shields? Or are we good if we implement it like this?

hmm, yes, the json endpoint requires authentication for private repositories.

byCedric:
@calebcartwright I see both Jenkins and TeamCity uses an authentication mechanism like:

```
if (serverSecrets.jenkins_user) {
  options.auth = {
    user: serverSecrets.jenkins_user,
    pass: serverSecrets.jenkins_pass,
  }
}
```

Is it ok if I use it something like this? Have no experience with that whatsoever. 😅  

```
if (serverSecrets.drone_user) {
  options.auth = {
    user: serverSecrets.drone_user,
    pass: serverSecrets.drone_pass,
  }
}
```

First I'll grab something to eat and will finish (at least) the refactor to JSON service tonight! Again, thanks for the quick reply and support. Absolutely love Shields and Drone ❤️

calebcartwright:
> Is it ok if I use it something like this

Yes, `serverSecrets` is the mechanism you'll want to use (I don't know what sort of auth Drone needs though)

Make sure to also update the [server secret docs](https://github.com/badges/shields/blob/master/doc/server-secrets.md) to include an entry for Drone, a bit more guidance on that [here](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#46-update-the-docs) and also update the config [here](https://github.com/badges/shields/blob/master/config/custom-environment-variables.yml#L44) to allow the Drone auth info to be set via env vars.

byCedric:
@bradrydzewski Sorry to ping you again, but right now I've got the basics up and running. You can define the cloud-hosted and self-hosted instances through the urls I mentioned earlier. Now, regarding to the authentication, is it possible to simply use a basic authentication request on the `builds/latest` endpoint?

The only documentation I could find is from [version `0.8.0`](https://0-8-0.docs.drone.io/api-authentication/) I guess. It tells me that I should use the access token provided from the user page. If this is the case, is this a long-term access token or should we fetch one before requesting the `builds/latest` endpoint?

Again, thanks for all the help guys! ❤️

byCedric:
Well, I think that's it! I just added the `DRONE_TOKEN`/`drone_token` to authenticate for self-hosted Drone instances. Did this based on the only [API documentation (0.8.0)](https://0-8-0.docs.drone.io/api-authentication/) I could find, so I think its best if Brad would take a look before merging. Thanks for the info related to the authentication! I think I did it correctly, but hey it's my first time 😄 

Let me know if something requires any change or if you just want to talk! 😄 

Cheers, ❤️ Cedric

bradrydzewski:
@byCedric looks great, including authentication

services/drone/drone-build.service.js
```diff
@@ -0,0 +1,123 @@
...
+
+module.exports = class DroneBuild extends BaseJsonService {
+  static get category() {
+    return 'build'
+  }
+
+  static get route() {
+    return {
+      base: 'drone',
+      pattern: ':scheme(http|https)?/:host/build/:user/:repo/:branch*',
```
calebcartwright: I assume this pattern is constructed this way to attempt to support Cloud & self-hosted instances conditionally, but I'm not sure that this is the best approach.

With scheme being optional, this is a "valid" route:
https://shields-staging-pr-3240.herokuapp.com/drone/mydrone-deployment.com/build/drone/drone.svg 

that ends up trying to make a protocol-less request to `://mydrone-deployment.com/apis/repos/drone/drone/builds/latest` which obviously is not a valid url. Accordingly it returns an `inaccessible` badge:



It would be fine to make the protocol and host mandatory in all cases (users can always enter `https` and `cloud.drone.io` respectively), this pattern is done for other badges (see [Jira](https://github.com/badges/shields/blob/master/services/jira/jira-issue.service.js) for example) . Or if you want to provide a separate route for badges targeting the cloud/SaaS instance to _not_ require users specify those values, then you could split this into two separate service classes (feel free to extract common functionality into separate functions and/or a base class) each with a more explicit route.

Also, let's change the route base to `drone/build` to be consistent with the Shields route conventions

services/drone/drone-build.service.js
```diff
@@ -0,0 +1,123 @@
...
+      options.headers = {
+        Authorization: `Bearer ${serverSecrets.drone_token}`,
+      }
+    }
+    const json = await this._requestJson({
+      schema,
+      url: `${scheme}://${host}/api/repos/${user}/${repo}/builds/latest`,
+      options,
+    })
+    return this.constructor.render({
```
calebcartwright: Not a big deal, but you could just call the `renderBuildStatusBadge` helper here (and in the examples)

services/drone/drone-build.service.js
calebcartwright: And yet another option to consider.. 😄 some of our badge routes use a default endpoint (typically the cloud/SaaS service), and accept a queryParam to support custom self-hosted endpoints. 

[Bitbucket Pull Requests](https://github.com/badges/shields/blob/master/services/bitbucket/bitbucket-pull-request.service.js#L92) is a good example of this. 

With that approach you could change the route here to:
```
  static get route() {
    return {
      base: 'drone/build',
      pattern: ':user/:repo/:branch*',
    }
  }
```

and the routes for build badges for cloud Drone would be:
`https://img.shields.io/drone/build/drone/drone.svg`

and for a self-hosted Drone instance:
`https://img.shields.io/drone/my-user/my-repo/drone.svg?server=https://my-self-hosted-drone.io`

byCedric: Ooh! I think I like the last approach the most! If it's ok with you I'll try to rip off bitbucket tonight 😄 This solution has the best readability vs functionality vs DRY in my opinion.

services/drone/drone-build.service.js
byCedric: Done! 😄

services/drone/drone-build.service.js
byCedric: Done! 🚀 Thanks for the detailed options here, it helps a lot understanding the idea/logic within Shields ❤️ Hope you like it!

byCedric:
I can squash everything to a single commit if you want, but I think it's handy-er if you do that when merging. But let me know if you want me to do that! 😄 #excited to use this! ❤️

calebcartwright:
> I can squash everything to a single commit if you want, but I think it's handy-er if you do that when merging

We always squash on merge so don't worry about it. Commit away 😄

calebcartwright:
I'll take a look at the updates in a few hours

byCedric:
Don't rush it, whenever you are available is fine 😄

services/drone/drone-build.service.js
```diff
@@ -0,0 +1,131 @@
...
+        },
+      },
+      {
+        title: 'Drone (self-hosted) with branch',
+        pattern: ':user/:repo/:branch',
+        queryParams: { server: 'https://drone.shields.io' },
+        namedParams: {
+          user: 'badges',
+          repo: 'shields',
+          branch: 'feat/awesome-thing',
```
calebcartwright: This sounds like an awesome branch! 😄

services/drone/drone-build.tester.js
```diff
@@ -0,0 +1,71 @@
...
+  .expectBadge({
+    label: 'build',
+    color: 'lightgrey',
+    message: Joi.alternatives().try(isBuildStatus, Joi.equal('invalid')),
+  })
+
+// Drone build (self-hosted)
+
+t.create('self-hosted build status on default branch')
+  .before(mockDroneCreds)
```
calebcartwright: nice 👍

services/drone/drone-build.tester.js
```diff
@@ -0,0 +1,71 @@
...
+const t = (module.exports = require('../tester').createServiceTester())
+const { mockDroneCreds, token, restore } = require('./drone-test-helpers')
+
+// Drone build (cloud)
+
+t.create('cloud build status on default branch')
+  .get('/drone/drone.json')
+  .expectBadge({
+    label: 'build',
+    color: 'brightgreen',
```
calebcartwright: I'd suggest removing the color checks from the expect statements here, as the colors are associated to a specific value that will (inevitably) fluctuate.

services/drone/drone-build.service.js
```diff
@@ -0,0 +1,131 @@
...
+
+  static get route() {
+    return {
+      queryParamSchema,
+      base: 'drone/build',
+      pattern: ':user/:repo/:branch*',
+    }
+  }
+
+  static get staticPreview() {
```
calebcartwright: Think we can remove this function entirely right?

services/drone/drone-build.service.js
```diff
@@ -0,0 +1,131 @@
...
+  static get examples() {
+    return [
+      {
+        title: 'Drone (cloud)',
+        pattern: ':user/:repo',
+        namedParams: {
+          user: 'drone',
+          repo: 'drone',
+        },
+        staticPreview: {
```
calebcartwright: You can call the `renderBuildStatusBadge` helper directly here

services/drone/drone-build.service.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'build',
+    }
+  }
+
+  static render({ status }) {
```
calebcartwright: can also remove this function, since it's no longer being called from anywhere (`handle` is using the `renderBuildStatusBadge` helper function directly which is 👍 )

services/drone/drone-build.tester.js
```diff
@@ -0,0 +1,71 @@
+'use strict'
+
+const Joi = require('joi')
+const { isBuildStatus } = require('../build-status')
+const t = (module.exports = require('../tester').createServiceTester())
+const { mockDroneCreds, token, restore } = require('./drone-test-helpers')
+
+// Drone build (cloud)
```
calebcartwright: just a minor nit, but you can probably remove these comments. the test names you've used sufficiently describe the tests IMO

services/drone/drone-build.service.js
```diff
@@ -0,0 +1,131 @@
...
+      }
+    }
+    if (!server) {
+      server = 'https://cloud.drone.io'
+    }
+    const json = await this._requestJson({
+      options,
+      schema: DroneBuildSchema,
+      url: `${server}/api/repos/${user}/${repo}/builds/latest`,
+    })
```
calebcartwright: It looks like the API will return a 401 status code if a user specifies an unknown/invalid repo or if they specify a private repo (at least based on some quick spot testing). It may be useful to include a custom error message to handle those scenarios:

```suggestion
      errorMessages: {
        401: 'repo not found or not authorized',
      },
    })
```

Otherwise I believe folks will see a generic `invalid` badge, even if they just had a typo in their repo name

services/drone/drone-build.service.js
byCedric: Yes, I think I copied it from Travis. It _was_ used in the examples, but not anymore. My bad for not removing this earlier.

services/drone/drone-build.service.js
byCedric: Sounds great!

byCedric:
Thanks Caleb! It's a team effort 😄 I'll process your comments later today, around 18:00 (CEST). Thanks for everything 🚀

services/drone/drone-build.tester.js
byCedric: Done!

services/drone/drone-build.tester.js
byCedric: Done!

services/drone/drone-build.service.js
byCedric: Done!

services/drone/drone-build.service.js
byCedric: Done!

services/drone/drone-build.service.js
byCedric: It's a synonym for every shields/drone branch 😉

services/drone/drone-build.service.js
byCedric: Done!

byCedric:
😄 go team! 🚀

andrewrynhard:
Hey guys, excited to see this. How long after merging can I start using this?

calebcartwright:
@andrewrynhard - no exact time frame but should be deployed to prod soon. The deployment bot will update the above comment https://github.com/badges/shields/pull/3240#issuecomment-483736477 once it has been deployed

paulmelnikow:
I'm running the deploy now.

Thank you all for your work and for tackling this long-standing feature request!

PR #3241, opened 2019-03-30
refactor [twitter] service

chris48s:


services/twitter/twitter.tester.js
```diff
@@ -15,25 +15,13 @@ t.create('Followers')
...
   })
 
-t.create('No connection')
-  .get('/follow/shields_io.json?label=Follow')
-  .networkOff()
-  .expectBadge({ label: 'Follow', message: 'inaccessible' })
-
 t.create('URL')
   .get('/url/https/shields.io.json')
   .expectBadge({ label: 'tweet', message: '' })
```
chris48s: This test isn't really testing the most important thing, which is the content of the links array, but this PR isn't making that worse, so I think I'm going to punt the job of improving the `expectBadge()` helper to another issue

services/twitter/twitter.service.js
```diff
@@ -116,49 +94,31 @@ class TwitterFollow extends LegacyService {
...
-        const user = match[1] // eg, shields_io
-        const format = match[2]
-        const options = {
-          url: `http://cdn.syndication.twimg.com/widgets/followbutton/info.json?screen_names=${user}`,
-        }
-        const badgeData = getBadgeData(`follow @${user}`, data)
+  async fetch({ user }) {
+    return this._requestJson({
+      schema,
+      url: `http://cdn.syndication.twimg.com/widgets/followbutton/info.json?screen_names=${user}`,
```
PyvesB: Should `user` be URL encoded? We may want to specify it as a query string parameter like we do in the [Wercker service](https://github.com/badges/shields/blob/f0e037e0404c40271f9e72c49cdd12d5e33d53fc/services/wercker/wercker.service.js#L31) for instance.

Issue #672, opened 2016-03-18
Handle large GitHub releases (use caching!?)


mfonville:
For OpenGApps I would like to include a badge with # downloads of our releases.
But already even one single release of one architecture often gives timeouts, like: [![Latest ARM](https://img.shields.io/github/downloads/opengapps/arm/latest/total.svg)](http://opengapps.org/arch=arm)

While I would prefer to even do:
[![ARM Releases](https://img.shields.io/github/downloads/opengapps/arm/total.svg)](http://opengapps.org/arch=arm)

And I would like to do that for our other architectures (arm64,x86,x86_64) preferably even with a grand-total for all of the 4 repositories together.
We do daily releases, so we are the "weird" one and not the typical use-case, but I would really like to use the badges.

Some caching of non real-time data would not hurt the use-case, since we are well above 3 million downloads...

espadrine:
I think the timeouts are actually caused by GitHub rate limiting, for which we have an issue here: #529. I suspect it was a lot, lot less often recently.

Still, we very recently have added functionality to enforce custom browser caching: #534. You can set it to, say, 86400 seconds (a day): https://img.shields.io/github/downloads/opengapps/arm/total.svg?maxAge=86400. I will add some documentation about it on the front page soon.

Do you feel that this addresses your issues?

mfonville:
@espadrine Indeed, the rate-limiting is very probably the cause of the problem. I will add the browser-caching, I hope that helps with the issue a bit, but I do fear that with the amount of unique visitors the rate-limit is hit nonetheless.
Thanks for the support!

mfonville:
I have btw the feeling that the 'total' is not correct.
I fetched all number of downloads data from the Python API and I see many millions more of downloads there. When discarding the .md5s and .txts and only counting the .zips the number is around 14 million.

**Does the shields.io take into account that the GitHubAPI uses pagination for large result sets?**

espadrine:
Based on the raw request's Link header, having the complete information would require 10 requests: `curl -I 'https://api.github.com/repos/opengapps/arm/releases' | grep ^Link`. We indeed only perform one, and that request is already pretty expensive and slow for GitHub's servers.

If GitHub offered a fast way to access the complete total download, we would be able to implement it for a live badge. Us offering a badge that takes ages to load is not very practical, and, as you noted, the current badge already is a pretty time-consuming badge for GitHub to compute.

[...]

PR #3256, opened 2019-04-01
Fixed GitHub release downloads, run [GithubDownloads] 

denisdulici:
By default, GH API gives 30 releases per page which causes wrong download counter if you have more than 30 releases.


calebcartwright:
@denisdulici Thanks for the PR!

The format/lint check jobs are failing. Can you please run `npm run prettier` and push the changes?

paulmelnikow:
@denisdulici Thanks for picking this up! The last commit picked up lockfile changes which need to be reset: `git checkout master package-lock.json` should do it.

Does this resolve the issue described in #672?

denisdulici:
@paulmelnikow Yes, it should resolve that issue. Not sure about the lockfile, should I reset it?

PR #3259, opened 2019-04-02
refactor [GithubIssueDetail]

calebcartwright:
Some inline comments/questions added

### DIFF
--- a/services/github/github-issue-detail.service.js
+++ b/services/github/github-issue-detail.service.js
@@ -1,27 +1,137 @@
 'use strict'
 
-const LegacyService = require('../legacy-service')
-const {
-  makeLabel: getLabel,
-  makeBadgeData: getBadgeData,
-} = require('../../lib/badge-data')
-const { makeLogo: getLogo } = require('../../lib/logos')
+const Joi = require('joi')
+const { InvalidResponse } = require('..')
+const { nonNegativeInteger } = require('../validators')
 const { formatDate } = require('../text-formatters')
-const { age: ageColor } = require('../color-formatters')
+const { age } = require('../color-formatters')
+const { GithubAuthService } = require('./github-auth-service')
 const {
   documentation,
-  stateColor: githubStateColor,
-  commentsColor: githubCommentsColor,
-  checkErrorResponse: githubCheckErrorResponse,
+  errorMessagesFor,
+  stateColor,
+  commentsColor,
 } = require('./github-helpers')
 
-// This legacy service should be rewritten to use e.g. BaseJsonService.
-//
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class GithubIssueDetail extends LegacyService {
+const getDefaultLabel = ({ json }) => {
+  if (json.pull_request) {
+    return `pull request ${json.number}`
+  }
+
+  return `issue ${json.number}`
+}
+
+const commonSchemaFields = {
+  number: nonNegativeInteger,
+  pull_request: Joi.any(),
+}
+
+const stateSchema = Joi.object({
+  ...commonSchemaFields,
+  state: Joi.string()
+    .allow('open', 'closed')
+    .required(),
+}).required()
+
+const stateMap = {
+  schema: stateSchema,
+  getLabel: getDefaultLabel,
+  getColor: ({ value }) => stateColor(value),
+  transform: ({ json }) => json.state,
+}
+
+const titleSchema = Joi.object({
+  ...commonSchemaFields,
+  title: Joi.string().required(),
+}).required()
+
+const titleMap = {
+  schema: titleSchema,
+  getLabel: getDefaultLabel,
+  transform: ({ json }) => json.title,
+}
+
+const authorSchema = Joi.object({
+  ...commonSchemaFields,
+  user: Joi.object({
+    login: Joi.string().required(),
+  }).required(),
+}).required()
+
+const authorMap = {
+  schema: authorSchema,
+  getLabel: () => 'author',
+  transform: ({ json }) => json.user.login,
+}
+
+const labelSchema = Joi.object({
+  ...commonSchemaFields,
+  labels: Joi.array()
+    .items(
+      Joi.object({
+        name: Joi.string().required(),
+        color: Joi.string().required(),
+      })
+    )
+    .required(),
+}).required()
+
+const labelMap = {
+  schema: labelSchema,
+  getColor: ({ json }) => {
+    if (json.labels.length === 1) {
+      return json.labels[0].color
+    }
+  },
+  getLabel: () => 'label',
+  transform: ({ json }) => {
+    if (json.labels.length === 0) {
+      throw new InvalidResponse({ prettyMessage: 'no labels found' })
+    }
+    return json.labels.map(l => l.name).join(' | ')
+  },
+}
+
+const commentsSchema = Joi.object({
+  ...commonSchemaFields,
+  comments: nonNegativeInteger,
+}).required()
+
+const commentsMap = {
+  schema: commentsSchema,
+  getColor: ({ value }) => commentsColor(value),
+  getLabel: () => 'comments',
+  transform: ({ json }) => json.comments,
+}
+
+const ageUpdateSchema = Joi.object({
+  ...commonSchemaFields,
+  created_at: Joi.date().required(),
+  updated_at: Joi.date().required(),
+}).required()
+
+const ageUpdateMap = {
+  schema: ageUpdateSchema,
+  formatMessage: date => formatDate(date),
+  getColor: ({ value }) => age(value),
+  getLabel: ({ which }) => (which === 'age' ? 'created' : 'updated'),
+  transform: ({ json, which }) =>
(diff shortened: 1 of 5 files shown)


services/github/github-issue-detail.service.js
```diff
@@ -1,27 +1,137 @@
...
 } = require('./github-helpers')
 
-// This legacy service should be rewritten to use e.g. BaseJsonService.
-//
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class GithubIssueDetail extends LegacyService {
+const getDefaultLabel = ({ json }) => {
```
calebcartwright: I decided to go with a solution of leveraging objects for the various `which` types to enable usage of more granular/stricter schemas (vs a single schema with all properties as optional) and to avoid needing multiple switch statements.

services/github/github-issue-detail.service.js
```diff
@@ -30,7 +140,7 @@ module.exports = class GithubIssueDetail extends LegacyService {
     return {
       base: 'github/issues/detail',
```
calebcartwright: I left the route here as it was previously defined, but wanted to highlight that the Legacy Service supported both `issues` and `pulls` (`github/issues|pulls/detail`). Is the desire to narrow the official path to just be `issues`?

If so I'll add a redirector service to prevent breaking anyone using `github/pulls/detail/...`

services/github/github-issue-detail.service.js
```diff
@@ -30,7 +140,7 @@ module.exports = class GithubIssueDetail extends LegacyService {
...
       pattern:
-        ':which(s|state|title|u|author|label|comments|age|last-update)/:user/:repo/:number(0-9+)',
+        ':which(s|state|title|u|author|label|comments|age|last-update)/:user/:repo/:number',
```
calebcartwright: Two items here:

1) do we want to move to the full words and not promote the one letter aliases (`s` and `u` in particular)? The existing  example below from the legacy service had overridden the pattern, presumably to remove those two shortcuts from the modal window
2) I forgot I had to drop the the `number` param expression, was getting 404 badge not found errors with that in place. Not sure what was going on there 🤔

services/github/github-issue-detail.spec.js
```diff
@@ -0,0 +1,84 @@
...
+    given({
+      which: 'label',
+      json: { labels: [{ name: 'bug', color: 'ee0701' }] },
+    }).expect({
+      value: 'bug',
+    })
+  })
+
+  context('transform()', function() {
+    it('throws InvalidResponse error when issue has no labels', function() {
```
calebcartwright: It wasn't immediately clear to me with `sazerac` how to do an `expect.throws` sort of test, so I just went with default mocha/chai here

services/github/github-issue-detail.spec.js
```diff
@@ -0,0 +1,84 @@
+'use strict'
+
+const { expect } = require('chai')
+const { test, given } = require('sazerac')
+const { InvalidResponse } = require('..')
+const GithubIssueDetail = require('./github-issue-detail.service')
+const { stateColor } = require('./github-helpers')
+
```
calebcartwright: Will happily add some other unit tests in here to validate the `render` and `transform` functions once there's no objections to the approach I took in the service implementation

services/github/github-issue-detail.service.js
paulmelnikow: I'm fine with that.

services/github/github-issue-detail.service.js
paulmelnikow: 1. Yes to all of that! I was trying to nudge people away from the single-character aliases when I set up the rest of those, and feel like that's a good direction to push in. Using a redirect sounds good.
2. Does it need to be `:number([0-9]+)`?

services/github/github-issue-detail.spec.js
paulmelnikow: Right, I don't believe sazerac supports that.

services/github/github-issue-detail.service.js
paulmelnikow: This is a bit difficult to follow, though it doesn't look too hard to maintain – and this code is doing a lot of stuff.

One suggestion that I think would make this easier to understand is to implement render functions inside your objects to make them more direct, instead of breaking up the render responsibility into separate functions that get called later. There might be the slightest bit more repetition with render functions, though I think it will be quite a bit easier to follow.

calebcartwright: > to implement render functions inside your objects to make them more direct

I like the sound of that, will do.

The other thing i considered was moving all of this `which` specific transform/render into a separate file, but decided to keep it all together for now. Any thoughts/preferences on that (in-file w/service class vs. separate file)?

paulmelnikow: Probably in with the service.

calebcartwright: Done! I like the way this looks a lot better, it was a really good idea. I believe it's far more readable now with each map object containing its respective `schema`, and functions for `render` and `transform`

services/github/github-issue-detail.service.js
calebcartwright: Done!

services/github/github-issue-detail.service.js
calebcartwright: 👍

services/github/github-issue-detail.service.js
```diff
@@ -1,27 +1,142 @@
...
+    updated_at: Joi.date().required(),
+  }).required(),
+  transform: ({ json, which }) =>
+    which === 'age' ? json.created_at : json.updated_at,
+  render: ({ which, value }) => ({
+    color: age(value),
+    label: which === 'age' ? 'created' : 'updated',
+    message: formatDate(value),
+  }),
+}
```
paulmelnikow: In the future it might be nice to design a mini-framework that can live inside of a service and deal with child-service logic that crosses boundaries of fetch (sometimes), schema, transform, and render. I know we've got these sprinkled about and it might be nice to design a single way to guide these implementations.

services/github/github-issue-detail.service.js
```diff
@@ -1,27 +1,142 @@
...
+
+const whichMap = {
+  state: stateMap,
+  title: titleMap,
+  author: authorMap,
+  label: labelMap,
+  comments: commentsMap,
+  age: ageUpdateMap,
+  'last-update': ageUpdateMap,
```
paulmelnikow: Just a thought that you could consider all of these, and then mutating the result to add the one synonym. Not sure that would make this better; just an option to consider.

services/github/github-issue-detail.service.js
```diff
@@ -1,27 +1,142 @@
...
+    user: Joi.object({
+      login: Joi.string().required(),
+    }).required(),
+  }).required(),
+  transform: ({ json }) => json.user.login,
+  render: ({ value }) => ({
+    label: 'author',
+    message: value,
+  }),
+}
```
paulmelnikow: This is much easier to read. Nice improvement! 👍

services/github/github-issue-detail.service.js
```diff
@@ -1,27 +1,142 @@
...
+      )
+      .required(),
+  }).required(),
+  transform: ({ json }) => {
+    if (json.labels.length === 0) {
+      throw new InvalidResponse({ prettyMessage: 'no labels found' })
+    }
+    return json.labels.map(l => l.name).join(' | ')
+  },
+  render: ({ value, json }) => {
```
paulmelnikow: What you've written here is fine, though for the purpose of polishing or thinking about ways to structure future work, you could consider this.

It might be a little nicer to have your transform function return an object of props, and your render function to be passed `which` and `props` (and not `json`). That way you can avoid doing rendering in transform and transforming in render. Here your transform function could return e.g. an array of label names and a second array of colors.

Where it really looks odd is the call to `this.render` in the examples, the arguments to which were unexpected.

All totally optional.

calebcartwright: I now realize that the `labels` badge is doing an odd blending of transform/render 

> It might be a little nicer to have your transform function return an object of props

Not sure whether I'm correctly visualizing what that would look like. I'm thinking that the proposal is for this particular `render` function to return an object (`props`) that's extracted the relevant data portions from the response, and passing that along to `render` (i.e. here the `value` param would be that object instead of the formatted/concatenated label name string). 

The example render felt odd to me too, so let me take another pass at simplifying this.

services/github/github-issue-detail.service.js
calebcartwright: I'm not sure I follow. What does a 'consider' of those look like?

services/github/github-issue-detail.service.js
paulmelnikow: I’m suggesting replacing `value` with an object that contains the values you need.

services/github/github-issue-detail.service.js
paulmelnikow: Ha! Consider _inlining_, is what I meant to write 😆

services/github/github-issue-detail.service.js
calebcartwright: Updated accordingly. Another solid improvement IMO

Issue #2026, opened 2018-08-30
Ensure consistent usage of renderVersionBadge


PyvesB:
The following function was introduced in #1922: https://github.com/badges/shields/blob/302c8606ff4c975b6310fe2cdfe58f92cede290b/lib/version.js#L144-L150 

We should be using it where appropriate.

See https://github.com/badges/shields/pull/1965#discussion_r213526512.

paulmelnikow:
Mostly done in #1998, though we need to tackle it for f-droid.

To expand on this a little bit: I'm thinking we should make a pass through _all_ the new-style services for consistency of style. Having a consistent style in place means contributors will refer to code that are exemplars of what we want, minimizing the amount of "please make this more the way we want it," and creating a more positive contributor experience.

It seems a little fussy to do this on existing code which is "good enough." However, I think more positive contributor experience is a good reason to make the extra effort up front.

What do you think about that idea?

PyvesB:
Yes, I think going through all our new services to make sure the style is consistent would make sense.

paulmelnikow:
> For example, there are some programming communities where a lot of projects use alternative versioning schema like [CalVer](https://calver.org/) or just don't specify any particular way of versioning.

👍 

Let's rename the helper so it's clear it's intended for communities which use semver, or better yet, add a `usesSemver` parameter to it.

> Similarly there are some programming communities where it is not really appropriate to try and assign meaning to licenses via colour, or where a license is "good" or "bad" may not be about permissive vs copyleft but more about whether it is "GPL-compatible" (which is a property that can apply to some subset of both permissive and copyleft licences) for example.

Are there any programming communities where copyleft is seen radically differently from others?

By choosing orange for GPL I've tried to sidestep the value judgement. I've wandered from one of these schools of thought to the other in the last 10 years, and tend to think folks who like GPL can train themselves to look for orange license badges, likewise folks who like permissive licenses and green.

Alternatively, perhaps we could try to refine the colors.

I participate in more than one programming community and FWIW I'd find the consistent colors helpful.

This overlaps with #1093; going to close that so let's wrap up this discussion here.

PR #3265, opened 2019-04-04
[pypi] Improve formatting of trove-based licenses

paulmelnikow:
Currently the trove-based license classifiers generate different formatting than the legacy license classifiers. This brings them into alignment.

Production:

This PR:

Think this is a good improvement; more consistent with most other badges, and will look better on readmes (see https://github.com/lace/vg for an example).

### DIFF
--- a/services/pypi/pypi-helpers.js
+++ b/services/pypi/pypi-helpers.js
@@ -57,12 +57,18 @@ function getLicenses(packageData) {
   if (license) {
     return [license]
   } else {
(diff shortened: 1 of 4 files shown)


calebcartwright:
Semi-related, I noticed that the color schemes applied on license badges vary between license services as well. Is that intentional or worth potentially trying to make them consistent at some point in the future?

paulmelnikow:
Yes! There's some discussion of that at #2026.

Issue #3267, opened 2019-04-04
In CI, run e2e tests on the build


paulmelnikow:
#3262 added support for e2e tests. As I'm looking at some of the dependency bump PRs, I realize I could speed up review of those if I didn't have to open the review apps.

If e2e tests ran on the build (i.e. the same site I would see in the review app), then their passing would give me confidence the site is basically rendering, which is the main thing I check when reviewing a PR like #3248.

calebcartwright:
I think there is a job that runs the e2e tests now (at least it's run on the most recent PRs i've opened like #3266). I wonder what endpoint the tests are targeting

calebcartwright:
I bet whatever solution is applied here could also be a big help with #2816

platan:
e2e job runs [`e2e`](https://github.com/badges/shields/blob/master/package.json#L111) script, which starts a server and runs e2e tests. We can easily run e2e tests targeting any instance by `CYPRESS_baseUrl=https://shields.io npm run test:e2e`. 
Do we create a staging instance for every PR automatically? If so, we can easily target staging instance because it have consistent naming (https://shields-staging-pr-3240.herokuapp.com/). We just need PR id. 
@paulmelnikow Is this what you want to achieve?

paulmelnikow:
Is it starting the server using `npm start`? I think what we'd want to do instead is first run `npm run build`, and then `npm run start:server`. The tests could still use the server that's running locally, but cypress would get the same static build as a user would see in production.

I think Docker needs something slightly different: a way to start the container, run tests, then shut down the container.

calebcartwright:
> I think Docker needs something slightly different: a way to start the container, run tests, then shut down the container.

I just meant that whatever we do to run the tests against an endpoint (like a staging app) could also be used as the validation piece against the container (just a different endpoint target)

platan:
Yes, the CI job starts server using `start`. It shouldn't be a lot of work to change it to `npm run build` + `npm run start:server`. I can do it.

paulmelnikow:
> I just meant that whatever we do to run the tests against an endpoint here (like a staging app) could also be leveraged in a similar fashion to do the validation piece against the container (just a different endpoint target). Just a matter of specifying the right value for `CYPRESS_baseUrl` if I follow correctly

[...]

PR #3268, opened 2019-04-04
In CI, run e2e tests on the build

platan:
Fixes https://github.com/badges/shields/issues/3267

`npm run e2e` still starts server using `npm run start`, which is convenient for development.


platan:
Again have some problems with CI. Closing in favour of https://github.com/badges/shields/pull/3269

PR #3272, opened 2019-04-06
JSON format modernisation and _shields_test removal

PyvesB:
Related to #2686 and #2687.

There wasn't much information about the JSON badge format, so I added a small piece of documentation which, amongst other things, highlights that `name` and `value` are now deprecated fields.


core/service-test-runner/icedfrisby-shields.js
```diff
@@ -32,18 +33,21 @@ const factory = superclass =>
...
-        expectedBadge = { name: label, value: message, color }
-      }
-
-      if (typeof message === 'string') {
-        return this.expectJSON(expectedBadge)
-      } else {
-        return this.expectJSONTypes(Joi.object().keys(expectedBadge))
-      }
+      return this.afterJSON(json => {
+        expect(json.label, 'label mismatch').to.equal(label)
```
paulmelnikow: One option to consider here is using multiple Frisby expectations instead of `afterJSON()`. The advantage of Frisby `expect`s is that they all run, so if multiple things go wrong, they are all reported. If they’re all chained into an `after` the test stops after the first failure.

I haven’t used that part of the API much myself so I can’t vouch for its ergonomics, though we can certainly work to improve it if it turns out there are problems.

In particular, I think you can call [`expectJSON()`](https://github.com/IcedFrisby/IcedFrisby/blob/master/API.md) (or `expectJSONTypes()`) multiple times, using the optional `path` argument:

```js
this.expectJSONTypes(‘label’, label)
this.expectJSONTypes(‘message’, message)
if (typeof color !== undefined) {
  this.expectJSONTypes(‘color’, color)
}
return this
```

I’m certainly fine with what you’ve done here as well! This could definitely be tackled / considered in a follow-up.

gh-badges/lib/make-badge.js
```diff
@@ -164,7 +164,6 @@ function makeBadge({
     logoPosition,
     logoWidth,
     logoPadding,
-    // `color` and `labelColor` are included for the `_shields_test` template.
```
paulmelnikow: Probably this comment should be kept, and updated to reference the json template.

gh-badges/templates/default-template.json
```diff
@@ -1,4 +1,7 @@
 {
   "name": {{=JSON.stringify(it.text[0])}},
-  "value": {{=JSON.stringify(it.text[1])}}
+  "label": {{=JSON.stringify(it.text[0])}},
+  "value": {{=JSON.stringify(it.text[1])}},
+  "message": {{=JSON.stringify(it.text[1])}},
+  "color": {{=JSON.stringify(it.color || null)}}
```
paulmelnikow: Could you transfer over the other properties from _shields_test-template? `logoWidth` and `labelColor`

The reasoning is that I’d like the output of the JSON badge to as closely as possible match the input to the `makeBadge` function. It should make the server logic and the badge generation code testable in complete isolation. It also makes the JSON endpoint more flexible because it contains all the data for the badge, you can hit that endpoint (on e.g. the live server), manipulate it the response, and then pipe it to your own copy of `makeBadge`. Shields has a ton of layers – though good on us, we keep peeling them away! I think the more we can make these schemas consistent, the lower the mental overhead will be.

services/endpoint/endpoint.tester.js
```diff
@@ -30,11 +30,8 @@ t.create('color and labelColor')
         labelColor: '#e6e6fa',
       })
   )
-  .expectJSON({
-    name: 'hey',
-    value: 'yo',
-    color: '#f0dcc3',
-    labelColor: '#e6e6fa',
+  .after(body => {
+    expect(body).to.include('fill="#e6e6fa"')
```
paulmelnikow: I haven’t re-read all these tests, though judging from the test title I think this needs to test `color` as well. If you add `labelColor` to the JSON body, you could revert this to be a JSON test.

core/service-test-runner/icedfrisby-shields.js
PyvesB: Using `path` was my initial attempt before the [issue with afterJSON](https://github.com/IcedFrisby/IcedFrisby/issues/206) was fixed, but if I remember correctly it didn't work as the second argument (`jsonTest`) must be an object rather than a plain string.

I also tried bundling everything into a single expect, but that didn't work nicely either:
* `expectJSON` does strict matching, so we would have to specify additional fields such as `color` for every single test. Services use a lot of different colors for the messages, so there's no sensible default value we could set. In my opinion, adding colors everywhere would add noise to our test suite rather than improve it.
* `expectContainsJSON` is a looser alternative, but when the match fails the error message isn't as clear as "expected message to be foo but was bar". In particular, it will highlight the deprecated fields `value` and `name` which we don't care about. Even though I did get the assertions working, the developer experience was quite confusing.

services/endpoint/endpoint.tester.js
PyvesB: Extended `expectBadge` to handle `labelColor` and `logoWidth`.

core/service-test-runner/icedfrisby-shields.js
paulmelnikow: Make sense. Want to open an issue documenting some of this?

paulmelnikow:
Could I suggest a full service-test run before merging? If you put `[*****]` in the title it’ll run in CI. My concern is legacy services; they’re always doing unexpected things.

core/service-test-runner/icedfrisby-shields.js
PyvesB: Will do so tomorrow.

PyvesB:
I hadn't realised some of the tests used Joi schemas for the color, will fix that in the morning. I might as well add support for both strings and schemas for all fields.

PyvesB:
The remaining failures seem unrelated.

PR #3275, opened 2019-04-06
refactor [puppetforge] service, add tests

chris48s:
This PR migrates puppetforge-modules from legacy services to the new service arch.
There are also some changes to the puppetforge-users badges, but its just moving code around to make the directory structure a bit more consistent.

Sorry this is quite a large diff.


services/puppetforge/puppetforge-modules.service.js
```diff
@@ -1,281 +0,0 @@
...
-          }
-          try {
-            if (info === 'v') {
-              if (json.current_release) {
-                const version = json.current_release.version
-                badgeData.text[1] = versionText(version)
-                badgeData.colorscheme = versionColor(version)
-              } else {
-                badgeData.text[1] = 'none'
-                badgeData.colorscheme = 'lightgrey'
```
chris48s: I've just ignored this case. If we hit a response that doesn't have the `current_release` key, we'll fail to validate with "unparseable json repsonse". I think this is probably reasonable.

services/puppetforge/puppetforge-module-pdk-version.service.js
```diff
@@ -0,0 +1,46 @@
...
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'pdk version' }
+  }
+
+  async handle({ user, moduleName }) {
+    const data = await this.fetch({ user, moduleName })
+    if (data.current_release.pdk) {
```
calebcartwright: is there anyway `pdk` could be true, but the `data.current_release.metadata` object key could be undefined? I'm not familiar with the service/api response but am assuming there's a relationship between those two keys based on the schema and this logic?

chris48s: Not sure, but I've tightened up the schema a bit so that if that condition does happen, the validation will fail

services/puppetforge/puppetforge-modules.service.js
calebcartwright: 👍

PR #3290, opened 2019-04-11
refactor [mavencentral]

Prouser123:


services/maven-central/maven-central.service.js
```diff
@@ -56,52 +63,26 @@ module.exports = class MavenCentral extends LegacyService {
...
-      })
-    )
+  static get defaultBadgeData() {
+    return { label: 'maven-central' }
+  }
+
+  async fetch({ groupId, artifactId }) {
+    const url = `${'https://repo1.maven.org/maven2/'}${encodeURIComponent(
+      groupId
+    ).replace(/\./g, '/')}/${encodeURIComponent(artifactId)}/maven-metadata.xml`
```
chris48s: I found that doing this as multiple function calls inline in a template string broken over 3 lines makes it quite difficult to read. What do you reckon to something more like:

```js
const group = encodeURIComponent(groupId).replace(/\./g, '/')
const artifact = encodeURIComponent(artifactId)
const url = `https://repo1.maven.org/maven2/${group}/${artifact}/maven-metadata.xml`
```

Its still 3 vertical lines, but easier to understand what's going on.

services/maven-central/maven-central.tester.js
```diff
@@ -22,7 +22,7 @@ t.create('latest 0.8 version')
 
 t.create('inexistent artifact')
   .get('/v/inexistent-group-id/inexistent-artifact-id.json')
-  .expectBadge({ label: 'maven-central', message: 'invalid' })
+  .expectBadge({ label: 'maven-central', message: 'not found' })
 
 t.create('connection error')
```
chris48s: You can get rid of the "connection error" test as this case is handled in `BaseService` now

calebcartwright:
Review app deployed https://shields-staging-pr-3290.herokuapp.com/

services/maven-central/maven-central.service.js
```diff
@@ -56,52 +63,26 @@ module.exports = class MavenCentral extends LegacyService {
...
+  }
+
+  async handle({ groupId, artifactId, versionPrefix }) {
+    const data = await this.fetch({ groupId, artifactId })
+    const versions = data.metadata.versioning.versions.version.reverse()
+    let version = versions[0]
+    if (versionPrefix !== undefined) {
+      version = versions.filter(v => v.toString().startsWith(versionPrefix))[0]
+      // if the filter returned no results, throw a NotFound
+      if (version === undefined) throw new NotFound()
```
calebcartwright: Would it be helpful to include some kind of a `prettyMessage` property on the `NotFound` error to indicate that it's the _version prefix_ that's not found (as opposed to the artifact itself)?

 

vs.

services/maven-central/maven-central.tester.js
```diff
@@ -22,18 +22,13 @@ t.create('latest 0.8 version')
...
   .get('/v/inexistent-group-id/inexistent-artifact-id.json')
-  .expectBadge({ label: 'maven-central', message: 'invalid' })
-
-t.create('connection error')
-  .get('/v/com.github.fabriziocucci/yacl4j.json')
-  .networkOff()
-  .expectBadge({ label: 'maven-central', message: 'inaccessible' })
+  .expectBadge({ label: 'maven-central', message: 'not found' })
 
 t.create('xml parsing error')
```
calebcartwright: I think we can probably get rid of this test altogether too, as it's testing functionality in the Base class, which has its own tests elsewhere

services/maven-central/maven-central.service.js
```diff
@@ -56,52 +63,27 @@ module.exports = class MavenCentral extends LegacyService {
...
+  }
+
+  async handle({ groupId, artifactId, versionPrefix }) {
+    const data = await this.fetch({ groupId, artifactId })
+    const versions = data.metadata.versioning.versions.version.reverse()
+    let version = versions[0]
+    if (versionPrefix !== undefined) {
+      version = versions.filter(v => v.toString().startsWith(versionPrefix))[0]
+      // if the filter returned no results, throw a NotFound
+      if (version === undefined)
```
calebcartwright: Can we add a test for this scenario as well? Should be able to use a live test with a prefix that doesn't exist (perhaps something like `99.99`)

services/maven-central/maven-central.tester.js
```diff
@@ -22,18 +22,8 @@ t.create('latest 0.8 version')
```
calebcartwright: One last ask from me (sorry I missed this earlier) 😄 

Let's switch to the `createServiceTester` shorthand:
`const t = (module.exports = require('../tester').createServiceTester())`

And I'd also suggest we reuse one of existing test validator for versions (they're used for other maven version related badges)
`isVPlusDottedVersionAtLeastOne` or `isVPlusDottedVersionNClausesWithOptionalSuffix` from `test-validators` should work here too

Prouser123: I'll probably go with `isVPlusDottedVersionNClausesWithOptionalSuffix`.
Might wanna add some nock tests for that.

paulmelnikow:
Hey @Prouser123 I'd like to get this merged by or on Friday, which is our target for getting the refactoring done (!) and also the two-year anniversary of when we _started_ refactoring (!!).

Would you have time to address these comments in the next day or two?

Prouser123:
Ah, with the new shorthand the service name is just `mavencentral` instead of `maven-central`.

PR #3291, opened 2019-04-11
[Jitpack] Added Downloads Badge

erayerdin:
Refer to #3221 and #3220.

I have messed up my local git repository so I had to reclone and build development environment. I started from a clean branch with upstream update.


calebcartwright:
Review app deployed at https://shields-staging-pr-3291.herokuapp.com/

services/jitpack/jitpack-downloads.tester.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+const t = (module.exports = require('../tester').createServiceTester())
+
+const isMetric = /^.*\/(week|month)/
```
calebcartwright: I'd suggest using the existing `isMetricOverTimePeriod` validator from `test-validators` instead. I think this would technically pass on `abc/month` which is not ideal 😄

erayerdin: I must have seen `isMetric` somewhere in the codebase (probably in pypi service` and have tried to imitate that. Whatever, I will use `isMetricOverTimePeriod`. :+1:

PR #3294, opened 2019-04-12
README: better visibility of discord icon in badge

tooomm:
The original blue color blends in too much with the dark background on the small icon.

- old: chat on Discord</a>
- new: chat on Discord</a>


paulmelnikow:
I don't feel strongly about whether we do this, though if we do it probably should be by tweaking the code used to auto-color the badges that were added in #2833.

You can see the results of the auto-coloring here: https://shields-staging.herokuapp.com/dev/logos

Though it rate limits; so it's easier to run that page locally.

paulmelnikow:
Hey @tooomm could you please open an issue to discuss? If these colors are changed, it will need a different approach than what's in this PR.

tooomm:
Done, see #3301

PR #3298, opened 2019-04-13
Rewrite [dynamicxml] tests, and fix issue with attribute filters

### LINKED ISSUE #3089: Use fixtures for tests of dynamic badges

paulmelnikow:
This fixes queries with an attribute filter in the middle, such as `//book[@id='bk102']/title`.

Close #3089

### DIFF
--- a/services/dynamic/dynamic-xml.tester.js
+++ b/services/dynamic/dynamic-xml.tester.js
@@ -1,9 +1,14 @@
 'use strict'
 
-const Joi = require('joi')
-const { expect } = require('chai')
-const { isSemver } = require('../test-validators')
+const queryString = require('query-string')
 const t = (module.exports = require('../tester').createServiceTester())
+const { exampleXml } = require('./dynamic-response-fixtures')
+
+const exampleUrl = 'https://example.test/example.xml'
+const withExampleXml = nock =>
+  nock('https://example.test')
+    .get('/example.xml')
+    .reply(200, exampleXml)
 
 t.create('No URL specified')
   .get('.json?query=//name&label=Package Name')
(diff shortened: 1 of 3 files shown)


services/dynamic/dynamic-xml.tester.js
```diff
@@ -111,46 +133,18 @@ t.create('XML from url | invalid url')
...
-  .get('.json?url=https://xml-test/api.xml&query=/name')
   .intercept(nock =>
-    nock('https://xml-test')
-      .get('/api.xml')
-      .reply(200, function(uri, requestBody) {
-        headers = this.req.headers
-        return '<?xml version="1.0" encoding="utf-8" ?><name>dynamic xml</name>'
-      })
+    nock('https://example.test', {
+      reqheaders: { accept: 'application/xml, text/xml' },
```
paulmelnikow: This causes a "Mocks not net satisfied" assertion to be thrown if the Accept header doesn't match.

calebcartwright:
Nice. Just to validate my understanding of this/#3089, using a live endpoint to retrieve XML has been making the tests problematic, and there's no real value in retrieving the XML from a live endpoint anyway (not like there's a static CircleCI, etc. service endpoint that live tests would be validating)

paulmelnikow:
Yea, that's exactly it.

services/dynamic/dynamic-xml.tester.js
```diff
@@ -25,76 +28,95 @@ t.create('No query specified')
...
   .expectBadge({
     label: 'custom badge',
-    message: Joi.string().regex(
-      /^Firefox( for Android)?,\sFirefox( for Android)?$/
-    ),
+    message:
+      "XML Developer's Guide, Midnight Rain, Maeve Ascendant, Oberon's Legacy, The Sundered Grail, Lover Birds, Splish Splash, Creepy Crawlies, Paradox Lost, Microsoft .NET: The Programming Bible, MSXML3: A Comprehensive Guide, Visual Studio 7: A Comprehensive Guide",
   })
 
-t.create('XML from url | caching with new query params')
```
paulmelnikow: I removed this test ^^ Frankly I have no idea what it is for. It looks like it was copied and pasted, or modified into oblivion.

Issue #1956, opened 2018-08-23
Our public Jenkins builds show up as 'inaccessible'


LeoNatan:
Hello,

Our Jenkins server is publicly available, but for some reason, it shows up as 'inaccessible':

![Build Status](https://img.shields.io/jenkins/s/https/jenkins-oss.wixpress.com/job/multi-detox-master.svg)

As I understand, the renderer uses the json API for Jenkins builds?

This is our json:
https://jenkins-oss.wixpress.com/job/multi-detox-master/api/json 

It is available publicly, but for some reason I am assuming the shields server is unable to reach it?

Thanks

paulmelnikow:
Hi! Thanks for reporting this.

Does the fix from #1809 help?

LeoNatan:
Interesting, using http and 8080 does work: ![Test](https://img.shields.io/jenkins/s/http/jenkins-oss.wixpress.com:8080/job/multi-detox-master.svg).

If possible, we would like to keep SSL. Do you have a log on your end that could shed light on what breaks when attempting to connect through SSL?

paulmelnikow:
Heh, yea, agreed. Only one person has access to the server logs. If the problem is reproducible in dev  and [on staging](https://shields-staging.herokuapp.com/) (it probably is) we could troubleshoot it that way, simply enough.

Btw, did you happen to try using https and specifying port 443?

LeoNatan:
Specifying 443 doesn't work (still 'inaccessible'). 

Reproduces on staging too:
![Test](https://shields-staging.herokuapp.com/jenkins/s/https/jenkins-oss.wixpress.com/job/multi-detox-master.svg)

LeoNatan:
Thanks for the quick responses, btw!

calebcartwright:
Underlying error the Shields server is getting when trying to hit your Jenkins instance:
`Inaccessible: unable to verify the first certificate`

calebcartwright:
I suppose we could set `strictSSL` to false in our request options to resolve this, but curious what others think

paulmelnikow:
Is this the issue from #2812? If so upgrading Node seems like a better solution.

calebcartwright:
I'm not sure if they're related, but either way I still get this same `inaccessible` error for Jenkins on Node v10.9.0

paulmelnikow:
[...]

PR #3302, opened 2019-04-14
refactor [JenkinsBuild]

calebcartwright:
Refs #2863


services/jenkins/jenkins-base.js
```diff
@@ -0,0 +1,29 @@
...
+const serverSecrets = require('../../lib/server-secrets')
+
+module.exports = class JenkinsBase extends BaseJsonService {
+  async fetch({
+    url,
+    schema,
+    qs,
+    errorMessages = { 404: 'instance or job not found' },
+  }) {
+    const options = { qs, strictSSL: false }
```
calebcartwright: I added the `strictSSL` setting while debuging https://github.com/badges/shields/issues/1956. This does technically fix the issue in #1956 but I'm not convinced this is necessarily the path we want to go. 

Curious what others think

paulmelnikow:
Haven't given this a full review, though was playing around with https://github.com/badges/shields/issues/1956#issuecomment-483057062 and I noticed a disparity and thought I'd mention it.

This on master:
```
npm run badge https://img.shields.io/jenkins/s/https/jenkins-oss.wixpress.com/job/multi-detox-master.svg
```
is equivalent to
```
npm run badge https://img.shields.io/jenkins/s/https/jenkins-oss.wixpress.com/multi-detox-master.svg
```
on this branch.

calebcartwright:
Fair point on the paths. It looks to me as if the legacy service did its best to support users whether they simply mentioned a job name `multi-detox-master` or whether they specified a full path that ran through views, like `view/Precise/view/All%20Precise/job/precise-desktop-amd64_default`

There is an issue here with the route path though, currently:
`pattern: ':protocol(http|https)/:host+/:job+'`

As I believe the consecutive `+` params are problematic. The problem I was trying to address is that some instances may be on their own url path, like: https://jenkins.ubuntu.com/server/ that need to support one or more additional `/`, but similarly the user provided values for `job` can have their own N number of `/` as well 🤔

calebcartwright:
I've updated the route to mirror the existing/current route

services/jenkins/jenkins-base.js
chris48s: In terms of shields.io itself, I think we should ideally fix this by upgrading node, but I realise this is currently problematic for...reasons. We could do this as a workaround while that is problematic.

For self-hosting users who want to use this option to deal with self-signed certs, I think we should make this a configurable option rather than turn `strictSSL` off by default

services/jenkins/jenkins-build.service.js
```diff
@@ -1,107 +1,107 @@
...
+    'green',
+    'green_anime',
+    'grey',
+    'grey_anime',
+    'disabled',
+    'disabled_anime',
+    'aborted',
+    'aborted_anime',
+    'notbuilt',
+    'notbuilt_anime'
```
chris48s: Instead of declaring the full list twice, you should be able to `Joi.allow(...Object.keys(colorStatusMap))` here :)

services/jenkins/jenkins-base.js
calebcartwright: > In terms of shields.io itself, I think we should ideally fix this by upgrading node,

This issue still persists on Node 10.x btw, so I think this may be a longer term issue. I through this idea out there in #1956 but I think I'd like to add a query param to these Jenkins badges that folks could use in these exceptional circumstances. I'll only include the `strictSSL` option here when the query param was set. 

Still not ideal IMO but if we end up going down to give Shields users a work-around when there's fix cert chain issues (note this is also happening for Coverity refs #3334) then that would at least keep it pretty narrow.

Thoughts?

services/jenkins/jenkins-build.service.js
calebcartwright: Good call!

services/jenkins/jenkins-base.js
chris48s: OK yes agreed. If we make `strictSSL` on by default but all the user to disable with a URL param I think that is a sensible default/viable workaround :+1:

calebcartwright: Done

services/jenkins/jenkins-base.js
```diff
@@ -0,0 +1,30 @@
...
+
+module.exports = class JenkinsBase extends BaseJsonService {
+  async fetch({
+    url,
+    schema,
+    qs,
+    errorMessages = { 404: 'instance or job not found' },
+    disableStrictSSL,
+  }) {
+    const options = { qs, strictSSL: disableStrictSSL === undefined }
```
calebcartwright: See below comment in `jenkins-common.js` for more info on param name

services/jenkins/jenkins-build.service.js
```diff
@@ -1,107 +1,95 @@
...
 
+  static get defaultBadgeData() {
+    return {
+      label: 'build',
+    }
+  }
+
   static get route() {
     return {
       base: 'jenkins/s',
```
calebcartwright: I don't have any PR #'s handy, but we seem to be trending away from single letter aliases to more explicit route param names. 

I'd like to do the same thing here (`build` or `status` instead of `s`). Any thoughts/objections?

services/jenkins/jenkins-common.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+const Joi = require('joi')
+
+const queryParamSchema = Joi.object({
+  disableStrictSSL: Joi.equal(''),
```
calebcartwright: Although I usually try to refrain from using negated variable names, naming this with the `disable` seemed to be the best approach given the default behavior we want here and the way boolean query params are handled (empty string when set, undefined when not set).

With this approach, a user will have to explicitly include the `disableStrictSSL` query param:
https://shields-staging-pr-3302.herokuapp.com/jenkins/s/https/jenkins-oss.wixpress.com/job/detox-android-56-master.svg?disableStrictSSL


As the default behavior (no query param provided) will still enforce strict ssl, which is why the below badge will fail (there's a cert chain issue with the target jenkins instance)
https://shields-staging-pr-3302.herokuapp.com/jenkins/s/https/jenkins-oss.wixpress.com/job/detox-android-56-master.svg

services/jenkins/jenkins-common.js
```diff
@@ -0,0 +1,20 @@
...
+}).required()
+
+const buildUrl = ({ protocol, host, job, lastBuild = true, plugin }) => {
+  const jobPrefix = job.indexOf('/') > -1 ? '' : 'job/'
+  return `${protocol}://${host}/${jobPrefix}${job}/${
+    lastBuild ? 'lastBuild/' : ''
+  }${plugin ? `${plugin}/` : ''}api/json`
+}
```
calebcartwright: I'm setting this function up to be re-used in all the other Jenkins services (Tests, etc.) which is why it contains things not explicitly used for the Build service badge

services/jenkins/jenkins-build.service.js
```diff
@@ -1,107 +1,95 @@
...
 
   static get examples() {
     return [
       {
         title: 'Jenkins',
-        pattern: ':scheme/:host/:job',
         namedParams: {
-          scheme: 'https',
+          protocol: 'https',
           host: 'jenkins.qa.ubuntu.com',
```
calebcartwright: I am also intentionally not including the new `disableStrictSSL` query param here as I do not believe this is something we want to advertise; it's a work around for a handful of exceptional circumstances only.

services/jenkins/jenkins-build.service.js
chris48s: Happy if you want to change it, but we'll want to maintain an alias/redirect for `/s` for existing users

We talked about formalising some of this stuff in docs in https://github.com/badges/shields/pull/3144#issuecomment-470665539 and then.. haven't got round to doing it yet. There's several things in that discussion that I want to get done at some point but haven't found the time. Having this stuff documented would make it easier when we're thinking

> wait, what **is** our convention for a "build status" URL?

:slightly_smiling_face:

calebcartwright: I'll leave it in place for now. I plan on using some of the common bits in here to refactor the Jenkins test service so I'll consider the alias redirecting in that one

PR #3303, opened 2019-04-14
fix [gratipay] test

calebcartwright:
Not really sure what happened, but the test for the deprecated gratipay/gittp service has been failing for a while with a 404 badge not found error. For example

`AssertionError: label mismatch: expected '404' to equal 'gratipay'`

This tweaks the rout pattern so the test is passing again.


services/gratipay/gratipay.service.js
```diff
@@ -12,14 +12,14 @@ module.exports = [
   deprecatedService({
     route: {
       base: 'gittip',
-      format: '(?:/user|/team|/project)?/(?:.*)',
+      pattern: ':various?',
```
paulmelnikow: I'm in the process of trying to use this in another service. Should this be `:various*`?

calebcartwright: elsewhere yeah it probably should be. the existing test route just used the base `gratipay` so i went with `?` to keep the tests passing without _completely_ dropping the other stuff. Probably doesn't matter all that much for a deprecated service anyway though 😄

paulmelnikow: Hehe it's true. Though I still see a lot of those "no longer available" badges still on readmes!

I think `:various*` (or `:various+`) would match e.g. `/user/foobar` like the old regex. I'll open a PR to update it.

Issue #2547, opened 2018-12-17
Caching layer for discardable credentials


paulmelnikow:
API credentials for the Matrix badge work by creating a guest user. It would be better for performance reasons of the badge and the server not to create a guest user for each request, but to create one once and use it for all requests.

In https://github.com/badges/shields/pull/2527#issuecomment-447990925 I described an adaptation of the GitHub `FsTokenPersistence`/`RedisTokenPersistence` storage layer to support Matrix.

The other places caching has arisen are for Gitlab #1838 which is similar to GitHub, and #1545 which is about caching of intermediate API responses. As I think about the Matrix use case some more, it seems possible that it's better to think of these as intermediate responses, since unlike GitHub and Gitlab API keys, they are discardable.

See also this discussion: https://github.com/badges/shields/issues/1848#issuecomment-434898047

Dependencies: #1848 #1806

PR #3316, opened 2019-04-15
Rewrite [codeclimate] coverage

paulmelnikow:
Attacking this in two pieces for ease of review. The legacy implementation for coverage is still there, though I disabled it via the route. That whole file will be removed in the next PR.

Ref #2863


core/base-service/redirector.spec.js
```diff
@@ -24,6 +24,15 @@ describe('Redirector', function() {
     expect(redirector(attrs).name).to.equal('VeryOldServiceRedirect')
   })
 
+  it('overrides the name', function() {
```
paulmelnikow: This was necessary to avoid a name conflict when adding two redirects with the same base URL (it had caused a validation test to fail).

calebcartwright:
Nice! Especially on pulling it out into parts like this. I was worried that all of the Codeclimate content would have to be refactored together, but this definitely makes it easier to review 👍

core/base-service/redirector.spec.js
calebcartwright: I was going to ask about the `name` addition above in redirector but I get it now 👍

services/codeclimate/codeclimate-common.js
```diff
@@ -0,0 +1,48 @@
...
+          latest_default_branch_test_report: Joi.object({
+            data: Joi.object({
+              id: Joi.string().required(),
+            }).allow(null),
+          }).required(),
+        }).required(),
+      })
+    )
+    .required(),
+}).required()
```
calebcartwright: this looks like it was a super fun schema to figure out 😄

services/codeclimate/codeclimate-coverage.service.js
```diff
@@ -0,0 +1,94 @@
...
+    const {
+      id: repoId,
+      relationships: {
+        latest_default_branch_test_report: { data: testReportInfo },
+      },
+    } = await fetchRepo(this, { user, repo })
+    if (testReportInfo === null) {
+      throw new NotFound({ prettyMessage: 'test report not found' })
+    }
+    const { data } = await this._requestJson({
```
calebcartwright: I did not know we had to make two API calls to get this data from Code Climate. That's unfortunate they don't make it available in one pass

services/codeclimate/codeclimate-coverage.tester.js
```diff
@@ -1,53 +1,33 @@
...
 
-t.create('test coverage percentage alternative top-level URL')
-  .get('/jekyll/jekyll.json')
-  .expectBadge({
-    label: 'coverage',
-    message: isIntegerPercentage,
-  })
-
 t.create('test coverage letter')
-  .get('/c-letter/jekyll/jekyll.json')
```
calebcartwright: 👀 we sure had a lot of alternate route patterns to get coverage data!

services/codeclimate/codeclimate-coverage.service.js
calebcartwright: Perhaps as a future enhancement we could consider allowing users to enter their repo id directly to avoid making that first API call (we could update the docs with steps on how users can find their respective repo id's)

I imagine that might help a bit with the rate limiting issue described in #1518

paulmelnikow: That’s a nice idea! A decent caching layer (a la #2547) could also help with stuff that without making the interface more opaque.

paulmelnikow:
Thanks for reviewing!

PR #3101, opened 2019-02-25
Move "good" badge helpers from lib/ to services/

paulmelnikow:
This moves a few helpers from `lib/` to `services/`:

build-status.js
build-status.spec.js
color-formatters.js
color-formatters.spec.js
contributor-count.js
licenses.js
licenses.spec.js
php-version.js
php-version.spec.js
text-formatters.js
text-formatters.spec.js
version.js
version.spec.js

And one from `lib/` to `core/`:

unhandled-rejection.spec.js

The diff is long, but the changes are straightforward.

Ref #2832

PR #3317, opened 2019-04-15
Migrated [JenkinsPlugin] to new service architecture

PyvesB:
See #2863.


calebcartwright:
Wow you turned that around fast!

paulmelnikow:
> The only question I have is around `regularUpdate` since the file containing it currently resides within the `legacy` directory hierarchy. If that helper will continue to be leveraged in the new service architecture, should that file (eventually) be moved to a different directory?

I think regular-update needs a total rewrite using async/await. Caching code is notoriously difficult to test and so I'd rather base the cache part on `node-cache`.

If this is the last legacy service to use it, the time to rewrite may be upon us!

https://github.com/badges/shields/pull/3101#issuecomment-468089472

PyvesB:
Yes, I believe this is the last legacy service that is using `regularUpdate`.

calebcartwright:
> I think regular-update needs a total rewrite using async/await. Caching code is notoriously difficult to test and so I'd rather base the cache part on node-cache

Do we want to do that here as part of refactoring this service? Or should we go ahead and merge this and then in a follow on tackle rewriting regular-update (and updating this service accordingly)?

paulmelnikow:
Think that can definitely be a follow-on.

PR #3318, opened 2019-04-16
Consolidate [NpmDownloads] into one service

paulmelnikow:
This moves the four npm download services into a single class, in line with #3174, and other service implementations we've written in the last couple months. 
1. Change the suffixes from **/m** to **/month** for consistency.
2. Add tests of one of the intervals besides total.
3. Rewrite mocked service tests as unit tests.
4. Use (and work with) the `intervalMap` pattern that I think @calebcartwright started us using.

### DIFF
--- a/services/npm/npm-downloads.service.js
+++ b/services/npm/npm-downloads.service.js
@@ -10,95 +10,92 @@ const pointResponseSchema = Joi.object({
   downloads: nonNegativeInteger,
 }).required()
 
-// https://github.com/npm/registry/blob/master/docs/download-counts.md#output-1
-const rangeResponseSchema = Joi.object({
-  downloads: Joi.array()
-    .items(pointResponseSchema)
-    .required(),
-}).required()
-
-function DownloadsForInterval(interval) {
-  const { base, messageSuffix = '', query, isRange = false, name } = {
-    week: {
-      base: 'npm/dw',
-      messageSuffix: '/w',
-      query: 'point/last-week',
-      name: 'NpmDownloadsWeek',
-    },
-    month: {
-      base: 'npm/dm',
-      messageSuffix: '/m',
-      query: 'point/last-month',
-      name: 'NpmDownloadsMonth',
-    },
-    year: {
-      base: 'npm/dy',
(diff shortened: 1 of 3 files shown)


services/npm/npm-downloads.service.js
```diff
@@ -10,95 +10,92 @@ const pointResponseSchema = Joi.object({
...
-      })
-      if (isRange) {
-        downloads = downloads
-          .map(item => item.downloads)
-          .reduce((accum, current) => accum + current)
-      }
-      return this.constructor.render({ downloads })
-    }
+  async handle({ interval, scope, packageName }) {
+    const { query, schema, transform } = intervalMap[interval]
```
paulmelnikow: I like using destructuring here. It makes this code particularly readable and I think a bit less error prone.

services/npm/npm-downloads.service.js
```diff
@@ -10,95 +10,92 @@ const pointResponseSchema = Joi.object({
...
+  dw: {
+    query: 'point/last-week',
+    schema: pointResponseSchema,
+    transform: json => json.downloads,
+    messageSuffix: '/week',
+  },
+  dm: {
+    query: 'point/last-month',
+    schema: pointResponseSchema,
+    transform: json => json.downloads,
```
paulmelnikow: A little bit of repetition here, though it helps with clarity.

services/npm/npm-downloads.service.js
calebcartwright: 👍 I like this a lot too

services/npm/npm-downloads.service.js
```diff
@@ -10,95 +10,92 @@ const pointResponseSchema = Joi.object({
...
+  }
 
-    static get route() {
-      return {
-        base,
-        pattern: ':scope(@.+)?/:packageName',
-      }
-    }
+  // For testing.
+  static get _intervalMap() {
```
calebcartwright: Good idea! This makes things easy to test without having to define (and unnecessarily call) a class level render/transform function

PR #3320, opened 2019-04-16
Fix [codeclimatecoverage] coverage-percentage route

paulmelnikow:
When I started rewriting the analysis badge I realized I’d broken this part of the route in #3316.

In the long run we might want to deprecate both `coverage-percentage` and `coverage-letter` and move the format argument into a query parameter, though I think for the moment it suffices to get it working again.

Also adds a comment with an issue link from the legacy implementation.

### DIFF
--- a/services/codeclimate/codeclimate-coverage-redirector.service.js
+++ b/services/codeclimate/codeclimate-coverage-redirector.service.js
@@ -2,6 +2,7 @@
 
 const { redirector } = require('..')
 
+// http://github.com/badges/shields/issues/1387
 module.exports = [
   redirector({
     name: 'CodeclimateCoverageShortcutRedirect',
--- a/services/codeclimate/codeclimate-coverage-redirector.service.js
+++ b/services/codeclimate/codeclimate-coverage-redirector.service.js
@@ -4,7 +4,7 @@ const { redirector } = require('..')
 
 module.exports = [
   redirector({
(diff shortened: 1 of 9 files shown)


paulmelnikow:
As I dig into this more carefully, I realize this route was never advertised and that may be why it wasn't tested. IMO this can be closed; though leaving open in case others have different opinions!

cc @PyvesB; I think you worked on these last.

PyvesB:
Here's what we were supporting and advertising in examples around the time I last heavily worked on the Code Climate service:

| Endpoint | Badge | Should be advertised by example |
|:-------------:|:-------------:|:-------------:|
_(top-level)_ |   |:x:
/c |  |:x:
/c-letter |  |:x:
/coverage |  |:heavy_check_mark:
/coverage-letter |  |:heavy_check_mark:
/maintainability |  |:heavy_check_mark:
/maintainability-percentage |  |:heavy_check_mark:
/issues |  | :heavy_check_mark:
/tech-debt |  |:heavy_check_mark:

The ones not advertised were still supported for backwards compatibility purposes.

calebcartwright:
I'd vote we not display/advertise multiple options in the UI editor dropdown that do the same thing. As such I'd suggest we leave the `which` pattern as-is with `coverage|coverage-letter`, but we should update the redirector config to maintain BC with `coverage-percentage` if we go that route

Issue #2632, opened 2019-01-04
Add not found schema to base services


paulmelnikow:
This code is from the clojar badge:

```js
    if (Object.keys(json).length === 0) {
      /* Note the 'not found' response from clojars is:
          status code = 200, body = {} */
      throw new NotFound()
    }
```

We have a bunch of other similar examples, where the schema is written to accommodate the not found case, and then there's code written to detect the alternate case and throw an error.

This responsibility is a kind of schema validation. We'd have clearer guarantees, particularly in the "found" case, if this responsiblity were delegated to Joi.

A really clean way to express this would be to optionally provide two schemas to `_requestJson` and friends. One would be the `schema` for the success case (like what we're passing now) and the other a `notFoundSchema`. If the `notFoundSchema` matches, we automatically throw `NotFound`, using the 404 error message.

calebcartwright:
Something I've seen (which I think is along the very same lines), is that for not found scenarios some services will return an HTTP 200 with an html page (that contains the _text_ '404 Not Found'). Would it be feasible to have a `notFoundSchema` that can handle that html scenario too?

paulmelnikow:
> some services will return an HTTP 200 with an html page (that contains the _text_ '404 Not Found')

😝 

Are these JSON services? `notFoundSchema` would require valid JSON, though we could have a `notFoundRegex` which we test if the JSON won't parse, before returning the typical "unparseable json".

calebcartwright:
Two places where I've seen this recently are:
* Snyk (for npm package) https://github.com/badges/shields/blob/master/services/snyk/snyk-vulnerability-npm.service.js#L52
* Coverity (https://github.com/badges/shields/blob/master/services/coverity/coverity-scan.service.js#L73)

> Are these JSON services?

😆  They'll return JSON if the query target exists, but doing that HTML nonsense otherwise. I believe the current behavior with those would be `invalid response` for the badge message when the projects are not found since that HTML isn't valid JSON

paulmelnikow:
> I believe the current behavior with those would be `invalid response` for the badge message when the projects are not found since that HTML isn't valid JSON

It'd be nice to improve on that.

PR #3321, opened 2019-04-17
Rewrite [codeclimate] analysis

paulmelnikow:


services/codeclimate/codeclimate-coverage.service.js
```diff
@@ -58,7 +58,6 @@ module.exports = class CodeclimateCoverage extends BaseJsonService {
       url: `https://api.codeclimate.com/v1/repos/${repoId}/test_reports/${
         testReportInfo.id
       }`,
-      errorMessages: { 404: 'test report not found' },
```
paulmelnikow: This was unused.

services/codeclimate/codeclimate-analysis.tester.js
```diff
@@ -77,6 +71,6 @@ t.create('malformed response for inner specific repo query')
   ) // No data.
   .networkOn() // Combined with allowUnmocked: true, this allows the outer user repos query to go through.
   .expectBadge({
-    label: 'maintainability',
-    message: 'invalid',
+    label: 'analysis',
+    message: 'invalid response data',
```
paulmelnikow: It might be worth keeping these two tests. They did catch errors in the way I had things wired up.

services/codeclimate/codeclimate-analysis.service.js
```diff
@@ -0,0 +1,178 @@
...
+    }),
+  },
+}
+
+module.exports = class CodeclimateAnalysis extends BaseJsonService {
+  static get route() {
+    return {
+      base: 'codeclimate',
+      pattern:
+        ':which(maintainability|maintainability-percentage|tech-debt|issues)/:user/:repo',
```
calebcartwright: Do we know what the relationship is between these 3 metrics? Based on the above calculation it seems maintainability and tech-debt are directly related, but what about issues? 

Thinking along the lines of https://github.com/badges/shields/issues/3174, I sometimes find myself wondering when/how to group certain metrics together (note I'm not suggesting this class be split, this class just made me remember that question/thought)  

I could potentially see down the line someone wanting to display different forms of issues (for example based on [severity](https://docs.codeclimate.com/docs/issues#section-issue-severity)) . In that hypothetical, what would the considerations be for adding that inline vs. creating a new class? Would we only split the class if there was a need for a new route parameter?

services/codeclimate/codeclimate-analysis.tester.js
```diff
@@ -43,29 +37,29 @@ t.create('maintainability letter')
...
   })
 
 t.create('malformed response for outer user repos query')
   .get('/maintainability/angular/angular.js.json')
   .intercept(nock =>
     nock('https://api.codeclimate.com')
-      .get('/v1/repos?github_slug=angular/angular.js')
+      .get('/v1/repos?github_slug=angular%2Fangular.js')
       .reply(200, {
         data: [{}], // No relationships in the list of data elements.
```
calebcartwright: I keep going back and forth on this test. What sort of scenarios in real life would result in us getting this API response? If the `/v1/repos` call includes a repo slug that doesn't exist, data is an empty array and we'd display a `repo not found` badge.

I can't picture what would cause this response but if we know of one then maybe we could update this comment to call that out for future observers? Something like `// No relationships in the list of data elements. This happens when the repo...`

calebcartwright: Just realized your below comment covered this test too 😄

services/codeclimate/codeclimate-coverage.service.js
calebcartwright: 👍

services/codeclimate/codeclimate-analysis.tester.js
paulmelnikow: Having thought about this a bit more, I realize the bugs these tests caught were a result of not correctly detecting the not found case (which the schema explicitly lets in). Implementing #2632 would eliminate that error surface.

services/codeclimate/codeclimate-analysis.service.js
paulmelnikow: I've never used Code Climate, though I think issues is showing the number of issues with the code, and the tech debt and maintainability scores are derived from the issue count. So they seem closely related. I was thinking that post-#3174 it would be sufficient to have a single entry in Analysis.

I think the alternate formatting for issues could be done with a query parameter, though if different `which`es support different query parameters, we may need to consider that as part of #3174. Alternatively it may make more sense to make them separate services.

calebcartwright: Thanks that helps!

services/codeclimate/codeclimate-coverage.service.js
```diff
@@ -3,14 +3,14 @@
 const Joi = require('joi')
 const { BaseJsonService, NotFound } = require('..')
 const { coveragePercentage, letterScore } = require('../color-formatters')
-const { keywords, fetchRepo } = require('./codeclimate-common')
+const { keywords, isLetterGrade, fetchRepo } = require('./codeclimate-common')
```
calebcartwright: Just to note a potential follow up that it might even be worth elevating this to broader validators at some point, as we do (or at least soon will) have some other grade data (Sonar comes to mind again)

services/codeclimate/codeclimate-analysis.tester.js
calebcartwright: I still can't wrap my head around what would make the scenario in this first test actually occur (I thought not found would always be `data: []`), but happy to defer to your judgement on keeping these i👍

services/codeclimate/codeclimate-analysis-redirector.service.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+const { redirector } = require('..')
+
+module.exports = [
+  // http://github.com/badges/shields/issues/1387
+  // https://github.com/badges/shields/pull/3320#issuecomment-483795000
+  redirector({
+    name: 'CodeclimateCoverageMaintainabilityRedirect',
+    category: 'coverage',
```
calebcartwright:
```suggestion
    category: 'analysis',
```

Guessing categories don't matter much for redirectors but probably should be `analysis` no?

calebcartwright:
Nice job with this by the way. I really like the way implementation for the different metrics and believe they will be a good reference

services/codeclimate/codeclimate-coverage.service.js
paulmelnikow: When we do this, one thought: some of these have "E"'s in them, and some don't.

services/codeclimate/codeclimate-analysis-redirector.service.js
paulmelnikow: Yea, this category and title were both wrong. Copy pasta!

PR #3324, opened 2019-04-17
Refactor [Wordpress] services

paulmelnikow:
- Prefer inline transforms to take place in `handle()` rather than `render()`
- Avoid inversion of control by removing `BaseWordpress#handle()`, passing `extensionType` into `fetch()`, and removing one layer of subclassing
- Move “not found” checks into `fetch()`
- Cache wordpress versions instead of fetching on each request
- Start to convert aliases to redirects (there are more of these which could be tackled in a follow-on)
- Replace at least one route `format` with a `pattern` (ref #3329)
- Partially reorder: name, category, route, examples, defaultBadgeData, render, fetch, handle

Some of this is in line with our established patterns or makes it clearly easier to follow; some of it is arguably stylistic.


paulmelnikow:
This pull request **introduces 3 alerts** when merging 6481b41d439229a1ad019c70781b1b8e9f41a552 into 5bedbbd9c1c19bbc70e2ecd7f4133ca9d24b78d7 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-86d35a449939774380f2dee575c9d2625636d4c4)

**new alerts:**

* 2 for Unused variable, import, function or class
* 1 for Assignment to constant

---

services/wordpress/wordpress-version-color.js
```diff
@@ -0,0 +1,55 @@
...
+    intervalMillis: 24 * 3600 * 1000,
+    json: true,
+    scraper: json => json.offers.map(v => v.version),
+  })
+}
+
+async function versionColorForWordpressVersion(version) {
+  const offeredVersions = await getOfferedVersions()
+
+  // What is this?
```
calebcartwright: 🤣

services/wordpress/wordpress-base.js
```diff
@@ -25,16 +25,14 @@ module.exports = class BaseWordpress extends BaseJsonService {
     throw new Error(`extensionType() function not implemented for ${this.name}`)
```
calebcartwright: is this still necessary? I _think_ I saw it removed from all the base classes

services/wordpress/wordpress-platform.service.js
```diff
@@ -56,59 +22,38 @@ class WordpressPluginRequiresVersion extends BaseWordpressPlatform {
...
-  static render({ version, color }) {
+  static get defaultBadgeData() {
+    return { label: 'wordpress' }
+  }
+
+  static render({ wordpressVersion }) {
     return {
-      message: `${addv(version)} tested`,
-      color,
+      message: addv(wordpressVersion),
```
calebcartwright: I don't personally have any objections to not including a `tested` suffix in the message, but any concerns about changing a message like this on an existing badge?

services/wordpress/wordpress-base.js
paulmelnikow: Ah, you're correct, it can be removed.

services/wordpress/wordpress-platform.service.js
paulmelnikow: Ack, this is a really gnarly diff, probably easier to inspect in split view. 😝 

The **tested** is still there in `WordpressPluginTestedVersion.render()` but the added code you're seeing is part of `WordpressPluginRequiresVersion.render()`.

services/wordpress/wordpress-version-color.js
paulmelnikow: I think this code has now been moved twice, but probably it should be completely rewritten. This should also have some of its own tests. There is already a lot changing in this PR. I'm not sure it needs to be dealt with now.

services/wordpress/wordpress-platform.service.js
calebcartwright: Oops! Thanks for that

services/wordpress/wordpress-downloads.service.js
```diff
@@ -29,13 +29,21 @@ function DownloadsForExtensionType(extensionType) {
       return `Wordpress${capt}Downloads`
     }
```
calebcartwright: Nothing specific to this line, but just want to note that there seems to be a whole lot happening in this file. If I'm not mistaken it ends up generating six different classes?

Could you envision future changes here, such as combining the plugin download intervals?

services/wordpress/wordpress-downloads.service.js
```diff
@@ -74,21 +79,25 @@ function InstallsForExtensionType(extensionType) {
...
-    }
-
     static get category() {
       return 'downloads'
     }
 
-    static render({ response }) {
+    static render({ installCount }) {
       return {
-        message: `${metric(response.active_installs)}+`,
```
calebcartwright: I may be getting tricked by GH's web dif again, but is this changing the message to drop the trailing `+` suffix?

services/wordpress/wordpress-rating.service.js
```diff
@@ -67,6 +54,22 @@ function RatingForExtensionType(extensionType) {
         },
       ]
     }
+
+    static render({ rating, numRatings }) {
+      const scaled = (rating / 100) * 5
+      return {
+        message: `${scaled.toFixed(1)}/5 (${metric(numRatings)})`,
+        color: floorCount(scaled, 2, 3, 4),
```
calebcartwright: It looks like the legacy service used the fixed/rounded scaled value in the color scale. I think that would technically mean there's a handful of boundary cases with rating values _really_ close to the scale steps that would end up colored differently.

For example if `scaled` has a value of `2.99` then I believe that would be yellowgreen, but in the legacy service (due to the `.toFixed` that same badge would have been green.

services/wordpress/wordpress-downloads.service.js
paulmelnikow: Yes, that change I’ll own up to. It’s the only badge that shows this, and it’s something that’s true of all download and install badges (for various reasons), so I don’t think it makes sense to include it.

services/wordpress/wordpress-rating.service.js
paulmelnikow: Sure, it seems reasonable to use the rounded display value for the color.

services/wordpress/wordpress-downloads.service.js
paulmelnikow: Yea, agreed the download intervals should be combined.

services/wordpress/wordpress-downloads.service.js
calebcartwright: Agreed. It was a little confusing to have that in here before, at least relative to all the other download badges

services/wordpress/wordpress-downloads.service.js
calebcartwright: 👍

PR #3326, opened 2019-04-17
Use url `pattern` for [nexus] + minor cleanup

paulmelnikow:
Ref #3329


services/nexus/nexus.service.js
```diff
@@ -114,10 +113,10 @@ module.exports = class Nexus extends BaseJsonService {
         }),
         documentation: `
         <p>
-          Note that you can use query options with any Nexus badge type (Releases, Snapshots, or Repository)
+          Note that you can use query options with any Nexus badge type (Releases, Snapshots, or Repository).
         </p>
         <p>
-          Query options should be provided as key=value pairs separated by a semicolon
+          Query options should be provided as key=value pairs separated by a colon.
```
calebcartwright: 🤦‍♂️ good catch!

services/nexus/nexus.service.js
```diff
@@ -46,14 +46,13 @@ module.exports = class Nexus extends BaseJsonService {
       base: 'nexus',
       // API pattern:
       // /nexus/(r|s|<repo-name>)/(http|https)/<nexus.host>[:port][/<entry-path>]/<group>/<artifact>[:k1=v1[:k2=v2[...]]]
-      format:
-        '(r|s|[^/]+)/(https?)/((?:[^/]+)(?:/[^/]+)?)/([^/]+)/([^/:]+)(:.+)?',
-      capture: ['repo', 'scheme', 'host', 'groupId', 'artifactId', 'queryOpt'],
+      pattern:
+        ':repo(r|s|[^/]+)/:scheme(http|https)/:hostAndPath+/:groupId/:artifactId([^/:]+):queryOpt(:.+)?',
```
calebcartwright: I'm updating the `hostAndPath` param name in the examples too as I think it'll be helpful for folks to see that in the editor window

calebcartwright:
The changes in here look good and I'm ready to approve. However, while playing with the examples I noticed we're getting invalid response data on many of them.

I don't _think_ anyone has raised an issue yet on this specific item, but definitely needs a review. I'd suggest doing so later on/outside this PR.

Thoughts?

calebcartwright:
https://developers.redhat.com/auth/realms/rhd/protocol/openid-connect/auth?scope=openid&response_type=code&redirect_uri=https%3A%2F%2Frepository.jboss.org%2Fnexus%2Fsec&state=6a99060e-0c6e-480c-b968-b938ae92aba4&login=true&client_id=nexus 

Looks like RH is requiring auth now for https://repository.jboss.org/nexus 🤔

paulmelnikow:
Is that why the tests are failing too?

calebcartwright:
> Is that why the tests are failing too?

Yes I believe so. We could swap the test targets to the OSS Sonatype instance (although that's where I was having trouble with the example).

The failing service tests are all now showing the below `inaccessible` error:
```
AssertionError: message mismatch: expected 'inaccessible' to equal 'artifact or version not found'
```

calebcartwright:
I found the issue with the schema that was rejecting our examples (like https://oss.sonatype.org/service/local/lucene/search?g=com.google.guava&a=guava). 

TL;DR that query will include every published version for that artifact, and the format for one of the fields can be in a different format (like `HEAD-android-SNAPSHOT`)

I've updated the schema accordingly, and I'm going to update the examples and tests to use the the OSS Sonatype instance (as the RH instance would require auth and a self-hosted Shields server)

services/nexus/nexus.tester.js
```diff
@@ -18,32 +18,36 @@ function mockNexusCreds() {
 }
 
 t.create('live: search release version valid artifact')
-  .get('/r/https/repository.jboss.org/nexus/jboss/jboss-client.json')
+  .timeout(15000)
```
calebcartwright: Now I remember why we were using the jboss instance for these tests.. the sonatype instance is periodically really slow leading to test timeouts.

calebcartwright:
I've approved but I am going to defer to another maintainer to merge since I've also added some commits to this PR

services/nexus/nexus.service.js
```diff
@@ -16,7 +16,10 @@ const searchApiSchema = Joi.object({
       Joi.object({
         latestRelease: optionalDottedVersionNClausesWithOptionalSuffix,
         latestSnapshot: optionalDottedVersionNClausesWithOptionalSuffix,
-        version: optionalDottedVersionNClausesWithOptionalSuffix,
+        version: Joi.alternatives().try(
+          optionalDottedVersionNClausesWithOptionalSuffix,
+          Joi.string()
```
chris48s: Given `optionalDottedVersionNClausesWithOptionalSuffix` is a special case of `string()`, I don't think we're actually gaining anything here by testing against `optionalDottedVersionNClausesWithOptionalSuffix` first if we'll fall back to accepting any string anyway.

Also should it be `Joi.string().required()`?

calebcartwright: > Also should it be Joi.string().required()?

No I don't think it can be required. I will review some previous issues we had with this schema to find out defnitively, but I believe these fields were set to optional intentionally because they are not always present on each object in the response array. 

> Given optionalDottedVersionNClausesWithOptionalSuffix is a special case of string(), I don't think we're actually gaining anything here by testing against optionalDottedVersionNClausesWithOptionalSuffix first if we'll fall back to accepting any string anyway.

That's a logical point. I do think the scenario of one of the previously published versions for an artifact having a version that doesn't include a digit is much more the exception than the norm (to my knowledge no one has reported the issue I saw while working on this).

I do think there's value in a future reader looking at this code and understanding that this field is generally expected to follow the same version pattern, but that there are a couple exceptional circumstances that necessitate a less-strict option. 

Perhaps it would be better change the schema to `Joi.string()` as you suggest, and include an inline comment to convey the above rationale?

chris48s: Right yeah clue's in the name **optional**DottedVersion.. :)

I do see your point on making it clear that `version` will usually follow the same pattern as `latestRelease`/`latestSnapshot` (except when it doesn't). I think I have a slight preference for just explaining it in a comment, but its fine to present that in either of the ways you've suggested really.

PR #2275, opened 2018-11-05
Rewrite [codacy] badges; rm unused svg helpers

paulmelnikow:

PR #3327, opened 2019-04-17
Use route `pattern`s in [snap-ci nsp gratipay githubmanifest codacy cauditor bithound]

paulmelnikow:
Ref #3329


services/codacy/codacy-grade.service.js
```diff
@@ -14,8 +14,7 @@ module.exports = class CodacyGrade extends BaseSvgScrapingService {
   static get route() {
     return {
       base: 'codacy/grade',
-      format: '(?:grade/)?(?!coverage/)([^/]+)(?:/(.+))?',
-      capture: ['projectId', 'branch'],
+      pattern: ':projectId/:branch*',
```
chris48s: Is the behaviour change here intentional or an oversight?

https://img.shields.io/codacy/grade/grade/59d607d0e311408885e418004068ea58.svg is a valid call (admittedly this is really odd - I'd never noticed it before) but https://shields-staging-pr-3327.herokuapp.com/codacy/grade/grade/e27821fb6289410b8f58338c7e0bc686.svg isn't.

paulmelnikow: Deliberate at this point.

Prior to #2275 it's clear we used to support some other variations. Not completely clear to me what they were, though maybe `/codacy` was an alias for `/codacy/coverage`?

I made a mistake in #2275. I don't think we ever intended `/codacy/grade/grade` to work, and we're not advertising that.

Whatever I did break has been broken since November without a complaint, so I'm inclined to let it go until someone asks.

Issue #3033, opened 2019-02-18
Badge suggestion feature not working


paulmelnikow:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)

:beetle: **Description**

Entering e.g. http://example.com into the suggest box and clicking **Suggest badges** throws an exception and generates this stack trace:

```
Uncaught (in promise) TypeError: Cannot read property 'replace' of undefined
    at n (make-badge-url.js:53)
    at Array.map (<anonymous>)
    at staticBadgeUrl (make-badge-url.js:65)
    at a.r.renderExample (badge-examples.js:41)
    at badge-examples.js:90
    at Array.map (<anonymous>)
    at a.r.render (badge-examples.js:90)
    at Ri (react-dom.production.min.js:169)
    at Ci (react-dom.production.min.js:168)
    at Li (react-dom.production.min.js:176)
```

calebcartwright:
Sidenote, I never really knew the Suggest badges feature even existed 😄 I feel like I'd read mentions of it before but it never really resonated til now

paulmelnikow:
It's not very discoverable, is it?

calebcartwright:
It's almost like an easter egg!

salmanzafar40:
Still same error
![screenshot from 2019-03-01 14-17-40](https://user-images.githubusercontent.com/29972877/53628854-eeda7580-3c2d-11e9-8712-a8ba08f357c2.jpg)

mithro:
This still appears to be happening from what I can see...

paulmelnikow:
If you're experiencing this issue too, feel free to 👍 the top post. That's what those are for :)

platan:
After fixing this bug end-to-end tests (introduced in https://github.com/badges/shields/pull/3262) can be used to test badge suggestion feature :-).

salmanzafar40:
i'm facing this issue again.

```
react-dom.production.min.js:198 TypeError: Cannot read property 'replace' of undefined
    at o (make-badge-url.js:53)
    at Array.map (<anonymous>)
    at staticBadgeUrl (make-badge-url.js:65)
    at A (badge-examples.js:37)
    at Xo (react-dom.production.min.js:150)
    at Mi (react-dom.production.min.js:178)
    at $a (react-dom.production.min.js:232)
    at Ka (react-dom.production.min.js:233)
    at Ol (react-dom.production.min.js:249)
    at Cl (react-dom.production.min.js:248)
[...]

PR #3331, opened 2019-04-17
Badge suggestion feature fix

platan:
Fix for https://github.com/badges/shields/issues/3033

The last working commit with suggestion is 7955f460f37f7cfaa1e9588cb84aa050f2f14275

<details>
<summary>Screenshots of the last working version</summary>

Screenshot_2019-04-17 Shields io Quality metadata badges for open source projects
Screenshot_2019-04-17 Shields io Quality metadata badges for open source projects(1)
Screenshot_2019-04-17 Shields io Quality metadata badges for open source projects(2)
</details>

If you want to test this change locally please add a `private.gh_token` to `config/local.yml`. 

Changes in this PR:
- `BadgeExamples` requires `examples` instead of `definitions` (service definitions)
- `BadgeExamples` is able to prepare a real badge as a preview
- customization form fills inputs with details of suggested badges
- end to end tests

<details>
<summary>Screenshots after fix</summary>

Screenshot_2019-04-17 Shields io Quality metadata badges for open source projects(3)
Screenshot_2019-04-17 Shields io Quality metadata badges for open source projects(4)

Compare the form above with a form of a regular, non-suggested badge:
Screenshot_2019-04-17 Shields io Quality metadata badges for open source projects(5)

</defails>


services/suggest.js
```diff
@@ -102,7 +124,6 @@ async function findSuggestions(githubApiProvider, url) {
 //  - suggestions: list of objects of the form:
 //    - title: string
 //    - link: target as a string URL.
-//    - path: shields image URL path.
```
platan: I have to update this doc.

frontend/components/badge-examples.js
```diff
@@ -33,24 +30,29 @@ const ClickableCode = styled(StyledCode)`
...
     baseUrl,
-    path: pattern,
+    pattern,
     namedParams,
     queryParams,
   })
 
+  let previewUrl
+  if (preview.buildFromExample) {
+    previewUrl = exampleUrl
```
paulmelnikow: 👍

frontend/components/badge-examples.js
```diff
@@ -71,16 +73,11 @@ Example.propTypes = {
   onClick: PropTypes.func.isRequired,
 }
 
-export default function BadgeExamples({ definitions, baseUrl, onClick }) {
-  const flattened = definitions.reduce((accum, current) => {
-    const { examples } = current
-    return accum.concat(examples)
-  }, [])
-
+export default function BadgeExamples({ examples, baseUrl, onClick }) {
```
paulmelnikow: 👍

frontend/components/customizer/customizer.js
```diff
@@ -18,16 +18,12 @@ export default class Customizer extends React.Component {
     exampleNamedParams: objectOfKeyValuesPropType,
     exampleQueryParams: objectOfKeyValuesPropType,
     initialStyle: PropTypes.string,
+    showActualParams: PropTypes.bool,
```
paulmelnikow: Could you call this something like `isPrefilled`?

frontend/components/customizer/customizer.js
```diff
@@ -18,16 +18,12 @@ export default class Customizer extends React.Component {
...
+    link: PropTypes.string,
```
paulmelnikow: Would it be better to pass this in through `exampleQueryParams`?

frontend/components/customizer/path-builder.js
```diff
@@ -211,9 +215,11 @@ export default class PathBuilder extends React.Component {
...
           </NamedParamLabelContainer>
           {this.renderNamedParamInput(token)}
-          <NamedParamCaption>
-            {namedParamIndex === 0 ? `e.g. ${exampleValue}` : exampleValue}
-          </NamedParamCaption>
+          {!this.props.showActualParams && (
+            <NamedParamCaption>
+              {namedParamIndex === 0 ? `e.g. ${exampleValue}` : exampleValue}
+            </NamedParamCaption>
+          )}
```
paulmelnikow: 👍

frontend/lib/service-definitions/service-definition-prop-types.js
```diff
@@ -10,18 +10,20 @@ const objectOfKeyValuesPropType = PropTypes.objectOf(PropTypes.string)
...
     label: PropTypes.string,
-    message: PropTypes.string.isRequired,
-    color: PropTypes.string.isRequired,
+    message: PropTypes.string,
+    color: PropTypes.string,
     style: PropTypes.string,
     namedLogo: PropTypes.string,
-  }).isRequired,
+    buildFromExample: PropTypes.bool,
+  }),
```
paulmelnikow: The goal of this schema is that it's kept in sync with `core/base-service/service-definitions.js`. Would it be possible to create a new schema that for `<BadgeExamples />`? You could refactor this one if you want to reuse part of it.

frontend/lib/service-definitions/service-definition-prop-types.js
```diff
@@ -10,18 +10,20 @@ const objectOfKeyValuesPropType = PropTypes.objectOf(PropTypes.string)
 
 const examplePropType = PropTypes.exact({
   title: PropTypes.string.isRequired,
+  link: PropTypes.string,
```
paulmelnikow: It may be appropriate to pass `link` through `example.queryParams`.

services/suggest.integration.js
```diff
@@ -63,29 +63,52 @@ describe('GitHub badge suggestions', function() {
           {
             title: 'GitHub issues',
             link: 'https://github.com/atom/atom/issues',
-            path: '/github/issues/atom/atom',
+            example: {
+              pattern: '/github/issues/:user/:repo',
+              namedParams: { user: 'atom', repo: 'atom' },
+              queryParams: {},
```
paulmelnikow: Love this 😁

services/suggest.js
```diff
@@ -22,8 +22,14 @@ function twitterPage(url) {
     link: `https://twitter.com/intent/tweet?text=Wow:&url=${encodeURIComponent(
```
paulmelnikow: As mentioned elsewhere, link could go in `example.queryParams`.

frontend/components/customizer/customizer.js
platan: Thanks. Naming is a tough task :-)

services/suggest.js
platan: done

frontend/components/customizer/customizer.js
platan: Good idea. The `link` is moved to query params in all places. This have one consequence - `link` is automatically added to a badge url and to a customization form:

image
image

Of course we can change it somehow, but I would like to hear your opinion.

paulmelnikow: This is just in the suggested badges, right? I think that is just fine! 😁

frontend/lib/service-definitions/service-definition-prop-types.js
platan: Thanks for this suggestion. This part of code wasn't perfect and I was hoping for your ideas. I created a separate schema.

frontend/components/customizer/query-string-builder.js
```diff
@@ -175,10 +180,11 @@ export default class QueryStringBuilder extends React.Component {
         <td>
           {isStringParam ? (
             <QueryParamInput
-              checked={value}
               name={name}
               onChange={this.handleServiceQueryParamChange}
+              size={30}
               type="text"
+              value={value}
```
platan: This is a `text` input, so we should have a `value` instead of `checked` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text).

platan:
I've addressed all issues.

frontend/components/customizer/customizer.js
```diff
@@ -129,24 +124,37 @@ export default class Customizer extends React.Component {
     this.setState({ queryString })
   }
 
+  constructor(props) {
+    super(props)
+    const { link } = { ...this.props.exampleQueryParams }
+    this.state = {
+      link: link || '',
```
paulmelnikow: I think this is now stored in `state.queryString.link`, and if it's edited by the user, that's the value that will be updated. Is that right? Should that replace `this.state.link`?

frontend/components/markup-modal/markup-modal-content.js
```diff
@@ -31,7 +31,8 @@ export default class MarkupModalContent extends React.Component {
       example: {
         title,
         example: { pattern, namedParams, queryParams },
-        preview: { style: initialStyle },
+        link,
```
paulmelnikow: I think this `link` property can be removed. As far as I can tell `<Customizer>` doesn't (or doesn't need to) use `props.link` since that value is in `props.exampleQueryParams.link` instead.

frontend/components/customizer/customizer.js
platan: There are two `links` actually and we mixed them by moving link to `queryParams` :-)
- the first `link` was is used in markdown markup to link badge image to some resource https://github.com/badges/shields/blob/07a8d2ecbe8b178a4612e698deb275cf19ed096e/frontend/components/customizer/customizer.js#L77-L82 in https://user-images.githubusercontent.com/1526000/56317336-c1557700-615c-11e9-98b3-c25b5eb0fec4.png (the last working version https://github.com/badges/shields/commit/7955f460f37f7cfaa1e9588cb84aa050f2f14275) we had a separate input for link used as a markup link
- the second `link` is a query param used as a link in svg badge

Which one `link` should be the suggested link?

paulmelnikow: Ahhh I see! That makes a lot of sense! Sorry for forgetting these were two different things.

Frankly I don't think the embedded SVG links are helpful except in very specific situations. So the suggested link should be used in the markdown, not in the query params.

platan:
@paulmelnikow thank you for the review

chris48s:
Well done both for working on this - it looks like it was a bit of a monster.

One question:
Since this was merged, we seem to have changed from showing URLs as placeholder in the badge index e.g: `:user/:repo` to showing example URLs again e.g: `badges/shields` regardless of whether we're using the suggest feature or not.
Was that an intended change in this PR, or an accident? I can't see a comment on in reading through the review notes here.

Before:

Screenshot at 2019-05-06 00-19-12

After:

Screenshot at 2019-05-06 00-19-56

paulmelnikow:
I noticed that too though TBH did not attribute it to this PR. So at least _I_ didn't intend it.

Assuming we're good with returning to patterns, I'd suggest we drop both the hostname and the multiple-choice bits from patterns while we're at it. So `:vcsName(github|gh|bitbucket|bb|gl|gitlab)` becomes `:vcsName`, and `http://localhost:8080/` goes away.

platan:
@chris48s well spotted. I wasn't intended. I will fix this and I will apply Paul's suggestions as well.

PR #3336, opened 2019-04-19
add strictSSL query param toggle to [coverity]

### LINKED ISSUE #3334: Coverity badge always gets "inaccessible" status

- [ ] My own instance

:beetle: **Description**

The Coverity Scan badge for [MD4C](https://github.com/mity/md4c) project always shows in a gray color with the status message as "inaccessible". Noticed this few days ago for the first time. But before that it used to work correctly for a long time.

I've just checked https://scan.coverity.com/projects/mity-md4c/badge.json and it seems to get the following JSON correctly:
```
{"project":"mity/md4c","status":"passed","message":"passed"}
```

:link: https://img.shields.io/coverity/scan/mity-md4c.svg?label=coverity%20scan

calebcartwright:
Close #3334 

As detailed in #3334, Coverity has a cert chain issue which prevents us from being able to render our Coverity badges (internally, we get `Inaccessible: unable to verify the first certificate` errors on cert chain issues, the badges render as `inaccessible`). We've seen this with a couple other services (or instances of self-hosted tools, like #1956)

Ideally the upstream providers would resolve, but our Coverity badge users have expressed understandable concern around the timeframe for if/when Coverity (Synopsys) would be able to resolve this. We receive the cert verification error on both Node 8.x and 10.x so I believe it's an issue that we'll need to address. 

This proposed solution adds a new a new query param `disableStrictSSL` that will allow the user to opt-in to disabling the strict ssl check, thus re-enabling Coverity badges. It's the same solution we seem to be going with for [Jenkins](https://github.com/badges/shields/pull/3302#discussion_r276838069)

Default behavior (strict ssl check enabled per `request`'s defaults):
https://shields-staging-pr-3336.herokuapp.com/coverity/scan/3997.svg

https://shields-staging-pr-3336.herokuapp.com/coverity/scan/3997.svg?disableStrictSSL

### DIFF
--- a/services/coverity/coverity-scan.service.js
+++ b/services/coverity/coverity-scan.service.js
@@ -10,10 +10,6 @@ const schema = Joi.object({
     .required(),
 }).required()
 
-const queryParamSchema = Joi.object({
-  disableStrictSSL: Joi.equal(''),
-}).required()
-
 module.exports = class CoverityScan extends BaseJsonService {
   static render({ message }) {
     let color
@@ -48,7 +44,6 @@ module.exports = class CoverityScan extends BaseJsonService {
     return {
       base: 'coverity/scan',
       pattern: ':projectId',
-      queryParamSchema,
     }
   }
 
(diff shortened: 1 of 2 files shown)


chris48s:
With Jenkins CI, I think its right for this to be user-configurable because its not a centralised service. We should assume that usually we can contact a jenkins instance with `strictSSL` on, but some people may be running an instance where there is an SSL cert issue that warrants turning it off. In that case they've got an option.

Coverity is different because its a centralised service. We know we're going to get an SSL error attempting to call https://scan.coverity.com/ in all cases (at leats until the cert changes upstream). There is no user who _doesn't_ need to add `?disableStrictSSL` to fix their badge, so I think in this case we should just `disableStrictSSL` by default, as opposed to being broken by default.

calebcartwright:
That works for me 👍

Issue #3328, opened 2019-04-17
Modernizing the server


paulmelnikow:
Since the legacy services have almost all been rewritten and the modern services are well abstracted from the server, I'd like to start work on modernizing the server itself.

The server is not a lot of code, though it is a bit tricky, and it's code that I'd like to be as transparent as possible to new people who are reading it.

Scoutcamp has served this project well for a long time, and much respect is due to @espadrine and @jankeromnes for creating something that has performed so well for so long. However there are some downsides, mainly that the framework isn't fully documented, the code is difficult to read, and it hasn't gotten wide adoption. I'd like to move forward with something else.

Criteria would be:

- Good documentation, active maintenance, and high test coverage.
- Native support for async/await support.
- Works well with [path-to-regexp](https://github.com/pillarjs/path-to-regexp) and provides an implementation path for trie-based routing.

I've started playing around with [micro](https://github.com/zeit/micro), [micro-router](https://github.com/pedronauck/micro-router), and [routington](https://github.com/pillarjs/routington). One idea is to create a `micro-trie-router` using `routington` and based on code from `micro-router`.

paulmelnikow:
One piece that should be tackled as part of this task is generating JSON badges directly, instead of indirecting through a template.

chris48s:
The concept of replacing scoutcamp with something else has been on my mind again.

I think last time we discussed it in much detail we said one of the bits of work we wanted to get done first was replacing request with a modern async/await based http client. This was done (slowly) under https://github.com/badges/shields/issues/4655

In common with that issue, I guess there are actually two jobs here: One is the legwork of replacing scoutcamp with $SOMETHINGELSE, which is going to be a big and difficult job. However before that, we have to decide what $SOMETHINGELSE is. Tbh, working that out could take as long so I figured it is useful to start that discussion.. has anyone on the core team got any thoughts on this?

paulmelnikow:
A couple weeks back, had an offline discussion with Chris about this, and wanted to propose a next step.

Basically our proposal is that $SOMETHINGELSE be Express. It’s not the hottest or fastest-growing framework, but it’s maintained, well-documented, and will be around for a long time.

Shields basically already has a custom framework, with few points of contact between scout camp and the service classes. Express is really flexible, so I think it’s a pretty good candidate for gluing in a higher-level framework like ours.

In my professional work, I use Express similarly – with a higher-level framework [Postgraphile](https://www.graphile.org/postgraphile/) – and it works very well.

[...]

PR #3337, opened 2019-04-20
refactor [JenkinsTests JenkinsCoverage], also run [JenkinsBuild]

calebcartwright:
Refs #2863 

Summary of changes:
* JenkinsTests badge refactored to new service arch. 
  * The badge message was also changed to match our other Test badges (details inline below)
  * Added a couple helpers for test-specific badge services with the intent that they be reused for our other test badges (currently Appveyor and Azure DevOps, CircleCI and Sonar in the future, etc.)
* JenkinsCoverage badge was refactored to leverage the new Jenkins related content added here and in #3302. 
  * I believe this also helps align more closely to the objectives in #3174  
  * Service tests updated to use Jenkins instances that appear to be more stable/performant
* The 3 Jenkins routes (Build, Coverage, and Tests) have been updated from their original shorthand/aliases to more explicit route names (with redirects to maintain backwards compatibility):
  * `/jenkins/s` --> `/jenkins/build`
  * `/jenkins/c` --> `/jenkins/coverage`
  * `/jenkins/t` --> `/jenkins/tests`

I know there's still ongoing discussion around what our url conventions are for `build`, but using `build`, `coverage`, and `tests` for Jenkins aligns the url's with the other service (Azure DevOps) we provide multiple metric badges for. 

Using `build` also seems to be _relatively_ common in our badge url's for other services where the service name doesn't automatically default to build status (like we do for TravisCI). Definitely open to feedback/suggestions though!


services/test-results.js
```diff
@@ -84,8 +84,31 @@ function renderTestResultBadge({
   return { message, color }
 }
 
+const getDocumentation = ({ route }) => `
```
calebcartwright: This is one of those things that I saw duplicated in [appveyor-tests.service.js](https://github.com/badges/shields/blob/master/services/appveyor/appveyor-tests.service.js#L9) and [azure-devops-tests.service.js](https://github.com/badges/shields/blob/master/services/azure-devops/azure-devops-tests.service.js#L32). 

Instead of replicating it again for JenkinsTests, I figured it would be better to put it here in one place and consume it from the other services (if there's no opposition to this, I will open a future PR to update Appveyor and Azure DevOps test services to leverage this). We've also got some additional test badges that have been requested (for example CircleCI and Sonar) that can re-use this as well.

services/test-validators.js
```diff
@@ -91,6 +91,39 @@ const isDependencyState = withRegex(
   /^(\d+ out of date|\d+ deprecated|up to date)$/
 )
 
+const getTestTotalsValidator = ({ passed, failed, skipped }) =>
```
calebcartwright: Same comment as above, as both Appveyor and Azure DevOps Tests were replicating this same functional code in multiple places.

Instead of replicating it again for JenkinsTests, I figured it would be better to put it here in one place and consume it from the other service testers (if there's no opposition to this, I will open a future PR to update Appveyor and Azure DevOps test services to leverage this). We've also got some additional test badges that have been requested (for example CircleCI and Sonar) that can re-use this as well.

services/jenkins/jenkins-tests.service.js
```diff
@@ -1,110 +1,154 @@
...
+    passed,
+    failed,
+    skipped,
+    total,
+    passedLabel,
+    failedLabel,
+    skippedLabel,
+    isCompact,
+  }) {
+    return renderTestResultBadge({
```
calebcartwright: This is a clear change to our badge message for Jenkins Tests (before the refactor the message was: `${passed} / ${total}`.

However, I believe this is worthwhile as reusing the `renderTestResultBadge` helper here provides a lot of new formatting capabilities and brings the Jenkins Test badge message format in alignment with our other test badges (Appveyor and Azure DevOps)

services/jenkins/jenkins-coverage.tester.js
```diff
@@ -1,176 +1,27 @@
...
-const t = new ServiceTester({
-  id: 'jenkins-coverage',
-  title: 'JenkinsCoverage',
-  pathPrefix: '/jenkins',
-})
-module.exports = t
+// The below page includes links to various publicly accessible Jenkins instances
+// although many of the links are dead, it is is still a helpful resource for finding
+// target Jenkins instances/jobs to use for testing.
+// https://wiki.jenkins.io/pages/viewpage.action?pageId=58001258
```
calebcartwright: I've updated the tests to use different Jenkins instances that seem to be much more performant and stable than the Apache instance. That gives us good test targets to validate both jacoco and cobertura coverage formats, so I've removed several of the mock tests accordingly (I believe we were mocking due to the lack of consistently available live jobs)

services/jenkins/jenkins-coverage.service.js
```diff
@@ -1,55 +1,60 @@
...
+              name: Joi.string().required(),
+              ratio: Joi.number()
+                .min(0)
+                .max(100)
+                .required(),
+            })
+          )
+          .has(Joi.object({ name: 'Lines' }))
+          .min(1)
+          .required(),
```
calebcartwright: Jenkins will return an HTTP 404 if a job is requested that did not publish code coverage in the cobertura format (which is why on a 404 we use the error message `job or coverage not found`). 

If the target job has cobertura coverage then the elements array will include an object with a `Lines` field, which is what we need.

services/jenkins/jenkins-build-redirect.service.js
```diff
@@ -6,11 +6,21 @@ module.exports = [
...
+    dateAdded: new Date('2019-04-20'),
+  }),
+  redirector({
+    category: 'build',
+    route: {
+      base: 'jenkins/s',
+      pattern: ':protocol(http|https)/:host+/:job+',
+    },
+    transformPath: ({ protocol, host, job }) =>
+      `/jenkins/build/${protocol}/${host}/${job}`,
```
paulmelnikow: It's not immediately obvious these are the same. Maybe you could define the `transformPath` function above to help clarify?

services/jenkins/jenkins-coverage-redirector.service.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+const { redirector } = require('..')
+
+module.exports = [
```
paulmelnikow: Since there's only one, you can `module.exports = redirector({ ... })` and then you can use `createServiceTester()`.

services/jenkins/jenkins-coverage.tester.js
```diff
@@ -1,176 +1,27 @@
...
```
paulmelnikow: 🙌

services/jenkins/jenkins-tests.service.js
```diff
@@ -1,110 +1,154 @@
...
+  }
 
-        if (serverSecrets.jenkins_user) {
-          options.auth = {
-            user: serverSecrets.jenkins_user,
-            pass: serverSecrets.jenkins_pass,
-          }
-        }
+  transform({ json }) {
+    const testsObject = json.actions.find(o => o.hasOwnProperty('failCount'))
```
paulmelnikow: `.hasOwnProperty` is a pretty archaic form that I don't think we need to use anymore.

```suggestion
    const testsObject = json.actions.find(o => 'failCount' in o)
```

services/test-results.js
paulmelnikow: Sounds good! I'm not sure it's essential to put the route in there, and leaving it out would simplify this, though I don't object.

services/test-validators.js
paulmelnikow: Looks nice! Appreciate the cleanup, definitely makes this more enjoyable to read.

I've a slight preference to call these `makeTestTotalsValidator`.

services/validators.js
```diff
@@ -2,11 +2,14 @@
...
+  .min(0)
+
 module.exports = {
-  nonNegativeInteger: Joi.number()
-    .integer()
-    .min(0)
-    .required(),
+  optionalNonNegativeInteger,
+
+  nonNegativeInteger: optionalNonNegativeInteger.required(),
```
paulmelnikow: Bet there are a lot of places where we can use this!

services/jenkins/jenkins-build-redirect.service.js
calebcartwright: I'm not following, could you rephrase? 

The legacy route accepted a base of `jenkins|jenkins-ci` and for the metric it accepted `s` as a shorthand for build status. Are you suggesting combining that into a single redirector, perhaps like the below, or something else entirely?

```js
  redirector({
    category: 'build',
    route: {
      base: '',
      pattern: ':alias(jenkins|jenkins-ci)/s/:protocol(http|https)/:host+/:job+',
    },
    transformPath: ({ protocol, host, job }) =>
      `/jenkins/build/${protocol}/${host}/${job}`,
    dateAdded: new Date('2019-04-20'),
```

paulmelnikow: Ah, sorry!

These two redirects are basically doing the same thing; one is for /jenkins/s` and the other for `/jenkins-ci/s`. This would be clearer if the `transformPath` functions, which are identical, weren't duplicated. Could you define the `transformPath` function above like:

```js
function transformPath({ protocol, host, job }) {
  ...
}
```

calebcartwright: oh also `host should be `.../:host/...` without the `+`

calebcartwright: Got it, thanks for the clarification. Do you think it would be more clear with two redirector services (leveraging a single transform function), or combining them into a single redirector?

services/test-validators.js
calebcartwright: Renamed accordingly

services/jenkins/jenkins-coverage-redirector.service.js
calebcartwright: Nice! For some reason I had it in my mind that `createServiceTester` couldn't be used for redirects but that's for deprecated services

services/jenkins/jenkins-build-redirect.service.js
paulmelnikow: I _think_ two would be more performant if we do #3328, so thinking ahead, I'd be inclined to stick with that!

services/jenkins/jenkins-coverage-redirector.service.js
calebcartwright: Done!

services/test-results.js
calebcartwright: Yeah I think having the full route in the doc was more helpful back before the window we have today that will auto populate the the query params for users. 

I can remove that section (and change this to a simple static prop) if we're good with no longer including that

services/jenkins/jenkins-build-redirect.service.js
```diff
@@ -2,15 +2,28 @@
...
 const { redirector } = require('..')
 
+const transformPath = ({ protocol, host, job }) =>
+  `/jenkins/build/${protocol}/${host}/${job}`
+
+const commonProps = {
+  category: 'build',
+  transformPath,
+  dateAdded: new Date('2019-04-20'),
+}
```
paulmelnikow: 👍

calebcartwright:
One last commit coming with a change to the doc

PR #3338, opened 2019-04-20
[Bit] add bit components count service

guysaar223:
add badge show Bit (bit.dev) components count


calebcartwright:
Deploying a review app @ https://shields-staging-pr-3338.herokuapp.com/

services/bit/bit-components.service.js
```diff
@@ -0,0 +1,59 @@
...
+}).required()
+
+module.exports = class bitComponents extends BaseJsonService {
+  async fetch({ owner, scope }) {
+    const url = `https://api.bit.dev/scope/${owner}/${scope}/`
+    return this._requestJson({
+      url,
+      schema: ownerSchema,
+      errorMessages: {
+        404: 'not found',
```
calebcartwright: `not found` is the default message on 404 errors, so you don't need to include this unless you want to customize that message (for example if you'd prefer 404 errors display something like `component not found`)

services/bit/bit-components.service.js
```diff
@@ -0,0 +1,59 @@
+'use strict'
+
+const Joi = require('joi')
+const { BaseJsonService } = require('..')
+const { metric } = require('../text-formatters')
+
+const ownerSchema = Joi.object({
+  payload: Joi.object({
+    totalComponents: Joi.number(),
+  }),
```
calebcartwright: Do you know if it's possible for the API to return a response object that is missing the `payload` object entirely, or the `payload` object to be missing the `totalComponents` key?

If not, then we should make `payload` a required object, and for `totalComponents` you should be able to use the `nonNegativeInteger` helper from `validators.js`

If those scenarios _are_ possible then we'll need to make sure those are handled in the code below

services/bit/bit-components.service.js
```diff
@@ -0,0 +1,59 @@
...
+  static render({ count }) {
+    return { message: metric(count), color: '#73398D' }
+  }
+
+  async handle({ owner, scope }) {
+    const json = await this.fetch({ owner, scope })
+    return this.constructor.render({ count: json.payload.totalComponents })
+  }
+
+  // Metadata
```
calebcartwright: Minor nit, but this comment line can be removed

services/bit/bit-components.service.js
```diff
@@ -0,0 +1,59 @@
...
+
+const ownerSchema = Joi.object({
+  payload: Joi.object({
+    totalComponents: Joi.number(),
+  }),
+}).required()
+
+module.exports = class bitComponents extends BaseJsonService {
+  async fetch({ owner, scope }) {
+    const url = `https://api.bit.dev/scope/${owner}/${scope}/`
```
calebcartwright: Is there any public documentation available for the API? If so it would be great to include the link in an inline comment (`// API Docs - https://docs.bit.dev/... `) to make it easier for folks to find that info.

services/bit/bit-components.tester.js
```diff
@@ -0,0 +1,15 @@
...
+t.create('collection (valid)')
+  .get('/ramda/ramda.json')
+  .expectBadge({
+    label: 'components',
+    message: Joi.string().regex(/^[0-9]+$/),
+  })
+
+t.create('collection (valid)')
+  .get('/bit/no-collection-test.json')
+  .expectBadge({ label: 'components', message: 'not found' })
```
calebcartwright: The expected message here will also need to be updated to reflect the change made to the 404 error 👍

services/bit/bit-components.tester.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+const Joi = require('joi')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('collection (valid)')
+  .get('/ramda/ramda.json')
+  .expectBadge({
+    label: 'components',
+    message: Joi.string().regex(/^[0-9]+$/),
```
calebcartwright: You'll want to use `isMetric` here from `test-validators.js`, especially since the service is using the `metric` formatter

services/bit/bit-components.service.js
```diff
@@ -0,0 +1,59 @@
...
+      url,
+      schema: ownerSchema,
+      errorMessages: {
+        404: 'collection not found',
+      },
+    })
+  }
+
+  static render({ count }) {
+    return { message: metric(count), color: '#73398D' }
```
calebcartwright: I'm not familiar with the bit.dev service, but I take it there's a reason to always use the `#73398D` color? With other count/download/etc. related metrics, it is common for the badges to use a color scale (for example `downloadCount` from `color-formatters.js`). 

Do you think it would be valuable here to have a color scale based on the count value, or does the static `#73398D` color make more sense?

services/bit/bit-components.service.js
guysaar223: done :)

services/bit/bit-components.service.js
guysaar223: it's should always return object with payload and totalComponents, unless the collection is not exists then it will return 404 with only payload object.
([https://api.bit.dev/scope/ramda/ramda2](https://api.bit.dev/scope/ramda/ramda2))

I have fixed the schema and using

services/bit/bit-components.service.js
guysaar223: there is no official API documentation.

services/bit/bit-components.service.js
guysaar223: you are right, I have changed it to dynamic color using downloadCount function.

services/bit/bit-components.service.js
calebcartwright: > there is no official API documentation

Ah that's disappointing but thanks for confirming. Do you know if any docs are pending? Or is this an unofficial/unsupported API? 

I ask because Shields is maintained by a small number of volunteers, but it supports hundreds of services. As such we like to have as much info available as possible to make it easier for us to maintain/support all those various integrations.

guysaar223: I think the official docs for the API will publish until the end of this year,
but I will need to confirm it, and update you.

calebcartwright:
Everything looks good! Can you envision any potential other `bit` related badges being of value down the road? 

It looks like there are are some component level data points (looks like there's a status, download count, license, etc.) that I could potential see folks wanting badges, and perhaps user/profile info as well (# of collections for a profile, # of contributors, etc.).

If so then we may want to think through the badge route a bit before merging to see if  `/bit/components` is still best route

guysaar223:
> Everything looks good! Can you envision any potential other `bit` related badges being of value down the road?
> 
> It looks like there are are some component level data points (looks like there's a status, download count, license, etc.) that I could potential see folks wanting badges, and perhaps user/profile info as well (# of collections for a profile, # of contributors, etc.).
> 
> If so then we may want to think through the badge route a bit before merging to see if `/bit/components` is still best route

I think we can change the route to  
`/bit/total-components
what do you think ?

services/bit/bit-components.service.js
calebcartwright: Thanks! No need to block on that, but feel free to drop us a note (or a PR 😄) when those docs become public

calebcartwright:
> I think we can change the route to
`/bit/total-components
what do you think ?

Is my understanding correct that the badge being added here is showing the total number of components in a collection? (again apologies as I am not familiar with bit.dev at all)

guysaar223:
> > I think we can change the route to
> > `/bit/total-components
> > what do you think ?
> 
> Is my understanding correct that the badge being added here is showing the total number of components in a collection? (again apologies as I am not familiar with bit.dev at all)

exactly

calebcartwright:
Our badge routes should follow the pattern `/service/noun/parameters/qualifiers` 

In this instance, the service is `bit` and the `noun` would be `collection` IMO, so I would suggest:

`/bit/collection/total-components/...` 

That pattern should hold up well if we need to add badges for other nouns down the road (`component`, `user`, etc.) as well as additional metrics/measures for collections

guysaar223:
> Our badge routes should follow the pattern `/service/noun/parameters/qualifiers`
> 
> In this instance, the service is `bit` and the `noun` would be `collection` IMO, so I would suggest:
> 
> `/bit/collection/total-components/...`
> 
> That pattern should hold up well if we need to add badges for other nouns down the road (`component`, `user`, etc.) as well as additional metrics/measures for collections

agree, I have been changed it to:  
`/bit/collection/total-components/...`

services/bit/bit-components.service.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  static get category() {
+    return 'other'
+  }
+
+  static get route() {
+    return {
+      base: 'bit/collection/total-components',
+      pattern: ':owner/:scope',
```
calebcartwright: One final question, is  `scope` the same as the collection name and is that something that bit.dev users will know? If so then this is good, but if not, would `:collection` be more clear?

guysaar223: scope and collection is the same, but the right name for this now is collection, 
I changed it to `collection`,
 thanks :)

Issue #3243, opened 2019-03-31
vaadin directory badges/tests flaky


chris48s:
:clock11: **When did the problem start?**

We've been seeing inconsistent failures on vaadin directory in the daily tests for at least a week

:camera: **Live badge**

https://img.shields.io/vaadin-directory/v/vaadinvaadin-grid.svg
![](https://img.shields.io/vaadin-directory/v/vaadinvaadin-grid.svg)

:wrench: **Is the live badge working?**

intermittent (same problem is affecting tests and live badges)

:beetle: **Stack trace**

```
     Error: Request timed out after 5000 ms
      at IcedFrisbyNock.run (node_modules/icedfrisby/lib/icedfrisby.js:1084:37)
      at <anonymous>
```

:bulb: **Possible solution**

I've raised an issue on the upstream repo https://github.com/vaadin/vaadin.com/issues/75

chris48s:
I've not seen this happen in the daily test runs lately, so I'm going to close it..

Issue #3329, opened 2019-04-17
Deprecate route `format`, replacing with `pattern`


paulmelnikow:
To make the code easier to understand and to support trie-based routing for #3328, it would be helpful to replace the several remaining instances of route `format`s with `patterns`.

Many of the remaining `format`s are designed to support aliases, and since we've started using redirects heavily, the likely solution is to rewrite the canonical form using a pattern, and add redirects for each of the alias forms.

Ideally by using multiple redirects we can rewrite everything using patterns.

paulmelnikow:
With the PRs for #3714 merged, outside of legacy redirectors there aren't many of these left!

calebcartwright:
One of the last remaining non-redirect ones is Wercker. I've sat down twice now to attempt to convert but it's a bit of an 👀 popper.

https://github.com/badges/shields/blob/ddbe19f6bd2f1b81c37712764a93edd9ac9c4b83/services/wercker/wercker.service.js#L22-L29

Based on the dialog in https://github.com/badges/shields/issues/2076, my sense is that there's two different types of build status we support from Wercker that require two different types of input (one an ID and the other the user/app name combo).

I'm thinking it might be easiest to split these in two, at least the service route definitions, with a base class/reusable functions for all the common bits. 

Does that sound reasonable?

chris48s:
There's an explanation of why splitting it is hard at https://github.com/badges/shields/pull/2279#discussion_r231661787 Once you split them into 2 services, we can't control the order the routes get matched in.

That said, as a related issue, what we're doing on the wercker badge is also a bit strange anyway (see point 3 on https://github.com/badges/shields/pull/2103 ). One of the routes that exists does an unexpected thing for legacy compatibility reasons.

At the time we did that, we had no idea how many users might be relying on the wercker badges, but now we have stats we know that we've served less than 1,000 total requests for the wercker badges in the last month.

Given all that, I'd be happy to:

* re-organise the routes for wercker so we can implement them all with patterns
* make the URLs match the content in all cases
* set up redirects as far as possible
* accept some badges are going to break for a small number of users

paulmelnikow:
Another of these where we're still using a `format` is Travis. We're using a negative lookahead to prevent matching `php-v`, along with an optional domain component. I'd suggest we resolve this one by making the domain `org|com` required, and adding a redirector with the legacy `format`.

PR #3339, opened 2019-04-20
refactor [vaadindirectory] badges

chris48s:
Side note: Not sure if #3243 is totally fixed, but all the tests have passed consistently while I've been working on these badges this evening


services/vaadin-directory/vaadin-directory-rating-count.service.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  static render({ ratingCount }) {
+    return {
+      message: `${metric(ratingCount)} total`,
+      color: floorCountColor(ratingCount, 5, 50, 500),
+    }
+  }
+
+  async handle({ which, packageName }) {
```
calebcartwright: Looks like we aren't using `which` here, so I take it the `rc` and `rating-count` there are for BC?

chris48s: Yes exactly. There are several vaadin badges which have more than one alias for the same badge:

* `rc|rating-count`
* `star|stars`
* `rd|release-date`
* `v|version`

all do the same thing.

Maybe it would be better to call the param `:alias` instead of `:which`?
I'm not massively keen to do it with a redirect class in this case as it requires quite a bit more boilerplate to do it that way.

calebcartwright: My initial thought here at first glance was that eslint would bark at `which` being unused but looks like I was wrong there. 

Given that putting the aliases in the route param doesn't prevent using a `pattern` (as articulated in #3329), I don't have any objections with this over redirectors. 👍

services/vaadin-directory/vaadin-directory-rating.service.js
```diff
@@ -0,0 +1,62 @@
...
+        namedParams: { packageName: 'vaadinvaadin-grid' },
+        staticPreview: this.render({ which: 'rating', score: 4.75 }),
+        keywords: ['vaadin-directory', 'rating'],
+      },
+      {
+        title: 'Vaadin Directory',
+        pattern: 'stars/:packageName',
+        namedParams: { packageName: 'vaadinvaadin-grid' },
+        staticPreview: this.render({ which: 'star', score: 4.75 }),
+        keywords: ['vaadin-directory', 'star', 'stars'],
```
calebcartwright: What would you think about a single example with a `:which(rating|stars)/:packageName`? That would make it easier for a user to toggle between the two formats from within the window without having to go back and forth between the editor and listing page. 

Not a blocker, but wanted to suggest it along the lines of the goals proposed in #3174

services/vaadin-directory/vaadin-directory-version.service.js
```diff
@@ -0,0 +1,42 @@
...
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'vaadin directory' }
+  }
+
+  static render({ version }) {
+    return { message: version, color: '#00b4f0' }
```
calebcartwright: I know the color is consistent with the current/legacy implementation, but out of curiosity do you know why we use a static `#00b4f0` color here instead of the `version` colorFormatter used for other version badges?

calebcartwright:
I'm still a little confused by when/if GH reflects the staging app deployments (currently shows as not deployed), but the apps is here https://shields-staging-pr-3339.herokuapp.com/

services/vaadin-directory/vaadin-directory-version.service.js
chris48s: I just reproduced the existing behaviour, but I don't think there's a clear reason for it. Having just done a bit of reading, Vaadin Directory does seem to at least assume SemVer if not enforce it, so I think its more useful to standardize on `renderVersionBadge()`

chris48s:
cheers for reviewing

PR #3350, opened 2019-04-22
Rewrite [githubcommitssince]; also test [githubrelease]

paulmelnikow:
Ref #2863


services/github/github-commits-since.service.js
```diff
@@ -1,20 +1,14 @@
...
+const { documentation, errorMessagesFor } = require('./github-helpers')
 
-// This legacy service should be rewritten to use e.g. BaseJsonService.
-//
-// Tips for rewriting:
-// https://github.com/badges/shields/blob/master/doc/rewriting-services.md
-//
-// Do not base new services on this code.
-module.exports = class GithubCommitsSince extends LegacyService {
+const schema = Joi.object().required()
```
calebcartwright: Is the schema still in progress or do we need to use this?

paulmelnikow: Oops! Fixed.

PR #3351, opened 2019-04-22
 Rewrite [GithubDownloads]

paulmelnikow:
This one was hairier than I expected!

For consistency with other download badges, I changed some formatting:

-  **downloads | 24k total** -> **downloads | 24k**
- **downloads | 3k** -> **downloads@latest | 3k**
- **downloads | 3k v0.29.0** -> **downloads@v0.29.0 | 3k**

I think the other behavior is the same as before.

Ref #2863


services/github/github-downloads.service.js
```diff
@@ -36,11 +41,10 @@ module.exports = class GithubDownloads extends LegacyService {
           user: 'atom',
           repo: 'atom',
         },
```
calebcartwright: any objections to adding `download` as a keyword? Happy to push a commit myself if there's no objections.

I know it wasn't in there before, but I just noticed it when I went into the staging app and tried to search for `github downloads`

calebcartwright:
I'm getting an `invalid response data` on the shields repo in the staging app whereas prod is showing `0 downloads`

https://shields-staging-pr-3351.herokuapp.com/github/downloads/badges/shields/total.svg


https://img.shields.io/github/downloads/badges/shields/total.svg

services/github/github-downloads.service.js
paulmelnikow: Yes please!

paulmelnikow:
It seems to me showing an error is reasonable, since we don't have downloadable assets. I'm not sure we have releases. It should be something more helpful, though. Do you agree?

calebcartwright:
> It seems to me showing an error is reasonable, since we don't have downloadable assets. I'm not sure we have releases. It should be something more helpful, though. Do you agree?

Agreed. I'm not sure if that's a `no releases found`, `no assets found`, etc. but if it's a detectable scenario it would be great to throw a better error

services/github/github-downloads.service.js
calebcartwright: done

services/github/github-downloads.service.js
```diff
@@ -22,7 +22,7 @@ const releaseArraySchema = Joi.alternatives().try(
   Joi.array().length(0)
 )
 
-const keywords = ['github download']
+const keywords = ['download']
```
calebcartwright: this is what i originally went with, but at least in my local env I was hoping to search with `github download` to narrow the results down all the way

paulmelnikow: Oh, sorry! Did that work? If so I’ll revert it. I thought the query got split on spaces and then matched against all the keywords, so didn’t think spaces in the keywords made any difference.

calebcartwright: Maybe we should have both terms? Right now in the review app I can search on `download` which gives a better result than prod today, but if I search for `github d..` I get no results

paulmelnikow: I feel like searching for "github download" should do the expected thing and we should fix that in a broader way. I've dug in a bit and it seems to be a bit involved, so I'll leave what you did for now.

PR #3353, opened 2019-04-23
Unify order of properties and methods in services

paulmelnikow:
I find having these in a consistent order makes the servicees much faster to read.

This is the order I’ve generally been using:

1. Category
2. Route
3. Examples
4. Rendering
5. Other helpers (`fetch()`, `transform()`)
6. `handle()`

How does this order look to people?

If we decide to keep this order, there are many more services which will need updating.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -17,6 +17,7 @@ overrides:
     parserOptions:
       sourceType: 'module'
   - files:
+      - 'core/base-service/**/*.js'
       - 'services/**/*.js'
     rules:
       sort-class-members/sort-class-members:
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -12,10 +12,33 @@ parserOptions:
   sourceType: 'script'
 
 overrides:
-  files:
-    - gatsby-browser.js
-  parserOptions:
-    sourceType: 'module'
+  - files:
+      - gatsby-browser.js
+    parserOptions:
+      sourceType: 'module'
+  - files:
+      - 'services/**/*.js'
+    rules:
+      sort-class-members/sort-class-members:
+        [
+          'warn',
+          {
+            order:
+              [
+                'name',
+                'category',
+                'isDeprecated',
+                'route',
+                'examples',
+                'defaultBadgeData',
+                'render',
+                'constructor',
+                'fetch',
+                'transform',
+                'handle',
+              ],
+          },
+        ]
 
 plugins:
   - mocha
@@ -48,26 +71,6 @@ rules:
   func-style: ['error', 'declaration', { 'allowArrowFunctions': true }]
   new-cap: ['error', { 'capIsNew': true }]
   import/order: ['error', { 'newlines-between': 'never' }]
-  sort-class-members/sort-class-members:
-    [
-      'warn',
-      {
-        order:
-          [
-            'name',
-            'category',
-            'isDeprecated',
-            'route',
-            'examples',
-            'defaultBadgeData',
-            'render',
-            'constructor',
-            'fetch',
-            'transform',
-            'handle',
-          ],
-      },
-    ]
 
   # Mocha-related.
   mocha/no-exclusive-tests: 'error'
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -20,7 +20,8 @@ overrides:
 plugins:
   - mocha
   - no-extension-in-require
-  - 'chai-friendly'
+  - chai-friendly
+  - sort-class-members
 
 rules:
   # Disable some rules from eslint:recommended.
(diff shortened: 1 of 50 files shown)


chris48s:
As long a telling people "your methods are in the wrong order" is ESLint's job and not ours, I don't have a strong preference on what the actual order is :)

calebcartwright:
I've thought about this a few times before myself, but ultimately decided there was just much variation of ordering across our different services to know whether there was a "right" ordering. 

I agree with @chris48s on both points: I don't personally have a strong preference on what the ordering is, but if we go down the path of establishing and then enforcing an order then that enforcement has to be automated.

paulmelnikow:
Yea, I'm happy to have these checked by the linter. Pretty much agreed it's awful business to be correcting people in PRs on this stuff. This change isn't autofixable, but I don't mind rearranging them once, and think uniformity will be worth the minor hassle of arranging them correctly going forward. Shall we merge this and I'll follow up with more fixes?

I've got the linter rule on warning to help with the manual fixing, though unfortunately it spews lots of messages, so we'll want to get these all fixed ASAP.

calebcartwright:
SGTM. Will give it a quick look over and approve later today (unless someone else beats me to it)

.eslintrc.yml
```diff
@@ -47,6 +48,26 @@ rules:
...
+            'category',
+            'isDeprecated',
+            'route',
+            'examples',
+            'defaultBadgeData',
+            'render',
+            'constructor',
+            'fetch',
+            'transform',
+            'handle',
```
chris48s: I'm not sure about applying this order globally. It definitely makes sense to apply to `services/**/*.js` and the base classes, but it feels like applying this in dirs like `/lib` for example that don't contain service classes is a bit odd. For example I notice that running `lint` locally is throwing some warnings on files in `frontend` but its a coincidence that we're using the same method names.

paulmelnikow: Yea, agreed that should be adjusted. Scoping it to `services` seems like the right thing to do.

Mostly we've dealt with that using separate eslintrcs, though I think there's also way to keep the config centralized and apply rules only to certain files.

chris48s: Yes you can do it all in one config file with glob patterns:

https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns

calebcartwright:
I convinced myself it would be easier to run the service tests than it would be to review each line change, but now I'm not so sure.

Any idea why the StackExchange tests are failing here? It looks like the labels are off

https://circleci.com/gh/badges/shields/51214
```
StackExchangeMonthlyQuestions Monthly Questions for StackOverflow Momentjs [ GET /stackoverflow/qm/momentjs.json ] - [ GET /stackoverflow/qm/momentjs.json ]
/home/circleci/project/core/service-test-runner/cli.js
AssertionError: label mismatch: expected 'stackoverflow' to equal 'stackoverflow momentjs questions'
    at Function._expectField (core/service-test-runner/icedfrisby-shields.js:47:51)
    at IcedFrisbyNock.afterJSON.json (core/service-test-runner/icedfrisby-shields.js:37:26)
    at IcedFrisbyNock.<anonymous> (node_modules/icedfrisby/lib/icedfrisby.js:924:10)
    at invokeNextHook (node_modules/icedfrisby/lib/icedfrisby.js:973:24)
    at Promise (node_modules/icedfrisby/lib/icedfrisby.js:987:7)
    at new Promise (<anonymous>)
    at IcedFrisbyNock._runHooks (node_modules/icedfrisby/lib/icedfrisby.js:946:12)
    at IcedFrisbyNock.run (node_modules/icedfrisby/lib/icedfrisby.js:1238:20)
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:229:7)
```a

**Comment by paulmelnikow** [comment]
Hmm, I would guess it's a transient failure. That test seems to be passing locally.

**Comment by calebcartwright** [comment]
Yeah it's really weird. It's passing for me locally too (and I see absolutely nothing in this PR that would cause the tests to fail) so I'm just going to remove the service tests from the title and re-run circle

File: services/wercker/wercker.service.js [hunk 2]
```diff
@@ -15,53 +15,6 @@ const werckerSchema = Joi.array()
...
-      })
-    }
-    const { result } = json[0]
-    return this.constructor.render({ result })
-  }
-
-  // Metadata
   static get category() {
     return 'build'
   }
```
calebcartwright: Side note on the route below... that is one helluva `format` 😄 Should be a whole lot of fun converting that to `pattern`

paulmelnikow: Indeed! Thank goodness for redirects.

PR #3355, opened 2019-04-23
Make the config from env more obvious at first glance; explain cascading config

paulmelnikow:


calebcartwright:
Oops! Will re-approve once prettier is satisfied 😄

PR #3358, opened 2019-04-24
Unify order of more services

paulmelnikow:
Ref #3353


paulmelnikow:
Thanks for reviewing these!

PR #3360, opened 2019-04-24
Rewrite [GithubTag]

paulmelnikow:
Ref #2863

### DIFF
--- a/services/github/github-tag.spec.js
+++ b/services/github/github-tag.spec.js
@@ -1,6 +1,8 @@
 'use strict'
 
 const { test, given } = require('sazerac')
+const { addv } = require('../text-formatters')
+const { version: versionColor } = require('../color-formatters')
 const GithubTag = require('./github-tag.service')
 
 describe('GithubTag', function() {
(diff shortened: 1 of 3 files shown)


calebcartwright:
I added a unit test for `render` and have approved 👍 (will defer to someone else to merge since I've added a commit)

services/github/github-tag.spec.js
```diff
@@ -0,0 +1,37 @@
...
+      'v1.3-beta3'
+    )
+    given({ json: tagFixture, usingSemver: false, includePre: false }).expect(
+      'cheese'
+    )
+  })
+
+  test(GithubTag.render, () => {
+    given({ usingSemver: false, version: '1.2.3' }).expect({
+      message: addv('1.2.3'),
```
paulmelnikow: Any reason not to replace these with literals?

calebcartwright: I've no issues with replacing them. Rationale very much related to https://github.com/badges/shields/pull/3266#discussion_r278374394, the latter paragraph in particular

calebcartwright: `message`'s inlined, still calling the `versionColor` helper below but happy to change that to `blue` if you prefer (and re 👍 the PR)

paulmelnikow: I'm happy to continue the discussion at https://github.com/badges/shields/pull/3266#discussion_r278374394 though that's a bit more long-ranging; we can revisit this. For now I'd prefer we inline the color as well.

PR #3362, opened 2019-04-25
Fix [LGTM] badges for non-GitHub projects.

chrisgavin:
LGTM supports multiple repository hosts, including Bitbucket and GitLab. Currently it's only possible to generate badges for GitHub projects, but by adding a `host` parameter, badges for other projects should also work.

To give an example, [this badge](https://img.shields.io/lgtm/alerts/b/atlassian/confluence-business-blueprints.svg?logo=lgtm&logoWidth=18) does not work, even though [the project exists on LGTM](https://lgtm.com/projects/b/atlassian/confluence-business-blueprints/).


services/lgtm/lgtm-alerts.service.js
```diff
@@ -6,8 +6,8 @@ const LgtmBaseService = require('./lgtm-base')
 module.exports = class LgtmAlerts extends LgtmBaseService {
   static get route() {
     return {
-      base: 'lgtm/alerts/g',
-      pattern: ':user/:repo',
+      base: 'lgtm/alerts',
+      pattern: ':host/:user/:repo',
```
calebcartwright: Do you know the set of allowed hosts that LGTM supports? I'm guessing GitHub, Bitbucket, and GitLab which are expressed via `g`, `b`, and probably `gl`, respectively. 

It would be better to scope the `:host` param accordingly (i.e. `:host(g|b|bl)/...`) so that it will display properly in the UI window. Personally, I'd prefer that we actually use `:host(github|bitbucket|gitlab)/...` in the route to present the full name to end users, and internally map that to the corresponding letter/key we need for the API

chrisgavin: Yes, I believe `g`, `b` and `gl` are the supported VCS hosts. Adding an internal mapping and a fixed list of hosts to display in the UI sounds good. I think for the actual routing it would make sense to fall back to just using whatever the input was to look up the project if it is not in the map. This would have the benefit of keeping existing badges working and it would allow for people to generate URLs manually if more hosts are added in the future, even if the map is not kept up to date.

It doesn't look like this would be too difficult to do, I'll try and update this pull request later today with if there are no objections to this implementation.

calebcartwright: Sounds like a plan! Regarding some of the specifics on implementation:

> I think for the actual routing it would make sense to fall back to just using whatever the input was to look up the project if it is not in the map

Currently the only existing route that is in use is `/lgtm/(alerts or grade)/g/...`, so if we update the routes to `/lgtm/alerts/:host(github|bitbucket|gitlab/...` then all we need to maintain backwards compatibility is to add a redirector to map `g` to `github`. 

We've got a special service class type that handles this redirect functionality that can be created via the `redirector` helper function. Here's an idea of what that might look like here for LGTM:

```js
'use strict'

const commonAttrs = {
  category: 'analysis',
  dateAdded: new Date('2019-04-26'),
}

module.exports = [
  redirector({
    route: {
      base: 'lgtm/alerts/g',
      pattern: ':user/:repo',
    },
    transformPath: ({ user, repo }) => `/lgtm/alerts/github/${user}/${repo}`,
    ...commonAttrs,
  }),
  redirector({
    route: {
      base: 'lgtm/grade/g',
      pattern: ':user/:repo',
    },
    transformPath: ({ user, repo }) => `/lgtm/grade/github/${user}/${repo}`,
    ...commonAttrs,
  }),
]
```

services/lgtm/lgtm-base.js
```diff
@@ -25,8 +31,17 @@ module.exports = class LgtmBaseService extends BaseJsonService {
     return { label: 'lgtm' }
   }
 
-  async fetch({ user, repo }) {
-    const url = `https://lgtm.com/api/v0.1/project/g/${user}/${repo}/details`
+  static get pattern() {
+    return ':host/:user/:repo'
```
calebcartwright: Thanks for the updates! We should still make this route pattern more restrictive to only allow the valid hosts we will accept. The only option we'll receive from existing badges is `g` which we can handle via our `redirector` functionality (details [here](https://github.com/badges/shields/pull/3362#discussion_r279043952) for reference).

services/lgtm/lgtm-base.js
```diff
@@ -25,8 +31,17 @@ module.exports = class LgtmBaseService extends BaseJsonService {
...
+  }
+
+  static get examplePattern() {
```
calebcartwright: Once the pattern is made more explicit as noted above, this (and usage of it) should be deleted as the examples will be able to use the default route pattern defined for the services

chrisgavin:
Sorry for the delay. I've managed to add a redirector, so hopefully this now works as you would like it to. :smile:

calebcartwright:
It looks good! The last piece is to add tests for those redirectors (side note, I've opened #3402 to document this process). I'd suggest something like the below snippet:

lgtm-redirect.tester.js
```js
'use strict'

const { ServiceTester } = require('../tester')

const t = (module.exports = new ServiceTester({
  id: 'LgtmRedirect',
  title: 'LgtmRedirect',
  pathPrefix: '/lgtm',
}))

t.create('alerts')
  .get('/alerts/g/badges/shields.svg', {
    followRedirect: false,
  })
  .expectStatus(301)
  .expectHeader(
    'Location',
    '/lgtm/alerts/github/badges/shields.svg'
  )

t.create('grade')
  .get('/grade/java/g/apache/cloudstack.svg', {
    followRedirect: false,
  })
  .expectStatus(301)
  .expectHeader(
    'Location',
    '/lgtm/grade/java/github/apache/cloudstack.svg'
  )
```

services/lgtm/lgtm-base.js
```diff
@@ -25,8 +31,13 @@ module.exports = class LgtmBaseService extends BaseJsonService {
...
   }
 
-  async fetch({ user, repo }) {
-    const url = `https://lgtm.com/api/v0.1/project/g/${user}/${repo}/details`
+  static get pattern() {
+    return `:host(${Object.keys(hostMappings).join('|')})/:user/:repo`
+  }
+
+  async fetch({ host, user, repo }) {
+    const mappedHost = host in hostMappings ? hostMappings[host] : host
```
calebcartwright: This conditional is no longer necessary given the pattern restriction (will always be one of the keys in `hostMappings`

Could either do
```suggestion
    const mappedHost = hostMappings[host]
```

Or delete that line and inline it in the `url` template  string

calebcartwright:
Thanks for this!

chrisgavin:
No problem. Thanks for reviewing! :smile:

PR #3299, opened 2019-04-13
Rewrite [GitHubForks]

paulmelnikow:
Fix #3053

PR #3366, opened 2019-04-27
Color fix for [GitHubForks] badge

PyvesB:
Small detail that was left out in #3299. The non social version of the badge is now grey rather than [the previous blue](https://github.com/badges/shields/blob/11cd0bdbc59d7d244aa61112102224aabdb7ea5b/services/github/github-forks.service.js#L80).


paulmelnikow:
Thanks for fixing this!

Issue #2674, opened 2019-01-07
SVG by default


paulmelnikow:
It'd be nice to simplify our badge URLs by dropping `.svg` from the end of the URLs. While it's common to include a `.svg` extension on a badge URL, there are a bunch of native status badges which don't, and it's hardly a requirement.

This would be easy to do in the modern badges because it's only handled in one place. For the legacy badges… well, maybe it'll motivate us to get the refactor done! 😁

We could support extensionless SVG in the modern badges as an undocumented feature, and hold off on advertising to the world and it in the UI until everything is fixed. If we're going to create a new static badge URL (#2673) it might make sense to advertise it for that.

I _think_ our Cloudflare rules apply regardless of file extension, though we'd need to verify that as well.

conao3:
I am not in a position to give an opinion, but I will write information for reference.

codacy badge has no `.svg` suffix.
When writing Readme in .org document, it is necessary to execute workaround to suffix `?.svg` due to GitHub parser problem. (Perhaps should fix GitHub's renderer, but I do not have that skill...)

I think that it is a good thing to be able to omit the `.svg` suffix in general situations, but I think it would be nice if you do not lose the current specification for the .org user.

This codacy badge issue has already reported.

Rendered: https://github.com/conao3/playground/tree/codacy-support
Raw .org document: https://raw.githubusercontent.com/conao3/playground/codacy-support/Readme.org

paulmelnikow:
It's useful to know there are some unusual cases where the extension is helpful. This is about [org mode](https://orgmode.org/), correct?

conao3:
Yes! `.org` document is plain text along the format of org mode.

paulmelnikow:
Since #2863 is done, we can do this now!

PR #3372, opened 2019-04-27
Add [Fedora] package version support

scop:
https://apps.fedoraproject.org/mdapi/


services/fedora/fedora.service.js
```diff
@@ -0,0 +1,49 @@
...
+
+module.exports = class Fedora extends BaseJsonService {
+  static get category() {
+    return 'version'
+  }
+
+  static get route() {
+    return {
+      base: 'fedora/v',
+      pattern: ':branch/:packageName',
```
calebcartwright: Is it possible for `branch` to contain a `/` character (like a `feature/abc` for branch)?

scop: As far as I know, no. The branch name occurs in various filenames in their distros, and that wouldn't work with a slash in it. I find it highly unlikely that they'd ever go through all the work that would make it possible. Their own API uses the branch as a path component as well, see https://apps.fedoraproject.org/mdapi/

scop:
Revised to make branch optional, defaulting to rawhide (== development). No way to permalink to current stable at the moment, see comments in code.

scop:
Note that due to the epoch version validator, requires #3397 to be in first.

services/fedora/fedora.tester.js
```diff
@@ -0,0 +1,30 @@
...
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Fedora package (default branch, valid)')
+  .get('/rpm.json')
+  .expectBadge({
+    label: 'fedora',
+    message: isVPlusDottedVersionNClausesWithOptionalSuffixAndEpoch,
+  })
+
+t.create('Fedora package (valid, mocked response)')
```
calebcartwright: Is this mocked test necessary? It seems like it's exactly the same as the above live test just with a mock response

scop: I initially copied the homebrew service tester as basis of this and it had a similar live/mocked same test pair in it. It's also now in the [Debian] and [ArchLinux] testers.

Maybe it in some cases is a good thing in the sense that if the API ever changes, we have some kind of literal example what the expectation was. For [Fedora] that's very well already covered in the schema though, but for example in [Debian] the schema consists of more or less arbitrary patterns, and a written out example helps some.

No strong opinions here, but I think I'd just leave it in for above reason and consistency.

calebcartwright: You're right that this same thing is done in the tester for `homebrew` but it's still uncommon for us to have have an identical live/mocked test pair. I think the [similar mocked test for Debian](https://github.com/badges/shields/blob/master/services/debian/debian.tester.js#L22) is a bit different (it specifies a non-default value for the distribution), but agree that [ArchLinux](https://github.com/badges/shields/blob/master/services/archlinux/archlinux.tester.js#L15) has an identical live/mock test pair

I'm still inclined to say we should remove this, but @paulmelnikow @chris48s since you looked at some of those other PRs, what's your take?

paulmelnikow: > Maybe it in some cases is a good thing in the sense that if the API ever changes, we have some kind of literal example what the expectation was. For [Fedora] that's very well already covered in the schema though, but for example in [Debian] the schema consists of more or less arbitrary patterns, and a written out example helps some.

Yea, indeed, it's helpful to document what the expectation was. And yea, this is a big part of what the schema is for. I'm not totally sure what you mean by arbitrary patterns. Do you mean the form of the version number?

How about removing the test, and adding a comment alongside the schema, like `// e.g. '4.14.2.1'`?

services/fedora/fedora.service.js
paulmelnikow: In almost all of our badges, we put the branch after the package name. If it's optional, I'm inclined to do the same thing here. We usually try to stick with consistency of our own conventions over matching the conventions of other projects and services. Every service does things a little differently, and it's helpful to give users a consistent experience.

services/fedora/fedora.tester.js
scop: The arbitrary `pattern`s are in [Debian], not here.

I don't think the `version: Joi.string().required()` line here deserves that kind of a comment, but sure, if you think the mock test is more harmful than good, let me know and I'll get rid of it.

services/fedora/fedora.service.js
scop: At least [Debian] and [Ubuntu] already have the optional branch before the package name, and [ArchLinux] has that kind of stuff there even though it's not optional.

In my opinion it's good this way, because then the svg filename is always `package.svg`, not e.g. `branch.svg`. And it's consistent between services no matter if the stuff before it is optional or not.

services/fedora/fedora.tester.js
paulmelnikow: > The arbitrary `pattern`s are in [Debian], not here.

Ah, gotcha.

Yea, I think the test can be removed.

services/fedora/fedora.service.js
paulmelnikow: > In my opinion it's good this way, because then the svg filename is always `package.svg`, not e.g. `branch.svg`. And it's consistent between services no matter if the stuff before it is optional or not.

To be honest I agree with you. It does seem nicer that way.

However these Gem downloads routes, for example, have been established for years:

Screen Shot 2019-05-01 at 12 45 38 PM

All the Code Coverage badges put branch after the repo: https://shields.io/category/coverage

We've done a lot of work with redirects lately to update our routes, though I don't I see being able to do that with these since we'd need a way to detect whether it's branch-first or branch-second.

Since Debian just came online and Ubuntu hasn't yet, and they aren't being used by too many people yet, it might be a good time to change them.

Also, a smaller thing, we're planning to drop the `.svg` or at least make it optional (#2674) which should clean up these routes, too!

services/fedora/fedora.tester.js
scop: Done.

services/fedora/fedora.service.js
scop: Fair enough. But how about this: I'll leave it as is here, then let's merge. And if someone wants to change the order for Debian and Ubuntu, they can do it for this too at the same time -- I kind of would not want to spend time that because I still think it would be better the other way around :)

scop: Nevermind, moved, see also #3409

paulmelnikow: Thank you!

paulmelnikow:
@calebcartwright Is this good to merge?

calebcartwright:
@paulmelnikow - I'm good with it 🚀

paulmelnikow:
Thank you both!

PyvesB:
@scop you've done some amazing work on all these Linux package version badges, thanks! I would love to spread the word about these new badges and thank you publicly on Shields's Twitter account, do you have an @ handle I could mention there? 😄

scop:
On Sun, May 5, 2019 at 4:49 PM Pierre-Yves B. <notifications@github.com>
wrote:

> I would love to spread the word about these new badges and thank you
> publicly on Shields's Twitter account, do you have an @ handle I could
> mention there? 😄
>
Sorry about the delay. But no, I don't have a meaningful handle for that.
If you want to point to some direction, I suggest linking to my GitHub
account some way you feel is appropriate.

PyvesB:
Here's the [Tweet](https://twitter.com/Shields_io/status/1127594478594920448) I came up with. I linked to your GitHub profile as suggested. 👍

PR #3373, opened 2019-04-28
Add [Debian] package version support

scop:
https://wiki.debian.org/Services/api.ftp-master.debian.org

### DIFF
--- a/services/debian/debian.service.js
+++ b/services/debian/debian.service.js
@@ -1,0 +1,63 @@
+'use strict'
+
+const Joi = require('joi')
+const { latest, renderVersionBadge } = require('../version')
+const { BaseJsonService, NotFound } = require('..')
+
+const schema = Joi.array()
+  .items(
+    Joi.object().pattern(
+      /./,
+      Joi.object()
+        .pattern(/./, Joi.object().pattern(/./, Joi.object()))
+        .required()
+    ) // Optional, missing means not found
+  )
+  .required()
+
+const defaultDistribution = 'stable'
+
+module.exports = class Debian extends BaseJsonService {
+  static get category() {
+    return 'version'
+  }
+
+  static get route() {
+    return {
+      base: 'debian/v',
+      pattern: ':distribution?/:packageName',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Debian package',
+        namedParams: { distribution: 'unstable', packageName: 'apt' },
+        staticPreview: renderVersionBadge({ version: '1.8.0' }),
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'debian' }
+  }
+
+  async handle({ distribution = defaultDistribution, packageName }) {
+    const data = await this._requestJson({
+      schema,
+      url: `https://api.ftp-master.debian.org/madison?f=json&s=${distribution}&package=${packageName}`,
+    })
+    if (!data.length) {
+      throw new NotFound()
+    }
+    // Distribution can change compared to request, for example requesting
+    // "stretch" may map to "stable" in results, so can't use value from
+    // request as is. Just take the first one instead.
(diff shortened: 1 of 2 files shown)


services/debian/debian.service.js
```diff
@@ -0,0 +1,63 @@
...
+      throw new NotFound()
+    }
+    // Distribution can change compared to request, for example requesting
+    // "stretch" may map to "stable" in results, so can't use value from
+    // request as is. Just take the first one instead.
+    for (const resultdist of Object.keys(data[0][packageName])) {
+      const versions = Object.keys(data[0][packageName][resultdist])
+      return renderVersionBadge({ version: latest(versions) })
+    }
+    throw new NotFound()
```
chris48s: Is this line reachable?

services/debian/debian.service.js
```diff
@@ -0,0 +1,63 @@
...
+      schema,
+      url: `https://api.ftp-master.debian.org/madison?f=json&s=${distribution}&package=${packageName}`,
+    })
+    if (!data.length) {
+      throw new NotFound()
+    }
+    // Distribution can change compared to request, for example requesting
+    // "stretch" may map to "stable" in results, so can't use value from
+    // request as is. Just take the first one instead.
+    for (const resultdist of Object.keys(data[0][packageName])) {
```
chris48s: I'm not familiar with this API. Is there any valid case where `data.length` can be >1?

services/debian/debian.tester.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
+
+const { isVPlusDottedVersionNClauses } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Debian package (default distribution, valid)')
+  .get('/apt.json')
+  .expectBadge({ label: 'debian', message: isVPlusDottedVersionNClauses })
```
chris48s: Its reasonably common for Debian packages to have some kind of suffix on the version e.g: https://api.ftp-master.debian.org/madison?f=json&s=testing&package=python-dev Is `apt` always going to be a version without a suffix? If not, can we go with `isVPlusDottedVersionNClausesWithOptionalSuffix` to make the test less brittle?

services/debian/debian.service.js
scop: It is, if a package is returned without any versions. I suppose that shouldn't happen, but if it does, "not found" seems appropriate behavior to me.

services/debian/debian.service.js
scop: Probably not. Would you like a case where it is handled somehow differently?

services/debian/debian.tester.js
scop: I don't think there are any such guarantees, so done. See also #3378

scop:
A couple of improvements I thought of but didn't get around to implementing as I didn't immediately spot what would be the best way to do them:
- Maybe mark versions containing `~` as prereleases -- "maybe" because that's what they usually are, but not strictly always; can be backports and such. 
- Implementing epoch aware ordering, possibly also stripping epochs from displayed versions; ref https://www.debian.org/doc/debian-policy/ch-controlfields.html#version, example https://packages.debian.org/stretch/bash-completion

services/debian/debian.service.js
chris48s: If we're getting back arrays which should only contain zero or one elements (and we don't know what do do if we get more), can we validate that in the schema.

I can't remember off the top of my head, but I think chaining `.single()` without `required()` will pass zero or one elements. If not, chaining `.min(0).max(1)` will do the job.

scop: I think `.single()` is the other way around (get single item as if it was an array). `.max(1).required()` seems to do the right thing.

paulmelnikow:
This is a really nice addition! Thank you!

PR #3374, opened 2019-04-28
Add [ArchLinux] package version support

scop:


services/archlinux/archlinux.tester.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+const { isVPlusDottedVersionNClauses } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Arch Linux package (valid)')
+  .get('/core/x86_64/pacman.json')
+  .expectBadge({ label: 'arch linux', message: isVPlusDottedVersionNClauses })
```
chris48s: Similar question here as on the Debian PR. Is a `pacman` version always going to match this regex, or should we be a little more lenient?

scop: Same as in Debian, done here too.

Issue #3242, opened 2019-03-30
allow expectBadge helper to assert links[] content


chris48s:
:clipboard: **Description**

If we could call

```
.expectBadge({ label: 'foo', message: 'bar', links: ['https://foo.bar/baz'] })
```

that would allow us to write more complete tests.

refs https://github.com/badges/shields/pull/3241#discussion_r270643671

PyvesB:
My current thinking is to break down the expectations inside `expectBadge` into separate assertions as part of #2686 (this was also suggested in #2687). Tackling this issue would become very easy, we would add the `link` to the JSON response and add a new assertion inside `expectBadge` which would only be checked if `link` is defined in the input of `expectBadge`. I was planning on putting all the assertions inside an `afterJSON` block within `expectBadge`, however, [this bug](https://github.com/IcedFrisby/IcedFrisby/issues/206) in IcedFrisby is currently a blocker.

Issue #3370, opened 2019-04-27
Bring serviceData and makeBadge schemas into full alignment


paulmelnikow:
Since #2863 is done, the schema translation being performed in `coalesceBadge()` is no longer useful. If we bring the `makeBadge()` schema into full alignment with `serviceData`, we won't need to translate schema in `coalesceBadge()`.

The function would still exist, but it would only coalesce, not _also_ translate schema.

This is dependent on resolving named icons within `makeBadge()` and making simple-icons an optional dependency.

This work is related to #3369, though separable.

PR #3376, opened 2019-04-28
Added link test expectations, affects [Dependabot GitHub Twitter]

PyvesB:
See #3242.

I was not sure whether to use the word link as singular or plural. Even though it makes slightly less sense to me, I went for `link` in the end, as this is the form used in [base.js](https://github.com/badges/shields/blob/2806eb8a00bf2b40db5ce3e6910a0b58a1a6fc8e/core/base-service/base.js#L52).


gh-badges/lib/make-badge.spec.js
```diff
@@ -94,7 +94,11 @@ describe('The badge generator', function() {
 
   describe('JSON', function() {
     it('should always produce the same JSON (unless we have changed something!)', function() {
-      const json = makeBadge({ text: ['cactus', 'grown'], format: 'json' })
+      const json = makeBadge({
+        text: ['cactus', 'grown'],
+        format: 'json',
+        links: ['https://example.com/', 'https://other.example.com/'],
```
chris48s: should this be `link:`, or is this one different on purpose?

gh-badges/templates/default-template.json
```diff
@@ -5,5 +5,6 @@
   "message": {{=JSON.stringify(it.text[1])}},{{?it.logoWidth}}
   "logoWidth": {{=JSON.stringify(it.logoWidth)}},{{?}}{{?it.labelColor}}
   "labelColor": {{=JSON.stringify(it.labelColor)}},{{?}}
-  "color": {{=JSON.stringify(it.color || null)}}
+  "color": {{=JSON.stringify(it.color || null)}}{{?(it.links[0] && it.links[0].length || it.links[1] && it.links[1].length)}},
```
chris48s: Should the comma on this line be inbetween the 2 template expressions, like it is on the previous lines?

gh-badges/lib/make-badge.spec.js
PyvesB: This one is different on purpose, as the plural form is used in [make-badge.js](https://github.com/badges/shields/blob/75def85eca5f045b147191e8142f3ce146051c4f/gh-badges/lib/make-badge.js#L107).

gh-badges/templates/default-template.json
PyvesB: `color` may be the final element in the JSON payload if there isn't a `link` element, therefore the comma has to be part of the last template expression. 😉

gh-badges/lib/make-badge.spec.js
paulmelnikow: We can fix that as part of #3370 😀

PR #3378, opened 2019-04-28
Allow tilde as optional suffix separator

scop:
For example [Debian] uses them, typically for pre-releases.


scop:
For example https://api.ftp-master.debian.org/madison?f=json&s=testing&package=wireless-tools currently has `30~pre9-13`.

paulmelnikow:
Will this help keep the tests green for #3373? If so, 👍 from me 😀

Issue #3365, opened 2019-04-26
`npm run test:entrypoint` fails


platan:
An interesting part of a log of `npm run test:entrypoint`:
```
Configuration:
{ bind: { address: '--no-config', port: 1111 },
...
[1] "address" must be a valid ip address with a optional CIDR
[2] "address" must be a valid hostname
      at Object.exports.process (node_modules/joi/lib/errors.js:203:19)
      at internals.Object._validateWithOptions (node_modules/joi/lib/types/any/index.js:764:31)
      at module.exports.internals.Any.root.validate (node_modules/joi/lib/index.js:147:23)
      at module.exports.internals.Any.root.attempt (node_modules/joi/lib/index.js:177:29)
      at new Server (core/server/server.js:116:30)
```
<details>
<summary>full log</summary>

```
> shields.io@0.0.0 test:entrypoint /home/m/workspaces/github/shields-platan
> cross-env NODE_CONFIG_ENV=test mocha entrypoint.spec.js



raven@2.6.4 alert: no DSN provided, error reporting disabled
Configuration:
{ bind: { address: '--no-config', port: 1111 },
  metrics: { prometheus: { enabled: false } },
  ssl: { isSecure: false },
  cors: { allowedOrigin: [] },
  persistence: { dir: './private' },
  services: 
   { github: 
      { baseUri: 'https://api.github.com/',
        debug: { enabled: false, intervalSeconds: 200 } },
     trace: false },
  profiling: { makeBadge: false },
  cacheHeaders: { defaultCacheLengthSeconds: 120 },
  rateLimit: false,
  handleInternalErrors: false,
  fetchLimit: '10MB',
  redirectUrl: 'http://badge-server.example.com' }
  1) "before all" hook for "should render a badge"
  2) "after all" hook: shut down the server for "should render a badge"

  0 passing (1s)
  2 failing

  1) "before all" hook for "should render a badge":
     ValidationError: {
  "bind": {
    "port": 1111,
    "address" [1, 2]: "--no-config"
  },
  "metrics": {
    "prometheus": {
      "enabled": false
    }
  },
  "ssl": {
    "isSecure": false
  },
  "cors": {
    "allowedOrigin": []
  },
  "persistence": {
    "dir": "./private"
  },
  "services": {
    "github": {
      "baseUri": "https://api.github.com/",
      "debug": {
        "enabled": false,
        "intervalSeconds": 200
      }
    },
    "trace": false
  },
  "profiling": {
    "makeBadge": false
[...]

PR #3390, opened 2019-04-29
Run entrypoint tests in CI

platan:
This PR:
- fixes `entrypoint.spec.js` by removing args from mocha
- adds `test:entrypoint` to CI

Issue https://github.com/badges/shields/issues/3365


platan:
This is a simple and working solution. @paulmelnikow suggested an alternative approach - running server in a subprocess - this would be more complicated, because is requires waiting for server to start. https://github.com/jeffbski/wait-on can be used in this approach.

PR #3392, opened 2019-04-29
Remove legacy helper makeBadgeData

### LINKED ISSUE #3369: Remove legacy badge-data functions
There are three places in production code where we're still using `makeBadgeData`:

```
core/base-service/legacy-request-handler.js:const { makeBadgeData: getBadgeData } = require('../../lib/badge-data')
core/base-service/legacy-request-handler.js:// are used by makeBadgeData (see `lib/badge-data.js`) and related functions.
core/server/server.js:const { makeBadgeData } = require('../../lib/badge-data')
```

And a couple more in tests:

```
core/base-service/base-svg-scraping.spec.js:const { makeBadgeData } = require('../../lib/badge-data')
core/base-service/legacy-request-handler.spec.js:const { makeBadgeData: getBadgeData } = require('../../lib/badge-data')
```

It would be great to rewrite these. It could be done as part of #3328, though it might be possible to think ahead to that and replace the `badgeData`-related code with calls to `coalesceBadge`.

paulmelnikow:
Close #3369


calebcartwright:
Ha! Nice branch name

paulmelnikow:
None of the services use it 😁

Issue #2428, opened 2018-11-29
Move badge templates from doT → Template literals


RedSparr0w:
:clipboard: **Description**
https://github.com/badges/shields/pull/2394#issuecomment-442651526
Rewrite the badge templates to use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

:dart: **Objective**
- Improve readability
- Improve performance
- Merge some parts of the templates to reduce duplication

:ballot_box_with_check: **TODO**
- [ ] Write basic template
- [ ] Test performance vs doT
- [ ] Optimize

paulmelnikow:
Thanks for opening this.

I'm marking this "good first issue." This is not the most straightforward task, though for someone who is comfortable with JavaScript and unit testing, they probably would be able to accomplish this without a lot of knowledge of Shields. It'd be a good introduction to the project, and also very high value. And, of course it's a great thing for an existing contributor or maintainer to take on, too. 😄

RedSparr0w:
Made a quick example with the template strings here:
https://codepen.io/redsparr0w/full/pQGvwz/

RedSparr0w:
I've made an example here using the converted templates:
[Shields.io Badge Generator](https://codepen.io/redsparr0w/full/QJYbmd)

Convert the doT template files to template strings: 
```javascript
template.replace('<svg', 'return `<svg')
  .replace(/\{\{=(.+?)\}\}/g, "${$1 ? `$2` : ''}")
  .replace(/\{\{\?(.+?)\}\}(.+?)\{\{\?\}\}/g, "${$1}")
  .replace(/\{\{\?(.+?)\}\}\r?\n(\s+)/g, "${$1 ?\n$2`")
  .replace(/\r?\n(\s+)\{\{\?\}\}/g, "`\n$1: ''}")
  .replace('</svg>', '</svg>`;');
```

Not sure how to go about it in shields though, will look into it tomorrow _(hopefully)_

RedSparr0w:
I've updated the CodePen using a similar approach, Not sure how to tidy it up more though though:
https://codepen.io/redsparr0w/full/PxgPdG/

paulmelnikow:
Looks like progress!

[...]

PR #3395, opened 2019-04-29
Generate JSON badges without using a template [GithubSearch]

paulmelnikow:
Started this as a warm-up for #2428, so it would be great to get it merged quickly if possible.


services/dependabot/dependabot.tester.js
```diff
@@ -9,7 +9,7 @@ t.create('semver stability (valid)')
     label: 'semver stability',
     message: isIntegerPercentage,
     link: [
-      'https://dependabot.com/compatibility-score.html?package-manager=bundler&amp;dependency-name=puma&amp;version-scheme=semver',
+      'https://dependabot.com/compatibility-score.html?package-manager=bundler&dependency-name=puma&version-scheme=semver',
```
paulmelnikow: I'd argue the change I made fixed an encoding bug. There is not a good reason for these ampersands to be HTML-encoded in the JSON response.

gh-badges/lib/make-badge.js
```diff
@@ -21,58 +21,56 @@ templateFiles.forEach(async filename => {
   const extension = path.extname(filename).slice(1)
   const style = filename.slice(0, -`-template.${extension}`.length)
   // Compile the template. Necessary to always have a working template.
-  templates[`${style}-${extension}`] = dot.template(templateData)
-  if (extension === 'svg') {
```
paulmelnikow: I changed the key in the array from ``${style}-${extension}`` to `style` since the extensions are now all svg. I also removed the if statement.

paulmelnikow:
Remaining failures seem to be unrelated.

calebcartwright:
> Remaining failures seem to be unrelated.

The tests failing on label mismatches on the GitHub badges has me a bit puzzled 🤔

gh-badges/lib/make-badge.js
```diff
@@ -164,18 +173,13 @@ function makeBadge({
     logoPosition,
     logoWidth,
     logoPadding,
-    // `color` and `labelColor` are included for the `default` JSON template.
-    color: normalizeColor(color),
-    labelColor: normalizeColor(labelColor),
     colorA: toSvgColor(labelColor),
     colorB: toSvgColor(color),
```
calebcartwright: `color` and `labelColor` are both now set to the return value of `normalizeColor`. Was it just an oversight before that `normalizeColor` wasn't used for `colorA` and `colorB`?

gh-badges/lib/make-badge.spec.js
```diff
@@ -93,34 +92,18 @@ describe('The badge generator', function() {
   })
 
   describe('JSON', function() {
-    it('should always produce the same JSON (unless we have changed something!)', function() {
```
calebcartwright: Best. Test. Name. Ever!

gh-badges/lib/make-badge.js
paulmelnikow: For the purpose of the rendered badge, it doesn't matter whether or not you normalize. Capital or lowercase color codes are both recognized by the browser.

I guess it seems less surprising for the badge to use the same, normalized color, as the JSON response.

gh-badges/lib/make-badge.spec.js
paulmelnikow: Too bad I killed it…

gh-badges/lib/make-badge.js
calebcartwright: > I guess it seems less surprising for the badge to use the same, normalized color, as the JSON response.

Completely agreed!

gh-badges/lib/make-badge.spec.js
calebcartwright: Still gave me a good chuckle

paulmelnikow:
> The tests failing on label mismatches on the GitHub badges has me a bit puzzled 🤔

We probably could try to improve the error messages our test helper is generating. I think what's happening is that we're getting something like **issues | inaccessible**.

calebcartwright:
> I think what's happening is that we're getting something like issues | inaccessible.

I think you're right.

gh-badges/lib/make-badge.js
```diff
@@ -109,12 +107,27 @@ function makeBadge({
...
+
+  // This ought to be the responsibility of the server, not `makeBadge`.
+  if (format === 'json') {
+    return JSON.stringify({
+      label: left,
+      message: right,
+      logoWidth,
+      color,
+      labelColor,
+      link: links,
```
calebcartwright: Is there a need to do any of the encoding `escapeXml` is doing for the json response? I can't think of anything off hand but figured I'd ask anyway

paulmelnikow: Nope! Though now that you mention it, I think that is what's caused the (desirable) change in the link where `&` are no longer escaped as `&amp;`.

paulmelnikow:
Thanks for reviewing!

PR #3396, opened 2019-04-29
[Wordpress] Support tested version without trailing ".0"

paulmelnikow:
See https://circleci.com/gh/badges/shields/52378 for an example failure, where a value of `5.2` is triggering "TypeError: Invalid Version: 5.2"


services/wordpress/wordpress-version-color.spec.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const { expect } = require('chai')
+const { versionColorForWordpressVersion } = require('./wordpress-version-color')
+
+describe('Wordpress version color helpers', function() {
+  it('versionColorForWordpressVersion()', async function() {
+    expect(await versionColorForWordpressVersion('11.2.0')).to.equal(
```
calebcartwright: I wish it was easier to test the logic here without having to make the API call, but understand why 👍

paulmelnikow: We could mock it, though I think we probably also need a live test.

We do have at least one other core test that hits an API, which is the test of Node's `async render`.

I could see moving these into the integration test suite. Not sure if it is worth it, until it starts causing a problem…

calebcartwright: > Not sure if it is worth it

That's my personal sense on all points here; I don't think it's worth making any other changes.

> We could mock it, though I think we probably also need a live test.

I believe the service tests (the ones in platform tester for example) are already hitting this to provide some live validation. Again I wouldn't suggest any changes, just noting that from my point of view it'd be nice if we could unit test that logic sans live API call.

calebcartwright:
I went to do a final look over in the staging app but noticed it had errored out. 

Logs claim to be missing a dependency `Error: Cannot find module 'escape-string-regexp'`? 

Going to update the branch and see if that goes away.

PR #3397, opened 2019-04-30
Distro version improvements

scop:

PR #3398, opened 2019-04-30
Add [Ubuntu] package version support

scop:
https://api.launchpad.net/1.0/

Requires #3397 first for epoch version validation support.

### DIFF
--- a/services/ubuntu/ubuntu.service.js
+++ b/services/ubuntu/ubuntu.service.js
@@ -40,7 +40,7 @@ module.exports = class Ubuntu extends BaseJsonService {
     return { label: 'ubuntu' }
   }
 
-  async handle({ series, packageName }) {
+  async fetch({ series, packageName }) {
     const seriesParam = series
       ? {
           distro_series: `https://api.launchpad.net/1.0/ubuntu/${encodeURIComponent(
@@ -48,7 +48,7 @@ module.exports = class Ubuntu extends BaseJsonService {
           )}`,
         }
       : {}
-    const data = await this._requestJson({
+    return this._requestJson({
       schema,
       url: 'https://api.launchpad.net/1.0/ubuntu/+archive/primary',
       options: {
@@ -65,6 +65,10 @@ module.exports = class Ubuntu extends BaseJsonService {
         400: 'series not found',
       },
     })
+  }
+
+  async handle({ series, packageName }) {
+    const data = await this.fetch({ series, packageName })
     if (!data.entries.length) {
       throw new NotFound()
     }
--- a/services/ubuntu/ubuntu.service.js
+++ b/services/ubuntu/ubuntu.service.js
@@ -61,6 +61,9 @@ module.exports = class Ubuntu extends BaseJsonService {
           ...seriesParam,
         },
       },
+      errorMessages: {
+        400: 'series not found',
+      },
     })
     if (!data.entries.length) {
       throw new NotFound()
--- a/services/ubuntu/ubuntu.service.js
+++ b/services/ubuntu/ubuntu.service.js
@@ -1,0 +1,72 @@
+'use strict'
+
+const Joi = require('joi')
+const { renderVersionBadge } = require('../version')
+const { BaseJsonService, NotFound } = require('..')
+
+const schema = Joi.object({
+  entries: Joi.array()
+    .items(
+      Joi.object({
+        source_package_version: Joi.string().required(),
+      })
+    )
+    .required(),
+}).required()
+
+module.exports = class Ubuntu extends BaseJsonService {
+  static get category() {
+    return 'version'
+  }
+
+  static get route() {
+    return {
+      base: 'ubuntu/v',
+      pattern: ':series?/:packageName',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Ubuntu package',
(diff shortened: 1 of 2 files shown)


services/ubuntu/ubuntu.service.js
```diff
@@ -0,0 +1,72 @@
...
+        qs: {
+          'ws.op': 'getPublishedSources',
+          exact_match: 'true',
+          order_by_date: 'true',
+          status: 'Published',
+          source_name: packageName,
+          ...seriesParam,
+        },
+      },
+    })
```
calebcartwright: It might be worth extracting all the above into a separate `fetch` function which is a pretty common pattern we use in similar circumstances in other service classes

services/ubuntu/ubuntu.service.js
```diff
@@ -0,0 +1,72 @@
...
+
+  async handle({ series, packageName }) {
+    const seriesParam = series
+      ? {
+          distro_series: `https://api.launchpad.net/1.0/ubuntu/${encodeURIComponent(
+            series
+          )}`,
+        }
+      : {}
+    const data = await this._requestJson({
```
calebcartwright: What would you think about including an http 400 error message check to catch the invalid series (similar to what you did in in #3372 at https://github.com/badges/shields/pull/3372/files#diff-339abe4445a00826d7df94d8429ccf9fR49). 

Perhaps the message could be something like `invalid series` or `series not found`?

scop: Ah, somehow I was under the impression that the service wouldn't actually respond with 400 for such requests. But it in fact does, thanks for pointing it out! And done.

services/ubuntu/ubuntu.service.js
scop: I can't immediately come up with something that would make it any better than  `_ requestJson` already is. It's just about constructing a URL with an optional query string, that's what we'd have to do for such a `fetch` function as well, and at that point it'd be pretty much just a pass-through to `_requestJson`.

Other service classes where that is done are different in the sense that they have multiple different services use the same API but utilize different things from its data. There it makes sense to have a common fetch, but in this case there's only a version service, and nothing else that could share the functionality. At least for now.

calebcartwright: > I can't immediately come up with something that would make it any better than _ requestJson already is

I don't follow what you mean by that, so let me try to rephrase my suggestion. 

A really common pattern that you'll find in most service classes is the utilization of `transform` and `fetch` methods, as [highlighted in our docs on writing service classes](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#43-querying-an-api). It is indeed often used in base classes/service families, but it is also commonly used in individual service classes like this one. It's mostly stylistic, but helps with readability/responsibility. 

Here's a few example classes that are using this `fetch` pattern, many with much simpler arguments being passed to `_requestJson`:

* [homebrew](https://github.com/badges/shields/blob/fafb22efee005122be095c4f74212f01df92c139/services/homebrew/homebrew.service.js#L39)
* [dub-version](https://github.com/badges/shields/blob/f8aeb56129954e613e24112c94b3eb4e02a4dff9/services/dub/dub-version.service.js#L35)
* [cookbook](https://github.com/badges/shields/blob/fafb22efee005122be095c4f74212f01df92c139/services/cookbook/cookbook.service.js#L35)
* [jsDeliver](https://github.com/badges/shields/blob/f8aeb56129954e613e24112c94b3eb4e02a4dff9/services/jsdelivr/jsdelivr-hits-github.service.js#L27)
* [bintray](https://github.com/badges/shields/blob/fafb22efee005122be095c4f74212f01df92c139/services/bintray/bintray.service.js#L44)
* [maven-metadata](https://github.com/badges/shields/blob/3f4097b355360f274356a0d2cdbfd97b06b087f5/services/maven-metadata/maven-metadata.service.js#L52)
* [github-repo-size](https://github.com/badges/shields/blob/6e203d2455144943c740335426b3fb55215425d7/services/github/github-repo-size.service.js#L53)
* [jitpack-version](https://github.com/badges/shields/blob/c7b37a5de835d0b95433b15f9e0ddc8d28bca05b/services/jitpack/jitpack-version.service.js#L45)

It's not a blocker, but I think that extracting these first ~25 lines into a separate `fetch` function:

```js
  async fetch({ series, packageName }) {
    const seriesParam = series
      ? {
          distro_series: `https://api.launchpad.net/1.0/ubuntu/${encodeURIComponent(
            series
          )}`,
        }
      : {}
    return this._requestJson({
      schema,
      url: 'https://api.launchpad.net/1.0/ubuntu/+archive/primary',
      options: {
        qs: {
          'ws.op': 'getPublishedSources',
          exact_match: 'true',
          order_by_date: 'true',
          status: 'Published',
          source_name: packageName,
          ...seriesParam,
        },
      },
      errorMessages: {
        400: 'series not found',
      },
    })
  }
```

and then updating `handle` accordingly to something like the below would improve readbility IMO
```js
  async handle({ series, packageName }) {
    const data = await this.fetch({ series, packageName })
    if (!data.entries.length) {
      throw new NotFound()
    }
    return renderVersionBadge({
      version: data.entries[0].source_package_version,
    })
  }
```

scop: Ah, you mean readability, I thought you were after reuse. Sure, I have nothing against that (and it could help with extending to other services besides version), will change.

PR #3409, opened 2019-05-02
Move optional path components to the end for [Debian] and [Ubuntu]

scop:
Refs https://github.com/badges/shields/pull/3372


paulmelnikow:
Thanks for this! Much appreciate it, especially since we disagreed.

Issue #3393, opened 2019-04-29
Inject serverSecrets


paulmelnikow:
See https://github.com/badges/shields/issues/2832#issuecomment-469550829:

`server-secrets.js` should be eliminated in favor of injecting the secrets through `BaseService.register()`.

This will require touching all the affected service implementations.

mrahmadalshammari12:
> 

1.

mrahmadalshammari12:
[مستند بلا عنوان.docx](https://github.com/badges/shields/files/3316595/default.docx)

paulmelnikow:
The important parts of this are complete.

These are the remaining references to `serverSecrets`:

```
core/server/secret-is-valid.js:const serverSecrets = require('../../lib/server-secrets')
core/server/secret-is-valid.js:    serverSecrets.shields_secret &&
core/server/secret-is-valid.js:    constEq(secret, serverSecrets.shields_secret)
dangerfile.js:        (diff.includes('authHelper') || diff.includes('serverSecrets')) &&
services/github/auth/acceptor.js:const serverSecrets = require('../../../lib/server-secrets')
services/github/auth/acceptor.js:  } = serverSecrets
services/github/auth/acceptor.spec.js:const serverSecrets = require('../../../lib/server-secrets')
services/github/auth/acceptor.spec.js:    serverSecrets.shields_ips = undefined
services/github/auth/acceptor.spec.js:    serverSecrets.shields_secret = undefined
services/github/auth/acceptor.spec.js:    sinon.stub(serverSecrets, 'shields_ips').value([])
services/github/auth/acceptor.spec.js:    sinon.stub(serverSecrets, 'shields_secret').value(fakeShieldsSecret)
services/github/auth/admin.spec.js:const serverSecrets = require('../../../lib/server-secrets')
services/github/auth/admin.spec.js:    serverSecrets.shields_secret = undefined
services/github/auth/admin.spec.js:      .stub(serverSecrets, 'shields_secret')
services/github/github-constellation.js:const serverSecrets = require('../../lib/server-secrets')
services/github/github-constellation.js:    const globalToken = serverSecrets.gh_token
services/github/github-helpers.js:const serverSecrets = require('../../lib/server-secrets')
services/github/github-helpers.js:  return Boolean(serverSecrets.gh_token)
```

Getting rid of these will require a little bit more refactoring, which shouldn't be terribly difficult, though may not have quite as much payoff as it did in the services.

paulmelnikow:
[...]

Issue #3368, opened 2019-04-27
Modernizing the request code


paulmelnikow:
I opened an issue for modernizing the server #3328, though I think probably we could have a separate issue for refactoring the request code and timeout code.

I've mentioned in previous discussions that I'd like to experiment with disabling the in-memory cache. It adds complexity, is untested, and I think it is not buying us a lot, especially now that we have a downstream CDN. If we see the hit rate is low, and/or disabling it doesn't send performance through the floor, I think we can drop it as part of rewriting the request code.

(It may be difficult to untangle the decision about dropping the in-memory badge cache from other questions about caching. In brief, the cache that I think would be most helpful right now would be of API responses (#1545), either raw or transformed. Those could be better shared between different badges. However our most critical performance issues probably are #3027 and adding a little more capacity, and most critical ops issues are #3332 and the deployment bottleneck.)

The other thing the legacy request handler does it handle timeouts. However timeouts are much simpler, and are already built into client libraries like [got](https://github.com/sindresorhus/got).

In Discord @calebcartwright also mentioned request's EOL, and that request has good support for proxying which he uses in his self-hosting environment. We'll have to find a solution for that, too.

paulmelnikow:
I opened #7877 which removes this code.

PR #963, opened 2017-04-26
[RFC] New API for registering services

Daniel15:
This is a proof-of-concept for what a simplified API for registering services could look like. It's actually not that much code - Essentially it's just a simple facade over all the stuff that exists today. To keep things simple, I just kept everything in `server.js` for now. I converted over the AppVeyor service as an example.

Some notes:
- Code duplication is mostly removed. 
  - Services no longer need to manually call `cache` or `sendBadge`. The service handler simply returns the badge data, and the wrapper around it handles caching
  - Service URLs no longer need to manually append `.(svg|png|gif|jpg|json)` to the end, it's handled automatically
- Metadata (name, type, and example URLs) are included so that we can automatically generate the `try.html` page in the future

Note that this does use async functions, which is a new feature in Node.js 7.x.  This means we'd need to bump the Node.js version from 6 to 7. However, I think the code clarity is worth it! The handler has a nice linear flow with a regular try-catch block for handling network errors. We could use promises instead, but async function are the future! 😃 

Tested the following URLs on my computer (using Node.js 7.8) and it works nicely:
http://localhost:8080/appveyor/ci/gruntjs/grunt.svg
http://localhost:8080/appveyor/ci/gruntjs/grunt/master.svg
http://localhost:8080/appveyor/ci/kittens/yarn.svg

cc @paulmelnikow
References #948

PR #3410, opened 2019-05-02
RFC for injecting server secrets + refactoring auth code [azuredevops bintray bitbucket]

paulmelnikow:
I started some work on #3393 and thought I'd ask for some feedback before I applied it to all the services.

Come to think of it, it may actually make sense to keep the old version working so this can be done in a more gradual way.


paulmelnikow:
@calebcartwright @chris48s would be great to hear your thoughts on this.

services/azure-devops/azure-devops-helpers.js
```diff
@@ -23,15 +23,8 @@ async function fetch(serviceInstance, { url, qs = {}, errorMessages }) {
   return { status }
 }
 
-function getHeaders() {
-  const headers = {}
-  if (serverSecrets.azure_devops_token) {
-    const pat = serverSecrets.azure_devops_token
-    const auth = Buffer.from(`:${pat}`).toString('base64')
```
calebcartwright: In my experience working with the AzDO APIs, this has always been a necessary step for using PATs vs. username/password. At first glance, I suspect that this change would actually break self-hosted Shields instances connecting to AzDO. 

When I get a chance I'll check out this branch and test it though to find out definitively

paulmelnikow: It'd be great if you could try it.

I _think_ this code is manually constructing an HTTP Basic auth header with an empty username: https://stackoverflow.com/a/43843163/893113

That's ultimately [what request is doing, too](https://github.com/request/request#http-authentication).

calebcartwright: Yeah, and it _should_ work but for whatever reason I could never get things to work this way. Hopefully that's just a user error/fluke though (or something that's since been resolved) 🤞

calebcartwright: It works! This is one of those times that I am thrilled that I was wrong 😄

services/auth.js
```diff
@@ -0,0 +1,35 @@
...
+    throw Error('serviceInstance should be an instance of BaseService')
+  }
+
+  return {
+    user: userKey ? serviceInstance.getSecret(userKey) : '',
+    pass: passKey ? serviceInstance.getSecret(passKey) : '',
+  }
+}
+
+function optionalAuth(serviceInstance, userKey, passKey) {
```
calebcartwright: Could we use named params/object here, or do we want consumers to explicitly declare value for the `userKey` arg when they don't need (like in the case of Azure DevOps)?

services/bitbucket/bitbucket.tester.js
```diff
@@ -152,35 +145,6 @@ t.create('pr (server, not found)')
     message: 'not found',
   })
 
-t.create('pr (auth)')
```
calebcartwright: What's the thinking behind removing these tests, and would you anticipate removing the same types of tests in other services?

calebcartwright: Sorry I was reading from the bottom up, I see this validation is now handled above in the new `spec` file 👍

calebcartwright:
I've still not gone through this fully yet, but wanted to go ahead and share that I like the looks of it and the direction!

I don't really have a preference one way or the other, but what do we see as the pros/cons of implementing the auth/secret access capabilities via helper functions that are imported (which is very similar to the existing `server secrets` approach) vs. incorporating these capabilities in service class hierarchy (for example adding the `auth` functions somewhere in the BaseService class tree, or extending the existing `_request*` functions, etc.)?

paulmelnikow:
Hmm, interesting, I could see how adding `requiredAuth` and `optionalAuth` as methods on `BaseService` could make calling this a bit easier. On the other hand, I'm reluctant to add to any more lines of code than necessary to `base.js`, especially for something that's so specific to a minority of the services.

I also thought about adding an auth helper object that would tuck all this stuff into a namespace, but wasn't thrilled about adding allocation overhead to every request. Though as I think about it more, in the context of all the work that is done for every request, this is probably a pointless optimization to consider.

Something that goes back to #963 that I'd like to do is create one instance of a service for the life of the server. Since all the request-specific state is already passed through method calls, this is fine; the stuff being passed into the constructor is mostly the same. The reason this hasn't been possible is that the `_sendAndCacheRequest` function has a timeout built into it, starting from the time a specific request comes in. Though after the timeout is reworked in #3368 I expect that could change. After that's done, the auth helper could be created in the constructor, and reused across each invocation of the service.

services/auth.js
paulmelnikow: Nice idea! Updated.

paulmelnikow:
I could try a second version of this that:

1. Requires services to declare the secrets they use. This is good for documentation!
2. When a service declares it uses secrets, the BaseService constructor could attach to new instances an auth helper object.
3. The auth helper object would have access to the declared secrets (`getSecret` and `tryGetSecret`) and would also provide `optionalAuth` and `requiredAuth` as convenience methods.
4. As a further optimization, after #3368 this code will only run once in the lifetime of the server.

calebcartwright:
Sounds intriguing!

Issue #3260, opened 2019-04-02
TypeError: s.replace is not a function


paulmelnikow:
https://sentry.io/organizations/shields/issues/966629668/?project=251614&referrer=github_plugin

```
TypeError: s.replace is not a function
  File "/home/m/shields/core/badge-urls/make-badge-url.js", line 53, in encodeField
    return encodeURIComponent(s.replace(/-/g, '--').replace(/_/g, '__'))
  File "/home/m/shields/core/badge-urls/make-badge-url.js", line 65, in staticBadgeUrl
    const path = [label, message, color].map(encodeField).join('-')
  File "/home/m/shields/core/server/server.js", line 209, in camp.route
    const redirectUrl = staticBadgeUrl({
  File "/home/m/shields/core/server/monitor.js", line 57, in Array.server.handle
    next()
...
(28 additional frame(s) were not displayed)

TypeError: s.replace is not a function
```

Probably this happens when that badge gets passed a number?

paulmillr:
this happens if you enter https://github.com/paulmillr/chokidar into the first input with any browser (safari, chrome)

paulmelnikow:
Hi! We're tracking the issue you're mentioning in #3033.

It's unlikely to be related to this issue which is happening at runtime on the server.

PR #3412, opened 2019-05-05
Redirect an old png badge with a number as a color; test on [static]

platan:
Fixes https://github.com/badges/shields/issues/3260

Problem happens when a value of a color in an old PNG static badge is a number: http://localhost:8080/my-label/my-message.png?color=1. In this case `color` in `queryParams` is a number. 
https://github.com/badges/shields/blob/0a0b5b3f031178851c9d295b9967d81ede052723/core/server/server.js#L203-L212

Surprisingly service test listed below is passing currently on master - value `1` is represented in `queryParams` as a String (only in test). 
`services/static-badge/static-badge.tester.js`
```js
t.create('Old static badge with a number as a color')
  .get('/foo/bar.png?color=1', { followRedirect: false })
  .expectStatus(301)
  .expectHeader('Location', '/badge/foo-bar-1.png')
```

Moreover I added some code + description allowing to debug server.


.vscode/launch.json
```diff
@@ -0,0 +1,13 @@
+{
```
platan: This configuration comes from https://github.com/Microsoft/vscode-recipes/tree/master/nodemon#configure-vs-code-debugging-with-a-launchjson-file

core/badge-urls/make-badge-url.js
```diff
@@ -62,7 +62,10 @@ function staticBadgeUrl({
   namedLogo,
   format = 'svg',
 }) {
-  const path = [label, message, color].map(encodeField).join('-')
+  const path = [label, message, color]
+    .map(s => s.toString())
```
paulmelnikow: Could you move the coercion to the calling side instead? I feel like it's reasonable for this function to expect strings. It's really the web framework that is a bit unreasonable, about giving us different types depending on the input.

services/static-badge/static-badge.tester.js
```diff
@@ -62,3 +62,8 @@ t.create('Old static badge')
   .get('/foo/bar.png?color=blue', { followRedirect: false })
   .expectStatus(301)
   .expectHeader('Location', '/badge/foo-bar-blue.png')
+
+t.create('Old static badge without a color')
```
paulmelnikow: Could you add a test of the behavior that you're fixing?

core/badge-urls/make-badge-url.js
platan: Thanks! Done.

services/static-badge/static-badge.tester.js
platan: I can. But as I wrote in PR description, this test is passing without a fix:
```js
t.create('Old static badge with a number as a color')
  .get('/foo/bar.png?color=1', { followRedirect: false })
  .expectStatus(301)
  .expectHeader('Location', '/badge/foo-bar-1.png')
```
A `color` value (in `queryParams`) is a string, not a number.

platan: In my opinion is it not worth to add a test which gives a false sense of working solution.

core/server/server.js
```diff
@@ -207,7 +207,8 @@ module.exports = class Server {
       const redirectUrl = staticBadgeUrl({
         label,
         message,
-        color,
+        // Fixes https://github.com/badges/shields/issues/3260
+        color: color && color.toString(),
```
paulmelnikow:
```suggestion
        color: color ? color.toString() : undefined,
```

services/static-badge/static-badge.tester.js
paulmelnikow: Ahhh I understand now. For some reason when this request is made from the test, Scoutcamp is returning the query string param as a string, but when it's made to the real server it gets a number. That's really puzzling.

I'm hoping to replace Scoutcamp (#3328) so I don't feel like it's worth digging further into why it's happening.

platan:
Thanks for merging it :-)

Issue #3402, opened 2019-04-30
Add redirector documentation


calebcartwright:
We've been using the `redirector`s more and more, and it'd be helpful to document the how/when/why/etc. for redirectors that folks can reference.

PR #3413, opened 2019-05-05
 switch [circleci] service to scrape SVG badge, reorganise routes

### LINKED ISSUE #1064: CircleCI : use only Status permission for the API calls

### LINKED ISSUE #1792: Our test badge incorrectly stays green

### LINKED ISSUE #1872: Circle CI badge should show status of master branch by default
As noted in #1859 the standard behaviour for our Circle CI badge is to show the status of the latest build on _any_ branch. However, for consistency with the behaviour of other build service badges and [Circle CI's own badges](https://circleci.com/docs/1.0/status-badges/) the standard behaviour should be to show the status of the latest build on the repo's default branch.

We should also provide some way to maintain the current behaviour as some users may rely on this.

chris48s:
Closes #1064
Closes #1792
Closes #1872

Default URL is now `circleci/build/gh/badges/shields/master?token=abc123`
There are redirects on the legacy routes to provide backwards compatibility for existing users

For review, it is easier to read this diff if you read each commit in isolation


services/circleci/circleci.service.js
```diff
@@ -2,70 +2,48 @@
...
   }
 
   static get route() {
     return {
-      base: 'circleci',
-      format:
-        '(?:token/(\\w+)/)?project/(?:(github|bitbucket)/)?([^/]+/[^/]+)(?:/(.*))?',
-      capture: ['token', 'vcsType', 'userRepo', 'branch'],
+      base: 'circleci/build',
+      pattern: ':vcsType(gh|bb)/:user/:repo/:branch*',
```
calebcartwright: I think the vcs options should be the more user friendly terms no?
```suggestion
      pattern: ':vcsType(github|bitbucket)/:user/:repo/:branch*',
```

chris48s:
I quite like allowing `gh|bb` as it allows you to easily take the badge URL Circle CI gives you and convert it to a shields one.

Screenshot at 2019-05-08 20-52-44

but you're right that across shields we more commonly use the longhand versions.
I've switched it to allow `(github|gh|bitbucket|bb)`. Does that seem reasonable?

calebcartwright:
>  Does that seem reasonable?

Sounds good to me!

paulmelnikow:
> it allows you to easily take the badge URL Circle CI gives you and convert it to a shields one.

On a side note, it would be great to add this capability to the frontend: where you can paste in a URL from a service and it transforms it into a badge URL. It's a similar idea to #2420.

services/build-status.js
```diff
@@ -27,6 +27,7 @@ const otherStatuses = [
   'cancelled',
   'expired',
   'initiated',
+  'no builds',
```
calebcartwright: We should probably update the corresponding test in [build-status.spec.js](https://github.com/badges/shields/blob/master/services/build-status.spec.js#L67) with this entry since that test includes the other status

services/circleci/circleci.tester.js
```diff
@@ -9,62 +9,58 @@ const t = (module.exports = new ServiceTester({
...
+t.create('legacy route with token')
+  .get(
+    '/token/b90b5c49e59a4c67ba3a92f7992587ac7a0408c2/project/github/RedSparr0w/node-csgo-parser/master.json'
   )
   .expectBadge({
     label: 'build',
-    message: 'invalid response data',
-    color: 'lightgrey',
+    message: isBuildStatus,
   })
```
calebcartwright: We admittedly don't have anything documented around our practices for redirectors yet (I opened #3402 to track creating those docs), but so far we've been testing them by disabling `followRedirect` and then checking the status and target url, example below from the vso redirector:

```js
t.create('Build: default branch')
  .get('/build/totodem/8cf3ec0e-d0c2-4fcd-8206-ad204f254a96/2.svg', {
    followRedirect: false,
  })
  .expectStatus(301)
  .expectHeader(
    'Location',
    '/azure-devops/build/totodem/8cf3ec0e-d0c2-4fcd-8206-ad204f254a96/2.svg'
  )
```

The approach you've taken here obviously validates the redirect route as well. This isn't a blocker for this PR by any means, but I just wanted to make note of it. Happy to shift this topic over to #3402 for discussion in the event folks have opinions on whether we have any preference for redirector testers (leave it up to the developer, prefer one or the other, etc.)

chris48s: Yeah I can see reasons to go both ways on this.
On the one hand, the thing we're testing is a redirector, so checking it issues a 301 is a unit testing approach to making sure that component does exactly what its supposed to do.
On the other hand, our service tests are really high-level integration tests, so testing the end-to-end behaviour also makes sense in this context.
There probably isn't a single correct answer to that, but it does make sense to pick one approach and document it as a convention. I'd be happy to go either way on it tbh.

paulmelnikow: Agreed there are reasons to go both ways here. When I wrote the first of these I picked this way, though to be honest neither way seems perfect.

I'm inclined to do what's easy to maintain.

A sort of middle-ground option is adding a helper that does a bit of both: check the redirect, then follow it, and finally do a basic check that the final result is not an error badge. Part unit test, part integration test, with a very lightweight assertion.

paulmelnikow:
Thanks for taking this on, Chris! And thanks Caleb for reviewing!

PR #3417, opened 2019-05-06
Convert remaining uses of node-fetch to got

paulmelnikow:
We're using a mix of node-fetch and got because some of what we need to do (specifically not follow redirects) is impossible in node-fetch.

This converts the rest of it over to got.


services/github/auth/admin.spec.js
```diff
@@ -60,11 +54,13 @@ describe('GitHub admin route', function() {
 
   context('the password is correct', function() {
     it('returns a valid JSON response', async function() {
-      const res = await fetch(`${baseUrl}/$github-auth/tokens`, {
-        headers: createAuthHeader(validCredentials),
+      const { username, password } = validCredentials
+      const { statusCode, body } = await got(`${baseUrl}/$github-auth/tokens`, {
+        auth: `${username}:${password}`,
```
PyvesB: Nice little simplification!

Issue #3404, opened 2019-05-01
Node version support policy


paulmelnikow:
Tangentially related to #3403:

We should set a policy of which _additional_ Node versions we support for:

1. The npm package
2. Self-hosting
3. Development

For the **npm package**, I'm inclined to do the most popular thing in the JS community, which is supporting LTS versions. That would mean today we'd start requiring 8+, at the end of 2019 we'd require 10+, and a year from now in April 2021, 12+. This strategy is allowing the community to keep things moving forward, and avoiding spending lots of resources supporting very old versions.

For **self-hosting** it would be great if we could support only the ["current" version](https://github.com/nodejs/Release). Or maybe – to give self-hosting users time to upgrade – any version that has been "current" within the last three months. Or maybe – to reduce the frequency of those upgrades – the latest even-numbered version.

The advantage of keeping this window really small is that it gives us the most options with modern syntax, without introducing a transpilation step.

For example, Node 12 supports public static class fields. We can replace:

```js
static get route() {
  return {
    base: 'github/license',
    pattern: ':user/:repo',
  }
}
```

with 

```js
static route = {
  base: 'github/license',
  pattern: ':user/:repo',
}
```

I'm pretty excited to be able to do that.

I want to make sure it's reasonable to self-host this project, and I feel like in many environments it is no big deal to specify whatever node version you might need. Though I don't want to _overinvest_ in self-hosting either.

If there's anyone who's likely to have a problem it could be enterprise users with tightly controlled environments. I'm not anti-enterprise and I'm fine supporting those users (maybe through transpilation) though if the work isn't widely benefiting the community, I'd want to fund it through support contracts.

calebcartwright:
What you've articulated above seems reasonable to me on all counts

I think that providing Shields via a Docker image (presumably through Dockerhub) would help a lot with the self-hosting scenarios. I think it's highly likely that most self-hosting users just want to have a Shields instance up and running and a working Docker image is a quick and easy means. 

I think how we version/tag our artifacts (the package, the Shields app, the potential Docker image, etc.) could be of use here as well (for example if someone _must_ use Node 6.x for whatever reason, then they  have to use older version X of the package, tag X of the repo, tag X of the docker image, etc.)

chris48s:
[...]

PR #3428, opened 2019-05-06
Clarify Node version support policy for package, and test in all supported versions

paulmelnikow:
Based on #3404, the four versions we support for the npm package are 8, 10, 11, and 12. Currently Node 8 is tested in our **main** build and Node 12 in **main@node-latest**. I added package-specific builds for 10 and 11.

We'll have to keep an eye on this as we fiddle with other builds. It could be easier if we added separate builds for each one, but having so many builds seems a bit unwieldy. Alternatively we could make a single build for the package, and use nvm or something like it to switch versions within that build.


.circleci/config.yml
```diff
@@ -109,36 +109,67 @@ services_steps: &services_steps
     - store_test_results:
         path: junit
 
-jobs:
-  npm-install:
```
paulmelnikow: I commented out the **npm-install** job, which we're not using right now.

paulmelnikow:
> Happy to merge this as it is, but I think this in the long run it will be easier if the range of node versions we test against is explicitly defined as opposed to implicit.

Pretty much agreed with that.

> FWIW, I don't see an issue with having an individual build for each node version.

My concern is that our build latency goes up even higher. (I think CircleCI gives us four simultaneous containers, which is why we often see "waiting behing your other queued builds".) Also scrolling through a long list in github is a bit harder.

Maybe we should attempt the nvm version?

> if we can only run the package tests on all the node versions if the PR touches files in the `gh-badges` dir.

I'm not sure there's a way to do that in Circle without actually starting a container. Also the list of files might be hard to contain. There's the root package.json, package-lock.json, and anything else that might affect the test toolchain.

calebcartwright:
> I think CircleCI gives us four simultaneous containers, which is why we often see "waiting behing your other queued builds"

I was wondering about this the other day too as it seems like the PR checks have been slower of late (although I know Circle had some platform issues recently too that may have been a factor). Do they offer any additional containers/benefits for OSS projects under any circumstances? 

Also, should we look into seeing if we can skip the cypress binary download (or ensure it is cached with the other contents), outside the job that's running the e2e tests? It seems like the npm install step is often the longest step in those jobs (45-60 seconds), and I suspect that may be a factor

chris48s:
> I'm not sure there's a way to do that in Circle without actually starting a container

I was thinking the same sort of thing as we do for the service tests, but you're right that initializing the container to decide whether to run the tests probably takes longer than running the tests in this case :D

paulmelnikow:
Superseded by #3498.

Issue #3433, opened 2019-05-07
WordPress rating examples broken


paulmelnikow:
I just noticed via #3432 that the WordPress rating example badges aren't rendering correctly:

<img width="926" alt="Screen Shot 2019-05-07 at 10 16 56 AM" src="https://user-images.githubusercontent.com/1487036/57306716-53440600-70b1-11e9-8c40-ad8373b6b612.png">

These badges shouldn't be displaying NaN, and I think some of the stars are supposed to be filled in.

PR #3434, opened 2019-05-07
Update references to "WordPress"

jeffpaul:
Fixes: badges#3432 and badges#3433


paulmelnikow:
Thanks!

I think #3433 is still an issue.

PR #3435, opened 2019-05-07
Improvements in the TUTORIAL.md file

JosepBernad:
While I was trying to install shields for the first time, I ran into a problem that I solved.
I've added the solution in the doc/TUTORIAL.md file in order to ease the process for the newbies like me.
Source of the solution: https://medium.com/andrewmmc-io/node-js-error-getaddrinfo-enotfound-localhost-b7ee35e1bb60


calebcartwright:
Thanks for this! Do you think it might be easier to just have the docs suggest trying to hit `127.0.0.1:3000` instead?

JosepBernad:
What do you think now @calebcartwright?

paulmelnikow:
HI! Did you run into this issue? What OS was that on?

Usually OS's will include this alias by default; it's rare that I've seen something like this happen.

I'm with Caleb… not sure I'd want to direct new users to edit their `/etc/hosts` file 🙈

calebcartwright:
I'd echo @paulmelnikow question about your OS if you experienced this issue. We want to make sure anyone can get Shields running on their local machine regardless of their OS, without having to manually edit the hosts file. 

If there's a particular OS/environment where our npm `start` script won't work unless the hosts file is manually updated, then I'd be inclined to review our scripts/config to see if we can address that instead.

JosepBernad:
So, do you guys think that the best option is to replace localhost with 127.0.0.1:3000?

paulmelnikow:
Can you say more about how you encountered this issue? What OS are you using?

JosepBernad:
I'm using macOS 10.12.6 (Sierra), node 10.15.3 and npm 6.4.1.
I was just following the TUTORIAL.md when I had this problem: 
`Error: getaddrinfo ENOTFOUND localhost.`
I researched it and found that the system can't find localhost. So, I decided to add the explanation to add localhost manually. 
Some folks told me that that's not the best way to do it so I think that the best way to do it is accessing straight to the IP address (instead of the localhost address).

What do you think, @paulmelnikow?

JosepBernad:
Also, I've just added the method in the first example, because it didn't compile:
`static get category() {`
  `return 'build';`
`  }`

paulmelnikow:
Thanks for adding the category!

Re /etc/hosts:

Here's an example of this issue reported on another project: angular/angular-cli#2227. I'm gleaning from that issue that this line is in /etc/hosts by default which certainly is also my expectation. So I think this issue occurs when the original contents of /etc/hosts is manually changed. Probably it is better if most users do not edit /etc/hosts for exactly this reason.

We could use 127.0.0.1 instead, though again, I'm not aware of any systems that don't have this mapping by default.

We shouldn't cover every eventuality in our tutorial and I don't want to give people the impression that they should be editing /etc/hosts (which seems to be what _caused_ this issue). I suppose we could add a troubleshooting section, and maybe link out to that issue. (particularly https://github.com/angular/angular-cli/issues/2227#issuecomment-358036526)

JosepBernad:
Nice, let me modify the file it with your proposal.

JosepBernad:
I've updated the TUTORIAL.md file with your proposals. Could someone review it in order to close it?
Thanks for your help. I'm really proud of my very first issue!

doc/TUTORIAL.md
```diff
@@ -37,8 +37,9 @@ install node and npm: https://nodejs.org/en/download/
    `npm ci`
 5. Run the badge server and the frontend dev server
    `npm start`
-6. Visit the website to check the front-end is loaded:
-   [http://localhost:3000/](http://localhost:3000/)
+6. Visit the website to check the front-end is loaded: [http://localhost:3000/](http://localhost:3000/).
+
+In case you get the _"getaddrinfo ENOTFOUND localhost"_ error, visit [http://127.0.0.1:3000/](http://localhost:3000) instead or take a look at [this issue](https://github.com/angular/angular-cli/issues/2227#issuecomment-358036526).
```
calebcartwright: Make sure to update the link part on this line too (it's currently still pointing to `localhost`)

```suggestion
In case you get the _"getaddrinfo ENOTFOUND localhost"_ error, visit [http://127.0.0.1:3000/](http://127.0.0.1:3000) instead or take a look at [this issue](https://github.com/angular/angular-cli/issues/2227#issuecomment-358036526).
```

calebcartwright:
This LGTM now. @paulmelnikow any other thoughts on this one?

JosepBernad:
I've just added the explanation for both `get category()` functions.

JosepBernad:
Can we close this PR or is there anything else to do?

calebcartwright:
> Can we close this PR or is there anything else to do?

Each PR has to be reviewed, and approved by at least one member of the maintainer team before it can be merged. Even after a PR has been approved, additional changes (like the one you made a few hours ago) restart the process. 

We'll take a look at your latest commit when we can and potentially provide additional feedback and/or request changes.

JosepBernad:
Thanks @calebcartwrigh! I I didn't want to seem to hurry, sorry if I did!

calebcartwright:
No worries!

doc/TUTORIAL.md
```diff
@@ -184,33 +196,38 @@ const schema = Joi.object({
 // (5)
 module.exports = class GemVersion extends BaseJsonService {
   // (6)
+  static get category() {
+    return 'build'
```
paulmelnikow:
```suggestion
    return 'version'
```

PR #3438, opened 2019-05-08
Show a badge pattern instead of a badge url

platan:
This PR contains changes:
- show a badge pattern instead of a badge url (https://github.com/badges/shields/pull/3331#issuecomment-489472413)
- remove a hostname from the badge url (https://github.com/badges/shields/pull/3331#issuecomment-489472748)

This PR does not remove hostname for suggested badges. Do we want to remove it there as well? In my opinion we do not have to do this, because suggested urls are working urls. 

This PR does not change patterns like `:vcsName(github|gh|bitbucket|bb|gl|gitlab)` to `:vcsName`. Patterns (https://github.com/badges/shields/blob/master/services/codecov/codecov.service.js#L55) contains these options (`(github|gh|bitbucket|bb|gl|gitlab)`), so we have to convert `/codecov/c/:vcsName(github|gh|bitbucket|bb|gl|gitlab)/:user/:repo/:branch` to `/codecov/c/:vcsName/:user/:repo/:branch`. I do not know what's the best way to do this. Any ideas?


frontend/components/markup-modal/markup-modal-content.js
```diff
@@ -16,7 +16,8 @@ export default function MarkupModalContent({ example, baseUrl }) {
     documentation,
     example: { pattern, namedParams, queryParams },
     link,
-    preview: { style: initialStyle, buildFromExample } = {},
+    preview: { style: initialStyle } = {},
+    realBadge,
```
paulmelnikow: Would it be clearer to call this property `isBadgeSuggestion`?

paulmelnikow:
> This PR does not change patterns like `:vcsName(github|gh|bitbucket|bb|gl|gitlab)` to `:vcsName`. Patterns (https://github.com/badges/shields/blob/master/services/codecov/codecov.service.js#L55) contains these options (`(github|gh|bitbucket|bb|gl|gitlab)`), so we have to convert `/codecov/c/:vcsName(github|gh|bitbucket|bb|gl|gitlab)/:user/:repo/:branch` to `/codecov/c/:vcsName/:user/:repo/:branch`. I do not know what's the best way to do this. Any ideas?

I haven't run this, though maybe something like it? It's adapted from `PathBuilder.constructPath`.

```js
const tokens = pathToRegexp.parse(pattern)
const path = tokens
  .map(token => {
    if (typeof token === 'string') {
      return token
    } else {
      const { delimiter, name, optional } = token
      if (optional) {
        return ''
      } else {
        return `${delimiter}:${name}`
      }
    }
  })
  .join('')
```

frontend/components/markup-modal/markup-modal-content.js
platan: It would be :-) Fixed.

platan:
@paulmelnikow thanks for suggestion. It helped me a lot. Now change is ready for the next round of review.

platan:
https://shields-staging-pr-3428.herokuapp.com does not show hostname for suggested url, because GATSBY_BASE_URL is not set. Similar behavior can be seen at https://shields-staging-pr-3438.herokuapp.com.

paulmelnikow:
> https://shields-staging-pr-3428.herokuapp.com does not show hostname for suggested url, because GATSBY_BASE_URL is not set. Similar behavior can be seen at https://shields-staging-pr-3438.herokuapp.com.

That will always be the case in a Heroku deploy, which typically doesn't know its deploy address at build time. You could default to the host from `window.location.href` when no base URL was configured. (I think there was/is some existing code that does that.)

platan:
@paulmelnikow would you like me to add something in this PR? I prefer not to deal with hostname at staging env in this PR.

paulmelnikow:
Okay! Then I will read through what is here.

frontend/lib/pattern-helpers.spec.js
```diff
@@ -1,10 +1,28 @@
...
   test(patternToOptions, () => {
     given('[^\\/]+?').expect(undefined)
     given('abc|[^\\/]+').expect(undefined)
     given('abc|def|ghi').expect(['abc', 'def', 'ghi'])
   })
+
+  test(removeRegexpFromPattern, () => {
+    given('/appveyor/ci/:user/:repo').expect('/appveyor/ci/:user/:repo')
+    given('/discourse/:scheme(http|https)/:host/topics').expect(
+      '/discourse/:scheme/:host/topics'
```
chris48s: This is a nice enhancement. Previously routes like `:variant(maintainability|maintainability-percentage|tech-debt|issues)` were generating some really long URLs on the index page and we don't need to do that now that we have drop-downs in the modal.

frontend/lib/pattern-helpers.js
```diff
@@ -11,4 +13,30 @@ function patternToOptions(pattern) {
...
+        if (typeof name === 'number') {
+          return `${delimiter}(${pattern})`
+        } else {
+          let modifier = ''
+          if (optional && !repeat) {
+            modifier = '?'
+          } else if (!optional && repeat) {
+            modifier = '+'
+          } else if (optional && repeat) {
+            modifier = '*'
```
paulmelnikow: I feel like in the future we may want to consider how we handle the *'s and ?'s. IIRC we have some patterns which have been added solely to avoid having the *'s show up in the frontend.

paulmelnikow:
Thanks for the nudge @chris48s and thanks @platan too!

platan:
@paulmelnikow thanks for review and suggestions. @chris48s thanks for review. Now it's finally merged.

Issue #3185, opened 2019-03-09
Reddit subreddit subscribers badge


crazy-max:
:clipboard: **Description**

> Which service is this badge for e.g: GitHub, Travis CI

Reddit

> What sort of information should this badge show?
> Provide an example in plain text e.g: "version | v1.01" or as a static badge
> (static badge generator can be found at https://shields.io)

Subreddit subscribers count badge like :

[![](https://img.shields.io/badge/follow%20r%2FPortapps-8-red.svg)](https://www.reddit.com/r/Portapps)

:link: **Data**

> Is there a public API?

Here : https://www.reddit.com/r/Portapps/about.json (replace _Portapps_ with another subreddit)

> Does the API requires an API key?

Seems not required

> Link to the API documentation.

https://www.reddit.com/dev/api/#GET_r_{subreddit}_about

:microphone: **Motivation**

Like Twitter badge, display subscribers count for a subreddit :)
I have already worked on previous badges, I can work on it if you want ;)

calebcartwright:
> I have already worked on previous badges, I can work on it if you want ;) 

Awesome!

RedSparr0w:
Just adding a quick note that you can also create this badge with the `dynamic-json` badge:
`https://img.shields.io/badge/dynamic/json.svg?label=r/Portapps%20subscribers&query=$.data.subscribers&url=https://www.reddit.com/r/Portapps/about.json`
![](https://img.shields.io/badge/dynamic/json.svg?label=r/Portapps%20subscribers&query=$.data.subscribers&url=https://www.reddit.com/r/Portapps/about.json)

But i think this could be a good addition to add to shields.

JosepBernad:
Guys, today I'm starting with this issue. I have an initial question: should we allow the user to change the color o use the reddit's color? Thanks!

JosepBernad:
I don't know exactly how to do it, but I've created a new branch in my fork in order to show you my progress. 
Am I doing the workflow correctly?

[That](https://github.com/JosepBernad/shields/commit/13c6ed8e2bde06487f77f73d163dba3be9587143)'s the last commit where I've just pushed what I've done. I'm a little bit stuck because I don't know how to solve the issues. 

I've been using other services in order to learn how it works.

calebcartwright:
#3443 is ready to merge to add this badge, but there is one item we'd like to get your thoughts on @crazy-max as the OP of this issue. 

The new badge #3443 will add the badge using our social badge style, so the badges will look like:
[...]

PR #3442, opened 2019-05-11
Add [Subreddit] subscribers - old

JosepBernad:
That's my first commit to solve this issue #3185. I need some help in order to solve the problems I currently have:
ValidationError: Example for SubredditSubscribers at index 0 {
  "title": "Subreddit followers",
  "namedParams": {
    "subreddit": "drums"
  },
  "staticPreview": {
    "label": "follow r/drums",
    "message": "15k",
    "style": "flat",
    "color": "red",
    "link" [1]: [
      "https://www.reddit.com/r/drums/"
    ]
  }
}

[1] "link" is not allowed

Any thoughts?

### DIFF
--- a/doc/TUTORIAL.md
+++ b/doc/TUTORIAL.md
@@ -39,7 +39,7 @@ install node and npm: https://nodejs.org/en/download/
    `npm start`
 6. Visit the website to check the front-end is loaded: [http://localhost:3000/](http://localhost:3000/).
 
-In case you can't see the front-end page, visit [http://127.0.0.1:3000/](http://localhost:3000) instead or take a look at [this issue](https://github.com/angular/angular-cli/issues/2227#issuecomment-358036526).
+In case you get the _"getaddrinfo ENOTFOUND localhost"_ error, visit [http://127.0.0.1:3000/](http://localhost:3000) instead or take a look at [this issue](https://github.com/angular/angular-cli/issues/2227#issuecomment-358036526).
 
 You may also want to install
 [ImageMagick](https://www.imagemagick.org/script/download.php).
--- a/doc/TUTORIAL.md
+++ b/doc/TUTORIAL.md
@@ -38,9 +38,10 @@ install node and npm: https://nodejs.org/en/download/
 5. Run the badge server and the frontend dev server
    `npm start`
 6. Visit the website to check the front-end is loaded:
-   [http://localhost:3000/](http://localhost:3000/).
-   If you get _"Error: getaddrinfo ENOTFOUND localhost"_, add the next line to the /etc/host file:
-   `127.0.0.1 localhost`
+   [http://localhost:3000/](http://localhost:3000/) or [http://127.0.0.1:3000/](http://127.0.0.1:3000/).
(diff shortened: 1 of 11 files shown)


services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,67 @@
...
+
+  static render({ subreddit, subscribers }) {
+    console.log('Render')
+    return {
+      label: `follow r/${subreddit}`,
+      message: metric(subscribers),
+      style: 'flat',
+      color: 'red',
+      link: [`https://www.reddit.com/r/${subreddit}/`],
+    }
```
calebcartwright: Try removing the `style` and `link` for now

services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,67 @@
...
+    return {
+      base: 'reddit/subreddit-subscribers',
+      pattern: ':subreddit',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Subreddit followers',
```
calebcartwright: This should probably also use the `subscribers` verbiage to be consistent with everything else

doc/TUTORIAL.md
```diff
@@ -37,8 +37,9 @@ install node and npm: https://nodejs.org/en/download/
    `npm ci`
```
calebcartwright: Let's keep these two changes separate. You've got #3435 opened to cover the tutorial changes, so let's not duplicate them

calebcartwright:
As to the issue described in your PR description, `staticPreview` currently doesn't support `link` which is why you're getting that error.

services/reddit/subreddit-subscribers.service.js
JosepBernad: Done, thanks!

doc/TUTORIAL.md
JosepBernad: Thanks @calebcartwright , actually I wanted to avoid the changes made in my local copy previously, but I couldn't. Do you know how to delete those commits?

services/reddit/subreddit-subscribers.service.js
JosepBernad: Done, thanks!

doc/TUTORIAL.md
calebcartwright: Ah I see. Looks like you did the original tutorial work in #3435 off your local master branch, and then you subsequently created this feature/subreddit-subscriber branch off of your master branch correct?

JosepBernad: Yes, I created the branch from my last commit instead of the last commit of the master of the original repository. 
Do you want me to remove this PR, and create another one properly?

calebcartwright: Yeah I think that would be best, thanks! For reference I'd suggest always using feature branches in your forked repositories to simplify this process, especially if you're going to work on separate efforts in parallel. 

I'll keep an eye out for a new PR

PR #3443, opened 2019-05-11
Add [Subreddit] subscribers

JosepBernad:
PR for the issue #3185.
It's not working properly but the main code is written. Tests need to be done too.
Thanks to @calebcartwright for the help!


services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,60 @@
+'use strict'
+
+const { BaseJsonService } = require('..')
+const Joi = require('joi')
+const { nonNegativeInteger } = require('../validators')
+const { metric } = require('../text-formatters')
+
+const schema = Joi.object({ subscribers: nonNegativeInteger })
```
calebcartwright: I believe that the `subscribers` field is contained within the `data` object key in the response isn't it?

For example
```json
{
  "kind":"t5",
  "data": {
      ...
     "subscribers": x,
      ...
   }
}
```

JosepBernad: Yes, I'm filtering the response here:
`async handle({ subreddit }) {`
`const json = await this.fetch({ subreddit })`
`return this.constructor.render({`
`subreddit,`
`subscribers: json.data.subscribers,`
`})`
}`

Am I doing it wrong?

calebcartwright: The schema is used to validate the raw response, but you've currently defined the schema structured to expect the `subscribers` field on the root response object. That does not match the actual response, so it's being rejected and the badge is not working:

https://shields-staging-pr-3443.herokuapp.com/reddit/subreddit-subscribers/drums.svg


You'll need to update the schema to match the response, which I think would be something like this:
```js
const schema = Joi.object({
  data: Joi.object({
    subscribers: nonNegativeInteger,
  }).required()
}).required()
```

JosepBernad: You're right, now it's working properly. I thought the `schema` worked in another way. Pushing the new version. Thanks!

JosepBernad:
I'm struggling with adding the link to the badge, but I don't really know where to place it.
My first thought was to add it to the render function, but I get the compilation error:
`[1] "link" is not allowed`.
Does anyone know where to place it? 
Thanks in advance!

calebcartwright:
Let's get the core part working, tests passing, etc. first, and then look at adding the link.

services/reddit/subreddit-subscribers.tester.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const { isMetric } = require('../test-validators')
+const { ServiceTester } = require('../tester')
```
calebcartwright: We can use the standard `createServiceTester` helper here, see the [tests tutorial](https://github.com/badges/shields/blob/master/doc/service-tests.md#1-boilerplate) for more info

calebcartwright:
As far as making the badges with the social style and link, the typical pattern with these sorts of badges is to use a literal for the example preview instead of invoking the `render` function. 

That would change the `examples` function to something like:
```js
  static get examples() {
    return [
      {
        title: 'Subreddit subscribers',
        namedParams: { subreddit: 'drums' },
        staticPreview: {
          label: 'follow r/drums',
          message: '15k',
          color: 'red',
          style: 'social',
       },
      },
    ]
  }
```

and then in your `render` function you can include the `link` field:
```js
  static render({ subreddit, subscribers }) {
    return {
      label: `follow r/${subreddit}`,
      message: metric(subscribers),
      color: 'red',
      link: [...],
    }
  }
```

JosepBernad:
@calebcartwright, do you prefer the social style or the original one proposed in the issue (flat and red).

services/reddit/subreddit-subscribers.tester.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+const { isMetric } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('subreddit-subscribers (valid subreddit)')
+  .get('/subreddit-subscribers/drums.json')
```
calebcartwright: You don't need the additional path reference here anymore, as the `createServiceTester` helper does a lot of the path prefixing for you
```suggestion
  .get('/drums.json')
```

services/reddit/subreddit-subscribers.tester.js
```diff
@@ -0,0 +1,18 @@
...
+  .expectBadge({
+    label: 'follow r/drums',
+    message: isMetric,
+  })
+
+t.create('subreddit-subscribers (invalid subreddit)')
+  .get('/subreddit-subscribers/not-a-real-subreddit.json')
```
calebcartwright:
```suggestion
  .get('/not-a-real-subreddit.json')
```

calebcartwright:
My apologies. I assumed since you were trying to add the links into the badge definition (and given the social nature of this badge) that the plan was to add this badge in the social format as our other social badges. 

Take our twitter follow badge for example:


The badge has embedded links: the label to has the clickable link to "follow" and the message/count has a link to view the followers. I assumed you were trying to do something similar for this badge, i.e.:


If that badge style is **not** the intent for the Reddit Subreddit Subscribers badge (i.e. if you just want to do the standard badge format) then you can forget about adding the links and switch to the more typical `render` and `staticPreview` usage from the examples:

```js
  static render({ subreddit, subscribers }) {
    return {
      label: `follow r/${subreddit}`,
      message: metric(subscribers),
      color: 'red',
    }
  }
```

```js
  static get examples() {
    return [
      {
        title: 'Subreddit subscribers',
        namedParams: { subreddit: 'drums' },
        staticPreview: this.render({
          subreddit: 'drums',
          subscribers: 77000,
        }),
      },
    ]
  }
```

services/reddit/subreddit-subscribers.service.js
calebcartwright: 👍

JosepBernad:
Well, actually your proposal makes more sense and it's more consistent with the other social badges. I would stick to your idea!

services/reddit/subreddit-subscribers.tester.js
```diff
@@ -0,0 +1,18 @@
...
+
+t.create('subreddit-subscribers (valid subreddit)')
+  .get('/drums.json')
+  .expectBadge({
+    label: 'follow r/drums',
+    message: isMetric,
+  })
+
+t.create('subreddit-subscribers (invalid subreddit)')
+  .get('/not-a-real-subreddit.json')
```
calebcartwright: I'm not entirely sure why, but https://www.reddit.com/r/not-a-real-subreddit/about.json is returning an HTTP 200 response for me, with a response object:

```json
{
  "kind": "Listing",
  "data": {
    "modhash": "",
    "dist": 9,
    "children": [...],
    "after": null,
    "before": null
  }
}
```

This is causing this test to fail in the CI job, so you may want to try changing it to a different value

JosepBernad: Nice, done. Locally I can't run the tests, and I don't know if I'm doing something wrong.
I type 
`npm run tests:services -- --only=reddit`
and I get
`Error: Unknown services: reddit`

calebcartwright: you'll want to use the same value you used for the path so try

`npm run tests:services -- --only=subreddit `

calebcartwright: it looks like that http 200 response is pretty consistent 🤔 I was _finally_ able to get the typical 404 with `abadfbasdfadfadfadfadfasdf` as the subreddit 😄 

This does make me think we may need to tweak the service (schema and `handle` functions) a bit to better handle the subreddit not found scenario. Otherwise users are just going to get the `invalid` or `invalid response data` errors which aren't exactly indicative of the underlying issue in this case

JosepBernad: Let me research on this then! Thanks for the help!

services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+
+const { BaseJsonService } = require('..')
+const Joi = require('joi')
+const { nonNegativeInteger } = require('../validators')
+const { metric } = require('../text-formatters')
+
+const schema = Joi.object({
+  data: Joi.object({
+    subscribers: nonNegativeInteger,
```
calebcartwright: Let's try making this optional to account for the way the API response is representing not found scenarios.
```suggestion
    subscribers: optionalNonNegativeInteger,
```

services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+
+const { BaseJsonService } = require('..')
```
calebcartwright:
```suggestion
const { BaseJsonService, NotFound } = require('..')
```

services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,70 @@
...
+      color: 'red',
+      link: [`https://www.reddit.com/r/${subreddit}`],
+    }
+  }
+
+  async fetch({ subreddit }) {
+    return this._requestJson({
+      schema,
+      url: `https://www.reddit.com/r/${subreddit}/about.json`,
+    })
```
calebcartwright:
```suggestion
      errorMessages: {
        404: 'subreddit not found',
      },
    })
```

services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  async handle({ subreddit }) {
+    const json = await this.fetch({ subreddit })
+    return this.constructor.render({
+      subreddit,
+      subscribers: json.data.subscribers,
+    })
+  }
```
calebcartwright:
```js
  transform(json) {
    const subscribers = json.data.subscribers
    if (subscribers === undefined) {
      throw new NotFound({ prettyMessage: 'subreddit not found' })
    }
    return { subscribers }
  }

  async handle({ subreddit }) {
    const json = await this.fetch({ subreddit })
    const { subscribers } = this.transform(json)
      subreddit,
      subscribers,
    })
  }
```

services/reddit/subreddit-subscribers.tester.js
calebcartwright: I added several inline suggestions/code blocks that I think should handle this. May be worth giving that approach a try

JosepBernad:
Thanks, @calebcartwright for the suggestions. It worked well, so I created the appropriate tests for each case.

services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,82 @@
...
+          style: 'social',
+        },
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'reddit',
+    }
```
calebcartwright: This will stick the logo in there by default
```suggestion
      namedLogo: 'reddit',
    }
```

services/reddit/subreddit-subscribers.service.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  static render({ subreddit, subscribers }) {
+    return {
+      label: `follow r/${subreddit}`,
+      message: metric(subscribers),
+      color: 'red',
+      link: [`https://www.reddit.com/r/${subreddit}`],
```
calebcartwright: is there a link that can be used for users to perform the action (i.e. to actually subscribe)?

JosepBernad: Hi @calebcartwright. I've been researching this, and I think it's not possible. There's no available link to subscribe to the subreddit. 😔

calebcartwright: No worries, thanks for looking into it!

crazy-max:
Thanks @JosepBernad :rocket:

PR #3457, opened 2019-05-14
Fix code review guidelines link in Contributing guide

calebcartwright:
I just noticed this but looks like the original link (http://amyciavolino.com/assets/MindfulCommunicationInCodeReviews.pdf) in our Contributing guide for code review guidelines no longer works and redirects here: https://amy.tech/. That site then has a link to what I assume is the original doc here: https://kickstarter.engineering/a-guide-to-mindful-communication-in-code-reviews-48aab5282e5e

Updating the link accordingly in our Contributing doc


calebcartwright:
Okay it actually looks like there's a longer/different version here: https://amy.tech/assets/MindfulCommunicationInCodeReviews.pdf

I'm guessing _that_ is what we want to link to?

calebcartwright:
I agree, thanks @PyvesB!

PR #3460, opened 2019-05-16
lastCompletedBuild for [Jenkins] badges and better discoverability

PyvesB:
This pull request covers three small improvements to our Jenkins badges:
* switching the API endpoints from `lastBuild` to `lastCompletedBuild`. `lastBuild` may point to a build in progress or cancelled and for which no coverage nor test data is available. Consequently, the user experience with some of the Jenkins badges is currently quite poor, as the badges will break each time a new build is started. `lastCompletedBuild` will keep on pointing to the last completed build until the current one has finished, therefore guaranteeing continuity of the badges in these cases.
* adding the `jacoco` and `cobertura` keywords to the coverage badge. If someone is looking for badges for these specific plugins, they will be more likely to find them as they don't currently show up when searched with their names.
* making the tests more reliable.


PyvesB:
Some tests are timing out and increasing the timeouts is not good enough. Several of the daily Jenkins tests were already failing for some time now, I'll see if I can use a more reliable Jenkins instance for the tests later today.

PyvesB:
Okay, all tests have passed without any `timeout` overrides! Ready for review.

services/jenkins/jenkins-coverage.tester.js
```diff
@@ -9,19 +9,17 @@ const t = (module.exports = require('../tester').createServiceTester())
 // https://wiki.jenkins.io/pages/viewpage.action?pageId=58001258
 
 t.create('jacoco: job found')
-  .get('/jacoco/https/wso2.org/jenkins/view/All%20Builds/job/archetypes.json')
-  .timeout(10000)
+  .get('/jacoco/https/builds.apache.org/job/Derby-JaCoCo.json')
```
calebcartwright: I'm hesitant to change this one.  We've had recurring issues with the builds.apache.org instance in the past, and it kept timing out for me when I was working on #3337. Obviously the builds.apache.org instance is working at the moment but it's been way too flaky in the past.  It doesn't look like this particular test was failing/timing out, so I think we might be better off keeping this test with the wso2.org instance. 

the ubuntu and updates.jenkins-ci.org instances (including the ones below) are definitely problematic and need to be replaced, but I'd suggest we find a target jenkins instance other than builds.apache.org. 

What do you think?

PyvesB: Fair enough, I've removed all builds.apache.org as well a few of the ubuntu and updates.jenkins-ci.org. The remaining ones seem to be working fine for now. 👍

PyvesB:
Note for future reference: the examples for code coverage are always somewhat tricky to find on public servers, here are two additional jobs which use JaCoCo:
https://ci.eclipse.org/acceleo/job/acceleo-master/
https://ci.eclipse.org/emf-parsley/job/emf-parsley-gerrit/

calebcartwright:
> the examples for code coverage are always somewhat tricky to find on public servers,

so true! good thought on posting them

Issue #3478, opened 2019-05-23
jsDelivr badges/tests are flaky 


calebcartwright:
:clock11: **When did the problem start?**
intermittent timeout issues occur for `daily`, `weekly`, and `yearly` hits badges. Has been occurring for a while, typically once every 2-3 of our service test runs

:camera: **Live badge**
n/a - monthly badges are fine, others fail intermittently

:wrench: **Is the live badge working?**
see above

:link: **CircleCI link**
Multiple/recurring intermittently, a few examples:
https://circleci.com/gh/badges/daily-tests/347
https://circleci.com/gh/badges/daily-tests/345
https://circleci.com/gh/badges/daily-tests/343
https://circleci.com/gh/badges/daily-tests/341
https://circleci.com/gh/badges/daily-tests/339

:beetle: **Stack trace**

```
Error: Request timed out after 5000 ms
```

:bulb: **Possible solution**
The API on jsDelivr intermittently times out on those non-default (i.e. non-monthly) queries. Increasing the test time out does not help, as during these intermittent issues the API takes an exceptionally long time before returning with an HTTP 524 error.

I've opened https://github.com/jsdelivr/data.jsdelivr.com/issues/13 to see if they can help resolve

calebcartwright:
Upstream API has stabilized, tests have been running fine for 3+ days now so closing this out

PR #3479, opened 2019-05-23
increase [jsdelivr] test timeouts

calebcartwright:
Refs #3478 - increasing the timeout on these tests to 10s to see if that helps resolve the recurring test timeouts. There are time windows (00:00-1:00 UTC) when the underlying API is known to have issues but hopefully this will fix the service test timeouts 🤞


PyvesB:
Doesn't seemed to have helped much, several tests have failed again in the daily run. 😕

calebcartwright:
> Doesn't seemed to have helped much, several tests have failed again in the daily run

Yeah and I'm not too surprised. This helps us show the jsDelivr team that the window of the API being unresponsive is longer than the 00-01 UTC window they are currently aware of, since we are consistently seeing this same issue ~03:45 UTC

PR #3495, opened 2019-05-27
Stop calling variables "which"; affects [chocolatey codeclimate conda crates debug gem github mozillaobservatory nuget powershellgallery pub resharper vaadindirectory]

chris48s:
Woohoo - variable name bikeshedding :bike: :house: 

There are a various cases where we support multiple slight variations of what is roughly the same badge. Sometimes there is obvious descriptive terminology we can use for this e.g: `:period(dd|dw|dm)`, but often we end up using something like:

* `:which(stars|rating)`
* `:which(dt|dtv|dv)`
* `:which(d|dv)`
* `:which(grade|grade-score)`

I understand that the reason we aren't using a great var name is because we generally fall back to it when there isn't obvious meaningful terminology to use. That said, I think this is slightly more problematic now that we directly expose the variable names we use in URL patterns to users via the builder. Now that it has shifted from an internal implementation detail to part of the public interface, it seems worthwhile putting a bit more effort into giving these parameters a better name.

Having put a bit of thought into this, I propose that where there are multiple slighty different versions of the same badge and we're going to expose that to the end user, we should use the word `variant` if there is not a more meaningful option available, but I'm also open to other better ideas if anyone has any.

I've started off this PR by changing a couple of services, but happy to go through the rest of the codebase (about another 20 services) to finish this job once we establish a word we're happy with.

### DIFF
--- a/services/vaadin-directory/vaadin-directory-rating-count.service.js
+++ b/services/vaadin-directory/vaadin-directory-rating-count.service.js
@@ -12,7 +12,7 @@ module.exports = class VaadinDirectoryRatingCount extends BaseVaadinDirectorySer
   static get route() {
     return {
       base: 'vaadin-directory',
-      pattern: ':which(rc|rating-count)/:packageName',
+      pattern: ':alias(rc|rating-count)/:packageName',
     }
   }
 
@@ -39,7 +39,7 @@ module.exports = class VaadinDirectoryRatingCount extends BaseVaadinDirectorySer
     }
   }
 
-  async handle({ which, packageName }) {
+  async handle({ alias, packageName }) {
     const { ratingCount } = await this.fetch({ packageName })
     return this.constructor.render({ ratingCount })
   }
--- a/services/codeclimate/codeclimate-analysis.service.js
+++ b/services/codeclimate/codeclimate-analysis.service.js
@@ -40,7 +40,7 @@ const issueColorScale = colorScale(
   ['brightgreen', 'green', 'yellowgreen', 'yellow', 'red']
 )
 
-const whichMap = {
+const variantMap = {
   maintainability: {
     transform: data => ({
       maintainabilityLetter: data.attributes.ratings[0].letter,
@@ -96,7 +96,7 @@ module.exports = class CodeclimateAnalysis extends BaseJsonService {
     return {
       base: 'codeclimate',
       pattern:
-        ':which(maintainability|maintainability-percentage|tech-debt|issues)/:user/:repo',
+        ':variant(maintainability|maintainability-percentage|tech-debt|issues)/:user/:repo',
     }
   }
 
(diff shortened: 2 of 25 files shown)


paulmelnikow:
I'm happy with variant in the case where there's nothing more descriptive we can use. I'd prefer `interval` to `period` for time periods / intervals.

For `grade|grade-score` and `stars|rating` the data is the same and what we're choosing is the display format. `format` is overloaded to mean the image file format (svg vs gif) though `extension` might be a better name for that. I'd be okay using `format` in the pattern. 🤷‍♂

services/vaadin-directory/vaadin-directory-rating-count.service.js
```diff
@@ -12,7 +12,7 @@ module.exports = class VaadinDirectoryRatingCount extends BaseVaadinDirectorySer
   static get route() {
     return {
       base: 'vaadin-directory',
-      pattern: ':which(rc|rating-count)/:packageName',
+      pattern: ':alias(rc|rating-count)/:packageName',
```
paulmelnikow: These Vaadin Directory aliases should be changed to redirects, though that doesn't need to be handled in this PR.

chris48s: In a case like this, do you think there's a big advantage to defining an extra object which issues a redirect so that we can change the route from `'vaadin-directory/:alias(rc|rating-count)/:packageName'` to `'vaadin-directory/rc/:packageName'`? Given it doesn't make the route excessively complex, I'd tend towards leaving them as they are and avoiding the extra request.

paulmelnikow: We have a lot of API surface and I think it's nice to pare it down where we can, and particularly it seems nice to advertise a single canonical route rather than two options. Though I guess that is already the case in the published examples.

calebcartwright:
We should probably also update the documentation to cover these scenarios so contributors have some guidance as to the standards we're going to follow.

calebcartwright:
In the spirit of meaningful, and consistent, param names, should we also tackle the various param names used in badges that support GitHub/Bitbucket/GitLab/etc.? We've got a mix of `vcs`, `vcsType`, `vcsName`, `host`, and perhaps others I don't remember offhand.

chris48s:
Yes there are probably some other places where we've got some variations like that which can be cleaned up :+1: In general it would be nice to have some docs for common naming conventions (e.g: `/v` for version, `/l` for licence, `/dt`, `/dm`, `/dw`, `/dd` for downloads etc). Its one of the thing mentioned in [this](https://github.com/badges/shields/pull/3144) (long) discussion. Its on my mental list to extract tasks from that and write up issues/do something about them, but I haven't done it yet.

PR #3498, opened 2019-05-28
Run package tests on multiple node versions with nvm 

chris48s:
This is an alternative approach to #3428 using NVM to run the tests on multiple node versions in a single container based on the discussion in that thread.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,7 +138,7 @@ package_steps: &package_steps
         <<: *run_package_tests
         environment:
           mocha_reporter: mocha-junit-reporter
-          MOCHA_FILE: junit/gh-badges/results.xml
+          MOCHA_FILE: junit/gh-badges/v8/results.xml
           NODE_VERSION: v8
         name: Run package tests on Node 8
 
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -109,6 +101,74 @@ services_steps: &services_steps
...
+          mocha_reporter: mocha-junit-reporter
+          MOCHA_FILE: junit/gh-badges/results.xml
+          NODE_VERSION: v11
+        name: Run package tests on Node 11
+
+    - run:
+        <<: *run_package_tests
+        environment:
+          mocha_reporter: mocha-junit-reporter
+          MOCHA_FILE: junit/gh-badges/results.xml
```
paulmelnikow: Could you use a different directory for each test run? e.g. `junit/gh-badges/v12/results.xml`

.circleci/config.yml
```diff
@@ -109,6 +101,74 @@ services_steps: &services_steps
...
+          set +e
+          export NVM_DIR="/opt/circleci/.nvm"
+          [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+          nvm install v12
+          nvm use v12
+          npm install -g npm
+          npm ci
+
+    - run:
+        <<: *run_package_tests
```
paulmelnikow: Could you include a comment here referencing the version policy?

PR #3503, opened 2019-05-29
(attempt to) fix 'failing service test' issue template

chris48s:
Had a poke about in https://github.com/badges/shields/issues/templates/edit and I've got a hunch that the reason this isn't being correctly picked up as an issue template is because the colons in `:green_heart:` are stopping the header from being parsed correctly:

Screenshot at 2019-05-29 15-42-49

vs

Screenshot at 2019-05-29 15-43-00

This should probably fix it :crossed_fingers:


chris48s:
magic

Screenshot at 2019-05-29 16-30-46

PR #3505, opened 2019-05-29
Upgrade Joi

chris48s:
[Joi](https://www.npmjs.com/package/joi) is dead
long live [Joi](https://www.npmjs.com/package/@hapi/joi)


paulmelnikow:
Hmm, don't we also need to change `require('joi')` to `require('@hapi/joi')`?

It looks like we are still installing another copy of Joi (and that's probably what's being used now). Any idea where the `joi` requirement is being pulled in?

For IcedFrisby Joi is a peer dependency (which definitely should be switched over to `@hapi/joi`). Probably that needs to happen for us to actually that in our service tests.

chris48s:
Right, yes of course - good point. I assumed there was some backwards-compatibility magic going on there :sparkles: 

> Any idea where the joi requirement is being pulled in?

`icedfrisby` requires `joi": "*"` and then `start-server-and-test` (used in the end-to-end tests) transitively requires `"joi": "^13.0.0"`, so we're going to end up with 2 copies of Joi in the tree somewhere until both of those are updated to use `@hapi/joi`.

paulmelnikow:
It's an optional dep in icedfrisby so that's not what's triggering it being pulled in.

jeffbski/wait-on#37 is the issue for the transitive dep of start-server-and-test.

How about we fix the optional dep in IcedFrisby to use `@hapi/joi` and update all our require's to do the same. That way we'll still have two copies of the dependency, but at least our code will all be using the same version.

chris48s:
Having done the epic find & replace, it looks like we are also blocked on https://github.com/dszakallas/joi-extension-semver/issues/2 to get this working.

paulmelnikow:
I shipped [icedfrisby@2.0.0](https://www.npmjs.com/package/icedfrisby) 🎉 which includes this update.

chris48s:
Did a full test run. Some timeouts and transient failures, but nothing related to this

PR #3506, opened 2019-05-29
Fix BadgeFactory.create() docblock

chris48s:
After pushing a new release earlier, I realised that we've [changed the spec for the format object](https://github.com/badges/shields/blob/master/gh-badges/CHANGELOG.md#deprecations) but didn't update the docblock. This PR updates the docblock, but it reminds me of a related issue which I've addressed in PR #3507


paulmelnikow:
Think it'll be easy to miss this again since the doc block is in a different file from the definition. Maybe worth adding a comment with a reminder?

PR #3507, opened 2019-05-29
WIP: Validate input to BadgeFactory.create() with Joi

chris48s:
Refs #3506

This adds schema validation to the format object accepted by `BadgeFactory.create()` and throws `Joi.ValidationError` on invalid inputs (we could also consider wrapping this to avoid exposing an implementation detail here, if we wanted). I think this is a useful change to make because it will provide the user with more predictable behaviour and better error behaviour on invalid inputs, but it requires a major version bump because it is a backwards-incompatible change. For example, previously `bf.create({ text: ['build', 'passed'], format: 'some invalid format' })` would render a badge in SVG format, whereas now it will throw `Joi.ValidationError`.

Also, more subtly, `bf.create({})` would previously throw `TypeError: Cannot read property 'map' of undefined` whereas now it throws `Joi.ValidationError`.

For this reason, I've split this out from #3506 as I think its more useful for that fix to go into a 2.2.1 patch release instead of the next major.

Given this requires a major version bump, if we're happy with making this change, its probably worth putting this PR on hold or turning this PR into a 3.x branch so we can bundle it with the other breaking changes we've previously discussed and put on hold. I probably need to read some old PR threads to refresh my memory on those, but broadly they are:

* Remove deprecated properties
* Replace the `BadgeFactory` class with a single function
* Publish the package to a new name and mark the old one as dead

TODO: update the README to explain throwing `Joi.ValidationError`


paulmelnikow:
Really like the idea of adding validation here. Helpful error messages FTW 👍

I was wondering how much this would affect our bundle size, and also if it would prevent using gh-badges in the frontend.

Looks like the answer to bundle size is, it adds about 10% ([link](https://bundlephobia.com/result?p=gh-badges@2.1.0) and [link](https://bundlephobia.com/result?p=@hapi/joi@15.0.3)).

The other issue is that I'm not sure Joi bundles correctly in the browser. I didn't use it in the Shields frontend for that reason. See e.g. https://www.npmjs.com/package/joi-browser, though I'm not sure if that situation has changed.

We could use a different validation lib for this little bit of code. I'm pretty familiar with [rho-contracts](https://github.com/paulmelnikow/rho-contracts-fork) and would be happy to push that up to npm and use it (I can probably get access to the original fork on npmjs as well).

Though maybe [prop-types](https://www.npmjs.com/package/prop-types) would be a better alternative?

I feel like optimizing this for browsers is moderately important; could be useful in certain tools and use cases including possibly the Shields frontend (for e.g. an endpoint badge tester page).

chris48s:
That's fair. Its quite a heavy library given we only want to use it in one place. I'll have a think about the other options.

chris48s:
Probably not going to get a chance to look at this again until next week, but I had a quick look at doing this using prop-types. I think there are a few reasons why prop-types isn't a good fit:

prop-types doesn't have a mode which throws an exception or returns true/false. It only logs console warnings.
We could maybe work around that. There are a couple of projects which add that behaviour on top of prop-types e.g:
https://github.com/ratehub/check-prop-types
https://github.com/gajus/react-strict-prop-types
but also, prop-types has is set up to onl check in devevelpment and the checks are off in prod. This is hard-coded into the library with a bunch of conditionals on `process.env.NODE_ENV !== 'production'`. This makes it a bad choice to use as a general purpose validation library in this sort of case.

Having looked at rho-contracts that should do the job. Happy to go with that. The other thing I've had a quick look at which is probably worth considering is [v8n](https://github.com/imbrn/v8n). Its smaller than rho-contracts, plays nice in the browser and has quite a nice fluent Joi-style interface.

Given compact is desirable in this context, I'm leaning slightly towards v8n but I'm appy to go with rho-contracts if you have a strong preference for it or see it being useuful in other contexts.. Any thoughts before I press ahead?

paulmelnikow:
Makes sense.

v8n looks nice! May as well give it a shot.

One thing I'd be curious your though on is whether we might want to use it instead of Joi, in the long run.

chris48s:
My instinct is that Joi has a lot of more esoteric validators, a plugin ecosystem etc that are saving us time on the server-side where bundle size/browser compat isn't a big issue. Replacing with something more lightweight there would probably result in us having to reinvent the wheel a bit. I think the tradeoffs are a bit different in these two places.

paulmelnikow:
That seems reasonable.

chris48s:
closing in favour of #3586

PR #3508, opened 2019-05-29
Add [Mastodon] service

### LINKED ISSUE #3458: Mastodon follow
A badge, like the twitter follow, but for mastodon. Also shows how many mastodon followers there are. 

:link: **Data**
It comes from the open mastodon API. https://docs.joinmastodon.org/api/libraries/

:microphone: **Motivation**

I do updates for a project of mine on mastodon, this badge would allow ppl to follow the account from the readme.

Prouser123:
Closes #3458.

The API requires a user ID to get an account instead of their username.
I found the ID in a `<link>` tag in the header of a user's profile page.

Example: `<link href="{domain}/api/salmon/2214" rel="salmon">` (2214 is the user ID)

We should either add this to the documentation for the badge or find another solution to get the user ID.

Also posted on discord.

### DIFF
--- a/services/mastodon/mastodon-follow.service.js
+++ b/services/mastodon/mastodon-follow.service.js
@@ -3,11 +3,11 @@
 const Joi = require('joi')
 const { BaseJsonService, NotFound } = require('..')
 const { metric } = require('../text-formatters')
-const { optionalUrl } = require('../validators')
+const { optionalUrl, nonNegativeInteger } = require('../validators')
 
 const schema = Joi.object({
   username: Joi.string().required(),
-  followers_count: Joi.number().required(),
+  followers_count: nonNegativeInteger,
 })
 
 const queryParamSchema = Joi.object({
--- a/services/mastodon/mastodon-follow.service.js
+++ b/services/mastodon/mastodon-follow.service.js
@@ -1,0 +1,82 @@
+'use strict'
+
+const Joi = require('joi')
+const { BaseJsonService, NotFound } = require('..')
+const { metric } = require('../text-formatters')
+const { optionalUrl } = require('../validators')
+
+const schema = Joi.object({
+  username: Joi.string().required(),
+  followers_count: Joi.number().required(),
+})
+
+const queryParamSchema = Joi.object({
+  domain: optionalUrl,
+}).required()
+
+module.exports = class MastodonFollow extends BaseJsonService {
+  static get category() {
+    return 'social'
+  }
+
+  static get route() {
+    return {
+      base: 'mastodon/follow',
+      pattern: ':id',
+      queryParamSchema,
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Mastodon Follow',
+        namedParams: {
+          id: '26471',
+        },
+        queryParams: { domain: 'https://mastodon.social' },
+        staticPreview: {
+          label: 'Follow',
+          message: '862',
+          style: 'social',
+        },
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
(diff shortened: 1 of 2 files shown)


calebcartwright:
Review app @ https://shields-staging-pr-3508.herokuapp.com/

services/mastodon/mastodon-follow.service.js
```diff
@@ -0,0 +1,82 @@
+'use strict'
+
+const Joi = require('joi')
+const { BaseJsonService, NotFound } = require('..')
+const { metric } = require('../text-formatters')
+const { optionalUrl } = require('../validators')
+
+const schema = Joi.object({
+  username: Joi.string().required(),
+  followers_count: Joi.number().required(),
```
calebcartwright: Am I correct in assuming that this will always be a non-negative integer value? If so we should reuse that respective helper validator

calebcartwright:
> The API requires a user ID to get an account instead of their username.
I found the ID in a <link> tag in the header of a user's profile page.

> Example: <link href="{domain}/api/salmon/2214" rel="salmon"> (2214 is the user ID)

> We should either add this to the documentation for the badge or find another solution to get the user ID.

See also: https://github.com/tootsuite/mastodon/issues/4588#issuecomment-348667202

I think my recommendation would be to add documentation to the badge that lists both ways for folks to figure out their user id. I know it would be easier for user's to use their username instead of id, but I'm disinclined to go down the path of us accepting user names and then making an additional API call on our end to figure out the user id on their behalf.

Prouser123:
From that issue, it looks like we can do `{domain}.well-known/webfinger?resource=acct:{user}@{hostname}` and parse it to get the user id.

If we don't want to add it to the badge, we could make a little web tool to get it.

Prouser123:
https://prouser123.me/misc/mastodon-userid-lookup.html
Proof of concept lookup tool

services/mastodon/mastodon-follow.service.js
```diff
@@ -0,0 +1,82 @@
...
+        namedParams: {
+          id: '26471',
+        },
+        queryParams: { domain: 'https://mastodon.social' },
+        staticPreview: {
+          label: 'Follow',
+          message: '862',
+          style: 'social',
+        },
+      },
```
calebcartwright: I think the only thing left is to add documentation to the examples here. I'd suggest listing your lookup tool as the first option, and then also list the other two (the lookup process and the html tag) as fallback options folks can use.

calebcartwright:
Looking good! Once documentation has been added I think this will be ready to merge

Prouser123:
Looks like we've got an integration timeout.

Prouser123:
Looks like something in that update (maybe #3510???) has interfered with the modal scrolling.

Edit: Yep, reverting that commit solves the issue.

image

calebcartwright:
> Edit: Yep, reverting that commit solves the issue.

Can you elaborate a bit? 

It looked to me like there was a spurious integration test failure on your last commit, and that after merging in the latest from master into this branch that test (unsurprisingly) is passing again. I don't think either of those was related to any other commits/changes, just a random test failure.

Are you saying you've discovered a new bug that you think was introduced in #3510?

Prouser123:
After the integration timeout, I merged the latest changes in the branch to get it to re-run.

After that, I went on the staging site and found that bug.
I think it was introduced in #3510 as after I reverted it on my local machine, it was fixed.

calebcartwright:
Gotcha, thanks for the extra detail. If you wouldn't mind, could you create a new bug that details the issue you've found? That way it can be tracked and tackled in a follow on

Issue #2806, opened 2019-01-18
Popup menu gets hidden behind rest of modal content


paulmelnikow:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)

:beetle: **Description**

After clicking the drop-down menu to copy badge markup, the markup options appear but are partly obscured.

<img width="1067" alt="screen shot 2019-01-18 at 6 11 24 am" src="https://user-images.githubusercontent.com/1487036/51383885-a9803f80-1ae8-11e9-9a1b-b13a8999801a.png">

It's easily workaroundable, just not quite as nice as if the menu were fully visible.

:bulb: **Possible Solution**

Probably something involving z-axis?

paulmelnikow:
Fix had to be reverted; see #3513 and #3517.

paulmelnikow:
As I mentioned at https://github.com/badges/shields/pull/3510#issuecomment-498045780:

> I'm guessing to deal with this specific issue we'd need to find out why the `menuPlacement` prop is sometimes not working as intended and fix it upstream. Feel free to dig further if you'd like… or to find something else to tackle!

chris48s:
We have just launched a completely redesigned frontend https://shields.io/
Closing this issue as it relates to the previous frontend

PR #3510, opened 2019-05-30
Make Dropdown Visible When Overflows

mackmmiller:
This PR addresses the dropdown visibility issue described in https://github.com/badges/shields/issues/2806. 

Screen Shot 2019-05-29 at 9 53 47 PM

### DIFF
--- a/frontend/components/markup-modal/index.js
+++ b/frontend/components/markup-modal/index.js
@@ -19,6 +19,7 @@ export default function MarkupModal({ example, baseUrl, onRequestClose }) {
       contentLabel="Example Modal"
       isOpen={isOpen}
       onRequestClose={onRequestClose}
+      style={{ content: { overflow: 'visible' } }}
     >
       {isOpen && (
         <ContentContainer>


paulmelnikow:
Hi! Thanks for picking this up!

Have you been able to reproduce the problem (i.e. before the fix)?

When I try it now I'm seeing the popup is not obscured as in the screenshot in #2806, but rather that it opens in the opposite direction. 

Screen Shot 2019-05-29 at 11 09 32 PM

I wonder if a subsequent release of react-select fixed it?

mackmmiller:
@paulmelnikow, I have been able to reproduce the problem!

The following image is a screenshot from the current live site:
Screen Shot 2019-05-30 at 7 22 09 PM

And this is an image of the site, at the same size, with the changes:
Screen Shot 2019-05-30 at 7 24 20 PM

The behavior you've described seems to be controlled by the `menuPlacement` prop on the [`<MarkupFormatSelect/>`](https://github.com/badges/shields/blob/98f380b2549fdd3c765cdf96ad453584c606c4f8/frontend/components/customizer/request-markup-button.js#L104) component. According to the [documentation](https://react-select.com/props), it should move the dropdown above the button when there isn't enough room, though clearly that is only working some of the time.

paulmelnikow:
Ah, gotcha. Thanks for digging into that!

calebcartwright:
Nice!

paulmelnikow:
P.S. Thanks for jumping in @mackmmiller! We've got lots of things we'd like to do with the frontend and would welcome your help!

paulmelnikow:
Hey @mackmmiller, just a heads up that we had to revert this (see #3513).

I'm guessing to deal with this specific issue we'd need to find out why the `menuPlacement` is sometimes not working as intended and fix it upstream. Feel free to dig further if you'd like… or to find something else to tackle!

PR #3511, opened 2019-05-30
improve logo escaping

chris48s:
Also prepares a 2.2.1 release for NPM

### DIFF
--- a/lib/logos.js
+++ b/lib/logos.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const Joi = require('joi')
 const { toSvgColor } = require('../gh-badges/lib/color')
 const coalesce = require('../core/base-service/coalesce')
 const { svg2base64 } = require('./svg-helpers')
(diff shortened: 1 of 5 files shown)


lib/logos.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+const Joi = require('joi')
```
chris48s: This will want updating to `const Joi = require('@hapi/joi')` in the context of #3505 I'll sort it out in one or the other depending on which order these get merged in.

lib/logos.js
```diff
@@ -31,7 +32,12 @@ function prependPrefix(s, prefix) {
 }
 
 function isDataUrl(s) {
-  return s !== undefined && /^(data:)([^;]+);([^,]+),(.+)$/.test(s)
+  try {
+    Joi.assert(s, Joi.string().dataUri())
```
chris48s: Turns out we don't need wheel 2.0 here :) https://github.com/hapijs/joi/blob/master/API.md#stringdataurioptions

paulmelnikow:
Cheers, thanks for picking this up!

platan:
@chris48s thank you!

Issue #3529, opened 2019-06-05
Libraries.io: Bad processing of API Response


corebreaker:
Are you experiencing an issue with...

- [X] [shields.io](https://shields.io/#/)
- [X] My own instance
- [X] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

Libraries.io dependencies on a repository reports no problems but the badge can show the good state.

:link: **Link to the badge**

[![Dependency status](https://img.shields.io/librariesio/github/corebreaker/goerrors.svg?style=plastic)](https://img.shields.io/librariesio/github/corebreaker/goerrors.svg)

:bulb: **Possible Solution**

I tracked the Libraries.io API response: 
```
{
    ...
    dependencies:
        [ { project_name: 'github.com/google/uuid',
            name: 'github.com/google/uuid',
            platform: 'go',
            requirements: 'v1.1.1',
            latest_stable: null,
            latest: '0.2',
            deprecated: null,
            outdated: null,
            filepath: 'go.sum',
            kind: 'runtime',
            normalized_licenses: [Array] }]
}
```

The `deprecated` and `outdated` field can be null, not only boolean as defined in Joi schema.
It posted a PR to add a `allowNull`.

PR #3530, opened 2019-06-05
Allow nulls for boolean fields on Libraries.io APIs

corebreaker:
Fix the issue #3529

### DIFF
--- a/services/librariesio/librariesio-dependencies.service.js
+++ b/services/librariesio/librariesio-dependencies.service.js
@@ -11,8 +11,12 @@ const schema = Joi.object({
   dependencies: Joi.array()
     .items(
       Joi.object({
-        deprecated: Joi.boolean().required(),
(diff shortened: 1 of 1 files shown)


calebcartwright:
Hi @corebreaker, thanks for reporting and opening a PR with a fix! I updated the PR title so that the tests associated with the Libraries.io service would be run, and as you can see the changes added here are causing the tests to fail. 

Can you take a look and update/fix the tests accordingly?

calebcartwright:
Actually looks like it was just a timeout. I'll re-start the tests and see if they pass this time

calebcartwright:
timeout errors are spurious, tests pass locally (with increased timeout) so going to disable the service test runs accordingly

corebreaker:
Hi @calebcartwright . Good news ! Thank you for the merge :+1: .

PR #3531, opened 2019-06-06
add gitee support to [JitpackVersion]

okou19900722:
add gitee

### DIFF
--- a/services/jitpack/jitpack-version.service.js
+++ b/services/jitpack/jitpack-version.service.js
@@ -19,7 +19,7 @@ module.exports = class JitPackVersion extends BaseJsonService {
   static get route() {
     return {
       base: 'jitpack/v',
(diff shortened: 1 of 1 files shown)


okou19900722:
jitpack has already supported gitee. https://jitpack.io/docs/#other-git-hosts

calebcartwright:
Thanks @okou19900722 - do you have a gitee repo/project that we can use to test?

calebcartwright:
Review app @ https://shields-staging-pr-3531.herokuapp.com/

PyvesB:
It might be a good idea to add tests for the different Git hosts: we'll be alerted if support for one of them is removed or if there's an issue with the API for that particular host. Probably not a blocker for this PR though. 👍

calebcartwright:
I wasn't proposing adding an automated test. I'm not familiar with gitee, so I was just hoping to get a project name that we could use to test with in the review app

PR #3533, opened 2019-06-06
restore [JitpackDownloads] badge

### LINKED ISSUE #3532: re-enable JitpackDownloads badge
Refs #3493 and #3394 

Seems there is a new/different API that provides the download data for Jitpack, and from some testing I've done it seems to be both fast and stable enough for our use (the stats api we were using before is still down). 

I believe we can re-enable this badge using the new api

calebcartwright:
Closes #3532, also refs #3493, #3493,  and #3220 

Basically just reverts #3493 which temporarily deprecated this badge due to issues with upstream API. Also switches to a new/different API to get download data which is more stable


services/jitpack/jitpack-downloads.service.js
```diff
@@ -1,14 +1,83 @@
...
+module.exports = class JitpackDownloads extends BaseJsonService {
+  static get category() {
+    return 'downloads'
+  }
+
+  static get route() {
+    return {
+      base: 'jitpack',
+      pattern:
+        ':variant(dw|dm)/:vcs(github|bitbucket|gitlab|gitee)/:user/:repo',
```
calebcartwright: @chris48s & @paulmelnikow, did we land on using `variant` for time period/intervals in #3495? The original (pre-#3493 deprecation) version of this badge used `period`  but changed this to `variant` based on #3495

services/jitpack/jitpack-downloads.service.js
```diff
@@ -1,14 +1,83 @@
...
+      message: `${metric(downloads)}${variantMap[variant].suffix}`,
+      color: downloadCount(downloads),
+    }
+  }
+
+  async fetch({ vcs, user, repo }) {
+    return this._requestJson({
+      schema,
+      url: `https://jitpack.io/api/downloads/com.${vcs}.${user}/${repo}`,
+      errorMessages: { 401: 'project not found or private' },
```
calebcartwright: One downside of this new API is that it doesn't return a 404/401 response when projects are not found like the previous (albeit unusable API) did, this one just returns a value of `0` for downloads. Keeping this in here though in case this new API adds 401/404 responses in the future

services/jitpack/jitpack-downloads.tester.js
```diff
@@ -1,23 +1,22 @@
...
+const { isMetricOverTimePeriod } = require('../test-validators')
 
-const t = (module.exports = new ServiceTester({
-  id: 'JitpackDownloads',
-  title: 'JitpackDownloads',
-  pathPrefix: '/jitpack',
-}))
+t.create('weekly (github)')
+  .get('/dw/github/jitpack/maven-simple.json')
+  .timeout(10000)
```
calebcartwright: In my local testing, this API and these tests are all returning well within 500ms. However, I thought it best to include an increased timeout given the historical timeout issues we've had with the Jitpack service. Happy to remove them if anyone is opposed though

services/jitpack/jitpack-downloads.service.js
chris48s: I think where it is specifically a timeframe like `dm|dw`, we've got a mix of `interval` and `period` in various places, but I think Paul is right that `interval` is more descriptive and we should standardise on that.

calebcartwright:
@erayerdin just an FYI that the Jitpack Downloads badges are back

PR #3534, opened 2019-06-07
Improvement-3509: Add 1 place rounding to 7 digit numbers, run all service tests [*****]

### LINKED ISSUE #3509: Display decimals when over 1 million
:clipboard: **Description**

Once downloads hit 1 million, the resolution of downloads is simply an integer, e.g. 1M. Pretty big milestone for some projects, but getting to display 2M might take just as long if not longer. What about showing a single decimal digit, e.g. 1.2M which still fits within limits. Perhaps this drops off after 10M. Same would go for billions, etc.

Orlandster:
fixes #3509


calebcartwright:
Thanks @Orlandster! Looks like we'll also need to update the regex used in the tests to account for the updated formatting. The validator regex used is in the tests is [here](https://github.com/badges/shields/blob/master/services/test-validators.js#L69) for reference. Below is a link to the test run with this change and a snippet from one of the failing tests

https://circleci.com/gh/badges/shields/55793
```
ValidationError: message mismatch: "value" with value "6.2M" fails to match the required pattern: /^[1-9][0-9]*[kMGTPEZY]?$/
```

services/text-formatters.spec.js
```diff
@@ -44,6 +44,11 @@ describe('Text formatters', function() {
     given(1000).expect('1k')
     given(999499).expect('999k')
     given(999500).expect('1M')
+    given(1049999).expect('1M')
+    given(1050000).expect('1.1M')
+    given(1549999).expect('1.5M')
+    given(10499999).expect('10M')
+    given(10500000).expect('11M')
```
paulmelnikow: What happens if you call e.g. `metric(1234567.89')`? I think what we want is 1.2M though I'm not sure it will currently do the right thing.

services/text-formatters.js
```diff
@@ -56,7 +56,10 @@ function metric(n) {
   for (let i = metricPrefix.length - 1; i >= 0; i--) {
     const limit = metricPower[i]
     if (n >= limit) {
-      n = Math.round(n / limit)
+      const digits = n.toString().length
+      // if number has 7 digits round to 1 decimal place, otherwise to no decimal place
+      n =
+        digits === 7 ? Math.round((n / limit) * 10) / 10 : Math.round(n / limit)
```
paulmelnikow: Would [`toFixed`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) accomplish this more clearly?

PyvesB:
@Orlandster it would be nice to get this landed, do you have some time to look into it again? 😉

PyvesB:
It's been over a month at this point, I'll pick this up in the coming days if there aren't any updates.

PR #3540, opened 2019-06-09
Update CI config to cache/skip Cypress binary

calebcartwright:
Updates Circle config to include the Cypress binary in the cache on the e2e job that runs the functional tests with Cypress, and skips the Cypress binary install on the other jobs (which don't run Cypress at all)

@platan - I'm not very familiar with Cypress so if you get a chance let me know if this looks and sounds reasonable

https://docs.cypress.io/guides/getting-started/installing-cypress.html#Skipping-installation
https://docs.cypress.io/guides/getting-started/installing-cypress.html#Binary-cache


platan:
Thanks for this PR @calebcartwright!

Skipping binary installation works great, in logs we can see:
```
Note: Skipping binary installation: Environment variable CYPRESS_INSTALL_BINARY = 0.
```

Let's compare execution time of 'main' job:
other PRs:
https://github.com/badges/shields/pull/3553: 2:19
https://github.com/badges/shields/pull/3551: 1:45
this PR: 1:09

Now let's move to caching. I'm not sure if it works:

In the first build after merging master to this PR: https://circleci.com/gh/badges/shields/56766 we have in "Saving Cache":
```
Creating cache archive...
Uploading cache archive...
Stored Cache to v2-cypress-dependencies-sSI6ZkI2pdZz4LInclgNsyMj8nAaZZpGaFzcrwSaslc=
  * /root/.cache/Cypress
```
But the second build (a manual rerun) https://circleci.com/gh/badges/shields/56769 is not restoring the cache in my opinion, in "Restoring Cache" we have:
```
Found a cache from build 56595 at v2-dependencies-sSI6ZkI2pdZz4LInclgNsyMj8nAaZZpGaFzcrwSaslc=
Size: 67 MB
Cached paths:
  * /home/circleci/project/node_modules

Downloading cache archive...
Validating cache...

Unarchiving cache...
```
and "Saving Cache" says:
```
Skipping cache generation, cache already exists for key: v2-cypress-dependencies-sSI6ZkI2pdZz4LInclgNsyMj8nAaZZpGaFzcrwSaslc=
Found one created at 2019-06-10 19:56:27 +0000 UTC
```
I don't know how to fix it. 

Moreover at https://docs.cypress.io/guides/guides/continuous-integration.html#We-recommend-users
we have (but I don't know if it is related to our problem):
> - Do not cache node_modules across builds. This bypasses more intelligent caching packaged with npm or yarn, and can cause issues with Cypress not downloading the Cypress binary on npm install.

> - If you are using npm install in your build process, consider switching to npm ci and caching the ~/.npm directory for a faster and more reliable build.

I will try to look at this again tomorrow.

calebcartwright:
Thanks @platan, and good spot on the cache. I believe the saving of the node_modules directory in cache is saved elsewhere, so i've updated the e2e job with an explicit cache restore and save step for the cypress binary in the e2e job

🤞

PR #3559, opened 2019-06-11
Test for "Build(deps-dev): bump danger from 7.1.4 to 8.0.0"

platan:
This is a test for https://github.com/badges/shields/pull/3553. We should get a warning "This PR modified the server but none of its tests. That's okay so long as it's refactoring existing code.". This PR should be closed if danger works properly.

### DIFF
--- a/core/server/server.js
+++ b/core/server/server.js
@@ -300,3 +300,4 @@ module.exports = class Server {
     }
   }
 }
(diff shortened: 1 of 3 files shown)


platan:
Danger is diabled for branches `/dependabot\/.*/`. I will create a new PR.

PR #3563, opened 2019-06-12
fix [Bitbucket] Issues badges

### LINKED ISSUE #3561: Bitbucket issues badge uses the old 1.0 API

- [ ] My own instance

:beetle: **Description**

The Bitbucket issues badge uses the old 1.0 API and no longer works. The pull requests badge can also be affected.

:link: **Link to the badge**

https://img.shields.io/bitbucket/issues/vx68k/netbeans-bitbucket-plugin.svg

:bulb: **Possible Solution**

Use the new 2.0 API instead. I will try to make a pull request.

calebcartwright:
Fixes #3561 

Atlassian has deprecated the v1 APIs that we were using for the Bitbucket Issues badges
https://developer.atlassian.com/cloud/bitbucket/deprecation-notice-v1-apis/ 

This switches to the v2 API.

https://developer.atlassian.com/bitbucket/api/2/reference/resource/
https://developer.atlassian.com/bitbucket/api/2/reference/meta/filtering#query-issues

### DIFF
--- a/services/bitbucket/bitbucket-issues.service.js
+++ b/services/bitbucket/bitbucket-issues.service.js
@@ -6,7 +6,7 @@ const { BaseJsonService } = require('..')
 const { nonNegativeInteger } = require('../validators')
 
 const bitbucketIssuesSchema = Joi.object({
-  count: nonNegativeInteger,
+  size: nonNegativeInteger,
 }).required()
 
 function issueClassGenerator(raw) {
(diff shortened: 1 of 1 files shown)


chris48s:
This is fine to merge as it is.

As a related issue, on the pull requests badge, we support both server and cloud versions of BB (for some reason, we only support cloud on the other BB badges, which is maybe an issue in itself) and we're assuming that we use the 1.0 API version for self-hosted users:

https://github.com/badges/shields/blob/b6fe3a3f9c06240541a2ed959e341ae813d96640/services/bitbucket/bitbucket-pull-request.service.js#L88-L90

I don't really know what the state of 1.0 vs 2.0 API support is like in the self-hosted space, but we'll probably need to deal with users who host their own BB instance which do/don't support one or other API version at some point too.

calebcartwright:
Thanks @chris48s! 

> for some reason, we only support cloud on the other BB badges, which is maybe an issue in itself

We only added support for BBs (Bitbucket Server, the self-hosted version) PR badges in the relatively recent past because a contributor was motivated and kind enough to send us a PR that added support. Agreed it would be great to add BBs support for the other badges too, although I'm not sure if all the badge types we support for BB cloud all applicable. For example I've never seen the Issues construct in BBs, BBs has always been linked with a Jira instance and issues were maintained in Jira.

> and we're assuming that we use the 1.0 API version for self-hosted users

Maintaining support for a breadth of versions of self-hosted tools can get tricky, but we do it for other tools (Sonar comes to mind) so perhaps those could serve as reference points

chris48s:
> although I'm not sure if all the badge types we support for BB cloud are applicable with BBs. For example I've never seen the Issues construct in BBs, BBs has always been linked with a Jira instance and issues were maintained in Jira

That's a good point which I hadn't considered.

> Maintaining support for a breadth of versions of self-hosted tools can get tricky, but we do it for other tools (Sonar comes to mind) so perhaps those could serve as reference points

Lets ignore it for now, but my prediction is that at some point we'll have some users who are on a BBS version that drops support for the v1 API and some who haven't upgraded yet. If/when we see users reporting an issue at least then we'll have a concrete example rather than speculation.

kazssym:
It would be better to use the `https://api.bitbucket.org/2.0` prefix as specified in the [API documentation](https://confluence.atlassian.com/bitbucket/rest-apis-222724129.html).

PR #3573, opened 2019-06-16
Add default labels to github issue templates

calebcartwright:


.github/ISSUE_TEMPLATE/1_Bug_report.md
```diff
@@ -1,6 +1,7 @@
 ---
 name: 🐛 Bug Report
 about: Report errors and problems
+labels: 'question'
```
PyvesB: Should this actually be `bug`?

calebcartwright: That's what my initial thought was too, but I ended up changing it to `question` based on the descriptions we've applied to the two labels. In the description for question we have `.., unconfirmed bugs, ...` whereas it feels like we use the `bug` label for a confirmed bug. 

Maybe we should a new label for `unconfirmed-bug` and remove that part of the description from the `question` label?

PyvesB: You're right, it makes sense from the description's point of view! We've got enough labels as it is in my opinion, I suggest we leave it like this!

PR #3574, opened 2019-06-16
Unused xml2js dependency

PyvesB:
Something I noticed whilst digging around with recent discussions around our XML handling.


PyvesB:
Thanks @platan! Will have a quick go at checking the other dependencies as well.

paulmelnikow:
Nice one, thanks for catching this!

PR #3586, opened 2019-06-18
WIP: Validate input to BadgeFactory.create() with rho-contracts

chris48s:
This replaces #3507

In the end I went with rho-contracts over v8n. v8n looked quite nice on the surface, but once I got into it, I found when it throws an exception, it doesn't provide the user with very useful information about what value failed which rule. That was one of the main things I wanted to achieve here and rho-contracts makes a better job of it.

TODO: update the README

### DIFF
--- a/gh-badges/lib/index.js
+++ b/gh-badges/lib/index.js
@@ -56,6 +56,7 @@ class BadgeFactory {
       if (e instanceof c.ContractError) {
         const wrappedError = new ValidationError()
         Object.assign(wrappedError, e)
+        wrappedError.name = 'ValidationError'
         throw wrappedError
       }
       throw e
--- a/gh-badges/lib/index.js
+++ b/gh-badges/lib/index.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const c = require('rho-contracts-fork')
+const c = require('@rho-contracts/rho-contracts')
 const makeBadge = require('./make-badge')
 const schema = c.object({
   text: c.tuple(c.string, c.string).strict(),
--- a/gh-badges/lib/index.js
+++ b/gh-badges/lib/index.js
@@ -1,6 +1,29 @@
 'use strict'
 
+const c = require('rho-contracts-fork')
 const makeBadge = require('./make-badge')
+const schema = c.object({
+  text: c.tuple(c.string, c.string).strict(),
+  labelColor: c.optional(c.string),
+  color: c.optional(c.string),
+  colorA: c.optional(c.string),
+  colorscheme: c.optional(c.string),
+  colorB: c.optional(c.string),
+  format: c.optional(c.oneOf('svg', 'json')),
+  template: c.optional(
+    c.oneOf(
+      'plastic',
+      'flat',
+      'flat-square',
+      'for-the-badge',
+      'popout',
+      'popout-square',
+      'social'
+    )
+  ),
+})
+
+class ValidationError extends Error {}
 
 class BadgeFactory {
   constructor(options) {
@@ -27,10 +50,22 @@ class BadgeFactory {
    * @see https://github.com/badges/shields/tree/master/gh-badges/README.md
    */
   create(format) {
+    try {
+      schema.check(format)
+    } catch (e) {
+      if (e instanceof c.ContractError) {
+        const wrappedError = new ValidationError()
+        Object.assign(wrappedError, e)
+        throw wrappedError
+      }
+      throw e
+    }
+
     return makeBadge(format)
   }
 }
 
 module.exports = {
   BadgeFactory,
+  ValidationError,
 }
--- a/gh-badges/lib/index.spec.js
+++ b/gh-badges/lib/index.spec.js
@@ -30,16 +30,31 @@ describe('BadgeFactory class', function() {
   })
 
   it('should throw a ValidationError with invalid inputs', function() {
-    expect(() => bf.create({})).to.throw(ValidationError)
-    expect(() => bf.create({ text: ['build'] })).to.throw(ValidationError)
+    expect(() => bf.create({})).to.throw(
+      ValidationError,
+      'Field `text` required, got {}'
+    )
+    expect(() => bf.create({ text: ['build'] })).to.throw(
+      ValidationError,
+      "Expected tuple of exactly size 2, but got [ 'build' ]\nfor the field `text` of the object\nThe full value being checked was:\n{ text: [ 'build' ] }"
+    )
     expect(() =>
       bf.create({ text: ['build', 'passed', 'something else'] })
-    ).to.throw(ValidationError)
+    ).to.throw(
+      ValidationError,
+      "Expected tuple of exactly size 2, but got [ 'build', 'passed', 'something else' ]\nfor the field `text` of the object\nThe full value being checked was:\n{ text: [ 'build', 'passed', 'something else' ] }"
+    )
     expect(() =>
       bf.create({ text: ['build', 'passed'], format: 'png' })
-    ).to.throw(ValidationError)
+    ).to.throw(
+      ValidationError,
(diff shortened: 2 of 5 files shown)


gh-badges/lib/index.js
```diff
@@ -27,10 +50,22 @@ class BadgeFactory {
    * @see https://github.com/badges/shields/tree/master/gh-badges/README.md
    */
   create(format) {
+    try {
+      schema.check(format)
+    } catch (e) {
+      if (e instanceof c.ContractError) {
+        const wrappedError = new ValidationError()
+        Object.assign(wrappedError, e)
+        throw wrappedError
```
chris48s: Instead of directly throwing `ContractError` here, I've wrapped it in a `ValidationError`.

My rationale here is that if we wanted to change the implementation of the validation we could do that without it being a breaking change, but I'm in 2 minds about whether its actually a useful abstraction if we just expose all the fields of the `ContractError` anyway. Thoughts?

paulmelnikow: I've no strong feelings either way.

gh-badges/lib/index.spec.js
```diff
@@ -16,5 +21,25 @@ describe('BadgeFactory class', function() {
...
+    expect(() => bf.create({ text: ['build'] })).to.throw(ValidationError)
+    expect(() =>
+      bf.create({ text: ['build', 'passed', 'something else'] })
+    ).to.throw(ValidationError)
+    expect(() =>
+      bf.create({ text: ['build', 'passed'], format: 'png' })
+    ).to.throw(ValidationError)
+    expect(() =>
+      bf.create({ text: ['build', 'passed'], template: 'something else' })
+    ).to.throw(ValidationError)
```
paulmelnikow: Is it worth asserting something about the error message?

gh-badges/lib/index.js
```diff
@@ -1,6 +1,29 @@
...
+const c = require('rho-contracts-fork')
 const makeBadge = require('./make-badge')
+const schema = c.object({
+  text: c.tuple(c.string, c.string).strict(),
+  labelColor: c.optional(c.string),
+  color: c.optional(c.string),
+  colorA: c.optional(c.string),
+  colorscheme: c.optional(c.string),
+  colorB: c.optional(c.string),
+  format: c.optional(c.oneOf('svg', 'json')),
```
paulmelnikow: I thought this accepted raster formats too. Am I wrong about that? Do we have an API for making raster formats or are they only available in the CLI?

chris48s: It doesn't. At the moment, if you pass `format: 'gif'` (or `'png'` or `'jpg'`), you just get a SVG badge back:

```js
> const { BadgeFactory } = require('./gh-badges')
undefined

> const bf = new BadgeFactory()
undefined

> bf.create({text: ['foo', 'bar']})
'<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="56" height="20"><linearGradient id="b" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="a"><rect width="56" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#a)"><path fill="#555" d="M0 0h27v20H0z"/><path fill="#4c1" d="M27 0h29v20H27z"/><path fill="url(#b)" d="M0 0h56v20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="110"> <text x="145" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="170">foo</text><text x="145" y="140" transform="scale(.1)" textLength="170">foo</text><text x="405" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="190">bar</text><text x="405" y="140" transform="scale(.1)" textLength="190">bar</text></g> </svg>'

> bf.create({text: ['foo', 'bar'], format: 'json'})
'{"label":"foo","message":"bar","link":["",""],"name":"foo","value":"bar"}'

> bf.create({text: ['foo', 'bar'], format: 'png'})
'<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="56" height="20"><linearGradient id="b" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><clipPath id="a"><rect width="56" height="20" rx="3" fill="#fff"/></clipPath><g clip-path="url(#a)"><path fill="#555" d="M0 0h27v20H0z"/><path fill="#4c1" d="M27 0h29v20H27z"/><path fill="url(#b)" d="M0 0h56v20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="110"> <text x="145" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="170">foo</text><text x="145" y="140" transform="scale(.1)" textLength="170">foo</text><text x="405" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="190">bar</text><text x="405" y="140" transform="scale(.1)" textLength="190">bar</text></g> </svg>'
```

The code to do raster formats is only in the CLI wrapper:

https://github.com/badges/shields/blob/c312eae3e3cfa28d8442b9e691ff063a85b3e3e5/gh-badges/lib/badge-cli.js#L69-L78

We could easily add that feature to the library too, which would be more consistent with the CLI. Simultaneously, we're generally trying to move away from raster format support so I'm not sure we want to introduce that here given we've either never supported it (or possibly we removed it and nobody cared).

paulmelnikow:
Hey, just a heads up that I re-published rho-contracts at https://www.npmjs.com/package/@rho-contracts/rho-contracts. Let me know if you're interested in working on it at all. I'm trying to get a better un-scoped name donated though it's not looking likely that's going to happen.

chris48s:
> Hey, just a heads up that I re-published rho-contracts..

Cheers. I'm going to aim to circle back and update this one tomorrow

chris48s:
Made a couple of updates.. jury's still out on raster formats..

I've changed the base of this PR so its targeted at `v3-dev` instead of `master`. I reckon it makes sense to do that as I work though https://github.com/badges/shields/issues/3587 That way if we need to do a 2.2.2 or something in the meantime, it shouldn't be a problem.

gh-badges/lib/index.js
chris48s: Have you had any further thoughts on this in the context of the work you've been doing on moving raster support out to another service? I'd quite like to get this settled one way or the other so we can finish this PR off and have a look at the other v3 roadmap tasks..

paulmelnikow:
Not to put a wrench in this, though it occurs to me we may want to reconsider whether to include a MPL-licensed dependency in this package.

chris48s:
AFAIK, there isn't an issue with us depending on MPL licenced code in our permissive licenced package as long as the package is distributed in different files with its original source and licence (which it is if you `npm install`). What do you reckon the issue is? Do you think it limits the way downstream consumers can use the package?

Relevant reading: https://tldrlegal.com/license/mozilla-public-license-2.0-(mpl-2)

If its a problem, at this point, I've at least spiked this feature using 4 different validation packages:

- Joi - rejected because its too "heavyweight"/might not be browser-compatible (is this still true in most recent versions?)
- prop-types - not suitable as a general-purpose validation library
- v8n - looked nice, but doesn't produce meaningful error messages for the user (which is the main thing we want)
- rho-contracts - does everything we want, but may be a license issue?

What's implementation number 5? Try another library, or just hard-code the checks? To do every check we want here, we basically need 4 checks:

* "object contains key x"
* "x is a string"
* "x is an array of 2 strings"
* "x is one of [array of strings]"

At this point, I'm leaning towards just writing those statements. I'm not sure looking for another package is actually going to save us time over building from scratch, which is really the reason why we would use a package.

chris48s:
3rd bite of the cherry in #3875

PR #3588, opened 2019-06-19
add [CodeFactor] grade badge

### LINKED ISSUE #3518: Support for CodeFactor
Can you add support for CodeFactor badges?
The default badges for CodeFactor has round edges but I like the `flat-square` style.

badge

calebcartwright:
Closes #3518 

I've never used CodeFactor before but created an account to check it out as they don't really have any public documentation. There's a variety of metrics it shows/tracks in addition to the Grade (cyclomatic complexity, hotspots, duplication, etc.) so I named the classes/files with the idea that we may add other CodeFactor badges down the road

### DIFF
--- a/services/codefactor/codefactor-grade.service.js
+++ b/services/codefactor/codefactor-grade.service.js
@@ -1,0 +1,60 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { BaseSvgScrapingService } = require('..')
+const { letterGrades } = require('./codefactor-helpers')
+
+const schema = Joi.object({
+  message: Joi.allow(...Object.keys(letterGrades)),
+}).required()
+
+module.exports = class CodeFactorGrade extends BaseSvgScrapingService {
+  static get category() {
+    return 'analysis'
+  }
+
+  static get route() {
+    return {
+      base: 'codefactor/grade',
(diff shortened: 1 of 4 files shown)


services/codefactor/codefactor-grade.service.js
```diff
@@ -0,0 +1,60 @@
...
+
+module.exports = class CodeFactorGrade extends BaseSvgScrapingService {
+  static get category() {
+    return 'analysis'
+  }
+
+  static get route() {
+    return {
+      base: 'codefactor/grade',
+      pattern: ':vcsType(github|bitbucket)/:user/:repo/:branch*',
```
calebcartwright: As far as I can tell, CodeFactor only appears to work with GitHub and Bitbucket repos

ryan-ccn: Yep. Thanks!!!

services/codefactor/codefactor-grade.service.js
```diff
@@ -0,0 +1,60 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { BaseSvgScrapingService } = require('..')
+const { letterGrades } = require('./codefactor-helpers')
+
+const schema = Joi.object({
+  message: Joi.allow(...Object.keys(letterGrades)),
```
PyvesB: Clever! 👍

Issue #3608, opened 2019-06-25
Add support for GitHub Package Registry: NPM variants


JamesMGreene:
:clipboard: **Description**

Recently, GitHub has launched a new beta feature: the [GitHub Package Registry (GPR)](https://github.com/features/package-registry)! :tada:

It currently supports alternative registries for NPM, Docker, Maven, NuGet, and RubyGems. For this issue, I would like to focus specifically on the NPM registry variant of GPR.

Ideally, Shields would support GPR support for all NPM-based badge URLs, including those that start with `/npm/, `/node/`, `/jsdelivr/npm/`, `/snyk/vulnerabilities/npm/`, etc.

:link: **Data**

As far as I know, the GitHub Package Registry for NPM supports all of the usual NPM Registry API endpoints (as it is fully compliant with the NPM CLI). This might lead one to believe that we could make use of the `registry_uri` query param that is available on some of the NPM services, e.g. https://img.shields.io/npm/v/@whitesource-yossi/npm-plugin.svg?label=GPR&logo=github&registry_uri=https%3A%2F%2Fnpm.pkg.github.com

However, the GitHub Package Registry's API does not support anonymous access, and so that approach fails with `404` responses.

Looking through the Shields code for the `npm-base` service, I see there is [already support for a Bearer token](https://github.com/badges/shields/blob/8621fe42d76b809e9ba4ae844920f8ed0373817a/services/npm/npm-base.js#L80-L82), so I believe you could establish a new set of `/npm`-equivalent URIs for a separate instance of the `npm-*` services with a [GitHub Personal Access Token (PAT)](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line) as the Bearer token (`npm_token`).

However, you may instead want to consider modifying the code to detect when `registry_uri` is set to `https%3A%2F%2Fnpm.pkg.github.com` and use an alternative secret name like `gpr_token`. This may feel less clean but it should still require minimal changes and would prevent the need to add and maintain alternative URLs &mdash; instead just ensuring that the support for `registry_uri` is available on all NPM-related endpoints.

:microphone: **Motivation**

GitHub Package Registry is likely to become a popular alternative to the NPM registry due to its focus on security, traceability, and mandatory package `@scope`. If a package author chooses to publish their packages to both GPR and NPM, then they could just get badges using the existing NPM endpoints. However, if they only chose to publish to GPR, then the existing endpoints will not suffice.

calebcartwright:
[...]

PR #3610, opened 2019-06-25
move [githubtag] and [githubrelease] under /v, move variants to query params

chris48s:
I've been meaning to get into this for quite a while but strugged to carve out the time. This follows on from this (now quite old) discussion: https://github.com/badges/shields/pull/3144#issuecomment-470344146 about resolving the odd inconsistent behaviours of `/tag` and `/release` which have evolved over time.

I've merged both the tag and release badges into a single version badge on `github/v` with URL params for `include_tags`, `include_prereleases` and `sort=date`/`sort=semver` and redirects on the legacy routes for compatibility. This gives us consitent behaviour for tags and releases with sensible defaults.

Note  I've only looked at Github tag/release version. I haven't addressed the issue of `v/vpre` on other badges here. That can be done in another PR.


services/github/github-version.service.js
```diff
@@ -0,0 +1,239 @@
...
+  .required()
+
+module.exports = class GithubVersion extends GithubAuthService {
+  static get category() {
+    return 'version'
+  }
+
+  static get route() {
+    return {
+      base: 'github/v',
```
calebcartwright: In light of [GitHub's addition of a Package Registry](https://github.com/features/package-registry), and requests like #3608 to add badge support for GH Package Registry, do you think it would it be helpful to include an additional qualifier in the route to differentiate this badge (versions for releases/tags) from the likely future badge for the package registry (version for a published package on GH)?

I suppose one option would be to leave this as is, with an implicit reference to tags/releases, and then do something like `github/package/v` for the package version. It's admittedly a little trickier to be more explicit here since the whole objective is to remove confusion around tags vs. releases.

chris48s: Great question. Looking at an example of a project that has tags, releases and packages: https://github.com/Codertocat/hello-world-npm/packages

I wonder if actually it would be better to keep tags and releases express it as:

* `/v/release/`
* `/v/tag/`
* `/v/package/`

instead of making `include_tags` a GET param (but keep `include_prereleases` and `sort=date`/`sort=semver` as params).

calebcartwright: That sounds like a good plan to me!

chris48s:
I've now split everything back out into tag and release again and rebased the commits so its a bit easier to review.

services/github/github-release.service.js
```diff
@@ -39,22 +92,106 @@ module.exports = class GithubRelease extends GithubAuthService {
...
+    color = sort === 'semver' ? versionColor(version) : color
+    color = isPrerelease ? 'orange' : color
+    return { label: 'release', message: addv(version), color }
+  }
+
+  async fetchReleases({ user, repo }) {
+    return this._requestJson({
+      url: `/repos/${user}/${repo}/releases`,
+      schema: releaseInfoArraySchema,
+      errorMessages: errorMessagesFor('no releases or repo not found'),
```
calebcartwright: For `/releases` I think we get an http 200 with an empty array correct? IIRC we only get the 404 on `/releases/latest` if there's no releases 

```suggestion
      errorMessages: errorMessagesFor('repo not found'),
```

services/github/github-release.service.js
```diff
@@ -39,22 +92,106 @@ module.exports = class GithubRelease extends GithubAuthService {
...
+      latestRelease = await fetchLatestRelease(this, { user, repo })
+      return this.constructor.render({
+        version: latestRelease.tag_name,
+        sort,
+        isPrerelease: latestRelease.prerelease,
+      })
+    }
+
+    const releases = await this.fetchReleases({ user, repo })
+    if (releases.length === 0) throw new NotFound({ prettyMessage: 'none' })
```
calebcartwright: Minor error message suggestion:
```suggestion
    if (releases.length === 0) throw new NotFound({ prettyMessage: 'no releases found' })
```

services/github/github-release.service.js
```diff
@@ -39,22 +92,106 @@ module.exports = class GithubRelease extends GithubAuthService {
...
+
+  static getLatestRelease({ releases, sort, includePrereleases }) {
+    if (sort === 'semver') {
+      const latestRelease = latest(releases.map(release => release.tag_name), {
+        pre: includePrereleases,
+      })
+      const kvpairs = Object.assign(
+        ...releases.map(release => ({ [release.tag_name]: release.prerelease }))
+      )
+      return { tag_name: latestRelease, prerelease: kvpairs[latestRelease] }
```
calebcartwright: Nice 👍 

Side note, but it might be nice one day if we could have the `latest` helper return the prerelease indicator along with the latest version it found (something like `return { version, prerelease: ... }`  instead of `return version`) to avoid the need for callers to have to make that check themselves

services/github/github-release.service.js
calebcartwright: Also, if we moved this line into the `getLatestRelease` function then I think this scenario would be easier to test via a unit test in the spec file.

services/github/github-release.service.js
```diff
@@ -39,22 +92,106 @@ module.exports = class GithubRelease extends GithubAuthService {
...
+    category: 'version',
+    route: {
+      base: 'github/release',
+      pattern: ':user/:repo',
+    },
+    transformPath: ({ user, repo }) => `/github/v/release/${user}/${repo}`,
+    dateAdded: new Date('2019-06-25'),
+  }),
+  GithubReleasePreRedirect: redirector({
+    category: 'version',
```
calebcartwright:
```suggestion
    name: 'GithubReleasePreRedirect',
    category: 'version',    
```

services/github/github-tag.service.js
```diff
@@ -84,39 +76,81 @@ module.exports = class GithubTag extends GithubAuthService {
...
-      json,
+    return tags[0].name
+  }
+
+  async handle({ user, repo }, queryParams) {
+    const sort = queryParams.sort
+    const includePrereleases = queryParams.include_prereleases !== undefined
+
+    const tags = await this.fetch({ user, repo })
+    if (tags.length === 0) throw new NotFound({ prettyMessage: 'none' })
```
calebcartwright: Same suggestions as above in the `release` service to update the message and consider moving this check into `getLatestTag` for easier testing.

```suggestion
    if (tags.length === 0) throw new NotFound({ prettyMessage: 'no tags found' })
```

services/github/github-tag.service.js
```diff
@@ -84,39 +76,81 @@ module.exports = class GithubTag extends GithubAuthService {
...
+    }),
+    dateAdded: new Date('2019-06-25'),
+  }),
+  GithubTagDateRedirect: redirector({
+    category: 'version',
+    route: {
+      base: 'github/tag-date',
+      pattern: ':user/:repo',
+    },
+    transformPath: ({ user, repo }) => `/github/v/tag/${user}/${repo}`,
```
calebcartwright: Just want to confirm that we don't need the `include_prereleases` query param here correct? If I'm following the original code correctly, `tag-date` excluded pre-releases so this redirect maintains consistency.

services/github/github-release.service.js
calebcartwright: I forgot that the `name` would be automatically set so not sure if this is necessary

services/github/github-tag.service.js
chris48s: On the `/releases` endpoint, the `include_prereleases` is meaningful regardless of whether we're sorting by date or semver because there is an "prerelease" boolean.

On the `/tag` endpoint, that doesn't exist - we've just got a string - so `include_prereleases` is only meaningful if `sort=semver` (because we're inferring whether a release is a pre-release ot not based on parsing the string as a semantic version). If `sort=date`, the output is exactly the same regardless of whether you pass `include_prereleases` or not. That's also why there isn't an example for that case.

Instead of `/v/tag/:user/:repo?sort=date` and `/v/tag/:user/:repo?sort=date&include_prereleases` we could set it so that `/v/tag/:user/:repo?sort=date&include_prereleases` throws an error (e.g: "invalid parameter") to make this clearer?

services/github/github-release.service.js
chris48s: I set the names explicitly on the other two redirects because if I didn't do that, they'll both auto-generate the same name which causes an error. I didn't bother explicitly setting a name here because the auto-generated name is unique.

calebcartwright: Gotcha, makes sense

services/github/github-tag.service.js
calebcartwright: Thanks that helps!

I think I'd be inclined to keep the implementation as-is (not throwing an error) given that it's just unused/the default behavior anyway. I could envision someone down the road asking for the most recent, non-prerelease tag by date (not entirely sure what the use case would be, but could definitely see it being requested anyway 😄)

In order to support that, we'd need users to be able to use the `include_prereleases` param to indicate their preference, and could then use it with a slightly updated definition for `getLatestTag`:

```js
static getLatestTag({ tags, sort, includePrereleases }) {
  if (sort === 'semver') {
    return latest(tags.map(tag => tag.name), { pre: includePrereleases })
  }

  if (includePrereleases) {
    return tags[0].name
  }

  // use semver package to find first non-prerelease tag
  const tag = tags.find(t => !semver.prerelease(t.name))
  if (!tag) {
    throw new NotFound({ prettyMessage: 'No release tags found' })
  }

  return tag.name
}
```

calebcartwright: Also, does the GitHub API for tags automatically sort tags by semver? 

When I take a look at the most recent tags for pytest: https://github.com/pytest-dev/pytest/tags, it seems pretty clear that v4.6.4 was published ~3 hours _after_ the v5.0.0 tag. However, the API response for pytest tags appears to have them sorted by semver:
```json
[
  {
    "name": "5.0.0",
    "zipball_url": "https://api.github.com/repos/pytest-dev/pytest/zipball/5.0.0",
    "tarball_url": "https://api.github.com/repos/pytest-dev/pytest/tarball/5.0.0",
    "commit": {
      "sha": "58bfc7736fc4f88eca669157822e00715c67a9bf",
      "url": "https://api.github.com/repos/pytest-dev/pytest/commits/58bfc7736fc4f88eca669157822e00715c67a9bf"
    },
    "node_id": "MDM6UmVmMzc0ODk1MjU6NS4wLjA="
  },
  {
    "name": "4.6.4",
    "zipball_url": "https://api.github.com/repos/pytest-dev/pytest/zipball/4.6.4",
    "tarball_url": "https://api.github.com/repos/pytest-dev/pytest/tarball/4.6.4",
    "commit": {
      "sha": "d3549df5b94137453b832345a9b8074f518731b0",
      "url": "https://api.github.com/repos/pytest-dev/pytest/commits/d3549df5b94137453b832345a9b8074f518731b0"
    },
    "node_id": "MDM6UmVmMzc0ODk1MjU6NC42LjQ="
  },
  {
    "name": "4.6.3",
    "zipball_url": "https://api.github.com/repos/pytest-dev/pytest/zipball/4.6.3",
    "tarball_url": "https://api.github.com/repos/pytest-dev/pytest/tarball/4.6.3",
    "commit": {
      "sha": "b8e65d03bf563d5de3a285af187fd00a2f202e5e",
      "url": "https://api.github.com/repos/pytest-dev/pytest/commits/b8e65d03bf563d5de3a285af187fd00a2f202e5e"
    },
    "node_id": "MDM6UmVmMzc0ODk1MjU6NC42LjM="
  },
  ...
]
```

And accordingly our badges show v5.0.0 for the latest tag by date:
prod


pr review app

chris48s: Well, that's an exciting development, isn't it..

Turns out the GitHub API sorts tags... in (reverse) alphabetical order :open_mouth: 

That's not in the API docs anywhere, but after reading that [on stackoverflow](https://stackoverflow.com/a/19615235) (and thinking "wtf?") I set up a test repo just to prove it to myself: https://api.github.com/repos/chris48s/tags-test/tags

```
a24edac  2019-06-29 22:06:10 +0100  (tag: e)
4ca5cd6  2019-06-29 21:59:53 +0100  (tag: f)
bfa97ef  2019-06-29 21:54:52 +0100  (tag: d)
b3614ac  2019-06-29 21:51:20 +0100  (tag: 1)
b645016  2019-06-29 21:51:12 +0100  (tag: 2)
3309766  2019-06-29 21:49:47 +0100  (tag: c)
8d7c78b  2019-06-29 21:49:42 +0100  (tag: b)
5c5a8d3  2019-06-29 21:49:05 +0100  (tag: a)
```

Additionally, the tags page sorts them differently to the API, but also seemingly not chronologically either: https://github.com/chris48s/tags-test/tags

Of course, thinking it through a bit more, @RedSparr0w is actually at least 4 months ahead of me on this: https://github.com/badges/shields/issues/3080#issuecomment-468486956 if not more :laughing: 

Not really sure what to do with this now.. let me have a think about it :D

chris48s: Having done a bit of reading, the only way to order tags by date in a performant way is to use the v4/graphql API. At the moment, we don't have any badges that use the v4 API, but API calls for new GitHub features (e.g: package registry, dependency graph, etc) are only being exposed via the v4 API so we'll need to get into it at some point.

Given that, we do need to look into the v4 API, maybe I'll put this on hold for the moment and use this as an excuse to have a poke at the v4 API and see what the implications would be for the token pool, etc. That might be the solution here, bit its also probably something we need to look into anyway.

calebcartwright: Sounds like a good plan to me

paulmelnikow:
This was just posted to [an old thread](https://github.com/badges/shields/issues/2117#issuecomment-509018718):

> If i use `https://img.shields.io/github/release/xbmc/xbmc.svg` it will show v18.2-Leia.
> If i use `https://img.shields.io/github/release-pre/xbmc/xbmc.svg` it will show v18.2-Leia.
> If i use `https://img.shields.io/github/tag/xbmc/xbmc.svg` it will show Gotham _alpha11..
> If i use `https://img.shields.io/github/tag-date/xbmc/xbmc.svg` it will show `legacy_drop_vs` as i can find at the bottom of this page `https://release-monitoring.org/project/20547/`.
> 
> Should it really be like this?
> 
> I want to to show v18.3 that is the latest release `https://github.com/xbmc/xbmc/releases` or at least the correct tag `https://github.com/xbmc/xbmc/tags`.

I just re-deployed the review app to http://shields-staging-pr-3610.herokuapp.com/. @Sopor can you give it a try to see if it works for you?

Sopor:
> I just re-deployed the review app to http://shields-staging-pr-3610.herokuapp.com/. @Sopor can you give it a try to see if it works for you?

@paulmelnikow I can't see any difference
`http://shields-staging-pr-3610.herokuapp.com/github/v/tag/xbmc/xbmc.svg` will still show
image

and i have tested them all but no `v18.3-Leia`. Maybe i do something wrong?

RedSparr0w:
>and i have tested them all but no `v18.3-Leia`. Maybe i do something wrong?

Only the first 30 tags are shown currently,
We can kind of work around this by using `?per_page=100` _(currently used on the downloads badge for some reason)_,
which will just make the problem slightly less likely to happen.

But as @chris48s [said above](https://github.com/badges/shields/pull/3610#discussion_r298811430),
We may need to swap over to using [GitHub GraphQL API v4](https://developer.github.com/v4/), But i'm not entirely sure how to use it all yet.
[Example tag query, sorted by date, xmbc repo](https://developer.github.com/v4/explorer/?query=%7B%0A%20%20repository(owner%3A%20"xbmc"%2C%20name%3A%20"xbmc")%20%7B%0A%20%20%20%20refs(refPrefix%3A%20"refs%2Ftags%2F"%2C%20first%3A%20100%2C%20orderBy%3A%20%7Bfield%3A%20TAG_COMMIT_DATE%2C%20direction%3A%20DESC%7D)%20%7B%0A%20%20%20%20%20%20edges%20%7B%0A%20%20%20%20%20%20%20%20node%20%7B%0A%20%20%20%20%20%20%20%20%20%20name%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A)

Not sure why that isn't showing as a proper release though, It doesn't even appear on the [API](https://api.github.com/repos/xbmc/xbmc/releases?per_page=100)
And the "Latest" release is still `1.18.2-Leia`:
Screenshot from 2019-07-10 15-34-48

---------------------

I think something may be wrong with the `include_prereleases` flag:
`https://shields-staging-pr-3610.herokuapp.com/github/v/release/xbmc/xbmc.svg?include_prereleases=1&sort=date`
[image](https://shields-staging-pr-3610.herokuapp.com/github/v/release/xbmc/xbmc.svg?include_prereleases=1&sort=date)

-----------------

paulmelnikow:
In the long run, switching over to the GraphQL API is a great idea. I've a little bit of experience querying GraphQL endpoints from "plain" apps (without a GraphQL framework) though I'm not sure what the best approach for Shields would be. I'd have to dig back in a bit.

chris48s:
I haven't done anything on this in the last few days (I got a bit sucked into the JSDoc thing, which grew in complexity). I'm going to try and find a bit of time to get into this later in the week.

chris48s:
Switching this to the GraphQl endpoint will produce the desired result in this case. For example, this query

```
query {
  repository(owner: "xbmc", name: "xbmc") {
    refs(refPrefix: "refs/tags/", first: 1, orderBy: {field: TAG_COMMIT_DATE, direction: DESC}) {
      edges {
        node {
          name
        }
      }
    }
  }
}
```

will return `v18.3-Leia`.


I did a bit of work on this last week. I've got a generic `BaseGraphqlService` class together and making queries.

I haven't got too deep into it at this point, but I've played with it enough to understand that for Github specifically, the token pool layer is going to need to work differently for the v3 (REST) and v4 (GraphQl) APIs. I'm happy to continue plugging away at that as I have time, but its a bit involved so I need to try and carve out a decent chunk of time to work on it. Will update once I've got something worth reviewing.

paulmelnikow:
> I haven't got too deep into it at this point, but I've played with it enough to understand that for Github specifically, the token pool layer is going to need to work differently for the v3 (REST) and v4 (GraphQl) APIs.

Is it a matter of using separate counters? We already do that for the `/search` endpoints so it should be possible to add a third category. If you want to push a branch with your work in progress I'd be happy to add the token-pool changes.

paulmelnikow:
By the way, which GraphQL fetch library are you using?

The code I was using to make a simple GraphQL request is here:

https://github.com/metabolize/apollo-resolver-gcs/blob/fe02778aacfaba4c8d9ab0bd65aefec5e54a8f44/src/apollo-server.test.js#L36-L47

It's using this package: https://www.npmjs.com/package/graphql-request

It looks like [Lokka](https://www.npmjs.com/package/lokka) is another possibility.

chris48s:
I'm not using any special library. I've just implemented it as a child class of `BaseJsonService`. I think its more useful for every base class to use the same underlying HTTP library (and hence accept the same options, share the same underlying code/helpers, behave in the same way behind a proxy for self-hosting users + so on).

It definitely needs seperate counters. That might be all it needs, but I've not got deep enough into it yet to have worked that out yet. I'll have a look at the code for `/search` I hadn't clocked that we already have 2 ways of counting GH token usage. If I get stuck, I'll tidy things up enough to push a branch and hand that bit over, but I wouldn't mind having a go myself first, when I get a chance.

services/github/github-tag.service.js
```diff
@@ -84,39 +80,101 @@ module.exports = class GithubTag extends GithubAuthV3Service {
...
   }
 
-  static transform({ usingSemver, includePre, json }) {
-    const versions = json.map(({ name }) => name)
-    if (usingSemver) {
-      return latest(versions, { pre: includePre })
-    } else {
-      return versions[0]
+  async fetch({ user, repo, sort }) {
+    const limit = sort === 'semver' ? 100 : 1
```
chris48s: Regardless of how many we request here, the query cost is still 1.

100 is an arbitrary number, so open to feedback but I reckon it keeps the response size reasonable. Grabbing the latest 100 releases should be plenty unless you want latest stabe semver and you released version 1.9 and then put out 101 betas for 2.0 ;)

This should be an improvement over what we were doing before using the first 30 in alphabetical order anyway, and fixes issues like #3080 :)

chris48s:
having shaved the graphql yak, I've now finished this off.

services/github/github-release.service.js
```diff
@@ -39,22 +92,107 @@ module.exports = class GithubRelease extends GithubAuthV3Service {
...
+    route: {
+      base: 'github/release-pre',
+      pattern: ':user/:repo',
+    },
+    transformPath: ({ user, repo }) => `/github/v/release/${user}/${repo}`,
+    transformQueryParams: params => ({ include_prereleases: null }),
+    dateAdded: new Date('2019-06-25'),
```
calebcartwright:
```suggestion
    dateAdded: new Date('2019-08-01'),
```

calebcartwright:
I'm not sure why the deployment is showing as outdated 🤔 will try to get the review app redeployed

services/github/github-release.service.js
```diff
@@ -39,22 +92,107 @@ module.exports = class GithubRelease extends GithubAuthV3Service {
...
+const redirects = {
+  GithubReleaseRedirect: redirector({
+    name: 'GithubReleaseRedirect',
+    category: 'version',
+    route: {
+      base: 'github/release',
+      pattern: ':user/:repo',
+    },
+    transformPath: ({ user, repo }) => `/github/v/release/${user}/${repo}`,
+    dateAdded: new Date('2019-06-25'),
```
calebcartwright:
```suggestion
    dateAdded: new Date('2019-08-02'),
```

calebcartwright:
The new functionality looks to be providing the desired results for the repo @Sopor shared 👍 

https://shields-staging-pr-3610.herokuapp.com/

https://shields-staging-pr-3610.herokuapp.com/github/v/tag/xbmc/xbmc


The latest release (by date) is still showing 18.2, but I believe that is actually the correct result beceause there's a tag for 18.3, but no associated release
https://shields-staging-pr-3610.herokuapp.com/github/v/release/xbmc/xbmc

services/github/github-release.service.js
```diff
@@ -1,32 +1,85 @@
...
+  releaseInfoSchema,
+} = require('./github-common-fetch')
+const { documentation, errorMessagesFor } = require('./github-helpers')
+const { NotFound, redirector } = require('..')
 
-module.exports = class GithubRelease extends GithubAuthV3Service {
+const queryParamSchema = Joi.object({
+  include_prereleases: Joi.equal(''),
+  sort: Joi.string()
+    .valid('date', 'semver')
```
calebcartwright: another wishlist item from me not directly related to this PR, but it'd be cool if the modal window for the query params could have a dropdown for cases like these like we do for route params

services/github/github-release.service.js
```diff
@@ -39,22 +92,107 @@ module.exports = class GithubRelease extends GithubAuthV3Service {
...
 
-  static render({ version, isPrerelease }) {
-    return {
-      message: addv(version),
-      color: isPrerelease ? 'orange' : 'blue',
+  static render({ version, sort, isPrerelease }) {
+    let color = 'blue'
+    color = sort === 'semver' ? versionColor(version) : color
+    color = isPrerelease ? 'orange' : color
+    return { label: 'release', message: addv(version), color }
```
calebcartwright: We don't need to set the label to `release` again here since it's already set as the default, or do you think there's value in having it explicitly inline here with the message and color?

services/github/github-tag.service.js
calebcartwright: The `per_page` query param for the V3 API had a max value of 100, so that feels like a natural choice to me

services/github/github-tag.service.js
```diff
@@ -84,39 +80,101 @@ module.exports = class GithubTag extends GithubAuthV3Service {
...
+const redirects = {
+  GithubTagRedirect: redirector({
+    category: 'version',
+    route: {
+      base: 'github/tag',
+      pattern: ':user/:repo',
+    },
+    transformPath: ({ user, repo }) => `/github/v/tag/${user}/${repo}`,
+    transformQueryParams: params => ({ sort: 'semver' }),
+    dateAdded: new Date('2019-06-25'),
```
calebcartwright:
```suggestion
    dateAdded: new Date('2019-08-02'),
```

services/github/github-tag.service.js
```diff
@@ -84,39 +80,101 @@ module.exports = class GithubTag extends GithubAuthV3Service {
...
+    route: {
+      base: 'github/tag-pre',
+      pattern: ':user/:repo',
+    },
+    transformPath: ({ user, repo }) => `/github/v/tag/${user}/${repo}`,
+    transformQueryParams: params => ({
+      sort: 'semver',
+      include_prereleases: null,
+    }),
+    dateAdded: new Date('2019-06-25'),
```
calebcartwright:
```suggestion
    dateAdded: new Date('2019-08-02'),
```

services/github/github-tag.service.js
```diff
@@ -84,39 +80,101 @@ module.exports = class GithubTag extends GithubAuthV3Service {
...
+  }),
+  GithubTagDateRedirect: redirector({
+    category: 'version',
+    route: {
+      base: 'github/tag-date',
+      pattern: ':user/:repo',
+    },
+    transformPath: ({ user, repo }) => `/github/v/tag/${user}/${repo}`,
+    dateAdded: new Date('2019-06-25'),
```
calebcartwright:
```suggestion
    dateAdded: new Date('2019-08-02'),
```

chris48s:
Oh well spotted. Thanks. That is unexpected behaviour to me based on the description of `TAG_COMMIT_DATE`: https://developer.github.com/v4/enum/reforderfield/

```gql
query {
  repository(owner: "chris48s", name: "tags-test") {
    refs(refPrefix: "refs/tags/", first: 30, orderBy: {field: TAG_COMMIT_DATE, direction: DESC}) {
      edges {
        node {
          name,
          target {
            ...on Commit {
              committedDate
            }
          }
        }
      }
    }
  }
}
```

Yields the response:

```json
{
  "data": {
    "repository": {
      "refs": {
        "edges": [
          {
            "node": {
              "name": "2",
              "target": {
                "committedDate": "2019-06-29T20:51:12Z"
              }
            }
          },
          {
            "node": {
              "name": "1",
              "target": {
                "committedDate": "2019-06-29T20:51:20Z"
              }
            }
          },
          {
            "node": {
              "name": "f",
              "target": {
                "committedDate": "2019-06-29T20:59:53Z"
              }
            }
          },
          {
            "node": {
              "name": "e",
              "target": {
                "committedDate": "2019-06-29T21:06:10Z"
              }
            }
          },
          {
            "node": {
              "name": "d",
              "target": {
                "committedDate": "2019-06-29T20:54:52Z"
              }
            }
          },
          {
            "node": {
              "name": "c",
              "target": {
                "committedDate": "2019-06-29T20:49:47Z"
              }
            }
          },
          {
            "node": {
              "name": "b",
              "target": {
                "committedDate": "2019-06-29T20:49:42Z"
              }
            }
          },
          {
            "node": {
              "name": "a",
              "target": {
                "committedDate": "2019-06-29T20:49:05Z"
              }
            }
          }
        ]
      }
    }
  }
}
```

I'd expect this query to sort such that the tag object 'e' is first as it has the latest commit timestamp.

I've raised a ticket with GH support about this by email. Leave it with me and I'll see what they say. I'll hold off on updating the timestamps until I see what the outcome of this is.

chris48s:
Reply from GitHub support:

> Hey there Chris,
>
> Sorry for the trouble here, you are indeed right - this behaviour doesn't seem to be correct.
>
> I've managed to reproduce this using the example you have provided and have gone ahead and escalated this to the engineering team for you.
>
> I'll let you know as soon as we have an update that we can share.

chris48s:
I've not heard anything more from GH support on this since I raised it a couple of weeks ago.

Given that if/when this gets fixed in the upstream API, there is no change we need to make to accommodate the fix, what do you think to just merging it? Although we have hit an edge case that isn't handled by the V4 API, I reckon the effect of merging this is probably net-positive

calebcartwright:
> what do you think to just merging it?

I had a similar thought as well, so I'm good with that. My only hesitation was that I wasn't sure if there was an off chance they could/would make any changes to the v4 API to address this that we'd need to take into account (I wasn't sure if the v4 API was subject to changes at any time).

Agreed on merging though so I'll give it a 👍 and then hopefully this edge case will get sorted automatically later on when they fix whatever is going on behind the scenes.

chris48s:
As far as I'm aware, v4 should be considered a stable/production-ready API, but it would be fair to say that at this point it has limited adoption in comparison to v3.

calebcartwright:
> As far as I'm aware, v4 should be considered a stable/production-ready API

Excellent and agreed. I also came across this: https://developer.github.com/v4/breaking_changes/ which may be useful for us down the road as we expand our v4 usage.

chris48s:
Cheers. This one has taken a long time to come together since we initially discussed it. Feels good to finally get it over the line :)

PR #3615, opened 2019-06-28
Bump icedfrisby and icedfrisby-nock

paulmelnikow:
This fixes a bug in icedfrisby that emerged with nock beta.15. I'm not sure if this will allow updating to nock beta.20, though it's worth merging regardless.


calebcartwright:
Only a handful of service tests failed when running all of them, and all those failures were timeouts or spurious/unrelated

👍

PR #3630, opened 2019-07-01
JSDoc comments for service test runner

chris48s:
I've not actually added a huge amount here - mainly just JSDoc-ifying comments that were already there.


chris48s:
Good question. I think that linter plugins do exist to enforce that new code has documentation or check that JSDoc comments are syntactically valid or match a particular style.

I don't think there is anything that can throw an error if you say

```js
/**
 * This totally returns a bool - honest, guv!
 *
 * @return {boolean}
 */
function foo() {
    return 'lol, nope!'
}
```

unless you're using typescript.

paulmelnikow:
> Good question. I think that linter plugins do exist to enforce that new code has documentation or check that JSDoc comments are syntactically valid or match a particular style.

Checking the syntax of the comments, and ideally auto-formatting them, seems good enough for now.

I'm using TypeScript more and more though I'm not sure it's worth adopting in the backend. I feel like it creates more learning overhead than we necessarily would benefit from; that the syntax comments you're adding get us most of the way.

I think it'd make working on the frontend way easier, though.

chris48s:
Agreed - I wasn't advocating adopting typescript - just using an example to show what we can/can't do with static analysis.

I've started working through the issues raised by [eslint-plugin-jsdoc](https://www.npmjs.com/package/eslint-plugin-jsdoc). Will aim to submit a PR in the next couple of days with that.

PR #3633, opened 2019-07-01
Support other [Bugzilla] instances

PyvesB:
As stated on Bugzilla's Twitter account:

> We are not http://bugzilla.mozilla.org , we are the software they run on that site (which is also used many other places)

Bugzilla is a widely used piece of open-source software, here's [a list](https://www.bugzilla.org/installation-list/) of other projects using it. Our badges should not be single-sided, this pull request adds support for any Bugzilla instance.


services/bugzilla/bugzilla.service.js
```diff
@@ -29,22 +29,41 @@ module.exports = class Bugzilla extends BaseJsonService {
...
           resolution: '',
         }),
         documentation,
       },
+      {
+        title: 'Bugzilla bug status (with path)',
+        namedParams: {
+          protocol: 'https',
+          host: 'bugs.eclipse.org',
+          path: 'bugs',
```
paulmelnikow: I wonder if it would be cleaner to put the URL in the query string. It makes for a bit more consistent way of handling these.

PyvesB: Do you have an example of a badge using query strings on top of your head?

I mainly based this on Jenkins's badges, which also use paths to target different instances. We may have inconsistent approaches in different places.

calebcartwright: I think the majority of services are using path params for custom urls/instances, but a couple of the newest/recently refactor services are using a query param for the url. Few examples with query param for url that I remember:

- [BitbucketPullRequest](https://github.com/badges/shields/blob/master/services/bitbucket/bitbucket-pull-request.service.js#L13)
- [Drone](https://github.com/badges/shields/blob/8621fe42d76b809e9ba4ae844920f8ed0373817a/services/drone/drone-build.service.js#L15)
- [Mastodon](https://github.com/badges/shields/blob/8621fe42d76b809e9ba4ae844920f8ed0373817a/services/mastodon/mastodon-follow.service.js#L14)

PyvesB: Thanks for those pointers. I'll propose a revised implementation later in the week!

paulmelnikow: Yea, I think we are handling these in different ways. Some services have a host, some have a host and path, and then there's an optional scheme, which makes the approaches less consistent. I like using the query parameter for this as it's the clearest to understand and easiest to implement, though as a negative, it makes for less pretty URLs.

PyvesB: Alternative implementation ready for review! 😁

PR #3637, opened 2019-07-03
Template literal templates

### LINKED ISSUE #2428: Move badge templates from doT → Template literals

paulmelnikow:
This has a good bit of work left to go, though I figured maybe if I put up the PR it might encourage me to nudge it along.

http://shields-staging-pr-3637.herokuapp.com/dev/styles

Closes #2428


chris48s:
How far do you think you are from having this ready for review? I can see this is a big job, but given this one has a big impact on the other package v3 work, it would be really useful if we could try and get this one landed. Do you have a handle on what the remaining tasks are? If its useful to provide comments on the WIP, I'm happy to find some time over the weekend.

paulmelnikow:
So, I’d tried to refactor this while keeping everything working, but I didn’t run tests along the way so what I’ve got is substantially refactored code which neither works nor is as clear as I’d like. The work left is to get it working, and more elegant, at the same time.

There’s some discussions we could start having around SVGO and how we want to handle that. I’m thinking we should not run SVGO at all in production, and instead encode the optimizations into the template, and write a test that SVGO is a no-op.

Put another way, I’m aiming for pixel perfect badges, maybe ones that work better than before, rather than total parity.

paulmelnikow:
I actually have a few hours on a ✈️ tomorrow and was thinking I might look at this again then.

chris48s:
I picked this up in #4459 which is now ready for review, so I'm going to close this.

Issue #3112, opened 2019-02-27
Any objections to deprecating the raster badges?


paulmelnikow:
Shields was founded to produce SVG badges which look really good, though it's long kept the ability to generate raster badges. However, the functionality requires a fair amount of work to maintain, is CPU- and time-intensive, and is probably not very useful.

I suggest we consider removing support for raster badges.

Another option would be to move out raster badges to their own microservice. It would simply fetch an SVG badge from Shields, and then rasterize it. That way the complexity can be completely removed from this codebase, and if there are performance issues they will only affect other raster users. It could be at `png.shields.io` or something like that.

What to do with these requests? Some options:

1. Redirecting to static PNGs that say "raster badges are no longer available".
2. Sneakily redirecting to the `.svg` equivalent. This should "just work" for readmes, websites, and the like. It'll be a unexpected if someone fetches a raster image manually from img.shields.io, though I think if we documented this behavior we'd be okay.
3. Redirecting to a raster service running on a different domain.

Also, a process for getting community feedback on this decision:

1. Add a deprecation notice on https://shields.io/ linking to this issue.
2. Gather feedback for 30 days.
3. Make a decision.

paulmelnikow:
This is the CDN's content-type breakdown for the last month:

Type | Requests
-- | --
svg | 280,896,691
empty | 23,300,961
json | 5,899,672
png | 2,481,242
html | 552,030
other | 474,690
gif | 234
javascript | 210
css | 74
plain | 1

I'm not sure those results are reliable. (What are these javascript requests through the CDN?) It might be good to add some onboard stats for this, along the lines of #3093. That said, it seems safe to say that gif and jpg are not really needed.

paulmelnikow:
Debugging flaky raster tests, one example of work this feature costs us: https://circleci.com/gh/badges/shields/43861

chris48s:
Agree we can dump gif and jpeg without too much ceremony. Although png usage is small proportionally, ~2.5 million requests per month is definitely a non-trivial number. I think we do need to collect feedback to form an opinion on this. If a bunch of people all appear and say "I am using png badges because [insert sensible reason]", that is important to take on board.

henrypoydar:
Sensible reason for supporting raster: embedding badges in emails. The Gmail image proxy, most notably, chokes on SVGs.

paulmelnikow:
[...]

Issue #3587, opened 2019-06-18
version 3 roadmap for npm package


chris48s:
As mentioned in https://github.com/badges/shields/pull/3507 there are a number of changes we've discussed to the NPM package which are not backwards-compatible. It makes sense to batch them into a version 3.

- [x] Validate inputs (done in #3875 )
- [x] Remove legacy properties deprecated in 2.2.0 (done in #3881 )
- [x] Replace the `BadgeFactory` class with a single function (done in #3884 )
    - Now that we're using anafanafo for text measurement, we don't need the class wrapper anymore
    - See https://github.com/badges/shields/pull/2336
- [ ] Expose the concept of logos, add SimpleIcons as a peer dependency (this is actually just a new feature not a BC break, so we could push this back to a minor, but its probably nice to think about in this block of work)
- [ ] Depend on the public interface internally instead of relative imports
- [ ] Publish the package to a new/better name, change repo dir, mark gh-badges dead/unmaintained
    - See https://github.com/badges/shields/pull/2300#issuecomment-438034266

Anything else that should be on that list?

calebcartwright:
If we change the package name (which I'm in favor of), does it make sense to start it with major version 3? I'm not sure what the semver protocols are for renamed packages. 

It's probably best to keep it v3.x to avoid  confusing existing users of `gh-badges` w/2.x or 1.x though right?

paulmelnikow:
Your list looks good! 👍 

Two I might add:

- [ ] Make dependencies for raster support optional (this will make the installed / bundled footprint much smaller)
- [ ] Finish converting to template strings, dropping DoT (#2428)

I agree with keeping the version scheme the same so the new package would still be 3.x. It'll be the same library, with a new name.

chris48s:
one more..

- [ ] Either drop raster support from the CLI or add it to the library (see https://github.com/badges/shields/pull/3586#discussion_r295802787 , https://github.com/badges/shields/pull/3875#discussion_r314467517 for discussion )

calebcartwright:
What about adding a typedef file to the package to make it easier to use from TypeScript (assuming there isn't already one under DefinitelyTyped)?

chris48s:
Presumably adding typescript defs is a non-breaking change, in that consuming them is optional? If so, we could add it in a minor.

As I've got into the v3 work, I'm already starting to think logos might be better pushed back to a minor too so we can limit 3.0 to just the non-backwards-compatible stuff.

calebcartwright:
[...]

Issue #1788, opened 2018-07-19
CLI raster badges are broken


paulmelnikow:
As observed at https://github.com/badges/shields/pull/1774/files#r203522813, the CLI isn't generating raster badges correctly.

```sh
node lib/badge-cli.js '' '' '#bada55' .png > favicon.png
```

Observed output:
![favicon](https://user-images.githubusercontent.com/1487036/42918125-535afb52-8ac1-11e8-88a4-b2918125ebd6.png)

Expected output:
![](https://img.shields.io/:--bada55.png?style=plastic)

platan:
```bash
# prepare a test
cat test.sh 
node lib/badge-cli.js '' '' '#bada55' .svg > favicon-tested.svg
grep 'width="22"' favicon-tested.svg

# find a bad commit
git bisect start HEAD 438e47adfb74e9a775e3fed7a5547959f343c009
git bisect run ./test.sh

# result
b216bcc17e667a2fb09bede7be0d1c67e731d767 is the first bad commit
```
the bad commit -> https://github.com/badges/shields/commit/b216bcc17e667a2fb09bede7be0d1c67e731d767

platan:
But:
```
node lib/badge-cli.js '' '' '#bada55' .png > favicon.png
```

gives me: 
- ![favicon](https://user-images.githubusercontent.com/1526000/44352988-05955800-a4a6-11e8-9060-549e0ad61969.png) in master
- ![favicon](https://user-images.githubusercontent.com/1526000/44353002-1514a100-a4a6-11e8-8a5e-1c71e555fdfc.png) in 438e47adfb74e9a775e3fed7a5547959f343c009 (random old commit)
- ![favicon](https://user-images.githubusercontent.com/1526000/44353142-763c7480-a4a6-11e8-8d67-56397945818e.png) in b216bcc17e667a2fb09bede7be0d1c67e731d767
- ![favicon](https://user-images.githubusercontent.com/1526000/44353163-83f1fa00-a4a6-11e8-9b70-97f11be80757.png) in b216bcc17e667a2fb09bede7be0d1c67e731d767 (one commit before b216bcc17e667a2fb09bede7be0d1c67e731d767)

paulmelnikow:
Seems like this may be a "wontfix," given discussion in #3587 and #3112.

paulmelnikow:
As discussed at https://github.com/badges/shields/pull/3644#issuecomment-508947740 it seems these work fine on some systems.

[...]

PR #3644, opened 2019-07-05
Drop gif + png, and redirect png to raster.shields.io

### LINKED ISSUE #3631: Replace badge server onboard rasterization with rasterizing proxy
As discussed in #3112 and badges/svg-to-image-proxy#25 we're in the process of deploying a new proxy server. This will replace the onboard generation of raster badges on the badge server which will redirect.

This is a feature that has real but relatively obscure use cases (https://github.com/badges/shields/issues/3112#issuecomment-468738270, https://github.com/badges/shields/issues/3112#issuecomment-480422502) and the simplification and performance benefits make it seem worth breaking this out.

It makes things slightly less convenient for self-hosting users who might need raster badges, though it is not hard to also self-host the rasterizer.

Our current usage 2.5M invocations will cost about $1/month (or less if the downstream cache handles a chunk of the load). I'm not sure how much execution time we should expect. I doubt the cost will add up but if it does we could reach out to Zeit about a donation.

Here are the steps to be taken:

- [x] Deploy rasterizing proxy to raster.shields.io
- [x] Make sure other people have access to deploy.
- [x] Document the production hosting. (#3643 badges/svg-to-image-proxy#27)
    - [ ] Design a better process and address security (#3647)
- [x] Get the server running stably.
- [ ] Update the main server to redirect the `.png` badges to the production URL of the proxy. (#3644)
- [ ] Remove the raster capability and its dependencies from the main server. (#3644)
- [ ] Document how to self-host the rasterizing proxy. (#3644)
    - [ ] Improve the self-hosting docs (#3648)

paulmelnikow:
This PR does a few things:

1. Remove rasterization support from the server. This responsibility is delegated to a raster server which proxies the SVG badges and renders them.
2. When a raster server URL is configured, 301 redirect all .png badges to the identical URL on the raster server.
    `https://img.shields.io/npm/v/express.png?style=flat-square` ↪️`https://raster.shields.io/npm/v/express.png?style=flat-square`
3. For configured redirects, redirect to the canonical URL on the raster server.
    `https://img.shields.io/vso/build/totodem/8cf3ec0e-d0c2-4fcd-8206-ad204f254a96/2.png?style=flat-square`
    ↪️`https://img.shields.io/azure-devops/build/totodem/8cf3ec0e-d0c2-4fcd-8206-ad204f254a96/2.png?style=flat-square`
4. Redirect the "legacy badge old version" to the appropriate URL on the raster server.
5. When no raster server is configured (e.g. PRs), render an SVG containing **404 | raster badges not available** for all `.png` badges. (Note that the raster server can be self-hosted; however, this is deferred to a later PR.)
5. Drop support for jpg and gif which are very infrequently used (see #3112). Render an SVG containing **410 | jpg no longer available**.
7. ~~Remove raster dependencies.~~ Remove the raster cache (which is only used in the CLI, and therefore pointless).
8. Move the LRUCache code out of the npm package.
8. A wee bit of refactoring in `server.js`.

Ref #3112
Close #3631


chris48s:
I've not read over all this, but giving it a cursory look, it seems like you're completely removing raster support from the package/CLI in this PR whereas a couple of weeks ago I think you were looking at leaving support in but converting `gm` to a peer dependency: https://github.com/badges/shields/issues/3587#issuecomment-504014841

Is that an oversight or has your opinion on raster support changed since we established that its only the CLI that supports them and not the library: https://github.com/badges/shields/pull/3586#discussion_r295802787 ?

paulmelnikow:
> I've not read over all this, but giving it a cursory look, it seems like you're completely removing raster support from the package/CLI in this PR whereas a couple of weeks ago I think you were looking at leaving support in but converting `gm` to a peer dependency: [#3587 (comment)](https://github.com/badges/shields/issues/3587#issuecomment-504014841)
> 
> Is that an oversight or has your opinion on raster support changed since we established that its only the CLI that supports them and not the library: [#3586 (comment)](https://github.com/badges/shields/pull/3586#discussion_r295802787) ?

Ah, right, I forgot about that. Thanks for the reminder! 😝

I'm fine with restoring that code it and keeping it as a peer dep, though another alternative would be to swap in the more modern rasterization lib [convert-svg-to-png](https://www.npmjs.com/package/convert-svg-to-png) that's being used in svg-to-image-proxy.

This also touches lru-cache. I don't want to smush your work on v3; not really sure the best way forward with these changes at the boundary of the library.

To keep this smallish, what do you think about my adding the raster capability back, using the new dependency, in a separate PR?

paulmelnikow:
Come to think of it, it wouldn't be so hard to put the raster support back into `gh-badges`. I could revert the removal of the `gm` dependency and the `lru-cache` changes. Pretty much all the other removals (e.g. references to imagemagick in the tutorial) could stay as they are.

chris48s:
I don't think the change to the LRU cache makes any difference because it doesn't affect the public interface (or at least, the documented interface) of the package - its just internals.

The bits that change the documented functionality (i.e: the CLI) are best done in a separate PR targetted at the `v3-dev` branch (same as I've done with #3586 ).

paulmelnikow:
I just remembered that raster badges are broken in the CLI, which makes this sort of moot, though I reverted it for now. I kept the lru-cache in its new location and removed internal references to it instead, as even if the CLI worked, the cache would be pointless.

chris48s:
> I just remembered that raster badges are broken in the CLI

Are they? If I run `badge build passed :green .png > mybadge.png` it spits out:

mybadge

paulmelnikow:
Huh, after upgrading imagemagick, this is what I get, which is obviously wrong, though different from what I was seeing when I opened #1788.

mybadge

config/custom-environment-variables.yml
```diff
@@ -14,6 +14,8 @@ public:
 
   redirectUri: 'REDIRECT_URI'
 
+  rasterUrl: 'RASTER_URL'
```
chris48s: Do you think it makes sense to configure a `RASTER_URL` in staging so that something like https://shields-staging-pr-3644.herokuapp.com/appveyor/ci/gruntjs/grunt.png works, or does it just become confusing to mix staging and production architecture?

core/base-service/base.spec.js
```diff
@@ -316,7 +316,7 @@ describe('BaseService', function() {
   })
 
   describe('ScoutCamp integration', function() {
-    const expectedRouteRegex = /^\/foo\/([^/]+?)\.(svg|png|gif|jpg|json)$/
+    const expectedRouteRegex = /^\/foo\/([^/]+?)\.(svg|json)$/
```
chris48s: While we're fiddling with this, is this also a sensible point in time to introduce extensionless SVG https://github.com/badges/shields/issues/2674 or not?

doc/self-hosting.md
```diff
@@ -93,6 +93,23 @@ machine.
...
+which points to your badge server.
+
+- In your raster instance, set `BASE_URL` to your Shields instance, e.g.
+  `https://shields.example.co`.
+- Optionally, in your Shields, instance, configure `RASTER_URL` to the base
+  URL, e.g. `https://raster.example.co`. This will send 301 redirects
+  for the legacy raster URLs instead of 404's.
+
+If anyone has set this up, more documentation on how to do this would be
+welcome! It would also be nice to ship a Docker image that includes a
```
chris48s: Do you think its worth adding a sentence just saying its designed to run as a lambda on now/AWS/etc, or is that stuff just best covered in the README for `svg-to-image-proxy`?

Dockerfile
```diff
@@ -1,7 +1,5 @@
 FROM node:8-alpine
 
-RUN apk add --no-cache gettext imagemagick librsvg git
```
chris48s: do we definitely not need git anymore?

config/custom-environment-variables.yml
paulmelnikow: Yea, I was thinking it would be nice to have one for the staging environment, though perhaps it's not necessary for PRs. I _think_ we could configure the env var in the staging app and that way it'll apply only to staging (not PRs).

doc/self-hosting.md
paulmelnikow: I think that's a fine thing to add. If it's information that isn't going to change, it's not the worst thing to mention it in two places.

Dockerfile
paulmelnikow: It looks like the automated Docker build ran successfully on this branch, and I can't think of a reason why it would be needed at runtime.

Screen Shot 2019-07-06 at 3 47 19 PM

Link: https://cloud.docker.com/u/shieldsio/repository/registry-1.docker.io/shieldsio/shields/timeline

core/base-service/base.spec.js
paulmelnikow: Yea, it's a good point. I was thinking about that, but I'm eager to deploy this so we can find out how the raster server is performing. Extensionless SVG will need a few other changes made including some CDN config tweaks. I'm curious to find out how stable the raster server is and don't really want to hold this up.

chris48s: Ok. Lets stick to changing one thing at a time :+1:

config/custom-environment-variables.yml
chris48s: its something we can deal with as an env var in heroku anyway, rather than something that needs changing in this PR

paulmelnikow:
Gosh, shall we ship this? 🚀

paulmelnikow:
1. Working: https://shields-raster-staging.now.sh/badge/foo-bar-blue.png?style=flat-square

2. Not redirecting correctly: http://shields-staging.herokuapp.com/badge/foo-bar-blue.png?style=flat-square
    It's dropping `?style=flat-square` and redirecting to a nonsense static badge:
    https://shields-raster-staging.now.sh/badge/badge-foo--bar--blue-lightgray.png

3. Also not redirecting correctly: http://shields-staging.herokuapp.com/npm/v/express.png?style=flat-square

paulmelnikow:
Confirmed fixed in #3659. (If it's not fixed for you, the bad redirects are probably stuck in your cache.)

PR #3645, opened 2019-07-05
adopt JSDoc, eslint-plugin-jsdoc

chris48s:
I've mostly managed to appease the lint gods, but there are some remaining warnings to fix. Probably worth opening a conversation at this point. I'll leave some comments inline. Some of these can be fixed by me just understanding the code properly so I can document it. Some of them are a bit more tricky..


.eslintrc.yml
```diff
@@ -82,3 +84,6 @@ rules:
   # Chai friendly.
   no-unused-expressions: 'off'
   chai-friendly/no-unused-expressions: 'error'
+
+  # Disable some rules from jsdoc/recommended.
+  jsdoc/require-jsdoc: 0
```
paulmelnikow: You could set this to `warn` which would give us results in CI while this is open.

chris48s:
Turns out you can't leave a line comment on a line that isn't touched in the diff, so I'll do it like this instead:

---

These are just about understanding the code so I can document it properly:

https://github.com/badges/shields/blob/f693ef019d5cd93262d2750c25760babdcd155b5/core/token-pooling/token-pool.js#L198-L204

* What is a sensible type hint for `id` here? Is the comment actually right? We say it can be a primitive value or an object reference, but realistically it looks like it has to be a string because we call `sanitizeToken()` on it. The only examples I can find of this code in use, `id` is always string.
* What is a sensible type/description for `data` here? I'm not able to find any examples of code where we're actually using this param for anything.

---

https://github.com/badges/shields/blob/0d3266a0adf6f990361023e9e56133ddc91b0ca0/core/service-test-runner/service-tester.js#L107-L114

I'm not really sure what to do with situations like this. Sometimes we define a function which takes an object which doesn't really have a sensible description - we've just made it an object so we've got named params. Not giving it a description throws `Missing JSDoc @param "attrs" description`.

Options are:

* Disable the check globally and just allow params with no description
* Wrap the comment in

    ```js
    /* eslint-disable jsdoc/require-param-description */
    ...
    /* eslint-enable jsdoc/require-param-description */
    ```

    everywhere we've done this

* Give it a pointless description just to keep the linter happy e.g: `@param {object} attrs attrs` or something

core/service-test-runner/infer-pull-request.js
```diff
@@ -64,13 +64,13 @@ function _inferPullRequestFromCircleEnv(env) {
...
- * @return {string} pr.repo
- * @return {string} pr.pullRequest PR/issue number
- * @return {string} pr.slug owner/repo/#pullRequest
+ * @param {object} [env=process.env] Environment variables
+ * @returns {object} pr
+ * @returns {string} pr.baseUrl (returned for travis CI only)
+ * @returns {string} pr.owner
+ * @returns {string} pr.repo
+ * @returns {string} pr.pullRequest PR/issue number
+ * @returns {string} pr.slug owner/repo/#pullRequest
```
chris48s: I'm not really sure what to do about these. Its useful to be able to describe the structure of a soft object. This is quite human-readable and some JSDoc --> HTML/markdown/whatevz tools do accept this multiple return syntax and render it validly. For example, doxdox is perfectly happy to render this: https://doxdox.org/badges/shields#core-service-test-runner-infer-pull-request.js-inferpullrequest

That said:

* doxdox does have some other issues - I've just picked it here because it can render docs straight from github which is convenient
* some tools do only render the first/last return

..but doing this throws `Found more than one @returns declaration` in the linter. The other options are:

* `@returns {object} pr`, which is less helpful
* Convert it to a `@typedef` which does render in some tools, but also `@typedef` is not supported by some others and defining the types outside of the function definition is less human-redable/editor hinting-friendly.

There doesn't seem to be a great solution to this. I guess it kind of depends on which JSDoc --> HTML/markdown/whatevz tool we opt for (spoiler: every single one I've tried so far doesn't do something important we want).

The other thing that is annoying is that multiple returns isn't a single check. You have to disable both `require-returns` and `require-returns-check` to not throw it.

.eslintrc.yml
chris48s: maybe one day ;)

paulmelnikow:
> I'm not really sure what to do with situations like this. Sometimes we define a function which takes an object which doesn't really have a sensible description - we've just made it an object so we've got named params. Not giving it a description throws `Missing JSDoc @param "attrs" description`.

I wonder if we could contribute a rule or option to the plugin that lets us handle this case in a smarter way. It could be a global exception for object parameters (not great) or better yet, one that validates that each of the destructured object parameters is documented.

> * Give it a pointless description just to keep the linter happy e.g: `@param {object} attrs attrs` or something

Not the worst option in the meantime…

`@param {object} attrs Refer to individual attrs`

paulmelnikow:
> https://github.com/badges/shields/blob/f693ef019d5cd93262d2750c25760babdcd155b5/core/token-pooling/token-pool.js#L198-L204

You brave soul, wading into TokenPool.

> * What is a sensible type hint for `id` here? Is the comment actually right? We say it can be a primitive value or an object reference, but realistically it looks like it has to be a string because we call `sanitizeToken()` on it. The only examples I can find of this code in use, `id` is always string.
> * What is a sensible type/description for `data` here? I'm not able to find any examples of code where we're actually using this param for anything.

I'm honestly not sure if other services that might use a token pool in the future (e.g. GitLab) would need more than just the token string. I kind of hate to wade into that code just to drop abstractions we might use sometime, though it's even worse to fix + maintain abstractions we don't need; it's throwing good time after bad.

I'd support renaming `id` to `tokenString` and removing `data`.

core/service-test-runner/infer-pull-request.js
paulmelnikow: A lint ignore is another option.

I can't say this particular helper function feels all that important, in the scheme of things. Are there many others of these? If it's just this function, I'm not sure it matters all that much which solution you pick.

chris48s: I picked this one to leave a comment on, but we do have several occurrences of this.
Its something we're going to run into as we document more stuff though. In general,

```js
return {
  foo: 73,
  bar: 'hello',
  baz: false,
}
```

is a very common thing to do in javascript and in any sort of meaningful docs, we're going to want to describe what `foo`, `bar` and `baz` are. As I say, I think it partly depends on what tooling we end up using so maybe that's the next sensible thing to look at before writing more docs.

core/token-pooling/token-pool.js
```diff
@@ -198,11 +198,8 @@ class TokenPool {
   /**
    * Add a token with user-provided ID and data.
    *
-   * @param id
-   *    The ID can be a primitive value or an object reference, and is used
-   *    (with `Set`) for deduplication. If a token already exists with a
-   *    given id, it will be ignored.
-   * @param data
+   * @param {string} id token string
+   * @param {*} data reserved for future use
```
paulmelnikow: 👍

core/service-test-runner/infer-pull-request.js
chris48s: One final way to think about this may be to say that if we've got complex objects we're going to pass around that need describing, maybe we should formalise them as classes. I tend to be of the opinion that the purpose of documentation is to describe what we've done rather than to decide how we do it. That said, maybe there are some situations where the act of describing the code is also helping to see how we should improve it.

For the moment, I've just disabled the lint rules to get rid of the remaining warnings so we can merge this, but I'll continue to think about this.

calebcartwright: Another approach I've seen folks used in the past is defining a doc typedef for the return type:
```javascript
/**
 * @typedef {object} ThatReturnedThing
 * @property {number} foo - Lorem ipsum.
 * @property {string} bar - Lorem ipsum.
 * @property {boolean} baz - Lorem ipsum.
 */
```

and then using it on the function doc
```js
 * ...
 * @returns {ThatReturnedThing} 
 */
```

chris48s: Yep. I did mention `@typedef`, but support for it seems to be non-universal - hence next thing to do is think about tooling..
Out of interest, do you have any suggestions for JSDoc renderers worth having a look at?

calebcartwright: > Yep. I did mention @typedef

Sorry 🤦‍♂ And that's why it is important to read thoroughly!

> Out of interest, do you have any suggestions for JSDoc renderers worth having a look at?

Not really. I kinda like the output of Toast UI template (https://www.npmjs.com/package/tui-jsdoc-template) that I've seen used in a few places, like Yeoman https://yeoman.github.io/generator/, but I don't have much experience nor strong opinions with any of them.

paulmelnikow: Have you looked at https://pypi.org/project/sphinx-js/ by chance? I'm not the hugest fan of Sphinx. I find it a pain to set up and I don't really know RsT. But I do use it; and once it's set up it works.

chris48s: No worries + cheers for the suggestions, both.

I think having tried out a few different tools, JSDoc3 (which could be used with the toast theme if we want) + typedefs is the right way to go, but I think JSDoc3 also requires some additional annotations for modules and stuff. I reckon we might as well aim to get it right in this PR because the tooling we adopt slightly affects how we want to do the annotations.

Leave this PR open. I'll expand the scope slightly to also adopting a renderer as well as some lint rules. I'll do some more work on it and push some more commits.

paulmelnikow: Seems like we ought to deploy the developer docs separately from the shields.io app so they are always in sync with master. Ideally we'd get review apps for PRs as well. I've been playing with Netlify which is extraordinarily easy to set up and might be a nice option; would also keep it out of the standard Heroku build for self-hosting users.

Not sure if you want to pull in the hosting for this PR too. I'm happy to help with that in this branch if you'd like.

core/base-service/base.js
```diff
@@ -95,47 +97,16 @@ module.exports = class BaseService {
...
-   *    A parameter requiring a numeric string can use
-   *    `Joi.number()`. A parameter that receives only non-numeric
-   *    strings can use `Joi.string()`. A parameter that never
-   *    receives numeric can use `Joi.string()`. A boolean
-   *    parameter should use `Joi.equal('')` and will receive an
-   *    empty string on e.g. `?compact_message` and undefined
-   *    when the parameter is absent. (Note that in,
-   *    `examples.queryParams` boolean query params should be given
-   *    `null` values.)
+   * @type {module:core/base-service/base~Route}
```
chris48s: We have to use this syntax in quite a few places because of this https://github.com/jsdoc/jsdoc/issues/969

Its fine.. until we move stuff about, at which point we'll probably forget and break links. Hopefully, that is where the `jsdoc/no-undefined-types` lint rule steps in to save us :crossed_fingers:

jsdoc.json
```diff
@@ -0,0 +1,20 @@
...
+      "build",
+      "coverage",
+      "logo",
+      "node_modules",
+      "private",
+      "public"
+    ]
+  },
+  "plugins": ["plugins/markdown"],
+  "opts": {
```
chris48s: We could also specify a theme here too, if we wanted but I've gone with the default for now

jsdoc.json
```diff
@@ -0,0 +1,20 @@
...
+    "destination": "api-docs/",
+    "readme": "README.md",
+    "recurse": true,
+    "tutorials": "doc/"
```
chris48s: This feature is really nice because it means we can embed our markdown tutorials in with the API docs.
Unfortunately the logo tutorial has one line that renders a bit odd. Its not the end of the world, and we can probably sort that with a bit of custom CSS, but I'll wait and see what we do about theming first.

package.json
```diff
@@ -111,7 +111,8 @@
     "start": "concurrently --names server,frontend \"npm run start:server\" \"cross-env GATSBY_BASE_URL=http://localhost:8080 gatsby develop --port 3000\"",
     "e2e": "start-server-and-test start http://localhost:3000 test:e2e",
     "e2e-on-build": "cross-env CYPRESS_baseUrl=http://localhost:8080 start-server-and-test start:server:e2e-on-build http://localhost:8080 test:e2e",
-    "badge": "cross-env NODE_CONFIG_ENV=test TRACE_SERVICES=true node scripts/badge-cli.js"
+    "badge": "cross-env NODE_CONFIG_ENV=test TRACE_SERVICES=true node scripts/badge-cli.js",
+    "build-docs": "rm -rf api-docs/ && jsdoc --pedantic -c ./jsdoc.json . && mkdir -p ./api-docs/frontend/images/ && cp ./frontend/images/logo.svg ./api-docs/frontend/images/logo.svg"
```
chris48s: Unfortunately this script isn't very cross-platform :( There's a few couple of things going on here:

1. JSdoc doesn't clear the output dir to start with by default and it doesn't seem to have a `--clear` option or anything, but if you move files about, you get stale docs left around so it _is_ useful to clear the output dir first.
2. I've set `"readme": "README.md",` in the JSDoc config file so that the home page of the docs has something on it. I'm copying this image so that the top image on the home page isn't broken. Other solutions here might be to host the image somewhere and link directly to the image with absolute path (instead of a relative path), or we could write a bespoke markdown page to use as the documentation index instead of using the repo README, which would probably be better in the long-run.

core/service-test-runner/infer-pull-request.js
chris48s: Yes - agree that we should get docs building + deploying sooner rather than later as its definitely going to be useful to:

a) get them somewhere contributors can see/use them
b) check how they render in staging when reviewing PRs that contribute docs

Thanks for the offer of help. I haven't got as far as thinking about the best way to host it, so happy to defer to your ideas on this. Normally I'd bung the docs on the gh-pages branch, but we're already using that on this repo.

I think it would probably be better to do that in a separate PR though as this one is already touching a lot of files. 2 smaller diffs will be easier to review.

chris48s:
I have a couple of small merge conflicts to resolve - I'll sort them out tomorrow. Accepting that issue, everything else should be good for review/feedback

package.json
paulmelnikow: To clean the output cross-platformly, you can use `rimraf` like in `coverage:clean`.

For the logo, I think you could use this link:

https://raw.githubusercontent.com/badges/shields/master/frontend/images/logo.svg?sanitize=true

(see https://stackoverflow.com/a/16462143/893113)

chris48s: > cross-platformly

perfectly cromulent word ;)


> https://stackoverflow.com/a/16462143/893113

Useful - I didn't know about `?sanitize=true`

package.json
```diff
@@ -111,7 +111,8 @@
     "start": "concurrently --names server,frontend \"npm run start:server\" \"cross-env GATSBY_BASE_URL=http://localhost:8080 gatsby develop --port 3000\"",
     "e2e": "start-server-and-test start http://localhost:3000 test:e2e",
     "e2e-on-build": "cross-env CYPRESS_baseUrl=http://localhost:8080 start-server-and-test start:server:e2e-on-build http://localhost:8080 test:e2e",
-    "badge": "cross-env NODE_CONFIG_ENV=test TRACE_SERVICES=true node scripts/badge-cli.js"
+    "badge": "cross-env NODE_CONFIG_ENV=test TRACE_SERVICES=true node scripts/badge-cli.js",
+    "build-docs": "rimraf api-docs/ && jsdoc --pedantic -c ./jsdoc.json . && mkdir -p ./api-docs/frontend/images/ && cp ./frontend/images/logo.svg ./api-docs/frontend/images/logo.svg"
```
paulmelnikow: I wonder if this needs `mkdirp` to be cross-platform.

calebcartwright: I think we will need to use `mkdirp` too 👍

core/base-service/base.js
calebcartwright: That makes sense and good that it will (probably/hopefully) get caught in the linting process. 

This is another one of those times when I'd hope danger could do some automagic to detect but that seems like it would be pretty challenging to check precisely. _Maybe_  danger could detect file renames/directory changes under `./core` and print a warning like we do for [serverSecrets](https://github.com/badges/shields/blob/master/dangerfile.js#L115)?

package.json
chris48s: Ah yes I forgot to remove the mkdir/cp commands when I changed the image to an absolute URL. Have just removed them.

PR #3652, opened 2019-07-06
Inject secrets into the services [azuredevops bintray drone jenkins jira npm sonar symfony teamcity wheelmap]

paulmelnikow:
This is a reworking of #3410 based on some feedback @calebcartwright left on that PR.

The goals of injecting the secrets are threefold:

1. Simplify testing
2. Be consistent with all of the other config (which is injected)
3. Encapsulate the sensitive auth-related code in one place so it can be studied and tested thoroughly

- Rather than add more code to BaseService to handle authorization logic, it delegates that to an AuthHelper class.
- When the server starts, it fetches the credentials from `config` and injects them into `BaseService.register()` which passes them to `invoke()`.
- In `invoke()` the service's auth configuration is checked (`static get auth()`, much like `static get route()`).
- If the auth config is present, an AuthHelper instance is created and attached to the new instance.
- Then within the service, the password, basic auth config, or bearer authentication can be accessed via e.g. `this.authHelper.basicAuth` and passed to `this._requestJson()` and friends.
- Everything is being done very explicitly, so it should be very clear where and how the configured secrets are being used.
- Testing different configurations of services can now be done by injecting the config into `invoke()` in `.spec` files instead of mocking global state in the service tests as was done before. See the new Jira spec files for a good example of this.

Ref #3393

Need adding in this PR:

- [x] Nexus
- [x] Symfony

I've deferred a few odds and ends to a later PR:

- [ ] Bitbucket. It has two sets of credentials in the same service so I have to think about how to solve that.
- [ ] GithubConstellation etc.
- [ ] `suggest`, and various other places in the server where the secrets are used
- [ ] Wheelmap service test config check


calebcartwright:
Know this is still a WIP, but I just finished a quick browse through the current set of changes and wanted to say it's looking good!

paulmelnikow:
Thanks for reading I'd like to Nexus and Symfony though I think I'm at a pausing point for today. The rest of this is ready for review, so if this gets merged I can open a new PR with those additional bits.

core/base-service/auth-helper.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  get isConfigured() {
+    return (
+      (this._userKey ? Boolean(this.user) : true) &&
+      (this._passKey ? Boolean(this.pass) : true)
+    )
+  }
+
+  get isValid() {
```
calebcartwright: Is this going to be used when the service classes (the ones using auth) are instantiated or when requests are made? I _believe_ it's the latter if I'm following the PR summary and some of the below files correctly

calebcartwright: The use case I originally had in the back of my mind (trying to catch config errors as early as possible) may not be as applicable. However, I'm not sure that's all that relevant. For example, SymfonyInsight service requires authorization, but I don't set that in my self-hosted Shields deployment (we don't need it) and I wouldn't want to _have_ to set it either. 

IIRC we have a similar setup in the PR review/staging apps and CI jobs where services like Wheelmap and Symfony don't have the required auth configured either. It might be nice to catch an invalid config at deploy time for a Product Shields.io deployment though (if we don't already have something like that).

core/base-service/errors.js
```diff
@@ -72,6 +72,23 @@ class Inaccessible extends ShieldsRuntimeError {
   }
 }
 
+class ImproperlyConfigured extends ShieldsRuntimeError {
```
calebcartwright: Do you think we'd ever have other, non-auth misconfiguration types of "configuration" related errors? IMO this is good as is, just wanted to put the question out there.

paulmelnikow: Hmm, I'm not sure. Usually we quit on bad configuration. What makes this one unusual is that we have started the server, and just aren't supporting this request.

I borrowed the name [from Django](https://docs.djangoproject.com/en/2.2/ref/exceptions/#improperlyconfigured) (which admittedly has a lot of configuration).

core/base-service/auth-helper.js
paulmelnikow: > Is this going to be used when the service classes (the ones using auth) are instantiated or when requests are made? I _believe_ it's the latter if I'm following the PR summary and some of the below files correctly

For now it's when requests are made, though I agree it would be better to examine the config at load time. That would mean moving it to `register()`, where I had it initially, but it made testing really annoying because you couldn't call `invoke()` without running the auth config first.

It would be nice to bail on startup if services are _partially_ configured (i.e. failing as loudly as possible).

I'd be open to improving/revisiting this after #3328.

calebcartwright:
Side note, I love the new PR feature that keeps track of which files I've viewed. Makes things a lot easier on you when you don't have time to go through everything in one sitting

dangerfile.js
```diff
@@ -108,10 +108,10 @@ if (allFiles.length > 100) {
   allFiles.forEach(file => {
     // eslint-disable-next-line promise/prefer-await-to-then
     danger.git.diffForFile(file).then(({ diff }) => {
-      if (/serverSecrets/.test(diff) && !secretsDocs.modified) {
+      if (/authHelper/.test(diff) && !secretsDocs.modified) {
```
calebcartwright: Should we check for both of these for now, and then remove `serverSecrets` once all the services have been updated to work with the new model?

core/base-service/errors.js
calebcartwright: No worries, was wondering if we'd ever need/want different error types for Auth Configuration errors vs. other config errors, but doesn't seem necessary at all considering.

core/base-service/base.spec.js
```diff
@@ -482,4 +482,43 @@ describe('BaseService', function() {
...
+
+    it('when auth is not configured properly, invoke() returns inacessible', async function() {
+      expect(
+        await AuthService.invoke({}, defaultConfig, {
+          namedParamA: 'bar.bar.bar',
+        })
+      ).to.deep.equal({
+        color: 'lightgray',
+        isError: true,
+        message: 'credentials have not been configured',
```
calebcartwright: I like this 👍

services/azure-devops/azure-devops-helpers.js
```diff
@@ -23,15 +22,4 @@ async function fetch(serviceInstance, { url, qs = {}, errorMessages }) {
...
-function getHeaders() {
-  const headers = {}
-  if (serverSecrets.azure_devops_token) {
-    const pat = serverSecrets.azure_devops_token
-    const auth = Buffer.from(`:${pat}`).toString('base64')
-    headers.Authorization = `basic ${auth}`
-  }
-
-  return headers
-}
```
calebcartwright: 🎉 🍾 The `auth()` approach feels so much simpler/cleaner in these service classes, glad to see the removal of this

services/symfony/symfony-insight-base.js
```diff
@@ -40,49 +39,41 @@ const keywords = ['sensiolabs', 'sensio']
...
 const gradeColors = {
   none: 'red',
-  bronze: '#C88F6A',
-  silver: '#C0C0C0',
-  gold: '#EBC760',
-  platinum: '#E5E4E2',
+  bronze: '#c88f6a',
+  silver: '#c0c0c0',
+  gold: '#ebc760',
+  platinum: '#e5e4e2',
```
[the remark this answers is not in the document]
calebcartwright: I don't have any objections to this. IIRC, When I added these back in #2971, I just copied in the hex values from some tool which capitalized them and didn't think about it too much.

services/drone/drone-build.service.js
```diff
@@ -85,11 +88,7 @@ module.exports = class DroneBuild extends BaseJsonService {
       qs: {
         ref: branch ? `refs/heads/${branch}` : undefined,
       },
-    }
-    if (serverSecrets.drone_token) {
-      options.headers = {
-        Authorization: `Bearer ${serverSecrets.drone_token}`,
-      }
+      headers: this.authHelper.bearerAuthHeader,
```
calebcartwright: If there's no `drone_token` set in the config, then this ends up as `headers: undefined` which is why we no longer need the conditional check that was done before, correct?

services/jira/jira-issue.tester.js
```diff
@@ -161,26 +160,3 @@ t.create('blue-gray status color')
...
-      .reply(200, {
-        fields: {
-          status: {
-            name: 'in progress',
-          },
-        },
-      })
-  )
-  .finally(restore)
-  .expectBadge({ label: 'secure-234', message: 'in progress' })
```
calebcartwright: 👏 woohoo!

services/jira/jira-issue.service.js
```diff
@@ -1,7 +1,8 @@
 'use strict'
 
 const Joi = require('@hapi/joi')
-const JiraBase = require('./jira-base')
```
calebcartwright: Can the `jira-base.js` file be deleted now?

services/drone/drone-build.service.js
paulmelnikow: Correct!

services/azure-devops/azure-devops-helpers.js
paulmelnikow: Thanks! This version came out really well. I'm glad for the back-and-forth we had on #3410; think it helped a lot in getting to this point.

services/sonar/sonar-base.js
```diff
@@ -1,11 +1,10 @@
 'use strict'
 
 const Joi = require('@hapi/joi')
-const serverSecrets = require('../../lib/server-secrets')
-const { isLegacyVersion } = require('./sonar-helpers')
 const { BaseJsonService } = require('..')
+const { isLegacyVersion } = require('./sonar-helpers')
 
-const schema = Joi.object({
+const modernSchema = Joi.object({
```
calebcartwright: Definitely more descriptive names for thease schemas 👍

services/symfony/symfony-insight-base.js
```diff
@@ -40,49 +39,41 @@ const keywords = ['sensiolabs', 'sensio']
...
-      },
-    }
-
-    if (
-      !serverSecrets.sl_insight_userUuid ||
-      !serverSecrets.sl_insight_apiToken
-    ) {
-      throw new Inaccessible({
-        prettyMessage: 'required API tokens not found in config',
-      })
```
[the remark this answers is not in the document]
calebcartwright: Awesome!

services/symfony/symfony-insight-base.js
```diff
@@ -40,49 +39,41 @@ const keywords = ['sensiolabs', 'sensio']
...
 }
 
 class SymfonyInsightBase extends BaseXmlService {
   static get category() {
     return 'analysis'
   }
 
+  static get auth() {
+    return {
+      userKey: 'sl_insight_userUuid',
```
calebcartwright: just noticed the capital `U` in these keys, whereas the vast majority of others are all lowercase with underscores. Should the key checks in the auth helpers be case sensitive?

services/symfony/symfony-insight-grade.tester.js
```diff
@@ -37,114 +19,10 @@ createTest(t, 'live: valid project grade', { withMockCreds: false })
     ).required(),
   })
 
-createTest(t, 'live: nonexistent project', { withMockCreds: false })
-  .before(prepLiveTest)
+t.create('live: nonexistent project')
```
calebcartwright: 🤔 Should we remove the `live:` prefix from these tests in light of #3661? These conditional skip tests are a little tricky since sometimes their live but others mocked.

services/symfony/symfony-insight.spec.js
```diff
@@ -0,0 +1,255 @@
+'use strict'
+
```
calebcartwright: This makes things a lot cleaner 👍 Haven't checked out this branch locally yet, but I assume the coverage levels are still the same for these services?

services/symfony/symfony-test-helpers.js
```diff
@@ -78,53 +77,24 @@ const multipleViolations = createMockResponse({
...
-  // In the case of the live tests we want to use the "real" creds if they
-  // exist otherwise we need to use the same stubbed creds as all the mocked tests.
-  if (!originalUuid) {
+function checkShouldSkip() {
+  const noToken =
+    !serverSecrets.sl_insight_userUuid || !serverSecrets.sl_insight_apiToken
+  if (noToken) {
     console.warn(
-      'No token provided, this test will mock Symfony Insight API responses.'
+      'No Symfony credetnials configured. Service tests will be skipped. Add credentials in local.yml to run these tests.'
```
calebcartwright:
```suggestion
      'No Symfony credentials configured. Service tests will be skipped. Add credentials in local.yml to run these tests.'
```

services/test-helpers.js
```diff
@@ -0,0 +1,24 @@
...
+const { promisify } = require('../core/base-service/legacy-request-handler')
+
+function cleanUpNockAfterEach() {
+  afterEach(function() {
+    nock.restore()
+    nock.cleanAll()
+    nock.enableNetConnect()
+    nock.activate()
+  })
+}
```
calebcartwright: 💯

services/symfony/symfony-insight-base.js
paulmelnikow: All this stuff passes through the Joi schema in server.js so if we were gonna be lenient we’d need to do so there. As is people get loud failures when things are wrong, though a case error is admittedly very picky. I’d be fine if we updated the case in a later PR, don’t feel strongly one way or another.

services/symfony/symfony-insight-grade.tester.js
paulmelnikow: Now they’re always live, right? Live, or skipped. But yea, I could update these titles.

services/symfony/symfony-insight-base.js
calebcartwright: SGTM

services/symfony/symfony-insight.spec.js
paulmelnikow: I’m hoping the coverage is good via test:core now! I don’t usually run coverage locally, I’m curious though.

services/symfony/symfony-insight-grade.tester.js
calebcartwright: > Now they’re always live, right? Live, or skipped

Ah yes you are correct. I didn't pay close enough attention the change

services/symfony/symfony-insight.spec.js
calebcartwright: It may be even higher now. Coverage was pretty high before but I think there were a couple blocks in the base service that I didn't bother to hit, and those blocks may be gone now (covered via test:core like you noted).

calebcartwright:
I'm ready to approve and excited about this change! Are you working on those lint errors already? If not I'll fix and add a commit

calebcartwright:
That nexus test timeout is unrelated (it's one of more more notoriously slow tests), so i'm going to remove it from the title

paulmelnikow:
Thanks for the review! This was a big one 😄

PR #3659, opened 2019-07-06
Fix raster redirect and drop the legacy static .png route

paulmelnikow:
This should fix the problem I noted at https://github.com/badges/shields/pull/3644#issuecomment-508952678 though it notes a bug, which is that the legacy static badge route `/label/message.png?color=green` does not work. That route is very, very old, and has never supported SVG, so I'm inclined to drop it, though if we want it to keep working, this redirect code is going to have to be changed in a more fundamental way. That route has always been registered last because it's difficult to distinguish from real badges.

I'm inclined to just drop the legacy static badge route since it probably isn't being used much (since it doesn't support SVG) and results in the server behaving differently with .svg and .png extensions at what is otherwise the same URL.

e.g. https://img.shields.io/foo/12345.svg shows 404 not found (not rendering, like a well behaved 404), but https://img.shields.io/badge/foo-12345-lightgray.png shows .

### DIFF
--- a/config/test.yml
+++ b/config/test.yml
@@ -5,8 +5,8 @@ public:
 
   rateLimit: false
 
-  redirectUrl: 'http://badge-server.example.com'
+  redirectUrl: 'http://frontend.example.test'
 
-  rasterUrl: 'http://raster.example.com'
+  rasterUrl: 'http://raster.example.test'
 
   handleInternalErrors: false
--- a/core/server/server.js
+++ b/core/server/server.js
@@ -16,10 +16,7 @@ const {
   clearRequestCache,
 } = require('../base-service/legacy-request-handler')
 const { clearRegularUpdateCache } = require('../legacy/regular-update')
-const {
-  staticBadgeUrl,
-  rasterRedirectUrl,
-} = require('../badge-urls/make-badge-url')
+const { rasterRedirectUrl } = require('../badge-urls/make-badge-url')
 const log = require('./log')
 const sysMonitor = require('./monitor')
 const PrometheusMetrics = require('./prometheus-metrics')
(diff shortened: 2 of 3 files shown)


paulmelnikow:
Okay, I updated this to drop the legacy .png static badge route. 🤞

paulmelnikow:
This pull request **introduces 1 alert** when merging a957a98d0a8485b3f931a853b28ccf207763f0dd into 66c7f13e38c76bfdbdb96107bc92e607b3dd5acc - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-68399174e9ee3963ec97f680fdd30219b9dea6bd)

**new alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

chris48s:
Well spotted. Sorry - I should have caught this in review of #3644 :/

Having had a play with it, I think the only way to continue catching the old `/label/message.png` route would be to catch `/my/route.(svg|json|png)$/` for every registered route, redirect for `.png` and serve it ourselves for `.svg` and `.json` and then we could keep the legacy static route.

On balance its probably better to drop the static legacy route. I think we were going to do it at some point. While we're dropping JPEG and GIF is probably the right time to do it..

paulmelnikow:
> Having had a play with it, I think the only way to continue catching the old `/label/message.png` route would be to catch `/my/route.(svg|json|png)$/` for every registered route, redirect for `.png` and serve it ourselves for `.svg` and `.json` and then we could keep the legacy static route.

Yea, that was my conclusion too.

PR #3661, opened 2019-07-07
Auto-annotate live and mock tests e.g. [nodeping]

### LINKED ISSUE #2555: Prefix the titles of mocked tests in the runner
See https://github.com/badges/shields/pull/2520#discussion_r242751074 for a suggestion about printing *[mock]* before all mocked tests so it doesn't have to be added verbatim to the title. The inference can be handled by inspecting [`intercepted`](https://github.com/badges/shields/blob/master/services/icedfrisby-no-nock.js#L9) which we're already doing in [`ServiceTester#toss`](https://github.com/badges/shields/blob/5c7d07f5be6c0b8f96f1a4de6001d007144c3022/services/service-tester.js#L95-L108).

paulmelnikow:
Close #2555

To avoid conflicts with #3652 I haven’t removed all the explicit annotations, though I’ve removed a bunch of them here.

### DIFF
--- a/core/service-test-runner/service-tester.js
+++ b/core/service-test-runner/service-tester.js
@@ -119,6 +119,9 @@ class ServiceTester {
     // eslint-disable-next-line mocha/prefer-arrow-callback
     fn(this.title, function() {
       specs.forEach(spec => {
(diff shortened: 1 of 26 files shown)


paulmelnikow:
Think I've gotten all of those. I also updated the examples in the service testing doc.

Issue #3332, opened 2019-04-17
Migrate GitHub tokens to cloud Redis


paulmelnikow:
As discussed in #1848, let's migrate the Shields runtime data store from the server file system to cloud Redis.

This centralized / hub-and-spoke architecture is simpler and easier-to-reason about than today's peer-to-peer setup, which requires each each server to know about all of the other servers. Bringing a server online requires reconfiguring all of them.

Switching to cloud Redis also makes it possible to move to a PaaS.

A reason to do this _now_ is that we may need to add token rotation to support Gitlab (https://github.com/badges/shields/issues/541#issuecomment-484264221). It would reduce the complexity of supporting Gitlab if we make the migration first.

The code is already in place for this (see #1939, #1906, and #1848), so the only thing left is to operationalize this:

- [x] Choose a cloud provider
- [x] Provision it
- [x] Configure backups
- [x] Load the tokens into the database
- [x] Test the config locally (#3705)
- [x] Push config to a canary server (s0)
- [x] Push config to the remaining servers

paulmelnikow:
Two options from https://github.com/badges/shields/issues/1848#issuecomment-410557894:

- [Compose](https://www.compose.com/pricing#redis), which I've used before, has Redis with persistence starting at $18.50/month
- [RedisLabs](https://redislabs.com/pricing/cloud/) has it for $7/month, or if we want it highly available, $9/month

We could ask for sponsorship, though this isn't a lot of money and we do have it in our budget.

calebcartwright:
I'd be good with either of those. I think Heroku offers Redis-as-a-service too

paulmelnikow:
Interesting! Heroku did agree to give Shields a credit, which we could use for our production hosting, though since Redis is a shareable add-on, we probably could use it for this as well.

One concern is there doesn't seem to be an offline way to access the backup:

https://devcenter.heroku.com/articles/heroku-redis#persistence

calebcartwright:
I think that's only applicable to the hobby (free) tier though.

I haven't looked at any of these providers in detail, but I believe any would be a solid choice. I've not used Compose myself before but i've heard a lot of good things, and you know RedisLabs will do Redis well 😄 If we end up using Heroku for production hosting then there's probably something to be said for using their Redis offering as well

paulmelnikow:
It does say:

> The hobby tier for Heroku Redis does not have any persistence.

Though it also says:

[...]

PR #3668, opened 2019-07-07
Switch to ioredis, and update RedisTokenPersistence to use a set instead of a list

paulmelnikow:
A set is the right way to deal with unique values.

Also add a script for importing a batch of tokens.

Ref #3332


paulmelnikow:
This pull request **introduces 1 alert** when merging c94e06258fb6d035db418706309ccfd6aa11b2b6 into aa135d370fadea22f0f642466a3d17c52ffb00c9 - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-7c3016e4ebdfdf4da84391381a4880ff5e753951)

**new alerts:**

---

This project has automated code review enabled, but doesn't use the <a href="https://github.com/apps/lgtm-com">LGTM GitHub App</a>. Migrate over by <a href="https://github.com/apps/lgtm-com/installations/new">installing the app</a>. Read about the <a href="https://lgtm.com/blog/announcing_github_apps_migration">benefits of migrating to GitHub Apps</a> in the blog.

---

core/token-pooling/redis-token-persistence.integration.js
```diff
@@ -18,17 +17,15 @@ describe('Redis token persistence', function() {
...
   const key = 'tokenPersistenceIntegrationTest'
 
-  let client
+  let redis
   beforeEach(async function() {
-    client = redis.createClient()
-    const del = promisify(client.del).bind(client)
-    await del(key)
+    redis = new Redis()
+    await redis.del(key)
```
calebcartwright: nice to not have to promisify everything here 👍

paulmelnikow: For sure!

core/token-pooling/redis-token-persistence.integration.js
```diff
@@ -84,8 +75,8 @@ describe('Redis token persistence', function() {
         await persistence.initialize()
         await persistence.noteTokenAdded(newToken)
 
-        const savedTokens = await lrange(key, 0, -1)
```
calebcartwright: i'm not familiar enough with these redis clients to follow some of the changes like this. assume it's all correct but would need to read up on those apis to offer any hepful feedback

paulmelnikow: It's better to have more eyes, though we don't necessarily need to block on it. The worst thing that happens is we have bugs that need fixing. I have the tokens backed up so we can roll back the data to this point if need be.

calebcartwright: Works for me. Will give this a quick final look later today and approve

scripts/import-github-tokens.js
```diff
@@ -0,0 +1,50 @@
...
+async function loadTokens() {
+  const contents = await fs.readFile('all_tokens_uniq.json', 'utf8')
+  const tokens = JSON.parse(contents)
+  console.log(`${tokens.length} tokens loaded`)
+  return tokens
+}
+
+function createClient() {
+  const redis = new Redis(process.env.REDIS_URL, {
+    tls: { servername: new URL(process.env.REDIS_URL).hostname },
```
calebcartwright: I take it this script is only intended to be used with compose, since it doesn't have the sample conditional `rediss` check?

scripts/import-github-tokens.js
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    // await load()
+    await list()
+  } catch (e) {
+    console.error(e)
+    process.exit(1)
+  }
+})()
+
+// Appease the linter.
```
calebcartwright: 🤣

scripts/import-github-tokens.js
paulmelnikow: Yea, correct… also I wanted to be super sure I was using TLS ;)

The script is also sort of throwaway, but keeping-around-for-reference.

PR #3683, opened 2019-07-09
add [twitch] user status badge

andyli:
This is a Twitch badge that I want to use, basically just showing whether a user is streaming or not. I'm open to url/param/badge style change suggestions.

Example:
 * when one is streaming: 
 * when one is offline: 

The badge link would be pointing to the Twitch user channel, `https://www.twitch.tv/$user`.

Thank @paulmelnikow for helping me on this :)


services/twitch/twitch.service.js
```diff
@@ -0,0 +1,83 @@
...
+      pattern: ':user',
+    }
+  }
+
+  static get examples() {
+    const preview = this.render({
+      user: 'andyonthewings',
+      isLive: true,
+    })
+    delete preview.link
```
calebcartwright: Nice approach to still be able to use the render function in the examples 👍

services/twitch/twitch.service.js
```diff
@@ -0,0 +1,83 @@
...
+      schema,
+      url: `https://api.twitch.tv/helix/users`,
+      options: {
+        headers,
+        qs: { login: user },
+      },
+    })
+    if (users.data.length < 1) {
+      throw new NotFound({ prettyMessage: 'invalid user' })
+    }
```
calebcartwright: Is it necessary to make these two API calls? What kind of response would we get from the https://api.twitch.tv/helix/streams API if `user` is invalid?

andyli: If `user` does not exist on Twitch, the result would be an empty array, which is the same as when a user is offline.

services/twitch/twitch.service.js
```diff
@@ -0,0 +1,83 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const serverSecrets = require('../../lib/server-secrets')
+const { BaseJsonService, NotFound } = require('..')
+
+const schema = Joi.any()
```
calebcartwright: This should be built out to represent the API responses. If it's really necessary to make two API calls (see my question below) then we can define two schemas. 

Making a few guesses based on the existing code, but suspect you'd want something like these:

```js
const userSchema = Joi.object({
  data: Joi.array().required()
})

const streamsSchema = Joi.object({
  data: Joi.array().required()
})
```

services/twitch/twitch.service.js
calebcartwright: Ah okay, that makes sense. Maybe it would be good to add an inline comment so that future readers of the code have that context?

services/twitch/twitch.service.js
calebcartwright: Actually, if the API responses are the same, then you could get by with a single schema:
```js
const schema = Joi.object({
  data: Joi.array().required()
})
```

andyli:
I've just worked on the comments a bit and define a proper joi schema as suggested.

I also found that we have to request and use a token, such that we can have a higher rate limit. See https://dev.twitch.tv/docs/api/guide/#rate-limits

services/twitch/twitch-base.js
```diff
@@ -0,0 +1,86 @@
...
+        // https://dev.twitch.tv/docs/api/guide/#rate-limits
+        if (err.name === 'InvalidResponse' && err.response.statusCode === 429) {
+          const resetTimestamp = err.response.headers['ratelimit-reset']
+          await sleep(Math.abs(resetTimestamp - Date.now()) + 100)
+          continue
+        }
+
+        // cannot recover
+        throw err
+      }
```
calebcartwright: There's a pretty narrow time window for Shields to return the requested badge, especially in the primary environment where badges are displayed (SCM hosts like GitHub). If it takes Shields more than 2-3 seconds to respond, then the badges won't be displayed on GitHub. 

I've not gone through this in call/sleep/retry flow in detail yet, but at first glance it does strike me as something that could potentially exceed our narrow window to respond to a twitch badge request, especially in combination with the other API calls that have to be made to get the data.

andyli: That's a reasonable concern, though I don't think we can tell whether it's a real issue before we deploy it to the wild.

For the flow here, `_getNewToken()` + `_requestJson()` need < 2 seconds on my machine. I can add a setInterval to do `_getNewToken()` before a token expire, to reduce the chance of `_requestJson()` being blocked by `_getNewToken()`.

For the case of API limit is exceeded, I don't think there is a better solution than the current one of waiting till `ratelimit-reset`. If it ever happen, we should ask Twitch for a higher rate limit. Twitch's [doc](https://dev.twitch.tv/docs/api/guide/#rate-limits) mentioned:

> If you need a higher rate limit, please fill out the form at https://dev.twitch.tv/limit-increase.

calebcartwright: A service class making more than one API call to an upstream service to render a single badge is very, very uncommon within the Shields codebase for the reasons I mentioned above. Having to make a minimum of 3 (and potentially more) API calls in sequence to render a single badge is a concern for me. 

With other services that apply rate limiting, our preferrence is to work with the vendor/upstream service provider to get a token/auth mechanism for Shields with an increased limit (as you noted above, thanks for sharing the doc), or having a rotating pool of tokens, etc. 

IMO we need to pursue one of those alternative approaches, instead of trying to do token management as part of the badge request/response workflow

andyli: Thanks for the explanation. I will add something to avoid having to go though the token flow when serving a badge. However, the token flow still need to be there in `_requestJson()` as a fallback, for the reason described in https://dev.twitch.tv/docs/authentication/#refresh-in-response-to-server-rejection-for-bad-authentication.

andyli:
I've made some improvement with the objective of reducing response delay. The timing info of the Twitch tests can be observed in the log as follows:

```
  Twitch
    [live] Status of andyonthewings
      √        [ GET /status/andyonthewings.json ] (1204ms)
    [live] Status of noopkat
      √
        [ GET /status/noopkat.json ] (199ms)
    [live] Invalid Username Specified
      √
        [ GET /status/invalidusernamethatshouldnotexist.json?label=Follow ] (209ms)


  3 passing (3s)
```

The first request took longer since we need to wait for a token.

calebcartwright:
@andyli - Just wanted to let you know that we've not forgotten about this! Things are pretty busy right now and there's some additional work required on our end to get the auth setup with twitch for our prod environment

andyli:
Hey, it has been a while. Is there anything blocking this?

calebcartwright:
Hi @andyli - thanks for the bump on this. Nothing else needed from your end, but I do want to chat with some of the other maintainers before deploying this. Will try to get back to you as soon as possible.

calebcartwright:
@andyli - we haven't lost track of this, just going to be a bit longer. Thanks for bearing with us!

andyli:
I've removed the user existence check as suggested. I didn't add/modify the badge text, so an invalid user will just give an offline badge. I think it's not too bad since the badge user will probably test the badge by clicking on it. The badge link will tell whether the user name specified is correct or not.

calebcartwright:
> @calebcartwright shall we get this merged?

Let's do it! I've triggered a new redeploy of the review app about of an abundance of caution, but as long as that goes through this should be good to go 👍

paulmelnikow:
@PyvesB Yea, can you send me what needs to be added?

PyvesB:
Will do so in the evening.

PyvesB:
The badge on the staging environment is not working correctly: 

I noticed that the two Twitch variables weren't added to `apps.json` or `config/*` as part if this PR, am I correct in thinking that this is what's missing?

PyvesB:
Preparing new PR to address the above.

PyvesB:
@paulmelnikow, I shared what you'll need via Keybase.

andyli:
Yeah, it's https://twitter.com/andy_li
Thanks!

PyvesB:
Tweet posted: https://twitter.com/Shields_io/status/1174037594851688448

Issue #3685, opened 2019-07-10
License badge colors


chintal:
:clipboard: **Description**

The choice of color for license badges does not seem to be specified anywhere. It would be nice to have that listed in the website somewhere. 

For instance, while most GPL badges seem to be blue in color, the badge for GPLv3+ is grey. 

`https://img.shields.io/pypi/l/fpvgcc.svg`
<img src=https://img.shields.io/pypi/l/fpvgcc.svg>

I can't find any specification to explain why this is so, or if there is some problem with the specific spelling of the license used. 'GPLv3+' seems to be more or less standard on PyPI in general.

calebcartwright:
You can see where the license colors are currently being derived from [here](https://github.com/badges/shields/blob/master/services/licenses.js). 

You can also override the color via the query param if you'd prefer something different:
https://img.shields.io/pypi/l/fpvgcc.svg?color=blue

![](https://img.shields.io/pypi/l/fpvgcc.svg?color=blue)

chintal:
@calebcartwright Thanks. I did know about the color option.

Would a pull request adding GPLv3 and possibly other common notations (I'm considering looking for the python setuptools list) be acceptable, or is the current list kept short on purpose, do you know?

calebcartwright:
The license list (and aliases) evolves frequently, so we'd definitely be receptive to a PR! I suspect that this is mostly likely just a missing aliases vs. a deliberate decision.

balibebas:
![DeepinScreenshot_select-area_20190718191038](https://user-images.githubusercontent.com/51995133/61453052-c0b4c280-a98f-11e9-9bd2-66635e0db03f.png)

Looks like there are other SPDX licenses which also have this color issue. FYI

chintal:
#3686 only fixed it for *GPL licences. SPDX has a pretty long list of licenses compared to the previous version, and it's likely that someone will have to go through and figure out which goes into which color basket, and which have to be supported at all.

I'd be fine with volunteering to do it, but beyond GPL, I can only do an all-or-nothing type thing. I wouldn't want to get into a discussion about whether a license is relevant or not.

calebcartwright:
Thanks everyone. If there's a desire to do a similar enhancement of the colors for other sets of licenses, please open a new issue to track there.

PR #3686, opened 2019-07-10
Updates *GPL licenses to new SPDX ids and adds additional aliases, run service tests for [apm aur bowerlicense cocoapodslicense cpanlicense cran crates ctan dublicense eclipsemarketplacelicense githublicense hexpm npmlicense packagistlicense pypilicense]

chintal:
[SPDX v3.5](https://spdx.org/licenses/) changes the IDs for GPL, LGPL, and AGPL licenses. This PR changes the list of licenses in the spexLicenceIds list to conform to the current standard, while adding the old forms to the aliases list to make sure nothing breaks.

Additionally, the following aliases are added based on the form in which these licenses are represented in [Python's classifiers](https://pypi.org/pypi?%3Aaction=list_classifiers). These old classifiers have informed python users and developers for years, and it is more common to come across one of those formats (ex. `GPLv3+`) in python packages than spdx licence IDs, as reported (by me) in #3685 .


PyvesB:
I believe you forgot to run `npm run prettier` to get the formatting right. 😉

chintal:
Fixed it. Took me a while to swallow how many dependencies npm has.

calebcartwright:
Oops! Thanks for fixing that service name @PyvesB

balibebas:
I cannot attach screenshots here like I can in Gitea using a 📱browser but wanted to mention how pleased I am my AGPL-3.0-or-later badge [shows up with color](https://www.npmjs.com/package/after-dark) now. Thanks, @chintal.

PR #3695, opened 2019-07-11
Colorize [OSSLifecycle] badge

PyvesB:
Rather than always displaying grey, I suggest we introduce colors for the three most used keywords.


paulmelnikow:
Looks good!

It might also be nice to update the examples to pick pretty much any repo except osstracker itself 😉

PyvesB:
I updated one of the examples to add a bit of diversity. 😃

PR #3696, opened 2019-07-11
Update [Sonar] routes: metric first, and server in the query string

paulmelnikow:
This is the preferred way we’re handling server URLs: in the query string.

While we still have a mix of these, I’ve argued this way is better for these reasons:

1. It allows us to make the URLs a bit more standardized across services.
2. It makes the routes unambiguous.
3. It requires less code.

While it introduces a URL-encoding requirement on the parameter, I think these tradeoffs are worth it.


services/sonar/sonar-base.js
```diff
@@ -43,13 +43,17 @@ module.exports = class SonarBase extends BaseJsonService {
     return { userKey: 'sonarqube_token' }
   }
 
-  async fetch({ sonarVersion, protocol, host, component, metricName }) {
-    let qs, url, schema
+  async fetch({ sonarVersion, server, component, metricName }) {
+    if (!server) {
```
PyvesB: Why do we need this check explicitly? Can we not use `optionalUrl.required()` when specifying the query parameter schema?

paulmelnikow: Fixed, thanks!

paulmelnikow:
The sonar routes are unique in putting the component name before the desired metric. It occurs to me that we could align this a bit further with the standard, by moving the component to the end of the route. If we're going to do that, this seems almost certainly the best time. I'm not sure if there is some reason not to do this… maybe there was something related to the host and path handling, or maybe there's a different reason… 🤔

calebcartwright:
> maybe there was something related to the host and path handling, or maybe there's a different reason… 🤔

It struck me as a little odd too the last time I was working on it, but I elected to not change it given how big the rest of the PRs were for Sonar on their own (didn't want to make them even bigger with a bunch of redirects)

calebcartwright:
> While we still have a mix of these, I’ve argued this way is better for these reasons:

I'm 100% on board with this. I suspect the original route structure was because the Sonar URL was always _required_. IIRC, some of the guidelines we've discussed elsewhere around path vs. query parameters is whether or not the parameter was required or optional, and leaning towards query params for optional.

paulmelnikow:
> I'm 100% on board with this. I suspect the original route structure was because the Sonar URL was always _required_. IIRC, some of the guidelines we've discussed elsewhere around path vs. query parameters is whether or not the parameter was required or optional, and leaning towards query params for optional.

Yea, that's absolutely right. Required URLs are an exception to the general rule of "put required things into the path." Similarly with e.g. the dynamic badges, which require not only URLs but also queries, which would be difficult to encode into a URL.

paulmelnikow:
> It struck me as a little odd too the last time I was working on it, but I elected to not change it given how big the rest of the PRs were for Sonar on their own (didn't want to make them even bigger with a bunch of redirects)

Gotcha. Yes… I remember they were huuuuge.

On this plus side, there's a single redirect in this PR which is redirecting _all_ of the current Sonar URLs. It should be easy to update that.

All of the tests and all of the routes in this PR will need to be updated again. I think that's totally worth it though I need a little break from this, so I'll probably let this linger a moment before picking it back up. (Meanwhile if anyone else wants to tackle it, feel free!)

services/sonar/sonar-redirector.service.js
```diff
@@ -4,15 +4,33 @@ const { redirector } = require('..')
...
+  }),
+  redirector({
+    name: 'SonarServerRedirector',
+    category: 'coverage',
+    route: {
+      base: 'sonar',
+      pattern: ':protocol(http|https)/:host(.+)/:component(.+)/:metric',
+    },
+    transformPath: ({ component, metric }) => `/sonar/${component}/${metric}`,
+    transformQueryParams: ({ protocol, host, sonarVersion }) => ({
```
calebcartwright: Do we need `sonarVersion` here? I believe this redirector will only be for the "current" routes (badges using the route created between when the Sonar service was refactored and now) and `sonarVersion` will already be specified as a query param.

I think this helper is only for converting route params to query params

paulmelnikow:
These also use `:` which is Sonar's separator in the component name. Should we change that to a `/`? Technically colons should be encoded as `%3A`.

so

`/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/coverage.svg`

would become

`/sonar/coverage/org.ow2.petals/petals-se-ase.svg?server=http://sonar.petalslink.com`

calebcartwright:
> These also use `:` which is Sonar's separator in the component name. Should we change that to a `/`? Technically colons should be encoded as `%3A`.
> 
> so
> 
> `/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/coverage.svg`
> 
> would become
> 
> `/sonar/coverage/org.ow2.petals/petals-se-ase.svg?server=http://sonar.petalslink.com`

Is the propsoal to split the component name into two different route params? If so I'd probably vote against that. Users can copy/paste their component name from Sonar directly into the modal today, and although it's not a ton of effort to split that up into two different text boxes in the modal window, I'd still prefer to not have to.

paulmelnikow:
Okay, makes sense.

services/sonar/sonar-redirector.service.js
calebcartwright: The current URL folks would be using for one of these Sonar badges with an older SonarQube version would be:
https://img.shields.io/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/coverage.svg?sonarVersion=4.2

(or in this PR review app)
https://shields-staging-pr-3696.herokuapp.com/sonar/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/coverage.svg?sonarVersion=4.2

I believe that's going to result in a redirect of:
(for the review app)
https://shields-staging-pr-3696.herokuapp.com/sonar/org.ow2.petals:petals-se-ase/coverage.svg?server=http%3A%2F%2Fsonar.petalslink.com

As the `sonarVersion` query param will be set to undefined by this redirect.  Note that particular badge will still render since the http://sonar.petalslink.com instance is Sonar 5.4 (which supports the new/modern API schema and the old/legacy API). However, I believe including `sonarVersion` here will break badges/users with legacy Sonar instances

services/sonar/sonar-redirector.service.js
```diff
@@ -4,15 +4,33 @@ const { redirector } = require('..')
...
+      server: `${protocol}://${host}`,
+      sonarVersion,
```
calebcartwright:
```suggestion
```

services/sonar/sonar-redirector.service.js
calebcartwright:
```suggestion
    transformQueryParams: ({ protocol, host }) => ({
```

services/sonar/sonar-redirector.tester.js
```diff
@@ -18,5 +19,27 @@ t.create('sonar version')
...
+  )
+  .expectStatus(301)
+  .expectHeader(
+    'Location',
+    `/sonar/org.ow2.petals:petals-se-ase/alert_status.svg?${queryString.stringify(
+      {
+        server: 'http://sonar.petalslink.com',
+      }
+    )}`
   )
```
calebcartwright:
```suggestion
t.create('sonar host parameter with version')
  .get(
    '/http/sonar.petalslink.com/org.ow2.petals:petals-se-ase/alert_status.svg?sonarVersion=4.2',
    {
      followRedirect: false,
    }
  )
  .expectStatus(301)
  .expectHeader(
    'Location',
    `/sonar/org.ow2.petals:petals-se-ase/alert_status.svg?${queryString.stringify(
      {
        server: 'http://sonar.petalslink.com',
        sonarVersion: '4.2',
      }
    )}`
  )
```

calebcartwright:
Everything looks good, save for the one redirector query param noted above. I _think_ I've got the right suggested changes but a fresh pair of eyes on it would be great 😄 

Ready to approve this PR otehrwise

paulmelnikow:
Even though we're keeping the colons, I think all these routes and tests still need to be updated for the move from

`/sonar/org.ow2.petals/petals-se-ase/coverage.svg?server=http://sonar.petalslink.com`

to

`/sonar/coverage/org.ow2.petals/petals-se-ase.svg?server=http://sonar.petalslink.com`

services/sonar/sonar-redirector.service.js
paulmelnikow: Ah, yes, I believe you're right!

calebcartwright:
> Even though we're keeping the colons, I think all these routes and tests still need to be updated for the move from
> 
> `/sonar/org.ow2.petals/petals-se-ase/coverage.svg?server=http://sonar.petalslink.com`
> 
> to
> 
> `/sonar/coverage/org.ow2.petals/petals-se-ase.svg?server=http://sonar.petalslink.com`

This is for moving the metric name from the end of the route to the beginning to be more consistent with typical shields routes correct? The examples you've listed moves the metric `coverage` in front of the component name, but also seem to replace the `:` with `/` which is confusing me a bit. Did you mean:

From 
`/sonar/org.ow2.petals:petals-se-ase/coverage.svg?server=http://sonar.petalslink.com`

to
`/sonar/coverage/org.ow2.petals:petals-se-ase.svg?server=http://sonar.petalslink.com`

?

calebcartwright:
I'll fix the prettier complaint since it was my suggested change that angered it in the first place 😄

paulmelnikow:
> Did you mean:
> 
> From
> `/sonar/org.ow2.petals:petals-se-ase/coverage.svg?server=http://sonar.petalslink.com`
> 
> to
> `/sonar/coverage/org.ow2.petals:petals-se-ase.svg?server=http://sonar.petalslink.com`
> 
> ?

Yes, exactly that. I must be multitasking a bit toooo much today.

paulmelnikow:
This is updated. Wasn't as bad as I thought! I think I've addressed all the review comments, though if not, please let me know!

services/sonar/sonar-redirector.service.js
```diff
@@ -26,7 +26,7 @@ module.exports = [
       base: 'sonar',
       pattern: ':protocol(http|https)/:host(.+)/:component(.+)/:metric',
     },
-    transformPath: ({ component, metric }) => `/sonar/${component}/${metric}`,
+    transformPath: ({ component, metric }) => `/sonar/${metric}/${component}`,
```
calebcartwright: To avoid redirecting twice, we should make this same change to the above redirector too (line 15)

PR #3701, opened 2019-07-12
Split [Bitbucket] tests

### LINKED ISSUE #3692: Split Bitbucket tests into separate files
Currently there is one file `bitbucket.tester.js` for two services. After #3691 is merged it would be good to split this up for consistency with the other services.

PyvesB:
Fixes #3692.


PyvesB:
I did a quick check and spotted two or three other services in the same case, will raise other PRs.

PR #3704, opened 2019-07-12
Split [pypi] tests

PyvesB:
See https://github.com/badges/shields/pull/3701#issuecomment-511017547. Quite a bit of  splitting to do here.


services/pypi/pypi-djversions.service.js
```diff
@@ -3,7 +3,7 @@
 const PypiBase = require('./pypi-base')
 const { sortDjangoVersions, parseClassifiers } = require('./pypi-helpers')
 
-module.exports = class PypiDjangoVersions extends PypiBase {
+module.exports = class PypiDjVersions extends PypiBase {
```
PyvesB: To be consistent with the route path and avoid having to override the `id` in the `ServiceTester`.

calebcartwright: I think I'd have had a slight preference for making the filenames more explicit to match the class names, but this is fine too

paulmelnikow: For future reference, I don't think the class names need to match the route path, even for `createServiceTester()`. We've quite a few that don't.

PyvesB: I think I'll go with @calebcartwright suggestion and change the filenames instead. But I've got a slight preference for sleep right now, so will do that sometime in the morning! :grin:

Issue #3714, opened 2019-07-14
Finish moving hostnames from routes to query strings


paulmelnikow:
As a number of services were migrated to the modern service architecture, the component of the route that held a URL – the hostname, and sometimes path and protocol – was moved into a single `?server=` query parameter.

This is the argument I made when I updated Sonar in #3696 :

> 1. It allows us to make the URLs a bit more standardized across services.
> 2. It makes the routes unambiguous.
> 3. It requires less code.
>
> While it introduces a URL-encoding requirement on the parameter, I think these tradeoffs are worth it.

There are a few more of these which need updating:

- [x] Discourse
- [x] Jira
- [x] Maven-metadata
- [x] Nexus
- [x] Teamcity
- [x] Twitter
- [x] Website

One of these, Teamcity, is still using the deprecated `format` argument with a regex string. It should be converted to use a `pattern`, probably at the same time.

Anyone is welcome to pick one of these up! You can look at the Sonar PR as a (relatively complicated) example of what this looks like. The services should be updated to reflect the new route and query params, and a redirector added to support backward compatibility with the old route.

If you want to tackle one of these, post a note here first so we don't duplicate effort! Also feel free to post if you have any questions!

PyvesB:
Another simpler example of such migration is [this commit](https://github.com/badges/shields/pull/3633/commits/9d564acf6c8fa5992915a92b31974d5937ed4928) I did in #3633. :wink:

chris48s:
Sorry - diving in late to the party on this one. In general, I think there are 2 different cases to consider when we're thinking about this issue.

For something like BitBucket or NPM there is a canonical instance ( https://bitbucket.org , https://www.npmjs.com/ etc), but you can also host your own. In that case, it makes sense to default to the canonical instance but allow a self-hosted instance to be passed in the query string.

e.g:

https://img.shields.io/npm/v/npm.svg
https://img.shields.io/npm/v/npm.svg?registry_uri=https%3A%2F%2Fregistry.npmjs.com

The other case is something which is inherently self-hosted or where there isn't a centralised or default version. Something like Jenkins or JIRA fits into this box. In that case, you always need to specify the instance, so we've tended to use the `scheme/host/path` style

e.g:

https://img.shields.io/jira/issue/https/issues.apache.org/jira/KAFKA-2896.svg

Are you looking to move to always putting the host/instance in the query string for (even when its always required) now?

paulmelnikow:
[...]

PR #3715, opened 2019-07-14
Add warnings alongside deprecated URL patterns

paulmelnikow:
Ref #3714


services/nexus/nexus.service.js
```diff
@@ -44,6 +44,8 @@ module.exports = class Nexus extends BaseJsonService {
       // API pattern:
       // /nexus/(r|s|<repo-name>)/(http|https)/<nexus.host>[:port][/<entry-path>]/<group>/<artifact>[:k1=v1[:k2=v2[...]]]
       pattern:
+        // Do not base new services on this route pattern.
+        // See https://github.com/badges/shields/issues/3714
         ':repo(r|s|[^/]+)/:scheme(http|https)/:hostAndPath+/:groupId/:artifactId([^/:]+):queryOpt(:.+)?',
```
calebcartwright: This is _precisely_ why 3714 will be so helpful. Some of these patterns get so gnarly, especially when there's a need for a conditional/optional route params

PR #3717, opened 2019-07-14
SVG by default

### LINKED ISSUE #2674: SVG by default

paulmelnikow:
Make cleaner badge URLs by omitting the `.svg` extension.

Closes #2674

Todo:

- [x] Update Cloudflare rules so all extensions are treated the same

After merging:

- [x] Update the frontend badge generation to omit the extension, and document that it can optionally be included


paulmelnikow:
As far as I can tell, Cloudflare is already caching everything under `img.shields.io`, regardless of extension.

paulmelnikow:
Feeling pretty good about this!

Would be great for some folks to take this for a spin on the staging app, just to make sure things seem like they're workin okay.

There's a weird failure in Redmine in this test run. Maybe it's just erroring out. Those tests work fine on my machine and almost certainly not related to this PR. https://circleci.com/gh/badges/shields/61383

calebcartwright:
> There's a weird failure in Redmine in this test run.

Some of the redmine tests have been failing intermittently in the daily test runs too, so it's almost certainly just that

calebcartwright:
Haven't gone through all the code yet, but have spent a little time poking around the review app. It looks like `.png` extensions aren't being forwarded to the proxy server. Is that just because this is a review app, or caused by this change?

core/server/server.js
```diff
@@ -201,7 +195,7 @@ class Server {
     })
 
     if (!rasterUrl) {
```
calebcartwright: I believe this answers the question I asked in the main PR thread regarding raster proxies, correct?

frontend/components/usage.js
```diff
@@ -394,8 +394,19 @@ export default function Usage({ baseUrl }) {
...
+
+      <p>
+        While we highly recommend using SVG, we also support <code>.png</code>
+        for use cases where SVG will not work. These requests should be made to
+        our raster server <code>https://raster.shields.io</code>. For example,
+        the raster equivalent of
+        <code>https://img.shields.io/v/npm/express</code> is
+        <code>https://raster.shields.io/v/npm/express</code>. For backward
+        compatibility, the badge server will redirect <code>.png</code> badges
+        to the raster server.
```
calebcartwright: 👍

services/nexus/nexus.service.js
```diff
@@ -46,7 +46,7 @@ module.exports = class Nexus extends BaseJsonService {
       pattern:
         // Do not base new services on this route pattern.
         // See https://github.com/badges/shields/issues/3714
-        ':repo(r|s|[^/]+)/:scheme(http|https)/:hostAndPath+/:groupId/:artifactId([^/:]+):queryOpt(:.+)?',
+        ':repo(r|s|[^/]+)/:scheme(http|https)/:hostAndPath+/:groupId/:artifactId([^/:]+?):queryOpt(:.+?)?',
```
calebcartwright: I'm getting a `404 badge not found` on these nexus badges when trying to use the optional `queryOpt`, although that's the same behavior in prod 🤔 

It's service patterns like this that I think will really benefit from moving the instance/server host url to a query param, especially when the instance/server can have an optional path segment and the badge route has a trailing optional param

paulmelnikow: Hmm, probably we should add a test. I leaned pretty heavily on the tests for these. Can you give me an example of a URL that should work?

paulmelnikow:
> Haven't gone through all the code yet, but have spent a little time poking around the review app. It looks like `.png` extensions aren't being forwarded to the proxy server. Is that just because this is a review app, or caused by this change?

That's right – the review apps aren't configured for raster. Though you could definitely set `RASTER_URL` on this review app if you want!

services/nexus/nexus.service.js
calebcartwright: I was just using the values we have listed in the example for the `Sonatype Nexus (Query Options)` listing. Altogether, that generates the below badge url (from prod):

https://img.shields.io/nexus/fs-public-snapshots/https/repository.jboss.org/nexus/com.progress.fuse/fusehq/:c=agent-apple-osx:p=tar.gz.svg

paulmelnikow: That URL is 404 in prod too, though if you remove the `/` before `:c=agent-apple-osx` it works:

https://img.shields.io/nexus/fs-public-snapshots/https/repository.jboss.org/nexus/com.progress.fuse/fusehq:c=agent-apple-osx:p=tar.gz.svg

calebcartwright: > That URL is 404 in prod too,

Yup, not related to this PR but worth digging into. Maybe that example listing could be updated in the nexus service to combine the artifactId and query into a single route param?

I'll open a new issue to discuss/track

paulmelnikow: Ahh gotcha. Maybe the optional queryOpt should be moved into a query param, too.

calebcartwright: That sounds like a good fix!

services/static-badge/static-badge.service.js
```diff
@@ -11,7 +11,7 @@ module.exports = class StaticBadge extends BaseStaticService {
   static get route() {
     return {
       base: '',
-      format: '(?::|badge/)((?:[^-]|--)*?)-?((?:[^-]|--)*)-((?:[^-]|--)+)',
+      format: '(?::|badge/)((?:[^-]|--)*?)-?((?:[^-]|--)*)-((?:[^-.]|--)+)',
```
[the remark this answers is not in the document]
calebcartwright: > It would be nice to replace this with a route with a more reasonable pattern, and maybe a second service for the :: form. Or, we could consider redirecting to /badge as the canonical format.

I find that pattern pretty tough on the eyes, so all on board with anything that helps simplify that

services/teamcity/teamcity-build.service.js
```diff
@@ -24,7 +24,7 @@ module.exports = class TeamCityBuild extends TeamCityBase {
       base: 'teamcity',
       // Do not base new services on this route pattern.
       // See https://github.com/badges/shields/issues/3714
-      format: '(?:codebetter|(http|https)/(.+)/(s|e))/([^/]+)',
+      format: '(?:codebetter|(http|https)/(.+)/(s|e))/([^/]+?)',
```
calebcartwright: As an FYI for anyone looking at this in the review app, the badge example listings incorrectly include an `https` prefix in the `hostAndPath` param (pretty sure I was the one that caused that 😉). If you copy/paste the example values make sure to trim that leading `https`. 

I will open a PR to fix the example listings

services/wercker/wercker.service.js
```diff
@@ -23,7 +23,7 @@ module.exports = class Wercker extends BaseJsonService {
     return {
       base: 'wercker',
       format:
-        '(?:(?:ci/)([a-fA-F0-9]{24})|(?:build|ci)/([^/]+/[^/]+))(?:/(.+))?',
+        '(?:(?:ci/)([a-fA-F0-9]{24})|(?:build|ci)/([^/]+/[^/]+?))(?:/(.+?))?',
```
calebcartwright: I'm getting a `404 | badge not found` with the wercker ci badge (no branch) in the review app. Not sure what the cause is

https://shields-staging-pr-3717.herokuapp.com/wercker/ci/59e33c8e982fc615500b357

paulmelnikow: This is 404 too:

https://img.shields.io/wercker/ci/59e33c8e982fc615500b357.svg

calebcartwright: well that's a bummer for the wercker badges, but at least not related to this PR 👍

calebcartwright:
Not sure when you'd like to merge and deploy this, but changes LGTM

paulmelnikow:
Seems to be working well on the review app, so unless other folks want to test it, we may as well go for it!

services/wercker/wercker.service.js
chris48s: its because `59e33c8e982fc615500b357` is only 23 characters long. https://img.shields.io/wercker/ci/559e33c8e982fc615500b357.svg does work

calebcartwright: Ha! User error on my part not being able to correctly copy and paste 🤦‍♂ 

Good spot

PR #3719, opened 2019-07-15
Rewrite CopiedContentIndicator using hooks

paulmelnikow:


frontend/components/customizer/copied-content-indicator.js
```diff
@@ -29,39 +29,34 @@ const PosedContentContainer = posed(ContentContainer)({
 
 // When `trigger()` is called, render copied content that floats up, then
 // disappears.
-export default class CopiedContentIndicator extends React.Component {
```
calebcartwright: Where in the UI does this appear? Is it in the badge modal window, on the `Copy badge url` button?

paulmelnikow: It’s the “Copied” text that appears and floats up after clicking “Get Badge URL.”

calebcartwright: Gotcha thanks 👍 The code LGTM, just wanted to check it out in the review app too (and know where to look)

paulmelnikow: I'm curious if, as someone who is new to React, if the new code is more understandable than the old code. Or, if it feels in any way more connected to the behavior of the component.

To hook or not to hook, (that is) the question of 2019.

frontend/components/customizer/copied-content-indicator.js
```diff
@@ -29,39 +29,34 @@ const PosedContentContainer = posed(ContentContainer)({
...
+      <PosedContentContainer onPoseComplete={handlePoseComplete} pose={pose}>
+        {copiedContent}
+      </PosedContentContainer>
+      {children}
+    </ContentAnchor>
+  )
 }
+// eslint-disable-next-line no-func-assign
+CopiedContentIndicator = forwardRef(CopiedContentIndicator)
 CopiedContentIndicator.propTypes = {
```
paulmelnikow: Aw, I meant to remove the prop-type. (I'd keep them around in the interim but I don't think they'll be long-lived, and TBH beyond documentation, they are not that useful.)

frontend/components/customizer/copied-content-indicator.js
calebcartwright: TBH, not much of a difference in grokability for me between the new and the old. _Perhaps_ something like `useImperativeHandle` looks/feels a bit more like React tribal knowledge. 

This style doesn't make it any _less_ understandable, so if there's value in this approach vs. the before, and/or if folks just prefer this style then 👍 from my perspective

chris48s: Reading over this PR, both styles are fairly hard work for me, but I find the older class-based style is a bit easier as it makes the component's relationship to the state a bit clearer than in the excapsulated imperative style. Hope your plan works out :crossed_fingers:

paulmelnikow: Hope so! 😝

PR #3721, opened 2019-07-15
Rewrite PathBuilder using hooks

paulmelnikow:


frontend/components/customizer/path-builder.js
```diff
@@ -68,25 +68,41 @@ const NamedParamCaption = styled(BuilderCaption)`
...
+export default function PathBuilder({
+  pattern,
+  exampleParams,
+  onChange,
+  isPrefilled,
+}) {
+  const [tokens] = useState(() => pathToRegexp.parse(pattern))
+  const [namedParams, setNamedParams] = useState(() => {
+    if (isPrefilled) {
+      return exampleParams
```
calebcartwright: Before, after this general block (if I follow correctly 😄)  `this.state.namedParams` was being set to `exampleParams`. Does that still get achieved with this approach, both if `isPrefilled` is true and if false? Am I ask a crazy/invalid question?

paulmelnikow: As far as I can tell, the old and new code do the same thing, which is to set namedParams to `exampleParams` when `isPrefilled` is true, and otherwise to an object containing an empty string for each named `token`.

The `isPrefilled` case is for badge suggestions, if you want to give it a shot. The logic _looks_ right to me, though I'm not sure I ran it.

calebcartwright: I wasn't sure if declaring a local `namedParams` on line 105 would impact how everything bubbles up, but will definitely take a look in the UI. Will be much easier for me to poke around the review app

paulmelnikow: Oh yea, I see, it's shadowing the other local. Could rename it to `result` for clarity.

paulmelnikow: I ended up replacing it with a conditional expression and `.reduce()`. Seems to work okay for regular badges and for suggestions.

calebcartwright:
Side note/observation on the suggested badges feature: if you proceed with an incomplete url (like https://github.com/badges instead of https://github.com/badges/shields), then the page goes blank (in chrome on windows ATM)

This happens in prod today too, so unrelated to this PR

paulmelnikow:
> Side note/observation on the suggested badges feature: if you proceed with an incomplete url (like https://github.com/badges instead of https://github.com/badges/shields), then the page goes blank (in chrome on windows ATM)
> 
> This happens in prod today too, so unrelated to this PR

Yikes, can reproduce. We can't use Joi in the frontend though probably should either add some schema validation for that response or a try block and log the error for this condition.

The underlying problem is a bad response from the suggest endpoint:

<details>

```
{
  "suggestions": [
    {
      "title": "GitHub issues",
      "link": "https://github.com/badges/undefined/issues",
      "example": {
        "pattern": "/github/issues/:user/:repo",
        "namedParams": {
          "user": "badges"
        },
        "queryParams": {}
      }
    },
    {
      "title": "GitHub forks",
      "link": "https://github.com/badges/undefined/network",
      "example": {
        "pattern": "/github/forks/:user/:repo",
        "namedParams": {
          "user": "badges"
        },
        "queryParams": {}
      }
    },
    {
      "title": "GitHub stars",
      "link": "https://github.com/badges/undefined/stargazers",
      "example": {
        "pattern": "/github/stars/:user/:repo",
        "namedParams": {
          "user": "badges"
        },
        "queryParams": {}
      }
    },
    {
      "title": "GitHub license",
      "link": "https://github.com/badges/undefined",
      "example": {
        "pattern": "/github/license/:user/:repo",
        "namedParams": {
          "user": "badges"
        },
        "queryParams": {}
      }
    },
    {
      "title": "Twitter",
      "link": "https://twitter.com/intent/tweet?text=Wow:&url=https%3A%2F%2Fgithub.com%2Fbadges",
      "example": {
        "pattern": "/twitter/url/:protocol(https|http)/:hostAndPath+",
        "namedParams": {
          "protocol": "https",
          "hostAndPath": "github.com/badges"
        },
        "queryParams": {}
      },
      "preview": {
        "style": "social"
      }
    }
  ]
}
```

</details>

Issue #3657, opened 2019-07-06
Proposal: Adopt TypeScript in the frontend


paulmelnikow:
I've been doing a lot of frontend work in TypeScript these days, and it's clear that it's dramatically easier to make behavior changes with types in place. Making major changes to the Shields frontend is always very effortful – and the support of types should make it both faster to work on, and a lot more fun.

Although there is a learning curve, I'm convinced the tradeoff is worth it. Here's my reasoning:

1. We haven't attracted a huge number of contributors to our frontend.
2. A type system will speed up experienced frontend developers, not slow them down.
3. We're more likely to attract good frontend developers if the developer experience is good. Types make that experience way better.
4. As code changes, PropTypes take even longer to maintain than TypeScript types, and they provide much less value.

TypeScript plays well with the tools we're already using: React, Prettier, ESLint, and styled-components. Unsurprisingly, it works great in VS Code, which might be the most popular editor for frontend work these days.

It also looks like [it works with Gatsby](https://www.gatsbyjs.org/packages/gatsby-plugin-typescript/) too, though I don't have personal experience using it with Gatsby. I expect most of our frontend libraries (like react-pose and react-select) have types, though there may be bit of work writing type defs for some of the smaller ones.

On the server side I still think we want to keep things tipped toward welcoming new contributors who are experienced programmers but not necessarily JS developers. I think our server framework provides a good experience, even for people who aren't familiar with JavaScript, and think @chris48s's work adding documentation with type annotations to our core library is a better way to help them.

calebcartwright:
SGTM. I'm a big fan of TypeScript, but like everything in life, there's pros and cons associated with it that should be considered in context.

The rationale you've outlined above makes a strong case that the pros outweigh any cons in the context of the frontend IMO, so I'm on board!

PyvesB:
I'm not familiar with Typescript and barely ever touch the frontend code anyway. No strong opinion either way.

paulmelnikow:
This is done!

PR #3722, opened 2019-07-15
Initial adoption of TypeScript in the frontend

paulmelnikow:
In this PR I’ve started adding types to the frontend. For the most part I started with the leaf nodes, and backed out if things got tricky.

A few notes about typings:

- I opened a PR to fix typings on node-walkdir: soldair/node-walkdir#39.
- Sazerac needs typings added for `forCases`.
- `@mapbox/react-click-to-select` does not have typings.

Ref #3657


package.json
```diff
@@ -147,7 +148,13 @@
     "@babel/preset-env": "^7.5.2",
     "@babel/register": "7.4.4",
     "@mapbox/react-click-to-select": "^2.2.0",
-    "babel-eslint": "^10.0.0",
+    "@types/chai": "^4.1.7",
+    "@types/enzyme": "^3.10.2",
+    "@types/mocha": "^5.2.7",
+    "@types/react-helmet": "^5.0.8",
+    "@types/styled-components": "4.1.8",
```
paulmelnikow: This is pinned to an old version due to DefinitelyTyped/DefinitelyTyped#33311.

.eslintrc-frontend.yml
```diff
@@ -26,3 +28,9 @@ rules:
   react/jsx-sort-props: 'error'
   react-hooks/rules-of-hooks: 'error'
   react-hooks/exhaustive-deps: 'error'
+
+  '@typescript-eslint/no-unused-vars': ['error', { 'args': 'none' }]
+
+  # Argh.
```
calebcartwright: :rofl:

package.json
```diff
@@ -74,15 +74,15 @@
     "coverage:report:generate": "nyc report",
     "coverage:report:open": "open-cli coverage/lcov-report/index.html",
     "coverage:report": "run-s --silent coverage:report:generate coverage:report:open",
-    "lint": "eslint \"**/*.js\"",
-    "prettier": "prettier --write \"**/*.@(js|md|json|yml)\"",
-    "prettier:check": "prettier --check \"**/*.@(js|md|json|yml)\"",
+    "lint": "eslint \"**/*.@(js|ts|tsx)\"",
```
calebcartwright: I've never used eslint for typescript source before, but now that I know eslint is capable of it, do you know any pros/cons of using the eslint w/the typescript parser vs. tslint? 

As long as we're linting the typescript source, I don't have a strong preference on the tooling. More a question of curiosity

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "include": ["frontend/**/*"],
+  "exclude": [],
+  "compilerOptions": {
+    "target": "esnext",
+    "module": "commonjs",
+    "declaration": true,
+    "outDir": "lib",
```
calebcartwright: Is this where we want the transpiled output to go?

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
calebcartwright: Do we need to include source maps to get accurate code coverage on the typescript source from nyc?

package.json
calebcartwright: I suppose another question would be whether we're linting anything twice:  first pass eslint against the .ts/.tsx files, and then a second pass (at some point) against the emitted .js/.jsx files

.eslintrc-frontend.yml
paulmelnikow: Comment duly copied and pasted from
some other project.

tsconfig.json
paulmelnikow: No idea! I haven’t yet the pleasure of running coverage with TS.

I can say that in my experience coverage is relatively less important, especially compared to the “does this code run or not” stage we’ve currently reached in the frontend.

tsconfig.json
paulmelnikow: With this setup the compiled code is never emitted and this setting is ignored unless you were manually to run tsc. Since we’re still using lib for a few things, maybe we should change this to “null” or “trash” or “tsc_output.”

package.json
paulmelnikow: ESLint works well on TypeScript. (And sort of off topic, but Prettier works _extraordinarily_ well with TypeScript.) There's been good velocity of development of the TS–ESLint tooling in recent months and I'm pretty sure Create React App is using ESLint for TS as well as JS. I like that it's something I know (I've implemented ESLint rules) and that we don't have to adopt yet another tool.

As nothing is being emitted, it's the TS source code that is linted by this change, not the ephemerally generated JS.

tsconfig.json
paulmelnikow: Not that I'm necessarily arguing _against_ testing, just that I think the wins from typing are big, and while I think most behaviors in the frontend will still need to be tested (and manually), TS React codebases are much less at risk of breaking during refactors or unrelated changes. I've never had a React app with a stable enough structure that the test-writing paid off, especially compared to the benefit from compile-time type checking.

tsconfig.json
paulmelnikow: I updated this to `unused_ts_output` which will hopefully convey the fact that it can be safely deleted.

tsconfig.json
calebcartwright: Understood. I only mentioned it since we're currently capturing and publishing code coverage, including the frontend, to coveralls. 

We don't need to block, but I would like to see that we continue to produce coverage for the frontend (and can hopefully avoid our project-wide coverage of 93% dropping 😄)

tsconfig.json
calebcartwright: I'm a little confused then, probably because of lack of familiarity with this portion of the Shields codebase.

I assume we must have _something_ that's transpiling our frontend source (which will be TS after this is merged) into some browser-compatible JS. If we're not using the TS compiler to do that, is that being done via something else (babel perhaps?)

paulmelnikow: Yes, you're right. This happens in three places:

- `gatsby dev` (handled by `gatsby-plugin-typescript` which is configured here: https://github.com/badges/shields/pull/3722/files#diff-b9e136416b90437fa1dac910280b45fc)
- `gatsby build` (same)
- Running tests, where it's handled by `ts-mocha`. I'm not sure which tools it might be using under the hood.

The dev output gets cached by Gatsby (I think in `.cache`) and the generated build goes into `public/` and generates [a tree that looks like this](https://github.com/badges/shields/tree/gh-pages). We definitely don't lint or run coverage on either of those. And since we never invoke `tsc`, the transpiled output is never written to the configured TypeScript `outDir`.

Does that make sense?

calebcartwright: > Does that make sense?

💯 indeed, thanks so much!

tsconfig.json
paulmelnikow: I think what's happening right now is that no coverage is reported for the TS files. I just tried for a few minutes but didn't manage to get it working. 

Could be worth opening an issue with https://github.com/piotrwitek/ts-mocha which is the way we're running tests (the easiest way I've found) but doesn't have any mention of using it with istanbul/nyc.

Alternatively I found these resources, not using `ts-mocha`. Probably the place to start on those would be to remove `ts-mocha` and try to get the tests running again.

- https://azimi.me/2016/09/30/nyc-mocha-typescript.1.html
- https://istanbul.js.org/docs/tutorials/typescript/
- https://github.com/john-cheesman/ts-mocha-chai-nyc

You're saying we don't need to block… should we open a new issue?

calebcartwright: > You're saying we don't need to block… should we open a new issue?

Will defer to you. If there's a big value add in getting TS for the frontend sooner rather than later, then I don't want to hold that up while we try to figure out the coverage story. 

I've always used istanbul/nyc with mocha, but also transpile beforehand via tsc. I'll check out this branch locally and poke around the code coverage/tests as well.

paulmelnikow: Yea, I'm itching to use this 😁

My inclination would be to drop frontend coverage until it feels pressing, in part because I continue to be surprised at how difficult it is to get babel/ts/mocha/istanbul happy with each other. I spent a surprising amount of time getting it working in the first place… it's frustrating. Coverage should be easier to accomplish than it is.

I'm also happy to leave it in place for the part that's still in .js and hopefully get around to fixing it sooner rather than later. Or to wait a little bit if you want to poke around!

calebcartwright: >My inclination would be to drop frontend coverage until it feels pressing

> I'm also happy to leave it in place for the part that's still in .js and hopefully get around to fixing it sooner rather than later. Or to wait a little bit if you want to poke around!

I'm sold! Let's deal with coverage in a follow on. I'll open a new issue after this gets merged

frontend/components/common.tsx
```diff
@@ -1,5 +1,4 @@
 import React from 'react'
-import PropTypes from 'prop-types'
 import styled, { css, createGlobalStyle } from 'styled-components'
```
calebcartwright: Question on the file extensions for my own edification: seems like a lot of these are going from `.js` to `.tsx`, but I suppose I was expecting `.jsx` --> `.tsx`. Are there some React-related changes we're adding here in addition to the TS switch? Or is `.tsx` extension just the way to go for TS files in React?

paulmelnikow: > Or is `.tsx` extension just the way to go for TS files in React?

^^ Yes, TS files which include JSX need to be `.tsx`.

Once upon a time JS with JSX were `.jsx` but no longer. I'm not sure why TS has kept it this way.

frontend/lib/generate-image-markup.ts
```diff
@@ -43,15 +47,21 @@ function quoteAsciiDocAttribute(attr) {
 }
 
 // lodash.mapvalues is huge!
-function mapValues(obj, iteratee) {
-  const result = {}
+function mapValues(
+  obj: { [k: string]: string | null },
+  iteratee: (value: string | null) => string
+): { [k: string]: string } {
```
calebcartwright: Fun types!

frontend/lib/generate-image-markup.ts
```diff
@@ -1,8 +1,8 @@
-export function bareLink(badgeUrl, link, title = '') {
+export function bareLink(badgeUrl: string, link?: string, title = '') {
```
calebcartwright: Do we really still need `link` and `title` parameters? Doesn't look like they were ever used.

paulmelnikow: It gives these functions a common signature, which I think is helpful.

calebcartwright: Ah so consistency for the callers

paulmelnikow: Right, and now that types are checked it makes that `generateMarkup` function work.

paulmelnikow:
Thanks for reviewing, both counts!

PR #3738, opened 2019-07-18
Rewrite Main using hooks

paulmelnikow:


frontend/components/main.js
```diff
@@ -29,55 +29,30 @@ const AppContainer = styled(BaseFont)`
   text-align: center;
 `
 
-export default class Main extends React.Component {
-  constructor(props) {
-    super(props)
+export default function Main({ pageContext }) {
+  const [searchIsInProgress, setSearchIsInProgress] = useState(false)
```
calebcartwright: Where do these `set*` functions come from? Is it a React thing?

paulmelnikow: Yea, it's what `useState` does:

https://reactjs.org/docs/hooks-state.html

Basically it's a way to persist stuff between multiple calls to the function across the lifetime of a component, and to automatically trigger a re-render after the values change.

PR #3740, opened 2019-07-18
Rewrite QueryStringBuilder using hooks

paulmelnikow:


frontend/components/customizer/query-string-builder.js
```diff
@@ -36,255 +36,246 @@ function getBadgeOption(name) {
...
+  }
+}
+
+function BadgeOption({ name, value, handleBadgeOptionChange }) {
+  const {
+    label = humanizeString(name),
+    shieldsDefaultValue: hasShieldsDefaultValue,
+  } = getBadgeOption(name)
+  return (
+    <tr>
```
calebcartwright: Apologies as I think I've asked this same question before in other threads 😄, but we no longer need the `key={name}` attribute with this new approach right?

frontend/components/customizer/query-string-builder.js
```diff
@@ -36,255 +36,246 @@ function getBadgeOption(name) {
...
+function ServiceQueryParam({
+  name,
+  value,
+  exampleValue,
+  isStringParam,
+  stringParamCount,
+  handleServiceQueryParamChange,
+}) {
+  return (
+    <tr>
```
calebcartwright: Apologies as I think I've asked this same question before in other threads 😄, but we no longer need the `key={name}` attribute with this new approach right?

frontend/components/customizer/query-string-builder.js
```diff
@@ -36,255 +36,246 @@ function getBadgeOption(name) {
...
+        ))}
+      </select>
+    )
+  } else {
+    return (
+      <QueryParamInput
+        name={name}
+        onChange={handleBadgeOptionChange}
+        type="text"
+        value={value}
```
calebcartwright: What's the difference between `checked={value}` (before) and this?

paulmelnikow: `checked` is used for the state of a check box. This is a text input, so I think it should always have been using `value` instead of `checked`.

frontend/components/customizer/query-string-builder.js
paulmelnikow: The key is used to disambiguate components at the same level. Since this returns only one component, it doesn't have any effect. The `<ServiceQueryParam>` invocation sets a key instead, as that's where the array of components is being rendered.

PR #3742, opened 2019-07-18
More TypeScript in the frontend

paulmelnikow:
A couple issues:

- [x] Lint config for `.d.ts` files outside of `frontend/`
- [ ] ~~Fix import of svg in tests (or find a different way to get the types right)~~


paulmelnikow:
I'm trying to get my head around what needs to happen to fix this failing test. [babel-plugin-inline-react-svg](https://github.com/airbnb/babel-plugin-inline-react-svg) is supposed to turn the logo into a React component, and that seems to be working correctly in Gatsby. I added a `declare module` for `.svg` files which appeased the TypeScript type checker. So now it's just ts-mocha that's not working.

What I'm not clear on is:

- Whether the Babel plugin is actually running on the svg file
- The best way to fix this

paulmelnikow:
Since the tests of the frontend components are just testing renders and making the most basic assertion about contents, the main value they provide is checking that the code still runs after something has changed.

TypeScript checks for an absence of type errors – which is close enough to "this code runs without errors" that I suggest we might remove the rest of these component tests for now, rather than pour more effort into maintaining them.

I'm not opposed to the idea of having tests for the frontend code. I do think there is value in testing the helper functions, and think we should keep these.

Though personally, I would rather spend time improving the functionality of the site – literally, changing its behavior – rather in than testing today's behavior, and I'm not yet convinced that writing tests of the components actually accelerates development of features.

calebcartwright:
I'm still working my way through some of the files in here, but could you elaborate a bit more on the removal of the tests?

Are they making the switch to TypeScript problematic? Hypothetically, if we weren't moving the frontend to TypeScript, do you still think it would be best to remove them?

I don't have strong feelings one way or the other on those specific tests, but do want to make sure I fully understand the rationale.

frontend/components/category-headings.tsx
```diff
@@ -1,24 +1,26 @@
 import React from 'react'
-import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { Link } from 'gatsby'
 import { H3 } from './common'
 
-export function CategoryHeading({ category: { id, name } }) {
+export interface Category {
```
calebcartwright: More of a strategy question, but is the thinking with the switch to TS for any type (interfaces, classes, aliases, etc.) that's only used in one file (presumably like `Category` here) to define that within the file, and then for a type used in multiple places that we'll define that in something like `common.tsx`?

frontend/types/mapbox__react-click-to-select/index.d.ts
```diff
@@ -0,0 +1,11 @@
+declare module '@mapbox/react-click-to-select' {
```
calebcartwright: does the double underscore in the file name have any particular meaning (either in TypeScript and/or React)?

frontend/lib/service-definitions/index.ts
```diff
@@ -0,0 +1,58 @@
...
+    style?: string
+    namedLogo?: string
+  }
+  keywords: string[]
+  documentation?: {
+    __html: string
+  }
+}
+
+export interface ModernRoute {
```
calebcartwright: Thoughts on just calling this `Route` since the other has the `Legacy` prefix?

frontend/components/usage.tsx
```diff
@@ -111,16 +117,19 @@ function StyleExamples({ baseUrl }) {
...
   const [first, ...rest] = shieldsLogos
-  return [renderLogo(first)].concat(
-    rest.reduce((result, logo) => result.concat([', ', renderLogo(logo)]), [])
+  const result = ([renderLogo(first)] as (JSX.Element | string)[]).concat(
+    rest.reduce(
+      (result, logo) => result.concat([', ', renderLogo(logo)]),
+      [] as (JSX.Element | string)[]
+    )
   )
+  return <>{result}</>
```
calebcartwright: I think I am starting to see what you mean with some of the tests we had for these components. Perhaps it's just the diff and/or the addition of the types, but seems like there's a bit of logic involved here that I personally wouldn't mind having a test for.

However, I also see that the existing spec file we had for this component wasn't testing this at all (at least AFAICT)

frontend/components/snippet.tsx
```diff
@@ -15,22 +18,39 @@ const CodeContainer = styled.span`
       max-width: 40%;
       overflow: hidden;
       text-overflow: ellipsis;
-    `};
```
calebcartwright: 👀 surprised our linter didn't catch this trailing semi before

frontend/components/footer.tsx
```diff
@@ -56,7 +51,6 @@ export default function Footer({ baseUrl }) {
         />{' '}
         {}
         <object
-          alt="chat on Discord"
```
calebcartwright: What's the thinking behind removing these?

paulmelnikow: It throws an error in TypeScript. I think the error is correct, since [`<object>` doesn't seem to support `alt` attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object).

frontend/components/snippet.tsx
paulmelnikow: Yea, not sure! Maybe because it's inside a (`styled`) tagged template literal?

frontend/components/usage.tsx
paulmelnikow: Across all the components, the existing tests were checking that the code could run, and running a basic confidence check for a substring in the result. None of them were testing logic, per se.

To take this component as an example: there's a lot of stuff in it, though it's generating what is essentially static content. When this code changes we have to look at the page to test it, and I don't think it's going to spontaneously break unless it's something related to common styling, which tests aren't likely to catch. If this code were to break it would likely be type-related (e.g. a component changing which props it accepts) and the type checker will now catch that.

We do have components with a significant amount of logic. It might be useful to have tests for that logic, though I'm more interested in making the frontend more impressive and the behaviors better. I think the effort is better put into doing that than in writing tests, and I'm also not sure that the tests will help that process.

frontend/lib/service-definitions/index.ts
paulmelnikow: Works for me!

paulmelnikow:
> I'm still working my way through some of the files in here, but could you elaborate a bit more on the removal of the tests?
> 
> Are they making the switch to TypeScript problematic? Hypothetically, if we weren't moving the frontend to TypeScript, do you still think it would be best to remove them?

Yes, there's a tooling issue relating to babel-plugin-inline-react-svg. I'm not sure the exact cause but I wrote a bit about it in https://github.com/badges/shields/pull/3742#issuecomment-513327029. I spent a while  trying to solve it and I'm not sure it's creating a lot of value so I'm not really interested in digging in further, though if anyone else feels like fixing it, I wouldn't object to keeping these tests around.

I posted my rationale on removing the tests here https://github.com/badges/shields/pull/3742#issuecomment-515710380 and sort of reiterated it in https://github.com/badges/shields/pull/3742#discussion_r314193804.

paulmelnikow:
Thanks for reviewing @calebcartwright!

frontend/components/usage.tsx
calebcartwright: Thanks for the context. Also just to clarify, my comment was not intended to be a suggestion (subtle or otherwise) that we needed a test here (or anywhere else) as part of this work to switch over to TypeScript. It was really just me thinking "out loud" 😄 

I still have not made the time to dig into this portion of the codebase to get familiar with it and find ways to contribute. When I am working with new code and/or refactoring, I like having an automated test to give me confidence that everything is functioning correctly and that I didn't break anything.

My thought here while looking at some of these changes was that if I had been the one working on them, that a test would have helped me personally in my dev loop, especially as someone not familiar with React/this part of our codebase.

frontend/types/mapbox__react-click-to-select/index.d.ts
paulmelnikow: Yes, the double underscore is the convention for a file with type defs for a scoped package. If it were named differently, TypeScript would not find the types.

paulmelnikow:
Cool, thanks for reviewing and testing. Glad to be getting this merged!

Issue #3635, opened 2019-07-03
Remove the popout badge formats?


paulmelnikow:
I think the popout badge formats look cool, and that it is possible to arrange them side-by-side in readmes in a way that looks good, and that they could even be used to convey useful information like distinguishing one platform for another (you could argue that logo + text does this better than text).

However, I don't think they are used very often. We've also gotten a complaint about them from one of the founders of shields, on the basis that consistency is a goal of the project, and offering lots of formats works against that. (#1478)

I certainly don't feel very strongly about them, one way or another.

However, I was in the process of working on #2428 (which I lost steam on a couple months ago) and wasn't personally feeling it was necessarily worth the effort to maintain these formats. So I thought I would raise the question. Would anyone be upset if we dropped these formats?

I don't think we have stats on the badge formats, though perhaps we should turn that on for a few days to inform this question.

calebcartwright:
> Would anyone be upset if we dropped these formats

I assume this question is directed to everyone (Shields end users, contributors, maintainers, etc.) correct? I don't feel strongly either. I don't use the popout format myself.

paulmelnikow:
> Shields end users, contributors, maintainers, etc.

Yes, to everyone. 😁 

I mean, if there's one person who wants to use it that probably is not enough to justify keeping it. It's not terribly hard to put something like RunKit in front of the Shields JSON badge, for a small number of people who are really passionate. But if I'm wrong and it's actually really popular, or really important to a significant group of maintainers or contributors, that could be reason to keep it.

PyvesB:
I don't have any strong opinions on this myself. Do we have any precise usage statistics? I've pinned the issue in an attempt to gather more feedback from our users.

paulmelnikow:
I agree it would be helpful to collect usage statistics.

chris48s:
One thing to consider when you collect stats:
For some reason (copy+pasteing, I guess), quite a lot of READMEs have badges that are using the `popout` or `popout-square` styles are using them without a logo. For example all these READMEs are using a shields badge with `style=popout-square` and no logo:

* https://github.com/daniu310/python/blob/a10aafe24787978ec3a676bd6d460655db424958/themes/theme-ad-master/README.md
* https://github.com/Nuglif/NGLoggerKit/blob/12dd5acbe8b78b2afd38c24b6eca379246846be4/README.md
* https://github.com/maxkrasnov/react-multilevel-select/blob/babfe26af46893113f1911893b9ab0e10c06cb42/README.md


[...]

PR #3747, opened 2019-07-21
Remove popout styles

PyvesB:
Removes the popout styles, as discussed in #3635.

I've left the `logoPosition` code for now, we can delete in a few weeks when we're confident no one has complained about the popout removal and we go ahead with closing #3635 for good.


PyvesB:
Good call, that case had slipped out of my mind!

Issue #1550, opened 2018-03-06
GitHub dependency graph


Alberthaff:
It would be great to have a shield displaying stats from GitHub's dependency graph (dependent repos / packages):

https://github.com/badges/shields/network/dependents

chris48s:
Thanks for the suggestion. Our data for the GitHub badges comes from their API. I don't think GitHub surface this information via an API endpoint yet, do they?

Alberthaff:
Yes, you are right.
Unfortunately, their API doesn't seem to have an endpoint to retrieve the dependency graph.

paulmelnikow:
Is there another service that could provide this? For example, we use https://libraries.io/ for bower. Would they have a relevant endpoint?

chris48s:
If the project is a package indexed by libraries.io, the [project-dependencies](https://libraries.io/api/#project-dependencies), [project-dependents](https://libraries.io/api/#project-dependents) (these are both about other _packages_ that your package depends on, or is dependent on) and the [dependent-repositories](https://libraries.io/api/#project-dependent-repositories) endpoint (which is about _repositories_ that depend on your package) are all relevant in this problem domain but won't produce the same numbers as the GH dependency graph. See https://github.com/librariesio/libraries.io/issues/2005

paulmelnikow:
This came up again in #1858, where @sindresorhus pointed out there is a GraphQL endpoint for the dependency graph: https://developer.github.com/changes/2018-04-24-preview-dependency-graph-and-vulnerability-hooks/

However:

> That endpoint doesn't seem to provide access to the dependents number, only the dependencies number.
> 
> IIRC scraping github violates their ToS so I'm not sure the best way forward here.

As @chris48s points out above, Libraries.io has a different number available, but it doesn't match Github. I added a badge for that in #1895.

chris48s:
If anyone is interested in working on this, there is now a `GithubAuthV4Service` base class for working with the Github GraphQL API. https://github.com/badges/shields/blob/master/services/github/github-forks.service.js is an example of another bad that uses it.

nedbat:
I haven't tried it, but does this GraphQL object provide the data needed?  https://docs.github.com/en/graphql/reference/objects#dependencygraphmanifest

GitHub is now displaying the number of dependents on the front page of a repo as "Used by", which will make more people want a badge for it:
[...]

PR #3763, opened 2019-07-23
Add BaseGraphqlService, support [github] V4 API

chris48s:
This bit of work sprouted off from #3610 but I've decided to do it in its own PR because this will require some review to get right and trying to load all this into a PR that is already a ~500 line diff is going to make that annoying. Lets get this bit right then I'll go back, rebase #3610 onto this work and pick that up.

Main things going on here are:

* Add a generic `BaseGraphqlService` - we'll probably need this in future anyway
* Add a `GithubAuthV4Service` service for badges which query the Github V4 API - most of the work here was to accommodate one change in the V4 API: [rate limit usage info is now returned in the response body](https://developer.github.com/v4/guides/resource-limitations/#returning-a-calls-rate-limit-status) on request. As well as supporting #3610 this will allow us to address other open feature requests requiring the V4 API. e.g: #1550, #3608
* Just for illustrative purposes, I've switched the github forks badge to use the graphql API. There's no real reason we need to do this. I just picked a really simple "get one value from the API, stick it on a badge" example to migrate as I think it helps to see a concrete implementation of the base class (both in dev and review). If we want to switch this back to use the V3/REST API, I can revert 160569e before merging.


core/base-service/base-graphql.js
```diff
@@ -0,0 +1,48 @@
...
+    // Attempting to parse the query string
+    // will throw a descriptive exception if it isn't valid
+    parse(query)
+  }
+
+  async _requestGraphql({
+    schema,
+    url,
+    query,
+    variables = {},
```
chris48s: I've made `query` and `variables` top-level params here. I'm in 2 minds about whether it makes more sense to collect them into an object. Also note here I deliberately haven't implemented mutations. In general, calling a shields badge should read data from upstream service, not write data to it, so we shouldn't need mutations.

core/base-service/base-graphql.js
```diff
@@ -0,0 +1,48 @@
...
+    options = {},
+    httpErrorMessages = {},
+    graphqlErrorHandler = defaultGraphqlErrorHandler,
```
chris48s: In general, Error handling for graphql is going to require a mixed approach:

* `httpErrorMessages`: Certain errors (e.g: `401 Unauthorized`, `500 Internal Server Error`) are going to be returned as status codes so we do need to be able to pass these.
* `graphqlErrorMessageHandler`: With a graphql, a single request can do several things. Some might error, some might not so we need to handle errors returned in an `errors` object (possibly accompanied by a partial valid response) with a `200 OK` status. All the spec says is that an error needs a `.message` property, so these are going to be a bit of a mixed bag from API to API.

services/github/github-api-provider.js
```diff
@@ -34,6 +51,7 @@ class GithubApiProvider {
     if (this.withPooling) {
       this.standardTokens = new TokenPool({ batchSize: 25 })
       this.searchTokens = new TokenPool({ batchSize: 5 })
+      this.graphqlTokens = new TokenPool({ batchSize: 25 })
```
chris48s: Is 25 a sensible number here? I don't really have a handle on the reasons for picking 5 and 25 above.

services/github/github-api-provider.js
```diff
@@ -52,33 +71,70 @@ class GithubApiProvider {
...
+          'x-ratelimit-reset': res.headers['x-ratelimit-reset'],
+        }
+        console.error(
+          `Invalid GitHub rate limit headers ${JSON.stringify(
+            logHeaders,
+            undefined,
+            2
+          )}`
+        )
+        log.error(e)
```
chris48s: Note that I am changing the behaviour here. Previously we were just swallowing this error in production, but it seems like if we end up here, we should be logging something to sentry so we can go and look at it. Taking this path means we are making calls using tokens but we're not updating the `usesRemaining` value. I assume this isn't happening routinely because we're not maxing out everyone's tokens, but I'd still like to know about it if it does happen.

services/github/github-auth-service.js
```diff
@@ -39,7 +41,34 @@ class ConditionalGithubAuthService extends BaseJsonService {
...
+    /*
+    The Github v4 API requires us to query the rateLimit object to return
+    rate limit info in the query body instead of the headers:
+    https://developer.github.com/v4/guides/resource-limitations/#returning-a-calls-rate-limit-status
+    This appends the relevant rateLimit query clause to each
+    call to the GH v4 API so we can keep track of token usage.
+    */
+    attrs.query = mergeQueries(
+      attrs.query,
+      'query { rateLimit { limit cost remaining resetAt } }'
```
chris48s: Explicitly asking the GH API to just tell us what our remaining usage is in the body seems like an easier approach than attempting to parse the query and calculate the projected cost of each call and less likely to result in a discrepency: https://developer.github.com/v4/guides/resource-limitations/#calculating-nodes-in-a-call

That said, we probably do need to watch out for query count on contributions. With the V3 API, its pretty hard for someone to contribute a badge that smashes our usage limit because 1 http request = 1 point, but with a graphql a single http request can cost you a whole truckload of points. Not sure what the solution is for that other than to watch out for it in review.

services/github/github-helpers.js
```diff
@@ -24,6 +25,14 @@ function errorMessagesFor(notFoundMessage = 'repo not found') {
   }
 }
 
+function graphqlErrorHandler(errors) {
+  if (errors[0].type === 'NOT_FOUND') {
+    throw new NotFound({ prettyMessage: 'repo not found' })
+  } else {
```
chris48s: In time, it might be useful to add other cases in here, but I don't know what they are yet.

services/github/github-api-provider.js
paulmelnikow: I think that's fine. 25 was picked as a reasonable number of requests to make before re-evaluating how much quota was left, in order to share fairly across other tokens and because the way this is implemented you can't accurately evaluate how much quota is left while requests are in flight. For search I picked 5 because the limit is something tiny like 50 requests per token.

services/github/github-api-provider.js
```diff
@@ -52,33 +71,70 @@ class GithubApiProvider {
...
+  getRateLimitFromHeaders(headers) {
+    const h = Joi.attempt(headers, headerSchema)
+    return {
+      rateLimit: h['x-ratelimit-limit'],
+      totalUsesRemaining: h['x-ratelimit-remaining'],
+      nextReset: h['x-ratelimit-reset'],
+    }
+  }
+
+  getRateLimitFromBody(body) {
```
paulmelnikow:
```suggestion
  getV4RateLimitFromBody(body) {
```

services/github/github-api-provider.js
```diff
@@ -52,33 +71,70 @@ class GithubApiProvider {
...
       this.standardTokens.add(tokenString)
       this.searchTokens.add(tokenString)
+      this.graphqlTokens.add(tokenString)
     } else {
       throw Error('When not using a token pool, do not provide tokens')
     }
   }
 
-  updateToken(token, headers) {
+  getRateLimitFromHeaders(headers) {
```
paulmelnikow:
```suggestion
  getV3RateLimitFromHeaders(headers) {
```

services/github/github-api-provider.js
```diff
@@ -52,33 +71,70 @@ class GithubApiProvider {
...
+    const h = Joi.attempt(headers, headerSchema)
+    return {
+      rateLimit: h['x-ratelimit-limit'],
+      totalUsesRemaining: h['x-ratelimit-remaining'],
+      nextReset: h['x-ratelimit-reset'],
+    }
+  }
+
+  getRateLimitFromBody(body) {
+    const parsedBody = JSON.parse(body)
```
paulmelnikow: Hmm, it would be nice to avoid parsing the body twice for all these responses. Any idea if there's a good way to do that?

services/github/github-api-provider.js
```diff
@@ -103,7 +161,7 @@ class GithubApiProvider {
   // Act like request(), but tweak headers and query to avoid hitting a rate
   // limit. Inject `request` so we can pass in `cachingRequest` from
   // `request-handler.js`.
-  request(request, url, query, callback) {
+  request(request, url, options = {}, callback) {
```
paulmelnikow: Was `query` always empty in the current GitHub badges?

services/github/github-api-provider.js
```diff
@@ -120,15 +178,12 @@ class GithubApiProvider {
...
-    const options = {
-      url,
-      baseUrl,
-      qs: query,
-      headers: {
-        'User-Agent': 'Shields.io',
-        Accept: 'application/vnd.github.v3+json',
-        Authorization: `token ${tokenString}`,
-      },
+    options.url = url
```
paulmelnikow: Rather than mutate the input, could you copy it into a new object? It seems safer that way (in case e.g. an object is reused for that parameter).

services/github/github-auth-service.js
```diff
@@ -39,7 +41,34 @@ class ConditionalGithubAuthService extends BaseJsonService {
...
 
+class GithubAuthV4Service extends BaseGraphqlService {
+  constructor(context, config) {
+    super(context, config)
+    this._requestFetcher = createRequestFetcher(context, config)
+    this.staticAuthConfigured = true
+  }
+
+  async _requestGraphql(attrs) {
+    attrs.url = `/graphql`
```
paulmelnikow: Same comment as elsewhere; could you copy the input into a new object?

services/github/github-auth-service.js
paulmelnikow: How about establishing a max number of queries, which could be overridden per service? If the response exceeds the configured number it could error out rather than rendering. That way we'll see it right in the code, and can even search for it.

On the other hand, we have a rather huge amount of GitHub quota – I want to say something like 2.2k tokens – so maybe we don't need to worry about this.

services/github/github-forks.service.js
```diff
@@ -67,11 +73,17 @@ module.exports = class GithubForks extends GithubAuthService {
   }
 
   async handle({ user, repo }) {
-    const { forks_count: forkCount } = await this._requestJson({
-      url: `/repos/${user}/${repo}`,
+    const json = await this._requestGraphql({
+      query:
+        'query ($user: String!, $repo: String!) { repository(owner: $user, name: $repo) { forks { totalCount } } }',
```
paulmelnikow: To make this easier to read, what do you think about using a backquote string with line breaks?

Related, would using the `gql` tagged template literal, and passing a parsed query into `requestGraphql` allow us to parse this once when the code is parsed instead of on every request?

core/base-service/base-graphql.js
paulmelnikow: It seems like the job of this function is to take an `errors` object and (hopefully) throw an error. I wonder if we could set this up instead as a `transformErrors` function, which gets an errors object and is responsible for returning an instance of `ShieldsRuntimeError`. We could verify that an error is returned, and then throw it. It reduces the chance that a bad error handler accidentally lets control slip back to the handler. Since there is not really much to do besides translate the errors object into a Shields error, I like the idea of calling it transform, which is more narrowly scoped than handle.

core/base-service/graphql.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Utility function to merge two graphql queries together
+ * This is basically copied from
+ * [graphql-query-merge](https://www.npmjs.com/package/graphql-query-merge)
+ * but can't use that due to incorrect packaging.
+ *
+ * @param {...string} queries queries to merge
+ * @returns {string} merged query
+ */
+function mergeQueries(...queries) {
```
paulmelnikow: 👍

core/base-service/base-graphql.js
chris48s: I see where you're going with that. I've phrased my follow-up question as a code snippet:

```js
if (json.errors) {
  const exception = transformErrors(json.errors)
  if (exception instanceof ShieldsRuntimeError) {
    throw exception
  } else {
    // what happens here?
    // do we throw a runtime error (i.e: log it to sentry if it happens in prod)
    // or throw a generic ShieldsRuntimeError (i.e: render an error badge if it happens in prod)
  }
}
```

services/github/github-auth-service.js
chris48s: Are you thinking of that as a runtime check or a static check? Either way, I'm not against adding it, but I think it can be raised as an issue to address in a later PR.

services/github/github-api-provider.js
chris48s: No, but before we were taking `query` and putting it into an `options` object, like this:

```js
request(request, url, query, callback) {

//bla

const options = {
  //bla
  qs: query,
  //bla
}

request(options...
```

Whereas now `options.qs` will already contain the query.

I've modified the call stack though `requestAsPromise()` and `createRequestFetcher()` appropriately. I don't think any of the actual badge code needs to change.

chris48s:
> As for porting the service or not, one strategy might be to reserve the GraphQL endpoints for the things that really need it, and the other might be to port things over to GraphQL as much as possible.

> Is there a performance difference?

I thought about this a bit already. As I see it, there are a few tradeoffs to think about here (some performance-related and some not). Predictably there is not a single clear-cut answer in all cases:

* In most cases, a sensible graphql query will return a smaller JSON response than a REST call. For example, in the forks example we need to request this object with the V3 API to get one value: https://api.github.com/repos/badges/shields but with graphql the response is only like `{ data: { repository: { forks: { totalCount: 1984 } }, rateLimit: { limit: 5000, cost: 1, remaining: 4999, resetAt: '2019-07-24T21:23:26Z' } } }` That's less bytes over the wire and a faster object to parse.
* Beyond that, I think we can probably assume that making one HTTP request to the V3 API and making one HTTP request to the V4 API are basically identical in terms of performance (even if theoretically one or other takes longer to calculate/serve on their end) because the main thing we're waiting for is network latency.
* With the V3 API, one request = one point off the usage limit. With the V4 API one request may be many points off the usage limit (but will be a minimum of one). Assuming we can produce the same badge with one REST call or a graphql request: In the best-case scenario, the graphql query will be the same cost as the REST request. In all other cases, graphql will be more expensive in this respect. Given we won't accept a badge that makes more than one V3 request, using the V3 request will sometimes be more efficient on our token usage.
* At this point in time, REST is more familiar to contributors and more well-understood.
* I don't think we're in any danger of GitHub deprecating the V3 API in the near future. They haven't published any info about when it will be deprecated or retired: https://developer.github.com/v3/versions/ . That said, presumably at some point in time it will go away and we'll need to migrate everything to V4 one day anyway, unless graphql goes out of fashion or something.

services/github/github-api-provider.js
chris48s: Yeah this is a bit annoying.. I think we could, but I think its also a fairly major change. At the moment we're essentially doing all this token work inside the call to `_request()`, which means we can deal with manipulating the request and processing the response in the same place. We can also deal with `standardTokens`, `searchTokens` and `graphqlTokens` in the same `GithubApiProvider` class. In order to avoid double-parsing, we'd need to restructure so we can still manipulate the request, but then update the token pool after we call `_parseJson()` in either `_requestJson()` or `_requestGraphql()`.

I've not really completely thought through the consequences of all this, but here's a couple of thoughts:

One way to deal with that would be to split this up so we deal with `standardTokens` / `searchTokens` in by overriding the `_requestJson()` method in `GithubAuthV3Service`.. and `graphqlTokens` by overriding the `_requestGraphql()` method of `GithubAuthV4Service`. Although at that point if `GithubAuthV4Service` overrides `_requestGraphql()`, its not really a subclass of `BaseGraphqlService` anymore. 

Maybe another way would be to allow the `_requestJson()`/`_requestGraphql()`/`_requestFomat()` methods in base classes to accept an optional callback param or something that cna be used to run a post-process after parsing but before we return the response.

Any further thoughts or ideas?

core/base-service/base-graphql.js
paulmelnikow: I think it can should work like an assertion:

```js
throw Error(`transformErrors() must return a ShieldsRuntimeError; got ${exception}`)
```

In development or service tests that will bubble right up, and in production it will get handled as an internal error.

services/github/github-auth-service.js
paulmelnikow: Since we don't know the answer until the response comes back, it'd have to be a runtime check. Agreed, it can be addressed later.

services/github/github-api-provider.js
paulmelnikow: Ahh yes, I gotcha.

There are a lot of layers of abstraction here! Would be nice to shave a couple of these off at some point. (Not that you're adding more. They were here to begin with.)

services/github/github-api-provider.js
paulmelnikow: How about having GithubAuthV4Service rely on BaseGraphqlService to do the json parsing, but pass in an empty schema so the validation is a no-op. Then the token-related stuff can happen, then GithubAuthV4Service can invoke _validate?

----

Partly related, I dislike deep inheritance because the indirection makes it hard for people unfamiliar with the code to dig through. Each layer is one more thing to keep in your head at once.

So I think it would be good to make BaseGraphqlService extend directly from BaseService. All that would take is exporting `parseJson` as its own function (or moving it to its own file).

There are a couple places in GitHub where we invoke `_parseJson` from within a service, and those will be made cleaner by importing it directly.

In Gerritt there's also a place where we override it, though it'd be less magical and easier to understand if we used BaseService directly (invoking _request, doing the trim, then invoking parse and validate).

The refactor is a bit involved so maybe it's best to leave it for a follow-on.

paulmelnikow:
> I thought about this a bit already. As I see it, there are a few tradeoffs to think about here (some performance-related and some not). Predictably there is not a single clear-cut answer in all cases:
> 
> * In most cases, a sensible graphql query will return a smaller JSON response than a REST call. For example, in the forks example we need to request this object with the V3 API to get one value: https://api.github.com/repos/badges/shields but with graphql the response is only like `{ data: { repository: { forks: { totalCount: 1984 } }, rateLimit: { limit: 5000, cost: 1, remaining: 4999, resetAt: '2019-07-24T21:23:26Z' } } }` That's less bytes over the wire and a faster object to parse.
> * Beyond that, I think we can probably assume that making one HTTP request to the V3 API and making one HTTP request to the V4 API are basically identical in terms of performance (even if theoretically one or other takes longer to calculate/serve on their end) because the main thing we're waiting for is network latency.
> * With the V3 API, one request = one point off the usage limit. With the V4 API one request may be many points off the usage limit (but will be a minimum of one). Assuming we can produce the same badge with one REST call or a graphql request: In the best-case scenario, the graphql query will be the same cost as the REST request. In all other cases, graphql will be more expensive in this respect. Given we won't accept a badge that makes more than one V3 request, using the V3 request will sometimes be more efficient on our token usage.
> * At this point in time, REST is more familiar to contributors and more well-understood.
> * I don't think we're in any danger of GitHub deprecating the V3 API in the near future. They haven't published any info about when it will be deprecated or retired: https://developer.github.com/v3/versions/ . That said, presumably at some point in time it will go away and we'll need to migrate everything to V4 one day anyway, unless graphql goes out of fashion or something.

Everything you're saying makes sense.

Assuming the GraphQL API is otherwise fast, since network latency is the key issue I'd guess the smaller responses probably will have a bigger impact than anything else.

Could you paste what you wrote into a comment in github-auth-service.js? Or else distill it down to what you think contributors need to know?

services/github/github-api-provider.js
chris48s: Yes I found this part a bit tricky to understand while I was working on it.

services/github/github-api-provider.js
chris48s: As a first step on this, I've moved `parseJson()` out to a helper and switched `BaseGraphqlService` to inherit from `BaseService`. That seems sensible anyway, but I've not taken it further.

Not sure I agree that overriding a method from the base class is "magical" in the context of gerrit though. Its what pretty much every other function we declare in a badge service class is doing - seems like a pretty consistent pattern :man_shrugging:

services/github/github-auth-service.js
chris48s: Right, ok. I misunderstood and thought you were still talking about pre-parsing the query before we send it.

Throwing the exception after we've already made the query is obviously simpler, but also of more limited value (esp if it makes it through to production) in that we've already made the request we don't want to make before deciding we didn't want to make it, if you see what I mean :)

Maybe its a useful MVP to start with, in that if we start getting exceptions in sentry saying "lol - just rinsed our token and now we're not even going to render a badge with all the expensive data we just requested" we can at least be reactive. Bear in mind it might lull us into a false sense of security though as we'd be unlikely to throw it in dev because we treat global tokens differently from pooled ones. My instinct is most contributors probably use a global token as opposed to a pool with one token in it (I know that's what I usually do - I think working on this PR is the first time I've set up a pool locally).

Unfortunately I think we do still actually need to work out the query cost up-front to implement it in a useful way.

chris48s:
Cheers. I'll raise issues for the additional bits of work

PR #3764, opened 2019-07-23
Upgrade to new Sentry SDKs; fix error reporting detail

### LINKED ISSUE #3708: Upgrade to new Sentry SDKs
Sentry recommends upgrading to their new SDKs.

Here is a little bit of info about them:

https://github.com/getsentry/sentry-javascript/blob/master/CHANGELOG.md#400

paulmelnikow:
Closes #3708


core/server/log.js
```diff
@@ -30,21 +29,10 @@ module.exports = function log(...msg) {
...
   listeners.forEach(f => f(d, ...msg))
-  Raven.captureException(msg, sendErr => {
-    if (sendErr) {
-      throttledConsoleError(
-        'Failed to send captured exception to Sentry',
-        sendErr.message
-      )
-    }
-  })
+  Sentry.captureException(msg)
```
calebcartwright: I take it this new SDK function is supposed to gracefully handle any errors that occur while trying to send the info?

paulmelnikow: While I can't find any complete documentation of the API, I can't find any examples that use this parameter so I think it's been removed. I think the errors are logged to the console (though that's not a real help to us).

server.js
```diff
@@ -8,9 +8,12 @@ require('dotenv').config()
...
 // Set up Sentry reporting as early in the process as possible.
 const config = require('config').util.toObject()
-const Raven = require('raven')
-Raven.config(process.env.SENTRY_DSN || config.private.sentry_dsn).install()
-Raven.disableConsoleAlerts()
+const Sentry = require('@sentry/node')
+Sentry.init({
+  dsn: process.env.SENTRY_DSN || config.private.sentry_dsn,
+  integrations: integrations =>
+    integrations.filter(integration => integration.name !== 'Console'),
```
paulmelnikow: This removes console log output which seems to be unrelated to the request that caused the error.

core/server/log.js
```diff
@@ -30,22 +29,11 @@ module.exports = function log(...msg) {
...
-    if (sendErr) {
-      throttledConsoleError(
-        'Failed to send captured exception to Sentry',
-        sendErr.message
-      )
-    }
-  })
-  console.error(d, ...msg)
+  listeners.forEach(f => f(d, err))
+  Sentry.captureException(err)
```
paulmelnikow: Calling this with the error instead of an array containing the error fixes the problem where the stack trace was truncated.

paulmelnikow:
The error reporting detail is working really well now! Very exciting.

PR #3770, opened 2019-07-25
Disable more default Sentry integrations

paulmelnikow:
This one is spewing a bunch of recent http calls into each error log.

I get the benefit of these plugins though I'm curious what resources they might be using. And in this case, don't want the requests to end up e.g. in a GitHub issue.

### DIFF
--- a/server.js
+++ b/server.js
@@ -9,10 +9,22 @@ require('dotenv').config()
 // Set up Sentry reporting as early in the process as possible.
 const config = require('config').util.toObject()
 const Sentry = require('@sentry/node')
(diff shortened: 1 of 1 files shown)


calebcartwright:
> And in this case, don't want the requests to end up e.g. in a GitHub issue.

I wonder if there's a Discord integration where we could have sentry drop a note over there

paulmelnikow:
> I wonder if there's a Discord integration where we could have sentry drop a note over there

Looks like it's possible: https://github.com/getsentry/sentry-plugins/issues/249

I get email notifications so presumably when we add other maintainers, they could also do the same…

PR #3792, opened 2019-07-26
move [Nexus] server and queryOpt from route to query params

### LINKED ISSUE #3745: Example/url builder wrong for Nexus (Query Options) badge

- [ ] My own instance

:beetle: **Description**

Refs https://github.com/badges/shields/pull/3717#discussion_r305450656

The defined [route pattern](https://github.com/badges/shields/blob/master/services/nexus/nexus.service.js#L49) for the Sonatype Nexus badges has an optional trailing route param (query options) that's _not_ separated from the preceeding route parameter with a `/`.

However, the [example listing route pattern](https://github.com/badges/shields/blob/master/services/nexus/nexus.service.js#L102) used for the Sonatype Nexus (Query Options) badge incorrectly inserts a `/`, resulting in generating a bad url/404

:link: **Link to the badge**

This badge url works:
https://img.shields.io/nexus/fs-public-snapshots/https/repository.jboss.org/nexus/com.progress.fuse/fusehq:c=agent-apple-osx:p=tar.gz.svg 
 

However, the badge url generated from the modal window does not (returns 404 badge not found):
https://img.shields.io/nexus/fs-public-snapshots/https/repository.jboss.org/nexus/com.progress.fuse/fusehq/:c=agent-apple-osx:p=tar.gz.svg 

:bulb: **Possible Solution**
We'll likely move the optional trailing `queryOpt` from a route param to a query param, with a respective redirector to maintain backwards compatibility

**Edit** - Fixed typos

calebcartwright:
Closes #3745


services/nexus/nexus.service.js
```diff
@@ -33,6 +34,15 @@ const resolveApiSchema = Joi.object({
   }).required(),
 }).required()
 
+// https://repository.sonatype.org/nexus-restlet1x-plugin/default/docs/path__artifact_maven_resolve.html
+// https://repository.sonatype.org/nexus-indexer-lucene-plugin/default/docs/path__lucene_search.html
+const queryParamSchema = Joi.object({
+  server: optionalUrl.required(),
+  queryOpt: Joi.string()
+    .regex(/(:(?:q|g|a|v|p|c|cn|sha1|from|count|repositoryId|e|r)=[\w-. ]+)+/i)
```
calebcartwright: This is a stricter regex than the previous one (`(:.+?)?`), but is based on the query params that the two Sonatype APIs actually accept (details at the links in the above inline comments). 

I can swap this back to the more lenient one if anyone has concerns with this new one

services/nexus/nexus.tester.js
```diff
@@ -7,34 +7,34 @@ const t = (module.exports = require('../tester').createServiceTester())
 
 t.create('search release version valid artifact')
   .timeout(15000)
-  .get('/r/https/oss.sonatype.org/com.google.guava/guava.json')
+  .get('/r/com.google/bitcoinj.json?server=https://oss.sonatype.org')
```
calebcartwright: Our Nexus tests are notoriously bad about timeouts. I found using this target was noticeably faster though (~1 second vs. 10-12 seconds)

PR #3829, opened 2019-08-06
tweak build schema for [AzureDevOpsBuild] to handle never built use case

### LINKED ISSUE #3819: Update AzureDevOps build schema with `never built` status
Currently, requests for an Azure DevOps build that has never been executed return are returning "invalid response data". I believe at some point a new status was added on the Azure DevOps side (`never built`) that's not included in [our schema](https://github.com/badges/shields/blob/master/services/azure-devops/azure-devops-helpers.js#L10). 

We should update the schema to include the `never built` status so that we don't display the `invalid response data` badge which may be a bit confusing for folks.

calebcartwright:
Closes #3819

### DIFF
--- a/services/azure-devops/azure-devops-build.tester.js
+++ b/services/azure-devops/azure-devops-build.tester.js
@@ -20,6 +20,10 @@ t.create('named branch')
     message: isBuildStatus,
   })
 
+t.create('never built definition')
+  .get('/swellaby/opensource/112.json')
+  .expectBadge({ label: 'build', message: 'never built' })
+
 t.create('unknown definition')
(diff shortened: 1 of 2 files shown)


services/azure-devops/azure-devops-build.tester.js
```diff
@@ -20,6 +20,10 @@ t.create('named branch')
     message: isBuildStatus,
   })
 
+t.create('never built definition')
+  .get('/swellaby/opensource/112.json')
+  .expectBadge({ label: 'build', message: 'never built' })
```
calebcartwright: I figured it'd be worth adding a test for this scenario, but can certainly remove it if folks feel strongly otherwise. I've got the target build pipeline here locked so that it will never run, keeping this test passing.

PR #3875, opened 2019-08-15
Validate input to BadgeFactory.create()

chris48s:
Following on from #3507 and #3586 here is another attempt. See https://github.com/badges/shields/pull/3586#issuecomment-513375146 for background.

This time I haven't used any library - I've just coded the validation rules by hand. Note again, the target branch is `v3-dev`.

### DIFF
--- a/gh-badges/lib/index.js
+++ b/gh-badges/lib/index.js
@@ -5,6 +5,8 @@
 
 const makeBadge = require('./make-badge')
 
+class ValidationError extends Error {}
+
 /**
  * BadgeFactory
  */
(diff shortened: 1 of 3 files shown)


gh-badges/lib/index.js
```diff
@@ -17,6 +19,56 @@ class BadgeFactory {
...
+      'colorscheme',
+      'colorB',
+    ]
+    stringFields.forEach(function(field) {
+      if (field in format && typeof format[field] !== 'string') {
+        throw new ValidationError(`Field \`${field}\` must be of type string`)
+      }
+    })
+
+    const formatValues = ['svg', 'json']
```
chris48s: We never came to an agreement on this. The discussion in https://github.com/badges/shields/pull/3586#discussion_r295802787 still applies here.

gh-badges/lib/index.js
```diff
@@ -17,6 +19,56 @@ class BadgeFactory {
...
+    ) {
+      throw new ValidationError('Field `text` must be an array of 2 strings')
+    }
+
+    const stringFields = [
+      'labelColor',
+      'color',
+      'colorA',
+      'colorscheme',
+      'colorB',
```
calebcartwright: Would it be worth another check to print a warning message when any of these deprecated fields are used, similar to the constructor check on `option`?

gh-badges/lib/index.spec.js
```diff
@@ -16,5 +21,37 @@ describe('BadgeFactory class', function() {
         template: 'flat',
       })
     ).to.satisfy(isSvg)
+    expect(
+      bf.create({
+        text: ['build', 'passed'],
+        foo: 'bar',
```
calebcartwright: I was confused by this one at first, but the intent is just a check that we don't want to throw a ValidationError when the user supplies extraneous keys right?

gh-badges/lib/index.js
calebcartwright: I don't have any strong feelings on this one myself. I suppose that either approach would be a breaking API change. 

I'd probably vote for the behavior you've implemented here (throwing a validation error on raster format). Obviously the lib doesn't support raster today, and AFAICT it doesn't seem like there's been any requests for it.

However, I'd be okay with adding raster support in the lib if we intend to keep raster support in the CLI, and especially if the process of fixing/upgrading the CLI support (refs https://github.com/badges/shields/issues/1788) could be aided by doing so.

chris48s: I agree for v3 we should either add raster to the library or remove it from the CLI so the two are consistent.

re: breaking changes, adding validation at all is a breaking change as explained in https://github.com/badges/shields/pull/3507#issue-283411297 so I'm not really worried about making a non-backwards-compatible change here. Lets just pick the way we want it to be moving forwards.

Basically what I want to do is work on each of the things listed in #3587 as a seperate PRs, merge them all to `v3-dev` and then when all that is done, merge that to master. That way we can still do a patch release off master in the meantime if we need to but we don't have to load all the v3 changes into one PR. I appreciate that for various reasons, bullet point one on that list has dragged out over about 3 months, so we've kinda lost sight of the high-level stuff a bit.

gh-badges/lib/index.js
chris48s: One of the things on the v3 roadmap is to remove the params we deprecated in 2.2.0. I was aiming to keep each of those points a separate PR, but maybe it makes sense to just remove the deprecated params in this PR rather as well than add validation for them here and then remove them later.

gh-badges/lib/index.spec.js
chris48s: exactly

chris48s: I've added  a comment to make this clearer

PR #3879, opened 2019-08-17
Typescripterize BadgeExamples and SuggestionAndSearch

paulmelnikow:
I realized the two different kinds of data that can be passed to `<BadgeExample />` were a bit less similar than I thought, so this includes a little refactor related to that which isn't perfect, but leaves things in a cleaner place than before.


calebcartwright:
Error details from the failed deployment of the reveiw app:

```
Error: Minified React error #130; visit https://reactjs.org/docs/error-decoder  .html?invariant=130&args[]=undefined&args[]= for the full message or use the n  on-minified dev environment for full errors and additional helpful warnings.
```

calebcartwright:
By the way, this also seems to fix the issue we've had historically with suggestions crashing on an invalid url 👍 🎉 💯

frontend/lib/service-definitions/index.ts
```diff
@@ -7,20 +7,24 @@ export interface Category {
...
 }
 
+export interface ExampleSignature {
+  pattern: string
+  namedParams: { [k: string]: string }
+  queryParams: { [k: string]: string }
+}
+
+export interface Preview {
+  label?: string
```
calebcartwright: Could you remind me again why label is optional (feel like I remember asking this before sorry 😄)? Is it because it's often set in the default badge data for a service or a different reason?

frontend/components/badge-examples.tsx
```diff
@@ -29,39 +42,52 @@ const ClickableCode = styled(StyledCode)`
...
-      label,
+      label: label || '',
       message,
       color,
       style,
       namedLogo,
     })
     exampleUrl = badgeUrlFromPath({
       path: removeRegexpFromPattern(pattern),
-      namedParams,
```
calebcartwright: Were these unused/unnesccessary before?

frontend/components/suggestion-and-search.tsx
```diff
@@ -54,18 +82,21 @@ export default function SuggestionAndSearch({
...
+        },
+        preview: preview || {},
         isBadgeSuggestion: true,
-        documentation,
       })
     )
 
     return (
       <BadgeExamples
+        areBadgeSuggestions
```
calebcartwright: I take it this results in a `false` or some other falsy value being specified?

frontend/components/badge-examples.tsx
paulmelnikow: Yea, they were ignored by that function.

frontend/lib/service-definitions/index.ts
paulmelnikow: I think this was to match the schema of the service definition export. If that always returns an empty string when there is no label, we could update this.

frontend/components/suggestion-and-search.tsx
paulmelnikow: The opposite; Including it results in `true`. This is a convention JSX inherited from HTML.

paulmelnikow:
Thanks for helping me keep this moving!

frontend/lib/service-definitions/index.ts
calebcartwright: Gotcha thanks

PR #3884, opened 2019-08-18
Remove BadgeFactory

chris48s:
Another one targetting v3-dev. This PR removes the `BadgeFactory` class (which we don't need now that all the text-measuring is being done by anafanafo instead of PDFKit) and just replaces it with a single function call `makeBadge()`.

The diff changes quite a lot, but its mostly just indentation. Probably the key thing here is: is `makeBadge()` the function name we want to expose?

I haven't switched our internal stuff to use the public interface of the package here yet, as there are still things (logos, links) that we support on shields.io but aren't exposed via the documented interface of the package yet.

I think I'm probably going to hold off on the other v3 changes (expose logos, rename the package, etc) until #3637 is done, as those jobs are going to create a big merge conflict with that work.

### DIFF
--- a/gh-badges/lib/index.js
+++ b/gh-badges/lib/index.js
@@ -3,85 +3,85 @@
  * @module gh-badges
  */
 
-const makeBadge = require('./make-badge')
+const _makeBadge = require('./make-badge')
 
 class ValidationError extends Error {}
 
-/**
- * BadgeFactory
- */
-class BadgeFactory {
-  constructor(options) {
-    if (options !== undefined) {
-      console.error(
-        'BadgeFactory: Constructor options are deprecated and will be ignored'
-      )
-    }
+function _validate(format) {
+  if (!('text' in format)) {
+    throw new ValidationError('Field `text` is required')
   }
 
-  _validate(format) {
-    if (!('text' in format)) {
-      throw new ValidationError('Field `text` is required')
-    }
-
-    if (
-      !Array.isArray(format.text) ||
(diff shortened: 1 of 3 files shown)


gh-badges/lib/index.js
```diff
@@ -3,85 +3,85 @@
...
-      'for-the-badge',
-      'popout',
-      'popout-square',
-      'social',
-    ]
-    if ('template' in format && !templateValues.includes(format.template)) {
-      throw new ValidationError(
-        `Field \`template\` must be one of (${templateValues.toString()})`
-      )
+function _clean(format) {
```
chris48s: I've added this additional function here which:

* strips any `null` or `undefined` values
* strips any values we don't want to pass on to the other `makeBadge()` ( in `lib/make-badge` )

(without mutating the input object) before we validate the object and pass it on. Its not really strictly related to the rest of this PR, but I realised I should have done this in the validation PR while I was working on this.

calebcartwright:
> Probably the key thing here is: is makeBadge() the function name we want to expose?

I don't have any opinions/assistance to offer there. Were there any other proposed names under consideration?

chris48s:
`createBadge()`, just `create()`? `badge()`?
not sure its deeply important, but once we pick a name we're stuck with it for a bit :)

paulmelnikow:
I like `makeBadge` and am fine with `createBadge` if you prefer that.

chris48s:
seems like we're down with `makeBadge()`

PR #3885, opened 2019-08-19
Typescripterize QueryStringBuilder and PathBuilder

paulmelnikow:


frontend/components/customizer/path-builder.tsx
```diff
@@ -106,10 +129,14 @@ export default function PathBuilder({
         // objects.
         tokens
           .filter(t => typeof t !== 'string')
-          .reduce((accum, { name }) => {
-            accum[name] = ''
-            return accum
-          }, {})
+          .map(t => t as Key)
```
calebcartwright: I'm not familiar with `path-to-regexp`, but I take it that any strings in the `tokens` array are still removed/handled?

paulmelnikow: Yea, they’re being filtered out just above.

calebcartwright: 🤦‍♂ 

Sorry about that! for some reason my brain read that (line 108/131) as a removed line

frontend/components/customizer/query-string-builder.tsx
```diff
@@ -24,20 +27,44 @@ const QueryParamCaption = styled(BuilderCaption)`
...
+  queryParams,
+  badgeOptions,
+}: {
+  queryParams: Record<string, string | boolean>
+  badgeOptions: Record<BadgeOptionName, string | undefined>
+}): {
+  queryString: string
+  isComplete: boolean
+} {
+  // Use `string | null`, because `query-string` renders e.g.
```
calebcartwright: Should this comment be `string | boolean | null` to match the `Record` line below?

frontend/components/customizer/query-string-builder.tsx
```diff
@@ -176,43 +228,59 @@ export default function QueryStringBuilder({
...
+    )
   )
 
-  function handleServiceQueryParamChange(event) {
-    const { name, type } = event.target
-    const value =
-      type === 'checkbox' ? event.target.checked : event.target.value
-    setQueryParams({ ...queryParams, [name]: value })
+  function handleServiceQueryParamChange({
+    target: { name, type, checked, value },
```
calebcartwright: Any potential issues we could run into here since `type` is a reserved/key word in TypeScript?

frontend/components/customizer/query-string-builder.tsx
paulmelnikow: Ha! Actually `boolean` should not be in there. I added that to silence a type warning but when I read the comment it's clear that the entire point is to convert the booleans to nulls.

frontend/components/customizer/query-string-builder.tsx
paulmelnikow: Will update that, though if that really is a problem, wonder if we should open a new issue to have the linter check for it.

calebcartwright: Everything was working correctly as far as I could tell but it wouldn't surprise me if at some point the TypeScript compiler would start to complain

Issue #2574, opened 2018-12-22
Badge request: GitHub actions


xiaoxiangmoe:
:clipboard: **Description**

badges for github actions 

:link: **Data**

Being consulted, this is just an idea. No API available now.

https://developer.github.com/actions/

paulmelnikow:
Some questions for you:

1. What's the problem you're wanting to solve?
2. How do you want to use github actions to solve it?
3. How would the badge be part of that?

xiaoxiangmoe:
* Q: What's the problem you're wanting to solve?
  A: Badge for unit test status and coverages in github actions.
* Q: How do you want to use github actions to solve it?
  A: I'll use github actions as runner for testing.
* Q: How would the badge be part of that?
  A: Badge will show result of testing..

devblackops:
Here is an example: https://github.com/CultureHQ/github-actions-badge

dougnukem:
@devblackops I'm using the badge from @kddeisz and it works great, they added functionality to link to github action results as well, maybe that can be added to `shields.io` (right now its a Lambda JS function).

- https://github.com/cedrickring/golang-action/blob/6fb2e21b8a312a444d7ce88ebb93783739d82f8f/README.md

[![Actions Status](https://wdp9fww0r9.execute-api.us-west-2.amazonaws.com/production/badge/cedrickring/golang-action)](https://wdp9fww0r9.execute-api.us-west-2.amazonaws.com/production/results/cedrickring/golang-action)

gr2m:
See also https://github.com/JasonEtco/action-badges

imbsky:
+1

chris48s:
It looks like the other example posted is calling a URL like https://api.github.com/repos/badges/shields/commits/master/check-suites with the header `Accept: application/vnd.github.antiope-preview+json`, e.g:

```
curl -H "Accept: application/vnd.github.antiope-preview+json" https://api.github.com/repos/badges/shields/commits/master/check-suites
```

..and then looking for a check-suite with the name "GitHub Actions". GitHub's docs on this API call say

> The API may change without advance notice during the preview period. Preview features are not supported for production use

( see https://developer.github.com/v3/checks/suites/#list-check-suites-for-a-specific-ref )

but tbh, if someone's interested in working on this, I'd be inclined to take a PR for it even though the API is beta at this stage.

qoomon:
FYI official badge api https://github.com/{github_id}/{repository}/workflows/{workflow_name}/badge.svg

[...]

PR #3898, opened 2019-08-19
add support for [GitHubWorkflowStatus] (Actions) using BaseSvgScraping service implementation

calebcartwright:
Refs #2574 - I'm not sure if this is the way we want to go, but opening for discussion. 

~~I also don't know enough about GitHub Actions to know the full scope of values we may see in the badges, and whether `isBuildStatus` will sufficiently cover those~~ **Edit** - This is scraping GH's workflow status badge which AFAICT uses typical result statuses we use `isBuildStatus` for


calebcartwright:
The Heroku review apps seem to no longer like me 😄 

```sh
Error: Cannot find module 'is-css-color'
Require stack:
- /tmp/build_c23bc303820126400ec74975c33f173e/gh-badges/lib/color.js
- /tmp/build_c23bc303820126400ec74975c33f173e/lib/logos.js
- /tmp/build_c23bc303820126400ec74975c33f173e/core/base-service/coalesce-badge.js
- /tmp/build_c23bc303820126400ec74975c33f173e/core/base-service/base.js
- /tmp/build_c23bc303820126400ec74975c33f173e/core/base-service/loader.js
- /tmp/build_c23bc303820126400ec74975c33f173e/scripts/export-service-definitions-cli.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:713:15)
    at Function.Module._load (internal/modules/cjs/loader.js:618:27)
    at Module.require (internal/modules/cjs/loader.js:771:19)
    at require (internal/modules/cjs/helpers.js:68:18)
    at Object.<anonymous> (/tmp/build_c23bc303820126400ec74975c33f173e/gh-badges/lib/color.js:3:20)
    at Module._compile (internal/modules/cjs/loader.js:868:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:879:10)
    at Module.load (internal/modules/cjs/loader.js:731:32)
    at Function.Module._load (internal/modules/cjs/loader.js:644:12)
    at Module.require (internal/modules/cjs/loader.js:771:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/tmp/build_c23bc303820126400ec74975c33f173e/gh-badges/lib/color.js',
    '/tmp/build_c23bc303820126400ec74975c33f173e/lib/logos.js',
    '/tmp/build_c23bc303820126400ec74975c33f173e/core/base-service/coalesce-badge.js',
    '/tmp/build_c23bc303820126400ec74975c33f173e/core/base-service/base.js',
    '/tmp/build_c23bc303820126400ec74975c33f173e/core/base-service/loader.js',
    '/tmp/build_c23bc303820126400ec74975c33f173e/scripts/export-service-definitions-cli.js'
  ]
}
npm ERR! code ELIFECYCLE
```

chris48s:
In terms of a validator, the scope of 'actions' is much larger than a CI service. They can be used to deploy, build/publish packages, send notifications, etc ( see https://github.com/marketplace?type=actions for examples ) so I'd expect we'd see a much wider range of values here.

Beyond that, the code for this is fine. My main reservation with this is relying on this endpoint given we don't have any docs from GitHub acknowledging it exists. At least we know the beta API endpoint is documented as being in beta. Anyone else have any thoughts on this?

paulmelnikow:
I don’t feel strongly, though I’d lean toward using the API. We have a lot of quota so that doesn’t seem like a main concern.

calebcartwright:
> At least we know the beta API endpoint is documented as being in beta. Anyone else have any thoughts on this?

> I don’t feel strongly, though I’d lean toward using the API.

I'd always prefer using an API 😄 I had intended to open up an implementation using the API as well for comparison discussion but got distracted. 

> We have a lot of quota so that doesn’t seem like a main concern.

I don't disagree, and I'm definitely not pushing for the SvgScraping implementation (I'd vote for the API as well). My thought process was that long term we may have some questions around quota (for example what the GraphQL API impact may or may not be), and whether Shields being able to provide a badge for GitHub data without hitting the quota would be a plus. IMO it is a plus, but a pretty small one that wouldn't change my vote. 

> In terms of a validator, the scope of 'actions' is much larger than a CI service. They can be used to deploy, build/publish packages, send notifications, etc

I realized that the upstream badge GitHub is providing and used here is really more of a _Workflow Status_ badge vs. individual actions, so the PR title/Service class name/etc. here should be updated to be more precise.

It would not surprise me at all if there are additional states that may be involved here, but I would note that a GitHub workflow in that context really isn't any different than Jenkins, Azure DevOps, Circle, etc. Those can/are used to do many of the same sort of steps with orchestration, but at the end of the day the top level pipeline/workflow/etc. still has some final status that conveys a more wholistic pass/fail/etc. state, and so accordingly do our badges.

calebcartwright:
Now that GH Actions has come out of beta, the native status badges support an optional query param for branch specification (which was the biggest feature gap with this approach). 

I'll check to see if the CheckSuites API now provides a way to query/filter based on workflow name now too. If not, I think we'll need to go with this approach so that users can specify both the workflow and branch names

calebcartwright:
In the spirit of not allowing perfection to be the enemy of the good, I think we should go ahead with this approach to support GitHub Actions/Workflows. I know there's a collective preference for APIs over badge scraping (myself included), but I've still not found away to use the GitHub Checks Suite API (which is much broader and more generic than GH Actions) that supports getting the status of a specific workflow by workflow-name. 

If/when we discover an alternative way to to get the data, we can always swap out our implementation at that time. However, there's a ton of interest for GH Actions and IMHO we should proceed with delivering this badge to the communtiy. 

With GitHub recently adding support for branch querying, the only outstanding item (AFAIK) with this PR is a question/concern @chris48s raised regarding using the `isBuildStatus` helper in the schema validation.

While I agree that GH actions can/are used for all sorts of things, the badge being scraped here is the `workflow status` badge: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

> Status badges show whether a workflow is currently failing or passing. 

Based on the text from those docs, as well as what I've seen, I'm fairly confident that this status badge will follow the standard `pass/fail` variant status message that we see from other CI/pipeline platforms. Yes there will be all sorts of actions and activities that occur within a workflow, but I expect the final status value that our schema will validate will align well with `isBuildStatus`. 

On the off chance that GH's native status badge _can_ contain wildly different status values (for example a test summary message like `10 passed, 1 failed`, etc.) then I feel it'd probably be better for us to fail loudly with a schema validation error and have users report the specifics so we can handle it accordingly.

calebcartwright:
Here's an example of a repo with multiple-workflows that illustrate why it's important for user's to be able to specify both the branch and workflow they want a badge for

https://github.com/rust-lang/mdBook/tree/master/.github/workflows

https://shields-staging-pr-3898.herokuapp.com/github/actions/rust-lang/mdBook/CI/master


https://shields-staging-pr-3898.herokuapp.com/github/actions/rust-lang/mdBook/Deploy/master


https://shields-staging-pr-3898.herokuapp.com/github/actions/rust-lang/mdBook/Deploy/gh-pages

imbsky:
I can't wait this! Thanks for your effort @calebcartwright!

calebcartwright:
I'm also leaning towards renaming the route/class/etc. to refer to this as `GitHubWorkflowStatus` and `/github/workflow/status` (or `GitHubActionsWorkflowStatus` and `/github/actions/workflow/status`) and will include `actions` in the keywords for searchability. 

That's because there can be multiple jobs within a workflow, and it's highly likely IMO that we'll want to support job status badges in the future (`github/workflow/job/status` etc.)

Thoughts?

PyvesB:
> I'm also leaning towards renaming the route/class/etc. to refer to this as `GitHubWorkflowStatus` and `/github/workflow/status` (or `GitHubActionsWorkflowStatus` and `/github/actions/workflow/status`) and will include `actions` in the keywords for searchability.
> 
> That's because there can be multiple jobs within a workflow, and it's highly likely IMO that we'll want to support job status badges in the future (`github/workflow/job/status` etc.)
> 
> Thoughts?

I would go with the shorter `/github/workflow/status` form and include "actions" as a keyword as you suggested.

services/github/github-actions.tester.js
```diff
@@ -0,0 +1,32 @@
...
+  .expectBadge({
+    label: 'build',
+    message: 'repo, branch, or workflow not found',
+  })
+
+t.create('valid workflow')
+  .get('/actions/toolkit/Main%20workflow.json')
+  .expectBadge({
+    label: 'build',
+    message: isBuildStatus,
```
PyvesB: Maybe worth adding the `no status` case to avoid flaky tests?

calebcartwright: Good catch!

imbsky:
👏

svoop:
Big thanks, @calebcartwright 🎉

calebcartwright:
Just a heads up for anyone following this thread.. it's going to be a bit longer before this is deployed and available

chris48s:
Thanks for getting this one over the line. Sorry I didn't get a chance to post earlier, but now that GitHub are documenting this exists, I think its fine to rely on it. Also agree on validation schema :+1: 

As a slight tangent, it is interesting to see GH starting to serve and support their own badges. I wonder if we will see more popping up. One to keep an eye on :eyes:

calebcartwright:
I stand corrected 😆  this is live! Thanks @paulmelnikow for getting this shipped so quickly 🎉 

Some examples:
https://img.shields.io/github/workflow/status/apache/arrow/Python?style=flat-square


And one with all sorts of Shields-customization goodness
https://img.shields.io/github/workflow/status/apache/arrow/Python?color=orange&label=actions&logo=github&logoColor=orange&style=for-the-badge

imbsky:
WOW! Thank you @paulmelnikow!

arcanis:
GitHub Actions now have a fresh new API, might be useful to get the workflow names and statuses without scrapping: https://developer.github.com/v3/actions/

calebcartwright:
Thanks for letting us know @arcanis!

PR #3902, opened 2019-08-22
Add service for [GitHubLabels]

pfaion:
Hi, I created a small service for displaying labels of GitHub repos.
This is nice for mentioning labels across GitHub, as GitHub-flavored markdown does not offer the option to reference a label. It keeps the color in sync and you will easily recognize if a label does not exist anymore.

The endpoint is `/github/labels/{user}/{repo}/{label}`.
You can e.g. reference the **documentation** label of this repo with: `/github/labels/badges/shields/documentation`, which would look similar to:

A nonexistent label `/github/labels/badges/shields/gibberish` would look like:
-red)

Please feel free to make suggestions to my code as I am not used to writing JS or developing for node.

Thanks for offering such a cool service!

### DIFF
--- a/services/github/github-labels.service.js
+++ b/services/github/github-labels.service.js
@@ -10,7 +10,7 @@ const schema = Joi.object({
 
 module.exports = class GithubLabels extends GithubAuthV3Service {
   static get category() {
-    return 'other'
+    return 'issue-tracking'
   }
 
   static get route() {
--- a/services/github/github-labels.service.js
+++ b/services/github/github-labels.service.js
@@ -5,7 +5,9 @@ const { GithubAuthV3Service } = require('./github-auth-service')
 const { documentation, errorMessagesFor } = require('./github-helpers')
 
 const schema = Joi.object({
-  color: Joi.string().hex().required(),
+  color: Joi.string()
+    .hex()
+    .required(),
 }).required()
 
 module.exports = class GithubLabels extends GithubAuthV3Service {
@@ -52,7 +54,7 @@ module.exports = class GithubLabels extends GithubAuthV3Service {
     return this._requestJson({
       url: `/repos/${user}/${repo}/labels/${name}`,
       schema,
-      errorMessages: errorMessagesFor(`${name} (repo or label not found)`),
+      errorMessages: errorMessagesFor(`repo or label not found`),
     })
   }
 
--- a/services/github/github-labels.service.js
+++ b/services/github/github-labels.service.js
@@ -1,0 +1,63 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { GithubAuthV3Service } = require('./github-auth-service')
+const { documentation, errorMessagesFor } = require('./github-helpers')
+
+const schema = Joi.object({
+  color: Joi.string(),
+}).required()
+
+module.exports = class GithubLabels extends GithubAuthV3Service {
+  static get category() {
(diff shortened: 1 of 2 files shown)


services/github/github-labels.service.js
```diff
@@ -0,0 +1,63 @@
...
+const { GithubAuthV3Service } = require('./github-auth-service')
+const { documentation, errorMessagesFor } = require('./github-helpers')
+
+const schema = Joi.object({
+  color: Joi.string(),
+}).required()
+
+module.exports = class GithubLabels extends GithubAuthV3Service {
+  static get category() {
+    return 'other'
```
calebcartwright:
```suggestion
    return 'issue-tracking'
```

services/github/github-labels.service.js
```diff
@@ -0,0 +1,63 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { GithubAuthV3Service } = require('./github-auth-service')
+const { documentation, errorMessagesFor } = require('./github-helpers')
+
+const schema = Joi.object({
+  color: Joi.string(),
```
calebcartwright: IIRC, all GitHub labels must have a color value (the system automatically assigns one if the user does not specify) so I think we can make the schema a bit more restrictive. Let me know if I've got those constraints wrong though 👍 

```suggestion
  color: Joi.string().hex().required(),
```

pfaion: Makes total sense, including the hex.
I didn't know how to use Joi properly.

services/github/github-labels.service.js
```diff
@@ -0,0 +1,63 @@
...
+      message: name,
+      color,
+    }
+  }
+
+  async fetch({ user, repo, name }) {
+    return this._requestJson({
+      url: `/repos/${user}/${repo}/labels/${name}`,
+      schema,
+      errorMessages: errorMessagesFor(`${name} (repo or label not found)`),
```
calebcartwright: I'd prefer to not inject the user provided label name here and just use `repo or label not found` error message. That'd be more consistent with the other GH badges and may also avoid any potential confusion. 

For example if a user has a typo in the GH repo, but has the label name correct:
https://shields-staging-pr-3902.herokuapp.com/github/labels/bagdes/shields/keep-service-tests-green

services/github/github-labels.service.js
calebcartwright: No worries! Looks like one of the inline suggestions I gave was not properly formatted for prettier, so if you could run an `npm run prettier` and then push that should resolve the failing prettier checks

services/github/github-labels.service.js
pfaion: My reasoning for injecting the name was that labels might get renamed/removed while a repository grows and old references might thereby become unresolveable. In this case I thought it might be nice to know which label was referenced, even when the reference becomes broken at some later point.

However, I agree about consistency. Additionally, confusion due to a typo might be a more common problem than confusion due to old references being not valid anymore (I just literally spent 10 minutes trying to understand why your example with /ba**gd**es/shields did not work).

pfaion: Fixed in d30d301472252ef462679ad37e72f1524f2e9b7c

calebcartwright: > I just literally spent 10 minutes trying to understand why your example with /bagdes/shields did not work

😉

Issue #3901, opened 2019-08-21
Unhandled Rejection (SecurityError): The operation is insecure.


Hardik500:
Are you experiencing an issue with...

- [ ] [shields.io](https://shields.io/#/)
- [X] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

I just started running the development using the steps in TUTORIAL.md after running `npm start`. After that the browser tries to render the webpage but provides an error,
Here is the screen shot file:

![Screenshot_1](https://user-images.githubusercontent.com/28431356/63434359-8df46300-c442-11e9-8ecc-b964179523e8.png)

The logs produced were:
`
[server] (node:10068) UnhandledPromiseRejectionWarning: Error: Token pool is exhausted
[server]     at TokenPool._nextBatch (C:\Users\User\Documents\User\git\shields\core\token-pooling\token-pool.js:263:11)
[server]     at TokenPool.next (C:\Users\User\Documents\User\git\shields\core\token-pooling\token-pool.js:300:20)
[server]     at GithubApiProvider.tokenForUrl (C:\Users\User\Documents\User\git\shields\services\github\github-api-provider.js:155:33)
[server]     at GithubApiProvider.request (C:\Users\User\Documents\User\git\shields\services\github\github-api-provider.js:171:22)
[server]     at Promise (C:\Users\User\Documents\User\git\shields\services\github\github-api-provider.js:210:12)
[server]     at new Promise (<anonymous>)
[server]     at GithubApiProvider.requestAsPromise (C:\Users\User\Documents\User\git\shields\services\github\github-api-provider.js:209:12)
[server]     at GithubForks._requestFetcher (C:\Users\User\Documents\User\git\shields\services\github\github-auth-service.js:13:23)
[server]     at GithubForks._request (C:\Users\User\Documents\User\git\shields\core\base-service\base.js:223:40)
[server]     at GithubForks._requestGraphql (C:\Users\User\Documents\User\git\shields\core\base-service\base-graphql.js:72:35)
[server] (node:10068) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
[server] (node:10068) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
`

[...]

PR #3904, opened 2019-08-22
fix package-lock.json to restore heroku deploys

calebcartwright:
Heroku deploys are failing again with missing dependencies that seem to be related to the package-lock.json file

https://github.com/badges/shields/pull/3898#issuecomment-523230312
https://dashboard.heroku.com/pipelines/d8fd42e8-8e42-466e-98fc-f5f50edbdd8f


calebcartwright:
No such luck 🤔 I'm gonna see if I can get dependabot to open a PR with a fix

paulmelnikow:
@calebcartwright Have you tried updating to the latest npm? I’m not following development super closely but there were some recent regressions on Windows that were fixed in the latest version.

paulmelnikow:
I'll give this a second shot. I think the last known working deploy was 316749bd69b684274b2ffd952f9c0872849d82af so I'll start by reverting to that.

paulmelnikow:
I wonder if this is not a problem with the lockfile, but with the npm installer. If I blow away my local `node_modules` and re-run `npm install`, I can reproduce the problem locally.

paulmelnikow:
From the build log:

```       
       We're sorry this build is failing! You can troubleshoot common issues here:
       https://devcenter.heroku.com/articles/troubleshooting-node-deploys
       
       Some possible problems:
       
       - Dangerous semver range (>) in engines.node
         https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version
```

paulmelnikow:
I wonder if I was wrong about which was the last known working version.

calebcartwright:
I'm baffled. I know this is the most unhelpful thing commonly said by engineers, but it works on my machine!

paulmelnikow:
Can you give me steps to produce a good install on this commit? I can try it on my machine.

The commit on #3909 is _not_ working on my machine.

calebcartwright:
For me a `rm -rf` on `node_modules` and `gh-badges/node-modules` and then an `npm i` and `npm start`. I can also run an `npm ci` in there

calebcartwright:
I finally just gave up and added an explicit `npm i` in the `gh-badges` directory as part of the npm `build` script. This is admittedly a hack that should not be neccessary IMO, but I'm not sure what else to try.

It still bugs me not knowing what's happening here though. It looks like heroku is using node v12.9.0 which is admittedly a big jump from my local 8.x environment:
```
 Resolving node version >= 8...
       Downloading and installing node 12.9.0...
       Bootstrapping npm >= 5 (replacing 6.10.2)...
```

Maybe all of this will just go away once we start consuming `gh-badges` as a package vs. the local reference?

gh-badges/package.json
```diff
@@ -36,7 +36,7 @@
   "dependencies": {
     "anafanafo": "^1.0.0",
     "dot": "^1.1.2",
-    "gm": "^1.23.0",
+    "gm": "^1.23.1",
```
calebcartwright: Thought updating this file might help (it did not). I can roll this back if anyone has concerns, but I don't see any issues with the patch bump on this dependency

calebcartwright:
I also tried to add some environment variables to skip the husky git hook setup and the cypress binary install via the steps outlined [here](https://devcenter.heroku.com/articles/nodejs-support#environment-variables), but they did not seem to flow through to the review app

paulmelnikow:
You could try adding the env bar in app.json.

calebcartwright:
> You could try adding the env bar in app.json.

Ah! Thanks for that. I was looking for a `buildpack` file for some reason, and when i didn't find it I just used the `heroku cli`

paulmelnikow:
Cool! Yea, by default env vars don’t propagate from the staging app to the review apps (which is a good thing, really).

paulmelnikow:
Do you think we should put this in a postinstall script temporarily? The problem isn’t specific to Heroku.

calebcartwright:
> Do you think we should put this in a postinstall script temporarily? The problem isn’t specific to Heroku.

Will do

calebcartwright:
> Cool! Yea, by default env vars don’t propagate from the staging app to the review apps (which is a good thing, really).

Definitely! I had explicitly specified my review app though (`heroku config:set CYPRESS_INSTALL_BINARY=0 --app shields-staging-pr-3904`) which I figured would have worked. 

I also tried setting them in `app.json` but they don't seem to be working their either 🤷‍♂

app.json
```diff
@@ -6,6 +6,16 @@
   "repository": "https://github.com/badges/shields",
   "logo": "http://shields.io/favicon.png",
   "env": {
+    "CYPRESS_INSTALL_BINARY": {
+      "description": "Disable the cypress binary installation",
+      "value": 0,
```
paulmelnikow:
```suggestion
      "value": "0",
```

Not sure if this will help, but it's worth a try!

app.json
```diff
@@ -6,6 +6,16 @@
...
+      "required": false
+    },
+    "HUSKY_SKIP_INSTALL": {
+      "description": "Skip the husky git hook setup",
+      "value": 1,
```
calebcartwright:
```suggestion
      "value": "1",
```

calebcartwright:
Even using strings for the env values in app.json didn't seem to have an impact
```
...
[19:16:11]  Unzipping Cypress        98% 1s [title changed]
       [19:16:11]  Unzipping Cypress        99% 1s [title changed]
       [19:16:12]  Unzipping Cypress        100% 0s [title changed]
       [19:16:12]  Unzipped Cypress        [title changed]
       [19:16:12]  Unzipped Cypress        [completed]
       [19:16:12]  Finishing Installation  [started]
       [19:16:12]  Finished Installation   /app/.cache/Cypress/3.4.1 [title changed]
       [19:16:12]  Finished Installation   /app/.cache/Cypress/3.4.1 [completed]
       [?25h
       You can now open Cypress by running: node_modules/.bin/cypress open
       
       https://on.cypress.io/installing-cypress
```

paulmelnikow:
Gosh, so much strangeness!

I'm guessing the problem we are running into is a recent regression in npm, and something we ought to report so it can be fixed. I haven't had much time at the keyboard in recent weeks and haven't had too much time to dig into this myself. Though its effects are kind of far-reaching (e.g. #3901) and I feel like we need to get to the bottom of it.

calebcartwright:
Perhaps this is related? https://npm.community/t/6-11-1-some-dependencies-are-no-longer-being-installed/9586

paulmelnikow:
It looks like that is the solution: `npm install -g npm@6.10.3` makes the problem go away.

Steps to reproduce the problem:

```sh
npm install -g npm@6.11.2
rm -rf node_modules
npm install
ls -d node_modules/is-css-color
```

calebcartwright:
I can't approve this as the author but 👍 from me

PR #3909, opened 2019-08-23
Fix lockfile to restore Heroku deploys

paulmelnikow:
Same idea as #3904.

Last known working was 316749bd69b684274b2ffd952f9c0872849d82af so I reverted to that and then ran npm install.

I'm on npm@6.11.2.


paulmelnikow:
Big nope!

PR #3913, opened 2019-08-25
add support for [GitHubActions] via API

calebcartwright:
Refs #2574. An alternative approach to #3898 that leverages the GH v3 Check Suites API instead of badge scraping the native GH Workflow result badge.

Although I still prefer this approach to #3898, there is one downside in instances where a repo has more than one Workflow running on a given ref (branch, etc.) vs. #3898. 

In such a scenario, the response from the [check-suites](https://developer.github.com/v3/checks/suites/#list-check-suites-for-a-specific-ref) will include multiple records for each Workflow but there's no way to associate the record in the API response to a named Workflow, at least not that I've been able to find. For example, https://api.github.com/repos/actions/setup-python/commits/master/check-suites

The badge scraping approach in #3898 does allow the user to explicitly specify which Workflow they want by workflow name, but it only works on the master branch. The API approach allows the user to pick the branch/ref, but not the specific workflow by name


services/github/github-api-provider.js
```diff
@@ -185,7 +185,8 @@ class GithubApiProvider {
         baseUrl,
         headers: {
           'User-Agent': 'Shields.io',
-          Accept: 'application/vnd.github.v3+json',
+          Accept:
+            'application/vnd.github.v3+json, application/vnd.github.antiope-preview+json',
           Authorization: `token ${tokenString}`,
         },
```
calebcartwright: Was the overriding of provided options deliberate here? I had to add the alternative media type required for this API here as it wasn't working from the service class due to the merge order

services/github/github-actions.service.js
```diff
@@ -0,0 +1,107 @@
...
+const { GithubAuthV3Service } = require('./github-auth-service')
+const { documentation, errorMessagesFor } = require('./github-helpers')
+const { NotFound } = require('..')
+
+const schema = Joi.object({
+  total_count: nonNegativeInteger,
+  check_suites: Joi.array()
+    .items(
+      Joi.object({
+        status: Joi.string().required(),
```
calebcartwright: We _could_ probably try to use a stricter validation here for status and conclusion, although I've intentionally kept these at `Joi.string()`. AFAICT, GitHub apps can update the list of possible status values that they use, which would flow through here (since this API returns all Check Suites, not just GH Actions Check Suites). 

I believe that if we tried to restrict the allowed strings for these fields that we'd inevitably end up with users receiving "invalid response data" badges

calebcartwright:
Failing service tests were other unrelated GitHub badge tests (and didn't fail in `services@node-latest`) so I'm going to update the PR title to only run the tests for this new service

services/github/github-api-provider.js
chris48s: I'm sure there is an open issue about the way we're merging options, but I can't find it for the life of me.

We often merge in the other order e.g: https://github.com/badges/shields/blob/c9cd77a7468f495575743fbb5d37ddd217c91098/core/base-service/base-json.js#L43-L46

but the downside of this is if we want to pass an extra header, we have to re-add `Accept: 'application/json'` (because the spread operation doesn't nest, if you see what I mean)

I suspect the reason we've done it in the other order here is so that if we pass any header, its not going to overwrite the `Authorization` header.

In general, we need a more sophisticated merge starategy for these `options` objects.

services/github/github-actions.service.js
chris48s: yes I think GH Actions is sufficiently flexible that its hard to reliably require something more restrictive here

services/github/github-actions.service.js
```diff
@@ -0,0 +1,107 @@
...
+  static get defaultBadgeData() {
+    return {
+      label: 'workflow',
+    }
+  }
+
+  static render({ status, conclusion }) {
+    if (status !== 'completed') {
+      return {
+        message: 'in progress',
```
PyvesB: `in progress` should probably be added to the list of `otherStatuses` in _build-status.js_. The test 'valid repo and ref' may occasionally fail otherwise. I would also suggest using `renderBuildStatusBadge` for the 'in progress' value as well: it won't change anything with the current implementation, but if ever we decide to change the colour or something else, this build badge will automatically be covered as well. 😉

services/github/github-actions.service.js
```diff
@@ -0,0 +1,107 @@
...
+          Accept: 'application/vnd.github.antiope-preview+json',
+        },
+      },
+      schema,
+      errorMessages: errorMessagesFor('repo or ref not found'),
+    })
+  }
+
+  transform(json) {
+    if (json.total_count === 0) {
```
PyvesB: Do we need this extra check? Wouldn't the `if (!actionsSuite)` get us covered as well?

calebcartwright:
Thanks for reviewing @PyvesB! 

I'm most likely going to just close this PR. I understand (and agree with) the preference for the API over badge scraping, but this approach simply isn't viable without the ability to utilize the workflow name (it's as core as being able to specify the GH owner and repo). 

The badge scraping approach is the only option that's available right now, and I've updated that PR with more detail in https://github.com/badges/shields/pull/3898#issuecomment-554768572.

The Check Suite and Status APIs are not specific Actions/Workflow, but perhaps GitHub will eventually roll out Actions/Workflow APIs that we could switch to down the road. However, IMO we should proceed with the badge scraping approach as there is a huge community demand for this badge that we can meet, today, with that approach.

PyvesB:
Could we simply ask for the workflow name feature on the GitHub community forums? GitHub staff have been quite good at solving issues and responding to user queries regarding GitHub actions. I will look at the other PR in the evening.

PyvesB:
This feature request may well do the trick when implemented: https://github.community/t5/GitHub-API-Development-and/How-to-retrieve-the-full-list-of-Actions-workflow-runs-shown-on/m-p/39171

calebcartwright:
Closing this as #3898 has been merged and deployed

arcanis:
GitHub Actions now have a fresh new API: https://developer.github.com/v3/actions/

PR #3933, opened 2019-09-01
Handle missing [SymfonyInsightGrade] gracefully

### LINKED ISSUE #3784: ValidationError: "message" is required
Sentry Issue: [SHIELDS-2W](https://sentry.io/organizations/shields/issues/1125304010/?referrer=github_integration)

```
ValidationError: {
  "label": "grade",
  [41m"message"[0m[31m [1]: -- missing --[0m
}
[31m
[1] "message" is required[0m
  File "/home/m/shields/core/base-service/base.js", line 379, in Function.invoke
    Joi.assert(serviceData, serviceDataSchema)
  ?, in null.<anonymous>
...
(3 additional frame(s) were not displayed)
```

PyvesB:
Fixes  #3784.


services/symfony/symfony-insight-grade.service.js
```diff
@@ -30,7 +30,7 @@ module.exports = class SymfonyInsightGrade extends SymfonyInsightBase {
     ]
   }
 
-  static render({ status, grade }) {
+  static render({ status, grade = 'none' }) {
```
calebcartwright: Though I think this would indeed prevent the occurence of the issue in #3784, I wonder if there's not an issue in the schema that's somehow resulting in `grade` having a falsy value

For example: https://github.com/badges/shields/blob/3874fa861e98d274a0bd649beeeb66c4547be7ec/services/symfony/symfony-insight-base.js#L22 could probably use a `.required()` right?

If I'm following the code correctly, in order for this to have caused the issue in #3784, the `status` of the scan would have to be `finished` but the `grade` would have to be `undefined` which seems odd based on what I remember from the Symfony grade scale

PyvesB: I thought about modifying the schema as well, however, without additional changes to how the base class works, this would also impact the other Symfony badges that don't care about the grade and can cope with responses that are missing that parameter.

> which seems odd based on what I remember from the Symfony grade scale

I wanted to try to build a real example to confirm this, but I couldn't even manage to get API responses from the test project I set up, it kept on reject my credentials for some reason... So yeah, a bit in the dark here! 😄

calebcartwright: > this would also impact the other Symfony badges that don't care about the grade and can cope with responses that are missing that parameter.

Ahhh that's right, good call!

calebcartwright: > I wanted to try to build a real example to confirm/deny this, but I couldn't even manage to get API responses from the test project I set up, it kept on reject my credentials for some reason

Yeah I remember having a lot of trouble with it myself. It's a tough one to work with

PR #3934, opened 2019-09-01
Tweak Gitpod config

paulmelnikow:
This disables the comments like https://github.com/badges/shields/pull/3933#issuecomment-526905743 which seem a bit chatty, as the review apps can be accessed from the checks.

This and also turns off builds for branches. For the most part we keep PRs open when branches are at a stage of being ready for useful review, so I think that should be sufficient for us.


PyvesB:
Maybe a future refinement would be to include the Gitpod link in the shields-ci message generated by Danger? Like that we would have one automated message per PR, but something easier to see than the PR status.

paulmelnikow:
> Maybe a future refinement would be to include the Gitpod link in the shields-ci message generated by Danger? Like that we would still have only one automated message per PR, but something easier to find than the PR status.

We could do that. Could also be worth highlighting that a review app can be deployed. I also think it's pretty easy to find in the checks, so that could work well enough, too.

Issue #3874, opened 2019-08-15
Intermittent 502 responses from camo.githubusercontent.com


adamjstone:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

Badges intermittently fail to load. Periodically, I am seeing 502 responses from camo.githubusercontent.com for random badges. The behavior resolves itself, usually, after one or two page refreshes, which leads me to think there are timeouts happening somewhere which eventually resolve themselves via caching.

:link: **Link to the badge**

No specific badges are implicated in this. I see it across numerous badges on numerous pages, including the GitHub README for the badges/shields repo itself. See the screen captures below.

- - -
![shields-repo-readme](https://user-images.githubusercontent.com/8525409/63095802-c3c0b580-bf31-11e9-9c9c-fb18ef28cb36.png)
- - -
![si-repo-readme](https://user-images.githubusercontent.com/8525409/63095808-cc18f080-bf31-11e9-850f-50adb67c15dd.png)
- - -
![shieds-response-info](https://user-images.githubusercontent.com/8525409/63095818-d20ed180-bf31-11e9-8bb2-1139a4e9ad0a.jpg)

adamjstone:
May or may not be related to:

https://github.com/badges/shields/issues/1245

paulmelnikow:
I'm seeing these too. They happen frequently while deploys are happening, and off and on and other times. To fix them for the deployment, we need to switch to a zero-downtime deployment. To fix them the rest of the time, I wonder if we need to add server capacity. We started adding a fourth server but that got held up.

I wonder if migrating to Heroku and bumping up our concurrency would solve both problems.

paulmelnikow:
I keep seeing these, even while deploys aren't running. I wonder if we should experiment with adding a single dyno on Heroku to see if more capacity solves the problem.

paulmelnikow:
Using Cloudflare with Heroku requires a CNAME so it's not an option to add a fourth server to our existing round-robin DNS.

We could run an experiment for several hours, though. We could set Heroku to four dynos, and point Cloudflare to that, and see how the reliability compares.

To use Heroku as a long-term option, I think there are three problems we'd need to solve:

1. Decide who will have access to see the configuration and manage the deployments (needs to be more than one active maintainer, but probably should _not_ be the entire core team) (#2577)
[...]

PR #3662, opened 2019-07-07
Metrics for a response size of an upstream service

platan:
This PR adds a metric for response size of an upstream service. I chose a histogram metric, mainly because in the histogram "Observations are very cheap as they only need to increment counters." (https://prometheus.io/docs/practices/histograms/). In histograms observations fall into buckets. Here we have buckets: 64kB, 128kB, ...,4MB, 8MB. Moreover I added a separate metric (reasons described in https://github.com/badges/shields/pull/3656#pullrequestreview-258613471).

How to test it?
```bash
METRICS_PROMETHEUS_ENABLED=true npm run start

# dynamic XML
curl "http://localhost:8080/badge/dynamic/xml.svg?color=red&label=label1&query=%2F%2Fname&url=https%3A%2F%2Fraw.githubusercontent.com%2Fhsz%2Fidea-gitignore%2Fmaster%2Fresources%2FMETA-INF%2Fplugin.xml"

# dynamic YAML
curl "http://localhost:8080/badge/dynamic/yaml.svg?color=red&label=label1&query=%24.trailingComma&url=https%3A%2F%2Fraw.githubusercontent.com%2Fbadges%2Fshields%2Fmaster%2F.prettierrc.yml"

# dynamic JSON
curl "http://localhost:8080/badge/dynamic/json.svg?color=red&label=label1&query=%24.name&url=https%3A%2F%2Fraw.githubusercontent.com%2Fbadges%2Fshields%2Fmaster%2Fpackage.json"

# non-dynamic
curl "http://localhost:8080/appveyor/ci/gruntjs/grunt.svg"

# results
curl "http://localhost:8080/metrics" -s | grep "service_response_bytes"
```
<details>
<summary>example result</summary>

```
# HELP service_response_bytes Service response size in bytes
# TYPE service_response_bytes histogram
service_response_bytes_bucket{le="65536",type="dynamic-xml"} 1
service_response_bytes_bucket{le="131072",type="dynamic-xml"} 1
service_response_bytes_bucket{le="262144",type="dynamic-xml"} 1
service_response_bytes_bucket{le="524288",type="dynamic-xml"} 1
service_response_bytes_bucket{le="1048576",type="dynamic-xml"} 1
service_response_bytes_bucket{le="2097152",type="dynamic-xml"} 1
service_response_bytes_bucket{le="4194304",type="dynamic-xml"} 1
service_response_bytes_bucket{le="8388608",type="dynamic-xml"} 1
[...]

PR #3947, opened 2019-09-03
Refactor existing metrics support into MetricHelper

paulmelnikow:
This completes the refactor suggested at https://github.com/badges/shields/pull/3662#issuecomment-509011229 in anticipation of adding more metrics support, such as response size of an upstream service, or response time.

Marking blocker as it's related to measuring response time for #3874.

I haven't tested this yet though we should follow "how to test it" in the top comment of https://github.com/badges/shields/pull/3662#issue-295109085.


paulmelnikow:
Closing in favor of #3948.

PR #3948, opened 2019-09-03
Add a response-time metric

paulmelnikow:
This completes the refactor suggested at https://github.com/badges/shields/pull/3662#issuecomment-509011229 and adds a response-time metric for #3874.

I haven't tested this yet though we should follow "how to test it" in the top comment of https://github.com/badges/shields/pull/3662#issue-295109085.

Supersedes #3947.

### DIFF
--- a/core/base-service/metric-helper.js
+++ b/core/base-service/metric-helper.js
@@ -3,14 +3,13 @@
 class MetricHelper {
   constructor({ metricInstance }, { category, serviceFamily, name }) {
     if (metricInstance) {
-      this._serviceRequestCounter = metricInstance.createServiceRequestCounter({
+      this.serviceRequestCounter = metricInstance.createServiceRequestCounter({
         category,
         serviceFamily,
         name,
       })
     } else {
-      // When metrics are disabled, use a mock counter.
-      this._serviceRequestCounter = { inc: () => {} }
+      this.serviceRequestCounter = undefined
     }
   }
 
@@ -20,7 +19,10 @@ class MetricHelper {
   }
 
   noteResponseSent() {
-    this._serviceRequestCounter.inc()
+    const { serviceRequestCounter } = this
+    if (serviceRequestCounter) {
+      serviceRequestCounter.inc()
+    }
   }
 }
 
--- a/core/base-service/metric-helper.js
+++ b/core/base-service/metric-helper.js
@@ -3,7 +3,7 @@
 class MetricHelper {
   constructor({ metricInstance }, { category, serviceFamily, name }) {
     if (metricInstance) {
-      this.serviceRequestCounter = metricInstance.createServiceRequestCounter({
+      this.serviceRequestCounter = metricInstance.createNumRequestCounter({
         category,
         serviceFamily,
         name,
--- a/core/base-service/metric-helper.js
+++ b/core/base-service/metric-helper.js
@@ -1,0 +1,27 @@
+'use strict'
+
+class MetricHelper {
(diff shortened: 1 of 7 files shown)


paulmelnikow:
I turned on metrics for the review app and you can see the result here:

https://shields-staging-pr-3948.herokuapp.com/metrics

core/base-service/metric-helper.js
```diff
@@ -0,0 +1,45 @@
...
+    const requestStartTime = performance.now()
+
+    return {
+      noteResponseSent() {
+        if (metricInstance) {
+          const elapsedTime = performance.now() - requestStartTime
+          metricInstance.noteResponseTime(elapsedTime)
+        }
+
+        if (serviceRequestCounter) {
```
calebcartwright: if `metricInstance`  is falsy then `serviceRequestCounter` will always be falsy too right? If so then could we move this inside the above `if` block?

core/base-service/metric-helper.js
```diff
@@ -0,0 +1,45 @@
+'use strict'
+
+const { performance } = require('perf_hooks')
```
calebcartwright: i had no idea this existed in Node 👍

core/base-service/metric-helper.js
paulmelnikow: That's true, though they'll both be truthy in production. 🤷‍♂

core/base-service/metric-helper.js
paulmelnikow: I did a quick test that it exists in Node 9. 😁

paulmelnikow:
@platan In the interest of moving forward with #3874 I'm going to deploy this, though if you have any input on how this was done, please feel free to comment and/or open a follow-on PR!

paulmelnikow:
This is live and I tried to add it to Grafana. I wasn't able to update the live version – I got this message:

Screen Shot 2019-09-03 at 7 01 02 PM

However I was able to make a copy for now: https://metrics.shields.io/d/service-response-time/service-response-time?orgId=1

I _think_ it's working correctly.

calebcartwright:
Awesome! 

One minor note, it looks like the response time has an entire object:
```
{env="shields-io-production",instance="s0.shields-server.com:443",job="shields-io-production",service="shields"}
```
instead of just the instance like we have on the other dashboard

paulmelnikow:
I'm not sure how to fix that, though if you figure it out I'd happily make the change!

platan:
@calebcartwright you want to see an instance name only? You have to edit a chart and put `{{instance}}` as a `Legend` value. 

Screenshot_2019-09-04 service-response-time - Grafana. 

[The result](https://metrics.shields.io/d/service-response-time/service-response-time?orgId=1&var-env=shields-io-production&var-instance=All&from=now-5m&to=now&fullscreen&panelId=11).

calebcartwright:
> @calebcartwright you want to see an instance name only? You have to edit a chart and put `{{instance}}` as a `Legend` value.

Perfect thanks @platan!

platan:
@paulmelnikow This change looks really nice and gives a simple option to create more metrics! I will try to create more charts using this new data. 

Dashboards are provisioned from https://github.com/platan/metrics-shields-io-config. I will move new charts from `service-response-time` dashboard to `services` dashboard.

platan:
I've moved "Service response time" chart to "services" dashboard.

jsheckells78:
https://github.com/badges/shields/commit/8868cdc503c465cfaabe9d101cc71cb1966bd10f

PR #3178, opened 2019-03-07
[WIP] [Netlify] status badge

AnandChowdhary:
For #3129

PR #3953, opened 2019-09-05
Add [Netlify] (fixed #3129)

### LINKED ISSUE #3129: Netlify status badge
:clipboard: **Description**

The popular CDN and static site host Netlify has added support for badges with build information (e.g., 
"success", "building" or "failed").

Badges can be added like: "netlify | success", just like the current "build | passing".

Details: https://www.netlify.com/docs/continuous-deployment/#status-badges

Example:

Netlify badge example

:link: **Data**

A general API is available at https://www.netlify.com/docs/api/. This issue will be updated once the documentation for the particular build status API are published too.

:microphone: **Motivation**

Lots of websites are now using Netlify as opposed to GitHub Pages. According to their website, over 400,000 "developers trust Netlify", probably since it's free and has a great toolkit (e.g., Netlify CMS).

Since they have badges available now, they should be added to Shields.io too!

AnandChowdhary:
This PR adds a status badge for Netlify, the static CDN and fixes #3953.

I tried making a PR in #3178 from scratch by using a "closest color matcher" many months ago, but I wasn't able to complete it. This time, it uses the SVG scraping service, and works pretty well (as long as Netlify doesn't update their badge colors, after which we'll update the service).

Thanks for the community for being so patient!

### DIFF
--- a/services/build-status.js
+++ b/services/build-status.js
@@ -21,8 +21,9 @@ const redStatuses = [
   'infrastructure_failure',
 ]
 
+const yellowStatuses = ['building']
+
 const otherStatuses = [
-  'building',
   'canceled',
   'cancelled',
   'expired',
@@ -56,6 +57,9 @@ function renderBuildStatusBadge({ label, status }) {
   if (greenStatuses.includes(status)) {
     message = 'passing'
     color = 'brightgreen'
+  } else if (yellowStatuses.includes(status)) {
+    message = 'building'
+    color = 'yellow'
   } else if (orangeStatuses.includes(status)) {
     message = status === 'partially succeeded' ? 'passing' : status
     color = 'orange'
--- a/services/netlify/netlify.service.js
+++ b/services/netlify/netlify.service.js
@@ -58,7 +58,7 @@ module.exports = class Netlify extends BaseSvgScrapingService {
 
   static render({ status }) {
     status = statusMap[status] || status
-    let result = renderBuildStatusBadge({ status })
+    const result = renderBuildStatusBadge({ status })
     if (result.message === 'building' && !result.color) {
       result.color = 'yellow'
     }
--- a/services/netlify/netlify.service.js
+++ b/services/netlify/netlify.service.js
@@ -34,16 +34,17 @@ module.exports = class Netlify extends BaseSvgScrapingService {
         title: 'Netlify',
         namedParams: {
           projectId: 'e6d5a4e0-dee1-4261-833e-2f47f509c68f',
-          documentation: 'To locate your project id, visit your project settings, scroll to "Status badges" under "General", and copy the ID between "/api/v1/badges/" and "/deploy-status" in the code sample'
         },
+        documentation:
+          'To locate your project id, visit your project settings, scroll to "Status badges" under "General", and copy the ID between "/api/v1/badges/" and "/deploy-status" in the code sample',
         staticPreview: renderBuildStatusBadge({ status: 'passing' }),
-      }
+      },
     ]
   }
 
   static get defaultBadgeData() {
     return {
-      label: 'netlify'
+      label: 'netlify',
     }
   }
 
--- a/services/netlify/netlify.service.js
+++ b/services/netlify/netlify.service.js
@@ -1,0 +1,77 @@
+'use strict'
+
+const { renderBuildStatusBadge } = require('../build-status')
+const { BaseSvgScrapingService } = require('..')
+
+const pendingStatus = 'building'
+const notBuiltStatus = 'not built'
+
+const statusMap = {
+  building: pendingStatus,
+  waiting: pendingStatus,
+  initiated: pendingStatus,
+  stopped: notBuiltStatus,
+  ignored: notBuiltStatus,
+  blocked: notBuiltStatus,
+  infrastructure_failure: 'failed',
+}
+
+module.exports = class Netlify extends BaseSvgScrapingService {
+  static get category() {
+    return 'build'
+  }
+
+  static get route() {
+    return {
+      base: 'netlify',
+      pattern: ':projectId/:branch*',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Netlify',
+        pattern: ':projectId',
+        namedParams: {
+          projectId: 'e6d5a4e0-dee1-4261-833e-2f47f509c68f',
+        },
+        staticPreview: renderBuildStatusBadge({ status: 'passing' }),
+      },
+      // {
+      //   title: 'Netlify (branch)',
+      //   pattern: ':projectId/:branch',
+      //   namedParams: {
+      //     projectId: '0bdb0440-3af5-0133-00ea-0ebda3a33bf6',
+      //     branch: 'master',
+      //   },
+      //   staticPreview: renderBuildStatusBadge({ status: 'passing' }),
+      // },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'netlify' }
+  }
+
+  static render({ status }) {
+    status = statusMap[status] || status
+    return renderBuildStatusBadge({ status })
+  }
+
+  async fetch({ projectId, branch }) {
+    const url = `https://api.netlify.com/api/v1/badges/${projectId}/deploy-status`
+    const { buffer } = await this._request({
+      url,
+    })
+    if (buffer.includes('#EAFAF9')) return { message: 'passing' }
+    if (buffer.includes('#FFF3F4')) return { message: 'failing' }
+    if (buffer.includes('#FEFAEA')) return { message: 'building' }
+    return { message: 'project not found' }
+  }
+
+  async handle({ projectId, branch }) {
+    const { message: status } = await this.fetch({ projectId, branch })
+    return this.constructor.render({ status })
+  }
+}
--- a/services/netlify/netlify.service.js
+++ b/services/netlify/netlify.service.js
@@ -32,7 +32,6 @@ module.exports = class Netlify extends BaseSvgScrapingService {
     return [
       {
         title: 'Netlify',
-        pattern: ':projectId',
         namedParams: {
           projectId: 'e6d5a4e0-dee1-4261-833e-2f47f509c68f',
         },
--- a/services/netlify/netlify.service.js
+++ b/services/netlify/netlify.service.js
@@ -50,7 +50,10 @@ module.exports = class Netlify extends BaseSvgScrapingService {
   }
 
   static get defaultBadgeData() {
-    return { label: 'netlify' }
+    return {
+      label: 'netlify',
+      namedLogo: 'netlify',
+    }
   }
 
   static render({ status }) {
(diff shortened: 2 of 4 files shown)


AnandChowdhary:
Hi @paulmelnikow, I gave this another shot after #3178. I'm sure this also will need some changes, but it works-ish?

services/netlify/netlify.service.js
```diff
@@ -0,0 +1,77 @@
...
+
+module.exports = class Netlify extends BaseSvgScrapingService {
+  static get category() {
+    return 'build'
+  }
+
+  static get route() {
+    return {
+      base: 'netlify',
+      pattern: ':projectId/:branch*',
```
ExE-Boss: Shouldn’t this just be `:projectId`?

```suggestion
      pattern: ':projectId',
```

services/build-status.js
```diff
@@ -21,8 +21,9 @@ const redStatuses = [
   'infrastructure_failure',
 ]
 
+const yellowStatuses = ['building']
+
 const otherStatuses = [
-  'building',
```
calebcartwright: There's a lot of badges using this helper and I think this change will have negative impacts on other badges. If you'd prefer to have a `yellow` color for the `building` status in Netflify then IMO you should override that color directly within the Netflify service vs. changing it here

services/netlify/netlify.service.js
```diff
@@ -0,0 +1,77 @@
...
+      base: 'netlify',
+      pattern: ':projectId',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Netlify',
+        pattern: ':projectId',
```
calebcartwright: You can remove this line. You can use `pattern` in an example if you need to do something different than the route pattern, but no need to redfine pattern here if it's the same

ExE-Boss: This `pattern` used to be different before https://github.com/badges/shields/pull/3953/commits/4369658193b48011d582701ae01d96d8b1916c26, which changed the one in `route`.

services/netlify/netlify.service.js
```diff
@@ -0,0 +1,77 @@
...
+      //     projectId: '0bdb0440-3af5-0133-00ea-0ebda3a33bf6',
+      //     branch: 'master',
+      //   },
+      //   staticPreview: renderBuildStatusBadge({ status: 'passing' }),
+      // },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'netlify' }
```
ExE-Boss: I recommend adding the Netlify icon here to match what Netlify is doing:
```suggestion
    return {
      label: 'netlify',
      namedLogo: 'netlify',
    }
```

AnandChowdhary: Understood, thanks!

services/build-status.js
ExE-Boss: Also, this is causing the build to fail.

```suggestion
  'building',
```

services/netlify/netlify.tester.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+const { isBuildStatus } = require('../build-status')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('netlify (valid, no branch)')
+  .get('/e6d5a4e0-dee1-4261-833e-2f47f509c68f.json')
+  .expectBadge({
+    label: 'build',
```
calebcartwright:
```suggestion
    label: 'netflify',
```

services/netlify/netlify.tester.js
```diff
@@ -0,0 +1,15 @@
...
+    message: isBuildStatus,
+  })
+
+t.create('netlify (repo not found)')
+  .get('/not-a-repo.json')
+  .expectBadge({ label: 'build', message: 'project not found' })
```
calebcartwright:
```suggestion
  .expectBadge({ label: 'netlify', message: 'project not found' })
```

ExE-Boss:
I’ve opened https://github.com/AnandChowdhary/shields/pull/6 to fix the failing tests.

services/netlify/netlify.tester.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+const { isBuildStatus } = require('../build-status')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('netlify (valid, no branch)')
+  .get('/e6d5a4e0-dee1-4261-833e-2f47f509c68f.json')
+  .expectBadge({
+    label: 'netflify',
```
ExE-Boss:
```suggestion
    label: 'netlify',
```

services/netlify/netlify.service.js
```diff
@@ -0,0 +1,83 @@
...
+  static get defaultBadgeData() {
+    return {
+      label: 'netlify',
+      namedLogo: 'netlify',
+    }
+  }
+
+  static render({ status }) {
+    status = statusMap[status] || status
+    let result = renderBuildStatusBadge({ status })
```
ExE-Boss: The linter requires this to be `const`.

```suggestion
    const result = renderBuildStatusBadge({ status })
```

calebcartwright:
hey @jankeromnes - looks like the Gitpod workspace here has been stuck in a pending state for the last few hours. It's not a problem for us or anything, just sharing for awareness

services/netlify/netlify.service.js
```diff
@@ -0,0 +1,83 @@
...
+
+  async fetch({ projectId, branch }) {
+    const url = `https://api.netlify.com/api/v1/badges/${projectId}/deploy-status`
+    const { buffer } = await this._request({
+      url,
+    })
+    if (buffer.includes('#EAFAF9')) return { message: 'passing' }
+    if (buffer.includes('#FFF3F4')) return { message: 'failing' }
+    if (buffer.includes('#FEFAEA')) return { message: 'building' }
+    return { message: 'project not found' }
```
calebcartwright: IIRC, a 404 response on the request to the above URL will automatically result in a `not found` badge being returned (I think this is why that not found test was failing before). You can also customize that not found message via something like `errorMessages: { 404: 'project not found' },` above alongside the `url` arg to `_request`. 

I'm not familiar with Netlify at all, but if we get down to this line here, is it really a case of the project not being found? The request would've had to have returned an HTTP 200 to get here, so just want to confirm this really is a project not found scenario and not more of an 'unknown' status. 

If it truly is a not found scenario, then I'd suggest throwing a NotFound error here instead of returning a 'project not found' status and then going through the render logic. To do that would be a:

update line 4 above to:
```js
const { BaseSvgScrapingService, NotFound } = require('..')
```

and then here line 76 would be:
```js
    throw new NotFound({ prettyMessage: 'project not found' })
```

ExE-Boss: This should instead probably be:
```suggestion
    return { message: 'unknown' }
```

In case **Netlify** changes the colours of their badge icons.

AnandChowdhary: It turns out Netlify returns a JSON object instead of the SVG when a project is not found: https://api.netlify.com/api/v1/badges/not-found/deploy-status. Maybe we can use that for "not found" and "error" as the fallback.

AnandChowdhary: It also correctly returns HTTP `404` so @calebcartwright you're right, it automatically goes to the "not found" badge. Therefore the fallback should be "unknown" as @Exe-Boss suggested.

services/netlify/netlify.service.js
```diff
@@ -0,0 +1,83 @@
...
+      pattern: ':projectId',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Netlify',
+        namedParams: {
+          projectId: 'e6d5a4e0-dee1-4261-833e-2f47f509c68f',
```
calebcartwright: Are Netlify users well versed in how to retrieve their project id? If not, would it be helpful to include some inline docs/link to docs to give them guidance on how to retrieve it? 

Adding a documentation key like the example below would make the info visible to users directly in Shields (in the badge creation window for Netlify)
```suggestion
          projectId: 'e6d5a4e0-dee1-4261-833e-2f47f509c68f',
          documentation: 'To locate your project id, go to....',
```

services/netlify/netlify.service.js
```diff
@@ -0,0 +1,83 @@
...
+      },
+      // {
+      //   title: 'Netlify (branch)',
+      //   pattern: ':projectId/:branch',
+      //   namedParams: {
+      //     projectId: '0bdb0440-3af5-0133-00ea-0ebda3a33bf6',
+      //     branch: 'master',
+      //   },
+      //   staticPreview: renderBuildStatusBadge({ status: 'passing' }),
+      // },
```
calebcartwright: Can we remove this commented out block?

services/netlify/netlify.service.js
```diff
@@ -0,0 +1,83 @@
...
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'netlify',
+      namedLogo: 'netlify',
```
calebcartwright: Let's remove this and let folks that want the logo be able to opt into it via the standard query param

For example:
https://shields-staging-pr-3953.herokuapp.com/netlify/e6d5a4e0-dee1-4261-833e-2f47f509c68f?logo=netlify

services/netlify/netlify.spec.js
```diff
@@ -0,0 +1,19 @@
...
+  test(Netlify.render, () => {
+    given({ status: 'building' }).expect(building)
+    given({ status: 'stopped' }).expect(notBuilt)
+    given({ status: 'infrastructure_failure' }).expect({
+      message: 'failing',
+      color: 'red',
+      label: undefined,
+    })
+  })
+})
```
calebcartwright: thanks for including these! 👍

services/netlify/netlify.service.js
ExE-Boss: It looks like this got disabled recentlish for non‑social‑style badges, as I don’t see this applied to any non‑social badges, despite those badges having `namedLogo` specified.

services/netlify/netlify.service.js
AnandChowdhary: Makes sense, I'm adding that.

AnandChowdhary:
Thanks for being so accommodating, @calebcartwright and @ExE-Boss! I think we're ready for merging, please let me know if you'd like me to change anything else. Very excited for this! 😄

AnandChowdhary:
Seems like that previous comment was premature and the tests are all failing. Turns out `namedParams` doesn't accept `documentation`.

calebcartwright:
> Edit: It's not a namedParam, I misunderstood @calebcartwright's review.

Actually that was my fault, apologies, I put the suggestion on the wrong line 🤦‍♂

jankeromnes:
Super cool contribution! 💯

> hey @jankeromnes - looks like the Gitpod workspace here has been stuck in a pending state for the last few hours. It's not a problem for us or anything, just sharing for awareness
> 
> Edit - it went through after some new commits so nevermind 😄

Oops, thanks for letting us know! May have been a temporary issue, I'll take a look. Hopefully this doesn't happen too frequently.

PR #3955, opened 2019-09-07
[dynamicxml] wrap xpath errors

### LINKED ISSUE #3796: [DynamicXml] Error: XPath parse error
Sentry Issue: [SHIELDS-35](https://sentry.io/organizations/shields/issues/1129443635/?referrer=github_integration)

```
Error: XPath parse error
  File "/home/m/shields/services/dynamic/dynamic-xml.service.js", line 45, in DynamicXml.handle
    .select(pathExpression, parsed)
  ?, in null.<anonymous>
...
(2 additional frame(s) were not displayed)
```

chris48s:
Closes #3796

Usually we don't directly bounce the exception message back to the user like this, but I reckon if you're trying to construct an xpath query, this is probably the most useful output and better than a runtime error.

### DIFF
--- a/services/dynamic/dynamic-xml.service.js
+++ b/services/dynamic/dynamic-xml.service.js
@@ -41,9 +41,15 @@ module.exports = class DynamicXml extends BaseService {
 
     const parsed = new DOMParser().parseFromString(buffer)
 
-    const values = xpath
-      .select(pathExpression, parsed)
-      .map((node, i) => (pathIsAttr ? node.value : node.firstChild.data))
+    let values
+    try {
+      values = xpath.select(pathExpression, parsed)
+    } catch (e) {
+      throw new InvalidResponse({ prettyMessage: e.message })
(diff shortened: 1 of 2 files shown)


services/dynamic/dynamic-xml.service.js
```diff
@@ -41,9 +41,15 @@ module.exports = class DynamicXml extends BaseService {
...
     const parsed = new DOMParser().parseFromString(buffer)
 
-    const values = xpath
-      .select(pathExpression, parsed)
-      .map((node, i) => (pathIsAttr ? node.value : node.firstChild.data))
+    let values
+    try {
+      values = xpath.select(pathExpression, parsed)
+    } catch (e) {
+      throw new InvalidResponse({ prettyMessage: e.message })
```
paulmelnikow: Should this be InvalidParameter as it's as likely a problem with the query as the data?

PR #3956, opened 2019-09-07
[npmversion] require key 'latest' to exist

### LINKED ISSUE #3794: [NpmVersion] Error: Can't generate a version color for undefined
Sentry Issue: [SHIELDS-34](https://sentry.io/organizations/shields/issues/1127061701/?referrer=github_integration)

```
Error: Can't generate a version color for undefined
  File "/home/m/shields/services/color-formatters.js", line 11, in version
    throw new Error(`Can't generate a version color for ${version}`)
  File "/home/m/shields/services/version.js", line 142, in renderVersionBadge
    color: versionColor(version),
  File "/home/m/shields/services/npm/npm-version.service.js", line 71, in Function.render
    return renderVersionBadge({
  File "/home/m/shields/services/npm/npm-version.service.js", line 101, in NpmVersion.handle
  ?, in null.<anonymous>
...
(1 additional frame(s) were not displayed)
```

chris48s:
Closes #3794

I think the reason we're seeing this is because (for some reason) sometimes we get a response back from NPM with no `latest` key. This modifies the schema to require the `latest` key to be present in order to accept the response.

### DIFF
--- a/services/npm/npm-version.service.js
+++ b/services/npm/npm-version.service.js
@@ -9,6 +9,7 @@ const keywords = ['node']
 
 // Joi.string should be a semver.
 const schema = Joi.object()
+  .keys({ latest: Joi.string().required() })
   .pattern(/./, Joi.string())
   .required()
 
--- a/services/npm/npm-version.tester.js
+++ b/services/npm/npm-version.tester.js
@@ -49,3 +49,16 @@ t.create(
 t.create('invalid package name')
   .get('/frodo-is-not-a-package.json')
   .expectBadge({ label: 'npm', message: 'package not found' })
+
+t.create("Response doesn't include a 'latest' key")
+  .intercept(nock =>
+    nock('https://registry.npmjs.org')
+      .get('/-/package/npm/dist-tags')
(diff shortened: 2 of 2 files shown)


paulmelnikow:
Hmm, isn't it legal not to have a `latest` tag? I wonder if it would be better to say **no releases found for @latest**.

chris48s:
> isn't it legal not to have a `latest` tag?

Not according to the docs: https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#package

> `dist-tags`: an object with at least one key, `latest`, representing dist-tags

That's why we're expecting `latest` to exist in the default case. I think its reasonable to treat a `dist-tags` object without `latest` as an invalid/unexpected response as opposed to a valid condition if it doesn't match the documented behaviour.

Given NPM version is one of our most popular badges and we've seen this exception 16 times in the last month, my hypothesis is that we're seeing this as a result of some very exceptional behaviour from the upstream api. Unforuntately we're both debugging blind here, so we don't actually have an example of what that response looks like. I just made one up for the test.

paulmelnikow:
Makes sense. Thanks for explaining!

Issue #3929, opened 2019-08-28
Badge Request: securityheaders.com


srtab:
:clipboard: **Description**
 * Which service is this badge for
   * https://securityheaders.com/;
 * What sort of information should this badge show?
   * Grade `securityheaders | D`;
   * With the appropriate grade color
   * Link should point to https://securityheaders.com/?followRedirects=on&hide=on&q=github.com
 * Example:
   * ![SecurityHeaders](https://img.shields.io/badge/securityheaders-D-652770)

:link: **Data**
 * Is there a public API?
   * Kind of, it's possible to obtain the grade through HTTP headers. See API link.
 * Does the API requires an API key?
   * No
 * Link to the API documentation.
    * https://github.com/securityheaders/securityheaders-bugs/issues/17#issuecomment-246103211

:microphone: **Motivation**
Show Website HTTP security headers rating/score, and promote HTTP header configuration for a safer internet. See real time rating, and detect miss configurations.

PyvesB:
Hello @srtab!

Thanks for the suggestion, sounds like a good idea. At some point our Code Climate badges were parsing HTTP headers to render our badge information. Even though this no longer is the case, I believe this would be an acceptable approach. Worth mentioning that another issue regarding a proper API was raised [here](https://github.com/securityheaders/securityheaders-bugs/issues/57).

Would you like to contribute to Shields and have a go at implementing this new badge? We've got a nice [tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) to help you get started. The [website badge](https://github.com/badges/shields/blob/master/services/website/website.service.js) also shows you how to do HEAD operations, which will probably be useful for the current use case. 😉 

Cheers,

Pyves

PR #3958, opened 2019-09-08
Added [SecurityHeaders] badge

srtab:
Pull request with #3929 contribution.


PyvesB:
Thanks for your contribution @srtab!

I've just edited the title and kicked the CI so that the new service tests get run as well.

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,91 @@
+'use strict'
+
+const { BaseService } = require('..')
+
+const documentation = `
+<p>
+  The <a href="https://securityheaders.com/">Security Headers</a>
+  provide a easy mechanism to analyze HTTP response headers and give information on how to deploy missing headers.
```
PyvesB: Typo: _an_ easy.

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,91 @@
...
+          grade: 'A+',
+        }),
+        documentation,
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'securityheaders',
```
PyvesB: I'm wondering whether the badge label should be in two words as well.

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,91 @@
...
+
+module.exports = class SecurityHeaders extends BaseService {
+  static get category() {
+    return 'monitoring'
+  }
+
+  static get route() {
+    return {
+      base: 'security-headers',
+      pattern: ':protocol(https|http)/:host',
```
PyvesB: I would suggest using a query parameter instead, like I did for the [Bugzilla service](https://github.com/badges/shields/blob/master/services/bugzilla/bugzilla.service.js). It has become our preferred approach for new badges. 😉

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,91 @@
...
+      color: colorMap[grade],
+    }
+  }
+
+  async handle({ protocol, host }) {
+    let grade
+
+    try {
+      const { res } = await this._request({
+        url: `https://securityheaders.com/?q=${protocol}%3A%2F%2F${host}&hide=on&followRedirects=on`,
```
PyvesB: You can use `qs` options like what is done in the [Wercker service](https://github.com/badges/shields/blob/b3be4d94d5ef570b8daccfd088c343a958988843/services/wercker/wercker.service.js#L87). Like that, no issues with URL encoding!

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,91 @@
...
+        options: {
+          method: 'HEAD',
+        },
+      })
+
+      grade = res.headers['x-grade']
+    } catch (e) {
```
PyvesB: No need to handle errors explicitly here: they would be handled by the base classes in a consistent way!

srtab:
Thanks for the review @PyvesB, i will make the changes as soon as possible.

services/security-headers/security-headers.service.js
srtab: Can i define an empty pattern?

services/security-headers/security-headers.service.js
srtab: Switched to "security headers".

services/security-headers/security-headers.service.js
srtab: Changed to use `qs` options :)

services/security-headers/security-headers.service.js
srtab: Removed error handling.

services/security-headers/security-headers.service.js
PyvesB: Sorry, probably wasn't very clear. In the Bugzilla service, `baseUrl` is not only used for tests: it's used in production to select the Bugzilla host you're interested in, for instance: https://img.shields.io/bugzilla/545425?baseUrl=https%3A%2F%2Fbugs.eclipse.org%2Fbugs

So what I had in mind was something along the lines of: https://img.shields.io/security-headers?domain=https%3A%2F%2Fshields.io

@paulmelnikow do you have a strong opinion on this one?

services/security-headers/security-headers.tester.js
```diff
@@ -0,0 +1,20 @@
...
+
+t.create('grade of http://shields.io')
+  .get('/http/shields.io.json')
+  .expectBadge({ label: 'security headers', message: 'F', color: 'red' })
+
+t.create('grade of https://shields.io')
+  .get('/https/shields.io.json')
+  .expectBadge({ label: 'security headers', message: 'F', color: 'red' })
+
+t.create('grade when no network')
```
PyvesB: This test can now go away as it's behaviour that's already tested within `BaseService`.

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,84 @@
...
+        method: 'HEAD',
+        qs: {
+          q: `${protocol}://${host}`,
+          hide: 'on',
+          followRedirects: 'on',
+        },
+      },
+    })
+
+    const grade = res.headers['x-grade']
```
PyvesB: Is it worth checking whether grade is defined? Something along the lines of:
````
if (!grade) {
  throw new NotFound({ prettyMessage: 'not available' }
}

````

File: services/security-headers/security-headers.service.js (hunk 3 above)
Comment (calebcartwright) [reply services/security-headers/security-headers.service.js]: IMO we should use a queryParam here. We decided to leverage queryParams vs. route params to take in a user-provided url input to simplify the patterns, and we should be consistent here. 

There's probably still a couple of existing services that haven't been updated yet, but for a net new service like this one we should follow the queryParam approach. 

See https://github.com/badges/shields/pull/3633#discussion_r299231749 for more info.

Comment (srtab) [reply services/security-headers/security-headers.service.js]: @PyvesB Is it possible to create a service without `pattern`? I tried with empty `pattern` on route, and `namedParams: {}` on examples but i obtain this error: "Error: Example for SecurityHeaders at index 0 does not declare a pattern". Can you help me with this?

File: services/security-headers/security-headers.tester.js (hunk 6 above)
Comment (srtab) [reply services/security-headers/security-headers.tester.js]: Sure, completely agree!

File: services/security-headers/security-headers.service.js (hunk 3 above)
Comment (calebcartwright) [reply services/security-headers/security-headers.service.js]: That's a good question. I've added some inline suggestions you can make that should get it running again

File: services/security-headers/security-headers.service.js [hunk 8]
```diff
@@ -0,0 +1,84 @@
+'use strict'
+
+const { BaseService } = require('..')
```
calebcartwright:
```suggestion
const Joi = require('@hapi/joi')
const { optionalUrl } = require('../validators')
const { BaseService } = require('..')

const queryParamSchema = Joi.object({
  baseUrl: optionalUrl.required(),
}).required()
```

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,84 @@
...
+`
+
+module.exports = class SecurityHeaders extends BaseService {
+  static get category() {
+    return 'monitoring'
+  }
+
+  static get route() {
+    return {
+      base: 'security-headers',
```
calebcartwright:
```suggestion
      base: '',
```

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,84 @@
...
+      pattern: ':protocol(https|http)/:host',
```
calebcartwright:
```suggestion
      pattern: 'security-headers',
      queryParamSchema,
```

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Security Headers',
+        namedParams: {
```
calebcartwright:
```suggestion
```

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,84 @@
...
+          protocol: 'https',
```
calebcartwright:
```suggestion
```

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,84 @@
...
+          host: 'shields.io',
```
calebcartwright:
```suggestion
```

services/security-headers/security-headers.service.js
```diff
@@ -0,0 +1,84 @@
...
+        },
```
calebcartwright:
```suggestion
        namedParams: {},
        queryParams: { baseUrl: 'https://shields.io' },
```

services/security-headers/security-headers.service.js
calebcartwright: AFAIK GitHub does not support suggested changes on multi-line blocks, so all of these are basically to convert what you currently have for `namedParams` to `namedParams: {},`

services/security-headers/security-headers.service.js
srtab: @calebcartwright Thanks for the suggestions, it works just fine! I already made the suggested changes.

services/security-headers/security-headers.tester.js
srtab: Done, i removed the test.

services/security-headers/security-headers.service.js
srtab: I agree. I added the suggested code.

PyvesB:
@srtab do you have a Twitter handle I can give credit to in a Tweet about this new badge?

srtab:
> @srtab do you have a Twitter handle I can give credit to in a Tweet about this new badge?

Sorry for late response. I do have an account, but i haven't used it. But thanks anyway.

PyvesB:
Tweet posted: https://twitter.com/Shields_io/status/1174037594851688448

PR #3985, opened 2019-09-10
Add Twitch keys to configuration files

PyvesB:
We missed out on these, see https://github.com/badges/shields/pull/3683#issuecomment-530070731.


PyvesB:
Looking much better now:

PR #3987, opened 2019-09-11
Change CI to Cirrus CI

RDIL:
The goal of this PR is to change the CI to be faster, easier to configure, and just plain more simple without losing functionality.


paulmelnikow:
Agree with what Caleb wrote, though just wanted to  add that we really care about making the experience of contributing to Shields a fun one!

If you're interested in working on that, some good places to start would be the discussion threads for the open issues tagged [developer-experience](https://github.com/badges/shields/issues?q=is%3Aopen+is%3Aissue+label%3Adeveloper-experience) or [documentation](https://github.com/badges/shields/issues?q=is%3Aopen+is%3Aissue+label%3Adocumentation)! And if you've new tooling ideas to propose, it's always good to start with a discussion!

Daniel15:
I'm also curious as to what the benefits are of Cirrus over something like Azure Pipelines or GitHub Actions?

PR #3990, opened 2019-09-13
move [Jira] hostname to query param

calebcartwright:
Refs #3714


services/jira/jira-issue.service.js
```diff
@@ -1,9 +1,14 @@
 'use strict'
 
 const Joi = require('@hapi/joi')
+const { optionalUrl } = require('../validators')
 const { authConfig } = require('./jira-common')
 const { BaseJsonService } = require('..')
 
+const queryParamSchema = Joi.object({
+  hostUrl: optionalUrl.required(),
```
paulmelnikow: Should this be `server`? Or `baseUrl`?

calebcartwright: I think `baseUrl` would probably be better. 

IIRC Atlassian uses "Server" in their product branding, including for Bitbucket and Jira, so `server` as the name  _could_ potentially connote to some users that they could only use this with a "server" instance, and not Jira cloud

PR #4020, opened 2019-09-14
move [Discourse] url to query param

calebcartwright:
Refs #3714

### DIFF
--- a/services/discourse/discourse.service.js
+++ b/services/discourse/discourse.service.js
@@ -14,7 +14,7 @@ const schema = Joi.object({
 }).required()
 
 const queryParamSchema = Joi.object({
-  hostUrl: optionalUrl.required(),
+  server: optionalUrl.required(),
 }).required()
 
 class DiscourseBase extends BaseJsonService {
@@ -25,7 +25,7 @@ class DiscourseBase extends BaseJsonService {
   static buildRoute(metric) {
     return {
       base: 'discourse',
-      pattern: `${metric}`,
+      pattern: metric,
       queryParamSchema,
     }
   }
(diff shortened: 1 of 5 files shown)


calebcartwright:
Oops! Didn't realize that we were using the Discourse badge as part of the e2e test suite. Fixing that now

services/discourse/discourse.service.js
```diff
@@ -13,6 +13,10 @@ const schema = Joi.object({
   like_count: nonNegativeInteger,
 }).required()
 
+const queryParamSchema = Joi.object({
+  hostUrl: optionalUrl.required(),
```
paulmelnikow: I don't feel strongly, though another option for naming this parameter might be `server`.

services/discourse/discourse.service.js
```diff
@@ -21,20 +25,19 @@ class DiscourseBase extends BaseJsonService {
   static buildRoute(metric) {
     return {
       base: 'discourse',
-      // Do not base new services on this route pattern.
-      // See https://github.com/badges/shields/issues/3714
-      pattern: `:scheme(http|https)/:host/${metric}`,
+      pattern: `${metric}`,
```
paulmelnikow:
```suggestion
      pattern: metric,
```

services/discourse/discourse.service.js
calebcartwright: I like that better 👍 I struggled a bit naming these, so I'm not attached to `hostUrl` either

services/discourse/discourse.service.js
calebcartwright: Good spot! That's what I get for being lazy with the delete key 😉

PR #4021, opened 2019-09-14
move [MavenMetadata] url to query param

calebcartwright:
Refs #3714


services/maven-metadata/maven-metadata-redirect.service.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+const { redirector } = require('..')
+
+module.exports = [
+  redirector({
+    category: 'version',
+    route: {
+      base: 'maven-metadata/v',
+      pattern: ':protocol(http|https)/:hostAndPath(.*?)',
```
calebcartwright: I believe the lazy pattern for `hostAndPath` is necessary to prevent it from sucking in an extension format (`.json`, `.svg`) if one is specified. Let me know if that's not the case though

paulmelnikow: Hmm, is that a limitation specific to redirects?

services/maven-metadata/maven-metadata.service.js
```diff
@@ -1,9 +1,14 @@
 'use strict'
 
 const Joi = require('@hapi/joi')
+const { optionalUrl } = require('../validators')
 const { renderVersionBadge } = require('../version')
 const { BaseXmlService } = require('..')
 
+const queryParamSchema = Joi.object({
+  metadataUrl: optionalUrl.required(),
```
paulmelnikow: This could probably just be `url`. Thoughts?

services/maven-metadata/maven-metadata.tester.js
```diff
@@ -6,11 +6,12 @@ const { isVPlusDottedVersionAtLeastOne } = require('../test-validators')
 const t = (module.exports = new ServiceTester({
```
paulmelnikow: Any reason this couldn't use `createServiceTester()`?

services/maven-metadata/maven-metadata-redirect.service.js
calebcartwright: > Hmm, is that a limitation specific to redirects?

I think it may be. I traced it to this point and found that the `.json` extension from the tests was getting pulled into the `hostAndPath` match so this was always injecting an extra `.svg`. Gave up after a while though and just changed the pattern 😄 

https://github.com/badges/shields/blob/9fe4d739b0071ec960f4a4704f1b63c82c1fee6f/core/base-service/redirector.js#L113-L116

services/maven-metadata/maven-metadata.service.js
calebcartwright: SGTM

services/maven-metadata/maven-metadata.tester.js
calebcartwright: I'm not sure why it wasn't used before, but I elected to leave this in place so I could explicitly set `pathPrefix` since the `maven-metadata` and `v` portions of the route had to be split across `base` and `pattern`. 

I could certainly switch this to `createServiceTester` though and add the leading `/v` to the tests

paulmelnikow: I'm 👍 on switching, eager to keep paring down these exceptions.

calebcartwright: Done and done!

services/maven-metadata/maven-metadata-redirect.service.js
calebcartwright: I took another look at this, I switched it `:hostAndPath+` and everything is working as expected. When I first created the redirector I copied in the pattern from the service which had `:hostAndPath(.+)` which wasn't working

PR #4022, opened 2019-09-15
switch [TeamCity] to pattern and move url to query param

calebcartwright:
Refs #3714


services/teamcity/teamcity-build.service.js
```diff
@@ -14,53 +15,47 @@ const buildStatusSchema = Joi.object({
     .required(),
 }).required()
 
+const queryParamSchema = Joi.object({
+  hostUrl: optionalUrl,
```
paulmelnikow: `server`?

PR #4027, opened 2019-09-15
move [Twitter] url to query param

calebcartwright:
Refs #3714


calebcartwright:
> Could you update the badge suggestion endpoint too?

Good catch!

PR #4041, opened 2019-09-19
fix type errors in [DynamicXml] service

### LINKED ISSUE #3814: [DynamicXml] TypeError: Cannot read property 'data' of null
Sentry Issue: [SHIELDS-3B](https://sentry.io/organizations/shields/issues/1134174969/?referrer=github_integration)

```
TypeError: Cannot read property 'data' of null
  File "/home/m/shields/services/dynamic/dynamic-xml.service.js", line 46, in xpath.select.map
    .map((node, i) => (pathIsAttr ? node.value : node.firstChild.data))
  ?, in Array.map
  File "/home/m/shields/services/dynamic/dynamic-xml.service.js", line 46, in DynamicXml.handle
    .map((node, i) => (pathIsAttr ? node.value : node.firstChild.data))
  ?, in null.<anonymous>
```

### LINKED ISSUE #4017: [DynamicXml] TypeError: xpath.select(...).map is not a function
Sentry Issue: [SHIELDS-3C](https://sentry.io/organizations/shields/issues/1134194706/?referrer=github_integration)

```
TypeError: xpath.select(...).map is not a function
  File "/home/m/shields/services/dynamic/dynamic-xml.service.js", line 46, in DynamicXml.handle
    .map((node, i) => (pathIsAttr ? node.value : node.firstChild.data))
  ?, in null.<anonymous>
```

calebcartwright:
I believe this fixes #4017 and fixes #3814


services/dynamic/dynamic-xml.service.js
```diff
@@ -47,14 +36,50 @@ module.exports = class DynamicXml extends BaseService {
...
+    values = values.reduce((accum, node) => {
+      if (pathIsAttr) {
+        accum.push(node.value)
+      }
+
+      if (node.firstChild) {
+        accum.push(node.firstChild.data)
+      }
+
+      return accum
```
calebcartwright: One of the runtime errors was trying to access `data` field when `firstChild` was null/undefined. My thinking is that it would be better to just ignore the node if all the conditions are false vs. throwing an exception. I'm not sure if that is the "correct" behavior, but it should get rid of the runtime error either way.

paulmelnikow:
Hi! Sorry this is has been sitting a bit. Were you able to identify badge URLs that produced the TypeErrors?

calebcartwright:
> Were you able to identify badge URLs that produced the TypeErrors?

Nope, and I imagine that'd be fairly tricky to do for the dynamic badges. However, I don't think identifying the original urls that produced the errors is necessary, as it's clear in the code when these could have occurred

services/dynamic/dynamic-xml.service.js
```diff
@@ -47,14 +36,48 @@ module.exports = class DynamicXml extends BaseService {
...
       throw new InvalidResponse({ prettyMessage: 'no result' })
     }
 
-    return renderDynamicBadge({ value: values, prefix, suffix })
+    return { values }
+  }
+
+  async handle(namedParams, { url, query: pathExpression, prefix, suffix }) {
+    // e.g. //book[2]/@id
+    const pathIsAttr = (
```
platan: What do you think about moving this to `transform` function? `pathIsAttr` it only used inside that function.

services/dynamic/dynamic-xml.service.js
```diff
@@ -27,18 +27,7 @@ module.exports = class DynamicXml extends BaseService {
...
-      pathExpression.split('/').slice(-1)[0] || ''
-    ).startsWith('@')
-
-    const { buffer } = await this._request({
-      url,
-      options: { headers: { Accept: 'application/xml, text/xml' } },
-      errorMessages,
-    })
-
+  transform({ pathExpression, buffer, pathIsAttr }) {
```
platan: I really like to see this as a separate function - it's more readable and easier to test.

services/dynamic/dynamic-xml.service.js
```diff
@@ -47,14 +36,48 @@ module.exports = class DynamicXml extends BaseService {
     } catch (e) {
       throw new InvalidParameter({ prettyMessage: e.message })
     }
-    values = values.map((node, i) =>
-      pathIsAttr ? node.value : node.firstChild.data
-    )
+
+    if (!Array.isArray(values)) {
+      throw new InvalidResponse({
+        prettyMessage: 'must be a collection of elements',
```
platan: At this stage we know that an xpath expression and an XML content are valid, but we are not able to return a result. What do you think about a `unsupported query` message?

calebcartwright: I'm not familiar with the `xpath` lib we are using to know what would result in the `select` function returning a non-array, but `unsupported query` sounds better to me anyway 😄

services/dynamic/dynamic-xml.service.js
platan: Query `//book[1]/title` for an empty element ` <title></title>` also results with `Cannot read property 'data' of null`.

platan:
`TypeError: xpath.select(...).map is not a function` is caused by query like `string(//book[1]/title)`.

calebcartwright:
> `TypeError: xpath.select(...).map is not a function` is caused by query like `string(//book[1]/title)`.
> BTW, https://github.com/goto100/xpath/issues?utf8=%E2%9C%93&q=is%3Aissue+array+ is a really good source of problematic queries :-)

Oops! I added a unit test that's stubbing out `xpath.select` before I saw this

calebcartwright:
Thanks for the review @platan!

> But for //book[1]/title/text() we get no result which can suggest, that given element does not exist. It would be better to return unsupported query or an actual result

Sure, that can be tackeld in a follow up for anyone so inclined 😄

PR #4060, opened 2019-09-20
upgrade Joi and related packages

### LINKED ISSUE #4031: Tracking Issue for @hapi/joi v16.x upgrade
Seems like the upgrade from v15.x to v16.x is going to be rather involved. Opening this so we can track this item, especially as I expect the dependabot PRs will continue to be closed/superseeded. 

Release notes for Joi 16: hapijs/joi#2037

https://github.com/badges/shields/pull/4010
https://github.com/badges/shields/pull/4008#issuecomment-531303928

Tasks:

- [x] Update joi-extension-semver for Joi 16 (dszakallas/joi-extension-semver#12)
- [x] Update IcedFrisby for Joi 16 (IcedFrisby/IcedFrisby#304)

chris48s:
Suspect we'll get a few timeouts on a full service test run, but I have had them all passing locally. Also note the full test run in Circle CI skips the Twitch and Symfony test suites because we've got no credentials set up, but you can check them locally with tokens, if you want. I'm going to raise another issue about this.

Beyond that I think I've done an OK job of explaining what's going on at each stage in the commit messages.

Closes #4031


core/base-service/base.js
```diff
@@ -39,14 +39,17 @@ const defaultBadgeDataSchema = Joi.object({
   namedLogo: Joi.string(),
 }).required()
 
-const optionalStringWhenNamedLogoPrsent = Joi.alternatives().when('namedLogo', {
-  is: Joi.string().required(),
-  then: Joi.string(),
-})
+const optionalStringWhenNamedLogoPrsent = Joi.alternatives().conditional(
```
calebcartwright:
```suggestion
const optionalStringWhenNamedLogoPresent = Joi.alternatives().conditional(
```

calebcartwright:
> Suspect we'll get a few timeouts on a full service test run, but I have had them all passing locally

Yup, couple unrelated failures, but basically all tests passing 👍

services/bstats/bstats-players.service.js
```diff
@@ -5,7 +5,7 @@ const { metric } = require('../text-formatters')
 const { BaseJsonService } = require('..')
 
 const schema = Joi.array()
-  .items(Joi.array().items([Joi.number().required(), Joi.number().required()]))
+  .items(Joi.array().items(Joi.number().required(), Joi.number().required()))
```
paulmelnikow: Purely for my own edification, what is this? ^^

services/clojars/clojars-release.tester.js
```diff
@@ -1,12 +1,15 @@
...
 const t = (module.exports = require('../tester').createServiceTester())
 
 t.create('clojars (valid)')
   .get('/prismic.json')
   .expectBadge({
     label: 'clojars',
-    message: /^\[prismic "([0-9][.]?)+"\]$/, // note: https://github.com/badges/shields/pull/431
+    message: Joi.string()
+      .regex(/^\[prismic "([0-9][.]?)+"\]$/)
+      .required(), // note: https://github.com/badges/shields/pull/431
```
paulmelnikow: What would you think about pushing this functionality into `expectBadge`?

services/sonar/sonar-tests.tester.js
```diff
@@ -100,7 +100,12 @@ t.create('Test Failures Count')
   )
   .expectBadge({
     label: 'test failures',
-    message: Joi.alternatives(isMetric, 0),
+    message: Joi.alternatives(
+      isMetric,
+      Joi.number()
+        .valid(0)
+        .required()
+    ),
```
paulmelnikow: This rather log schema is repeated a few times in this file. Would you mind refactoring it?

services/twitch/twitch.tester.js
```diff
@@ -26,9 +27,7 @@ t.create('Status of andyonthewings')
   .get('/status/andyonthewings.json')
   .expectBadge({
     label: 'twitch',
-    message: function(v) {
-      return ['live', 'offline'].indexOf(v) >= 0
-    },
+    message: Joi.equal('live', 'offline').required(),
```
paulmelnikow: 👍

services/static-badge/static-badge.tester.js
```diff
@@ -24,7 +24,11 @@ t.create('All one color')
 
 t.create('Not a valid color')
   .get('/badge/label-message-notacolor.json')
-  .expectBadge({ label: 'label', message: 'message', color: null })
+  .expectBadge({
+    label: 'label',
+    message: 'message',
+    color: undefined,
```
calebcartwright: This one puzzled me a bit. Why's the color value different? 

I'm asking out of curiosity, not suggesting a change or anything

services/sonar/sonar-tests.tester.js
calebcartwright: I had the same thought (although I also thought I should have done that to begin with when I wrote the original 😄)

services/bstats/bstats-players.service.js
chris48s: It is looking for is an array of arrays of exactly 2 numbers e.g:

```
[
  [ 1500346800000, 30679 ],
  [ 1567774800000, 110761 ],
  [ 1510439400000, 40127 ]
]
```

We could write it like

```js
const schema = Joi.array()
  .items(
    Joi.array()
      .items(Joi.number().required())
      .length(2)
      .required()
  )
  .required()
```

if you think that is clearer?

services/clojars/clojars-release.tester.js
chris48s: In principle, we could do. Any ideas on how to check that `expected` is a regex given


```
> typeof /^[0-9]+$/
'object'
```

?

services/static-badge/static-badge.tester.js
chris48s: Its a good question because I thought this was sensible but looking at it again, I've unwittingly changed what is going on here.

In javascript

```
> typeof null
'object'
```

so previously, this test would have hit the `if (typeof expected === 'object')` case. Now, there is no case for if `expected` is a generic `object`, so `.expectBadge({ label: 'label', message: 'message', color: null })` throws an exception. I changed it, but changing it to `undefined` actually changes the meaning of the test because then we're making _no assertion at all_ about the value of `color` rather than explicitly asserting that it isn't set.

On reflection, the correct way to explicitly assert that `color` is undefined is to use [`.forbidden()`](https://github.com/hapijs/joi/blob/master/API.md#anyforbidden), which I've done in 48a5afc

calebcartwright: Ah okay, thanks for the explanation!

services/clojars/clojars-release.tester.js
paulmelnikow: Yea, `/asdf/ instanceof RegExp` is `true`.

chris48s: I've added the code in 66344c0 to allow `expected` to just be a regex and reverted 0b18d17 to restore the 3 occurrences where we were just passing a regex.

There are many more occurrences throughout the codebase where we're passing `message: Joi.string().regex(/foo/)` e.g:

https://github.com/badges/shields/blob/8621fe42d76b809e9ba4ae844920f8ed0373817a/services/jenkins/jenkins-plugin-version.tester.js#L22
https://github.com/badges/shields/blob/51d864efc106dbfc7bae4d4aa2d306517a18e323/services/discourse/discourse.tester.js#L95-L97
https://github.com/badges/shields/blob/76b2e2634a358606426607b39a24b5de4d98e266/services/github/github-downloads.tester.js#L48-L50
https://github.com/badges/shields/blob/6cbc7b587a532d70a0f3f13d1d61a93b2ff7387b/services/lgtm/lgtm-alerts.tester.js#L11
...and so on

or where we are acheiving the same thing using `withRegex()` e.g:

https://github.com/badges/shields/blob/1f29c22d3dd47794da72254d2b8527dc0232fea2/services/keybase/keybase-btc.tester.js#L10
https://github.com/badges/shields/blob/1f29c22d3dd47794da72254d2b8527dc0232fea2/services/spiget/spiget-latest-version.tester.js#L12
https://github.com/badges/shields/blob/83c3c709086f73f4ec1576a0121d188c21ce5982/services/puppetforge/puppetforge-module-endorsement.tester.js#L10
...and so on

We could just replace all of those with `message: /foo/` as there is no reason to wrap them in a Joi schema, but given it is not necessary to perform the upgrade to Joi 16, its probably better to do that in another PR if you prefer that style. There are quite a lot of them

calebcartwright: > its probably better to do that in another PR if you prefer that style

Agreed!

Issue #4068, opened 2019-09-24
GitHub Occurances Badge


cloewen8:
:clipboard: **Description**

A badge for GitHub that counts the occurrences of a sequence in a file.

For example, given `/github/occurances/badges/shields/README.md/badge`, the badge would show *22* (badge occurs in README.md 22 times).

Optimally, sequence should be a regular expression (escape sequences, word boundaries, character classes).

:link: **Data**

The data required for this can be retrieved from `https://raw.githubusercontent.com/:user/:repo/:branch/:path`. It only requires authentication for private repositories.
Unfortunately I don't know of any official documentation for this endpoint, only that it is the destination when pressing "Raw" on a file on GitHub.

Additional processing is required and would need to be limited.

:microphone: **Motivation**

I personally want to use it to count the number of facts in a text file (each fact is on its own line).
A badge for counting lines would work, but being able to count anything opens the door for a lot more opportunity:
- How many times is "as you know" mentioned in a story?
- How many times, if at all, is `goto` used?
- How many code blocks are present?
- How many references to a shutdown API exist?

paulmelnikow:
Hi, thanks for your request. We have something similar which searches for files within a repo that match a specific pattern, using the GitHub Search API, however it’s not able to do this.

I like the idea of a dynamic text badge, and can see doing lines or string matching, however I feel like for a lot of things you’d want a regex (and not sure we should run arbitrary regexes, since they can be crafted in a way that they use a large amount of compute resource).

Can you share a link to the file? Sometimes seeing the specific case really solidifies why a feature should exist. It might also surface a creative way to use what is already there!

cloewen8:
I absolutely agree that arbitrary regex (or any user submitted code) should not be blindly trusted! For computation, a timeout can be used. I know regex101.com uses this strategy.
![image](https://user-images.githubusercontent.com/1569808/65548990-d2f43500-ded9-11e9-91cb-f97bce5c41fa.png)

Here is were I want to use it: https://github.com/cloewen8/dolphin-fact/blob/master/README.md
Currently, I'm using /github/size, but this isn't very helpful, but better than nothing. I want to use it as a form of progress counter, get people interested as the project grows.

paulmelnikow:
Huh, since it's a list, what would you think about using YAML instead? That way you could use the Dynamic YAML badge and a JSONPath expression.

[...]

Issue #3960, opened 2019-09-08
Badge Request: Go (golang) module version from GitHub


JulienBreux:
:clipboard: **Description**

- Which service is this badge for: `GitHub`
- What sort of information should this badge show?
`Go | v1.13`

:link: **Data**

- `go.mod` (Eg: https://github.com/JulienBreux/baleia/blob/master/go.mod#L3)

:microphone: **Motivation**

I love Go, I love Shield.io 😉 

👀 **Example**

![Example](https://img.shields.io/badge/go-v1.13-blue)

calebcartwright:
For anyone interested in implementing this, you may want to take a look at the similar service that provides the version of an npm package from the package.json file in a GitHub repo

https://github.com/badges/shields/blob/master/services/github/github-package-json.service.js

pedrofracassi:
I'll work on it later today.

hjagodzinski:
I see @pedrofracassi you didn't make it. If you don't mind I will look into it.

pedrofracassi:
@hjagodzinski please do.

calebcartwright:
In my admittedly limited experience with Go, I've always found the mod file to be in the root of the repo. Is that something that's enforced/required, or would be it possible for someone to have the mod file in a subdirectory of the repo?

pedrofracassi:
Only if it's a monorepo with multiple go packages, I believe. Maybe let people add the path to the badge URL?

calebcartwright:
Thanks @pedrofracassi - that's exactly what I was thinking

PR #4096, opened 2019-09-27
Add dependency badge for Pipenv applications [GithubPipenv]

paulmelnikow:
I recently published https://github.com/metabolize/rq-dashboard-on-heroku and want to add badges to show the locked version of Python and rq-dashboard, the main dependency it’s wrapping.

This is along the lines of #2259, which was for package.json-based applications, and also included some discussion of a Python application that used `requirements.txt`. It’s useful for showing the pinned version of any dependency in a Python application that uses a lockfile.

In the future, as an alternative to reading Pipfile.lock, I could see expanding this to read Pipfile. However for my purposes I prefer to show the locked dependency, since that’s the version that a user of my package would actually get if they ran it on Heroku.


services/github/github-common-fetch.js
```diff
@@ -24,9 +24,9 @@ const contentSchema = Joi.object({
   encoding: Joi.equal('base64').required(),
 }).required()
 
-async function fetchJsonFromRepo(
+async function fetchRepoContent(
   serviceInstance,
-  { schema, user, repo, branch = 'master', filename }
+  { user, repo, branch = 'master', filename }
```
paulmelnikow: I hadn't realized that `Pipfile` is TOML but `Pipfile.lock` is JSON, so I unnecessarily refactored this. However it'd help with #3960 and https://github.com/badges/shields/issues/4068#issuecomment-534805506 so seems worth keeping.

services/pypi/pypi-python-versions.tester.js
```diff
@@ -3,7 +3,6 @@
 const Joi = require('@hapi/joi')
 const t = (module.exports = require('../tester').createServiceTester())
 
-// These regexes are the same, but declared separately for clarity.
```
paulmelnikow: This probably referred to `isPipeSeparatedDjangoVersions` but these are no longer in the same file.

services/pipenv-helpers.js
```diff
@@ -0,0 +1,49 @@
...
+
+function getDependencyVersion({
+  kind = 'prod',
+  wantedDependency,
+  lockfileData,
+}) {
+  let dependenciesOfKind
+  if (kind === 'dev') {
+    dependenciesOfKind = lockfileData.develop
+  } else if (kind === 'prod') {
```
chris48s: The pipenv lockfile and docs refer to 'default' and 'dev' dependencies. As opposed to 'prod' and 'dev'. Lets use that terminology here too.

services/pipenv-helpers.js
```diff
@@ -0,0 +1,49 @@
...
+    dependenciesOfKind = lockfileData.default
+  } else {
+    throw Error(`Not very kind: ${kind}`)
```
chris48s: This seems like a bit of an odd error. I'd be confused by it if it popped up in Sentry :thinking:

services/pipenv-helpers.js
```diff
@@ -0,0 +1,49 @@
...
+}).required()
+
+const isLockfile = Joi.object({
+  _meta: Joi.object({
+    requires: Joi.object({
+      python_version: Joi.string().required(),
+    }).required(),
+  }).required(),
+  default: Joi.object().pattern(Joi.string().required(), isDependency),
+  develop: Joi.object().pattern(Joi.string().required(), isDependency),
```
chris48s: Requiring _every_ dependency to have the version key will cause us to render `'invalid response data'` on lockfile that use any VCS dependencies. Example: https://github.com/DemocracyClub/aggregator-api/blob/5dc48baa0a45effadc238cf93954200bfdcc855b/Pipfile.lock#L134-L138

services/pipenv-helpers.js
```diff
@@ -0,0 +1,49 @@
...
+const { InvalidParameter } = require('.')
+
+const isDependency = Joi.object({
+  version: Joi.string().required(),
+}).required()
+
+const isLockfile = Joi.object({
+  _meta: Joi.object({
+    requires: Joi.object({
+      python_version: Joi.string().required(),
```
chris48s: Can we only require this key for the python version badge? Although its in by default, `python_version` is not required. If my `Pipfile` is 

```toml
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]

[packages]

```

`pipenv lock` will happily produce

```json
{
    "_meta": {
        "hash": {
            "sha256": "ebffa69a1fa192d1cef7cb42ad79231ca976565c5ce371a70160b3048d3cbc06"
        },
        "pipfile-spec": 6,
        "requires": {},
        "sources": [
            {
                "name": "pypi",
                "url": "https://pypi.org/simple",
                "verify_ssl": true
            }
        ]
    },
    "default": {},
    "develop": {}
}
```

which would also prevent the `dependency-version` badge from working.

Unfortunately removing `python_version` is sometimes needed for projects which need to work on more than one python version because `python_version` can't be multiple versions or a range.

services/pipenv-helpers.js
paulmelnikow: Are you thinking in the badge URLs or just internally?

I don't mind using `default` though the pipenv lockfile uses `develop`. I'd prefer to use `dev` in the URL for more cross-project consistency, though I don't feel that strongly about it.

Another option is to translate from the URL terms to the Pipenv terms in `handle()`… which isn't a bad idea, though if we do end up using this function for another service (e.g. "Pipfile.lock from URL") that code would end up being duplicated so we'd probably need a second helper function.

services/pipenv-helpers.js
paulmelnikow: Heh. Copy-pasted from package-json-helpers. There was some chat about it then; basically it's an assertion failure which is unlikely ever to happen. Can change it to something more specific though.

services/pipenv-helpers.js
paulmelnikow: Rather than try to handle everything I started with a naive implementation and figured we could improve it.

Thanks for the example. Will add a test case that covers this.

services/pipenv-helpers.js
paulmelnikow: Sounds good. I'd also rather present a better error for the python version on a project where it's missing. Any idea of a project that looks like that?

If not I could probably write it in a unit test instead.

services/pipenv-helpers.js
chris48s: I was just thinking about internally using 'default' here instead of 'prod'.

I agree that `:user/:repo/:packageName` should be the user-facing URL. You shouldn't have to specify `:user/:repo/default/:packageName`. Agree, `dev` is fine in the user-facing URL, they are `[dev-packages]` in the Pipfile, the flag is `--dev`, etc https://pipenv.readthedocs.io/en/latest/cli/#cmdoption-pipenv-install-d

services/pipenv-helpers.js
chris48s: It just happened the first repo I thought of to test with contains a VCS dependency, so I hit this issue immediately :)

> Is that reliably a VCS dependency?

What do you mean by that?

services/pipenv-helpers.js
chris48s: Sure. Here's one: https://github.com/fikovnik/ShiftIt/blob/master/Pipfile.lock

services/pipenv-helpers.js
paulmelnikow: > It just happened the first repo I thought of to test with contains a VCS dependency, so I hit this issue immediately :)

Nice 😁 

I'm thinking I'll add a service test that shows the version of a VCS dependency. Would that dependency on that repo be a good one to use?

paulmelnikow: Have used that one for now.

services/github/github-common-fetch.js
calebcartwright: Nice!

services/pipenv-helpers.js
calebcartwright: This would only happen in the event of a programming error, correct?

services/github/github-pipenv.service.js
```diff
@@ -0,0 +1,167 @@
...
+    return [
+      {
+        title: 'GitHub Pipenv locked Python version',
+        pattern: ':user/:repo',
+        namedParams: {
+          user: 'metabolize',
+          repo: 'rq-dashboard-on-heroku',
+        },
+        staticPreview: this.render({ version: '3.7' }),
+        documentation,
```
calebcartwright: Will it be evident to end users (either from the `locked` word in the example and/or route) that this is will only work with a lockfile? Would it be helpful to include some additional text in `documentation` to convey that?

I've never worked with `Pipenv` so not sure whether that additional info would helpful or just duplicative clutter

services/pipenv-helpers.js
chris48s: oh. I see what you mean. Yeah - its fine to point a live test at that repo

services/github/github-pipenv.service.js
chris48s: I reckon it will probably be OK, but we could change from "GitHub Pipenv locked Python version" to "GitHub Pipenv Python version from Pipfile.lock" (etc) just to be super clear.

paulmelnikow: Yea, agreed, I approached this badge more from a developers perspective than usual and was thinking documentation would be helpful. 

In the long run some explanatory documentation could be useful on all our badges.

paulmelnikow: I ended up not adding Pipfile.lock to the title, rather writing some pretty extensive docs of what the badge is for, designed for people who don't necessarily have any context.

h1nk:
It seems like if the default branch name is not `master` than the badge option with no branch specified fails. I'm assuming that's why there is a second option with the branch name, but is there a simple way to make it so the default branch is used on the badge with no branch name specified?

paulmelnikow:
Hi @h1nk, thanks for the question. I just dug into this for a minute and while I think we can fetch repo contents from the default branch using the API, those queries are only used when you self-host with your own GitHub credentials. On the production server we fetch repo contents using raw.githubusercontent.com URLs which as far as I can tell need the branch hard-coded (i.e. you can't ask for the default branch).

We do have a lot of API quota at our disposal so if pulling content from the default branch seems important, we could drop this quota optimization and always use the API. It'd also simplify the code which is a plus.

h1nk:
I think the badges that already have options to specify branch names explicitly should probably be kept for user who wish to use them, but it would definitely be great to see automatic default branch detection for the badges with no branch name specified if it doesn't incur a significant API quota penalty and isn't too problematic to change in the codebase.

I'd love to hear some input from others about whether this would be a good or bad addition. :)

paulmelnikow:
Probably not blocking this, and it's not the only badge affected. Would you mind opening a new issue to discuss it?

I left this open for a few days in case @chris48s wanted to take another pass but I think I'll go ahead and merge.

PR #4101, opened 2019-09-28
add [GitHubPackageRegistryDownloads] badges

### LINKED ISSUE #4071: GitHub Package Registry - Downloads Badge
:clipboard: **Description**
Download count badges for packages hosted on the GitHub Package Registry

- Which service is this badge for e.g: GitHub Package Registry 
- What sort of information should this badge show? Download counts, in whichever intervals the GitHub API makes available

:link: **Data**

- Is there a public API?  Yes
- Does the API requires an API key? Auth should be used like our other GH services, specifically leveraging our [BaseGraphQlService](https://github.com/badges/shields/blob/master/core/base-service/base-graphql.js). The [RegistryPackageStatistics](https://developer.github.com/v4/object/registrypackagestatistics/) will be of use
- Link to the API documentation. https://developer.github.com/v4/

:microphone: **Motivation**
Because people want them! 😄 The motivation/benefits for the download badges for the GH Package registry is the same as the other registries we support

Refs https://github.com/badges/shields/issues/3608

calebcartwright:
Closes #4071 

Access to the GitHub Package Registry requires the tokens used have the `read:packages` scope enabled. I'm not sure what all needs to be done to ensure the tokens in our pool are properly scoped, but that's why the service tests are failing:

```sh
AssertionError: message mismatch: expected 'Your token has not been granted the required scopes to execute this query. The \'downloadsThisMonth\' field requires one of the following scopes: [\'read:packages\'], but your token has only been granted the: [\'\'] scopes. Please modify your token\'s scopes at: https://github.com/settings/tokens.' to equal 'version not found'
```


services/github/github-package-registry-downloads.service.js
```diff
@@ -0,0 +1,295 @@
...
+// https://developer.github.com/v4/object/registrypackageversionstatistics/
+const packageStatistics = Joi.object({
+  statistics: Joi.object({
+    downloadsToday: nonNegativeInteger,
+    downloadsThisWeek: nonNegativeInteger,
+    downloadsThisMonth: nonNegativeInteger,
+    downloadsThisYear: nonNegativeInteger,
+    downloadsTotalCount: nonNegativeInteger,
+  }).required(),
+})
```
calebcartwright: I'm not really sure why the GitHub GraphQL Schema has two entirely separate objects with identical fields for this but oh well 🤷‍♂

services/github/github-package-registry-downloads.service.js
```diff
@@ -0,0 +1,295 @@
...
+        },
+        staticPreview: this.render({
+          interval: 'total',
+          count: 5,
+          version: '1.0.6',
+        }),
+        documentation,
+      },
+      {
+        title: 'GitHub Package Registry Downloads (Latest Version)',
```
calebcartwright: This is obviously the same as the above example in that it allows the user to specify their target version, but figured I'd include this example to explicitly show users they can specify `latest`. I don't feel strongly about it though if anyone feels it should be removed

services/github/github-package-registry-downloads.service.js
```diff
@@ -0,0 +1,295 @@
...
+  static render({ interval, count, version }) {
+    const { messageSuffix } = intervalMap[interval]
+    return {
+      label: this._getLabel(version),
+      message: `${metric(count)}${messageSuffix}`,
+      color: downloadCount(count),
+    }
+  }
+
+  async fetch({ user, repo, packageName, version }) {
```
calebcartwright: This function got a little long, but that's mostly because of the formatting and the fact that we need three different GraphQL queries. 

I suspect there's away that the queries/variables could be collapsed into one that is conditionally constructed based on the params (for example including the `version` filter), but I suppose having them inline like this helps with readability.

services/github/github-package-registry-downloads.service.js
```diff
@@ -0,0 +1,295 @@
...
+
+  transform({ json, interval, version }) {
+    const ghPackage = json.data.repository.registryPackages.nodes[0]
+    if (!ghPackage) {
+      throw new NotFound({ prettyMessage: 'package not found' })
+    }
+
+    let statistics
+    if (version === 'latest') {
+      statistics = ghPackage.latestVersion.statistics
```
calebcartwright: AFAICT, if a package exists then it by definition must have a `latestVersion`, so I'm not worried about any `falsy` values for `latestVersion` here

services/github/github-package-registry-downloads.service.js
```diff
@@ -0,0 +1,295 @@
...
+}).required()
+
+const versionDownloadsSchema = Joi.object({
+  data: Joi.object({
+    repository: Joi.object({
+      registryPackages: Joi.object({
+        nodes: Joi.array()
+          .items(
+            Joi.object({
+              version: packageStatistics.allow(null),
```
calebcartwright: This is `null` when a non-existent value is specified for `version` in the query

services/github/github-package-registry-downloads.tester.js
```diff
@@ -0,0 +1,148 @@
+'use strict'
+
```
calebcartwright: There's 18 tests in here, some of which folks may find duplicative. I don't feel too terribly strongly, but here was my thinking behind including them:

* There's three different GraphQL queries that are issued depending on whether the route includes a `version`, and whether the `version` value is `latest`
* The schema has two different contracts for package statistics, that currently have the same field names: [RegistryPackageStatistics](https://developer.github.com/v4/object/registrypackagestatistics/) and [RegistryPackageVersionStatistics](https://developer.github.com/v4/object/registrypackageversionstatistics/). Obviously those can change/evolve independently so I think it's important that our tests cover each interval variant for each of those contracts
* GitHub Package Registry is still in a beta (seemingly a private one) which AFAICT means that the contracts are still subject to change, and the GitHub V4 GraphQL API that's used by this service is still [undergoing contract changes](https://developer.github.com/v4/breaking_changes/) as well. This is one of those instances where I'd prefer to have a few extra tests so that we're more likely to see breaking API changes more quickly.

services/github/github-package-registry-downloads.tester.js
```diff
@@ -0,0 +1,148 @@
...
+const Joi = require('@hapi/joi')
+const { isMetric, isMetricOverTimePeriod } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+const downloadsToday = Joi.alternatives(
+  isMetricOverTimePeriod,
+  Joi.allow('0/today')
+)
```
calebcartwright: Apparently `isMetricOverTimePeriod` doesn't support `0` values for the count, but the first few packages I tried to use for these tests had values of `0` for certain intervals. 

Given that GitHub Package Registry is still pretty new/beta, I felt including a `0..` alternative for these tests would help make them less brittle for the short term.

paulmelnikow:
> Access to the GitHub Package Registry requires the tokens used have the `read:packages` scope enabled. I'm not sure what all needs to be done to ensure the tokens in our pool are properly scoped, but that's why the service tests are failing

I think to do this, we will need to start collecting new tokens 😞 

I wonder if it's possible to keep the old tokens working for the original grants while including a new grant in the new tokens. We also don't want people to be giving access to their private packages. Do you know if this permissions can be scoped to public packages?

Also, I've been meaning to follow up with @espadrine about transferring the OAuth app into the Shields org, which we should do regardless 😀

calebcartwright:
> Do you know if this permissions can be scoped to public packages?

No, at least not that I've seen. 
https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/#available-scopes

image

Although FWIW, it seems that for private packages there may be [additional scope(s)](https://help.github.com/en/articles/configuring-npm-for-use-with-github-package-registry#authenticating-to-github-package-registry) required for the token to give it access to the private repo, so perhaps it won't be a problem?

calebcartwright:
> I wonder if it's possible to keep the old tokens working for the original grants while including a new grant in the new tokens.

I have seen a couple notifications from GitHub where an existing OAuth app I use was requesting new scopes/access so maybe we'll luck out 🤷‍♂

services/github/github-package-registry-downloads.service.js
```diff
@@ -0,0 +1,295 @@
...
+module.exports = class GithubPackageRegistryDownloads extends GithubAuthV4Service {
+  static get category() {
+    return 'downloads'
+  }
+
+  static get route() {
+    return {
+      base: 'github/packages/downloads',
+      pattern:
+        ':user/:repo/:interval(today|week|month|year|total)/:packageName/:version?',
```
calebcartwright: Based on what I've seen, all packages have to be associated with a specific repository, even though GitHub does now include a `Packages` menu item at the user/org level that catalogs all the pacakges.

chris48s:
> Do you know if this permissions can be scoped to public packages?

> FWIW, it seems that for private packages there may be additional scope(s) required for the token to give it access to the private repo

I've not actually tried publishing a private package to test this, but this [blog post](https://medium.com/better-programming/everything-you-need-to-know-about-the-github-package-registry-b6e91c321bfd) also implies that this is based on the the visibility of the repo the package is associated with:

> You must use a personal access token with the `read:packages` and `write:packages` scopes to publish and delete public packages in the GitHub Package Registry with npm.
> Your personal access token must also have the `repo` scope when the repository is private.

My reading of that is also that if a token has `repo:public_repo` scope and `read:packages` then it will only be able to read packages which are associated with public repos, not that the `read:packages` scope allows access to all of the owner's packages, but we should probably verify that.

> I have seen a couple notifications from GitHub where an existing OAuth app I use was requesting new scopes/access

I think if we did this, we'd still need to maintain two pools for tokens that can read packages and tokens that can't every user who has donated us a token grants the additional access (or the vast majority). Otherwise, the badge will sometimes work and sometimes won't depending on which token we pull out from the pool. Maybe we could try requesting the additional scope before deploying this and if enough users were to enable it, drop all the tokens that don't have it out of the pool? Its a bit of a pain though.

calebcartwright:
I'm going to open a new issue to track the work of figuring out our app/token situation as that's got broader implications than this one particular PR

calebcartwright:
In light of current cleanup efforts I'm going to go ahead and close this. Can circle back whenever the token blockers are resolved

PR #4103, opened 2019-10-01
Make search work with category names

### LINKED ISSUE #4100: Make search work with category names
:clipboard: **Description**

I was reading through #4090 and wanted to see the state of our license badges. Instead of clicking on the category, I searched for the "license" keyword. Only three badges showed up out of eighteen in total.

In my opinion, we should include the category name as a search keyword for our badges. What do you think?

If we go forward with this, we should go through all our examples and make sure that none of them explicitly already use the category name in their keyword list.

PyvesB:
Closes #4100.

This approach does the processing in the backend by appending to the list of existing keywords in the examples we output. I've also got a stashed solution that does the processing in the frontend, however, it was somewhat less flexible and probably slightly slowed down the filtering.

My first Hacktoberfest 2019 contribution! 🎉


core/base-service/examples.js
```diff
@@ -154,7 +155,9 @@ function transformExample(inExample, index, ServiceClass) {
       style: style === 'flat' ? undefined : style,
       namedLogo,
     },
-    keywords,
+    keywords: keywords.concat(
+      categories.find(c => c.id === ServiceClass.category).keywords
+    ),
```
calebcartwright: Is it possible for `keywords` to ever be falsy here? Or do we initialize this as an empty array (if there's no explicitly defined keywords in the service def) beforehand?

services/vaadin-directory/vaadin-directory-rating-count.service.js
```diff
@@ -23,7 +23,7 @@ module.exports = class VaadinDirectoryRatingCount extends BaseVaadinDirectorySer
         pattern: 'rating-count/:packageName',
         namedParams: { packageName: 'vaadinvaadin-grid' },
         staticPreview: this.render({ ratingCount: 6 }),
-        keywords: ['vaadin-directory', 'rating-count', 'rating count'],
```
calebcartwright: I suppose this _technically_ would make this badge no longer search-discoverable via `rating count`, though that's fine IMO because that term seems to be very specific to this service, which is search-discoverable through a multitude of other terms 👍

core/base-service/examples.js
PyvesB: The Joi schema specifies `default([])`, so I _think_ it should be okay.

calebcartwright: Makes sense, everything looks good in the review app too 👍

services/categories.js
```diff
@@ -1,21 +1,25 @@
...
+  { id: 'social', name: 'Social', keywords: ['social'] },
+  { id: 'version', name: 'Version', keywords: ['version'] },
+  {
+    id: 'platform-support',
+    name: 'Platform & Version Support',
+    keywords: ['platform'],
+  },
+  { id: 'monitoring', name: 'Monitoring', keywords: ['monitoring'] },
+  { id: 'activity', name: 'Activity', keywords: ['activity'] },
+  { id: 'other', name: 'Other', keywords: [] },
```
paulmelnikow: 💯

PyvesB:
Thanks for reviewing both. If ever we want to change the implementation in the future and move more of this logic to the frontend, I'm happy to pick that up then. 👍

Issue #4104, opened 2019-10-01
GitHub Contributors Service Test Failing


calebcartwright:
:clock11: **When did the problem start?**
~September 29th 2019

:camera: **Live badge**

https://img.shields.io/github/contributors/badges/shields
![](https://img.shields.io/github/contributors/badges/shields)

:wrench: **Is the live badge working?**
Yes

:link: **CircleCI link**
https://circleci.com/gh/badges/daily-tests/621

:beetle: **Stack trace**

```
GithubContributors [live] 1 contributor [ GET /contributors/paulmelnikow/local-credential-storage.json ] 
  AssertionError: message mismatch: expected 'repo not found' to equal '1'
```

:bulb: **Possible solution**
Looks like Paul may have deleted his repo that this test was using 😄 

Changing the test target (perhaps to use `badges/shields`) should resolve this

seetd:
Hi,

I would like to attempt to create a PR for this. Thanks.

calebcartwright:
> I would like to attempt to create a PR for this. Thanks.

That'd be awesome, thanks so much! Let us know if you have any questions

paulmelnikow:
> Looks like Paul may have deleted his repo that this test was using 😄

Ha! It's true :) I got control of the original and deleted the fork. 

I think the test wants a repo with a single contributor. The new copy is here:

https://github.com/metabolize/local-credential-storage

It still has one contributor, though it's soon to have Dependabot as a second contributor. I'm not totally sure what repo to suggest that would have more staying power.

Actually, come to think of it, this repo could work:

https://github.com/badges/shields-tests

seetd:
I was using https://github.com/badges/shields-pushgateway but I can do https://github.com/badges/shields-tests as well. I think I almost got the PR ready and I will submit it in a moment.

PR #4105, opened 2019-10-02
fix: use repos from badges organization for [GithubContributors]  (#4104)

seetd:
Modified the tests for GithubContributors service to make use of repos under the badges organization.


calebcartwright:
Thanks @seetd! Just a heads up that I've tweaked the PR name so that the service tests will be run as part of the CI process.

Issue #4102, opened 2019-10-01
Bugzilla badge broken


Jack-Works:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

<img src="https://img.shields.io/bugzilla/1536094?style=flat-square" />
https://img.shields.io/bugzilla/1536094?style=flat-square

calebcartwright:
Looks like that API response for that bug has an emptry string for the `resolution` field which is being rejected

https://bugzilla.mozilla.org/rest/bug/1536094

For anyone interested in working on this one, I believe adding an `.allow('')` to the schema for the resolution field would resolve this

https://github.com/badges/shields/blob/362da5262ba0a46d59f37030891b84afcd80b0a5/services/bugzilla/bugzilla.service.js#L16

Jack-Works:
Does it mean bugzilla is returned a invalid response or the scheme is maybe out of date?

calebcartwright:
> Does it mean bugzilla is returned a invalid response or the scheme is maybe out of date?

No issues with the API, I just think we were expecting the API response to always include some value but that's clearly not always the case!

seetd:
I would like to try fixing this as well

calebcartwright:
Closed via #4106 4106

PR #4106, opened 2019-10-02
 fix: allow empty but required resolution property for [Bugzilla] (#4102)

seetd:
This was tested against a test case with the bug id 1536094 as shown in the attached.
Screen Shot 2019-10-02 at 1 48 40 AM

The test was not committed since the bug can be modified in future with the resolution field populated which will break the test. There is a way to query the the Bugzilla API and search by status providing a limit of 1 in order to create a robust test. But that involves changing the behavior of the service that is not much to do just for the test. So I am leaving out the unit test for this change

### DIFF
--- a/services/bugzilla/bugzilla.service.js
+++ b/services/bugzilla/bugzilla.service.js
@@ -13,7 +13,9 @@ const schema = Joi.object({
     .items(
       Joi.object({
         status: Joi.string().required(),
-        resolution: Joi.string().required(),
+        resolution: Joi.string()
+          .allow('')
+          .required(),
       }).required()
     )
     .min(1)
--- a/services/github/github-contributors.tester.js
+++ b/services/github/github-contributors.tester.js
@@ -4,14 +4,14 @@ const t = (module.exports = require('../tester').createServiceTester())
 const { isMetric } = require('../test-validators')
 
 t.create('Contributors')
-  .get('/contributors/cdnjs/cdnjs.json')
+  .get('/contributors/badges/shields.json')
   .expectBadge({
     label: 'contributors',
     message: isMetric,
   })
 
 t.create('1 contributor')
-  .get('/contributors/paulmelnikow/local-credential-storage.json')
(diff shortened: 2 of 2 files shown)


services/bugzilla/bugzilla.service.js
```diff
@@ -13,7 +13,7 @@ const schema = Joi.object({
     .items(
       Joi.object({
         status: Joi.string().required(),
-        resolution: Joi.string().required(),
+        resolution: Joi.string().allow(''),
```
calebcartwright: I think we still want to have the `.required()` to ensure the field is there. 

Would you mind validing that something like the below still solves the bug?
```suggestion
        resolution: Joi.string().allow('').required(),
```

seetd: It worked and I updated the PR to `resolution: Joi.string().required().allow('')` to make it look closer to the status property. It works as `resolution: Joi.string().allow('').required()` as well if you want that

calebcartwright: Awesome, thanks for confirming. IIRC we typically put `.required()` on the end, though as you noted it doesn't make a function difference. 

It's up to you if you'd like to swap those 👍

seetd: I swapped them and I also squash the commits into a single one. Hope that does not mess up anything.

calebcartwright: > I swapped them and I also squash the commits into a single one. Hope that does not mess up anything.

Thanks, and no worries. We always squash on merge so feel free to structure your commits in whichever way works best for you/makes the changes easier to review

seetd: Thanks I am done for this PR now

calebcartwright:
Fix is confirmed in the review app!

calebcartwright:
Also, if you're ever in Discord, drop us a note and we'll add you to the Contributor role

PR #4107, opened 2019-10-02
Optimize [GithubIssues] using GraphQL

paulmelnikow:
While playing around with a badge for Hacktoberfest I noticed the GitHub Issues badge fetches a huge amount of JSON to render just the total count. There doesn’t seem to be a way to limit the response size through the REST API, so I thought I’d switch this to use GraphQL instead.


calebcartwright:
Nice!

services/github/github-issues.service.js
```diff
@@ -238,35 +255,98 @@ module.exports = class GithubIssues extends GithubAuthV3Service {
...
+        transformErrors,
+      })
+      return { issueCount: totalCount }
+    }
   }
 
   async handle({ variant, raw, user, repo, label }) {
-    const json = await this.fetch({ variant, user, repo, label })
+    const isPR = isPRVariant[variant]
+    const isClosed = isClosedVariant[variant]
```
paulmelnikow: I also moved the logic of converting `variant` to `isPR` and `isClosed` to `handle()`. IMO as this concern is related to interpreting the badge URL, it doesn't really belong to fetching or rendering.

calebcartwright:
Random question, but did you see/try https://developer.github.com/v4/explorer/ while working on this? 

I found it was a huge help for me while trying to figure out the queries for the Package Registry

services/github/github-issues.service.js
calebcartwright: Makes sense, this also does the lookup once instead too which is a plus 👍

paulmelnikow:
> Random question, but did you see/try https://developer.github.com/v4/explorer/ while working on this?
> 
> I found it was a huge help for me while trying to figure out the queries for the Package Registry

Yea, it's super useful!

PR #4108, opened 2019-10-02
Added [VisualStudioAppCenter] badges 🎉 

### LINKED ISSUE #2127: Badge request: Visual Studio App Center
It would be awesome to have shields support for Visual Studio App Center build badges. Notably because they don't support custom labels at this time like VSTS builds. I don't believe you would be able to have separate badges for builds/distributions since there doesn't appear to be a distribution badge available currently. However, builds in App Center can be set to distribute too so arguably a user could have both included into one badge if their build is configured for that.

I did some poking around yesterday, and I'm not sure it's entirely possible to fulfill this request since they don't have an API for pulling the information. Maybe someone could reverse engineer their badge URL to make it work, but that's a bit beyond my understanding of the project.

App Center API documentation:
https://openapi.appcenter.ms/#/

Example of their badge URL:
https://build.appcenter.ms/v0.1/apps/d3124a04-aa78-4963-8489-20e54dc01ba8/branches/DEV/badge

I saw some progress on Hockey App, which App Center is the successor of I believe, but I don't think that code would work for App Center.

Prouser123:
Closes #2127.


services/visual-studio-app-center/visual-studio-app-center-base.js
```diff
@@ -0,0 +1,40 @@
...
+    return this._requestJson({
+      schema,
+      options: {
+        headers: {
+          'X-API-Token': token,
+        },
+      },
+      errorMessages: {
+        401: 'invalid token',
+        403: 'project not found',
```
calebcartwright: I'm not familiar with the API here, but are they really returning 403 and 404 for project not found scenarios? I would've guessed 403 would be some kind of forbidden/trying to access a project the token is not authorized to access

services/visual-studio-app-center/visual-studio-app-center-builds.service.js
```diff
@@ -0,0 +1,62 @@
...
+})
+
+module.exports = class VisualStudioAppCenterBuilds extends BaseVisualStudioAppCenterService {
+  static get category() {
+    return 'build'
+  }
+
+  static get route() {
+    return {
+      base: 'vsac/builds',
```
calebcartwright: Although it admittedly makes for a longer url, I think it'd be better to use the full `visual-studio-app-center` here like we do for the Visual Studio Marketplace badges: https://github.com/badges/shields/blob/master/services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js#L25

services/visual-studio-app-center/visual-studio-app-center-builds.service.js
```diff
@@ -0,0 +1,62 @@
...
+      pattern: ':owner/:app/:branch/:token',
```
calebcartwright: Are both `branch` and `token` required values here? I'm not familiar with this service but it seems with other build services that `branch` is optional, and if so we'd want to have the route params reflect that with an optional branch param

services/visual-studio-app-center/visual-studio-app-center-builds.service.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Visual Studio App Center Builds',
+        namedParams: {
+          owner: 'jct',
+          app: 'my-amazing-app',
```
calebcartwright: lol

services/visual-studio-app-center/visual-studio-app-center-base.js
Prouser123: Surprisingly, yes.
401s are always returned for invalid API tokens and 403/404s for invalid projects.

Interestingly, very occasionally the 404 message will display 'user not found' but this is not consistent so I did not include it.

services/visual-studio-app-center/visual-studio-app-center-builds.service.js
Prouser123: Obviously

services/visual-studio-app-center/visual-studio-app-center-base.js
calebcartwright: Fun!

services/visual-studio-app-center/visual-studio-app-center-builds.service.js
Prouser123: Token is always required, but builds is the only endpoint that requires (or can even use) the branch variable.

Releases just use sequential IDs.

calebcartwright: So a branch has to be specified explicitly? The API doesn't use a default branch if none is specified?

Prouser123: No, it requires a branch to be specified.

services/visual-studio-app-center/visual-studio-app-center-builds.service.js
Prouser123: Done in 10bf8ea

services/visual-studio-app-center/visual-studio-app-center-releases-size.service.js
```diff
@@ -0,0 +1,59 @@
...
+  static get defaultBadgeData() {
+    return {
+      label: 'size',
+      color: 'blue',
+    }
+  }
+
+  static render({ size }) {
+    return {
+      message: `${(size / 1048576).toString().substring(0, 4)} MB`,
```
calebcartwright: Could you see if using `pretty-bytes` like some of our other download size badges would be viable here and help simplify this?

For example:
https://github.com/badges/shields/blob/master/services/github/github-repo-size.service.js

services/visual-studio-app-center/visual-studio-app-center-builds.tester.js
```diff
@@ -0,0 +1,59 @@
+'use strict'
+
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('[fixed] Passing Build')
```
calebcartwright: I think you could probably collapse these first four tests into a single test, and just use the `isBuildStatus` validator for the message vs. checking for an explicit message with fixed status builds/mocked responses. 

The core focus for the service tests is to validate integration with the upstream API and to try to catch integration/contract issues quickly, and we can cover that with a single test. The rest are basically just re-testing the logic of the `renderBuildStatusBadge` function which is heavily tested already elsewhere

services/visual-studio-app-center/visual-studio-app-center-releases-osversion.service.js
```diff
@@ -0,0 +1,60 @@
...
+
+module.exports = class VisualStudioAppCenterReleasesOSVersion extends BaseVisualStudioAppCenterService {
+  static get category() {
+    return 'version'
+  }
+
+  static get route() {
+    return {
+      base: 'visual-studio-app-center/releases/osver',
+      pattern: ':owner/:app/:token',
```
calebcartwright: Again you'll have to pardon my lack of familiarity with this service 😄 but is there just a single min_os version for an app, or is there one for iOS and one for Android? If so, how would users make that distinction, if they wanted a badge for each?

services/visual-studio-app-center/visual-studio-app-center-releases-osversion.service.js
```diff
@@ -0,0 +1,60 @@
...
+    return {
+      label: 'min version',
+      color: 'blue',
+    }
+  }
+
+  static render({ app_os, min_os }) {
+    return {
+      message: `${app_os} ${min_os}`,
+    }
```
calebcartwright: What would you think about using the app os in the label here? For example something like:



I'm not sure if that makes sense in this context, but I think that may be a bit more aligned with our other platform/version support badges.

services/visual-studio-app-center/visual-studio-app-center-releases-osversion.tester.js
```diff
@@ -0,0 +1,53 @@
...
+  // This application will never have a new release created.
+  .get(
+    '/jct/test-fixed-android-react/8c9b519a0750095b9fea3d40b2645d8a0c24a2f3.json'
+  )
+  .expectBadge({
+    label: 'min version',
+    message: 'Android 4.1',
+  })
+
+t.create('[fixed] Example Release')
```
calebcartwright: Similar to my comment in the downloads tester, I think we should remove this test

services/visual-studio-app-center/visual-studio-app-center-releases-size.service.js
```diff
@@ -0,0 +1,59 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const {
+  BaseVisualStudioAppCenterService,
+  keywords,
+  documentation,
+} = require('./visual-studio-app-center-base')
+
```
calebcartwright:
```suggestion
const { nonNegativeInteger } = require('../validators')
```

services/visual-studio-app-center/visual-studio-app-center-releases-size.service.js
```diff
@@ -0,0 +1,59 @@
...
+const schema = Joi.object({
+  size: Joi.number().required(),
```
calebcartwright:
```suggestion
  size: nonNegativeInteger,
```

services/visual-studio-app-center/visual-studio-app-center-releases-size.tester.js
```diff
@@ -0,0 +1,38 @@
+'use strict'
+
```
calebcartwright:
```suggestion

const { isFileSize } = require('../test-validators')
```

services/visual-studio-app-center/visual-studio-app-center-releases-size.tester.js
```diff
@@ -0,0 +1,38 @@
...
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('8368844b to 7.98mb')
+  .get('/nock/nock/nock.json')
+  .intercept(nock =>
```
calebcartwright: Can we switch this to a live/real test, and then use the `isFileSize` validator on the badge message instead of an explicit size?

services/visual-studio-app-center/visual-studio-app-center-releases-size.tester.js
```diff
@@ -0,0 +1,38 @@
...
+    nock('https://api.appcenter.ms/v0.1/apps/')
+      .get('/nock/nock/releases/latest')
+      .reply(200, {
+        size: 8368844,
+      })
+  )
+  .expectBadge({
+    label: 'size',
+    message: '7.98 MB',
```
calebcartwright:
```suggestion
    message: isFileSize,
```

services/visual-studio-app-center/visual-studio-app-center-releases-version.service.js
```diff
@@ -0,0 +1,61 @@
...
+
+  async handle({ owner, app, token }) {
+    const { version, short_version } = await this.fetch({
+      owner,
+      app,
+      token,
+      schema,
+    })
+    return renderVersionBadge({
+      version: `${short_version} (${version})`,
```
calebcartwright: Based on the example and some of the service tests, I'm not sure I follow the difference between these two fields, as it seems like `short_version` is actually longer than `version`?

Should we really display both of them together like this, or perhaps provide a param/option to the user to allow them to specify which one they'd prefer? Perhaps a query param like `?versionFormat=short`

Prouser123: `short_version` is the actual application version, whereas `version` is the release number.

Although I'm not personally sure how useful the release number would be on it's own.

services/visual-studio-app-center/visual-studio-app-center-releases-osversion.service.js
Prouser123: There is a single `min_os` version for each app. Each platform / OS is a separate appcenter project.

services/visual-studio-app-center/visual-studio-app-center-releases-size.tester.js
Prouser123: I'd like to keep this mocked test as the App Center displays in `mebibytes` instead of `megabytes` but I will add a live test with the `isFileSize` validator.

services/visual-studio-app-center/visual-studio-app-center-releases-version.service.js
calebcartwright: Ah okay. So just weird/somewhat misleading field names in the API response. 

Is that the common format users see in App Center? If so then agreed this makes the most sense

services/visual-studio-app-center/visual-studio-app-center-releases-osversion.service.js
calebcartwright: > Each platform / OS is a separate appcenter project.

I assume that `project` and `app` are synonymous? For example if I had my app for iOS and Android, I'd have badges like

visual-studio-app-center/releases/osver/me/my-android-app/mytoken
visual-studio-app-center/releases/osver/me/my-ios-app/mytoken

services/visual-studio-app-center/visual-studio-app-center-releases-size.service.js
```diff
@@ -0,0 +1,62 @@
...
+const prettyBytes = require('pretty-bytes')
+const { nonNegativeInteger } = require('../validators')
+const {
+  BaseVisualStudioAppCenterService,
+  keywords,
+  documentation,
+} = require('./visual-studio-app-center-base')
+
+const schema = Joi.object({
+  size: nonNegativeInteger.required(),
```
calebcartwright: No need to include the `.required()` here as that's already attached:
https://github.com/badges/shields/blob/11836f55de1541a7c55d077c6f359c20b82e7570/services/validators.js#L15

```suggestion
  size: nonNegativeInteger,
```

services/visual-studio-app-center/visual-studio-app-center-releases-size.service.js
```diff
@@ -0,0 +1,62 @@
...
+  static get defaultBadgeData() {
+    return {
+      label: 'size',
+      color: 'blue',
+    }
+  }
+
+  static render({ size }) {
+    return {
+      // AppCenter displays in mebibytes.
```
calebcartwright: Are you sure that the value in the API response is not just in bytes? Based on the [swagger spec](https://api.appcenter.ms/preview/swagger.json), `size` seems to be bytes:

```json
...
        "size": {
          "description": "The release's size in bytes.",
          "type": "number"
        },
...
```

The AppCenter UI may display the value in the mebibyte format, but that doesn't necessarily mean the API response uses that format. If you have any other info indicating that the API is mebibytes, then let's include URLs to that documentation in an inline comment

services/visual-studio-app-center/visual-studio-app-center-releases-osversion.service.js
Prouser123: Yep, sorry for not making it clearer before.

services/visual-studio-app-center/visual-studio-app-center-releases-version.service.js
Prouser123: Yeah, that is the format displayed in App Center right now.

calebcartwright: Thanks!

services/visual-studio-app-center/visual-studio-app-center-releases-osversion.service.js
calebcartwright: No worries, thanks for explaining

services/visual-studio-app-center/visual-studio-app-center-releases-size.service.js
Prouser123: Sorry about that. I know that all UI elements display it in mebibytes, but I don't have any sources for the API also being in mebibytes. I'm currently just converting the response to match what the UI displays.

calebcartwright: Gotcha, I thought that was the case. I don't think we need to be doing that conversion. We're using `pretty-bytes` for all our size related badges, which uses base 10. AFAICT the API response is in bytes which we can (and should) pass directly into `pretty-bytes` for formatting so that this is consistent with Shields formatting for sizes

services/visual-studio-app-center/visual-studio-app-center-releases-osversion.service.js
calebcartwright: I'm not entirely sure how to interpret the confused emoji in the context of the suggestion 😄 

Given the `min_os` naming convention, this badge actually sounds like a better fit for the `platform-support` category instead of `version`, but with both `platform-support` and `version` badges, the badge label is representative of the platform/service. 

What we currently have here:  doesn't really fit into that mold, so the label/message need to be tweaked a bit IMO to be consistent with Shields conventions. 

The ones I listed in my last comment were just some suggested alternatives.

https://shields.io/category/platform-support
image

https://shields.io/category/version
image

paulmelnikow: A couple other options:

- **android | >=4.1**
- **android | 4.1+**

Prouser123: I'd like to go with  if you guys don't mind. I personally think that it would be the least confusing.

calebcartwright: SGTM!

m-ruhl:
What's the status? I would like these badges 😍

calebcartwright:
@m-ruhl - there's still some changes that need to be made before this can be merged

calebcartwright:
hey @Prouser123 just wanted to bump this to see where things are? 

IIRC the main two outstanding changes are the update to the os version badge format, and the removal of the unneeded division in the release size badge.

PyvesB:
I'm keen on getting this landed, so I'll pick the PR up and address the last outstanding comments if that's okay with you @Prouser123 ?

PyvesB:
Unfortunately, I don't think `Allow edits from maintainers` was enabled. Either I can submit a brand new PR, or we have to wait.

calebcartwright:
The e2e job failure is unrelated and should be fixed via #4614

PyvesB:
I've addressed the format of the OS version badge, the redundant `required()` call and the bytes format for the size badge.

The _VisualStudioAppCenterBuilds [live] Valid Build_ test is however failing, because `no builds found` is not a valid build status. @Prouser123 is this expected, did you set up that test app yourself?

Prouser123:
Hi everyone,

**First of all, I'm really sorry about all this mess.** I really don't feel good about this. I'm really sorry to keep you waiting. - As it turns out being overrun with the mess that is a school life doesn't leave much time, not to mention all the other problems I've had.

Anyways, enough with me. With the `VisualStudioAppCenterBuilds [live] Valid Build` test, (@PyvesB) it **was** a valid build but it looks like Microsoft have removed it, probably due to it's age. I've since built a new one, but this is something that we should probably test via nock instead until we get a suitable replacement, and yes, all the valid live tests use a test account that I setup whilst making this.

Also just a note about `allow edits from maintainers`, I never received this option. - maybe because I'm using a org for forks?

PyvesB:
> **First of all, I'm really sorry about all this mess.** I really don't feel good about this. I'm really sorry to keep you waiting. - As it turns out being overrun with the mess that is a school life doesn't leave much time, not to mention all the other problems I've had.

Nah, don't worry about it, there's no mess at all! With little effort on my side we're pretty much there in terms of landing this pull request. And that's thanks to your amazing work! 😉 

> Anyways, enough with me. With the `VisualStudioAppCenterBuilds [live] Valid Build` test, (@PyvesB) it **was** a valid build but it looks like Microsoft have removed it, probably due to it's age. I've since built a new one, but this is something that we should probably test via nock instead until we get a suitable replacement, and yes, all the valid live tests use a test account that I setup whilst making this.

Using nock for builds sounds like a reasonable solution, I've gone with that.

> Also just a note about `allow edits from maintainers`, I never received this option. - maybe because I'm using a org for forks?

Yes, I checked with some other PRs of mine and I believe that option isn't available when using an organisation.

@calebcartwright I thin your outstanding comments are addressed, would you please give it a final look?

calebcartwright:
Heroku deployment is failing though seems to be spurious 🤷‍♂ 

https://dashboard.heroku.com/pipelines/d8fd42e8-8e42-466e-98fc-f5f50edbdd8f/review-app/cf8fcb0d-6e8e-4914-876b-46dbf24ee7e8
```bash
 !
 !   Build timed out while waiting to start.
 !
```

PyvesB:
> I've forgotten the specifics around the API key/token that we're requiring in the route param, but may be worth a follow up PR to include inline instructions on the examples if that's not an obvious thing for users of the AppCenter service

Makes sense, let's keep an eye on any user feedback we get about these badges!

calebcartwright:
Thank you both @PyvesB and @Prouser123! Glad to see this one landed

PR #4112, opened 2019-10-02
Add [SourceForgeOpenTickets] badge

### LINKED ISSUE #1573: Badge request: SourceForge open ticket count
It would be nice if there were a badge for the number sourceforge tickets similarly to the number of github issues badge.

There is a stats page that should be easy to parse:

https://sourceforge.net/p/texstudio/bugs/stats/
https://sourceforge.net/p/texstudio/feature-requests/stats/

There is also an API, but it only seems to give the number of total tickets, while the number of open issues is far more interesting

https://sourceforge.net/rest/p/texstudio/bugs?limit=1
https://sourceforge.net/rest/p/texstudio/feature-requests?limit=1

Prouser123:
Closes #1573.

Unfortunately, as [previously mentioned](https://github.com/badges/shields/issues/1573#issuecomment-412820033), we are dangerously close to the 5s limit, and there are no faster API endpoints to use. 😢 

The only other available endpoint returns the total amount of tickets, which isn't useful for us.

Test excerpt below:
```md
    SourceforgeTickets
    [live] bugs
      √ 
        [ GET /bugs/sevenzip.json ] (984ms)
    [mocked] bugs (1500)
      √ 
        [ GET /bugs/nock.json ] (51ms)
    [live] feature requests
      √ 
        [ GET /feature-requests/sevenzip.json ] (4611ms)
    [mocked] feature requests (500)
      √ 
        [ GET /feature-requests/nock.json ]
    [live] invalid project
      √ 
        [ GET /bugs/invalid.json ] (870ms)


  5 passing (8s)
```

This badge currently has two params for either `bugs` or `feature-requests`.


services/sourceforge/sourceforge-tickets.service.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  static render({ count }) {
+    return {
+      message: metric(count),
+    }
+  }
+
+  async fetch({ type, project }) {
+    const url = `https://sourceforge.net/rest/p/${project}/${type}/search?limit=1&q=status:open`
```
calebcartwright: Think that colon may need to be encoded. Could either do it manually 
```suggestion
    const url = `https://sourceforge.net/rest/p/${project}/${type}/search?limit=1&q=status%3Aopen`
```

Or
```suggestion
    const url = `https://sourceforge.net/rest/p/${project}/${type}/search?limit=1&q=${encodeURIComponent('status:open')}`
```

services/sourceforge/sourceforge-tickets.service.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { metric } = require('../text-formatters')
+const { BaseJsonService } = require('..')
```
calebcartwright:
```suggestion
const { nonNegativeInteger } = require('../validators')
const { BaseJsonService } = require('..')
```

services/sourceforge/sourceforge-tickets.service.js
```diff
@@ -0,0 +1,65 @@
...
+
+const schema = Joi.object({
+  count: Joi.number().required(),
```
calebcartwright:
```suggestion
  count: nonNegativeInteger,
```

services/sourceforge/sourceforge-tickets.service.js
```diff
@@ -0,0 +1,65 @@
...
+
+module.exports = class SourceforgeTickets extends BaseJsonService {
+  static get category() {
+    return 'other'
+  }
+
+  static get route() {
+    return {
+      base: 'sourceforge/tickets',
+      pattern: ':type(bugs|feature-requests)/:project',
```
calebcartwright: I feel like `project` should probably come before the type
```suggestion
      pattern: ':project/:type(bugs|feature-requests)',
```

services/sourceforge/sourceforge-tickets.tester.js
```diff
@@ -0,0 +1,49 @@
...
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('bugs')
+  .get('/bugs/sevenzip.json')
+  .expectBadge({
+    label: 'open tickets',
+    message: isMetric,
+  })
+
+t.create('bugs (1500)')
```
calebcartwright: Let's remove this test

services/sourceforge/sourceforge-tickets.tester.js
```diff
@@ -0,0 +1,49 @@
...
+  })
+
+t.create('feature requests')
+  .get('/feature-requests/sevenzip.json')
+  .expectBadge({
+    label: 'open tickets',
+    message: isMetric,
+  })
+
+t.create('feature requests (500)')
```
calebcartwright: Let's remove this test too

services/sourceforge/sourceforge-tickets.service.js
Prouser123: From sourceforge's rules:

```
It must begin with a letter.
It must be between 3 and 15 characters in length.
It can only contain only lowercase letters (a-z), numbers (0-9), and dashes ("-") or underscores ("_")
```

I think we can get away without encoding it.

services/sourceforge/sourceforge-tickets.service.js
Prouser123: I did it this way as that was how it was done in the base sourceforge badge:
`:interval(dt|dm|dw|dd)/:project/:folder`

But if you'd rather it be before the type then that's fine :)

Prouser123:
Looks like it timed out or something?

Prouser123:
Still timing out :(

calebcartwright:
Try adding a `.timeout(10000)` to the tests

Prouser123:
There we go!

services/sourceforge/sourceforge-tickets.service.js
calebcartwright: So the difference there is that the interval represents the noun portion of that route. https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-urls

The badge for that service is for downloads, whereas the badge here is for open ticket counts, so the `interval` is the noun there, and the noun here is `tickets`.  Both `project` and `folder` in the SourceForge downloads route and `project` and `type` here in this service are params/qualifiers. I personally think having `project` as the first "param" (in the context of the full badge url) makes more sense

services/sourceforge/sourceforge-tickets.service.js
```diff
@@ -0,0 +1,66 @@
...
+}).required()
+
+module.exports = class SourceforgeTickets extends BaseJsonService {
+  static get category() {
+    return 'other'
+  }
+
+  static get route() {
+    return {
+      base: 'sourceforge/tickets',
```
calebcartwright: Also, should this perhaps use `open-tickets` for the noun? I know the example listing include "open", but I think it's probably worth making that explicit in the badge route:

```suggestion
      base: 'sourceforge/open-tickets',
```

Prouser123: If you don't mind, I'd prefer to integrate this in a future PR.

calebcartwright: Sure that sounds good. In that case though, I do still think that the noun here should be `open-tickets` or `tickets-open` to make that more explicit. If there's any follow up work to support open/closed/total then we can add redirectors if needed

PyvesB: @Prouser123 could you please make this small change? I believe it's the only blocker for merging this. 😄

Prouser123: Really sorry about this. I made this change locally when offline and never pushed it to the remote. Oops!

Prouser123:
Interesting build error on 551e98c4ed6ddc990f96549ca760032b0800202a

PyvesB:
> 
> 
> Interesting build error on [551e98c](https://github.com/badges/shields/commit/551e98c4ed6ddc990f96549ca760032b0800202a)



> 
> 
> Interesting build error on [551e98c](https://github.com/badges/shields/commit/551e98c4ed6ddc990f96549ca760032b0800202a)

I think you need to change the filename to be consistent with the new route base prefix.

PR #4141, opened 2019-10-05
update contributing docs on url/query params

calebcartwright:
We pivoted to always using query parameters to accept server/instance/url type params, so I've updated this section in our Contributing guidelines to reflect that


CONTRIBUTING.md
```diff
@@ -105,12 +105,11 @@ There are three places to get help:
 ## Badge URLs
 
 - The format of new badges should be of the form
-  `/SERVICE/NOUN/PARAMETERS/QUALIFIERS.format`. For instance,
-  `/gitter/room/nwjs/nw.js.svg`. The vendor is gitter, the
-  badge is for rooms, the parameter is nwjs/nw.js, and the format is svg.
-- For services which require a hostname, the badge should be of the form
-  `/SERVICE/SCHEME/HOST/NOUN/PARAMETERS/QUALIFIERS.format`. For instance,
-  `/discourse/https/discourse.example.com/topics.svg`.
+  `/SERVICE/NOUN/PARAMETERS/QUALIFIERS`. For instance,
```
calebcartwright: Figured we should probably remove `.svg` from this too, especially to avoid confusing new contributors who won't necessarily be aware of the format aspects now that we default to svg.

Issue #3833, opened 2019-08-08
Request: W3C Markup Validation Service


jelle619:
:clipboard: **Description**

- Which service is this badge for e.g: GitHub, Travis CI

The W3C Markup Validation Service

- What sort of information should this badge show?

Whether a page is validated (the validator does not return any errors or warnings), has warnings and/or errors.

Examples:
![Validated](https://img.shields.io/badge/W3C-validated-brightgreen)
![Warning](https://img.shields.io/badge/W3C-3%20warnings-yellow)
![Error](https://img.shields.io/badge/W3C-2%20errors-red)

:link: **Data**

- Is there a public API?
Please read down below.
- Does the API requires an API key?
It's free to use for everyone.
- Link to the API documentation.
Some documentation can be found [here](https://validator.w3.org/docs/). Please read below for more information.

The W3C Validator can be found [here](https://validator.w3.org). A link to its documentation is mentioned above.
One particular page that might be of interest is [this one](https://validator.w3.org/nu/about.html). Under "extras", a bookmark outputting the results in JSON can be found. Its JavaScript code is as follows:

```javascript
(function(){function%20c(a,b){var%20c=document.createElement("textarea");c.name=a;c.value=b;d.appendChild(c)}var%20e=function(a){for(var%20b="",a=a.firstChild;a;){switch(a.nodeType){case%20Node.ELEMENT_NODE:b+=a.outerHTML;break;case%20Node.TEXT_NODE:b+=a.nodeValue;break;case%20Node.CDATA_SECTION_NODE:b+="<![CDATA["+a.nodeValue+"]]\>";break;case%20Node.COMMENT_NODE:b+="<\!--"+a.nodeValue+"--\>";break;case%20Node.DOCUMENT_TYPE_NODE:b+="<!DOCTYPE%20"+a.name+">\n"}a=a.nextSibling}return%20b}(document),d=document.createElement("form");d.method="POST";d.action="https://validator.w3.org/nu/";d.enctype="multipart/form-data";d.target="_blank";d.acceptCharset="utf-8";c("showsource","yes");c("out","json");c("content",e);document.body.appendChild(d);d.submit()})();
```

This could potentially serve as the "API" that is needed to make this work with Shields.io.

:microphone: **Motivation**

[...]

PR #4148, opened 2019-10-06
Add [W3C] Markup Validation Service Badge (#3833)

seetd:
This is my first badge and the logic is getting a little complex due to how the API goods. So I thought it will be a good idea to get a review to see if I am going in the wrong direction. The code works in a pretty close manner to https://validator.nu/

This is a sample of how the badges will look.
Screen Shot 2019-10-06 at 7 10 24 PM

This is how the example dialog will look like
Screen Shot 2019-10-06 at 7 13 23 PM


calebcartwright:
Thanks for this! The build was failing due to a weird (and still unexplained) issue with one of our tests in Circle's Node 12 environment - See https://github.com/badges/shields/pull/4146 so I've merged the latest from master to get those passing again. 

I'll take a look through the changes here tomorrow

calebcartwright:
👀 My initial reaction is that there's a lot more going on here than I'd anticipated, so I have a couple clarification questions for you:

1. For the purposes of Shields being able to provide W3C Validation Badges, we need to pass a minimum of two parameters in our API call to the W3C Checker Web Service: 
    - `doc`: this is the page/url that will be checked
    - `out`: this the response format that we'd like from the Web Service (and we're using `json` here of course)
2. There are some [additional, optional, parameters](https://github.com/validator/validator/wiki/Service-%C2%BB-Common-params) that can be passed to the W3C Checker Web Service, and it looks like you've decided to try to support two of those:
    - [`parser`](https://github.com/validator/validator/wiki/Service-%C2%BB-Common-params#parser): which basically has 4 allowed values: `html`, `xml`, `xmldtd`, and `none`, and only one value can be specified
    - [`schema`](https://github.com/validator/validator/wiki/Service-%C2%BB-Common-params#schema): 0 or more allowed values can be specified.

Do I have that correct?

seetd:
Hey @calebcartwright, 

That is right, as I was working on it. That was my thought about it. But for schema, the validator is actually using a preset. Then that actually selects a set of schemas to validate against.

Presets and Schema
Screen Shot 2019-10-07 at 3 27 14 PM
Screen Shot 2019-10-07 at 3 40 41 PM

Parser
Screen Shot 2019-10-07 at 3 27 22 PM

Thanks

calebcartwright:
Yeah that schema support is a nightmare of complexity 😄

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,269 @@
...
+const Joi = require('@hapi/joi')
+const { optionalUrl } = require('../validators')
+const { BaseJsonService } = require('..')
+
+const schema = Joi.object({
+  url: optionalUrl.required(),
+  messages: Joi.array()
+    .required()
+    .items(Joi.object()),
+}).required()
```
calebcartwright: This is the schema that is used to validate the response from the the upstream API (The W3C Checker Web Service in this case) so it should model the response objects that we care about. 

As such, I don't think we need `url` here right? The `messages` are the only things we care about. We should also be more explicit about those message objects, including the properties we expect to see on the objects in the `messages` array

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,269 @@
...
+
+  // Return the value unchanged
+  return value
+}
+
+const queryParamSchema = Joi.object({
+  doc: optionalUrl.required(),
+  preset: Joi.string()
+    .allow('')
+    .custom(queryParamDecode(presetRegex), 'preset validation'),
```
calebcartwright: Let's move `preset` to a route parameter, since only one option can be specified and there's only ~4 options we need to support. I've added an inline suggestion on the route definition on this note

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,269 @@
...
+
+module.exports = class W3cValidation extends BaseJsonService {
+  static get category() {
+    return 'analysis'
+  }
+
+  static get route() {
+    return {
+      base: 'w3c-validation',
+      pattern: 'doc',
```
calebcartwright:
```suggestion
      pattern: ':prefix(default|html|xml|xmldtd)',
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,269 @@
...
+          color: 'brightgreen',
+        }),
+        documentation,
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return {
+      namedLogo: 'w3c',
```
calebcartwright: We only use logos by default in social badges so we shouldn't have `namedLogo` here, but we should have a default `label`
```suggestion
      label: 'w3c',
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,269 @@
...
+        qs,
+      },
+    }
+  }
+
+  async fetch(jsonUrl) {
+    return this._requestJson(jsonUrl)
+  }
+
+  async handle(_routeParams, { doc, preset, parser }) {
```
calebcartwright:
```suggestion
  async handle({ preset }, { doc, parser }) {
```

Let's also refactor a lot of the content in this function out based on responsibilities. For example, a really common pattern we use is to have a `transform` function that takes the raw API response (which would be `data` here) and inspecting/analyzing that response to return the results.

There's also no need to have `getUrl` as a separate function that's called from `handle`, let's move all of that into the `fetch` function. 

That would then result in `handle` looking something like this: 
```suggestion
  async handle({ preset }, { url, parser }) {
    const data = await this.fetch({ url, preset, parser })
    const messages = this.transform(data)
    ...
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,269 @@
...
+  if (regExpression && !decodeURI(value).match(regExpression)) {
+    return helpers.error('regex')
+  }
+
+  // Return the value unchanged
+  return value
+}
+
+const queryParamSchema = Joi.object({
+  doc: optionalUrl.required(),
```
calebcartwright: I'd also like to rename this. We have the benefit of not having to use the exact same params that the upstream API uses, and I'd prefer to not use `doc` here (it's a little confusing IMO). I'd suggestion something like

```suggestion
  url: optionalUrl.required(),
```

Or

```suggestion
  targetUrl: optionalUrl.required(),
```
Or

```suggestion
  targetPageUrl: optionalUrl.required(),
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,269 @@
...
+      color: this.getColor(messageTypes),
+    })
+    badge.link = [
+      url
+        .format({
+          host: jsonUrl.url,
+          query: jsonUrl.options.qs,
+        })
+        .toString(),
+    ]
```
calebcartwright: This is only for `social` type badges, so let's remove this
```suggestion
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,269 @@
...
+        accumulator[type] = 0
+      }
+      accumulator[type] += 1
+      return accumulator
+    }
+    const messageTypes = data.messages.reduce(reducer, {})
+    const badge = this.constructor.render({
+      message: this.getMessage(messageTypes),
+      color: this.getColor(messageTypes),
+    })
```
calebcartwright: Determining the badge `message` and `color` are rendering responsibilities, so this should be shifted into the `render` function.

seetd:
Thanks for the extensive review, I will look through them.

seetd:
With regards to the schema support, I was thinking of writing unit tests for it. Is there an example or documentation I can refer it. I found the service testing documentation but I thought unit test to the schema generation function will be cleaner

calebcartwright:
> With regards to the schema support, I was thinking of writing unit tests for it. Is there an example or documentation I can refer it. I found the service testing documentation but I thought unit test to the schema generation function will be cleaner

Great question! We don't really have any documentation (at least not that I can think of right now) specifically on unit tests for service classes/functions. IIRC, I volunteered to do this several months ago but haven't gotten around to it 😄 

I forgot to mention this in my earlier review, but I think another good refactoring would be to extract some content (perhaps the documentation and the W3C param schema functions/variables) into a separate file, perhaps something like `w3c-helpers.js`, and then import that in your service class. 

That should simplify the unit testing a bit. We use `mocha` and `chai` for unit testing throughout the project, with `sazerac` as well for the data driven unit tests. Here's one example of unit tests for a service class that I remember off hand https://github.com/badges/shields/blob/9bf7220260d85ec9c74c992693d68da71eaff010/services/liberapay/liberapay-goal.spec.js

but you can also poke around at the various `*.spec.js` files within the `services` directory

services/w3c/w3c-validation.service.js
seetd: Remove url from schema

services/w3c/w3c-validation.service.js
seetd: moved the parser (default,html,xml,xmldtd) instead of the preset

services/w3c/w3c-validation.service.js
seetd: Done

services/w3c/w3c-validation.service.js
seetd: This is the new pattern `pattern: ':parser(default|html|xml|xmldtd)',`

services/w3c/w3c-validation.service.js
seetd: New signature `async handle({ parser }, { targetUrl, preset }) {`

services/w3c/w3c-validation.service.js
seetd: Used `targetUrl `

services/w3c/w3c-validation.service.js
seetd: Removed

services/w3c/w3c-validation.service.js
seetd: `getMessage` and `getColor` have been moved to a helper file and the caller is now the `render` method

seetd:
This is the new popup dialog after the changes
Screen Shot 2019-10-08 at 9 59 22 AM

I added unit test, I hope that is not too much code. But the unit tests help me identified some bugs that I have fix with the latest commit

calebcartwright:
> I added unit test, I hope that is not too much code

Unit tests are always welcome!

services/w3c/w3c-validation-helper.spec.js
```diff
@@ -0,0 +1,381 @@
...
+      const { preset, expectedResult } = test
+      it(`should return ${expectedResult} for ${preset}`, function() {
+        // Act
+        const actualResult = presetRegex.test(preset)
+
+        // Assert
+        expect(actualResult).to.equal(expectedResult)
+      })
+    })
+  })
```
calebcartwright: Cheers for following the Arrange/Act/Assert pattern, though we should remove the explicit comments/spacing calling it out in these tests.

seetd: Sure I will do that now

services/w3c/w3c-validation-helper.js
```diff
@@ -0,0 +1,157 @@
+'use strict'
+
```
calebcartwright: Nice. Extracting these items in here really helps the main service file read more succinctly 👍

services/w3c/w3c-validation.service.js
calebcartwright: Excellent. What about the schema for the message objects inside `messages`? 

I'm guessing it would result in the schema looking something like:
```js
const schema = Joi.object({
  messages: Joi.array()
    .required()
    .items(Joi.object({
      message: Joi.allow('error', 'warning', '...').required(),
    })),
}).required()
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,118 @@
...
+
+  // Return the value unchanged
+  return preset
+}
+
+const queryParamSchema = Joi.object({
+  targetUrl: optionalUrl.required(),
+  preset: Joi.string()
+    .allow('')
+    .custom(presetDecode(presetRegex), 'preset validation'),
```
calebcartwright: Is it necessary to use a custom validator here? 
Would something like the below work? Or is the custom validator required due to the query params being uri encoded?
```js
preset: Joi.string().regex(presetRegex).allow(null),

File: services/w3c/w3c-validation.service.js (hunk 8 above)
Comment (calebcartwright) [reply services/w3c/w3c-validation.service.js]: Perfect! Next step is to extract the json response (`data`) inspection into a separate `transform` function

File: services/w3c/w3c-validation.service.js (hunk 1 above)
Comment (seetd) [reply services/w3c/w3c-validation.service.js]: Added validation for the `type` property

File: services/w3c/w3c-validation.service.js (hunk 8 above)
Comment (seetd) [reply services/w3c/w3c-validation.service.js]: Move MessageTypes generation into a `transform` method

File: services/w3c/w3c-validation.service.js (hunk 11 above)
Comment (seetd) [reply services/w3c/w3c-validation.service.js]: Using the regex validator but doing `allow('')` instead of `allow(null)`

File: services/w3c/w3c-validation.service.js [hunk 12]
```diff
@@ -0,0 +1,115 @@
...
+  getSchema,
+} = require('./w3c-validation-helper')
+const { BaseJsonService } = require('..')
+
+const schema = Joi.object({
+  messages: Joi.array()
+    .required()
+    .items(
+      Joi.object({
+        type: Joi.any()
```
calebcartwright:
```suggestion
        type: Joi.string()
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,115 @@
...
+
+  async fetch(targetUrl, preset, parser) {
+    const qs = { doc: targetUrl, out: 'json' }
+    if (preset) qs['schema'] = getSchema(preset)
+    if (parser && parser.toLowerCase() != 'default') qs['parser'] = parser
+    return this._requestJson({
+      url: 'http://validator.nu',
+      schema,
+      options: {
+        qs,
```
calebcartwright:
```suggestion
        qs: {
          schema: getSchema(preset),
          parser: parser === 'default' ? undefined : parser,
          doc: targetUrl,
          out: 'json,
        },
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,115 @@
...
+  static render({ messageTypes }) {
+    return {
+      label: 'w3c',
+      message: getMessage(messageTypes),
+      color: getColor(messageTypes),
+    }
+  }
+
+  async fetch(targetUrl, preset, parser) {
+    const qs = { doc: targetUrl, out: 'json' }
```
calebcartwright:
```suggestion
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,115 @@
...
+    if (preset) qs['schema'] = getSchema(preset)
```
calebcartwright: Just make sure that `getSchema` returns undefined if `preset` is falsy. You can set a param to undefined in `qs` and it'll be removed/handled accordingly on the backend automatically, so we don't have to worry about checking that here.
```suggestion
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,115 @@
...
+    if (parser && parser.toLowerCase() != 'default') qs['parser'] = parser
```
calebcartwright: Since `parser` is a route param with a fixed set of values, it will always be one of those allowed values (and all are lowercase), so we can simplify this a bit. I've inlined the conditional below in the suggested changes for `qs`
```suggestion
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  async handle({ parser }, { targetUrl, preset }) {
+    const { messages } = await this.fetch(targetUrl, preset, parser)
+
+    const messageTypes = this.transform(messages)
+    const badge = this.constructor.render({
+      messageTypes,
+    })
+    return badge
```
calebcartwright:
```suggestion
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,115 @@
...
+      accumulator[type] += 1
+      return accumulator
+    }, {})
+  }
+
+  async handle({ parser }, { targetUrl, preset }) {
+    const { messages } = await this.fetch(targetUrl, preset, parser)
+
+    const messageTypes = this.transform(messages)
+    const badge = this.constructor.render({
```
calebcartwright: We can just return this directly, no need to create the local var
```suggestion
 return this.constructor.render({ messageTypes })
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,115 @@
...
+      messageTypes,
```
calebcartwright:
```suggestion
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,115 @@
...
+    })
```
calebcartwright:
```suggestion
```

services/w3c/w3c-validation.service.js
seetd: Made the changes for the handle function

services/w3c/w3c-validation.service.js
seetd: Made this change as well

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,114 @@
...
+
+  static get defaultBadgeData() {
+    return {
+      label: 'w3c',
+    }
+  }
+
+  static render({ messageTypes }) {
+    return {
+      label: 'w3c',
```
calebcartwright: We can remove this since the label is set in the default badge data
```suggestion
      label: 'w3c',
```

services/w3c/w3c-validation.tester.js
```diff
@@ -0,0 +1,87 @@
+'use strict'
+
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('W3C Validation page conforms to standards with no preset and parser')
+  .get(
+    '/default.json?targetUrl=https://hsivonen.com/test/moz/messages-types/no-message.html&preset='
```
calebcartwright: Let's drop the empty query param from the end
```suggestion
    '/default.json?targetUrl=https://hsivonen.com/test/moz/messages-types/no-message.html'
```

services/w3c/w3c-validation.service.js
seetd: Done

services/w3c/w3c-validation.tester.js
seetd: Done. And sorry to take so much of your time in reviewing this

services/w3c/w3c-validation-helper.js
```diff
@@ -0,0 +1,157 @@
...
+  <p>
+    This badge relies on the https://validator.nu service to perform the validation. Please refer to https://about.validator.nu/ for the full documentation and Terms of service.
+    The following are required from the consumer for the badge to function.
+
+    <ul class="note">
+      <li>
+        Path:
+        <ul>  
+          <li>
+            parser (Optional can be left as blank): The parser that is used for validation. This is a passthru value to the service
```
calebcartwright: This text should probably be updated or removed now correct? Perhaps we could mention how `default` works, but with the current badge route definition one of those four values must be specified

services/w3c/w3c-validation-helper.js
```diff
@@ -0,0 +1,157 @@
...
+const html4Expression =
+  '^HTML\\s?4\\.01\\s?(Strict|Transitional|Frameset)\\s?,\\s?URL\\s?\\/\\s?XHTML\\s?1\\.0\\s?(Strict|Transitional|Frameset)\\s?,\\s?URL$'
+const xhtmlExpression =
+  '^(XHTML\\s?,\\s?SVG\\s?1\\.1\\s?,\\s?MathML\\s?3\\.0(\\s?,\\s?RDFa\\s?Lite\\s?1\\.1)?)|(XHTML\\s?1\\.0\\s?Strict\\s?,\\s?URL\\s?,\\s?Ruby\\s?,\\s?SVG\\s?1\\.1\\s?,\\s?MathML\\s?3\\.0)$'
+const svgExpression =
+  '^SVG\\s?1\\.1\\s?,\\s?URL\\s?,\\s?XHTML\\s?,\\s?MathML\\s?3\\.0$'
+const presetRegex = new RegExp(
+  `(${html5Expression})|(${html4Expression})|(${xhtmlExpression})|(${svgExpression})`,
+  'i'
+)
```
calebcartwright: I haven't tested these out to confirm, but they appear to be case sensitive. Is that by design? Does the W3C Checker API require case-correctness for the preset param?

services/w3c/w3c-validation-helper.js
```diff
@@ -0,0 +1,157 @@
...
+
+const getMessage = messageTypes => {
+  const messageTypeKeys = Object.keys(messageTypes)
+  messageTypeKeys.sort() // Sort to make the order error, warning for display
+
+  if (Object.keys(messageTypes).length === 0) {
```
calebcartwright:
```suggestion
  if (messageTypeKeys.length === 0) {
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,113 @@
...
+    return this._requestJson({
+      url: 'http://validator.nu',
+      schema,
+      options: {
+        qs: {
+          schema: getSchema(preset),
+          parser: parser === 'default' ? undefined : parser,
+          doc: targetUrl,
+          out: 'json',
+        },
```
calebcartwright: We may need to do uri encoding of the values for `schema` and `doc`, just to be safe

services/w3c/w3c-validation.tester.js
```diff
@@ -0,0 +1,87 @@
+'use strict'
+
+const t = (module.exports = require('../tester').createServiceTester())
+
```
calebcartwright: In order to minimize brittleness in our service tests, we don't typically assert an _exact_ value for the message. The common approach is to use `@hapi/joi` to assert that the message is what we expected. 

Here's a few examples from other services:
https://github.com/badges/shields/blob/master/services/sonar/sonar-violations.tester.js#L6-L11
https://github.com/badges/shields/blob/master/services/github/github-hacktoberfest.tester.js
https://github.com/badges/shields/blob/ce0ddf93fc75e905743ac40db55c62a9bf44a53c/services/symfony/symfony-insight-violations.tester.js

services/w3c/w3c-validation.tester.js
calebcartwright: No worries!

calebcartwright:
Sorry for all the different reviews on this one. There's a lot going on in order to implement this service that takes a while to read through. 

It's coming along great though, thanks for sticking with it!

services/w3c/w3c-validation-helper.js
seetd: The API takes a list of space delimited urls representing the schema. An example is     `http://s.validator.nu/svg-xhtml5-rdf-mathml.rnc, http://s.validator.nu/html5/assertions.sch, http://c.validator.nu/all/`. This seems too much to configure for the badge. So I took the preset concept from the GUI version of the validator that maps the presets to the string of schema urls. They are using a Dropdown so I guess there is no casing issue for them. For the Regex, I am using the `i` flag that ignores casing. The following are unit test cases for the preset.
```
      { preset: 'html,svg 1.1,mathml 3.0', expectedResult: true },
      { preset: 'HTML,SVG 1.1,MathML 3.0', expectedResult: true },
```
The only difference in my implementation from the GUI version is instead of doing `HTML + SVG 1.1 + MathML 3.0` I replaced the `+` with the `,`. As `+` has a special meaning in an url and I was having a hard time fighting to get that work well in both the testing and running the dialog help and also the actual badge. The other thing is that for the spaces surrounding `,` I made them optional. So the version we are using is a lot more forgiving.


Screen shots of the GUI can be found at this comment 
> Hey @calebcartwright,
> 
> That is right, as I was working on it. That was my thought about it. But for schema, the validator is actually using a preset. Then that actually selects a set of schemas to validate against.
> 
> Presets and Schema
> Screen Shot 2019-10-07 at 3 27 14 PM
> Screen Shot 2019-10-07 at 3 40 41 PM
> 
> Parser
> Screen Shot 2019-10-07 at 3 27 22 PM
> 
> Thanks

services/w3c/w3c-validation.service.js
seetd: I can do it for the doc. But it apparently no longer recognize the schema urls when I do either `encodeURI` or `encodeURIComponent`. The tests start returning the wrong results

services/w3c/w3c-validation.tester.js
seetd: Done I made the messages use Joi.string().regex checks

services/w3c/w3c-validation.service.js
seetd: Update I manage to encode the individual url for the schema property. The encodeURI lives in the getSchema function for schema.

services/w3c/w3c-validation-helper.js
seetd: Updated. Sorry forgot to remove that Object.keys call

services/w3c/w3c-validation-helper.js
seetd: Updated the text. I left it in for better explanation. I can remove it if required

services/w3c/w3c-validation-helper.js
calebcartwright: > So the version we are using is a lot more forgiving.

Oops! That's my fault for not reading down far enough. That's also great that we're more forgiving (I wanted to make sure we weren't being _more_ strict) 👍

services/w3c/w3c-validation.service.js
calebcartwright: Nice!

services/w3c/w3c-validation-helper.js
seetd: No problem. That was a long one.

services/w3c/w3c-validation.tester.js
calebcartwright: Excellent. Could we take this another step and do something like:

```js
const isW3CMessage = Joi.alternatives().try('validated', isErrorOnly, isWarningOnly, isErrorAndWarning),
```

I'm not familiar with the targetUrl's that are being used in the tests, nor if/when the results of the W3C checker may change for this service. If you feel pretty confident that the results for those targets will be static for a long period of time, then it would be good to include some inline comments/links to details about those targets, etc. for future readers.

If the results for those pages are a bit more dynamic (may change at any time/will likely change within the next several months/etc.), then it might be better to remove the `color` checks from these tests, and change the message to be `message: isW3CMessage,` and then create a .spec file and add some unit tests for the `render` function to validate the color

services/w3c/w3c-validation-helper.js
calebcartwright: Thanks, I like this as is. It provides helpful context to folks, especially the explanation of `default` 👍

services/w3c/w3c-validation.tester.js
seetd: The pages being tested again are supposed to be static. But how about making the colors `const isW3CColor = Joi.alternatives().try('brightgreen','red','yellow')`

calebcartwright: Sure, though I do think it would be good to have some tests that validate the color and messaging logic

seetd: I made the changes to the tests

seetd: There are unit tests in w3c-validation-helper.spec.js under the `getColor` and `getMessage` test suites to cover the combinations of MessageTypes derived from the response of the API for both the colors and messages.

Example

```
    it('returns "2 errors, 2 warnings" if 2 error and 2 warning message is provided', function() {
      const messageTypes = { error: 2, warning: 2 }

      const actualResult = getMessage(messageTypes)

      expect(actualResult).to.equal('2 errors, 2 warnings')
    })
```

calebcartwright: > There are unit tests in w3c-validation-helper.spec.js under the getColor and getMessage test suites to cover the combinations of MessageTypes derived from the response of the API for both the colors and messages.

Gotcha, don't worry about the render function then

> I am thinking of pulling out the logic from the transform method into a helper function and unit test that. What do you think?

it's fine to leave that inline in the same file as the service class definition. it's helpful to have the schema defining what the response object will look like in the same place as the code that's dealing with that response

services/w3c/w3c-validation-helper.spec.js
```diff
@@ -0,0 +1,321 @@
+'use strict'
+const { expect } = require('chai')
+const {
+  presetRegex,
+  getMessage,
+  getColor,
+  getSchema,
+} = require('./w3c-validation-helper')
+
```
calebcartwright: The tests look good as is, but you may want to also check out `sazerac` which we use elsewhere. It basically helps you get to the same endstate without having to define a lot of the boilerplate yourself. 

For example:
https://github.com/badges/shields/blob/master/core/base-service/auth-helper.spec.js#L19-L45
https://github.com/badges/shields/blob/4c159eff004f78b176ae72bb20337d9e95757074/core/base-service/auth-helper.spec.js#L19-L45

and
https://github.com/badges/shields/blob/master/services/jenkins/jenkins-build.spec.js
https://github.com/badges/shields/blob/4c159eff004f78b176ae72bb20337d9e95757074/services/jenkins/jenkins-build.spec.js#L11-L40

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,113 @@
...
+} = require('./w3c-validation-helper')
+const { BaseJsonService } = require('..')
+
+const schema = Joi.object({
+  messages: Joi.array()
+    .required()
+    .items(
+      Joi.object({
+        type: Joi.string()
+          .allow('info', 'error', 'non-document-error')
```
calebcartwright: I was poking around the review app for this PR (https://shields-staging-pr-4148.herokuapp.com/) and tried testing with a non-existent URL for `targetUrl`. It looks like the response we get from the W3C service for the 'not found' scenario is an HTTP 200 with a single `non-document-error` error that has a message like `"message": "fakefakefakefakefakefakefake.com: Name or service not known"`. 

Internally Shields has defaults that help service classes deal with the 'not found' scenarios, but that's driven off the HTTP response codes (for example if an usptream API returns an HTTP 404, the default badge is something like ) so we'll have to handle this scenario explicitly here.

I'd suggest augmenting the `transform` function to check for this scenario (looks like we can check for a single error with a message that ends with `Name or service not known`), and to then throw Shields' `NotFound` error. 

For example:
```js
const { NotFound } = require('..')
...
  transform(messages) {
    ... // new logic to detect not found scenario
    if (notFound) {
      throw new NotFound({  prettyMessage: 'target url not found' })
    }
    return messages.reduce((accumulator, message) => {
....

```

seetd: I tried implementing this but there is a false positive error from the validator. It handles the use case well when you pass it a URL that does not exist. But I found that it returns the same error message `{"url":"http://www.shields.io","messages":[{"type":"non-document-error","subType":"io","message":"c.validator.nu: Name or service not known"}]}` in one use case. The use case is when you pass the validator gets an url that exists but does not contain an iframe in the document and the schema for `HTML 4.01 Frameset, URL / XHTML 1.0 Frameset, URL`. Not sure if that is a bug or in the context of the validator it might not be wrong. Since it was trying to find an iframe which does not exists. I can change the service test to cover that false positive if that is the intention. To test the use case please run the following 3 calls and you can see the response.

1. https://validator.w3.org/nu/?doc=http://www.shields.io&out=json&schema=http://s.validator.nu/xhtml10/xhtml-frameset.rnc%20http://s.validator.nu/html4/assertions.sch%20http://c.validator.nu/all-html4/ (with frameset schemas that return an error of type 'non-document-error')
2. https://validator.w3.org/nu/?doc=http://www.shields.io&out=json (no schema and returns normal validation errors and warnings)

seetd: deleted last message. I was wrong about the output

seetd: Made the change to throw NotFound

services/w3c/w3c-validation-helper.spec.js
seetd: Replace forEach tests with `sazerac`

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,124 @@
...
+  static render({ messageTypes }) {
+    return {
+      message: getMessage(messageTypes),
+      color: getColor(messageTypes),
+    }
+  }
+
+  async fetch(targetUrl, preset, parser) {
+    return this._requestJson({
+      url: 'https://validator.w3.org/nu/',
```
seetd: Using the w3c URL instead of what happens to be an old version. The functionality appears to be similar at least for the params we are passing. And w3c will appear to be more lightly to be still around for a long time

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,124 @@
...
+const Joi = require('@hapi/joi')
+const { optionalUrl } = require('../validators')
+const {
+  documentation,
+  presetRegex,
+  getColor,
+  getMessage,
+  getSchema,
+} = require('./w3c-validation-helper')
+const { BaseJsonService } = require('..')
```
calebcartwright:
```suggestion
const { BaseJsonService, NotFound } = require('..')
```

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,124 @@
...
+const { NotFound } = require('..')
```
calebcartwright:
```suggestion
```

services/w3c/w3c-validation.service.js
calebcartwright: Thanks! Are you sure the check you've added fully covers the not found scenario? I'm still not seeing the `target url not found` message:
https://shields-staging-pr-4148.herokuapp.com/w3c-validation/default?targetUrl=https%3A%2F%2Fadfasdfasdfasdfadfadfadfasdfadf.com


and the json response from the w3c service does not seem to mention the 404 text in the error:
```json
{
  "url": "https://adfasdfasdfasdfadfadfadfasdfadf.com",
  "messages": [
     {
      "type": "non-document-error",
      "subType": "io",
      "message": "adfasdfasdfasdfadfadfadfasdfadf.com: Name or service not known"
    }
  ],
}
```

It'd also be good to add a service test to cover the not found scenario, since that's going to be entirely driven off the values in the API response from the upstream service

seetd: Looks implement the NotFound is going to be tricky then. This validation `https://validator.w3.org/nu/?doc=http://hsivonen.com/test/moz/messages-types/warning.html&out=json&schema=http://s.validator.nu/xhtml10/xhtml-frameset.rnc%20http://s.validator.nu/html4/assertions.sch%20http://c.validator.nu/all-html4/` which checks an existing page with an info in normal HTML DTD returns the same error message as your test case. That is kind of a false positive.

calebcartwright: Since you changed the W3C checker from `https://validator.nu/` to `https://validator.w3.org/nu/`, are there any updates needed to the values we're providing for the schema? I ask because that error message is complaining about a value we are providing to the upstream API, not the user provided value for the target url

`"message": "c.validator.nu: Name or service not known"`

seetd: There was no difference from what I tested so far. The same message appears on the validator.nu if I don't remember wrongly. There is a workaround I think. For the testcase you provided the hostname of the provided url is in the message instead of `c.validator.nu`. We could use that as a check

calebcartwright: Perhaps I'm doing something wrong, but I feel like I'm getting the expected response with https://validator.nu:

https://validator.nu/?doc=http://hsivonen.com/test/moz/messages-types/warning.html&out=json&schema=http://s.validator.nu/xhtml10/xhtml-frameset.rnc%20http://s.validator.nu/html4/assertions.sch%20http://c.validator.nu/all-html4/
```json
{ 
  "url":"http://hsivonen.com/test/moz/messages-types/warning.html",
  "messages":[ 
    { 
      "type":"error",
      "lastLine":6,
      "lastColumn":6,
      "firstColumn":1,
      "message":"Element “body” not allowed as child of element “html” in this context. (Suppressing further errors from this subtree.)",
      "extract":">\n</head>\n<body>\n<p cl",
      "hiliteStart":10,
      "hiliteLength":6
    },
    { 
      "type":"error",
      "lastLine":7,
      "lastColumn":13,
      "message":"“<” in an unquoted attribute value. Probable cause: Missing “>” immediately before.",
      "extract":"y>\n<p class=foo<bar></p>\n</bod",
      "hiliteStart":15,
      "hiliteLength":1
    },
    { 
      "type":"error",
      "lastLine":9,
      "lastColumn":7,
      "firstColumn":1,
      "message":"Element “html” is missing a required instance of child element “frameset”.",
      "extract":">\n</body>\n</html>",
      "hiliteStart":10,
      "hiliteLength":7
    }
  ]
}
```

but with `https://validator.w3.org/nu/` we're getting an error complaining about the values we are passing for the `schema` param, specifically for`c.validator.nu`:
https://validator.w3.org/nu/?doc=http://hsivonen.com/test/moz/messages-types/warning.html&out=json&schema=http://s.validator.nu/xhtml10/xhtml-frameset.rnc%20http://s.validator.nu/html4/assertions.sch%20http://c.validator.nu/all-html4/
``` json
{ 
  "url":"http://hsivonen.com/test/moz/messages-types/warning.html",
  "messages":[ 
    { 
      "type":"non-document-error",
      "subType":"io",
      "message":"c.validator.nu: Name or service not known"
    }
  ]
}
```

It seems like there's some correlation between the validator endpoint we are using and whether the schema values work

seetd: Ok looks like the preset and schema might not be the best way forward. I am going to look into this more. Sorry for the troubles

calebcartwright: > Ok looks like the preset and schema might not be the best way forward

Well, you're definitely closer to the details on this, but my initial question is simply whether we'd need to use different schema _values_ with https://validator.w3.org/nu/ (maybe instead of  `c.validator.nu` we'd need something like `c.validator.w3...`)? 

It also seems like things are working as-is with this implementation with https://validator.nu, so before we go down an entirely different road I think it would be worth revisiting using `https://validator.nu/` vs. `https://validator.w3.org/nu/` as our upstream endpoint. FWIW, Istarted getting HTTP 429 errors with https://validator.w3.org/nu/ after only a handful of requests which has me worried about rate limiting with that endpoint. I've not seen any such issues with https://validator.nu

seetd: It was interesting since `https://validator.nu/` or `https://validator.w3.org/nu/` appear to be running off the same source `https://github.com/validator/validator`.  https://about.validator.nu actually asks you to report issues to https://github.com/validator/validator. Given that you faced the limiting issue. I redirect the code to point back to https://validator.nu. I also found a new param doctype. This appears to be cleaner then using the schema which is what the `https://validator.nu/` UI is using. Although I tested against the API version of it `https://validator.nu/?doc=https%3A%2F%2Fhsivonen.com%2Ftest%2Fmoz%2Fmessages-types%2Fwarning.html&charset=%28detect+automatically%29&doctype=ISO%2FIEC+15445%3A2000+%28"ISO+HTML"%29&out=json` and it works. I am working if we should move of schema and use that instead.

services/w3c/w3c-validation-helper.js
```diff
@@ -51,11 +51,9 @@ const getSchema = preset => {
     } else {
       schema.push('http://s.validator.nu/xhtml10/xhtml-frameset.rnc')
     }
-    schema.push('http://s.validator.nu/html4/assertions.sch')
```
seetd: According this https://github.com/validator/validator/issues/559 it appears that this is being removed

services/w3c/w3c-validation.service.js
calebcartwright: very interesting indeed. if we've got things working here, then i'd be inclined to finish this out with the current-implementation (plus support for 'not found' scenario). if you're willing and interested in investigating an alternative then we can tackle that in a follow up

seetd: I think the last commit has the changes required for NoFound then. I added a new testcase for it as well. Thanks.

services/w3c/w3c-validation.service.js
```diff
@@ -0,0 +1,118 @@
...
+          out: 'json',
+        },
+      },
+    })
+  }
+
+  transform(messages) {
+    return messages.reduce((accumulator, message) => {
+      let { type } = message
+      if (type === 'non-document-error') {
```
calebcartwright: Is this check sufficient on its own? Is it possible to have a `non-document-error` that's _not_ due to an invalid/inaccessible url for the `doc` param?

seetd: Not sure if there might be others but the last 4 was the frameset error. That was fixed with the url change. The other option will be to do some form or regex check for the 404 message and another regex for the `"adfasdfasdfasdfadfadfadfasdfadf.com: Name or service not known"` check

calebcartwright: I suppose it is a matter of what the response looks like from the upstream API when the target url cannot be found, and under what circumstances that API response may include an object in `messages` array that has the `non-document-error` type.

My sense from the little bit of experimentation I've done is that we'll get a response with a single object in `messages` that will look something like this:

```json
{
  "messages": [
    {
      "type": "non-document-error",
      "subType": "io",
      "message": "{{target url domain name}}: Name or service not known"
    }
  ]
}
```

What I'm wondering is if we could ever see a response from the API that looks something like this:
```json
{
  "messages": [
    {
      "type": "non-document-error",
      "subType": "{{something other than io}}",
      "message": "{{some other type of message that's not a 404 on the url for doc"
    },
    {
      "type": "error",
      "message": "Element “meta” is missing one or more of the following attributes..."
    }
  ]
}
```
Because if so then I suspect it would be pretty confusing for users to see the target url not found badge 

Based on what you've seen so far, do you think that's a possible scenario? Or if the `messages` array has _any_ `non-document-error` type messages can we safely assume that's a target url "not found" scenario?

seetd: I did a search through the codebase at `https://github.com/validator/validator`, from a high level view. I think implementing the usual 404 might not be a good idea like what you mention. Reading the documentation in detail as well. There is a high chance that there are more types of non-document-error besides those indicating that the url is not valid

```
"non-document-error" signifies an error that causes the checking to end in an indeterminate state because the document being validated could not be examined to the end. Examples of such errors include broken schemas, bugs in the validator and IO errors. (Note that when a schema has parse errors, they are first reported as errors and then a catch-all non-document-error is also emitted.)
```

calebcartwright: Thanks for digging into it. My takeaway is that we need to do some additional checks to more accurately detect the target url not found scenario. I'd suggest the following:

* include `subType` and `message` in the schema on the objects in the messages array so that we'll have access to those values
* In the `transform` function, include the following conditions to better detect target url not found: 
    1. (optional) check `messages.length` for a single message. my assumption is that if there's an io error with the target url (specified via the `doc` param) that there wouldn't be any other messages
   2. check `type` for `non-document-error`
   3. check `subType` for `io`
   4. check `message` for a `Name or service not known` suffix
   5. (optional) check the full `message`, expecting something like `{{targetUrl with http(s) and trailing paths stripped}}: Name or service not known` For example: if the targetUrl was `https://notrealgithub.com/foo` then `message` should be `notrealgithub.com: Name or service not known`

services/w3c/w3c-validation.service.js
```diff
@@ -86,13 +88,29 @@ module.exports = class W3cValidation extends BaseJsonService {
...
-  transform(messages) {
-    return messages.reduce((accumulator, message) => {
-      let { type } = message
-      if (type === 'non-document-error') {
-        throw new NotFound({ prettyMessage: 'target url not found' })
+  transform(url, messages) {
+    if (messages.length === 1) {
+      const { subType, type, message } = messages[0]
+      if (type === 'non-document-error' && subType === 'io') {
+        let notFound = false
```
seetd: More lines of code but I thought this is the easiest to understand instead of some crazy regex

services/w3c/w3c-validation.service.js
seetd: Sorry for the radio silence had to swing back to work on a Nanodegree. I made the notfound tests stricter.

services/w3c/w3c-validation.service.js
calebcartwright: https://shields-staging-pr-4148.herokuapp.com/w3c-validation/default?targetUrl=https%3A%2F%2Fgasdfasdfithub.com

🙌

services/w3c/w3c-validation.tester.js
```diff
@@ -0,0 +1,100 @@
...
+  isErrorOnly,
+  isWarningOnly,
+  isErrorAndWarning
+)
+const isW3CColors = Joi.alternatives().try('brightgreen', 'red', 'yellow')
+t.create(
+  'W3C Validation page conforms to standards with no preset and parser with brightgreen badge'
+)
+  .get(
+    '/default.json?targetUrl=https://hsivonen.com/test/moz/messages-types/no-message.html'
```
calebcartwright: By the way where did you find these pages? Seems like they were setup explicitly to be used as test targets

services/w3c/w3c-validation.service.js
calebcartwright: Agreed!

seetd:
No problem it was a fun learning experience

services/w3c/w3c-validation.tester.js
seetd: They were actually test pages that was provided in the documentation of the validator project
https://github.com/validator/validator/wiki/Service-%C2%BB-HTTP-interface

chris48s:
Yes - well done and thanks to both of you for persisting with the review one. I think this must be one of the most commented PRs on the repo.

Issue #4025, opened 2019-09-15
libraries.io dependents badge with scoped NPM packages does not work


johnwebbcole:
libraries.io dependents badge with scoped NPM packages does not work

I'm trying to add a badge using libraries.io dependents for a scoped package.  I noticed that [libraries.io encodes the scoped name](https://libraries.io/npm/@jwc%2Fjscad-utils) into the package name, like `@jwc%2Fjscad-utils`, but this doesn't work with shields.io.

```
![Dependents](http://img.shields.io/librariesio/dependents/npm/@jwc%2Fjscad-utils
```

![Dependents](http://img.shields.io/librariesio/dependents/npm/@jwc%2Fjscad-utils)

If there is already a way to use scoped NPM packages, I wasn't able to find it in the documentation or another issue pointing to more info.

Thanks!

calebcartwright:
Thanks for reporting this @johnwebbcole! If anyone is interested in fixing this bug, the solution here will probably be very similar to the ones in #3274 and #3570

paulmelnikow:
@johnwebbcole Any chance you'd be interested in taking a look at this?

netsgnut:
Just stumbled upon this problem right now. Will try and see if I could hack something out of it.

calebcartwright:
Closed via #4151

PR #4151, opened 2019-10-09
Allow [librariesio] badges referencing scoped npm package (#4025)

netsgnut:
This PR proposes an implementation of allowing libraries.io badges (dependents, sourcerank...) to work with scoped npm packages. It is inspired by how the current [bundlephobia service](https://github.com/badges/shields/blob/2.2.1/services/bundlephobia/bundlephobia.service.js#L23) handles it.

Once merged, it should solve #4025.

Please see if it looks good to you. Thanks!

### DIFF
--- a/services/librariesio/librariesio-common.js
+++ b/services/librariesio/librariesio-common.js
@@ -11,10 +11,12 @@ const projectSchema = Joi.object({
   rank: anyInteger,
 }).required()
 
-async function fetchProject(serviceInstance, { platform, packageName }) {
+async function fetchProject(serviceInstance, { platform, scope, packageName }) {
   return serviceInstance._requestJson({
     schema: projectSchema,
-    url: `https://libraries.io/api/${platform}/${packageName}`,
+    url: `https://libraries.io/api/${encodeURIComponent(platform)}/${
+      scope ? encodeURIComponent(`${scope}/`) : ''
+    }${encodeURIComponent(packageName)}`,
     errorMessages: { 404: 'package not found' },
   })
 }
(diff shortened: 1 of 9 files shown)


netsgnut:
@calebcartwright Thanks for heads-up! In my local environment (with both node 10 and 12), I could not reproduce the `ValidationError` apart from the occasional timeout. Could it be some hiccups upstream instead?

calebcartwright:
I've kicked off a new run of the service tests, and things are indeed looking better. If you're interested and willing, you may want to add an increased timeout on those service tests, as we do have fairly recurring timeout issues with the libraries.io API response in our daily service test runs

calebcartwright:
Also could you do a rebase or merge in the latest from the `master` branch? We have a check configured that blocks merges to master unless the branch is up to date

netsgnut:
> I've kicked off a new run of the service tests, and things are indeed looking better. If you're interested and willing, you may want to add an increased timeout on those service tests, as we do have fairly recurring timeout issues with the libraries.io API response in our daily service test runs

Good idea, perhaps let's make all those tests timeout in 10s?

> Also could you do a rebase or merge in the latest from the `master` branch? We have a check configured that blocks merges to master unless the branch is up to date

No problem :ok_hand:

calebcartwright:
> Good idea, perhaps let's make all those tests timeout in 10s?

Sounds good!

netsgnut:
Looks like server hiccups again :(

calebcartwright:
Thanks so much @netsgnut!

netsgnut:
@calebcartwright No problem, it was fun :)

Issue #3202, opened 2019-03-13
dynamic xml badge: XPath functions not supported?


tooomm:
Are you experiencing an issue with...

- [shields.io](https://shields.io/#/)

:beetle: **Description**

XML: https://raw.githubusercontent.com/flathub/io.github.Cockatrice.cockatrice/master/io.github.Cockatrice.appdata.xml
Query: `string(//releases/release/@version)`

I used https://www.freeformatter.com/xpath-tester.html as helper with the query and it works there.

I want to display `2.7.0` from the above xml source.
The value/string is located in the version attribute of the release node/key.

:link: **Link to the badge**

![](https://img.shields.io/badge/dynamic/xml.svg?color=4ac41d&label=version%20in%20appdata.xml&query=string%28%2F%2Freleases%2Frelease%2F%40version%29&url=https%3A%2F%2Fraw.githubusercontent.com%2Ftooomm%2Fio.github.Cockatrice.cockatrice%2Fmaster%2Fio.github.Cockatrice.appdata.xml)

```
https://img.shields.io/badge/dynamic/xml.svg?color=4ac41d&label=version%20in%20appdata.xml&query=string%28%2F%2Freleases%2Frelease%2F%40version%29&url=https%3A%2F%2Fraw.githubusercontent.com%2Ftooomm%2Fio.github.Cockatrice.cockatrice%2Fmaster%2Fio.github.Cockatrice.appdata.xml
```

paulmelnikow:
Hi, thanks for the report.

Huh, I wonder what is happening here. It’s strange that it says “internal error” and not “shields internal error.”

The Dynamic XML tests are currently failing (see #3089) so fixing that is probably the first order of business.

RedSparr0w:
It seems like it may just be the XPath functions which are not working:

[![XML Badge](https://img.shields.io/badge/dynamic/xml.svg?color=4ac41d&label=version%20in%20appdata.xml&query=//releases/release/@version&url=https://raw.githubusercontent.com/tooomm/io.github.Cockatrice.cockatrice/master/io.github.Cockatrice.appdata.xml)](https://img.shields.io/badge/dynamic/xml.svg?color=4ac41d&label=version%20in%20appdata.xml&query=//releases/release/@version&url=https://raw.githubusercontent.com/tooomm/io.github.Cockatrice.cockatrice/master/io.github.Cockatrice.appdata.xml)
```
[...]

PR #4153, opened 2019-10-10
Support for XPath functions in [dynamicxml] badge

platan:
Fixes https://github.com/badges/shields/issues/3202. 
XPath functions are well described at https://developer.mozilla.org/en-US/docs/Web/XPath/Functions. https://devhints.io/xpath is also a nice source of examples (with comparison to CSS selectors).


calebcartwright:
Thanks @platan! I'll take a look at this in a few hours

platan:
Badge reported in https://github.com/badges/shields/issues/3202 now works properly:

https://shields-staging-pr-4153.herokuapp.com/badge/dynamic/xml?label=version&query=string%28%2F%2Freleases%2Frelease%2F%40version%29&url=https%3A%2F%2Fraw.githubusercontent.com%2Fflathub%2Fio.github.Cockatrice.cockatrice%2Fmaster%2Fio.github.Cockatrice.appdata.xml

services/dynamic/dynamic-xml.service.js
```diff
@@ -41,21 +41,25 @@ module.exports = class DynamicXml extends BaseService {
...
-        accum.push(node.firstChild.data)
-      }
+    if (
+      typeof values === 'string' ||
+      typeof values === 'number' ||
+      typeof values === 'boolean'
+    ) {
+      values = [values]
+    } else {
+      values = values.reduce((accum, node) => {
```
calebcartwright: Are we sure that `xpath.select(...)` will always return a `string`, `number`, `boolean`, or an array? This seems like a superior fix for the type errors we were getting before, but I just want to make sure we can't still end up trying to call `.reduce` on an undefined `values` somehow

platan: @calebcartwright  I wasn't able to find a query which returns something different than types mentioned above. But we can revert previous code to support these unusual values returned by `select`.

platan:
@calebcartwright Thanks for a review and a good suggestion!

Issue #2688, opened 2019-01-08
Experiment with retry() for failed service tests


paulmelnikow:
Often service tests fail once, only to run correctly the next time through. IcedFrisby has a [`retry()` function](https://github.com/IcedFrisby/IcedFrisby/blob/master/API.md#retrycount-backoff) which may be a good solution for isolated failures. It could make the daily suite, and especially PR builds, more reliable, with a teeny tiny amount of effort.

We could experiment with:

- Turning this on for all the tests
- Attaching it to flaky tests
- Turning it on for all the tests, but only in the daily tests (or PRs?)

platan:
Retries are enabled in daily tests and in PR tests. I'm closing this issue now, but I'm happy to experiment with different options if needed.

paulmelnikow:
Thanks @platan!

PR #4166, opened 2019-10-12
An option to retry a failed service; test on [dynamicxml]

platan:
This PR adds an option to retry all failed service tests. It uses IcedFrisby's [retry()](https://github.com/IcedFrisby/IcedFrisby/blob/master/API.md#retrycount-backoff) function. 
Usage:
```bash
RETRY_COUNT=3 RETRY_BACKOFF=0 npm run test:services -- --only=dynamicxml
```
Retried tests are marked with `R` in logs. An example:
```
    [live] XML from url | invalid url
RR      ✓ 
	[ GET .json?url=https://github.com/badges/shields/raw/master/notafile.xml&query=//version ] (413ms)
```
Default values:
- count: 0
- backoff: 0

I've tested it manually by setting different values of a timeout for specific service tests. 
I would like to use the functionality from this PR for https://github.com/badges/shields/issues/2688.

### DIFF
--- a/core/service-test-runner/cli.js
+++ b/core/service-test-runner/cli.js
@@ -15,6 +15,11 @@
 // Run tests on a given instance:
 //   SKIP_INTERCEPTED=TRUE TESTED_SERVER_URL=https://test.shields.io npm run test:services --
 //
+// Run tests with given number of retries and backoff:
+//   RETRY_COUNT=3 RETRY_BACKOFF=100 npm run test:services --
+// Retry option documentation:
+// https://github.com/IcedFrisby/IcedFrisby/blob/master/API.md#retrycount-backoff
+//
 // Service tests are run in CI in two cases: scheduled builds and pull
 // requests. The scheduled builds run _all_ the service tests, whereas the
 // pull requests run service tests designated in the PR title. In this way,
@@ -59,6 +64,9 @@ const Runner = require('./runner')
 
 require('../unhandled-rejection.spec')
 
+const retry = {}
+retry.count = process.env.RETRY_COUNT || 0
+retry.backoff = process.env.RETRY_BACKOFF || 0
 let baseUrl, server
 if (process.env.TESTED_SERVER_URL) {
   baseUrl = process.env.TESTED_SERVER_URL
@@ -77,7 +85,7 @@ if (process.env.TESTED_SERVER_URL) {
 }
 
 const skipIntercepted = envFlag(process.env.SKIP_INTERCEPTED, false)
-const runner = new Runner({ baseUrl, skipIntercepted })
+const runner = new Runner({ baseUrl, skipIntercepted, retry })
 runner.prepare()
 
 // The server's request cache causes side effects between tests.
--- a/core/service-test-runner/cli.js
+++ b/core/service-test-runner/cli.js
@@ -65,8 +65,8 @@ const Runner = require('./runner')
 require('../unhandled-rejection.spec')
 
 const retry = {}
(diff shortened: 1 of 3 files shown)


core/service-test-runner/service-tester.js
```diff
@@ -116,7 +116,7 @@ class ServiceTester {
    * @param {string} attrs.baseUrl base URL for test server
    * @param {boolean} attrs.skipIntercepted skip tests which intercept requests
```
calebcartwright: I don't recall offhand how to best handle a nested object param, but we should probably update the jsdoc here to account for the new retry param

core/service-test-runner/cli.js
```diff
@@ -15,6 +15,11 @@
 // Run tests on a given instance:
 //   SKIP_INTERCEPTED=TRUE TESTED_SERVER_URL=https://test.shields.io npm run test:services --
 //
+// Run tests with given number of retries and backoff:
+//   RETRY_COUNT=3 RETRY_BACKOFF=100 npm run test:services --
```
PyvesB: Shall we indicate that it's in milliseconds here?

core/service-test-runner/cli.js
```diff
@@ -59,6 +64,9 @@ const Runner = require('./runner')
 
 require('../unhandled-rejection.spec')
 
+const retry = {}
+retry.count = process.env.RETRY_COUNT || 0
```
calebcartwright: Will these need to be explicitly cast to numbers/integers? I'm not familiar with IcedFrisby's retry, but I've run into issues with env vars being stringified and then libs that are expecting numbers bark at those values

paulmelnikow:
This is awesome! Thanks so much for taking this on, @platan!

Review seems well under control though I do have one question. Will we set these env vars for the service tests in CI and daily tests? Should we turn on the retries on by default so they run in dev too?

core/service-test-runner/service-tester.js
platan: Thanks! https://jsdoc.app/tags-param.html describes this case:

> If a parameter is expected to have a specific property, you can document that property by providing an additional @param tag. For example, if an employee parameter is expected to have name and department properties, you can document it as follows:
> Documenting a parameter's properties
> 
> /**
>  * Assign the project to an employee.
>  * @param {Object} employee - The employee who is responsible for the project.
>  * @param {string} employee.name - The name of the employee.
>  * @param {string} employee.department - The employee's department.
>  */
> Project.prototype.assign = function(employee) {
>     // ...
> };

I expected ESLint+JSDoc plugin to catch it. But JSDoc plugin doesn't support a destructuring parameter: https://github.com/gajus/eslint-plugin-jsdoc/issues/11

core/service-test-runner/cli.js
platan: Thanks. Done.

core/service-test-runner/cli.js
platan: [Documentation](https://github.com/IcedFrisby/IcedFrisby/blob/master/API.md#retrycount-backoff) says that count and backoff should be numbers, but `retry` works with strings. Nevertheless, I converted it to integers.

platan:
@paulmelnikow I wanted to start with retries in daily tests. But we can add them to service tests in CI as well. 
But I think we shouldn't turn it on in dev. If you get a timeout (it's the only error after which retry can help) you should probably increase the default timeout and rerun the test manually.

paulmelnikow:
That sounds good to me!

PR #4168, opened 2019-10-12
Support [JenkinsCoverage] through the Code Coverage API plugin

### LINKED ISSUE #3785: JenkinsCoverage from Unified Coverage Plugin
:clipboard: **Description**
Jenkins has a number of plugins available for code coverage information. We currently support coverage badges for the [JaCoCo](https://plugins.jenkins.io/jacoco) and [Cobertura](https://plugins.jenkins.io/cobertura) plugins, but not the newer [Unified Coverage Plugin](https://jenkins.io/blog/2018/08/17/code-coverage-api-plugin-1/)

Refs #3761 

:link: **Data**
-  **Is there a public API?** Yes
- **Does the API requires an API key?** Same as other Jenkins badges (no for public instances, yes for private instances)
- **Link to the API documentation**: https://github.com/jenkinsci/code-coverage-api-plugin#rest-api

:microphone: **Motivation**
Allow users that leverage the coverage plugin to get coverage badges from Shields (without requiring them to use a different plugin)

PyvesB:
Closes #3785.

The plugin is called "Code Coverage API", so we could use `code-coverage-api` in the badge's path to mirror what we do with the JaCoCo and Cobertura plugins, but that gets quite lengthy and repeats the `coverage` word already present in the base path. Therefore, I went for the short and simple `api`.

### DIFF
--- a/services/jenkins/jenkins-coverage.service.js
+++ b/services/jenkins/jenkins-coverage.service.js
@@ -21,6 +21,7 @@ const formatMap = {
     }).required(),
     treeQueryParam: 'instructionCoverage[percentage]',
     transform: json => ({ coverage: json.instructionCoverage.percentage }),
+    pluginSpecificPath: 'jacoco',
   },
   cobertura: {
     schema: Joi.object({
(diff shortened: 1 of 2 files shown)


services/jenkins/jenkins-coverage.service.js
```diff
@@ -47,9 +48,48 @@ const formatMap = {
...
+const documentation = `
+<p>
+  We support coverage metrics from a variety of Jenkins plugins:
+  <ul>
+    <li><a href="https://plugins.jenkins.io/jacoco">JaCoCo</a></li>
+    <li><a href="https://plugins.jenkins.io/cobertura">Cobertura</a></li>
+    <li>Any plugin which integrates with the <a href="https://plugins.jenkins.io/code-coverage-api">Code Coverage API</a> (e.g. llvm-cov, Cobertura 1.13+, etc.)</li>
+  </ul>
+</p>
+`
```
calebcartwright: 👍

calebcartwright: Actually I just noticed some weird formatting on the list items in the review app:
image

PyvesB: Hmmmm, interesting, I've got a different result: jenkins-coverage

Do you have strange results with other badges that have lists in their documentation text (e.g. Matrix)?

calebcartwright: Yeah I am seeing something similar, although it is slightly less pronounced since the individual list items are closer in length. Not a blocker for this by any means, but perhaps something we want to dig into later on. 

I'm seeing this in Chrome on Win 10 FWIW

PyvesB: I managed to reproduce this in Opera so it's probably a Chromium vs. Firefox thing. Will open a separate issue.

PyvesB: See #4170.

PyvesB:
Thanks for reviewing! The implementation was easy on this one, but figuring out what to do in the first place was slightly more tricky. 😄

Issue #3273, opened 2019-04-06
Blocked by CORS policy


ApertureDevelopment:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

When I try using the badges in a Steam addon description, I get this error and the badge doesnt show up:

```
Access to image at 'https://img.shields.io/github/issues-raw/Aperture-Development/MSync-2.png' from origin 'https://steamcommunity.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

:link: **Link to the badge**

https://img.shields.io/github/issues-raw/Aperture-Development/MSync-2.png

Steps for reproduction:

1. Generate a badge
2. Add it like this: ``[img]{badge}[/img]`` to your addon description

paulmelnikow:
Hi, thanks for the report. I don’t see any reason we couldn’t add CORS headers.

ApertureDevelopment:
When will that be fixed?

paulmelnikow:
The maintainers and contributors all work on this project in their spare time, so there’s no timeline I’m afraid! This is a community project; anyone who is interested in trying to solve the problem is welcome to. The maintainers are happy to provide guidance and review code. (If you don’t want to or aren’t able to, no worries! Maybe someone else will come along and work on it.)

In terms of technical advice, I’ve used the [cors package](https://www.npmjs.com/package/cors) before, which is the standard way to go about this. Though I’ve never done it with Scoutcamp (that’s the web framework Shields uses) and I’m not sure if there is a CORS handler built in. The relevant bits of code would probably belong in `core/server`, and a corresponding test would belong there as well.

paulmelnikow:
It may also be worth researching whether anything has to be configured in Cloudflare for CORS.

chris48s:
I think all the other headers we set are passed through by CloudFlare without having to whitelist them. The same-origin policy is implemented in the browser sandbox, so we should only need to add the header.

Prouser123:
We **should** just be able to add `Access-Control-Allow-Origin: *` as a header to every request.

CloudFlare, no matter what settings you set, will not interfere with CORS, and will pass it through unaltered.

paulmelnikow:
Fixed in #4171 🎉

PR #4171, opened 2019-10-13
Add CORS Header to every request

TagnumElite:
In Issue #3273 Prouser123 said to just `Access-Control-Allow-Origin: *` have as a header to every request.

I did exactly that. I have confirmed that this works by using an [heroku app](https://shields-cors.herokuapp.com/) and a [workshop file](https://steamcommunity.com/sharedfiles/filedetails/?id=1875923653) running the changes.

I don't know whether or not this is the correct method but it works for me.

### DIFF
--- a/core/server/server.js
+++ b/core/server/server.js
@@ -220,11 +220,6 @@ class Server {
         })
       )
     })
-
-    camp.handle((req, res, next) => {
-      res.setHeader('Access-Control-Allow-Origin', '*')
-      next()
-    })
   }
 
   /**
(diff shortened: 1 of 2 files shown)


core/server/server.js
```diff
@@ -323,6 +323,11 @@ class Server {
     const { apiProvider: githubApiProvider } = this.githubConstellation
     suggest.setRoutes(allowedOrigin, githubApiProvider, camp)
 
+    camp.handle((req, res, next) => {
```
paulmelnikow:
```suggestion
    # https://github.com/badges/shields/issues/3273
    camp.handle((req, res, next) => {
```

PR #4176, opened 2019-10-14
Enable retries for service tests on CI

platan:
Agreed in https://github.com/badges/shields/pull/4166#issuecomment-541414778, this PR enables retries in service tests on CI. Similar change in daily test: https://github.com/badges/daily-tests/pull/15
I think it solves https://github.com/badges/shields/issues/2688.


platan:
@chris48s Thanks!

PR #4180, opened 2019-10-15
Refactor [wordpress] downloads and add interval to theme downloads

TagnumElite:
I've cleaned up the `wordpress-downloads` service

I've also added an interval for theme downloads and removed unnecessary `DownloadsForInterval` function.
As well as added tests for the interval theme downloads and some missing year interval tests.


chris48s:
Nice. Good to have more consistency on the variants we support for plugins/themes. This also makes these nicer to edit in the modal:

Screenshot at 2019-10-15 19-24-10

Issue #4169, opened 2019-10-12
GitHub OAuth app/token pool updates


calebcartwright:
In order for shields Shields to support integrating with GitHub's new(ish) Package Registry offering, we'll need the GitHub tokens in our pool to have the `read:packages` permission (which they currently do not). 

Some action items:

* [ ] Transferring the current OAuth App. There's been past discussion about transferring the OAuth App Shields.io uses to Shields org and I think it probably makes sense to do that as part of this work
* [ ] Determine how to best get the required token permissions for Shields.io. There's relevant discussion over in #4101 (starting at https://github.com/badges/shields/pull/4101#issuecomment-536309798 and continuing below)

chris48s:
Another relevant question here:

At the moment, we've got a pile of tokens. Is there any mechanism to find out which permissions are granted to a token (other than "try and do a thing that needs permission X and see if it works"). I'm thinking if we ask users to grant a token an extra permission, do we then have any mechanism to know which tokens in our pool now have the extra permission and which ones don't?

paulmelnikow:
Hmm, good question. I would think “try it and see if it works”, though maybe there is an endpoint that checks the rights of individual tokens?

calebcartwright:
It looks like there is some token scope checking endpoints

https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/ 

```bash
curl -H "Authorization: token OAUTH-TOKEN" https://api.github.com/users/codertocat -I
HTTP/1.1 200 OK
X-OAuth-Scopes: repo, user
X-Accepted-OAuth-Scopes: user
```

paulmelnikow:
I wrote a little CLI utility for working with big piles of GitHub tokens. We could add the scope checking to that: https://github.com/paulmelnikow/github-limited

chris48s:
We talked about this a bit earlier today.

Right now, we've got zero tokens that have the necessary permissions to do this and no ability to collect tokens that have it or request it. Regardless of how we're going to manage tokens once we have them, one of the first things we need to do is give ourselves the ability to collect tokens which do have the permissions we need. This is a necessary first step before we can work out how to manage the pool or add badges that need special permissions.

At the moment, tokens are collected via a GitHub OAuth app https://img.shields.io/github-auth which is tied to @espadrine 's account.

One possible approach we could take would be:

[...]

PR #4184, opened 2019-10-17
add [GitHubPackageRegistryVersion] badge

### LINKED ISSUE #4183: Badge Request: GitHub Package Registry - Version Badge
:clipboard: **Description**
Download count badges for packages hosted on the GitHub Package Registry

- Which service is this badge for e.g: GitHub Package Registry 
- What sort of information should this badge show? Download counts, in whichever intervals the GitHub API makes available

:link: **Data**

- Is there a public API?  Yes
- Does the API requires an API key? Auth should be used like our other GH services, specifically leveraging our [BaseGraphQlService](https://github.com/badges/shields/blob/master/core/base-service/base-graphql.js). The [RegistryPackageStatistics](https://developer.github.com/v4/object/registrypackagestatistics/) will be of use
- Link to the API documentation. https://developer.github.com/v4/

:microphone: **Motivation**
Because people want them! 😄 The motivation/benefits for the version badges for the GH Package registry is the same as the other registries we support

Related: #3608 and #4071

calebcartwright:
Closes #4183 

Much like #4101, this is blocked on #4169, but I figure it'd be better to implement some of these GitHub Package Registry badges so we're ready to go upon the resolution of #4169


services/github/github-package-registry-version.service.js
```diff
@@ -0,0 +1,185 @@
...
+
+module.exports = class GithubPackageRegistryVersion extends GithubAuthV4Service {
+  static get category() {
+    return 'version'
+  }
+
+  static get route() {
+    return {
+      base: 'github/packages',
+      pattern: ':type(v|vpre)/:user/:repo/:packageName',
```
calebcartwright: Either GitHub's Package Regisry does not yet correctly identify "prerelease" package versions, or the GraphQL is just always treating all package versions as non-pre 🤷‍♂ 

The end result, at least for now, is that we'll end up displaying the same version value (the one GitHub tells us is the latest) regardless of whether the user selects `v` or `vpre` for `type`

calebcartwright:
In light of current cleanup efforts I'm going to go ahead and close this. Can circle back whenever the token blockers are resolved

PR #4187, opened 2019-10-17
Add Crates.io recent downloads badge (fixes #4186), run [Crates]

### LINKED ISSUE #4186: Badge Request - Crates.io Recent Downloads
:clipboard: **Description**

It would be helpful if Shields.io exposed a badge for recent downloads of a crate on Crates.io, as well as the existing total downloads badge.

The display would look something like:

I'd like to pick this up as a PR later today/over the weekend, but I thought it'd be a good idea to raise an issue for feedback first.

:link: **Data**

The data is retrievable from the same API endpoint that the current 'downloads' badge uses, via the `recent_downloads` field.

It currently can only be accessed for the crate as a whole, not on a per-version basis - therefore the `:version` URL parameter would need to be either disallowed or ignored for this variant of the downloads badge.

:microphone: **Motivation**

Recent downloads are a better metric of the *current* popularity of a library than total downloads.

The concrete usecase for this is for an ecosystem guide that the Rust Game Development Working Group are working on - we'd like to display the recent downloads of various crates in a wiki page.

17cupsofcoffee:
Summary of the changes:

* Added a 'recent downloads' badge for Crates.io.
  * I exposed it at `/crates/dr/:crate` (r for recent), as this seemed consistent with the existing routes in that namespace.
  * The data for recent downloads is only available for the crate as a whole, not individual versions, so I made the service bail out early with an error if you try to do that, to avoid unneccecary calls to Crates.io.
* Cleaned up some of the existing code.
* Updated the examples.
    * I left `/d/:crate/:version` out of the examples, because it does the same thing as `/dt/:crate/:version` and the latter makes more sense to me. The URL still works, though.
    * The URLs in the examples now include the relevant variants instead of `:variant` - this seems to play nicer with the modal editor.

Hopefully this is okay, first time contributing :)

image

### DIFF
--- a/services/crates/crates-downloads.service.js
+++ b/services/crates/crates-downloads.service.js
@@ -2,6 +2,7 @@
 
 const { downloadCount: downloadCountColor } = require('../color-formatters')
 const { metric } = require('../text-formatters')
+const { InvalidParameter, NotFound } = require('../../core/base-service/errors')
 const { BaseCratesService, keywords } = require('./crates-base')
 
 module.exports = class CratesDownloads extends BaseCratesService {
@@ -97,7 +98,9 @@ module.exports = class CratesDownloads extends BaseCratesService {
       /* crates.io doesn't currently expose
          recent download counts for individual
          versions */
-      return { message: 'no data' }
+      throw new InvalidParameter({
+        prettyMessage: 'recent downloads not supported for specific versions',
+      })
     }
 
     const json = await this.fetch({ crate, version })
@@ -108,7 +111,7 @@ module.exports = class CratesDownloads extends BaseCratesService {
          or
          https://crates.io/api/v1/crates/libc/0.1.76
          returns a 200 OK with an errors object */
-      return { message: json.errors[0].detail }
+      throw new NotFound({ prettyMessage: json.errors[0].detail })
     }
 
     const downloads = this.transform({ variant, json })
--- a/services/crates/crates-downloads.service.js
+++ b/services/crates/crates-downloads.service.js
@@ -12,7 +12,7 @@ module.exports = class CratesDownloads extends BaseCratesService {
   static get route() {
     return {
       base: 'crates',
-      pattern: ':variant(d|dv)/:crate/:version?',
+      pattern: ':variant(d|dv|dr)/:crate/:version?',
     }
   }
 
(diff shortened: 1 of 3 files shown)


services/crates/crates-downloads.service.js
```diff
@@ -59,7 +81,25 @@ module.exports = class CratesDownloads extends BaseCratesService {
     }
   }
 
+  _extractDownloads({ variant, json }) {
```
calebcartwright: Our convention is to use the name `transform` for this type of function that takes the raw response and returns the data points, so let's rename this
```suggestion
  transform({ variant, json }) {
```

services/crates/crates-downloads.service.js
```diff
@@ -71,14 +111,8 @@ module.exports = class CratesDownloads extends BaseCratesService {
...
 
-    let downloads
-    if (variant === 'dv') {
-      downloads = json.version
-        ? json.version.downloads
-        : json.versions[0].downloads
-    } else {
-      downloads = json.crate ? json.crate.downloads : json.version.downloads
-    }
+    const downloads = this._extractDownloads({ variant, json })
```
calebcartwright:
```suggestion
  const downloads = this.transform({ variant, json })
```

services/crates/crates-downloads.service.js
17cupsofcoffee: Done!

services/crates/crates-downloads.service.js
```diff
@@ -59,7 +81,25 @@ module.exports = class CratesDownloads extends BaseCratesService {
...
+    }
+  }
+
   async handle({ variant, crate, version }) {
+    if (variant === 'dr' && version) {
+      /* crates.io doesn't currently expose
+         recent download counts for individual
+         versions */
+      return { message: 'no data' }
+    }
```
calebcartwright: There's a couple ways we could handle this scenario:

1. Basically the same as what's here, but with our `InvalidParameter` error, perhaps something like:
```js
      throw new InvalidParameter { prettyMessage: 'recent downloads not supported for versions' }
      // Note we'd also need to import
      // const { InvalidParameter } = require('..')
```
2. Just return the recent downloads for crate itself. 

I think I'd prefer option 1, as option 2 may lead to some confusion. Thoughts?

17cupsofcoffee: I agree that option 2 is confusing! I like your first suggestion better.

Might it also be worth me wrapping the other error (the one that gets returned if the crate version doesn't exist) in a `NotFound` or an `Inaccessible` while I'm at it?

calebcartwright: > Might it also be worth me wrapping the other error (the one that gets returned if the crate version doesn't exist) in a NotFound or an Inaccessible while I'm at it?

Definitely (and thanks!). `NotFound` would be perfect for that use case 👍

17cupsofcoffee: Done :)

services/crates/crates-downloads.service.js
```diff
@@ -2,6 +2,7 @@
 
 const { downloadCount: downloadCountColor } = require('../color-formatters')
 const { metric } = require('../text-formatters')
+const { InvalidParameter, NotFound } = require('../../core/base-service/errors')
```
calebcartwright: We import and rexport a lot of `core` in the `services/index.js` so that they can be easily imported in all the various service modules, and to simplify refactorings when we shuffle/rename things in `core`
```js
const { InvalidParameter, NotFound } = require('..')
```

17cupsofcoffee: When I saw that in your last suggestion I thought the `..` was just you leaving a placeholder 😆 Fixed!

17cupsofcoffee:
No problem, thank you for the helpful review comments 😄

Issue #658, opened 2016-02-25
Bintray badge is flaky


agemooij:
The Bintray badge seems to be broken; all versions are rendered as a grey "invalid".

Perhaps the API has changed? I've tested with multiple repositories but the example on shields.io itself is also broken:

[ ![Current Version](https://img.shields.io/bintray/v/asciidoctor/maven/asciidoctorj.svg) ](https://bintray.com/asciidoctor/maven/asciidoctorj/_latestVersion)

migmruiz:
Having the same issue here

xvik:
It works sometimes, but mostly not. Could it be caused by [bintray api limits](https://bintray.com/docs/api/#_limits)?

nwalters512:
Can confirm that this is still an issue. It was working yesterday, but all Bintray badges (at least the one for my project and the example from shields.io) are currently showing "invalid".

raniejade:
Badge is broken for my project as well. Rate limiting is very likely the culprit.

espadrine:
The request performed is `curl -v 'https://bintray.com/api/v1/packages/asciidoctor/maven/asciidoctorj/versions/_latest'`. I cannot find a `X-RateLimit-Limit` header in the response. Is this issue still ongoing? Did they switch their rate limiting system?

cdeszaq:
As of today, the issue is still present. Does anyone have contacts at Bintray to see if we can resolve this?

xvik:
Here is what bintray support answered about this issue:

> Thank you for contacting us.
> We were not able to reproduce this behaviour using a custom shield.
> We also verified that the 'getLatestVersion' API is not rate limited so this does not appear to be the issue.

So there are no limits on bintray side and probably the problem is on the shields side. 

Dear shield maintainers, please look actual error messages in your logs and try to investigate this.

P.S. I also asked bintray support to allow bintray badge customization (different message), but don't know if this ever happen:
> We have forwarded your request to our Product Team and will consider adding the option to customise the badge.

Fi5t:
The badge has broken again...

paulmelnikow:
It looks like the badge has been flaky for a long time.

Unfortunately the server doesn't log all the API calls. So I think the way to resolve this long-standing issue is to set up a recurring test over a week or more, and send documentation of the failures to their support.

If anyone is game, would be great to have your help!

laughedelic:
Still a problem. From the example on shields.io:

[...]

PR #4217, opened 2019-10-18
fix [bintray] badges

calebcartwright:
Refs (perhaps fixes?) #658

### DIFF
--- a/services/bintray/bintray.service.js
+++ b/services/bintray/bintray.service.js
@@ -23,7 +23,11 @@ module.exports = class Bintray extends BaseJsonService {
   }
 
   static get auth() {
-    return { userKey: 'bintray_user', passKey: 'bintray_apikey' }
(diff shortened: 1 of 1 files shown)


calebcartwright:
https://shields-staging-pr-4217.herokuapp.com/bintray/v/asciidoctor/maven/asciidoctorj

calebcartwright:
I'm a little confused now 😕 i was originally able to reproduce the error we have in prod in my local env, and then things started working after the change in this PR. 

However, when I tried reverting my local env back to the current master, it still works. Also, the bade is working fine in staging

https://shields-staging.herokuapp.com/bintray/v/asciidoctor/maven/asciidoctorj

calebcartwright:
I think my initial findings were just a fluke, so this change is not necessary/helpful. Closing accordingly

PR #4218, opened 2019-10-19
[gitlab] Add documentation for publicity state of pipelines

Bl4d3s:
As requested by @paulmelnikow 
If anything should be written in other words etc feel free to tell me I will change it accordingly


Bl4d3s:
I just looked at the staging: Maybe the image should be a lot smaller or completely removed?

calebcartwright:
> I just looked at the staging: Maybe the image should be a lot smaller or completely removed?

Visuals are always helpful, though it is rather large. I think it may also be helpful to highlight that **Public pipelines** checkbox (perhaps with an arrow pointing to, or box around it?)

Bl4d3s:
> > I just looked at the staging: Maybe the image should be a lot smaller or completely removed?
> 
> Visuals are always helpful, though it is rather large. I think it may also be helpful to highlight that **Public pipelines** checkbox (perhaps with an arrow pointing to, or box around it?)

I did revise the picture now its much smaller (I ommited the heading and not necessary parts) and I added a red circle /arrow to mark the important setting

services/gitlab/gitlab-pipeline-status.service.js
```diff
@@ -15,6 +15,25 @@ const queryParamSchema = Joi.object({
   gitlab_url: optionalUrl,
 }).required()
 
+const documentation = `
+<p>
+  Important: If your project is publicly visible, but the badge is like this:
+  <img src="https://img.shields.io/badge/build-not&nbsp;found-red" alt="build not found"/>
+</p>
+<p>
+  Check, if your pipelines are publicly visible as well.<br />
```
calebcartwright:
```suggestion
  Check if your pipelines are publicly visible as well.<br />
```

PR #4222, opened 2019-10-20
fix [nuget] version build metadata handling

### LINKED ISSUE #4219: NuGet badges with build metadata treated as pre-release

:beetle: **Description**

I've been using build metadata in my NuGet package versioning since mid-May (eg. `0.11.0+294`) and that has been accessible via the "/v/" path for NuGet. My last successful version number that appeared under the "/v/" path was `0.16.0+388`.

In my next version, my build information is `0.17.0+1b81349-429` which follows [semver rule #10 for Build Metadata](https://semver.org/#spec-item-10). This build does not appear under "/v/" but under "/vpre/".

I can obviously just use the "/vpre/" URLs but the thing is, that build number isn't any less of a pre-release than `0.16.0+388` is (in that both are major version zero).

If there is a change that major version zero now needs "/vpre/", that's fine. I think that `0.16.0+388` shouldn't appear under "/v/" though.

EDIT: Just to add to the confusion, I have another library (see the second 2 badges) which use the same version number structure as what I've mentioned already but it appears fine under "/v/".

[Should I be blaming caching?](https://shouldiblamecaching.com/)

:link: **Link to the badge**

calebcartwright:
Fixes #4219


services/nuget-fixtures.js
```diff
@@ -34,9 +34,24 @@ const nuGetV3VersionJsonFirstCharNotZero = JSON.stringify({
...
+const nuGetV3VersionJsonBuildMetadataWithDash = JSON.stringify({
+  data: [
+    {
+      totalDownloads: 0,
+      versions: [
+        {
+          version: '1.16.0+388',
+        },
+        { version: '1.17.0+1b81349-429' },
+      ],
```
calebcartwright: Interesting prettier, interesting 👀

chris48s: Prettier takes into account the formatting of the input file - its not only the AST structure that determines the output. I think with objects, prettier only switches you to/from single/multi-line objects if it also violates some other rule. If you change this to either:


```js
versions: [
  {
    version: '1.16.0+388'
  },
  {
    version: '1.17.0+1b81349-429'
  },
],
```

or

```js
versions: [
  { version: '1.16.0+388' },
  { version: '1.17.0+1b81349-429' },
],
```

and then run `prettier`, you won't get the result


```js
versions: [
  {
    version: '1.16.0+388',
  },
  { version: '1.17.0+1b81349-429' },
],
```

services/nuget/nuget-v3-service-family.js
```diff
@@ -184,16 +185,19 @@ function createServiceFamily({
...
-
       let latest = versions.slice(-1).pop()
       const includePrereleases = which === 'vpre'
       if (!includePrereleases) {
-        const filtered = versions.filter(item => !item.version.includes('-'))
+        const filtered = versions.filter(item => {
+          if (semver.valid(item.version)) {
+            return !semver.prerelease(item.version)
+          }
+          return !item.version.includes('-')
```
chris48s: Not clear how likely it is for us to hit this line as I'm not really clear what the constraints are on a NuGet version string, but how about


```js
return !item.version.split('+')[0].includes('-')
```

here? (i.e: we only say its a pre-release if the `-` character appears before a `+` character)

calebcartwright: > I'm not really clear what the constraints are on a NuGet version string

Me either, but my sense is that there aren't really any constraints, or at least that the version formats vary wildly.  I wanted to handle the semver specific use case without risking rocking the boat 😄 

Are you suggesting replacing the entire check with the `return !item.version.split('+')[0].includes('-')`, or just replacing line 195?

Turnerj: If you were curious for details on package versioning for NuGet, there is some documentation here: https://docs.microsoft.com/en-au/nuget/concepts/package-versioning

The tl;dr, it is close to (if not exactly) semver v1 for some packages and semver v2 otherwise.

calebcartwright: thanks @Turnerj! Do you know if that is enforced by all the various registries that support nuget packages though? Or is it just convention/just enforced by some, like nuget.org?

Turnerj: I don't believe it is enforced by all registries. NuGet is definitely the biggest registry though (its the default in Visual Studio), probably followed by MyGet (though likely way behind NuGet) before then counting the more ad-hoc things like CI's providing NuGet feeds of builds.

I can't find any specific details about what MyGet does with their NuGet feed support but from what I gleam from their docs, I think they likely follow closely what NuGet do. Regarding other ad-hoc ones, I doubt there is any specific constraint handling for package versions.

services/nuget-fixtures.js
calebcartwright: That's my own fault for not checking it aftewards. I pasted it in expecting prettier to make them the same, but makes sense that it didn't (especially since it didn't exceed the max line length)

services/nuget/nuget-v3-service-family.js
calebcartwright: I suppose it all depends on whether the `-` before `+` precedence applies in non-semver compatible versions. 

Admittedly a heavily contrived example, but let's say there's a nuget package that contains one or more non-semver compliant versions like:
`1b+abc-123`

We'd currently classify that as a pre-release version, and I don't know if/how non-disruptively we can change how we treat that to say it's not a pre-release because the SemVer spec says that everything after the `+` is build metadata which can include a `-` character without the version being pre-release.

TBH I'm not sure what the correct answer is, and I don't feel strongly one or the other. If we want to say that when determining whether a nuget package version is pre-release we won't include any `-` characters that occur after a `+` character then I'm 👍 with that, and I'll add some inline comments with links to this thread. 

And if we do, then I think we can get by with just using `!item.version.split('+')[0].includes('-')` since we're going to apply that rule whether the version is semver compliant or not (`semver.prerelease` is functionally doing the same thing anyway).

chris48s: Yeah if version can be an arbitrary string, its probably meaningless trying to decide if its a pre-release or not in the fallback case where its not a semver.

Turnerj: That sounds fair at least - if you want to handle pre-releases, use semver for your versioning.

PyvesB: I agree with @chris48s and @Turnerj. However, to keep the badges consistent with how they are currently displayed for users who do not use semantic versioning, I suggest we leave line 195 as it is above, i.e. no changes to this part of the PR. If someone opens an issue for a non semantic versioning case, we can discuss it then and see if we can refine our handling. Does this sound good?

services/nuget-fixtures.js
PyvesB: Useful to know @chris48s ! @calebcartwright could you try fixing the formatting? 😄

calebcartwright:
Thanks @PyvesB!

Issue #4188, opened 2019-10-17
Maven-Metadata renders a version like "1.0" as "1"


loosebazooka:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

For the following metadata:

metadata: http://central.maven.org/maven2/com/google/guava/guava-beta-checker/maven-metadata.xml

The actual badge rendered
[![Maven Central](https://img.shields.io/maven-metadata/v/http/central.maven.org/maven2/com/google/guava/guava-beta-checker/maven-metadata.xml.svg?colorB=007ec6)](https://maven-badges.herokuapp.com/maven-central/com.google.guava/guava-beta-checker)

expected badge
![expected](https://img.shields.io/badge/maven-v1.0-blue.svg)

Presumably some kind of issue with how versions are handled (as strings vs numbers)?

calebcartwright:
Thanks @loosebazooka 

You could use the MavenCentral version badge in the interim which provides your expected format
https://img.shields.io/maven-central/v/com.google.guava/guava-beta-checker?label=maven)
![](https://img.shields.io/maven-central/v/com.google.guava/guava-beta-checker?label=maven)

I recall something similar in the [MavenCentral](https://github.com/badges/shields/blob/master/services/maven-central/maven-central.service.js) service (fixed in #3935), so I imagine the same fix could be applied to the [MavenMetadata](https://github.com/badges/shields/blob/master/services/maven-metadata/maven-metadata.service.js) service to fix this

netsgnut:
Thank you @calebcartwright for the pointer! Let's see if I could prepare a PR for that.

calebcartwright:
Closed via #4224

PR #3935, opened 2019-09-02
[MavenCentral] service should not parse versions as number

reda-alaoui:
I am the maintainer of https://github.com/Cosium/maven-git-code-format.
The current latest version is `1.30`.
Despite this, shield thinks the latest version is `1.3`.

By parsing maven-metadata.xml version node values as number, maven-central service strips the trailing zero from `1.30`, turning it into `1.3`.

This PR disables number parsing for the maven-central service.

Closes #3642

PR #4224, opened 2019-10-22
[MavenMetadata] should use version as string (#4188), also run [MavenCentral]

netsgnut:
This PR proposes a fix to #4188, where version numbers from `maven-metadata.xml` get trimmed erroneously in the `maven-metadata` service.

The fix is directly adapted from #3935 (done to `maven-central` by @reda-alaoui).

### DIFF
--- a/services/maven-metadata/maven-metadata.tester.js
+++ b/services/maven-metadata/maven-metadata.tester.js
@@ -12,6 +12,33 @@ t.create('valid maven-metadata.xml uri')
     message: isVPlusDottedVersionAtLeastOne,
   })
 
+t.create('version ending with zero')
+  .get(
+    '/v.json?metadataUrl=http://central.maven.org/maven2/mocked-group-id/mocked-artifact-id/maven-metadata.xml'
+  )
+  .intercept(nock =>
+    nock('http://central.maven.org/maven2')
+      .get('/mocked-group-id/mocked-artifact-id/maven-metadata.xml')
+      .reply(
+        200,
+        `
+      <metadata>
+        <groupId>mocked-group-id</groupId>
+        <artifactId>mocked-artifact-id</artifactId>
+        <versioning>
+          <latest>1.30</latest>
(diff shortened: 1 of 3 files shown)


services/maven-metadata/maven-metadata.tester.js
```diff
@@ -12,6 +12,33 @@ t.create('valid maven-metadata.xml uri')
...
+          <versions>
+            <version>1.30</version>
+          </versions>
+          <lastUpdated>20190902002617</lastUpdated>
+        </versioning>
+      </metadata>
+      `
+      )
+  )
+  .expectBadge({ label: 'maven', message: /^v1\.30$/ })
```
calebcartwright: Any reason the regex is needed for `message`? Would the below not work?

```suggestion
  .expectBadge({ label: 'maven', message: 'v1.30' })
```

netsgnut: Not at all, apart from keeping the same convention from the source this snippet is adapted from. It should work without the regex, indeed. Shall we remove it then?

(And BTW should we leave the [same line](https://github.com/badges/shields/blob/54bcedc0f42a211e8b802001e28af27da024a621/services/maven-central/maven-central.tester.js#L53) at `maven-central` as-is in this PR? If you like this PR to change that line for `maven-central.test.js`, we can include it as well.)

calebcartwright: Sure that sounds good

netsgnut: Updated via a69a796 and da735da.

calebcartwright:
When you get a chance can you rebase or merge in the lastest from master?

netsgnut:
@calebcartwright Right away! Should have done so when I pushed the new commit :(

calebcartwright:
> @calebcartwright Right away! Should have done so when I pushed the new commit :(

No worries, thanks for the quick fix!

PR #4228, opened 2019-10-24
add [homebrew] cask service

### LINKED ISSUE #1638: Badge request: Homebrew Cask
[Homebrew Cask](https://caskroom.github.io/) is now officially part of Homebrew. Now that we already have a badge for Homebrew, it makes sense to also include Homebrew Cask.

pnowaczyk:
I went for the '/' route to distinguish regular and cask formulas, much like many other services

both regular and cask services are similar in implementation, but different schemas and api routes make it quite hard to find any common code to refactor, so I decided to leave it as it

this closes #1638


calebcartwright:
That's odd, I've never seen the Docker Cloud tests fail before. I've kicked off a re-try

paulmelnikow:
Huh, odd. For the record it doesn’t actually run any tests, just builds an image.

Issue #2435, opened 2018-11-30
Packagists Version numbers not being updated


danmcadams:
Are you experiencing an issue with...
 - [X] [shields.io](https://shields.io/#/)
 - [ ] My own instance
 - [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**
The version number on my packagist shield is not being updated, it is however the most recent version when you go to the page on packagist.  I checked the other issues here and they don't seem to line up with my issue. My versions are tagged exactly the same, and I have no version listed in composer.

Current version is v0.1.0 and my shield shows v0.0.1

The repository in question is here: https://github.com/The-Road-Bunch/command-menu

:link: **Link to the badge**
https://img.shields.io/packagist/v/theroadbunch/command-menu.svg

:bulb: **Possible Solution**

paulmelnikow:
Hi! Thanks for the report.

Confirmed. It looks like there is some lag. The new version is not reflected in the endpoint we're hitting:

https://packagist.org/packages/theroadbunch/command-menu.json

Based on the doc here https://packagist.org/apidoc#get-package-data it's cached for 12 hours, which explains the issue.

Maybe the version badge could be modified to use the "always up to date" endpoint those docs mention, instead:

http://repo.packagist.org/p/theroadbunch/command-menu.json

paulmelnikow:
@chris48s Do you know any reason (other than historical) that we might be using the dynamic endpoint for the version?

chris48s:
No. Some of the packagist badges use `/p` and some use `/packages` but I don't think there is any solid reason for that (although note that the format of those 2 endpoints is _slightly_ different so we'd need to modify the parsing code a little bit to change).

platan:
I think we can switch to "Composer metadata" (https://packagist.org/apidoc#get-package-data). I will be beneficial for Shields and for Packagist. About Composer metadata:
> This is the preferred way to access the data as it is always up to date, and dumped to static files so it is very efficient on our end.

calebcartwright:
Sounds like a win-win

PyvesB:
I believe we're using the composer API nowadays, so we no longer have this 12-hour cache issue. Will go ahead and close the issue.

PR #4231, opened 2019-10-24
#2435 [packagist] composer metadata api

piekar294:
Hi, this PR implements improvements proposed in #2435


services/packagist/packagist-base.js
```diff
@@ -3,33 +3,62 @@
...
-  }).required(),
+  packages: Joi.object()
+    .pattern(/^/, packageSchema)
+    .required(),
 }).required()
-
 const keywords = ['PHP']
 
 class BasePackagistService extends BaseJsonService {
   async fetch({ user, repo, schema, server = 'https://packagist.org' }) {
```
calebcartwright: I _think_ I've got the gist of the difference between the JSON API and the Composer Metadata: TL;DR - JSON API has everything but is subject to 12 hour latency wheras Composer Metadata is always up to date but only has metadata about the package (no registry stats, etc.).

The docs you added to `fetchByJsonAPI` were very helpful (thanks!), but I still found myself bouncing between the two. Could we add docs to both functions to speak to their respective use cases? That way if a dev is working on one of the Packagist services they can get the doc for either function vs. having to always go to `fetchByJsonAPI`

services/packagist/packagist-downloads.service.js
```diff
@@ -101,7 +101,7 @@ module.exports = class PackagistDownloads extends BasePackagistService {
   async handle({ interval, user, repo }, { server }) {
     const {
       package: { downloads },
-    } = await this.fetch({ user, repo, schema, server })
+    } = await this.fetchByJsonAPI({ user, repo, schema, server })
```
calebcartwright: It may also be helpful if we include a note in the documentation for the badges like this one that are using the JSON API that the data may be delayed by 12 hours.

services/packagist/packagist-base.js
piekar294: I thinks it's a very good idea.
Missing documentation is added :)

services/packagist/packagist-downloads.service.js
piekar294: Did you refer to https://shields.io/category/downloads when you mentioned documentation?

I am not quite sure about the correct way to present information about possible delay.
Right now, badges are described just with short name with optional data within brackets.
Maybe `cached` keyword could be introduced (example attached).
This way, label would remain compact.
On the other hand, I am not convinced if this `cache` keyword would be understandable by users.

Maybe you already have another idea?

Screenshot 2019-10-28 at 18 40 03

services/packagist/packagist-base.js
calebcartwright: Thanks!

services/packagist/packagist-downloads.service.js
calebcartwright: Apologies for the confusion. I was referring to the documentation which appears to users within the modal window after clicking one of the sample listing which shows up above the parameter fields:

image

We've got this on a few badges where it's helpful to convey additional info to users, for example:

Jenkins Coverage
image



Discord
image


This is done by defining the desired documentation text within the service class, and attaching it to the example listing objects using the `documentation` key

https://github.com/badges/shields/blob/7e0976cd8cc65f012ee86f962aad08ce3aa16cb9/services/jenkins/jenkins-coverage.service.js#L82-L91

https://github.com/badges/shields/blob/7e0976cd8cc65f012ee86f962aad08ce3aa16cb9/services/jenkins/jenkins-coverage.service.js#L107-L122

piekar294: Thank you for complex explanation. I've totally missed how documentation property is working.

I've submitted change with additional documentation for json api delay.

calebcartwright:
I'm getting an `invalid response data` badge for the license with this PR

https://shields-staging-pr-4231.herokuapp.com/packagist/l/doctrine/orm


but it's working fine in prod:
https://img.shields.io/packagist/l/doctrine/orm

calebcartwright:
Thanks for the quick fix!

piekar294:
> I'm getting an `invalid response data` badge for the license with this PR
> 
> https://shields-staging-pr-4231.herokuapp.com/packagist/l/doctrine/orm
> 
> 
> but it's working fine in prod:
> https://img.shields.io/packagist/l/doctrine/orm
> 

Thank you for catching this out.
I investigated it further and I surprisingly found out that example from tests was working just fine (`https://img.shields.io/packagist/l/symfony/symfony`).
`npm run badge -- /packagist/l/doctrine/orm` showed problem with schema, so I fix it out.
I haven't found out the reason why schema is not working fine for `doctrine/orm` yet.

I will try to figure out what exactly went wrong. Moreover, I think it would be appropriate to describe this situation in tests, so i would like to add some tests from schema validation :)

calebcartwright:
> I will try to figure out what exactly went wrong. Moreover, I think it would be appropriate to describe this situation in tests, so i would like to add some tests from schema validation :)

That'd be great! Feel free to open a new issue to keep track of that

calebcartwright:
Actually just to clarify, did you mean you want to add some additional packagist tests as part of this PR?

piekar294:
> Actually just to clarify, did you mean you want to add some additional packagist tests as part of this PR?

I thought about adding tests to this PR, but I may open new PR as well.
Probably I wont be able to work on tests before Tuesday/Friday, so
feel free to merge this PR, if you found current changes to be sufficient :)

calebcartwright:
> I thought about adding tests to this PR, but I may open new PR as well.
Probably I wont be able to work on tests before Tuesday/Friday, so
feel free to merge this PR, if you found current changes to be sufficient :)

Thanks for clarifying. I'm going to go ahead and merge, and if you decide to look into adding more tests we can do so in a future PR

PR #4232, opened 2019-10-24
Added [GithubGoModGoVersion GitHub] badge

### LINKED ISSUE #3960: Badge Request: Go (golang) module version from GitHub

hjagodzinski:
Closes #3960.


services/github/github-go-mod.service.js
```diff
@@ -0,0 +1,81 @@
...
+
+module.exports = class GithubGoModGoVersion extends ConditionalGithubAuthV3Service {
+  static get category() {
+    return 'version'
+  }
+
+  static get route() {
+    return {
+      base: 'github/go-mod/go-version',
+      pattern: ':user/:repo/:branch*',
```
calebcartwright: We should also include an optional query parameter folks can use to specify the path to the mod file for cases where it's not in the root of the repo/multiple packages/etc.

https://github.com/badges/shields/issues/3960#issuecomment-546084081

paulmelnikow: Coincidentally I just did something really similar for the package.json dependency badge in #4233! Feel free to refer to that to see how to do this.

services/github/github-common-fetch.js
```diff
@@ -45,10 +45,11 @@ async function fetchRepoContent(
       throw new InvalidResponse({ prettyMessage: 'undecodable content' })
     }
   } else {
-    return serviceInstance._request({
+    const response = await serviceInstance._request({
```
paulmelnikow: Thanks for fixing this bug! I'm a bit surprised this hasn't come up before. I guess this code path is not used. 

One suggestion for consistency with the way most of the shared code is written:

```suggestion
    const { buffer } = await serviceInstance._request({
```

services/github/github-common-fetch.js
```diff
@@ -45,10 +45,11 @@ async function fetchRepoContent(
...
       url: `https://raw.githubusercontent.com/${user}/${repo}/${branch}/${filename}`,
       errorMessages,
     })
+    return response.buffer
```
paulmelnikow:
```suggestion
    return buffer
```

hjagodzinski:
@calebcartwright @paulmelnikow thank you guys for taking time to have a look at my PR. If not today than tomorrow I will address your comments.

services/github/github-go-mod.service.js
hjagodzinski: Similar as in #4233 I have added a query param allowing to specify a filename. I guess accepting only a path would suffice though.

services/github/github-common-fetch.js
hjagodzinski: At first I was a bit surprised as well but going deeper I realized that block of code was actually never called :) `fetchJsonFromRepo` called `fetchRepoContent` only if `serviceInstance.staticAuthConfigured` was evaluated to true.

services/github/github-common-fetch.js
hjagodzinski: Sorry, instead of commiting your suggestion I pushed a new commit 😄

services/github/github-go-mod.service.js
```diff
@@ -0,0 +1,95 @@
...
+          branch: 'master',
+        },
+        staticPreview: this.render({ version: '1.12', branch: 'master' }),
+        documentation,
+      },
+      {
+        title: 'GitHub go.mod Go version (subfolder of monorepo)',
+        pattern: ':user/:repo',
+        namedParams: { user: 'golang', repo: 'go' },
+        queryParams: { filename: 'src/go.mod' },
```
hjagodzinski: I was not able to come up with better example at the moment. IMHO it does the work as `go.mod` is under subfolder.

services/github/github-go-mod.service.js
paulmelnikow: How would that be different?

I thought about calling this parameter `path` but thought `filename` seemed clearer.

hjagodzinski: Instead of requiring a client to pass a path containing each time `go.mod` part, you could allow him to pass only a path to a directory where `go.mod` resides. Provided the filename `go.mod` never changes, this would mean skipping redundant information as it's known upfront (the error messages already assume the filename is `go.mod`). For example instead of `filename=src/go.mod` you would have `path=src`. What do you think?

paulmelnikow: Ahh, I gotcha. I think it's clearer to include `/go.mod` in the parameter. It makes it self-documenting.

hjagodzinski: Oki doki, let's keep it that way.

services/github/github-go-mod.tester.js
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+t.create('Go version (repo not found)')
+  .get('/badges/not-existing-repo.json')
+  .expectBadge({
+    label: 'Go',
+    message: 'repo not found, branch not found, or go.mod missing',
+  })
+
+t.create('Go version (missing Go version in go.mod)')
```
paulmelnikow: Could this be changed to a live test, now that there's a way to specify an alternate filename?

hjagodzinski: It would require a `go.mod` file or in fact any kind of file that exists and and does not contain a line similar to `go 1.14`. Personally, in this case, I would stick to a mock server as it allows control over the returned content. A live test gives you certainty the badge communicates correctly with the server. This test verifies how badge behaves when a file doesn't contain expected line. In my opinion a "traditional" unit test suits here better.

If you guys feel it's better having a live test I can make it by pointing to `shields`'s repo `package.json` for instance.

calebcartwright: I just forked a repo and removed the go version, feel free to use this as the target for a live test

https://github.com/calebcartwright/ci-detective/blob/master/go.mod

hjagodzinski: Thanks. I've changed the test to live one.

services/github/github-go-mod.service.js
calebcartwright: That works just fine!

services/github/github-go-mod.service.js
```diff
@@ -0,0 +1,95 @@
...
+        staticPreview: this.render({ version: '1.14' }),
+        documentation,
+      },
```
calebcartwright: Can we update the example listings to include supporting a custom file path/subfolder with a specific branch too?

services/github/github-go-mod.service.js
```diff
@@ -0,0 +1,95 @@
...
+
+  static get route() {
+    return {
+      base: 'github/go-mod/go-version',
+      pattern: ':user/:repo/:branch*',
+      queryParamSchema,
+    }
+  }
+
+  static get examples() {
```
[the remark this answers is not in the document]
hjagodzinski: Done.

services/github/github-go-mod.service.js
```diff
@@ -0,0 +1,95 @@
...
+      user,
+      repo,
+      branch,
+      filename,
+    })
+    const { go } = this.constructor.parseContent(content)
+    return this.constructor.render({ version: go, branch })
+  }
+
+  static parseContent(content) {
```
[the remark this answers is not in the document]
hjagodzinski: Done.

services/github/github-go-mod.service.js
hjagodzinski: Done.

PR #4233, opened 2019-10-24
Add filename override to [GithubPackageJson]

paulmelnikow:
Refs https://github.com/badges/shields/issues/2259#issuecomment-545485159


services/github/github-package-json.tester.js
```diff
@@ -60,13 +60,24 @@ t.create('Dev dependency version')
     message: semverRange,
   })
 
-t.create('Prod prod dependency version')
```
calebcartwright: lol, good spot

PR #4251, opened 2019-10-26
fix [Sonar] legacy API tests

### LINKED ISSUE #4221: SonarQube service tests failing
~October 18th

:camera: **Live badge**

https://img.shields.io/sonar/coverage/swellaby:letra?server=https%3A%2F%2Fsonarcloud.io

Yes, there's no issue with the badges, just with the target SonarQube instance a subset of the services tests were using

:link: **CircleCI link**
https://circleci.com/gh/badges/daily-tests/672

:beetle: **Stack trace**

```
SonarCoverage [live] Coverage [ GET /org.ow2.petals%3Apetals-se-ase.json?server=http://sonar.petalslink.com ] 
  ValidationError: message mismatch: "value" with value "inaccessible" fails to match the required pattern: /^[1-9][0-9]?%|^100%|^0%$/
```

:bulb: **Possible solution**
For context (especially for those not familiar with Sonar), Sonar is available in two modes: the SaaS/cloud mode (referred to as SonarCloud - https://sonarcloud.io) and the download/self-host mode (referred to as SonarQube - https://www.sonarqube.org/downloads/); note that SonarCloud is the same thing, just as a SaaS offering of Sonar that uses the latest-and-greatest version. The current/latest version of SonarQube is 8.0. 

Several years ago there was a major API contract change in SonarQube with the release of version 5.4. Shields supports providing badges for both legacy and modern Sonar versions: the sonar badges accept an optional query param `sonarVersion` to allow users to indicate they have a "legacy" (pre v5.4) SonarQube instance. If that query param is provided to Shields then Shields will use the legacy API, otherwise Shields will use the "modern" API (Sonar 5.4+).

As such, we've historically had service tests for our Sonar badges that cover both cases. We have service tests that are targeting the SonarCloud instance (latest/modern API) as well as service tests that are targeting a self-hosted SonarQube instance that was the legacy <5.4 version. 

The legacy SonarQube instance used in those service tests is http://sonar.petalslink.com/ which has been offline the last few days thus resulting in our failed service tests. 

One option for us is to wait to see if http://sonar.petalslink.com/ comes back online (I suppose we could also try to get in contact with them, but I'm not really sure how to best do so)

Another option, and the one I'd suggest, is that we remove the subset of our service tests for the Sonar badges that are targeting that legacy SonarQube instance. The main purpose of our service tests is to validate our integrations with external services and to catch upstream changes quickly. However, that's not relevant IMO for SonarQube <= v5.4; it's a package/released version of software from several years ago, not a live service that can be changed by the vendor at will. The API contracts for those older SonarQube instances cannot change. The service tests for the legacy API will only ever fail if the target instance is offline.

The service tests we have that are targeting SonarCloud will cover the integration we have with that service that _can_ change, and we can replace the service tests targeting the legacy-api endpoint with unit tests to validate any legacy-api specific logic (i.e. `transform`). 

Thoughts?

calebcartwright:
Resolves #4221 by mocking the service tests that target the legacy SonarQube APIs


services/sonar/sonar-base.js
```diff
@@ -14,8 +14,9 @@ const modernSchema = Joi.object({
             Joi.number().min(0),
             Joi.allow('OK', 'ERROR')
           ).required(),
-        }).required()
```
calebcartwright: I found a few scenarios where the measures/metrics array in the response can actually be empty, details below

services/sonar/sonar-base.js
```diff
@@ -83,12 +84,22 @@ module.exports = class SonarBase extends BaseJsonService {
     const metrics = {}
 
     if (useLegacyApi) {
-      json[0].msr.forEach(measure => {
+      const [{ msr: measures }] = json
+      if (!measures.length) {
+        throw new NotFound({ prettyMessage: 'metric not found' })
+      }
```
calebcartwright: It turns out that the API (both current and legacy) will return a 404 for an unsupported metric, but it can also return an http 200 with an empty array if the metric is valid, but the target component does not publish the corresponding data (for example if the project is not publishing test results). This also happens sometimes when the metric key was previously valid, but has since been deprecated

calebcartwright:
> Could you add some links to #4221 (comment) to indicate why these tests are mocked? Maybe it's even worth explaining that this should generally not be used as an example for other services

Sure, I'll probably linke directly to https://github.com/badges/shields/issues/4221#issuecomment-546611598 which IMO explains things

services/sonar/sonar-base.js
paulmelnikow: Is it worth adding that in a comment, too?

calebcartwright: 😄 it'd definitely be more useful in the long term to be inline the code instead of an aside in a PR wouldn't it.

updating that now

Issue #4227, opened 2019-10-23
Yesterday's server resource incident


paulmelnikow:
In https://github.com/badges/shields/issues/4223#issuecomment-545127052 @platan wrote:

> There is something strange happening with our servers in the last hours:
> https://metrics.shields.io/d/aESRBSjmz/services?orgId=1&from=now-3h&to=now
> https://metrics.shields.io/d/g_1B7zhik/prom-client-default-metrics?orgId=1&refresh=10s&from=now-3h&to=now

See also some discussion in the #ops channel.

I rolled back to 8a21c7746fc131bd8f2c621c331bc357c61e3b69 and so far things have been stable. It's not 100% clear if the rollback fixed the problem, or if one of these commits leading up to 54bcedc0f42a211e8b802001e28af27da024a621 caused it. We should try to re-deploy those commits and/or determine which one is causing the resource leak.

I'm a little busy in the next couple days. Wishing someone could help with this!

paulmelnikow:
I'm curious, has anyone been tracking whether the problem has gone away since the rollback. The badges in the Shields repo seems okay, though there have been some more reports in #4223.

calebcartwright:
I was keeping tabs on the stats yesterday and didn't notice anything in particular. Haven't been able to do much today though

PR #4253, opened 2019-10-26
revert cors header injection #4171

calebcartwright:
Refs #4227 #4173 #4252

This is one of the two code changes that touched a lot of services which were part of the last deployment that preceeded the resource spikes/slow response times on the badge servers. 

Though we still don't know whether _any_ of the commits in that deployment played a role, opening this in case we want to try a deployment without this change to test/validate

cc @paulmelnikow

### DIFF
--- a/core/server/server.js
+++ b/core/server/server.js
@@ -323,12 +323,6 @@ class Server {
     const { apiProvider: githubApiProvider } = this.githubConstellation
     suggest.setRoutes(allowedOrigin, githubApiProvider, camp)
 
-    // https://github.com/badges/shields/issues/3273
-    camp.handle((req, res, next) => {
-      res.setHeader('Access-Control-Allow-Origin', '*')
-      next()
-    })
-
     this.registerErrorHandlers()
     this.registerRedirects()
     this.registerServices()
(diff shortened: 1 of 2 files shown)


paulmelnikow:
Seems worth a shot!

calebcartwright:
As far as trying to test whether these changes were the cause of the issuese we saw.. what would be the most convenient approach for you on the deployment side? 

I was thinking I'd go ahead and merge them both now, and then whenever the next deployment goes in, we can monitor the servers/response times for a while afterwards. If everything looks okay then we could re-deploy the original changes in isolation (I can open up PR's that restore these changes so that they are ready to go when we're ready) to see whether or not they are problematic.

Thoughts?

paulmelnikow:
> I was thinking I'd go ahead and merge them both now, and then whenever the next deployment goes in, we can monitor the servers/response times for a while afterwards. If everything looks okay then we could re-deploy the original changes in isolation (I can open up PR's that restore these changes so that they are ready to go when we're ready) to see whether or not they are problematic.

That sounds good. I can do a deploy today or tomorrow.

PR #4265, opened 2019-10-28
Add [GithubLerna] badge

### LINKED ISSUE #2413: Badge request: Lerna
:clipboard: **Description**

- Which service is this badge? Lerna

- What sort of information should this badge show? `lerna | v1.2.3` or `lerna | independent`

:link: **Data**

- Is there a public API?

No need for public API, it's inside the `version` key in `lerna.json`,  in the root folder.

Example: https://github.com/babel/babel/blob/master/lerna.json

pedrofracassi:
- [x] Add Lerna badge
- [x] Make it work when the version is `independent`
- [x] Write tests

Closes #2413


services/github/github-lerna.service.js
```diff
@@ -0,0 +1,66 @@
...
+        documentation,
+      },
+    ]
+  }
+
+  static render({ version, branch }) {
+    return renderVersionBadge({
+      version,
+      tag: branch,
+      defaultLabel: 'lerna',
```
calebcartwright: It'd be better to extract the label out into the `defaultBadgeData` getter function. This way the label will be set even in the event of errors, such as 404/not found scenarios

```js
    return {
      label: 'lerna'
    }
  }
```

pedrofracassi:
What's the best way to display `independent` instead of the version when needed?

calebcartwright:
> What's the best way to display `independent` instead of the version when needed?

I'm not familiar with lerna, but what if the render function was something like the below?

```js
  static render({ version, branch }) {
    if (version === 'independent') {
      return {
        label: branch ? `lerna@${branch}` : undefined,
        message: version,
      }
    }

    return renderVersionBadge({
      version,
      tag: branch,
      defaultLabel: 'lerna',
    })
  }
```

pedrofracassi:
That looks good

PyvesB:
@pedrofracassi we're really close to getting this merged! It would be great if you could implement @calebcartwright's suggestion above. 😉

PyvesB:
I'm keen on getting this landed, so I'll pick the PR up, add some tests and implement @calebcartwright if that's okay with you @pedrofracassi?

pedrofracassi:
@PyvesB yes, please. Some personal stuff got in the way and I wasn't able to get it done.

services/github/github-lerna.service.js
PyvesB: Actually there's a specific reason why it's done like this. If you extracted to `defaultBadgeData`, badges that don't specify a tag (i.e. branch in this case) would become `lerna | v1.0.0`. If you just specify as `defaultLabel`, they remain `version | v1.0.0`, and only tagged ones will contain the Lerna keyword, e.g.  `lerna@master | v1.0.0`.

Not sure whether it's a good reason though, but it's consistent with the behaviour in our _manifest.json_ and _package.json_ badges.

PyvesB:
> > What's the best way to display `independent` instead of the version when needed?
> 
> I'm not familiar with lerna, but what if the render function was something like the below?
> 
> ```js
>   static render({ version, branch }) {
>     if (version === 'independent') {
>       return {
>         label: branch ? `lerna@${branch}` : undefined,
>         message: version,
>       }
>     }
> 
>     return renderVersionBadge({
>       version,
>       tag: branch,
>       defaultLabel: 'lerna',
>     })
>   }
> ```

Actually this is not needed, things are handled automatically in `renderVersionBadge`. However, a change in the Joi schema was required to accept `independent`.

PyvesB:
Okay, quite a few changes in the end:
* added a test suite.
* added support for `independent` version.
* modified the base path `github/lerna` -> `github/lerna-json/v`. This is consistent with our _package.json_ and _manifest.json_ badges, and will allow us to easily add non-version Lerna badges in the future.
* renamed the file and class to follow the same structure as _package.json_.

@calebcartwright could you please review again when you have a minute? 😄

services/github/github-lerna-json.service.js
```diff
@@ -0,0 +1,66 @@
...
+    })
+  }
+
+  async handle({ user, repo, branch }) {
+    const { version } = await fetchJsonFromRepo(this, {
+      schema: versionSchema,
+      user,
+      repo,
+      branch,
+      filename: 'lerna.json',
```
calebcartwright: Do we know if the `lerna.json` file must always be in the root of the repo?

PyvesB: To be completely honest with you: no idea. I assumed so with the few examples I saw.

calebcartwright: We can probably go with this as-is, and if it's indeed possible and our users request it, we can add an optional query param to let users specify the repo path to the file

calebcartwright: Will give this a final look-over in a bit and then approve 👍

services/github/github-lerna.service.js
calebcartwright: > Not sure whether it's a good reason though, but at least it's consistent with the behaviour in our manifest.json and package.json badges.

True for manifest.json, but isn't the label for package.json always `version` even when the user specifies a branch? And for another flavor, the go.mod version badges from GH repos are `go | v1.0.0` and `go@master | v1.0.0`. 🤷‍♂ 

I'm not sure which of those patterns the lerna badge should follow 🤔 Should it mimic the package.json pattern or do we want to have `lerna` explicitly in the label?

PyvesB: Changed. Shall we do something about our _manifest.json_ and _package.json_ badges in a separate issue? I've got a feeling _package_ and _manifest_ are very overloaded terms, so I would suggest simply sticking with `version` everywhere for those. Or should we just ignore the inconsistency?

calebcartwright: > Shall we do something about our manifest.json and package.json badges in a separate issue? I've got a feeling package and manifest are very overloaded terms, so I would suggest simply sticking with version everywhere for those. Or should we just ignore the inconsistency?

I think it's worth reviewing and potentially updating yes. `version` makes sense to me for the package.json, but I don't even remember offhand what `manifest` is used for

PR #4266, opened 2019-10-28
When serviceData fails validation, include the service class in the stack trace

paulmelnikow:
This should make it easier to debug #3784.

### DIFF
--- a/core/base-service/base.spec.js
+++ b/core/base-service/base.spec.js
@@ -192,7 +192,7 @@ describe('BaseService', function() {
       })
     })
 
-    it('Throws a validation error on invalid data', async function() {
+    context('On invalid data', function() {
       class ThrowingService extends DummyService {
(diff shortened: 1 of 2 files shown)


core/base-service/base.spec.js
```diff
@@ -192,27 +192,45 @@ describe('BaseService', function() {
...
+          expect(e.name).to.equal('ValidationError')
+          expect(e.details.map(({ message }) => message)).to.deep.equal([
+            '"message" is required',
+          ])
+        }
+      })
+
+      // Ensure debuggabillity.
+      // https://github.com/badges/shields/issues/3784
+      it('Includes the service class is present in the stack trace', async function() {
```
paulmelnikow:
```suggestion
      it('Includes the service class in the stack trace', async function() {
```

PR #4267, opened 2019-10-30
Tutorial: Update the method order to be canonical

paulmelnikow:
This makes the explanation read slightly out-of-order relative to the code, though it means if people are copy and pasting they won't end up with lint errors. It seems like a net win.


calebcartwright:
> This makes the explanation read slightly out-of-order relative to the code

We could always update the numbers and explanation ordering later on to line them back up if it becomes problematic right?

paulmelnikow:
Yea, that’s true. Though I suspect explaining the code in the order it’s calked is more helpful, even if our method order puts the dependencies before the dependants. Ideally I suppose we could rework the writing to explain that we jump to the bottom and then work our way back up.

PR #4271, opened 2019-10-31
fix [PackagistLicense] type error

### LINKED ISSUE #4268: Packagist: TypeError: Cannot read property 'license' of undefined
Sentry Issue: [SHIELDS-4G](https://sentry.io/organizations/shields/issues/1305757780/?referrer=github_integration)

```
TypeError: Cannot read property 'license' of undefined
  File "/home/m/shields/services/packagist/packagist-license.service.js", line 72, in PackagistLicense.transform
    json.packages[this.getPackageName(user, repo)]['dev-master'].license,
  File "/home/m/shields/services/packagist/packagist-license.service.js", line 78, in PackagistLicense.handle
    const { license } = this.transform({ json, user, repo })
  ?, in null.<anonymous>
```

calebcartwright:
Fixes #4268 

I deleted the empty .spec file for the Version service as well


services/packagist/packagist-license.service.js
```diff
@@ -67,10 +68,16 @@ module.exports = class PackagistLicense extends BasePackagistService {
   }
 
   transform({ json, user, repo }) {
-    return {
-      license:
-        json.packages[this.getPackageName(user, repo)]['dev-master'].license,
+    const {
+      packages: {
+        [this.getPackageName(user, repo)]: { 'dev-master': branch },
```
paulmelnikow: This is some extreme destructuring! 😆

Would it be clearer, I wonder, the old-fashioned way?

```js
const packageName = this.getPackageName(user, repo)
const branch = json.packages[packageName]['dev-master']
```

calebcartwright: Lol, never thought I'd hear you suggest _not_ destructuring 😆 

I don't really have a preference, though I agree the old-fashioned way here is a bit easier on the eyes (although I don't think this is the only instance in our codebase where that's true)

paulmelnikow: > Lol, never thought I'd hear you suggest _not_ destructuring 😆

I know, right? 😝 

Well I'm nothing if not nuanced. Is there a Goldilocks emoji?

> although I don't think this is the only instance in our codebase where that's true

It'd be interesting to take a pass through, thoroughly desensitized from exposure to this codebase, and see what now seems better and worse. I've never really tried to codify it, or express it in terms of linter rules. Could be interesting.

calebcartwright: I can't remember specific instances off hand, but I know there's been a few others I wrote that looked a bit gnarly. I was under the impression (mistakenly it seems) that we wanted to use destructuring everywhere, so this discussion has been helpful for me.

Sounds like we want to prefer destructuring, but not be dogmatic about it. If the code is more readible the old-fashioned way, then that'd be the general preference, correct? 

Would also be good to include that guidance in the Contributing/Tutorial docs once we finalize it.

paulmelnikow: > Sounds like we want to prefer destructuring, but not be dogmatic about it. If the code is more readible the old-fashioned way, then that'd be the general preference, correct?

I think ideally we'd express it in terms of rules or a lint rule, though I think if we say that and give a few examples, that's fine.

calebcartwright: "which looks better" would take an awesome lint rule wouldn't it? maybe someone's already solved it with an eslint plugin though

PR #4272, opened 2019-10-31
Refactor JSONPath based services, run [DynamicJson DynamicYaml]

platan:
[Dynamic YAML service](https://github.com/badges/shields/blob/master/services/dynamic/dynamic-yaml.service.js) and [dynamic JSON service](https://github.com/badges/shields/blob/master/services/dynamic/dynamic-json.service.js) allows to query using JSONPath. Both have a lot of common code. 
An both have similar bugs: 
- https://github.com/badges/shields/search?q=%22obj+needs+to+be+an+object%22&type=Issues
- https://github.com/badges/shields/search?q=Lexical+error+on+line+1.+Unrecognized+text.&type=Issues

This PR extracts a piece of code which allows to build a JSONPath based service.  
This PR fixes https://github.com/badges/shields/issues/3788 and fixes https://github.com/badges/shields/issues/4261


services/dynamic/json-path.js
```diff
@@ -0,0 +1,86 @@
...
+const { renderDynamicBadge, errorMessages } = require('../dynamic-common')
+const { InvalidParameter, InvalidResponse } = require('..')
+
+/**
+ * Dynamic service class factory which wraps {@link module:core/base-service/base~BaseService} with support of {@link https://jsonpath.com/|JSONPath}.
+ *
+ * @param {Function} superclass class to extend
+ * @returns {Function} wrapped class
+ */
+module.exports = superclass =>
```
platan: http://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/ is a nice resource about "Subclass Factories".

services/dynamic/json-path.js
```diff
@@ -0,0 +1,86 @@
...
+  class extends superclass {
+    static get category() {
```
platan: Alternatively we can move `category()` and `defaultBadgeData()` to DynamicJson/DynamicYaml. Then this factory become more universal.

services/dynamic/json-path.spec.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const chai = require('chai')
+const { expect } = chai
+const jsonPath = require('./json-path')
+
+chai.use(require('chai-as-promised'))
```
platan: `chai-as-promised` was removed in https://github.com/badges/shields/pull/2359, but it's great for testing rejected promises. I migrated several other tests to `chai-as-promised`. But not all of them, because `rejectedWith` can verify an error class and a message only.

core/base-service/base.spec.js
```diff
@@ -101,27 +103,21 @@ describe('BaseService', function() {
       )
     })
 
-    it('Should throw if route is not overridden', async function() {
-      try {
-        await BaseService.invoke({}, {}, {})
```
calebcartwright: Though I personally prefer using `async/await`, it's more of a stylistic opinion and not anything i feel strongly about

services/dynamic/dynamic-json.service.js
```diff
@@ -1,62 +1,19 @@
...
-const { BaseJsonService, InvalidParameter, InvalidResponse } = require('..')
-
-module.exports = class DynamicJson extends BaseJsonService {
-  static get category() {
-    return 'dynamic'
-  }
+const jsonPath = require('./json-path')
+const { BaseJsonService } = require('..')
 
+module.exports = class DynamicJson extends jsonPath(BaseJsonService) {
```
calebcartwright: Admittedly just another personal stylistic preference, but this feels like a case where favoring composition over inheritance might be a bit more readable/grokable

services/dynamic/json-path.js
```diff
@@ -0,0 +1,86 @@
...
+     * @param {object} attrs Refer to individual attrs
+     * @param {Joi} attrs.schema Joi schema to validate the response transformed to JSON
+     * @param {string} attrs.url URL to request
+     * @param {object} [attrs.errorMessages={}] Key-value map of status codes
+     *    and custom error messages e.g: `{ 404: 'package not found' }`.
+     *    This can be used to extend or override the
+     *    [default](https://github.com/badges/shields/blob/master/services/dynamic-common.js#L8)
+     * @returns {object} Parsed response
+     */
+    async _getData({ schema, url, errorMessages }) {
```
calebcartwright: Could we use the `fetch` naming convention here?

platan: What exactly do you mean? Would you like to change `_getData` to `fetch`? Should I change parameters?

calebcartwright: Sorry I should have been more clear, but yes I was suggesting changing the function name from `_getData` to `fetch`. I know `_getData` is consistent with what was being used before in the json/yaml service classes, but I feel like we use the function name `fetch` in other service classes for this responsibility and I think it'd be good to do so in these dynamic badge services for consistency

core/base-service/base.spec.js
```diff
@@ -101,27 +103,21 @@ describe('BaseService', function() {
...
-        expect.fail('Expected to throw')
-      } catch (e) {
-        expect(e.message).to.equal('Route not defined for BaseService')
-      }
+    it('Should throw if route is not overridden', function() {
+      return expect(BaseService.invoke({}, {}, {})).to.be.rejectedWith(
+        'Route not defined for BaseService'
```
calebcartwright: Should this use a regex instead? It seems to be doing partial string/contains validation (if you change the string message to just `BaseService` it will still pass)
```suggestion
        /^Route not defined for BaseService$/
```

platan: Good spot! I wasn't aware of this behavior. I fixed 3 other places using `throw()` or `rejectedWith()`.

services/dynamic/json-path.js
platan: Thanks for clarification. Done.

platan:
@calebcartwright Thanks for the review. I agree with you that favoring composition over inheritance might be good approach. But I have limited time recently I couldn't propose a better version.

calebcartwright:
> @calebcartwright Thanks for the review. I agree with you that favoring composition over inheritance might be good approach. But I have limited time recently I couldn't propose a better version.

No worries!

Issue #4178, opened 2019-10-14
Move remaining hostnames from route params to query params


calebcartwright:
A follow up to #3714 with the last services that still have url/server/host parameters specified as route params instead of query params:

* [ ] [Jenkins (build, coverage, and tests)](https://github.com/badges/shields/tree/master/services/jenkins)
* [ ] ~~[Mozilla Observatory](https://github.com/badges/shields/blob/master/services/mozilla-observatory/mozilla-observatory.service.js#L61)~~
* [x] [Swagger](https://github.com/badges/shields/blob/master/services/swagger/swagger.service.js)

I _think_ that's all of them but if anyone sees any others please update this list/leave a comment

paulmelnikow:
Thanks for opening this.

For what it’s worth I don’t mind it on the Observatory badge as it’s just a hostname, and thus fits nicely into the URL. Though if you prefer to make them all query params I could get behind that.

calebcartwright:
> For what it’s worth I don’t mind it on the Observatory badge as it’s just a hostname, and thus fits nicely into the URL.

Ah okay, I didn't actually look that closely 😄 

> Though if you prefer to make them all query params I could get behind that.

WFM

PyvesB:
Will pick up the Jenkins badges.

PR #4294, opened 2019-11-02
[Swagger] Specs are valid, when only warnings are shown

Bl4d3s:
If the official example  from [here](https://petstore3.swagger.io/api/v3/openapi.json) is used against the swagger Validator it generates a valid badge, but does also generate some warning level errors. (See [valid badge](http://validator.swagger.io/validator?url=https://petstore3.swagger.io/api/v3/openapi.json) and [debug messages](http://validator.swagger.io/validator/debug?url=https://petstore3.swagger.io/api/v3/openapi.json))

I adjusted the parsing of the endjust endpoint so warnings don't generate an invalid swagger Spec.

I also added a test for this scenario.


Bl4d3s:
Unfortunatley I don't know why some tests are now failing, if someone can point me in the right direction, I'll gladly fix it. (with npm test locally all went well)

services/swagger/swagger.service.js
```diff
@@ -61,15 +61,24 @@ module.exports = class SwaggerValidatorService extends BaseJsonService {
       },
     })
   }
-
   async handle({ scheme, url }) {
     const json = await this.fetch({ scheme, urlF: url })
     const valMessages = json.schemaValidationMessages
 
-    if (!valMessages || valMessages.length === 0) {
```
calebcartwright: Thanks for this! While you're updating this service, would you mind also extracting this logic (and the helper function) into a `transform` function that returns valid/invalid, and also move the color determination to the `render` function?

Bl4d3s: Will do, for clarification: the helper method should also be inside the transform or can it stay outside?

calebcartwright: I think it can probably all go inside a single transform function

calebcartwright:
> Unfortunatley I don't know why some tests are now failing, if someone can point me in the right direction, I'll gladly fix it. (with npm test locally all went well)
> 
> Edit: Now they have passed ?

Sorry I forgot to mention it earlier, but the original problem was the title of the PR (which started with `[Service]`. After I changed it to put the brackets around `Swagger` the tests passed. The way the [CI process works in Shields for PRs](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#tests) is that the service tests for services listed inisde `[]` are executed. Since you had `[Service]` in the PR it was trying to run the tests for a service named `Service` which does not exist

services/swagger/swagger.service.js
```diff
@@ -62,14 +66,22 @@ module.exports = class SwaggerValidatorService extends BaseJsonService {
     })
   }
 
+  transform(valMessages) {
+    if (
+      !valMessages ||
+      valMessages.length === 0 ||
+      valMessages.every(msg => msg.level === 'warning')
```
calebcartwright: Do we know what the full scope of potential `level` options are here? If it's something binary like, `warning` and `error` then we could probably use `.any(..)` instead. If we aren't sure what the potential options are, then this is probably the best approach 👍

Bl4d3s: Taken from the [github site of the validator](https://github.com/swagger-api/validator-badge/blob/master/src/main/swagger/swagger.yaml)

 SchemaValidationError:
      type: object
      properties:
        level:
          type: string
        domain:
          type: string
        keyword:
          type: string
        message:
          type: string
        schema:
          $ref: '#/components/schemas/Schema'
        instance:
          $ref: '#/components/schemas/Instance'
        required:
          type: array
          items:
            type: string
        missing:
          type: array
          items:
            type: string

calebcartwright:
If I'm following correctly, the objective was to make the resulting badge consistent with the Swagger validator by handling warning-only level "'errors". 

As noted in the PR description, using the sample https://petstore3.swagger.io/api/v3/openapi.json returns valid with the Swagger validator, but invalid for Shields

http://validator.swagger.io/validator?url=https://petstore3.swagger.io/api/v3/openapi.json


Prod Shields
https://img.shields.io/swagger/valid/2.0/https/petstore3.swagger.io/api/v3/openapi


However, AFACIT the staging app for this PR is still saying the sample is invalid
https://shields-staging-pr-4294.herokuapp.com/swagger/valid/2.0/https/petstore3.swagger.io/api/v3/openapi.json


Did I get any of that wrong? If not, does that mean the original problem still exists?

Bl4d3s:
> If I'm following correctly, the objective was to make the resulting badge consistent with the Swagger validator by handling warning-only level "'errors".
> 
> As noted in the PR description, using the sample https://petstore3.swagger.io/api/v3/openapi.json returns valid with the Swagger validator, but invalid for Shields
> 
> http://validator.swagger.io/validator?url=https://petstore3.swagger.io/api/v3/openapi.json
> 
> 
> Prod Shields
> https://img.shields.io/swagger/valid/2.0/https/petstore3.swagger.io/api/v3/openapi
> 
> 
> However, AFACIT the staging app for this PR is still saying the sample is invalid
> https://shields-staging-pr-4294.herokuapp.com/swagger/valid/2.0/https/petstore3.swagger.io/api/v3/openapi.json
> 
> 
> Did I get any of that wrong? If not, does that mean the original problem still exists?

You are right, I tested with my own specification, located [here](https://rest.cloud-in.space/swaggerDefinitions/empire.yaml), where it works. Ill look into it and update the PR

calebcartwright:
I wonder if it's an issue with the file extension (when it is a json based spec) getting sucked in and dropped from being passed along to the validator service. 

TBH we should probably move the spec location param from a route param to a query param, refs #3714 and #4178

Bl4d3s:
> I wonder if it's an issue with the file extension (when it is a json based spec) getting sucked in and dropped from being passed along to the validator service.
> 
> TBH we should probably move the spec location param from a route param to a query param, refs #3714 and #4178

It most definitly is.
 See attached log of badge debug output

> npm run badge -- swagger/valid/2.0/https/petstore3.swagger.io/api/v3/openapi.json          

> shields.io@0.0.0 badge /home/michael/GitRepos/shields
> cross-env NODE_CONFIG_ENV=test TRACE_SERVICES=true node scripts/badge-cli.js >"swagger/valid/2.0/https/petstore3.swagger.io/api/v3/openapi.json"

>1103204445 FsTokenPersistence configured with /home/michael/GitRepos/shields/private/github->user-tokens.json
1103204445 Server is starting up: http://localhost:1111/
 Service class  👩‍🍳 
 SwaggerValidatorService
 Named params  🎫 
 { scheme: 'https', url: 'petstore3.swagger.io/api/v3/openapi' }
 Query params  🖍 
 {
  color: undefined,
  colorA: undefined,
  colorB: undefined,
  label: undefined,
  labelColor: undefined,
  link: undefined,
  logo: undefined,
  logoColor: undefined,
  logoPosition: undefined,
  logoWidth: undefined,
  style: undefined
}
 Request  🏹 
 http://validator.swagger.io/validator/debug
 Response status code  🎯 
 200
 Response JSON (before validation)  🎯
{
  schemaValidationMessages: [
    {
      level: 'error',
      message: "Can't read from file https://petstore3.swagger.io/api/v3/openapi"
    }
  ]
}
 Response after validation  🛁
{
  schemaValidationMessages: [
    {
      level: 'error',
      message: "Can't read from file https://petstore3.swagger.io/api/v3/openapi"
    }
  ]
}
 Service data  🛡 
 { message: 'invalid', color: 'red' }
 Rendered badge  🛡 
 {
  label: 'swagger',
  message: 'invalid',
  color: 'red',
  link: [],
  name: 'swagger',
  value: 'invalid'
}

Bl4d3s:
I'm not entierly sure if this is the correct approach, as it will be a breaking change. Is this how you intended it or how should I do it? Locally now both the json and yaml spec show a valid spec, the tests Im currently fixing

services/swagger/swagger.service.js
```diff
@@ -22,20 +22,21 @@ module.exports = class SwaggerValidatorService extends BaseJsonService {
   static get route() {
     return {
       base: 'swagger/valid/2.0',
-      pattern: ':scheme(http|https)?/:url*',
+      pattern: ':scheme(http|https)?/:fileExtension(json|yaml)?/:url*',
```
calebcartwright: We should move the spec url path from route parameters to a single query param, with a redirector to maintain backwards compatibility for existing badges. 

The Website service is probably a decent reference, https://github.com/badges/shields/pull/4028/files

https://github.com/badges/shields/blob/7f54e09fe542578f8d4488aab9d020b54d0f0e8c/services/website/website.service.js#L41-L47
https://github.com/badges/shields/blob/master/services/website/website-redirect.service.js

calebcartwright: Actually the TwitterUrl badge may be a simpler/better reference, as neither it nor this Swagger service had the same complexity the Website badge did 

https://github.com/badges/shields/blob/master/services/twitter/twitter.service.js
https://github.com/badges/shields/blob/master/services/twitter/twitter-redirect.service.js

Bl4d3s: I managed to move it to a query param, will look into the redirector in the following week

Bl4d3s: So I wrote the redirector, and theoretically its working, but I've got one problem
`module.exports = [
  redirector({
    category: 'other',
    name: 'SwaggerUrlRedirect',
    route: {
      base: 'swagger/valid/2.0',
      pattern: ':scheme(http|https)?/:url*',
    },
    transformPath: () => `/swagger/valid/3.0/spec`,
    transformQueryParams: ({ scheme, url }) => ({
      url: `${scheme}://${url}`
    }),
    dateAdded: new Date('2019-11-03'),
  }),
]`
This
`http://localhost:8080/swagger/valid/2.0/https/raw.githubusercontent.com%2FOAI%2FOpenAPI-Specification%2Fmaster%2Fexamples%2Fv2.0%2Fjson%2Fp.etstore-expanded.json`
will get
`http://localhost:8080/swagger/valid/3.0/spec.json?url=https%3A%2F%2Fraw.githubusercontent.com%2FOAI%2FOpenAPI-Specification%2Fmaster%2Fexamples%2Fv2.0%2Fjson%2Fp.etstore-expanded`

So the file extension does get appended on my path instead of begin part of the pattern. Any idea why?

calebcartwright: For the route `pattern`, try this instead:

`pattern: ':scheme(http|https)?/:url+',`

Also, I don't think the `scheme` portion should really be optional (I know it was already, just think we should also change that too given what we are issuing to the upstream API)

https://github.com/badges/shields/blob/7f54e09fe542578f8d4488aab9d020b54d0f0e8c/services/swagger/swagger.service.js#L52-L63

calebcartwright:
It might also be helpful to add a NotFound check here, since presumably, everyone that's tried to get a badge for a json based spec would've received `invalid` regardless of whether the spec was valid. It looks like the Validator API will return an HTTP 200 response with a single error like the one below, and it would be great if we could detect that and convey that in the badge (for example ) instead of `invalid`

```json
{ 
  "schemaValidationMessages":[ 
    { 
      "level":"error",
      "message":"Can't read from file {{spec url}}"
    }
  ]
}
```

We could do that by detecting that condition as part of the `transform` function (single `error` level error with that message) and then throwing a Shields error:
```js
  throw new NotFound({ prettyMessage: 'spec not found' })
```

services/swagger/swagger.service.js
Bl4d3s: Unfortunately the changed pattern does not do any difference. Its just with json, a yaml file ending does get translated correctly

calebcartwright: Did you also remove the `?` from the scheme?

Bl4d3s: Tried with removing and without yes

calebcartwright: Ah okay, would you mind trying `:url(.+)` as well? Sorry I know I've seen this before recently just don't remember off hand what the fix was

Bl4d3s: No change also tried with * and (:url.+) and :url(.+)+

Bl4d3s: Could this be some sort of issue with the badge formats? Supported is svg and json ... and the .json is interpreted as the format of the badge when I see correctly

Bl4d3s:
I did implement the not found feature, and pushed the (currently not working) redirect, maybe you can see the issue better this way

calebcartwright:
> I did implement the not found feature, and pushed the (currently not working) redirect, maybe you can see the issue better this way

Thanks that should help! My understanding is that a swagger spec will always have a .json or a .yml/.yaml extension, do you know if that is correct?

Bl4d3s:
> > I did implement the not found feature, and pushed the (currently not working) redirect, maybe you can see the issue better this way
> 
> Thanks that should help! My understanding is that a swagger spec will always have a .json or a .yml/.yaml extension, do you know if that is correct?

Yes, quote from the [website ](https://swagger.io/docs/specification/about/)

> API specifications can be written in YAML or JSON.

calebcartwright:
Okay I managed to get this working, though with one caveat that IMO we'll need to accept. I'd started making some local changes so the service class diverged just a bit from what you have here. Feel free to incorporate or toss, but something like this should get us to a point of having this solved.

<details><summary>swagger.service.js</summary>

```js
'use strict'

const { optionalUrl } = require('../validators')

const schema = Joi.object({
  schemaValidationMessages: Joi.array().items(
    Joi.object({
      level: Joi.string().required(),
      message: Joi.string().required(),
    }).required()
  ),
}).required()

  specUrl: optionalUrl.required(),
}).required()

module.exports = class SwaggerValidatorService extends BaseJsonService {
  static get category() {
    return 'other'
  }

  static get route() {
    return {
      base: 'swagger/valid',
      pattern: '2.0',
      queryParamSchema,
    }
  }

  static get examples() {
    return [
      {
        title: 'Swagger Validator',
        staticPreview: this.render({ status: 'valid' }),
        namedParams: {},
        queryParams: {
          specUrl:
            'https://raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v2.0/json/petstore-expanded.json',
        },
      },
    ]
  }

    return { label: 'swagger' }
  }

  static render({ status }) {
    if (status === 'valid') {
      return { message: status, color: 'brightgreen' }
    } else {
      return { message: status, color: 'red' }
    }
  }

  async fetch({ specUrl }) {
    return this._requestJson({
      url: 'http://validator.swagger.io/validator/debug',
      schema,
      options: { qs: { url: specUrl } },
    })
  }

  transform({ json, specUrl }) {
    const valMessages = json.schemaValidationMessages

    if (!valMessages || valMessages.length === 0) {
      return { status: 'valid' }
    }

    if (valMessages.length === 1) {
      const { message, level } = valMessages[0]
      if (level === 'error' && message === `Can't read from file ${specUrl}`) {
        throw new NotFound({ prettyMessage: 'spec not found or unreadable' })
      }
    }

    if (valMessages.every(msg => msg.level === 'warning')) {
      return { status: 'valid' }
    }

    return { status: 'invalid' }
  }

  async handle(_routeParams, { specUrl }) {
    const json = await this.fetch({ specUrl })
    const { status } = this.transform({ json, specUrl })
    return this.constructor.render({ status })
  }
}
```

</details>

<details><summary>swagger-redirect.service.js</summary>

```js
'use strict'

module.exports = [
  redirector({
    category: 'other',
    name: 'SwaggerRedirect',
    route: {
      base: 'swagger/valid/2.0',
      pattern: ':scheme(https|http)/:url*',
    },
    transformPath: () => '/swagger/valid/2.0',
    transformQueryParams: ({ scheme, url }) => {
      const suffix = /(yaml|yml|json)$/.test(url) ? '' : '.json'
      return { specUrl: `${scheme}://${url}${suffix}` }
    },
    dateAdded: new Date('2019-11-03'),
  }),
]
```
</details>

Currently, no one is able to use the Swagger badge with the default/extensionless mode when using a json-based spec.

So this doesn't actually work, as the filename is misinterpreted as the requested `.json` format from Shields
https://img.shields.io/swagger/valid/2.0/https/raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v2.0/json/petstore-expanded.json
Badge doesn't work because it returns json --> , but the json response is also misleading (it's really a case of the spec not being found)
```json
{"label":"swagger","message":"invalid","color":"red","link":[],"name":"swagger","value":"invalid"}
```


Shields users would have to explicitly include the `.svg` extension to get a working badge
https://img.shields.io/swagger/valid/2.0/https/raw.githubusercontent.com/OAI/OpenAPI-Specification/master/examples/v2.0/json/petstore-expanded.json.svg


Using the approach I've shared above, this edge case (json-based spec without explicit badge extension) is still providing a json response, but at least now it is the _correct_ json response. 
```json
{"label":"swagger","message":"valid","color":"brightgreen","link":[],"name":"swagger","value":"valid"}
```

I do not see a way to magically correct that original edge case as part of the redirector, so I believe this the best we can do. We knew that there would be a handful of edge cases that would be problematic when we updated the badge formats to `.svg` by default (and dropped the need to always explicitly include `.svg`). 

Everything works fine when explicitly including the extension with the legacy routes, and everything will work fine going forward (since the url of the spec has been moved to the query param)

calebcartwright:
Also, I'd love to actually get some live service tests for this. If there's no publicly accessible specs that are relatively consistent (ones that will be valid or invalid for an extended duration) then we can have the tests point to a live spec with an assertion that uses `Joi.allow('valid', 'invalid')` for the `message`. 

We can also add some tests for the not found scenario now that we're properly detecting them.

Bl4d3s:
Now json should be usable with json format by default, svg needs to be specified in addition just as you said. 

I also added live tests and tests for the redirector.
One live test I'm not sure though: The invalid spec I just used the schema for a spec, but its not a spec as such. (but valid json)

services/swagger/swagger.tester.js
```diff
@@ -21,6 +24,27 @@ t.create('Valid')
     color: 'brightgreen',
   })
```
calebcartwright: Let's remove any of the mocked tests that are duplicative of the real/live tests. You acn also remove the `Live` prefix from the test names, as our test runner automatically denotes whether a test is mocked or live in the output

services/swagger/swagger.tester.js
```diff
@@ -41,3 +65,73 @@ t.create('Invalid')
...
+t.create('Live valid with warnings')
+  .get(`${getURLBase}https://petstore3.swagger.io/api/v3/openapi.json`)
+  .expectBadge({
+    label: 'swagger',
+    message: 'valid',
+    color: 'brightgreen',
+  })
+
+// Isn't a spec, but valid json
+t.create('Live invalid')
```
calebcartwright: > One live test I'm not sure though: The invalid spec I just used the schema for a spec, but its not a spec as such. (but valid json)

I'm guessing this is the test you were referring to?

Bl4d3s: Yes exactly

services/swagger/swagger.tester.js
Bl4d3s: I removed all mocked ones besides the invalid one

services/swagger/swagger.tester.js
```diff
@@ -41,3 +30,52 @@ t.create('Invalid')
...
+
+t.create('Valid with warnings')
+  .get(`${getURLBase}https://petstore3.swagger.io/api/v3/openapi.json`)
+  .expectBadge({
+    label: 'swagger',
+    message: 'valid',
+    color: 'brightgreen',
+  })
+
+// Isn't a spec, but valid json
```
calebcartwright: I think this is fine, as it still allows us to validate our integration with/handling of the response from the validator API 👍

PR #4305, opened 2019-11-08
Don't break the back button when redirecting legacy routes

fregante:
## Problem

1. Visit https://shields.io/#/examples/size (LEFT CLICK, not in a new tab)
2. Browser redirects you to https://shields.io/examples/size
3. Press the browser's `Back` button

## Expected

You'll be back on this issue

## Actual

The browser will load the first URL and Gatsby will redirect back to the second one

## Notes

I followed the [Gatsby's `navigate` docs](https://www.gatsbyjs.org/docs/gatsby-link/#how-to-use-the-navigate-helper-function) but this is untested.


calebcartwright:
hi @fregante, thanks for the PR!

Would you run prettier on your branch when you get a chance? CI is failing due to formatting issues

paulmelnikow:
Out of curiosity, where are you encountering these links?

fregante:
I think I found it on Bundlephobia’s repo

calebcartwright:
Interestingly, I can reproduce the problem as described in both Firefox and Edge, but not in Chrome 🤷‍♂ 

Hopefully there's not a ton of those legacy urls still out there, but I'm 👍 on getting this fixed in Firefox & Edge.

For reference here's the equivalent legacy url in the review app for this pr:
https://shields-staging-pr-4305.herokuapp.com/#/examples/size

calebcartwright:
Thanks again @fregante

PR #4321, opened 2019-11-12
fix flaky [mozillaobservatory] service tests

calebcartwright:
These tests have been pretty flaky, but I think this will finally stabilize them

Some recent failures:
https://circleci.com/gh/badges/daily-tests/722
https://circleci.com/gh/badges/daily-tests/720

### DIFF
--- a/services/mozilla-observatory/mozilla-observatory.tester.js
+++ b/services/mozilla-observatory/mozilla-observatory.tester.js
@@ -3,7 +3,6 @@
 const Joi = require('@hapi/joi')
 const t = (module.exports = require('../tester').createServiceTester())
 
-const validColors = ['brightgreen', 'green', 'yellow', 'orange', 'red']
 const isMessage = Joi.alternatives()
   .try(
     Joi.string().regex(/^[ABCDEF][+-]? \([0-9]{1,3}\/100\)$/),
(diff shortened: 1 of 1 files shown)


services/mozilla-observatory/mozilla-observatory.tester.js
```diff
@@ -3,32 +3,27 @@
...
-  .valid(...validColors)
-  .required()
-
 t.create('request on observatory.mozilla.org')
+  .timeout(10000)
   .get('/grade-score/observatory.mozilla.org.json')
   .expectBadge({
     label: 'observatory',
     message: isMessage,
-    color: isColor,
```
calebcartwright: There's mocked tests below for every possible API response that include color validation, no need to duplicate that here with a catch-all color check

paulmelnikow:
Thanks for keeping up on these!

Have you found the retries have helped reduce transient failures?

calebcartwright:
> Thanks for keeping up on these!

I'm hoping to get to a point where they pass more often than they fail. Not sure about others, but I personally find I tend to ignore the daily test run due to how frequently they fail 😞 

> Have you found the retries have helped reduce transient failures?

Yeah I think they have helped, at least with tests that fail due to the upstream API taking longer for the initial request for a project/repo/etc. My next wish list item would be to add a backoff/increased timeout period as part of the failure retries

PR #4367, opened 2019-11-27
[Packagist] user/repo must be lowercase

g105b:
An issue has developed recently where the package name and repo name of the packagist packages have an added requirement of being lowercase. This has broken the Shields.io badge for some of my projects where their repo name has capital letters, and the badges have been automatically added by a script.


calebcartwright:
> An issue has developed recently where the package name and repo name of the packagist packages have an added requirement of being lowercase

Do you know if this is only being applied to net-new packages, or does it impact existing packages too? I'm assuming the latter based on your description but just want to confirm.

calebcartwright:
you'll also need to run prettier to fix the formatting issues that are causing the ci to fail

g105b:
It's affecting repos that have otherwise been working for a number of years.

OK, bear with me, I forgot...

g105b:
@calebcartwright prettier done :+1:

calebcartwright:
> @calebcartwright prettier done 👍

Perfect thanks! Do you happen to know of any announcements, posts, etc. on the Packagist change? It'd probably be good to add an inline comment with the links/details about the change so future viewers of the code have the context, and we could include a link to the announcement in the comment.

g105b:
I have not found any news/changelog to be honest. In `composer.json` (the config file that packagist uses, similar to `package.json` used by npm), there has always been a lowercase rule to package names, so technically all packages in packagist must already be lowercase, but my personal system for building Shields.io into my ReadMe files didn't take that into account. 

Then about two weeks ago, the repos that use capital letters (e.g. https://github.com/PhpGt/DomTemplate) stopped showing their packagist version in the Shields.io.

I assumed it was a bug in Shields, but I noticed that changing the URL to lowercase fixed the issue, hence the fix in this PR.

calebcartwright:
Can you provide some specific Packagist badge examples from Shields.io that aren't working for you? I know your change in this PR is minor, but there's a couple things puzzling me.

In the example repo you shared, https://github.com/PhpGt/DomTemplate, the readme badges are coming from `https://badge.status.php.gt` and not Shields.io (at least not directly) 

Also when I look at other Packagist badges on Shields.io, I'm not seeing the case sensitivity issues you refer to:

https://img.shields.io/packagist/dm/DOCtRINE/ORM


https://img.shields.io/packagist/dm/doctrine/orm

calebcartwright:
My guess is that one of the two Packagist-related endpoints that Shields hits is case insensitive, while the other is case sensitive. However, IMO we'd need to be careful to ensure that auto-lowercasing is actually a safe thing to do that won't break other badges. 

Many of the composer schema requirements are not strictly enforced on the packagist side which has been the source of several challenges we've faced in Shields.io and it would not surprise me in the least if there are other packages on Packagist that explicitly need uppercase support for their badges to work (i.e. maybe packages that would need https://img.shields.io/packagist/v/Foo/Bar that would be broken if `Foo` and `Bar` were lowercased by Shields)

calebcartwright:
Okay I've found the underlying issue so scratch my last two comments 😄 

We made a switch to use the [Composer Metadata API](https://packagist.org/apidoc#get-package-data) a few weeks ago for as many of the Packagist badges as possible, and that includes the Version badge. 

Both the Composer Metadata API and the Packagist JSON API are case insensitive for the vendor and project name params, so the current fix proposed here that lowercases those values for the API call won't resolve the issue.

For example, note that both of these calls to the Composer Metadata API return the same result:
https://packagist.org/packages/AURA/iNTl.json
https://packagist.org/packages/aura/intl.json

The problem is that the response from the Composer Metadata API normalizes everything to lowercase (even packages/package versions that were published prior to the ~2012 lower case enforcement). Shields then uses the user-provided user values to select the correct package from the API response (as the responses from the Composer Metadata API can include multiple packages if there are similarly-named packages) but we're doing so currently using the case of the user-provided values:

https://github.com/badges/shields/blob/603b9ada7a48c0a2c9ab3a19507d50eccdd84716/services/packagist/packagist-base.js#L69-L71

It's there in `getPackageName` where we'd need to lowercase the user/repo values to solve this

g105b:
> In the example repo you shared, https://github.com/PhpGt/DomTemplate, the readme badges are coming from https://badge.status.php.gt and not Shields.io (at least not directly)

Sorry I should have explained that. There used to be a lot of trouble with the shields.io server which was causing Github's camo.githubusercontent.com caching mechanism to time out, so I have badge.status.php.gt set up to proxy shields.io with a long timeout (Github's is only 5 seconds).

> Can you provide some specific Packagist badge examples from Shields.io that aren't working for you?

There is nothing better than a hard example. Obviously this is affecting my repositories. For example:

The actual badge is here: https://img.shields.io/packagist/v/PhpGt/DomTemplate.svg 

Compare to the lowercase version: https://img.shields.io/packagist/v/phpgt/domtemplate.svg 

My view is the following: 

+ Composer does not allowing uppercase characters in package names
+ All shields.io badges work with all-lowercase characters
+ Some shields.io badges do not work with mixed upper and lower case 

I accept my view above is possibly naive, and I hugely respect your discipline in making sure there are not any backwards breaking changes introduced, so I would look forward to hearing your comments.

calebcartwright:
> so I would look forward to hearing your comments

Thanks @g105b, my [last comment](https://github.com/badges/shields/pull/4367#issuecomment-559863831) details the underlying issue as well as the fix that'd be needed to resolve

g105b:
@calebcartwright thanks, I've changed where the lowercasing occurs now.

calebcartwright:
Great, thanks! Also I was wrong about one of those API calls 🤦‍♂ sorry for that. 

The `fetch` function will need the `user` and `repo` params to be lowercased too like you originally had it

g105b:
Ahh, I'll get there in a minute! Here's the code added back in, but I would be interested in learning how we could add a test for this case - or is this more of an integration test?

calebcartwright:
Seems to be working now 👍 
https://shields-staging-pr-4367.herokuapp.com/packagist/v/PhpGt/DomTemplate

g105b:
Thank you, Caleb, for your open source work.

Issue #4391, opened 2019-12-02
App fails when using a named pipe for the port


hannesne:
Are you experiencing an issue with...

- [ ] [shields.io](https://shields.io/#/)
- [x] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

When hosting the app using iisnode (such as on an Azure App Service plan using Windows), the system exposes a named pipe for the port. This makes the validation logic in core/server/server.js fail, as it expects the port to be a tcp port number (integer 0 - 65560). 

:bulb: **Possible Solution**

Changing this validation rule to accept strings fixes the issue. I will post a PR.

paulmelnikow:
A named pipe? That seems a bit odd!

Will Azure App Service allow running it in Linux instead?

hannesne:
I had a quick go trying to run it in Linux, and couldn't get it to work. It didn't seem to respond appropriately to the pings from the underlying service fabric, and I've not yet figured out why. I'll try to have another look at it later. It did work on a Docker container in a Linux app service, but the container was 168MB big, which caused very long startup times when using a free app service plan. Some Windows features are also not necessarily always available in Linux.

PyvesB:
According to Wikipedia regarding [named pipes in Windows](https://en.wikipedia.org/wiki/Named_pipe#In_Windows):

> They are very rarely seen by users, but there are notable exceptions. The VMware Workstation PC hardware virtualization tool, for instance, can expose emulated serial ports to the host system as named pipes

So ISS is maybe simply built using VMWare underneath the hood, which would explain this?

hannesne:
I would be incredibly surprised to find any vmware in that stack. It all runs on hyper-v. This is the request handler that iis uses in app services when hosting node apps: https://github.com/Azure/iisnode

PyvesB:
Closed by #4392 .

PR #4392, opened 2019-12-02
Changed configuration validation to allow using a named pipe for port.

hannesne:
Fixes issue #4391


core/server/server.js
```diff
@@ -28,7 +28,7 @@ const requiredUrl = optionalUrl.required()
 
 const publicConfigSchema = Joi.object({
   bind: {
-    port: Joi.number().port(),
+    port: Joi.optional(),
```
calebcartwright: Does the below work in the App Service Windows plan env? I'd prefer it if so to keep more explicit validation while supporting that use case

```suggestion
    port: Joi.alternatives().try(
      Joi.number().port(),
      Joi.string(),
    ),
```

paulmelnikow: We could also try using https://www.npmjs.com/package/unc-path-regex.

hannesne: That should work fine. I've got a unit test that confirms it as well.

calebcartwright: @paulmelnikow were you thinking of using `unc-path-regex` as part of a custom `@hapi/joi` validator vs the more lenient `Joi.string()`?

core/server/server.spec.js
```diff
@@ -23,6 +23,14 @@ describe('The server', function() {
     server = undefined
   })
 
+  it('should allow strings for port', async function() {
+    //fixes #4391 - This allows the app to be run using iisnode, which uses a named pipe for the port.
+    const pipeServer = createTestServer({
+      port: '\\\\.\\pipe\\9c137306-7c4d-461e-b7cf-5213a3939ad6',
+    })
+    expect(pipeServer).to.not.undefined
```
calebcartwright: Not sure if this is a Chai shortcut that I'm just not familiar with or if it should have a `.be` in the chain?

```suggestion
    expect(pipeServer).to.not.be.undefined
```

core/server/server.js
PyvesB: (unresolving this conversation as there were still open questions)

Named pipes happen to be a very specific subset of Windows file paths, the aforementioned `unc-path-regex` will accept them but is much more general. In my opinion, adding a dependency for a single edge-case check on startup would be overkill, especially if that dependency does not do exactly what we want.

Why not write a simple regex that validates against the Windows pipe naming specs? The [Microsoft docs](https://docs.microsoft.com/en-us/windows/win32/ipc/pipe-names) indicates `\\ServerName\pipe\PipeName` and even `\\.\pipe\PipeName` if it's the same server.

core/server/server.js
```diff
@@ -28,7 +28,7 @@ const requiredUrl = optionalUrl.required()
 
 const publicConfigSchema = Joi.object({
   bind: {
-    port: Joi.number().port(),
+    port: Joi.alternatives().try(Joi.number().port(), Joi.string()),
```
PyvesB:
```suggestion
    port: Joi.alternatives().try(
      Joi.number().port(),
      Joi.string().pattern(/^\\\\\.\\pipe\\.+$/),
    ),
```
Would something like this do the trick?

core/server/server.js
paulmelnikow: If we use a regex we can use `Joi.string().regex()`; we don't need to make an extension. That package is basically one line, though we could add our own regex that is more limited to `\\.\pipe\PipeName`.

core/server/server.js
hannesne: Ok I've tested and applied the change.

Issue #1955, opened 2018-08-23
Extension to "commits since" badge: allow counting from latest version tag


Lucretiel:
Many projects don't use Github Releases to track releases, but simply ordinary git tags. These show up in the GitHub releases page (for example: https://github.com/Lucretiel/joinery/releases), but aren't marked as releases in the API. It'd be nice if there was another endpoint– perhaps `/commits-since/USER/REPO/latest-tag.svg`– which counts commits after the most recent tag matching `/\d+(\.\d+)/`

Lucretiel:
Prior art: #993

paulmelnikow:
Hi, thanks for the suggestion! It's a nice idea.

To get the latest tag reliably currently requires many API calls (#875), so we may need a caching solution for that first.

Lucretiel:
Yeah, at the end of the day we'd need to implement GitHub paginating. We can fetch up to 100 tags per repo per call, and unfortunately I don't see any obvious way to only fetch new tags.

agilgur5:
So `/github/tag` and `/github/tag-pre` (and `/github/tag-date`) currently get the latest SemVer tag (and date tag). Is it now possible that `commits-since` could extend that logic to allow for commits since latest SemVer?

PR #4395, opened 2019-12-04
add sort and prerelease query params to [GitHubCommitsSince] (also refactors [GitHubRelease], [GitHubTag])

agilgur5:
## Description

Uses the new query params logic from the GitHubRelease Service to provide the same functionality when for the GitHubCommitsSince Service when using the `latest` release (so one can use e.g.  `latest.json?sort=semver`).

## Motivation

I have some libraries where I have a commits since badge meant to track number of commits since the "latest" release. Since "latest" by GitHub API is by date, this means if I release a backport, the badge will start tracking from that backport since it's latest by date. Tracking commits since latest by semver is what I would like to do.

The workaround I've been having to do without this is to put the latest SemVer version into the badge and updating it whenever I make a new release (tedious, and has to already be pinned before any backport is released, otherwise above bug will occur until the badge is pinned, and if, say it's released on NPM before the badge is pinned, it will be tracking the backport until another release is added)

## Related

Related to #1955 ; upon second read, I realized that asks for the addition of _tags_ on top of just _releases_, what I was looking for in https://github.com/badges/shields/issues/1955#issuecomment-495466824 is actually different -- the addition of SemVer. Guess I read that wrong when I made that comment and never filed a separate issue for SemVer support.

## Review Notes

Some things to look at in particular:
- I renamed `github-release.spec.js` to `github-common-fetch.spec.js` because I moved the logic there so it could be shared. `getLatestRelease` isn't normally exposed, so I exported it as `_getLatestRelease` so that the tests could import it. Not sure what the optimal/preferred way would be to do this (other than rewriting the tests to run against the new `fetchLatestRelease`).
  - **EDIT**: I also don't know how to run `github-common-fetch.spec.js` as it itself is not a service. It's currently causing CI to error because it's not found 😕 
- The examples and tests for commits since. I could add more permutations of how one could combine the query params with each other and with branches, but not sure if that is necessary.

## Potential Future Opportunities

In addition to `latest`, could extending tracking for specific SemVer line -- e.g. `^1.0.0` would track latest `1.x.x` releases. This seems far from trivial (and unsure of efficiency on GitHub API), but it would allow say, a backport branch to show how many commits since the latest backport, instead of latest by SemVer or latest by date. Could also be used to track a certain SemVer line of dependencies or something


calebcartwright:
Thanks for the PR @agilgur5! 

> EDIT: I also don't know how to run github-common-fetch.spec.js as it itself is not a service. It's currently causing CI to error because it's not found 😕

That's correct, the `[ServiceClassName..]` notation in the PR title's are used to specify which subset of the service tests to run in the service test stage (more details [here](https://github.com/badges/shields/blob/master/doc/service-tests.md#pull-requests)). That's utilized because the of the nature of the service tests (integration testing with the upstream platforms/APIs) and how long it takes to run the full suite.

There's other jobs that will run all of the unit tests, including those defined in `github-common-fetch.spec.js`, so those don't have to be called out explicitly in the PR title.

I've updated the PR title accordingly and everything looks green now 👍

agilgur5:
Thanks for the super-fast response!! 😮 

> There's other jobs that will run all of the unit tests, including those defined in `github-common-fetch.spec.js`, so those don't have to be called out explicitly in the PR title.

Ah I see, gotcha. I read the `service-tests` docs and didn't have trouble understanding those, but I didn't know what the Jasmine tests ran under. Somehow I didn't realize they were unit tests when I was reading through (despite looking up what the `sazerac` package does) -- that makes a lot more sense now 😅

calebcartwright:
> Thanks for the super-fast response

Fortunate timing 😄 

CI looks good and the review app is up at https://shields-staging-pr-4395.herokuapp.com/ if you'd like to poke around. 

I'm signing out for the night but this'll get a more thorough review from a maintainer soon. Thanks again!

PyvesB:
Thanks for the contribution @agilgur5! Will review it over the week-end if no one picks it up before. 😉

services/github/github-common-fetch.js
```diff
@@ -1,8 +1,9 @@
 'use strict'
```
calebcartwright: Would it make sense to move the GitHub release specific content into a separate helper module? I know there was previously domain-specific artifacts in here for Issues and a bit for Releases, but it's getting pretty Release heavy and I wonder if splitting would be helpful.

services/github/github-commits-since.service.js
```diff
@@ -77,6 +81,52 @@ module.exports = class GithubCommitsSince extends GithubAuthV3Service {
         }),
         documentation,
       },
+      {
+        title: 'GitHub commits since latest including pre-releases',
```
calebcartwright: This is the example for latest by date right? If so then it'd be good to mention that explicitly in the title to ensure the difference is clear for end users. Check out the example titles used in the GH Release service for a reference

services/github/github-commits-since.tester.js
```diff
@@ -33,6 +33,20 @@ t.create('Commits since by latest release (branch)')
     message: isMetric,
   })
```
calebcartwright: We have some schema validation helpers defined in a few helper modules within the `services` directory that help reduce the `@hapi/joi` boilerplate/repetition, including `withRegex` in `test-validators.js` that you could use here

services/github/github-commits-since.tester.js
```diff
@@ -33,6 +33,20 @@ t.create('Commits since by latest release (branch)')
...
 
+t.create('Commits since by latest SemVer release')
+  .get('/microsoft/typescript/latest.json?sort=semver')
+  .expectBadge({
+    label: Joi.string().regex(/^(commits since){1}[\s\S]+$/),
```
calebcartwright:
```suggestion
    label: isCommitsSince,
```

services/github/github-commits-since.tester.js
agilgur5: ok sure done, added another commit to refactor this in

services/github/github-commits-since.service.js
agilgur5: I copy+pasted from GH Release into here and then edited them. The pure latest (no pre-releases) is above this, so I left this with the same wording. Just changed them all to be explicit about "by date", though they're not explicitly using the `sort=date` query param, just that it's the default.

services/github/github-common-fetch.js
agilgur5: Possibly, the release stuff seems to be exclusively used by the 3 services I refactored (`commits-since`, `release`, and `tag`), but I don't have enough context on this codebase to really know 🤷‍♂ I can do whichever.

I thought the audience for question was other collaborators

calebcartwright: > I thought the audience for this question was other maintainers

Sorry I could've been more clear, but I intended it as an open question for anyone, maintainer, author, or otherwise 😄 

> Possibly, the release stuff seems to be exclusively used by the 3 services I refactored

That was my sense as well and why I posed the question. There's a mix of other existing GitHub-related services, plus more that are planned/requested, that will utilize the non-release specific content from `github-common-fetch`. 

I'll leave it up to you, we don't need to block this on splitting them out, but if you feel like doing it that'd be great!

agilgur5: Ok, I've split it off into `github-common-release` now. Think that's the last remaining comment here

calebcartwright:
Thanks @agilgur5! Will try to take a final look at this one within the next couple days

PR #4398, opened 2019-12-05
Remove blank space for examples with no named parameters

### LINKED ISSUE #4394: Excessive blank space for examples with no named parameters
Following #4390, we now have a few examples with no named parameters, only query parameters. In the UI, this leads to some excessive blank space above the path:
jenkins

We could reduce the box's size and centre the path string in these cases.

Not the end of the world and not an area of the code I'm familiar with, but I'm happy to have a dig around.

PyvesB:
An attempt at changing some frontend code, which I rarely do. Our amazing Gatsby setup and my IDE made it smooth and interesting though! Closes #4394.

Results on my machine:
* Firefox:
firefox

* Opera:
opera


frontend/components/customizer/path-builder.tsx
```diff
@@ -229,11 +243,13 @@ export default function PathBuilder({
   }
 
   let namedParamIndex = 0
+  const onlyLiterals =
+    tokens.filter(token => typeof token === 'string').length === tokens.length
```
calebcartwright: What would you think about using `every` here instead? Believe it'd look something like:
```suggestion
    tokens.every(token => typeof token === 'string')
```

PyvesB: Thanks, I was unaware of that method.

calebcartwright:
I checked the review app in Firefox, Chrome, and Edge on my machine and def seems to have addressed #4394

The code LGTM but I don't know enough about the frontend code myself to offer useful feedback one way or the other, so I'd defer to any of our more frontend-versed maintainers 😄

frontend/components/customizer/path-builder.tsx
```diff
@@ -153,10 +155,22 @@ export default function PathBuilder({
...
     return (
-      <PathBuilderColumn key={`${tokenIndex}-${literal}`}>
-        <PathLiteral isFirstToken={tokenIndex === 0}>{literal}</PathLiteral>
+      <PathBuilderColumn
+        height={onlyLiterals ? '18px' : '78px'}
+        key={`${tokenIndex}-${literal}`}
+      >
+        <PathLiteral
+          isFirstToken={tokenIndex === 0}
+          marginTop={onlyLiterals ? '0px' : '39px'}
```
paulmelnikow: I'd suggest passing `onlyLiterals` into `PathBuilderColumn` and `PathLiteral` and moving the conditional expression there. That way the layout components (i.e. the columns and the literals) are better encapsulated. Also it avoids duplicating the `'78px'` which should make it easier to maintain if this changes.

frontend/components/customizer/path-builder.tsx
```diff
@@ -153,10 +155,22 @@ export default function PathBuilder({
     })
   }
 
-  function renderLiteral(literal: string, tokenIndex: number) {
+  function renderLiteral(
+    literal: string,
+    tokenIndex: number,
+    onlyLiterals: boolean
```
paulmelnikow: It's a bit odd to have a parameter `onlyLiterals` on a function called `renderLiteral()`.

```suggestion
    pathContainsOnlyLiterals: boolean
```


Alternatively you could invert it and call it `hasNamedParams`.

PR #4409, opened 2019-12-11
add [AppVeyorJobBuild] status badge

### LINKED ISSUE #4393: Badge Request: AppVeyor Job Level Status Badges
:clipboard: **Description**
I'm using dynamic badges to display the status of separate AppVeyor jobs since AppVeyor does not support per-job status badges. It is working cool, here is an example Windows build status.

The only problem the color is static, it does not indicate the build status (e.g. green for success, red for failure). I hope it can be made dynamic.

There are two options that I see:
- Add `colorQuery` parameter, but the problem here is that JSON does not contain a color name to be displayed, so this is not appropriate option.
- Add `message-to-color-map` parameter. I think its value should be JSON object, for example:
```JavaScript
{
  "success": "success",
  "failed": "critical",
  "<some placeholder to match other values>": "inactive"
}
```

The only problem here is placeholder value, I hope we can just use empty string for it. But I think it would be better to just use current `color` parameter as fallback.

So, the final idea:
- Add `message-to-color-map` parameter, if the message matches one of the key then use the matched value as badge color.
- If it does not match any key then fallback to the `color` parameter.

calebcartwright:
Resolves #4393


services/appveyor/appveyor-build-job.service.js
```diff
@@ -0,0 +1,68 @@
+'use strict'
+
+const { renderBuildStatusBadge } = require('../build-status')
+const AppVeyorBase = require('./appveyor-base')
+const { NotFound } = require('..')
+
+module.exports = class AppVeyorBuildJob extends AppVeyorBase {
+  static get route() {
+    return {
+      base: 'appveyor/job/build',
```
calebcartwright: I was originally planning on making this `appveyor/build/job` but was unable to do so given the badge for the overall build status starts with `appveyor/build/...` and has the trailing branch route param (the route params were being incorrectly interpreted as being part of the branch name)

services/appveyor/appveyor-build-job.service.js
```diff
@@ -0,0 +1,68 @@
...
+      // this project exists but no builds have been run on it yet
+      return { status: 'no builds found' }
+    }
+
+    const {
+      build: { jobs },
+    } = data
+    if (!jobs) {
+      throw new NotFound({ prettyMessage: 'no jobs found' })
+    }
```
calebcartwright: TBH I'm skeptical there's a real scenario where we'd ever see this, but given the array of TypeError's we see in prod I figured better safe than sorry here

services/appveyor/appveyor-build-redirect.service.js
```diff
@@ -0,0 +1,16 @@
+'use strict'
+
+const { redirector } = require('..')
+
+module.exports = [
+  redirector({
+    category: 'build',
+    route: {
+      base: 'appveyor/ci',
```
calebcartwright: In other services/platforms where we provide multiple CI-related like badges build, test, coverage (Azure DevOps, Jenkins, TeamCity, etc.) we've been using `build`, `tests`, `coverage`, so I figured I'd change appveyor to use `build` instead of `ci` to be consistent.

calebcartwright:
Frontend unit test failing somehow - https://circleci.com/gh/badges/shields/82569 (i can't reproduce locally though 😕) 

```bash
AssertionError: expected [ Array(50) ] to have a length below 50 but got 50
```

Looks like we have a frontend unit test that's verifying `getDefinitionsForCategory` by validating the number of service definitions in the `build` categeory is between 10 and 50, exclusively, and I'm guessing this PR may be adding the 50th `build` service class which is making this fail?
https://github.com/badges/shields/blob/05a0881670040200f7d5e134560ae1211775abea/frontend/lib/service-definitions/index.spec.ts#L11-L15

services/appveyor/appveyor-build-job.service.js
```diff
@@ -0,0 +1,68 @@
+'use strict'
+
+const { renderBuildStatusBadge } = require('../build-status')
+const AppVeyorBase = require('./appveyor-base')
+const { NotFound } = require('..')
+
+module.exports = class AppVeyorBuildJob extends AppVeyorBase {
```
PyvesB: Maybe `AppVeyorJobBuild` to match the ordering of the path?

services/appveyor/appveyor-build-job.service.js
calebcartwright: That was my initial thought as well, and if we were rolling out the AppVeyor support for the very first time I would've wanted to structure the schema that way.

I decided not to do that here though because the existing AppVeyor badges that have been in use do not enforce this at the schema level, and the [API documentation](https://www.appveyor.com/docs/api/projects-builds/#get-project-last-build) I could find does not provide much details around the response schema (just one example response). 

Given that uncertainty, my thinking is that it'd be better to handle this case programmatically instead of potentially causing `invalid response data` badges, especially on the existing AppVeyor Build/Test badges.

services/appveyor/appveyor-build-job.service.js
calebcartwright: I suppose it'd be best to have the file name, class name, and route as aligned as possible 😄 

The last minute route base change is what caused the discrepancy here but will update the others accordingly

PR #4434, opened 2019-12-14
A response size metric of an upstream service for dynamic badges like [dynamicyaml], (but not [jetbrains])

platan:
This PR adds a metric for response size of an upstream service for dynamic badges. I chose a histogram metric, mainly because in the histogram "Observations are very cheap as they only need to increment counters." (https://prometheus.io/docs/practices/histograms/). In histograms observations fall into buckets. Here we have buckets: 64kiB, 128kiB, ...,4MiB, 8MiB. 

Dynamic badges fetch data from any source and for performance, reliability and security reasons fetch limit should be lower than 10MB (current limit). This metric should help us in determining the new limit. 

How to test it?
```bash
METRICS_PROMETHEUS_ENABLED=true npm run start

# dynamic XML
curl "http://localhost:8080/badge/dynamic/xml.svg?color=red&label=label1&query=%2F%2Fname&url=https%3A%2F%2Fraw.githubusercontent.com%2Fhsz%2Fidea-gitignore%2Fmaster%2Fresources%2FMETA-INF%2Fplugin.xml"

# dynamic YAML
curl "http://localhost:8080/badge/dynamic/yaml.svg?color=red&label=label1&query=%24.trailingComma&url=https%3A%2F%2Fraw.githubusercontent.com%2Fbadges%2Fshields%2Fmaster%2F.prettierrc.yml"

# dynamic JSON
curl "http://localhost:8080/badge/dynamic/json.svg?color=red&label=label1&query=%24.name&url=https%3A%2F%2Fraw.githubusercontent.com%2Fbadges%2Fshields%2Fmaster%2Fpackage.json"

# results
curl "http://localhost:8080/metrics" -s | grep "service_response_bytes"
```

<details>
<summary><code>service_response_bytes_bucket{le="65536",category="dynamic",family="dynamic",service="dynamic_xml"} 1 ...</code></summary>

```
service_response_bytes_bucket{le="65536",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="131072",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="262144",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="524288",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="1048576",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="2097152",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="4194304",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="8388608",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="+Inf",category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_sum{category="dynamic",family="dynamic",service="dynamic_xml"} 32529
service_response_bytes_count{category="dynamic",family="dynamic",service="dynamic_xml"} 1
service_response_bytes_bucket{le="65536",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="131072",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="262144",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="524288",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="1048576",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="2097152",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="4194304",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="8388608",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="+Inf",category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_sum{category="dynamic",family="dynamic",service="dynamic_yaml"} 84
service_response_bytes_count{category="dynamic",family="dynamic",service="dynamic_yaml"} 1
service_response_bytes_bucket{le="65536",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_bucket{le="131072",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_bucket{le="262144",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_bucket{le="524288",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_bucket{le="1048576",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_bucket{le="2097152",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_bucket{le="4194304",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_bucket{le="8388608",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_bucket{le="+Inf",category="dynamic",family="dynamic",service="dynamic_json"} 1
service_response_bytes_sum{category="dynamic",family="dynamic",service="dynamic_json"} 10959
service_response_bytes_count{category="dynamic",family="dynamic",service="dynamic_json"} 1
```
</details>

You may want to compare it with the previous approach https://github.com/badges/shields/pull/3662


core/base-service/metric-helper.js
```diff
@@ -11,9 +11,17 @@ class MetricHelper {
         serviceFamily,
         name,
       })
+      this.serviceResponseSizeHistogram = metricInstance.createServiceResponseSizeHistogram(
```
calebcartwright: I'm definitely out of my element here on the promethus content, and this is not a blocker for me by any means, more some general questions/food for thought.

I find myself wondering if would it be possible to use more generic names in here (i.e. not including the `Histogram` suffix on `this.serviceResponseSizeHistogram`) and/or if perhaps aspects of the `prom-client` API could be more fully encapsulated within `prometheus-metrics.js`? 

The theoretical use cases I was thinking of (and def not suggesting these) were around what changes would be required here if for example there was an API change `prom-client` (such as `observe(..)` being renamed to `note(..)` in a new major version), if `serviceResponseSize` was changed to a different metric type in `prometheus-metrics.js`, etc.

core/base-service/base.js
```diff
@@ -214,20 +214,43 @@ class BaseService {
...
     this.authHelper = authHelper
     this._handleInternalErrors = handleInternalErrors
+    this._metricHelper = metricHelper
   }
 
   async _request({ url, options = {}, errorMessages = {} }) {
     const logTrace = (...args) => trace.logTrace('fetch', ...args)
     logTrace(emojic.bowAndArrow, 'Request', url, '\n', options)
     const { res, buffer } = await this._requestFetcher(url, options)
+    await this._meterResponse(res, buffer)
```
calebcartwright: Is it necessary to `await` this one or is it possible to process the response/return the badge to the user and send the response size data to prometheus in parallel/afterwards?

platan: Thanks for this accurate comment! 
`await` is it not necessary at all. Prometheus uses pull model over HTTP and we expose our metrics at `/metrics`. We store metrics in memory and we do not send them anywhere (no network calls). This particular metric is currently a histogram and collecting data in histograms is not a CPU intensive operation (in the histogram "Observations are very cheap as they only need to increment counters."). But this may change in the future - we may change histogram to something else or event replace Prometheus with an other monitoring solution which can do network calls. 
Therefore we should gather metrics in the background and return a response to the user  without a delay. I will remove `await`. 

After removing `await` `service response size metric is optional` test is passing, but it should wait for `expect` to be true. Do you have a suggestion how we can test it? My proposition is to use https://github.com/TheBrainFamily/wait-for-expect.

core/base-service/metric-helper.js
platan: > I find myself wondering if would it be possible to use more generic names in here (i.e. not including the `Histogram` suffix on `this.serviceResponseSizeHistogram`) and/or if perhaps aspects of the `prom-client` API could be more fully encapsulated within `prometheus-metrics.js`?
>
I agree that encapsulating is a good idea. I would like to compare our code with another metrics API which I'm familiar with - Java library Metrics (https://metrics.dropwizard.io/4.1.2/getting-started.html). Metrics uses similar concepts like Prometheus (https://prometheus.io/docs/concepts/metric_types/): counters, gauges, histograms. Name `serviceResponseSizeHistogram` give an info what kind of metric it is and what kind of data it stores. In my opinion we should leave `serviceResponseSizeHistogram` and `serviceRequestCounter` unchanged. 
 
> The theoretical use cases I was thinking of (and def not suggesting these) were around what changes would be required here if for example there was an API change `prom-client` (such as `observe(..)` being renamed to `note(..)` in a new major version), if `serviceResponseSize` was changed to a different metric type in `prometheus-metrics.js`, etc.

Prometheus's counter and Metrics's counter uses `#inc()`. Prometheus's histogram uses `#observe()` and Metrics's histogram uses `#update()` for updates. I just want to show that despite using Prometheus API directly in `metric-helper.js` method we are calling have quite general an universal names. 
Moreover `prometheus-metrics.js` is used only in `metric-helper.js` and this limits possible places requiring changes. 

What do you think about it? :-)

calebcartwright:
Thanks for the ping. I'll be travelling for a couple more days and still mostly AFK, but will follow up as soon as I can

core/base-service/metric-helper.js
calebcartwright: > What do you think about it? :-)

I still think that some refactoring could be beneficial, but definitely agreed that everything metric-related is well confined within `core` and that's why I think this is more of a potential follow up (if anything) to this PR. 

My main observation is that when viewing `metric-helper.js` directly, there seem to be some assumptions about the internals of things within `prometheus-metrics.js`. That for me is an indicator that there's the potential opportunity to encapsulate more within `prometheus-metrics.js`

core/base-service/base.js
calebcartwright: > After removing await service response size metric is optional test is passing, but it should wait for expect to be true

I'm not sure I follow, could you elaborate a bit more? Are you saying that the test is passing, but that the validation/assertion statement is basically invalid (because the test is skipping it)? 

Also am I correct that this is the test you are referring to?
https://github.com/badges/shields/pull/4434/files#diff-0567f28266920ae49126b66b59269cdfR516-R539

platan: That's right. https://github.com/badges/shields/pull/4434/files#diff-0567f28266920ae49126b66b59269cdfR516-R539 is testing `_request` method. In the test we are waiting for the result of the `await serviceInstance._request({ url, options })`, but `_request` is collecting metrics by async method `_meterResponse` in parallel. In the test we are not waiting for `_meterResponse` to finish. Currently metrics are collected in memory and test is passing. But in the future this can change. We can simulate it by:
```diff
@@ -243,4 +243,5 @@ class BaseService {
 
   async _meterResponse(res, buffer) {
+    await new Promise((resolve) => setTimeout(resolve, 1))
     if (
       this._metricHelper &&
```

core/base-service/metric-helper.js
platan: OK, I will try to do some refactoring in a separate PR :-)

core/base-service/base.js
calebcartwright: Gotcha, thanks for the clarification. I hadn't looked closely enough at the innards of `_meterResponse`. 

By any chance do we have any profiling/measurement available? I'd venture to say that the easiest thing for now would be to keep that `await` in there, as I'm guessing there's only a very very small amount of time needed right now to cache/queue the metric in the current setup. 

Similar to the other comments here, this is something that could be revisited if there's any refactoring here.

platan: I restored this `await`. Saving this data should take negligible amount of time.

platan:
Thank you @calebcartwright! Anyone else want to look at this or can I merge it?

PR #4435, opened 2019-12-15
Add documentation for designing badge URLs

### LINKED ISSUE #3716: Update documentation for designing badge URLs
https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-urls should be expanded and updated based on the decision documented in #3693 and #3714.

There might also be some useful bits in https://docs.google.com/document/d/11RCKeEqMQmKnh8Ij0sywRZPE66wxQBwHqiE9-OAZqhQ/edit that can be incorporated.

chris48s:
Closes #3716
This documents the rules discussed in https://github.com/badges/shields/pull/3144#issuecomment-470336042 and formalises some of the work we've recently done on moving certain params into the query string or making routes more consistent.


doc/badge-urls.md
```diff
@@ -0,0 +1,27 @@
...
+  - The parameter is related to formatting. e.g: `/appveyor/tests/:user/:repo?compact_message`.
+  - The parameter is for an uncommon optional attribute, like an alternate registry URL.
+  - The parameter triggers application of alternative logic, like version semantics. e.g: `/github/v/tag/:user/:repo?sort=semver`.
+  - Services which require a url/hostname parameter always should use a query string parameter to accept that value. e.g: `/discourse/topics?server=https://meta.discourse.org`.
+
+It is convention to use the following standard routes and abbreviations across services:
+
+- Coverage: `/coverage`
+- Downloads or Installs:
+  - Total: `/dt`
```
calebcartwright: Do we have a position on the convention for services that don't provide interval/period breakdowns of the download data for example a service that just has total downloads? I imagine we'd want to encourage `/dt` regardless (for best-case future compatibility if and when a service _does_ add interval/period data), and if so, perhaps we could state that explicitly?

```suggestion
  - Total: `/dt` - Use this even for services that only provide the total download/install data 
```

doc/badge-urls.md
```diff
@@ -0,0 +1,27 @@
...
+  - Per month: `/dm`
+  - Per week: `/dw`
+  - Per day: `/dd`
+  - For version: `/dv`
```
calebcartwright: I believe ATM this is used in the download badges for gems and crates, but couldn't this be a bit limiting for services that support the daily/weekly/monthly/yearly/etc. download counts both for latest/current version and a specific version? For example if a user wanted the monthly download count of v1.0.0 of package foo.

(I can't think of a service off hand that supports download counts for versions but figured I'd ask anyway 😄)

doc/badge-urls.md
chris48s: Agreed. There may be some cases where we have just `/d`but I think we should encourage `/dt` as a baseline for the reasons you've said there

doc/badge-urls.md
chris48s: Good question. Given we don't yet have an example of a service where we've done this (AFAIK), there is no convention to document yet.

I think in that case, we would use `dd`/`dw`/`dm` but give all of them an optional version param, so

* `service/dd/mypackage` - downloads per day (all versions)
* `service/dd/mypackage/1.0.0` - downloads per day (version 1.0.0)

as opposed to having `/dd` and `/ddv` (or whatever) as 2 seperate endpoints.

Does that clarify your thinking on what we should document at all?

calebcartwright: > I think in that case, we would use dd/dw/dm but give all of them an optional version param, so
> 
> * service/dd/mypackage - downloads per day (all versions)
> * service/dd/mypackage/1.0.0 - downloads per day (version 1.0.0)

That's exactly what I had in mind as well. I'd vote for that approach vs. recommending `/dv` for the same future-compatible reasons as recommending `/dt` discussed above

chris48s: Yep sounds reasonable. I've raised #4454 to move the routes we've got in line with this.

PR #4436, opened 2019-12-15
Convert vpre routes to include_prereleases param in [bower chocolatey clojars packagist powershellgallery pub resharper]

chris48s:
This migrates most of the `vpre` routes to use and `include_prereleases` query param as discussed in https://github.com/badges/shields/pull/3144#issuecomment-470684713 and in line with the conventions documented in #4435 .
The exception is myget and nuget which still use regex routes. I'm going to ignore thise for now. We can get those two done when we convert to `:param` routes.


platan:
I'm happy to look at this PR during the weekend.

services/bower/bower-version.tester.js
```diff
@@ -44,10 +49,15 @@ t.create('Version label should be `no releases` if no stable version')
...
-  .get('/vpre/bootstrap.json')
+  .get('/v/bootstrap.json?include_prereleases')
   .intercept(nock =>
     nock('https://libraries.io')
       .get('/api/bower/bootstrap')
       .reply(200, { normalized_licenses: [], latest_release_number: null })
   )
   .expectBadge({ label: 'bower', message: 'no releases' })
+
+t.create('Version (legacy redirect: vpre)')
```
platan: Instead of testing `.json` resource we can test `/vpre/bootstrap.svg` or `/vpre/bootstrap` (because we are verifying a header, not a JSON content) - both are redirecting to `/bower/v/bootstrap.svg?include_prereleases`. Properly working redirects for SVG badges are more important than JSON resources.

services/bower/bower-version.tester.js
```diff
@@ -44,10 +49,15 @@ t.create('Version label should be `no releases` if no stable version')
...
+  .get('/vpre/bootstrap.json', { followRedirect: false })
+  .expectStatus(301)
```
platan: Badges are verified by `.expectBadge()` assertion. How about adding a new custom assertion for redirects?
```diff
diff --git a/core/service-test-runner/icedfrisby-shields.js b/core/service-test-runner/icedfrisby-shields.js
index ec1748da..4b1070f9 100644
--- a/core/service-test-runner/icedfrisby-shields.js
+++ b/core/service-test-runner/icedfrisby-shields.js
@@ -53,0 +54,4 @@ const factory = superclass =>
+    expectRedirect(location) {
+      return this.expectStatus(301).expectHeader('Location', location)
+    }
+
```
and
```diff
diff --git a/services/bower/bower-version.tester.js b/services/bower/bower-version.tester.js
index 613c4900..58502377 100644
--- a/services/bower/bower-version.tester.js
+++ b/services/bower/bower-version.tester.js
@@ -62,2 +62 @@ t.create('Version (legacy redirect: vpre)')
-  .expectStatus(301)
-  .expectHeader('Location', '/bower/v/bootstrap.json?include_prereleases')
+  .expectRedirect('/bower/v/bootstrap.json?include_prereleases')
```

services/bower/bower-version.tester.js
```diff
@@ -44,10 +49,15 @@ t.create('Version label should be `no releases` if no stable version')
...
+  .get('/v/bootstrap.json?include_prereleases')
   .intercept(nock =>
     nock('https://libraries.io')
       .get('/api/bower/bootstrap')
       .reply(200, { normalized_licenses: [], latest_release_number: null })
   )
   .expectBadge({ label: 'bower', message: 'no releases' })
+
+t.create('Version (legacy redirect: vpre)')
+  .get('/vpre/bootstrap.json', { followRedirect: false })
```
platan: We need `{ followRedirect: false }` if we want to test redirects, because `followRedirect` is `true` by default. But do we actually need following redirects in our tests? In my opinion we can change `followRedirect` to `false` by default in all service tests. What's your opinion?

services/bower/bower-version.service.js
```diff
@@ -37,10 +42,11 @@ module.exports = class BowerVersion extends BaseBowerService {
     return { label: 'bower' }
   }
 
-  async handle({ vtype, packageName }) {
+  async handle({ packageName }, queryParams) {
     const data = await this.fetch({ packageName })
+    const includePrereleases = queryParams.include_prereleases !== undefined
 
-    if (vtype === 'v') {
+    if (!includePrereleases) {
```
platan: Isn't more readable without this negated condition? (https://eslint.org/docs/rules/no-negated-condition)
```js
    if (includePrereleases) {
      if (data.latest_release_number) {
        return renderVersionBadge({ version: data.latest_release_number })
      }
    } else {
        if (data.latest_stable_release) {
          return renderVersionBadge({ version: data.latest_stable_release.name })
        }
    }
```

services/bower/bower-version.tester.js
chris48s: Agreed - I've changed it for the services already touched in this PR.
I'll defer doing the same on other services to another PR.

services/bower/bower-version.tester.js
chris48s: Again, agreed. I've added that helper here and changed it for the services already touched in this PR. I'll leave updating the other service tests to use it for a different PR (maybe several) to avoid scope creep. There are about 30 other services to update.

services/bower/bower-version.tester.js
chris48s: I'm going to leave this one for now as it needs a full test suite run. In the abstract, I think you're probably right but I'd want to do a full service test run with it off to see what happens.

services/resharper/resharper.tester.js
```diff
@@ -41,12 +41,20 @@ t.create('version (not found)')
...
   })
 
 t.create('version (pre) (not found)')
-  .get('/vpre/not-a-real-package.json')
+  .get('/v/not-a-real-package.json?include_prereleases')
   .expectBadge({ label: 'resharper', message: 'not found' })
+
+t.create('version (legacy redirect: vpre)')
+  .get('/vpre/ReSharper.Nuke.json', { followRedirect: false })
+  .expectStatus(301)
```
platan: This is the last place in this PR where you can use `expectRedirect` + `.svg`.

platan:
I've added a new chart with redirectors only https://metrics.shields.io/d/aESRBSjmz/services?orgId=1&from=now-15m&to=now: 
image

PR #4459, opened 2019-12-23
Template literal templates

### LINKED ISSUE #2428: Move badge templates from doT → Template literals

chris48s:
This turned into an absolute beast, but I think we're finally good for review. I suspect the review process for this one is going to be pretty involved too. I'll mark up the diff with a couple of specific comments, but here's some high level notes:

- The core purpose of this PR is to remove use of the doT template library and move to generating SVG output using javascript template literals. The rendered SVG should be unchanged in all cases.
- Paul started this work in #3637 and then it stalled. I've pretty much squashed all the work from that PR into 058fa48 (apart from the bit we already merged in #3395) and then iterated from there.
- This PR targets the `v3-dev` branch, not `master`. Getting this merged will unblock the rest of the work we need to do on V3 so hopefully we'll be able to get the rest of the breaking changes done and merged to master once we can draw a line under this.
- A secondary consequence of this work is we needed to re-consider our approach to SVG optimisation. Roughly in line with Paul's comments in https://github.com/badges/shields/pull/3637#issuecomment-524078874 I decided to drop SVGO at the library layer and _mostly_ manually implement the optimisations (running SVGO won't be an _exact_ no-op). See 286636d , 8e62a86 , b877bc9 and 2f62ab7 . The exception to this is that sometimes SVGO replaces a `<rect>` with a `<path>` element if the `<path>` takes up less characters. I haven't bothered implementing this optimisation. This does mean that we'll serve a few extra characters for every badge (the re-generated snapshots in ed24253 give an idea of the impact of this), but I think the tradeoff is OK.
- I decided that `flat`, `flat-square` and `plastic` are all sufficiently similar that they are best thought of as slight variations of the same thing so they are implemented as child classes of a base class. This allows most of the implementation to be shared. `for-the-badge` and `social` styles were so different that they don't really shoehorn into that pattern, so those two are just implemented as completely bespoke render functions, as opposed to as subclasses of `Badge`.
- Given this PR is really ripping out the guts of the core rendering logic and starting again, this PR could use review from more than one of the core maintainers.
- There's a staging deploy at https://shields-staging-pr-4459.herokuapp.com/
- Things that would be particularly useful to get feedback on:
  - Are we escaping all the things that need escaping - have I missed any calls to `escapeXml()`?
  - Are there any cases or situations I've forgotten to test?

Closes #2428
Refs #3587


__snapshots__/badge-renderers.spec.js
```diff
@@ -0,0 +1,655 @@
...
+        <rect width="45" height="20" fill="#0f0"/>
+        <rect x="45" width="45" height="20" fill="#b3e"/>
+        <rect width="90" height="20" fill="url(#s)"/>
+      </g>
+
+      <g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="110">
+        
+        <text x="235" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="350">cactus</text><text x="235" y="140" transform="scale(.1)" textLength="350">cactus</text>
+        <text x="665" y="150" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="350">grown</text><text x="665" y="140" transform="scale(.1)" textLength="350">grown</text>
+      </g>
```
chris48s: I'm in 2 minds about the best way to run these tests/record the test snapshots. I've done it with `minify=false` because I figured that if something changes its easier to read the diff than if its all smashed into one line, but the flipside of this is the tests are more brittle because they're sensitive to any whitespace changes (which isn't the case with `minify=true`). Any thoughts?

gh-badges/lib/make-badge.spec.js
```diff
@@ -139,6 +139,48 @@ describe('The badge generator', function() {
     })
   })
 
+  describe('"flat" template badge generation', function() {
+    it('should match snapshot', function() {
```
chris48s: I'm also in 2 minds about these tests. I added them in fcf6c68 because I thought they would help but they now essentially duplicate one case of the more extensive tests that exist in `gh-badges/lib/badge-renderers.spec.js`, but one layer further out in the application. They're not doing any harm, but I'm not totally convinced its useful to duplicate them. Again, any thoughts?

calebcartwright:
I don't think I've ever looked at the code for the npm package, and I'm _way_ out of my depth here 😄 

I'll try to review it soon though to see if there's anything helpful I can add

paulmelnikow:
Thanks so much for picking this up, Chris! I will look for some time to review it, either tonight or in the next week or so.

gh-badges/package.json
```diff
@@ -35,10 +35,8 @@
   },
```
calebcartwright: Couldn't get this up above on the `name` field, but where did we land on potentially renaming the package (or is that out of scope for this PR)?

gh-badges/lib/badge-renderers.js
```diff
@@ -0,0 +1,596 @@
...
+}
+
+function renderLogo({
+  logo,
+  logoWidth = 14,
+  logoPadding = 0,
+  extraPadding = 0,
+}) {
+  if (logo) {
+    const x = 5 + extraPadding
```
calebcartwright: For my own edificiation, where does the extra +5 and +3 come from/why those specific values?

gh-badges/lib/make-badge.js
```diff
@@ -1,104 +1,21 @@
...
 it is likely this will impact on the package's public interface in index.js
 */
 module.exports = function makeBadge({
   format,
-  template,
+  template = 'flat',
   text,
+  colorscheme,
   color,
+  colorA,
```
calebcartwright: At first glance I am surprised to see `colorscheme`, `colorA`, and `colorB` added here since they were deprecated and (I presumed) were potentially going to be dropped entirely as part of 3.x

gh-badges/lib/make-badge.spec.js
calebcartwright: > I'm also in 2 minds about these tests

Just to ensure I'm looking at the right set of tests, how many of the below tests fall into this category for you? Just down to line 182 or all the `it('should match snapshot', ...)` tests?

gh-badges/lib/badge-renderers.js
```diff
@@ -0,0 +1,596 @@
...
+  if (params.minify) {
+    return stripXmlWhitespace(badge.render())
+  }
+  return badge.render()
+}
+
+function flatSquare(params) {
+  const badge = new FlatSquare(params)
+  if (params.minify) {
+    return stripXmlWhitespace(badge.render())
```
paulmelnikow: Not necessary for merging this, though it would be nice to reduce the duplication of the minify logic. Perhaps it could go in the Badge superclass or in the calling code.

gh-badges/lib/badge-renderers.js
```diff
@@ -0,0 +1,596 @@
...
+    throw new Error('Not implemented')
+  }
+
+  static get verticalMargin() {
+    throw new Error('Not implemented')
+  }
+
+  static get shadow() {
+    throw new Error('Not implemented')
+  }
```
paulmelnikow: Nice!

gh-badges/lib/badge-renderers.js
```diff
@@ -0,0 +1,596 @@
...
+    labelColor = '#555',
+  }) {
+    const { hasLogo, totalLogoWidth, renderedLogo } = renderLogo({
+      logo,
+      logoWidth,
+      logoPadding,
+    })
+    const hasLabel = label.length
+
+    labelColor = hasLabel || (hasLogo && labelColor) ? labelColor : color
```
paulmelnikow: This is fairly puzzling, particularly the `(hasLogo && labelColor)` bit.

paulmelnikow:
Do you think we should test the performance difference? IIRC the script in `scripts/benchmark-performance.sh` is designed to test how much time we spend in `makeBadge()`.

gh-badges/lib/make-badge.spec.js
chris48s: > how many of the below tests fall into this category for you?

All the ones added in fcf6c68

> my preference would be to test it at this layer rather than one level down

Yeah that makes sense. That kind of answers the other question too.

gh-badges/package.json
chris48s: Yep - changing the name is one of the things to do for version 3. Its one of the points in #3587
The scope of this PR is just to remove doT, as opposed to completing _all_ the remaining breaking changes for V3.

gh-badges/lib/badge-renderers.js
chris48s: Broadly because they were hard-coded values in the original templates we're trying to match.

I can't link to a line here because GitHub won't render the files, but if you look at say line 8 of
https://github.com/badges/shields/blob/v3-dev/gh-badges/templates/flat-square-template.svg
the X and Y values for the image are just hard-coded to 5 and 3. This function generalises so that we can pass additional offsets for `for-the-badge` style.

Any thoughts on whether there is a more intuitive way to express it?

gh-badges/lib/make-badge.js
chris48s: Good catch :+1: #3637 was originally targetted at `master`. When I rebased that stuff onto `v3-dev`, I hadn't realised I was accidentally undoing some of the work from https://github.com/badges/shields/pull/3881 I'll fix this.

gh-badges/lib/make-badge.js
```diff
@@ -108,75 +25,47 @@ module.exports = function makeBadge({
...
-      template = 'flat'
-    }
-  }
-  if (template === 'social') {
-    left = capitalize(left)
-  } else if (template === 'for-the-badge') {
-    left = left.toUpperCase()
-    right = right.toUpperCase()
+  template = camelcase(template)
+  if (template.startsWith('popout')) {
```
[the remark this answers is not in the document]
chris48s: Yeah good question.. I hadn't really questioned the behaviour of this for library users. As implemented, this will do the same thing for library users and shields.io users. i.e: as implemented a call to `makeBadge({text: ['build', 'passed'], color: 'green', template: 'popout-square' })` will render a `flat-square` badge, but we could break it and force lib users to make that choice as part of their upgrade instead of making an implicit assumption. Its more explicit/less "magic" but also maybe less convenient. I suppose given this realase changes the public interface by replacing `BadgeFactory.create()` with `makeBadge()`, you're going to have to change all your code anyway..

Would that be your preference?

gh-badges/lib/make-badge.js
```diff
@@ -108,75 +25,47 @@ module.exports = function makeBadge({
...
-  if (leftWidth % 2 === 0) {
-    leftWidth++
-  }
-  let rightWidth = (anafanafo(right) / 10) | 0
-  // Increase chances of pixel grid alignment.
-  if (rightWidth % 2 === 0) {
-    rightWidth++
+  let render = badgeRenderers[template]
+  if (render === undefined) {
+    render = badgeRenderers.flat
```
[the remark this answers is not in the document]
chris48s: As it stands, if you're a library user template is currently optional: https://github.com/badges/shields/blob/c05394d023e89334eaa46b38fd8c276a094d62a0/gh-badges/lib/index.js#L73
Are you saying we should make template a required param in the library, or do you just mean we should keep the existing behaviour but push the responsibility for allowing this to be optional out a layer in the application?

If we were going to get rid of the default on template for package users, now is the time to do that because its a breaking change but I don't see it as a sensible thing to do tbh. I think defaulting to flat if template is not supplied is sensible behaviour and I see no reason to break backwards compatibility here.

Alternatively, maybe you just mean we should keep the behaviour but push the responsibility for this out a layer? In that case: if we want the npm package and server to have the same behaviour, it probably makes more sense to do it in one place instead of two places but I don't have a really strong opinion on it.

gh-badges/lib/badge-renderers.js
```diff
@@ -0,0 +1,596 @@
...
+  extraPadding = 0,
+}) {
+  if (logo) {
+    const x = 5 + extraPadding
+    const y = 3 + extraPadding
+    return {
+      hasLogo: true,
+      totalLogoWidth: logoWidth + logoPadding,
+      renderedLogo: `<image x="${x}" y="${y}" width="${logoWidth}" height="14" xlink:href="${escapeXml(
+        escapeXml(logo)
```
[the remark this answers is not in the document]
chris48s: nope - fixed in 735c03c . Will review the tests we've got..

gh-badges/lib/badge-renderers.js
```diff
@@ -0,0 +1,596 @@
...
+      renderedText = `<text x="${x}" y="${shadowMargin}" fill="#010101" fill-opacity=".3" transform="scale(.1)" textLength="${outTextLength}">${escapedContent}</text>`
+    }
+    renderedText += `<text x="${x}" y="${textMargin}" transform="scale(.1)" textLength="${outTextLength}">${escapedContent}</text>`
+
+    return {
+      renderedText,
+      width: textLength,
+    }
+  }
+  return { renderedText: '', width: 0 }
```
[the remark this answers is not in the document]
chris48s: done in a47f9cc

gh-badges/lib/badge-renderers.js
chris48s: Yeah its pretty weird isn't it. Again I can't link to a line here because GitHub won't render the files, but if you look at line 12 of https://github.com/badges/shields/blob/v3-dev/gh-badges/templates/flat-template.svg that's where it originally came from. Any idea what the original intent of that is?

chris48s:
> Do you think we should test the performance difference? IIRC the script in `scripts/benchmark-performance.sh` is designed to test how much time we spend in `makeBadge()`.

This took a bit of fiddling (looks like somewhere in setting up the config schema in `server.js` we broke `PROFILE_MAKE_BADGE`). I don't want to shave that yak just now, but I got it to work by manually setting the vars. My prediction was that this would slightly speed up the process because:
- SVGO is no longer parsing the XML string into an object, manipulating it and serializing it back into a string
- We've removed a template library and replaced it with native JS code
..so I was surprised to find that I ran the benchmarks and found the new code is slightly slower :(

before (branch:master): 0.0436251
after (branch:template-literals2): 0.0585933

gh-badges/lib/make-badge.js
```diff
@@ -108,75 +25,47 @@ module.exports = function makeBadge({
...
-    colorA: toSvgColor(labelColor),
-    colorB: toSvgColor(color),
-    escapeXml,
-  }
-
-  const templateFn = templates[template]
-
-  // The call to template() can raise an exception.
-  return templateFn(context)
+    logoPadding: logo && label.length ? 3 : 0,
```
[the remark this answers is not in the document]
chris48s: The text gets trimmed at the start of this function: https://github.com/badges/shields/pull/4459/files#diff-c3370bf5c3927aea8d8e020f1933a493L109 so its already happened by the time we reach this line

paulmelnikow:
Disappointing that it's slower, though @mbarkhau had predicted as much (see https://github.com/badges/shields/issues/2428#issuecomment-466771373). I'm sure we could improve on it in the future if we wanted to. Though for now, I don't think it's much to worry about. I'd be more concerned if it was 10x slower. It's still a teeny, tiny part of the work of rendering a badge.

gh-badges/lib/badge-renderers.js
paulmelnikow: Maybe it was meant to be `(hasLabel || hasLogo) && labelColor`, which would make more sense.

gh-badges/lib/make-badge.js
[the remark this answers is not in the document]
paulmelnikow: Ahh, sorry I should have explained that better.

I think it is a good idea to keep `flat` the default for the library, as it is now. I think this layer is a fine place for that to happen.

The behavior I'm questioning is the library accepting non-existent styles and rendering them as `flat`. I think it should throw an error in that case.

Then the server will need to know which styles are valid, and perform its own validation, to avoid passing a nonexistent style to the renderer.

gh-badges/lib/make-badge.js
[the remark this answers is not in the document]
paulmelnikow: Yea. I doubt library users are using popout to begin with, though if they are I think it's a better idea to fail loudly than silently give them what they aren't expecting.

paulmelnikow:
By the way, I'm guessing that the best way to optimize that code is to:

1. Reduce the use of dots by using variables.
2. Reduce the complexity of the call stack.
3. Maybe see if using functions instead of classes speeds things up.

gh-badges/lib/make-badge.js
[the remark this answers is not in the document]
paulmelnikow: > In that case: if we want the npm package and server to have the same behaviour, it probably makes more sense to do it in one place instead of two places but I don't have a really strong opinion on it.

My suggestion is that the package and the server should reject non-existent styles (e.g. `socail`). That may go against this principle, since it means both the server and the package are doing this validation. I guess my counter-principle is this: it is a good idea for web APIs to validate their inputs early, rather than passing around possible garbage values and catching the errors later. We have a schema for the query params and it's easy to provide a good error message if we reject an invalid value then.

gh-badges/lib/badge-renderers.js
calebcartwright: > Any thoughts on whether there is a more intuitive way to express it?

Not really, other than perhaps an inline comment or something like that. I hadn't really looked at the svg template files before so when perusing the js here they seemed like semi-magical numbers

gh-badges/lib/make-badge.spec.js
chris48s: 62529b2

gh-badges/lib/make-badge.js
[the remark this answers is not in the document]
chris48s: 93115dd

gh-badges/lib/make-badge.js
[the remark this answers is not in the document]
chris48s: yep - that all makes much more sense: 619d850

gh-badges/lib/badge-renderers.js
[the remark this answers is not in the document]
chris48s: Done in 9470b82 but also while I was looking at this realised some of our existing tests made no sense..

gh-badges/lib/badge-renderers.js
chris48s: ccdc468 - on close inspection, checking `labelColor` is redundant here anyway

gh-badges/lib/badge-renderers.js
chris48s: Maybe instead of assuming a base position and allowing it to be modified with an offset, we just explicitly pass in X and Y for each badge size, then we can make it more obvious how the positions relate to each badge format. That might be clearer..

gh-badges/lib/badge-renderers.js
```diff
@@ -0,0 +1,596 @@
...
+    <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="${width}" height="${height}">
+    ${main}
+    ${renderLinks({ links, leftWidth, rightWidth, height })}
+    </svg>`
+}
+
+function stripXmlWhitespace(xml) {
+  return xml
+    .replace(/>\s+/g, '>')
+    .replace(/<\s+/g, '<')
```
RedSparr0w: I'm not 100% sure what the function is for, but would it be better to strip all whitespace before and after the `<>` chars?
```suggestion
    .replace(/\s*<\s*/g, '<')
```

On another note, would this affect someone who wanted to make a badge that said something like label | less < more?

gh-badges/lib/badge-renderers.js
paulmelnikow: That seems like a nice idea. Also, we could call it `horizPadding` even if it’s an `x` value.

gh-badges/lib/badge-renderers.js
chris48s: > It's been awhile so i'm not 100% sure what the function is for

It allows us to generate

```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <linearGradient id="b" x2="0" y2="100%">
    ...
</svg>
```

(or whatever)..but then serve it as

```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><linearGradient id="b" x2="0" y2="100%">...</svg>
```

without using SVGO, or otherwise parsing into an XML object and writing back out to a string.


> But would it be better to strip all whitespace before and after the <> chars?

We're calling `.trim()` at the end which will strip any leading/trailing whitespace. Is that what you mean?



> would this affect someone who wanted to make a badge that said something like label | less < more?

I had thought of this. As long as we've run everything through `escapeXml()`, we'll replace `<` with `&lt;` and so on, so its fine as long as everything is correctly escaped. This would cause problems if we've got an unescaped `<` or `>` though

gh-badges/lib/badge-renderers.js
chris48s: cool. I'll aim to push a change for this tomorrow. Having chewed it over a bit more, I've actually realised there is an error I never noticed before.

We hard-code y=3 for `flat` and `plastic`, but y should really be derived from the badge height, so it should be ([badge height] - [logo height])/2 so for the flat badge that is (20 - 14)/2 = 3, but then the plastic badge is slightly shorter which means when we hard-code the y pos to 3, the logo actually renders a bit too low vertically i.e:

https://img.shields.io/badge/plastic-style-blue?style=plastic&logo=javascript

gh-badges/lib/badge-renderers.js
RedSparr0w: > We're calling `.trim()` at the end which will strip any leading/trailing whitespace. Is that what you mean?

Just for the spacing in between,  like if there was an extra space before the `>` it wouldn't be removed with the current code,
```diff
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" >
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
but thinking about it, It's probably not actually an issue seeing as all that is already hard coded.



> I had thought of this. As long as we've run everything through `escapeXml()`, we'll replace `<` with `&lt;` and so on, so its fine as long as everything is correctly escaped. This would cause problems if we've got an unescaped `<` or `>` though

Good point, forgot about this 👍

gh-badges/lib/make-badge.js
```diff
@@ -108,75 +22,43 @@ module.exports = function makeBadge({
...
-  if (leftWidth % 2 === 0) {
-    leftWidth++
-  }
-  let rightWidth = (anafanafo(right) / 10) | 0
-  // Increase chances of pixel grid alignment.
-  if (rightWidth % 2 === 0) {
-    rightWidth++
+  const methodName = camelcase(template)
+  if (!(methodName in badgeRenderers)) {
+    throw new Error(`Unknown template: '${template}'`)
```
paulmelnikow: Not blocking, though at some point this should be consolidated with the style validation logicin gh-badges/lib/index.js.

paulmelnikow:
This is looking good!

gh-badges/lib/badge-renderers.js
chris48s: done in c5ef23a

chris48s:
> Something I haven't done is look at a bunch of badges. What do you think about restoring the test page

Yep. Good call - that's now at https://shields-staging-pr-4459.herokuapp.com/dev/styles

Having looked at all those, I spotted another issue which I've fixed in 34db0f2

chris48s:
Cheers for review all :)

I've merged this onto the v3-dev branch. Over the weekend I'll review where I got to with the v3 roadmap: what else needs to go into v3.0/what can wait till v3.1 and pick up the rest of this. I'd like to get to a stage where we can ship this sooner rather than later.

PR #4475, opened 2019-12-28
Updates [Discord] badge for API changes

night:


services/discord/discord.service.js
```diff
@@ -4,9 +4,7 @@ const Joi = require('@hapi/joi')
 const { BaseJsonService } = require('..')
 
 const discordSchema = Joi.object({
-  members: Joi.array()
-    .allow(null)
-    .required(),
+  presence_count: Joi.number().required(),
```
calebcartwright: We've got some helper validators we can use here. If we import `validators` up above (https://github.com/badges/shields/blob/master/services/validators.js) then we can use one of those validators here:

```suggestion
  presence_count: nonNegativeInteger,
```

Issue #4477, opened 2019-12-28
GitHub Deployments


TrueBrain:
:clipboard: **Description**

GitHub has (basic) support for deployments. We started to use this, and I have been trying to get a badge for this. Sadly, the API is a bit too complicated to simply use the dynamic badge, or at least as far as I understand it. So having support for this, or helping me out how to do it, would be greatly appreciated :)

What I would like the badge to show, per environment, what the current state is of the latest deployment. So for example:

![GitHub Deployments](https://img.shields.io/badge/staging-success-green)

:link: **Data**

For example:
https://api.github.com/repos/OpenTTD/website/deployments?environment=staging

This gives an overview of all the deployments. We need the latest. In here is `statuses_url` which leads to:
https://api.github.com/repos/OpenTTD/website/deployments/191234207/statuses

Here the top result is the latest update of the deployment, the one I would like to see. The `status` field is the important part here.

See https://developer.github.com/v3/repos/deployments/ for more information.

:microphone: **Motivation**

As we are slowly migrating to having GitHub the center of the universe, having a badge with the latest deployment status is the next logic step.
The current issue I am having is that I have to make two dynamic calls for a single badge. I have no idea how I could be doing that.

calebcartwright:
I wonder if the v4 API provides the data more readily available via a single request 🤔 

Definitely worth looking into supporting this badge, though in the immediate short term one possible solution would be to use our endpoint badge (https://shields.io/endpoint) type with a custom endpoint that makes those two API calls and returns the data in the required format

calebcartwright:
Yup, looks like the latest deployment status data can be retrieved via a single request using the v4 GraphQL endpoint (https://developer.github.com/v4/object/deployment/)

Sample query:
```graphql
query($user: String!, $repo: String!, $environment: String!) {
  repository(owner: $user, name: $repo) {
    deployments(last: 1, environments: [$environment]) {
      nodes {
        latestStatus {
          state
        }
      }
    }
  }
}
```

If anyone is interested in working on this, we've a great [tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) on implementing new badges!

There's a handful of our other GitHub badges that are using the new v4 endpoints that can be used as references as well: https://github.com/search?utf8=%E2%9C%93&q=extends+GithubAuthV4Service+repo%3Abadges%2Fshields+path%3A%2Fservices&type=Code&ref=advsearch&l=&l=

TrueBrain:
[...]

PR #4478, opened 2019-12-29
Add [GithubDeployments] badge (#4477)

TrueBrain:
My first attempt to add a badge here :) Few things worth mentioning:

- I could not find a good category, so I picked 'other'.
- I could not find a colouring scheme that fits deployments, so I created one.
- I tried to figure out how `*.tester.js` works, but the [documentation](https://github.com/badges/shields/blob/master/doc/service-tests.md) was not clear enough for me to do anything with (I still have no clue what `get()` is doing, and how to add anything to it). So I did not add any.

Btw, the [tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) is very awesome and well written; thank you for that.

Curious what you think of this!

In case you want to test how the badge looks, you can use https://github.com/OpenTTD/website ([environments](https://github.com/OpenTTD/website/deployments) 'staging' or 'production' works). It can be difficult to find a GitHub project with deployments enabled.

### DIFF
--- a/services/github/github-deployments.service.js
+++ b/services/github/github-deployments.service.js
@@ -1,0 +1,138 @@
+'use strict'
+
+const gql = require('graphql-tag')
+const Joi = require('@hapi/joi')
+const { GithubAuthV4Service } = require('./github-auth-service')
+const { documentation, transformErrors } = require('./github-helpers')
+const { NotFound } = require('..')
+
+const greenStates = ['SUCCESS']
+const redStates = ['ERROR', 'FAILURE']
+const blueStates = ['INACTIVE']
+const otherStates = ['IN_PROGRESS', 'QUEUED', 'PENDING']
+
+const allState = greenStates
+  .concat(redStates)
+  .concat(blueStates)
+  .concat(otherStates)
+
+const isDeploymentState = Joi.equal(...allState)
+
+const schema = Joi.object({
+  data: Joi.object({
+    repository: Joi.object({
+      deployments: Joi.object({
+        nodes: Joi.array()
+          .items(
+            Joi.object({
+              latestStatus: Joi.alternatives([
+                Joi.object({
+                  state: isDeploymentState,
+                }),
+                null,
+              ]),
+            })
+          )
+          .required(),
+      }).required(),
+    }).required(),
+  }).required(),
+}).required()
+
+module.exports = class GithubDeployments extends GithubAuthV4Service {
+  static get category() {
+    return 'other'
+  }
+
+  static get route() {
+    return {
+      base: 'github/deployments',
+      pattern: ':user/:repo/:environment',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'GitHub deployments',
+        namedParams: {
+          user: 'badges',
+          repo: 'shields',
+          environment: 'shields-staging',
+        },
+        staticPreview: this.render({
+          state: 'success',
+        }),
+        documentation,
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'state' }
+  }
+
+  static render({ state }) {
+    let color
+    if (greenStates.includes(state)) {
+      color = 'brightgreen'
+    } else if (redStates.includes(state)) {
+      color = 'red'
+    } else if (blueStates.includes(state)) {
+      color = 'blue'
+    }
+
+    let message
+    if (state == 'IN_PROGRESS') {
+      message = 'in progress'
+    } else {
+      message = state.toLowerCase()
+    }
+
+    return {
+      message,
+      color,
+    }
+  }
+
+  async fetch({ user, repo, environment }) {
+    return await this._requestGraphql({
+      query: gql`
+        query($user: String!, $repo: String!, $environment: String!) {
+          repository(owner: $user, name: $repo) {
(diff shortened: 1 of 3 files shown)


chris48s:
I've also edited this PR title so that when you push service tests to the branch, they'll be run in CI

services/github/github-deployments.service.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'GitHub deployments',
+        namedParams: {
+          user: 'badges',
+          repo: 'shields',
+          environment: 'production',
```
calebcartwright: let's using the staging env instead so the example provides a better sample
```suggestion
          environment: 'shields-staging',
```

TrueBrain: Had not seen this repo also uses deployments :D Sweet, that makes the test-case also a lot easier, tnx :)

TrueBrain:
@chris48s: thank you very much! Now adding the test was easy :) Rereading the page now also makes more sense. What totally put me on the wrong track, were the llinks to wercker itself. To me it read like we were calling their API, instead of the shields API. Totally makes sense now.

Related to the colouring: `renderBuildStatusBadge` has one drawback: it changes the message of the `greenStatus` always to `passing`. `passing` from a deployment point of view is a silly word to use. But I guess I can work around that relatively easy. At least avoids a ton of code duplication :)

services/github/github-deployments.service.js
```diff
@@ -0,0 +1,122 @@
...
+
+const schema = Joi.object({
+  data: Joi.object({
+    repository: Joi.object({
+      deployments: Joi.object({
+        nodes: Joi.array()
+          .items(
+            Joi.object({
+              latestStatus: Joi.object({
+                state: Joi.string(),
```
calebcartwright: We should use a stricter set of validations here based on the known/expected status values

services/github/github-deployments.service.js
```diff
@@ -0,0 +1,122 @@
...
+    })
+
+    let message = build_message
+
+    // renderBuildStatusBadge() does everything we need except it rewrites the
+    // message for status in 'greenStatuses' to 'passing'. This makes little
+    // sense in the context of a deployment. So rewrite it back to 'success'.
+    if (state == 'success') {
+      message = 'success'
+    }
```
calebcartwright: Given the above, is it really worth using the `renderBuildStatusBadge` helper here? Seems like all we'd be using it for here is to map a color to the `state` value from the deployment object

services/github/github-deployments.service.js
```diff
@@ -0,0 +1,122 @@
...
+    if (data.repository.deployments.nodes.length == 0) {
+      return { state: 'unknown' }
+    }
+
+    let state = data.repository.deployments.nodes[0].latestStatus.state
+    state = state.toLowerCase()
+
+    if (state == 'in_progress') {
+      return { state: 'in progress' }
+    }
```
calebcartwright: This (and the lowercasing) feel more like rendering responsibilities than transform responsibilities IMO. Generally speaking those are, respectively, presenting the data to the user in a pretty/friendly way, and extracting the relevant data portions from the raw response

services/github/github-deployments.service.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+      `,
+      variables: { user, repo, environment },
+      schema,
+      transformErrors,
+    })
+  }
+
+  transform({ data }) {
+    if (data.repository.deployments.nodes.length == 0) {
```
calebcartwright: In what scenarios would this occur? Is it when the user specifies an invalid environment value, when there were no deployments to the specified environment, something else?

services/github/github-deployments.service.js
calebcartwright: For reference, here's the raw values from GitHub that Shields will see (note the all upper case) https://developer.github.com/v4/enum/deploymentstatusstate/

services/github/github-deployments.service.js
TrueBrain: That sounds a lot better :D One other corner case is when there is a deployment which did not receive a status update yet. I am not sure we can differentiate between "env not found" and "deployment without status". Also not sure it is worth the effort (given it is a rare corner case in normal operations).

chris48s:
My initial though was that this maps well onto "Build Status" as most of the status values already existed there, but if we're trying to bash a square peg into a round hole, lets keep it custom :+1:

services/github/github-deployments.tester.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Deployments')
+  .get('/badges/shields/shields-staging.json')
+  .expectBadge({
+    label: 'state',
+    message: Joi.string(),
```
chris48s: Can we define a more specific validation here.
For example, if the badge said `'environment not found'`, that would pass `Joi.string()`, but it shouldn't

TrueBrain: I do apologize, as you mentioned that before. For some reason .. I forgot again :) Fixed!

calebcartwright:
Awesome, thanks @TrueBrain!

PR #4489, opened 2020-01-04
Update to eslint 6.8.0 [appveyor githubissuedetail packagist]

znarf:
- update to eslint 6.8.0
- turn off new or updated rules generating warning and errors
- update related packages
- removed related unnecessary packages


.eslintrc.yml
```diff
@@ -1,5 +1,4 @@
 extends:
-  - standard-jsx
```
calebcartwright: Can you provide more detail on why this ruleset and associated dependencies are being removed?

znarf: `eslint-config-standard-react` extends `eslint-config-standard-jsx` https://github.com/standard/eslint-config-standard-react/blob/master/eslintrc.json#L21 so I think it's not justified to add it separately.

.eslintrc.yml
```diff
@@ -48,8 +46,13 @@ overrides:
     parser: '@typescript-eslint/parser'
     rules:
       # Argh.
-      '@typescript-eslint/explicit-function-return-type': 'off'
+      '@typescript-eslint/explicit-function-return-type':
+        ['error', { 'allowExpressions': true }]
+      '@typescript-eslint/no-empty-function': 'error'
+      '@typescript-eslint/no-var-requires': 'error'
```
znarf: If these are fixed, maybe we can remove these rules and use the default configuration?

znarf:
@paulmelnikow awesome follow up, definitely the right thing to do!

package.json
```diff
@@ -163,9 +163,8 @@
     "@types/react-modal": "^3.10.1",
     "@types/react-select": "^3.0.8",
     "@types/styled-components": "4.1.8",
-    "@typescript-eslint/eslint-plugin": "^1.13.0",
-    "@typescript-eslint/parser": "^1.13.0",
-    "acorn": "^7.1.0",
```
znarf: This was required at some point as peer dependency of a sub-sub-dependency of `@typescript-eslint` but it's not anymore with the upgrade.

package.json
```diff
@@ -183,23 +182,19 @@
     "danger-plugin-no-test-shortcuts": "^2.0.0",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
-    "eslint": "^5.16.0",
+    "eslint": "^6.8.0",
     "eslint-config-prettier": "^6.9.0",
-    "eslint-config-standard": "^12.0.0",
```
znarf: It's not used anywhere but maybe that's a mistake?

package.json
```diff
@@ -183,23 +182,19 @@
...
     "eslint-plugin-import": "^2.19.1",
     "eslint-plugin-jsdoc": "^18.6.0",
     "eslint-plugin-mocha": "^6.2.2",
     "eslint-plugin-no-extension-in-require": "^0.2.0",
-    "eslint-plugin-node": "^11.0.0",
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-react": "^7.17.0",
     "eslint-plugin-react-hooks": "^2.3.0",
     "eslint-plugin-sort-class-members": "^1.6.0",
-    "eslint-plugin-standard": "^4.0.1",
```
znarf: Was required as peer of `eslint-config-standard`

package.json
```diff
@@ -183,23 +182,19 @@
...
-    "eslint-config-standard": "^12.0.0",
-    "eslint-config-standard-jsx": "^8.1.0",
     "eslint-config-standard-react": "^9.0.0",
     "eslint-plugin-chai-friendly": "^0.5.0",
     "eslint-plugin-cypress": "^2.8.1",
     "eslint-plugin-import": "^2.19.1",
     "eslint-plugin-jsdoc": "^18.6.0",
     "eslint-plugin-mocha": "^6.2.2",
     "eslint-plugin-no-extension-in-require": "^0.2.0",
-    "eslint-plugin-node": "^11.0.0",
```
znarf: Was required as peer of `eslint-config-standard`

package.json
```diff
@@ -183,23 +182,19 @@
     "danger-plugin-no-test-shortcuts": "^2.0.0",
     "enzyme": "^3.11.0",
     "enzyme-adapter-react-16": "^1.15.2",
-    "eslint": "^5.16.0",
+    "eslint": "^6.8.0",
     "eslint-config-prettier": "^6.9.0",
-    "eslint-config-standard": "^12.0.0",
-    "eslint-config-standard-jsx": "^8.1.0",
```
znarf: Removed from the updated configuration because it's not needed, `eslint-config-standard-react` is extending it.

.eslintrc.yml
paulmelnikow: Because these rules throw errors on .js files, they are turned off at the top level, so I think they need to be turned back on for .ts files.

package.json
paulmelnikow: We use `standard-react` –  does it declare its own dependency on `standard`?

paulmelnikow: Ah yes: https://github.com/badges/shields/pull/4489#discussion_r363082988

znarf: @paulmelnikow When I checked, seemed `eslint-config-standard-react` was NOT bringing `eslint-config-standard`. I'm not too familiar with it so not sure what to do at this point.

znarf: @paulmelnikow I just checked and `eslint-config-standard` is not enforced right now in the codebase, if we activate it, we get many errors all over the place. Most of it can be automatically fixed with `--fix` but then the syntax is not compatible with prettier rules (`space-before-function-paren`, `comma-dangle`).

znarf:
Updated the PR with an activation of `eslint-config-standard`.

package.json
paulmelnikow: eslint-config-prettier should disable those, maybe check the order in which those two are listed in extends?

paulmelnikow: Ah, you must have solved that.

znarf: I adjusted the config in https://github.com/badges/shields/pull/4489/commits/4bd3ec8abf4bc58c479b1aa3f08dbe9ce64f3808

calebcartwright:
@znarf when you get a chance, could you rebase or resolve the merge conflicts?

frontend/components/customizer/request-markup-button.tsx
```diff
@@ -3,14 +3,17 @@ import styled from 'styled-components'
...
 
-const ClickableControl = (props: any) => (
-  <components.Control
-    {...props}
-    innerProps={{
-      onMouseDown: props.selectProps.onControlMouseDown,
-    }}
-  />
-)
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
calebcartwright: Why are these inline overrides needed? Is there ruleset conflicts on the function signature, something else?

calebcartwright:
Also as a general observation, seems there are a few changes that are switching away from arrow functions.

We'd previously had discussions around preferring arrow functions where applicable for stylistic preferences ( for example: https://github.com/badges/shields/pull/2803#discussion_r249011621), so I was just curious if the changes here are different circumstances, or if this represents a change in our style preferences

znarf:
@paulmelnikow can you look at @calebcartwright comments?

paulmelnikow:
> We'd previously had discussions around preferring arrow functions where applicable for stylistic preferences ( for example: [#2803 (comment)](https://github.com/badges/shields/pull/2803#discussion_r249011621)), so I was just curious if the changes here are different circumstances, or if this represents a change in our style preferences

My reading of that discussion is not so clear. The cleanup I found that referenced back to it was #2813, which prevents using the `function` keyword in function expressions, but allows arrow function expressions, which we do use all over the place. Most of that PR was converting declarations of the form

```js
const foo = () => ...
```

to

```js
function foo() {
  ...
```

I see one change like that in https://github.com/badges/shields/pull/4489/files#diff-ca6e5284d51981f3d13d88efe70eab4c – is that the one you mean @calebcartwright?

It seems consistent with the stylistic cleanup that was done in #2813. We do use arrow functions a lot, but don't usually assign them to variables.

frontend/components/customizer/request-markup-button.tsx
paulmelnikow: Perhaps we should add a comment. Basically I don't think we should be using `any` or `ts-ignore`, however the correct type of these Control props were difficult to track down and irrelevant to what we're doing here, since we just pass in the ClickableControl class which gets instantiated by react-select.

calebcartwright: Ahhh okay. I'll confess I've used `ts-ignore` a few times but did feel a little guilty 😄

calebcartwright:
> It seems consistent with the stylistic cleanup that was done in #2813. We do use arrow functions a lot, but don't usually assign them to variables.

If you're good with it @paulmelnikow then I'm good! I hadn't looked at it too closely so just wanted to ask

package.json
paulmelnikow: I'm seeing this message in the logs from `npm install` and I'm thinking we should restore this.

```
npm WARN eslint-config-react-app@5.2.0 requires a peer of eslint-plugin-react-hooks@1.x but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-standard@14.1.0 requires a peer of eslint-plugin-node@>=9.1.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-standard@14.1.0 requires a peer of eslint-plugin-standard@>=4.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN rollup-plugin-typescript2@0.25.2 requires a peer of rollup@>=1.26.3 but none is installed. You must install peer dependencies yourself.
```

PR #4523, opened 2020-01-12
drop raster support in package CLI

chris48s:
Refs https://github.com/badges/shields/issues/3587#issuecomment-573446622
This is one of the approaches we could take..
The other approach would be #4524
If we decide we like this one, I'll make sure the the README/docblock is up-to-date for this change before marking for review


chris48s:
Ha. The integration test failure is because `v3-dev` isn't up to date with `master` so doesn't have #4616 on it. One of the things to do after this when I look at the final PR to merge `v3-dev` onto `master` and deal with changelog, final docs pass, packaging mods + etc is to bring `v3-dev` up to date with `master` again.

chris48s:
Would it be possible to get a review on this seeing as we've had no showstoppers in #4564
Then I'll rebase `v3-dev` and get the last PR for version 3 done

calebcartwright:
Sorry! For some reason I mistakenly thought we were all set here. Will do a quick final look now

PR #4524, opened 2020-01-12
support raster formats in makeBadge

chris48s:
Refs https://github.com/badges/shields/issues/3587#issuecomment-573446622
This is one of the approaches we could take..
The other approach would be #4523
If we decide we like this one, I'll make sure the the README/docblock is up-to-date for this change before marking for review

### DIFF
--- a/gh-badges/lib/index.js
+++ b/gh-badges/lib/index.js
@@ -5,6 +5,12 @@
 
 const _makeBadge = require('./make-badge')
 const svg2img = require('./svg-to-img')
+let gm
+try {
+  gm = require('gm')
+} catch (e) {
+  gm = null
+}
 
 class ValidationError extends Error {}
 
@@ -29,14 +35,14 @@ function _validate(format) {
     }
   })
 
-  try {
-    require('gm')
-  } catch (e) {
-    if ('format' in format && ['png', 'jpg', 'gif'].includes(format.format)) {
-      throw new ValidationError(
-        `peerDependency gm is required for output in .${format.format} format`
-      )
-    }
+  if (
+    !gm &&
+    'format' in format &&
+    ['png', 'jpg', 'gif'].includes(format.format)
+  ) {
+    throw new ValidationError(
+      `peerDependency gm is required for output in .${format.format} format`
+    )
   }
 
   const formatValues = ['svg', 'json', 'png', 'jpg', 'gif']
--- a/gh-badges/lib/index.js
+++ b/gh-badges/lib/index.js
@@ -4,6 +4,7 @@
  */
 
 const _makeBadge = require('./make-badge')
+const svg2img = require('./svg-to-img')
 
 class ValidationError extends Error {}
 
@@ -28,7 +29,17 @@ function _validate(format) {
     }
   })
 
-  const formatValues = ['svg', 'json']
+  try {
+    require('gm')
+  } catch (e) {
+    if ('format' in format && ['png', 'jpg', 'gif'].includes(format.format)) {
+      throw new ValidationError(
+        `peerDependency gm is required for output in .${format.format} format`
+      )
+    }
+  }
+
+  const formatValues = ['svg', 'json', 'png', 'jpg', 'gif']
   if ('format' in format && !formatValues.includes(format.format)) {
     throw new ValidationError(
       `Field \`format\` must be one of (${formatValues.toString()})`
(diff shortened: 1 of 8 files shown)


gh-badges/lib/index.js
```diff
@@ -72,10 +83,14 @@ function _clean(format) {
  * @returns {string} Badge in SVG or JSON format
  * @see https://github.com/badges/shields/tree/master/gh-badges/README.md
  */
-function makeBadge(format) {
+async function makeBadge(format) {
```
chris48s: This change would require us to switch `makeBadge()` to an async function

gh-badges/lib/index.js
```diff
@@ -28,7 +29,17 @@ function _validate(format) {
     }
   })
 
-  const formatValues = ['svg', 'json']
+  try {
+    require('gm')
+  } catch (e) {
+    if ('format' in format && ['png', 'jpg', 'gif'].includes(format.format)) {
+      throw new ValidationError(
+        `peerDependency gm is required for output in .${format.format} format`
```
chris48s: I'm not really sure how to test this case.
I've moved `gm` (which is now a `peerDependency`) to a dev dependency of the application. That means we can run the tests for raster formats, but then it means we can't test this case. The 2 methods I can see would be:

- Somehow mock the call to require to force it to throw
- Hive off this one test and run it in its own CI build without `gm` installed (and remove it from the main `npm test` run)

I don't really have any experience testing stuff with peer dependencies. Anyone know what is the usual tactic for this? In one sense, the simple approach is just to ignore this and go with the approach in #4523 , but its concept we'll need to find an approach for if we expose logos in the library and make `simple-icons` a peer dependency anyway.

calebcartwright: I wonder if using [require.resolve](https://nodejs.org/api/modules.html#modules_require_resolve_request_options) would be easier to test than a try/catch 🤔 

```js
  if (!require.resolve('gm')) {
    // conditional format check and throw
  }
```

and then in the unit tests use Sinon to stub conditionally?
```js
  Sinon.stub(require, 'resolve').withArgs('gm').callsFake(() => ...)
```

Disclaimer though, haven't tried that and have absolutely no clue if that'll actually work

chris48s: did a bit of reading and the consensus seems to be: use proxyquire for this
https://github.com/thlorenz/proxyquire#using-proxyquire-to-simulate-the-absence-of-modules

calebcartwright:
I'd vote for this one over #4523. 

I don't feel strongly one way or the other though, just think it'd be good for the package to continue to support at least png if possible

PyvesB:
Have any of our users asked for support in the npm package? Do we know if any of them currently use this in the CLI?

Given that we've moved the raster service out of the main Shields.io service and given how easy it is to wrap the API or pipe a command line utility to turn an SVG into a JPEG or PNG, I would argue we're not adding much value and could drop support.

Maybe it's worth reaching out to the community and ask what they think? We could create a new issue with a basic thumbs up/down voting system, ask for feedback on Discord, create a Twitter poll, etc.

chris48s:
> given how easy it is to wrap the API or pipe a command line utility to turn an SVG into a JPEG or PNG, I would argue we're not adding much value and could drop support.

Exactly. This is why my preferred approach is to drop it, as in https://github.com/badges/shields/pull/4523

If we completely removed raster support from shields.io, there's no workaround for those that rely on it (hence the proxy). If you're currently using raster support in the CLI and we drop it, the workaround is to pipe the output of `badge` to to imagemagick (which I think you'd need to have to have installed anyway if you were relying on that functionality).

calebcartwright:
> We could create a new issue with a basic thumbs up/down voting system, ask for feedback on Discord, create a Twitter poll, etc.

Given our experience with dropping the popout styles (which we also believed to be low/minimal usage) and comments raised in #3635 about the lack of notification/communication, I definitely think this would be necessary

PyvesB:
> 
> 
> > We could create a new issue with a basic thumbs up/down voting system, ask for feedback on Discord, create a Twitter poll, etc.
> 
> Given our experience with dropping the popout styles (which we also believed to be low/minimal usage) and comments raised in #3635 about the lack of notification/communication, I definitely think this would be good (and necessary!) part of dropping raster support

Done all three:
* pinned issue with a new label type, `user-poll`: #4564
* Twitter poll: https://twitter.com/Shields_io/status/1220459422171959297
* poll in the Discord support channel, using [Poll-Bot](https://github.com/stayingqold/Poll-Bot).

Issue #929, opened 2017-04-03
Repology badges


AMDmi3:
Repology.org tracks package versions across a lot of package repositories.

It supports badges with number of repositories which package a given project (examples below for Firefox):

![Number of packages for Firefox](https://repology.org/badge/tiny-repos/firefox.svg)

And per-repository badges with version (of given project package in given repository):

![Arch package version](https://repology.org/badge/version-for-repo/arch/firefox.svg)
![FreeBSD port version](https://repology.org/badge/version-for-repo/freebsd/firefox.svg)
![Ubuntu Zesty package version](https://repology.org/badge/version-for-repo/ubuntu_zesty/firefox.svg)

Would be nice to support these on shields.io.

PS. It also supports a big badge with packaging summary in all known repos, but I guess that's not exactly sields.io format.

![Packaging status](https://repology.org/badge/vertical-allrepos/firefox.svg)

paulmelnikow:
Hi, thanks for bringing this up. Do they have an API?

AMDmi3:
Yes, https://repology.org/api/v1

paulmelnikow:
Great. Would you like to make a PR? I could make some design suggestions and give you pointers if needed.

AMDmi3:
Not really, sorry. This is only a suggestion.

avylove:
Why was this closed? Was it integrated? I don't see anything on shields.io

AMDmi3:
I don't see neither interest nor a real need in this. Repology has its own badges anyway.

avylove:
The main reasons are to have the formatting consistency and caching provided by shields.io. Many of the projects on shields.io have their own badges, but it's often preferable to use the ones provided by shields.io for these reasons.

I did start to look into this. I think some basic badges can be easily incorporated, but the api for Repology currently doesn't seem to have support for querying a single repo for a specific package (all repos are always returned) and there doesn't seem to be a way to resolve repo identifiers like 'epel_7' to repo names like 'EPEL 7'. This can be worked around, but support from the api would be a big help.

PyvesB:
Just to clarify the status here: we now support the number of repositories badge, but are still lacking the versions badges.

zero77:
The also have:
https://repology.org/badge/latest-versions/firefox.svg

PR #4529, opened 2020-01-14
Add [Repology] repositories badge

matchai:
This PR implements a badge that is functionally equivalent to the Repology "Tiny badge with number of repositories", as they call it:

[Packaging status](https://repology.org/project/starship/versions)

This PR partly addresses #929.

Originally I tried using the [Repology API](https://repology.org/api), but realized that the badge displays the number of repository families that a project is in, rather than simply the number of repositories it is in.
> Repository family is an aggregation of related repositories, e.g. Debian Stable, Testing, Unstable and Experimental all belong to a single family and are thus only counted once.

The API doesn't provide any information about the repository families that a repository is in, so this can't be achieved with the API alone.

This is my first time contributing to the project, so any feedback is greatly appreciated! 🙏
Once this PR is merged in, I'll try to tackle some of the other Repology badges. 😊


services/repology/repology-repositories.service.js
```diff
@@ -0,0 +1,52 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { BaseSvgScrapingService } = require('..')
+
+const schema = Joi.object({
+  message: Joi.number(),
```
calebcartwright: We've got some helpers for common schema fields like this one (non-negative number) that we can use here. These reside in [validators](https://github.com/badges/shields/blob/master/services/validators.js)

Just import the validator
```js
```
and then change the schema to use it
```suggestion
  message: nonNegativeInteger,
```

services/repology/repology-repositories.service.js
```diff
@@ -0,0 +1,52 @@
...
+    return {
+      message: repositoryCount,
+      color: 'blue',
+    }
+  }
+
+  async handle({ projectName }) {
+    const { message: repositoryCount } = await this._requestSvg({
+      schema,
+      url: `https://repology.org/badge/tiny-repos/${projectName}.svg`,
```
calebcartwright: what's the response from repology when the specified project doesn't exist/isn't found? HTTP 404 or something else?

services/repology/repology-repositories.tester.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Existing project')
+  .get('/starship')
+  .expectBadge({
+    label: 'in repositories',
+    message: Joi.number().greater(0),
```
calebcartwright: Can also use the same validator here on the schema

services/repology/repology-repositories.service.js
matchai: It simply provides a badge with 0 repositories.

services/repology/repology-repositories.spec.js
```diff
@@ -0,0 +1,11 @@
...
+
+const { test, given } = require('sazerac')
+const RepologyRepositories = require('./repology-repositories.service')
+
+describe('Repology repositories count badge', function() {
+  test(RepologyRepositories.render, () => {
+    given({ repositoryCount: 0 }).expect({ message: 0, color: 'blue' })
+    given({ repositoryCount: 18 }).expect({ message: 18, color: 'blue' })
+  })
+})
```
calebcartwright: We can remove these altogether. For the classes that integrate with the upstream services/platforms, we typically only add unit/function level tests like these for transformation (parsing/extraction) and rendering logic which we fortunately don't have here!

services/repology/repology-repositories.service.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Repology - Repositories',
+        namedParams: { projectName: 'starship' },
+        staticPreview: this.render({ repositoryCount: '18' }),
+        keywords: ['repositories'],
```
calebcartwright: All words in the example title are automatically searchable keywords, so you can remove the `keywords` key from here (unless there's other search keywords you think could be helpful)

services/repology/repology-repositories.service.js
```diff
@@ -0,0 +1,53 @@
...
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'in repositories' }
+  }
+
+  static render({ repositoryCount }) {
+    return {
+      message: repositoryCount,
```
calebcartwright: By any chance do you know what the realistic ceiling would be for values we could see? Shields uses a common set of logic for metric/numerical values in the rendering (for example 1,045 would be rendered as 1k) and I'm thinking it'd probably be best to go ahead and use taht same formatter here.

Similar to the validators for schemas, we've got a lot of helpers for various formatting scenarios. I believe you'd want to use `metric` here from https://github.com/badges/shields/blob/master/services/text-formatters.js

```suggestion
      message: metric(repositoryCount),
```

services/repology/repology-repositories.service.js
```diff
@@ -0,0 +1,53 @@
...
+      color: 'blue',
```
calebcartwright: and if the color is going to be static, you can move this up to `defaultBadgeData`

services/repology/repology-repositories.service.js
matchai: The most I've seen is 55, but I think it'd best to be safe and use it. 👍

services/repology/repology-repositories.tester.js
```diff
@@ -0,0 +1,32 @@
...
+  })
+
+t.create('Non-existent project')
+  .get('/invalidprojectthatshouldnotexist.json')
+  .expectBadge({
+    label: 'in repositories',
+    message: '0',
+  })
+
+t.create('Project with repositories')
```
calebcartwright: Last item 😄 We can get rid of this test too (and the mocked svg). the svg parsing logic is well tested elsewhere in the codebase already, and if repology were to make any changes to their svg structure then the other two tests would fail anyway

services/repology/repology-repositories.service.js
```diff
@@ -0,0 +1,55 @@
...
+        title: 'Repology - Repositories',
+        namedParams: { projectName: 'starship' },
+        staticPreview: this.render({ repositoryCount: '18' }),
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'in repositories',
```
calebcartwright: this is probably okay if you feel strongly about it, but i'd suggest just using `repositories` instead to be more consistent with other Shields badges (understand that repology use `in repositories` in their own badges). 

folks can always use the `?label=...` query param with Shields badges to change it to their liking, but just about all of our badges have single word messages/don't include prepositions

matchai:
Thank you for taking the time to patiently review my PR. 🙏😊

calebcartwright:
> Thank you for taking the time to patiently review my PR 🙏😊

My pleasure! Really appreciate the PR. 

The CI failure is spurious, but will give it one-more re-run to see 🤞

PR #4553, opened 2020-01-17
Add [MicroBadger] version/docker tag badge

### LINKED ISSUE #4530: Microbadger Versions/Docker Tag
:clipboard: **Description**

This badge is for MicroBadger.
The information it should show is the version which is effectively the Docker tag.
An example can be found at: https://images.microbadger.com/badges/version/alpine.svg

:link: **Data**

There is a public API located at https://api.microbadger.com this is the same API that is currently utilised for the existing MicroBadger size based shields (layers and size). Documentation can be found [here](https://microbadger.com/api).

:microphone: **Motivation**

As far as I can tell there's no Docker tag based shield, given that MicroBadger has already been implemented for the other shields I figured this should be low hanging fruit to provide a shield which can be customised for Docker tags.

nightah:
This badge provides the ability to look up Docker tags via the MicroBadger API.

I've tested my changes manually against all the MicroBadger badges, I can confirm both the new badge and existing ones are working as intended.


calebcartwright:
Thanks for the PR @nightah! Can you took a look at some of the CI failures? The integration one is spurious and can be ignored, but the other jobs are failing due to formatting issues (you can resolve this by running `npm run prettier`) and it looks like a lot of the service tests are failing for MicroBadger with these changes:

https://circleci.com/gh/badges/shields/86527?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link 

You can run that same subset of MicroBadger service tests locally via:
`npm run test:services -- --only=microbadger`

nightah:
So even if I revert my PR it looks like some of the tests fail.
Seems there's some issue(s) outside of just this PR.

calebcartwright:
> So even if I revert my PR it looks like some of the tests fail.
Seems there's some issue(s) outside of just this PR.

Everything is passing for me with the latest on the master branch, but part of what you are seeing is likely reflective of some of the challenges that @PyvesB was referring to.

At this point I also have some reservations about adding more badges that rely on MicroBadger, and I think the better course of action would be to pursue the Docker hub api to provide version/tag badges for images

nightah:
> At this point I also have some reservations about adding more badges that rely on MicroBadger, and I think the better course of action would be to pursue the Docker hub api to provide version/tag badges for images

Again, I believe this is purely related to the size badges which I believe is quite well documented.

I've sorted out the tests and rebased/submitted. I don't disagree with you that pursuing the Docker API directly for the tag and size badges makes the most sense, it's just outside of my expertise.

The version badge in my experience is actually the most stable of the three, but as the owners, please address as you see fit, I will respect your wishes either way.

Thanks again!

calebcartwright:
> Again, I believe this is purely related to the size badges which I believe is quite well documented.

Do you know if MicroBadger is still supported? The broader concern I have is that it doesn't seem to be an actively developed/maintained project based on the repo activity, maintainer comments, etc.

nightah:
I believe the maintainers have said that it's a side project now, so very much best efforts per the linked issue: https://github.com/microscaling/microbadger/issues/38.

calebcartwright:
@nightah - by any chance would you have any interest/bandwidth in an implementation of Docker version badges that leverages the Docker hub APIs?

No worries if not! It'd be fantastic to add a Docker image version/tag badge to Shields and having the two different implementations (this one + Docker APIs) to compare would be great. 

Either way, thank you again for the PR and the research you've already done on the API side!

nightah:
@calebcartwright I guess the key difference with Microbadger vs DockerHub is that the former performs all the logic to determine the `latest` tag and all of the associated matching tags to present accordingly in the `LatestVersion` field.
Then when looking up a specific tag, for example `alpine:3.10` it also determines based on the `SemVer` what the last minor patch on said specified version is which in this case is `3.10.3`.

Also it's worth keeping in mind that there's no native way via DockerHub (at least that I can see or is documented) to be able to determine number of layers. So MicroBadger might stay in the interim for that alone.

I can probably look to start something, but it might take a little while. Somebody much more experienced can probably knock this off pretty quickly!

calebcartwright:
> Also it's worth keeping in mind that there's no native way via DockerHub (at least that I can see or is documented) to be able to determine number of layers. So MicroBadger might stay in the interim for that alone.

At this time we're not talking about any changes to the existing MicroBadger badges. It's only about adding support for Docker version abdges within Shields, and a question about the best way to do that.

nightah:
@calebcartwright I understand.

I have also submitted another PR #4562 to work alongside, or replace the MicroBadger size badge.
As previously mentioned that badge is specifically the one that suffers from the metadata issues.

Can I suggest that we utilise that for Docker image size and we can continue to utilise the layer badge and the versions as proposed within this PR. I'm suggesting this primarily because there's a bit of logic for the matching versions that MicroBadger performs in the backend that I will probably struggle to replicate.

nightah:
@calebcartwright I've submitted a DockerHub variant of the versions/tags badge in #4562.
Happy for you to take your pick here, I've written a number of tests which should cover the cases and tested extensively across a number of library and user based repos.

Please let me know if you have any questions.

PyvesB:
> I'm suggesting this primarily because there's a bit of logic for the matching versions that MicroBadger performs in the backend that I will probably struggle to replicate.

We've got quite quite a few helper functions around our codebase that do more or less clever version logic. I'm not familiar with the DockerHub API and am not yet clear on the exact problem we would have to solve, but I'm wondering whether we could reuse one of our existing functions!

PyvesB:
Ignore me - you've probably found a solution in the other aforementioned PR. 😄

nightah:
Yep you’re right.

I still think that maybe having both badges could make sense because of the limitations with the Docker API.

Let’s discuss in the other PR and agree on the approach.

Issue #3926, opened 2019-08-27
NuGet version badges use full version including semver build metadata


kzu:
- [x] [shields.io](https://shields.io/#/)

:beetle: **Description**

NuGet version badges use the full version of the package, which makes for quite an ugly long badge. In [SemVer 2.0](https://semver.org), the build metadata is never used (to filter or otherwise distinguish versions), so it can be safely ignored. 

:link: **Link to the badge**

![AutoCodeFix](https://img.shields.io/nuget/vpre/AutoCodeFix.svg)

:bulb: **Possible Solution**

If a `+` is found in the version string, just ignore it and the remaining characters when rendering the badge.

kzu:
Got it! Yep, agreed that dropping if by default doesn't make sense, and I
also personally think the build metadata is starting to become more common
in nuget.org, but it's not widespread yet either.

Thanks!

PR #4556, opened 2020-01-19
[nuget myget] defer decision on which version is latest to registry

### LINKED ISSUE #4494: Shields.io badge doesn't show last package from Nuget.org

- [ ] My own instance

:beetle: **Description**

Hi, we are using  shields.io in our project to point to a nuget.org package , but it seems it stopped showing our latest package 4.4.0.991265

:link: https://img.shields.io/nuget/v/Xamarin.Forms.svg?style=flat-square&label=nuget

https://www.nuget.org/packages/Xamarin.Forms/

 [NuGet](https://www.nuget.org/packages/Xamarin.Forms/)

should show  4.4.0.991265

:bulb: **Possible Solution**

chris48s:
Closes #4494 Discussion in #4494 + on discord

tl;dr summary:
We're trying to sort nuget versions as semver. While nuget _advises_ users to use semver2, it doesn't enforce it and there's a bunch of stuff in the ecosystem that uses other versioning schemes. The nuget registry itself has some logic to apply some ordering to versions which we can't easily reproduce and users (completely reasonably) expect shields to report the same latest version that the registry does.

### DIFF
--- a/services/nuget/nuget-v3-service-family.js
+++ b/services/nuget/nuget-v3-service-family.js
@@ -2,7 +2,6 @@
 
 const { promisify } = require('util')
 const Joi = require('@hapi/joi')
-const semver = require('semver')
 const { regularUpdate } = require('../../core/legacy/regular-update')
 const RouteBuilder = require('../route-builder')
 const { renderVersionBadge, renderDownloadBadge } = require('./nuget-helpers')
@@ -82,15 +81,9 @@ const schema = Joi.object({
   data: Joi.array()
     .items(
       Joi.object({
-        versions: Joi.array()
-          .items(
-            Joi.object({
-              version: Joi.string().required(),
-            })
-          )
-          .default([]),
         totalDownloads: Joi.number().integer(),
         totaldownloads: Joi.number().integer(),
+        version: Joi.string().required(),
       })
     )
     .max(1)
@@ -104,14 +97,14 @@ async function fetch(
   serviceInstance,
   { baseUrl, packageName, includePrereleases = false }
 ) {
+  const prerelease = includePrereleases ? 'true' : 'false'
   const json = await serviceInstance._requestJson({
     schema,
     url: await searchQueryServiceUrl(baseUrl),
     options: {
       qs: {
         q: `packageid:${encodeURIComponent(packageName.toLowerCase())}`,
-        // Include prerelease versions.
-        prerelease: 'true',
+        prerelease,
         // Include packages with SemVer 2 version numbers.
         semVerLevel: '2',
       },
@@ -184,21 +177,26 @@ function createServiceFamily({
         withFeed,
         feed,
       })
-      const { versions } = await fetch(this, { baseUrl, packageName })
-      let latest = versions.slice(-1).pop()
       const includePrereleases = which === 'vpre'
-      if (!includePrereleases) {
-        const filtered = versions.filter(item => {
-          if (semver.valid(item.version)) {
-            return !semver.prerelease(item.version)
-          }
-          return !item.version.includes('-')
+      let data
+      try {
+        data = await fetch(this, {
+          baseUrl,
+          packageName,
+          includePrereleases,
         })
-        if (filtered.length) {
-          latest = filtered.slice(-1).pop()
+      } catch (e) {
+        if (e instanceof NotFound && !includePrereleases) {
+          data = await fetch(this, {
+            baseUrl,
+            packageName,
+            includePrereleases: true,
+          })
+        } else {
+          throw e
         }
       }
-      const { version } = latest
+      const version = data.version
       return this.constructor.render({ version, feed })
     }
   }
@@ -232,7 +230,11 @@ function createServiceFamily({
         withFeed,
         feed,
       })
-      const packageInfo = await fetch(this, { baseUrl, packageName })
+      const packageInfo = await fetch(this, {
+        baseUrl,
+        packageName,
+        includePrereleases: true,
+      })
 
       // Official NuGet server uses "totalDownloads" whereas MyGet uses
       // "totaldownloads" (lowercase D). Ugh.
--- a/services/myget/myget.tester.js
+++ b/services/myget/myget.tester.js
@@ -70,7 +70,7 @@ t.create('version (valid)')
     message: isVPlusDottedVersionNClausesWithOptionalSuffix,
   })
 
-t.create('total downloads (tenant)')
+t.create('version (no stable versions)')
   .get('/dotnet.myget/dotnet-coreclr/v/Microsoft.DotNet.CoreCLR.json')
   .expectBadge({
     label: 'dotnet-coreclr',
@@ -87,7 +87,7 @@ t.create('version (yellow badge)')
   .intercept(nock =>
     nock('https://api-v2v3search-0.nuget.org')
       .get(
-        '/query?q=packageid%3Amongodb.driver.core&prerelease=true&semVerLevel=2'
+        '/query?q=packageid%3Amongodb.driver.core&prerelease=false&semVerLevel=2'
       )
       .reply(200, nuGetV3VersionJsonWithDash)
   )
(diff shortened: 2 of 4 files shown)


services/nuget/nuget-v3-service-family.js
```diff
@@ -184,21 +177,26 @@ function createServiceFamily({
...
+      } catch (e) {
+        if (e instanceof NotFound && !includePrereleases) {
+          data = await fetch(this, {
+            baseUrl,
+            packageName,
+            includePrereleases: true,
+          })
+        } else {
+          throw e
         }
```
chris48s: The downside of this approach is that if a user requests a `/v` badge and there are no stable releases, we throw a `NotFound`. Normally if there are no stable versions, but the user requests a `/v` badge, we just show the latest unstable version. This is the case for other version badges on shields and its also what we've done in the past for nuget/myget. In order to reproduce that behaviour (and avoid making a breaking change), we now need to make 2 API requests. This means in the case where we're requesting `/v` but there aren't any stable versions published, this is slower to get the data. Its probably inside the critical 4 second window, but its definitely noticeably slower.

The other thing we could do here is throw an error in the case where the user requests a `/v` badge and there are no stable releases, but its a breaking change and inconsistent with what we do for other services.

I suspect this problem may be one of the reasons we decided to try and decide ourselves on what is the latest version ourselves in the first place instead of just deferring to the registry's opinion in the first place?

calebcartwright:
Thanks Chris! 

Just to add some additional clarifications on the tl;dr
1. this is explicitly about the determination of whether a package version is a release version or pre-release version 
2. `users (completely reasonably) expect shields to report the same latest version that the registry does.` 
the nuget.org registry (and perhaps others) also drop metadata-like portions of versions in the UI, whereas Shields presents the exact version we get from the nuget APIs (refs #3926). This is not addressed in this PR, so even with this change Shields would still not be displaying the exact same version values that the nuget.org UI displays

services/nuget/nuget-v3-service-family.js
calebcartwright: > its a breaking change and inconsistent with what we do for other services.

Definitely agreed, though this is already a potentially breaking change anyway isn't it?

My sense is that nuget, and accordingly how we handle nuget-related badges, just seems to be different than the other registries we integrate with. We use different rendering logic/color schemes, there's an undocumented/uncommon logic in version processing on the registry side, etc. 

Perhaps the question ultimately boils down to: should we try to treat nuget like all the other package/artifact type registries that we integrate with to be more consistent with our other package/artifact version badges, or do we handle it differently to account for all of its unique aspects to be more consistent with the upstream provider and the experience/expectations of users.

It seems that our current state was an attempt to satisfy both which is likely why we're having these issues. IMO any change we make is going to cause some amount of a breaking change, so I'd prefer we decide which way we want to go (consistency with nuget or consistency with our other badges), and communicate/publicize any associated changes.

chris48s: > Definitely agreed, though this is already a potentially breaking change anyway isn't it?

I think the behaviour change as proposed just brings us into line with user expectations and is best just viewed as a bugfix. I'd be surprised if anyone is knowingly reliant on our sorting being different than nuget's.

> My sense is that nuget, and accordingly how we handle nuget-related badges, just seems to be different than the other registries we integrate with.

Meh.. kinda. I broadly view it as being like PyPI. We don't try and maintain our own implementation of PEP 440. We just say "hey PyPI - what's the latest (pre-)release of this package?" and report what the registry tells us. If the shields badge said a different thing than the registry, that would be unexpected. We should do the same thing for nuget.

> should we try to treat nuget like all the other package/artifact type registries that we integrate with to be more consistent with our other package/artifact version badges, or do we handle it differently to account for all of its unique aspects to be more consistent with the upstream provider and the experience/expectations of users.

In general, our badges should reflect the norms of the community they serve, but I see the decision on what to do in this case as a practicality relating to performance/render time. The fact that we would have to make 2 HTTP requests in this situation is a consequence of the (clunky) API implementation as opposed to the norms of nuget as a packaging community or a desire for consistency across registries.

This has all got a bit philosophical tbh :smile:

calebcartwright: > This has all got a bit philosophical tbh 😄

I posed the question, with some context around my thinking, because it directly impacts how I'd respond to your inline comment/question. My intent was not to philosophize anything, so a more explicit attempt below.

>  and inconsistent with what we do for other services

^ My main point is that I don't think that neccessarily matters since there's other aspects of the nuget version badge that are already inconsistent with the other version badges. 

I still think it would be better to make a decision on how to handle the nuget version badges holistically with specific actions (I understand most of these are outside the scope of this PR, but enumerating for completeness)

Option 1 - consistent with the nuget registries
1. Update our rendering logic to automatically strip all build metadata from the version values, just like the nuget registries UI and clients do
    * #3926 
2. Keep the unique 3-color rendering logic for our badge
3. Defer to the registry to make pre-rel vs. rel determination on versions
4. For packages with no release versions, we should feel free to throw an error on the `/v` route if there's concerns that attempting to fallback to pre-release versions could lead to timeouts; we should not feel obligated to mimic this aspect of the other version badges, though it'd be great if we could.

Option 2 - consistent with the other version badges
1. Display the raw version values the registry provides in the API response
2. Drop the custom rendering logic and use the same version badge rendering that all the other badges do
3. Fallback to pre-release versions on the `/v` route if there's no release versions
4. Defer to the registry to make pre-rel vs. rel determination, unless that risks (performance or otherwise) the ability to satisfy the third bullet to fallback to pre-release versions. 
    * If the fallback pattern can't be supported by deferring to the registry, then we should attempt to mimic the logic ourselves to ensure the fallback _can_ be supported

I still feel like the current state of our nuget version badges is selectively consistent, and I'd still prefer we pick one or the other and be fully consistent with one or the other.

chris48s: > For packages with no release versions, we should feel free to throw an error on the `/v` route if there's concerns that attempting to fallback to pre-release versions could lead to timeouts; we should not feel obligated to mimic this aspect of the other version badges, though it'd be great if we could.

I think if we were starting from a position of "we don't have a badge for nuget and we're adding one", I'd be on board with that, but bear in mind that the position we're starting from is that we already have a lot of nuget badges in the wild (nuget is one of the more popular services on shields - we're serving over 100,000 requests a day just on the nuget version badge) and this change would take a lot of badges like this one:

https://img.shields.io/nuget/v/EasyCaching.Core.svg

and replace them with an error. Even if this is the decision we'd make if we were starting from scratch, we do need to take that into account.

Maybe we should just try and patch up our existing implementation if the performance hit is too much?


> Update our rendering logic to automatically strip all build metadata from the version values the registry API gives us

In terms of dropping the metadata portion of the version, is it always seperated by a `+` i.e: `version+metadata`, or is this another one where its complicated/we don't really know? There are some [docs on versioning](https://docs.microsoft.com/en-us/nuget/create-packages/prerelease-packages), but build metadata isn;t mentioned.

calebcartwright: > we're serving over 100,000 requests a day just on the nuget version badge) and this change would take a lot of badges like this one
> and replace them with an error.

Understood, but especially at scale, we could also be replacing a chunk of those existing/working `/v` badges with timeouts instead, especially if a large portion of those requests are for packages that don't have any release versions. 

While I understand that we wouldn't want to make a change that replaced working badges with an error overnight/without warning, I do want to pose the general question of whether we could make such a change after sufficient communication to users (apologies for the semi-OT question, probably best discussed in Discord)

> Maybe we should just try and patch up our existing implementation if the performance hit is too much?

That's probably the safest approach if fully maintaining the status quo for the forseeable future is required (pre-rel fallback, no timeouts/errors, etc.).

I wish we had a way to do some sort of load testing to get a better feel of how likely the change was to cause timeouts in prod, or even some generic telemetry like the percentage of `/v` version badges that fell back to a pre-release version (which I don't think would violate the no-tracking promise at all).

We could give this approach a go for a trial period to see to what degree, if any, it causes timeouts, and then revert back to patching the existing implementation if the timeouts are pervasive. 

The actual code changes here look fine so if you think the odds of timeouts are low then I can give the PR a 👍

chris48s: I reckon we try this first: #4582
If we are still getting issues raised with that approach, lets revisit it.

PR #4582, opened 2020-01-25
always strip build metadata from [nuget myget] versions

### LINKED ISSUE #3926: NuGet version badges use full version including semver build metadata

### LINKED ISSUE #4494: Shields.io badge doesn't show last package from Nuget.org

chris48s:
Here's another approach to #4494. Nuget is sufficiently opaque that I don't really know if this is a rock solid assumption, but given the problems discussed in https://github.com/badges/shields/pull/4556#discussion_r368289174 I reckon we give this a approach a try first.

closes #4494
closes #3926


chris48s:
Cheers for the review :+1:  I'm going to leave this one and see if @calebcartwright has any thoughts before I merge.

calebcartwright:
The code changes LGTM, just a few clarifying questions:

* This change means there are some set of users that will see a different badge value after deployment, even though there's been no change to their nuget package. Are we considering that change to be more of a bug fix/attempt to align the version values with nuget.org vs. a "breaking change"?
* If there are current users of our nuget version badge that _want_ the build metadata in their badge, do we have any thoughts on what we'd tell them/how we'd respond to those folks after this change is deployed?

chris48s:
1. Yes
2. Good question.. Do you think we should ignore the metadata part for the purposes of determining whether its a pre-release or not, but then make whether we show it on the badge or not configurable via a flag? Nuget itself does seem to _mostly_ present it without and it seems like its not necessary to use the full version (with metadata portion) on the console or in your manifest.

calebcartwright:
> Do you think we should ignore the metadata part for the purposes of determining whether its a pre-release or not, but then make whether we show it on the badge or not configurable via a flag

Yes I think that's probably what we'll need, and was my proposed solution for metadata #3926, albeit with the flag used to opt-out of having the metadata included in the value displayed on the badge to maintain the current state.

However, given the spirit of this PR to have better alignment with users expectations, I'd suggest we proceed with this as-is so that the default version value on our badges best matches the value on the registries (at least as far as we can tell). If and when a user comes to us and wants the metadata bits displayed, we could add a flag at that time to allow them to opt-in to that behavior

PR #4583, opened 2020-01-25
[AzureDevOpsBuild] Support badge on a specific stage or job

Thilas:
Per [Azure DevOps documentation](https://docs.microsoft.com/en-us/rest/api/vsts/build/status/get) on builds, here are some changes that allow to create badges for specific stages and/or jobs.


services/azure-devops/azure-devops-build.service.js
```diff
@@ -58,22 +65,60 @@ module.exports = class AzureDevOpsBuild extends BaseSvgScrapingService {
...
+      {
+        title: 'Azure DevOps builds (stage)',
+        pattern:
+          ':organization/:projectId/:definitionId/:branch*\\?stage=:stage',
+        namedParams: {
+          organization: 'totodem',
+          projectId: 'Shields.io',
+          definitionId: '5',
+          branch: 'master',
+          stage: 'Successful Stage',
```
PyvesB: You can include it inside `queryParams` instead, and simplify the pattern. See [this GitHub badge](https://github.com/badges/shields/blob/0f332413dfc4bfd4feeca7088749c4949f9e980f/services/github/github-release.service.js#L67) for example.

services/azure-devops/azure-devops-build.service.js
```diff
@@ -46,7 +53,7 @@ module.exports = class AzureDevOpsBuild extends BaseSvgScrapingService {
         pattern: ':organization/:projectId/:definitionId',
         namedParams: {
           organization: 'totodem',
-          projectId: '8cf3ec0e-d0c2-4fcd-8206-ad204f254a96',
+          projectId: 'Shields.io',
```
PyvesB: Could you please explain this change? I'm not familiar with how this service works.

services/azure-devops/azure-devops-build.service.js
```diff
@@ -58,22 +65,60 @@ module.exports = class AzureDevOpsBuild extends BaseSvgScrapingService {
...
       },
+      {
+        title: 'Azure DevOps builds (stage)',
+        pattern:
+          ':organization/:projectId/:definitionId/:branch*\\?stage=:stage',
+        namedParams: {
+          organization: 'totodem',
+          projectId: 'Shields.io',
+          definitionId: '5',
+          branch: 'master',
```
PyvesB: Maybe it's worth dropping the `branch` for these two new examples?

services/azure-devops/azure-devops-build.service.js
```diff
@@ -58,22 +65,60 @@ module.exports = class AzureDevOpsBuild extends BaseSvgScrapingService {
...
+    { stage, job }
+  ) {
     // Microsoft documentation: https://docs.microsoft.com/en-us/rest/api/vsts/build/status/get
     const { status } = await fetch(this, {
       url: `https://dev.azure.com/${organization}/${projectId}/_apis/build/status/${definitionId}`,
-      qs: { branchName: branch },
+      qs: {
+        branchName: branch,
+        stageName: stage,
+        jobName: job,
```
PyvesB: Given that `stageName` must be specified in order to use `jobName`, maybe it's worth validating against that? The Joi schema can probably support this kind of logic, but an explicit check throwing a `InvalidParameter` with a message along the lines of "stage must be specified with job" would do the trick as well!

services/azure-devops/azure-devops-build.service.js
Thilas: Good point: I noticed that Azure DevOps was now accepting the project name for Builds whereas it used to expect a project id. But I guess we should probably stick to the project id, especially because on Releases, Azure DevOps still require a project id. I'll rollback that part.

services/azure-devops/azure-devops-build.service.js
Thilas: I hesitated to keep branch or not... But if you too think that it would be better, let's drop it indeed.

services/azure-devops/azure-devops-build.service.js
Thilas: There is nothing preventing from getting a badge for a given job without specifying its stage in Azure DevOps API. I tested it and it works perfectly. That's why I did not implement any validation for that.

services/azure-devops/azure-devops-build.service.js
Thilas: That seems indeed way better (I didn't know about it!).

Thilas:
@PyvesB, thanks for the quick reply and for your support! I just pushed some changes according to your comments.

services/azure-devops/azure-devops-build.service.js
PyvesB: You're right, I had read the documentation a bit too fast! You only have to specify stage when using job if there a more than one stages: "If there are more than one, then it is required to pass in a stageName value when specifying a jobName".

PR #4584, opened 2020-01-25
Yo dawg, I heard you like bots 🤖

chris48s:
This PR is an attempt to reduce the manual overhead of reviewing weekly dependabot PRs.

Dependabot has an [automerged_updates](https://dependabot.com/docs/config-file/#automerged_updates) feature which allows you to specify dependencies or patterns which will be automatically merged if the CI build passes. In general, this is quite a useful feature, and I'd like to use it to auto-merge some testing/linting dependencies, but unfortunately we can't use it out-of-the-box due to our branch protection rules.

So, following the advice in https://github.com/dependabot/feedback/issues/86 and based on the code in https://github.com/hmarr/auto-approve-action I have made a little Github Action: https://github.com/chris48s/approve-bot/blob/master/index.js If this action finds a PR which:

- Was opened by dependabot
- Contains the text "If all status checks pass Dependabot will automatically merge this pull request" in the PR body (this text is added by dependabot if the dependency matches a rule defined in `automerged_updates` )

the action will:

- Approve the PR
- Add the label "squash when passing" (which will then invite repo ranger to do its thing and start updating the branch/merge it).

That will mean we can keep our branch protection rules but allow bots auto-merge a subset of dependencies and hopefully reduce manual review overhead, allowing us to focus on dependencies where a human intervention adds value.

The downside of this is that if we mark a bunch of dependencies as "squash when passing" all at the same time, repo ranger really crunches through build minutes (this is a problem for both heroku and netlify). Doing this is might make that problem worse rather than better, particularly if we get a week with a lot of matching updates :grimacing:


.dependabot/config.yml
```diff
@@ -4,6 +4,31 @@ update_configs:
...
     update_schedule: 'weekly'
+    automerged_updates:
+      - match:
+          dependency_name: 'chai*'
+          update_type: 'semver:minor'
+      - match:
+          dependency_name: 'cypress'
+          update_type: 'semver:minor'
+      - match:
+          dependency_name: 'eslint*'
```
chris48s: Note the wildcards on some of these. For example, this will match `eslint`, but also `eslint-config-prettier`, `eslint-plugin-standard` etc

.dependabot/config.yml
```diff
@@ -4,6 +4,31 @@ update_configs:
...
+          update_type: 'semver:minor'
+      - match:
+          dependency_name: 'sazerac'
+          update_type: 'semver:minor'
+      - match:
+          dependency_name: 'sinon*'
+          update_type: 'semver:minor'
+      - match:
+          dependency_name: 'snap-shot-it'
+          update_type: 'semver:minor'
```
chris48s: In general, I think this is a fairly sensible set of patterns to auto-merge if the tests pass, but obviously I'm open to feedback on it and it can evolve as time goes on..

The only thing I'm really checking when I review PRs for these packages is whether the build is green, so I figure we might as well automate that rule instead of having to do it myself. Major version updates will still require manual review. Even if a bad update does slip the net, only testing/linting tools will break - no production dependencies.

.github/workflows/auto-approve.yml
```diff
@@ -0,0 +1,10 @@
+name: Auto approve
+on: pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: chris48s/approve-bot@1.0.0
```
chris48s: I made this repo under my own namespace because I don't have permission to create new repos under the @badges org, but I'd be happy to transfer it to the @badges org if someone can accept the transfer if we adopt this approach.

.dependabot/config.yml
```diff
@@ -4,6 +4,31 @@ update_configs:
   - package_manager: 'javascript'
     directory: '/'
     update_schedule: 'weekly'
+    automerged_updates:
+      - match:
+          dependency_name: 'chai*'
+          update_type: 'semver:minor'
```
calebcartwright: Is the `update_type` specified here literal or more like a semver ceiling? For example will this only auto-merge on a literal minor version bump, or will it work on both patch _and_ minor?

.dependabot/config.yml
calebcartwright: This seems like a reasonable set to me as well. I still do a quick manual spot check on the review apps even for these dev-only deps, but acknowledge that the E2E test job really covers that scenario anyway.

.dependabot/config.yml
chris48s: My reading of the docs is its the second one: `update_type: 'semver:minor'` should auto-merge patch and minor updates (like the inverse of `@dependabot ignore this minor`).

.dependabot/config.yml
```diff
@@ -4,6 +4,31 @@ update_configs:
   - package_manager: 'javascript'
     directory: '/'
     update_schedule: 'weekly'
```
platan: > The downside of this is that if we mark a bunch of dependencies as "squash when passing" all at the same time, repo ranger really crunches through build minutes (this is a problem for both heroku and netlify). Doing this is might make that problem worse rather than better, particularly if we get a week with a lot of matching updates 😬

Is it possible to update some (automerged) dependencies daily and other dependencies weekly? If not we can update all dependencies daily.

chris48s: I think everything has to be on the same schedule.

I think moving to daily updates is going to give us a higher volume of dependency PRs requiring manual review for stuff like the gatsby* and babel* dependencies which are cranking out multiple releases per week.

calebcartwright: Another thing that will (eventually) be a big help is the ability to have dependency updates grouped together within a single PR - https://github.com/dependabot/feedback/issues/5

chris48s:
Since this is approved, I'm going to go ahead and merge it as I'd like to try it out when dependabot submits a load of PRs tomorrow.

If we want to tweak it some more as time goes on, we can do.

runxel:
Wouldn't be the simple-icons package also a good candidate to auto merge?

PR #4590, opened 2020-01-30
always take worst-cases usesRemaining in token pool

### LINKED ISSUE #4561: Intermittent failures of integration@node-latest build in CI
see https://github.com/badges/shields/pull/4535#issuecomment-575849981 for discussion

example build: https://circleci.com/gh/badges/shields/86544

failing test is:
> Github API provider with token pool should update the token with the final limit remaining and reset time - should update the token with the final limit remaining and reset time

chris48s:
Closes #4561 :tada: 

Having spent some time on this, I discovered the reason for these intermittent failures is because the GitHub API is returning an inconsistent timestamp for `X-RateLimit-Reset`. You should be able to reproduce this behaviour using cURL and a valid GH API token e.g:

```sh
$ curl -I "https://api.github.com/rate_limit" -H "Authorization: Bearer f00b42" | grep X-RateLimit-Reset
X-RateLimit-Reset: 1580421718

$ curl -I "https://api.github.com/rate_limit" -H "Authorization: Bearer f00b42" | grep X-RateLimit-Reset
X-RateLimit-Reset: 1580421718

$ curl -I "https://api.github.com/rate_limit" -H "Authorization: Bearer f00b42" | grep X-RateLimit-Reset
X-RateLimit-Reset: 1580421717

$ curl -I "https://api.github.com/rate_limit" -H "Authorization: Bearer f00b42" | grep X-RateLimit-Reset
X-RateLimit-Reset: 1580421717

$ curl -I "https://api.github.com/rate_limit" -H "Authorization: Bearer f00b42" | grep X-RateLimit-Reset
X-RateLimit-Reset: 1580421718

$ curl -I "https://api.github.com/rate_limit" -H "Authorization: Bearer f00b42" | grep X-RateLimit-Reset
X-RateLimit-Reset: 1580421717

$ curl -I "https://api.github.com/rate_limit" -H "Authorization: Bearer f00b42" | grep X-RateLimit-Reset
X-RateLimit-Reset: 1580421717

$ curl -I "https://api.github.com/rate_limit" -H "Authorization: Bearer f00b42" | grep X-RateLimit-Reset
X-RateLimit-Reset: 1580421718
```

This was causing us to sometimes not decrement `usesRemaining` because ~half the time we assume that the value we already hold is newer than the new value. This probably wasn't causing a _huge_ issue in production because the `X-RateLimit-Reset` seems to fluctuate between 2 values (I assume there's some kind of rounding issue going on upstream) so after a few requests, we catch up and adopt the new `usesRemaining` value, but it is giving us issues under test.

This PR changes the logic to ignore the timestamp and just always take `Math.min(this._usesRemaining, usesRemaining)`, even if we think `this._usesRemaining` is newer than `usesRemaining`. This means we'll always either decrement the token or update `nextReset`. This should get the test passing consistently :crossed_fingers: 

If someone gets a chance to review this before tomorrow and no changes are needed, feel free to just merge it. It would be really useful to get this one in before dependabot opens a pile of PRs tomorrow.

### DIFF
--- a/services/github/github-api-provider.integration.js
+++ b/services/github/github-api-provider.integration.js
@@ -95,7 +95,7 @@ describe('Github API provider', function() {
 
       const [token] = tokens
       expect(token.usesRemaining).to.equal(usesRemaining)
-      expect(token.nextReset).to.equal(nextReset)
+      expect(token.nextReset).to.be.within(nextReset - 1, nextReset)
     })
   })
 })
--- a/services/github/github-api-provider.integration.js
+++ b/services/github/github-api-provider.integration.js
@@ -95,6 +95,14 @@ describe('Github API provider', function() {
 
       const [token] = tokens
       expect(token.usesRemaining).to.equal(usesRemaining)
+
+      /*
+      There is some kind of rounding or off-by-one error
+      which causes the GitHub API to return an inconsistent
+      timestamp value for `X-RateLimit-Reset`.
+
+      see https://github.com/badges/shields/pull/4590 for more details
+      */
       expect(token.nextReset).to.be.within(nextReset - 1, nextReset)
     })
   })
--- a/core/token-pooling/token-pool.js
+++ b/core/token-pooling/token-pool.js
@@ -109,10 +109,8 @@ class Token {
     ) {
       this._nextReset = nextReset
       this._usesRemaining = usesRemaining
-    } else if (nextReset === this._nextReset) {
-      this._usesRemaining = Math.min(this._usesRemaining, usesRemaining)
     } else {
-      // Discard the new update; it's older than the values we have.
+      this._usesRemaining = Math.min(this._usesRemaining, usesRemaining)
     }
   }
 


calebcartwright:
So glad to see you figured this one out! For some reason I thought we were also running integration tests on multiple versions of node and was curious why it was only failing on `latest`, but see now that wasn't the case 👍

services/github/github-api-provider.integration.js
```diff
@@ -95,7 +95,7 @@ describe('Github API provider', function() {
 
       const [token] = tokens
       expect(token.usesRemaining).to.equal(usesRemaining)
-      expect(token.nextReset).to.equal(nextReset)
+      expect(token.nextReset).to.be.within(nextReset - 1, nextReset)
```
paulmelnikow: Can’t say this modified assertion makes a ton of sense as written. Worth adding a comment?

chris48s:
so.. turns out its not an off-by-one:

`AssertionError: expected 1580508686 to be within 1580508684..1580508685`.

I wonder if we should just drop that assertion entirely?

chris48s:
I guess the other option would be to just always set `this._nextReset = nextReset` in the token pool.. any downside to that?

paulmelnikow:
Hmmm, Let me give this a little more thought, and study more carefully your example in the top post. It's making more sense now why this was written the way it was…

chris48s:
Did you get a chance to think about/revisit this at all?

calebcartwright:
Seems like the failing test here has started failing again :thinking: 

https://github.com/badges/shields/pull/4810#issuecomment-605993769
https://circleci.com/gh/badges/shields/96063?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

paulmelnikow:
@chris48s Sorry I missed the last bump. Will take a look again at the substantive question here.

chris48s:
I had another look at this today just to see if GitHub has magically started returning a consistent `X-RateLimit-Reset` value.. it hasn't. Given this has been an issue for a solid 5 months, I think we do need to accommodate this behaviour (even though its annoying).

paulmelnikow:
Apologies for letting this fester so long.

The purpose of the original logic was to correctly handle the case of the rate limit resetting. A typical reset might look like this:

```
1.
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 18
X-RateLimit-Reset: 1580421717

2.
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 59
X-RateLimit-Reset: 1580421777
```

Imagine, however, this sequence of requests:

```
1.
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 18
X-RateLimit-Reset: 1580421717

2.
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 59
X-RateLimit-Reset: 1580421777

3.
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 17
X-RateLimit-Reset: 1580421717
```

What this would suggest to me is the third request's rate limit was evaluated before the reset, while the second request's was evaluated after the reset. I'm not totally sure this happens, though it seems very likely, given that different requests could take different amounts of time to complete on GitHub's side, and similarly different amounts of time to be sent across the wire. We don't even know when during the request the quota is evaluated. We can't guarantee we will process the responses in the same order in which GitHub applied the quota headers.

In that case, I think it's indisputable that we'd want to record the quota information from the second request `{ usesRemaining: 59, nextReset: 1580421777 }` and ignore the third request's obsolete quota information.

If I understand correctly, #4561 is a different problem. Instead of nextReset leaping by the amount of the rate-limit interval, it's wiggling around by small amounts.

```
X-RateLimit-Reset: 1580421717
X-RateLimit-Reset: 1580421718
X-RateLimit-Reset: 1580421718
X-RateLimit-Reset: 1580421717
X-RateLimit-Reset: 1580421717
X-RateLimit-Reset: 1580421718
X-RateLimit-Reset: 1580421717
X-RateLimit-Reset: 1580421717
X-RateLimit-Reset: 1580421718
```

These are all one second differences. However it seems likely, based on 

> `AssertionError: expected 1580508686 to be within 1580508684..1580508685`.

that the variation might be as much as two seconds. (Maybe it's +/- 1 second?)

I guess what I'd suggest is keeping the three conditions of the if-else, but instead of comparing `nextReset` exactly, comparing within a tolerance.

```js
// Since GitHub returns slightly inconsistent values for `X-RateLimit-Reset`, 
// instead of comparing the values directly, we use a tolerance threshold.
// https://github.com/badges/shields/pull/4561
const TOLERANCE_SECONDS = 2

if (
    this._nextReset === this.constructor.nextResetNever ||
    nextReset > this._nextReset + TOLERANCE_SECONDS
) {
    // Either it's the first time we've used the token, or its quota has reset
    // since the last request. Overwrite the information we have.
    this._nextReset = nextReset
    this._usesRemaining = usesRemaining
} else if (nextReset < this._nextReset - TOLERANCE_SECONDS) {
    // On a prior request, this token reset, and we've updated our record.
    // However the request we're currently processing was issued _before_ the
    // reset, so its quota information is outdated. Discard it.
} else {
    // The token has not reset since the last request. We don't know whether the
    // request we're processing was newer or older, so to be conservative, keep
    // the lower value for `usesRemaining`.
    this._usesRemaining = Math.min(this._usesRemaining, usesRemaining)
}
```

chris48s:
I've forgotten basically everything about this, but I tried a bunch of requests to https://api.github.com/rate_limit just now and got.. the exact same value for `X-RateLimit-Reset` on every request. Its possible we've ignored this for long enough that the problem has been fixed upstream (or possibly it is intermittent). We could just switch those tests back on and see if we get failures again. If not, then :tada: 

Failing that, I don't know what a correct value would be for `TOLERANCE_SECONDS` if the issue still exists. Looking back through some old builds from around the time we started having problems with this, the largest differences I can find are:

* `AssertionError: expected 3161 to equal 3157` on https://app.circleci.com/pipelines/github/badges/shields/jobs/88526
* `AssertionError: expected 2553 to equal 2546` on https://app.circleci.com/pipelines/github/badges/shields/jobs/88921

I guess we'd just have to pick a number, turn the tests back on and keep increasing the number until we stop getting failed builds.

chris48s:
Actually, those differences in the linked builds above are discrepancies are in the `usesRemaining` value, not the `nextReset` value, so they don't actually indicate what the value of `TOLERANCE_SECONDS` would need to be.

paulmelnikow:
> Actually, those differences in the linked builds above are discrepancies are in the `usesRemaining` value, not the `nextReset` value, so they don't actually indicate what the value of `TOLERANCE_SECONDS` would need to be.

Oh, interesting. I think CI is sharing a GitHub token with staging, so I guess this can happen either from (1) concurrent integration test runs, or (2) someone using staging to generate GitHub badges.

> I guess we'd just have to pick a number, turn the tests back on and keep increasing the number until we stop getting failed builds.

I think that's right. This threshold is good for keeping the builds green, but is also the value we want to be using in production.

> Its possible we've ignored this for long enough that the problem has been fixed upstream (or possibly it is intermittent). We could just switch those tests back on and see if we get failures again. If not, then 🎉

I'm fine with that. So to recap, you're suggesting: re-enable the tests, see if the problem comes back, and if it does, merge the new logic from https://github.com/badges/shields/pull/4590#issuecomment-703005006?

chris48s:
> you're suggesting: re-enable the tests, see if the problem comes back, and if it does, merge the new logic from https://github.com/badges/shields/pull/4590#issuecomment-703005006 ?

Yeah - exactly :+1:

calebcartwright:
> > you're suggesting: re-enable the tests, see if the problem comes back, and if it does, merge the new logic from [#4590 (comment)](https://github.com/badges/shields/pull/4590#issuecomment-703005006) ?
> 
> Yeah - exactly +1

I'm good with this as well

paulmelnikow:
Sounds like a plan!

PR #4614, opened 2020-02-02
deps: pin `start-server-and-test` to 1.10.7

calebcartwright:
Apparently v1.10.8 of `start-server-and-test` requires Node 10.x+ due to transitive dependencies

https://github.com/badges/shields/pull/4600#issuecomment-581162776


PyvesB:
Slightly worried that one of these problems may sneak into production. 😕

PR #4615, opened 2020-02-03
Added [OPM] version badge

fffonion:
Adds the [Openresty Package Manager](https://opm.openresty.org/) badge.


services/opm/opm.service.js
```diff
@@ -0,0 +1,75 @@
+'use strict'
+
+const { addv } = require('../text-formatters')
+const { BaseService, NotFound, InvalidResponse } = require('..')
+
+module.exports = class Luarocks extends BaseService {
```
calebcartwright: Will need to rename this class as we've already got an existing `Luarocks` service. Maybe `OpmVersion`? It'd probably be best to rename the file to something like `opm-version.service.js` too
https://github.com/badges/shields/blob/d5af9f884b6db2f4bba9c2582123d08a2c97486a/services/luarocks/luarocks.service.js#L17

fffonion: Ah yes that was a typo... I've renamed the file and class name.

fffonion:
Friendly ping on this PR : ) Is there any blockers I can fix at this time?

calebcartwright:
Thanks for the bump @fffonion. No known blockers or issues at this time, just need to get around to reviewing this one in more detail 😄 

Thanks again for the PR, will try to get back to you soon!

services/opm/opm-version.service.js
```diff
@@ -0,0 +1,75 @@
...
+  static get defaultBadgeData() {
+    return {
+      label: 'opm',
+    }
+  }
+
+  static render({ version }) {
+    const color = '#599059'
+
+    return { message: addv(version), color }
```
calebcartwright: Would you explain the thinking for using a fixed color for the badge instead of the `renderVersionBadge` helper?

services/opm/opm-version.service.js
```diff
@@ -0,0 +1,75 @@
...
+      options: {
+        method: 'HEAD',
+        qs: {
+          account: user,
+          name: moduleName,
+        },
+      },
+      errorMessages: {
+        404: 'module not found',
+      },
```
calebcartwright: So I take it there's no available JSON api we could integrate with?

fffonion: Yes for now there's no JSON API available.

services/opm/opm-version.service.js
fffonion: I was copying from luarocks.service and found it has this fixed color behaviour. Is it a better practice to use renderVersionBadge? If so I can make the change.

calebcartwright: In general, I'd always prefer to use our helper functions to ensure that all of our badges within a given category are consistent. However, the Luarocks version badge does use a distinct/uncommon (at least within Shields) color scheme. 

I'm unclear on what the relationship is between OPM and Luarocks TBH, but I do think we should either use the common `renderVersionBadge` helper here, or the same coloring logic that is used for the Luarocks version badges if there's a strong connection/commonality between the two and users of OPM badges would expect the same coloring as the Luarocks badges. 

https://github.com/badges/shields/blob/98a12c97176650a27792749f5137aeba91321966/services/luarocks/luarocks.service.js#L48-L62

fffonion: 👍 I've changed to use `renderVersionBadge` instead.

services/opm/opm-version.service.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  static render({ version }) {
+    const { label: defaultLabel } = this.defaultBadgeData
+    return renderVersionBadge({
+      label: defaultLabel,
+      version,
+    })
+  }
```
calebcartwright: We can remove this function altogether and just leverage the `renderVersionBadge` helper everwhere (details inline elsewhere on how)

services/opm/opm-version.service.js
```diff
@@ -0,0 +1,77 @@
...
+
+  static get examples() {
+    return [
+      {
+        title: 'OPM',
+        namedParams: {
+          user: 'openresty',
+          moduleName: 'lua-resty-lrucache',
+        },
+        staticPreview: this.render({ version: 'v0.08' }),
```
calebcartwright:
```suggestion
        staticPreview: renderVersionBadge({ version: 'v0.08' }),
```

services/opm/opm-version.service.js
```diff
@@ -0,0 +1,77 @@
...
+    if (!version) {
+      throw new InvalidResponse({ prettyMessage: 'version invalid' })
+    }
+    return version
+  }
+
+  async handle({ user, moduleName }) {
+    const version = await this.fetch({ user, moduleName })
+
+    return this.constructor.render({ version })
```
calebcartwright:
```suggestion
    return renderVersionBadge({ version })
```

PR #4617, opened 2020-02-03
tell dependabot not to offer incompatible upgrades

chris48s:
I'm starting to lose track of which dependencies we want to avoid merging.

Until we can sort out our production environment, lets add them all to [ignored_updates](https://dependabot.com/docs/config-file/#ignored_updates). This will stop dependabot from opening PRs to upgrade to these dependency versions. As we find more, we can add additional entires.

Once we get that upgrade done, we can remove them all and get back to these upgrades.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -29,6 +29,28 @@ update_configs:
       - match:
           dependency_name: 'snap-shot-it'
           update_type: 'semver:minor'
+    ignored_updates:
+      - match:
+          dependency_name: cross-env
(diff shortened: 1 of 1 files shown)


calebcartwright:
Great idea! I forget how dependabot handles pinned versions, so wondering if we need to include `start-server-and-test` in the ignore list as well (v1.10.8 sneakily requires Node 10.x via a transitive dep)? 

https://github.com/badges/shields/blob/master/package.json#L253

PR #4659, opened 2020-02-16
[Nexus] 3: set prerelease=false for /r/* badges

### LINKED ISSUE #4640: Nexus 3 "releases" badge searches all repositories

- [x] My own instance

:beetle: **Description**

Building a badge with the "Sonatype Nexus (releases)" UI results in a request to Nexus that doesn't restrict the repository in which the search is performed. This means that it can also return snapshot artifacts.

For example, on my self-hosted instance, badge URL like http://localhost/nexus/r/foo.bar/baz?nexusVersion=3&server=http%3A%2F%2Fnexus.example.com results in `GET /service/rest/v1/search?group=foo.bar&name=baz&sort=version`. As you can see, there is no repository parameter.

:link: **Link to the badge**

This badge returns a snapshot (timestamped) version: https://img.shields.io/nexus/r/org.pentaho.adaptive/daemon?nexusVersion=3&server=https%3A%2F%2Fnexus.pentaho.org

:bulb: **Possible Solution**

I'm not sure what can be done in the general case, and what the Nexus 2 implementation does. From the user side, the workaround is to explicitly specify the repository - however, this is complicated by #4639.

nigredo-tori:
Fixes #4640.

### DIFF
--- a/services/nexus/nexus.tester.js
+++ b/services/nexus/nexus.tester.js
@@ -285,9 +285,11 @@ t.create('Nexus 3 - repository version with query')
 t.create('Nexus 3 - search release version without snapshots')
   .get(
     // Limit the version from above, so that any later artifacts don't break this test.
-    '/r/org.pentaho.adaptive/daemon.json?server=https://nexus.pentaho.org&nexusVersion=3&maven.version=<=9.0.0.0-20180423.155306-4'
+    `/r/org.pentaho.adaptive/daemon.json?server=https://nexus.pentaho.org&nexusVersion=3&queryOpt=${encodeURIComponent(
+      ':maven.baseVersion=<8.1.0.1'
+    )}`
   )
   .expectBadge({
     label: 'nexus',
-    message: 'v8.1.0.6-682',
+    message: 'v8.1.0.0-365',
   })
--- a/services/nexus/nexus.tester.js
+++ b/services/nexus/nexus.tester.js
@@ -281,3 +281,13 @@ t.create('Nexus 3 - repository version with query')
     label: 'nexus',
     message: isVersion,
   })
+
+t.create('Nexus 3 - search release version without snapshots')
+  .get(
+    // Limit the version from above, so that any later artifacts don't break this test.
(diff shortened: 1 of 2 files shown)


services/nexus/nexus.tester.js
```diff
@@ -281,3 +281,13 @@ t.create('Nexus 3 - repository version with query')
     label: 'nexus',
     message: isVersion,
   })
+
+t.create('Nexus 3 - search release version without snapshots')
+  .get(
+    // Limit the version from above, so that any later artifacts don't break this test.
+    '/r/org.pentaho.adaptive/daemon.json?server=https://nexus.pentaho.org&nexusVersion=3&maven.version=<=9.0.0.0-20180423.155306-4'
```
PyvesB: Using `maven.version` like this will be interpreted as a Shields query parameter and will not be forwarded to Nexus. You should use something like this instead:
```
    `/r/org.pentaho.adaptive/daemon.json?server=https://nexus.pentaho.org&nexusVersion=3&queryOpt=${encodeURIComponent(
      ':maven.version=<=9.0.0.0-20180423.155306-4'
    )}`
```

However:
* the `queryParamSchema` pattern may need some tweaking.
* I don't think `maven.version` is a valid parameter (https://nexus.pentaho.org/service/rest/v1/search?group=org.pentaho.adaptive&name=daemon&prerelease=false&maven.version=%3C=9.0.0.0-20180423.155306-4 returns no results). `version` or `maven.baseVersion` may be what we're looking for?

nigredo-tori: You are correct on all counts.
1. `maven.version` is not a valid Nexus 3 parameter. I've changed it to `maven.baseVersion`.
1. Since I had used the last available version as the limit, the test still worked, even though the parameter was ignored. I've lowered the limit, so the most recent versions are not considered (to make sure the limit works), but the latest filtered component is still a snapshot: https://nexus.pentaho.org/service/rest/v1/search?group=org.pentaho.adaptive&name=daemon&sort=version&maven.baseVersion=%3C8.1.0.1
1. I've passed `maven.baseVersion` via `queryOpt`.
1. I've relaxed the value part of the `queryOpt` regex - anything is allowed, as long as it doesn't contain a colon.

Issue #4641, opened 2020-02-12
pkgreview.dev: Average Ratings Badge


KumarAbhirup:
:clipboard: **Description**
Heya! I am the creator of [pkgreview.dev](https://pkgreview.dev)... Launched yesterday. [See this tweet](https://twitter.com/kumar_abhirup/status/1226549898163740673)!

[pkgreview.dev](https://pkgreview.dev) is a Ratings and Reviews website for #NPM packages where people could thank maintainers enough for their work! ⭐️⭐️⭐️

- **Which service is this badge for?** 👉 [pkgreview.dev](https://pkgreview.dev)

- **What sort of information should this badge show?**

   Example Sample Static Badge
   [![Example Badge](https://img.shields.io/badge/pkgreview.dev-%E2%98%85%E2%98%85%E2%98%85%E2%98%85%E2%98%86%20%C2%B7%209%20reviews-%23000?style=for-the-badge&logo=npm)](https://pkgreview.dev/npm/react)

🖼 **pkgreview.dev**

![Landing Thumbnail](https://miro.medium.com/max/3000/1*XqaCZgJzV_lV1KOIS9lLYg.jpeg)

:link: **Data**

- **Is there a public API?**
  Yes.

- **Does the API requires an API key?**
  No.

- **Link to the API documentation.**
  I haven't created documentation for it yet.

  So, I will explain the API here.

  ### **API Endpoint** 🔰 
  ```
  GET https://pkgreview.dev/api/v1/:packageManager/:packageName
  ```

  > Currently, the only `packageManager` it supports is `npm`. Soon, it will support `pip`, `composer` and `docker-hub`.

  ### **Example Responses** 🔰

  #### 1
  Endpoint: `https://pkgreview.dev/api/v1/npm/@withkoji%2Fvcc`
  ```json
  {
    "name": "@withkoji/vcc",
    "type": "npm",
    "reviewsCount": 1,
    "rating": 0.6,
    "icon": "npm",
    "starString": "★★★☆☆"
  }
  ```

  #### 2
  Endpoint: `https://pkgreview.dev/api/v1/npm/react`
  ```json
  {
    "name": "react",
    "type": "npm",
    "reviewsCount": 3,
    "rating": 1,
    "icon": "npm",
    "starString": "★★★★★"
  }
  ```

  #### 3
  Endpoint: `https://pkgreview.dev/api/v1/npm/pkg-that-is-not-available`
  ```json
  {
    "error": true,
[...]

PR #4660, opened 2020-02-16
Add [pkgreview].dev service

KumarAbhirup:
Referring to issue #4641 

In the issue, I and @calebcartwright discussed adding the [pkgreview.dev](https://pkgreview.dev) service.

This is the Pull Request Regarding that.

Currently, draft, as I am fixing a few bugs...

Help appreciated 🙌


KumarAbhirup:
READY FOR REVIEW :)

If you find something wrong or want to change badge labels to more generalized words, you can do that. 😇

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
+'use strict'
```
calebcartwright: We typically (maybe even always) are able to utilize a single class to provide both the rating and star flavors of a badge. This eliminates a lot of the boilerplate code and makes it a bit easier to maintain. We should do the same here.

See a couple other service classes that provide the same types of badges for references:

https://github.com/badges/shields/blob/master/services/visual-studio-marketplace/visual-studio-marketplace-rating.service.js

https://github.com/badges/shields/blob/master/services/spiget/spiget-rating.service.js

KumarAbhirup: Okay! I will take a look and will try to merge them both!

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+
+const Joi = require('@hapi/joi')
+const { BaseJsonService } = require('..')
+
```
calebcartwright: We've got a lot of helper utility that encapsulate functionality common across many badges, and those can (and should) be used here as well to simplify the formatting.

```js
const { starRating, metric } = require('../text-formatters')
const { floorCount } = require('../color-formatters')
```

KumarAbhirup: Okay, so does this mean I would use the `starRating` text-formatter instead of the `starString` provided by the API?

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+const schema = Joi.object({
+  name: Joi.string().required(),
+  type: Joi.string().required(),
+  icon: Joi.string().required(),
+  starString: [Joi.string().optional(), Joi.allow(null)],
+  reviewsCount: Joi.number(),
+  rating: [Joi.number().optional(), Joi.allow(null)],
+}).required()
```
calebcartwright: We use the schemas to ensure the data points that are required to render the badge are there (without having to check individual keys later on), so these don't need to mimic every single entry in the response; just the ones the service class needs.

As such, we can drop every key in the above schema that's not needed (that would be removing everything except `rating` and `reviewsCount`). Also some inline comments above on those two

KumarAbhirup: Ok, would do 🙌

**Question:** I am using `name` and `type` for the badge link, so they need to included in the schema, right?

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+
+const Joi = require('@hapi/joi')
+const { BaseJsonService } = require('..')
+
+const schema = Joi.object({
+  name: Joi.string().required(),
+  type: Joi.string().required(),
+  icon: Joi.string().required(),
+  starString: [Joi.string().optional(), Joi.allow(null)],
+  reviewsCount: Joi.number(),
```
calebcartwright: I would expect that this value will *always* be zero or a positive integer, in which case we can use one of our validators to make that a required field (once the formatter is imported as mentioned above)
```suggestion
  reviewsCount: nonNegativeInteger,
```

KumarAbhirup: Oops, looks like I committed the suggestion without implementing the formatter. Fixing it very soon! Thank you for patience!

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+  rating: [Joi.number().optional(), Joi.allow(null)],
```
calebcartwright: Based on the info you shared in the original issue, this seems like it will always be a decimal number between 0 and 5, so let's reflect that in the schema:

```suggestion
  rating: Joi.number().min(0).max(5).required(),
```

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+      pattern: ':pkgManager/:pkgSlug+',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'pkgreview.dev Package Ratings',
+        namedParams: { pkgManager: 'npm', pkgSlug: 'react' },
+        staticPreview: {
```
calebcartwright: You can make a call to the `render` function here instead of specifying individual keys

KumarAbhirup:
Few PR checks aren't successful it seems, fixing it.

Also, I feel like I might have broken something due to committing a suggestion blindly, fixing it all soon.

Also, you may keep reviewing the code so that I do better in the next review.

Thank you guys for the patience! Really appreciate it! Learning a lot while contributing. I am new here 😅

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+          message: '4/5 (125)',
+          color: '#4F78FE',
+        },
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return {
+      label: 'pkgreview.dev',
```
calebcartwright: Remember that our labels need to be non-promotional and reflect the data point they are conveying. This should be
```suggestion
      label: 'rating',
```

KumarAbhirup: Sure! Doing it!

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+      color: '#4F78FE',
```
calebcartwright: This can be removed. We only really specify a default color when the badges will always be the same color, regardless of the data/content in the API response which is not the case here

services/pkgreview/package-rating.service.js
KumarAbhirup: Makes sense!

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  static render({ type, name, rating, reviewsCount }) {
+    return {
+      label: `pkgreview.dev`,
```
calebcartwright: Can remove this line as the label is already set via `defaultBadgeData`
```suggestion
```

services/pkgreview/package-rating.service.js
KumarAbhirup: Doing it!

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+
+  static get defaultBadgeData() {
+    return {
+      label: 'pkgreview.dev',
+      color: '#4F78FE',
+    }
+  }
+
+  static render({ type, name, rating, reviewsCount }) {
+    return {
```
calebcartwright: We have conventions for service classes around separation of concerns and which function implements which. Currently, the render function is doing a combination of extraction/transformation of the raw data from the API _and_ rendering of the badge which is not what we want. 

At a high level, there's 3 core things that happen from within the `handle` function when a badge is requested:
1. retrieve data from the upstream provider (responsibility of the `fetch` function)
2. transformation/extraction of the relevant data from the response (this is a `transform` action that's either handled within the `handle` function or commonly within a separate `transform` function)
3. rendering (`render` function).

The conversion of raw rating should be extracted from render function accordingly

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+      label: `pkgreview.dev`,
+      message: rating
+        ? `${+parseFloat(rating * 5).toFixed(2)}/5 (${reviewsCount})`
+        : 'Be the first to review',
+      color: '#4F78FE',
```
calebcartwright:
```suggestion
      color: floorCount(averageRating, 2, 3, 4),
```

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  static render({ type, name, rating, reviewsCount }) {
+    return {
+      label: `pkgreview.dev`,
+      message: rating
+        ? `${+parseFloat(rating * 5).toFixed(2)}/5 (${reviewsCount})`
+        : 'Be the first to review',
```
calebcartwright: Please remember the information I shared on the issue around the need for these badges to be consistent with the way Shields handles such badges 😄 We do not inject additional text like this with links to a service/platform (see below)

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,80 @@
...
+      color: '#4F78FE',
+      link: [
```
calebcartwright: Need to remove this as we do not bake in links into the these types of badges

KumarAbhirup: Sure! Removing the link!

services/pkgreview/package-rating.service.js
calebcartwright: Correct, Shields will use the raw, numeric, values from the pkgreview.dev API response

services/pkgreview/package-rating.service.js
calebcartwright: > so they need to included in the schema, right?

No, because the badge link needs to be removed

KumarAbhirup:
Busy for a few days. Will code here soon after I am back ⚡

services/pkgreview/package-rating.service.js
KumarAbhirup: When the package isn't reviewed by anyone, there wouldn't be any rating, so `rating: null` in such case.

KumarAbhirup:
I haven't written nice test cases yet,

but would like to know if other stuff is following the shields.io guidelines.

Writing tests currently. I am not very sure what to test here, would love to get some information about it :)

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,99 @@
...
+
+const pkgReviewColor = colorScale([2, 3, 4])
+
+const schema = Joi.object({
+  rating: [
+    Joi.number()
+      .min(1)
+      .max(5)
+      .optional(),
+    Joi.allow(null),
```
calebcartwright: Shouldn't this actually be a range of 0-1? I forgot that the API response was a percentage. Also, you can add the `.allow(null)` onto the schema chain, doesn't need to be a separate option. Something like:

```js
    rating: Joi.number()
      .min(0)
      .max(1)
      .required()
      .allow(null),
```
I believe that the `optional` chain would allow for a response without the `rating` key to get through our response validation which I don't think is what we want. IIUC, the `rating` key will always be in the response with a value of `null` in the case of a package that's never had any reviews, or a value between 0-1 (inclusive on the upper end).

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,99 @@
...
+
+module.exports = class PkgreviewRating extends BaseJsonService {
+  static get category() {
+    return 'rating'
+  }
+
+  static get route() {
+    return {
+      base: 'pkgreview',
+      pattern: ':format(rating|stars)/:pkgManager/:pkgSlug+',
```
calebcartwright: 👍

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,99 @@
...
+    })
+  }
+
+  transform({ pkgSlug, ...json }) {
+    return {
+      ...json,
+      pkgSlug: encodeURIComponent(pkgSlug),
+    }
+  }
+
```
calebcartwright: `transform` receives the raw response, knows the details of the response, and extracts/transforms the relevant values. I'm guessing what's in here right now is leftover from when you were embedding the link into the badge, but right now this is basically just injecting a key (`pkgSlug`) to the raw response that's not used anywhere else. 

The only real "transformation" that's happening within this service class is the conversion of raw rating percentage to the 5 star scale (I mentioned this above above in the inline comments on the `render` function) so you could probably get rid of this `transform` altogether and just add that one line in handle (see below) 

```suggestion
```

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,99 @@
...
+  async handle({ format, pkgManager, pkgSlug }) {
+    const json = await this.fetch({ pkgManager, pkgSlug })
+
+    const transformedJson = this.transform(json)
+
+    return this.constructor.render({ ...transformedJson, format })
```
calebcartwright:
```suggestion
    const { reviewsCount, rating } = await this.fetch({ pkgManager, pkgSlug })
    return this.constructor.render({ reviewsCount, format , rating: rating * 5 })
```

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,99 @@
...
+      color: pkgReviewColor(rating * 5), // floorCount(rating * 5, 2, 3, 4),
+    }
+  }
+
+  async fetch({ pkgManager, pkgSlug }) {
+    return this._requestJson({
+      schema,
+      url: `https://pkgreview.dev/api/v1/${pkgManager}/${pkgSlug}`,
+      errorMessages: {
+        400: 'package not found',
```
calebcartwright: Does the service ever return a 404 for not founds, or is it always 400?

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,99 @@
...
+        408: 'response timed out',
+        502: 'response timed out',
+        500: 'response timed out',
```
calebcartwright: Can you elaborate a bit on the inclusion of all these response codes, and mapping them to a timed out response? Are you seeing performance/responsiveness issues with the API?

calebcartwright:
I'm also having a lot of difficulty hitting the API right now (getting 500 errors/timeouts/etc.) Is the service down right now?

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,99 @@
...
+        staticPreview: this.render({
+          format: 'stars',
+          rating: 0.3,
+          reviewsCount: 200,
+        }),
+      },
+    ]
+  }
+
+  static render({ rating, reviewsCount, format }) {
```
calebcartwright: I've added some additional items below for `transform` (namely to remove it) and `handle` to refactor things a bit along with simplifying this function (for example the `rating * 5` expression currently exists in three places within this function). Just make sure to update the example listing aboves to use the corresponding `rating` value (i.e. `0.7` -> `3.5`, `0.3` -> `1.5`)

```js
  static render({ rating, reviewsCount, format }) {
    const message =
      format === 'rating'
        ? `${rating.toFixed(1)}/5 (${metric(reviewsCount)})`
        : starRating(rating)

    return {
      message,
      label: format,
      color: floorCount(rating, 2, 3, 4),
    }
  }
```

services/pkgreview/package-rating.tester.js
```diff
@@ -0,0 +1,63 @@
+'use strict'
+
+const t = (module.exports = require('../tester').createServiceTester())
+
```
calebcartwright: Similarly to the helpers for the service class, we've got a lot of helpers for test validation as well. If you import these suggseted ones, then the they can be used to complete these tests

```suggestion
const { withRegex, isStarRating } = require('../test-validators')

const isRatingWithReviews = withRegex(/[0-5]\.[0-9]{1}\/5?\s*\([0-9]*\)$/)

```

services/pkgreview/package-rating.tester.js
```diff
@@ -0,0 +1,63 @@
...
+t.create('Stars Badge renders')
+  .get('/stars/npm/react.json')
+  .expectBadge({ label: 'stars' })
```
calebcartwright:
```suggestion
  .expectBadge({
    label: 'stars',
    message: isStarRating,
  })
```

services/pkgreview/package-rating.tester.js
```diff
@@ -0,0 +1,63 @@
...
+
+t.create('Rating Badge renders')
+  .get('/rating/npm/react.json')
+  .expectBadge({ label: 'rating' })
```
calebcartwright:
```suggestion
  .expectBadge({
    label: 'rating',
    message: isRatingWithReviews,
  })
```

services/pkgreview/package-rating.tester.js
```diff
@@ -0,0 +1,63 @@
...
+
+t.create('Badge fails to render (package not found)')
+  .get('/rating/npm/icuwowgnbneiugvneiurfgrsfghw.json')
+  .expectBadge({
+    label: 'rating',
+    message: 'package not found',
+    color: 'lightgrey',
+  })
+
+t.create('Badge fails to render (package not found)')
```
calebcartwright: You can delete this test and all the ones below

services/pkgreview/package-rating.service.js
calebcartwright: Note that this doesn't take into account the scenario of a package that has no reviews, need to think through the desired behavior for that, probably either a `throw new NotFound({ prettyMessage: 'no reviews' })` or just treat it as a `0` rating

services/pkgreview/package-rating.tester.js
```diff
@@ -0,0 +1,63 @@
...
+
+t.create('Stars Badge renders')
+  .get('/stars/npm/react.json')
+  .expectBadge({ label: 'stars' })
+
+t.create('Rating Badge renders')
+  .get('/rating/npm/react.json')
+  .expectBadge({ label: 'rating' })
+
+t.create('Badge fails to render (package not found)')
```
calebcartwright:
```suggestion
t.create('nonexistent package')
```

KumarAbhirup:
Thank you so much for the review!

The API pkgreview.dev depends on for fetching NPM Data: `npms.io`, is facing some downtime. 

I hope it gets fixed. I have posted the issue on [https://github.com/npms-io/npms/issues/41](https://github.com/npms-io/npms/issues/41) to report the API bug. If it does not get fixed, I will use some other API or will do some web-scraping. :)

I don't know if it is a GitHub bug, but for some reason, I am not able to reply to your review. I can't see any textbox to do that.

Well, will tweak the code according to the review... Thanks again!

services/pkgreview/package-rating.tester.js
KumarAbhirup: I tried this in commit `88d64b517d5fb5a15280872a04196a0391544249`... But for a reason, it fails.

According to CircleCI logs, `5/5 (3)` does not match the regex.

KumarAbhirup: Well, the regex `\^[0-5](.?([0-9]))?\/5?\s*\([0-9]*\)$\` matches `5/5 (3)`.

There's just one bug with this... that is... even `5.5/5 (3)` matches though the maximum is 5.

services/pkgreview/package-rating.service.js
KumarAbhirup: `encodeURIComponent(pkgSlug)` is needed in `fetch` for fetching packages containing `%2f` in the slug.

I have updated the code regarding that in commit 135eaf49d37b10385359b17c937bc1524bf133ab. Please check out if it follows the guidelines. ♥

KumarAbhirup:
Thank you so much for your patience!

I have updated the code according your review.

I am still not sure if a few snippets in the code are following the guidelines...

## 1
```js
async handle({ format, pkgManager, pkgSlug }) {
    const { reviewsCount, rating } = await this.fetch({
      pkgManager,
      pkgSlug: encodeURIComponent(pkgSlug),
    })
      reviewsCount,
      format,
      rating: +parseFloat(rating * 5).toFixed(1),
    })
}
```

I have updated the `handle()` method so that it does not need the `transform()` method.

The `pkgSlug: encodeURIComponent(pkgSlug)` is needed for packages that contain `%2f` in the slug. 

**Is it okay to pass the modified slug value directly to the `fetch()` method?**

---

## 2

The `isRatingWithReviews` regex you suggested in the `package-rating.tester.js` was working for `5.0/5 (3)` but not for `5/5 (3)`...

So, I updated the regex to 👇
```js
const isRatingWithReviews = withRegex(/^[0-5](.?([0-9]))?\/5?\s*\([0-9]*\)$/)
```

It works for `4.3/5 (5)` and `5/5 (3)`. 

But a bug that I found with both the regex was that `5.6/5 (3)` also matches the regex. Which should not, because the minimum value is 1 and the maximum is 5.

I would love to get help here.

---

Asking for review again ❤️

calebcartwright:
> 1. Is it okay to pass the modified slug value directly to the fetch() method?

Would be better to just pass in `pkgSlug` as is to `fetch`, and have `fetch` do the `encodeURIComponent` as part of constructing the `url` param

> 2. But a bug that I found with both the regex was that 5.6/5 (3) also matches the regex. Which should not, because the minimum value is 1 and the maximum is 5.

Try updating the regex then. I just copied one from another service class as an example. Don't feel you have to use that _exact_ regex, but the test does need to validate the badge message and a `withRegex(...)` schema is the way to go. 

What about this one?
```js
const isRatingWithReviews = withRegex(/^(([0-4](.?([0-9]))?)|5)\/5?\s*\([0-9]*\)$/)
```

> 3. API Downtime

👍 Please do keep an eye on it and keep us posted. Shields has a very tight window in which it has to retrieve the data needed to create and return the requested badge in order for the badges to actually display within the GitHub UI (where they are most often used, embedded in readmes)

> 4. pkgreview.dev themed colorScale() >> Is there any way I could get the dynamic shades of purple for the badges?

Remember Shields.io's goals around consistency, and that all of our badges, including this one, need to be consistent. So no, please don't try to introduce a different color scale for this badge that would make it differ from our other rating/stars badges.

services/pkgreview/package-rating.service.js
```diff
@@ -74,26 +73,22 @@ module.exports = class PkgreviewRating extends BaseJsonService {
       schema,
       url: `https://pkgreview.dev/api/v1/${pkgManager}/${pkgSlug}`,
       errorMessages: {
-        400: 'package not found',
+        400: 'bad request',
```
calebcartwright: What circumstances (namely, what user-provided params to Shields.io) would result in Shields receiving a 400 response?

services/pkgreview/package-rating.service.js
```diff
@@ -74,26 +73,22 @@ module.exports = class PkgreviewRating extends BaseJsonService {
...
+        404: 'package not found',
         408: 'response timed out',
```
calebcartwright: I think this should be removed. The Shields infrastructure handles timeouts already and has a pattern for consistency handling scenarios where the upstream data provider does not respond/timesout. As such, individual service classes don't need to handle that directly.

services/pkgreview/package-rating.service.js
```diff
@@ -74,26 +73,22 @@ module.exports = class PkgreviewRating extends BaseJsonService {
...
-
-    return this.constructor.render({ ...transformedJson, format })
+    const { reviewsCount, rating } = await this.fetch({
+      pkgManager,
+      pkgSlug: encodeURIComponent(pkgSlug),
+    })
+    return this.constructor.render({
+      reviewsCount,
+      format,
+      rating: +parseFloat(rating * 5).toFixed(1),
```
calebcartwright: I'd deliberately put `toFixed` within the `render` function in the code suggestions as that's really only necessary as part of the rendering when format is `rating` then actual rating value is included in the rendered badge message (because we wouldn't want to display a message like `4.458323413/5 (9k)` 😉)

services/pkgreview/package-rating.tester.js
calebcartwright: Yup, sorry about that. Copied from a different service and needs a different regex, see my response in https://github.com/badges/shields/pull/4660#issuecomment-589889940 which has one that should work here

services/pkgreview/package-rating.service.js
```diff
@@ -0,0 +1,94 @@
...
+
+module.exports = class PkgreviewRating extends BaseJsonService {
+  static get category() {
+    return 'rating'
+  }
+
+  static get route() {
+    return {
+      base: 'pkgreview',
+      pattern: ':format(rating|stars)/:pkgManager/:pkgSlug+',
```
calebcartwright: Since the acceptable values for `pgkManager` are constrained by the upstream service (pkgreview), we can tighten up this route to avoid Shields unnecessarily accepting bogus values and making API calls to pkgreview.

Below suggestion only includes `npm` as the allowed value, though if you'd prefer to go ahead and include the other package managers you're planning to support that's fine too

```suggestion
      pattern: ':format(rating|stars)/:pkgManager(npm)/:pkgSlug+',
```

services/pkgreview/package-rating.service.js
KumarAbhirup: For unsupported `pkgManager`, the API responds with `400 Bad Request`.

For eg. `https://pkgreview.dev/api/v1/lol/react`, i.e. `/lol/react`

services/pkgreview/package-rating.service.js
KumarAbhirup: Will do!

services/pkgreview/package-rating.service.js
KumarAbhirup: Cool! Will do!

services/pkgreview/package-rating.service.js
KumarAbhirup: Will do!

services/pkgreview/package-rating.service.js
KumarAbhirup: Removed `408: 'response timed out',` :)

services/pkgreview/package-rating.service.js
KumarAbhirup: Done!

services/pkgreview/package-rating.service.js
KumarAbhirup: Fixed :)

services/pkgreview/package-rating.service.js
KumarAbhirup: The `Bad Request` from `pkgreview.dev API` is now handled with your pattern syntax: `/:pkgMananger(npm)`... So, yes, it is no more needed.

Removed the `400: 'bad request'` error. :)

KumarAbhirup:
Thanks a lot again!

Made the changes as asked (hopefully 🤞).

If there's something more I can do or fix, please let me know 🙌

This project helped me get better at reading code also this was the first time someone reviewed my code, so thanks ❤️

services/pkgreview/package-rating.service.js
calebcartwright: As you add support for new package managers to the service, drop us a note (or even better a PR 😄) so that we can update the route accordingly to accept those additional package managers

services/pkgreview/package-rating.tester.js
```diff
@@ -0,0 +1,24 @@
...
+t.create('Stars Badge renders')
+  .get('/stars/npm/react.json')
+  .expectBadge({ label: 'stars', message: isStarRating })
+
+t.create('Rating Badge renders')
+  .get('/rating/npm/react.json')
+  .expectBadge({ label: 'rating', message: isRatingWithReviews })
+
+t.create('nonexistent package')
+  .get('/rating/npm/ohlolweallknowthispackagewontexist.json')
```
calebcartwright: i like your fake package name 😆

calebcartwright:
> This project helped me get better at reading code also this was the first time someone reviewed my code, so thanks ❤️

Awesome! Congrats on your first contribution to Shields 🚀

PR #4662, opened 2020-02-16
Extend [Nexus] fallback logic to cover repository version badge

### LINKED ISSUE #4639: Nexus badges with custom repository don't automatically detect Nexus 3

- [x] My own instance

:beetle: **Description**

Shields created for Nexus 3 servers using the "Sonatype Nexus (repository)" UI say that the artifact is not found. Explicitly setting the `nexusVersion` parameter fixes the issue - but the parameter itself is not present in the UI.

:link: **Link to the badge**

Without `nexusVersion`: https://img.shields.io/nexus/omni/org.apache.commons/commons-lang3?server=https%3A%2F%2Fnexus.pentaho.org

With `nexusVersion`: https://img.shields.io/nexus/omni/org.apache.commons/commons-lang3?server=https%3A%2F%2Fnexus.pentaho.org&nexusVersion=3

:bulb: **Possible Solution**

Temporary workaround: manually append this to the badge's URL: `&nexusVersion=3`.

PyvesB:
Fixes #4639.

@nigredo-tori feel free to have a look! 😉


nigredo-tori:
The changes look good to me. This PR doesn't address the fact that only one out of four examples includes the `nexusVersion` input - but, as long as the fallback logic performs as expected, I'm fine with it.

PyvesB:
@nigredo-tori I've only included the parameter in the one Nexus 3 example we have - others target Nexus 2 servers. Given that we only currently only act on the parameter when it is equal to `3`, I don't want to encourage users to pointlessly specify  `nexusVersion=2` especially if there's a high likelihood that their server will migrate at some point. However, the documentation does talk about that parameter on all examples, with a bit more explanations on its behaviour. Does that make sense?

nigredo-tori:
@PyvesB, sure, that makes sense. My confusion mostly stems from how these examples are presented in the UI - it's easy to mistake them for comprehensive query builders, rather than a few arbitrary samples of a larger API. But that's a separate topic.

At the moment, this PR looks good to me. 👍

PR #4720, opened 2020-02-29
Remove a couple of unused dependencies

znarf:
Removed in https://github.com/badges/shields/pull/2906/files
- @babel/plugin-proposal-class-properties
- @babel/plugin-proposal-object-rest-spread
- @babel/preset-env

Built in:
- url

No usage detected:
- mkdirp
- ~react-error-overlay~
- walkdir

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -58,6 +58,9 @@ update_configs:
           dependency_name: prom-client
           version_requirement: '>=12.0.0'
       - match:
+          dependency_name: react-error-overlay
+          version_requirement: '>=3.0.0'
+      - match:
           dependency_name: sinon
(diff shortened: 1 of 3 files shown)


znarf:
@PyvesB I checked and `react-error-overlay` is effectively required. Looks like [this file](https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby/cache-dir/error-overlay-handler.js) from `gatsby` is copied in the `.cache` directory. It's smelling a bit wrong but I don't have enough knowledge in `gatsby` internals to dig more into it.

I'm adding it back, however I think it's best to add it in version "^3.0.0" so that it matches what current version of `gatsby` (2.19.22) is expecting.

Issue #4114, opened 2019-10-03
Make badges that rely on repository contents use default branch instead of "master"


h1nk:
:clipboard: **Description**

Continuation of the discussion started in PR #4096 (see [comment](https://github.com/badges/shields/pull/4096#issuecomment-537343737))

Essentially I asked whether it would be possible to automatically detect the default repository branch name for badges that rely on GitHub repository content and have the option to omit the branch name rather than simply defaulting to `master`.

paulmelnikow:
Basically we'd need to switch these from fetching from githubusercontent.com, which does not require quota, to using the API.

Advantages:

- The implementation would get slightly simpler since it'd use the same code whether in prod/staging/self-hosting.

Disadvantages:

- We'd need to use GitHub API calls to fetch the contents, meaning it'd use our quota.

h1nk:
Yes, I suppose the question is whether the API quota cost is worth it or not and also whether changing the default behavior of the branch-less option for the badges that have it is really necessary. I'm also going to assume that this will put an additional strain on badge load times vs. just simply using githubusercontent.com due to the added API query. Although some amount of caching could probably be used to reduce the added load time caused by having to check the default branch with the API on subsequent badge requests.

The only real reason I can see users wanting this is if they change the default branch name, which frankly doesn't happen very often I think.

Would love to hear some input from others if they think this would be a good or bad addition.

h1nk:
It actually might be better to just do the branch name lookup via the API if githubusercontent.com returns a 404 for `master`, otherwise skip the unnecessary API query. I think this would be an excellent compromise and solve the problem of changed default branch names while minimizing unnecessary API requests.

paulmelnikow:
It’d be better for quota, though in the 404 and non-master cases it’d double the number of requests and delay the response.

chris48s:
One thing to bear in mind with the contents API is that it only allows files under 1Mb. For example, if I call https://raw.githubusercontent.com/jukbot/smart-industry/master/package-lock.json that gives me a file, but https://api.github.com/repos/jukbot/smart-industry/contents/package-lock.json throws an error. `package-lock.json` in particular can become quite large (ours is just under the 1Mb limit). You might say its not such a bad thing to refuse to parse a file over 1Mb, but it would probably be a break for some users.

> it’d double the number of requests and delay the response

Yeah I wouldn't like to introduce a second API call to every request of this type to cover this case :grimacing:

h1nk:
[...]

PR #4722, opened 2020-03-01
Add [GithubPackageLockJson GitHubPackageJson npm] badge

### LINKED ISSUE #4693: Node module version from package-lock.json
:clipboard: **Description**

[npm](https://img.shields.io/npm/v/mermaid?color=teal&label=mermaid&logo=npm)
Shows the exact version of a package found in `package-lock.json`.

:link: **Data**

Retrieving the data should function pretty much the same way as it already works for `package.json`.

:microphone: **Motivation**

This badge should co-exist with the already existing `package.json` badge. The only difference would be that it gets the module's version from `package-lock.json` which allows you to show on which exact version of a package your code currently depends on.

vonas:
Closes #4693

### DIFF
--- a/services/github/github-package-lock-json.service.js
+++ b/services/github/github-package-lock-json.service.js
@@ -1,0 +1,281 @@
+'use strict'
+
+const { renderVersionBadge } = require('../version')
+const {
+  transformAndValidate,
+  renderDynamicBadge,
+} = require('../dynamic-common')
+const {
+  isPackageLockJsonWithDependencies,
+  getLockDependencyVersion,
+} = require('../package-lock-json-helpers')
+const { semver } = require('../validators')
+const { ConditionalGithubAuthV3Service } = require('./github-auth-service')
+const { fetchJsonFromRepo } = require('./github-common-fetch')
+const { documentation } = require('./github-helpers')
+const Joi = require('@hapi/joi')
+
+const keywords = ['npm', 'node']
+
+const versionSchema = Joi.object({
+  version: semver,
+}).required()
+
+// Leave this for completion's sake.
+// Usually you'll go ahead and get the version from package.json instead.
+class GithubPackageLockJsonVersion extends ConditionalGithubAuthV3Service {
+  static get category() {
+    return 'version'
+  }
+
+  static get route() {
+    return {
+      base: 'github/package-lock-json/v',
+      pattern: ':user/:repo/:branch*',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'GitHub package-lock.json version',
+        pattern: ':user/:repo',
+        namedParams: { user: 'IcedFrisby', repo: 'IcedFrisby' },
+        staticPreview: this.render({ version: '3.0.0' }),
+        documentation,
+        keywords,
+      },
+      {
+        title: 'GitHub package-lock.json version (branch)',
+        pattern: ':user/:repo/:branch',
+        namedParams: {
+          user: 'IcedFrisby',
+          repo: 'IcedFrisby',
+          branch: 'master',
+        },
+        staticPreview: this.render({ version: '3.0.0' }),
+        documentation,
+        keywords,
+      },
+    ]
+  }
+
+  static render({ version, branch }) {
+    return renderVersionBadge({
+      version,
+      tag: branch,
+      defaultLabel: 'version',
+    })
+  }
+
+  async handle({ user, repo, branch }) {
+    const { version } = await fetchJsonFromRepo(this, {
+      schema: versionSchema,
+      user,
+      repo,
+      branch,
+      filename: 'package-lock.json',
+    })
+    return this.constructor.render({ version, branch })
+  }
+}
+
+const dependencyQueryParamSchema = Joi.object({
+  filename: Joi.string(),
+}).required()
+
+class GithubPackageLockJsonDependencyVersion extends ConditionalGithubAuthV3Service {
+  static get category() {
+    return 'platform-support'
+  }
+
+  static get route() {
+    return {
+      base: 'github/package-lock-json/dependency-version',
+      pattern: ':user/:repo/:scope(@[^/]+)?/:packageName/:branch*',
+      queryParamSchema: dependencyQueryParamSchema,
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'GitHub package-lock.json dependency version',
+        pattern: ':user/:repo/:packageName',
+        namedParams: {
+          user: 'developit',
+          repo: 'microbundle',
+          packageName: 'rollup',
+        },
+        staticPreview: this.render({
+          dependency: 'rollup',
+          range: '1.29.1',
+        }),
+        documentation,
+        keywords,
+      },
+      {
+        title: 'GitHub package-lock.json dependency version (on branch)',
+        pattern: ':user/:repo/dev/:scope?/:packageName/:branch*',
+        namedParams: {
+          user: 'octokit',
+          repo: 'rest.js',
+          branch: 'master',
+          scope: '@babel',
+          packageName: 'core',
+        },
+        staticPreview: this.render({
+          dependency: '@babel/core',
+          range: '7.8.4',
+        }),
+        documentation,
+        keywords,
+      },
+      {
+        title:
+          'GitHub package-lock.json dependency version (subfolder of monorepo)',
+        pattern: ':user/:repo/:packageName',
+        namedParams: {
+          user: 'OriginProtocol',
+          repo: 'origin',
+          packageName: 'davidshimjs-qrcodejs',
+        },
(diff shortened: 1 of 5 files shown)


services/github/github-package-lock-json.service.js
```diff
@@ -0,0 +1,281 @@
...
+
+const keywords = ['npm', 'node']
+
+const versionSchema = Joi.object({
+  version: semver,
+}).required()
+
+// Leave this for completion's sake.
+// Usually you'll go ahead and get the version from package.json instead.
+class GithubPackageLockJsonVersion extends ConditionalGithubAuthV3Service {
```
calebcartwright: It seems like this badge is explicitly to provide the `version` key value from the package lockfile of the package itself (not dependnecies). What's the use case or need for this? It seems duplicative of existing functionality Shields already provides.

services/github/github-package-lock-json.service.js
```diff
@@ -0,0 +1,281 @@
...
+    return this.constructor.render({
+      dependency: wantedDependency,
+      range,
+    })
+  }
+}
+
+// This must be exported after GithubPackageLockJsonVersion in order for the
+// former to work correctly.
+class DynamicGithubPackageLockJson extends ConditionalGithubAuthV3Service {
```
calebcartwright: Same question as above on the `GithubPackageLockJsonVersion` class, what's the use case for this? Other than dependency version information, what's the use case when you'd need to use the lockfile instead of the package.json file?

services/github/github-package-lock-json.service.js
```diff
@@ -0,0 +1,281 @@
...
+  transformAndValidate,
+  renderDynamicBadge,
+} = require('../dynamic-common')
+const {
+  isPackageLockJsonWithDependencies,
+  getLockDependencyVersion,
+} = require('../package-lock-json-helpers')
+const { semver } = require('../validators')
+const { ConditionalGithubAuthV3Service } = require('./github-auth-service')
+const { fetchJsonFromRepo } = require('./github-common-fetch')
```
calebcartwright: Also want to note that this using the GH contents v3 API that has a file limit of 1MB, which will be problematic in a lot of instances given the sizes of the lockfiles

https://github.com/badges/shields/issues/4114#issuecomment-538531437

services/github/github-package-lock-json.service.js
vonas: I added it for completeness, there isn't any significant use case I can think of.

services/github/github-package-lock-json.service.js
vonas: Same here.

services/github/github-package-lock-json.service.js
calebcartwright: Thanks! Let's remove it then, we don't want to have to maintain code and tests unnecessarily, and removing it should simplify this file a bit too

calebcartwright:
Oops! For some reason I didn't notice those updated commits. Have deployed a review app for this PR and seems to be working well. Will take a closer look at the code changes tomorrow

https://shields-staging-pr-4722.herokuapp.com/github/package-lock-json/dependency-version/jukbot/smart-industry/@babel/core

services/github/github-package-lock-json.tester.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
+const { ServiceTester } = require('../tester')
+const { semverRange } = require('../validators')
+
+const t = (module.exports = new ServiceTester({
+  id: 'GithubPackageLockJson',
+  title: 'GithubPackageLockJson',
+  pathPrefix: '/github/package-lock-json',
+}))
```
calebcartwright: Let's use the standard service helper
```suggestion
const t = (module.exports = require('../tester').createServiceTester())
```

services/github/github-package-lock-json.tester.js
```diff
@@ -0,0 +1,65 @@
...
+    message: semverRange,
+  })
+
+t.create('Dependency version')
+  .get('/dependency-version/paulmelnikow/react-boxplot/simple-statistics.json')
+  .expectBadge({
+    label: 'simple-statistics',
+    message: semverRange,
+  })
+
```
calebcartwright: Lets get rid of these two tests as they are unnecessary. The label specification is common across any and every service and is well tested elsewhere already. The second test with `simple-statistics` is duplicative of the one above with `react`
```suggestion
```

services/github/github-package-lock-json.service.js
```diff
@@ -0,0 +1,126 @@
...
+const { documentation } = require('./github-helpers')
+const { BaseJsonService } = require('..')
+
+const keywords = ['npm', 'node']
+
+const dependencyQueryParamSchema = Joi.object({
+  filename: Joi.string(),
+}).required()
+
+class GithubPackageLockJsonDependencyVersion extends BaseJsonService {
```
calebcartwright: GitHub has a pretty strict throttling policy on non-authenticated requests, so all of the requests Shields issues to GitHub need to include authentication headers. We have additional base service class infrastructure to support this, which we'll want to use here (specifically `GithubAuthV3Service` to avoid the size limitations on the contents api)

```suggestion
class GithubPackageLockJsonDependencyVersion extends GithubAuthV3Service {
```

services/github/github-package-lock-json.service.js
```diff
@@ -0,0 +1,126 @@
...
+  static get category() {
+    return 'platform-support'
```
calebcartwright: Think this probably be under the `dependency` category. I know several of the other similar badges (npm dep version, github dep version from package.json, etc.) are in the `platform-support` category.

However, the pipenv locked dep version badge is under the `dependency` category and FWIW that feels like the better category for this badge (and those other ones currently under `platform-support`). 

@paulmelnikow  any thoughts on this?

services/github/github-package-lock-json.service.js
```diff
@@ -0,0 +1,126 @@
...
+        title:
+          'GitHub package-lock.json dependency version (subfolder of monorepo)',
+        pattern: ':user/:repo/:packageName',
+        namedParams: {
+          user: 'OriginProtocol',
+          repo: 'origin',
+          packageName: 'davidshimjs-qrcodejs',
+        },
+        queryParams: {
+          filename: 'origin/dapps/marketplace/package-lock.json',
```
calebcartwright: the path to the file shouldn't include a repo/remote/branch name, so we should remove `origin` from here so the live example values presented in the badge builder window will work

```suggestion
          filename: 'dapps/marketplace/package-lock.json',
```

https://shields-staging-pr-4722.herokuapp.com/github/package-lock-json/dependency-version/OriginProtocol/origin/davidshimjs-qrcodejs?filename=origin%2Fdapps%2Fmarketplace%2Fpackage-lock.json


https://shields-staging-pr-4722.herokuapp.com/github/package-lock-json/dependency-version/OriginProtocol/origin/davidshimjs-qrcodejs?filename=dapps%2Fmarketplace%2Fpackage-lock.json

services/github/github-package-lock-json.service.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  static render({ dependency, range }) {
+    return {
+      label: dependency,
+      message: range,
+      color: 'blue',
+    }
+  }
```
calebcartwright: We should use our `renderVersionBadge` helper here so that this badge will be consistent with our other version badges. 

You can grab the helper from `version.js`:
```js
const { renderVersionBadge } = require('../version')
```

and then remove this function altogether and just use that helper down in `handle` and in the examples

```suggestion
```

services/github/github-package-lock-json.service.js
```diff
@@ -0,0 +1,126 @@
...
+    })
+
+    return this.constructor.render({
+      dependency: wantedDependency,
+      range,
+    })
+  }
+}
+
+module.exports = [GithubPackageLockJsonDependencyVersion]
```
calebcartwright:
```suggestion
```

services/github/github-package-lock-json.service.js
paulmelnikow: Hmm, interesting question.

This one is a lot like the Pipenv badge so whatever we decide, these probably belong in the same place.

Maybe we ought to narrow "Platform & Version Support" to "Platform Support" and reserve it for things like the Node version / php version, and move the others into Dependencies.

paulmelnikow:
Hey @vonas, would you be up for incorporating those comments? It would be great to get this merged!

PyvesB:

If someone else wants to pick this up, you can simply fetch the commits by running `git fetch origin pull/4722/head:pr-4722`. Address the pending review comments and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

PR #4723, opened 2020-03-01
add inline wercker ci docs

### LINKED ISSUE #4715: Wercker shields reporting "application not found"
When trying the Wercker shields they are not finding the application using the application id from the official wercker shields badge.

eg. these 2 addresses on wercker work with this id

https://app.wercker.com/status/c46a517909343e278c9ce354164394ae/s/master

https://app.wercker.com/project/byKey/c46a517909343e278c9ce354164394ae

so I infer the application ID to be `c46a517909343e278c9ce354164394ae`, but this doesn't work on shield.io as generated by the https://shields.io/category/build page:

https://img.shields.io/wercker/build/harisekhon/c46a517909343e278c9ce354164394ae

I've also tried this:

https://img.shields.io/wercker/ci/c46a517909343e278c9ce354164394ae

calebcartwright:
Close #4715 

Adds inline documentation on the Wercker CI badges to provide guidance to users on the need to use the application id (and not the wercker badge key) and how they can lookup the app id

### DIFF
--- a/services/wercker/wercker.service.js
+++ b/services/wercker/wercker.service.js
@@ -14,6 +14,24 @@ const werckerSchema = Joi.array()
   .max(1)
   .required()
 
+const werckerCIDocumentation = `
+<p>
+  Note that Wercker badge Key (used in Wercker's native badge urls) is not the same as
+  the Application Id and the badge key will not work.
+</p>
+<p>
+  You can use the Wercker API to locate your Application Id:
+  <br />
+  <br />
+  https://app.wercker.com/api/v3/applications/:username/:applicationName
+  <br />
+  For example: https://app.wercker.com/api/v3/applications/wercker/go-wercker-api
+  <br />
+  <br />
+  Your Application Id will be in the 'id' field in the API response.
+</p>
+`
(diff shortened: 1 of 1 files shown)


services/wercker/wercker.service.js
```diff
@@ -14,6 +14,24 @@ const werckerSchema = Joi.array()
   .max(1)
   .required()
 
+const werckerCIDocumentation = `
+<p>
+  Note that Wercker badge Key (used in Wercker's native badge urls) is not the same as
+  the Application Id and the badge key will not work.
+</p>
+<p>
+  You can use the Wercker API to locate your Application Id:
```
calebcartwright: I've no idea if there's a UI/more user friendly means to look this up, so pointing folks to the API since we know that'll work

PyvesB: Hmmm... Given that it's the main ID used for quite a few of the API calls and given that users haven't complained about this, I would be surprised if it wasn't available in the UI. But no way to verify that without a real account, so let's go with that for now. 😃

PR #4753, opened 2020-03-07
Trim user input when building path and query string

### LINKED ISSUE #4725: Ignore whitespaces in frontend input
:clipboard: **Description**

From @calebcartwright  in #4715:
> You may want to make sure there's no leading nor trailing spaces in the application id box, especially if you copy/pasted the value in there

To be honest, this happens to me quite often when using the UI, and the error messages generally aren't helpful (e.g. 404 badge not found, invalid query parameter, etc.). Could we make the frontend code trim user inputs when generating the preview and the URL? Do you have cases in mind where leading or trailing spaces are actually valid?

PyvesB:
Fixes #4725.

### DIFF
--- a/frontend/components/customizer/path-builder.tsx
+++ b/frontend/components/customizer/path-builder.tsx
@@ -89,12 +89,12 @@ export function constructPath({
   const path = tokens
     .map(token => {
       if (typeof token === 'string') {
-        return token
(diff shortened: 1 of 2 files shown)


paulmelnikow:
Nice refinement!

PR #4756, opened 2020-03-08
📦 version 3

### LINKED ISSUE #3480: NPM audit vulnerability on dependency of gh-badges package

- [ ] My own instance
- [x] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

NPM audit is reporting a moderate severity vulnerability in the "dot" package, which is a dependency of the gh-badges package:

```
                       === npm audit security report ===


                                 Manual Review
             Some vulnerabilities require your attention to resolve

          Visit https://go.npm.me/audit-guide for additional guidance


  Moderate        Command Injection

  Package         dot

  Patched in      No patch available

  Dependency of   gh-badges

  Path            gh-badges > dot

  More info       https://npmjs.com/advisories/798

found 1 moderate severity vulnerability in 97 scanned packages
  1 vulnerability requires manual review. See the full report for details.
```

This was raised on the dot package's repo: https://github.com/olado/doT/issues/281

However, it seems like that package is not being maintained anymore?

As a result of the dependency on the "dot" package, the gh-badges package will always introduce an NPM audit failure.

:link: **Link to the badge**

:bulb: **Possible Solution**

### LINKED ISSUE #4564: PNG/JPG/GIF support: to raster or not to raster?
Users of Shields, we need your opinion!

We currently support generating PNG/JPG/GIF badges through our CLI tool. We were wondering if this feature was useful to anyone - there's not point in us maintaining and bloating our codebase with something that isn't used. We've got a PR to clean things up: #4523.

If it's a valued feature, we would look at adding it to our NPM package for feature parity. Would any of you be interested in such an option? We've got a PR to achieve this: #4524.

Simply add a reaction to this post to let us know what you think:
**👍 to keep support in the CLI and add support to the NPM package.**
**👎 to drop support from the CLI.**
**👀 what, Shields has a CLI and an NPM package??**

chris48s:
There's a lot going on in this pull request. Fortunately, because we've worked on it in small chunks, most of it has already been reviewed :)

Commits edbad19, cf3612d, 544f4e6, f3f95b7 and fc16600 have all been reviewed already. The relevant PRs are linked in the commit messages, although they've now been though some rebasing to get the branch up to date with master.

Commit 2c02630 and onwards are all new in this PR, but its mostly moving stuff about and a final review/update on the docs.

The next steps I think we should make on this are:

- Merge this and deploy it. Then shields.io will be finally running off the new package code, including new template code introduced in #4459
- Publish a beta/release candidate to NPM using the new package name
- Monitor production for errors. We're the biggest user of our own package. I suspect if we've overlooked any bugs we'll spot them pretty fast once we're using it in production
- Publish a stable release using the new package name
- Publish a final gh-badges 2.2.2 release advising users to upgrade
- Mark gh-badges dead/unmaintained on NPM
- Start work on v3.1

but happy to discuss it.

Just to review the roadmap/issues:

- Merging this closes #3480
- Merging this closes #4564
- This PR contains all the changes discussed in #3587 which require a BC-break. When we merge this I will close that and raise new issues for:
    - Expose the concept of logos, add SimpleIcons as a peer dependency
    - Depend on the public interface internally instead of relative imports
    - Add typescript definitions to make package easier to use with TypeScript
    We can make those changes later/in a feature release without breaking backwards compatibility.

Usual review process aside, it would be particularly useful to get another set of eyes on the documentation to make sure we're all up-to-date and the migration advice in the changelog is correct.


badge-maker/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "gh-badges",
-  "version": "2.2.1",
+  "name": "badge-maker",
+  "version": "3.0.0-rc1",
```
chris48s: * I've already published a package that does nothing at https://www.npmjs.com/package/badge-maker just to stake a claim on the name.
* I'm suggesting the first release is `-rc1` here, but I don't really have strong opinions on where it should be `-beta` or whatever..

badge-maker/CHANGELOG.md
```diff
@@ -1,6 +1,55 @@
...
+  will throw a `ValidationError`.
+- Raster support has been removed from the CLI. On the console, the output of `badge` can be piped to a utility like [imagemagick](https://imagemagick.org/script/command-line-processing.php). If you were previously using
+  ```sh
+  badge build passed :green .gif
+  ```
+  this could be replaced by
+  ```sh
+  badge build passed :green | magick svg:- gif:-
+  ```
+- Deprecated parameters have been removed. In version 2.2.0 the `colorA`, `colorB` and `colorscheme` params were deprecated. In version 3.0.0 these have been removed.
```
calebcartwright: 👍

badge-maker/CHANGELOG.md
```diff
@@ -1,6 +1,55 @@
...
+
+### Security
+
+- Removed dependency on doT library which has known vulnerabilities.
```
calebcartwright: 🔒 🎉

badge-maker/README.md
```diff
@@ -1,42 +1,41 @@
...
+badge build passed :green .svg > mybadge.svg
 ```
 
 ### As a library
 
 ```js
-const { BadgeFactory } = require('gh-badges')
-
-const bf = new BadgeFactory()
+const { makeBadge, ValidationError } = require('badge-maker')
```
calebcartwright: is there a need to import `ValidationError` here?

calebcartwright:
Only a handful of service test failures that were unrelated 👍

frontend/components/development/style-page.tsx
```diff
@@ -0,0 +1,145 @@
+import React, { Fragment } from 'react'
```
calebcartwright: will this page be enabled in production?

badge-maker/README.md
```diff
@@ -1,42 +1,41 @@
...
 
 const format = {
   text: ['build', 'passed'],
   color: 'green',
   template: 'flat',
 }
 
-const svg = bf.create(format)
+makeBadge(format) // <svg...
+makeBadge({}) // ValidationError: Field `text` is required
```
calebcartwright: not a big deal, but may be helpful to include the assignment here at least on the initial use case like in the changelog and 2.x docs
```suggestion
const badge = makeBadge(format) // <svg...
makeBadge({}) // ValidationError: Field `text` is required
```

badge-maker/README.md
```diff
@@ -47,28 +46,19 @@ The format is the following:
 
 ```js
 {
-  text: [ 'build', 'passed' ],  // Textual information shown, in order
+  text: [ 'build', 'passed' ],  // Textual information to be shown on the badge, in order
 
-  format: 'svg',  // Also supports json
+  format: 'svg',  // One of 'svg' or 'json'
```
calebcartwright: 'One of' is reading a bit strange to me with just two options (though I can't really put my finger on why 🤔) I don't think there's any ambiguity around what the two possible options are, and I'm happy to approve as is, but suggesting the below in case that reads a bit better to anyone else
 
```suggestion
  format: 'svg',  // Either 'svg' or 'json'
```

frontend/components/development/style-page.tsx
chris48s: Nope. Its a bit buried, but if you have a look at [gatsby-node.js](https://github.com/badges/shields/pull/4756/files#diff-fda05457e393bada716f508859bfc604) its conditional on the `INCLUDE_DEV_PAGES` env var which is off in production. Same with `frontend/components/development/logo-page.tsx`.

badge-maker/README.md
chris48s: I think we do need to make it clear somewhere in the README that you can import and handle `ValidationError` from the module. I've pushed another commit amending the example a bit so we actually do something with `ValidationError` (and incorporating your other suggestions). Do you think that is a bit clearer?

calebcartwright: Yup, like it!

badge-maker/README.md
```diff
@@ -126,3 +122,12 @@ There are three ways to specify `color` and `labelColor`:
...
 [css/svg color]: http://www.w3.org/TR/SVG/types.html#DataTypeColor
+
+## Raster Formats
+
+Conversion to raster formats is no longer directly supported. In javascript
+code, SVG badges can be converted to raster formats using a library like
+[gm](https://www.npmjs.com/package/gm). On the console, the output of `badge`
+can be piped to a utility like
+[imagemagick](https://imagemagick.org/script/command-line-processing.php)
+e.g: `badge build passed :green | magick svg:- gif:-`.
```
paulmelnikow: 👍

badge-maker/lib/badge-cli.js
```diff
@@ -18,7 +17,7 @@ if (process.argv.length < 4) {
   console.log('    #xxxxxx (six hex digits)')
   console.log('    color (CSS color)')
   console.log('  output:')
-  console.log('    svg, png, jpg, or gif')
+  console.log('    svg or json')
```
paulmelnikow: Is there a reason the CLI ought to support JSON? I'm not sure I see the value of that.

badge-maker/package.json
```diff
@@ -35,10 +35,7 @@
   },
   "dependencies": {
     "anafanafo": "^1.0.0",
-    "dot": "^1.1.2",
```
paulmelnikow: 🎉

badge-maker/lib/badge-cli.js
chris48s: This isn't actually new - it worked before it just wasn't documented in the console help.

Getting rid of the raster support allowed us to drop some additional dependencies but maintaining this is zero effort/overhead so I'm inclined to just keep it around. In fact, we'd have to add an extra validation rule to disallow it in this case.

paulmelnikow: It doesn't really make sense as a feature, though. If it's not documented, we could remove it later without calling it a breaking change. This could be considered in a follow-on PR, post rc1.

chris48s: > This could be considered in a follow-on PR, post rc1.

Meh. Its taken the thick end of a year to get as far as opening this PR. If we need another evening or two to get this detail right I think we can all cope.

> If it's not documented, we could remove it later without calling it a breaking change

This is an interesting point. I hadn't focussed that we've never previously documented that this exists at all. I'm not deeply concerned about a breaking change given the laundry list of other breaking changes in this release. That fact we've never documented it does substantially reduce the chance that anyone is using it though :slightly_smiling_face: 

Thinking about it a bit further, if we drop JSON from the CLI, we should just drop the format argument completely. If the only possible valid argument you can pass is `.svg` and its optional, `badge build passed :green .svg` serves no purpose at all.

I'll have a look at this later..

paulmelnikow:
Tangential question to https://github.com/badges/shields/pull/4756#discussion_r390566136: are we still using the package to generate the JSON badges? At some point after this is merged, I'd like to refactor the server parts which touch `makeBadge()` and stop running the JSON badges through it.

chris48s:
Hrmmm.. So first lets start with a bit of disambiguation. There's 2 `makeBadge()` functions in the package - one public and one internal:

https://github.com/badges/shields/blob/40bfc6ec51598589c63a83f6499b50eab69d9361/badge-maker/lib/make-badge.js#L11
https://github.com/badges/shields/blob/40bfc6ec51598589c63a83f6499b50eab69d9361/badge-maker/lib/index.js#L75

Lets call them `lib/make-badge.makeBadge` and `lib/index.makeBadge` just to avoid confusion.
Internally, we depend on relative imports to `lib/make-badge.makeBadge` (so we're depending on the internals of our package as opposed to its public interface). See https://github.com/badges/shields/pull/4756/commits/32ed0531bd0c563e37915dc0abc47940bca62b8e . At the moment we have to do that because there's functionality we rely on for shields.io which is not exposed via the NPM package (logos and links). This means we have a somewhat poorly defined interface boundary.

What that means is we're not in the starting place where you think we are, but we're already in the place you think we should move to.

I've been broadly thinking in the opposite direction though. I think we should try to move to a situation where shields.io depends on the public interface of the `badge-maker` package (`lib/index.makeBadge`) so we're "eating our own dog food". Moving away from relative imports to `lib/make-badge.makeBadge` was one of the original jobs in the v3 roadmap https://github.com/badges/shields/issues/3587 but its predicated on adding logos to the package which can be done in a future release with no BC break. That said, I'm not completely wedded to that course of action. Maybe its just tinkering we don't really need to do.

Based on your comment, I'm guessing your next suggestions is going to be that we drop JSON support from the library. While I'm on board that its a bit pointless in the CLI, I'm less convinced on the library side. On the face of it,

```js
const { makeBadge } = require('badge-maker')

const format = {
  text: ['build', 'passing'],
  color: 'green',
  format: 'json',
}

const badge = makeBadge(format)
console.log(badge)
// {"label":"build","message":"passing","color":"green","link":[],"name":"build","value":"passing"}
```

is a bit wierd, but broadly the reason we expose this is to make testing easier. We write tests. So do our users (or at least I hope they do :D ) so it feels like there is a legit use-case here and no real benefit to us in removing it.

calebcartwright:
I don't have any strong opinions on json support in the lib one way or the other, but happy to give this another 👍 if and when needed for merging

chris48s:
This one has kind of festered for a bit, but I'd quite like to try and resolve it one way or another. Partly because there's quite a large chunk of work here blocked on one decision. Also partly because if I have to resolve a merge conflict between this and reformatting every file in the codebase to upgrade prettier, I might go insane.

PyvesB:
Don't feel strongly either, keeping JSON support in the library (but not the CLI) doesn't hurt much and comes in handy whilst testing or simply experimenting with the library. Generally speaking, your roadmap and future plans of "eating our own dog food" make perfect sense! :)

Overall this looks like good work to me, thanks for all the efforts and documentation! 👍 

I've got one observation that's making me slightly nervous. I tried running that new benchmark tooling around the `makeBadge` call in _base-static.js_, and performance is now ~50% slower (from 0.054ms to 0.082ms on my machine, averaged over 5 executions of `npm run benchmark:badge`). It's probably not the end of the world given that we're talking about ~30μs, however, given that `makeBadge` can be invoked hundreds of times under heavy load and that our servers are already maxed out in terms of capacity, I thought I would mention it.

chris48s:
Its a good point. In the depths of the review comments for #4459 we did observe that switching away from doT does reduce performance: https://github.com/badges/shields/pull/4459#issuecomment-570706539

In the grand scheme of things, I don't think it really matters that its _slightly_ slower when we're just thinking about one request. Potentially it does become important given our current infrastructure constraints. I'm not really sure how to gauge how big the impact will be when it comes to making lots of calls at peak load.

I did have a think about Paul's comments on optimisation in that thread, but I tested some benchmark runs using the default style (which uses a class and a lot of function calls) against some using the for-the-badge style (where almost all of the code is in one function) and it seems to make no difference to performance so I think that's probably a dead end..

Do you have any other thoughts on possible optimisations?

PyvesB:
> I'm not really sure how to gauge how big the impact will be when it comes to making lots of calls at peak load.

Ideally, it would have been nice to do a blue/green deployment on one of the servers, and take a look at the Grafana dashboard after a few hours. But we would need some level of coordination with Paul and make sure it's the only effective change compared to the previous deployment (i.e. we wouldn't want one of the servers offering a new badge and not the others). Not sure it's worth the effort. Will see whether I can generate some Flamegraphs locally instead.

> Do you have any other thoughts on possible optimisations?

None yet, but I'm happy to do a more focused read through in the coming days and see if any easy wins stick out! But I wouldn't treat this as a blocker, again I'm only _slightly_ nervous, I've got a feeling that even a 50% slowdown on `makeBadge` will still have a small overall impact. :+1:

chris48s:
I think you're right that this probably won't make a huge difference, but that's not a bad plan. If we were to deploy this change only on to one of the three servers then it should hopefully become clear if that server is performing substantially worse than the other two. If so, we could revert.

calebcartwright:
> Ideally, it would have been nice to do a blue/green deployment on one of the servers

>  If we were to deploy this change only on to one of the three servers then it should hopefully become clear if that server is performing substantially worse than the other two

One thing I'm still unclear on (or keep forgetting) is our current load balancing strategy. Do we have a GTM-based strategy splitting between the two in Canada and one in France, and/or is there a round robin within a region, etc.

Admittedly I've still not had a chance to take a closer look at our existing measurement capabilities. However, questions like load balancing strategies, plus OVH's recurring issues of late in the Canada data centers (which have a strong correlation with the timing of our responsiveness issues FWIW), have me somewhat skeptical of how reliably we could measure/attribute performance issues to a single function like `makeBadge` especially considering all the other moving parts involved in servicing the badge request.

I'm a 100% on board with the principle of blue/green or canary deployments, but I'm not sure how readily we can do that with the current prod runtime. IMO there's a lot of uncontrollable/unknown variables in prod that can impact response times, and I'd expect there to be even more to consider if the change was only on a single server.

I'd personally lean towards deploying this changeset (in isolation/no other changes) via the typical deploy process (all servers) if we want to do some experimentation with this.

My two cents

paulmelnikow:
I could update this with master and then deploy the branch to s0. That's probably the simplest way to get it deployed.

As far as checking performance, what I'd probably do is watch for (1) a bunch of bad badges on GitHub or (2) a spike in response times on that server. Maybe @platan has ideas on other things that can be monitored.

I missed that this was blocking on a substantive question, sorry! Let me respond to that next.

paulmelnikow:
> Based on your comment, I'm guessing your next suggestions is going to be that we drop JSON support from the library. While I'm on board that its a bit pointless in the CLI, I'm less convinced on the library side. On the face of it,
> 
> ```js
> const { makeBadge } = require('badge-maker')
> 
> const format = {
>   text: ['build', 'passing'],
>   color: 'green',
>   format: 'json',
> }
> 
> const badge = makeBadge(format)
> console.log(badge)
> // {"label":"build","message":"passing","color":"green","link":[],"name":"build","value":"passing"}
> ```
> 
> is a bit wierd, but broadly the reason we expose this is to make testing easier. We write tests. So do our users (or at least I hope they do :D ) so it feels like there is a legit use-case here and no real benefit to us in removing it.

It seems to me a consumer of the library writing a test is basically writing a test like we'd write of one of our "transform" functions:

```js
expect(JSON.parse(makeBadge({
  text: ['build', 'passing'],
  color: 'green',
  format: 'json',
}))).to.include({
  label: 'build',
  message: 'passing',
  color: 'green',
})
```

Ideally I think we'd go one step further in the library API, adding `label` and `message`, and deprecating the `text` parameter. Then the JSON really becomes a pass through and there's nothing left to test (except that valid SVG comes out).

paulmelnikow:
@calebcartwright There isn't any true load balancing; it's just round-robin DNS.

chris48s:
So you've kind of answered a small question there, but passed over the big one :D

I think there is a big design question about what the badge-maker package should be which we don't all agree on:

One way of thinking about it is:
Option 1 - Badge-maker should be a stand-alone package. It happens to live in the same repo as shields.io, but in principle it could live in its own repository if we wanted. Its a package with many users, but its biggest and main user is shields.io. Any functionality we depend on should be exposed and documented via the public interface of the package and we consume that functionality in the same way as any other user, eating our own dogfood.

Another way of thinking about it is:
Option 2 - Badge-maker is a package which exposes only a subset of the badge rendering capibilites of shields.io to a wider audience. It lives in and is inseparable from the shields.io repo. The interface we give our users is different to the interface we consume internally.

Option 2 is where we are now. I've been thinking we should be heading towards option 1. I'm not necessarily _right_ about that, but that's definitely the direction I've been thinking in. Given that removing a feature that our own application depends on from the public interface of the package actively reverses that direction of travel, I think its worthwhile to be explicit about the reasons for that.

---

Independently of that, and whichever way we go on it, this a very interesting point:

> Ideally I think we'd go one step further in the library API, adding label and message, and deprecating the text parameter. Then the JSON really becomes a pass through

This has caused me to step back and think about the interface we're exposing in a way I haven't previously, but really really should have.

At the moment, our public interface is:

```js
{
  text: ['build', 'passing'],
  labelColor: '#4c1',
  color: '#555',

  // we don't currently expose this, but internally we have
  links: ['https://foo/', 'https://bar/'],
}
```

If you take your point into account, but just think of it from the outside, the input object should probably really be:

```js
{
  label: 'build', /*..or maybe */ labelText: 'build',
  message: 'passing', /*..or maybe */ messageText: 'passing',
  labelColor: '#4c1',
  messageColor: '#555', // as opposed to just color
  labelLink: 'https://foo/',
  messageLink: 'https://bar/',
}
```

or perhaps even:

```js
{
  label: {
    text: 'build',
    color: '#4c1',
    link: 'https://foo/',
  },
  message: {
    text: 'passing',
    color: '#555',
    link: 'https://bar/',
  },
}
```

Zooming out, both of those feel more obvious/consistent. So whereas you're thinking modifying the input object to match our internals, this makes me think we should think about what is an intuitive interface from a user's perspective and then provide that..

Changing that would move us from one state where the public interface doesn't quite match the internals to another state where the public interface doesn't quite match the internals, but it makes more sense.

While we're making numerous non-BC changes, perhaps now is a good time to look at this too.

---

Its taken a while to respond to this, but getting more eyes and opinions on this is forcing me to think things through more and see stuff we've missed first time round, which is useful. I've also realised that:

* Links are trivial to (re-)enable, so I might as well do that now once we decided on the interface and
* By requiring both label and message in https://github.com/badges/shields/pull/3875 we break the ability to make a "message only" badge like  https://img.shields.io/badge/-message-blue so when I make changes based on discussion of the above, I'll fix that too!

paulmelnikow:
@chris48s What do you think about merging this, shipping it as `3.0.0beta1` (or not, even) and continuing to address these open questions?

chris48s:
Sure. I think at this point there is value in being able to test the performance impact of the template changes and not have this huge changeset hanging over us as I think we're pretty settled on the changes to internals and badge rendering logic proposed in this PR, which is the meat of it.

I think I would like to try and make a reasonably final-ish decision about the names of the input object keys and fix the known issue before we publish a beta to NPM.

paulmelnikow:
Sounds good!

platan:
Actually I have one idea how can we test a performance of this change. 
1. Prepare a massive amount of URLs to different badges
2. Deploy shields from the `master` branch and for the `v3-dev` branch somewhere, e.g. at Heroku
3. Send requests to both deployments
4. Compare the response times and a server load using our [metrics](http://metrics.shields.io). 


Regarding point 1, several weeks ago I started analysing [GitHub public dataset available at GCP](https://console.cloud.google.com/marketplace/details/github/github-repos?filter=solution-type:dataset), https://medium.com/google-cloud/github-on-bigquery-analyze-all-the-code-b3576fd2b150. I haven't finish this (yet), but we are able to extract a lot of URLs to shields.io used by our users.

chris48s:
Following on from the conversation in #4849 I've updated this PR

paulmelnikow:
Is the proposal in https://github.com/badges/shields/pull/4756#issuecomment-608061009 blocking on #4874?

chris48s:
@paulmelnikow - As discussed above, for the purposes of evaluating performance/easily rolling back if necessary it is useful to be able to deploy this in isolation.
As such, I'll leave it with you to actually press the merge button on this one so you can co-ordinate that. I'll pick up publishing the beta package once that's done.

paulmelnikow:
This branch, specifically e356258, is live on s0.

chris48s:
I don't know exactly what times you deployed this/rolled it back, so its difficult to tell from the graphs but given we think the outage was linked to OVH do you fancy giving this another test next week and see what happens?

paulmelnikow:
Yea, sounds good. I’ll do a deploy tomorrow and we can take this branch for a spin on Monday.

paulmelnikow:
While trying to deploy this on s0, I got a message that the disk filled up again. I wrote Thaddée asking him to clean it. (I'm assuming it's Cypress installs, as before.)

As of now s0 has the same code as the other machines.

paulmelnikow:
Okay, the disk on s0 is back to an okay place. I'll deploy this tonight.

paulmelnikow:
13366fc5b01a4d66fc3bfbebdb70f9081059601b is live on s0!

paulmelnikow:
It looks like unknown templates are causing internal errors which somehow aren't getting caught, causing requests to hang.

e.g. https://s0.shields-server.com/badge/foo-bar-blue?style=foobar

There are a good few dozen of these in the last couple minutes, so I'm thinking I'll probably roll this back.

calebcartwright:
Did you rollback already?
https://s0.shields-server.com/badge/foo-bar-blue?style=foobar

chris48s:
Whoops - have updated with a fix.

chris48s:
I've slightly lost track of what's deployed at this point, but are we pretty happy we've now ironed out the kinks with this?

paulmelnikow:
The live one is 4999e55. I think we're confident it doesn't have major bugs.

I'm okay with merging this.

Has anyone evaluated how s0 is performing compared to historical?

chris48s:
Cool. In that case I'm going to crack on and merge this, push the release candidate and review the next steps..

platan:
After merging this PR I got an email from Heroku: "Automatic deployment of shields-staging failed".

chris48s:
Hmm. Thanks - good spot. The error it is failing with is

```
npm ERR! syscall access
npm ERR! path /tmp/build_213196dfe99649b0be38d9bc6dc323c7/node_modules/gh-badges
npm ERR! errno -2
npm ERR! enoent ENOENT: no such file or directory, access '/tmp/build_213196dfe99649b0be38d9bc6dc323c7/node_modules/gh-badges'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 
```

I wonder if there is some cache that needs to be cleared? The staging PRs were deploying correctly.

chris48s:
I've set `NODEMODULESCACHE=true` in the env vars for the staging pipeline, but I can't see a way to retry the deploy without pushing another commit to master. I'm going to check back on it tomorrow (inevitably we'll merge some dependency bumps tomorrow). If that fixes it, I'll unset the env var to re-enable the cache. If that doesn't fix it then I'll worry about it some more.

paulmelnikow:
I think you can do that here, want to give it a try?

Screen Shot 2020-04-23 at 5 13 49 PM

PR #4759, opened 2020-03-13
feat: add badge for .nycrc config based coverage, run [nycrc]

bcoe:
Adds a badge for folks who are using [nyc](https://github.com/istanbuljs/nyc) or [c8](https://github.com/bcoe/c8) directly to enforce coverage, via the [--check-coverage](https://github.com/bcoe/c8#checking-coverage) flag.

`lines` and `branches` thresholds can be set in a `.nycrc`, `.nycrc.json`, or `package.json` (in the case of `package.json`, in the `nyc` or `c8` stanza), these stanzas will in turn be reflected in the badge, see: https://github.com/yargs/yargs/blob/master/.nycrc


services/nycrc/nycrc.tester.js
```diff
@@ -0,0 +1,58 @@
+'use strict'
+
+const { isIntegerPercentage } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('valid .nycrc')
+  .get('/yargs/yargs.json?config=.nycrc')
+  .intercept(nock =>
+    nock('https://api.github.com')
```
PyvesB: We tend to use real data as much as possible, especially for the "happy case" tests. This helps us spot changes in behaviour/performance of upstream APIs early on. Could `nock` be removed from this test and the `package.json with nyc stanza` one?

services/nycrc/nycrc.tester.js
```diff
@@ -0,0 +1,58 @@
...
+        ).toString('base64'),
+        encoding: 'base64',
+      })
+  )
+  .expectBadge({
+    label: 'min coverage',
+    message: '"branches" or "lines" threshold missing',
+  })
+
+t.create('package.json with nyc stanza')
```
PyvesB: Would a test for the `no nyc or c8 stanza found` case be useful? You could simply use _badges/shields_ as an example for this.

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,129 @@
...
+        message: `${coverage.toFixed(0)}%`,
+        color: coveragePercentage(coverage),
+      }
+    }
+  }
+
+  async handle({ user, repo }, queryParams) {
+    let coverage = 'unknown'
+    const { config } = queryParams
+    const missingFields = new InvalidParameter({
```
PyvesB: `InvalidParameter` is generally used for invalid user input, as a path or query parameter in the badge for example. Would `InvalidResponse` be more appropriate here? Or maybe `NotFound`?

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,129 @@
...
+      const pkgJson = await fetchJsonFromRepo(this, {
+        schema: pkgJSONSchema,
+        user,
+        repo,
+        branch: 'master',
+        filename: config,
+      })
+      const nycConfig = pkgJson.c8 || pkgJson.nyc
+      if (!nycConfig) {
+        throw new InvalidParameter({
```
PyvesB: Probably best to use `NotFound` here.

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,129 @@
...
+      }
+    } else {
+      return {
+        message: `${coverage.toFixed(0)}%`,
+        color: coveragePercentage(coverage),
+      }
+    }
+  }
+
+  async handle({ user, repo }, queryParams) {
```
PyvesB: Would extracting part of the logic inside a `transform` function make things a bit more readable? Something like this:
````
  transform({ branches, lines }) {
    const missingFields = new InvalidParameter({
      prettyMessage: '"branches" or "lines" threshold missing',
    })
    if (branches || lines) {
      return branches || lines
    } else {
      throw missingFields
    }
  }
````
You could even pass a config object in and extract `branches` and `lines` in there as these variable names are the same regardless of the configuration type. 😉

File: services/nycrc/nycrc.service.js [hunk 6]
```diff
@@ -0,0 +1,129 @@
...
+    return 'coverage'
+  }
+
+  static get route() {
+    return {
+      base: 'nycrc',
+      // TODO: eventually add support for .yml and .yaml:
+      pattern: ':user/:repo',
+      queryParamSchema: Joi.object({
+        config: Joi.string().valid('.nycrc', '.nycrc.json', 'package.json'),
```
PyvesB: 👍

@paulmelnikow also suggested using a default option in his email, I think you can achieve that by chaining `default(...)` at the end if you wanted to.

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,129 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { coveragePercentage } = require('../color-formatters')
+const { GithubAuthV3Service } = require('../github/github-auth-service')
+const { fetchJsonFromRepo } = require('../github/github-common-fetch')
+const { InvalidParameter } = require('..')
+
+const nycrcSchema = Joi.object({
```
PyvesB: The schemas look good 👍

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,129 @@
...
+  }).optional(),
+}).required()
+
+const documentation = `<p>
+  Create a code coverage badge, based on thresholds stored in a
+  <a href="https://github.com/istanbuljs/nyc#common-configuration-options">.nycrc config file</a>
+  on GitHub.
+</p>`
+
+module.exports = class Nycrc extends GithubAuthV3Service {
```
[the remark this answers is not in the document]
bcoe: Done 👍

services/nycrc/nycrc.tester.js
bcoe: I've switched the happy paths to hit the real `yargs/yargs` and, based on your suggestion added a test that hits `badges/shields`.

I left a couple `nock` tests for the unhappy paths 😄

services/nycrc/nycrc.tester.js
bcoe: Done, used `badges/shields` like recommended.

services/nycrc/nycrc.service.js
bcoe: done, the refactor cleaned things up 👍

bcoe:
@PyvesB thanks for the quick code review 👍I'm excited for this badge.

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  static get route() {
+    return {
+      base: 'nycrc',
+      // TODO: eventually add support for .yml and .yaml:
+      pattern: ':user/:repo',
+      queryParamSchema: Joi.object({
+        config: Joi.string()
+          .valid('.nycrc', '.nycrc.json', 'package.json')
```
calebcartwright: It's common with other badges to allow users to specify the repo path to the file, since they're not always in the root of the repo. I suspect we'll need to do something similar here, and could use two different query params (one for directory path and one for file name) if we want to have validation on the specific file names that'll be accepted

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,131 @@
...
+  static render({ coverage }) {
+    return {
+      message: `${coverage.toFixed(0)}%`,
+      color: coveragePercentage(coverage),
+    }
+  }
+
+  transform(config) {
+    const { branches, lines } = config
+    if (branches || lines) {
```
calebcartwright: Would it be worth allowing users to specify whether to use branches or lines via a badge parameter? Can still keep branches as the default, but I can envision folks asking at some point to have control over which is used for their badge

services/nycrc/nycrc.service.js
bcoe: perhaps we should just default to .nycrc, and allow an arbitrary path?

bcoe:
@PyvesB @calebcartwright I believe I've addressed all your suggestions, thanks for the reviews 👌

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,137 @@
...
+      base: 'nycrc',
+      // TODO: eventually add support for .yml and .yaml:
+      pattern: ':user/:repo',
+      queryParamSchema: Joi.object({
+        config: Joi.string()
+          .regex(/(.*\.nycrc)|(.*\.json$)/)
+          .default('.nycrc'),
+        // Allow the default threshold detection logic to be overridden, .e.g.,
+        // favoring lines over branches:
+        preferredThreshold: Joi.string().optional(),
```
calebcartwright: Should probably restrict this to only support the supported values

```suggestion
        preferredThreshold: Joi.string().optional().allow('branches', 'lines'),
```

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,137 @@
...
+    return {
+      message: `${coverage.toFixed(0)}%`,
+      color: coveragePercentage(coverage),
+    }
+  }
+
+  extractThreshold(config, preferredThreshold) {
+    const { branches, lines } = config
+    if (preferredThreshold && config[preferredThreshold]) {
+      return config[preferredThreshold]
```
calebcartwright: I think this is fine if the schema is tightened for `preferredThreshold` values, but if `preferredThreshold` will be accepting any string value then I'd rather throw an `InvalidParameter` here on a bogus value (for example `?preferredThreshold=foobar`)

calebcartwright:
Few final thoughts from me added inline, but not a hard stopper though if you'd rather go ahead and merge @PyvesB

bcoe:
@calebcartwright take a look, ended up adding validation for `preferredThreshold`, and also adding the `functions` threshold to the schema; I can't think of too many reasons why you'd enforce functions, vs., branches, or lines, but might as well make it valid.

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,160 @@
...
+      const nycConfig = pkgJson.c8 || pkgJson.nyc
+      if (!nycConfig) {
+        throw new NotFound({
+          prettyMessage: 'no nyc or c8 stanza found',
+        })
+      } else {
+        coverage = this.extractThreshold(nycConfig, preferredThreshold)
+      }
+    }
+    return this.constructor.render({ coverage })
```
calebcartwright: Since an arbitrary file with a `.json` extension is accepted for the `config` query parameter, we should handle the case where someone provides something like `foo.json` (either by throwing an error or treating it like an nycrc config file and seeing if the key exists). 

As it stands currently, with such a value for `config` we'd display a `NaN%` message like .

For example I deployed a [review app](https://shields-staging-pr-4759.herokuapp.com/) for this PR to test it out live, and was able to reproduce the NaN by using a repo that contains a shared nyc config module I use elsewhere:

https://shields-staging-pr-4759.herokuapp.com/nycrc/swellaby/nyc-config?config=partial-coverage.json


If it doesn't make sense for this badge to support any arbitrary json file (and I can understand why if so), then it'd be better IMO to throw an error instead of rendering NaN

calebcartwright: I think this scenario could also reasonably happen to end users when there's a typo in their config file name (`?config=.ncyrc.json` vs `?config=.nycrc.json`)

calebcartwright:
Sorry for the multiple passes @bcoe. Overall this is in good order, though I did just notice a path that'd result in a potentially confusing coverage value that should be addressed, detailed inline

services/nycrc/nycrc.service.js
bcoe: @calebcartwright good catch. I opted to simply default to assuming `.json` files are in `.nycrc` format, unless we explicitly see `package.json`.

calebcartwright:
@PyvesB - I'm going to go ahead and merge as there's been no major changes since your 👍

bcoe:
@calebcartwright @PyvesB awesome \o/ thanks for landing. Anything we're waiting on for a production deployment? I'm excited to work on a blog post.

calebcartwright:
In our current production setup @paulmelnikow is the only one who can run a deploy to prod, so will have to defer to him on the timing.

Hopefully soon though!

paulmelnikow:
I’ll take care of it tonight!

PyvesB:
@bcoe's blog post about this: https://dev.to/bcoe/enforce-javascript-code-coverage-with-github-actions-36lg
Our Tweet: https://twitter.com/Shields_io/status/1241390201370443776

services/nycrc/nycrc.service.js
```diff
@@ -0,0 +1,160 @@
...
+  }).optional(),
+}).required()
+
+const documentation = `<p>
+  Create a code coverage badge, based on thresholds stored in a
+  <a href="https://github.com/istanbuljs/nyc#common-configuration-options">.nycrc config file</a>
+  on GitHub.
+</p>`
+
+const validThresholds = ['branches', 'lines', 'functions']
```
rosingrind, line 54: @bcoe thank you for the badge, I'm curious if there was any reason not to add `'statements' here?

PR #4780, opened 2020-03-15
Make it easier to benchmark and profile the code

PyvesB:
My grand idea coming into the week-end was to optimise Shields' code and miraculously solve all the performance problems that we've been observing recently.

Admittedly, things did not quite go according to plan. However, I've collated some work that will hopefully prove useful. 😄

Previously, we only had a _benchmark-performance.sh_ script to gather performance data during development. It was unpractical for several reasons:
* Bash is not readily available on Windows, so users like myself could not run it.
* new timings were difficult to add: 4 or 5 different files needed to be modified to do things cleanly.
* development timing code was mixed with the rest of the production code. If we want proper production metrics, we now have Grafana.

I did the following:
* removed anything related to profiling in Shields' production code.
* wrote two small scripts that allow to easily benchmark during development by adding `console.time` statements anywhere in the code. These are in Javascript, therefore they will work on Windows as well. They can be chained together with an NPM script.
* added another NPM script to generate profiling data.
* documented how to run and use things.

Let me know what you think!


calebcartwright:
> My grand idea coming into the week-end was to optimise Shields' code and miraculously solve all the performance problems that we've been observing recently.
>
>Admittedly, things did not quite go according to plan. However, I've collated some work that will hopefully prove useful. 😄

I've still not had a chance to read through this in detail yet, but just so I have the right context when doing so, does this PR have changes that will help with measuring and determining attributes of the app in the production runtime env?

PyvesB:
@calebcartwright no, the changes are mostly aimed at measuring performance locally during development.

.gitignore
```diff
@@ -113,3 +113,6 @@ service-definitions.yml
 
 # Rendered API docs
 /api-docs/
+
+# Flamebearer
+flamegraph.html
```
chris48s: I think you also want to gitignore `flamegraph.png` and remove it from this PR

doc/performance-testing.md
```diff
@@ -0,0 +1,44 @@
...
+
+If you want to change the number of iterations in the benchmark, you can modify
+the values specified by the `benchmark:badge` script in _package.json_. If
+you want to benchmark a specific code path not covered by the static badge, you
+can modify the badge URL in _scripts/benchmark-performance.js_.
+
+## Profiling the full code
+
+Want to have an overview of how the entire application is performing? Simply
+run `npm run profile:server` in your terminal. This will start the server
```
chris48s: I was slightly confused by this at first because it only starts the badge server - not the front-end too. Do you reckon its worth either explaining this in the docs, or also starting the front-end with `concurrently` like we do with `start`?
(don't have a strong preference on which one)

package.json
```diff
@@ -106,6 +106,8 @@
     "start:server:e2e-on-build": "node server 8080",
     "start:server": "cross-env NODE_CONFIG_ENV=development nodemon server 8080",
     "debug:server": "cross-env NODE_CONFIG_ENV=development nodemon --inspect server.js 8080",
+    "profile:server": "cross-env NODE_CONFIG_ENV=development node --prof server 8080",
+    "benchmark:badge": "node scripts/benchmark-performance.js --iterations 10100 | node scripts/capture-timings.js --warmup-iterations 100",
```
chris48s: I tried running this and for me it throws

```
events.js:288
      throw er; // Unhandled 'error' event
      ^

Error: listen EACCES: permission denied 127.0.0.1:80
    at Camp.setupListenHandle [as _listen2] (net.js:1292:21)
    at listenInCluster (net.js:1357:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1496:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:69:10)
Emitted 'error' event on WebSocketServer instance at:
    at Camp.emit (events.js:311:20)
    at Camp.EventEmitter.emit (domain.js:482:12)
    at emitErrorNT (net.js:1336:8)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  code: 'EACCES',
  errno: 'EACCES',
  syscall: 'listen',
  address: '127.0.0.1',
  port: 80
}
Average time over -100 iterations: 0ms
```

Before I spend a lot of time trying to track this down, can you reproduce, or is it just me?
Looking at the diff, I'm surprised we're trying to bind to port 80 here..

doc/performance-testing.md
paulmelnikow: I'd suggest _not_ starting the frontend, as it'll load a bunch of random static badges. I imagine you would use this in conjunction with a script that fires specific requests against localhost (like the current performance benchmark script).

.gitignore
paulmelnikow: I think it's being linked to from docs. If we need to check this in for doc purposes but don't want it in master, another option is the solution I outlined here: https://stackoverflow.com/a/42677655/893113. I don't feel strongly either way.

chris48s: oh yes - sorry. I hasn't clocked it is used in the docs

doc/performance-testing.md
chris48s: Good point :+1:  Lets just note it in the docs then

PyvesB: The docs are hopefully clearer now!

package.json
PyvesB: Could you please try again? This was probably because `NODE_CONFIG_ENV` was missing - I've got a _local.yml_ config which probably hid that problem.

chris48s: Thanks. I still get the same issue with the latest changes, but it works if I set the port in an env var e.g: `PORT=1111 npm run benchmark:badge`. I think you need to add that to the NPM command because we're not explicitly passing a port to `node server` like we usually would.

chris48s:
If I run `PORT=1111 npm run benchmark:badge` with one set of `time`/timeEnd` statements e.g:

```js
console.time('makeBadge')
const svg = makeBadge(badgeData)
console.timeEnd('makeBadge')
```

I get
Average time over **10000** iterations: some number

If I run it with more than one set e.g:

```js
console.time('setCacheHeadersForStaticResource')
setCacheHeadersForStaticResource(ask.res)
console.timeEnd('setCacheHeadersForStaticResource')

console.time('makeBadge')
const svg = makeBadge(badgeData)
console.timeEnd('makeBadge')
```

Average time over **20100** iterations: some number

What's the expected behaviour in that case?

PyvesB:
> What's the expected behaviour in that case?

I mainly had the "one timing" case in mind initially, but I've now made the code more clever to deal with any amount! 😉

package.json
PyvesB: I've switched to `NODE_CONFIG_ENV=test` - that one contains a port and seems to be what is used for most scripts that actually need to start the server.

PR #4804, opened 2020-03-20
don't upgrade any gatsby libs

chris48s:
All the gatsby stuff needs node 10+ now


calebcartwright:
We're rapidly approaching a point where Dependabot will be ignoring all of our dependencies because everything will require Node 10+ 😄

PR #4807, opened 2020-03-21
Refactor checkShouldSkip, affects [Twitch SymfonyInsight Wheelmap]

PyvesB:
Refactor of the `checkShouldSkip` function:
* removes duplicated code in the Twitch, Wheelmap and Symfony Insight tester classes by extracting the function to _test-helpers.js_.
* avoids duplicated log statements: the "service tests will be skipped" warning will now appear once for the entire service rather than for every single test.
* makes things a tiny bit more readable? `.skipWhen(noSymfonyToken)` makes more sense than `.skipWhen(checkShouldSkip)` in my opinion.


PyvesB:
@paulmelnikow thanks for your thoughts!

Things are a bit more complicated: `.skipWhen(!haveSymfonyToken)` isn't valid code as `skipWhen` requires a predicate. Three solutions I can think of:
* use something like `.skipWhen(() => !haveSymfonyToken())`, not very readable IMHO.
* switch to using `.skipIf()`, the drawback being that it is evaluated much earlier in the test lifecycle. This leads to logging "I'm skipping Twitch tests because I don't have a token" then "I'm running Twitch tests" rather then the other way around.
* adding a new negative predicate function in icedfrisby, though the name `skipUnless` to me is ambiguous, I would tend to think it's the opposite of `skipIf` (the boolean-based one), rather then the opposite of `skipWhen`. Having too many variants may lead to confusion.

paulmelnikow:
Right, I forgot these were predicates that got evaluated later. I'd probably suggest one of these:

- Add a `negate` or `not` function which negates a predicate, returning a new predicate. Then you can `.skipIf(not(haveSymphonyToken))`
- Leave this how it is.

I'm fine with either, and don't want to hold this up!

PyvesB:
The simplest is probably to keep it as it is for now. We can always iterate and improve at a later point if need-be! :)

Issue #4808, opened 2020-03-22
dockercloud builds are failing


chris48s:
Looks like this has been consistently failing for the last couple of days.
I don't have a login for dockercloud to get any more info on this.

calebcartwright:
I'm not sure what is up with the link from the PR check, but seems to be failing during dependency install due to Cypress (for some reason :thinking:)
 
https://hub.docker.com/repository/registry-1.docker.io/shieldsio/shields/builds/30289177-3125-4f59-afcc-343c188829cf

```bash
Step 5/14 : COPY package.json package-lock.json /usr/src/app/
---> 43eaac92919c
Step 6/14 : COPY gh-badges /usr/src/app/gh-badges/
---> eb54dfede451
Step 7/14 : RUN NODE_ENV=development CYPRESS_INSTALL_BINARY=0 npm ci
---> Running in f6120315507e
[91mnpm ERR! code ENOENT
npm ERR! syscall spawn git
npm ERR! path git
npm ERR! errno ENOENT
npm ERR! enoent Error while executing:
npm ERR! enoent undefined ls-remote -h -t ssh://git@github.com/cypress-io/request.git
npm ERR! enoent
npm ERR! enoent
npm ERR! enoent spawn git ENOENT
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent
[0m
[91m
npm ERR! A complete log of this run can be found in:
npm ERR! /root/.npm/_logs/2020-03-22T19_55_06_655Z-debug.log
[0m
Removing intermediate container f6120315507e
The command '/bin/sh -c NODE_ENV=development CYPRESS_INSTALL_BINARY=0 npm ci' returned a non-zero code: 1
```

paulmelnikow:
Right. I think the issue is that git it not installed.

Although, the purpose of `CYPRESS_INSTALL_BINARY=0` is to prevent the Cypress install – maybe something changed with that flag? We could also try `--no-scripts`.

calebcartwright:
> Although, the purpose of CYPRESS_INSTALL_BINARY=0 is to prevent the Cypress install

Exactly, that's the bit that's puzzling me. 

> maybe something changed with that flag?

I haven't seen anything and their docs still say that's the way to do it :man_shrugging: 
https://docs.cypress.io/guides/getting-started/installing-cypress.html#Environment-variables

paulmelnikow:
Looks like a regression in Cypress: cypress-io/cypress#6752

[...]

PR #4810, opened 2020-03-22
fix: [node npm] service has bad colors #4809

### LINKED ISSUE #4809: badge request: Node LTS supported engines badge

- [ ] My own instance

:beetle: **Description**

When adding a badge for node supported engines like
If only puts a green color if the package supports node 13 which is wrong in my opinion.
Node odd versions are not stable and most libraries don't bother supporting them

:link: **Link to the badge**

node

:bulb: **Possible Solution**

We need to figure out the latest LTS version of node and not the latest release which might be a non LTS version.
This should be done here:
https://github.com/badges/shields/blob/a2c35a69c0150a7e1d9df94ab14b94d0eb3af473/services/node/node-version-color.js#L9

I found out that you can get the latest version by parsing:
https://github.com/nodejs/Release/blob/master/schedule.json
I will try to work on a quick PR later on today

regevbr:
Fix #4809


regevbr:
Instead of getting the current version (which can be an odd version which is unstable and will never be LTS), I get all the current LTS versions and make sure that the package matches to all of them

regevbr:
The dockercloud tests failed for some reason and the link is broken so I can't check why...

calebcartwright:
> The dockercloud tests failed for some reason

Don't worry about that one, it's unrelated #4808

calebcartwright:
The failing service test is due to changes in this PR though. Can you please take a look whenever you get a chance @regevbr 

https://circleci.com/gh/badges/shields/94808?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```bash
NodeVersion [live] gets the node version of passport [ GET /passport.json ] - [ GET /passport.json ]
  AssertionError: expected [Function] to not throw an error but 'TypeError: Invalid comparator: intermediate' was thrown
```

regevbr:
@calebcartwright this is actually doesn't seem related to me - the message the live tests return is "credentials are misconfigured" and it is also mentioned in the link you sent me.
I'm trying to run it locally and it fails as well, but I'm not sure what those credentials should be... Can you please help?

regevbr:
@calebcartwright I managed to find the reason for this failure, and it is not due to my changes - one or both of NPM_ORIGINS and NPM_TOKEN env variables are missing in the ci process...
After setting those in my dev env, the tests are passing.

calebcartwright:
Yeah, seems like the node failure was spurious. Thanks for checking!

Will take a closer look at this PR over the weekend

regevbr:
@calebcartwright I have implemented a new badge.
Tested in locally and it works great.
Can you please CR?

regevbr:
seems like tests are failing due to issues with github integration and not related to my code

services/node/node-lts.service.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+const NodeVersionBase = require('./node-base')
+const { versionColorForRangeLts } = require('./node-version-color')
+
+module.exports = class NodeLtsVersion extends NodeVersionBase {
+  static get path() {
+    return 'v-lts'
+  }
+
```
calebcartwright: Should probably have different default badge labels between these two services, `node lts` for this one for example, so `defaultBadgeData` may be better off specified explicitly in each respective badge class

services/node/node-version-color.spec.js
```diff
@@ -0,0 +1,54 @@
...
+  getStableVersions,
+  getCurrentVersion,
+  versionColorForRangeLts,
+  versionColorForRangeCurrent,
+} = require('./node-version-color')
+
+describe('node version color', function() {
+  describe(`stable`, function() {
+    it('should only return lts versions', async function() {
+      this.timeout(4000)
```
calebcartwright: I'd really like to avoid running tests like this that are making live API calls within our unit test suite. It'd probably better to test the logic either by setting up [service tests using nock to mock the API responses](https://github.com/badges/shields/blob/master/doc/service-tests.md#5-mocking-responses) or perhaps keeping these as unit tests but using Sinon to mock out the relevant function calls

regevbr: Will do

services/node/node-lts.service.js
regevbr: Will do

calebcartwright:
Thanks for the quick response! Will review later today

regevbr:
@calebcartwright thanks,
 it is not ready yet though, will let you know when it is. It is a lot of work to get all the mockings done as there are many mutations to it.

You should know I found a bug in the npm verison manager where when asked to get version of package with a certain tag (and not scoped) it actually got the latest version info.
I fixed the bug in this PR as well.

regevbr:
@calebcartwright I'm done, waiting for your CR

calebcartwright:
> You should know I found a bug in the npm verison manager where when asked to get version of package with a certain tag (and not scoped) it actually got the latest version info.
I fixed the bug in this PR as well.

Is it possible to open a separate PR with that change, or do you need those changes here to update the node platform support badges?

regevbr:
Sadly I can't separate...

calebcartwright:
> Sadly I can't separate...

Ah sorry, forgot the node classes extended the npm base class

calebcartwright:
> You should know I found a bug in the npm verison manager where when asked to get version of package with a certain tag (and not scoped) it actually got the latest version info.
I fixed the bug in this PR as well

Would you elaborate a bit more on what you believe to be a bug. Is it just for the node service and/or with any npm badges? As noted in the inline comment, the call with `/latest` in `fetchPackageData` was done deliberately for those npm badges to minimize the response size and because the npm badges that consume that function (collaborators, license, dependency version, etc.) don't need or support the `tag` filtering.

The npm version badge is working correctly, but it's not using that same function:

https://github.com/badges/shields/blob/5de4334ea709f9e8f0b464b18c03009bdc3f4b50/services/npm/npm-version.service.js#L92-L96

https://img.shields.io/npm/v/express/next

https://img.shields.io/npm/v/express/latest


If the bug is only for the node platform support badges, then we should find a different way to fix that which doesn't impact all the other npm badges unnecessarily.

services/node/testUtils/packageJsonVersionsTemplate.json
```diff
@@ -0,0 +1,662 @@
+{
```
calebcartwright, line 1: Could we trim this down to the minimal set of keys that are required? It'd really increase readability to only have the core pieces in here that are needed for the tests

services/node/testUtils/test-utils.js
```diff
@@ -0,0 +1,184 @@
...
+
+const getTemplate = template => JSON.parse(templates[template])
+
+const mockPackageData = (
+  packageName,
+  engines,
+  scope,
+  tag,
+  registry
+) => nock => {
```
calebcartwright, line 28: Let's destructure here given the quantity of params so that we don't have to rely on position and passing a lot of falsy values unnecessarily when there's a need to provide a value for `tag` or `registry`
```suggestion
const mockPackageData = ({
  packageName,
  engines,
  scope,
  tag,
  registry,
}) => nock => {
```

services/node/node-current.tester.js
```diff
@@ -0,0 +1,112 @@
...
+const {
+  mockPackageData,
+  mockCurrentSha,
+  mockNonExistingPackageData,
+} = require('./testUtils/test-utils')
+
+function expectSemverRange(message) {
+  expect(() => new Range(message)).not.to.throw()
+}
+
```
calebcartwright, line 15: Let's make sure we do still have live tests as well to provide that level of integration validation our service tests usually do. The mocked tests are great for this uncommon case where it's pretty tricky to test some of the internal logic within `node-version-color` with unit tests, but we definitely want to have the live tests as well.

services/node/node-lts.tester.js
```diff
@@ -0,0 +1,166 @@
...
+  mockPackageData,
+  mockNonExistingPackageData,
+  mockReleaseSchedule,
+  mockVersionsSha,
+} = require('./testUtils/test-utils')
+
+function expectSemverRange(message) {
+  expect(() => new Range(message)).not.to.throw()
+}
+
```
calebcartwright, line 16: Same comment as in `node-current.tester.js` regarding live tests

services/node/node-lts.tester.js
```diff
@@ -0,0 +1,166 @@
...
+t.create('engines satisfies all lts node versions')
+  .get('/passport.json')
+  .intercept(mockReleaseSchedule())
+  .intercept(mockPackageData(`passport`, `10 - 12`))
+  .intercept(mockVersionsSha())
+  .expectBadge({ label: 'node lts', message: `10 - 12`, color: `brightgreen` })
+  .afterJSON(json => {
+    expectSemverRange(json.message)
+  })
```
calebcartwright, line 25: Do we need this for these mocked tests given that `expectBadge` is already doing an exact match on the expected message?

services/node/testUtils/packageJsonVersionsTemplate.json
regevbr: sure thing

services/node/testUtils/test-utils.js
regevbr: yes my bad, will fix it

services/node/node-lts.tester.js
regevbr: Nope, I will remove them in mocked tests

services/node/node-current.tester.js
regevbr: sure thing

services/node/node-lts.tester.js
regevbr: sure thing

regevbr:
@calebcartwright I fixed all your remarks.

Regarding your question:

> > You should know I found a bug in the npm verison manager where when asked to get version of package with a certain tag (and not scoped) it actually got the latest version info.
> > I fixed the bug in this PR as well
> 
> Would you elaborate a bit more on what you believe to be a bug. Is it just for the node service and/or with any npm badges? As noted in the inline comment, the call with `/latest` in `fetchPackageData` was done deliberately for those npm badges to minimize the response size and because the npm badges that consume that function (collaborators, license, dependency version, etc.) don't need or support the `tag` filtering.
> 
> The npm version badge is working correctly, but it's not using that same function:
> 
> https://github.com/badges/shields/blob/5de4334ea709f9e8f0b464b18c03009bdc3f4b50/services/npm/npm-version.service.js#L92-L96
> 
> https://img.shields.io/npm/v/express/next
> 
> https://img.shields.io/npm/v/express/latest
> 
> 
> If the bug is only for the node platform support badges, then we should find a different way to fix that which doesn't impact all the other npm badges unnecessarily.

In order to get npm package information on a specific version or tag, one needs to use the `${registryUrl}/${packageName}` endpoint, which returns ALL the versions of the package and their info. In it, there is a list of tags (`dist-tags`) which you can use to resolve actual version of the latest release, or a custom tag and then get the info on it from the `versions` key.
This endpoint returns a lot of data, most of it is not relevant as you are looking for info about a specific version only.
If you know you want to get the package info only on the latest version, you can use the /latest endpoint which is much more efficient. The only catch is that you can't use it with scoped packages.
In addition, there is no endpoint to get info about a specific version.

My guess is that the first version of `npm-base` used the extended endpoint only. Then someone discovered the latest endpoint and that is why the comment is there. The only thing they missed is handling tags for non scoped packages.

Indeed, non of the code that uses `fetchPackageData` besides the node service, use tags and that is why the bug didn't really effect them.
The `npm-version` service needs refactoring to use the `fetchPackageData` so it will be more efficient and without my fix, this refactoring will introduce the bug to that service as well.

To sum up, I believe my fix is legit and will not harm any other services. Also, we need to open up a feature request to refactor the npm service to utilize the fixed version of `fetchPackageData` and also to add better unit tests to then `fetchPackageData` funciton.

regevbr:
Please note I also change the label to be `node-lts` as it will look better when using with tagged versions, e.g `node-lts@next` vs `node lts@next`

calebcartwright:
> To sum up, I believe my fix is legit and will not harm any other services.

So the other npm badges are still using the `/latest` endpoint right? if so then i'm good, just want to ensure this change wasn't going to cause the all of those other npm badges to use the other endpoint and larger response size

regevbr:
They will not be affected

regevbr:
@calebcartwright sorry for the short answer previously.
The other packages will only be affected as they are not passing tag at all those the condition I added will not be invoked. They will use /latest when the package being checked is not scoped, otherwise they will use the extended endpoint, but that is the current behavior in any case.

Even if in the future they will support tag, they will use the extended endpoint, but that is what they should use, otherwise they will get false data.

calebcartwright:
> @calebcartwright sorry for the short answer previously.

No worries! You answered my question anyway. Will gives this a final lookover later today

services/node/node-current.tester.js
```diff
@@ -13,98 +13,164 @@ function expectSemverRange(message) {
   expect(() => new Range(message)).not.to.throw()
 }
 
-t.create('engines satisfies current node version')
+t.create('gets the node version of passport - live')
```
calebcartwright, line 16: you can drop the ` - live` suffix from the test name. our test runner actually detects whether tests are live or mocked and annotates them accordingly

services/node/node-current.tester.js
```diff
@@ -13,98 +13,164 @@ function expectSemverRange(message) {
...
   .expectBadge({ label: 'node@canary', message: `12`, color: `yellow` })
+
+t.create('gets the node version of passport from a custom registry - live')
+  .get('/passport.json?registry_uri=https://registry.npmjs.com')
+  .expectBadge({ label: 'node' })
   .afterJSON(json => {
     expectSemverRange(json.message)
   })
 
 t.create('engines satisfies current node version with custom registry')
```
calebcartwright, line 159: We really only need the mocked tests to validate logic within `node-version-color`, and we don't want to have any extra mocked tests that are duplicating scenarios which are explicitly validated by a live test.

As such let's remove the test at the bottom of the file that's mocking an invalid package, and I'm also wondering if this one can be deleted too given the live test above

services/node/node-lts.service.js
```diff
@@ -9,7 +9,7 @@ module.exports = class NodeLtsVersion extends NodeVersionBase {
   }
 
   static get defaultBadgeData() {
-    return { label: 'node lts' }
+    return { label: 'node-lts' }
```
calebcartwright, line 12: :+1:

services/node/node-current.tester.js
regevbr: will do

services/node/node-current.tester.js
regevbr: I agree will remove them

regevbr:
@calebcartwright thanks for your review, I have revised the code accordingly

services/node/node-current.tester.js
```diff
@@ -0,0 +1,157 @@
...
+      engines: '12',
+      scope: '@stdlib',
+      tag: '',
+      registry: '',
+    })
+  )
+  .intercept(mockCurrentSha(13))
+  .expectBadge({ label: 'node', message: `12`, color: `yellow` })
+
+t.create("gets the tagged release's node version version of ionic")
```
calebcartwright, line 76: I'm a little surprised that prettier didn't flag the double quotes here (and below) :thinking:

services/node/node-current.tester.js
```diff
@@ -0,0 +1,157 @@
...
+      engines: '>=0.4.0',
+      scope: '@stdlib',
+      tag: '',
+      registry: '',
+    })
+  )
+  .intercept(mockCurrentSha(13))
+  .expectBadge({ label: 'node', message: `>=0.4.0`, color: `brightgreen` })
+
+t.create('engines not satisfies current node version - scoped')
```
calebcartwright, line 62:
```suggestion
t.create('engines does not satisfy current node version - scoped')
```

services/node/node-current.tester.js
```diff
@@ -0,0 +1,157 @@
...
+  .intercept(
+    mockPackageData({
+      packageName: 'passport',
+      engines: '>=0.4.0',
+    })
+  )
+  .intercept(mockCurrentSha(13))
+  .expectBadge({ label: 'node', message: `>=0.4.0`, color: `brightgreen` })
+
+t.create('engines not satisfies current node version')
```
calebcartwright, line 30:
```suggestion
t.create('engines does not satisfy current node version')
```

services/node/node-current.tester.js
```diff
@@ -0,0 +1,157 @@
...
+    })
+  )
+  .intercept(mockCurrentSha(13))
+  .expectBadge({
+    label: 'node@testing',
+    message: `>=0.4.0`,
+    color: `brightgreen`,
+  })
+
+t.create('engines not satisfies current node version - tagged')
```
calebcartwright, line 99:
```suggestion
t.create('engines does not satisfy current node version - tagged')
```

services/node/node-current.tester.js
```diff
@@ -0,0 +1,157 @@
...
+    })
+  )
+  .intercept(mockCurrentSha(13))
+  .expectBadge({
+    label: 'node@canary',
+    message: `>=0.4.0`,
+    color: `brightgreen`,
+  })
+
+t.create('engines not satisfies current node version - scoped and tagged')
```
calebcartwright, line 135:
```suggestion
t.create('engines does not satisfy current node version - scoped and tagged')
```

services/node/node-lts.tester.js
```diff
@@ -0,0 +1,217 @@
...
+  .intercept(
+    mockPackageData({
+      packageName: 'passport',
+      engines: '10 - 12',
+    })
+  )
+  .intercept(mockVersionsSha())
+  .expectBadge({ label: 'node-lts', message: `10 - 12`, color: `brightgreen` })
+
+t.create('engines not satisfies all lts node versions')
```
calebcartwright, line 35:
```suggestion
t.create('engines does not satisfy all lts node versions')
```

services/node/node-lts.tester.js
```diff
@@ -0,0 +1,217 @@
...
+    mockPackageData({
+      packageName: 'stdlib',
+      engines: '10 - 12',
+      scope: '@stdlib',
+    })
+  )
+  .intercept(mockVersionsSha())
+  .expectBadge({ label: 'node-lts', message: `10 - 12`, color: `brightgreen` })
+
+t.create('engines not satisfies all lts node versions - scoped')
```
calebcartwright, line 79:
```suggestion
t.create('engines does not satisfy all lts node versions - scoped')
```

services/node/node-lts.tester.js
```diff
@@ -0,0 +1,217 @@
...
+    })
+  )
+  .intercept(mockVersionsSha())
+  .expectBadge({
+    label: 'node-lts@testing',
+    message: `10 - 12`,
+    color: `brightgreen`,
+  })
+
+t.create('engines not satisfies all lts node versions - tagged')
```
calebcartwright, line 129:
```suggestion
t.create('engines does not satisfy all lts node versions - tagged')
```

services/node/node-lts.tester.js
```diff
@@ -0,0 +1,217 @@
...
+    })
+  )
+  .intercept(mockVersionsSha())
+  .expectBadge({
+    label: 'node-lts@canary',
+    message: `10 - 12`,
+    color: `brightgreen`,
+  })
+
+t.create('engines not satisfies all lts node versions - scoped and tagged')
```
calebcartwright, line 180:
```suggestion
t.create('engines does not satisfy all lts node versions - scoped and tagged')
```

regevbr:
Thank @calebcartwright for the patience and thorough review!

regevbr:
We completely forgot to add text documentations to the badges. I will create a new pr

Issue #4784, opened 2020-03-19
Badge request: POEditor


Cimbali:
:clipboard: **Description**

Showing translation progress per language, from the online platform l10n platform <https://poeditor.com>.

Right now, I make my own static badges for this. For example:

![Chinese: 22%](https://img.shields.io/badge/%F0%9F%87%A8%F0%9F%87%B3%20Chinese-22%25-f60)
![Chinese (TW): 22%](https://img.shields.io/badge/%F0%9F%87%B9%F0%9F%87%BC%20Chinese%20%28TW%29-22%25-f60)
![Spanish: 22%](https://img.shields.io/badge/%F0%9F%87%AA%F0%9F%87%B8%20Spanish-22%25-f60)
![Czech: 98%](https://img.shields.io/badge/%F0%9F%87%A8%F0%9F%87%BF%20Czech-98%25-1f0)
![French: 100%](https://img.shields.io/badge/%F0%9F%87%AB%F0%9F%87%B7%20French-100%25-0f0)
![German: 98%](https://img.shields.io/badge/%F0%9F%87%A9%F0%9F%87%AA%20German-98%25-1f0)
![Polish: 84%](https://img.shields.io/badge/%F0%9F%87%B5%F0%9F%87%B1%20Polish-84%25-5f0)

These contain:
- language name
- flag of country’s language (encoded utf8 symbols)
- progress as a percentage, and color-coded from red (0%) to green (100%)

:link: **Data**

The API 
- is public, 
- requires:
   - a project id
   - a (read-only) API key,
- is described at <https://poeditor.com/docs/api>, and
- works using _HTTP POST_ requests. 

For example for one of my projects, and with my public read-only API key, the request for Chinese would be:

    $ curl -s https://api.poeditor.com/v2/languages/list -d id=323337 -d api_token=7a666b44c0985d16a7b59748f488275c | jq '.result.languages[] | select(.code == "zh-CN")'
    {
      "name": "Chinese",
      "code": "zh-CN",
      "translations": 22,
      "percentage": 22,
      "updated": "2020-03-12T12:29:04+0000"
    }

The equivalent jsonpath expression is `$.result.languages[?(@.code=="zh-CN")]`.

I suppose the basic URL could be something like `/poeditor/:project-id/:language-code`.

---

The flags are of course very optional but they look nice. They are just the 2 country code letters (in the chinese example case CN) encoded as utf-8 “REGIONAL INDICATOR SYMBOL LETTER”. E.g.:

[...]

PR #4812, opened 2020-03-23
[poeditor] Adding POEditor badge

Cimbali:
This is related to issue #4784.

From the steps in the tutorial, so far, I have:
- [x] queried the API
  - [x] added authentication
- [x] added an example
- [x] written tests
- [x] updated documentation

So far, it seems to work:
image

However I’m a little confused and would need help with some of the following considerations:
- Authentication on POEditor is done with public tokens, but per project. All of the authentication I see going on here is server-side. How does a user specify a token? Is there another badge that’s a good example for that?
- There’s a lot of static data in the file right now. Is there a better place for this? Should it be made more compact (even though this specific styling was automatically applied by commit hooks)?
   https://github.com/badges/shields/blob/5d6886f7a12f0b0539858ae3cef22a954841acca/services/poeditor/poeditor.service.js#L25-L365


chris48s:
Thanks for working on this feature. Looks like you're making good progress.

> Authentication on POEditor is done with public tokens, but per project. All of the authentication I see going on here is server-side. How does a user specify a token? Is there another badge that’s a good example for that?

In terms of the token auth:

* If every user has to supply their own token it should just be passed as a GET param. You could look at [CircleCI](https://github.com/badges/shields/blob/8621fe42d76b809e9ba4ae844920f8ed0373817a/services/circleci/circleci.service.js) or [codecov](https://github.com/badges/shields/blob/8621fe42d76b809e9ba4ae844920f8ed0373817a/services/codecov/codecov.service.js) for an example of another service that does this. Env vars are for when we need to set one token on the server which is used by everyone (e.g: to lift a rate limit on anonymous access).
* Like those badges, can you include a documentation key with some HTML help explaining to users how to generate a token with the correct scope.
* Is a token needed in all cases, or are there any public projects we could use in the service/integration tests?

> There’s a lot of static data in the file right now

Unfortunately, I think we're just heading in the wrong direction with this aspect and I'm :-1: on including flags as part of this service for several reasons:

* In general, our guidance is that logos/icons should not be shown by default https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines We don't really need them  https://img.shields.io/badge/German-20%25-yellow communicates all the info we need and the user can customise it if they want.
* We shouldn't take on the maintenance of a list of all countries
* In some cases, mapping languages to country flags may be controversial or divisive
* The emoji flags render correctly in some browsers but not others e.g:

Screenshot at 2020-03-24 22-00-16

Screenshot at 2020-03-24 22-00-31

so we can't deliver this feature in a consistent way.

Adding a shield to show the progress data from the POEditor API is a great feature and I'm looking forward to reviewing the final contribution, but I think we should leave the choice of whether to also customise that badge with a flag icon and how to do that (emoji or custom SVG) as a choice for the user to make if they want it.

Looking forward to reviewing the final version :+1:

calebcartwright:
> Unfortunately, I think we're just heading in the wrong direction with this aspect and I'm -1 on including flags as part of this service for several reasons

I agree, we should not stick the flags in there by default. Folks that really want to include the flags can do so optionally via the `?logo` param and corresponding base64 data (more info on our logo docs [here](https://github.com/badges/shields/blob/master/doc/logos.md#custom-logos))



https://img.shields.io/badge/German-20%25-yellow?logo=data:image/svg+xml;utf8;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iaXNvLTg4NTktMSI/Pgo8IS0tIEdlbmVyYXRvcjogQWRvYmUgSWxsdXN0cmF0b3IgMTkuMC4wLCBTVkcgRXhwb3J0IFBsdWctSW4gLiBTVkcgVmVyc2lvbjogNi4wMCBCdWlsZCAwKSAgLS0+CjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBpZD0iTGF5ZXJfMSIgeD0iMHB4IiB5PSIwcHgiIHZpZXdCb3g9IjAgMCA1MTIgNTEyIiBzdHlsZT0iZW5hYmxlLWJhY2tncm91bmQ6bmV3IDAgMCA1MTIgNTEyOyIgeG1sOnNwYWNlPSJwcmVzZXJ2ZSIgd2lkdGg9IjUxMnB4IiBoZWlnaHQ9IjUxMnB4Ij4KPHBhdGggc3R5bGU9ImZpbGw6IzQ2NDY1NTsiIGQ9Ik00NzMuNjU1LDg4LjI3NkgzOC4zNDVDMTcuMTY3LDg4LjI3NiwwLDEwNS40NDMsMCwxMjYuNjIxdjczLjQ3MWg1MTJ2LTczLjQ3MSAgQzUxMiwxMDUuNDQzLDQ5NC44MzMsODguMjc2LDQ3My42NTUsODguMjc2eiIvPgo8cGF0aCBzdHlsZT0iZmlsbDojRkZFMTVBOyIgZD0iTTAsMzg1LjM3OWMwLDIxLjE3NywxNy4xNjcsMzguMzQ1LDM4LjM0NSwzOC4zNDVoNDM1LjMxYzIxLjE3NywwLDM4LjM0NS0xNy4xNjcsMzguMzQ1LTM4LjM0NSAgdi03My40NzFIMFYzODUuMzc5eiIvPgo8cmVjdCB5PSIyMDAuMDkiIHN0eWxlPSJmaWxsOiNGRjRCNTU7IiB3aWR0aD0iNTEyIiBoZWlnaHQ9IjExMS44MSIvPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8Zz4KPC9nPgo8L3N2Zz4K

Cimbali:
Ah well. It looks my initial commit (5d3d21c35949006a81af45f1f7bdf5f72e07294f) was right on then: no flags and api token as GET param. Will just lose the other 2, add docs and tests, and push an update to this PR when I get around to it.

Cimbali:
NB the API key is always necessary to query the API. I’ve put one of my projects (Cimbali/pympress) with my public key for the live tests.

services/poeditor/poeditor.service.js
```diff
@@ -0,0 +1,115 @@
...
+  <p>
+    As per <a href="https://poeditor.com/docs/api">the POEditor API documentation</a>,
+    <q>all requests to the API must contain the parameter api_token. You can get this key from your POEditor account.
+    You'll find it in <a href="https://poeditor.com/account/api">My Account > API Access</a>.</q>
+  </p>
+`
+
+const schema = Joi.object({
+  response: Joi.object({
+    code: nonNegativeInteger.required(),
```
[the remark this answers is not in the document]
Cimbali: It is a success code that follows similar conventions than the HTTP response code (200 ok, 403 on permission errors) but also with also different values (4011 for invalid tolken).

The actual HTTP response code from their API is always 200 though, even when returning an error (as replicated on the mock tests).

services/poeditor/poeditor.service.js
```diff
@@ -0,0 +1,115 @@
...
+const { BaseJsonService } = require('..')
+
+const documentation = `
+  <p>
+    You must specify the read-only API token from the POEditor account to which the project belongs.
+  </p>
+  <p>
+    As per <a href="https://poeditor.com/docs/api">the POEditor API documentation</a>,
+    <q>all requests to the API must contain the parameter api_token. You can get this key from your POEditor account.
+    You'll find it in <a href="https://poeditor.com/account/api">My Account > API Access</a>.</q>
```
chris48s: I don't have a POEditor account so I don't know exactly what the wording should be, but can we make the instruction here _super clear_ about generating a read-only token.
e.g: "tick the 'read-only' box and click 'generate token' "(or whatever the step is)

services/poeditor/poeditor.service.js
```diff
@@ -0,0 +1,115 @@
...
+  static render({ code, message, language }) {
+    const color = 'red'
+
+    if (code !== 200) {
+      return { message, color }
+    }
+
+    if (language === undefined) {
+      return { message: `Language not in project`, color }
+    }
```
chris48s: If we need to render an error response, can we do this by throwing an `InvalidResponse()` exception rather than manually rendering the error badge.

services/poeditor/poeditor.service.js
Cimbali: I mean there’s nothing to do, it’s literally just there on that page:

image

services/poeditor/poeditor.service.js
Cimbali: Done.

services/poeditor/poeditor.service.js
chris48s: I've pushed a tiny wording change just to mention read-only again. It is important our docs try to help ensure users don't expose a token that allows write access (it still happens).

Issue #4728, opened 2020-03-04
Badge request: Offset Earth for carbon offset/trees


owenvoke:
:clipboard: **Description**

I was thinking of opening a pull request for Offset Earth badges. I know that they are not currently  a service that is used widely by developers, however with projects such as [Treeware](https://treeware.earth) I think that this may become more popular.

I was thinking that to start with this could be two badges, one for the Carbon Offset in tonnes, and one for the number of Trees for an account.

Badge previews (plain text):
- `carbon offset | 5.2 tonnes`
- `trees | 250`

Badge previews (image):

![Screenshot of badges](https://image.prntscr.com/image/lE8x1rqnTaKszJTXwp-1sw.png)

:link: **Data**

The data can be retrieved from their API, however I am awaiting confirmation whether this is ok to use [from their developers on Twitter](https://twitter.com/owenvoke/status/1234876112053948418).

The requests do not require an API key, and there don't appear to be docs (although I think there are for businesses?).

:microphone: **Motivation**

I think that this would be beneficial for companies and users who support Treeware or want to help the environment in general.

I have already created the necessary services for this on a development branch, if people are interested in adding these badges I will open a PR.
https://github.com/badges/shields/compare/master...owenvoke:feature/offset-earth

PyvesB:
Hi @owenvoke , thanks for opening this issue on the repository. Our [badge guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines) do state the following:
> Shields.io hosts integrations for services which are primarily used by developers or which are widely used by developers

However, as said on Twitter, I think we should consider your suggestion anyway. Given the [negative impact that part of our industry is having](https://www.bbc.co.uk/news/technology-48853230), I like the idea of Shields supporting some more socially responsible initiatives. We've already got a great example of this in our codebase, i.e. our [Wheelmap](https://wheelmap.org/) badges.

What do other maintainers think? Should we make another exception here?

paulmelnikow:
A rationale for including Wheelmap is that it can be used for developer events. Can we say something similar here? e.g. can [Offset Earth](https://offset.earth/) be used on software projects?

remotealex:
Hello! Alex here from Offset Earth 👋

Just thought I'd chime in and say that it's all good to use our API. Here are the three endpoints I suspect you'll be interested in:

Tree total
```
GET https://public.offset.earth/users/<username>/trees
=> JSON { total: number }
```

Carbon offset total
```
[...]

PR #4814, opened 2020-03-25
Add [Treeware] badge

owenvoke:
This is related to https://github.com/badges/shields/issues/4728, it adds a badge for the [Treeware](https://treeware.earth) project.

The badge this adds contains tests, and displays the trees that have been planted by a package (based on a referer hash).

### DIFF
--- a/services/treeware/treeware-trees.service.js
+++ b/services/treeware/treeware-trees.service.js
@@ -25,7 +25,7 @@ module.exports = class TreewareTrees extends BaseJsonService {
   static get examples() {
     return [
       {
-        title: 'Treeware',
+        title: 'Treeware (Trees)',
         namedParams: { owner: 'stoplightio', packageName: 'spectral' },
         staticPreview: this.render({ count: 250 }),
       },
--- a/services/treeware/treeware-trees.service.js
+++ b/services/treeware/treeware-trees.service.js
@@ -1,0 +1,63 @@
+'use strict'
+
+const crypto = require('crypto')
+const Joi = require('@hapi/joi')
+const { metric } = require('../text-formatters')
+const { floorCount } = require('../color-formatters')
+const { BaseJsonService } = require('..')
+
+const apiSchema = Joi.object({
+  total: Joi.number().required(),
+}).required()
+
+module.exports = class TreewareTrees extends BaseJsonService {
+  static get category() {
+    return 'other'
+  }
+
+  static get route() {
+    return {
+      base: 'treeware/trees',
+      pattern: ':owner/:packageName',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Treeware',
+        namedParams: { owner: 'jamesmills', packageName: 'laravel-timezone' },
+        staticPreview: this.render({ count: 50 }),
+      },
+    ]
+  }
+
+  static get defaultBadgeData() {
+    return { label: 'trees' }
+  }
+
+  static render({ count }) {
+    return { message: metric(count), color: floorCount(count, 10, 50, 100) }
+  }
+
+  async fetch({ reference }) {
+    const url = `https://public.offset.earth/users/treeware/trees?ref=${reference}`
+    return this._requestJson({
+      url,
+      schema: apiSchema,
+      errorMessages: {
+        404: 'repository not found',
+      },
+    })
+  }
+
+  async handle({ owner, packageName }) {
+    const reference = crypto
+      .createHash('md5')
+      .update(`${owner}/${packageName}`)
+      .digest('hex')
+    const { total } = await this.fetch({ reference })
+
+    return this.constructor.render({ count: total })
+  }
+}
--- a/services/treeware/treeware-trees.service.js
+++ b/services/treeware/treeware-trees.service.js
@@ -41,12 +41,12 @@ module.exports = class TreewareTrees extends BaseJsonService {
   }
 
   async fetch({ reference }) {
-    const url = `https://public.offset.earth/users/treeware/trees?ref=${reference}`
+    const url = `https://public.offset.earth/users/treeware/trees`
     return this._requestJson({
       url,
       schema: apiSchema,
-      errorMessages: {
-        404: 'repository not found',
+      options: {
+        qs: { ref: reference },
       },
     })
   }
--- a/services/treeware/treeware-trees.service.js
+++ b/services/treeware/treeware-trees.service.js
@@ -26,8 +26,8 @@ module.exports = class TreewareTrees extends BaseJsonService {
     return [
       {
         title: 'Treeware',
-        namedParams: { owner: 'jamesmills', packageName: 'laravel-timezone' },
-        staticPreview: this.render({ count: 50 }),
+        namedParams: { owner: 'stoplightio', packageName: 'spectral' },
+        staticPreview: this.render({ count: 250 }),
       },
     ]
   }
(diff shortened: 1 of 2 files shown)


services/treeware/treeware-trees.service.js
```diff
@@ -0,0 +1,63 @@
...
+  static get defaultBadgeData() {
+    return { label: 'trees' }
+  }
+
+  static render({ count }) {
+    return { message: metric(count), color: floorCount(count, 10, 50, 100) }
+  }
+
+  async fetch({ reference }) {
+    const url = `https://public.offset.earth/users/treeware/trees?ref=${reference}`
```
chris48s: Can we pass ref using `queryParams` instead of templating here please

services/treeware/treeware-trees.tester.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+const t = (module.exports = require('../tester').createServiceTester())
+const { isMetric } = require('../test-validators')
+
+t.create('request for existing package')
+  .timeout(10000)
```
chris48s: Is this required, or is it just copied from another test? If our starting assumption is that this will regularly take ~10 seconds to give us a response, I'm a bit concerned that these badges are mostly going to fail to render. That said, all the example calls I've made to this API returned a response quickly.

owenvoke: Ah, I copied it from another badge. Just assumed it was required, but I'll remove it. :+1:

services/treeware/treeware-trees.service.js
owenvoke: Didn't know about `queryParams` that sounds much better. :+1:

PyvesB: Example on how to achieve this here: https://github.com/badges/shields/blob/097bf87e418e0f4242a6d0a8f90c09104d02fadf/services/twitch/twitch.service.js#L64

services/treeware/treeware-trees.service.js
```diff
@@ -0,0 +1,63 @@
...
+const { floorCount } = require('../color-formatters')
+const { BaseJsonService } = require('..')
+
+const apiSchema = Joi.object({
+  total: Joi.number().required(),
+}).required()
+
+module.exports = class TreewareTrees extends BaseJsonService {
+  static get category() {
+    return 'other'
```
PyvesB: We also have a `funding`category, Not sure which one is more appropriate, what do you think?

services/treeware/treeware-trees.service.js
```diff
@@ -0,0 +1,63 @@
...
+    return { message: metric(count), color: floorCount(count, 10, 50, 100) }
+  }
+
+  async fetch({ reference }) {
+    const url = `https://public.offset.earth/users/treeware/trees?ref=${reference}`
+    return this._requestJson({
+      url,
+      schema: apiSchema,
+      errorMessages: {
+        404: 'repository not found',
```
PyvesB: I don't think there are any known cases where the API returns a 404 (i.e. using a random owner and package name return 0 trees): we can probably remove this `errorMessages`parameter.

services/treeware/treeware-trees.service.js
```diff
@@ -0,0 +1,63 @@
...
+      },
+    })
+  }
+
+  async handle({ owner, packageName }) {
+    const reference = crypto
```
PyvesB: Nice and clean way of creating the digest§ 👍

services/treeware/treeware-trees.tester.js
```diff
@@ -0,0 +1,33 @@
...
+  .get('/stoplightio/spectral.json')
+  .expectBadge({
+    label: 'trees',
+    message: isMetric,
+  })
+
+t.create('request for existing profile (mock)')
```
PyvesB: Do you mean "existing package"?

services/treeware/treeware-trees.tester.js
```diff
@@ -0,0 +1,33 @@
...
+  .get('/stoplightio/spectral.json')
+  .intercept(nock =>
+    nock('https://public.offset.earth')
+      .get('/users/treeware/trees?ref=65c6e3e942e7464b4591e0c8b70d11d5')
+      .reply(200, { total: 50 })
+  )
+  .expectBadge({
```
PyvesB: If this test is mocked, you should add the exact count to the expectation as well as the colour - otherwise there is little added value compared to the test that hits the real API.

services/treeware/treeware-trees.tester.js
```diff
@@ -0,0 +1,33 @@
...
+    label: 'trees',
+    message: isMetric,
+  })
+
+t.create('invalid profile (mock)')
```
PyvesB: I would suggest hitting the real API for this invalid package test.

services/treeware/treeware-trees.service.js
owenvoke: Thank you, I'll add that in shortly.

services/treeware/treeware-trees.service.js
owenvoke: I wasn't sure about this, I guess funding is probably more appropriate. 👍

@jamesmills, I was just interested what you'd suggest for this?

owenvoke:
I've applied the requested review comments, and commented where necessary. 👍

services/treeware/treeware-trees.service.js
Gummibeer: I think that `funding` would be okay, but I want to throw in `license` as well. Because it's part of the license and not an optional funding.
The point is that every user is **required** to be trees if he uses the Treeware licensed software in production.

jamesmills: We actually leave it to the package maintainer to decide how to work the Treeware "condition of use". We currently have two in the wild.

> This package is free and open-source, under the MIT license. Use it however you want. This package is Treeware. If you use it in production, then we ask that you buy the world a tree to thank us for our work.

> You're free to use this package, but if it makes it to your production environment you are required to buy the world a tree.

In the first example, it's attached to the licence. However, the maintainer can decide how to word it.

I think it's a licence and would be happiest with this option.

Gummibeer:
Thanks for merging it! 🚀 Is there any plan on when this will be deployed?

PyvesB:
@Gummibeer deployments happen on a regular basis, but it's a manual step that can only be done by a couple of the maintainers. So no precise ETA, but hopefully it won't be too long!

PyvesB:
@platan there's no _shields-deployment_ bot message here, any idea why? 🤔

platan:
@PyvesB Thanks for reporting this problem. 
I'm not sure what was the cause of the problem. Logs in the app which handles webhooks form GitHub are empty. A webhook request send after merging this PR timed out (`We couldn’t deliver this payload: Timed out` message is available in the `shields-deployment` app settings). I redelivered this particular webhook request (it responded in 0.38 seconds) and comment was added. 
Within the last year deployment-status worked properly 404 times (https://github.com/badges/shields/pulls?q=merged%3A%3E2019-03-28+base%3Amaster+-author%3Aapp%2Fdependabot-preview+%22This+pull+request+was+merged+to+master+branch.%22)
In 12 cases it didn't work (https://github.com/badges/shields/pulls?q=merged%3A%3E2019-03-28+base%3Amaster+-author%3Aapp%2Fdependabot-preview+NOT+%22This+pull+request+was+merged+to+master+branch.%22)

PR #4815, opened 2020-03-25
Add [OffsetEarth] badges

owenvoke:
This is related to https://github.com/badges/shields/issues/4728 and https://github.com/badges/shields/pull/4814.

It adds 2 new badges related to the [Offset Earth](https://offset.earth) service which powers [Treeware](https://treeware.earth) and can show both trees planted for an owner, and carbon that has been offset.


services/offset-earth/offset-earth-carbon.tester.js
```diff
@@ -0,0 +1,37 @@
...
+    color: 'green',
+  })
+
+t.create('invalid profile (mock)')
+  .get('/non-existent-username.json')
+  .intercept(nock =>
+    nock('https://public.offset.earth')
+      .get('/users/non-existent-username/carbon-offset')
+      .reply(404, {})
+  )
```
chris48s: Could this be a live test instead of mocked?

services/offset-earth/offset-earth-trees.tester.js
```diff
@@ -0,0 +1,33 @@
...
+    color: 'green',
+  })
+
+t.create('invalid profile (mock)')
+  .get('/non-existent-username.json')
+  .intercept(nock =>
+    nock('https://public.offset.earth')
+      .get('/users/non-existent-username/trees')
+      .reply(404, {})
+  )
```
chris48s: Same comment here.. do we actually need to mock it?

PyvesB:
I would probably add some of the same comments I had in #4814, in particular error messages and names of tests. 😉

services/offset-earth/offset-earth-carbon.tester.js
owenvoke: I'll update this now, apologies for the delay. 🙂

services/offset-earth/offset-earth-carbon.tester.js
```diff
@@ -0,0 +1,32 @@
...
+const { withRegex } = require('../test-validators')
+
+t.create('request for existing profile')
+  .get('/offsetearth.json')
+  .expectBadge({
+    label: 'carbon offset',
+    message: withRegex(/[\d.]+ tonnes/),
+  })
+
+t.create('request for existing profile (mock)')
```
calebcartwright: Do we need this test considering the live one above? If so what are we using this test to validate above and beyond the first one?

Our test runner also automatically flags each test as live or mock in the output so the trailing `(mock)` would be a bit duplicative

services/offset-earth/offset-earth-trees.service.js
```diff
@@ -0,0 +1,60 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { metric } = require('../text-formatters')
+const { floorCount } = require('../color-formatters')
+const { BaseJsonService } = require('..')
+
+const apiSchema = Joi.object({
+  total: Joi.number()
```
calebcartwright: Will these values be whole numbers? If so we can use the `nonNegativeInteger` helper from validators

https://github.com/badges/shields/blob/master/services/validators.js#L15

services/offset-earth/offset-earth-carbon.tester.js
owenvoke: I'll remove this test, and remove the `(mock)` text (it was just taken from another example service). 👍

owenvoke:
I've applied the requested changes. 👍 Let me know if there's anything else needed.

services/offset-earth/offset-earth-carbon.tester.js
calebcartwright: No worries, thanks for updating!

services/offset-earth/offset-earth-carbon.service.js
```diff
@@ -0,0 +1,61 @@
...
+
+const Joi = require('@hapi/joi')
+const { metric } = require('../text-formatters')
+const { floorCount } = require('../color-formatters')
+const { BaseJsonService } = require('..')
+
+const apiSchema = Joi.object({
+  total: Joi.number()
+    .positive()
+    .required(),
```
chris48s, line 11: Should this one also be `nonNegativeInteger`?

owenvoke: It's not an integer, but if there is a `nonNegativeFloat` or something I could use that?

PyvesB: I don't think there is - this is probably fine as it is, if we found ourselves using this too often, we can create one then :)

PyvesB:
I'm still confused as to why the error message says `profile not found`. Should it not be "owner" or "user" or something else?

owenvoke:
I guess I can change it to `owner`, it's just referred to as profile where I've seen it. :+1: I'll change it in a bit.

PyvesB:
If you've seen it referred to as profile, I would leave it as it is! But shouldn't we change the `:owner` name in the badge path as well to keep them consistent? @remotealex what do you think?

remotealex:
@PyvesB Yup. We refer to it as a `profile` on our end, never `owner`. That param is actually called `usernameOrUserId` on our end as it can take either a user's `username` or their user object's `userId`.  

In this case the most appropriate variable name would be `username` and the error message could be `No user found for this username` or similar?

owenvoke:
I'm up for changing it to whatever you guys think would make the most sense. 👍

PyvesB:
> I'm up for changing it to whatever you think would make the most sense. 👍

Based on @remotealex comment above, shall we simply change all occurrences of `owner` and `profile` to `username`?

philsturgeon:
@PyvesB yep please change both owner and profile to `username`. 👍

owenvoke:
Updated all references to use `username`. 👍

PR #4841, opened 2020-03-28
fix: [Spiget] invalid response for external downloads #4840

### LINKED ISSUE #4840: Spiget downloads badge doesn't work if download refers to external site

- [X] My own instance

:beetle: **Description**
On [SpigotMC.org](https://spigotmc.org) resources can be either uploaded directly to their site or the download-button can redirect to an external site. In both cases, it is counted as a download and is registered on SpigotMC and Spiget. The problem is that if resources the download on an external site the response from Spiget is interpreted as invalid and the badge only shows "invalid response data".

:link: **Link to the badge**
https://img.shields.io/spiget/downloads/66647
Spiget Downloads

:bulb: **Possible Solution**
After a short investigation using a local instance, it turns out that the Spiget API reports a file size of `0` and an empty string as the unit if the file is not hosted on SpigotMC but rather on an external site. The problem is that for Joi an empty string is a completely missing property in the object and thus deems the response as invalid.

[Here is the line that causes the problem](https://github.com/badges/shields/blob/77a819cefe113cab8c624f8dab3863fcfe4d0f15/services/spiget/spiget-base.js#L10)

A possible solution would be to split the schema validation up, so the file related stuff would be separate, but that's not a clean solution, is it?

Sytm:
Fixes #4840

Although I am not sure if I should use a mocked API response or a real one (obviously preferred), but I do not know if there are resources on SpigotMC that will stay externally hosted for the foreseeable future so the tests do not break.

Also, I am wondering if the color should change for the download size badge if the resource is hosted externally.

Looks like the service tests didn't run because some credentials are not configured. I had the same problem locally and hoped it would work if it runs in CI


PyvesB:
Thanks for the contribution, using the `type` parameter looks great!
 
> Although I am not sure if I should use a mocked API response or a real one (obviously preferred), but I do not know if there are resources on SpigotMC that will stay externally hosted for the foreseeable future so the tests do not break.

You can use [Advanced Achievements](https://www.spigotmc.org/resources/advanced-achievements.6239/) for that test. I'm the author and don't intend on bringing the downloads back in-house in the foreseeable future. 😉  

> Also, I am wondering if the color should change for the download size badge if the resource is hosted externally.

Would `lightgrey` be more appropriate? We often use the colour for cases where we can't convey the requested information for whatever reason.

> Looks like the service tests didn't run because some credentials are not configured. I had the same problem locally and hoped it would work if it runs in CI

No, I do think think they were successfully run. The warnings before the test suite are for another service, Symfony. Mocha then filters things out and only runs relevant tests within the whole suite.

Sytm:
> No, I do think think they were successfully run. The warnings before the test suite are for another service, Symfony. Mocha then filters things out and only runs relevant tests within the whole suite.

Over [here](https://github.com/badges/shields/blob/master/doc/service-tests.md#3-running-the-tests) it shows some additional output for the service tests that do not appear on my machine (also missing from ci).

I also modified the test code to make it deliberately fail, but the tests "passed" anyways.

PyvesB:
> Over [here](https://github.com/badges/shields/blob/master/doc/service-tests.md#3-running-the-tests) it shows some additional output for the service tests that do not appear on my machine (also missing from ci).
> 
> I also modified the test code to make it deliberately fail, but the tests "passed" anyways.

Hmmm... Locally things are running as expected, however, I'm running a slightly older version of the `master` branch. On the CI, I just noticed the following error message:

> Unable to save test results from /home/circleci/project/junit
> Error path is not valid /home/circleci/project/junit: error accessing path: /home/circleci/project/junit: lstat /home/circleci/project/junit: no such file or directory
> Found no test results, skipping

You're right, things may have broken in the meantime. Will investigate further.

PyvesB:
I've tested things manually with an older version of Mocha, and I believe you'll need to update the mocked data in _spiget-tested-versions.tester.js_, they no longer conform to the updated schema. 😉

Sytm:
> I've tested things manually with an older version of Mocha, and I believe you'll need to update the mocked data in _spiget-tested-versions.tester.js_, they no longer conform to the updated schema. wink

I no longer use mocked data? In 77d75f3 I updated it to use your plugin as the test.

PyvesB:
Sorry for not explaining more clearly. Existing tests, in _spiget-tested-versions.tester.js_, use mock data which happens to be the minimum data required to validate the Joi schema and make those tests pass. The implementation, in _spiget-tested-versions.service.js_, uses the common schema you modified in _spiget-base.js_. The schema now expects a mandatory `type` field in the response, so the mocked data from those other tests should change accordingly. Does this make sense?

Sytm:
> Sorry for not explaining more clearly. Existing tests, in _spiget-tested-versions.tester.js_, use mock data which happens to be the minimum data required to validate the Joi schema and make those tests pass. The implementation, in _spiget-tested-versions.service.js_, uses the common schema you modified in _spiget-base.js_. The schema now expects a mandatory `type` field in the response, so the mocked data from those other tests should change accordingly. Does this make sense?

Ah yes, I thought you were still talking about `spiget-download-size.tester.js`, my bad.

PR #4865, opened 2020-04-03
fix badge CLI debug script

chris48s:
We accidentally broke this when we dropped the .svg extension by default.


chris48s:
> Think we should test the script?

Ideally, yes but equally the impact of this breaking on end-users was nothing happened. I think the same is true for most of the other things in `/scripts`. I think I'm going to raise it as an issue, rather than hold off merging the fix.

PR #4869, opened 2020-04-04
Add optional :version to crates.io version

guswynn:
I noticed that download had support for tagging a specific version, but the simpler, crates version shield didn't
This seemed like a simple addition as the base crates service fetch already supported version

I am very new to this project, can someone help me test?

### DIFF
--- a/services/crates/crates-version.service.js
+++ b/services/crates/crates-version.service.js
@@ -12,7 +12,7 @@ module.exports = class CratesVersion extends BaseCratesService {
   static get route() {
     return {
       base: 'crates/v',
-      pattern: ':crate',
+      pattern: ':crate/:version?',
     }
   }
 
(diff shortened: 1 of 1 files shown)


guswynn:
I think my only failures are `prettier`, ill look into how to fix those

paulmelnikow:
Hi! Thanks for the contribution.

I'm not sure I understand what this is for. Why would you want to display a specific version on the version badge?

It's possible the static badge is a better way to accomplish what you're looking for:


https://img.shields.io/badge/crates.io-v0.3.23-orange

calebcartwright:
I agree with Paul. In such a scenario with a fixed/pre-known version, the static badge would be the way to go. There's no need for shields to have to make an API call to crates.io.

Thanks for the interest and the PR though @guswynn!

guswynn:
@paulmelnikow thats probably what I'm looking for!

curious, when I search for "crate" on https://shields.io/, there is no example of the fixed badge, is that because it generates the badge based on the string in the url? I wonder, would there be a way to make this more discover-able? maybe have a list of known-common static badges as examples so they show up in the search?

paulmelnikow:
Heh, yea, exactly. you don't get it by searching because it can be literally anything you like. It's there below the section headings:

Screen Shot 2020-04-04 at 11 25 55 PM


There's been some discussion of some prerendered static badges, though I think we'd probably reserve that for badges which are truly reusable, like . A badge listing a specific crate version won't be useful to anyone else.

calebcartwright:
I'm going to close this one given the discussion above.

Thanks again for the PR @guswynn and let us know if you have any questions and/or are interested in working on any other items!

PR #4872, opened 2020-04-05
node badges are missing descriptions #4871

### LINKED ISSUE #4871: node badges are missing descriptions

:beetle: **Description**

node badges are missing descriptions

:link: **Link to the badge**

https://shields.io/category/platform-support

:bulb: **Possible Solution**

add descriptions

regevbr:
Fix #4871

### DIFF
--- a/services/node/node-base.js
+++ b/services/node/node-base.js
@@ -17,7 +17,6 @@ module.exports = class NodeVersionBase extends NPMBase {
     const type = this.type
     const documentation = `
   <p>
-    This badge is designed to covey to users the node version support status of a package.
     ${this.documentation}
     The node version support is retrieved from the <code>engines.node</code> section in package.json.
   </p>
--- a/services/node/node-base.js
+++ b/services/node/node-base.js
@@ -17,10 +17,9 @@ module.exports = class NodeVersionBase extends NPMBase {
     const type = this.type
     const documentation = `
   <p>
-    This badge is designed for projects that release packages to <a href="https://www.npmjs.com/">npm</a><br/>
-    The badge can be added to the project readme to inform users of the node versions that the package is supported by<br/>
-    The information is retrieved from the <code>engines.node</code> section in package.json<br/>
+    This badge is designed to covey to users the node version support status of a package.
     ${this.documentation}
+    The node version support is retrieved from the <code>engines.node</code> section in package.json.
   </p>
 `
     const prefix = `node-${type}`
--- a/services/node/node-base.js
+++ b/services/node/node-base.js
@@ -15,6 +15,14 @@ module.exports = class NodeVersionBase extends NPMBase {
 
   static get examples() {
     const type = this.type
+    const documentation = `
+  <p>
+    This badge is designed for projects that release packages to <a href="https://www.npmjs.com/">npm</a>.
+    The badge can be added to the project readme to inform users of the node versions that the package is supported by.
+    The information is retrieved from the <code>engines.node</code> section in package.json.
+    ${this.documentation}
+  </p>
+`
     const prefix = `node-${type}`
     return [
       {
@@ -25,6 +33,7 @@ module.exports = class NodeVersionBase extends NPMBase {
           nodeVersionRange: '>= 6.0.0',
         }),
         keywords,
+        documentation,
       },
       {
         title: `${prefix} (scoped)`,
@@ -34,6 +43,7 @@ module.exports = class NodeVersionBase extends NPMBase {
           nodeVersionRange: '>= 6.0.0',
         }),
         keywords,
+        documentation,
       },
       {
         title: `${prefix} (tag)`,
@@ -44,6 +54,7 @@ module.exports = class NodeVersionBase extends NPMBase {
           tag: 'latest',
         }),
         keywords,
+        documentation,
       },
       {
         title: `${prefix} (scoped with tag)`,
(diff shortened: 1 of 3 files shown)


regevbr:
@calebcartwright I added descriptions to the node badges, which we forgot to add in #4810 
I'm not too good with texts, so please feel free to suggest a more user friendly description and I will change accordingly.

services/node/node-base.js
```diff
@@ -15,6 +15,14 @@ module.exports = class NodeVersionBase extends NPMBase {
 
   static get examples() {
     const type = this.type
+    const documentation = `
+  <p>
+    This badge is designed for projects that release packages to <a href="https://www.npmjs.com/">npm</a><br/>
+    The badge can be added to the project readme to inform users of the node versions that the package is supported by<br/>
+    The information is retrieved from the <code>engines.node</code> section in package.json<br/>
+    ${this.documentation}
+  </p>
```
calebcartwright, line 24: Thanks for following up with these! One thing to keep in mind with the inline badge help text is that there's a general assumption that once Shields users get to this point they are generally familiar with the corresponding service and the basic premise of Shields and utilizing badges.

The inline help text is primarily geared towards things like providing guidance on how users can locate or retrieve the parameter values (for example how to find your project id), explaining the difference between badge A and similarly-looking badge B, etc.

In this context, the most important thing to convey IMO is the distinction between the LTS badge and the current badge, and if possible with links out to relevant info on Node's release cycle. I know the last sentence in the text speaks to that, but I don't think that point is emphasized as much as it could and should be.

If you feel strongly that the additional information is important, I'd still suggest tweaking the text/structure of the text to get a better presentation of the material.

In the not-so-distant past we changed the help text alignment from centered to left which works best in most circumstances, but not so much in cases like this with multiple blocks of text.

Lastly, remember that these badges can work with any custom registry as well, not just the main npmjs registry. As such the current text of that first sentence could be a point of confusion

image

regevbr:
What do you suggest to do about the left align?

services/node/node-current.service.js
```diff
@@ -19,4 +19,8 @@ module.exports = class NodeCurrentVersion extends NodeVersionBase {
   static get colorResolver() {
     return versionColorForRangeCurrent
   }
+
+  static get documentation() {
+    return `The badge will be green if the selected package release supports the current (latest) node version`
```
calebcartwright, line 24: I'd caution against trying to describe the colors within the description (also FWIW the color is `brightgreen`) and instead describe the information that the badge conveys.

The intent of this badge is to allow users (presumably package owners) to convey the status of support their package has for the current/latest node release, and conveying that message with whatever wording makes sense (independent of color mapping) should suffice.

services/node/node-lts.service.js
```diff
@@ -19,4 +19,8 @@ module.exports = class NodeLtsVersion extends NodeVersionBase {
   static get colorResolver() {
     return versionColorForRangeLts
   }
+
+  static get documentation() {
+    return `The badge will be green if the selected package release supports all lts node versions`
```
calebcartwright, line 24: Same comment/theme as the current badge. It's a status of package support across all Node release versions that are in LTS state

calebcartwright:
> What do you suggest to do about the left align?

That's the alignment context of the help text, and not something that can be changed at this time, so you'll need to ensure that the text added for these badges looks decent in that context. 

I'd still suggest removing, or at least trimming, some of the existing text which should help. You can also split content into multiple paragraphs

paulmelnikow:
You can also include HTML tags in there. For example, I think you could wrap each paragraph in a `<p>`. Although these really are just sentences, and I think it would be better to combine it into one paragraph.

calebcartwright:
> For example, I think you could wrap each paragraph in a <p>. Although these really are just sentences, and I think it would be better to combine it into one paragraph.

That's what I was suggesting as well. "paragraphs" in the markup sense, not the grammar sense :smile: I think the main driver is the length of that second sentence which has to be wrapped, and that ends up making the whole block of content look off (at least to my eye)

regevbr:
Thanks guys for the comments. I gave it another shot, but as I said, expressing my self in writing was never a strong suite...

services/node/node-base.js
```diff
@@ -15,6 +15,13 @@ module.exports = class NodeVersionBase extends NPMBase {
 
   static get examples() {
     const type = this.type
+    const documentation = `
+  <p>
+    This badge is designed to covey to users the node version support status of a package.
```
calebcartwright, line 20: I think we can drop this first sentence as it's essentially covered by the second/class specific sentence
```suggestion
```

services/node/node-current.service.js
```diff
@@ -19,4 +19,8 @@ module.exports = class NodeCurrentVersion extends NodeVersionBase {
   static get colorResolver() {
     return versionColorForRangeCurrent
   }
+
+  static get documentation() {
+    return `This badge acts as an indicator that the package is supported by the <b>latest</b> release of node`
```
calebcartwright, line 24:
```suggestion
    return `This badge indicates whether the package supports the <b>latest</b> release of node`
```

services/node/node-lts.service.js
```diff
@@ -19,4 +19,8 @@ module.exports = class NodeLtsVersion extends NodeVersionBase {
   static get colorResolver() {
     return versionColorForRangeLts
   }
+
+  static get documentation() {
+    return `This badge acts as an indicator that the package is supported by <b>all</b> LTS node versions`
```
calebcartwright, line 24:
```suggestion
    return `This badge indicates whether the package supports <b>all</b> LTS node versions`
```

regevbr:
@calebcartwright I applied your suggestions.

regevbr:
Thanks @calebcartwright for helping out here

Issue #3946, opened 2019-09-03
PaaS-friendly metrics 


paulmelnikow:
As discussed at #3874 we're having a serious capacity problem which is having a negative impact on our users. I've proposed an experiment which is to run Shields for a day on four Herkou dynos and compare the performance. IMO the most useful metric would be onboard response time, and ideally we'd have a way to measure it both before and after.

If the Heroku experiment works well, it would be a convenient way to go forward as the deploy and scaling process is easy and transparent, and they have agreed to sponsor us at a level that I expect will fully cover the cost.

There's a challenge to overcome with using Prometheus on Heroku. In https://github.com/badges/shields/issues/3874#issuecomment-523061570 I mentioned:

> 2\. Because the individual servers can't be reached externally, metrics will need to be generated on each server and sent to the metrics server. [This SO post](https://stackoverflow.com/questions/48830134/how-to-monitor-heroku-multi-dynos-app-with-prometheus) outlines two options: one using [pushgateway](https://prometheus.io/docs/practices/pushing/), and the other scraping more frequently and including the `$DYNO` variable in the metrics.

I think my first suggestion would be to try to use pushgateway and see how well that works. What do you think about this option?

/cc @platan

(Related to previous effort at #1848)

paulmelnikow:
Prometheus really seems to discourage this setup and as far as I can tell doesn't have a recipe for a PaaS like Heroku. [In the readme](https://github.com/prometheus/pushgateway#non-goals) they state they chose not to implement TTL because they consider it an anti-pattern. [There is a fork we could consider](https://github.com/pkcakeout/pushgateway#stale-timers-extension) which has implemented a TTL extension.

Alternatively we could solve this on the calling side. From the [pushgateway docs](https://prometheus.io/docs/practices/pushing/):

> The latter point is especially relevant when multiple instances of a job differentiate their metrics in the Pushgateway via an instance label or similar. Metrics for an instance will then remain in the Pushgateway even if the originating instance is renamed or removed. This is because the lifecycle of the Pushgateway as a metrics cache is fundamentally separate from the lifecycle of the processes that push metrics to it. Contrast this to Prometheus's usual pull-style monitoring: when an instance disappears (intentional or not), its metrics will automatically disappear along with it. When using the Pushgateway, this is not the case, and you would now have to delete any stale metrics manually or automate this lifecycle synchronization yourself.

Heroku has a `DYNO` env var: https://devcenter.heroku.com/articles/dynos#local-environment-variables

[...]

PR #4874, opened 2020-04-05
📈 PaaS-friendly metrics

platan:
This PR tries to fix https://github.com/badges/shields/issues/3946 :-)

## Current metrics

- We are using https://github.com/siimon/prom-client to:
    - collect [default metrics](https://github.com/siimon/prom-client#default-metrics) (CPU and memory usage, Node.js-specific metrics), https://metrics.shields.io/d/g_1B7zhik/prom-client-default-metrics
    - collect [custom metrics](https://github.com/badges/shields/blob/557878edb6063934bddf6c0dfa6b30647fca432f/core/server/prometheus-metrics.js#L9-L69), https://metrics.shields.io/d/aESRBSjmz/services?orgId=1
- Each server expose metrics via `/metrics` resource, e.g. https://s0.servers.shields.io/metrics (the server is using self-signed SSL certificates)
- https://prometheus.io is running at metrics.shields.io is used to fetch metrics from all servers (s0, s1, s2) every 15 seconds
- https://grafana.com/grafana/ is running at metrics.shields.io and displays dashboard/charts with out metrics: http://metrics.shields.io
- http://metrics.shields.io is namaged via https://github.com/platan/metrics-shields-io-config

## Problem we want to solve
After moving to Heroku (or possibly other PaaS), we are not able to access metrics (`/metrics` endpoint) from specific instance. 

## Solution
Solution was proposed in this [comment](https://github.com/badges/shields/issues/3874#issuecomment-523061570)
> 2. Because the individual servers can't be reached externally, metrics will need to be generated on each server and sent to the metrics server. 
This PR is implementation of this idea.

- We are still using prom-client to collect all metrics (default + custom). No change in code is needed 
- All Shields servers convert metrics from returned by prom-client as a JSON to [influx format](https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_tutorial/) and push it to metrics.shields.io/telegraf every 15 seconds
- https://github.com/influxdata/telegraf is running at metrics.shields.io, it accepts metrics send by Shields servers in influx format, it is also converts metrics to prometheus format and expose them via HTTP, metrics expire in Telegraf after 20 seconds (Telegraf can expire metrics automatically, which pushgateway cannot)
- Prometheus collect metrics from the Telegraf instance
- Metrics are available in Grafana

### Why converting to Influx format?
Currently Telegraf cannot accept metrics in Prometheus format in [HTTP listener](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/http_listener_v2), but there are plans to add such feature. 

### How to identify instances?
Currently metrics are identified by 2 labels (can be observed at https://metrics.shields.io/d/g_1B7zhik/prom-client-default-metrics): 
- env - identifies service environment, value „shields-io-production” indicates http://shields.io
- instance - identifies particular instance, values: „s0.shields-server.com:443”, „s1.shields-server.com:443” and „s2.shields-server.com:443”

Both labels are added to metrics by Prometheus, metrics in `/metrics` endpoint do not contain such information. 

Now we have add information described above to metrics. 
The `instance` value source can be set using `public.metrics.influx.instanceIdFrom` configuration property. 3 values are allowed:
- `env-var`- a value will be read from an environment variable with a name defined by the `public.metrics.influx.instanceIdEnvVarName` configuration property
- `hostname` - a value will be equal to https://nodejs.org/api/os.html#os_os_hostname`. This value can be mapped using `hostnameAliases`
- `random`, e.g. `fyytvm7el`

The `env` value can be defined using `public.metrics.influx.envLabel` configuration property. 

There is also a new configuration property `public.metrics.prometheus.endpointEnabled`, which allows to enable/disable the `/metrics` endpoint. 

#### production (VPS at OVH)
We can leave current metrics as they are. But we can also use the new approach. This configuration should enable new metrics and disable the old one:
```yaml
public:
  metrics:
    prometheus:
      endpointEnabled: false
    influx:
      enabled: true
      url: https://metrics.shields.io/telegraf
      timeoutMilliseconds: 1000
      intervalSeconds: 15
      instanceIdFrom: hostname
      envLabel:  shields-io-production
      hostnameAliases:
        vps71670: s0.servers.shields.io
        vps244529: s1.servers.shields.io
        vps117870: s2.servers.shields.io
```
\+ `INFLUX_USERNAME` and `INFLUX_PASSWORD` environment variables. 

The code below shows what `os.hostname()` returns at OVH metrics server: 
```bash
m@vps580707:~$ hostname
vps580707
m@vps580707:~$ hostname -f
vps580707.ovh.net
m@vps580707:~$ cat hostname.js
const os = require('os');

console.log('hostname: ', os.hostname());
m@vps580707:~$ ./node-v12.16.1-linux-x64/bin/node hostname.js
hostname:  vps580707
```
I guess our servers should return vps71670, vps244529 and vps117870 for `os.hostname()` (https://github.com/badges/shields/blob/master/doc/production-hosting.md#badge-servers)

#### Heroku

I suggest to use https://devcenter.heroku.com/articles/dyno-metadata#dyno-metadata to identify instances. 
HEROKU_DYNO_ID can be used as an instance id.  

If we want to use Dyno metadata, we have to [enable it](https://devcenter.heroku.com/articles/dyno-metadata#usage). 

In the staging app(https://shields-staging.herokuapp.com) and in review apps we can use app name (`HEROKU_APP_NAME`) as an env value. This will give values `shields-staging` and `shields-staging-pr-XXXX`. We can send metrics from all env (production, staging, review apps) to http://metrics.shields.io/. 

Telegraf is already running at http://metrics.shields.io/ and at http://metrics-test.shields.platan.space (a separate instance I use for testing). Configuration can be found [here](https://github.com/platan/metrics-shields-io-config/blob/b0801cb15c95db084aa532e5cfa8ca8140b2e54c/shields-io-metrics.yml#L292). I already send some metric from my locally running app to https://metrics-test.shields.platan.space/d/g_1B7zhik/prom-client-default-metrics?orgId=1&from=1586112499614&to=1586113011638&var-env=development&var-instance=rpy2hz7cd

Follow ups:
- break out the config into a separate module - https://github.com/badges/shields/pull/4874#discussion_r408379031


platan:
Metrics from https://shields-staging-pr-4874.herokuapp.com/ can be found in https://metrics-test.shields.platan.space/ (`shields-staging-pr-4874` env).

core/server/influx-metrics.js
```diff
@@ -0,0 +1,72 @@
...
+          if (error) {
+            console.log(
+              `Cannot push metrics. Cause: ${error.name}: ${error.message}`
+            )
+          }
+          if (response && response.statusCode >= 300) {
+            console.log(
+              `Cannot push metrics. ${response.request.href} responded with status code ${response.statusCode}`
+            )
+          }
```
chris48s, line 44: If we're just going to log this, we should probably use `console.error()`

That said, we don't really have meaningful access to logs at the moment and this seems like something where if it happens once or twice, we don't really care but if it happens more than N times in a row, we want to know about it. How about keeping an error counter which resets to zero on a successful push and we count more than N errors, throw an exception so we log it to sentry?

platan: I agree, `console.error` is better. How about using https://github.com/badges/shields/blob/4ee26bfcdb8336812c85fb4ac72ebd2c1f76343d/core/server/log.js#L32-L37
which uses `console.error()` and sends exception to Sentry? I would like to start with something simpler by sending all errors to Sentry. 
Are you worried that we will sent too many exceptions to Sentry?
As a next step I would like to add metrics for the number of failed requests and for the response time (for the code above).

chris48s: I'm not too worried about the volume of notifications I just think its inevitable that occasionally we'll hit the timeout and generate a false-positive, but I think you're right - lets keep it simple for now. We can manage the notifications in sentry using "ignore until this occurs another N times" or whatever.
Lets use the `error()` helper as you suggest :+1:

platan: done

core/server/influx-metrics.js
```diff
@@ -0,0 +1,77 @@
...
+const log = require('./log')
+
+module.exports = class InfluxMetrics {
+  constructor(metricInstance, instanceMetadata, config) {
+    this._metricInstance = metricInstance
+    this._instanceMetadata = instanceMetadata
+    this._config = config
+  }
+
+  async registerMetricsEndpoint(server) {
```
paulmelnikow, line 14: I don't think we should have this endpoint turned on in production. Could we remove it? Or put it under a config flag?

core/server/influx-metrics.js
```diff
@@ -0,0 +1,77 @@
...
+    server.route(/^\/metrics-influx$/, (data, match, end, ask) => {
+      ask.res.setHeader('Content-Type', 'text/plain')
+      ask.res.end(this.metrics())
+    })
+  }
+
+  async startPushingMetrics() {
```
paulmelnikow, line 21:
```suggestion
  startPushingMetrics() {
```

core/server/influx-metrics.js
```diff
@@ -0,0 +1,77 @@
...
+            )
+          }
+        }
+      )
+    }
+    this._intervalId = setInterval(
+      sendMetrics,
+      this._config.intervalSeconds * 1000,
+      this._metricInstance,
+      this._instanceMetadata
```
paulmelnikow, line 57: `sendMetrics` doesn't take any arguments, so these last two arguments appear to be unused.

core/server/server.js
```diff
@@ -173,22 +206,25 @@ class Server {
...
-          ','
-        )}`
+    const privateConfig = this.validatePrivateConfig(
+      config.private,
+      privateConfigSchema
+    )
+    if (publicConfig.metrics.influx && publicConfig.metrics.influx.enabled) {
+      this.validatePrivateConfig(
+        config.private,
+        privateMetricsInfluxConfigSchema
```
paulmelnikow, line 227: Why does this need to be validated again?

core/server/influx-metrics.js
```diff
@@ -0,0 +1,77 @@
...
+    )
+  }
+
+  metrics() {
+    return promClientJsonToInfluxV2(this._metricInstance.metrics(), {
+      env: this._instanceMetadata.env,
+      application: 'shields',
+      instance: this._config.hostnameAsAnInstanceId
+        ? this._instanceMetadata.hostname
+        : this._instanceMetadata.id,
```
paulmelnikow, line 67: I like that these values are injected, however the way they flow through all these layers feels like a lot of indirection. I'd love to propose a more direct way of configuring these values and getting them here, though I don't have a better suggestion right now. Maybe I can think about this some more and suggest and improvement in a follow-on.

config/custom-environment-variables.yml
```diff
@@ -6,6 +6,12 @@ public:
   metrics:
     prometheus:
       enabled: 'METRICS_PROMETHEUS_ENABLED'
+    influx:
+      enabled: 'METRICS_INFLUX_ENABLED'
+      uri: 'METRICS_INFLUX_URI'
```
paulmelnikow, line 11: I think we've mostly converted over our usage of URI to URL.

```suggestion
      url: 'METRICS_INFLUX_URL'
```

core/server/server.js
```diff
@@ -82,6 +84,21 @@ const publicConfigSchema = Joi.object({
     prometheus: {
       enabled: Joi.boolean().required(),
```
paulmelnikow, line 85: I think we probably should add a second prometheus flag so you can have the prometheus endpoint turned off while still collecting prometheus metrics. Something like `endpointEnabled: Joi.boolean().default(true)`. That way, we can turn off the `/metrics` endpoint after we've passed the migration period, and also new instances we set up, which don't need the Prometheus endpoint can have it turned off.

core/server/server.spec.js
```diff
@@ -3,162 +3,336 @@
...
+      })
 
-  it('should return the 410 badge for obsolete formats', async function() {
-    const { statusCode, body } = await got(`${baseUrl}npm/v/express.jpg`, {
-      throwHttpErrors: false,
+      it('should require env', function() {
+        delete customInstanceMetadata.env
+        expect(
+          () => new Server(config.util.toObject(), customInstanceMetadata)
+        ).to.throw('"env" is required')
```
paulmelnikow, line 334: I can imagine these config and metadata tests providing a lot of confidence during development. Although, I'm not sure they seem worth keeping / maintaining. I'm open to being convinced.

Or, maybe the issue is that we should keep the tests, but the validation logic is in the wrong place, and that where it really belongs is in the influx object.

core/server/influx-metrics.spec.js
```diff
@@ -0,0 +1,204 @@
...
+              .and(
+                sinon.match.has(
+                  'message',
+                  'Cannot push metrics. http://shields-metrics.io/metrics responded with status code 400'
+                )
+              )
+          )
+        },
+        1000,
+        1
```
paulmelnikow, line 200: 👍

core/server/influx-metrics.spec.js
```diff
@@ -0,0 +1,204 @@
...
+        uri: 'http://shields-metrics.io/metrics',
+        timeoutMillseconds: 50,
+        intervalSeconds: 0,
+        username: 'metrics-username',
+        password: 'metrics-password',
+      })
+
+      influxMetrics.startPushingMetrics()
+
+      await waitForExpect(
```
paulmelnikow, line 186: This would be easier to test if you extract `sendMetrics` to a method which can be invoked directly and returns a promise.

core/server/metrics/format-converters.spec.js
```diff
@@ -0,0 +1,217 @@
...
+      const influx = promClientJsonToInfluxV2(json, {
+        instance: 'instance1',
+        env: 'production',
+      })
+
+      expect(influx).to.be.equal(
+        'prometheus,env=production,instance=instance1 counter1=11'
+      )
+    })
+  })
```
paulmelnikow, line 216: 😁 👍

config/custom-environment-variables.yml
platan: done

core/server/influx-metrics.js
platan: true, done

core/server/influx-metrics.js
platan: This one is only for testing - it's easy to check metrics in influx format (we can pull them instead of pushing it somewhere). I don't have a strong need to save it. Maybe it's not worth to maintain it. I will remove it.

platan:
Thanks @paulmelnikow for the review. I will apply your suggestions soon :-)

core/server/influx-metrics.js
platan: done

core/server/server.js
platan: I want to require `username` and `password` for influx metrics only if influx metrics are enabled. The private config schema and the public config schema are two separate schemas and I did't find a different way to do it.

core/server/influx-metrics.spec.js
platan: `sendMetrics` returning a promise is a good idea and test are simpler now. I left one test with `waitForExpect` to test `influxMetrics.startPushingMetrics`.

core/server/server.js
platan: Done. I also changed the `createTestServer` function, you can check it.

core/server/influx-metrics.js
platan: done

core/server/server.js
```diff
@@ -160,8 +181,20 @@ const privateConfigSchema = Joi.object({
   twitch_client_id: Joi.string(),
   twitch_client_secret: Joi.string(),
   wheelmap_token: Joi.string(),
+  metrics: {
+    influx: privateInfluxConfigSchema,
```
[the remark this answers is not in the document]
platan: done

core/server/server.js
```diff
@@ -82,6 +84,21 @@ const publicConfigSchema = Joi.object({
...
+        .when('enabled', { is: true, then: Joi.required() }),
+      timeoutMilliseconds: Joi.number()
+        .integer()
+        .min(1)
+        .when('enabled', { is: true, then: Joi.required() }),
+      intervalSeconds: Joi.number()
+        .integer()
+        .min(1)
+        .when('enabled', { is: true, then: Joi.required() }),
+      hostnameAsAnInstanceId: Joi.boolean(),
```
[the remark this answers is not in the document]
platan: > I think this could be made a little more self-documenting:
> 
> Maybe it could be `instanceIdFrom: Joi.string().equal('hostname', 'heroku-dyno-id', 'random')`.
> 

I agree this is more self-documenting. In this PR wanted to avoid anything related to Heroku in the code. That's why I added the `INSTANCE_ID` env var and proposed to set it in `Procfile` (`INSTANCE_ID=$HEROKU_DYNO_ID`). 
When someone set `instanceIdFrom = heroku-dyno-id` we will read `$HEROKU_DYNO_ID` as an instance id in the code. 

What do you thing about `instanceIdFrom: Joi.string().equal('hostname', 'env-var', 'random')` and a separate option `instanceIdEnvVarName`, which can be set to `HEROKU_DYNO_ID`? I this this is more general solution.

 
> Also while we're still on the current/legacy hosting setup, if there's a way to fetch the public IP for the production server, we could put the mapping somewhere in code. A little gross, but gets the job done.

Why do you want to fetch the public IP? Did you mean [Cnames](https://github.com/badges/shields/blob/master/doc/production-hosting.md#badge-servers)?
Currently we are reading a hostname. We can add a mapping inside our config. 
```
public:
  metrics:
    influx:
      influx: 
        instanceIdFrom: hostname
        hosnameAliases:
          vps71670: s0.servers.shields.io
          vps244529: s1.servers.shields.io
          vps117870: s2.servers.shields.io
```

core/server/server.spec.js
platan: I would like to keep them, because they are quite simple (but test more complicated cases using `.when()`) and shouldn't require a lot of work in the future. If you prefer I can try to parametrise them - this should shorten them. 

We can also move these tests to a separate file, but I'm not sure influx object is a good place for them. A config validation is performed in the sever object and tests for validation can be located in some file connected with the server.

core/server/server.js
platan: But we do not have to require the `username` and the `password` in the influx config. Maybe it's not worth to keep such complicated code for this validation.

core/server/server.spec.js
paulmelnikow: Let's leave it as is and consider it in a follow-on. If the validation really belongs in the server and not in the metrics, maybe it's time to break out the config into a separate module.

core/server/server.js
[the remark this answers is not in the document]
paulmelnikow: I think it is okay to carefully make references to Heroku in the code. It's not only a platform we are wanting to support for future production. We also support it for staging, review apps, and self-hosting. As long as we're careful to not assume we're on Heroku I don't see why it would be a problem.

That said, I think `instanceIdEnvVarName` in config is a fine compromise.

core/server/server.js
paulmelnikow: Ahh gotcha. Why don't you make the call, and if you think it's worth keeping, add a comment in the code explaining why it's implemented the way it is.

core/server/influx-metrics.spec.js
paulmelnikow: Would you be comfortable using a mock instead?

config/shields-io-production.yml
```diff
@@ -2,6 +2,7 @@ public:
   metrics:
     prometheus:
       enabled: true
+      endpointEnabled: true
```
paulmelnikow, line 5: 👍

core/got-test-client.js
```diff
@@ -3,4 +3,4 @@
 const got = require('got')
 
 // https://github.com/nock/nock/issues/1523
-module.exports = got.extend({ retry: 0 })
+module.exports = got.extend({ retry: 0, throwHttpErrors: false })
```
paulmelnikow, line 6: This client is imported in a lot of places, and I'm not sure I'm convinced `throwHttpErrors: False` should be the default everywhere. (Whereas `retry: 0` avoids a lot of surprises, especially when using mock requests.)

Would it work to define this in the file where it's needed?

core/server/influx-metrics.js
```diff
@@ -0,0 +1,68 @@
...
+      pass: this._config.password,
+    }
+    const request = {
+      uri: this._config.url,
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+      body: this.metrics(),
+      timeout: this._config.timeoutMillseconds,
+      auth,
+    }
+    const postAsync = promisify(post)
```
paulmelnikow, line 27: This could be moved outside of the class so it's not recreated for each call.

server.js
```diff
@@ -46,7 +47,11 @@ if (fs.existsSync(legacySecretsPath)) {
 }
 
 const Server = require('./core/server/server')
-const server = (module.exports = new Server(config))
+const server = (module.exports = new Server(config, {
+  id: process.env.INSTANCE_ID,
+  env: process.env.INSTANCE_ENV || process.env.NODE_CONFIG_ENV || 'unknown',
```
platan, line 52: @paulmelnikow would you like to handle `env` similarly to `instanceIdFrom`? I think we can simply use https://github.com/badges/shields/blob/master/config/custom-environment-variables.yml for reading `env`. 

If we want to remove `env` from the piece of code above, I suggest to remove instance metadata object.

paulmelnikow: Hmm, to keep things simple I'd suggest we do one of these two:

1. Always use NODE_CONFIG_ENV
2. Specify the influx env in config as a string literal, and rely on setting it using the env var set in custom_environment_variables

core/server/influx-metrics.spec.js
platan: A mock should be OK, but I think `sinon.useFakeTimers` + `tickAsync` is slightly better. What do you think about this approach? Of course I can change it to use mocks.

core/server/server.spec.js
platan: I will add this as a TODO task in the end of the PR description.

core/server/server.js
platan: I added a comment with an explanation.

core/server/server.js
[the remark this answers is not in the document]
platan: Done. 

> I think it is okay to carefully make references to Heroku in the code. It's not only a platform we are wanting to support for future production. We also support it for staging, review apps, and self-hosting. As long as we're careful to not assume we're on Heroku I don't see why it would be a problem.

I really like your way of thinking.

core/got-test-client.js
platan: Sure. I moved `throwHttpErrors` to the test which requires it.

core/server/influx-metrics.js
platan: 👍  done

server.js
platan: I chose option 2, because this will work for review apps - we want to set the env value to the PR review application name.

platan:
@paulmelnikow I think I fixed all issues. I will also update PR description. 

I would like to enable dyno metadata in our staging/review apps. Is it OK for you that I will do it? https://devcenter.heroku.com/articles/dyno-metadata#usage

platan:
@paulmelnikow and @chris48s, thank you for the review 👍 
Metrics from https://shields-staging.herokuapp.com are available at https://metrics.shields.io/d/g_1B7zhik/prom-client-default-metrics 🎉 (you have to change env to shields-staging) .

calebcartwright:
Glad to see this landed! Well done everyone :tada:

paulmelnikow:
@platan Do you want to send me some credentials for the production server through Keybase? Or should I snag them from staging?

platan:
I created a separate user for the production. I send you credentials through Keybase.

Issue #4684, opened 2020-02-24
Badge request: Conda license


lewisacidic:
:clipboard: **Description**

A badge displaying the license from conda

:link: **Data**

Conda API, just like the other conda services:

URL: api.anaconda.org/package/:channel/:package

field: license

:microphone: **Motivation**

Get license information for conda-only packages.

PyvesB:
Hello @lewisacidic, thanks for the suggestion! Would you like to submit a pull request? We've got a [nice little tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) to get you started. 😉

lewisacidic:
I've got no time for a few weeks - I'll put it on my list but it might take some time to get to it.

mondeja:
This issue can be closed.

PR #4875, opened 2020-04-05
[Conda] license badge added.

mondeja:
As was requested in #4684, I've added the Conda license badge.


PyvesB:
Thanks for your contribution @mondeja ! Everything looks good to me. 👍 

Does another maintainer want to have a look before I merged this?

services/conda/conda-base.js
```diff
@@ -5,6 +5,7 @@ const { nonNegativeInteger } = require('../validators')
 const { BaseJsonService } = require('..')
 
 const condaSchema = Joi.object({
+  license: Joi.string(),
```
chris48s, line 8: We should either make `license` a `.required()` key, or if it is optional in the response then `CondaLicense` should have its own schema that only requires `license`.

mondeja: Some packages have the license as null. For example [this](https://api.anaconda.org/package/conda-forge/python). I'm not secure if license key are included in all packages, neither if the schema match with null licenses. I'll take a look at it again.

chris48s: OK. If license may be null or missing, then we don't want to make it required to render the platform/version badges. Lets define a custom schema in `conda-license.service.js` for just the license:

```js
const schema = Joi.object({
  license: Joi.string().required(),
}).required()
```

and make it required for the license badge only.

calebcartwright: I forget offhand whether a `null` value would fail that schema validation, but as long as we can throw an error and present a helpful message (like no license found) as opposed to the `invalid response data` that'd be good

services/conda/conda-license.service.js
```diff
@@ -0,0 +1,60 @@
...
+  static render({ licenses }) {
+    return renderLicenseBadge({ licenses })
+  }
+
+  async handle({ channel, pkg }) {
+    const json = await this._requestJson({
+      schema,
+      url: `https://api.anaconda.org/package/${channel}/${pkg}`,
+    })
+    // const json = await this.fetch({ channel, pkg })
```
calebcartwright, line 55: Let's remove this

mondeja: Sorry, it's removed.

calebcartwright: No worries! We all do it :smile:

Issue #2754, opened 2019-01-13
All GitHub badges throw error with no token set


chris48s:
:question: **Question**

It used to be possible to run your dev copy with no `GH_TOKEN` set. The allowance is too low to do a full test run, but the anon allowance is enough to render a few badges in the front-end.
Now if no `GH_TOKEN` is set, all GH badges will fail to render. This is probably unexpected to new/casual contributors.

Is this change intentional, or is it a mistake that has been introduce in the recent work around server secrets/GH auth?

paulmelnikow:
Yes, this is a usability regression that should be fixed. (I mentioned it in the top post of #2733.) Let’s track it here.

chris48s:
Basically the issue here is that if `config.private.gh_token` is set, we'll use that. Otherwise, we'll use polling.

In reality, there should be 3 valid options:

- `config.private.gh_token` is set --> use global token
- `config.private.postgres_url` is set --> use token pooling
- neither `config.private.gh_token` nor `config.private.postgres_url` is set --> use anonymous access

The most important block is here
https://github.com/badges/shields/blob/528165f65c546af969dafa8aa0ccca89ee9740f8/services/github/github-constellation.js#L22-L44

Issue #4862, opened 2020-04-03
Locally Self hosted Docker returns internal error


apobekiaris:
I started the Docker container locally as per docks and the site runs fine. However none of the badges are created. Is there some component i miss? Are there any instructions to debug it?

![image](https://user-images.githubusercontent.com/159464/78384715-5f975280-75e3-11ea-91d3-d18ba728393b.png)

paulmelnikow:
Hi! Shields emits stack traces for the errors to standard output, so you should be able to get more details in the container.

apobekiaris:
Hi

this is what it reprots

```
2020-04-05T17:47:15.381086328Z: [ERROR]      at TokenPool._nextBatch (/usr/src/app/core/token-pooling/token-pool.js:268:11)
2020-04-05T17:47:15.381095828Z: [ERROR]      at TokenPool.next (/usr/src/app/core/token-pooling/token-pool.js:305:20)
2020-04-05T17:47:15.381099828Z: [ERROR]      at GithubApiProvider.tokenForUrl (/usr/src/app/services/github/github-api-provider.js:158:34)
2020-04-05T17:47:15.381103628Z: [ERROR]      at GithubApiProvider.request (/usr/src/app/services/github/github-api-provider.js:172:22)
2020-04-05T17:47:15.381107528Z: [ERROR]      at Promise (/usr/src/app/services/github/github-api-provider.js:211:12)
2020-04-05T17:47:15.381111228Z: [ERROR]      at new Promise (<anonymous>)
2020-04-05T17:47:15.381115328Z: [ERROR]      at GithubApiProvider.requestAsPromise (/usr/src/app/services/github/github-api-provider.js:210:12)
2020-04-05T17:47:15.381118928Z: [ERROR]      at GithubStars._requestFetcher (/usr/src/app/services/github/github-auth-service.js:13:23)
2020-04-05T17:47:15.381129928Z: [ERROR]      at GithubStars._request (/usr/src/app/core/base-service/base.js:230:40)
2020-04-05T17:47:15.381134528Z: [ERROR]      at GithubStars._requestJson (/usr/src/app/core/base-service/base-json.js:47:35)
```

apobekiaris:
here a more complete version just in case

```
Configuration:
{ bind: { address: '0.0.0.0' },
  metrics: { prometheus: { enabled: false } },
  ssl: { isSecure: false },
  cors: { allowedOrigin: [] },
  persistence: { dir: './private' },
  services:
   { github:
[...]

PR #4876, opened 2020-04-06
When token pool is empty, surface the error to the badge consumer; test [github]

paulmelnikow:
Should provide better debuggability in self-hosting cases like #4862 but also on the production server.

You could argue that this could be an ImproperlyConfigured error, though I think the issues are really #2754 and #3371.

### DIFF
--- a/core/token-pooling/token-pool.js
+++ b/core/token-pooling/token-pool.js
@@ -5,6 +5,7 @@
 
 const crypto = require('crypto')
 const PriorityQueue = require('priorityqueuejs')
+const { Inaccessible } = require('../base-service/errors')
 
 /**
  * Compute a one-way hash of the input string.
@@ -265,7 +266,10 @@ class TokenPool {
       }
     }
 
-    throw Error('Token pool is exhausted')
+    throw new Inaccessible({
+      underlyingError: Error('Token pool is exhausted'),
+      prettyMessage: 'no tokens available',
+    })
   }
 
   /**
--- a/core/token-pooling/token-pool.spec.js
+++ b/core/token-pooling/token-pool.spec.js
@@ -3,12 +3,13 @@
 const { expect } = require('chai')
 const sinon = require('sinon')
 const times = require('lodash.times')
+const { Inaccessible } = require('../base-service/errors')
 const { Token, TokenPool } = require('./token-pool')
 
 function expectPoolToBeExhausted(pool) {
   expect(() => {
     pool.next()
-  }).to.throw(Error, /^Token pool is exhausted$/)
+  }).to.throw(Inaccessible, /^Inaccessible: Token pool is exhausted$/)
 }
 
 describe('The token pool', function() {


chris48s:
> but also on the production server.

Won't doing this mean the error gets sent to end users rather than being logged to sentry for us?

paulmelnikow:
> Won't doing this mean the error gets sent to end users rather than being logged to sentry for us?

Yes.

I suppose we could log it in Sentry too.

chris48s:
At the moment, we've got this e.g: https://sentry.io/organizations/shields/issues/1143429301 I think we do want to retain visibility on this error. If we hit this a lot, its the admins that need to know about this problem, not the end user.

paulmelnikow:
> If we hit this a lot, its the admins that need to know about this problem, not the end user.

On the production server that's true, though it's because we know we have plenty of quota.

calebcartwright:
Is there a difference between Shields.io and self-hosting here? Kinda seems like we may not want to surface this to end-users for the main service, but it may make more sense to do for self-hosters

paulmelnikow:
We ask for quota donations. This is the message that, in theory, could mean that we're out of quota.

IMO "internal error" really should mean it's a bug in Shields. I'm not sure why we'd want to obscure the message, but if we did, we could replace it with something generic like "service temporarily unavailable" or "too many requests."

calebcartwright:
Just to clarify, my comment [above](https://github.com/badges/shields/pull/4876#issuecomment-609997912) was intended to be a question, not a statement. 

I think this makes for a better end user experience with more context vs. the current `internal error`. Would be good to also still have a way for the maintainer/ops team to get insight into this (via Sentry or whatever makes sense) should it occur as quickly as possible and not have to wait for a user to file an issue.

paulmelnikow:
Ah, sorry, I misunderstood that. Also I didn't mean to be short, was just trying to be expeditious in my response!

Here are some thoughts:

1. Let's reserve "internal error" for programmer errors.
2. Self-hosting users need a meaningful message displayed on the badge when their quota is up (or not configured, as is the case until #2754 is fixed). Likely they are not setting up Sentry, rather relying on reports from users when badges are broken. We probably should either fix #2754 to allow unauthenticated requests, or improve the error message when there's neither Redis nor a static token configured.
3. I think for users on shields.io, it would be good for the user to know if we do run out of quota, however that seems very unlikely given the amount of quota we have. The more likely case right now, if this message is displayed, is that the server is just starting up and doesn't have its tokens yet (#3371). We could say "tokens not ready or exhausted" which would cover both of those cases. Or maybe implement a "tokens are loading" state so we can provide a better message for that case.
4. I'd like to know when this condition happens, regardless of whether the tokens are exhausted or not loaded yet. Although I wonder if placing a metric on the total number of tokens and the total number of quota calls would be a better way to surface a meaningful statistic. Seeing it in Sentry is feels like mostly a distraction, especially since it's a "normal" part of the server starting up.

There may be more to discuss than we want to do in this PR. We could move this back to an issue if folks think that's better.

chris48s:
> I'd like to know when this condition happens, regardless of whether the tokens are exhausted or not loaded yet.

Yes - this. I don't have strong views on whether its grafana or sentry, but it feels like a step back if we completely lose visibility on it.

---

More generally, I think the reason its a bit unclear what to do with this error is because as far as possible, we want to put errors in a place where they are seen by the person who can do something about it. In this case, its a bit complicated:

For shields.io: On the one hand, _in theory_ as an end user if you're seeing this error you _sort of can_ go and fix it yourself by authorizing our github app (if you haven't already), but "no tokens available" isn't really a clear pathway/call-to-action there. It feels like a bit of a stretch.
I suspect a more realistic situation is actually that we as maintainers would notice that we're routinely hitting this (via sentry or grafana) and put out a call via our comms channels (discord/twitter/etc) that we need more tokens with some explanation of how to contribute a token/assurance we won't totally hammer your API allowance (which is more than you can really fit on a badge).

For self-hosting users: I know this comes out of #4862 and a desire to make this clearer to new users, but I think really if we assume the audience of this error is the _maintainer_ of the instance (as opposed to their end user - in this case, the call-to-action is "go and set `GH_TOKEN`"), this error belongs in logs rather than on a badge and maybe the job is to provide more helpful info in the log, rather than less information on a badge?

paulmelnikow:
> I suspect a more realistic situation is actually that we as maintainers would notice that we're routinely hitting this (via sentry or grafana) and put out a call via our comms channels (discord/twitter/etc) that we need more tokens with some explanation of how to contribute a token/assurance we won't totally hammer your API allowance (which is more than you can really fit on a badge).

This is true, although probably we should update our GitHub badge documentation to bring it more in line with this. Right now it says:

>  If your GitHub badge errors, it might be because you hit GitHub's rate limits. You can increase Shields.io's rate limit by adding the Shields GitHub application using your GitHub account. 

… which is true, though not as true as it could be.

> For self-hosting users: I know this comes out of #4862 and a desire to make this clearer to new users, but I think really if we assume the audience of this error is the _maintainer_ of the instance (as opposed to their end user - in this case, the call-to-action is "go and set `GH_TOKEN`"), this error belongs in logs rather than on a badge and maybe the job is to provide more helpful info in the log, rather than less information on a badge?

On the other hand, I imagine many self-hosting users will not use GitHub badges at all. So it might be a user who tries to use one, and then reaches out to the maintainer of the instance. Something like **github token not set** would be a pretty decent message for either of those audiences.

Also when we fix #2754, the GitHub badges will probably work for a while and then eventually stop working. I think at that point the really descriptive message would be **too many unauthenticated requests**.

calebcartwright:
> Ah, sorry, I misunderstood that.

No worries! I realized my comment was very unclear and just wanted to clarify. 

> Likely they are not setting up Sentry, rather relying on reports from users when badges are broken

Speaking from my own self-hosting experience, I've definitely found this to be the case. I never know there's an issue unless someone pings me with a question or issue with one of their badges.

It makes my job as the admin/maintainer of the self-hosted instance easier when those users have a badge message they can share which can often save me from debugging/checking logs (common example being expired creds and corresponding badge messages indicating auth issues)

calebcartwright:
> There may be more to discuss than we want to do in this PR. We could move this back to an issue if folks think that's better

My sense is that there's agreement and consensus on the core 4 points outlined [above](https://github.com/badges/shields/pull/4876#issuecomment-610511302).

We want something other than `internal error` for the badge message, but also want for the Shields server admins to have insight into token starvation without having to rely on a live badge/report.

Given the extreme unlikelihood that Shields.io will run out of quota, I wouldn't be opposed to merging this and addressing the additional items in a follow up (provided we could do so relatively quickly). However, I'd prefer to defer this for now and roll all the changes together at a later date if we'll need some time to discuss/solve those other points.

paulmelnikow:
> It makes my job as the admin/maintainer of the self-hosted instance easier when those users have a badge message they can share which can often save me from debugging/checking logs (common example being expired creds and corresponding badge messages indicating auth issues)

OMG, a testimony to the value of helpful error messages! 😁

paulmelnikow:
It sounds like the thrust here is to preserve the Sentry error in production, while providing a better error message to the user.

There are some improvements we'd like to make for self-hosting, although I think they can be deferred, especially if the message is actually helpful.

If folks agree, I'll update this accordingly.

calebcartwright:
Closing given the resolution provided by #9427

PR #4877, opened 2020-04-08
Fix TypeError in [SymfonyInsight] transform

### LINKED ISSUE #4816: TypeError: Cannot read property 'violations' of undefined
Sentry Issue: [SHIELDS-4R](https://sentry.io/organizations/shields/issues/1582780843/?referrer=github_integration)

```
TypeError: Cannot read property 'violations' of undefined
  File "/home/m/shields/services/symfony/symfony-insight-base.js", line 96, in SymfonyInsightGrade.transform
    const violationContainer = lastAnalysis.violations
  File "/home/m/shields/services/symfony/symfony-insight-grade.service.js", line 53, in SymfonyInsightGrade.handle
    const { grade, status } = this.transform({ data })
  ?, in null.<anonymous>
```

calebcartwright:
Fixes #4816

### DIFF
--- a/services/symfony/symfony-insight-base.js
+++ b/services/symfony/symfony-insight-base.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const Joi = require('@hapi/joi')
-const { BaseXmlService } = require('..')
+const { BaseXmlService, NotFound } = require('..')
 
 const violationSchema = Joi.object({
   severity: Joi.equal('info', 'minor', 'major', 'critical').required(),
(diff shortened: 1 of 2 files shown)


services/symfony/symfony-insight-base.js
```diff
@@ -87,6 +87,11 @@ class SymfonyInsightBase extends BaseXmlService {
 
   transform({ data }) {
     const lastAnalysis = data.project['last-analysis']
+
+    if (!lastAnalysis) {
+      throw new NotFound({ prettyMessage: 'no analyses found' })
+    }
```
calebcartwright, line 93: Considered addressing this by making `last-analysis` required in the schema, but figured a contextual error message on the badge would be better than the `invalid response data` message especially since we're not sure what sort of projects/scenarios were resulting in all those TypeErrors

calebcartwright:
> ..and I thought I was sooooo clever ;)

:laughing: sorry i should've left a note.

i think i actually went through the process of creating a symfony account back when we were refactoring all the services. assuming i can find the corresponding creds, i've no issues including those in the ci process so that we can actually run the live tests

Issue #4897, opened 2020-04-11
Badge Request: JetBrains Plugin Rating


ChrisCarini:
:clipboard: **Description**

JetBrains provides an ecosystem of plugins for their IDEs. `shields` already has two badges for this ecosystem:

1. `downloads` - https://github.com/badges/shields/blob/master/services/jetbrains/jetbrains-downloads.service.js
1. `versions` - https://github.com/badges/shields/blob/master/services/jetbrains/jetbrains-version.service.js

It would be great to have another one for plugin `ratings`.

Examples:

- `rating | 4.5/5`
- `rating | ★★★★½`

:link: **Data**

The public API can be found here: `https://plugins.jetbrains.com/plugins/list?pluginId=<pluginId>`.

This is already easily accessible via [/services/jetbrains/jetbrains-base.js](https://github.com/badges/shields/blob/master/services/jetbrains/jetbrains-base.js).

:microphone: **Motivation**

It would be awesome to be able to display a shield with a JetBrains plugin rating!

ChrisCarini:
Submitted PR #4898

paulmelnikow:
Resolved via #4898. Thank you!

ChrisCarini:
No problem! Glad to be able to contribute!

PR #4898, opened 2020-04-11
Adding badges for JetBrains Plugin Ratings, run [JetBrainsRating]. #4897

ChrisCarini:
Hello! Please review this PR - first time doing node development, but followed existing JetBrains badge patterns among other badges. Added tests and tested via the local web interface.

### DIFF
--- a/services/jetbrains/jetbrains-rating.service.js
+++ b/services/jetbrains/jetbrains-rating.service.js
@@ -1,0 +1,87 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { starRating } = require('../text-formatters')
+const { colorScale } = require('../color-formatters')
+const JetbrainsBase = require('./jetbrains-base')
+
+const pluginRatingColor = colorScale([2, 3, 4])
+
+const schema = Joi.object({
+  'plugin-repository': Joi.object({
+    category: Joi.object({
+      'idea-plugin': Joi.array()
+        .min(1)
+        .items(
+          Joi.object({
+            rating: Joi.string().required(),
+          })
+        )
+        .single()
+        .required(),
+    }),
+  }).required(),
+}).required()
+
+module.exports = class JetbrainsRating extends JetbrainsBase {
+  static get category() {
+    return 'rating'
+  }
+
+  static get route() {
+    return {
+      base: 'jetbrains/plugin/r',
+      pattern: ':format(rating|stars)/:pluginId',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'JetBrains IntelliJ Plugins',
+        pattern: 'rating/:pluginId',
+        namedParams: {
+          pluginId: '11941-automatic-power-saver',
+        },
+        staticPreview: this.render({
+          rating: '4.5',
+          format: 'rating',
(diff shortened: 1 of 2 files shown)


services/jetbrains/jetbrains-rating.service.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  static render({ rating, format }) {
+    const message =
+      format === 'rating'
+        ? `${+parseFloat(rating).toFixed(1)}/5`
+        : starRating(rating)
+    return {
+      message,
+      label: undefined,
```
chris48s, line 76: Is this `label: undefined` needed? I think you could just remove it.

services/jetbrains/jetbrains-rating.tester.js
```diff
@@ -0,0 +1,84 @@
+'use strict'
+
+const { withRegex, isStarRating } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+const isRatingWithReviews = withRegex(/^(([0-4](.?([0-9]))?)|5)\/5$/)
```
chris48s, line 6: What does "WithReviews" mean here?

Side note: We should probably have a validator for this in `test-validators` I'm going to raise this as an issue because having a quick look over the other rating tests, this seems to be something we're dealing with a bit inconsistently across services. We don't need to do it in this PR.

calebcartwright: Probably referenced from the pkgreview tests. I believe the distinction was made in the name to reflect the fact that some services provide the rating data as well as the count of reviews that contributed to that rating, whereas other services do not.

https://img.shields.io/wordpress/plugin/rating/bbpress


vs:

https://img.shields.io/amo/rating/dustman

services/jetbrains/jetbrains-rating.service.js
ChrisCarini: Hi @chris48s - it does not appear so. Apologies, I'm new to node development. I've gone ahead and removed this and will resolve your comment.

p.s. Thank you for the review! Very much appreciated!

services/jetbrains/jetbrains-rating.tester.js
ChrisCarini: > Probably referenced from the pkgreview tests.

@calebcartwright hit the nail on the head. This was directly lifted from the `pkgreview` tests.

I'll update this to `isRating` to more accurately reflect what I'm checking (not inclusive of 'with reviews') - I'll go ahead and resolve this comment.

ChrisCarini:
Awesome, thank *you*, @chris48s !!

I also saw in the contribution docs that the @shield-deployment bot should have come along with a message - but it seems that failed. I'm assuming that since this change was merged into master it'll go out with the next deploy, correct? :)

calebcartwright:
> I'm assuming that since this change was merged into master it'll go out with the next deploy, correct?

Correct. Sometimes the deployment bot message just doesn't get posted for whatever reason

ChrisCarini:
Ok cool - I saw a few others where the deployment bot also didn't post. Looking forward to seeing this out!

Issue #4900, opened 2020-04-11
github release with no stable includes beta releases without include_preleleases


othercorey:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: 

When a github repo has no stable release, the github release badge will include pre-release beta releases without setting `include_prereleases`. However, release candidate (RC) releases are properly ignored.

You end up with a badge that only shows the last beta release instead of no releases.

All the beta and RC releases on github are marked pre-release.

:link:

https://img.shields.io/github/v/release/cakephp/twig-view?sort=semver&style=flat-square

:bulb:

I assume that the badge should be 'no releases found' when `include_prereleases` is missing.

If the latest pre-release is supposed to show when there are no stable releases, then the badge should include RC releases.

chris48s:
If there is no latest stable semver release, the intent is we should fall back to the latest pre-release here, which should be `1.0.0-RC1` for this repo.
I've opened PR #4901 with one proposed solution.

chris48s:
I'm going to close this as fixed now that https://github.com/badges/shields/pull/4901 is merged and deployed, but its difficult to explicitly verify this particular example as the project now has a stable release anyway :)

othercorey:
Thanks! We ended up going from RC to stable pretty quickly, but will be nice to have for next project!

PR #4901, opened 2020-04-11
improve latest() fallback result if no stable versions available

chris48s:
Refs #4900 

This is one thought, but in the case where we've:

- removed all pre-releases
- tried to sort `[]` as semver
- ended up with `undefined`
- decided to show the "latest" pre-release

should we actually attempt a `semver` sort here instead of a straight string comparison?


calebcartwright:
I may confusing myself trying to eyeball this code :smile:

Am I correctly understanding that the scenario we are trying to address is the selection of the correct version in the case where a user requests a releases badge (without `include_prereleases`) in a repo that has only pre-release versions so we fallback to selecting a pre-rel version?

> should we actually attempt a semver sort here instead of a straight string comparison?

Assuming my understanding is correct, then yes IMO. Otherwise I think there would be the potential (however small) for Shields to show different badges for the latest pre-release version in a repo right?

chris48s:
Yeah, basically, and I think you're right.

One of the things that makes this function particularly confusing to reason about is that it is used in loads of different badges with varying degrees of "how confident are we this is a semver", so its slightly difficult to reason about all the knock-on effects of making a change here.

I'll have another go at it anyway..

services/version.js
```diff
@@ -66,11 +66,9 @@ function latestDottedVersion(versions) {
   return version
 }
 
-// Given a list of versions (as strings), return the latest version.
-// Return undefined if no version could be found.
-function latest(versions, { pre = false } = {}) {
+function latestMaybeSemVer(versions, pre) {
```
chris48s, line 69: Ideas for a better name welcome :D

calebcartwright: I was just chuckling at this :laughing: 

I like it, and it gets my vote!

services/version.js
```diff
@@ -93,10 +91,29 @@ function latest(versions, { pre = false } = {}) {
...
+
+  // if we've still got nothing,
+  // fall back to a case-insensitive string comparison
+  if (version == null) {
+    versions = versions.sort((a, b) =>
+      a.toLowerCase().localeCompare(b.toLowerCase())
+    )
+    version = versions[versions.length - 1]
+  }
+
```
calebcartwright, line 116: I like this flow from a readability perspective, but seems like it could add some additional iterations over the input `versions` in the fallback cases. 

I don't have a good handle on how frequently we have to drop to the fallback case, nor in those cases how mange pre-release versions there typically are (I'd guess/hope not significant amount).

I'd guess it's not likely to have a big impact, but would be good IMO if there was a clean way to derive the full set semver-compatible versions and, if need be, the single numbered versions one time and reuse them in the fallback case.

chris48s: Rather than using regex, we could check `semver.valid()` on every version, although implicitly that's already happening in `rcompare()`: if it fails, we throw.

calebcartwright: I'll preface with the usual disclaimer about code written inline within the GitHub Web UI, but I guess I'm wondering if we could do something like:

```js
function latest(versions, { pre = false } = {}) {
  let version = ''
  const origVersions = versions
  // return all results that are likely semver compatible versions
  versions = origVersions.filter(version => /\d+\.\d+/.test(version))
  // If no semver versions then look for single numbered versions
  if (!versions.length) {
    versions = origVersions.filter(version => /\d+/.test(version))
  }

  version = latestMaybeSemVer(versions, pre)

  if (version == null && !pre) {
    version = latestMaybeSemVer(versions, true)
  }

  // if we've still got nothing,
  // fall back to a case-insensitive string comparison
  if (version == null) {
    versions = versions.sort((a, b) =>
      a.toLowerCase().localeCompare(b.toLowerCase())
    )
    version = versions[versions.length - 1]
  }

  return version
}

function latestMaybeSemVer(versions, pre) {
  let version = ''

  if (!pre) {
    // remove pre-releases from array
    versions = versions.filter(version => !/\d+-\w+/.test(version))
  }
  try {
    // coerce to string then lowercase otherwise alpha > RC
    version = versions.sort((a, b) =>
      semver.rcompare(
        `${a}`.toLowerCase(),
        `${b}`.toLowerCase(),
        /* loose */ true
      )
    )[0]
  } catch (e) {
    version = latestDottedVersion(versions)
  }
  return version
}
```

calebcartwright: I'll also readily admit this could be an unnecessary/premature optimization consideration. I just wanted to point out that it seemed like we'd be adding some extra loops in the fallback case, and I'm thinking we should be able to solve the sorting issue without necessarily having to do so.

chris48s:
Yeah I think the service tests are almost universally checking that whatever comes back matches `isVPlusTripleDottedVersion` or `isVPlusDottedVersionNClauses` etc rather than specific validating ordering behaviour.

I think the tests in `services/version.spec.js` which are part of the core test suite should have our back here

PR #4905, opened 2020-04-13
Fix failing [Nexus] tests

### LINKED ISSUE #4903: Nexus Repository Version Service Tests Failing
~24th March 

:camera: **Live badge**

https://img.shields.io/nexus/proxy-public-3rd-party-release/com.fasterxml.jackson.core/jackson-databind?server=https%3A%2F%2Fnexus.pentaho.org&nexusVersion=3

Yes

:link: **CircleCI link**
https://circleci.com/gh/badges/daily-tests/1036

:beetle: **Stack trace**

```
Nexus [live] Nexus 3 - repository version [ GET /proxy-public-3rd-party-release/com.fasterxml.jackson.core/jackson-databind.json?server=https://nexus.pentaho.org&nexusVersion=3 ]
  ValidationError: message mismatch: "value" with value "v2.11.0.rc1" fails to match the required pattern: /^v\d+(\.\d+)*([-+~].*)?$/
```

:bulb: **Possible solution**
Looks like it just needs a tweak on the schema used for test validation

PyvesB:
Fixes #4903. Tests were failing because `v2.11.0.rc1` does not match `/^v\d+(\.\d+)*([-+~].*)?$/`.

I've changed the test data rather than the regex. Unless I'm missing an obvious trick, the latter would be quite hard to change without making validation significantly looser. The obvious solution, `/^v\d+(\.\d+)*([-+~.].*)?$/` (extra dot in the allowed char list) would start validating strings such as `v2.11.0€rc1`, as it would capture from the previous dot - it would be pretty much equivalent to `/^v\d+(\.\d+)*(.*)?$/`.

Versions in the `v2.11.0.rc1` format are quite rare anyway, I figured there was little value in changing the validator for such an edge-case, but I can change things if you disagree!

### DIFF
--- a/services/nexus/nexus.tester.js
+++ b/services/nexus/nexus.tester.js
@@ -264,7 +264,7 @@ t.create('Nexus 3 - search snapshot version for artifact without snapshots')
 
 t.create('Nexus 3 - repository version')
   .get(
-    '/proxy-public-3rd-party-release/com.fasterxml.jackson.core/jackson-databind.json?server=https://nexus.pentaho.org&nexusVersion=3'
+    '/proxy-public-3rd-party-release/com.h2database/h2.json?server=https://nexus.pentaho.org&nexusVersion=3'
   )
   .expectBadge({
     label: 'nexus',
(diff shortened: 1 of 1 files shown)


calebcartwright:
I don't want to bikeshed on this too much since the most important thing is that we get the tests working again, but is there any reason we can't use the same validator in the tests that the service class does? 

I recall there being some complexity in the Nexus schemas so is this a case where the the service class is very relaxed (a `Joi.string()`)?

PyvesB:
> I recall there being some complexity in the Nexus schemas so is this a case where the the service class is very relaxed (a `Joi.string()`) but we want something a bit more strict in the tests?

Yes, I think it's one of those, the service class uses `version: Joi.string()`

calebcartwright:
Added the label to have ranger go ahead and merge this. We can revisit if and when any issue actually recurs

PR #4907, opened 2020-04-15
fix TypeError in [DockerVersion] badge

calebcartwright:
Refs #4904. It should actually fix that issue but going to defer closing it for now to confirm there's no more issues reported in Sentry

### DIFF
--- a/services/docker/docker-version.service.js
+++ b/services/docker/docker-version.service.js
@@ -97,14 +97,15 @@ module.exports = class DockerVersion extends BaseJsonService {
       if (version !== 'latest') {
         return { version }
       }
-      if (Object.keys(data.results[0].images).length === 0) {
+      const imageTag = data.results[0].images.find(
+        i => i.architecture === 'amd64'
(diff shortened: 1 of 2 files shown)


chris48s:
In the case where there's some data in the array but none where `architecture === 'amd64'` but there is some other architecture, shall we just fall back and use another key? I'm not sure `amd64` is actually crucial, we've just used it on the assumption it is usually populated?

calebcartwright:
> I'm not sure amd64 is actually crucial, we've just used it on the assumption it is usually populated

When the badge was first added, we discussed including a query param to allow users to specify the target architecture (still defaulting to amd64 as that's by far the most common) but elected to defer that to a follow up. 

I'd suggest maintaining the current behavior (amd64 or nothing) with this change to fix the runtime error, and then implementing that arch param in a follow up vs. trying to select and codify a default fallback/secondary arch

calebcartwright:
Issue opened to track extended architecture support: #4911

Issue #4656, opened 2020-02-15
AUR Info


Raymo111:
:clipboard: **Description**

So much more info could be captured from AUR!
 - Popularity
 - Maintainer
 - Last Packager
 - Last Updater
 - Package base
 - Out-of-date or not
 - PKGBUILD Version

:link: **Data**

Same wasy as existing AUR badges.

:microphone: **Motivation**

I think it's cool. It's also informative. Let me know if you want help with this. I'd gladly implement all of the above.

calebcartwright:
hi @Raymo111 👋 thanks for reaching out, and the willingness to contribute!

As you probably already know, Shields currently provides `version`, `license`, and `votes` badges for AUR
https://img.shields.io/aur/license/android-studio
![](https://img.shields.io/aur/license/android-studio)

https://img.shields.io/aur/version/android-studio
![](https://img.shields.io/aur/version/android-studio)

https://img.shields.io/aur/votes/android-studio
![](https://img.shields.io/aur/votes/android-studio)

These badges are constructed utilizing the same AUR endpoint
`https://aur.archlinux.org/rpc.php?type=info&arg=<packageName>`

For example, here's the API call and response for `android-studio`
https://aur.archlinux.org/rpc.php?type=info&arg=android-studio
```json
{
  "version": 1,
  "type": "info",
  "resultcount": 1,
  "results": {
    "ID": 675651,
    "Name": "android-studio",
    "PackageBaseID": 69975,
    "PackageBase": "android-studio",
    "Version": "3.5.3.0-1",
    "Description": "The official Android IDE (Stable branch)",
    "URL": "https://developer.android.com/",
    "NumVotes": 893,
    "OutOfDate": null,
    "Maintainer": "kordianbruck",
    "FirstSubmitted": 1368647110,
    "LastModified": 1576097257,
    "License": "Apache",
    "URLPath": "/cgit/aur.git/snapshot/android-studio.tar.gz",
    "CategoryID": 1
  }
}
```

A couple of the data points enumerated in the OP do indeed appear available within the API response (`OutOfDate` for example) and seem viable, but there are some that I don't see a corresponding data point for in the response (like `Popularity` and `LastPackager`) and others that do seem to be represented in the response but that I have questions on (for example, how dynamic is the `maintainer` field/could that be more easily conveyed via a static badge?).

As such, I think it'd be a good idea to provide the how/why detail for each of the individual badges you've proposed so that we can figure out which ones are possible and which ones may not be.

[...]

PR #4952, opened 2020-04-26
Add [AUR] maintainer and last modified badges

mrf345:
Hello there! I'm adding two additional badges to the `AUR` service, [_related issue_]( https://github.com/badges/shields/issues/4656):

- image

-  image

Thanks for the awesome service!

### DIFF
--- a/services/aur/aur.service.js
+++ b/services/aur/aur.service.js
@@ -7,6 +7,7 @@ const { nonNegativeInteger } = require('../validators')
 const { BaseJsonService, NotFound } = require('..')
 const { formatDate } = require('../text-formatters')
 const { age: ageColor } = require('../color-formatters')
+const { InvalidResponse } = require('..')
 
 const aurSchema = Joi.object({
   resultcount: nonNegativeInteger,
@@ -45,6 +46,8 @@ class BaseAurService extends BaseJsonService {
   }
 
   async fetch({ packageName }) {
+    // Please refer to the Arch wiki page for the full spec and documentation:
+    // https://wiki.archlinux.org/index.php/Aurweb_RPC_interface
     return this._requestJson({
       schema: aurSchema,
       url: 'https://aur.archlinux.org/rpc.php',
@@ -176,7 +179,7 @@ class AurVersion extends BaseAurService {
 
 class AurMaintainer extends BaseAurService {
   static get category() {
-    return 'social'
+    return 'other'
   }
 
   static get route() {
@@ -205,7 +208,9 @@ class AurMaintainer extends BaseAurService {
   }
 
   async handle({ packageName }) {
-    const { results: { Maintainer: maintainer } } = await this.fetch({ packageName })
+    const {
+      results: { Maintainer: maintainer },
+    } = await this.fetch({ packageName })
     if (!maintainer) {
       throw new InvalidResponse({ prettyMessage: 'No maintainer' })
     }
--- a/services/aur/aur.service.js
+++ b/services/aur/aur.service.js
@@ -21,7 +21,9 @@ const aurSchema = Joi.object({
       NumVotes: nonNegativeInteger,
       Version: Joi.string().required(),
       OutOfDate: nonNegativeInteger.allow(null),
-      Maintainer: Joi.string().required(),
+      Maintainer: Joi.string()
+        .required()
+        .allow(null),
       LastModified: nonNegativeInteger,
     }).required()
   ),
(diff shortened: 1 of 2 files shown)


mrf345:
I guess that's a flaky [test result](https://circleci.com/gh/badges/shields/99008?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) 🤔  it's passing locally:

image

PyvesB:
Thanks for you contribution @mrf345 ! Let's see if re-running the tests makes them pass. 😉

mrf345:
Hi @PyvesB , my pleasure! I don't have write access to rerun the tests

image

so i'm just changing the ref and force pushing with `git commit --amend && git push -f` I tried it couple of times yesterday and just now but same [failing tests](https://circleci.com/gh/badges/shields/99027?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). yet the tests locally are passing 🤔

image

PyvesB:
Re-running the test suite for this PR did not help. That test did not fail in [last daily CI run](https://app.circleci.com/pipelines/github/badges/daily-tests/173/workflows/a0b8d1c9-17bb-4a66-893f-376ec1453b04/jobs/1068/tests) though. Maybe the API response for `gog-gemini-rue` is missing a `Maintainer` or `LastModified` field, which you've added to the AUR schema in this PR? Could you please check that?

mrf345:
Thanks man. You're 100% on point! turns out i was missing `.allow(null)` in `Maintainer` schema.
Just managed to reproduce it locally after running `npx mocha --delay core/service-test-runner/cli.js --only=aur`:

image

should be passing soon 🤞

services/aur/aur.service.js
```diff
@@ -168,8 +174,85 @@ class AurVersion extends BaseAurService {
...
+    return { label: 'maintainer' }
+  }
+
+  static render({ maintainer }) {
+    return { message: maintainer, color: 'blue' }
+  }
+
+  async handle({ packageName }) {
+    const json = await this.fetch({ packageName })
+    return this.constructor.render({ maintainer: json.results.Maintainer })
```
calebcartwright, line 209: Given that we had to update the schema to account for the `Maintainer` key having a null value, we'll need to check for that scenario and handle it accordingly. Something like the below could work (obligatory disclaimer about potential prettier/eslint/etc. issues with inline code typed in the GitHub UI), note you'd need to also add the import `InvalidResponse` from `..` up on line 7

```suggestion
    const { results: { Maintainer: maintainer } } = await this.fetch({ packageName })
    if (!maintainer) {
      throw new InvalidResponse({ prettyMessage: 'No maintainer' })
    }
    return this.constructor.render({ maintainer })
```

services/aur/aur.service.js
```diff
@@ -168,8 +174,85 @@ class AurVersion extends BaseAurService {
...
+
+  static render({ date }) {
+    const color = ageColor(date)
+    const message = formatDate(date)
+    return { color, message }
+  }
+
+  async handle({ packageName }) {
+    const json = await this.fetch({ packageName })
+    const date = 1000 * parseInt(json.results.LastModified)
```
calebcartwright, line 247: I don't suppose there's a spec/doc for this API anywhere is there? If so might be helpful to include that link in an inline comment for future eyes

services/aur/aur.service.js
```diff
@@ -168,8 +174,85 @@ class AurVersion extends BaseAurService {
   }
 }
 
+class AurMaintainer extends BaseAurService {
+  static get category() {
+    return 'social'
```
calebcartwright, line 179: The category for this one is a little tricky, but I'm thinking we should use a different one (maybe `other`). The social badges are a different breed and almost always have a very unique style (https://shields.io/category/social).

This is more of just an additional attribute badge for a package

services/aur/aur.service.js
mrf345: Thanks for the great insight and example! I've missed that fail-safe. although, i thought there's some sort of generic error handle that catch them.

services/aur/aur.service.js
mrf345: There's am arch [wiki page](https://wiki.archlinux.org/index.php/Aurweb_RPC_interface) it has a reference of the expected parameters. But no details regarding data-types.

services/aur/aur.service.js
mrf345: Good call! `others` is definitely a better fit.

PR #4956, opened 2020-04-27
Get data for [Discord] badges via OVH server proxies

calebcartwright:
There's more details on the underlying issue in our Discord channels, but for a tl;dr.. 

Discord throttles calls to their API, and they'd whitelisted our 3 OVH servers previously to allow us to get the requisite data. When we routed img.shields.io traffic over to our Heroku environment we lost that whitelisting and our badge servers (on Heroku) hit the rate limits pretty quickly resulting in `invalid` for all our Discord badges.

This is a temporary fix to get the Discord badges working by having our Heroku-based badge servers use the OVH servers as a proxy to get the Discord data.

### DIFF
--- a/services/discord/discord.service.js
+++ b/services/discord/discord.service.js
@@ -69,7 +69,7 @@ module.exports = class Discord extends BaseJsonService {
 
   constructor(context, config) {
     super(context, config)
-    this._runningOnHeroku = config.shieldsProductionHerokuHacks
+    this._shieldsProductionHerokuHacks = config.shieldsProductionHerokuHacks
   }
 
   async fetch({ serverId }) {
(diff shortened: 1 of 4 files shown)


services/discord/discord.service.js
```diff
@@ -62,6 +67,11 @@ module.exports = class Discord extends BaseJsonService {
     }
   }
 
+  constructor(context, config) {
+    super(context, config)
+    this._runningOnHeroku = config.shieldsProductionHerokuHacks
```
paulmelnikow, line 72: Should we sync up these names? e.g. use shieldsProductionHerokuHacks for the instance var?

paulmelnikow:
Thanks for addressing this, Caleb!

services/discord/discord.service.js
calebcartwright: yeah, missed that one somehow. updating now

dodocodes:
👀 thank you for this!

PR #4959, opened 2020-04-27
Delay start until the server is ready

### LINKED ISSUE #4958: Heroku-friendly startup
In the Heroku ops experiment (#4929) we're seeing a lot of dropped requests at the time the server restarts. This lasts only a few seconds. During a peak-time deployment today we dropped about 1000 requests, which is not great. Heroku kicks the servers every 24 hours even when nothing has happened… so it's pretty important that we can fail over smoothly!

With the legacy servers the changeover downtime lasted several minutes, so it's already a big improvement that we can deploy during peak hours.

Heroku is designed for smooth failover, however it requires that we do not bind the HTTP port until the server is completely set up. This requires a modification to Scoutcamp, the http server framework Shields is built on. Since we're hoping to phase it out (#3328) I think a quick and dirty patch is appropriate. I've forked Scoutcamp and opened badges/sc#1.

This will fix #3771 for the Heroku deployment, too.

paulmelnikow:
Fixes #4958

Fixes only in Heroku #3771.

I'd rather not keep in the GitHub dependency, so let's wait to merge this until we publish a fork with the change in badges/sc#1.


paulmelnikow:
I think the Docker build is failing because there is no Git in the container.

calebcartwright:
:tada: https://www.npmjs.com/package/@shields_io/camp

PR #4968, opened 2020-04-29
Support [Codecov] coverage badge with flag

### LINKED ISSUE #3512: Codecov monorepo support
:clipboard: **Description**

I have a code coverage badge for codecov. It takes the overall coverage value for the whole repository.
I would like to include the values per project (or directories).
[Codecov, at least in the UI displays like](https://codecov.io/gh/getsentry/sentry-dotnet/tree/master/src). I'm unsure of any APIs:

image

image

My goal is to have the coverage value on the table with the package links.
Is there a way already to do this? Or does it make sense to create a new badge for it?

MunifTanjim:
Instead of using the Codecov API, this change reads them by scrapping the Codecov badge itself (e.g.: https://codecov.io/github/codecov/example-node/branch/master/graph/badge.svg)

This is done because, Codecov doesn't exposes their [flags feature](https://docs.codecov.io/docs/flags) through API. But we can get the coverage for a specific flag by scrapping their badge. This is useful for monorepo setups.

resolves #3512


calebcartwright:
Hi @MunifTanjim, thanks for the PR! 

I've tweaked the PR title so that our CI process will execute the tests (more detail [here](https://github.com/badges/shields/blob/master/doc/service-tests.md#pull-requests) and as you can see there's quite a few test failures.

Could you take a look at those when you get a chance?

MunifTanjim:
@calebcartwright Thanks for adjusting the PR title, I missed reading that section of the docs.

I've fixed the failed tests. Could a take a look when you get a chance?

services/codecov/codecov.service.js
```diff
@@ -2,37 +2,34 @@
...
-    totals: Joi.object({
-      c: Joi.number().required(),
-    }),
-  }),
-}).required()
+const { BaseSvgScrapingService } = require('..')
 
 const queryParamSchema = Joi.object({
-  token: Joi.string(),
+  token: Joi.string().regex(/^\w{10}$/),
```
calebcartwright, line 8: Is there any documentation that states the possible valid token values?

If so it'd be great to include an inline comment with a link, but if not then I think we're better off leaving the schema as-is since it's just a pass through to the upstream service anyway and I'd prefer to not risk breaking any existing badges

services/codecov/codecov.service.js
```diff
@@ -2,37 +2,34 @@
...
+
 const documentation = `
   <p>
     You may specify a Codecov token to get coverage for a private repository.
   </p>
   <p>
-    See the <a href="https://docs.codecov.io/reference#authorization">Codecov Docs</a>
-    for more information about creating a token.
+    You can find the token for your badge in this url:
+    <code>https://codecov.io/{vcsName}/{user}/{repo}/settings/badge</code>
```
calebcartwright, line 28: Does Codecov use a different token mechanism for their API vs their badges? If so then I'm not sure if we'll be able to bundle the existing, flagless, badges and the flag-based badges into a single service class without breaking existing usage of the badge for private repos

services/codecov/codecov.service.js
```diff
@@ -100,40 +99,33 @@ module.exports = class Codecov extends BaseJsonService {
...
-      }
-    }
-    return this._requestJson({
+  async fetch({ vcsName, user, repo, branch, token, flag }) {
+    const url = `https://codecov.io/${vcsName}/${user}/${repo}${
+      branch ? `/branches/${branch}` : ''
+    }/graph/badge.svg`
+    const queryParams = [flag && `flag=${flag}`, token && `token=${token}`]
+      .filter(Boolean)
+      .join('&')
```
calebcartwright, line 108: Let's use `qs` down below which will handle this (plus any uri encoding) for us
```suggestion

```

services/codecov/codecov.service.js
```diff
@@ -100,40 +99,33 @@ module.exports = class Codecov extends BaseJsonService {
...
+    return this._requestSvg({
       schema,
-      options,
-      url,
-      errorMessages: {
-        401: 'not authorized to access repository',
-        404: 'repository not found',
-      },
+      valueMatcher: svgValueMatcher,
+      url: `${url}?${queryParams}`,
```
calebcartwright, line 112: Inline suggestion disclaimer that prettier/eslint/etc. may want to change parts of this
```suggestion
      url,
      qs: { flag, token }
```

services/codecov/codecov.service.js
```diff
@@ -100,40 +99,33 @@ module.exports = class Codecov extends BaseJsonService {
...
+      .filter(Boolean)
+      .join('&')
+    return this._requestSvg({
       schema,
-      options,
-      url,
-      errorMessages: {
-        401: 'not authorized to access repository',
-        404: 'repository not found',
-      },
```
calebcartwright, line 122: Is this because Codecov will always return a 200 response, but with an `unknown` in the case of an invalid repository? 

If so, it might be better to throw a `NotFound` error in the case of `unknown` for the user experience. A lot of times folks will have a minor typo in one of the route params (i.e. `/badges/sheilds/`) and it's helpful if we can provide a more meaningful badge message in those cases.

services/codecov/badge-100.svg
```diff
@@ -0,0 +1,23 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="122" height="20">
```
calebcartwright, line 1: I was a little confused by these files at first, though now realize they're being used as mock/test data. In similar scenarios elsewhere in the codebase where we need the svg xml for a test, we've done so via an inline string in a js file (for example [here](https://github.com/badges/shields/blob/612a43aaa9cd8435c4165a7d2ebd5bc44838a0da/services/snyk/snyk-test-helpers.js#L3-L6) or [here](https://github.com/badges/shields/blob/612a43aaa9cd8435c4165a7d2ebd5bc44838a0da/lib/svg-helpers.spec.js#L8)

I'd prefer to see something similar done here, or at least move these files under a subdirectory that clearly indicates their purpose (i.e. `mocks/badge-100.svg` ) although I don't feel _too_ strongly about it

services/codecov/codecov.service.js
MunifTanjim: It appears as an error message when you use token of different length: https://codecov.io/gh/codecov/example-node/graph/badge.svg?token=abc

Other than that, there's no documentation link really.

services/codecov/codecov.service.js
MunifTanjim: > Does Codecov use a different token mechanism for their API vs their badges?

Yes. Codecov provides a 10 character token specifically for their badges.

> If so then I'm not sure if we'll be able to bundle the existing, flagless, badges and the flag-based badges into a single service class without breaking existing usage of the badge for private repos

Yes, that's correct. We can handle this by falling back to scrapping svg when the `flag` query param is passed. But I couldn't find a "clean" way to use both `BaseJsonService` and `BaseSvgScrappingService` together for a specific service.

What do you suggest?

services/codecov/codecov.service.js
MunifTanjim: > Is this because Codecov will always return a 200 response, but with an unknown in the case of an invalid repository?

Yes. Their badge returns `unknown` when:
- repository doesn't exist
- coverage is yet not uploaded for a repo
- coverage for a specific flag doesn't exist or hasn't been uploaded yet

services/codecov/codecov.service.js
calebcartwright: In that case we'd have to use two separate classes in that case, one that extends BaseJson for flagless and a new one that extends BaseSvg (this one would need a different route). 

IMO it'd be be ideal if we could have a single implementation like you've done here, but our hand is probably going to be forced to maintain BC for private repo badges.

Curious if any maintainers have any thoughts here?

services/codecov/codecov.service.js
calebcartwright: Ah I see. That's a very minor downside in the svg scraping approach since we no longer have the ability to differentiate. We should probably leave this as you have it here :+1:

services/codecov/codecov.service.js
paulmelnikow: I haven't digested this PR, but here's a quick thought:

If this is a matter of separating out the fetching functionality from the base class, that's something we probably could do. We also use two formats in the F-Droid badge. #2518

If it's trying to do backward compatibility that's a different question (and I'd probably need to study this a bit more).

MunifTanjim: I've re-introduced the previous implementation (with some modifications) and fallback to it with `!flag && token && !badgeTokenPattern.test(token)` condition.

This should be enough to maintain backward-compatibility.

calebcartwright:
I did go ahead and setup a private repo and enabled Codecov to test this out. The good news is that I can confirm this is successfully maintaining backwards compatibility with private projects! 

However, it seems like this may no longer be working with the `flag` param

For example:

We're showing 92% with or without the flag (note that Shields conditionally rounds coverage up in all our coverage badge whereas Codecov always rounds down, hence the 92 vs 91)
https://shields-staging-pr-4968.herokuapp.com/codecov/c/github/ethereumjs/ethereumjs-vm?flag=vm


https://shields-staging-pr-4968.herokuapp.com/codecov/c/github/ethereumjs/ethereumjs-vm


As compared to the native Codecov badges

https://codecov.io/gh/ethereumjs/ethereumjs-vm/branch/master/graph/badge.svg?flag=vm


https://codecov.io/gh/ethereumjs/ethereumjs-vm/branch/master/graph/badge.svg

services/codecov/codecov.service.js
```diff
@@ -20,19 +21,34 @@ const schema = Joi.object({
...
   <p>
     You may specify a Codecov token to get coverage for a private repository.
   </p>
   <p>
-    See the <a href="https://docs.codecov.io/reference#authorization">Codecov Docs</a>
-    for more information about creating a token.
+    You can find the token for your badge in this url: <code>https://codecov.io/{vcsName}/{user}/{repo}/settings/badge</code>.
+  </p>
+  <p>
+    You can also use Codecov's API token (doesn't support <code>flag</code>). See the <a href="https://docs.codecov.io/reference#authorization">docs</a> for guidance to create API token.
```
calebcartwright, line 47: From what I'm seeing, one benefit of the Codecov badge token is that it is strictly read-only, whereas the API token (currently used for private repos) cannot be scoped within codecov; authenticating with the API token provides authorization to everything the token-owner has access to.

As I mentioned above, we do prefer API over SVG scraping, but in this instance badge scraping approach would be preferable in every case (except any existing private, flagless badges) because it allows users to use a badge-read-only token.


As such, let's not even mention the API auth token in the docs. We'll maintain support to avoid breaking existing badges, but lets steer folks to use the badge tokens.
```suggestion
    You can find the token under the badge section of your project settings page, in this url: <code>https://codecov.io/{vcsName}/{user}/{repo}/settings/badge</code>.
```

services/codecov/codecov.service.js
```diff
@@ -100,40 +118,77 @@ module.exports = class Codecov extends BaseJsonService {
...
+      branch ? `/branches/${branch}` : ''
+    }/graph/badge.svg`
+    return this._requestSvg({
+      schema,
+      valueMatcher: svgValueMatcher,
+      url,
+      options: {
+        qs: { token, flag },
+      },
+      errorMessages: token ? { 400: 'invalid token pattern' } : {},
```
calebcartwright, line 171: The 400 will handle the scenario where someone tries to use an API token (such as an existing private badge) with a flag right?

MunifTanjim:
> We're showing 92% with or without the flag (note that Shields conditionally rounds coverage up in all our coverage badge whereas Codecov always rounds down, hence the 92 vs 91):

This is strange 🤔 This is the result when I'm running it locally on my pc:

image

services/codecov/codecov.service.js
MunifTanjim: Yes, that's right.

calebcartwright:
Does indeed seem to be working now!

calebcartwright:
@paulmelnikow - I've approved but will wait another day or two to merge in case you want to review given the discussion in https://github.com/badges/shields/pull/4968#discussion_r418593467

Basically, this PR does two cool things for our Codecov badges: (1) enables badges for coverage with flags/monorepos and (2) allows us to switch the way support codecov badges for private repos to a read-only, project specific token.

Both of those require SVG scraping Codecov's native badges, whereas we were previously using Codecov's API. So while we've switched to the SVG scraping implementation, we have to keep support for Codecov's API which is required to maintain BC for existing usage of our codecov badges for private repos

paulmelnikow:
What you're saying makes sense to me! 👍

PR #4976, opened 2020-04-29
Upgrade to Mocha 7; test on [npmdownloads]

### LINKED ISSUE #4842: Service tests don't run with mocha 7
:beetle: **Description**

In #4841, @Sytm noticed that the Spiget service tests were not being run. I've checked this with several other services, and actually none of them are running anymore, which is confirmed by the latest daily test run on CircleCI. Tests are loaded, IcedFrisby's `skipWhen` checks for credentials are run and then the process terminates without any error message.

After a bit of digging, I managed to narrow down the culprit to #4822, bump of Mocha's version from 6.2.2 to 7.1.1. I checked things locally with older 7.0.0, it's broken there as well, so something changed in that release. The [changelog](https://github.com/mochajs/mocha/releases/tag/v7.0.0) lists a number of breaking changes, however, nothing that would impact us in this way as far as I can tell.

I've been digging around and trying various things for the past hour or so, but probably due to my lack of familiarity with the Mocha library, I've not made much progress. Any clues?

paulmelnikow:
For some reason the `--delay` version is not working in Mocha 7. I'm not sure why that is. Although, invoking Mocha the normal synchronous way seems to be working fine, so maybe `--delay` just isn't necessary anymore for what we're doing in Shields.

I don't see anything in the changelog about this :man_shrugging:

Closes #4842


paulmelnikow:
Did they run? I think so. It says 5 tests, although I can't see the output.

PR #4978, opened 2020-04-29
Lengthen timeout on a sometimes slow test

paulmelnikow:


services/wordpress/wordpress-version-color.spec.js
```diff
@@ -18,6 +18,8 @@ describe('toSemver() function', function() {
 
 describe('versionColorForWordpressVersion()', function() {
   it('generates correct colours for given versions', async function() {
+    this.timeout(5e3)
```
calebcartwright, line 21: had not seen this timeout arg type before

paulmelnikow: It’s a JavaScript thing! You can use it anywhere. It’d nice for numbers. I kind of forgot though, I think Mocha supports strings like ‘10s’ too.

Issue #4950, opened 2020-04-25
Depend on badge-maker's public interface instead of relative imports


chris48s:
:clipboard: **Description**

There are 3 features we need to expose in the public interface of badge-maker before we can remove our relative imports and depend on the badge-maker package.

- https://github.com/badges/shields/issues/4947 - logos
- https://github.com/badges/shields/issues/4948 - move JSON rendering from package to server
- https://github.com/badges/shields/issues/4949 - links

This also should help with https://github.com/badges/shields/issues/3370

Issue #4949, opened 2020-04-25
Expose links in badge-maker


chris48s:
:clipboard: **Description**

Fundamentally we need to add the ability to pass a couple of strings into `makeBadge()`, but we need to decide on the interface for that. One of the useful things we've done in v3 is bringing the terminology used in shields.io URL params, the endpoint badge and the NPM package into line. This is **a good thing** but the current interface for links in URL params is a bit weird. Now is probably a good time to re-think it. There is some discussion of this in:

https://github.com/badges/shields/issues/4849#issuecomment-609108141

Of the v3.1 issues, this one is currently the least clearly-defined.

chris48s:
This got done in https://github.com/badges/shields/pull/10283 by just exposing the current interface on to it with no changes. Improving it is still valid though.

Issue #4947, opened 2020-04-25
Expose the concept of logos, add SimpleIcons as a peer dependency


chris48s:
:clipboard: **Description**

Expose logo and logo-specific params (e.g: `logoColor`, `logoWidth`). There are 3 cases for logos:

- Named SimpleIcons logo
- Custom base64-encoded logo
- Named shields custom logo https://github.com/badges/shields/tree/master/logo

At a library level it probably makes sense to reduce that to two cases - SimpleIcons and base64. We can keep our handful of custom logos outside of badge-maker and just treat them as something we handle in server somewhere and pass to the renderer as custom base64.

This will allow us to remove final relative import to the package requiring anything other than `makeBadge()` https://github.com/badges/shields/blob/8621fe42d76b809e9ba4ae844920f8ed0373817a/lib/logos.js#L4

#4524 never happened, but there are some notes on testing code with peer dependencies in https://github.com/badges/shields/pull/4524#discussion_r365603567

This work should also remove all/most of the remaining code from https://github.com/badges/shields/tree/master/lib (and if there is anything left, we should work out where to move it to)

paulmelnikow:
> At a library level it probably makes sense to reduce that to two cases - SimpleIcons and base64. We can keep our handful of custom logos outside of badge-maker and just treat them as something we handle in server somewhere and pass to the renderer as custom base64.

That's an interesting thought! Though it would have a downside of the logos not being available to package users (i.e. the package and the service supporting different sets of logos).

paulmelnikow:
Though I'd pick this up and see if we can keep closing the gap toward dogfooding our own package.

I think there's one question we should resolve first, which is whether to handle the same three cases (simpleicons, custom base64, and shields logos) in the package, or handle simpleicons and base64 in the package and keep the responsibility for the shields logos with the server.

I'd lean toward supporting all three in the package, so that the service and the package support the same set of logos. If users are accustomed to using certain logos on the server, having the package support a smaller set seems like it would be confusing.

chris48s:
Some semi-random thoughts on this topic that don't necessarily all agree with each other..

* If you think of the package as being something that helps you build your own shields.io then yes its odd if the package and service behave differently, but I tend to assume that library users want to do something outside the scope of what using shields.io directly or self-hosting an instance can do for them rather than exactly replicate it.
[...]

PR #4980, opened 2020-04-30
Move JSON rendering from badge-maker to server [*****]

### LINKED ISSUE #4948: Move JSON output out of badge-maker and into server
:clipboard: **Description**

We need JSON output for testing reasons, but it is not a supported feature of the NPM package. Move this feature into the badge server.

paulmelnikow:
1. Move normalization from "internal" `makeBadge()` into a `_clean()` function which is privately exported by badge-maker. It's necessary that the server can access this functionality to do the JSON rendering. (Alternatively we could remove the underscore to make this part of the public API. However this may not be necessary, as I think it will only be needed for Shields.)
2. Some changes to help with #4950:
    1. Bring the "internal" badge-maker schema more in line with the badge-maker public schema.
    2. Bring the output of `coalesceBadge()` more in line with the badge-maker public schema.
    3. Since the badge-maker public interface requires strings, the package internals should assume a string. Accordingly, move numeric coercion from the the package internals to the server.

Closes #4948

### DIFF
--- a/badge-maker/lib/color.spec.js
+++ b/badge-maker/lib/color.spec.js
@@ -18,15 +18,50 @@ test(isHexColor, () => {
 })
 
 test(normalizeColor, () => {
+  // Shields named color.
   given('red').expect('red')
+  given('green').expect('green')
   given('blue').expect('blue')
-  given('4c1').expect('#4c1')
-  given('f00f00').expect('#f00f00')
-  given('ABC123').expect('#abc123')
-  given('#ABC123').expect('#abc123')
+  given('yellow').expect('yellow')
+
+  // valid hex
+  forCases([given('#4c1'), given('#4C1'), given('4C1'), given('4c1')]).expect(
+    '#4c1'
+  )
+  forCases([
+    given('#abc123'),
+    given('#ABC123'),
+    given('abc123'),
+    given('ABC123'),
+  ]).expect('#abc123')
+
+  // valid rgb(a)
+  given('rgb(0,128,255)').expect('rgb(0,128,255)')
+  given('rgba(0,128,255,0)').expect('rgba(0,128,255,0)')
+  // valid hsl(a)
+  given('hsl(100, 56%, 10%)').expect('hsl(100, 56%, 10%)')
+  given('hsla(25,20%,0%,0.1)').expect('hsla(25,20%,0%,0.1)')
+
+  // CSS named color.
   given('papayawhip').expect('papayawhip')
   given('purple').expect('purple')
+
   forCases([
+    // invalid hex
+    given('#123red'), // contains letter above F
+    given('#red'), // contains letter above F
+    // invalid rgb(a)
+    given('rgb(220,128,255,0.5)'), // has alpha
+    given('rgba(0,0,255)'), // no alpha
+    // invalid hsl(a)
+    given('hsl(360,50%,50%,0.5)'), // has alpha
+    given('hsla(0,50%,101%)'), // no alpha
+    // neither a css named color nor colorscheme
+    given('notacolor'),
+    given('bluish'),
+    given('almostred'),
+    given('brightmaroon'),
+    given('cactus'),
     given(''),
     given('not-a-color'),
     given(undefined),
@@ -36,6 +71,9 @@ test(normalizeColor, () => {
     given({}),
     given(() => {}),
   ]).expect(undefined)
+
+  // Semantic color alias
+  given('success').expect('brightgreen')
   given('lightgray').expect('lightgrey')
   given('informational').expect('blue')
 })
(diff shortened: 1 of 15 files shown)


paulmelnikow:
Failures seem unrelated, except that one triggered an error in an untested code path in the request handler, which I fixed in 0a68459098aa2a7c25f1798257f1e5791cdbc03c.

badge-maker/lib/color.spec.js
```diff
@@ -18,15 +18,50 @@ test(isHexColor, () => {
 })
 
 test(normalizeColor, () => {
+  // Shields named color.
   given('red').expect('red')
+  given('green').expect('green')
   given('blue').expect('blue')
-  given('4c1').expect('#4c1')
-  given('f00f00').expect('#f00f00')
```
calebcartwright, line 24: Was this removed because it was duplicative of the `ABC123` one?

paulmelnikow: Yes, exactly.

chris48s:
I'll have a look at this one over the weekend

badge-maker/lib/index.js
```diff
@@ -42,23 +52,26 @@ function _clean(format) {
...
-  // convert "public" format to "internal" format
-  cleaned.text = [cleaned.label || '', cleaned.message]
-  delete cleaned.label
-  delete cleaned.message
-  if ('style' in cleaned) {
-    cleaned.template = cleaned.style
-    delete cleaned.style
-  }
+  // Whitespace removal.
+  cleaned.label = `${cleaned.label}`.trim()
```
chris48s, line 67: For a message only badge, running
```
cleaned.label = `${cleaned.label}`.trim()
```

would result in the label property being the string literal `"undefined"`. This is not what we want.

paulmelnikow:
> Having seen the implementation: If you consider that the near-future goal is that we generate our own (SVG) badges via `index.makeBadge()` (and hence everything will go through `_validate()` and `_clean()` implicitly), do you reckon exporting an undocumented function is worse than exposing an undocumented `format="json"` param or depending on a relative import for the purpose of json output?
> 
> They all represent a sort of "backdoor", or a way in which we're _not_ going to eat our own dogfood, so if we're going to do one of those things we might as well pick the one that requires us to do the _least_ acrobatics elsewhere in the codebase achieve it, if you see what I mean.

The server needs to access the normalization capability, though not the validation code, of the badge maker, separately from invoking `makeBadge()`, so it can emit normalized values in the JSON response. 

I suspect we should export a public `normalize()` function for this. It's unlikely anyone other than Shields will use it, but it's still a reaonsably logical part of the API. It feels cleaner than having a `{ format: json }` parameter, particularly since we make these weird legacy modifications to the schema in the server anyway (and I definitely don't think we should have those in the package).

Maybe this function should assume the data is valid, or maybe it should validate it, not sure. I haven't thought carefully about what is validation vs. what is normalization. We could tackle that now, or defer it to a follow-on. Deferring might be the thing to do since we want to release 3.0.0… if it's private and undocumented, then we can make it public in 3.1.0.

Thoughts?

badge-maker/lib/index.js
paulmelnikow: Indeed. I'll fix this up tomorrow.

chris48s:
> The server needs to access the normalization capability, though not the validation code, of the badge maker, separately from invoking `makeBadge()`, so it can emit normalized values in the JSON response.

If the objective is to move `JSON.stringify()` from one place to another, yeah. Just to take a step back though..

Once we do #4947 #4949 and #4950 the code in `server.js` / `legacy-request-handler.js` becomes

```js
const { makeBadge } = require('badge-maker')
//...
makeBadge(badgeData)
```

..at which point everything for the SVG request is getting processed by `_clean()` and `_validate()` anyway. We shouldn't be passing anything to `makeBadge()` that doesn't pass `_validate()` - if we are, that's something we need to fix.

In principle, I agree that not exposing `format='json'` as an undocumented option on `makeBadge()` is conceptually nicer if it gives us a completely clean interface boundary, but if we trade off an undocumented "secret" param for an undocumented "secret" function, maybe the lesser evil is just to leave the `JSON.stringify()` in the package and go with and undocumented param if that means the server code is like:

```js
makeBadge({...badgeData, format='json'})
```

instead of doing the whole `makeBadgeOrJson()` dance?

paulmelnikow:
Having thought about this a while, I feel like it's pretty fine for the package to have non-secret `validate()` and `normalize()` functions.

paulmelnikow:
I've picked a couple bits of this out. I'm thinking I'll close it for now, and wait to address the rest when we're ready to dogfood our own package.

PR #4983, opened 2020-04-30
Fix production crash in [GithubStars] with trailing spaces

### LINKED ISSUE #4982: ValidationError in GitHub Stars
https://sentry.io/organizations/shields/issues/1322996930/?referrer=github_plugin

```
ValidationError: {
  "message": "redacted",
  "color": "blue",
  "link": [
    "https://github.com/redacted/redacted            ", [31m[1][0m
    "https://github.com/redacted/redacted            /stargazers"
  ]
}
[31m
[1] "link[0]" must be a valid uri[0m
  File "/app/core/base-service/base.js", line 296, in GithubStars._validateServiceData
    Joi.assert(serviceData, serviceDataSchema)
  File "/app/core/base-service/base.js", line 409, in Function.invoke
    serviceInstance._validateServiceData(serviceData)
...
(4 additional frame(s) were not displayed)
```

paulmelnikow:
Closes #4982


services/github/github-stars.tester.js
```diff
@@ -20,3 +20,17 @@ t.create('Stars (repo not found)')
     label: 'stars',
     message: 'repo not found',
   })
+
+// https://github.com/badges/shields/issues/4982
+// It doesn't seem important that this work with trailing spaces, though if it
+// did we'd probably want a more global fix.
+t.create('Stars (repo name with trailing spaces)')
+  .get('/badges/shields     .json')
```
calebcartwright, line 28: > It doesn't seem important that this work with trailing spaces, though if it did we'd probably want a more global fix.

Yeah I can't imagine GH allows trailing spaces in the repo name

paulmelnikow: Weirdly the API accepts them! And then we crash in rendering.

services/github/github-stars.tester.js
```diff
@@ -20,3 +20,17 @@ t.create('Stars (repo not found)')
...
+  .expectBadge({
+    label: 'stars',
+    message: isMetric,
+    link: [
+      'https://github.com/badges/shields%20%20%20%20%20',
```
chris48s, line 33: I don't understand this. https://github.com/badges/shields%20%20%20%20%20 is a 404. I get that it makes the validation pass, but why do we want to render an invalid link?

paulmelnikow: My goal here is to avoid this runtime error. The ideal error would be something like "repo param is invalid" or even 404 for not matching any of the routes. But I'd rather return a badge with a nonsense link than the internal error.

I'm sort of surprised that our path parameters allow trailing spaces to be sucked into the named params. I'm not sure why that is, how what's involved in fixing it. I'd hate to have to update all of our patterns to disallow spaces. Maybe there is a flag for path-to-regex that can do this globally?

I'd suggest we settle on a compromise for the moment, add a comment explaining that we know this is garbage, and open a new issue to investigate further and figure out what the perfect thing is.

chris48s: Hmm. I suppose it does mean

https://img.shields.io/github/stars/badges/shields%20?style=social

works like

https://img.shields.io/twitter/follow/espadrine%20?style=social

I'm not sure its _intrinsically wrong_ for a route param to have a trailing space. If something like `/user/repo/labels/good%20first%20issue` could be a valid route in your application, its probably not the router's job to say `/user/repo/labels/good%20first%20issue%20` can't be.

> I'd suggest we settle on a compromise for the moment, add a comment explaining that we know this is garbage, and open a new issue to investigate further and figure out what the perfect thing is.

I'm sure us-from-the future will know what to do with this. We'll be super smart by then :smile:

paulmelnikow: > I'm not sure its _intrinsically wrong_ for a route param to have a trailing space. If something like `/user/repo/labels/good%20first%20issue` could be a valid route in your application, its probably not the router's job to say `/user/repo/labels/good%20first%20issue%20` can't be.

Right, that's true.

I guess I have figured most of our query parameters are `a-zA-Z0-9^/` sort of things, basically domain-name characters.

You can't have spaces at the end of a GitHub username or org, for example. I think most of our route params (though not query params) should probably not have whitespace in them.

chris48s: I'll stick this on to merge - are you happy to write up an issue for thinking about this at a slightly higher level

paulmelnikow: Done: #5012

PR #4991, opened 2020-04-30
fix link width params

### LINKED ISSUE #4989: 'NaN' in 'width' attribute inside rect element of anchor in generated svg

- [ ] My own instance

:beetle: **Description**

Static Endpoint Combined with `<object>` tag fails : `NaN` in width element

:link: **Link to the badge**

```
<object data="https://img.shields.io/static/v1?labelColor=green&label=LABEL&message=MY%20MESSAGE&color=orange&style=flat-square&link=https%3a%2f%2fmy.domain.com"></object>
```

This code doesn't work but used to work.
It returns an invalid NaN in the `width` field of the rect element (child of anchor) of the generated svg

```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="130" height="20">

<rect width="45" height="20" fill="#97ca00"/>
<rect x="45" width="85" height="20" fill="#fe7d37"/>
</g>

<g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="110">
<text x="235" y="140" transform="scale(.1)" textLength="350">LABEL</text>
<text x="865" y="140" transform="scale(.1)" textLength="750">MY MESSAGE</text>
</g>

<a target="_blank" xlink:href="https://my.domain.com">
<rect width="NaN" height="20" fill="rgba(0,0,0,0)"/>
</a>

</svg>
```

:bulb: **Possible Solution**

chris48s:
closes #4989


badge-maker/lib/badge-renderers.js
```diff
@@ -99,23 +99,23 @@ function renderText({
 
 function renderLinks({
   links: [leftLink, rightLink] = [],
-  labelWidth,
-  messageWidth,
+  leftWidth,
+  rightWidth,
```
calebcartwright, line 103: I'm not entirely clear how to test this, though it does seem to fix an issue. 

I didn't go through this file in detail but did see various usage of left/right vs. label/message in different places, and I was wondering if it might be easier to try use one or the other more consistently? (if so, those changes could certainly be done in a followup vs tacking onto this)

paulmelnikow: Could we add an example for this case to the test page?

chris48s: > but did see various usage of left/right vs. label/message in different places

The convention I tried to follow in this is:

`labelWdith` is be the size in pixels of the text "build", whereas `leftWidth` is the size of the whole left side of the badge. `messageWidth` is the size of the text "passing", whereas `rightWidth` is the size of the entire right side of the badge.

The link is the whole left/right side of the badge. That's why the fix here is to change the params of `renderLinks()` to `leftWidth`, `rightWidth` rather than to do this:

```diff
diff --git a/badge-maker/lib/badge-renderers.js b/badge-maker/lib/badge-renderers.js
index 22adca8c..e64efd48 100644
--- a/badge-maker/lib/badge-renderers.js
+++ b/badge-maker/lib/badge-renderers.js
@@ -126,7 +126,7 @@ function renderBadge({ links, leftWidth, rightWidth, height }, main) {
   return `
     <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="${width}" height="${height}">
     ${main}
-    ${renderLinks({ links, leftWidth, rightWidth, height })}
+    ${renderLinks({ links, labelWidth: leftWidth, messageWidth: rightWidth, height })}
     </svg>`
 }
```

I'm not promising that if you trawl the entire of `badge-renderers.js` there are no exceptions, and there's probably some grey areas of "the size of this bit of text plus some padding but not the entire left/right side of the badge" but that's the thought behind it..

chris48s: > Could we add an example for this case to the test page?

I had a quick look at this. In order to render the badge with links in a useful way, we have to embed with `<object>` rather than ``. That's do-able, but at the moment the `<Badge>` component is hard-coded to use ``. (We don't actually render the social badges previews using `<object>`on shields.io - maybe this is one of the reasons we didn't spot this edge case).

I'm not saying its a bad idea to add this, but its also not trivial in the same way the last "can we add this to the examples page" was. I'm not 100% sure if its a massive job (maybe its simpler than I think), but my inclination is to kick `<object>` embeds in the front-end to an issue.

calebcartwright: > labelWdith is the size in pixels of the text "build", whereas leftWidth is the size of the whole left side of the badge. messageWidth is the size of the text "passing", whereas rightWidth is the size of the entire right side of the badge.

D'oh! That makes sense, sorry for the noise

paulmelnikow: > I'm not saying its a bad idea to add this, but its also not trivial in the same way the last "can we add this to the examples page" was. I'm not 100% sure if its a massive job (maybe its simpler than I think), but my inclination is to kick `<object>` embeds in the front-end to an issue.

It should be quick, but I'm happy to tackle it in a follow-on. (Probably adding a boolean flag to `<Badge>` would do the trick.)

Issue #4977, opened 2020-04-29
drop node < 10 testing/support


chris48s:
:clipboard: **Description**

We're now deployed on node 10 in production now. Some follow-up tasks:

:wrench:  Modify CI build to drop testing on node < 10 and test on sensible versions
:books:  Update docs for self-hosting users that node 10+ is required/supported
:books:  Update docs for package users that node 10+ is required/supported (get this done before we release stable v3 package)
:ghost:  Resurrect https://github.com/badges/shields/pull/4405 (was there anything else we started and had to abandon?)
:robot:  Update dependabot config to stop ignoring all the node 10 only dependencies
:package:  Upgrade ALL TEH PACKAGES (prioritising anything with security fixes first)

Refs https://github.com/badges/shields/issues/3404
Refs https://github.com/badges/shields/issues/4973

Anything I've forgotten there?

chris48s:
this is now done, except for the dependency updates which are in progress

Issue #4973, opened 2020-04-29
upgrade to node 12 in production


chris48s:
:clipboard: **Description**

We're currently on 10, but should now be no reason we can't adopt 12 soon.

paulmelnikow:
12 makes sense to me.

Looks like we're testing on Node 8 and node@latest currently, so probably we should update that to @12 and @latest.

calebcartwright:
Should we consider running on all LTS versions to cover the self-hosting cases, or should we just update the docs accordingly?

chris48s:
We probably _should_, but there's also a tradeoff of build minutes/number of simultaneous builds, which is why I went with 10 and 12 in https://github.com/badges/shields/pull/4992 .
We could do [the nvm dance](https://github.com/badges/shields/blob/7385739a43abea17f304e05e2f568add9180c22a/.circleci/config.yml#L89-L99) in more places..

paulmelnikow:
I think our call in #3404 for self-hosting is that we'll support current plus possibly whatever has been current in the past three months. (Right now I think that would be 12 + 14.)

> For **self-hosting** it would be great if we could support only the ["current" version](https://github.com/nodejs/Release). Or maybe – to give self-hosting users time to upgrade – any version that has been "current" within the last three months. Or maybe – to reduce the frequency of those upgrades – the latest even-numbered version.

chris48s:
Lets get our own house in shape first then track that schedule moving forwards :)

paulmelnikow:
Does this mean we can use static fields in the service classes?! 🤯

chris48s:
oh yeah - there are probably some things like static/private class properties that are now unlocked

PR #4992, opened 2020-04-30
Update test matrix and docs

chris48s:
Refs #4977
Refs #3404
Refs #4973

I haven't bothered to update the reference in https://github.com/badges/shields/blob/master/doc/production-hosting.md as that whole doc needs revamping. I'll deal with this as another PR

My thinking here is while we are deployed on node 10 and we want to upgrade to 12 soon, those are the two most important runtimes and that's what we should run our tests on. Once we're on to 12 in production, we can switch to tracking latest again. I think running the core/service tests against 3 runtimes bloats the matrix too much.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,6 +109,7 @@ package_steps: &package_steps
           export NVM_DIR="/opt/circleci/.nvm"
           [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
           nvm install v12
+          nvm install v14
           nvm use v12
           npm install -g npm
           npm ci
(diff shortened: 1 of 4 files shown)


chris48s:
We will also need to update the branch protection rules to account for the changes to test names, but I'll wait for feedback first..

chris48s:
Right, I've updated the branch protection rules to change some `-latest`s to `-12`s which makes this merge-able. Having looked at it, we should maybe review what checks are/aren't required. I'm not sure if we've just added more builds to circle over time and forgotten to add them as required in the branch setup or if there are some we intentionally don't require a pass on?

paulmelnikow:
> Having looked at it, we should maybe review what checks are/aren't required. I'm not sure if we've just added more builds to circle over time and forgotten to add them as required in the branch setup or if there are some we intentionally don't require a pass on?

I think we've chosen to make service and integration tests optional, along with LGTM and Dockercloud.

PR #5013, opened 2020-05-02
update production hosting/deploy docs

### LINKED ISSUE #4974: update documentation to reflect new production environment
:clipboard: **Description**

Review https://github.com/badges/shields/tree/master/doc and update

chris48s:
I attempted to update the table of who has access to what, but as I was doing it I realised the table we made in the document last week doesn't match up with the selection of invites I received and I don't know who else received what so I've ignored that section for now.

@paulmelnikow - do you feel like having a go at that table, or we can review it in the next call if you want..

closes #4974


doc/production-hosting.md
```diff
@@ -36,38 +36,7 @@
...
 
-## Badge servers
-
-There are three public badge servers on OVH VPS’s.
-
-| Cname                       | Hostname             | Type | IP             | Location           |
-| --------------------------- | -------------------- | ---- | -------------- | ------------------ |
-| [s0.servers.shields.io][s0] | vps71670.vps.ovh.ca  | VPS  | 192.99.59.72   | Quebec, Canada     |
-| [s1.servers.shields.io][s1] | vps244529.ovh.net    | VPS  | 51.254.114.150 | Gravelines, France |
-| [s2.servers.shields.io][s2] | vps117870.vps.ovh.ca | VPS  | 149.56.96.133  | Quebec, Canada     |
```
paulmelnikow, line 47: We're still using these as proxying servers so we probably should keep them in the docs, somewhere, until they are no longer part of the production setup.

doc/production-hosting.md
```diff
@@ -247,12 +177,3 @@ Request performance is monitored in two places:
 [monitor]: https://shields.redsparr0w.com/1568/
 [notifications]: http://shields.redsparr0w.com/discord_notification
 [monitor discord]: https://discordapp.com/channels/308323056592486420/470700909182320646
-
-## Known limitations
-
-1.  The only way to inspect the commit on the server is with `git ls-remote`.
```
paulmelnikow, line 253: This list item, about `git ls-remote` probably should be kept with the legacy servers until they are no longer part of production.

doc/production-hosting.md
```diff
@@ -147,64 +116,25 @@ hosted on [Zeit Now][]. It's managed in the
...
+$ npm install  # install dependencies (devDependencies are needed to build the frontend)
+$ make deploy-gh-pages  # build the frontend and push it to the gh-pages branch
+```
 
-1.  Pushes the same working tree to s1 and s2.
-2.  Creates a new working tree for the frontend.
-3.  Adds a commit cleaning out the index.
-4.  Adds another commit with the build frontend.
-5.  Pushes to `gh-pages`.
+No secrets are required to build or deploy the frontend.
```
paulmelnikow, line 137: 👍

doc/production-hosting.md
```diff
@@ -147,64 +116,25 @@ hosted on [Zeit Now][]. It's managed in the
...
-2.  In that tree, runs `features` and `examples` to generate data files
-    needed for the frontend.
-3.  Builds and checks in the built frontend.
-4.  Checks in `local-shields-io-production.yml`.
-5.  Pushes to s0, which updates dependencies and then restarts itself.
-
-`make push-s1 push-s2 deploy-gh-pages` does the following:
+```sh
+$ git pull  # update the working copy
+$ npm install  # install dependencies (devDependencies are needed to build the frontend)
```
paulmelnikow, line 133: To respect the lockfile, this should be `npm ci`.

paulmelnikow:
Thanks for taking this on! I can have a go at the access table in a follow-on.

doc/production-hosting.md
calebcartwright: Agreed, especially since this is apparently a point of reference for some vendors currently whitelisting our traffic

doc/production-hosting.md
```diff
@@ -36,38 +36,7 @@
...
-
-There are three public badge servers on OVH VPS’s.
-
-| Cname                       | Hostname             | Type | IP             | Location           |
-| --------------------------- | -------------------- | ---- | -------------- | ------------------ |
-| [s0.servers.shields.io][s0] | vps71670.vps.ovh.ca  | VPS  | 192.99.59.72   | Quebec, Canada     |
-| [s1.servers.shields.io][s1] | vps244529.ovh.net    | VPS  | 51.254.114.150 | Gravelines, France |
-| [s2.servers.shields.io][s2] | vps117870.vps.ovh.ca | VPS  | 149.56.96.133  | Quebec, Canada     |
-
-- These are single-core virtual hosts with 2 GB RAM [VPS SSD 1][].
```
calebcartwright, line 49: This always impresses me when I read it. Those 3 little boxes did a _lot_ for a long time

doc/production-hosting.md
chris48s: `npm install` doesn't ignore the lockfile, but it will update the lockfile if `package.json` doesn't match it. While `npm ci` would exit with a non-zero status code in that case, if you've made uncommited changes to `package.json`, `git pull` would have failed in the previous step, so.. meh :shrug: . I've changed it anyway.

paulmelnikow: If `package.json` specifies e.g. `^1.0.2` and `1.1.0` is available but we haven't upgraded, it will upgrade it and update the lockfile. We want that to happen when we're bumping the dependencies in a PR, not when we're deploying.

chris48s: Its not deeply important for merging this PR, but I'm just trying to understand what you're saying here..

If I create the following `package.json`:

```json
{
  "dependencies": {
    "is-png": "^1.0.0"
  }
}
```

and corresponding `package-lock.json`:

```json
{
  "requires": true,
  "lockfileVersion": 1,
  "dependencies": {
    "is-png": {
      "version": "1.0.0",
      "resolved": "https://registry.npmjs.org/is-png/-/is-png-1.0.0.tgz",
      "integrity": "sha1-PYA3P+m4nWX9NB9lnT/AoRNecYo="
    }
  }
}
```

even though [is-png 1.1.0](https://www.npmjs.com/package/is-png/v/1.1.0) exists, I can `npm install` in that project dir as many times as you like. `package-lock.json` is always going to have `1.0.0` in it, and so is `node_modules`.

Are you saying you think that just doing an `npm install` transparently bumps that to `1.1.0`, or am I misunderstanding a more subtle point here?

PR #4946, opened 2020-04-24
fix logo placement with custom logoWidth value

chris48s:
closes #4933

PR #5037, opened 2020-05-02
v3 stable

chris48s:
We've been running the v3 package code in production for a while and we've had a release candidate out for over a week. Some non-zero number of people have downloaded it and during that time we've fixed a couple of bugs that slipped the net in review thanks to community feedback:

* https://github.com/badges/shields/pull/4946
* https://github.com/badges/shields/pull/4991

We've also dropped testing on node < 10 and updated the docs for that and made a start on work slated for 3.1.

I'm going to suggest we tag a stable 3.0.0 release at this point, get it on NPM and look at deprecating the old package, unless anyone objects..


chris48s:
Thanks :+1:  I've pushed v3 to NPM. I'll have a look at how to pack a final release of the old gh-badges package documenting the deprecation in the next couple of days

paulmelnikow:
`npm deprecate` is a pretty decent option. You write a string of text that shows up at the top of the page. I think it also shows up in npm install logs.

chris48s:
Yes. I do plan to do that, but I want to do 1 more release with no new code, but documenting it in the chagelog. The reason for this is if you use dependabot, it will submit a PR upgrading you to a version with no changes but with a note in the changelog saying "go download this new package", rather than just never offering you an update ever again, which is what will happen if we just deprecate it.

PR #5043, opened 2020-05-03
Fix a label in the bug issue template

webknjaz:


webknjaz:
Looks like it was accidentally copied from the question template

calebcartwright:
Thank you for the PR! However, the usage of the `question` label on the bug template is deliberate. 

We consider issues created with the bug template to be unconfirmed bugs until a maintainer has been able to confirm, at which point we swap the labels. 

For reference the description for the `question` label is:
>  Support questions, usage questions, unconfirmed bugs, discussions, ideas

paulmelnikow:
If you'd like @webknjaz you could add a comment in the markdown explaining this.

webknjaz:
Ah, got it. Usually projects have some explicit triage label for such purpose. You can even use a bot like https://github.com/apps/triage-new-issues

calebcartwright:
Yup and we considered doing so, but decided against it as part of minimizing the quantity of labels in the repo.

Issue #3663, opened 2019-07-07
Announcing the raster server


paulmelnikow:
Shields has changed the way it presents **raster badges** – i.e. badges in PNG, GIF, and JPG format.

> `https://img.shields.io/v/npm/express.png`
> ↪️`https://raster.shields.io/v/npm/express.png`

**What has changed?**

- `img.shields.io` renders only SVG and JSON badges.
- Previously it also supported PNG, GIF, and JPG.
- PNG badges are rasterized downstream at `raster.shields.io`
    - For backward compatibility, the PNG URLs will 301 redirect to `raster.shields.io`
- Support for GIF and JPG has been removed.

**Note:** The legacy raster static badge route, https://img.shields.io/label/message.png has been discontinued. (#3659) URLs of that form should be updated to use one of the modern routes:

- SVG: https://img.shields.io/badge/label-message-color
- PNG: https://raster.shields.io/badge/label-message-color

**Why was this changed?**

[Shields has always been about SVG badges first](https://olivierlacan.com/posts/an-open-source-rage-diamond/), because they look better and perform better.

PNG badges are a fringe but meaningful use case for badges in emails, and services like Slack which don’t support SVG.

Moving rasterization off-board simplifies the responsibilities of the badge server and hardens the server, as it prevents someone from overloading the server with raster requests. (Each raster request used a subprocess, making them much more expensive than SVG badges.)

GIF and JPG were removed because, on a monthly basis, these were barely touched.

Read more: #3112

**How does it work?**

- The new service is implemented as a web function using the [micro](https://github.com/zeit/micro) framework.
- The server code is here: [rasterize.js](https://github.com/badges/svg-to-image-proxy/blob/master/rasterize.js)
- Rasterization is handled by [convert-svg-to-png](https://www.npmjs.com/package/convert-svg-to-png) which uses headless Chromium via [Puppeteer](https://pptr.dev/).
- The service is hosted on [Zeit Now 2.0](https://zeit.co/now).

**Notes**

- If you self-host a copy of Shields and need to support raster badges, you will need to host your own copy of the raster server. You can do that inexpensively on [Zeit Now or on your own hardware](https://github.com/badges/shields/blob/master/doc/self-hosting.md#raster-server).
    - If you've gotten set this up, we'd welcome better documentation! (#3648)
- The raster server will always return PNG, and will work even if you don't replace `.svg` with `.png` in the URL (e.g. `https://raster.shields.io/v/npm/express.svg`)

**Acknowledgements**

[...]

PR #5046, opened 2020-05-03
send s-maxage and public cache headers

chris48s:
Refs #3663

Reading over https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control
I don't think `public` should be a problem for self-hosted users. If your traffic is going through a shared or public cache in the first place, I don't think whether or not that cache can store the response or not is the main issue.

If you're worried about it, we could add an env var for `ALLOW_PUBLIC_CACHE_HEADER` or something so its configurable.


paulmelnikow:
> I don't think `public` should be a problem for self-hosted users. If your traffic is going through a shared or public cache in the first place, I don't think whether or not that cache can store the response or not is the main issue.

Is that still true on a (private) corporate proxy server?

chris48s:
Can you give an example of what you're getting at here?

paulmelnikow:
Yea. I'm thinking about on-server auth and auth within the URLs. Let me read a bit more about `public` and respond later today.

chris48s:
Right yeah I think I see where you're going with this..

If a self-hosted shields instance has been put behind some kind of auth (so essentially you have to log in to see an image), allowing the images to be cached could potentially leak images that are only available to an authenticated user to unauthenticated users via the cache, right?

Given that, what's your preference for how we make it configurable?

paulmelnikow:
I'm happy to merge the `s-maxage` bit.

I've been reading up more about `public`. There are three relevant sections in the HTTP spec:

- [13.4 Response Cacheability](https://tools.ietf.org/html/rfc2616#section-13.4)
- [14.9.1 What is Cacheable?](https://tools.ietf.org/html/rfc2616#section-14.9.1)
- [14.8 Authorization](https://tools.ietf.org/html/rfc2616#section-14.8)

13.4 says:

> Note that section 14.8 normally prevents a shared cache from saving and returning a response to a previous request if that request included an Authorization header.

14.8 explains that authenticated requests may be cached up to the max-age, and then not after that.

Requests to the public Shields server never have authentication headers, so I'm not sure I understand what the benefit would be of setting `public` vs leaving it alone. Basically max-age already implies `public`.

Requests to self-hosted Shields servers probably should cached according to the standard semantics. If they have Auth headers, those requests should be cached like any other requests with Auth headers.

This is based on what I'm getting from reading the spec, though I'm open to being convinced otherwise.

chris48s:
I've removed the `public` header. Do we have a good way to monitor the affect of merging this on how much traffic we're serving from cache? (particularly on the raster side)

paulmelnikow:
I'm not sure! I'm happy to give you access to the Zeit app if you want to check it out.

chris48s:
I think its the cache on now rather than cloudflare where we're hoping to see a difference as a result of this.

chris48s:
I just ran a deploy pushing this commit to production.
If I run `curl -I "https://img.shields.io/badge/build-passing-green.png?style=social"` a few times in a row, I'm still getting
```
cache-control: max-age=2592000
via: 1.1 vegur
cf-cache-status: MISS
```
on every request so I think we're still serving nothing from now's CloudFront cache - we're invoking the lambda for every request. I can't find anything in the now dashboard that tells us much cache usage. We could try explicitly sending `public` on the response from svg-to-image-proxy but not shields itself.. or we could give up on it and focus on alternative solutions

paulmelnikow:
I think that request is for the redirect:

```
$ curl -I "https://img.shields.io/badge/build-passing-green.png?style=social"
HTTP/2 301
date: Wed, 06 May 2020 19:45:36 GMT
content-length: 0
set-cookie: __cfduid=d03a0ee9dfbae65b456bc75960c6276551588794336; expires=Fri, 05-Jun-20 19:45:36 GMT; path=/; domain=.shields.io; HttpOnly; SameSite=Lax
location: https://raster.shields.io/badge/build-passing-green.png?style=social
cache-control: max-age=2592000
via: 1.1 vegur
cf-cache-status: MISS
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
server: cloudflare
cf-ray: 58f533dc6b2defe4-EWR
cf-request-id: 028d1ebdbf0000efe415278200000001
```

I guess the thing to check is the URL the raster proxy would be sending to Shields. After hitting this URL a few times I'm seeing cache hits:

```
~ $ curl -I "https://img.shields.io/badge/build-passing-green.svg?style=social"
HTTP/2 200
date: Wed, 06 May 2020 19:48:28 GMT
content-type: image/svg+xml;charset=utf-8
set-cookie: __cfduid=ddbc7d0fea7151cda0fd3fa2c213231301588794508; expires=Fri, 05-Jun-20 19:48:28 GMT; path=/; domain=.shields.io; HttpOnly; SameSite=Lax
cache-control: max-age=86400 s-maxage=86400
last-modified: Wed, 06 May 2020 19:16:11 GMT
via: 1.1 vegur
cf-cache-status: HIT
age: 42
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
server: cloudflare
cf-ray: 58f5380d3d43f051-EWR
cf-request-id: 028d215c410000f0511b9c0200000001
```

And now (though not as of a couple minutes ago) I'm seeing cache hits on the rasters themselves:

```
$ curl -I "https://raster.shields.io/badge/build-passing-green.svg?style=social"
HTTP/2 200
content-type: image/png
cache-control: public, max-age=0, must-revalidate
date: Wed, 06 May 2020 19:49:21 GMT
last-modified: Wed, 06 May 2020 19:16:12 GMT
content-length: 0
x-vercel-cache: HIT
x-now-cache: HIT
age: 2
x-now-trace: iad1
server: now
x-vercel-id: iad1::sfo1::fmrf2-1588794564358-7b0005b8989e
x-now-id: iad1::sfo1::fmrf2-1588794564358-7b0005b8989e
strict-transport-security: max-age=63072000
```

chris48s:
> I think that request is for the redirect

:facepalm: rookie error!

calebcartwright:
I'm thinking this may be cause behind the failing tests as described in #5046 :smile:

paulmelnikow:
It’d a little weird that the redirect request isn’t being cached. I wonder why that is.

Issue #4930, opened 2020-04-22
Daily Tests Job Failing (not service tests)


calebcartwright:
We've had a handful of persistently failing service tests (seems like each time we've fixed one a new one popped up :smile:) that I hadn't even noticed the `Run main tests` step was failing and has been for at least a month. 

```bash
0323034607 Error: I've made a huge mistake
    at ThrowingService.handle (/home/circleci/project/shields/core/base-service/base.spec.js:251:17)
    at Function.invoke (/home/circleci/project/shields/core/base-service/base.js:76:1710)
    at Context.<anonymous> (/home/circleci/project/shields/core/base-service/base.spec.js:255:31)
    at callFn (/home/circleci/project/shields/node_modules/mocha/lib/runnable.js:387:21)
    at Test.Runnable.run (/home/circleci/project/shields/node_modules/mocha/lib/runnable.js:379:7)
    at Runner.runTest (/home/circleci/project/shields/node_modules/mocha/lib/runner.js:535:10)
    at /home/circleci/project/shields/node_modules/mocha/lib/runner.js:653:12
    at next (/home/circleci/project/shields/node_modules/mocha/lib/runner.js:447:14)
    at /home/circleci/project/shields/node_modules/mocha/lib/runner.js:457:7
    at next (/home/circleci/project/shields/node_modules/mocha/lib/runner.js:362:14)
    at Immediate._onImmediate (/home/circleci/project/shields/node_modules/mocha/lib/runner.js:425:5)
    at runCallback (timers.js:810:20)
    at tryOnImmediate (timers.js:768:5)
    at processImmediate [as _immediateCallback] (timers.js:745:5)
```
https://circleci.com/gh/badges/daily-tests/1056

There's also other errors popping up in some of the logs, for example this one from a recent run:
```bash
0421034630 Error: Cannot push metrics. Cause: NetConnectNotAllowedError: Nock: Disallowed net connect for "shields-metrics.io:80/metrics"
    at InfluxMetrics.sendMetrics (/home/circleci/project/shields/core/server/influx-metrics.js:1:6674)
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:229:7)
```

Haven't really had a chance to dig into it but wanted to open this to track it

platan:
Errors regarding metrics was introduced recently. This is a fix https://github.com/badges/shields/pull/4931

PyvesB:
[...]

PR #5050, opened 2020-05-03
increase timeout for service name check (again)

calebcartwright:
Follow up to #5047 as another attempt to address the issue in #4930 related to the `coverage:test` npm script failing consistently in the daily tests job.

The last bump increased the timeout to 5 seconds but that [didn't seem to help](https://github.com/badges/shields/issues/4930#issuecomment-623181834)

I was hoping to be able to debug/test exclusively over in the daily tests repo. Unfortunately though, mocha does not support setting a default timeout via an environment variable (at least not that I could find), and because the script uses `run-s` I don't see a way to get a `--timeout` override to mocha in that context.

Hence an excessive timeout set here :smile:


paulmelnikow:
Thanks for digging into this!

PR #5051, opened 2020-05-04
upgrade to prettier 2

### LINKED ISSUE #4972: upgrade prettier
:clipboard: **Description**

- review open PRs: anything big we want to try and get over the line first? (also a good time to do a  bit of a clean out)
- can we adopt some settings that mean we don't have to change _every single file_ in the codebase to upgrade it?

chris48s:
Closes #4972

I've gone with `arrowParens: avoid`, but it doesn't help much.. the thing that would really help to ease this transition would be if `function() {` --> `function () {` was configurable, but AFAICT [its not](https://prettier.io/docs/en/options.html).


.prettierrc.yml
```diff
@@ -3,3 +3,4 @@ singleQuote: true
 trailingComma: es5
```
paulmelnikow, line 3: This line can be removed, as it's now the default.

PR #5053, opened 2020-05-06
Update `depcheck` script for minimum supported Node version

paulmelnikow:
Ref #4977


paulmelnikow:
Heh. The e2e failure will be fixed by a change to the Circle config in #5052.

paulmelnikow:
I think it should just work this time.

PR #5079, opened 2020-05-10
implement [Discord] online/total

### LINKED ISSUE #4500: `Discord online/total` badge
:clipboard: **Description**

A discord badge as the one currently available, with the only difference, that it shows the total number of people in the server as well:

E.g. `Discord: 15/50 online`

:link: **Data**

I am not sure if there is such an endpoint with the available data in the Discord API.

:microphone: **Motivation**

Showing the number of online users is good, but sometimes you want to also show the number of people that have joined the server in total. It's good to show the total interest in your project/server.

kenanchristian:
Fixes #4500


services/discord/discord.service.js
```diff
@@ -27,6 +32,9 @@ const documentation = `
   To use the Discord badge a Discord server admin must enable the widget setting on the server.
 </p>
 <iframe src="https://player.vimeo.com/video/364220040" width="640" height="210" frameborder="0" allow="autoplay; fullscreen" allowfullscreen></iframe>
+<p>
+  To get the total member count you need to create a invite that never ends, and then add the Invite ID after the Server ID in the URL
```
PyvesB, line 36: Maybe we could be _extra_ clear by adding something along the lines of "If you do not include the Invite ID, only the number of members presently online will de displayed."

services/discord/discord.service.js
```diff
@@ -37,16 +45,22 @@ module.exports = class Discord extends BaseJsonService {
...
-      pattern: ':serverId',
+      pattern: ':serverId/:inviteId?',
     }
   }
 
   static get examples() {
     return [
       {
-        title: 'Discord',
+        title: 'Discord Server',
```
PyvesB, line 55: The example titles don't really convey what the badges will be displaying. Would the following be more informative?
* Discord Online Members
* Discord Online & Total Members

Feel free to suggest something else as well! :)

services/discord/discord.tester.js
```diff
@@ -11,6 +11,14 @@ t.create('gets status for Reactiflux')
     color: 'brightgreen',
   })
 
+t.create('gets online presence for Reactiflux')
+  .get('/102860784329052160/NdsqWcj.json')
+  .expectBadge({
+    label: 'chat',
+    message: Joi.string().regex(/^[0-9]+\/[0-9]+ online$/),
```
PyvesB, line 18: 👍

services/discord/discord.service.js
kenanchristian: Sure I'll add that in!

services/discord/discord.service.js
kenanchristian: I like the `Discord Online & Total Members`, does it make sense to merge the examples into 1? Since the invite id is optional, and when it's not added it will show just like the old one

PyvesB: As you like, I think it's also fine to keep the two examples!

services/discord/discord.service.js
```diff
@@ -27,6 +32,10 @@ const documentation = `
   To use the Discord badge a Discord server admin must enable the widget setting on the server.
 </p>
 <iframe src="https://player.vimeo.com/video/364220040" width="640" height="210" frameborder="0" allow="autoplay; fullscreen" allowfullscreen></iframe>
+<p>
+  To get the total member count you need to create a invite that never ends, and then add the <code>INVITE ID</code> after the <code>SERVER ID</code> in the URL<br>
+  It will only show the online member when <code>INVITE ID</code> is not provided.
```
PyvesB, line 37: Is a word missing here, i.e. `online member count`?

calebcartwright:
One thing I noticed while looking at the invite api back when #4500 was opened is that while the presence count shown with guild api https://discordapp.com/api/guilds/${serverId}/widget.json seems to match up precisely with what I saw in reality in Discord, the _approximate_ presence count in the invite api was usually _way_ off. 

I haven't had a chance to see if that's still the case, but I'd suggest we try a few guilds/invites in the review app to find out. What I think would be particularly confusing for users that were using the modal badge window on the Shields site is seeing the online count vary wildly when they add/remove the invite code param

PyvesB:
Hmmm... I am not seeing the changes from this PR in the [review app](https://shields-staging-pr-5079.herokuapp.com). Is this the case for others?

PyvesB:
> Hmmm... I am not seeing the changes from this PR in the [review app](https://shields-staging-pr-5079.herokuapp.com). Is this the case for others?

Recreating the review app did not improve things. Other review apps, for example for  #5081, are working as expected. Will try starting the server locally and have a dig around.

PyvesB:
Everything starts up as expected locally, I can successfully use the new badge.

Here are the counts I got from a few Discord servers:

| Server name | Precise count ("old" badge) | Approximate count (PR badge) |
| ------------- | ------------- | ------------- |
| Shields (11:00 UTC)  | 30  | 30 |
| Shields (11:05 UTC)  | 32  | 31 |
| Shields (13:35 UTC)  | 33  | 34 |
| Reactiflux (11:00 UTC)  | 6400 | 6401 |
| Reactiflux (11:05 UTC)  | 6429 | 6438 |
| Reactiflux (13:35 UTC)  | 6911 | 6916 |
| Minimally Correct (13:40 UTC)  | 36 | 36 |
| Home Assistant (13:40 UTC)  | 4529 | 4532 |
| MideLevel (13:50 UTC)  | 145 | 144 |
| EndenJS (13:55 UTC)  | 8 | 8 |

Based on this small set of examples, there are indeed differences which seem to grow with the activity of the server. Unless I got lucky, they don't seem to be as big as you were describing @calebcartwright, but still potentially a blocker for this. What do people think?

Of course an obvious solution would be to do two API calls, one to retrieve the precise online count, the other to retrieve the total count, but that would add complexity and slow down badge rendering, so I don't think we should go down that route.

calebcartwright:
Thanks @PyvesB for digging into it and sharing the numbers! I think there's actually two comparison topics here:

1. invite code API vs. guild widget API
2. either/both of those API response values compared to the actual online counts users can see within the Discord app/website

Back when I was last looking at the two APIs in January I was seeing a higher variance than the numbers you shared (25 vs 15 online users for the Shields server for example), so things do indeed seem better. 

However, I am still seeing some larger discrepancies between the two APIs, Reactiflux shown below (~60 user difference, though I also saw one closer to 100 earlier) 
image

image

Then for the second item, I also saw a case earlier where the actual online count was < 3500 but the Invite Code API response had > 4000 approximate count (the widget setting is enabled on that server so I can't compare the two API responses)

image

Generally speaking, I think we all agree we should be really cautious about adding a new badge (or a new format for a badge) that impacts any existing badge(s).

PyvesB:
Thanks for your thoughts @calebcartwright .

I would say we go ahead and merge this, however I would suggest the following changes before doing so:
* indicate in the badge's documentation that the Discord API returns an approximate present count and that the numbers indicated in the badge may vary slightly compared to the actual count.
* remove the serverId parameter from this new badge, as it's not actually being used if I'm not mistaken. Server ID and invite ID have different formats, so we can probably build some sort of regex to differentiate between the two.

services/discord/discord.service.js
```diff
@@ -72,32 +93,45 @@ module.exports = class Discord extends BaseJsonService {
...
+        ? `https://legacy-img.shields.io/discord/${serverId}/${inviteId}.json`
+        : `https://legacy-img.shields.io/discord/${serverId}.json`,
       schema: proxySchema,
     })
   }
 
-  async handle({ serverId }) {
+  async handle({ serverId, inviteId }) {
     if (this._shieldsProductionHerokuHacks) {
       const { message, color } = await this.fetchOvhProxy({ serverId })
```
PyvesB, line 123: This line needs the `inviteId` as well.

PyvesB: No longer needed given that this line was removed since #5346.

PyvesB:
@kenanchristian any chance you could circle back to this and address the outstanding review comments? I don't think there's much remaining: a rebase, a bit of documentation polishing and removing a parameter from the badge path. 😉

sdabhi23:
@PyvesB I believe that having a single variable for both serverId and inviteId would confuse the users, naming it correctly would be a different problem altogether.

We can address this issue by naming the variable as **:id** and adding instructions in the document, but I would prefer to go with the original design with 2 variables or having different base paths for the badges. What are your thoughts on this?

PyvesB:
Do you know if serviceId and inviteId have different formats? If so, we could keep the same concise `/discord/some-id` path, but then have `pattern: ':serverId[some-regex]'` and `pattern: ':inviteId[some-other-regex]'` to differentiate between the two in the implementation. Like that you avoid having the same variable name for both. Not sure whether that's a good idea though. @calebcartwright any thoughts from your side?

calebcartwright:
> Not sure whether that's a good idea though. @calebcartwright any thoughts from your side?

Still trying to get back into the context of this one. My recollection was that we're going to go with invite code for the official/advertised badge that would be listed on Shields.io, though obviously for BC we need to make sure existing badges using serverid don't break.

I think it'd be good to have two Discord badge variants that users could choose between:




and the user-provided parameter would be the same (presumably the invite code), with either a route difference or query param to let users choose between the two. 

I feel like providing documentation around two different id's and actively advertising/promoting both will likely just lead to confusion, so my preference would be that the BC support for existing serverid badges be hidden on Shields.io and new users going forward. I don't really have any strong opinions around the route pattern beyond taht, other than I agree with the need to change what's currently here (`pattern: ':serverId/:inviteId?'`)

PyvesB:
> Still trying to get back into the context of this one. 

The joys of picking up a unattended PR several months later! 😄 

> My recollection was that we're going to go with invite code for the official/advertised badge that would be listed on Shields.io, though obviously for BC we need to make sure existing badges using serverid don't break.

Well, I think I had misunderstood things in that case. In my mind, we were going to keep the two badges advertised on Shields.io, leave the existing implementation as it is, and only make the new one use the invite API. Basically what is done in the current PR, with a few tweaks to the path and to the documentation. The considerations around the two API calls stuck in my head, it really seemed to me that we were debating whether we wanted to do invite-only or invite+server for the new badge.

One thing worth noting is that the Discord token I've been given has a rate-limit ten times smaller for the invite API than it has for the server API. My back of the envelope calculations show that even with that it should still be more than enough if we shifted the existing badge to the invite API, but I still think we should keep that in mind.

I think the three solutions we have are:
* switch to invite for both badges and only use that.
* stick with the current state of the PR with old -> server, new -> invite.
* stick with server for old and use server+invite for new.

Once we've decided which API integration we want, then we can design the badge paths, query params, how we orchestrate the docs and examples, etc.

sdabhi23:
> Do you know if serviceId and inviteId have different formats?

Yes. ServerId is a snowflake, which means that it is a numeric string. InviteId is a alphanumeric string

After going through all the discussions I was also under the same impression as Pierre, that we were going to keep the two badges advertised on Shields.io, leave the existing implementation as it is, and only make the new one use the invite API

>I think the three solutions we have are:
> 1. switch to invite for both badges and only use that.
> 1. stick with the current state of the PR with old -> server, new -> invite.
> 1. stick with server for old and use server+invite for new.

- For 1 we would have to separately manage BC. I am not sure about how to manage it without advertising the path on shields.io, but will figure it out if we want to implement it in this way
- The 2nd option makes more sense to me as it will help with the rate limits as well if we see huge increase in traffic in future
- I would prefer not to make 2 API calls for a single badge as it will practically cut our rate limits to half of what they are right now while the total member count will still be an approximate

calebcartwright:
> One thing worth noting is that the Discord token I've been given has a rate-limit ten times smaller for the invite API than it has for the server API

That's probably my misunderstanding. I didn't realize our increased rate limit was scoped to a single API. 

> we were debating whether we wanted to do invite-only or invite+server for the new badge

I'm opposed to the invite+server requirement for the new badge. I'm not following why we'd require user to provide two duplicative params for a single badge, especially when one is completely unused.

PyvesB:
Okay, thanks for clarifying. 👍🏻 

> The data points with both are imprecise estimates

Shall we double-check this to make sure we're headed in the right direction? You mentioned observing different variance between January and May, I don't think it would hurt to do another sanity check given that it's been several months. We're also using a slightly different API endpoint than we were at the time (different domain and an additional `/v6` bit is now present in the path); I wouldn't expect that to impact the backend processing, but you never know.

@sdabhi23 would you like to help with that analysis? Basically producing a similar table to what I did in https://github.com/badges/shields/pull/5079#issuecomment-633050648, with an additional column indicating the value displayed in the Discord app/website itself. Like that we'll be able to do an up-to-date comparison between:
* `presence_count` from `https://discord.com/api/v6/guilds/${serverId}`.
* `approximate_presence_count` from `https://discord.com/api/v6/invites/${inviteId}?with_counts=true)`.
* the value displayed by either the app or the website.

calebcartwright:
> Shall we double-check this to make sure we're headed in the right direction?

Yup, that's a good idea! It'd be different if the serverid value is consistent with real life while inviteid is still an ~estimate

PyvesB:
There hasn't been any activity for a while, so I'm going to go ahead and close this pull request. Anyone who is willing to do the analysis I suggested in https://github.com/badges/shields/pull/5079#issuecomment-690733584, feel free to do so and report your findings in  #5079.

Thanks again for your interest in the Shields.io project!

PR #5081, opened 2020-05-12
[Scoop] Add scoop badge service

0xced:
This pull request adds version badges for [Scoop](https://scoop.sh), a command-line installer for Windows.


PyvesB:
Thanks for your contribution @0xced !

I haven't yet done a detailed review of the code, but I noticed you were using https://github.com/lukesampson/scoop/blob/master/buckets.json to get some of the information. What are the guarantees around this file? Is there any risk of it going away or changing anytime soon? Do other applications rely on it? If we go ahead with this, it may be worth informing the Scoop team that we're making use of it for this badge.

0xced:
> What are the guarantees around this file? Is there any risk of it going away or changing anytime soon?

I think it's pretty safe to use it. It's been there since 2013 and got some updates to the bucket URLs but neither the file location nor the JSON schema has changed over the years.

> Do other applications rely on it?

I don't know about other applications, but [Scoop itself relies on it](https://github.com/lukesampson/scoop/blob/48f121e4668ff38fbbad99185244edb19a87f0d3/lib/buckets.ps1#L37-L41).

> If we go ahead with this, it may be worth informing the Scoop team that we're making use of it for this badge.

I just informed them here: https://github.com/lukesampson/scoop/issues/3978

services/scoop/scoop.service.js
```diff
@@ -0,0 +1,95 @@
+'use strict'
+
```
calebcartwright, line 2: Let's go ahead and name the file `scoop-version.service.js` to match the classname and because there's a reasonably good chance we'll add other scoop-related badges in the future (for example licenses, etc.)

services/scoop/scoop.service.js
```diff
@@ -0,0 +1,95 @@
...
+    const bucket = queryParams.bucket || 'main'
+    const bucketUrl = buckets[bucket]
+    if (!bucketUrl) {
+      throw new NotFound({ prettyMessage: `bucket "${bucket}" not found` })
+    }
+    const {
+      groups: { user, repo },
+    } = gitHubRepoRegExp.exec(bucketUrl)
+    try {
+      const { version } = await fetchJsonFromRepo(this, {
```
calebcartwright, line 78: I'm only vaguely familiar with scoop and don't know anything about buckets/etc. I take it the various buckets listed in https://github.com/lukesampson/scoop/blob/master/buckets.json are each registries of sorts where apps can be listed for consumption and thus installed via `scoop`.

If that's correct, do you know how dynamic that bucket list is? I'm assuming that the vast majority of apps would be registered/listed under the `main` bucket and I'm wondering if we can avoid having to make 2 separate API calls for at least the majority of cases.

services/scoop/scoop.tester.js
```diff
@@ -0,0 +1,44 @@
+'use strict'
+
```
calebcartwright, line 2: Same comment as above regarding the filename (`scoop-version.tester.js`)

services/scoop/scoop.service.js
calebcartwright: Also, is it possible for additional "3rd party" and/or private buckets? (could I have my own private/internal bucket within a corporate network for example)

services/scoop/scoop.service.js
0xced: Done in 97897c3f5001443a09e3677c8b660bbdd3daeab6.

services/scoop/scoop.tester.js
0xced: Also done in 97897c3f5001443a09e3677c8b660bbdd3daeab6.

services/scoop/scoop.service.js
PyvesB: Maybe we could simply do that second API call once and cache the result for all subsequent calls? Given that Shields instances are redeployed on a regular basis and given that the file changes rarely, we probably don't need to implement any fancy refresh strategy.

calebcartwright: > we probably don't need to implement any fancy refresh strategy.

Not sure what you were referring to by fancy refresh (we may be thinking the same thing), but I'm simply wondering if we can go ahead make the one API call for the app under https://github.com/ScoopInstaller/Main/tree/master/bucket/foo.json (assuming that's the base default/most common collection of apps) in the majority of cases instead of always first calling https://github.com/lukesampson/scoop/blob/master/buckets.json.

PyvesB: I was thinking keep most of the implementation as is, but do the request to https://github.com/lukesampson/scoop/blob/master/buckets.json only once and store its return value in a field on first successful retrieval.

PyvesB: What are your thoughts on my above suggestion @0xced? :)

PyvesB:
@0xced I'm keen on getting this pull request landed, I don't think there's much left to do. Any chance you can circle back to this in the coming weeks? 😉

services/scoop/scoop-version.service.js
```diff
@@ -0,0 +1,95 @@
...
+const Joi = require('@hapi/joi')
+const { NotFound } = require('..')
+const {
+  ConditionalGithubAuthV3Service,
+} = require('../github/github-auth-service')
+const { fetchJsonFromRepo } = require('../github/github-common-fetch')
+const { renderVersionBadge } = require('../version')
+
+const gitHubRepoRegExp = new RegExp(
+  'https://github.com/(?<user>.*)/(?<repo>.*)/?'
```
kidonng, line 12: Note this works well without the trailing slash, but if there is trailing slash you will get:

```js
gitHubRepoRegExp.exec('https://github.com/lukesampson/scoop/')
// user: "lukesampson/scoop"
// repo: ""
```

```suggestion
  'https://github.com/(?<user>.*?)/(?<repo>.*?)(/|$)'
```

PyvesB: Thanks for pointing this out @kidonng. Would you be interested in helping us finish this implementation?

kidonng: Did you mean finishing this PR? I see it's basically working but lacks the ability to cache `buckets.json`, I'd like to help but I can't find documentation about handling caching.

PyvesB: Yes, potentially finishing the PR if you're up for it!

Caching the `bucket.json` doesn't have to be very fancy. On the first request, get the file and store its contents in a class field that is reused for all subsequent requests. :)

PyvesB:
Given that there was little left to do and that quite a few people seemed interested in this, I've gone ahead and completed the pull request. Could someone else from @badges/shields-maintainers please have a final look at this piece of work?

calebcartwright (approved):
LGTM

PR #5105, opened 2020-05-21
update [discord] domain

chris48s:
One thing to note with this is that because discord uses OVH as a proxy in production, we need to deploy this to OVH as well some time before discordapp.com goes away


paulmelnikow:
> One thing to note with this is that because discord uses OVH as a proxy in production, we need to deploy this to OVH as well some time before discordapp.com goes away

That's actually a problem because of the Node version. If we need a quick solution, we could backport a fix to an old rev.

chris48s:
Yeah I guess we'll have to cherry-pick the commit.

Issue #2545, opened 2018-12-17
Print a danger warning when modified services are not in the PR title


paulmelnikow:
We already show a warning when services haven't been tested, though we don't compare that against the service names in the PR title. This should be pretty straightforward to do using the helper functions we have.

I'm not sure if Danger has access to the PR metadata, though we could generate it easily enough in an earlier step in the CI, using [`npm run test:services:pr:prepare`](https://github.com/badges/shields/blob/b4b29ab0f787069111db4bf332810493adf363c5/.circleci/config.yml#L226-L228).

platan:
Maybe we should run service tests for all modified services? Was it discussed before? With this approach you will not need to add service name to PR title and this would be a very user friendly.

chris48s:
> Maybe we should run service tests for all modified services? Was it discussed before?

Yes - https://github.com/badges/shields/pull/1936
I don't think any of the issues identified in that thread are impossible to solve, but there are definitely some tradeoffs to consider.

paulmelnikow:
One thing new since #1936 is that GitHub Actions have come online. I am not completely sure, though I think they provide a UI that lets users trigger actions. I'm not sure if you can set options for those actions. Might be worth an investigation, though! (Much easier than setting up Drone!)

Also along the original line of the issue, a danger warning is a nice place to start, because we can get a feel for how well the detection is working (false positives and false negatives) over a long period of time before we wire it up to the service-test run.

PR #5129, opened 2020-05-23
WIP use GH action to detect which service tests to run; also run [packagist]

chris48s:
refs https://github.com/badges/shields/issues/2545
refs https://github.com/badges/shields/pull/1936

I've been using github actions a fair bit on some other projects and I thought I'd revisit this idea.

The implementation I've got so far continues to allow us to specify tests to run in the PR title (as we do now). This is mainly useful for dependabot PRs where we aren't modifying any service or test code but still want to run some service tests.

It also uses the GH actions API to detect any `services/**/*.(service|tester).js` files modified and attempts to run relevant service tests based on the path. At the moment, the way I'm doing this is _fairly_ accurate, but I've realised that there are some cases where there's a bit of a mismatch between the directory names and service/tester names. For example `SubredditSubscribers` lives in `/reddit`, most of the tester names don't contain the `-` character, except `snap-ci` and `f-droid`, which do. Also some of the services that only export a redirector don't match with the path name. One thing we could do is rename some stuff to make it more consistent, but there's not really anything to stop further discrepancies appearing in future.

I think if we want to take this approach forward, what I'll probably do is write a second test loader that works based on providing paths and then use that to make this more reliable (but retain the current name-based loading for services specified in the PR title). This is a bit more involved though, so I figured I'd get some feedback on the WIP before I embark on that journey. There will also be some work on:
- docs
- testing
- removing Circle CI config
- moving secrets/keys from Circle env vars GH Actions
- updating branch protection rules, etc

to fully finish this..

Thoughts?

The service tests running on this PR are arbitrary - its just to demonstrate the action actually running some tests


chris48s:
boom: https://github.com/badges/shields/pull/5129/checks?check_run_id=702053286
Thinking about it, I suspect the `ci/circleci: services` builds will probably fail, as I've not disabled them but I've deleted some code they need..

calebcartwright:
I'll preface by saying that I've had generally positive experiences with GHA so far (though it's admittedly still fairly immature compared to the more established CI platforms) and am a fan. 

Is the thinking that long term we could migrate our CI/PR validation from Circle to GHA, or just moving the service test run jobs to GHA while keeping the rest in Circle?

chris48s:
> I'll preface by saying that I've had generally positive experiences with GHA so far (though it's admittedly still fairly immature compared to the more established CI platforms) and am a fan.

That exactly summarises my opinion on it too :)

> Is the thinking that long term we could migrate our CI/PR validation from Circle to GHA, or just moving the service test run jobs to GHA while keeping the rest in Circle?

Certainly at the moment I'd only be looking to move the service tests and leave everything else running on Circle CI. The reason I think GH actions is a good fit for this specific job is because its really easy to get an array of all the files changed in this PR in the CI build or otherwise interrogate the PR (for example, you'll notice I've been able to completely bin all the code in `infer-pull-request.js`). There's no reason why we _couldn't_ do that in Circle instead.. its just a bit more of a faff.

I don't see a problem with having some tests run in GH Actions and some in Circle, but if we have good experiences with it we could also look at moving more stuff in future.

chris48s:
this eventually got turned into https://github.com/badges/shields/pull/8421

Issue #5014, opened 2020-05-02
Move the frontend to heroku


chris48s:
:clipboard: **Description**

Currently we host the badge server on heroku and the front end on gh-pages.

This was originally done for performance reasons as it kept the traffic on shields.io's homepage(s) off the badge servers. This decision was made before we were behind a CDN.

The downside of this arrangement is that it means we don't benefit from a lot of the nice things that being on heroku buys us (e.g: automated deploys, easy rollbacks, etc).

Now that we are deployed on heroku behind a CDN, it should be possible to serve the front-end from the same dynos that serve the badges with minimal impact on performance. We can lean on the CDN to handle most of the traffic for us.

What would be the steps necessary to move the front-end off GH pages so we can move to fully automated (or one-click) deployment/rollbacks?

paulmelnikow:
I agree this would be pretty impactful. In addition to the benefits of rollbacking, etc., it makes deployment a one-click operation, rather than a one-click operation plus a `git pull && npm ci && make deploy-gh-pages`. It also ensures the deploy is pristine, not polluted by anything else that might be unusual in the local dev environment. It's more transparent and secure, too.

One blocking issue is to de-overlap the routes between the badge server and the frontend. I believe the only one which overlaps is https://shields.io/endpoint and https://img.shields.io/endpoint, which can be distinguished only by the presence of a query parameter. I would prefer not to change either of these, to be honest, as there are lots of links pointing to both. Though I would also rather not have to introduce complicated routing to accommodate this. Maybe we could put the exception logic in the endpoint badge? i.e. fall back to serving the generated endpoint html when there is no query parameter. This probably deserves discussion in its own issue.

Other than that… this already basically works on staging! So if we point shields.io to the same app as img.shields.io we're most of the way there.

We should think about whether we want to force badge users to user https://img.shields.io/ vs https://shields.io/, and vice versa. If we decide we want to force users to address the badge and the frontend separately, that could be another way to resolve the /endpoint conflict (i.e. examine the host on the request and route appropriately.) The conflict would remain unresolved in staging and other Heroku deploys but that's not a big deal.

There may be something special that has to be done because it's the zone apex.

Finally, we could think about caching, and what cache headers we want to send (from the badge server) for the frontend, and may need a solution for cache-busting at deploy time.

chris48s:
I had a bit more of a think about this/play with stuff after our conversation got cut short yesterday.

[...]

PR #5137, opened 2020-05-27
redirect /endpoint with no query params to endpoint docs

chris48s:
Refs #5014

Had a look at this.
I think I'm still in favour of just moving the endpoint docs, but if we're going to do a redirect, I think this is what the solution looks like.

The actual service badge class is sufficiently abstracted that we can't really issue the redirect there. This is mostly a good thing: we're working at a sufficiently high level in the service classes that we don't really think about issueing a HTTP response - just what text/colours/etc we want to render on a badge. It means we can't really do the redirect in the service class though.

The other approach I can see would be to define a custom `RedirectException`, throw it in `handle()` and then catch it and issue the redirect, but the handling code for that ends up in `legacy-request-handler` anyway. It seems like its probably not worth defining a mechanism to do a thing we don't really want to be doing.


PyvesB:
I've got no strong opinion on whether we should use a redirect or move the endpoint documentation. Should we just go with this approach? We can always choose to change the endpoint at a later point if we have second thoughts. :)

chris48s:
Yeah that's a good point. It is probably more valuable to do this to unblock moving the front-end to heroku (and hence automated deploys) then leave open the option to change it later.

PyvesB (approved):

PR #5154, opened 2020-05-30
Add the /steam/update-date badge, run [steam]

saibotk:
This adds the /steam/update-date badge, to also provide a badge that shows the date of a file on steam when it was last updated.

The code is pretty much straight forward, equal to the `steam/release-date` badge, except that it reads out `time_updated` instead of `time_created`.

If there is anything you would like to have changed, please let me know! 
Thank you for this amazing project!

### DIFF
--- a/services/steam/steam-workshop.service.js
+++ b/services/steam/steam-workshop.service.js
@@ -63,6 +63,7 @@ const steamFileSchema = Joi.object({
           Joi.object({
             file_size: Joi.number().integer().required(),
             time_created: Joi.number().integer().required(),
+            time_updated: Joi.number().integer().required(),
             subscriptions: Joi.number().integer().required(),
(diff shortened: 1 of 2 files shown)


calebcartwright:
Thank you for the PR @saibotk! A few of the CI checks are failing (including linter and service tests), so could you take a look and update accordingly?

saibotk:
Thanks @calebcartwright, but as it seems, I would have to log in to CircleCI to see the output, so it's hard to do. But otherwise I also don't really know why this would fail anyways, as I just did the same as the steam/release-date function :smile: 
But maybe I also need to add the new service endpoint to other tests or another list?

saibotk:
Nervermind @calebcartwright I found out what was wrong and it should now all work out.

Issue #5161, opened 2020-06-02
Badge Request: All Contributors


JoeIzzard:
:clipboard: **Description**

All Contributors is a specification designed to show all contributors no matter there contribution. Currently they have a bot that manually updates a manual Shield.io Badge in Repo's however this doesn't work too well. The alternative that works is the dynamic json endpoint with a query. This works but isn't the easiest to generate. A wrapper to make the URL nicer would be perfect. It would show the total contributors for a repo, as defined by that system. For examlple:

All Contributors | 29

:link: **Data**

The data already exists in repo's that use it as a JSON file that can be accessed via GitHub Raw. The url would be:

`https://raw.githubusercontent.com/*USER*/*REPO*/*BRANCH*/.all-contributorsrc`

As the Bot updates the master branch, it would be safe to use that branch to pull the stats from. Currently some are using Shields JSON Dynamic endpoint, with a query to count the contributors array. An example of Dynamic JSON URL is:

`https://img.shields.io/badge/dynamic/json?color=orange&style=flat-square&label=All%20Contributors&query=%24.contributors.length&url=https%3A%2F%2Fraw.githubusercontent.com%2FJoeIzzard%2F.github%2Fmaster%2F.all-contributorsrc`

This is dynamic url for my [.github](https://github.com/joeizzard/.github) repo.
 

:microphone: **Motivation**

Currently Contributor counts only include people that actually contribute code, where as the all contributors project also recognises those that contribute out side of committing code, such as graphics, bug reports and such. This allows you to show an accurate contributor count for projects that instead use that specification.

JoeIzzard:
I'm going to have a go at implementing this, but not done anything for shields before so any tips more than welcome!

chris48s:
We have a tutorial you can follow though for adding a new badge:
https://github.com/badges/shields/blob/master/doc/TUTORIAL.md

This badge falls into the category of "grab a file from a hard-coded location in a github repo and extract some data from it". Fortunately this is a well trodden path - there are several other badges that do this. Some examples to crib from:

* https://github.com/badges/shields/blob/master/services/github/github-lerna-json.service.js
* https://github.com/badges/shields/blob/master/services/github/github-manifest.service.js
* https://github.com/badges/shields/blob/master/services/github/github-package-json.service.js
* https://github.com/badges/shields/blob/master/services/github/github-pipenv.service.js

`.all-contributorsrc` is json, so you can make use of the `fetchJsonFromRepo` common helper.

[...]

PR #5163, opened 2020-06-04
Implement All Contributors Badge, run [GitHubAllContributors]

JoeIzzard:
Implements #5161

Adds the 'all contributors' badge, using the contributor count color scale.

### DIFF
--- a/services/github/github-all-contributors.service.js
+++ b/services/github/github-all-contributors.service.js
@@ -1,0 +1,59 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { renderContributorBadge } = require('../contributor-count')
+const { ConditionalGithubAuthV3Service } = require('./github-auth-service')
+const { fetchJsonFromRepo } = require('./github-common-fetch')
+const { documentation } = require('./github-helpers')
+
+const schema = Joi.object({
+  contributors: Joi.array(),
+}).required()
+
+module.exports = class GithubAllContributorsService extends ConditionalGithubAuthV3Service {
+  static get category() {
+    return 'activity'
+  }
+
+  static get route() {
+    return {
+      base: 'github/all-contributors',
+      pattern: ':user/:repo',
+    }
+  }
+
+  static get examples() {
+    return [
+      {
+        title: 'Github All Contributors',
+        namedParams: {
+          repo: 'all-contributors',
+          user: 'all-contributors',
+        },
+        staticPreview: this.render({ contributorCount: 66 }),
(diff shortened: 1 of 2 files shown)


services/github/github-all-contributors.service.js
```diff
@@ -0,0 +1,59 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { renderContributorBadge } = require('../contributor-count')
+const { ConditionalGithubAuthV3Service } = require('./github-auth-service')
+const { fetchJsonFromRepo } = require('./github-common-fetch')
+const { documentation } = require('./github-helpers')
+
+const schema = Joi.object({
+  contributors: Joi.array(),
```
calebcartwright, line 10: We need to ensure that the contributors array always exists as part of the response validation, so let's add a `.required()`
```suggestion
  contributors: Joi.array().required(),
```

JoeIzzard:
Thanks for pointing that out @calebcartwright, I thought the required at the end of the schema did that, but wasn't sure. I've implemented that change 👍

services/github/github-all-contributors.service.js
```diff
@@ -0,0 +1,59 @@
...
+
+module.exports = class GithubAllContributorsService extends ConditionalGithubAuthV3Service {
+  static get category() {
+    return 'activity'
+  }
+
+  static get route() {
+    return {
+      base: 'github/all-contributors',
+      pattern: ':user/:repo',
```
calebcartwright, line 21: Sorry one more question/thought regarding the branch. In the referenced issue you mentioned:

> As the Bot updates the master branch,

Is that always explicitly master, or is it the default branch in the repository?

JoeIzzard: Just done some testing, it does indeed push to the default branch not always master. I will look into adding a way to handle that, and a specified branch.

calebcartwright: We can just add the `branch` route param, looks like you've already got everything else wired up to handle it :+1: 

```suggestion
      pattern: ':user/:repo/:branch*',
```

JoeIzzard: I've implemented that, as well as update the example given to the Frontend, from the searching I have done it seems there's currently no way to use the default branch (Issue #4114) if a branch isn't specified, would the be correct?

calebcartwright: Correct. We do the best we can today, and for folks with different defaults they need to include an extra param in their badge url

Issue #5165, opened 2020-06-04
[Subreddit-subscribers] Doesn't recognise private subs


JoeIzzard:
Issue is with:

- [X] [shields.io](https://shields.io/#/)
- [X] My own instance
- [ ] [badge-maker NPM package](https://www.npmjs.com/package/badge-maker)

:beetle: **Description**

The badge Subreddit-Subscribers doesn't recognise a private sub differently to a none existing sub. This causes the error message to be 'invalid' when the sub does exist.

:link: **Link to the badge**

![Century Club](https://img.shields.io/reddit/subreddit-subscribers/centuryclub?style=social)

Here I have used the example of the CenturyClub Subreddit which will always be private.

:bulb: **Possible Solution**

Possible solution is to add an error message for a '403' Forbidden header, this could then result in a specific error message for that case.

JoeIzzard:
This has also caused the valid subreddit test to fail temporarily as the used subreddit has been made private for 24 hours. May also explain reported drop in code coverage for that service?

JoeIzzard:
Currently working on implementing the 403 error message, and adding a test to mock missing data to help code coverage. PR will be ready soon.

JoeIzzard:
Fixed by PR #5166, now in production.

PR #5166, opened 2020-06-05
[Reddit] New badge and additional testing

JoeIzzard:
Fixes Issue #5165, [Subreddit-Subscribers] not recognising a private subreddit. It now returns the following when met with a 403 error, which is what Reddit gives when the subreddit is private:

Reddit | subreddit is private

Also added tests to test for this, and a JSON reply without the required information.

While I was at it I have also added a new badge for Reddit User Karma, this supports link, comment or combined karma using the same basic method as the Subreddit Subscribers. Included tests for this new badge.

Finally, to make testing for the whole service easier, I have renamed the class for Subreddit-Subscribers to allow you to test the service 'reddit' instead of testing 'subreddit' and the new 'reddit' separately.


services/reddit/subreddit-subscribers.tester.js
```diff
@@ -23,3 +23,22 @@ t.create('subreddit-subscribers (not existing subreddit)')
...
   })
+
+t.create('subreddit-subscribers (private sub)')
+  .get('/centuryclub.json')
+  .expectBadge({
+    label: 'reddit',
+    message: 'subreddit is private',
+  })
+
+t.create('subreddit-subscribers (private sub)')
```
calebcartwright, line 34: could you provide some info about the purpose of this test?

JoeIzzard: The test 'subreddit-subscribers (private sub)' is testing a subreddit that is private and will return a 403 error code, to check that it is being handled correctly. CenturyClub is a club where you have to have a specific amount of karma to get access so will always be private to shields.

services/reddit/user-karma.service.js
```diff
@@ -0,0 +1,91 @@
...
+    if (variant === 'link') {
+      karma = json.data.link_karma
+      label = `${user}'s karma (link)`
+    } else if (variant === 'comment') {
+      karma = json.data.comment_karma
+      label = `${user}'s karma (comment)`
+    } else {
+      const total = json.data.link_karma + json.data.comment_karma
+      karma = total
+      label = `${user}'s karma`
```
calebcartwright, line 82: The role for constructing the `label` value is a rendering related concern so it should reside in the `render` function. I'm also not sure how grammatically safely we can (or should) attempt to pluralize the `user` (i.e. avoiding `foos's karam`)

services/reddit/user-karma.tester.js
```diff
@@ -0,0 +1,80 @@
...
+  .get('/link/user_simulator.json')
+  .intercept(nock =>
+    nock('https://www.reddit.com/u')
+      .get('/user_simulator/about.json')
+      .reply(200, { kind: 't2', data: { link_karma: 20, comment_karma: 80 } })
+  )
+  .expectBadge({
+    label: "user_simulator's karma (link)",
+    message: '20',
+  })
```
calebcartwright, line 44: a common pattern many of a service classes use is to leverage a `transform` function that accepts the raw json response data and returns the corresponding value. one of the benefits of that approach is that it makes it simpler to test logic that's extracting/transforming with a simple unit test instead of having to use a full service test (which are primarily used as integration tests with the myriad upstream platforms/providers we integrate with).

don't feel like you have to change that here, but sharing for awareness in case you want to/for any potential future use cases

services/reddit/subreddit-subscribers.tester.js
calebcartwright: how does this differ from the live/non-mocked test immediately above? i feel like the 403 use case is already covered and am still confused by this mocked test

services/reddit/user-karma.service.js
JoeIzzard: The main reason I kept the construction of `label` in the handle function was to prevent needlessly running the same if statement multiple times for a single badge, but can move it into the render function if that's preferred. 

I see what you mean about pluralising the `user`, and again can remove it if that the preferred shields option, I just felt it would be much more natural being plural. For example with my username, `JoeIzzard's Karma` seems more natural, especially for a social badge, than `JoeIzzard Karma`. I think either way we are assuming something so it's a case of what fits with shields better from and standard style point?

services/reddit/subreddit-subscribers.tester.js
JoeIzzard: I see what you mean, I was referring to a different test sorry. The mocked test returns no subscriber information, testing the part of the code that throws an error if it is missing. The live test's should in theory always return that information, or cause another error such as being private.

services/reddit/user-karma.tester.js
JoeIzzard: Noted, I will definitely look at doing that in the future.

services/reddit/user-karma.service.js
calebcartwright: > The main reason I kept the construction of label in the handle function was to prevent needlessly running the same if statement multiple times for a single badge, but can move it into the render function if that's preferred

I suspected that was the case. However, the benefits of the separation of concerns outweigh any minor performance gains and will be more consistent with the Shields style/approach. Also, the this can be expressed succinctly within `render` by including `variant` is a param and only needing one conditional expression evaluation.

Something like
```js
const label = variant === 'combined' ? `${user}'s karma` : `${user}'s karma (${variant})`
```

calebcartwright: > I just felt it would be much more natural being plural

Is there any kind of symbol prefix/suffix that could be used? I'm only an occasional and brief browser of reddit so not terribly familiar with the platform. For example `r/User karma` or even something like what we do for the Twitter badges `@User karma` (understood the `@` has a specific meaning on that platform that may not apply to Reddit)

JoeIzzard: That's a good idea, I will go that route (Reddit uses /u/username for users). Will also move the label construction to render, and fix a hard coded username in the handle function that shouldn't be there either.

services/reddit/user-karma.service.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  static render({ variant, karma, user }) {
+    const label =
+      variant === 'combined'
+        ? `u/${user} karma`
+        : `u/${user} karma (${variant})`
+    return {
+      label: `${label}`,
```
calebcartwright, line 55: Can just use `label` directly here
```suggestion
      label,
```

services/reddit/user-karma.service.js
```diff
@@ -0,0 +1,96 @@
...
+    return { karma }
+  }
+
+  async handle({ variant, user }) {
+    const json = await this.fetch({ user })
+    const { karma } = this.transform({ json, variant })
+
+    return this.constructor.render({
+      variant,
+      karma: `${karma}`,
```
calebcartwright, line 92: Let's keep `karma` numeric, especially given the utilization of the `metric` helper function.
```suggestion
      karma,
```

services/reddit/subreddit-subscribers.tester.js
calebcartwright: Ah okay, think I have a better sense after taking a look at the service class. I think it would be better to use a different test name here that better reflects what the test is validating, perhaps:

```suggestion
t.create('subreddit-subscribers (no subscribers)')
```

I also feel like the error message that's being thrown when the `subscribers` key is missing in the response is a little confusing, especially given the API responds with a proper 404 response code for an unknown subreddit (and we use the exact same error message for both). 

I feel like the error message should be something more like `no subscribers found` as an `InvalidResponse` error instead of a `NotFound` as well

calebcartwright:
Thanks for making the changes @JoeIzzard! Will give this a final pass later today if none of the other maintainers beat me to it

services/reddit/subreddit-subscribers.tester.js
calebcartwright: I'm probably going to go ahead and merge, but this will need a follow up change, as at a minimum we don't want to keep two tests with the exact same name as we have right now

Issue #5188, opened 2020-06-08
Rename master branch to main


paulmelnikow:
There's a recent movement to replace "master" with a word that is descriptive while being absent the racist history of master & slave. This [isn't a new idea](https://www.vice.com/en_us/article/8x7akv/masterslave-terminology-was-removed-from-python-programming-language), however it's gaining momentum on Twitter given the groundswell of action in the United States and globally after the police killing of George Floyd in Minneapolis two weeks ago. I stand with the protestors. 🖤 

The language we use matters. Changing this for the Shields-managed repos is a minor inconvenience for us, and mostly a one-time thing, and seems well worth it.

I'd propose we rename the branch in Shields to `main`, update the open PRs to target that, and identify whatever else we have to update, like the deploy bot, deploy scripts, badges, etc.

There's a tool here we could use for the PRs: https://github.com/ethomson/retarget_prs

(I think we should also review the places where we default to master in the badge code, and think about how to handle that, however let's discuss that in #4755 and open new issues as necessary.)

calebcartwright:
Thanks for bringing this up! I'm a big +1 on making this change

chris48s:
There is a practical consideration here.

First of all, lets assume that the shields.io repo default branch would be renamed from master to main but our application will continue to assume that the default branch on most git(hub) repos is called master for now. That can move to another issue. For now, it is an accurate assumption.

We frequently use `badges/shields` either as an example or in a service test for a badge that makes the assumption your default badge is called master, or explicitly using the branch name master. If we no longer have a master branch, there's going to be a whole load of service tests that fail or examples that don't work at the point the change is made. In order to make this change, we would need to first replace all places where `badges/shields` is used in a service test or example with the assumption (implicit or explicit) that a branch which is called master exists.

To be clear, I'm not opposing this change, but there is a bunch of stuff that would need to happen first to do it in an orderly way.

Other (less cumbersome) things that would need updating in order to make this change (do add more if you think of them):
- Impact on CI/coverage setup
- Heroku deploy process/Makefile
- Any badges in use in READMEs
- Documentation references (esp links)

calebcartwright:
> but our application will continue to assume that the default branch on most git(hub) repos is called master for now. That can move to another issue. For now, it is an accurate assumption

Agreed. I don't think we can change that unless and until GitHub changes the name the platform selects for default branch.

[...]

PR #5190, opened 2020-06-08
Avoid using racially linked language

paulmelnikow:
Related to #5188. Curious if there are other places in Shields we should consider revising, though this is the only one that came to mind.

There's a badge URL that references [this Arch Linux package](https://aur.archlinux.org/packages/dns-zone-blacklist-git/) but I can't find a replacement. Also it's supposed to be an example of a package with no license, however the website lists it as being MIT licensed. I'm not sure what's up with that.


chris48s:
We do also have one occurrence of blacklist in a service test:

https://github.com/badges/shields/blob/13d75e0607334b700fdb4a533768cde341619a50/services/aur/aur.tester.js#L54-L56

We'd need a different example of a package in the Arch User Repos which does not declare a license to replace it. Needn't block the merge of this though.

paulmelnikow:
Yea, and weirdly if you look at the website for that package (see link in top post), it shows MIT license. I poked around for a bit but everything had a license listed on the website.

PR #5213, opened 2020-06-14
update [codeclimate] service tests/example

chris48s:
CodeClimate coverage tests and example were both referring to repos that don't use codeclimate for coverage any more.


PyvesB:
Code Climate tests breaking: classic! 😄

PR #4781, opened 2020-03-15
Fix [Twitch] authorization

PyvesB:
Fixes #4778.

PR #4729, opened 2020-03-04
Check request origin before sending credentials, affects [AzureDevops Bintray Bitbucket Drone GitHub Jenkins Jira Nexus NPM Sonar Symfony TeamCity Wheelmap]

PyvesB:
This pull request adds a new `authorizedOrigins` configuration key for a number of services which have the ability to send credentials to user-defined endpoints. More information about the impact of this will be published shortly.

This pull request was a joint effort from @calebcartwright, @chris48s, @paulmelnikow  and myself.

PR #5217, opened 2020-06-16
Fix [Twitch] Client-ID header

### LINKED ISSUE #5214: Twitch badges failing

:beetle: **Description**

Looks like our twitch badges are failing in production and in the service tests. I grabbed the `TWITCH_CLIENT_ID` and `TWITCH_CLIENT_SECRET` from heroku and ran the tests locally. The error is:

```json
{"error":"Unauthorized","status":401,"message":"Client ID and OAuth token do not match"}
```

It looks like CircleCI is using different tokens than production, but I guess the failure is the same.

:link: **Link to the badge**

examples:
 - http://img.shields.io/twitch/status/andyonthewings
 - http://img.shields.io/twitch/status/noopkat

:bulb: **Possible Solution**

Clearly we need new tokens. @PyvesB I think the tokens are linked to your account, but also if these need rotating occasionally, maybe we need a better solution to this?

PyvesB:
Fixes #5214.

We broke Twitch badges back in #4729. I mostly fixed things in #4781, but I missed out on bit with led to the `Client-ID` not actually being set. The impact of this mistake only became visible recently for reasons outlined in #5214.

The improved spec test should prevent any regressions in the future.

### DIFF
--- a/core/base-service/base.js
+++ b/core/base-service/base.js
@@ -114,7 +114,7 @@ class BaseService {
    * credentials to the request. For example:
    * - `{ options: { auth: this.authHelper.basicAuth } }`
    * - `{ options: { headers: this.authHelper.bearerAuthHeader } }`
-   * - `{ options: { qs: { token: this.authHelper.pass } } }`
+   * - `{ options: { qs: { token: this.authHelper._pass } } }`
    *
    * @abstract
    * @type {module:core/base-service/base~Auth}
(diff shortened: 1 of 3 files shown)


core/base-service/base.js
```diff
@@ -114,7 +114,7 @@ class BaseService {
    * credentials to the request. For example:
    * - `{ options: { auth: this.authHelper.basicAuth } }`
    * - `{ options: { headers: this.authHelper.bearerAuthHeader } }`
-   * - `{ options: { qs: { token: this.authHelper.pass } } }`
```
PyvesB, line 117: Unrelated to Twitch, but a similar piece of code (or rather documentation) that was missed in #4729.

PR #5232, opened 2020-06-20
support custom manifest path for [GitHubManifest] and fix service tests

### LINKED ISSUE #5230: fix github manifest service tests

Probably 15th June

:link: **CircleCI link**

https://app.circleci.com/pipelines/github/badges/shields/1804/workflows/3f987aad-b393-40f4-b2aa-9b2d13ae18c2/jobs/107672

:beetle: **Stack trace**

```
ValidationError: message mismatch: "value" with value "repo not found, branch not found, or manifest.json missing" fails to match the required pattern: /^v\d+(\.\d+)?(\.\d+)?$/
```

:bulb: **Possible solution**

Tests are failing because the example repo we're using no longer has a `manifest.json` in the root (its moved to a subdir). We need to find another example repo that has a `manifest.json` in the root.

calebcartwright:
Resolves #5230

### DIFF
--- a/services/github/github-manifest.service.js
+++ b/services/github/github-manifest.service.js
@@ -15,6 +15,10 @@ const schema = Joi.object({
   version: individualValueSchema,
 }).required()
 
+const queryParamSchema = Joi.object({
+  file: Joi.string().regex(/.*manifest\.json$/),
+})
+
 const flexibleSchema = Joi.object().required()
 
 class GithubManifestVersion extends ConditionalGithubAuthV3Service {
@@ -26,6 +30,7 @@ class GithubManifestVersion extends ConditionalGithubAuthV3Service {
     return {
       base: 'github/manifest-json/v',
       pattern: ':user/:repo/:branch*',
+      queryParamSchema,
     }
   }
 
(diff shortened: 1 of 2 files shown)


services/github/github-manifest.service.js
```diff
@@ -63,13 +95,13 @@ class GithubManifestVersion extends ConditionalGithubAuthV3Service {
     })
   }
 
-  async handle({ user, repo, branch }) {
+  async handle({ user, repo, branch }, { file }) {
```
chris48s, line 98: Thanks for picking this up. Looks good, but a minor bikeshed on the param name: The other related service where we've done something similar is the github package.json service. On that one, we called the param `filename`. Lets use the same name for this param too.

services/github/github-manifest.tester.js
```diff
@@ -11,25 +11,36 @@ const t = (module.exports = new ServiceTester({
...
 
-t.create('Manifest object')
-  .get('/background/RedSparr0w/IndieGala-Helper.json')
+t.create('Manifest object (path)')
+  .get(
+    '/background/RedSparr0w/IndieGala-Helper.json?file=extension/manifest.json'
+  )
   .expectBadge({ label: 'manifest', message: 'invalid key value' })
 
 t.create('Manifest invalid json response')
```
chris48s, line 46: do you think its worth adding a test case here for the situation where the repo does exist and the filename is specified but the filename doesn't exist?

services/github/github-manifest.service.js
calebcartwright: I struggled with this a bit so happy to do a bit of bikeshedding. We're not consistently consistent with these (for example `nyc` has a `config` query param, snyk has a `manifestFilePath`, etc.), and `filename`, at least outside a Shields context, is usually the name of the file itself, not the entire file path. Plus for this service the name of the file must be `manifest.json`.

I don't feel strongly about this, and the majority (or at least plurality) of instances are using `filename`. However, do want to call it out as another area where I feel the only thing that's consistent is inconsistency :smile:

chris48s: The package.json/go.mod ones are harder to change now. Perhaps `path` would have been a better choice there too.

As I mentioned in #5234 (timing), nyc is a bit odd anyway as it should probably move under `/github` like all the other "get a file from github and extract a value from it" badges (same with osslifecycle). That would be an opportunity to change the param to `filename` when we move it and translate it in the redirect.

How about we adopt `filename`, bring nyc into line when we move it, document in https://github.com/badges/shields/blob/master/doc/badge-urls.md that we use `filename` and accept that snyk is different? At least then we are using the same param name within the github service family and if we want to add this feature to other "get a file from github and extract a value from it" badges which currently use a hard-coded location (e.g: pipenv, lerna, etc) we've established a pattern.

calebcartwright: > How about we adopt filename

Agreed, was planning on making that change anyway, especially since using `file` like I originally had it would have only increased inconsistency :laughing: 

> document in https://github.com/badges/shields/blob/master/doc/badge-urls.md that we use filename and accept that snyk is different

:+1: I think adding some guidance/documentation around our goals for consistency would be great in general. There's some things that are consistent without exception and strictly enforced (even down to the order of methods in service classes), while that are a lot fuzzier, and inconsistent consistency is a bit of a pet peeve of mine :smile: 

Even something simple like the below example would be pretty helpful in my opinion:

__For X, ensure you are consistent with (Everything else//all of Shields | Other badges for the same platform/service | other badges in the same category | ... )__

> The package.json/go.mod ones are harder to change now

That's probably true, and again want to stress that I don't really care about the name. Will note though that should we ever need to change that query params (as opposed to route params) give us more flexibility to potentially change things without strictly having to use a redirector (admittedly with other tradeoffs). For example:

```js
  filename: Joi.string(),
  // new param name
  filepath: Joi.string(),
})

...

async handle(routeParams, { filepath, filename }) {
    file: filepath || filename || 'some deault value',
}
...

Issue #5215, opened 2020-06-14
What do we do when(if?) major tools stop calling the default branch master?


chris48s:
⛈️ There's a storm brewing... time for a GitHub essay 📚

For the last ~15 years the standard name for a git default branch has been master. This convention was established by git and all major git-based ecosystems have inherited the decision. Shields inherits this norm: Branch is an optional param. We expect your default branch to be called master. If you want a badge for a non-default branch, or you've called your default branch something else, we require you to specify it. This assumption applies to source control services (github, bitbucket, gitlab, etc) and source control-adjacent services (CI, coverage tools, etc). There are currently a number of discussions going on in various git communities/businesses about changing the standard name for a default branch. For example:

- git mailing list: https://lore.kernel.org/git/CAOAHyQwyXC1Z3v7BZAC+Bq6JBaM7FvBenA-1fcqeDV==apdWDg@mail.gmail.com/
- github (allegedly): https://www.zdnet.com/article/github-to-replace-master-with-alternative-term-to-avoid-slavery-references/
- gitlab: https://gitlab.com/gitlab-org/gitlab/-/issues/220906
- git for windows: https://github.com/git-for-windows/git/issues/2674

It is worth noting that in all cases platforms are looking at changing the default for newly created repos. None are looking at retrospectively forcing a change on repos that already exist. Given GitHub and GitLab are working on it I wouldn't be surprised to see bitbucket follow suit, although that is entirely unsubstantiated speculation.

Speaking of unsubstantiated speculation, I figure there's probably 3 ways things might play out here..

## 🗞️  Read all about it - nothing happened [highly unlikely]

Anticlimax. None of this goes anywhere. All common git source control tools/platforms continue to call their default branch `master`.

## 🌈  Why can't we all just get along? [also quite unlikely]

The entire git community unites and reaches a consensus. All git clients and platforms unanimously decide on a single blessed solution. The gods of git decree "Thy default branch shall henceforth be called `primary`". The villagers rejoice.

## 💥  The great split [most likely]

Turns out naming things is hard. Who knew? 🤔 Nobody can agree on anything. The ecosystem splinters into a fragmented mess where everyone's default branch is variously called default, main, trunk, develop, release, base, primary etc. Defaults vary by platform, convention varies within platforms. Default vs main is the new tabs vs spaces. Master traditionalists rage against the Trunk reformers. You've met developers before, right?

---

So.. what are we going to do? Well first, we wait..

[...]

PR #5234, opened 2020-06-20
auto-detect default branch on [github nycrc osslifecycle]

chris48s:
..with this one wierd trick :sparkles: 

Shout out to @andersk for this [excellent suggestion](https://github.com/badges/shields/issues/5215#issuecomment-646375440) which allows us to solve this problem in a large number of cases with one very small code change and no impact on performance :smile: 

Arguably this makes a very slight backwards-incompatible change in that if your repo has a branch which **is** called master but which **isn't** the default, right now https://img.shields.io/osslifecycle/user/repo is looking at your master (by name) branch but this will change it to look at your default branch (whatever that is called). I'm not particularly concerned about this. I think it unlikely many users are consciously relying on that behaviour and we can regard this as a fix which brings us into line with how it should have worked in the first place/what most users will expect anyway.

As a side note, we should probably move nycrc and osslifecycle under the github service family and set up redirects, but that's another issue.

I've only addressed 

* github
* nycrc
* osslifecycle

in this PR. The other services mentioned in https://github.com/badges/shields/issues/5215#issuecomment-646293697 need a different solution (WIP), although solving this for all the affected services under the github family is not to be sniffed at.

I think this should also give us a fix for #4114 ?

There will be some failing tests due to #5230 . Once #5232 is merged, I will rebase to get the tests passing.


calebcartwright:
This is a fantastically simple solution, love it!

I suppose another rather fringe case: a repo that has both `master` and `HEAD` branches, with `master` as the current default (which i think GitHub will technically let you do) :eyes: 

IMO this is still the right approach, maybe we can just enumerate the handful of edge cases like this in the issue with workarounds should anyone run into these in "real life". Just the tradeoffs we're having to make to avoid doubling the badge response time and doubling our GH token quota usage that the alternatives would've required

chris48s:
> I suppose another rather fringe case: a repo that has both `master` and `HEAD` branches, with `master` as the current default (which i think GitHub will technically let you do)

You can't have a branch which is explicitly called `HEAD` in git:

```sh
$ git init
Initialised empty Git repository in /home/chris/temp/asdf/.git/
$ git checkout -b HEAD
fatal: 'HEAD' is not a valid branch name.
```

You also can't make a branch called `HEAD~1` or `HEAD@{1}` etc for the same reason.

..or do you mean something different?

calebcartwright:
GitHub (not git) will apparently let you do it from the web portal which I found surprising. To your point though, doing so would be spectacularly unhelpful given the client side implications and thus not real/valid badge use case. 

I would still maintain however that as part of this change it would be a good idea to create (and pin) and issue that succinctly summarizes the change we're making, and any known edge cases and workarounds. That way if anyone _does_ encounter an edge case they can easily find the workaround, and/or report any cases we haven't seen yet (as opposed to having to go through lengthy issues/PRs)

chris48s:
So you can :open_mouth: Every day's a school day :mortar_board: 

Screenshot at 2020-06-22 19-46-09

True story: I nearly just tried doing this on a repo I actually care about, then I realised that I might be about to make it impossible to check the repo out so I switched and did it on a test repo :rofl:  I find it incredibly surprising that github lets you do that. I might have a quick play and see what nonsense ensues if I actually try and do something with this.

Side note: I'm also working on another (less nice) PR for bitbucket, gitlab and travis ci which I'm aiming to submit tonight or tomorrow.

chris48s:
I had a quick play with this. You're totally right that if you do create a branch called `HEAD`, github's UI/API assumes you mean HEAD (the branch) not HEAD (the ref) so if you make a branch called HEAD and its not the default then this would result in unexpected behaviour. If you try and check it out locally (at least with the version of git I have), you basically can't interact with that branch at all though so it is quite pointless. I guess you could have a branch called HEAD on a repo that you only ever interact with via the github UI, or (more likely) a branch called HEAD that you created, decided was a terrible idea but then didn't delete.
I think we both still think its the right approach in the grand scheme of things, but dammit that is an annoying edge case.

calebcartwright:
> I think we both still think its the right approach in the grand scheme of things, but dammit that is an annoying edge case.

Absolutely, on both counts! IMHO this is much better than the alternative, and I'd rather have unexpected behavior for the hypothetical (and presumably nonexistent) HEAD branch case instead of impacting every single valid/standard badge use case. 

Have you had a chance to do any additional testing with this? The code changes and approach LGTM, though I think it'd be worth manually validating the behavior with a sample repo and contrived data if no one's done so yet

calebcartwright:
> As a side note, we should probably move nycrc and osslifecycle under the github service family and set up redirects, but that's another issue

Agreed. I keep thinking I've convinced myself one way or the other, only to then change my mind

chris48s:
> Have you had a chance to do any additional testing with this?

Yeah good question. I've tried making some ad-hoc requests on some repos where the default branch is/isn't master. It does appear to work as intended, but its not a super well documented feature.
In terms of automated testing, its essentially a feature of an external service so I haven't added any additional tests - I'm just relying on the existing tests to verify the change doesn't break anything really.

If you've got any other thoughts on stuff I should look at, I'm all ears..

calebcartwright:
>  I've tried making some ad-hoc requests on some repos where the default branch is/isn't master. It does appear to work as intended, but its not a super well documented feature.

That's all I had in mind. Was going to do the same but figured I'd check first.

> In terms of automated testing, its essentially a feature of an external service so I haven't added any additional tests - I'm just relying on the existing tests to verify the change doesn't break anything really.

> If you've got any other thoughts on stuff I should look at, I'm all ears..

Nah I don't think we need anything else, sounds like it's working as expected :+1:

PR #5236, opened 2020-06-23
Adopt Code of Conduct

### LINKED ISSUE #2671: Code of conduct
We've been fairly lucky in this project especially given its broad reach to have maintained a high degree of cordiality and civility. This is much thanks to @espadrine who has a great deal of equanimity, and we continue to have a team of very friendly maintainers and mostly civil contributors.

That said, it's a best practice to provide a [code of conduct](https://opensource.guide/code-of-conduct/). Projects without a code of conduct may be perceived as less safe / friendly to new contributors.

While I know that's not actually the case for _this_ project, I want to make sure no one gets that impression from this omission. I also think it's a good idea to have something written down that we're ready to enforce when the time comes.

[Contributor Covenant](https://www.contributor-covenant.org/) is a well respected template.

calebcartwright:
Closes #2671 :heart: 

We've had extensive discussion in #2671, seemingly with a convergence on adopting Contributor Covenant. There's a minor todo regarding the contact endpoints/which team(s) will be involved, but my hope is we can resolve that quickly, or adopt this-as is with an existing team (perhaps Core or Operations if we want a smaller group) taking the role until we can finalize the team/members that will own governance/enforcement.

Also, prettier made a few minor changes (formatting, ordered list items, etc.) so added that is a separate commit so folks could see the tiny differences we'll have.


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,130 @@
...
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+[conduct@shields.io][contact-method].
```
calebcartwright, line 63: AFAIK this mailbox doesn't actually exist, so we'd either need to create it or plug in one of the existing ones.

paulmelnikow:
I've set up conduct@shields.io, which forwards to me and Chris, along with paul@shields.io and chris@shields.io which go to one of us, and can be used for conduct complaints about the other. Probably we should add those to the text.

I'd also prefer if we obscure the addresses in some way to reduce the chance of getting spam (e.g. "conduct at shields.io").

calebcartwright:
> I'd also prefer if we obscure the addresses in some way to reduce the chance of getting spam (e.g. "conduct at shields.io").

Agreed, have a thought on how to do so, will update later today

calebcartwright:
:laughing: let me know if you all think this is too much, but i thought we could add our own badge flavor here to obscure the contact emails

paulmelnikow:
I put off doing this so long I forgot that it's Caleb and not Chris who agreed to handle this 🤣 

Chris has pointed this out. So I've updated conduct@ to deliver to me and Caleb, and added caleb@ to deliver to Caleb.

> 😆 let me know if you all think this is too much, but i thought we could add our own badge flavor here to obscure the contact emails
> 
> 
> 
> 

This works for me!

PyvesB:
> 
> 
> 😆 let me know if you all think this is too much, but i thought we could add our own badge flavor here to obscure the contact emails
> 
> 
> 
> 

I'm guessing you could even put `@shields.io` with the obfuscation on the right handside: I wouldn't expect any bot to be clever enough to piece the SVG back together!

calebcartwright:
> I'm guessing you could even put @shields.io with the obfuscation on the right hand-side: I wouldn't expect any bot to be clever enough to piece the SVG back together!

Good point, updated accordingly!

paulmelnikow (approved):

PyvesB (approved):

chris48s (approved):

PR #5237, opened 2020-06-23
[crates] recent_downloads can be null

obfusk:
The API apparently returns `null` for that key when the crate is so new that it doesn't have any downloads yet.


calebcartwright:
Thanks for the PR @obfusk! Did you encounter a badge with an `invalid response data` message for a specific crate?

services/crates/crates-base.js
```diff
@@ -9,7 +9,7 @@ const keywords = ['Rust']
 const crateSchema = Joi.object({
   crate: Joi.object({
     downloads: nonNegativeInteger,
-    recent_downloads: nonNegativeInteger,
+    recent_downloads: nonNegativeInteger.allow(null),
```
paulmelnikow, line 12: Instead of putting the default value in the transform, can you use `.default(0)` here?

obfusk: That only works for `undefined`, not for `null`; I tried that :(

obfusk:
> Thanks for the PR @obfusk! Did you encounter a badge with an `invalid response data` message for a specific crate?

Yes. But it's been downloaded by now. I think you can only run into this with a create you've only just created.

services/crates/crates-base.js
calebcartwright: > That only works for undefined, not for null

I didn't realize that was the case either but the [docs](https://hapi.dev/module/joi/api/?v=17.1.1#anydefaultvalue) seem to support that

services/crates/crates-downloads.service.js
```diff
@@ -87,7 +87,7 @@ module.exports = class CratesDownloads extends BaseCratesService {
       case 'dv':
         return json.crate ? json.versions[0].downloads : json.version.downloads
       case 'dr':
-        return json.crate.recent_downloads
+        return json.crate.recent_downloads || 0
```
calebcartwright, line 90: Could you also add a test case for this scenario, and/or a comment on the schema just to keep some in-code context on this scenario? 

Either a unit test targeting the transform function (like the one for the [crates version badges](https://github.com/badges/shields/blob/master/services/crates/crates-version.spec.js)) or a service test with a mocked response from the crates.io API (https://github.com/badges/shields/blob/master/doc/service-tests.md#5-mocking-responses)

obfusk:
I added a test case (and fixed it :sweat_smile:). No idea why the `main` test is failing now.

calebcartwright:
Looks like the formatting check is failing. You can fix this by running `npm run prettier` locally

More info here: https://github.com/badges/shields/blob/master/CONTRIBUTING.md#prettier

calebcartwright:
Looks like ranger bot is having trouble with this one. @obfusk can you do a rebase so we can get this merged? Thanks!

chris48s:
I think the issue here is the 'maintainers can edit' checkbox wasn't checked, which means if this branch ends up more than one commit behind master we can't update it, neither can the bot, and the branch protection rule won't let us merge it. @obfusk  Try pressing "update branch". That will update it without dismissing the approving review. If the bot doesn't merge it after that, I'll push these commits to another branch and we'll merge from there. Sorry this has been a bit of a pain.

PR #5241, opened 2020-06-24
auto-detect default branch name in [bitbucketpipelines gitlab travisphpversion]

chris48s:
refs #5215

In some cases, we can use the handy `HEAD` trick to query whatever the default branch is with one HTTP call. Sometimes we can't, so we have to use another method to find out the default branch using another network request. One option is we can make an API call to github/bitbucket/gitlab. The other is we can shell out to `git` and call `git ls-remote --symref https://github.com/badges/shields.git HEAD`. Inevitably there are some tradeoffs to consider. Not all of them necessarily come into play in this PR, but I'm going to list them all anyway as this is something we will have to think about more in future.

In general, it is best to make an API call if we can because:

* An HTTP API gives us nice error codes (e.g: 404 = repo not found, 403 = private repo, etc)
* An HTTP API call can be easily mocked under test with nock. The request made by `git` is harder to mock
* If we need to support private repos/self-hosting, the available credentials might influence what we can use. For example, `git` only allows HTTPS basic auth or SSH key, but if we only hold an API token for self-hosting users, we can't authenticate with either of those methods for private repos.
* A JSON response can be more reliably parsed than the text output of `git`
* This probably varies a bit, but in general it seems to be faster to use a JSON API than `git`

There is one advantage of using git directly though:

* If we use `git`, we don't need to worry about an API rate limit. This makes it more suitable if we don't have API credentials, or if we would be likely to exceed a rate limit.

So.. the upshot of that is I've chosen:

* Bitbucket pipelines: Call the BitBucket API
* Gitlab Pipelines: Shell out to git
* Travis PHP Version: Call the Github API

In slightly more detail:

### BitBucket pipelines

BitBucket's API allows us to pass a `ref_type`

> The following reference types are supported:
> * branch
> * named_branch
> * bookmark
> * tag

As far as I can tell, none of them allow us to pass `HEAD` as a ref, so this now makes two API calls - fairly straightforward implementation. Making two API calls does impact performance but both the calls are pretty quick.

### Gitlab pipelines

With GitLab we don't have an API token (or a pool of them), which is why pipelines is the only gitlab badge we support. This somewhat ties our hands. I've implemented a general helper which uses simple-git to call `ls-remote` (although gitlab pipelines is the only service which uses it). It requires git to be installed (I think the heroku buildpack does have this. We'll find out when we get this on staging). Again, there is some performance impact as we're now making two network calls but its far from our worst performing badge. If/when we implement a proper suite of gitlab badges using their API it would probably be good to review this approach.

### Travis PHP Version

Travis PHP version calls the travis-ci **.org** API, which I think means the source repo is hosted on GitHub, so I've used a call to the GitHub API here (AFAIK only travis.com supports bitbucket so this is something we would need to bear in mind if we generalised). In the abstract, adding a call to the Github API isn't the end of the world - as APIs go its pretty fast. Unfortunately, the Travis PHP Version badge already makes two API calls and this brings us up to three, which is a bit rubbish.
To offset this, I've done a bit of acrobatics to fire off two of the requests at the same time to keep the performance within reasonable bounds. I think we could probably look at caching the PHP versions in Redis instead of fetching it on every request, but this solution keeps the performance in check for now.

### DIFF
--- a/services/bitbucket/bitbucket-pipelines.service.js
+++ b/services/bitbucket/bitbucket-pipelines.service.js
@@ -25,6 +25,12 @@ const bitbucketPipelinesSchema = Joi.object({
     .required(),
 }).required()
 
+const bitbucketRepoSchema = Joi.object({
+  mainbranch: Joi.object({
+    name: Joi.string().required(),
+  }).required(),
+}).required()
+
 module.exports = class BitbucketPipelines extends BaseJsonService {
   static get category() {
     return 'build'
@@ -70,9 +76,22 @@ module.exports = class BitbucketPipelines extends BaseJsonService {
   }
 
   async fetch({ user, repo, branch }) {
-    const url = `https://api.bitbucket.org/2.0/repositories/${user}/${repo}/pipelines/`
+    let branchName
+    if (branch == null) {
+      const repoUrl = `https://api.bitbucket.org/2.0/repositories/${user}/${repo}/`
+      const repoResp = await this._requestJson({
+        url: repoUrl,
+        schema: bitbucketRepoSchema,
+        errorMessages: { 403: 'private repo' },
+      })
+      branchName = repoResp.mainbranch.name
+    } else {
+      branchName = branch
+    }
+
+    const pipelineUrl = `https://api.bitbucket.org/2.0/repositories/${user}/${repo}/pipelines/`
     return this._requestJson({
-      url,
+      url: pipelineUrl,
       schema: bitbucketPipelinesSchema,
       options: {
         qs: {
@@ -81,7 +100,7 @@ module.exports = class BitbucketPipelines extends BaseJsonService {
           pagelen: 2,
           sort: '-created_on',
           'target.ref_type': 'BRANCH',
-          'target.ref_name': branch,
+          'target.ref_name': branchName,
         },
       },
       errorMessages: { 403: 'private repo' },
@@ -99,7 +118,7 @@ module.exports = class BitbucketPipelines extends BaseJsonService {
   }
 
   async handle({ user, repo, branch }) {
-    const data = await this.fetch({ user, repo, branch: branch || 'master' })
+    const data = await this.fetch({ user, repo, branch })
     return this.constructor.render({ status: this.constructor.transform(data) })
   }
 }
--- a/services/git.js
+++ b/services/git.js
@@ -1,0 +1,65 @@
+'use strict'
+
+const sg = require('simple-git')()
+const { InvalidResponse, NotFound } = require('.')
+
+async function getDefaultBranch({
+  baseUrl,
+  user,
+  repo,
+  username = '',
+  password = '',
+  simpleGit = sg,
+}) {
+  /*
+  Note on authentication: if we pass the function a valid
+  username and password, they will be used:
+  https://username:password@github.com/badges/shields.git
+  This could be useful where we've got
+  BITBUCKET_USER / BITBUCKET_PASS set, for example.
+
+  If username and password aren't specified,
+  we will intentionally pass blank credentials:
+  https://:@github.com/badges/shields.git
+  This might seem like a bad idea at first, but if we pass no credentials at all
+  `git ls-remote --symref https://github.com/badges/shields.git HEAD`
+  will work, but
+  `git ls-remote --symref https://github.com/foobar/does-not-exist.git HEAD`
+  will leave us sitting at an interactive prompt
+  `Username for 'https://github.com':`
+  waiting for input that is never going to arrive.
+
+  If we pass a set of intentionally blank credentials,
+  `git ls-remote --symref https://:@github.com/badges/shields.git HEAD`
+  still works fine for a public repo, but
+  `git ls-remote --symref https://:@github.com/foobar/does-not-exist.git HEAD`
+  will instantly fail on the assumption that the repo either doesn't exist,
+  or it does but we don't have access to it. Either way simple-git
+  will throw a `GitError: remote: HTTP Basic: Access denied`
+  which we can catch and re-throw as NotFound rather than sitting at a
+  login prompt forever.
+
+  So from outside, omitting the username and password params allows us to
+  query public repos only, failing if the repo is not found/private,
+  which is what we'd expect.
+  */
+  const url = `${baseUrl
+    .replace(/\/$/, '')
+    .replace('://', `://${username}:${password}@`)}/${user}/${repo}`
+
+  let console_text
+  try {
+    console_text = await simpleGit.listRemote(['--symref', url, 'HEAD'])
+  } catch (e) {
+    // GitError
+    throw new NotFound({ prettyMessage: 'repo not found' })
+  }
+
+  const match = console_text.match(/^ref: refs\/heads\/([^\s]+)[\s]+HEAD/)
+  if (match) {
+    return match[1]
+  }
+  throw new InvalidResponse({ prettyMessage: 'invalid response data' })
+}
+
+module.exports = { getDefaultBranch }
--- a/services/git.js
+++ b/services/git.js
@@ -1,58 +1,34 @@
 'use strict'
 
 const sg = require('simple-git')()
+sg.env('GIT_TERMINAL_PROMPT', 0)
 const { InvalidResponse, NotFound } = require('.')
 
 async function getDefaultBranch({
   baseUrl,
   user,
   repo,
-  username = '',
-  password = '',
+  username,
+  password,
   simpleGit = sg,
 }) {
-  /*
-  Note on authentication: if we pass the function a valid
-  username and password, they will be used:
-  https://username:password@github.com/badges/shields.git
-  This could be useful where we've got
-  BITBUCKET_USER / BITBUCKET_PASS set, for example.
-
-  If username and password aren't specified,
-  we will intentionally pass blank credentials:
-  https://:@github.com/badges/shields.git
-  This might seem like a bad idea at first, but if we pass no credentials at all
-  `git ls-remote --symref https://github.com/badges/shields.git HEAD`
-  will work, but
-  `git ls-remote --symref https://github.com/foobar/does-not-exist.git HEAD`
-  will leave us sitting at an interactive prompt
-  `Username for 'https://github.com':`
-  waiting for input that is never going to arrive.
-
-  If we pass a set of intentionally blank credentials,
-  `git ls-remote --symref https://:@github.com/badges/shields.git HEAD`
-  still works fine for a public repo, but
-  `git ls-remote --symref https://:@github.com/foobar/does-not-exist.git HEAD`
-  will instantly fail on the assumption that the repo either doesn't exist,
-  or it does but we don't have access to it. Either way simple-git
-  will throw a `GitError: remote: HTTP Basic: Access denied`
-  which we can catch and re-throw as NotFound rather than sitting at a
-  login prompt forever.
-
-  So from outside, omitting the username and password params allows us to
-  query public repos only, failing if the repo is not found/private,
-  which is what we'd expect.
-  */
-  const url = `${baseUrl
-    .replace(/\/$/, '')
-    .replace('://', `://${username}:${password}@`)}/${user}/${repo}`
+  const url = new URL(baseUrl)
+  url.pathname = `${user}/${repo}`
 
   let console_text
   try {
-    console_text = await simpleGit.listRemote(['--symref', url, 'HEAD'])
+    const args = ['--symref', url.toString(), 'HEAD']
+    if (username != null && password != null) {
+      const authStr = Buffer.from(`${username}:${password}`).toString('base64')
+      args.push('-c')
+      args.push(`http.extraheader="Authorization: Basic ${authStr}"`)
+    }
+    console_text = await simpleGit.listRemote(args)
   } catch (e) {
-    // GitError
-    throw new NotFound({ prettyMessage: 'repo not found' })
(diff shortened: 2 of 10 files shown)


calebcartwright:
Thanks for putting this together! I'm hoping to carve out some time to dig through this over the weekend.

> Again, there is some performance impact as we're now making two network calls but its far from our worst performing badge

Besides the extra networking call, any concerns and/or config we may need to account for process spawning associated with this? (which I presume is happening somewhere)

chris48s:
> any concerns and/or config we may need to account for process spawning associated with this?

Hmm. Interesting question. I hadn't really given much thought to the gitlab badge, which invokes `git` from JS. I guess it is going to spawn an OS process for each badge call which could potentially stack up. I'm not really sure if that's something we need to worry about or not.. do you have any more detailed thoughts on it? Its pretty standard practice in most languages that every request spawns its own process, but then I don't reckon we'd be serving the level of traffic we do off just 4 servers in most languages.

For the ones that just make 2 API calls, its javascript, so its single thread and promises. Although we try and avoid them, we've already got a few services that make more than one API call (or can do) e.g: matrix, nuget, f-droid, codeclimate. Its basically the same setup here.

If we're not convinced on the gitlab approach but we're happy with the other two, I can split this up into multiple smaller PRs if needed.

Btw, I totally forgot to check this on a review app after I initially submitted the PR, but I've just set up https://shields-staging-pr-5241.herokuapp.com/ and the gitlab badge does work, which is useful.

calebcartwright:
> I'm not really sure if that's something we need to worry about or not.. do you have any more detailed thoughts on it?

I don't have anything concrete at the moment, but I do think that the process spawning (associated memory implications) is a factor we'll need to take into account (especially given our Heroku strategy of horizontally scaled small dynos).

We're not getting a ton of GitLab traffic today because as you noted we only support the one badge today. As such it's unlikely to have any noticeable impact. However, I imagine it could quickly become significant with higher volume badges.

calebcartwright:
Also, I'm still having my own internal debate to try to make up my mind on whether I think Shields should continue trying to do auto-magic default branch detection for services that require Shields to send them a branch param, but don't provide their own auto-magic solution like GitHub does with HEAD.

On one hand I understand the convenience for users to not have to specify a branch when setting up their badge. OTOH, setting up a given badge is really a one time action for end users (it's not like they have to take action and keep re-specifying their branch every the time the badge is loaded/viewed), but Shields has to do this auto-magic work on every (non-cached) badge request. 

For me it's a matter of tradeoffs and weighing the convenience for a user to not have to include a `/my-branch` when they add their badge to a readme/etc. against the increased workload for the badge servers (extra network calls, process spawning, etc.) and associated code maintenance.

services/bitbucket/bitbucket-pipelines.service.js
```diff
@@ -70,9 +76,22 @@ module.exports = class BitbucketPipelines extends BaseJsonService {
...
   async fetch({ user, repo, branch }) {
-    const url = `https://api.bitbucket.org/2.0/repositories/${user}/${repo}/pipelines/`
+    let branchName
+    if (branch == null) {
+      const repoUrl = `https://api.bitbucket.org/2.0/repositories/${user}/${repo}/`
+      const repoResp = await this._requestJson({
+        url: repoUrl,
+        schema: bitbucketRepoSchema,
+        errorMessages: { 403: 'private repo' },
+      })
```
calebcartwright, line 86: Completely orthogonal to this PR, but just hit me we support private repos with other BB badges in the self-hosted Shields scenario, and may be worth doing so here at some point in the future (though admittedly I don't think anyone's asked for it)

services/git.js
```diff
@@ -0,0 +1,65 @@
...
+  password = '',
+  simpleGit = sg,
+}) {
+  /*
+  Note on authentication: if we pass the function a valid
+  username and password, they will be used:
+  https://username:password@github.com/badges/shields.git
+  This could be useful where we've got
+  BITBUCKET_USER / BITBUCKET_PASS set, for example.
+
```
calebcartwright, line 20: Would it be feasible to use the auth header instead? I know the userinfo approach is still used, but feel it's generally fallen out of favor (https://tools.ietf.org/html/rfc3986#section-3.2.1)

not familiar with `simple-git` so not sure if/how, but something that results in
```
git -c http.extraheader="Authorization: Basic base64(username:password)"
```

services/git.js
```diff
@@ -0,0 +1,65 @@
...
+  const url = `${baseUrl
+    .replace(/\/$/, '')
+    .replace('://', `://${username}:${password}@`)}/${user}/${repo}`
+
+  let console_text
+  try {
+    console_text = await simpleGit.listRemote(['--symref', url, 'HEAD'])
+  } catch (e) {
+    // GitError
+    throw new NotFound({ prettyMessage: 'repo not found' })
```
calebcartwright, line 55: Will this always be strictly true? Would it be worth using a slightly different error message in this case to make it easier to tell from the badge message if the error is occurring in here vs. individual badge services that would leverage this helper and the output in the subsequent API calls?

services/bitbucket/bitbucket-pipelines.service.js
chris48s: Yes there are several inconsistencies within the BB service family. For example the Pull Requests badge works with both BB Cloud and BB Server, but Issues and Pipelines only work with BB Cloud. There's also a mix of V1 and V2 API calls. I don't think anything in we are doing in this PR prevents us from sorting these out at another time.

services/git.js
chris48s: Yeah good point. There are probably some other errors that could happen here. I've made the error case a bit more specific. The condition I've added here for not found  would also be hit if we were to pass bad credentials, but at the moment we have zero badges that actually pass this function credentials.

services/git.js
chris48s: We can pass any flags you want, so instead of

```js
console_text = await simpleGit.listRemote(['--symref', url, 'HEAD'])
```

we can call

```js
console_text = await simpleGit.listRemote(['--symref', url, 'HEAD', '-c', 'http.extraheader="Authorization: Basic Og=="'])
```

The issue with this is if it fails, it prompts, which was the exact thing I was trying to avoid. Passing the creds in the URL does prevent that prompting. However, this comment has motivated me to revisit the idea of whether there is a better way to do this and I found `GIT_TERMINAL_PROMPT=0`. The tradeoff is we lose a bit of granularity on the errors because all we know is we couldn't prompt for credentials. I think this is a better/safer option on balance though, and it allows us to rewrite this function so it doesn't require twice as much comment as code to explain what is going on :)

chris48s:
I know there are a couple of conflics on this PR but they are only on `package.json`/`package-lock`. I'll deal with resolving them if/when we're happy with the main body of code.

services/git.js
calebcartwright: > Passing the creds in the URL does prevent that prompting.

That seems like pretty odd behavior. I would've expected providing the same empty/bogus auth value in the header as in the uri userinfo component to have the same behavior :man_shrugging: Anyway, glad it's resolved :+1:

PyvesB:
> Also, I'm still having my own internal debate to try to make up my mind on whether I think Shields should continue trying to do auto-magic default branch detection for services that require Shields to send them a branch param, but don't provide their own auto-magic solution like GitHub does with HEAD.
> 
> On one hand I understand the convenience for users to not have to specify a branch when setting up their badge. OTOH, setting up a given badge is really a one time action for end users (it's not like they have to take action and keep re-specifying their branch every the time the badge is loaded/viewed), but Shields has to do this auto-magic work on every (non-cached) badge request.
> 
> For me it's a matter of tradeoffs and weighing the convenience for a user to not have to include a `/my-branch` when they add their badge to a readme/etc. against the increased workload for the badge servers (extra network calls, process spawning, etc.) and associated code maintenance.
> 
> > Although we try and avoid them, we've already got a few services that make more than one API call (or can do) e.g: matrix, nuget, f-droid, codeclimate
> 
> That's quite true although AFAIK in all of those cases (other codeclimate) the sequential API calls are strictly necessary to get the required data vs. Shields making assumptions/auto-populating a parameter value that the upstream service requires by making multiple networking calls.

This message is a good summary of the trade-offs we have to consider. 👍 

In my opinion, making the users specify a branch when setting up their badge is not the end of the world, especially as they'll rarely change the default branch name. It even has the nice side effect of making badges paths more consistent. I find it suboptimal that we specify most optional elements as query parameters in our badge URLs, but for branches we've historically been using optional path elements.

I don't think we should go down the route of increasing our code complexity. Doing multiple network calls to render our badges makes them less reliable and will impact performance. We've always been careful when implementing badges that do multiple network calls and as mentioned by @calebcartwright it's generally been as a last resort if we can't get the information by any other means or if the parameter value is not easily "guessable" by the user.

chris48s:
Just to clarify, what's your proposed way forward?

Are you saying we should:
- Convert branch to a required param on these badges and break badges for existing users who are using Bitbucket Pipeline, Gitlab Pipeline and Travis PHP version without a branch (which is probably most of them).
- Leave things as they are (assuming the default branch is called master) for these three services?

PyvesB:
My proposed way forward would be something in the middle: make the branch name a required parameter, but add a service redirect that assumes that the default branch is master to avoid breaking existing badges.

calebcartwright:
> My proposed way forward would be something in the middle: make the branch name a required parameter, but add a service redirect that assumes that the default branch is master to avoid breaking existing badges. What do you think?

That's what I had in mind as well

chris48s:
Two follow-up questions:

1. Do you think we should do that in all three of these cases, or do you think we should only do it on GitLab/BB (where we have to increase the render time of the badge in order to work out the default branch) but audo-detect on TravisPHP (where we can add the extra call without having to increase render time because we can issue two requests at once)?

2. We touched on this _slightly_ in #5215 , but in your view, what is the difference between:
    * `branch` is not a required param. If you don't specify it, we'll assume `master`
    * `branch` is a "required" param (but if you don't specify it we'll assume `master` for backwards-compatibility via a redirect - don't tell anyone though :wink: )

I'm not necessarily against this approach, but I'm interested in understanding your thinking.

calebcartwright:
> We touched on this slightly in #5215 , but in your view, what is the difference between:

Fair question. While I agree that from an end-user perspective the result is essentially the same, the motivation and justifications are different. Philosophically I think that our badge route parameters should require the same data points that the upstream service/platform requires, and option 2 moves us to that end.

Yes, we'd have to do the redirect thing because we don't have a time machine and can't break existing badges, but we're still moving to a point where our set of required params match the upstream set.

If anyone "discovers" the `master` branch assumption, IMO it's easier to explain that we had to do so with a redirector to avoid breaking a massive set of badges vs. it being a strategic decision that Shields is continuing to auto-assume master. The redirect behavior will also impact existing badges in the sense that they'll be ever so slightly slower due to the redirect, so there's motivation for existing users to update their badge routes (especially if/when folks start renaming their default branches).

The second option would presumably help avoid accruing any more "debt" of new users leveraging the assumed branch behavior, at least relative to the first option. 

Lastly, I think the second option helps avoid any future debates with other services, as it takes the sails out from the "well the XYZ service class route has an optional param with an assumed default even though XYZ requires it" 

<hr />

> Do you think we should do that in all three of these cases

I'd vote for all three for many of the same reasons as above.

<hr />

I realize that one impact of the approach I'm suggesting (not assuming/detecting a default branch) means that users that change their default branch name will likely see broken badges and have to correct their badge routes. 

I'm quite sure there will be some folks that will be angry/irritated by that, and can understand that initial frustration, but still feel this is the right way to go. It'll be a minor, one-time nuisance, but it's not Shields job IMO to assume/determine values for required params on behalf of users and the drawbacks of attempting to do so just far outweigh the pros.

calebcartwright:
Also, if we do end up going the route of auto-detecting branches then I do think the code and changes here are probably good to go. Would just want to keep an eye on the resource utilization on Heroku for a bit post-deployment

paulmelnikow:
Lots of great points all around! @chris48s thanks for taking on this work!

I think multiple requests are not ideal, but okay in circumstances where it is required to get the information we need, or where it makes the badge drastically easier to use. (Matrix is a good example of that, IIRC.)

A handful of the badges used to invoke a subprocess but I think we've now gotten those down to zero. I do think subprocesses are important to avoid, as it is a very, very easy way to overwhelm the server with a DoS.

If we need to `ls-remote` would it be worth looking at a pure JS implementation like [Isomorphic Git](https://isomorphic-git.org/)? (Also definitely want to be really careful with sending credentials. I also am not sure if the legacy servers might have private SSH keys on them which could have access to non-public repos.)

In terms of how we handle the defaulting, a redirect will very gently nudge people to the new URLs, mainly because when it's pasted into a browser, you will end up seeing the redirect target. I do think the redirect approach will reduce the number of new repos which will rely on the default. Someday we could deprecate the redirect and replace it with a "please specify a branch" message. Assuming we're not ready now to break existing badges and say, "please specify a branch," I think that'd be my preference.

chris48s:
Yeah I think I will close this and submit a new PR making branch "required" (but not actually required :wink:). While at the moment there are "only" three badges to think about, its not inconceivable we may end up with a full suite of bitbucket or gitlab badges at some later date. If we commit to auto-detect/two requests now, its hard to row back on it later. If we start publicising the existing badge as "requiring" a branch param now we can say its a hard requirement if/when we add more. We could always add a branchless option in future for example if BB/Gitlab allows it to be done in a single query (like GH does), but once it exists its basically impossible to take it away.
The replacement PR will be a lot simpler than this one, but I haven't got time tonight. I'll aim to have a look at the weekend.

PR #5262, opened 2020-07-03
[GitlabCoverage] Add badge for code coverage

### LINKED ISSUE #3333: Badge Request: Gitlab Code Coverage
:clipboard: **Description**

- Service: Gitlab
- Information: Code Coverage
- badge

:link: **Data**

- public API: Yes
- API key: no
- doc: https://docs.gitlab.com/ce/api/pipelines.html#get-a-single-pipeline

- Example: 
_10779832_ is the project's id

from https://gitlab.com/api/v4/projects/10779832/pipelines get the latest pipeline id and then go to https://gitlab.com/api/v4/projects/10779832/pipelines/57333607 to get the `coverage`

:microphone: **Motivation**

You propose already multiple services for Codecoverage but for people who want to use Gitlab's one it would be better

riskycase:
Fixes #3333 

 Todo:

- [x] Add code to render the badge that shows the code coverage as reported in GitLab CI
- [x] Add tests for above code
- [x] Update documentation -> Needs discussion on how to implement
- [x] Add support for selecting job to extract coverage from


riskycase:
Need a few opinions on what to do for following cases:
* GitLab has a section in the settings where we set a RegEx to capture the coverage percentage from the logs. Now what happens is that there is same behaviour for 
  1. Branch that exists but coverage is not set up
  2. Branch doesn't exist
* Non existent repo seems to fail silently (shows a grey 'inaccessible' instead of 'repo not found')
* There is one failing test for the pipeline badge config as well, maybe related to previous point

calebcartwright:
@riskycase when you get a chance could you rebase and grab the latest changes from the default branch? We have made some updates to how branches are handled with services, including GitLab, that require Shields to provide a branch in our request. 

Also regarding these points from your last comment
> * Non existent repo seems to fail silently (shows a grey 'inaccessible' instead of 'repo not found')
> * There is one failing test for the pipeline badge config as well, maybe related to previous point

Correct, this is a separate and still pending issue (refs #5282)

riskycase:
This is what I have gone with so far...

> * GitLab has a section in the settings where we set a RegEx to capture the coverage percentage from the logs. Now what happens is that there is same behaviour for
>   
>   1. Branch that exists but coverage is not set up
>   2. Branch doesn't exist

Added a line in documentation about setting up code coverage in GitLab UI which points to the GitLab docs

> * Users can also get coverage for a specific job (documented [here](https://docs.gitlab.com/ee/ci/pipelines/settings.html#test-coverage-report-badge)), so maybe we'll need to duplicate for cases where the badge is generated from a specific job rather than the default

Have no clue how to implement this.

calebcartwright:
> > * Users can also get coverage for a specific job (documented [here](https://docs.gitlab.com/ee/ci/pipelines/settings.html#test-coverage-report-badge)), so maybe we'll need to duplicate for cases where the badge is generated from a specific job rather than the default
> 
> Have no clue how to implement this.

Could you provide a little more backgruond on this one? I'm not familiar with GitLab's pipelines, and the docs didn't really help my understanding either. Specifically I want to make sure I understand what a "Job" is in this context.

riskycase:
A job in GitLab CI is the equivalent of a Step in Circle CI or job in Travis CI. It's just the group of commands to be run that will either build, test or deploy. Now coming to the GitLab feature, let's say if we have two Jobs which run tests for the lastest version and the LTS version of node (like the above services and services@node-12), there's an option to get the coverage from a specific one of those jobs by using a URL query param.

If you see [this](https://gitlab.com/riskycase/broadcastem-core/-/pipelines/168529455) page, I have two of these 'jobs' in the Testing stage, and I can get the badge for a specific job by using `?job=urlEncodedJobName`.

calebcartwright:
> A job in GitLab CI is the equivalent of a Step in Circle CI or job in Travis CI.

Gotcha, thanks for the explanation! I wanted to get clarification since a Job in some pipeline/automation platforms is a run or executed instance of a given pipeline and thus relatively static. 

In this case though it definitely seems like we'd want to give users the option to support getting coverage from a specific job. We can support this by updating our badge route here to accept an additional (optional) parameter from users, and then use it accordingly in the call to get the svg from Gitlab. 

Given that the branch is a required route parameter, and the fact that job will be optional anyway, we can add an additional param (perhaps `jobName`) to the query param schema

riskycase:
> accept an additional (optional) parameter from users
Right, and since this is what we do for supporting self hosted instances, I could probably add this here in this PR itself. Now given that parameter is optional, would something like comparing with undefined do?

chris48s:
Sorry to drive-by on this one, but I don't think we should add the redirect to `/master` for new badges. We need it on old ones to avoid a BC break, but when adding new badges for services that don't allow us to get a result for whatever the default branch is in one query, we should just make branch a required param moving forwards.

calebcartwright:
>  don't think we should add the redirect to /master for new badges

Agreed

> Sorry to drive-by on this one,

Ha, no worries, the more eyes the better. I actually haven't had a chance to look at the code changes yet myself and not sure if any one else has either.

riskycase:
TBH I too was waiting for some feedback, though this is more or less the code copied and modified from the pipeline status badge

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,121 @@
...
+      schema: badgeSchema,
+      url: `${baseUrl}/${user}/${repo}/badges/${branch}/coverage.svg?style=flat-square`,
+      errorMessages: {
+        401: 'repo not found',
+        404: 'repo not found',
+      },
+    })
+    if (!/[0-9.]+/.test(percentageText)) {
+      throw new NotFound({ prettyMessage: 'not set up' })
+    }
```
calebcartwright, line 101: Given the schema validation, we know `percentageText` will either be a coverage percentage string, or `unknown`. I assume this is intended to handle the latter case, but the benefits of using the schema for validation is that we don't have to re-check values again in the body. Something like the below would suffice

```suggestion
    if (percentageText === 'unknown')) {
      throw new NotFound({ prettyMessage: 'not set up' })
    }
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,121 @@
...
+    // Parse percentage to a number
+    const percentage = Number(/[0-9.]+/.exec(percentageText)[0]).toFixed(0)
+    return this.constructor.render({ percentage })
```
calebcartwright, line 104: Think this could be expressed a bit more simply, and leave the precision decision to the render function where it's better addressed. Note the arg rename for the render function to match the suggested param name change above. 

The service class for Codecov is a good example to reference, as it also uses the upstream Svg https://github.com/badges/shields/blob/master/services/codecov/codecov.service.js 

```suggestion
    return this.constructor.render({ coverage: Number(percentageText.slice(0, -1)) })
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,121 @@
...
+    pattern: ':user/:repo',
+  },
+  transformPath: ({ user, repo }) => `/gitlab/coverage/${user}/${repo}/master`,
+  dateAdded: new Date('2020-07-12'),
+})
+
+module.exports = {
+  GitlabCoverage,
+  GitlabCoverageRedirector,
+}
```
calebcartwright, line 121:
```suggestion
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,121 @@
...
+<img src="https://user-images.githubusercontent.com/12065866/67156911-e225a180-f324-11e9-93ad-10aafbb3e69e.png" alt="Setting Public pipelines set"/>
+<p>
+Also make sure you have set up code covrage parsing as described <a href="https://docs.gitlab.com/ee/ci/pipelines/settings.html#test-coverage-parsing">here</a>
+</p>
+<p>
+  Your badge should be working fine now.
+</p>
+`
+
+class GitlabCoverage extends BaseSvgScrapingService {
```
calebcartwright, line 45: Can (and should) export the class this way once the redirector is removed to allow for using the service test helpers

```suggestion
module.exports = class GitlabCoverage extends BaseSvgScrapingService {
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,121 @@
...
+      },
+    ]
+  }
+
+  static render({ percentage }) {
+    return {
+      message: `${percentage}%`,
+      color: coveragePercentageColor(percentage),
+    }
+  }
```
calebcartwright, line 85: bit of a nit, but would suggest an alternative param name like `coverage`. it's more explicit and also consistent with the majority of the other coverage-related service classes. note the `toFixed` moved here, more details below in the handle function

```suggestion
  static render({ coverage }) {
    return {
      message: `${coverage.toFixed(0)}%`,
      color: coveragePercentageColor(coverage),
    }
  }
```

services/gitlab/gitlab-coverage.tester.js
```diff
@@ -0,0 +1,44 @@
+'use strict'
+
+const { isIntegerPercentage } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Coverage').get('/gitlab-org/gitlab-runner.json').expectBadge({
+  label: 'coverage',
+  message: isIntegerPercentage,
+})
+
```
calebcartwright, line 10: This is reliant upon the redirector which we need to remove, so this test should go too

```suggestion
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,121 @@
...
+
+  async handle(
+    { user, repo, branch = 'master' },
+    { gitlab_url: baseUrl = 'https://gitlab.com' }
+  ) {
+    const { message: percentageText } = await this._requestSvg({
+      schema: badgeSchema,
+      url: `${baseUrl}/${user}/${repo}/badges/${branch}/coverage.svg?style=flat-square`,
```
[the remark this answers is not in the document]
riskycase: This style param returns only a single pair of coverage: percentage. We could theoretically do away with it but the svg elements returned are not labelled anything, so this seemed to be the best way to go with.

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,121 @@
...
+  coveragePercentage: coveragePercentageColor,
+} = require('../color-formatters')
+const { optionalUrl } = require('../validators')
+const { BaseSvgScrapingService, NotFound, redirector } = require('..')
+
+const badgeSchema = Joi.object({
+  message: Joi.alternatives()
+    .try(
+      Joi.string().regex(/^([0-9]+\.[0-9]+%)|unknown$/),
+      Joi.equal('unknown')
```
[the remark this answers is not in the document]
riskycase: The catch here is that GitLab just RegEx's the coverage from the log of the executed commands, so ideally this should cover most cases IMO

I think unknown was added when the badge didn't work for the non-existent branch/coverage not set up situation

riskycase:
Replied to the points which I feel required explanation, I'll get to work on removing redirector class and the other suggestions once I get back to my PC

services/gitlab/gitlab-coverage.service.js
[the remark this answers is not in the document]
calebcartwright: > I think unknown was added when the badge didn't work for the non-existent branch/coverage not set up situation

There shouldn't be a need to have multiple paths to catch `unknown` though. Either the alternative can be dropped since the regex will allow `unknown` or the regex could be simplified to drop `unknown` since there's an alternative schema option to handle

services/gitlab/gitlab-coverage.service.js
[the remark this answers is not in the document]
calebcartwright: Sorry, I'm not sure I follow. Are you saying that you must pass the extra `style` query param to Gitlab in order to be able to scrape the returned SVG, and that without `?style=flat-square` the returned svg cannot be scraped/is significantly more difficult to do so?

If there is such a requirement then please add some explicit inline documentation that explains that, otherwise this seems like a superfluous query param that could be removed at some point in the future

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,121 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const {
+  coveragePercentage: coveragePercentageColor,
+} = require('../color-formatters')
+const { optionalUrl } = require('../validators')
+const { BaseSvgScrapingService, NotFound, redirector } = require('..')
```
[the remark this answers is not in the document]
riskycase: [Done](https://github.com/badges/shields/pull/5262/commits/2fd761890095c18e503140a0f9fe2524eacd2c9d#diff-481a722248c7617406848306c7335ce8R6)

services/gitlab/gitlab-coverage.service.js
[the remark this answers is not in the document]
riskycase: I checked it without the style param and it worked right, I guess I assumed the code might break with multiple matches.

[Done](https://github.com/badges/shields/pull/5262/commits/2fd761890095c18e503140a0f9fe2524eacd2c9d#diff-481a722248c7617406848306c7335ce8R88)

services/gitlab/gitlab-coverage.service.js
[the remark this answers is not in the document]
riskycase: Removed line 14

services/gitlab/gitlab-coverage.service.js
riskycase: [Done](https://github.com/badges/shields/pull/5262/commits/2fd761890095c18e503140a0f9fe2524eacd2c9d#diff-481a722248c7617406848306c7335ce8R94)

services/gitlab/gitlab-coverage.service.js
riskycase: [Done](https://github.com/badges/shields/pull/5262/commits/2fd761890095c18e503140a0f9fe2524eacd2c9d#diff-481a722248c7617406848306c7335ce8R97)

riskycase:
Have resolved all the suggested changes, we're only left with the part where the user can select a job to get the coverage from.
One way could be to make the job name too a required parameter since this badge isn't deployed yet, or use query string like how it is done for the base GitLab url, but unlike that which has a default value (`https://gitlab.com`), we'll have to omit the qs completely while requesting

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,101 @@
...
+
+const Joi = require('@hapi/joi')
+const { coveragePercentage } = require('../color-formatters')
+const { optionalUrl } = require('../validators')
+const { BaseSvgScrapingService, NotFound } = require('..')
+
+const badgeSchema = Joi.object({
+  message: Joi.alternatives()
+    .try(Joi.string().regex(/^([0-9]+\.[0-9]+%)|unknown$/))
+    .required(),
```
calebcartwright, line 11: No need for `alternatives` now
```suggestion
  message: Joi.string().regex(/^([0-9]+\.[0-9]+%)|unknown$/).required(),
```

services/gitlab/gitlab-coverage.tester.js
```diff
@@ -0,0 +1,39 @@
...
+    label: 'coverage',
+    message: isIntegerPercentage,
+  })
+
+t.create('Coverage (existent branch but coverage not set up)')
+  .get('/gitlab-org/gitlab/master.json')
+  .expectBadge({
+    label: 'coverage',
+    message: 'not set up',
+  })
```
calebcartwright, line 18: This test is failing in the ci check, as the upstream badge does indeed return a coverage percentage.
https://gitlab.com/gitlab-org/gitlab/badges/master/coverage.svg


https://app.circleci.com/pipelines/github/badges/shields/2330/workflows/5e0300af-43b8-445a-a7c8-04bd47cefd2c/jobs/111783

```bash
GitlabCoverage [live] Coverage (existent branch but coverage not set up) [ GET /gitlab-org/gitlab/master.json ] - [ GET /gitlab-org/gitlab/master.json ]
/home/circleci/project/core/service-test-runner/cli.js

AssertionError: message mismatch: expected '70%' to equal 'not set up'
    at Function._expectField (core/service-test-runner/icedfrisby-shields.js:69:51)
    at IcedFrisbyNock.afterJSON.json (core/service-test-runner/icedfrisby-shields.js:54:26)
    at IcedFrisbyNock.<anonymous> (node_modules/icedfrisby/lib/icedfrisby.js:954:10)
    at invokeNextHook (node_modules/icedfrisby/lib/icedfrisby.js:1003:24)
    at Promise (node_modules/icedfrisby/lib/icedfrisby.js:1017:7)
    at new Promise (<anonymous>)
    at IcedFrisbyNock._runHooks (node_modules/icedfrisby/lib/icedfrisby.js:976:12)
    at IcedFrisbyNock.run (node_modules/icedfrisby/lib/icedfrisby.js:1276:20)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,101 @@
...
+      {
+        title: 'Gitlab code coverage (self-hosted)',
+        namedParams: { user: 'GNOME', repo: 'libhandy', branch: 'master' },
+        queryParams: { gitlab_url: 'https://gitlab.gnome.org' },
+        staticPreview: this.render({ coverage: 93 }),
+        documentation,
+      },
+    ]
+  }
+
```
calebcartwright, line 74: Let's add the defaultBadgeData function, which will address the failing tests by ensuring the expected label is always set.

```suggestion

  static get defaultBadgeData() {
    return { label: 'coverage' }
  }
  
```

(See item number 8 in [this section](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#43-querying-an-api) of the tutorial for more info)

calebcartwright:
FYI I've updated the PR title to only run the service tests for the new badge being added here and not the existing Gitlab Pipeline Status badges as there's a known issue with those (refs #5282).

> One way could be to make the job name too a required parameter since this badge isn't deployed yet, or use query string like how it is done for the base GitLab url, but unlike that which has a default value (https://gitlab.com), we'll have to omit the qs completely while requesting

This should be an optional param users can leverage, or not, so it needs to be a query parameter. And yes, it shouldn't have a default value because we only want to use it when a user has explicitly provided a value :+1:

riskycase:
Just had a look at the tests, the one that you have pointed out seems to be due to the changes upstream, as the exact same test had passed at the time of writing (https://github.com/badges/shields/blob/3e4b08054c9b9e0f4d7a1431b3df031273f6f432/services/gitlab/gitlab-coverage.tester.js#L18). To keep the tests passing we will need to find a repo that will never have coverage set up at all

riskycase:
Have modified the handle function this way:
```js
  async handle(
    { user, repo, branch },
    { gitlab_url: baseUrl = 'https://gitlab.com', job_name	}
  ) {
	job_name = job_name ? `?job=${job_name}` : '';
    const { message: percentage } = await this._requestSvg({
      schema: badgeSchema,
      url: `${baseUrl}/${user}/${repo}/badges/${branch}/coverage.svg${job_name}`,
      errorMessages: {
        401: 'repo not found',
        404: 'repo not found',
      },
    })
    if (percentage === 'unknown') {
      throw new NotFound({ prettyMessage: 'not set up' })
    }
      coverage: Number(percentage.slice(0, -1)),
    })
  }
```
But how do I add the `job_name` to `queryParamSchema`? The baseUrl key is tested against a validator, and I'm not sure if I should validate this too or try to add it as is

calebcartwright:
> But how do I add the job_name to queryParamSchema? The baseUrl key is tested against a validator, and I'm not sure if I should validate this too or try to add it as is

Sorry I'm not sure I understand the question here. You just need to add it as another key in the schema:


```js
  gitlab_url: optionalUrl,
  job_name: Joi.string(), // or Joi.string().optional(), though the optional piece isn't necessary
}).required()
```

Also, at this point you may want to consider extracting out a `fetch` function to handle the svg retrieval, and a `transform` function to extract the coverage value. https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#43-querying-an-api

riskycase:
I wasn't exactly sure I understood the transform function so I dropped the error handling in there.

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,143 @@
...
+  static transform({ coverage }) {
+    if (coverage === 'unknown') {
+      throw new NotFound({ prettyMessage: 'not set up' })
+    }
+    return this.render({
+      coverage: Number(coverage.slice(0, -1)),
+    })
+  }
```
calebcartwright, line 134: Exactly!

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,143 @@
...
+    return {
+      message: `${coverage.toFixed(0)}%`,
+      color: coveragePercentage(coverage),
+    }
+  }
+
+  async fetch(
+    { user, repo, branch },
+    { gitlab_url: baseUrl = 'https://gitlab.com', job_name }
+  ) {
```
calebcartwright, line 111: nit but we can use a single destructured object param in the signature here. the handle function has them split in two due to how the routing/server infra handles route vs. query params and passes them to these service classes, but we don't have to mimic that between these internal functions
```suggestion
  async fetch({
    user,
    repo,
    branch,
    gitlab_url: baseUrl = 'https://gitlab.com',
    job_name,
  }) {
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,143 @@
...
+    // Since the URLdoesn't return a usable value when an invalid job name is specified,
```
calebcartwright, line 112:
```suggestion
    // Since the URL doesn't return a usable value when an invalid job name is specified,
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+    return this.render({
+      coverage: Number(coverage.slice(0, -1)),
+    })
+  }
+
+  async handle({ user, repo, branch }, { gitlab_url, job_name }) {
+    const svg = await this.fetch(
+      { user, repo, branch },
+      { gitlab_url, job_name }
```
calebcartwright, line 139: See suggestion on the fetch function that would need this to be changed accordingly as well
```suggestion
      { user, repo, branch, gitlab_url, job_name },
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,143 @@
...
+    if (coverage === 'unknown') {
+      throw new NotFound({ prettyMessage: 'not set up' })
+    }
+    return this.render({
+      coverage: Number(coverage.slice(0, -1)),
+    })
+  }
+
+  async handle({ user, repo, branch }, { gitlab_url, job_name }) {
+    const svg = await this.fetch(
```
calebcartwright, line 137: Don't feel obligated to change this, but could go ahead and destructure this as well with a rename for use below on the call to the transform function
```suggestion
    const { message: coverage } = await this.fetch(
```

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,143 @@
...
+      { user, repo, branch },
+      { gitlab_url, job_name }
+    )
+    return this.constructor.transform({ coverage: svg.message })
```
calebcartwright, line 141: Our handle function still needs to call render, so this could either be split into two lines

```suggestion
    const { coverage } = this.constructor.transform({ coverage })
    return this.constructor.render({ coverage })
```

or oneline:
or one line
```suggestion
    return this.constructor.render({ coverage: this.constructor.transform({ coverage }) })
```

riskycase:
I guess this should wrap everything up

services/gitlab/gitlab-coverage.service.js
```diff
@@ -0,0 +1,149 @@
...
+        documentation,
+      },
+      {
+        title: 'Gitlab code coverage (specific job)',
+        namedParams: {
+          user: 'gitlab-org',
+          repo: 'gitlab',
+          branch: 'master',
+        },
+        queryParams: { job_name: 'rspec:coverage' },
```
calebcartwright, line 73: Out of curiosity, do we know of any jobs that have this setup? 

The upstream badge for this is returning `uknown` so the Shields badge is showing the corresponding `not set up`
https://gitlab.com/gitlab-org/gitlab/badges/master/coverage.svg?job=rspec:coverage

riskycase: This job did indeed exist at least at the time of writing (https://gitlab.com/gitlab-org/gitlab/-/jobs/667537958), but it's possible that it was recently removed

riskycase: We could use the gitlab-runner repo instead, that has lesser activity

https://gitlab.com/gitlab-org/gitlab-runner/-/jobs/670879955

calebcartwright: While we strive to have examples that work/fully demonstrate their respective features, I was primarily interested in finding a working instance of a coverage badge for a specific job. Would `test coverage report` be the correct job name for the gitlab-runner` repo? 

This one does indeed seem to work:
https://shields-staging-pr-5262.herokuapp.com/gitlab/coverage/gitlab-org/gitlab-runner/master?job_name=test%20coverage%20report

riskycase: Okay, I'll use that. While I would've liked to have the example show a job in case where different jobs had different coverages (`gitlab-org/gitlab` had different coverage for Ruby and JS), I think we can give the higher priority to a working badge instead.

riskycase:
I'll merge the upstream branch to make sure everything goes smoothly

services/gitlab/gitlab-coverage.service.js
calebcartwright: Yeah that would've been ideal, but for now just having something that works will suffice :smile:

PR #5263, opened 2020-07-03
Escape special characters in badge URL

hkva:
Hi. When generating markdown for a badge with spaces in the URL, the spaces need to be replaced with their HTML-encoded equivalent (`%20`) or the badge will not display on GitHub.

Here's a basic example of the issue in action: https://gist.github.com/hkva/159a4f5d20b70edc234d18e5081dfb45

This PR adds a one-line fix in the markdown generator to replace the spaces.


frontend/lib/generate-image-markup.ts
```diff
@@ -18,7 +18,8 @@ export function markdown(
   link?: string,
   title?: string
 ): string {
-  const withoutLink = `![${title || ''}](${badgeUrl})`
+  const withoutSpaces = badgeUrl.replace(/\s/g, '%20')
+  const withoutLink = `![${title || ''}](${withoutSpaces})`
```
paulmelnikow, line 22: Hi! Thanks for this and for looking into a fix!

I think the problem is not here, but in the badge URL, which should be a valid URL before it's passed to this function. Fixing it here would only fix it for Markdown though the problem exists for each of these formats since the URL is what's wrong.

hkva: @paulmelnikow Ok, that makes sense. I'm not too familiar with the code base. Would this be a better spot to implement a fix? https://github.com/badges/shields/blob/b4579adc44329e33c7491a78f14a53a7b30ce34c/frontend/components/customizer/customizer.tsx#L51

paulmelnikow: I think it needs to be here:

https://github.com/badges/shields/blob/b4579adc44329e33c7491a78f14a53a7b30ce34c/frontend/components/customizer/path-builder.tsx#L81-L108

Issue #3795, opened 2019-07-28
Provide the endpoint response's label and message as an alt text option


nodes777:
**Issue**
Currently, when using the [JSON endpoint method](https://shields.io/endpoint) of generating a badge, the returned img element has an alt="". A user should be able to use the endpoint's label and message to populate the alt attribute. This is a use case where presumably the badge information is being generated dynamically, and therefore a user can't reliably fill in the alt text themselves.

This creates an accessibility issue where screen reader users are not able to receive the same information. The information that is visually apparent is not programmatically present.

**Expected Behavior**
For example, if I'm grabbing some code coverage information from a third party, and return the following object:
`{
  "schemaVersion": 1,
  "label": "Line Coverage:",
  "message": "98%",
  "color": "orange"
}`

Then then generated badge img should be:

`<img alt="Line Coverage: 98%" src="<the-generated-url>">`

I'd suggest that the default behavior be that the label and message be the alt attribute, though providing the option to override that would also be appreciated.

Thank you!

paulmelnikow:
Hi! Thanks for your question.

What do you mean, the returned img element? Are you talking about the "Copy HTML" option under the "Customize and test" section of the frontend?

nodes777:
Hi Paul,

Thanks for your time!

I'm referring to the img element in the below picture from the [example JSON page](https://shields.io/endpoint). Where the "is it monday" img has an alt of null (alt=""). But I would expect the alt to contain the text in the image "is it monday no". And to be able to provide the label and message properties to the alt of any badge.

Let me know if that makes sense.

Thanks!

![Capture](https://user-images.githubusercontent.com/10222435/63356052-243a6300-c335-11e9-8d8a-12dd0eb3756c.PNG)

paulmelnikow:
Hmm, I'm not totally sure what you want to accomplish.

When you embed the badge in a web page using an <img> element, if the badge is dynamic, that means you don't know what the text will be until the endpoint is hit. If the badge is always the same, sure, you can put the text in the alt attribute, though you may as well use a static badge instead of the endpoint badge.

The web page you're looking at is solely for previewing so the only thing that would make sense to use for an alt is something like "endpoint badge preview".

And again, there's the same issue which is that the alt attribute needs to be set on the frontend, and the contents of the badge is not known by the frontend.

[...]

Issue #229, opened 2014-08-05
Accessibility of SVG images for screenreaders


finnp:
Hey,

I just read bout accessibility of SVG images: http://www.paciellogroup.com/blog/2013/12/using-aria-enhance-svg-accessibility/. I think having SVGs is a great chance to make the information available to blind people.

I might look into this later myself, just wanted to let you know about this.

Finn

PR #5289, opened 2020-07-06
feat: add accessible text alternative in when a graphic and link

evelynhathaway:
This PR adds accessible text to the SVG when used inline so that people using assistive technologies (ATs) can read the dynamic badge values.

It also fixes the tab order and content of links so they are read out to ATs.

This will not help GitHub for instance which post-processes SVGs into PNGs, though we can [advocate for this functionality](https://twitter.com/eeveedev/status/1280280343254163459). 😄♿

**Tested in:**
- Chrome + Narrator (Windows)
- Chrome + NVDA (Windows)
- IE 11 + Narrator (Windows)
- IE 11 + NVDA (Windows)
- Chrome Mobile + TalkBack (Android)

Parallel to https://github.com/badgen/badgen/pull/62

**Related issues:**
- https://github.com/badges/shields/issues/3795
- https://github.com/badges/shields/issues/229


evelynhathaway:
This PR doesn't fix the *pre-existing* compatibility issue with NVDA + IE not handling aria-hidden and SVG links correctly. I gave it a go, but it doesn't seem possible without upstream fixes with supporting SVG links.

Overall, this improves the AT-user experience when used with `<object>` embeds like on the footer of the website. Maybe in the future, GitHub will add <title> support to their markdown parser.

paulmelnikow:
Just wanted to say a quick thanks for this important contribution @evelynhathaway! I or one of the other maintainers will take a look at this soon!

chris48s:
Cheers for having a look at this. I haven't looked at the code properly yet, but I've had a quick test on the staging deploy ( https://shields-staging-pr-5289.herokuapp.com/ ) and generating a badge with links doesn't render any text for the flat/flat-square/pastic badges (social and FTB seem to be ok). So for example, this:

 - https://img.shields.io/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&link=https%3A%2F%2Fgithub.com%2Fbadges%2Fshields%2F&style=flat

renders as expected, but the equivalent call on the review app:

 - https://shields-staging-pr-5289.herokuapp.com/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&link=https%3A%2F%2Fgithub.com%2Fbadges%2Fshields%2F&style=flat

shows no text (the staging deploy might take a few seconds to spin up). Can you have a look at fixing that before we review this please.

This is going to be a bit of a fiddly diff to review, but I'll try and carve some time out to have a look at it. @calebcartwright - given we're close to changing our raster proxy, as part of the review, can you try running these through squint (once fixed) and see if there's any impact of these changed to the rendered SVGs on rasterization?

evelynhathaway:
@chris48s Fixed!

I know it's a big PR that reworks a lot, so I am definitely open to any suggestions on how to make it simpler. This was a bigger task than I thought for someone not familiar with the codebase. Haha.

badge-maker/lib/badge-renderers.js
```diff
@@ -87,46 +111,62 @@ function renderText({
...
+  const rightLink = escapeXml(links[1])
   const hasLeftLink = leftLink && leftLink.length
   const hasRightLink = rightLink && rightLink.length
-  const leftLinkWidth = hasRightLink ? leftWidth : leftWidth + rightWidth
 
-  function render({ link, width }) {
-    return `<a target="_blank" xlink:href="${link}"><rect width="${width}" height="${height}" fill="rgba(0,0,0,0)"/></a>`
-  }
+  const accessibleText = createAccessibleText({ label, message })
+  const excapedAccessibleText = escapeXml(accessibleText)
```
chris48s, line 154: There's a typo in the var name here
`excapedAccessibleText` --> `escapedAccessibleText`

badge-maker/lib/badge-renderers.js
```diff
@@ -87,46 +111,62 @@ function renderText({
...
 
-  return (
-    (hasRightLink
-      ? render({ link: rightLink, width: leftWidth + rightWidth })
-      : '') +
-    (hasLeftLink ? render({ link: leftLink, width: leftLinkWidth }) : '')
-  )
-}
+  if (hasLink === undefined) {
+    hasLink = links && links.length && links.every(link => link)
```
chris48s, line 157: I'm not following the intent of `links.every(link => link)` here.
Are we trying to achieve something different from `hasLeftLink || hasRightLink` ?

chris48s:
As it stands, this PR introduces a change in behaviour. Right now, if we create a social badge with only one link, only the left side of the badge is a link (this is different to the other 4 badge styles where the whole badge is a link in this situation). This PR changes that behaviour. Compare

http://img.shields.io/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&style=social
with
https://shields-staging-pr-5289.herokuapp.com/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&style=social

Bringing them into line probably simplifies things, but I have no idea if they're currently different by design or if its just inconsistency. Can anyone remember if they are different for _good reasons_ or not? I wouldn't be opposed to just changing it if this makes the code simpler tbh.

chris48s:
> I am definitely open to any suggestions on how to make it simpler. This was a bigger task than I thought for someone not familiar with the codebase

Having read it all over, I have written some notes on tidying this up, but I am also conscious reading over them that this is quite involved for a first contribution and where to go next with it is a bit subjective. If you're keen to get further into it, I'm happy to post some thoughts but also if you're not (my instinct is that this is a PR motivated by caring about accessibility in widely used projects rather than caring specifically about this project per-se?) I think it would be reasonable to review up to the point where we are happy with the core change/impact on the rendered outputs and one of the maintainers can pick up a bit of refactoring/tidy-up as a follow-up.

I'll leave it to you to lead us on where you want to go with it..

paulmelnikow:
> Bringing them into line probably simplifies things, but I have no idea if they're currently different by design or if its just inconsistency. Can anyone remember if they are different for _good reasons_ or not? I wouldn't be opposed to just changing it if this makes the code simpler tbh.

If I had to guess, it's that the difference wasn't by design. I'm content with the change 👍 

I noticed from these two links that the hover appearance and targets have changed slightly, though I think that's not blocking and could be looked at in a follow-on if anyone is interested. (You can compare these three links to see what I mean.)

1. Old one link: http://img.shields.io/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&style=social
2. Old two links: http://img.shields.io/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&link=http://example.com&style=social
3. New one link: https://shields-staging-pr-5289.herokuapp.com/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&style=social

chris48s:
Just to clarify as this is quite subtle stuff now.

At the moment on shields.io, hovering over the left side when there is a link slightly darkens the background fill, whereas on staging it doesn't change it (although there may be a slightly flicker just as you hover). I hadn't spotted that until you pointed it out.

so on live you've got

---

un-hovered
Screenshot at 2020-07-10 23-47-50
and hovered
Screenshot at 2020-07-10 23-47-44

---

whereas on staging it is like

Screenshot at 2020-07-10 23-47-50

with or without hover

---

Is that the only difference you're calling out, or is there another?

paulmelnikow:
Yea, quite subtle stuff (and definitely not blocking). No, it's about how the hover state activates and deactivates. There is actually a hover behavior on the left side but it only activates along the border instead of over the whole path. I think the easiest way to show this would be with a screencap, but I think you can see what I mean if e.g. you slowly move the mouse from outside the left side to the center of the left side. On the old one the hover activates and stays activated, and on the new one it's janky. On the new one, the behavior on the right side is a bit cleaner though I think ideally the hover behavior would apply to the whole area. (I'm on Firefox; it's possible this is different in different browsers.)

badge-maker/lib/badge-renderers.js
evelynhathaway: Yes, I will make this explicit.

badge-maker/lib/badge-renderers.js
evelynhathaway: Thanks for catching this!

evelynhathaway:
> activates along the border instead of over the whole path

You're correct. The left link's click target rectangle is also an overlay. I moved it to its correct order in the SVG so that it once again overlays over the text. I've also moved the `:hover` to match the new link behavior if that behavior is preferred. It now activates the hover styles when hovering over the `<a>`, rather than the overlay. (e.g. hover over either side of a [single link social badge](https://shields-staging-pr-5289.herokuapp.com/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&style=social))

If the old behavior works better, I can add the special cases for that just let me now.

evelynhathaway:
> Having read it all over, I have written some notes on tidying this up, but I am also conscious reading over them that this is quite involved for a first contribution and where to go next with it is a bit subjective.
> [...]
> I'll leave it to you to lead us on where you want to go with it..
> @chris48s

Either way, let me know what works best for the project, and I'll do my best to help out. 😄

chris48s:
Nice. In terms of the changes to the SVG output, I think this is now looking good :+1: Thanks

@calebcartwright - it would be useful to hear from you just to check there's nothing we need to tweak re: rasterization. My instinct is its probably fine, but I'd like to know if there's something here that is going to give us any additional problems with `librsvg` (beyond those we already have :grimacing: )

---

In terms of code structure, this is somewhat subjective. There's a tradeoff here between DRY, seperation of concerns and keeping the control flow easy to follow and we probably can't achieve perfection on all of those, but these are my thoughts:

I think the main problems here are we're doing too much directly in `renderBadge()` and the handling of links is too fragmented. Essentially the way this should ideally be layer-seperated is `Flat`/`FlatSquare`/`Plastic`/`forTheBadge()`/`social()` should accept the directly user-supplied params (`label`, `message`, `logo`, `links`, etc), turn them into XML fragments and then hand the XML strings (or variables calculated during that process) off to `renderBadge()` to assemble a valid SVG output. I think we were actually breaking our own rule here with the original implementation by passsing `links` straight to `renderBadge()` (we should probably have been calling `renderLinks()` elsewhere) but I think with these changes we're now leaking too many of the user input params all the way through here.

I think what I'd probably do is call `createAccessibleText()` in `Badge.constructor`/`social()`/`forTheBadge()`, extract a `renderTitle()` function (which returns `''` or `<title...`) and do the same with that (passing in `title`). The downside there is we're then calling those things in 3 places (your solution is more DRY), but that's true of `renderLogo()`, `renderText()` etc too.. Broadly the thinking here is `Flat`/`FlatSquare`/`Plastic` are slight variations on the same thing and should code-share as much as possible via inheritance whereas `social()` and `forTheBadge()` are just _different things_ and there is some duplication :shrug:

I think optionally wrapping the whole body in an `<a>` tag probably can live in `renderBadge()`, but we could do it outside in another function. Either way, the thing I found hardest to get my head round while reading this was this pattern in `social()` and `forTheBadge()`:

```js
if (hasLeftLink && hasRightLink) { ...

//bunch of other stuff

renderBadge(
  {
    links: hasOnlyLeftLink ? links : []
    ...
   }
)
```

..and then regardless of whether we supply `hasLink` to `renderBadge()`, its always `hasLeftLink || hasRightLink` (but sometimes we've manually fiddled with `links` before we calculate it).

I reckon what we probably want to do here is pull out all the `hasLeftLink` / `hasRightLink` stuff out of `renderBadge()`, maybe into another helper function (`shouldWrapBodyWithLink()`?? Naming things is hard). Maybe social/FTB need their own slightly different version of this logic? Another reason to pull it back a step in the proces.. Then we make the decision "are we going to wrap the whole thing in an `<a>` tag?" in `Badge.constructor`/`social()`/`forTheBadge()`. Then at the point we get to actually calling `renderBadge()` we've made that decsion and we just pass it a param `link` (which by this point is either a URL string or `null`/`undefined`/whatever) and the logic in `renderBadge()` becomes just

```js
${ link ? `<a target="_blank" xlink:href="${link}">${main}</a>` : main }
```

I think bringing all that back a layer into `Badge.constructor`/`social`/`forTheBadge` should allow us to clean up this whole `// Override for aria if links are handled elsewhere (i.e social style)` business and make the link rendering logic a bit easier to understand.

What do you think? If you agree/want to pick that up, that would be great. If you don't really have the appetite for it I could pick it up.

calebcartwright:
@chris48s 

> @calebcartwright - it would be useful to hear from you just to check there's nothing we need to tweak re: rasterization. My instinct is its probably fine, but I'd like to know if there's something here that is going to give us any additional problems with librsvg (beyond those we already have grimacing )

Have checked locally with the same battery of test badges I've been using with the upstream svg source pointed to the review app for this PR (https://shields-staging-pr-5289.herokuapp.com) instead of Shields.io; No change/issues observed

paulmelnikow:
> You're correct. The left link's click target rectangle is also an overlay. I moved it to its correct order in the SVG so that it once again overlays over the text. I've also moved the `:hover` to match the new link behavior if that behavior is preferred. It now activates the hover styles when hovering over the `<a>`, rather than the overlay. (e.g. hover over either side of a [single link social badge](https://shields-staging-pr-5289.herokuapp.com/badge/badges-shields-blue?link=https%3A%2F%2Fgithub.com%2Fbadges%2F&style=social))

Wow, this is looking great! 😍 Thank you!

evelynhathaway:
> What do you think? If you agree/want to pick that up, that would be great. If you don't really have the appetite for it I could pick it up.
> @chris48s 

I took a shot at it in 14edc13ff7ee347a6dd4b63ce6d938b8d443c1ea. Let me know what you think. I didn't move some things out entirely since some directly or indirectly depended on the SVG tag.

PR #5345, opened 2020-07-20
Add [Flathub] badge

kidonng:
Add a badge for [Flathub](https://flathub.org/), an app store and build service for Linux.


chris48s:
Hi thinks for working on this. Before we review it, can you resolve the build failures.

You can identify and resolve the formatting issues with `npm run lint` and `npm run prettier`

With the service tests, you need to call the `.json` endpoints under test, not the SVGs (have a look at another test module and you'll see what I mean). The nock failure is probably due to the rogue ` character on the start of

```js
nock('`https://flathub.org/')
```

but I've not tested it. Looking forward to reviewing this once your build is passing.

kidonng:
🙌

services/flathub/flathub.tester.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+const {
+  isVPlusDottedVersionNClausesWithOptionalSuffixAndEpoch,
+} = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Flathub (valid)').get('/org.mozilla.firefox.json').expectBadge({
+  label: 'flathub',
+  message: isVPlusDottedVersionNClausesWithOptionalSuffixAndEpoch,
```
chris48s, line 10: Do flatpak versions actually have the epoch clause?

services/flathub/flathub.service.js
```diff
@@ -0,0 +1,58 @@
...
+      const data = await this._requestJson({
+        schema,
+        url: `https://flathub.org/api/v1/apps/${encodeURIComponent(
+          packageName
+        )}`,
+      })
+      return renderVersionBadge({ version: data.currentReleaseVersion })
+    } catch (error) {
+      if (error instanceof InvalidResponse) {
+        throw new NotFound({
```
chris48s, line 51: After a bit experementing, I can see why we're doing this: The 'not found' response from FlatHub is a 200 OK with an empty body so we're re-throwing the parse error as a `NotFound`, but its a bit non-standard. When we have to deal with an odd API like this, we usually add a short comment to explain what we're doing e.g:
https://github.com/badges/shields/blob/478d14300c4203a134ddf8cbaf5bc1ee8150e5a3/services/cdnjs/cdnjs.service.js#L52-L56
https://github.com/badges/shields/blob/478d14300c4203a134ddf8cbaf5bc1ee8150e5a3/services/itunes/itunes.service.js#L51-L56
Can we do the same here.

services/flathub/flathub.tester.js
kidonng: I copied the code form archlinux 😅

services/flathub/flathub.service.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+const Joi = require('@hapi/joi')
+const { InvalidResponse, NotFound } = require('..')
+const { renderVersionBadge } = require('../version')
+const { BaseJsonService } = require('..')
+
+const schema = Joi.object({
+  currentReleaseVersion: Joi.string().required(),
```
calebcartwright, line 9: I'm not familiar with Flathub, so just want to confirm that we need the schema to be this relaxed? Can versions be basically anything?

chris48s:
Just having a look at some packages on flathub the version schema appears to be pretty lax
For example this project's "version" is just a git commit hash https://flathub.org/apps/details/io.gitlab.jstest_gtk.jstest_gtk
so I we should just accept an arbitrary string on the way in. For the test, we're testing against firefox, so lets validate it as `isVPlusDottedVersionNClauses`: https://www.mozilla.org/en-US/firefox/releases/

calebcartwright:
> so I we should just accept an arbitrary string on the way in. For the test, we're testing against firefox, so lets validate it as isVPlusDottedVersionNClauses

I don't have any objections to that, but have wondered in the past what value we gain from tests that use stricter schema validation that the service class actually accepts, since it has, on occasion, caused failing service tests

chris48s:
We shouldn't validate a version/build status/etc as `string()` under test because we'll miss genuine failures. "invalid" is a string, so is "inaccessible", etc..

calebcartwright:
> We shouldn't validate a version/build status/etc as `string()` under test because we'll miss genuine failures. "invalid" is a string, so is "inaccessible", etc..

Sure, and I'm not suggesting we shouldn't have any validation in the tests (which just switching to `string()` in the tests would functionally do), but I do wonder if there might be an option somewhere in between no validation in the test and extremely strict validation on the badge message.

For example, using `invalid()` or `disallow()` to with exclusions for `invalid`, `inaccessible`, etc. and/or checking the badge color. 

Again I'm fine with what's here for this badge, but I think there may be value in a reusable test validator that does a "make sure the service test worked and we received a valid badge, but i don't really care what the text of the badge message was"

services/flathub/flathub.service.js
calebcartwright: Confirmed below that version values are not constrained

services/flathub/flathub.tester.js
```diff
@@ -0,0 +1,25 @@
...
+  .get('/org.mozilla.firefox.json')
+  .intercept(nock =>
+    nock('https://flathub.org')
+      .get('/api/v1/apps/org.mozilla.firefox')
+      .reply(200, {
+        flatpakAppId: 'org.mozilla.firefox',
+        currentReleaseVersion: '78.0.1',
+      })
+  )
+  .expectBadge({ label: 'flathub', message: 'v78.0.1' })
```
calebcartwright, line 21: Not strictly necessary to have this test, especially given the live test above. I don't feel too strongly about it though so no need to block.

chris48s:
I think we can merge this as it is. Then issue of how strictly we validate versions under test is a bigger conversation than this PR. Lets move that to an issue, or a discussion or something.

Issue #4962, opened 2020-04-28
Strategy for avoiding rate limiting issues on Heroku


calebcartwright:
A few of the upstream services Shields integrates with had whitelisted the IP addresses of our [OVH badge servers](https://github.com/badges/shields/blob/master/doc/production-hosting.md#badge-servers) so that Shields would not hit rate limits. 

We started running an experiment hosting production on Heroku in response to some availability/performance issues we've had recently (#4878), but now our badge requests are no longer coming from static IP addresses which cause a few rate limiting issues (#4957) 

In the short term we've implemented a workaround by proxying the requests through OVH badge servers (#4956) but we're going to need a long term solution. 

There's more detailed discussions in various channels on the shields Discord server, but a few potential action items:

* Get a list of upstream services that are whitelisting the OVH Server IP addresses to avoid rate limits for Shields.io
* Investigate alternative approaches to the OVH proxy, such as using Heroku static IP addons, running our own proxy server, and/or working with the upstream providers to get a token Shields can use for authentication that will allow Shields.io to avoid those rate limiting levels

paulmelnikow:
At the cost of what these services provide, and the simplicity of what needs to be deployed, I'm inclined to run this ourselves on a vps, perhaps with CloudFlare in front of it. I'm not super familiar with how proxying works, or how to cache it privately, but it seems like we could assemble a solution that way.

I do think caching upstream responses is an area where we can achieve some performance growth, though getting this stabilized would be a higher priority.

paulmelnikow:
Besides OVH, did you identify any other affected services?

chris48s:
I think we said https://hex.pm/ might have whitelisted our IPs at some point, but we've not had any reports that their badges have stopped working since the move

Daniel15:
> working with the upstream providers to get a token Shields can use for authentication that will allow Shields.io to avoid those rate limiting levels

This seems the most sustainable. IP whitelisting isn't really sustainable long-term as it makes it more difficult to move to another server provider, add IPv6 in the future, etc. It's better to have some way to authenticate requests are coming from Shields.io, whether that be JWT tokens, OAuth'd requests, whatever.

PyvesB:
We've successfully implemented a token-based approach for Discord, and Hex.pm turned out not to be a problem in the end. Let's discuss decommissioning the old OVH servers during next Operations meeting!

PR #5346, opened 2020-07-20
Migrate [Discord] implementation to use bot token

PyvesB:
Part of #4962.

In the past few weeks, I've been working with some folks from Discord in order to get us away from the old IP whitelisting solution we had until now. Worth noting that they've been really helpful and friendly!

My Discord account holds a Shields.io app with an associated bot. Said bot has an API token for which they have configured higher  rate limits on their side for our use case. We simply have to send a header `Authorization: Bot the_bot_token` as part of the API requests we perform. This pull request covers the implementation.

One tricky aspect is that the Discord API still works when no token is provided or when a bogus token is provided: it simply has lower rate limiting thresholds. Due to firewall complexities on their side, nothing in the response allows us to tell whether we're correctly using the token or not. _discord.spec.js_ is particularly important to avoid any future regressions. In terms of moving this forward, I do not plan on using any token on the CI, but will include our special bot token on both the staging and production environments. Once this pull request is merged, I will organise some testing with them on the staging Heroku app - they can easily tell from their monitoring whether we're correctly sending well-formed authorization headers or not. Once we're happy with that, we can promote to production.


PyvesB:
> 
> 
> This pull request **introduces 1 alert** when merging [6f14503](https://github.com/badges/shields/commit/6f1450351a563e2176b7249aa810cefec3f4f2ab) into [b1841f2](https://github.com/badges/shields/commit/b1841f288eda4ac497c598f11133ebc2c1ee9526) - [view on LGTM.com](https://lgtm.com/projects/g/badges/shields/rev/pr-2b6b64284de2eed8d69b5aa269867a0f7f72a615)
> 
> **new alerts:**
> 
>     * 1 for Hard-coded credentials

Hmmm... This seems like a false positive.

calebcartwright:
> Hmmm... This seems like a false positive.

Yeah, Semmle's being smart in that it knows a hardcoded value is making it's way into the `Authorization` header value, but not quite smart enough to know that's the key and that substrings of that value like `Basic` or `Bearer` aren't actually an issue :thinking:

PyvesB:
> Yeah, Semmle's being smart in that it knows a hardcoded value is making it's way into the `Authorization` header value, but not quite smart enough to know that's the key and that substrings of that value like `Basic` or `Bearer` aren't actually an issue 🤔

I've rearranged the code for generating the bot authorization header a little, this seems to have sorted things out. There's a small amount of code duplication, but it ends up being more readable, so it's a good compromise if it makes LGTM happy as well.

That alert almost caused me to have a heart attack. When I saw it pop up, I thought I had accidentally committed the special Discord token and that I would have to urgently flag it up with them and shamefully ask for a new one to be set up. No worries in the end! 😄

core/base-service/auth-helper.js
```diff
@@ -151,6 +151,11 @@ class AuthHelper {
     return this.isConfigured ? { Authorization: `Bearer ${pass}` } : undefined
   }
 
+  get _botAuthHeader() {
+    const { _pass: pass } = this
+    return this.isConfigured ? { Authorization: `Bot ${pass}` } : undefined
```
night, line 156: to avoid duplicate code as well as platform specific code in this generic helper, you might consider refactoring this logic to accept the concept of a "token type", and then configure discord to be a `Bot` token type (and you could default it to `Bearer`)

calebcartwright: That's precisely what was there originally but the engine in LGTM was barking about hard coded credentials because of it

https://github.com/badges/shields/pull/5346#issuecomment-661238002
https://github.com/badges/shields/pull/5346#issuecomment-662007456

night: any reason not to just ignore it with `// lgtm [js/hardcoded-credentials]` on that line? unsure why we'd want a less optimal solution to impress a linter

calebcartwright: > unsure why we'd want a less optimal solution to impress a linter

Personally I think it's important for our project to not have screaming "hardcoded credential" errors that are displayed as part of the analysis of our project; it's not a matter of "impressing" a linter. Such warnings are legitimate causes of concerns for our users, especially if they don't have the context/bandwidth to dig into the causes of warnings.

That being said, I wasn't aware that lgtm supported inline disables (thanks for the tip!), and would be quite content to leverage that with the original solution and some inline explanatory comments.

PyvesB: Thanks for the feedback. Will revert the last commit and add an LGTM exclusion instead.

Note to myself for a future PR: we have some existing false positives which could benefit from this as well.

PyvesB: You're gonna laugh.

According to the [LGTM documentation](https://help.semmle.com/lgtm-enterprise/user/help/alert-suppression.html), "suppression comments only apply to the line they are on". So I shoved the `// lgtm [js/hardcoded-credentials]` on the same line.

Then I ran Prettier. It didn't like what I did, so it put the comment on the next line. Same for the shortened catch all `// lgtm` equivalent. So I had to add a `// prettier-ignore`  exclusion as well to the whole method.

Here's what I've ended up with:
```javascript
// prettier-ignore
withBearerAuthHeader(requestParams, bearerKey = 'Bearer') { // lgtm [js/hardcoded-credentials]
  return this._withAnyAuth(requestParams, requestParams =>
    this.constructor._mergeHeaders(
      requestParams,
      this._bearerAuthHeader(bearerKey)
    )
  )
}
```

Not the end of the world, but I'm not impressed by what I ended up with either. It's been a long day on my side, will shut down the IDE for now and sleep over this.

PyvesB: Well, I did come up with something that avoided the Prettier ignore, but the LGTM suppression doesn't seem to work. Anything obvious I've done wrong?

paulmelnikow: @s0 Thought I'd mention you: we're having difficulty placing a lgtm ignore on the same line as the problem due to conflicts with Prettier. I'm curious if you have encountered this elsewhere!

s0: Hehe, this is a funny problem. Reminds me of prettier/eslint fix loops!

@paulmelnikow I'm no longer working for Semmle/GitHub, but I think I can shed some light on this. What you have there is probably fine and will unlikely introduce an alert, if you look at the suppression instructions, it says:

> If you make changes to the suppression of alerts in a pull request, please note that theses changes will only be effective once that pull request is merged. See Alert suppression for more details.

I think the reasoning behind this was to prevent people hiding new problems at review time by also adding comments?

cc @chrisgavin ?

PyvesB: Thanks for the info @s0! As a small improvement suggestion, I feel it would be less confusing if the alert explicitly specified "this alert will be suppressed once the PR is merged", at least it proves that the suppression comment was well-formed. :)

paulmelnikow: > @paulmelnikow I'm no longer working for Semmle/GitHub

I didn't realize this @s0! Thanks for the pointers!

Issue #5282, opened 2020-07-04
GitLab Pipeline Status service test failing


calebcartwright:
:clock11: **When did the problem start?**
~8 days ago, ~26th of June

:camera: **Live badge**

https://img.shields.io/gitlab/pipeline/gitlab-org/gitlab
![](https://img.shields.io/gitlab/pipeline/gitlab-org/gitlab)

:wrench: **Is the live badge working?**
Yes

:link: **CircleCI link**

https://app.circleci.com/pipelines/github/badges/daily-tests/309/workflows/360cfa77-ba5c-422c-b2e6-dbabe6152878/jobs/1206

:beetle: **Stack trace**

```
GitlabPipelineStatus [live] Pipeline status (nonexistent repo) [ GET /this-repo/does-not-exist.json ] - [ GET /this-repo/does-not-exist.json ]
/home/circleci/project/shields/core/service-test-runner/cli.js

AssertionError: message mismatch: expected 'inaccessible' to equal 'repo not found'
    at Function._expectField (core/service-test-runner/icedfrisby-shields.js:69:51)
    at IcedFrisbyNock.<anonymous> (core/service-test-runner/icedfrisby-shields.js:54:26)
    at IcedFrisbyNock.<anonymous> (node_modules/icedfrisby/lib/icedfrisby.js:954:10)
    at invokeNextHook (node_modules/icedfrisby/lib/icedfrisby.js:1003:24)
    at /home/circleci/project/shields/node_modules/icedfrisby/lib/icedfrisby.js:1017:7
    at new Promise (<anonymous>)
    at IcedFrisbyNock._runHooks (node_modules/icedfrisby/lib/icedfrisby.js:976:12)
    at IcedFrisbyNock.run (node_modules/icedfrisby/lib/icedfrisby.js:1276:20)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Context.<anonymous> (node_modules/icedfrisby/lib/icedfrisby.js:1348:9)
```

:bulb: **Possible solution**
Not sure yet

calebcartwright:
Spent some time looking into this one, and seems to be a pretty frustrating issue. 

We have to do svg scraping for this badge (the only GitLab badge we currently support) due to GitLab's API rate limiting and the absence of any complex workarounds (token pool). 

[...]

PR #5347, opened 2020-07-22
bypass checkErrorResponse on [website] badge

### LINKED ISSUE #5291: Website service failed with a status code of 204

- [ ] My own instance

:beetle: **Description**

My website responed 204 to HEAD request, and the website service shown 'down'.

:link: **Link to the badge**

https://img.shields.io/website?label=backend&logo=nestjs&url=https%3A%2F%2Fehtt.azurewebsites.net%2Fdatabase

:bulb: **Possible Solution**

I've debuged the code and got following error
https://github.com/badges/shields/blob/964ad087674edf57ae329d1b1aecf5a7c10cefc7/services/website/website.service.js#L80-L95
```
InvalidResponse: Invalid Response: Got status code 204 (expected 200)
buffer:""
defaultPrettyMessage:"invalid"
message:"Invalid Response: Got status code 204 (expected 200)"
name:"InvalidResponse"
prettyMessage:"invalid"
response:IncomingMessage {_readableState: ReadableState, readable: false, domain: null, …}
stack:"Error: Got status code 204 (expected 200)
    at ~\shields\core\base-service\check-error-response.js:16:26
    at Website._request (~\shields\core\base-service\base.js:228:45)
    at process._tickCallback (internal/process/next_tick.js:68:7)"
```

chris48s:
closes #5291
I don't see any reason why this _should_ cause problems. If it does, somebody is not HTTP-ing properly, but I want to do a full service test run juuuuust in case


chris48s:
only failure is the one identified in #5282 :tada:

calebcartwright:
> If it does, somebody is not HTTP-ing properly, but I want to do a full service test run juuuuust in case

But my HTTP 418 is failing now :laughing:

chris48s:
Whatever we do here we're definitely addressing a corner case. We have hundreds of integrations and this is the first time we've ever really had the need to accept a non-200 2XX status and treat it as anything - either valid or invalid.

The 3XX case is kind of academic as we will call with `followRedirects` and we'll have already followed the 3XX by this point.

> This still feels a little weird to me

On the one hand, its completely normal. 200-299 is defined as success, so treating them as success is entirely conventional.
On the other hand, you are totally right that a lot of 2XX statuses would be unusual or unexpected given our use-case. For example, a `201 Created` is totally normal if you've just submitted a `POST` to `/create` but I wouldn't expect us to issue a request like that in the context of shields.

One thing we can do is (like you say) keep the existing logic, just make `Website` a special case and give it a custom `_request()` method which replaces the behaviour of `checkErrorResponse()` with a special case for that badge and we'll review it if it ever comes up again.

Another option could be do modify what I've done here to process `errorMessages` first. Then if there is some 2XX status that may be issued by an upstream API which we explicitly want to treat as a failure in some context, we can set

`errorMessages: { 202: 'API taking too long to respond' }`

but otherwise its treated as valid. Then we have flexibility and if in some future case we do need to accept a `204` or something for another service badge, we can. Both scenarios are unusual, but not completely inconceivable.

paulmelnikow:
> Another option could be do modify what I've done here to process `errorMessages` first. Then if there is some 2XX status that may be issued by an upstream API which we explicitly want to treat as a failure in some context, we can set
> 
> `errorMessages: { 202: 'API taking too long to respond' }`
> 
> but otherwise its treated as valid. Then we have flexibility and if in some future case we do need to accept a `204` or something for another service badge, we can. Both scenarios are unusual, but not completely inconceivable.

This is a nice idea. Basically the idea is we'll only accept some strange 2xx or 3xx code if it's asked for by the service.

Though I'm not sure it really solves the Website issue, does it?

chris48s:
I was thinking the other way round - we treat any 2XX as _valid_ by default, but we can choose to handle override that and treat specific "valid" codes as an error if we want

paulmelnikow:
Oh, I see. That makes less sense to me, as I think a weird 2xx response is probably due to a transient error or a CDN misconfiguration. We haven’t seen an API up to now that deliberate signals an error using a 201 or a 204.

chris48s:
At this point, both cases are essentailly fictional. We'll override `_request()` in website for now. Next time it comes up we can review

Issue #5259, opened 2020-06-26
F-Droid badge displays new version too early


kitsunyan:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [badge-maker NPM package](https://www.npmjs.com/package/badge-maker)

:beetle: **Description**

Basically, there are the following steps in F-Droid build process:

1. `checkupdates` — checks for updates, alters metadata, and commits it to the fdroiddata repo, ~6h delay after release.
2. `build` — builds packages from metadata, ~24h delay.
3. `deploy` — publishes new packages when build is complete, ~12h delay.

F-Droid version is fetched from metadata in repository, which means it will display a new version about a day and half earlier than the package is actually published.

I'm using this badge to show the difference between GitHub release and F-Droid release. But it doesn't make any sense because F-Droid release gets "updated" in 6 hours after GitHub release, but the package wasn't actually updated in F-Droid yet and won't be updated near ~24 hours or even more.

:link: **Link to the badge**

F-Droid release: ![](https://img.shields.io/f-droid/v/nya.kitsunyan.foxydroid)

:bulb: **Possible Solution**

I'm not quite sure how to do this best. It's possible to parse the response from the [web page](https://f-droid.org/en/packages/nya.kitsunyan.foxydroid/), but that's not a good solution. Another solution is parsing the [old XML repo index](https://f-droid.org/repo/index.xml), but it's a really huge XML file to parse.

PyvesB:
Hello @kitsunyan, thanks for flagging this up.

> I'm not quite sure how to do this best. It's possible to parse the response from the [web page](https://f-droid.org/en/packages/nya.kitsunyan.foxydroid/), but that's not a good solution. Another solution is parsing the [old XML repo index](https://f-droid.org/repo/index.xml), but it's a really huge XML file to parse.

Yes I think you're right by saying that these would not be good solutions as they would significantly slow down badge rendering and add complexity.

@niccokunzmann do you know if any progress was made on the F-Droid API that we requested a couple of years ago? Do you have any other ideas to solve this in a simple and efficient way?

niccokunzmann:
I could add something: The web page may also not be what the Android apps use, I think I read something about this.
The release process is in my view:
(1) metadata repository
(2) build
(3.1) website build in parallel to (3.2) index creation for apps (3.3) wiki page generation, ...

So, there are several stages.

I think about these steps and will not take them:
[...]

PR #5362, opened 2020-07-25
Migrate to [F-Droid] API

kitsunyan:
Reference: https://github.com/badges/shields/issues/5259 (this PR shouldn't close it yet)

F-Droid is working on the API which will be published at the official website https://gitlab.com/fdroid/fdroid-website/-/merge_requests/568. It will take some time to discuss everything and extend it for other needs.

For now I prototyped this API using GitLab CI, which uses the same data structures and works better even in the current state. In the future (1-2 weeks, I guess) when the official API will be released the endpoint will be changed.

This new badge deprecates `metadata_format` parameter, since metadata is not parsed anymore. Instead, ~~`suggested`~~ `include_prereleases` parameter is added. You can see the difference at `org.fdroid.fdroid.privileged` package: if ~~`suggested` is used~~ `include_prereleases` is not specified, the badge will display suggested version instead of the latest, which is the case when latest version is alpha/beta/etc and wasn't marked as suggested in F-Droid metadata.


services/f-droid/f-droid.tester.js
```diff
@@ -8,165 +8,77 @@ const t = (module.exports = new ServiceTester({
...
+  .get('/v/org.fdroid.fdroid.privileged.json')
+  .intercept(nock => nock(base).get(path).reply(403, 'some 403 text'))
+  .expectBadge({ label: 'f-droid', message: 'app not found' })
 
-t.create('Package is not found with "metadata_format" query parameter')
-  .get('/v/axp.tool.apkextractor.json?metadata_format=yml')
-  .intercept(nock => nock(base).get(`${path}.yml`).reply(404))
+t.create('Package is not found with 404')
+  .get('/v/org.fdroid.fdroid.privileged.json')
+  .intercept(nock => nock(base).get(path).reply(404, 'some 403 text'))
```
PyvesB, line 53: `some 404 text`? Alternatively, you could make this test hit the real API.

services/f-droid/f-droid.tester.js
```diff
@@ -8,165 +8,77 @@ const t = (module.exports = new ServiceTester({
...
 
-t.create('Package is not found due invalid metadata format')
-  .get('/v/axp.tool.apkextractor.json?metadata_format=xml')
-  .expectBadge({
-    label: 'f-droid',
-    message: 'invalid query parameter: metadata_format',
-  })
-
 /* If this test fails, either the API has changed or the app was deleted. */
 t.create('The real api did not change')
```
PyvesB, line 83: This test has a `timeout` of 10 seconds. Can it be removed, i.e. is the new API fast enough?

services/f-droid/f-droid.service.js
```diff
@@ -55,64 +64,31 @@ module.exports = class FDroid extends BaseYamlService {
...
-      result = await this.fetchYaml(url, fetchOpts)
+  async handle({ appId }, { suggested }) {
+    const json = await this.fetch({ appId })
+    const svc = suggested !== undefined && json.suggestedVersionCode
+    const filtered = (json.packages || []).filter(
+      ({ versionCode }) => !svc || versionCode <= svc
+    )
+    const packages = filtered.length === 0 ? json.packages : filtered
+    if (!packages || packages.length === 0) {
+      throw new NotFound({ prettyMessage: 'app not found' })
```
PyvesB, line 87: Would `no packages found` be a better error message in this case?

services/f-droid/f-droid.service.js
```diff
@@ -1,19 +1,28 @@
...
+  suggestedVersionCode: optionalNonNegativeInteger,
+  packages: Joi.array().items({
+    versionName: Joi.string().required(),
+    versionCode: nonNegativeInteger,
+  }),
 }).required()
 
 const queryParamSchema = Joi.object({
-  metadata_format: Joi.string().valid('yml', 'txt'),
+  suggested: Joi.equal(''),
```
PyvesB, line 22: Is it worth documenting this parameter in the second `F-Droid (latest by suggested)` example?

For example, here's another badge that has some documentation: https://github.com/badges/shields/blob/master/services/discord/discord.service.js

services/f-droid/f-droid.service.js
```diff
@@ -55,64 +64,31 @@ module.exports = class FDroid extends BaseYamlService {
...
-    return { version: match[1] }
-  }
-
-  async handle({ appId }, { metadata_format: metadataFormat }) {
-    const url = `https://gitlab.com/fdroid/fdroiddata/raw/master/metadata/${appId}`
-    const fetchOpts = {
-      options: {},
+      url,
       errorMessages: {
+        403: 'app not found',
```
PyvesB, line 73: When do we expect to get 403 response codes?

services/f-droid/f-droid.tester.js
kitsunyan: Yes, missed it when copypasted the test.
I thought it's better to avoid real API because who knows, I chose a package name which doesn't exist, and someone will decide to publish an application with the same package name in the future :eyes:

services/f-droid/f-droid.tester.js
kitsunyan: For now it's moved from gitlab to gitlab. Later it will move to f-droid.org, which is actually even slower than gitlab. I actually don't really see timeout is used for other tests, so I can remove it, I just kept is there because it already was set.

services/f-droid/f-droid.service.js
kitsunyan: Actually, there should not be cases when server responds with a valid json but without `packages` at all. So it's a sanity check, in which case default error message is returned. I can change if you want.

services/f-droid/f-droid.service.js
kitsunyan: Sometimes server may reply with 403 when API will be moved to f-droid.org. I can't imagine a "real" cases when it will reply this, but it's technically possible due to the nature of the API: it's not a real "REST" API, but instead it's just a bunch of statically generated files.

services/f-droid/f-droid.service.js
kitsunyan: So, I decided to move to this messages because I decided to change `suggested` with `include_prereleases`. So it will be normal when no packages found (this is the case for `org.fdroid.fdroid.privileged.ota`).

services/f-droid/f-droid.service.js
kitsunyan: Instead of documenting this option, I decided to switch to `include_prereleases` instead of `suggested`. This option is negative to another, and it's more clear and already widely used in other badges.

services/f-droid/f-droid.tester.js
PyvesB: If the tests consistently pass without the timeout, I suggest removing it.

services/f-droid/f-droid.service.js
PyvesB: Makes sense. 👍🏻

services/f-droid/f-droid.service.js
PyvesB: Indeed more clear, no documentation needed.

services/f-droid/f-droid.service.js
PyvesB: I've got no strong opinion. If you want to keep the 403 case, let's leave it alone. :)

services/f-droid/f-droid.tester.js
PyvesB: I would argue that there's little risk that someone publishes an app called "io.shiels.does.not.exist", and worse case scenario, we can trivially fix the test. Using the real API at least proves that a 404 is indeed being returned when an app is not found and gives us a little more confidence about the implementation. :)

PyvesB:
Tested a few example on the review app: everything looks good on that side. 👍🏻

services/f-droid/f-droid.tester.js
kitsunyan: Makes sense. Switched to live test.

services/f-droid/f-droid.tester.js
kitsunyan: For now it passes. When API will be moved to f-droid.org, it may be needed to bring it back. f-droid.org it known to be slow sometimes.

services/f-droid/f-droid.service.js
```diff
@@ -1,19 +1,28 @@
...
+  suggestedVersionCode: optionalNonNegativeInteger,
+  packages: Joi.array().items({
+    versionName: Joi.string().required(),
+    versionCode: nonNegativeInteger,
+  }),
 }).required()
 
 const queryParamSchema = Joi.object({
-  metadata_format: Joi.string().valid('yml', 'txt'),
+  include_prereleases: Joi.equal(''),
```
calebcartwright, line 22: Does this change of query params have the potential to impact the message that could be displayed on any existing user badges?

services/f-droid/f-droid.service.js
```diff
@@ -55,64 +64,30 @@ module.exports = class FDroid extends BaseYamlService {
     }
   }
 
-  async fetchYaml(url, options) {
-    const yaml = await this._requestYaml({
+  async fetch({ appId }) {
+    const url = `https://gitlab.com/kitsunyan/fdroidapi/raw/master/applications/${appId}`
```
calebcartwright, line 68: You mentioned in the PR description that this endpoint will be changing once this endpoint gets moved under F-Droid as an official API. What will happen with this endpoint once that transition occurs? Will it redirect to the new endpoint, or will it continue to work for some period after the transition, or will it be a hard cut over? 

Just want to make sure that this doesn't break on us in a couple weeks when that transition occurs

services/f-droid/f-droid.tester.js
```diff
@@ -8,169 +8,74 @@ const t = (module.exports = new ServiceTester({
...
-  .intercept(nock => nock(base).get(`${path}.yml`).reply(404))
-  .expectBadge({ label: 'f-droid', message: 'app not found' })
-
-t.create('The api changed')
-  .get('/v/axp.tool.apkextractor.json?metadata_format=yml')
-  .intercept(nock => nock(base).get(`${path}.yml`).reply(200, ''))
+t.create('Package is not found with invalid response"')
+  .get(`/v/${testPkg}.json`)
+  .intercept(nock => nock(base).get(path).reply(200, '{"field":"value"}'))
   .expectBadge({ label: 'f-droid', message: 'invalid response data' })
```
calebcartwright, line 74: We don't typically have mocked service tests that inject a schema-failing response. I suppose I don't feel too strongly about it, but just sharing for awareness in case you ever find yourself working in the Shields codebase again :smile:

services/f-droid/f-droid.service.js
```diff
@@ -55,64 +64,30 @@ module.exports = class FDroid extends BaseYamlService {
...
+      ({ versionCode }) => !svc || versionCode <= svc
+    )
+    if (packages.length === 0) {
+      throw new NotFound({ prettyMessage: 'no packages found' })
     }
-
-    return this.constructor.render(result)
+    const version = packages.reduce((a, b) =>
+      a.versionCode > b.versionCode ? a : b
+    ).versionName
```
calebcartwright, line 90: FWIW a common pattern with the service classes is to extract this sort of logic into a separate `transform` function that's responsible for interrogating the raw response to extract and transform the desired data points. Besides the separation of concerns, an additional benefit is that it's a lot easier to test a `transform` function with standard unit tests vs. trying to setup all the mocks with service tests. 

Don't feel like you need to change this though, mostly sharing for awareness

calebcartwright:
I left a comments/questions inline, but don't have any blocking issues with this. Thanks for all your work on this @kitsunyan (as well as the new upstream API) and for reviewing @PyvesB!

services/f-droid/f-droid.service.js
kitsunyan: In the worst case "package not found" error will be displayed for them. This might be the case for those who publish only beta versions into F-Droid and use the badge (I guess there no such cases).

I think the fact badge is moved from repo metadata to API can be worse, e.g. it won't work for archived applications (which doesn't have packages in the main repository), however they're also not available via F-Droid website.

Removing `metadata_format` shouldn't impact at all.

services/f-droid/f-droid.service.js
kitsunyan: Since I'm the one who maintain this API prototype, I'll maintain it until shields.io will be migrated to the official API.

services/f-droid/f-droid.tester.js
kitsunyan: So, I can just remove it, right?

services/f-droid/f-droid.service.js
calebcartwright: I'm not sure I followed all that (I know nothing about F-Droid :smile:). IIUC, the query param change won't have any impact but there's a very small risk that the change of data source/API could break existing badges in a very rare set of circumstances, and we wouldn't anticipate any such existing badges.

That SGTM

services/f-droid/f-droid.service.js
calebcartwright: That's fantastic. We'll try to keep an eye on the upstream incorporation of the API, but please feel free to drop us a note (either on a GitHub thread or in Discord) when that happens

PR #5387, opened 2020-08-01
VS Code Marketplace Release Date Badge added along with tests, run [VisualStudioMarketplace]

YashTotale:
## New badge for the VS Code Marketplace: Release Date.

Preview

_Tests Included_

PR #5388, opened 2020-08-01
VS Code Marketplace Last Updated Badge added along with tests, run [VisualStudioMarketplace]

YashTotale:
## New badge for the VS Code Marketplace: Last Updated
Preview

_Tests Included_


calebcartwright:
This looks to also contain the Release Date badge commits from #5387. These two should be sufficiently small to address in a single PR, so I am going to close #5387 and track the addition of both within this PR. If you'd still rather have them as separate PRs, then please be sure to make them truly distinct/remove duplicative changes and commits

YashTotale:
> This looks to also contain the Release Date badge commits from #5387. These two should be sufficiently small to address in a single PR, so I am going to close #5387 and track the addition of both within this PR. If you'd still rather have them as separate PRs, then please be sure to make them truly distinct/remove duplicative changes and commits

Yeah, I think having both in a single PR would be better.

services/visual-studio-marketplace/visual-studio-marketplace-base.js
```diff
@@ -27,6 +27,8 @@ const extensionQuerySchema = Joi.object({
                 )
                 .min(1)
                 .required(),
+              releaseDate: Joi.string().required(),
+              lastUpdated: Joi.string().required(),
```
calebcartwright, line 31: Are both of these field guaranteed to always be there for all packages, and never be null or undefined? `releaseDate` feels like a safe bet, but want to make sure that `lastUpdated` is there immediately after the initial publish and not just null until the first _update_

services/visual-studio-marketplace/visual-studio-marketplace-last-updated.tester.js
```diff
@@ -0,0 +1,79 @@
...
+            ],
+          },
+        ],
+      })
+  )
+  .expectBadge({
+    label: 'last updated',
+    message: 'april 2013',
+    color: 'red',
+  })
```
calebcartwright, line 50: This looks like a mocked version of the live test above, so let's remove it. Our service tests are largely a suite of integration tests with the myriad upstream services and platforms Shields integrates with, so we try to be mindful about having duplicative tests within here. 

I know some of the other VS Marketplace service tests have similar ones, but we don't need to mimic that here.

services/visual-studio-marketplace/visual-studio-marketplace-last-updated.tester.js
```diff
@@ -0,0 +1,79 @@
...
+    label: 'last updated',
+    message: 'extension not found',
+  })
+
+t.create('invalid')
+  .get('/visual-studio-marketplace/last-updated/yasht.terminal-all-in-one.json')
+  .intercept(nock =>
+    nock('https://marketplace.visualstudio.com/_apis/public/gallery/')
+      .post(`/extensionquery/`)
+      .reply(401)
```
calebcartwright, line 73: What circumstances result in us receiving a 401 response status code?

services/visual-studio-marketplace/visual-studio-marketplace-release-date.tester.js
```diff
@@ -0,0 +1,79 @@
...
+            ],
+          },
+        ],
+      })
+  )
+  .expectBadge({
+    label: 'release date',
+    message: 'april 2013',
+    color: 'red',
+  })
```
calebcartwright, line 50: Same comment as the similar test in the last updated service tester file, let's remove this

```suggestion
```

services/visual-studio-marketplace/visual-studio-marketplace-release-date.tester.js
```diff
@@ -0,0 +1,79 @@
...
+    label: 'release date',
+    message: 'extension not found',
+  })
+
+t.create('invalid')
+  .get('/visual-studio-marketplace/release-date/yasht.terminal-all-in-one.json')
+  .intercept(nock =>
+    nock('https://marketplace.visualstudio.com/_apis/public/gallery/')
+      .post(`/extensionquery/`)
+      .reply(401)
```
calebcartwright, line 73: Same question here on circumstances that cause a 401

services/visual-studio-marketplace/visual-studio-marketplace-base.js
YashTotale: Yes, `lastUpdated` is there right after the initial publish. You can try it out using this extension I found with only 1 release: `tscpp.node-dependencies`.

services/visual-studio-marketplace/visual-studio-marketplace-last-updated.tester.js
YashTotale: Got it! It's removed.

services/visual-studio-marketplace/visual-studio-marketplace-release-date.tester.js
YashTotale: Removed

services/visual-studio-marketplace/visual-studio-marketplace-last-updated.tester.js
YashTotale: Removed 401 tests

services/visual-studio-marketplace/visual-studio-marketplace-release-date.tester.js
YashTotale: Removed 401 tests.

PR #5390, opened 2020-08-03
VS Code Marketplace Publisher Badges added along with tests

YashTotale:
### Publisher ID Badge
Publisher ID Badge

### Publisher Badge
Publisher Bage

_Tests Included_


calebcartwright:
Thank you for the PR! However, I'm wondering what the value in having this as a dynamic badge really is, when users could already use the static badge to display their publisher ID should they want to do so.

https://img.shields.io/badge/publisher%20id-Yash%20T-brightgreen


AIUI, your publisher id is _essentially_ static except in the rarest of circumstances, but even in order to get it change you have to contact Microsoft directly.

YashTotale:
Oh shoot. I completely forgot about static badges. Sorry about that. 😆

calebcartwright:
No worries!

PR #5391, opened 2020-08-03
Changed [LGTM] Alerts Badge

YashTotale:
## Changes made to LGTM Alerts Badge

The LGTM Alerts badge has the label "lgtm" and the message "(some number) alerts".

However, this design is inconsistent with most other badges. Generally, the label specifies the parameter of the badge and the message is only a value defining that parameter.

**This is how it looked previously:**
Previous Badge

**This is how it looks now:**
New Badge

If a person using the new badge wants to specify that it is an LGTM badge, they can simply change the label of the badge from "alerts" to "LGTM alerts".
 
However, if a person using the old badge wants to change the message to only the number of alerts, they are not able to change "(some number) alerts" to "(some number)".

So, I think having only the number of alerts in the message allows for more flexibility when using the badge.


calebcartwright:
Thank you for the PR! First, just want to note that I've updated the PR title according to [our contributing info on PR titles](https://github.com/badges/shields/blob/master/doc/service-tests.md#pull-requests) so that the service tests will be executed (please keep this in mind for any future PRs!)

Philosophically, I'm in favor of this change, both because it improves consistency and also because the current message is arguably in conflict with our [badge specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md#bad).

However, since this is already an existing and used badge, we have to take into account the impact on existing badge users. I'm also not sure if there's any additional historical context around the implementation of the LGTM service that would be relevant.

Let's see what some of the other maintainers think about this one

services/lgtm/lgtm-alerts.service.js
```diff
@@ -35,7 +41,8 @@ module.exports = class LgtmAlerts extends LgtmBaseService {
 
   static render({ alerts }) {
     return {
-      message: metric(alerts) + (alerts === 1 ? ' alert' : ' alerts'),
+      label: 'alerts',
```
calebcartwright, line 44: FYI it's not necessary to set a static label value when you've already done so in `defaultBadgeData`. In some cases the label value is determined dynamically based on input/fetch response/etc. and so it's necessary to be able to set the `label` as part of rendering. However, in most cases this is a static value and handled via the default badge data :+1:

YashTotale: Got it!

YashTotale:
Maybe I can create another badge for LGTM alerts with the new style. Those using the current badge can continue to use it, and those who want to switch can use the new one.

calebcartwright:
> Maybe I can create another badge for LGTM alerts with the new style. Those using the current badge can continue to use it, and those who want to switch can use the new one.

Thank you for the offer, but no we definitely don't want to have duplicative service classes just to support different default badge messages.

chris48s:
Hi thanks for raising this :+1:

I think we need to label this as "LGTM alerts" rather than just alerts.
i.e: 

We should not consider this a BC break.

In some cases metrics are platform-agnostic. Code coverage is an established metric. Whether we are measuring it on codecov or coveralls, 80% coverage is 80% coverage. We know what it means for a build to pass or fail. Whether it passed or failed on travis or circle isn't the key point.

An "alert" isn't really _a thing_  - its tied to the platform so I think we need to include LGTM in the label in order to explain this value, otherwise its a bit of a 

Although badges should generally be non-promotional, sometimes it is necessary to name a product or service in the label to explain the value.

For example:

 - https://img.shields.io/chrome-web-store/v/ogffaloegjglncjfehdfplabnoondfjo
 - https://img.shields.io/docker/cloud/build/jrottenberg/ffmpeg

etc

There is possibly some debate over whether this is 100% consistent across our hundreds of integrations, but if there are other examples we need to look at, lets start another issue for it.

calebcartwright:
This was discussed earlier today, and Chris' summary above is a great summary of the collective decision/perspective. Glad to see this merged!

PR #5394, opened 2020-08-05
Fix all remaining LGTM alerts

PyvesB:
As promised in https://github.com/badges/shields/pull/5346#discussion_r459001172


gatsby-browser.js
```diff
@@ -12,8 +12,7 @@ function scrollToElementId(id) {
 
 export function onRouteUpdate({ location: { hash } }) {
   if (hash) {
-    if (!redirectLegacyRoutes()) {
-      window.setTimeout(() => scrollToElementId(hash), 10)
-    }
+    redirectLegacyRoutes()
+    window.setTimeout(() => scrollToElementId(hash), 10)
```
paulmelnikow, line 16: This seems like a change in behavior. What’s the rationale?

services/codecov/codecov.tester.js
```diff
@@ -75,7 +75,7 @@ t.create('handles unauthorized error (with api token)')
   .intercept(nock =>
     nock('https://codecov.io/api', {
       reqheaders: {
-        authorization: 'token a1b2c3d4e5f6g7h8',
+        authorization: 'token a1b2c3d4e5f6g7h8', // lgtm [js/hardcoded-credentials]
```
paulmelnikow, line 78: :+1:

frontend/lib/generate-image-markup.ts
```diff
@@ -45,7 +45,7 @@ function quoteAsciiDocAttribute(attr: string | null): string {
   if (attr == null) {
     return 'None'
   } else {
-    const withQuotesEscaped = attr.replace(/"/g, '\\"')
+    const withQuotesEscaped = attr.replace(/"/g, '\\"') // lgtm [js/incomplete-sanitization]
```
paulmelnikow, line 48: What’s the thinking here, that we don’t care about other things that might need to be escaped? Probably worth adding a comment explaining why the ignore is being placed.

gatsby-browser.js
PyvesB: `redirectLegacyRoutes` does not return any value, so `if (!redirectLegacyRoutes())` would always evaluate to true.

frontend/lib/generate-image-markup.ts
PyvesB: There are two aspects here:
* why the ignore is being placed: LGTM is flagging it up as a potential security vulnerability. Here the string values are prepared and returned to the user they want to use their badge in a AsciiDoc document. We're not using the value in any actual processing, so no security risk (e.g. SQL injection, cross site scripting) as far as I can tell.
* why the existing code only escapes quotes: there's a link to the AsciiDoc User Guide at the point of call of this method (line 69 of this same class) which explains that "To include a double quotation mark (") character in a quoted attribute value the the quotation mark must be escaped with a backslash. ".

paulmelnikow: That makes sense to me. (Heh, I think I wrote this code!) Could you add that in a comment?

PyvesB: Done! :)

PyvesB:
> 
> 
> No more red alerts! 🍏 📗 🟢 💚

Indeed:
lgtm-0

PR #5411, opened 2020-08-08
[VisualStudioMarketplace] Trending Badges

YashTotale:
Added Visual Studio Marketplace Trending Badges.

- Trending Daily Badge
- Trending Weekly Badge
- Trending Monthly Badge

### DIFF
--- a/services/visual-studio-marketplace/visual-studio-marketplace-base.js
+++ b/services/visual-studio-marketplace/visual-studio-marketplace-base.js
@@ -43,6 +43,9 @@ const statisticSchema = Joi.object().keys({
   onpremDownloads: Joi.number().default(0),
   averagerating: Joi.number().default(0),
   ratingcount: Joi.number().default(0),
+  trendingdaily: Joi.number().default(0),
+  trendingweekly: Joi.number().default(0),
+  trendingmonthly: Joi.number().default(0),
 })
 
 module.exports = class VisualStudioMarketplaceBase extends BaseJsonService {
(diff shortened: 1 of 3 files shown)


calebcartwright:
Thank you for the PR. Do you know if these `trending*` fields are based on the initial install measure, or if they also include updates? I imagine we're going to need to incorporate these as variants within our existing download/install badges vs. maintaining entirely separate classes and routes. 

YashTotale:
I'm not sure what the trending fields are based or what they are out of...

calebcartwright:
@YashTotale - In order to move forward with this, we're going to need at least two things:

* Determine whether these data points are associated with Downloads or Installs. We'll need this in order to be able to properly identify variants in the route param, since the API provides both of those numbers in total and we provided badges for both.
    * I suspect the easiest way to do this would be to look at the API response for a handful of packages. If any of the trending data points exceed the value for the install total, then we know the trends are for downloads
* This should not be implemented as a separate service with a non-compliant badge route. These additional variants need to be incorporated in the [existing service](https://github.com/badges/shields/blob/master/services/visual-studio-marketplace/visual-studio-marketplace-downloads.service.js), and will need to follow [our route conventions](https://github.com/badges/shields/blob/master/doc/badge-urls.md) as I mentioned earlier. This would likely end requiring adding a redirector to map the old `d` to `dt` (or `i` to `it` depending on which metric these map to)

PyvesB:

If someone else wants to pick this up, you can simply fetch the commits by running git fetch origin pull/5411/head:pr-5411. Address the pending review comments and open a new pull request with a co-authored trailer included in the commit message to give attribution to the original author.

PR #5432, opened 2020-08-15
auto-close 'version bump only' gatsby deps

chris48s:
As discussed in https://github.com/badges/shields/pull/5405 I have pushed an update to approve-bot which will attempt to detect gatsby-plugin releases where every changelog entry is "Version bump only for package X"

diff: https://github.com/chris48s/approve-bot/compare/1.0.0...2.0.0

This PR updates to use that.


calebcartwright:
`isPointlessGatsbyBump` might be my favorite function name of all time :laughing:

PyvesB:
> 
> 
> `isPointlessGatsbyBump` might be my favorite function name of all time 😆

Agreed!

Issue #5433, opened 2020-08-16
Offset Earth has been renamed


PyvesB:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [badge-maker NPM package](https://www.npmjs.com/package/badge-maker)

:beetle: **Description**

The service tests for Offset Earth have been failing for the past few days, and the example on our website now returns "username not found". Looking into things a bit more closely, it seems that [Offset Earth has been renamed to Ecologi](https://ecologi.com/articles/updates/ecologi-the-new-name-for-offset-earth).

:link: **Link to the badge**

https://img.shields.io/offset-earth/trees/offsetearth

![Offset Earth (Trees)](https://img.shields.io/offset-earth/trees/offsetearth)

:bulb: **Possible Solution**

A few things need to be done:
* fix the tests so that they use `ecologi` as the username instead of the old `offsetearth` one.
* fix the examples we display on the website in a similar way.
* modify the badge path so that it becomes `https://img.shields.io/ecologi/${username}`
* implement a service redirector to keep compatibility with existing badges ([existing example](https://github.com/badges/shields/blob/master/services/codeclimate/codeclimate-analysis-redirector.service.js) for another service and [its tests](https://github.com/badges/shields/blob/master/services/codeclimate/codeclimate-analysis-redirector.tester.js)).
* potentially change the API endpoint we're hitting. We're currently using https://public.offset.earth, however, https://public.ecologi doesn't (yet?) seem to be a thing. Is there another hostname available and will the old one just stop working at some point?

@remotealex, @owenvoke, @jamesmills, @philsturgeon you were all involved in the original discussions in #4728, could you please shed some light on the last bullet point? Would one of you be willing to help us out and pick up this piece of work? :)

owenvoke:
I'm happy to work on this when I'm home tonight. 👍🏻

owenvoke:
Added in https://github.com/badges/shields/pull/5434, hope that's all correct but let me know if not.

remotealex:
Hey all! Thanks @owenvoke for rustling up the PR.

Like Owen mentioned, changing everything `offset.earth` to `ecologi.com` should sort it out 👍

PR #5434, opened 2020-08-16
feat: rename [OffsetEarth] to [Ecologi]

owenvoke:
This updates the Offset Earth endpoints to use Ecologi, it also adds redirects for the previous badges.

﻿- Update Offset Earth (Carbon) to Ecologi
- Update Offset Earth (Trees) to Ecologi
- Add redirect for Offset Earth (Carbon)
- Add redirect for Offset Earth (Trees)

Closes https://github.com/badges/shields/issues/5433


owenvoke:
Thanks @PyvesB, updated it now. 👍🏻

PR #5436, opened 2020-08-17
Run on node 12 in production, update test matrix

### LINKED ISSUE #4973: upgrade to node 12 in production

chris48s:
Closes #4973

We've been testing and developing on NodeJS 12 for some time now.
Its time to migrate production to use it too.
This PR updates the config for the production environment and self-hosting users and switches the test matrix to test on node 12/14 instead of 10/12. We will continue to support node 10 in the badgge-maker library while it is still maintained and drop in the next major release.

There's a staging deploy up at https://shields-staging-pr-5436.herokuapp.com/ which Herkou has deployed on node 12:

```
-----> Installing binaries

       engines.node (package.json):  ^12.18.3

       engines.npm (package.json):   ^6.14.7

       

       Resolving node version ^12.18.3...

       Downloading and installing node 12.18.3...

       Bootstrapping npm ^6.14.7 (replacing 6.14.6)...

       npm ^6.14.7 installed
```

We can use this to do a bit of last minute testing if needed but I've been running node 12 locally for months without issue. I assume the same is true of other maintainers too.

TODO: we will need to update branch protection rules to account for updated build names before we merge


chris48s:
I think the Github Actions failure is unrelated to the changes in this PR:

Screenshot at 2020-08-17 20-30-10

.circleci/config.yml
```diff
@@ -163,7 +164,7 @@ package_steps: &package_steps
 jobs:
   npm-install:
```
paulmelnikow, line 165: Not related to this PR, but this job should be removed, right?

package.json
```diff
@@ -250,8 +250,8 @@
     "typescript": "^3.8.3"
   },
   "engines": {
-    "node": "^10.16.3",
-    "npm": "^6.11.3"
+    "node": "^12.18.3",
+    "npm": "^6.14.7"
```
paulmelnikow, line 254: The depcheck script should be updated, too.

PR #5440, opened 2020-08-18
Adopt static fields in [amo] and core

paulmelnikow:
Since we've upgraded production to Node 12 (#5436) we can finally adopt static fields!

This starts the process by updating core and one of the service families.

I've left the static fields alone in the badge library since for now we're targeting Node 10+.

### DIFF
--- a/core/base-service/base-graphql.spec.js
+++ b/core/base-service/base-graphql.spec.js
@@ -12,15 +12,8 @@ const dummySchema = Joi.object({
 }).required()
 
 class DummyGraphqlService extends BaseGraphqlService {
-  static get category() {
-    return 'cat'
-  }
-
-  static get route() {
-    return {
-      base: 'foo',
-    }
-  }
+  static category = 'cat'
+  static route = { base: 'foo' }
 
   async handle() {
     const { requiredString } = await this._requestGraphql({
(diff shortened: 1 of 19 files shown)


core/base-service/base-graphql.spec.js
```diff
@@ -12,15 +12,8 @@ const dummySchema = Joi.object({
...
-    return 'cat'
-  }
-
-  static get route() {
-    return {
-      base: 'foo',
-    }
-  }
+  static category = 'cat'
+  static route = { base: 'foo' }
```
calebcartwright, line 16: Looks like this will let us express things a bit more succinctly :+1: Will we still be able to use eslint to enforce a consistent ordering of these like we did for the functions?

paulmelnikow: Looks like yes!

```
/Users/pnm/code/shields/core/base-service/base-graphql.spec.js
  16:3  error  Expected static property category to come before static property route  sort-class-members/sort-class-members
```

PyvesB (approved):
Looks good to me as well. It does make the code more succinct. 👍🏻

chris48s:
:+1:  for updating the tutorial docs early. Doing this for hundreds of services is going to be a right pain. I wonder if there is s tool that can help automate this process..
If not, we could leave it 6 weeks and this year, we make the _only_ hacktoberfest issue "convert static properties in service classes" :smile:

paulmelnikow:
> Doing this for hundreds of services is going to be a right pain. I wonder if there is s tool that can help automate this process..

There might be an ESLint rule or a custom plugin for it. Or else maybe [codemod](https://github.com/facebook/codemod)?

PR #5442, opened 2020-08-19
[GithubWorkflowStatus] add event query string parameter

### LINKED ISSUE #5439: github-workflow-status: add event parameter
### Feature request

Currently, the github-workflow-status service doesn't support the event parameter:
https://github.com/badges/shields/blob/4059bdcdfa8b901c8c4c2676ad2fca02ff6b4489/services/github/github-workflow-status.service.js#L71-L74

The problem with this is that by not being able to filter for different events, a wrong build status can be returned. An example for this are scheduled workflow runs which cancel themselves if there's no new commit since the last scheduled run. Those canceled runs are being reported as "failed" by the unfiltered badge. Even if Github would report them as "canceled", the returned build status would be wrong, as it's showing a different event.

What I'm asking for is the addition of the `event` query string parameter, in addition to the already implemented `branch` parameter, so that users can filter for `push` events for example.
https://github.com/badges/shields/blob/4059bdcdfa8b901c8c4c2676ad2fca02ff6b4489/services/github/github-workflow-status.service.js#L74

Thank you!

bastimeyer:
This adds support for the `event` query string parameter to the GithubWorkflowStatus service.
Closes #5439 

According to the [Badge URL guidelines](https://github.com/badges/shields/blob/master/doc/badge-urls.md), parameters like `branch`, etc. should be part of the URL's path. Does this also include parameters like `event` here? The reason why I've added this as a query string is that an event can be specified without a branch. Having two optional route path parameters next to each other, namely branch and event, is a bit weird. I've checked other services for similar patterns and couldn't find anything.

Also, to make `npm run test:services -- --only=GithubWorkflowStatus` succeed, I had to fix the workflow name in the tests, as the "Main workflow" doesn't exist on the [`/actions/toolkit`](https://github.com/actions/toolkit/actions) repo (anymore?) that is being used for testing. I've changed the workflow name to "toolkit-unit-tests", one of the existing workflows, and this resolved it. See the following badge requests:

- https://github.com/actions/toolkit/workflows/Main%20workflow/badge.svg
- https://github.com/actions/toolkit/workflows/Main%20workflow/badge.svg?branch=master
- https://github.com/actions/toolkit/workflows/Main%20workflow/badge.svg?event=push
- https://github.com/actions/toolkit/workflows/Main%20workflow/badge.svg?branch=master&event=push
- https://github.com/actions/toolkit/workflows/toolkit-unit-tests/badge.svg
- https://github.com/actions/toolkit/workflows/toolkit-unit-tests/badge.svg?branch=master
- https://github.com/actions/toolkit/workflows/toolkit-unit-tests/badge.svg?event=push
- https://github.com/actions/toolkit/workflows/toolkit-unit-tests/badge.svg?branch=master&event=push

If this should be split up into two commits/PRs because of this change, please tell me. Should the examples also be changed to an existing workflow name?

### DIFF
--- a/services/github/github-workflow-status.service.js
+++ b/services/github/github-workflow-status.service.js
@@ -11,6 +11,10 @@ const schema = Joi.object({
     .required(),
 }).required()
 
+const queryParamSchema = Joi.object({
+  event: Joi.string(),
+}).required()
+
 const keywords = ['action', 'actions']
 
 module.exports = class GithubWorkflowStatus extends BaseSvgScrapingService {
@@ -22,6 +26,7 @@ module.exports = class GithubWorkflowStatus extends BaseSvgScrapingService {
     return {
       base: 'github/workflow/status',
       pattern: ':user/:repo/:workflow/:branch*',
+      queryParamSchema,
     }
   }
 
@@ -33,7 +38,7 @@ module.exports = class GithubWorkflowStatus extends BaseSvgScrapingService {
         namedParams: {
           user: 'actions',
           repo: 'toolkit',
-          workflow: 'Main workflow',
+          workflow: 'toolkit-unit-tests',
         },
         staticPreview: renderBuildStatusBadge({
           status: 'passing',
@@ -47,8 +52,25 @@ module.exports = class GithubWorkflowStatus extends BaseSvgScrapingService {
         namedParams: {
           user: 'actions',
           repo: 'toolkit',
-          workflow: 'Main workflow',
+          workflow: 'toolkit-unit-tests',
           branch: 'master',
+        },
+        staticPreview: renderBuildStatusBadge({
+          status: 'passing',
+        }),
+        documentation,
+        keywords,
+      },
+      {
+        title: 'GitHub Workflow Status (event)',
+        pattern: ':user/:repo/:workflow',
+        namedParams: {
+          user: 'actions',
+          repo: 'toolkit',
+          workflow: 'toolkit-unit-tests',
+        },
+        queryParams: {
+          event: 'push',
         },
         staticPreview: renderBuildStatusBadge({
           status: 'passing',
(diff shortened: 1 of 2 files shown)


calebcartwright:
Thanks for the PR!

> According to the Badge URL guidelines, parameters like branch, etc. should be part of the URL's path. Does this also include parameters like event here? 

`event` should indeed be a query parameter on the route. Optional parameters should almost always be query params, but there's a few exceptions (like `branch`) that are so commonly used that they're made available as a route param instead.

> Also, to make npm run test:services -- --only=GithubWorkflowStatus succeed, I had to fix the workflow name in the tests, as the "Main workflow" doesn't exist on the /actions/toolkit repo (anymore?) that is being used for testing

Interesting, I guess the tests are still working because GitHub is still providing a badge, though it does seem things have been restructured in the upstream test target.

https://github.com/actions/toolkit/workflows/Main%20workflow/badge.svg


> Should the examples also be changed to an existing workflow name?

Thanks! That would indeed be helpful. We strive to have example options that will work when folks plug them in, but it doesn't really matter what the example/test target is.

> If this should be split up into two commits/PRs because of this change, please tell me. 

Definitely one PR, and we always squash on merge so no worries on the commits

services/github/github-workflow-status.service.js
```diff
@@ -11,6 +11,10 @@ const schema = Joi.object({
     .required(),
 }).required()
 
+const queryParamSchema = Joi.object({
+  event: Joi.string(),
```
calebcartwright, line 15: Was wondering if it would be worth a more restrictive schema that had a list of allowed events, though seems like there's quite a few of them and probably not worth it

bastimeyer: My personal opinion is that some of the events are not particularly interesting for a project like shields, so restricting the user input doesn't really make sense and is unnecessary. If the user sets an invalid event name, they will see the "no status" response anyway. But if that's how it's handled in other services with a set of optional parameters, then I could of course add a list of the currently supported events by Github to the schema. If Github changes the events list in the future, you'll have to upgrade this though, adding maintenance burden.
https://docs.github.com/en/actions/reference/events-that-trigger-workflows

calebcartwright: That's always the tradeoff. In this case the tighter restriction mainly just helps users spot typos more easily vs. other cases where it ensures that only values the service class can deal with pass the filter. 

Given the long list I think it's better to go without it, can always reevaluate later

Issue #5455, opened 2020-08-23
Color of text not changeable


subnut:
:clipboard: **Description**
There is no option that I could find at [shields.io](https://shields.io/#styles) that allows to set the background color. This is important in cases where you might want the badge color to be a light one. The default white is not very visible in that case. Or say, someone wants to make the labelColor white!

If you believe that it is unreasonable to expect that, look at this:
![Linux](https://img.shields.io/badge/Linux-FCC624?style=for-the-badge&logo=linux&logoColor=white)
![Linux](https://img.shields.io/badge/Linux-FCC624?style=for-the-badge&logo=linux&logoColor=black)

You yourself judge which penguin looks better. Then look at the text to see if it is as readable as the text of the IDE you love :wink:
Don't you think the text would have been more readable only if the text was _BLACK_ ?

Oh, why I chose _that_ particular color? It was suggested by [simpleicons.org](https://simpleicons.org/?q=linux) for this logo
Which maniac would choose to do this? [ME!](https://subnut.github.io/ncm2-github-emoji/) :sweat_smile: 

<br>

#### TL;DR
Add a new `labelTextColor` attribute (for the left part) and `textColor` (for the right part)

subnut:
> I believe you are correct.

:disappointed: 

> Will not commit to it right now, but I'll keep it in mind

Thanks very much! :smile:

> No worries, though if it's okay with you I suggest we close this issue and leave the others open. :)

Of course! :innocent:

ooliver1:
![image](https://user-images.githubusercontent.com/34910574/198157280-6b011c07-d9dd-43f3-965d-09f9f2837933.png)
I don't know if its just me but this is like, pretty hard to read, at least on my monitor and bad eyes

PR #5467, opened 2020-08-29
Use black font when background color is too light

### LINKED ISSUE #4661: Color of badge text
:clipboard: **Description**

In terms of accessibility it would be great if there would be a threshold of lightness in color so we would get _dark_ text on light badge colors automatically.
They can be extremely hard to read otherwise, which means you need dark and saturated badge colors, which limits the users.

PyvesB:
Fixes #4661.

I've chosen quite a conservative threshold for switching to black text. None of the badges advertised on our homepage will be affected, this functionality will only kick in with very light background colours.

Let me know what you think!


PyvesB:
Cc @subnut, got some time to work on this following your ping in #5455. :)

subnut:
@PyvesB Great!
This made my day! (technically, night. It's 10PM here :sweat_smile: )
Someone actually working on something due to my request!
Thanks again! :smiling_face_with_three_hearts:

subnut:
@PyvesB can there be an option to override the auto-detection? Like ... `textColor=dark` for black and `textColor=light` for the default white ?
I personally think that the auto-detection should stay, but there should also be an option to manually override it.

Why? Well .. in #5455 .. I would have liked the badges having Black text. The current threshold still keeps it white.
Yes, you can always lower the threshold .. but then someone else would complain that his badge is having Black text instead of the White he expected !

So ... I think that the current threshold should be kept as it is. It seems well-thought-of. But it's _always_ the edge cases that annoy users. So why not provide an override for them?

Just my two cents
Cheers!

chris48s:
This is really nice.

https://shields-staging-pr-5467.herokuapp.com/badge/foo-bar-white
https://shields-staging-pr-5467.herokuapp.com/badge/foo-bar-FFFF00

I appreciate the sentiment of setting a very conservative threshold so we don't change the appearance for any of the standard badge colours. Do you have any idea what sort of threshold we'd be looking at to start changing the text colour on something like

https://shields-staging-pr-5467.herokuapp.com/badge/foo-bar-9cf
https://shields-staging-pr-5467.herokuapp.com/badge/foo-bar-aqua

and whether that would start impacting any of the standard colours?


Also did you consider using the dark grey colour we use when we re-colour light logos for social badges (`#333333`) for the text? It might be a bit less harsh than a straight black?

badge-maker/lib/badge-renderers.js
```diff
@@ -10,6 +11,19 @@ function capitalize(s) {
   return `${s.charAt(0).toUpperCase()}${s.slice(1)}`
 }
 
+function colorsForBackground(color) {
+  if (brightness(color) <= 0.8) {
```
chris48s, line 15: It would be clearer what we're doing here if we gave this a name e.g:
```js
const brightnessThreshold = 0.8;
```
We should probably do the same thing for the logo thresholds too - those are also magic numbers.

PyvesB:
> @PyvesB can there be an option to override the auto-detection? Like ... `textColor=dark` for black and `textColor=light` for the default white ?
> I personally think that the auto-detection should stay, but there should also be an option to manually override it.
> 
> Why? Well .. in #5455 .. I would have liked the badges having Black text. The current threshold still keeps it white.
> Yes, you can always lower the threshold .. but then someone else would complain that his badge is having Black text instead of the White he expected !
> 
> So ... I think that the current threshold should be kept as it is. It seems well-thought-of. But it's _always_ the edge cases that annoy users. So why not provide an override for them?
> 
> Just my two cents
> Cheers!

Thanks for your thoughts @subnut! You've touched upon a classic problematic within the Shields project, namely how much customisation should we allow within our badges? Shields makes two main promises, readability and consistency. There's always a careful balance to strike and we need to make sure new customisation options don't impede too much on one or the other - for example we got the balance wrong in #1478.

Here, this PR trades off some consistency (i.e. text is no longer always white) for some readability (i.e. badges are easier to read on light backgrounds). However, I'm concerned that adding a `textColor=dark|light` option for users would further degrade consistency, as the text font color would now possibly be different for the same background color, with no clear gain on the readability side. Yes, users will be able to fine tune some specific edge-cases, but on the other hand there's a likelihood that they'll make a poor decision, copy-paste a badge URL and change the background but forget to tweak `textColor` (this kind of thing happens more often that you would think!) or simply specify the override on a build badge currently showing light green without realising that the badge will become unreadable when their build fails and the badge turns red!

Beyond this debate about readability, consistency and customisation, I think we should be careful when it comes to adding new options, as having too many of them doesn't necessarily help users and may even overwhelm some of them. There are already approximately ten query parameters available for all badges!

If a lot of users end up requesting an override and we can't make them happy by tweaking the threshold, then I'd personally be open to considering such an option, but I'm not keen on implementing it initially as part of this PR. 😉 


> I appreciate the sentiment of setting a very conservative threshold so we don't change the appearance for any of the standard badge colours. Do you have any idea what sort of threshold we'd be looking at to start changing the text colour on something like
> 
> https://shields-staging-pr-5467.herokuapp.com/badge/foo-bar-9cf
> https://shields-staging-pr-5467.herokuapp.com/badge/foo-bar-aqua
> 
> and whether that would start impacting any of the standard colours?

0.75 would make the first one go black, 0.69 the second one. With 0.69 being used, I don't think our standard named colors change, but a few of the default badges displayed on the homepage do, "CII Best Practices Level", "SymfonyInsight Stars" and "9cf" from what I can tell from a quick browse through. With 0.75 only the "9cf" badge on the homepage changes from what I can see. Do you want me to lower the threshold to something in between so that only "9cf" is affected?

> Also did you consider using the dark grey colour we use when we re-colour light logos for social badges (`#333333`) for the text? It might be a bit less harsh than a straight black?

Good point, I've pushed a commit with that change.

chris48s:
Sorry for letting this rot for a bit. The code as it stands looks good. I think we could go to 0.69 no problem (sorry I should have just played with this locally myself).

That said, this PR has got me thinking about the level of contrast on our standard colour schemes from the perspective of accessibility. Some of our standard colour schemes actually score quite poorly and could be improved. For example

Screenshot at 2020-09-06 11-55-53

Screenshot at 2020-09-06 11-56-18

I'm aware this is scope creep, so unless you really want to dive into that now I think we should probably switch the threshold to 0.69 (if you agree with that) and merge this because its an improvement over what we have now. Then as another issue it might be worth reviewing our most commonly used styles from this perspective. We probably can't rely on only brightness to get amazing contrast on every bespoke colour, but if we can use 'presets' to ensure all of these have a good accessibility score that would be beneficial.

PyvesB:
> Then as another issue it might be worth reviewing our most commonly used styles from this perspective. We probably can't rely on only brightness to get amazing contrast on every bespoke colour, but if we can use 'presets' to ensure all of these have a good accessibility score that would be beneficial.

Alternatively there is a [color-contrast-checker](https://www.npmjs.com/package/color-contrast-checker) package available, I'm wondering whether we could make use of that or another similar package.

> I'm aware this is scope creep, so unless you really want to dive into that now I think we should probably switch the threshold to 0.69 (if you agree with that) and merge this because its an improvement over what we have now.

I've pushed the change. Agreed, let's continue to iterate over contrast in a separate issue.

chris48s (approved):

laowantong:
Before:

51002331bf55b8f4d4cc5cd9376a01ba

Now:

image

That just ruined my carefully crafted band-aid badge :cry:

PyvesB:
@laowantong if you make the background a bit darker, the font will automatically swap to white. :wink:

laowantong:
@PyvesB Sure, but the background color would no more evoke a classic band aid. Thanks anyway!

laowantong:
@PyvesB I have just switched to another service, sorry about that.

hugovk:
Thanks, much improved!

# Before

image


# After

image



https://github.com/NaNoGenMo/2019

Issue #5470, opened 2020-08-31
Badge request: Localizely


lzoran:
:clipboard: **Description**

The [Localizely](https://localizely.com/) is a translation management system (TMS) that helps teams to automate, manage, and translate content.

It would be nice to have badges for this platform for showing overall and per-language translation progress.
Translation progress here represents a number of translations that are marked as reviewed.

Examples (static badges):

[![Localizely](https://img.shields.io/badge/localized-98%25-green)](https://app.localizely.com/)
[![Localizely](https://img.shields.io/badge/English%20(US)-100%25-brightgreen)](https://app.localizely.com/)
[![Localizely](https://img.shields.io/badge/German-96%25-green)](https://app.localizely.com/)

:link: **Data**

There is a public API for fetching the data.

The API requires:
- project id
- user read-only API token
- branch name (optional)

API documentation: https://api.localizely.com/swagger-ui.html#/Translation_Status_API

Example (API usage):

```
curl -X GET "https://api.localizely.com/v1/projects/84498921-8d05-4327-9e41-bdda5deb11fe/status" -H  "accept: */*" -H  "X-Api-Token: 158e1dc48121415d9a19408923184117ef40be841f014d72bc72c0198fe66372"
{
  "strings": 5,
  "reviewedProgress": 60,
  "languages": [
    {
      "langCode": "en-US",
      "langName": "English (US)",
      "strings": 5,
      "reviewed": 3,
      "reviewedProgress": 60
    },
    {
      "langCode": "de",
      "langName": "German",
      "strings": 5,
      "reviewed": 3,
      "reviewedProgress": 60
    }
  ]
}
```

:microphone: **Motivation**

It would be nice if we could easily show these badges to the potential users and contributors of the open-source projects. That way we would provide simple insight into the current state of the project and potentially new ideas and contributions.

(Note, in case you agree, I'm willing to implement this.)

chris48s:
Sounds good. We'd be happy to review a PR for this. There's a tutorial for creating new badges at https://github.com/badges/shields/blob/master/doc/TUTORIAL.md

Issue #1495, opened 2018-02-12
CodeClimate badges for private repos


vineyardbovines:
Thanks for the great work all.

Has anyone has been able to get shields' CodeClimate badges working for private repos? I've tried a couple different ways to config the URL, but none work. I know CC has their own badges, but I'm not a fan of them and want all the badges I have to match, style-wise.

PyvesB:
Hello @gretzky ,

Even though I have never actually used CodeClimate, I recently rewrote all the badges relating to this service (see #1387 and #1368), so I may be able to shed some light on the issue and possibly work on a solution. The [API documentation](https://developer.codeclimate.com/#authentication) mentions that a token should be provided for authenticated calls. I'm guessing that doing so would probably allow to retrieve the relevant data from private repositories. With the current implementation, Shield's CodeClimate badges do not support providing a token, hence why all your attempts have failed so far.

Some other services such as CircleCI or CodeCov seem to have an optional token parameter that can be inserted in the badge's URL, I'm guessing we could add similar support for CodeClimate. We would need to have a look at what CodeClimate's token model is and how finely you can tune the token settings, as we wouldn't want sensitive information to be accessible by anyone looking at the badge's URL. :wink:

Cheers,

Pyves

gretzky:
Hey @PyvesB, thanks for the response. I found the CircleCI implementation, which is what I started digging into. I'd be happy to do some more further digging-- I'll make a fork and see if I can pull anything off.

paulmelnikow:
Passing secrets to the shields server is not a great idea. While we try to keep things secure, we don't guarantee it, and would like not to put other people's data at risk. The circle handling is an anomaly. Is is possible we could accomplish this use case through self-hosting, as we do for github?

paulmelnikow:
The codeclimate badge still needs to be refactored, though we've got a good pattern in place for secrets config now, so once that is done, it should be fairly easy to add support for this in self-hosted environments.

PyvesB:
Code Climate has been sunset in favour of the new Qlty Cloud service. Will consequently go ahead and close this issue.

PR #5472, opened 2020-09-01
Add [Localizely] badges

lzoran:
This pull request is related to the [#5470](https://github.com/badges/shields/issues/5470) issue.


services/localizely/localizely-base.js
```diff
@@ -0,0 +1,64 @@
...
+      url: `https://api.localizely.com/v1/projects/${projectId}/status`,
+      options: {
+        qs: { branch },
+        headers: { Accept: 'application/json', 'X-Api-Token': apiToken },
+      },
+      errorMessages: {
+        400: 'bad request',
+        401: 'unauthorized',
+        403: 'forbidden',
+      },
```
calebcartwright, line 54: The custom error message capability exists to allow service classes provide contextual error messages to the end user. The bad request scenario really shouldn't be possible. If there's inputs required by the upstream service, then we need to ensure our badge route, so I wouldn't expect the 401 case to be a real possibility either. 

If there's a desire to provide a custom error message for the 403 case, I'd suggest something that'll shed more light (`invalid API token` or `not authorized for project`, etc.)

services/localizely/localizely-lang-progress.service.js
```diff
@@ -0,0 +1,68 @@
...
+const { InvalidResponse } = require('..')
+const { coveragePercentage } = require('../color-formatters')
+const {
+  BaseLocalizelyService,
+  keywords,
+  documentation,
+  queryParamSchema,
+} = require('./localizely-base')
+
+module.exports = class LocalizelyLangProgress extends BaseLocalizelyService {
```
calebcartwright, line 12: Does this really need to be a separate service class? Seems like we could have a single service class that has `languageCode` as an optional param, and then have logic (probably in a new `transform` function) that will extract correct `reviewedProgress` value in kind, and a similar logic check in the `render` function to set the label accordingly

Something like this perhaps:
```js
  static render({ langName, reviewedProgress }) {
    return {
      label: langName || 'localized',
      message: `${reviewedProgress}%`,
      color: coveragePercentage(reviewedProgress),
    }
  }

  static transform(json, languageCode) {
    if (!languageCode) {
      return { reviewedProgress: json.reviewedProgress } 
    }
    
    const language = json.languages.find(lang => lang.langCode === languageCode)
    if (!language) {
      throw new InvalidResponse({ prettyMessage: 'Unsupported language' })
    }

    const { langName, reviewedProgress } = language
    return { langName,  reviewedProgress }
  }

  async handle({ projectId, branch }, { token: apiToken, languageCode }) {
    const json = await this.fetch({ projectId, branch, apiToken })
    const { langName, reviewedProgress } = this.constructor.transform(json, languageCode)
    return this.constructor.render({ langName, reviewedProgress })
  }
```

services/localizely/localizely-base.js
```diff
@@ -0,0 +1,64 @@
...
+const queryParamSchema = Joi.object({ token: Joi.string() }).required()
+
+class BaseLocalizelyService extends BaseJsonService {
+  async fetch({ projectId, branch, apiToken }) {
+    return this._requestJson({
+      schema,
+      url: `https://api.localizely.com/v1/projects/${projectId}/status`,
+      options: {
+        qs: { branch },
+        headers: { Accept: 'application/json', 'X-Api-Token': apiToken },
```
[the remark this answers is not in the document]
lzoran: Thanks for the feedback!

Maybe I'm missing something here, but according to the `_requestJson` implementation and the [querying-an-api doc](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#43-querying-an-api), headers will be overwritten in case they are provided.

```
// base-json.js
async _requestJson({ schema, url, options = {}, errorMessages = {} }) {
    const mergedOptions = {
      ...{ headers: { Accept: 'application/json' } },
      ...options,
    }
    const { buffer } = await this._request({
      url,
      options: mergedOptions,
      errorMessages,
    })
    const json = this._parseJson(buffer)
    return this.constructor._validate(json, schema)
  }
```

services/localizely/localizely-base.js
lzoran: Suggestion applied within [dcd1d](https://github.com/badges/shields/pull/5472/commits/dcd1d001676e386345e47b46782f60fe5b92339d) commit.

services/localizely/localizely-lang-progress.service.js
lzoran: It make sense!
Changes applied within [6c290e](https://github.com/badges/shields/pull/5472/commits/6c290e732a73ce2be8b163aca4a070f30e7dc741) commit.

services/localizely/localizely-base.js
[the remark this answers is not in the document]
calebcartwright: You're correct that the explicit json mode based that ultimately gets passed to our http client lib (request) does have the json mode wiped out due to the object merging.

However, it's not necessary to specify the header here because the server in this case (Localizely API) will only ever return a json response, so there's no negotiation needed. The base infra will also [handle the parsing of the response](https://github.com/badges/shields/blob/bf98260a4c41e69844a1fcba840a45c5e83d6aac/core/base-service/json.js#L8-L24) and handle any errors accordingly. 

The API doc you linked also states immediately after the referenced text:

> There is no need to specify json, as the JSON parsing is handled by _requestJson()

It's not a big deal one way or the other, so don't feel like you have to remove, just noting it's not explicitly necessary to include it. Some of our other json-based classes with additional headers include Accept anyway and others do not

calebcartwright:
Thanks for the updates @lzoran! I'm not sure what's going on with the CI checks, but I'll try to go through the latest changes tomorrow and see if we can't get Circle unstuck

services/localizely/localizely-base.js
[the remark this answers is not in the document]
lzoran: Applied suggestion for removing the `Accept` header within [9028aa](https://github.com/badges/shields/pull/5472/commits/9028aa1857e4887071e6b4a9f948df597ce44612) commit.

services/localizely/localizely.service.js
```diff
@@ -0,0 +1,135 @@
...
+    queryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'Localizely overall progress',
+      keywords,
+      documentation,
+      namedParams: {
+        projectId: '1349592f-8d05-4317-9f46-bddc5def11fe',
```
calebcartwright, line 60: This is currently returning a not authorized message in the review app. Not required but we generally like to have the example badges use values that'll work when users plug them in. 

What about using a different project?

```suggestion
        projectId: '5cc34208-0418-40b1-8353-acc70c95f802',
```

https://shields-staging-pr-5472.herokuapp.com/localizely/progress/1349592f-8d05-4317-9f46-bddc5def11fe?token=0f4d5e31a44f48dcbab966c52cfb0a67c5f1982186c14b85ab389a031dbc225a


vs.

https://shields-staging-pr-5472.herokuapp.com/localizely/progress/5cc34208-0418-40b1-8353-acc70c95f802?token=0f4d5e31a44f48dcbab966c52cfb0a67c5f1982186c14b85ab389a031dbc225a

services/localizely/localizely.service.js
```diff
@@ -0,0 +1,135 @@
...
+          '312045388bfb4d2591cfe1d60868ea52b63ac6daa6dc406b9bab682f4d9ab715',
+        languageCode: 'en-US',
+      },
+      staticPreview: this.render({
+        langName: 'English (US)',
+        reviewedProgress: 97,
+      }),
+    },
+  ]
+
```
calebcartwright, line 88: Let's include a default label as well

```suggestion

  static defaultBadgeData = { label: 'localized' }

```

services/localizely/localizely.service.js
```diff
@@ -0,0 +1,135 @@
...
+          '312045388bfb4d2591cfe1d60868ea52b63ac6daa6dc406b9bab682f4d9ab715',
+      },
+      staticPreview: this.render({ reviewedProgress: 93 }),
+    },
+    {
+      title: 'Localizely language progress',
+      keywords,
+      documentation,
+      namedParams: {
+        projectId: '1349592f-8d05-4317-9f46-bddc5def11fe',
```
calebcartwright, line 74:
```suggestion
        projectId: '5cc34208-0418-40b1-8353-acc70c95f802',
```

services/localizely/localizely.service.js
```diff
@@ -0,0 +1,135 @@
...
+  token: Joi.string().required(),
+  languageCode: Joi.string().regex(/^[a-z]{2}(-[A-Z][a-z]{3})?(-[A-Z]{2})?$/),
+}).required()
+
+module.exports = class Localizely extends BaseJsonService {
+  static category = 'other'
+
+  static route = {
+    base: 'localizely/progress',
+    pattern: ':projectId/:branch*',
```
calebcartwright, line 50: Do you think there's a chance that one day localizely could make the progress data available anonymously, without requiring the token? 

We usually want [required parameters to be route params](https://github.com/badges/shields/blob/master/doc/badge-urls.md#badge-url-conventions) and at least today, the token is required. However, we have had a case before where a token was once required (and thus we required it as a route param) and then later on the service dropped the token requirement.

I think I could personally live with `token` being a (required) query param here if there is such a possibility, but if not then `token` should actually be a route param again.

cc @badges/shields-maintainers for input from others

calebcartwright (commented):
Thanks for making the updates! A couple minor items left inline, and a question about route vs. query param for token, but otherwise this is just about ready for merging IMO

services/localizely/localizely.service.js
chris48s: In general, with other services where the user can pass a token (e.g: circle, codecov, bitrise, poeditor, etc) the token is passed as a query param.

calebcartwright: Good point. The tokens for circle and codecov are optional (... _may_ provide) whereas the tokens are also **required** for bitrise and poeditor. I think keeping the token as a query param makes sense here as well then to be consistent, but I also think we should update our badge route documentation to make it clear that tokens are the exception to the rule:

> Parameters should always be part of the route if they are required to display a badge e.g: :packageName.

Should instead be something like

> Parameters should always be part of the route if they are required to display a badge e.g: :packageName, with the exception of tokens which should be provided as query parameters

PyvesB: By the way, what does the token allow to do? Can it be used to expose some potentially sensitive data? I know we have a few existing badges that allow users to specify a token as a parameter, but we should probably be a bit careful about expanding the pattern any further. For reference, @paulmelnikow has advised against reusing the Circle/Codecov approach in https://github.com/badges/shields/issues/1495#issuecomment-370195236.

services/localizely/localizely.service.js
lzoran: Examples updated with real values within [a2803c](https://github.com/badges/shields/pull/5472/commits/a2803ccbce8465fb866db905d9e957b892b7caf6) commit.

services/localizely/localizely.service.js
lzoran: Updated within [a2803c](https://github.com/badges/shields/pull/5472/commits/a2803ccbce8465fb866db905d9e957b892b7caf6) commit.

services/localizely/localizely.service.js
lzoran: Included within [a2803c](https://github.com/badges/shields/pull/5472/commits/a2803ccbce8465fb866db905d9e957b892b7caf6) commit.

services/localizely/localizely.service.js
lzoran: Hi @calebcartwright, @chris48s and @PyvesB, 

The Localizely recently introduced the `read-only` API tokens for this case, so I believe that their API won't change.

Regarding security, with the `read-only` API tokens, users can access only to the translation status data and to download translations (already included in the project for localization). 
Having that in mind, I don't see any potential security risk.

calebcartwright (approved):
Thanks @lzoran! Everything LGTM, will hold off on merging for a bit longer to see if there's any blocking concerns on the token usage, otherwise should be good to go!

services/localizely/localizely.service.js
PyvesB: Thanks for the info @lzoran! Based on the information you've provided, I would say that this approach is fine in principle.

In practice, there's still opportunity for things to easily go wrong. I've just created a brand new Localizely account, and here's what I get in my profile section:
Localizely

Al token is automatically created for me and available by default, it has _write_ permissions as well. It's all too easy for users to not pay attention and simply press that copy button to use the default token, without realising that they're leaking write access to their project through the upload API.

At a minimum, I think we should make this explicit in the documentation: putting a clear warning telling users not to use the default token, and instead create a new one with only read permissions. Possibly an additional sentence briefly outlining the concrete security consequences of using a write token for their badge.

lzoran: You are right @PyvesB!
I've updated docs within [086c414](https://github.com/badges/shields/pull/5472/commits/086c41433e58a786b2ced608595a8baaa7de4957) commit.

PyvesB (approved):
Green light from me as well, thanks @lzoran ! 👍🏻

PyvesB:
@lzoran do you have a Twitter handle I could use in a Tweet about this new badge?

lzoran:
> 
> 
> @lzoran do you have a Twitter handle I could use in a Tweet about this new badge?

Yes, of course!
That would be awesome!

You can use [@localizely](https://twitter.com/localizely).

Thanks!

PyvesB:
I was doing some last sanity checks on the staging environment before pushing to production, and I noticed the following badge seems to render as expected:
https://shields-staging.herokuapp.com/localizely/progress/5cc34208-0418-40b1-8353-acc70c95f802/main?token=this-is-not-a-valid-token

Is this expected? I'm a bit confused given that the Localizely API states that "API requests without authentication will fail".

lzoran:
> 
> 
> I was doing some last sanity checks on the staging environment before pushing to production, and I noticed the following badge seems to render as expected:
> https://shields-staging.herokuapp.com/localizely/progress/5cc34208-0418-40b1-8353-acc70c95f802/main?token=this-is-not-a-valid-token
> 
> Is this expected? I'm a bit confused given that the Localizely API states that "API requests without authentication will fail".

Hi @PyvesB,

Yes, it is.

The Localizely recently introduced the read-only API tokens for that.
Before that, users only had read-write tokens, and they were not required for fetching translation status data.
After publishing and promoting this badge, the Localizely will inform all existing users of opensource projects to migrate their dynamic badges to the new one which requires a read-only API token.

PyvesB:
Thanks for clarifying! Badges are now live! 🎉

PR #5474, opened 2020-09-03
[Packagist] Use default branch as indicated by Packagist

ramsey:
Instead of assuming `dev-master` for the default branch, check the package data from Packagist to see which branch is set as the default, and use it.


ramsey:
> This PR modified service code for packagist but not its test code. 

Is this because `packagist-license` and `packagist-php-version` are considered separate services from `packagist`?

Ocramius (approved):

chris48s (approved):
nice. Thanks for picking this last one up

ramsey:
@chris48s Oh, nice! Glad to know I helped close out that other issue. I kind of stumbled on this by accident and wasn't aware of the other effort, so this was serendipitous. 😁

PR #5478, opened 2020-09-05
Deprecate [Codetally] service

PyvesB:
The services tests for Codetally have been failing for many weeks now. The service seems down, I reported the issue early last month (https://github.com/Codetally/codetally_core/issues/2#issuecomment-670851076) but there's been no response since. I suggest we deprecate this badge.


calebcartwright (approved):
Agreed

PR #5493, opened 2020-09-05
[WordPress] API Update & new badges

JoeIzzard:
This PR does a few things, not related to an issue as I didn't open one first.

Firstly, it updates the API version used to the latest version (1.2), and adjusts some schema, services and tests to work with the new version. It also changes how we detect an error as these are now provided in JSON, instead of as plain text.

Next it add more badges where possible. These include:
- WordPress Plugin Issues
    - Open or Closed
    - Raw Value, Percentage (eg 45% open), Out of (20 open out of 100 total) and comparison (20 open, 80 closed)
- WordPress Plugin & Theme Last Updated
- WordPress Theme Required WP Version
- WordPress Plugin & Theme Required PHP Version
- WordPress Plugin & Theme Author Social Badge
- WordPress Plugin Contributor Social Badge

### DIFF
--- a/services/wordpress/wordpress-base.js
+++ b/services/wordpress/wordpress-base.js
@@ -16,6 +16,22 @@ const themeSchema = Joi.object()
     requires: stringOrFalse,
     last_updated: Joi.string(),
     requires_php: stringOrFalse,
+    author: Joi.object()
+      .keys({
+        user_nicename: Joi.string(),
+        profile: Joi.string(),
+        avatar: Joi.string(),
+        display_name: Joi.string(),
+      })
+      .required(),
+  })
+  .required()
+
+const pluginContributor = Joi.object()
+  .keys({
+    profile: Joi.string(),
+    avatar: Joi.string(),
+    display_name: Joi.string(),
   })
   .required()
 
@@ -32,6 +48,9 @@ const pluginSchema = Joi.object()
     support_threads_resolved: nonNegativeInteger,
     requires_php: stringOrFalse,
     last_updated: Joi.string(),
+    author: Joi.string(),
+    author_profile: Joi.string(),
+    contributors: Joi.object().pattern(/^/, pluginContributor),
   })
   .required()
 
--- a/services/wordpress/wordpress-base.js
+++ b/services/wordpress/wordpress-base.js
@@ -13,6 +13,8 @@ const foundSchema = Joi.object()
     active_installs: nonNegativeInteger,
     requires: Joi.string(), // Plugin Only
     tested: Joi.string(), // Plugin Only
+    support_threads: nonNegativeInteger, // Plugin Only
+    support_threads_resolved: nonNegativeInteger, // Plugin Only
   })
   .required()
 
@@ -42,6 +44,8 @@ module.exports = class BaseWordpress extends BaseJsonService {
               tags: 0,
               screenshot_url: 0,
               downloaded: 1,
+              support_threads: 1,
+              support_threads_resolved: 1,
             },
           },
         },
--- a/services/wordpress/wordpress-base.js
+++ b/services/wordpress/wordpress-base.js
@@ -13,7 +13,7 @@ const themeSchema = Joi.object()
     num_ratings: nonNegativeInteger,
     downloaded: nonNegativeInteger,
     active_installs: nonNegativeInteger,
-    requires: Joi.string(),
+    requires: stringOrFalse,
     last_updated: Joi.string(),
     requires_php: stringOrFalse,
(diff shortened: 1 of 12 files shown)


JoeIzzard:
@calebcartwright I have updated my PR, using some of your current PR's as an example; specifically #5500, let me know if I missed anything that needs updating. From what I saw on yours, I have changed:

- category
- route
- examples
- defaultBadgeData

I also updated name as WordPress uses that too.

chris48s:
Thanks for working on this. One of the reasons nobody has reviewed it yet is because there is so much going on in one PR it is very hard to review.

If we were to do one PR to just update the badges that already exist to use the Wordpress v1.2 API and get that reviewed and merged, could all the other "add badge for [some stat]" then be rebased and split out into a number of smaller PRs based on that? That would make it much easier for us to deal with because we could split it across different maintainers at different points in time rather than one person having to find a clear day to grok all this at once.

My preference would be that we proceed as above if at all possible. It does require some extra work, but will make it way easier to actually get this work landed.

That said, I have had a quick skim over what you've done here so far. This isn't a complete review but one broad thing I would observe is that this PR introduces a lot of optional schema parameters (with Joi, everything is optional unless it is explicitly `.required()` so `Joi.string()` is an optional string and `Joi.string().required()` is a required string). It isn't necessarily _wrong_ to have optional params - if the upstream response sometimes returns a key and sometimes doesn't then it is necessary, but where we can't rely on the schema to validate something exists then we have to replicate that validation elsewhere. So for example, if we're going to say `last_updated` is optional ( `last_updated: Joi.string()` instead of `last_updated: Joi.string().required()` ) then the badge that uses the `last_updated` value needs to deal with the situation where `last_updated` is `undefined`, or whatever. Having had a quick read over this it looks like there are probably a bunch of new schema keys that are missing a `.required()` or there is a bunch of manual validation missing to cover the situations where those values are undefined.

calebcartwright:
> My preference would be that we proceed as above if at all possible. It does require some extra work, but will make it way easier to actually get this work landed.

Ditto!

Sorry for the delayed response @JoeIzzard but i've had trouble making the time to really get into this one. The only other thought I'll share is that something felt off about the combination of service classes in the `-social` file, though perhaps that was just due to the quick drive by pass. I guess it's relatively consistent with the other files for this service, but it didn't strike me as very intuitive that the `Author` service classes would be in a file with a `-social` name

JoeIzzard:
I can definitely look at splitting these out over more PR's if that's easier. It mostly spiralled as I started just wanting to add some more badges, then "Oh we should use the new API", "Oh that breaks X" and so on. 

@calebcartwright In terms of Author being in the `social` file, this just came from it being a social badge. I'm happy to move it or reorganise however is preferred. Happy to tackle that whilst splitting this into multiple PR's.

@chris48s So would I need to `.required` every key even if the object as a whole is required? I was mostly following what already existed so it could be I'm missing some testing. The WordPress API is fairly consistent, most of the fields we use I believe should also return something when requested as they must exist on the WP end. Those that don't (For example, `requires` I have set to alternatives with the real data and empty data type.) I have tried to deal with at the schema level. I'm happy to split this out into a new PR, then deal with it all as part of that review in a little more detail?

chris48s:
> I can definitely look at splitting these out over more PR's if that's easier. 

That would be a big help thanks

> So would I need to `.required` every key even if the object as a whole is required?

Yes exactly. Everything is optional unless it is `.required()`. Here's a worked example which hopefully makes things clearer:

```js

const schema = Joi.object()
  .keys({
    obj: Joi.object()
      .keys({
        string1: Joi.string().required(),
        string2: Joi.string(),
      })
      .required(),
  })
  .required()

Joi.assert({}, schema)                                          // fails with '[1] "obj" is required'
Joi.assert({ obj: { string1: 'foo', string2: 'bar' } }, schema) // passes
Joi.assert({ obj: { string1: 'foo' } }, schema)                 // this passes too - string1 is required - string2 is optional
Joi.assert({ obj: { string1: 'foo', string2: 3 } }, schema)     // fails with '[1] "obj.string2" must be a string' - string2 doesn't have to be there, but if it is there, it must be a string
```

In general, we want to nail down as much of our validation using the schema as possible as it cuts down on the amount of manual error handling we need to do, but sometimes we do have to account for keys which are sometimes in the response and sometimes aren't.

JoeIzzard:
That should be everything from this PR now split out into more manageable chunks.

PyvesB:
If my understanding is correct, this pull request is no longer relevant as it's been split into multiple smaller ones. Will close. 😉

PR #5496, opened 2020-09-06
convert some service classes to static props, run [ansible apm appveyor]

calebcartwright:
Not sure why but felt an inclination to tackle a few of these earlier. These conversions can cause some large diffs, so only did a handful of services for easier review


PyvesB (approved):
Looks good to me. 👍🏻

PR #5508, opened 2020-09-09
Rework isCSSColor check

PyvesB:
This PR makes the code in _color.js_ slightly more complex, but gets rid of a dependency both in the main Shields server and the NPM package.

I ran a micro-benchmark to check that performance wasn't affected negatively (careful, you have to change to badge URL in _scripts/benchmark-performance.js_ to be a non-standard Shields colour): the new `isCSSColor` is actually ~10% faster on my machine - though as we're only talking about a few μs, I wouldn't treat this as a strong argument in favour of the PR.

### DIFF
--- a/badge-maker/lib/color.js
+++ b/badge-maker/lib/color.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const isCSSColor = require('is-css-color')
 const cssColorConverter = require('css-color-converter')
 
 // When updating these, be sure also to update the list in `badge-maker/README.md`.
@@ -36,6 +35,13 @@ Object.entries(aliases).forEach(([alias, original]) => {
 const hexColorRegex = /^([\da-f]{3}){1,2}$/i
 function isHexColor(s = '') {
   return hexColorRegex.test(s)
+}
+
+function isCSSColor(color) {
+  return (
+    !Array.isArray(color) &&
+    typeof cssColorConverter(color).toRgbaArray() !== 'undefined'
+  )
 }
 
 function normalizeColor(color) {
--- a/badge-maker/lib/color.js
+++ b/badge-maker/lib/color.js
@@ -39,8 +39,8 @@ function isHexColor(s = '') {
 
 function isCSSColor(color) {
   return (
-    !Array.isArray(color) &&
-    typeof cssColorConverter(color).toRgbaArray() !== 'undefined'
+    typeof color === 'string' &&
+    typeof cssColorConverter(color.trim()).toRgbaArray() !== 'undefined'
   )
 }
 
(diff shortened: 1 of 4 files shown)


chris48s:
Are there any cases which are covered in the `is-css-color` test suite https://github.com/princejwesley/is-css-color/blob/master/test.js that we should add to our tests for `normalizeColor()` just to make sure we're not regressing on a corner case by making this change?

PyvesB:
@chris48s good point!

As far as I can tell from comparing with those unit tests, two edge cases are regressing within `normalizeColor`:
* when the specified color has leading white space. This could be solved with a `trim()` call if we think we should still handle such cases.
* RGBA colors such as `rgba(-100, 20, 111, 1)`, but they don't actually seem to work today anyway, at least on our running Shields.io server: https://img.shields.io/badge/hello-world-rgba(--100%2C%2020%2C%20111%2C%201). Maybe it's simply not supported by SVGs?

PyvesB:
Okay, this has gotten more involved than I expected, I've had to consult the CSS color specifications. After the latest commit and with some further digging:
* colors with trailing or leading whitespace now pass normalisation. No change in behaviour then.
* hex colors with four or eight digits (e.g. `#fffe` are now accepted), whereas they weren't previously. This seems to comply with the [new CSS draft spec](https://www.w3.org/TR/css-color-4/#hex-notation), and renders fine in badges, seems like a good addition to me.
* RGB colors with negative numbers (`rgb(-100, 20, 111)`) now fail normalisation. The [current CSS spec](https://www.w3.org/TR/2018/REC-css-color-3-20180619/#rgb-color) talks about clipping, the [new CSS draft spec](https://www.w3.org/TR/css-color-4/#hex-notation) seems to think they're invalid. They don't render properly today anyway (https://img.shields.io/badge/hello-world-rgb(--100%2C%2020%2C%20111)), so I would say this is a good thing.
* unlike what I said above, RGBA colors do work. The examples used in https://github.com/princejwesley/is-css-color/blob/master/test.js all have a negative value for one of the components, which is why they were failing when I tried them yesterday. Same as RGB colors then.
* RBG colors with a mix of percentages and absolute values now pass normalisation, but render weirdly in an SVG. According to the CSS specification, I don't think this should be allowed, I've raised https://github.com/andyjansson/css-color-converter/issues/4 but wouldn't treat it as a blocker.

chris48s:
> this has gotten more involved than I expected

The devil is in the detail  :japanese_goblin: 

Basically happy to merge this. I _think_ the test failures are unrelated to the changes in this PR but I tried restarting the build and they still failed.

chris48s:
oh yes the same test failures are being discussed in https://github.com/badges/shields/pull/5509

chris48s (approved):

PR #5509, opened 2020-09-10
convert some service classes to static props, run [cdnjs chromewebstore] 

calebcartwright:


paulmelnikow (dismissed):

paulmelnikow (commented):
Hmm, these are odd test failures. Not sure what the fix is, happy to re-:+1:.

calebcartwright:
> Hmm, these are odd test failures. Not sure what the fix is, happy to re-+1.

Yeah they are :thinking: Will probably let it sit for another day and then just try re-running

calebcartwright:
Well, at least these test on v14 are failing consistently (#5511). I can't reproduce anything locally, and the fact things are fine on v12 but timing out on v14 makes me think it could just be an issue with the Circle image in use.

PyvesB (approved):

PR #5512, opened 2020-09-12
Update to nextgen circle images

calebcartwright:
Circle apparently now has "Legacy" and "Next-Gen" images, and we've been using the former. This switches the Node v14 jobs to a Next-Gen Node v14 image because doing so fixes the failing tests on the `{main,integration}@node-14` jobs that are blocking other PRs (refs #5508 #5509 #5511)

https://circleci.com/docs/2.0/circleci-images/#next-gen-language-images
https://circleci.com/docs/2.0/circleci-images/#legacy-language-images

Note that I did set this to use the Node v14.9.0 because the next-gen node v14.10.0 also produces the same mass test time out as the legacy v14 image (my guess is they are the same). The same test are all passing fine for me locally with `v14.10.1`, so my best guess is there's something funky with the Circle v14.10.0 build. 

I'd suggest we go ahead and merge this to avoid blocking all the other PRs, and update in the near future to track the latest/current 14 release once the Circle images are fixed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,7 +170,7 @@ jobs:
 
   main@node-14:
     docker:
-      - image: cimg/node:14
+      - image: cimg/node:14.11
 
     <<: *main_steps
 
@@ -183,7 +183,7 @@ jobs:
 
   integration@node-14:
     docker:
-      - image: cimg/node:14
+      - image: cimg/node:14.11
       - image: redis
 
     <<: *integration_steps
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,7 +170,7 @@ jobs:
 
   main@node-14:
     docker:
-      - image: cimg/node:14.10
+      - image: cimg/node:14
 
     <<: *main_steps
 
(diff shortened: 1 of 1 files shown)


calebcartwright:
And of course the existing images are magically working again :laughing: Will leave this open for a bit for discussion, but it no longer seems like a necessary tactical move

PyvesB:
Does 14.10 now work as well?

PyvesB:
The got update from #5529 contains the following fix:

> Prepare a fix for hanging promise on Node.js 14.10.x 29d4e325b110ccf7571d4265d40760be4175f7ff

I'm wondering whether this may have been another cause of the failures with 14.10.

calebcartwright:
> I'm wondering whether this may have been another cause of the failures with 14.10.

That sounds like a highly likely explanation

> Does 14.10 now work as well?

Doesn't look like it. Browsing the image tags, my sense is that Circle's seems to only be focused on the minor versions for the Node releases. The "legacy" image we're currently using may receive patch updates (~14.10.1 which works fine for me as well) and or it's already been bumped to 14.11.x

https://circleci.com/developer/images/image/cimg/node

calebcartwright:
I'm going to close this one. Can revisit the next-gen images down the road

PR #5522, opened 2020-09-12
Add guidelines about including tokens in badge URLs

PyvesB:
Lack of clarity caused some questioning in #5472, let's document our agreed approach for the future.


CONTRIBUTING.md
```diff
@@ -107,6 +107,9 @@ There are three places to get help:
 - Badges should not obtain data by scraping web pages - these are likely to break frequently.
   Whereas API publishers are incentivised to maintain a stable platform for their users,
   authors of web pages have no such incentive.
+- Badges may require users to specify a token in the badge URL as long it is scoped only to
+  fetching status and doesn't expose any sensitive information. Generating a token with the
```
calebcartwright, line 111: May be worth more general verbiage here to account for other types of data points
```suggestion
  fetching information and doesn't expose any sensitive information. Generating a token with the
```

calebcartwright (approved):
LGTM, one suggested word change left inline, happy to re-:+1: if needed

CONTRIBUTING.md
PyvesB: Yup, makes sense!

calebcartwright (approved):

PR #5533, opened 2020-09-12
Remove camelcase dependency from badge-maker package

PyvesB:
Further slimming down the NPM package, I believe we can do without the _camelcase_ dependency in there.

I haven't benchmarked the change, but probably saves a few CPU cycles.


calebcartwright:
This would technically be a breaking change wouldn't it? I don't have any objections to requiring precise input from consumers (provide `plastic` instead of `PLASTIC` or `for-the-badge` instead of `For-The-Badge`) but I feel like we'd need a major version bump for it

PyvesB:
Actually, I don't believe it is a breaking change, given that the style is validated upstream: https://github.com/badges/shields/blob/bb2f827e000c7c26bbb8bf7b14dbb108e7c3e452/badge-maker/lib/index.js#L26

calebcartwright:
> Actually, I don't believe it is a breaking change, given that the style is validated upstream: 

Yes indeed, thanks for the correction!

calebcartwright (approved):

paulmelnikow (approved):

PR #5534, opened 2020-09-12
Remove duplicate normalizeColor calls

PyvesB:
In the non-JSON case, there is no need to call `normalizeColor`: `toSvgColor` already calls `normalizeColor` internally.


badge-maker/lib/make-badge.js
```diff
@@ -24,17 +24,14 @@ module.exports = function makeBadge({
 
   const [label, message] = text
 
-  color = normalizeColor(color)
-  labelColor = normalizeColor(labelColor)
```
calebcartwright, line 28: Would it be worth adding inline comments below in the json case (and/or on the `toSvgColor` signature) that there's a presumption that normalization will be handled internally by `toSvgColor` just for future reference? 

I know it's unlikely that we'd make a breaking change down the road, but I just find it helpful sometimes to document assumed invariants for future readers

calebcartwright (approved):
:+1:

Issue #4849, opened 2020-04-02
Design question for npm package


paulmelnikow:
Continued from https://github.com/badges/shields/pull/4756#issuecomment-607420151, where @chris48s wrote:

> I think there is a big design question about what the badge-maker package should be which we don't all agree on:
> 
> One way of thinking about it is:
> Option 1 - Badge-maker should be a stand-alone package. It happens to live in the same repo as shields.io, but in principle it could live in its own repository if we wanted. Its a package with many users, but its biggest and main user is shields.io. Any functionality we depend on should be exposed and documented via the public interface of the package and we consume that functionality in the same way as any other user, eating our own dogfood.
> 
> Another way of thinking about it is:
> Option 2 - Badge-maker is a package which exposes only a subset of the badge rendering capibilites of shields.io to a wider audience. It lives in and is inseparable from the shields.io repo. The interface we give our users is different to the interface we consume internally.
> 
> Option 2 is where we are now. I've been thinking we should be heading towards option 1. I'm not necessarily _right_ about that, but that's definitely the direction I've been thinking in. Given that removing a feature that our own application depends on from the public interface of the package actively reverses that direction of travel, I think its worthwhile to be explicit about the reasons for that.

paulmelnikow:
I agree with you that we should be heading toward option 1.

It has the lowest mental overhead, and also means we can harmonis/ze with the Endpoint badge spec. I think that could create a lot of flexibility for users.

It sounds like the gaps are:

1. Links
2. Logos
3. Possibly JSON, if we remove the JSON support as discussed at https://github.com/badges/shields/pull/4756#issuecomment-597782014

Is that right?

chris48s:
Yep - I'd agree with that assessment. Just to flesh those three points out slightly more. In order to move from using relative imports to being a consumer of our own package, we need:

### Links

Exposing links is pretty trivial (although we do need to decide on the public interface for it) its passing through a couple of strings really.

### Logos

Logos is slightly more involved as we need to move the logo code from `/lib` into the package, allow simpleicons to be installed as a peer dependency (if the user wants/needs it) and work out the interface for logos (to deal with simpleicons/custom/base64) in the input object. Doing this would also remove the final remaining relative import to the package requiring anything other than `makeBadge`:

https://github.com/badges/shields/blob/8621fe42d76b809e9ba4ae844920f8ed0373817a/lib/logos.js#L4

[...]

PR #5542, opened 2020-09-14
Expose the links feature in badge-maker

dmatis:
In order to utilize the 'links' feature in badge-maker, 'links' must be added to `expectedKeys`.

### DIFF
--- a/badge-maker/README.md
+++ b/badge-maker/README.md
@@ -67,6 +67,7 @@ The format is the following:
   message: 'passed',  // (Required) Badge message
   labelColor: '#555',  // (Optional) Label color
   color: '#4c1',  // (Optional) Message color
+  links: ['https://github.com/badges/shields'],  // (Optional) Badge links (can provide up to 2 links)
 
   // (Optional) One of: 'plastic', 'flat', 'flat-square', 'for-the-badge' or 'social'
   // Each offers a different visual design.
--- a/badge-maker/lib/index.js
+++ b/badge-maker/lib/index.js
@@ -38,7 +38,14 @@ function _validate(format) {
 }
 
 function _clean(format) {
-  const expectedKeys = ['label', 'message', 'labelColor', 'color', 'style']
+  const expectedKeys = [
+    'label',
+    'message',
+    'labelColor',
+    'color',
+    'style',
+    'links',
+  ]
 
   const cleaned = {}
   Object.keys(format).forEach(key => {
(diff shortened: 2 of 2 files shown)


chris48s:
Refs https://github.com/badges/shields/issues/4949
The main reason this isn't currently exposed in the package is because the way it works at the moment is a bit odd for legacy reasons. See related conversation (mixed in with some other discussion) in https://github.com/badges/shields/issues/4849#issuecomment-609108141
My current thinking on this (that I haven't actually done anything about) is that it would make more sense to expose it as `labelLink` and `messageLink` but as someone (maybe the only person :D ) who actually wants to use this feature for something, you might be a good person to give us some feedback on your use-case..

dmatis:
Hi @chris48s, we're building a dashboard with badges containing the deployed SHA for each of our environments. We would like the `links` feature so that we can click on the badge and redirect to the appropriate Github URL for that SHA. The links aren't static so unfortunately we're unable to statically add them to our Markdown.

We can't use Shields endpoint as described here: https://shields.io/endpoint because it's not accessible outside our VPN (so we need to create our own badge using `badge-maker`).
Screen Shot 2020-09-14 at 10 51 41 AM

chris48s:
So in terms of just purely _getting a thing done_, if the link isn't always the same, you are right in terms of the options available to you. That said, if you're going to stand up your own server to render badges with badge-maker, another option would be to have the process issue a redirect to a custom static badge, or proxy the response.

So at the moment, you've got a function that does _some stuff_ and then runs something like

```js
makeBadge({
  label: 'build',
  message: 'passing',
  color: 'brightgreen',
})
```

and returns that string with a `200 OK` response but then another way to achieve roughly the same end result is to generate the string `build-passing-brightgreen` instead of the JS object `{label: 'build', message: 'passing', color: 'brightgreen' }` and then issue a `302 Redirect` to https://img.shields.io/badge/build-passing-brightgreen (or proxy the response, if you want).
If you did that, then I think that would allow you to just achieve what you're trying to do right now with the `?link` param? 

---

In terms of this feature, if you're keen to work on it I think exposing the public API `labelLink` and `messageLink` probably makes the most sense in the package but its a bigger job than just setting `links = [labelLink, messageLink]` because currently `links[0]` affects the whole badge when `links.length == 1` but only the left side when `links.length == 2`. To make `labelLink` make sense as a param it should only ever affect the label (left) side of the badge but then just making that change in the package is then a BC break for the server so when we call out to `makeBadge()` in the server component we'll need to translate that to

```js 
labelLink = links[0]
messageLink = links[0]
```

if `links.length == 1` to retain the existing behaviour of `?link` for SaaS users. We'll also need tests for all that.

That said, if you _don't_ want to work on it (esp if the above approach solves your immediate issue), this has at least prompted me to think through this issue again so I may end up tackling it.

PyvesB:
@dmatis would you be keen on working on the solution that @chris48s outlined above? If not, I suggest we close this pull request, anyone can still feel free to pick this task up at a later stage. 😉

PR #5547, opened 2020-09-15
[Tokei] Implement a lines of code (LOC) badge

### LINKED ISSUE #2731: Badge request: Lines of code (LOC)
:clipboard: **Description**

It would be great to have a badge that just counts all lines of code (no markdown, or so obviously) and shows that number to the user.

service: GitHub
shows: "101|LOC" or "42|LOC"

:link: **Data**

GitHub, don't know the details here

:microphone: **Motivation**

That is useful for projects that aim to be minimal/small, or artistic stuff that does something in exactly 42 lines of code or any other "special number".

MonliH:
This PR implements a badge that uses [tokei](https://github.com/XAMPPRocky/tokei) to count the lines of code in your repository. It works with GitHub, GitLab, and many others (you can specify the URL you want to use).

Closes #2731, where this was originally requested.


services/tokei/tokei-github-lines.service.js
```diff
@@ -0,0 +1,65 @@
+'use strict'
+
```
calebcartwright, line 2: Few other thoughts below that could impact the desired filename, but either way I think we'll definitely want to drop `github` as there's nothing github specific about the class/badge here

services/tokei/tokei-github-lines.service.js
```diff
@@ -0,0 +1,65 @@
...
+const documentation = `
+<p>
+  The <code>provider</code> is the domain name of git host.
+  If no TLD is provided, <code>.com</code> will be added.
+  For example, setting <code>gitlab</code> or <code>bitbucket.org</code> as the provider also works.
+</p>
+`
+
+module.exports = class LinesOfCode extends BaseJsonService {
```
calebcartwright, line 20: I think it's reasonably likely that we'd get enhancement requests for this that tokei also supports (like `files` for example), so we should probably just use something like `Tokei` for the class name

```suggestion
module.exports = class Tokei extends BaseJsonService {
```

services/tokei/tokei-github-lines.service.js
```diff
@@ -0,0 +1,65 @@
...
+          repo: 'shields',
+        },
+        staticPreview: this.render({ lines: 119500 }),
+        keywords: ['loc', 'tokei'],
+        documentation,
+      },
+    ]
+  }
+
+  static defaultBadgeData = { label: 'total lines' }
```
calebcartwright, line 41: Since both the label and color will be static, we can just utilize default badge data. We should also use `blue` as the default color since this is informational, whereas brightgreen colored badges are used to indicate some kind of pass/success gate (build passing, high test/code coverage %, etc.) 
```suggestion
  static defaultBadgeData = {
    label: 'total lines',
    color: 'blue'
  }
```

services/tokei/tokei-github-lines.service.js
```diff
@@ -0,0 +1,65 @@
...
+
+  static render({ lines }) {
+    return {
+      label: 'total lines',
+      message: metric(lines),
+      color: 'brightgreen',
+    }
```
calebcartwright, line 48:
```suggestion
    return { message: metric(lines) }
```

services/tokei/tokei-github-lines.service.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  async fetch({ provider, user, repo }) {
+    return this._requestJson({
+      schema,
+      url: `https://tokei.rs/b1/${provider}/${user}/${repo}`,
```
calebcartwright, line 54: I think we should add a comment here to speak to the API behavior. All the info on the [tokei_rs](https://github.com/XAMPPRocky/tokei_rs) GitHub page speaks to badges, but it will return json provided the proper request headers.

This isn't a [violation of our guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines) or anything (at least not in my mind) but will be helpful for future observers. Will also be helpful to have a link to the tokei_rs repo since that's where the code for the endpoint is hosted

services/tokei/tokei-github-lines.service.js
```diff
@@ -0,0 +1,65 @@
...
+  The <code>provider</code> is the domain name of git host.
+  If no TLD is provided, <code>.com</code> will be added.
+  For example, setting <code>gitlab</code> or <code>bitbucket.org</code> as the provider also works.
+</p>
+`
+
+module.exports = class LinesOfCode extends BaseJsonService {
+  static category = 'size'
+
+  static route = { base: 'tokei/l', pattern: ':provider/:user/:repo' }
```
calebcartwright, line 23: I think it'd be better to use a more complete noun here, especially since elsewhere in Shields `l` has other meanings (licenses) 
```suggestion
  static route = { base: 'tokei/lines', pattern: ':provider/:user/:repo' }
```

services/tokei/tokei-github-lines.service.js
calebcartwright: I think `tokei.service.js` will probably be the best bet

services/tokei/tokei-github-lines.service.js
calebcartwright: This also applies for me to the file name (it'd be confusing to have `lines` in the file name as well if/when other metrics are supported)

calebcartwright (commented):
Thanks for the PR @MonliH! Few items left inline for your review, but overall looks pretty good. 

One thing we will need to keep in mind is the overall render time, because noncached requests for large repositories have a pretty good chance to cause users to see issues in the github readme environments due to the github/camo time limits. Not really anything we can do about that, though if it's noticeably sluggish we may want to mention it in the example documentation so it'll show to users browsing on Shields.io

services/tokei/tokei-github-lines.service.js
MonliH: Oh yeah, forgot to change that name! It was originally for just GitHub projects but then I decided to expand it a bit. I think `tokei-lines.service.js` might be the better option, though, because tokei can do more than just count lines of code (i.e. count of comments and individual files are also included in the API response). 

Maybe in the future, someone could implement a badge for a specific file as well? (in that case doesn't it make more sense to use `tokei-lines.service.js`?)

services/tokei/tokei-github-lines.service.js
MonliH: Sure, will do

MonliH: I'll also link [this](https://github.com/XAMPPRocky/tokei_rs/issues/8#issuecomment-475071147) in the comment.

services/tokei/tokei-github-lines.service.js
MonliH: Renamed.

services/tokei/tokei-github-lines.service.js
MonliH: Done.

MonliH:
Alright, I've just fixed everything requested in the review, let me know if there's anything else.

calebcartwright:
Thanks! Changes LGTM, but after playing with it a bit I think we may want to add some additional info in the user-facing docs that explain the default behavior and point them to the tokei repo.

I was initially surprised when testing with a few tiny sample repos (only had a couple lines of _code_) to see the badge showing numbers well north of 1k. It took me a while to realize that number was reflecting the lines from the package.json and package-lock.json files in the repo which were causing the inflation and discrepancy between what I was expecting and what I was seeing on the badge. 

Committing the ignore file in the repo made a huge difference for me, and now I'm seeing a count for lines of code that I'd expect:
https://shields-staging-pr-5547.herokuapp.com/tokei/lines/github/calebcartwright/mocha-config-sample


Since I'd anticipate most users won't be intimately familiar with tokei and plenty will have similar expectations, I think it'd be good if we could briefly describe that behavior in the docs and include relevant links, such as https://github.com/XAMPPRocky/tokei#excluding-folders

calebcartwright:
And just to clarify my last comment @MonliH, we'd want to update the documentation in the service class: https://github.com/badges/shields/pull/5547/files#diff-da36ef884f072a90c40ccbf4034c88f0R12-R17

Afterwards this should be good to merge

MonliH:
Yep, that's what I figured, thanks for the clarification though.

MonliH:
Alright, everything should be good! Again, let me know if you would like me to change anything regarding the docs.

calebcartwright (approved):
Awesome, thank you!

PR #4109, opened 2019-10-02
Add a GitHub project badge for Hacktoberfest [GithubHacktoberfest]

paulmelnikow:
Something I've wanted to offer is a nice badge for Hacktoberfest, which projects can use to advertise their participation, encourage contributions, and celebrate new contributions. There was some discussion about this at https://github.com/badges/shields/issues/1519#issuecomment-456267637 and a little bit of chat in Discord today.

Here's a proposal! Would be really curious for comments!

By default this counts open issues labeled **hacktoberfest** but you can pick a different label if you want. By default it counts all PRs merged in October but you can filter that with a label if you want.

This is hard-coded for Hacktoberfest 2019. At some point if people like the badge, we'll want to generalize it, though I think we could wait and do that sometime next year.

Alternatively we could include 2019 in the badge URL, which would make it clear that projects are participating in a given year (as opposed to some past year) and also to display stats from previous years.

PR #5549, opened 2020-09-17
[GitHubHacktoberfest] 2020 (and beyond)

### LINKED ISSUE #5545: Hacktoberfest 2020 Badge
:clipboard: **Description**

The current Hacktoberfest badge is for Hacktoberfest 2019. There's no option to change the year - `2019` is hardcoded.

Screenshot 2020-09-15 at 10 55 34 AM

GitHub Hacktoberfest combined status

:link: **Data**

This badge was already implemented in 2019, I hope the APIs exist with you.

:microphone: **Motivation**

Having a Hacktoberfest countdown badge will be great as [Hacktoberfest 2020](https://hacktoberfest.digitalocean.com/) is approaching us.

calebcartwright:
Resolves #5545 

I decided to remove the 2019 hardcoding, but also constrain the route param for the year to a set of allowed values that only includes 2019 and 2020 (so should this be merged, next year we'd need to add 2021). I also had to take into account what the badge message should be for an upcoming/future Hacktoberfest, and elected to show a countdown of days until Hacktoberfest would begin. It didn't make sense to me pre-Hacktoberfest kickoff for the badge to show the number of days remaining until the end of October (for example, that'd currently have looked something like ), so instead we have:

https://shields-staging-pr-5549.herokuapp.com/github/hacktoberfest/2020/badges/shields

https://shields-staging-pr-5549.herokuapp.com/github/hacktoberfest/2019/badges/shields

One thing I feel is nice with this is that the badge will serve a dual purpose: folks could go ahead and add the 2020 badge to their projects today and it will act as a countdown, and then on October first it will automatically convert to the issue/PR and days remaining count for them.

In this pre-Hacktoberfest start for specified year scenario, I also decided to skip running the query since there's not really any point trying to determine how many PRs were opened during the month of October before October has started. One side effect of this is that if a user adds the 2020 hacktoberfest badge pre-October 1st but has a typo in their org/repo name, then the countdown days will still show without telling the user the repo wasn't found.. until October 1st at which point it'll change to


services/github/github-hacktoberfest.service.js
```diff
@@ -65,8 +65,8 @@ module.exports = class GithubHacktoberfestCombinedStatus extends GithubAuthV4Ser
 
   static get route() {
     return {
-      base: 'github/hacktoberfest/2019',
-      pattern: ':user/:repo',
+      base: 'github/hacktoberfest',
+      pattern: ':year(2019|2020)/:user/:repo',
```
calebcartwright, line 69: As discussed when we first added this badge last year, [we want to keep the year as an explicit part of the route](https://github.com/badges/shields/pull/4109#issuecomment-537642828). I wasn't sure if we wanted to completely open this up just yet, or keep it constrained to a couple years, but obviously opted for the latter.

services/github/github-hacktoberfest.service.js
```diff
@@ -113,12 +116,24 @@ module.exports = class GithubHacktoberfestCombinedStatus extends GithubAuthV4Ser
...
+    suggestedIssueCount,
+    contributionCount,
+    daysLeft,
+    daysToStart,
+    year,
+    hasStarted = true,
+  }) {
+    if (!hasStarted) {
+      return {
+        message: `${daysToStart} ${maybePluralize('day', daysToStart)} til kickoff!`,
```
calebcartwright, line 129: I wasn't really sure what do with the message here, definitely open to suggestions/alternatives

services/github/github-hacktoberfest.service.js
```diff
@@ -59,66 +59,71 @@ const queryParamSchema = Joi.object({
...
+    daysToStart,
+    year,
+    hasStarted = true,
+  }) {
+    if (!hasStarted) {
+      return {
+        message: `${daysToStart} ${maybePluralize(
+          'day',
+          daysToStart
+        )} til kickoff!`,
```
PyvesB, line 119: This spelling caught my eye. Reading on various Internet resources, it seems that `till` should be preferred.

https://www.grammarly.com/blog/until-till-til/
https://writingexplained.org/til-or-till-difference
https://www.merriam-webster.com/words-at-play/should-you-use-until-or-till-or-til

PyvesB (commented):
Thanks for picking this up, mostly looks good, simply a spelling comment inline. 👍🏻

services/github/github-hacktoberfest.service.js
calebcartwright: TIL I learned that til... :laughing:

PyvesB:
> In this pre-Hacktoberfest start for specified year scenario, I also decided to skip running the query since there's not really any point trying to determine how many PRs were opened during the month of October before October has started. One side effect of this is that if a user adds the 2020 hacktoberfest badge pre-October 1st but has a typo in their org/repo name, then the countdown days will still show without telling the user the repo wasn't found.. until October 1st at which point it'll change to 

Didn't mention it in the first pass, but I don't think that side effect is a big problem.

PyvesB (approved):

PR #5574, opened 2020-09-19
Configure max-age header for frontend

chris48s:
Refs #5014

Turned out this did not require a PR to camp, but it took a long time to figure out this small diff :smile: 
What this actually does is apply `frontendMaxAge` to everything, but then for every other request we then overwrite it with subsequent calls to `setHeader()`. I think this is fine though.

I've decided to just serve the front-end with a 5 minute expiry. Its not massively sophisticated, but at the moment we deploy the front-end on GH pages. GH pages just serves everything with a non-configurable 5 min expiry and I don't recall ever seeing any major front-end/back-end out-of-sync bugs (and if there are, they get fixed in 5 mins). Our current deploy strategy means we deploy front-end and back-end a bit out of sync anyway, so.. opening gambit: lets just do that and see what happens :shrug: I guess the concern is not so much front-end/back-end out-of-sync bugs, but maybe taking on the extra traffic/load of serving the front-end with a relatively short max-age might be a performance hit. We have absolutely no idea how much traffic the frontend does, so I guess we just have to try it and see..

The other questions this throws up for me are:

* Assuming this is the missing piece of the jigsaw for #5014 what are the next steps to actually move to serving everything from heroku?
* Once we do this, is there actually any value to `localhost:3000` and `localhost:8080` being different in dev? i.e: Should `npm start` become `npm run build && npm run start:server` instead of `concurrently --names server,frontend "npm run start:server" "cross-env GATSBY_BASE_URL=http://localhost:8080 gatsby develop --port 3000"` ?

### DIFF
--- a/core/server/server.js
+++ b/core/server/server.js
@@ -149,6 +149,9 @@ const publicConfigSchema = Joi.object({
   rateLimit: Joi.boolean().required(),
   handleInternalErrors: Joi.boolean().required(),
   fetchLimit: Joi.string().regex(/^[0-9]+(b|kb|mb|gb|tb)$/i),
+  documentRoot: Joi.string().default(
+    path.resolve(__dirname, '..', '..', 'public')
+  ),
 }).required()
 
 const privateConfigSchema = Joi.object({
@@ -429,7 +432,7 @@ class Server {
     log(`Server is starting up: ${this.baseUrl}`)
 
     const camp = (this.camp = Camp.create({
-      documentRoot: path.resolve(__dirname, '..', '..', 'public'),
+      documentRoot: this.config.public.documentRoot,
       port,
       hostname,
       secure,
--- a/core/server/server.js
+++ b/core/server/server.js
@@ -20,6 +20,7 @@ const {
 } = require('../base-service/legacy-request-handler')
 const { clearRegularUpdateCache } = require('../legacy/regular-update')
 const { rasterRedirectUrl } = require('../badge-urls/make-badge-url')
+const { setHeadersForCacheLength } = require('../base-service/cache-headers')
 const log = require('./log')
 const sysMonitor = require('./monitor')
 const PrometheusMetrics = require('./prometheus-metrics')
@@ -186,6 +187,20 @@ const privateMetricsInfluxConfigSchema = privateConfigSchema.append({
   influx_username: Joi.string().required(),
   influx_password: Joi.string().required(),
 })
+
+function getHandlerIndex(camp, functionName) {
+  const stackFunctions = camp.stack.map(fn => fn.name)
+  const index = stackFunctions.findIndex(name => name === functionName)
+  if (index === -1) {
+    throw new Error(`Could not find stack function ${functionName}`)
+  }
+  return index
+}
+
+function addHandlerAtIndex(camp, index, fn) {
+  camp.stack.splice(index, 0, fn)
+}
+
 /**
  * The Server is based on the web framework Scoutcamp. It creates
  * an http server, sets up helpers for token persistence and monitoring.
@@ -421,6 +436,17 @@ class Server {
       cert,
       key,
     }))
+
+    const index = getHandlerIndex(this.camp, 'staticLayer')
+    addHandlerAtIndex(this.camp, index, function addCacheHeaders(
+      req,
+      res,
+      next
+    ) {
+      const frontendMaxAge = 300
+      setHeadersForCacheLength(res, frontendMaxAge)
+      next()
+    })
 
     const { metricInstance } = this
     this.cleanupMonitor = sysMonitor.setRoutes(
--- a/core/server/server.spec.js
+++ b/core/server/server.spec.js
@@ -45,6 +45,16 @@ describe('The server', function () {
         .and.to.include('apple')
     })
 
+    it('should serve front-end with default maxAge', async function () {
+      const { headers } = await got(`${baseUrl}/`)
+      expect(headers['cache-control']).to.equal('max-age=300, s-maxage=300')
+    })
+
+    it('should serve badges with custom maxAge', async function () {
+      const { headers } = await got(`${baseUrl}npm/l/express`)
+      expect(headers['cache-control']).to.equal('max-age=3600, s-maxage=3600')
+    })
+
     it('should redirect colorscheme PNG badges as configured', async function () {
       const { statusCode, headers } = await got(
         `${baseUrl}:fruit-apple-green.png`,
--- a/core/server/server.spec.js
+++ b/core/server/server.spec.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const path = require('path')
 const { expect } = require('chai')
 const isSvg = require('is-svg')
 const config = require('config')
@@ -13,7 +14,11 @@ describe('The server', function () {
     before('Start the server', async function () {
       // Fixes https://github.com/badges/shields/issues/2611
       this.timeout(10000)
-      server = await createTestServer()
+      server = await createTestServer({
+        public: {
+          documentRoot: path.resolve(__dirname, 'test-public'),
+        },
+      })
       baseUrl = server.baseUrl
       await server.start()
     })
--- a/core/server/test-public/index.html
+++ b/core/server/test-public/index.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>shields.io</title>
+  </head>
+  <body>
+    concise, consistent, legible
+  </body>
+</html>


core/server/server.js
```diff
@@ -421,6 +422,11 @@ class Server {
       cert,
       key,
     }))
+    camp.handle(function (req, res, next) {
+      const frontendMaxAge = 300
```
chris48s, line 426: I guess we could pull this out to a setting if we wanted so that self-hosting users can tweak it. We've already got

```yml
cacheHeaders:
  defaultCacheLengthSeconds: 120
```

so we would probably need to split out into

```yml
cacheHeaders:
  defaultBadgeMaxAge: 120
  defaultFrontendMaxAge: 300
```

For the moment I've just left it hard-coded.

chris48s:
Hmm. So I guess the reason the tests pass locally but are failing in CI is because I have the frontend built locally, but in CI we don't. One way to deal with this is to add `npm run build &&` to the front of `test:core` but its quite a heavyweight solution and slows that test suite down quite a bit. I wonder if there's a more lightweight approach to give us something to serve on / under test :thinking:

paulmelnikow:
Thanks for picking this up!

> Once we do this, is there actually any value to localhost:3000 and localhost:8080 being different in dev?

Yea, it’s useful when developing the frontend. Otherwise you’d have to rerun the build every time you change something.

core/server/server.js
paulmelnikow: Hard coded seems fine until someone asks.

calebcartwright: Agreed

calebcartwright:
>  I wonder if there's a more lightweight approach to give us something to serve on / under test thinking

Could we just temporarily copy one of the static files under `public` instead of running a full build as part of that test? Or maybe we could add a param to `start` on the server with the same value for the `documentRoot`, but allow tests to override it when needed?

chris48s:
I think I'm leaning towards the test `documentRoot` approach instead of writing to `public` under test.

> it’s useful when developing the frontend

:bulb:

core/server/server.js
```diff
@@ -421,6 +422,11 @@ class Server {
...
+      setHeadersForCacheLength(res, frontendMaxAge)
+      next()
```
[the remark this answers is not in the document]
chris48s: We only register the handler once at server startup - we're not re-registering it on every request
but the actual handler function _does_ run on every request. i.e: we are re-evaluating `new Date()` for the `Expires`/`Date` headers on every request - we're not just evaluating `new Date()` once on server startup and using that value for the server lifetime.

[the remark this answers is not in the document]
paulmelnikow: I guess that means these values get set on the requests first, and then it's up to the badge request handler to overwrite them when returning badges.

That seems okay for the most part, though for a few kinds of requests that might be a problem: static badges, redirects, and the suggest endpoint. Static badges and redirects invoke `setCacheHeadersForStaticResource()` which sets `Cache-Control` and `Last-Modified` and after this change they get a hodgepodge of headers. I haven't dug into the suggest endpoint, though I would guess those requests are not getting any cache headers in master. Probably it would be a good idea to add server tests which check all the headers which are set on those requests.

It seems better to set the new cache headers only on requests being fulfilled through the scoutcamp documentRoot (i.e. [`staticUnit`](https://github.com/espadrine/sc/blob/c1042e58fb849e1a2ec87dc22695d21329b0e8e1/lib/camp.js#L698-L703)), instead of on every request. Looking at the code, I don't see an easy way to do that without modifying scoutcamp.

In terms of the cache behavior itself, an alternative to the five-minute cache would be to use `setCacheHeadersForStaticResource()` (maybe with a shorter timeout) combined with `serverHasBeenUpSinceResourceCached()`, which most of the time will send 304s. That'd be closer aligned to the way we handle our existing static content.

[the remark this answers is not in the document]
chris48s: > I guess that means these values get set on the requests first, and then it's up to the badge request handler to overwrite them when returning badges.

Yeah that's what I was trying to explain when I said

> What this actually does is apply `frontendMaxAge` to everything, but then for every other request we then overwrite it with subsequent calls to `setHeader()`

clearly that failed.

Its useful you point out that there is an issue with static badges and redirects. I'll revisit. It has also occurred to me that we might want this on in production, but not for development.

> It seems better to set the new cache headers only on requests being fulfilled through the scoutcamp documentRoot

Yes that would be ideal, but not available out-of-the-box. Maybe we _do_ need a PR to scoutcamp after all. I will rethink it.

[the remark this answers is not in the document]
paulmelnikow: > Yeah that's what I was trying to explain when I said
> 
> > What this actually does is apply `frontendMaxAge` to everything, but then for every other request we then overwrite it with subsequent calls to `setHeader()`

Ahha sorry, I missed that in the top comment.

chris48s:
Right here's a revised approach.

Scoutcamp gives us a stack of handlers. Each handler either ends the request with a response, or hands off to the next handler in the stack. If we can't serve the request with `routeLayer`, we hand off to `staticLayer`. `handle()` only allows us to add a handler to the _front_ of the stack (at index 0), so in order to work round this, I've added a little helper `addHandlerAtIndex()` which allows us to add a handler _anywhere_ in the stack. This allows us to insert a handler between `routeLayer` and `staticLayer` so we can add some headers to only requests which weren't served by `routeLayer`. Does this seem better?

Having thought about it slightly more, we don't actually need to implement different behaviour for dev/prod because we already have it if we're going to stick with `gatsby develop` in dev :)

chris48s:
Would it be possible to get a second look at this one?

paulmelnikow:
> In terms of the cache behavior itself, an alternative to the five-minute cache would be to use `setCacheHeadersForStaticResource()` (maybe with a shorter timeout) combined with `serverHasBeenUpSinceResourceCached()`, which most of the time will send 304s. That'd be closer aligned to the way we handle our existing static content.

Any thoughts on this part?

core/server/server.js
```diff
@@ -414,14 +432,25 @@ class Server {
...
+      documentRoot: this.config.public.documentRoot,
       port,
       hostname,
       secure,
       cert,
       key,
     }))
 
+    const index = getHandlerIndex(this.camp, 'staticLayer')
+    addHandlerAtIndex(this.camp, index, function addCacheHeaders(
```
paulmelnikow, line 444: This will run for everything that isn't handled by a route, which I think mainly is the `camp.notfound` route defined in `server.js`.

An alternative could be to replace `staticLayer` with our own function, although I don't see a way to do that without copying in a bunch of code from `respondWithFile`.

Hmm, Scoutcamp is sending some cache headers on files…

https://github.com/espadrine/sc/blob/c1042e58fb849e1a2ec87dc22695d21329b0e8e1/lib/camp.js#L677-L685

chris48s: Yes - inserting the handler here will apply those headers to every subsequent handler in the stack (i.e: also `notFound`). Is there a problem if we serve 404s with these headers applied?

> Any thoughts on this part?

I've not tried it yet. IIUC, doing that would also require us to _replace_ `staticLayer`, yeah?

paulmelnikow: It makes me wonder if the headers Scoutcamp is setting (in that link) are the ones we want.

chris48s: ..so you're thinking what we actually want to do is just bin all this off an go with the out-of-the-box ScoutCamp behaviour?

paulmelnikow: This is from the output of `curl -v https://shields-staging.herokuapp.com/ > /dev/null`:

```
> GET / HTTP/1.1
> Host: shields-staging.herokuapp.com
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: Cowboy
< Connection: keep-alive
< Content-Type: text/html
< Last-Modified: Thu, 01 Oct 2020 02:21:59 GMT
< Date: Thu, 01 Oct 2020 19:21:10 GMT
< Transfer-Encoding: chunked
< Via: 1.1 vegur
```

Comparing that to our own static content (`curl -v https://shields-staging.herokuapp.com/badge/foo-bar-blue > /dev/null` …

```
> GET /badge/foo-bar-blue HTTP/1.1
> Host: shields-staging.herokuapp.com
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 200 OK
< Server: Cowboy
< Connection: keep-alive
< Cache-Control: max-age=86400, s-maxage=86400
< Last-Modified: Thu, 01 Oct 2020 02:24:01 GMT
< Content-Type: image/svg+xml;charset=utf-8
< Date: Thu, 01 Oct 2020 19:23:08 GMT
< Transfer-Encoding: chunked
< Via: 1.1 vegur
```

… the thing that's missing is an actual Cache-Control header. I think by default that means it's set to `private`, and that we'd do better by additionally setting `max-age`.

The changes you've made here for testing seem great.

I get not wanting to modify the work, although I still don't see a way to keep the code you're adding from running on the requests that are further down the stack, like `camp.notfound`. The support for If-Modified-Since + 304's in sc seems good. What would you think about adding a staticCacheSeconds param to Scoutcamp? Or even hard-coding it, since it's a just-for-us fork of something we basically want to deprecate anyway.

core/server/server.js
```diff
@@ -414,14 +432,25 @@ class Server {
...
-      documentRoot: path.resolve(__dirname, '..', '..', 'public'),
+      documentRoot: this.config.public.documentRoot,
       port,
       hostname,
       secure,
       cert,
       key,
     }))
 
+    const index = getHandlerIndex(this.camp, 'staticLayer')
```
calebcartwright, line 443: Could we add an inline comment/link with the relevant scoutcamp info? This feels a bit magical to the non-scoutcamp fluent like me :smile:

calebcartwright:
The code changes here seem simple enough themselves, but I definitely don't have enough context on the sc integration/potential implications to offer much.

Let me know if another pair of eyes is needed on anything, otherwise happy to defer to you all!

core/server/server.js
chris48s: > and that we'd do better by additionally setting `max-age`

OK.. so then, isn't that what this PR does at the moment? The current default behaviour plus setting a max-age (and s-max-age and expires) Are we having two different conversations?

I've said

> Yes - inserting the handler here will apply those headers to every subsequent handler in the stack (i.e: also `notFound`). Is there a problem if we serve 404s with these headers applied?

we've said a bunch more stuff that doesn't really answer that question (but feels like we've circled back round to "what this PR should do is what it does now", or I've misunderstood?) and now we're back to

> although I still don't see a way to keep the code you're adding from running on the requests that are further down the stack, like `camp.notfound`

Can you really clearly say what you're asking/saying/suggesting on this topic. There is something important missing. Fortunately cache invalidation is a notoriously easy topic, so it should be easy to sort out ;)

paulmelnikow: > Fortunately cache invalidation is a notoriously easy topic, so it should be easy to sort out ;)

🤣 

> Can you really clearly say what you're asking/saying/suggesting on this topic. There is something important missing. 

Sorry I've haven't been explaining it clearly.

The code you've added here causes `setHeadersForCacheLength()` to be invoked on requests for static content, as intended, and for requests that don't match static content which are handled by `camp.notfound()`, which is unintended.

We should ensure we're only adding headers to the requests fort static content.

The only correct way I see to do this is to modify scoutcamp's `replyWithFile()` function with additional code to set max-age headers. The clean way to configure it would be to pass a staticCacheSeconds option to the Camp constructor, set it as an instance variable (something like `server.staticCacheSeconds`), and then pass it as a parameter to `respondWithFile()` on https://github.com/badges/sc/blob/35f6b68ae890fd0c8f3771c74c9d7971b69cee41/lib/camp.js#L701.

chris48s: > The code you've added here causes `setHeadersForCacheLength()` to be invoked on requests for static content, as intended, and for requests that don't match static content which are handled by `camp.notfound()`

Exactly. This is what mean when I say:

> Yes - inserting the handler here will apply those headers to every subsequent handler in the stack (i.e: also `notFound`)

This is going round in the same circle again: I'm saying "yes - the max-age header will also be set on 404 responses" and then you're replying with "the max-age header will be set on 404 responses". :laughing: 

The question I'm asking is: If we serve a 404 with the same (fairly short) `max-age` as the rest of the front-end, is that actually a problem? i.e: What is the bad thing that happens if we do that? As far as I can tell, its not inherently wrong to set a `max-age` on a 404, but I'd like to understand what I'm missing about why its going to cause us a problem in this situation. I'm not saying it necessarily _isn't_ - I'm just not seeing what you're seeing.

paulmelnikow: > This is going round in the same circle again: I'm saying "yes - the max-age header will also be set on 404 responses" and then you're replying with "the max-age header will be set on 404 responses". 😆

😆 

> The question I'm asking is: If we serve a 404 with the same (fairly short) `max-age` as the rest of the front-end, is that actually a problem? i.e: What is the bad thing that happens if we do that? As far as I can tell, its not inherently wrong to set a `max-age` on a 404, but I'd like to understand what I'm missing about why its going to cause us a problem in this situation. I'm not saying it necessarily _isn't_ - I'm just not seeing what you're seeing.

What I can say now is: setting cache headers on 404 responses _is not the intention of this PR_, but would be an unintended effect.

I can investigate this effect further, and articulate whether or not there's a circumstance where it will cause a problem. (I'm mostly thinking about invalid badge URLs that are 404's.)

chris48s: I find this use of the word "unintended" annoying because it seems like you're still trying to say I don't know what the code I've written does, but I'm going to take the most charitable interpretation of this and assume you mean you're saying it doesn't do what you want, rather than I don't know what it does.

Either way, I care about getting to a stage where we can do one-step deploys more than I care what TTL we're serving the 404 badge with. Lets move this on and say we don't want to make that behaviour change at this time. I've submitted https://github.com/badges/sc/pull/3 and https://github.com/badges/sc/pull/4 . I'm going to close this PR and I'll submit a revised PR once we've made a change upstream.

We can defer the issue of whether we do or don't want to send an explicit `max-age` on 404s. At the CloudFlare level I think the behaviour we _should_ have if we send no `max-age` is that the 404 badge gets implicitly cached by CloudFlare with a 3m TTL (that's my reading of https://support.cloudflare.com/hc/en-us/articles/200172516-Understanding-Cloudflare-s-CDN#h_51422705-42d0-450d-8eb1-5321dcadb5bc anyway) so I don't think it should actually be a cache miss on every request anyway (at least at the CDN level).. but then I went to check that and found they are served with `cf-cache-status: DYNAMIC`. Then I started pulling the thread and realised there's loads of stuff not getting served from cache that should be so I guess they _are_ always a cache miss, but it seems like what TTL we've got on the 404s might not be the biggest issue in the grand scheme of things. Anyway, lets kick this 404 can down the road and leave them with no explicit `max-age` for now..

PR #5579, opened 2020-09-19
[WordPress] API Update

### LINKED ISSUE #5499: "invalid response data" on WordPress plugin badge

- [ ] My own instance

:beetle: **Description**

I'm trying to add a badge with the [WooCommerce plugin](https://wordpress.org/plugins/woocommerce-myparcel) version to our repository , but the shield doesn't work. I've tried several other plugins, `bbpress` (from the example), `woocommerce` and a random plugin I plucked from the WordPress website. All of them work, but our plugin's doesn't. Could there be something wrong in the plugin's settings? Or am I missing something?

:link: **Link to the badge**

https://img.shields.io/wordpress/plugin/v/woocommerce-myparcel

JoeIzzard:
Related to #5493, the API Update has been split out into its own PR, to reduce the demand on the maintainers. This should also fix #5499 

The API change, required a couple of changes to schema and request, and related testing in platform tester. I also took this chance to split the plugin and theme schemas into their own separate entities to make adding the new badges from #5493 easier, however it doesn't introduce any of the schema required for those badges.

I have also `.required()` the schema as we should always receive those items for the current badge set, however I am open to exploring the option of not requiring them at the schema but dealing with errors in the badges instead.

I will handle converting service classes to static props as I continue to submit PR's for other aspects of #5493

### DIFF
--- a/services/wordpress/wordpress-base.js
+++ b/services/wordpress/wordpress-base.js
@@ -9,10 +9,10 @@ const stringOrFalse = Joi.alternatives(Joi.string(), Joi.bool())
 const themeSchema = Joi.object()
   .keys({
     version: Joi.string().required(),
-    rating: nonNegativeInteger.required(),
-    num_ratings: nonNegativeInteger.required(),
-    downloaded: nonNegativeInteger.required(),
-    active_installs: nonNegativeInteger.required(),
+    rating: nonNegativeInteger,
+    num_ratings: nonNegativeInteger,
+    downloaded: nonNegativeInteger,
+    active_installs: nonNegativeInteger,
   })
   .required()
 
--- a/services/wordpress/wordpress-base.js
+++ b/services/wordpress/wordpress-base.js
@@ -19,10 +19,10 @@ const themeSchema = Joi.object()
 const pluginSchema = Joi.object()
   .keys({
     version: Joi.string().required(),
-    rating: nonNegativeInteger.required(),
-    num_ratings: nonNegativeInteger.required(),
-    downloaded: nonNegativeInteger.required(),
-    active_installs: nonNegativeInteger.required(),
+    rating: nonNegativeInteger,
+    num_ratings: nonNegativeInteger,
+    downloaded: nonNegativeInteger,
+    active_installs: nonNegativeInteger,
     requires: stringOrFalse.required(),
     tested: Joi.string().required(),
   })
--- a/services/wordpress/wordpress-base.js
+++ b/services/wordpress/wordpress-base.js
@@ -4,25 +4,48 @@ const Joi = require('@hapi/joi')
 const { nonNegativeInteger } = require('../validators')
 const { BaseJsonService, NotFound } = require('..')
 
-const foundSchema = Joi.object()
+const stringOrFalse = Joi.alternatives(Joi.string(), Joi.bool())
+
+const themeSchema = Joi.object()
   .keys({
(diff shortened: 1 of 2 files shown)


services/wordpress/wordpress-base.js
```diff
@@ -4,25 +4,48 @@ const Joi = require('@hapi/joi')
...
-    num_ratings: nonNegativeInteger,
-    downloaded: nonNegativeInteger,
-    active_installs: nonNegativeInteger,
-    requires: Joi.string(), // Plugin Only
-    tested: Joi.string(), // Plugin Only
+    version: Joi.string().required(),
+    rating: nonNegativeInteger.required(),
+    num_ratings: nonNegativeInteger.required(),
+    downloaded: nonNegativeInteger.required(),
+    active_installs: nonNegativeInteger.required(),
```
calebcartwright, line 15: the `nonNegativeInteger` helper already includes `required()` so there's no need to add a duplicative one here
```suggestion
    rating: nonNegativeInteger,
    num_ratings: nonNegativeInteger,
    downloaded: nonNegativeInteger,
    active_installs: nonNegativeInteger,
```

services/wordpress/wordpress-base.js
```diff
@@ -4,25 +4,48 @@ const Joi = require('@hapi/joi')
...
   .required()
 
-const notFoundSchema = Joi.string().allow(null, false)
+const pluginSchema = Joi.object()
+  .keys({
+    version: Joi.string().required(),
+    rating: nonNegativeInteger.required(),
+    num_ratings: nonNegativeInteger.required(),
+    downloaded: nonNegativeInteger.required(),
+    active_installs: nonNegativeInteger.required(),
```
calebcartwright, line 25:
```suggestion
    rating: nonNegativeInteger,
    num_ratings: nonNegativeInteger,
    downloaded: nonNegativeInteger,
    active_installs: nonNegativeInteger,
```

services/wordpress/wordpress-base.js
```diff
@@ -37,6 +60,7 @@ module.exports = class BaseWordpress extends BaseJsonService {
               homepage: 0,
               tags: 0,
               screenshot_url: 0,
+              downloaded: 1,
```
calebcartwright, line 63: What's the purpose/impact of this field? Does it filter out plugins/themes that have 0 downloads?

calebcartwright (commented):
Thanks @JoeIzzard! Couple minor inline tweaks needed, and one question for your review, but otherwise this is good to go

services/wordpress/wordpress-base.js
JoeIzzard: So the new API allows us to change which bits of information we need, and previously `downloaded` was returned by default. It's not anymore so we are explicitly asking for that bit of information. 

`1` is us requesting it
`0` is us telling the API to skip that

That block currently skips a lot of info we don't need, but can also be expanded to request additional information. Some of the new badges I will be submitting will also add to this block to request the information that they need. A similar change was made to the platform tester file, as it changes the expected url. 

It won't skip anything that has 0 downloads, but will just ensure we get that field returned in the response.

calebcartwright: Ah okay cool, thanks for the info (it actually makes more sense now that i think about it being a member of `fields` :laughing:)

calebcartwright (approved):
LGTM, thank you!

PR #5583, opened 2020-09-20
add a validator for 0/{{time period}}, run [GitHubCommitActivity]

calebcartwright:
I know there's some history behind why certain test validators like `isMetricOverTimePeriod` reject a value of `0` and I understand why we'd want a validator that required a positive integer. 

However, sometimes the value genuinely is 0 and historically the strategy to avoid those cases causing spurious service test failures in the past was to attempt to pick a test target that was unlikely to result in a 0. That doesn't always work though, for example from the daily test run a little earlier today: https://app.circleci.com/pipelines/github/badges/daily-tests/481/workflows/8627f386-28e0-46fb-b19c-d8c6d04b4cde/jobs/1379

I've opened this PR for consideration, which adds a _new_ validator for matching 0/timePeriod, that service tests could opt into using if needed with something `Joi.alternatives().try(isMetricOverTimePeriod, isZeroOverTimePeriod)`

### DIFF
--- a/services/github/github-commit-activity.tester.js
+++ b/services/github/github-commit-activity.tester.js
@@ -1,7 +1,16 @@
 'use strict'
 
-const { isMetricOverTimePeriod } = require('../test-validators')
+const Joi = require('@hapi/joi')
+const {
+  isMetricOverTimePeriod,
+  isZeroOverTimePeriod,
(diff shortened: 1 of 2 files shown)


PyvesB (approved):
👍🏻

PR #5586, opened 2020-09-21
Work around DT conflict; bump TS and @types/styled-components

### LINKED ISSUE #5294: The @types/styled-components mess
Wanted to surface this here rather than burying it in a dependabot PR…

As @calebcartwright observed at https://github.com/badges/shields/pull/5277#issuecomment-653836259 the version of `@types/styled-components` that we've pinned as a result of DefinitelyTyped/DefinitelyTyped#33311 is now blocking us from updating TypeScript.

This is untenable.

Part of me wants to try to dive into the DefTyped repo and fix this though I just don't have a ton of bandwidth.

Increasingly the [hacky workaround](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311#issuecomment-653938738) of deleting `node_modules/@types/react-native`, a package which of course should not be installed into Shields, is feeling like a reasonable way to handle this.

paulmelnikow:
@types/styled-components causes a conflict with @types/react-native. See DefinitelyTyped/DefinitelyTyped#33311.

Close #5294

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6839,6 +6839,16 @@
       "integrity": "sha512-2xtoL22/3Mv6a70i4+4RB7VgbDDORoWwjcqeNysojZA0R7NK17RbY5Gof/2QiFfJgX+KkWghbwJ+d/2SB8Ndzg==",
       "dev": true
     },
+    "@types/hoist-non-react-statics": {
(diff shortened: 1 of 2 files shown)


calebcartwright (approved):
:+1:

PR #5587, opened 2020-09-21
convert [lgtm liberapay librariesio localizely luarocks] classes to static props

jabbar86:
Converted Static Classes to Static Props.


calebcartwright (approved):
Thanks!

PR #5591, opened 2020-09-22
Completed [mavencentral mavenmetadata microbadger mozillaobservatory myget netlify] static classes into static props

jabbar86:


services/maven-central/maven-central.service.js
```diff
@@ -52,7 +52,9 @@ module.exports = class MavenCentral extends BaseXmlService {
     },
   ]
 
-  static defaultBadgeData = { label: 'maven-central' }
+  static defaultBadgeData = {
+    label: 'maven-central',
+  }
```
calebcartwright, line 57: Out of curiosity, why this change?

services/mozilla-observatory/mozilla-observatory.service.js
```diff
@@ -46,42 +46,32 @@ const documentation = `
 `
 
 module.exports = class MozillaObservatory extends BaseJsonService {
-  static get category() {
-    // TODO: Once created, change to a more appropriate category,
-    // see https://github.com/badges/shields/pull/2926#issuecomment-460777017
```
calebcartwright, line 51: Let's be sure that we don't drop the inline comment as part of the refactoring

services/maven-central/maven-central.service.js
jabbar86: @calebcartwright Just to align the object format as I have done in all files.

services/mozilla-observatory/mozilla-observatory.service.js
jabbar86: Sure will keep this in mind.Thanks

services/maven-central/maven-central.service.js
calebcartwright: prettier will automatically wrap lines when they need to be, though generally respects the original line wrapping as well. This one is really short so it could easily fit on one line. 

No need to revert this change, but also don't feel obligated to wrap these really simple/short object literals on multiple lines either. In some cases, such as larger route definitions, they're more readable, but otherwise keeping short fields like this one line helps make the files ever so smaller :smile:

services/mozilla-observatory/mozilla-observatory.service.js
calebcartwright: No worries. This should be good to go once you restore this comment

services/maven-central/maven-central.service.js
jabbar86: Okay @calebcartwright  understood thank you  😄

services/mozilla-observatory/mozilla-observatory.service.js
jabbar86: @calebcartwright changes done.

calebcartwright (approved):
Thanks!

PR #5620, opened 2020-09-28
fix [jsdelivr] service tests

calebcartwright:
https://app.circleci.com/pipelines/github/badges/daily-tests/495/workflows/4dcc5a2d-179a-40da-8956-0ffb7c3540fe/jobs/1393

Another batch of failing tests due to 0/week 

Leveraging a similar approach as #5583 for this one vs the approach in #5618 as I've found the jsdelivr service to be trickier to find good test targets


PyvesB (approved):
Looks good to me. 👍🏻

PR #5621, opened 2020-09-28
[Criterion] Add support for "no change" status

chmoder:
While adding support for the "no change" status I noticed the criterion api updates I made last week broke your tests.  We added a "noise level threshold" to the status calculation because the runner may have something going on while a benchmark is running which makes the performance _appear_ to have regressed.  Now if a benchmark is within, for example 2% of the previous measurement, we report "no change", instead of "regressed".

Thank you for helping out with the criterion badge, you are more than welcome to assign issues related to criterion to me in the future.


PyvesB (approved):
Thanks for proactively picking this up @chmoder , looks good to me! 👍🏻

PR #5625, opened 2020-09-29
package.json: sort dependency array

paulmelnikow:
This diff surfaces a lot when making updates to package.json.


calebcartwright (approved):
I don't suppose prettier or eslint has an automagic feature to do this for us do they?

paulmelnikow:
> I don't suppose prettier or eslint has an automagic feature to do this for us do they?

Prettier does not, though here are a couple options:

- https://github.com/keithamus/sort-package-json
- https://github.com/kellyselden/eslint-plugin-json-files/blob/master/docs/rules/sort-package-json.md (might be just for keys, not for dependency arrays)

It doesn't happen too often, only when dependencies are renamed by hand.

PR #5626, opened 2020-09-29
Remove enzyme which we aren’t using right now

paulmelnikow:
Our approach to preventing bugs on the frontend has mostly leveraged types, though we could restore these if we start writing frontend tests again. In the meantime it's one less thing to unnecessarily install and maintain.


calebcartwright (approved):
sgtm

PR #5628, opened 2020-09-30
Remove obsolete peer-to-peer GitHub token-sharing code

paulmelnikow:
Tokens are now shared exclusively through the Redis DB.

Ref #3393

### DIFF
--- a/config/shields-io-production.yml
+++ b/config/shields-io-production.yml
@@ -18,7 +18,3 @@ public:
   redirectUrl: 'https://shields.io/'
 
   rasterUrl: 'https://raster.shields.io'
-
-private:
-  # These are not really private; they should be moved to `public`.
-  shields_ips: ['192.99.59.72', '51.254.114.150', '149.56.96.133']
--- a/core/server/monitor.js
+++ b/core/server/monitor.js
@@ -59,10 +59,6 @@ function setRoutes({ rateLimit }, { server, metricInstance }) {
     next()
   })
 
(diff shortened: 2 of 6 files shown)


calebcartwright:
Not really familiar with this portion of the codebase, but no objections or concerns with the changes either. IIUC the acceptance of the donated tokens is still fully addressed via https://github.com/badges/shields/pull/5628/files#diff-caf3e8bee6a9ea794502322899cbbc7cR83 correct?

paulmelnikow:
> Not really familiar with this portion of the codebase, but no objections or concerns with the changes either. IIUC the acceptance of the donated tokens is still fully addressed via https://github.com/badges/shields/pull/5628/files#diff-caf3e8bee6a9ea794502322899cbbc7cR83 correct?

Yea, that's right.

If we are accepting new tokens on Heroku at the moment, I guess it's working by forwarding the tokens to the Shields servers which then add them to Redis. That's very strange, of course. After this change the (Heroku) server that gets the token should deposit it right into Redis.

calebcartwright:
> If we are accepting new tokens on Heroku at the moment, I guess it's working by forwarding the tokens to the Shields servers which then add them to Redis. That's very strange, of course.

Strange indeed, and good timing on this! Had we decommissioned the OVH servers before removing this we might've seen some weird errors

calebcartwright (dismissed):

calebcartwright (approved):

PR #5631, opened 2020-09-30
Inject shieldsSecret into GitHub token admin endpoint

paulmelnikow:
Ref #3393


core/server/secret-is-valid.js
```diff
@@ -13,9 +11,6 @@ function constEq(a, b) {
...
 }
 
-module.exports = function secretIsValid(secret = '') {
-  return (
-    serverSecrets.shields_secret &&
-    constEq(secret, serverSecrets.shields_secret)
-  )
+module.exports = function secretIsValid(secret = '', expected) {
+  return expected && constEq(secret, expected)
 }
```
calebcartwright, line 16: I know that in practice this still achieves the same functional behavior, and I can't think of any other way of doing it, though do want to note that this feels a bit like we're breaking the original intent of this function as a guaranteed/encapsulated comparison with the actual secret value (because we can now do `if (secretIsValid('my-fake-secret', 'my-fake-secret')....`

calebcartwright (dismissed):

core/server/secret-is-valid.js
paulmelnikow: Hmm, true.

Maybe it's better to remove this function and export only `constEq`?

Alternatively we could make a function like this:

```js
function makeSecretIsValid(serverSecret) {
  return function (secret = '') {
    ...
  }
}
```

calebcartwright: Either of those would work, and I'd personally prefer either. No strong objections to what's here though so have approved, but happy to re-:+1: if you decide to change

calebcartwright (approved):
Nice!

PR #5632, opened 2020-09-30
Refactor service classes [snyk sonar sourceforge sourcegraph spack]

vanhaaggen:

PR #5640, opened 2020-10-02
Remove EditorConfig from recommend VS Code plugins

paulmelnikow:
Frequently when I open Shields in VS Code it pops up a message asking if I want to install the recommended plugins. I think for Shields to recommend a plugin, we should agree it is pretty much essential for working on the project.

I have never used any form of EditorConfig. I imagine it may be useful for some users on some platforms, though for the majority of users, it is definitely not essential.

Perhaps the users who need it (Windows users?) know who they are and can take it on themselves to install the plugin if PRs cause them problems.

Certainly open to discussion if people feel otherwise!

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,7 +1,3 @@
 {
-  "recommendations": [
-    "esbenp.prettier-vscode",
-    "EditorConfig.EditorConfig",
-    "dbaeumer.vscode-eslint"
-  ]
+  "recommendations": ["esbenp.prettier-vscode", "dbaeumer.vscode-eslint"]
 }


calebcartwright:
I'm not necessarily opposed to this, though I suspect we have a line ending-related issue that some users are running into periodically (https://github.com/badges/shields/pull/5632#issuecomment-702008559)

It's possible to tell VS Code to hush when it comes to recommended extensions, via `showRecommendationsOnlyOnDemand` and/or `ignoreRecommendations` which may be a better option if you really don't want the EditorConfig extension (though it would also kinda beg the question of why we have this file: https://github.com/badges/shields/blob/master/.editorconfig).

The prettier extension is helpful for sure, but IIRC it also requires users have the global `formatOnSave` setting enabled which isn't specific to any extension whereas when it comes to the basic line ending, indent style, etc. EditorConfig will "just work" regardless of platform and without needing any other VS Code settings enabled.

One alternative (that may be worthwhile regardless) would be to add a .gitattributes file to the repo to prevent line ending issues sometimes seen by Windows users, presumably related to their local git's autocrlf configuration

paulmelnikow:
> I'm not necessarily opposed to this, though I suspect we have a line ending-related issue that some users are running into periodically ([#5632 (comment)](https://github.com/badges/shields/pull/5632#issuecomment-702008559))

It'd be helpful to know what's actually happening there. We have prettier configured to use `lf` for endOfLine (that's the default anyway starting in Prettier 2). Maybe that contributor could run it again and push a branch so we can see. Curious if it's line endings, semicolons, wrong version of Prettier…

> The prettier extension is helpful for sure, but IIRC it also requires users have the global `formatOnSave` setting enabled which isn't specific to any extension whereas when it comes to the basic line ending, indent style, etc.

I think we could set `formatOnSave` in the project, right?

> EditorConfig will "just work" regardless of platform and without needing any other VS Code settings enabled.

Do you use it? Is it useful?

I would think on existing files, editors would read the line endings in the file and use them. Prettier would enforce them on any new files created. For that matter, much of what's in our editorconfig seems redundant with Prettier. And doesn't git even keep line endings consistent these days?

> One alternative (that may be worthwhile regardless) would be to add a .gitattributes file to the repo to prevent line ending issues sometimes seen by Windows users, presumably related to their local git's autocrlf configuration

Ah, yea, maybe that's what I was thinking of!

I don't feel strongly about this either way, though if this popup is pulling my attention then it's probably doing the same for a bunch of other people. Maybe the best solution for _me_ is to just turn this off globally… though I do wonder if we should be using this feature at all. We could let contributors manage their own extensions.

calebcartwright:
> It'd be helpful to know what's actually happening there. 

Agreed! It's the second or third semi-recent incident that's been reported without much detail so I'm admittedly speculating, but a run of prettier impacting _every_ file strikes me as almost certainly line ending related.

> We have prettier configured to use lf for endOfLine (that's the default anyway starting in Prettier 2). Maybe that contributor could run it again and push a branch so we can see. Curious if it's line endings, semicolons, wrong version of Prettier…

My theory about this being Windows is because non-ancient versions of git have the auto crlf (check out Windows style, commit Unix style) enabled by default, but users can also change this at will during the install and of course afterwards. I'm guessing in these cases that git is somehow converting the in-source Unix style line endings to Windows on checkout in the user's local repo, and when the user runs prettier those are "converted back" to Unix style; in these cases git thinking those are actual diffs.

> I don't feel strongly about this either way, though if this popup is pulling my attention then it's probably doing the same for a bunch of other people. Maybe the best solution for me is to just turn this off globally… though I do wonder if we should be using this feature at all. We could let contributors manage their own extensions.

It's a fair point. There's a couple repos I have to open occasionally at my day job that went _way_ overboard with the recommended extensions including several that I've tried previously but genuinely dislike, and have no desire to enable. I find that same recommended install popup annoying in those situations as well :smile: 

> Do you use it? Is it useful?

I do and that's why I hesitate because I'm a big fan of EditorConfig, especially its ubiquity and availability in just about every editor/IDE one can conjure (many of which support it out of the box _without_ requiring an extension https://editorconfig.org/). VS Code is one of the instances where you actually need the extension installed.

I don't feel too strongly either. If there's a portion of the Shields contributors/maintainer community that doesn't want to install the extension then I'm okay with removing this and the accompanying dialog on project open experience :+1:

PyvesB:
> If there's a portion of the Shields contributors/maintainer community that doesn't want to install the extension

I don't use VSCode to develop the project, so no opinion on this one.

calebcartwright:
The other thing I need to keep in mind is the distinction between EditorConfig in general vs. whether the subset of Shields contributors that use VS Code need that recommendation in addition to the Prettier extension

calebcartwright (approved):
I've opened a separate issue to track the (suspected) line ending issue, so I'm okay with removing this

PR #5660, opened 2020-10-03
Completed [twitch twitter ubuntu uptimerobot vaadindirectory] from st…

jabbar86:


services/twitch/twitch.service.js
```diff
@@ -8,39 +8,28 @@ const helixSchema = Joi.object({
...
-    })
-    // link[] is not allowed in examples
-    delete preview.link
-    return [
-      {
-        title: 'Twitch Status',
-        namedParams: {
-          user: 'andyonthewings',
-        },
-        staticPreview: preview,
```
calebcartwright, line 35: hmm. I guess this okay since there's just this one example, but in general want to be mindful about having the examples be completely disconnected from the actual rendering function due to the potential for drift. I'm sure there's a handful of other cases within the codebase, just a bit worried a hypothetical future change could update the actual render function and not be aware of the need to update the preview

calebcartwright (approved):

services/twitch/twitch.service.js
jabbar86: @calebcartwright agreed. will keep this in mind. you can assign me in the future if any changes happen.

PR #5664, opened 2020-10-06
Finish removing server-secrets.js

### LINKED ISSUE #3393: Inject serverSecrets

paulmelnikow:
I’ve tested locally that setting `gh_token` still conditionally enables `ConditionalGithubAuthV3Service`.

Closes #3393


doc/self-hosting.md
```diff
@@ -184,9 +184,14 @@ Start the server using the Sentry DSN. You can set it:
 sudo SENTRY_DSN=https://xxx:yyy@sentry.io/zzz node server
 ```
 
-- or by `sentry_dsn` secret property defined in `private/secret.json`
+Or via [checked in config](server-secrets.md):
```
calebcartwright, line 187: Coming at this cold but I was thrown off by the `checked in..` wording. Is my understanding correct that the intent is to point folks to the server-secrets file for more info about how configs can be set, or encouraging them to check something in, in the VCS sense?

paulmelnikow: Yes, exactly. Maybe "checked-in config" isn't the best way to phrase it. Open to better wording…

calebcartwright, line 187: Bit verbose, but maybe something like
```suggestion
Or via [any of the other supported options](server-secrets.md):
```

PyvesB: This feels even more confusing to me.

PyvesB: Maybe `Or via config, as you would do with [server secrets](server-secrets.md):`. If they don't know how to set secrets via config, they can go to that documentation where there's a little bit more detail.

calebcartwright: It makes sense to me in the broader context, with the preceding talking about setting the environment variable and then the `or..` but I like your suggested text better any way @PyvesB

calebcartwright (approved):

Issue #3027, opened 2019-02-18
Require production traffic to pass through the CDN


paulmelnikow:
Requests that hit the production servers via `img.shields.io` come through Cloudflare. However, the server IP addresses are well known, and accept direct requests. This circumvents Cloudflare's DOS protections and makes it useless in an attack like the one documented in #2991. (There is some protection in place on the server, but it is trivial to bypass.)

It's nice to be able to send test requests directly to the server, and browse the UI on the server for testing, though perhaps we can authenticate those using a header instead. Perhaps we could require HTTP Basic auth for requests that don't come from the CDN, and share the credentials with all the maintainers.

Some useful reference here:

- https://support.cloudflare.com/hc/en-us/articles/201897700-Step-4-Recommended-First-Steps-for-all-Cloudflare-users

Related: we could consider letting Cloudflare handle our rate limiting instead of using our own code on the servers.

calebcartwright:
Totally agreed with all the above. Additionally, I also think it would be beneficial to have the ability to completely block all requests not in a whitelisted IP range (traffic routed via CloudFlare, certain maintainer IPs, etc.) from getting to the servers in the first place. I assume this could be handled via firewall/networking features of our infra provider.

paulmelnikow:
I'm not sure if OVH allows configuring that in the UI or if it would have to go in the firewall config on the VPS. I don't have access to either.

paulmelnikow:
A few links related to this:

- [An article about how to do this with Cloudflare](https://www.viget.com/articles/heroku-cloudflare-the-right-way/)
- [Some middleware we could adapt for that](https://github.com/clive-io/cloudflare-middleware/blob/master/index.js)
- [The Cloudflare IPv4 addresses](https://www.cloudflare.com/ips-v4) and [IPv6 addresses](https://www.cloudflare.com/ips-v6)

paulmelnikow:
I gave this a shot in #5603 

The Shields production Heroku hostname is pretty easy to guess if not already public, so it still seems like a very good idea to require that the production traffic actually come from Cloudflare.

This prevents us from firing requests to Heroku directly, but that seems fine. We already can't address requests to specific dynos, and I don't see any reason why e.g. the admin endpoints couldn't be accessed through the CDN.

calebcartwright:
> This prevents us from firing requests to Heroku directly, but that seems fine.

IIRC we're pushing metrics from the badge servers now vs. the Prometheus pull model correct?

[...]

PR #5666, opened 2020-10-06
When configured, require requests to come from Cloudflare

paulmelnikow:
This is the code part of #3027, following [this article](https://www.viget.com/articles/heroku-cloudflare-the-right-way/) and using [this middleware](https://github.com/clive-io/cloudflare-middleware).

I pulled in the `addHandlerAtIndex()` function @chris48s wrote for #5574.

The middleware isn't perfect for scoutcamp, since it relies on `req.ip` which is something set by Express. However, the other solutions I found were either explicitly deprecated ([cloudflare-ip](https://www.npmjs.com/package/cloudflare-ip)) or relied on dynamically fetching the list of Cloudflare hosts ([cloudflare-ips](https://www.npmjs.com/package/cloudflare-ips)), which seems unnecessary as this list has not changed in several years.

If folks prefer, we could also adapt the code from `cloudflare-middleware` and/or create our own package that does exactly what we want, rather than depend on it. I opened one downstream PR to check that the ranges are still up to date (they are) though I'm not sure if the author will be responsive to us.

I've left this off to start, so we can test it in production using an env var before we make it the production default.

Looking forward to feedback on this approach!


paulmelnikow:
Opened #5667 to fix this Danger error.

calebcartwright:
> The middleware isn't perfect for scoutcamp, since it relies on `req.ip` which is something set by Express. However, the other solutions I found were either explicitly deprecated ([cloudflare-ip](https://www.npmjs.com/package/cloudflare-ip)) or relied on dynamically fetching the list of Cloudflare hosts ([cloudflare-ips](https://www.npmjs.com/package/cloudflare-ips)), which seems unnecessary as this list has not changed in several years.
> 
> If folks prefer, we could also adapt the code from `cloudflare-middleware` and/or create our own package that does exactly what we want, rather than depend on it. I opened one downstream PR to check that the ranges are still up to date (they are) though I'm not sure if the author will be responsive to us.

Seems straightforward enough, even with the extra handler step. Also seems sufficiently small/simple that I'd be comfortable with pulling it in too

calebcartwright (approved):
LGTM

paulmelnikow:
I'll probably merge this tomorrow, and maybe take it for a spin on non-peak Sunday time?

calebcartwright:
> I'll probably merge this tomorrow, and maybe take it for a spin on non-peak Sunday time?

SGTM

calebcartwright (approved):

PR #5671, opened 2020-10-07
Remove FsTokenPersistence

paulmelnikow:
Closes #5665


core/token-pooling/redis-token-persistence.js
```diff
@@ -3,13 +3,13 @@
 const { URL } = require('url')
 const Redis = require('ioredis')
 const log = require('../server/log')
-const TokenPersistence = require('./token-persistence')
 
-module.exports = class RedisTokenPersistence extends TokenPersistence {
```
calebcartwright, line 8: At first glance I found myself wondering if it'd be worth trying to keep `TokenPersistence` as a base class or pulling those functions out into a utils that could be composed/injected into a hypothetical future non-redis based persistence implementation.

But after a bit of thinking I really don't think it is. IIRC the `noteToken*` functions are the only consumed/utilized functions, so any hypothetical future non-redis based persistence solution would just need to have those functions anyway like this one

services/github/github-constellation.js
```diff
@@ -106,13 +103,15 @@ class GithubConstellation {
   }
 
   onTokenInvalidated(tokenString) {
-    process.nextTick(async () => {
-      try {
-        await this.persistence.noteTokenRemoved(tokenString)
-      } catch (e) {
-        log.error(e)
-      }
-    })
```
calebcartwright, line 115: This will also get in cases of local dev env/self-hosted that doesn't have redis configured but has set a GH token in the config which has hit the rate limits correct? I imagine this gets hit in non-persistence enabled scenarios which is why we wouldn't want to throw an error in this case (like above in `onTokenAdded`)?

core/token-pooling/redis-token-persistence.js
paulmelnikow: Yea, it’s a reasonable question. I think my rationale for removing the abstraction layer is that we don’t need to support more than one way to persist dynamic tokens. It’s only for Shields production.

services/github/github-constellation.js
paulmelnikow: In practice we don’t generally hit the rate limit because of the `reserveFraction`, though IIRC `onTokenInvalidated` is not invoked for rate limit exceeded, only 403s. The reason to guard it here is that `this.persistence` might be unset, in which case this code would crash.

calebcartwright (approved):

PR #5694, opened 2020-10-11
Add [GitHubSponsors] badge

### LINKED ISSUE #4683: Badge request: GitHub Sponsors
Hey,
Can we add one for github sponsors

Thanks.

outsideris:
Close #4683 

스크린샷 2020-10-11 오후 7 23 31

I added a badge for GitHub sponsors count. It will display sponsors count and it will be `0` even if the user/org is not enabled the sponsor feature.

Please review my code because I don't think I understand the badge structure.


calebcartwright:
Thank you for the PR @outsideris! Could you please run prettier (`npm run prettier`) to address the formatting issues and failing CI checks?

calebcartwright:
I also noticed that this PR is adding a new logo which is something we generally try to avoid. It's not a design goal of ours to mimic the native buttons/badges/etc. of other services, so I really don't think we need to try to duplicate the look and feel of the GH sponsor button with this badge.

IMO this should use the same default GitHub logo as our other GH social badges do, and we shouldn't take on another custom logo.

https://github.com/badges/shields/blob/master/doc/logos.md#contributing-logos

outsideris:
@calebcartwright Ok, I see. I also worried about it. 
I made the GH sponsor logo from GH sponsor button and actually it is just heart icons. I can submit the logo to Simpleicons? I couldn't decide to submit the logo to Simpleicons because most of them are specific for companies or services, I think.

outsideris:
And I have a question.
Simpleicons looks like monochrome icons. There is a way to set a pink color with Simpleicon?

calebcartwright:
> I can submit the logo to Simpleicons? I couldn't decide to submit the logo to Simpleicons because most of them are specific for companies or services, I think

SimpleIcons is by far our preferred approach for incorporating logo support. I agree that you may have trouble getting a heart icon added upstream as I don't think that meets their requirements for inclusion. 

> Simpleicons looks like monochrome icons. There is a way to set a pink color with Simpleicon?

For a definitive answer I think you'd want to review the Simple Icons docs or reach out to them, but the coloring is by design particularly to allow for easy customization vs fixed, hardcoded colors. You can see this reflected in the examples below

https://img.shields.io/appveyor/build/gruntjs/grunt?logo=github


https://img.shields.io/appveyor/build/gruntjs/grunt?logo=github&logoColor=pink


https://img.shields.io/appveyor/build/gruntjs/grunt?logo=github&logoColor=blue

services/github/github-sponsors.service.js
```diff
@@ -0,0 +1,118 @@
...
+  GithubSponsorsRedirect: redirector({
+    category: 'social',
+    route: {
+      base: 'github/sponsors',
+      pattern: ':user',
+    },
+    transformPath: ({ user }) => `/github/sponsors/${user}`,
+    dateAdded: new Date('2020-10-09'),
+  }),
+}
```
calebcartwright, line 113: The redirector utility is something we use to allow us to update the routes of existing badges without breaking users that are currently leveraging those badges, but we don't need to anything here for a net-new badge.

Removing this will also allow for exporting the service class directly

outsideris:
Ok, I will try to send the GH sponsor icon to SimpleIcons, and back to this PR.

calebcartwright:
> Ok, I will try to send the GH sponsor icon to SimpleIcons, and back to this PR.

Sounds good. That being said, I still think that our default logo for this badge should be the same as the default logo for our other GitHub badges so IMO we could proceed with this once the custom logo is removed. If and when the pink heart logo gets incorporated in Simple Icons then Shields users would be able to use it `?logo=pink-heart-...`, but in the interim folks can always use custom logos on our badges via base64 

https://github.com/badges/shields/blob/master/doc/logos.md#custom-logos

outsideris:
I removed the custom sponsors logo and reformat it with prettier. 
And I submit a [PR](https://github.com/simple-icons/simple-icons/pull/3730) for SimpleIcons.

services/github/github-sponsors.service.js
outsideris: Removed.

calebcartwright (commented):
One minor item on using static props, then good to go!

outsideris:
> We just recently finished up (just about anyway) a refactor of all the service classes

Thank you. Honestly, I confused when I make a new badge. I accepted your suggestion.

calebcartwright (approved):
Thank you!

patrikhuber:
Great to see this!

Can I also just display the logo, without the sponsors count? That would be great.

calebcartwright:
@patrikhuber, as discussed on the original issue, the badge added here is for displaying sponsor count on a badge that conforms to the Shields specification. Shields is not really in the business of providing vendor/service-styled static buttons.

If you really want to use our service to create something that looks like GitHub's button, then you could try using our static badge feature (which has always been available). However, I don't think it's going to give you what you are looking for, which again isn't really related to this PR nor the core mission of our service.

https://img.shields.io/badge/Sponsor-white?logo=github&logoColor=pink
 

https://img.shields.io/badge/Sponsor-ffffff?logo=github&logoColor=pink&style=flat-square


If you really want to use that same heart logo, then you'd either have to provide it as a custom logo via a base64 value for the query param https://github.com/badges/shields/blob/master/doc/logos.md#custom-logos or you could wait to see if the upstream PR to Simple Icons gets accepted.

patrikhuber:
@calebcartwright Thank you very much for your reply!

services/github/github-sponsors.service.js
```diff
@@ -0,0 +1,88 @@
...
+    repositoryOwner: Joi.object({
+      sponsorshipsAsMaintainer: Joi.object({
+        totalCount: nonNegativeInteger,
+      }),
+    }).allow(null),
+  }).required(),
+}).required()
+
+module.exports = class GithubSponsors extends GithubAuthV4Service {
+  static category = 'social'
```
kernc, line 22: This probably should have been [`'funding'`](https://shields.io/category/funding)?

outsideris: Oops. I missed there is funding category.

marcomontalbano:
> @patrikhuber, as discussed on the original issue, the badge added here is for displaying sponsor count on a badge that conforms to the Shields specification. Shields is not really in the business of providing vendor/service-styled static buttons.
> 
> If you really want to use our service to create something that looks like GitHub's button, then you could try using our static badge feature (which has always been available). However, I don't think it's going to give you what you are looking for, which again isn't really related to this PR nor the core mission of our service.
> 
> https://img.shields.io/badge/Sponsor-white?logo=github&logoColor=pink
> 
> 
> https://img.shields.io/badge/Sponsor-ffffff?logo=github&logoColor=pink&style=flat-square
> 
> 
> If you really want to use that same heart logo, then you'd either have to provide it as a custom logo via a base64 value for the query param https://github.com/badges/shields/blob/master/doc/logos.md#custom-logos or you could wait to see if the upstream PR to Simple Icons gets accepted.

@calebcartwright, I just noticed that the `GitHub Sponsors` icon is available at https://simpleicons.org, so another option to display just the logo without the count is the following.

https://img.shields.io/badge/-Sponsor-fafbfc?logo=GitHub%20Sponsors
GitHub Sponsors

Issue #4691, opened 2020-02-26
For-the-badge ratio is off


b3u:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

:link: **Link to the badge**
The size of the for-the-badge style is not the same size as for-the-badge itself. The ratio is different.

Compare ![forthe-badge](https://forthebadge.com/images/badges/powered-by-electricity.svg) with ![shields](https://img.shields.io/badge/Powered_By-Electricity-5d9741?labelColor=c1d72f&style=for-the-badge)

:bulb: **Possible Solution**

paulmelnikow:
Hmm, it's a different typeface too, isn't it! 😁 

This code was refactored recently and hasn't yet been merged to master. I don't think that will fix the problem though if anyone would like to work on this, it should be on the [v3-dev branch](https://github.com/badges/shields/tree/v3-dev).

Happy to accept contributions (on that branch, for the time being) which bring our badges more inline with the design at https://forthebadge.com/.

If we need to add a new typeface that'll have to be done in this repo:
https://github.com/metabolize/anafanafo

PR #5696, opened 2020-10-11
Improve appearance of for-the-badge and social badges by measuring correct font

### LINKED ISSUE #4275: Letter spacing is incorrect on social badges

- [x] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**

The social badges display with incorrect letter spacing. 

:link: **Link to the badge**

https://img.shields.io/github/stars/badges/shields?style=social

By removing the `textWidth` you can see what the letter spacing would be if the width were correct:

Correct:

Screen Shot 2019-10-31 at 6 21 00 PM

Incorrect:

Screen Shot 2019-10-31 at 6 21 18 PM

:bulb: **Possible Solution**

- Measure the text using the font used for display (I suppose Helvetica or Helvetica Neue)
- Measure the text in the correct weight (700)
- Make sure nothing else is getting in the way 🤓

paulmelnikow:
Even before #2311, Shields has measured all badge text using 11px Verdana, which is the correct font for plastic, flat, and flat-square.

However the for-the-badge style uses 10px Verdana and bold 10px Verdana, and the social badge uses bold 11px _Helvetica_. This causes these badges to be wider or narrower than they should be, and the text to be stretched accordingly.

This came up as part of some work @calebcartwright is doing on [squint](https://github.com/badges/squint/), the reimplementation of the raster proxy, in which he was having to work around this bug in the text length on the SVG badges. I'm pretty confident this will fix his issue for the social badges.

Since the for-the-badge style relies on synthetic letter spacing (we just multiply the computed length in code) there is probably room to massage this further. Our test badges all look as good or better than before, so I think it should not block merging this, however we might need to get it right before the downstream problem in squint goes away. (There's also #4691 which this doesn't attempt to address.)

Fixes #4275

----

Before and after (Firefox OS X) https://shields-staging-pr-5696.herokuapp.com/dev/styles:

Screen Shot 2020-10-11 at 5 45 16 PM

Subtle, but you can see "Python" looks better (and maybe "Tests"?), and "passing" has the same spacing when it appears on its own as it does in "5 passing, 1 failed". On img.shields.io "passing" on its own is squished, and "5 passing, 1 failed" is stretched. The padding around the message is more consistent, too.

Screen Shot 2020-10-11 at 5 45 52 PM

Pretty subjective. I think **3.5 | 3.6 | 3.7** and **5 PASSING, 1 FAILED** look better than before. Will be curious to see how these changes shake out on longer examples.


paulmelnikow:
This is the example from #4275 before:

Screen Shot 2020-10-11 at 5 56 49 PM

And after:

Screen Shot 2020-10-11 at 5 56 44 PM

The label is spaced correctly now, and the padding looks good and similar. The message is the tiniest bit wider than before, which I think is also correct, although I seem to have added padding. I think I probably should dial the padding back down, just a hair.

calebcartwright:
Can confirm this makes a huge difference on the raster side with the Social badges. The png's are still, of course, not raster-perfect, but the Social bades are markedly better with this :heart:

badge-maker/lib/badge-renderers.js
```diff
@@ -139,7 +132,7 @@ function renderText({
     return { renderedText: '', width: 0 }
   }
 
-  const textLength = preferredWidthOf(content)
+  const textLength = preferredWidthOf(content, { font: '11px Verdana' })
```
calebcartwright, line 135: Where was the font used for measurement defined previously? Was that internal to anafanafo?

badge-maker/lib/badge-renderers.js
```diff
@@ -641,7 +639,9 @@ function forTheBadge({
 
   labelWidth += 10 + totalLogoWidth
   if (label.length) {
-    labelWidth += 10 + label.length * 1.5
+    // Add 10 px of padding, plus approximately 1 px of letter spacing per
+    // character.
+    labelWidth += 10 + 2 * label.length
```
calebcartwright, line 644: Should the comment say `approximately 2px of..` or should the calc be `10 + label.length` or perhaps am I just missing something?

badge-maker/lib/badge-renderers.js
paulmelnikow: Yea, previously it was always 110px Verdana.

badge-maker/lib/badge-renderers.js
paulmelnikow: Yes, you’re totally right. I think these numbers need to be tweaked too. I’ll bring the comment in sync when it stabilizes.

calebcartwright (approved):
There's a couple very minor numerical changes that I'm not quite following, but overall seems straightforward enough and definitely has a positive visual impact on the badges (and major improvement on rasterization). 

May beneficial for someone more familiar with this part of the codebase to give it a look over, but I'm :+1: on the change

paulmelnikow:
Making the for-the-badge widths match is going to take some more work and I think it's going to be kind of messy. I'd suggest we address that in a follow-on PR.

calebcartwright:
Did you want to pull in #5714 first or is this ready for a :+1: and merging?

paulmelnikow:
Oh, sure, let's pull that in first.

calebcartwright (dismissed):

calebcartwright:
Will be around for a bit this evening if there's a need to re-approve after updating the branch

paulmelnikow:
https://shields.io/category/social provides a pretty good illustration of the impact of this change on the consistency of the social badges.

Before:

Screen Shot 2020-10-15 at 7 42 03 PM

After:

Screen Shot 2020-10-15 at 7 42 43 PM

__snapshots__/make-badge.spec.js
```diff
@@ -969,15 +969,15 @@ exports['The badge generator "for-the-badge" template badge generation should ma
...
   height="28"
   role="img"
   aria-label="CACTUS: GROWN"
 >
   <title>CACTUS: GROWN</title>
   <g shape-rendering="crispEdges">
-    <rect width="74" height="28" fill="#0f0" />
-    <rect x="74" width="73" height="28" fill="#b3e" />
+    <rect width="73" height="28" fill="#0f0" />
+    <rect x="73" width="72.5" height="28" fill="#b3e" />
```
paulmelnikow, line 980: OMG ❤️ this diff

calebcartwright (approved):

Issue #5555, opened 2020-09-19
🎃 Convert Service Classes to Static Fields 🎃


calebcartwright:
We recently completed the [upgrade to Node 12 in our production environment](https://github.com/badges/shields/pull/5436) which, among other benefits, now allows us to leverage [static fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields) for defining core attributes like the category and route. 

We've completed the refactoring of several service classes to use static fields, but there's a lot of them remaining and we could use your help to convert the rest!

## What are the steps required?

This work involves updating the members in a service class (found in the *-service.js files) to convert the static getters to static fields,

For example:
```js
  static get category() {
    return 'version'
  }
```
 
becomes
```js
  static category = 'version'
```

If a service class doesn't have any getters (e.g. redirector classes like [this one](https://github.com/badges/shields/blob/master/services/offset-earth/offset-earth-trees-redirect.service.js), then there's nothing needed for that class!

Additional information and references can be found in the corresponding section of the [tutorial documentation](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#42-our-first-badge) as well previous PRs that did this conversion like #5552 or #5530 

## How to contribute?

Below is the list of remaining service class families (as a link to their location in source) that need to be refactored. If you're interested in working on this, leave a comment on this thread or drop us a note in [Discord](https://discord.gg/HjJCwm5) letting us know which one(s) you'd like to work on. Items that have been checked off have already been completed, but the others are up for grabs!

As a rule of thumb, we'd generally prefer to see PRs contain between 2-4 refactored service class families to keep the review and merge process streamlined. You can work on as many as you'd like, but please try to follow that guidance when contributing and split PRs accordingly!

- [x] [gemnasium](https://github.com/badges/shields/tree/master/services/gemnasium)
- [x] [gitlab](https://github.com/badges/shields/tree/master/services/gitlab)
- [x] [gitter](https://github.com/badges/shields/tree/master/services/gitter)
- [x] [gratipay](https://github.com/badges/shields/tree/master/services/gratipay)
- [x] [hackage](https://github.com/badges/shields/tree/master/services/hackage)
- [x] [hexpm](https://github.com/badges/shields/tree/master/services/hexpm)
- [x] [homebrew](https://github.com/badges/shields/tree/master/services/homebrew)
- [x] [hsts](https://github.com/badges/shields/tree/master/services/hsts)
[...]

PR #5697, opened 2020-10-12
Add back [ChromeWebStore] rating

### LINKED ISSUE #5475: Badges for Google Chrome Web store rating not working

- [ ] My own instance

:beetle: **Description**

Google Chrome Web Store rating badges seems not working since a few weeks ago.

:link: **Link to the badge**

https://img.shields.io/chrome-web-store/rating-count/oenllhgkiiljibhfagbfogdbchhdchml.svg

https://img.shields.io/chrome-web-store/rating/oenllhgkiiljibhfagbfogdbchhdchml.svg

https://img.shields.io/chrome-web-store/stars/oenllhgkiiljibhfagbfogdbchhdchml.svg

crimx:
Replace chrome-web-store-item-property with [webextension-store-meta](https://www.npmjs.com/package/webextension-store-meta).

fix #5475


PyvesB:
Thanks for the contribution @crimx! Do you have a feel of how webextension-store-meta compares to chrome-web-store-item-property in terms of performance? HTML scrapping can be quite expensive, we need to more cautious when changing this specific family of badges.

crimx:
You are right. Currently it's about 30 times slower. Let me see what I can do.

crimx:
webextension-store-meta has been refactored with htmlparser2 and cheerio. Performance is now the same as chrome-web-store-item-property. Average 7.8ms per page.

PyvesB:
> 
> 
> webextension-store-meta has been refactored with htmlparser2 and cheerio. Performance is now the same as chrome-web-store-item-property. Average 7.8ms per page.

Thanks for looking into those numbers and updating the pull request!

One thing I noticed is that `require('webextension-store-meta/lib/chrome-web-store')` takes significantly longer to execute than `require('chrome-web-store-item-property')` when the server starts up. The cheerio dependency seems to be quite heavy. Here's the kind of flame graph I get when I start Shields locally:
Screenshot 2020-10-14 221049

The big red rectangle in the middle-bottom part of the graph corresponds to the Chrome badges, roughly 15% of total start up time of the entire Shields.io app, between 0.5 and 1 seconds. My Windows machine is quite old and is not necessarily representative of what the servers actually run on, so it may be useful for someone else to pull out some numbers.

If it's only this single family of badges and the hit is only at startup, it's not the end of the world in my personal opinion, but I thought I would point this out in case there are easy wins for you to look into. Other maintainers may also have stronger feelings about this, so I'll let them weigh in as well.

crimx:
Thanks for the info. The cheerio does seem quite heavy. Let me see if I can get rid of it.

calebcartwright:
> Other maintainers may have stronger feelings about this though, so I'll let them weigh in as well.

I think this goes without saying but worth repeating anyway, that we object to html scraping as an approach to get data for badges. As such in the abstract this feels like a step in the wrong direction. However, given that these were only recently deprecated and the other badges in the family use the same technique, I won't object provided the change isn't negatively impactful on our runtime.

Our runtime utilizes small, resource constrained nodes that can be scaled horizontally, so in addition to the execution time performance discussed above, I would also like to be able to have a reasonable degree of confidence that some feasible X number of these badge requests hitting a node wouldn't negatively impact the environment.

crimx:
Cheerio and other redundant dependencies has been removed. I've also added lazy parsing and cache. Parsing speed is now down to 2ms per page.

image

Comparing to chrome-web-store-item-property:

image

PyvesB:
Looks much better! Busy end of the week for me, so I'll aim at doing a more proper review during the weekend if another maintainer doesn't pick it up in the meantime. :)

services/chrome-web-store/chrome-web-store-base.js
```diff
@@ -1,22 +1,19 @@
 'use strict'
 
-const chromeWebStore = require('chrome-web-store-item-property')
+const ChromeWebStore = require('webextension-store-meta/lib/chrome-web-store')
```
PyvesB, line 3: For consistency with the other Chrome classes, it's probably best to keep it as lower camelcase, i.e. `chromeWebStore `.

services/chrome-web-store/chrome-web-store-rating.service.js
```diff
@@ -1,33 +1,131 @@
...
+
+  static get defaultBadgeData() {
+    return { label: 'rating' }
+  }
 }
 
-module.exports = [
-  deprecatedService({
-    route: {
+class ChromeWebStoreRating extends BaseChromeWebStoreRating {
```
PyvesB, line 18: Good to see this come back! 🎉

services/chrome-web-store/chrome-web-store-rating.service.js
```diff
@@ -1,33 +1,131 @@
...
+const { metric, starRating } = require('../text-formatters')
+const { NotFound } = require('..')
+const BaseChromeWebStoreService = require('./chrome-web-store-base')
 
-const commonAttrs = {
-  category: 'rating',
-  label: 'rating',
-  dateAdded: new Date('2020-09-06'),
+class BaseChromeWebStoreRating extends BaseChromeWebStoreService {
+  static get category() {
```
PyvesB, line 9: Could you please convert this to static fields? See #5555 for more information.

PyvesB (commented):
I've double-checked performance at startup and when generating badges dynamically, nothing stands out anymore. Things actually seem faster than before, so thanks for your efforts on the optimisation side. 👍🏻 

The rest of the implementation looks good to me. I've left a couple of minor comments for you to address, but we're basically there. 😉

services/chrome-web-store/chrome-web-store-base.js
crimx: > For consistency with the other Chrome classes

What does it mean? The uppercase is because `ChromeWebStore` is a class. `ChromeWebStore.load` is a static method.

services/chrome-web-store/chrome-web-store-rating.service.js
crimx: @PyvesB Done.

services/chrome-web-store/chrome-web-store-base.js
PyvesB: I think I got a little confused. Let's ignore this one!

PyvesB (approved):

Issue #5389, opened 2020-08-01
GitHub Deployments Service Test schema missing variant


calebcartwright:
Apparently the latest deployment status can be `null` while there is a deployment in progress, 

![image](https://user-images.githubusercontent.com/13042488/89107794-0f959480-d3f9-11ea-83f7-65a5c1bf0893.png)

which we account for within the service class schema accordingly:

https://github.com/badges/shields/blob/ea24784b26fc175e63cbf2265101ee9aec9fa5cf/services/github/github-deployments.service.js#L28-L32

However, our service tests do not:

https://github.com/badges/shields/blob/ea24784b26fc175e63cbf2265101ee9aec9fa5cf/services/github/github-deployments.tester.js#L6-L15

Ran into this earlier while trying to test a Dependabot PR while Dependabot/Repo Ranger were merging other PRs leading to shields-staging being deployed to during. I know this hasn't exactly been a frequent issue, but probably wouldn't hurt to make some tweaks, either by handling the `null` case more deliberately in the service class (for example, in what circumstance(s) does this occur and could we map the raw `null` to a more user-friendly badge message) or updating the service test validation schema

dominikbrandon:
Hello there, I've just opened the PR addressing this issue - #5704. Please let me know if any further work will be necessary.

homersimpsons:
Shouldn't this one be close as #5704 is merged?

PR #5704, opened 2020-10-15
[Github] Handle the case when the Github deployment status returns null

dominikbrandon:
This PR addresses an issue #5389


dominikbrandon:
I did not find any information in [the documentation](https://docs.github.com/en/free-pro-team@latest/graphql/reference/objects#deployment) about when does it return null in the `latestStatus` field. Therefore I assumed that it's empty just for a moment for new jobs and that the API will eventually return something else.

calebcartwright (approved):
Thanks!

PR #5705, opened 2020-10-15
Add [GithubDirectoryFileCount] badge

### LINKED ISSUE #4899: Badge request: Github number of files in directory filtering by type and/or extension
:clipboard: **Description**

For Github.
Include number of files inside repository directory.
Must filter by file extension and/or type. Valid types would be `file`, `dir` or without filtering (both types included).

:link: **Data**

Data from [here](https://api.github.com/repos/badges/shields/contents/services) using the V3 Github API.
API key is optional.
[Endpoint documentation](https://developer.github.com/v3/repos/contents/#get-contents).

:microphone: **Motivation**
This feature will allow us to show how many services are supported currently for bages calling to [this endpoint](https://api.github.com/repos/badges/shields/contents/services) and filtering by `type="dir"`. For example:

hjagodzinski:
Resolves #4899.


services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+  static examples = [
+    {
+      title: 'GitHub repo directory file count',
+      pattern: ':user/:repo/:path',
+      namedParams: { user: 'badges', repo: 'shields', path: 'services' },
+      staticPreview: this.render({ count: 10 }),
+      documentation,
+    },
+    {
+      title: 'GitHub repo directory file count (dir type)',
```
calebcartwright, line 44:
```suggestion
      title: 'GitHub repo directory count',
```

Will probably also want a non-required `:path` param example for this one and the files as well

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+      pattern: ':user/:repo/:path',
+      namedParams: { user: 'badges', repo: 'shields', path: 'services' },
+      queryParams: { type: 'dir' },
+      staticPreview: this.render({ count: 8 }),
+      documentation,
+    },
+    {
+      title: 'GitHub repo directory file count (file type)',
```
calebcartwright, line 52:
```suggestion
      title: 'GitHub repo file count (file type)',
```

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+      pattern: ':user/:repo/:path',
+      namedParams: { user: 'badges', repo: 'shields', path: 'services' },
+      queryParams: { type: 'file' },
+      staticPreview: this.render({ count: 2 }),
+      documentation,
+    },
+    {
+      title: 'GitHub repo directory file count (file extension)',
```
calebcartwright, line 60:
```suggestion
      title: 'GitHub repo file count (file extension)',
```

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+      staticPreview: this.render({ count: 1 }),
+      documentation,
+    },
+  ]
+
+  static defaultBadgeData = { color: 'blue', label: 'files' }
+
+  static render({ count }) {
+    return {
+      message: count,
```
calebcartwright, line 73: Let's use the `metric` helper (can be imported from `../text-formatters`) here to ensure proper and consistent formatting of large numbers. I'd hope there's not a ton of directories that have more than 1,000 files but I'm sure there's some out there :smile:

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+      url: `/repos/${user}/${repo}/contents/${path}`,
+      schema,
+      errorMessages: errorMessagesFor('repo or directory not found'),
+    })
+  }
+
+  static transform(files, { type, extension }) {
+    if (type) {
+      files = files.filter(file => file.type === type)
+    }
```
calebcartwright, line 88: Technically a user could provide a `type` of `dir` but also provide an `extension` param value. In this case I feel like we should either throw an `InvalidParameter` error outright, or return early since there's not really a need to iterate through the collection again to further filter by file extension when the collection only consists of directories

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+    return {
+      count: files.length,
+    }
+  }
+
+  async handle({ user, repo, path }, { type, extension }) {
+    const content = await this.fetch({ user, repo, path })
+    if (!Array.isArray(content)) {
+      throw new InvalidParameter({ prettyMessage: 'not a directory' })
+    }
```
calebcartwright, line 101: Let's move this check to `transform`. It's a more logical fit from a separation of duties, and also simplifies the testing story since `transform` functions are much simpler to unit test

services/github/github-directory-file-count.tester.js
```diff
@@ -0,0 +1,75 @@
+'use strict'
+
```
calebcartwright, line 2: We can use the `isMetric` helper here for validating the badge messages
```suggestion

const { isMetric } = require('../test-validators')
```

services/github/github-directory-file-count.tester.js
```diff
@@ -0,0 +1,75 @@
...
+    label: 'files',
+    message: '3',
+  })
+
+t.create('directory file count (invalid type)')
+  .get('/badges/shields/services.json?type=submodule')
+  .expectBadge({
+    label: 'files',
+    message: 'invalid query parameter: type',
+  })
```
calebcartwright, line 67: We can remove this test, as the param validation infrastructure is already well tested elsewhere

services/github/github-directory-file-count.tester.js
```diff
@@ -0,0 +1,75 @@
...
+    message: '2',
+  })
+
+t.create('directory file count (file type)')
+  .get('/badges/shields/services.json?type=file')
+  .intercept(mockContents)
+  .expectBadge({
+    label: 'files',
+    message: '3',
+  })
```
calebcartwright, line 60: I think these (and the other mocked one at the bottom) would probably be better suited to unit tests vs. service tests with intercepts on mocks. For example from the Codecov service:

https://github.com/badges/shields/blob/master/services/codecov/codecov.spec.js

calebcartwright (commented):
Thank you for the PR @hjagodzinski! This is in pretty good shape, left a few minor things inline that I think we'll want to tweak and/or shuffle

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+  Joi.object({}).required(),
+  Joi.array()
+    .items(
+      Joi.object({
+        path: Joi.string().required(),
+        type: Joi.string().required(),
+      })
+    )
+    .required()
+)
```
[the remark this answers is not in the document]
hjagodzinski: I've explained it already well enough I used slightly altered you explanation in the comment.

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+
+  static route = {
+    base: 'github/directory-file-count',
+    pattern: ':user/:repo/:path*',
+    queryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'GitHub repo directory file count',
```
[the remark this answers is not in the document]
hjagodzinski: Sure, I've added the example.

services/github/github-directory-file-count.service.js
hjagodzinski: I've added the examples.

services/github/github-directory-file-count.service.js
hjagodzinski: Indeed `transform` seems more suitable 👍

services/github/github-directory-file-count.tester.js
hjagodzinski: My intension is rather to test that unexpected values (`valid('dir', 'file')`) passed to `type` will result in an error than test the infrastructure itself.

services/github/github-directory-file-count.tester.js
hjagodzinski: Unit tests will cover only `transform` method logic but won't validate the correct values are passed to it by `handle` (namely `type` and `extension`). Therefore service tests provide greater confidence the code behaves as expected. If I wrote the unit tests for `transform` solely and then removed passing `{ type, extension }` to `transform` altogether, no test would turn red.

services/github/github-directory-file-count.service.js
hjagodzinski: I intentionally did not restrict the use of extension to `file` type only. I thought someone might find it helpful to be able to filter directories by extension if needed. I have never come across such a need but I thought leaving it will keep code simpler and might help someone. Since I don't have strong feelings about this I can change the code as you suggested.

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,105 @@
...
+      staticPreview: this.render({ count: 2 }),
+      documentation,
+    },
+    {
+      title: 'GitHub repo directory file count (file extension)',
+      pattern: ':user/:repo/:path',
+      namedParams: { user: 'badges', repo: 'shields', path: 'services' },
+      queryParams: { extension: 'js' },
+      staticPreview: this.render({ count: 1 }),
+      documentation,
```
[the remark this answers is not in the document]
hjagodzinski: I've added notes to documentation on both `type` and `extension` parameters.

hjagodzinski:
@calebcartwright thank you for reviewing my PR and your feedback! I've covered most of your comments with changes to the code according to your suggestions. In the remaining ones I left my replies. Let me know what you think.

hjagodzinski:
I like to keep names in the code in accordance with the natural language I use to describe it. Since you suggested changing in the examples _directory file count_ to _file count_ I wonder wether not change `GithubDirectoryFileCount` to `GithubFileCount`. What do you think?

services/github/github-directory-file-count.tester.js
calebcartwright: I understand that, and what I'm saying is that is already well tested elsewhere within the service class and input param validation code. We'd have a _ton_ of duplicative service tests (that are primarily used as integration tests because of the myriad external platforms/services/APIs we call) if we included tests like this for each and every service that used params. 

I definitely understand the instinct to add such a test here with the lens of validating this one class, but in the broader context of Shields, each and every service test has an impact both in terms of code maintenance and execution during our daily test runs.

services/github/github-directory-file-count.tester.js
calebcartwright: I understand how the code flow works :wink: 

As stated above, our service tests are primarily used for validation of integrations with external endpoints and high level badge function. We integrate with a ton of platforms that can and do change their API contracts at will (we don't get notifications preemptively :smile:), and the service tests help us know whether or not we're still correctly interfacing with a service and able to render a badge.

We don't want to use service tests for automated verification of the interaction/call sites from one internal function to another. Again, I understand your reasoning and instinct, but it's a matter of tradeoffs. Within the Shields codebase and the contexts and constraints we work with, we favor and trust ourselves as maintainers to not merge and deploy such an obviously goofy change vs. utilizing our service tests to also verify all code-level component/function interactions.

This is fine for now, but please keep this in mind on any potential future PRs!

calebcartwright:
> Since you suggested changing in the examples _directory file count_ to _file count_ I wonder wether not change `GithubDirectoryFileCount` to `GithubFileCount`. What do you think?

My suggestion was to aid in cases where a user is browsing the example listings on Shields.io and wanting to have a clearer distinction between the examples targeting file counts vs. directory counts since the badge will display one or the other. I think having only of those words in the example listing helps better convey that and makes them more visually distinct.

The name of the service classes is typically more representative of the ServiceNoun pattern, and doesn't have to be strictly coupled with the text of examples which are targeted for end users and thus use more natural language. I guess the noun in this case is more of a `ContentsCount` or `ContentCount`, with Content having a file variant and directory variant.

I'm less concerned with the non-user facing class name and DirectoryFileCount is fine with me, though certainly open to an alternative if you can think of a better term!

services/github/github-directory-file-count.service.js
calebcartwright: Within the context of a GitHub repo, what will qualify as a folder/directory extension? Wouldn't this just functionally be a filter of directories that happened to have a name ending with some `.foobar`?

I do have some worries that this endpoint in particular could be abused, so if we can prevent an extra loop over the response, and processing step for each entry, then I'd really like to do so, especially since there's not a strong valid use case for it.

hjagodzinski: Exactly, it would boil down to filtering directories with extension. I have restricted the use of `extension` parameter to `type` `file`. Either passing no `type` altogether or passing `type` `dir` will result in an error.

services/github/github-directory-file-count.tester.js
hjagodzinski: You explanation have broadened my perspective. I see where you're coming from. The service tests comes with a trade offs. Your points as valid. I've removed this test.

services/github/github-directory-file-count.tester.js
hjagodzinski: I was not my intention to imply your lack of understanding. I was merely sharing my concerns :) Again, I have better understanding where you're coming from.

I've moved mocked service tests to unit tests.

hjagodzinski:
It's a valid point that the most important aspect of naming is keeping the external compatibility. A potential cost of an internal class name change is incomparably lower. Nevertheless I will keep the current class name.

Thank you for taking time to explain to me the reasoning behind your comments. I appreciate that.

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,158 @@
...
+<p>
+  <b>Note:</b><br>
+  1. Parameter <code>type</code> accepts either <code>file</code> or <code>dir</code> value. Passing any other value will result in an error.<br>
+  2. Parameter <code>extension</code> accepts file extension without a leading dot.
+     For instance for <code>.js</code> extension pass <code>js</code>.
+     Only single <code>extension</code> value can be specified.
+     <code>extension</code> is applicable for <code>type</code> <code>file</code> only.
+     Passing it either without <code>type</code> or along with <code>type</code> <code>dir</code> will result in an error.<br>
+  3. GitHub API has an upper limit of 1,000 files for a directory.
+     In case a directory contains files above the limit, a badge might present inaccurate information.<br>
```
calebcartwright, line 22: I think this is a reasonable limitation, especially since the `contents` API seems much like it'll be easier for us to work with than the alternative `trees` API and how uncommon a >1k path and badge scenario would be :+1:

services/github/github-directory-file-count.service.js
```diff
@@ -0,0 +1,158 @@
...
+</p>
+`
+
+const schema = Joi.alternatives(
+  /*
+   alternative empty object schema to provide a custom error message
+   in the event a file path is provided by the user instead of a directory
+  */
```
calebcartwright, line 30: :heart:

services/github/github-directory-file-count.service.js
calebcartwright: Still want to use `metric(count)` for the value here. I suppose that alternatively we could check if count is equal to 1000 and conditionally change the value to `1k` instead, but would much rather use our existing helper to ensure consistency across the project

calebcartwright (commented):
Final thing is using the `metric` formatter, then ready for merging!

calebcartwright (approved):
:+1:

Issue #4911, opened 2020-04-15
Support architecture specification in DockerVersion badge


calebcartwright:
:clipboard: **Description**
Our Docker Version badge currently defaults to leveraging the `amd64` architecture in the image tags to determine the correct version to show on the badge.

While this is a reasonable default, it'd also be good if users could override/specify the desired architecture to use in this process so that the version badge will still be useful in other scenarios.

Refs https://github.com/badges/shields/pull/4907#issuecomment-614200710
Refs https://github.com/badges/shields/pull/4562#discussion_r372161380

nightah:
I completely forgot about this!
I'm happy to address this soon, when I manage to find some spare time if nobody else picks it up in the interim.

vividh:
I'd like to work on this. Can I get some pointers regarding the changes to be made here, and how to verify the fix?

chris48s:
At the moment, amd64 is hard-coded here: https://github.com/badges/shields/blob/e03549d88d3a038db1ee3730ee4ca84a7abeea77/services/docker/docker-version.service.js#L85
I think it makes most sense to pull it out to a `queryParam`, like `sort`
https://github.com/badges/shields/blob/e03549d88d3a038db1ee3730ee4ca84a7abeea77/services/docker/docker-version.service.js#L116
https://github.com/badges/shields/blob/e03549d88d3a038db1ee3730ee4ca84a7abeea77/services/docker/docker-version.service.js#L30
which defaults to amd64 (but can accept to other architectures).

Error handling needs to cover invalid value supplied for architecture. 

I think a good example to test against would be something like raspbian/desktop which doesn't have have amd64 builds https://hub.docker.com/r/raspbian/desktop/tags that doesn't work at the moment but adding an arch param should enable us to generate a badge for that. You may also be able to find an example of a docker image that has different versions for different architectures but I don't know of one. You could add a mocked test that replicates that condition though.

vividh:
Thanks! I'll get cracking on this, and will revert soon with an update.

vividh:
@chris48s Made the changes in https://github.com/badges/shields/pull/5772. Let me know if it's acceptable or any changes are required there.

vividh:
@chris48s Ping!

calebcartwright:
Thank you for the PR and for your patience @vividh. There's quite a lot of PRs pending review at the moment, and as maintainers we're all volunteers with limited time :smile: One or more maintainers should get to your PR soon

PR #5707, opened 2020-10-15
fix: TypeError in [DockerVersion] badge

platan:
Fixes https://github.com/badges/shields/issues/5535. This solution is similar to https://github.com/badges/shields/pull/4907. Still waiting for https://github.com/badges/shields/issues/4911 :-)


services/docker/docker-version.spec.js
```diff
@@ -78,4 +78,31 @@ describe('DockerVersion', function () {
       .to.throw(InvalidResponse)
       .with.property('prettyMessage', 'digest not found for latest tag')
   })
+
+  it('throws InvalidResponse error with custom tag and no amd64 architecture digests', function () {
```
paulmelnikow, line 82:
```suggestion
  // https://github.com/badges/shields/issues/5535
  it('throws InvalidResponse error with custom tag and no amd64 architecture digests', function () {
```

paulmelnikow (dismissed):
Thanks for picking this up!

paulmelnikow (approved):

PR #5708, opened 2020-10-15
Added [GithubRPackageVersion] badge

### LINKED ISSUE #4368: Badge request for GH version of R packages from the DESCRIPTION file
:clipboard: **Description**

Thanks for the great work :)

It would be nice if R package developers could get public GH package version badges for their development versions from the DESCRIPTION file's line with structure: `Version: ${package_version} \n`

:link: **Data**

- The DESCRIPTION file's path is in https://github.com/${user}/${repo}/blob/${branch}/DESCRIPTION

- The policy that requires the aforementioned `Version: ` line structure can be found in http://r-pkgs.had.co.nz/description.html. 

- An example can be found [here](https://github.com/mixOmicsTeam/mixOmics/blob/RELEASE_3_10/DESCRIPTION).

:microphone: **Motivation**

It would be great if users could also have badges for their latest stable version not yet submitted on CRAN or Bioconductor so users can easily see if they need to update for the latest features or not.

hjagodzinski:
Resolves #4368.


services/github/github-go-mod.service.js
```diff
@@ -45,23 +45,14 @@ module.exports = class GithubGoModGoVersion extends ConditionalGithubAuthV3Servi
...
     },
-    {
-      title: 'GitHub go.mod Go version (branch & subfolder of monorepo)',
-      pattern: ':user/:repo/:branch',
-      namedParams: { user: 'golang', repo: 'go', branch: 'master' },
-      queryParams: { filename: 'src/go.mod' },
-      staticPreview: this.render({ version: '1.14' }),
-      documentation,
-      keywords,
-    },
```
calebcartwright, line 64: It's not clear to me why we'd remove the example that allows users to specify a custom branch. Could you explain the rationale for removing this?

services/github/github-go-mod.tester.js
```diff
@@ -15,7 +15,7 @@ t.create('Go version (from branch)')
     message: isVPlusDottedVersionAtLeastOne,
   })
 
-t.create('Go version (mongorepo)')
+t.create('Go version (monorepo)')
```
calebcartwright, line 18: :rofl: good catch!

services/github/github-r-package.tester.js
```diff
@@ -0,0 +1,60 @@
...
+  ].join('\n')
+
+  return nock('https://api.github.com')
+    .get(`/repos/mixOmicsTeam/mixOmics/contents/${path}`)
+    .query({ ref: 'HEAD' })
+    .reply(200, {
+      content: Buffer.from(content, 'utf-8').toString('base64'),
+      encoding: 'base64',
+    })
+}
```
calebcartwright, line 21: Similarly to my comment in https://github.com/badges/shields/pull/5705/files#r505937638, I think using standard unit tests against the `transform` function would probably be more lightweight and straightforward

calebcartwright (commented):
Awesome thank you for the PR! Few minor items left inline for your review

services/github/github-go-mod.service.js
hjagodzinski: Minimalizm :) I've contemplated that since the examples on specifying both `branch` and `filename` are already there, there is no need for this one. Reading your suggestions from other PR I realized someone might use this example to generate an URL. Hence I restored this example and added corresponding to the new badge.

services/github/github-go-mod.tester.js
hjagodzinski: My but was telling me it's about Mongo 🤣

services/github/github-r-package.tester.js
hjagodzinski: In the referenced comment I've shared my thought on that. Let's establish there the final way to go about those tests.

hjagodzinski:
@calebcartwright thank you for reviewing my PR and your feedback!

services/github/github-r-package.tester.js
hjagodzinski: Done.

calebcartwright (approved):
LGTM, thank you!

ajabadi:
@hjagodzinski Thank you for implementing this! I could not see any updates on usage examples. Can you please provide one?

calebcartwright:
@ajabadi - merged changes are not immediately deployed to production, as noted by the bot in the immediately preceeding comment 

https://github.com/badges/shields/pull/5708#issuecomment-711103593

You can browse the changes in staging though
https://shields-staging.herokuapp.com/

https://shields-staging.herokuapp.com/github/r-package/v/mixOmicsTeam/mixOmics

ajabadi:
Thank you @calebcartwright for clarifying and sorry I missed the bot's note.

calebcartwright:
no worries, they can be easy to miss!

ajabadi:
@hjagodzinski I think it would be nice if the badge read `github(@branch) | version` instead of `R(@branch) | version` because it would be rather general knowledge that is an `R` package and the badge version is intended to contrast the GitHub version to that of other hosting platforms (e.g. CRAN or Bioconductor).

calebcartwright:
@ajabadi - I appreciate you sharing your perspective, but what you are asking for does not conform to our badge specification and would also be inconsistent with the way we display this information on our badges. Please also note that a closed/completed PR is not really the place for requesting features/changes. You can always open a new issue, open a Discussion item, and/or ping us on Discord.

Remember that you can utilize the `?label=foo` query parameter to set the label to whatever you want, but the default label value here is what it is supposed to be

https://shields-staging.herokuapp.com/github/r-package/v/mixOmicsTeam/mixOmics/master?label=github(@master)
)

ajabadi:
@calebcartwright that's great! I did speculate that there might be a label feature but could not recall the trick. Thanks.

PR #5709, opened 2020-10-15
Add a deployment status using GitHub Actions

platan:
2 years ago we created a bot that adds a comment the end of the PR with deployment status (https://github.com/badges/shields/issues/1708). The bot (https://github.com/apps/shields-deployment) was a https://github.com/bkeepers/workflow app running at ZEIT Now 1.0. 

This year zeit.co announced Zeit Now 1.0 shutdown. On August 7, 2020, all Zeit Now 1.0 deployments were deleted. 

Just before the shutdown, I moved Workflow instance to AWS Lambda. On the one hand it was a good move, because:
- we have few requests to handle and serverless is great in this case
- it's "free" in our case - `The AWS Lambda free usage tier includes 1M free requests per month (https://aws.amazon.com/lambda/pricing/)
- https://github.com/probot/serverless-lambda allowed to easily move to AWS Lambda (Workflow uses Probot)

On the other hand, I have some problems with the Workflow/serverless-lambda, because they use incompatible versions of Probot and it required some hacks to make them work properly (the result is stored in my private repo).

In this PR I remove Probot's configuration and migrate the whole solution to GitHub Actions. 

https://github.com/platan/shields-deployment-status-example/pull/55 is an example PR using [similar](https://github.com/platan/shields-deployment-status-example/blob/master/.github/workflows/add_deployment_status.yml) GitHub Action 

Zrzut ekranu 2020-10-15 o 17 57 43


chris48s (approved):
nice. Simplifying this seems like a good move :+1:

calebcartwright:
Indeed, well done! I like having it defined and being able to manage entirely within repo too

PR #5711, opened 2020-10-15
unskip token integration test

chris48s:
refs #4590
Lets see if this is now stable


paulmelnikow (approved):

services/github/github-api-provider.integration.js
```diff
@@ -79,7 +79,7 @@ describe('Github API provider', function () {
       }
     })
 
-    it.skip('should update the token with the final limit remaining and reset time', function () {
+    it('should update the token with the final limit remaining and reset time', function () {
```
paulmelnikow, line 82: Is it worth including this link here, for the next time it fails on someone?

```suggestion
    // Is this test failing? See https://github.com/badges/shields/pull/4590#issuecomment-708551801
    it('should update the token with the final limit remaining and reset time', function () {
```

paulmelnikow (approved):

PR #5714, opened 2020-10-15
Prettify the SVG snapshots (and a couple cleanups)

paulmelnikow:
I’m running into some strange issues working on badge-maker in #5696, where the images look good but the snapshots have changed in unexpected ways. I’m finding them a little tricky to debug.

This saves the prettified snapshots instead, which I think will make them easier to compare and debug. They may have to be updated when we upgrade prettier.

I also inlined a few things in the test file.


__snapshots__/make-badge.spec.js
```diff
@@ -1,135 +1,1947 @@
 exports['The badge generator SVG should match snapshot 1'] = `
```
calebcartwright, line 1: I gotta be honest, can't see myself doing a detailed line-by-line review of this file :smile: I assume there's no additional delta beyond the prettification and that our tests give us confidence that prettier didn't somehow botch them?

paulmelnikow: Oh, yea, you mean in this PR. 😁 Yea, the changes to that file were all auto-generated.

calebcartwright (approved):

PR #5716, opened 2020-10-16
A few straight refactors in badge-renderers

paulmelnikow:
These are aimed at improving readability and a bit of DRY.

### DIFF
--- a/badge-maker/lib/badge-renderers.js
+++ b/badge-maker/lib/badge-renderers.js
@@ -84,7 +84,7 @@ function renderLogo({
     return {
       hasLogo: true,
       totalLogoWidth: logoWidth + logoPadding,
-      renderedLogo: `<image x="${x}" y="${y}" width="${logoWidth}" height="14" xlink:href="${escapeXml(
+      renderedLogo: `<image x="${x}" y="${y}" width="${logoWidth}" height="${logoHeight}" xlink:href="${escapeXml(
         logo
       )}"/>`,
     }
--- a/badge-maker/lib/badge-renderers.js
+++ b/badge-maker/lib/badge-renderers.js
@@ -14,14 +14,9 @@ function capitalize(s) {
 
 function colorsForBackground(color) {
   if (brightness(color) <= brightnessThreshold) {
-    return {
-      textColor: '#fff',
-      shadowColor: '#010101',
-    }
-  }
-  return {
-    textColor: '#333',
-    shadowColor: '#ccc',
+    return { textColor: '#fff', shadowColor: '#010101' }
+  } else {
+    return { textColor: '#333', shadowColor: '#ccc' }
   }
 }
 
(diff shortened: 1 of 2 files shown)


badge-maker/lib/badge-renderers.js
```diff
@@ -82,22 +77,19 @@ function renderLogo({
...
+    const logoHeight = 14
+    const y = (badgeHeight - logoHeight) / 2
+    const x = horizPadding
     return {
-      hasLogo: false,
-      totalLogoWidth: 0,
-      renderedLogo: '',
+      hasLogo: true,
+      totalLogoWidth: logoWidth + logoPadding,
+      renderedLogo: `<image x="${x}" y="${y}" width="${logoWidth}" height="14" xlink:href="${escapeXml(
```
calebcartwright, line 87: Not strictly related to the changes here, but is the height attribute intended to be aligned with `logoHeight` or do they both just coincidentally have a value of `14`? If the former, is there any reason why the attribute has an inline value instead of using the variable?

badge-maker/lib/badge-renderers.js
```diff
@@ -728,17 +697,12 @@ function forTheBadge({
       ${renderMessageText()}
     </g>`
   )
-
-  if (minify) {
-    return stripXmlWhitespace(badge)
-  }
```
calebcartwright, line 734: Could this change potentially have any behavioral impacts? Or was `minify` always truthy for all badge styles previously?

paulmelnikow: No, I don't see how. `badge-renderers.js` is only used in `make-badge.js`, and in the only spot where the function is invoked it gets `minify: true`.

badge-maker/lib/badge-renderers.js
paulmelnikow: Yes, I suspect it probably should be using `${logoHeight}`. It seems okay to make that change here.

badge-maker/lib/badge-renderers.js
chris48s: If I ever have to read the SVG outputs by eye, I usually flip `minify` off. How do you normally deal with that?

There was also a point where we were using it to make the snapshot tests more readable https://github.com/badges/shields/pull/4459#discussion_r362106126 but then we switched to minified snapshots later, although I notice you've taken another approach to that elsewhere which should be less sensitive to whitespace changes, so hopefully best of both worlds there.. :+1:

paulmelnikow: Ah, interesting. My workflow for that hasn't been great. Usually I save out the svg's from the browser and then open them in VS Code and format them.

I'd be happy to add an env var that turns off the minification, if you think that would be useful.

The formatted snapshots seem to be super nice, though! I imagine that may end up being sufficient.

chris48s: Nah. Thinking about it,

```diff
-  return stripXmlWhitespace(
+  return (
```

isn't exactly much more effort than

```diff
-    minify: true,
+    minify: false,
```

on the odd occasion we want that - Its not that common.

calebcartwright: I didn't make it down to `make-badge.js` last night, would've been good to do so beforehand as I would've seen `minify` was always true :laughing:

calebcartwright (approved):
:+1:

PR #5717, opened 2020-10-16
[Wordpress] Added Theme Required WP Version Badge & refactor

JoeIzzard:
Related to #5493, this PR refactors the existing class for Plugin required WP version to also cover themes. Also in this is a slight refactor of Plugin Tested Version class to convert to static props.

Sorry the #5493 split up has been pending a while, been fairly busy with work and starting uni!

### DIFF
--- a/services/wordpress/wordpress-base.js
+++ b/services/wordpress/wordpress-base.js
@@ -14,6 +14,7 @@ const themeSchema = Joi.object()
     downloaded: nonNegativeInteger,
     active_installs: nonNegativeInteger,
     requires_php: stringOrFalse.required(),
+    requires: stringOrFalse.required(),
   })
   .required()
 
--- a/services/wordpress/wordpress-platform.service.js
+++ b/services/wordpress/wordpress-platform.service.js
@@ -62,9 +62,7 @@ function WordpressRequiresVersion(extensionType) {
 }
 
 class WordpressPluginTestedVersion extends BaseWordpress {
-  static get category() {
-    return 'platform-support'
-  }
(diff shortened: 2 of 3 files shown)


services/wordpress/wordpress-base.js
```diff
@@ -13,6 +13,7 @@ const themeSchema = Joi.object()
     num_ratings: nonNegativeInteger,
     downloaded: nonNegativeInteger,
     active_installs: nonNegativeInteger,
+    requires: stringOrFalse.required(),
```
calebcartwright, line 16: Just want to confirm this key will always exist for themes? Know we had an overly strict schema issue here recently but can't remember if that was due to `requires` not existing or string vs string/bool

calebcartwright (commented):
Minor confirmation question on the schema, otherwise LGTM

services/wordpress/wordpress-base.js
JoeIzzard: It should always exist. If it has an actual value, it will return it, if not it will return false.

I think you may be thinking of issue #5499, which was a bug caused by the old schema only expecting a string and breaking if the theme wasn't set to usable data. That should have been resolved with PR #5579 which introduced the stringOrFalse.

calebcartwright:
@JoeIzzard  when you get a chance can you rebase and address the merge conflicts? I imagine there will be a few minor ones on most/all of these PRs given the touches to the shared schema. The changes LGTM to so should be good to merge once conflicts are resolved

calebcartwright (approved):
Thank you!

PR #5718, opened 2020-10-16
Fail in CI when package tests fail

### LINKED ISSUE #5713: Package tests fail silently in CI
I was looking at #5696 and had a hard time making sense of the snapshot diff, which seemed like it had totally wrong values in it.

It turns out the snapshot is out of date, and the tests are failing in CI but showing as succeeded.

https://app.circleci.com/pipelines/github/badges/shields/3473/workflows/df50bd0c-8a58-4f20-9a75-9fd279fd6ea1/jobs/121024

One pretty easy way to fix this would be to quit after the first failure, although it would mean not testing any of the subsequent Node versions.

paulmelnikow:
Fixes #5713

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,7 @@ run_package_tests: &run_package_tests
   when: always
   command: |
     # https://discuss.circleci.com/t/switch-nodejs-version-on-machine-executor-solved/26675/3
-    set +e
+    set -e
     export NVM_DIR="/opt/circleci/.nvm"
     [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
     nvm install $NODE_VERSION
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,39 +86,6 @@ services_steps: &services_steps
     - store_test_results:
         path: junit
 
-run_package_tests: &run_package_tests
-  when: always
-  command: |
-    # https://discuss.circleci.com/t/switch-nodejs-version-on-machine-executor-solved/26675/3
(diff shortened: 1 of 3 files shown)


calebcartwright:
I assume the test failures were deliberate? a test of the tests as it were :smile:

paulmelnikow:
@chris48s do you see any reason not to make this change?

calebcartwright (approved):
Approving as this LGTM, though agreed with getting feedback from Chris given the historical context on the setup

chris48s:
> @chris48s do you see any reason not to make this change?

Nope. I have a strongly pro-"build failing if the tests fail" stance :smile: Thanks for picking this one up. I spotted the issue and was going to have a look at it tonight. Looking at the git blame, I think `set +e` is basically a typo :grimacing: it should have always been `set -e`

I've actually very recently started using `set -euo pipefail` in preference to `set -e` since I read https://wizardzines.com/comics/bash-errors/ You could do that here too, if you wanted. Happy to re-:+1:  if needed

chris48s (dismissed):

chris48s:
maybe being extra strict was overly optimistic :laughing:

paulmelnikow:
This is working now. I think the script is a little easier to edit in its own file but I can put it back into the Circle config if you prefer.

chris48s:
>  I think the script is a little easier to edit in its own file but I can put it back into the Circle config

But bash-script-in-a-string-in-a-yaml-file is my favourite programming language :laughing: 

I'm going to manually squash this one rather than label it just to make some sense of the commit message

chris48s (approved):

PR #5719, opened 2020-10-16
Align `label` and `message` props of `coalesceBadge()` with internal `makeBadge()`

paulmelnikow:
There is an internal `makeBadge()` function which is called from a few places in the server and from the public `makeBadge()` function which is a light wrapper. (Eventually we want to dogfood the public API: that's the work of #4950, and this helps with it by aligning the interfaces.)

Related to that is #3370, which is about aligning the `serviceData` schema (i.e. the result of `handle()`) with the public `makeBadge()` function.

A legacy quirk of the _private_ `makeBadge()` function is accepting a `text: ['label', 'message']` array instead of separate `{ label, message }` props like the rest of the codebase. `coalesceBadge()` has to translate from `{ label, message }` to `text: ['label', message']`. This removes that bit of indirection.

It also rewrites most of the tests of `coalesceBadge()` to use `.includes()`, providing IMO a slight improvement in readability.

This is cherry-picked from #4980 but stops short of the more substantive (and maybe controversial) changes of that PR.

### DIFF
--- a/core/base-service/base-svg-scraping.spec.js
+++ b/core/base-service/base-svg-scraping.spec.js
@@ -6,10 +6,6 @@ const Joi = require('joi')
 const makeBadge = require('../../badge-maker/lib/make-badge')
 const BaseSvgScrapingService = require('./base-svg-scraping')
 
-function makeExampleSvg({ label, message }) {
-  return makeBadge({ text: ['this is the label', 'this is the result!'] })
-}
-
 const schema = Joi.object({
   message: Joi.string().required(),
 }).required()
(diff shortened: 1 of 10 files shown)


core/base-service/base-svg-scraping.spec.js
```diff
@@ -6,10 +6,6 @@ const Joi = require('joi')
 const makeBadge = require('../../badge-maker/lib/make-badge')
 const BaseSvgScrapingService = require('./base-svg-scraping')
 
-function makeExampleSvg({ label, message }) {
-  return makeBadge({ text: ['this is the label', 'this is the result!'] })
-}
```
paulmelnikow, line 11: 🙄

chris48s (approved):
LGTM :+1:

Issue #5573, opened 2020-09-19
Moment.js Maintenance Mode Announcement


calebcartwright:
Some interesting news on Moment.js being deprecated (sort of). I don't see a need for any action, at least in the short term, given that they've said they'll continue to address security issues. However, opening for awareness and discussion.

https://momentjs.com/docs/#/-project-status/

> We now generally consider Moment to be a legacy project in maintenance mode. It is not dead, but it is indeed done.

> In practice, this means:

>    - We will not be adding new features or capabilities.
>    - We will not be changing Moment's API to be immutable.
>    - We will not be addressing tree shaking or bundle size issues.
>    - We will not be making any major changes (no version 3).
>    - We may choose to not fix bugs or behavioral quirks, especially if they are long-standing known issues.

> However, since we understand that Moment is well established in millions of existing projects:

>   - We will address critical security concerns as they arise.
>   - We will release data updates for Moment-Timezone following IANA time zone database releases.

chris48s:
Yeah I also spotted this the other day. I'd agree with your perspective. There is no rush to migrate given it will continue to receive security updates. That said, our usage of moment is pretty light: we require it in less than 10 files, we're not heavily tied to it by another dependency which forces us to use it, and its not part of our service framework so if we do want to migrate to _something else_ its not an enormous job like request is, for example.

paulmelnikow:
Yea, it should be an easy swap.

I'm curious, does anyone have experience with Day.js or the other alternatives?

calebcartwright:
> I'm curious, does anyone have experience with Day.js or the other alternatives?

Nope, I've only ever used Moment

calebcartwright:
Nearly 2 years on and updates are still coming (albeit small and far between) #8170

chris48s:
Yes - moment does still receive major bugfixes, just not active development. It is one of the reasons why migrating away from it has fallen by the wayside (another being that it is not completely obvious what to migrate _to_ ).

DenverCoder1:
dayjs has a very similar API to moment so is likely the easiest to migrate to. I've migrated from moment to dayjs before and for many things it's as simple as just replacing `moment()` with `dayjs()`.

PR #5722, opened 2020-10-16
[Wordpress] Theme & Plugin Last Updated badge

JoeIzzard:
Related to #5493, adds the new last update badge for both plugins and themes.


services/wordpress/wordpress-base.js
```diff
@@ -13,6 +13,7 @@ const themeSchema = Joi.object()
     num_ratings: nonNegativeInteger,
     downloaded: nonNegativeInteger,
     active_installs: nonNegativeInteger,
+    last_updated: Joi.string().required(),
```
calebcartwright, line 16: More schema questions, but not familiar with the service/API details in this case. Will this field strictly and always exist, regardless of things like version, whether the theme (in this case, plugin below) has only been published once without any updates, etc.?

calebcartwright: Also, I feel like given the utilization of this value down below in the transform function of the service class, we probably want to tighten up this schema a bit to account for the shape of the string that the service class is expecting.

We're in the process of trying to write down our guidelines for schemas, but this is one of the circumstances we've been discussing.

cc @chris48s for any additional thoughts on schema structure

JoeIzzard: In terms of if this will always exist, I believe so. I think it just returns the last action date, so for a brand new plugin/theme with no updates, it should return the initial publish date. I can't say for sure though as most of the Wordpress API isn't documented past endpoints you can query.

As for the transform point, just to note that the plugin and theme return different formatting, so we need to keep that in mind.

calebcartwright: > As for the transform point, just to note that the plugin and theme return different formatting, so we need to keep that in mind.

It's not about what they return, but about the input. `date` here is the value of a key from the response object that currently has a very relaxed `Joi.string()` schema. However, the transform function is **clearly** expecting this string to be in a very specific format and it will cause runtime errors if `date` does not conform to that assumed shape. 

What I'm requesting here is that `Joi.string()` be updated (probably by adding a regex) to ensure that the value of `date` will conform to the structure that `transform` requires and is currently just assuming

```js
    transform(date, extensionType) {
      let d
      if (extensionType === 'plugin') {
        d = date.split(' ')
        d = d[0]
      } else {
        d = date
      }

      const ymd = d.split('-')

      const out = ymd[0] + ymd[1] + ymd[2]
      return out
    }
```

JoeIzzard: I see! 

So would it be better to do that at the schema level, or in the class that handles that badge? In theory, if it ever came back not matching, would it being in the schema also break other badges that don't rely on that piece of info?

calebcartwright: > So would it be better to do that at the schema level, or in the class that handles that badge?

We want to do input validation as part of the schema

> In theory, if it ever came back not matching, would it being in the schema also break other badges that don't rely on that piece of info?

Yes, if someone requested a badge for a given theme/plugin, and the API response failed to match the input validation then Shields would return an `invalid response data` badge (by design). Since all the Wordpress badges are using the same shared schema, yes, a request for any of the Wordpress badges would fail in this circumstance even if they don't care about that particular field. That's why it's important when using a single, shared schema for multiple badges instead of service/badge specific schemas that all of the fields included in the schema will always be validated regardless of the badge for that family.

JoeIzzard: Awesome. So I have added two patterns, one for themes, and one for plugins and added those to the schema. Also updated a mocked test which was using the wrong pattern and causing it to fail this new schema check.

chris48s: > cc @chris48s for any additional thoughts on schema structure

Sorry. Probably should have jumped in on this one a couple of days ago.

This regex/splitting strings up and joining them back together approach doesn't seem like the best way to deal with this either on the validation or formatting side tbh. We should really lean on a library here IMO. Lets put aside https://github.com/badges/shields/issues/5573 for now. Adding a couple more lines of moment.js code isn't going to cause us a huge issue in the grand scheme of things. How about we do something like:


```js
const moment = require('moment')
```

...

```js

last_updated: Joi.string().required()
```

...

```js
const date = moment(last_updated, "YYYY-MM-DD hh:mma GMT") // think this is the right format string for the plugin case

if (date.isValid()) {
  throw new InvalidResponse({prettyMessage: 'invalid date'});
}

// do something with date.format('YYYY-MM-DD')
```

While its nice to do the validation in the schema if possible rather than throwing an `InvalidResponse()` manually, `Joi.date()` isn't flexible enough for us here due to the format used.

I've got a much higher level of confidence that `const date = moment(last_updated, "YYYY-MM-DD hh:mma GMT")` does what we want than I have looking at `[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{1,2}:[0-5][0-9](am|pm) [A-Z]{3}`. For a start I can actually read it :) . By the looks of it you need two different format strings for themes than plugins, but the moment.js docs are great https://momentjs.com/docs/#/parsing/string-format/

Maybe dates/times are another topic for https://github.com/badges/shields/discussions/5619

calebcartwright: While I do have some minor reservations about increasing our moment.js footprint, I do agree that it's minor enough to not be an issue here and like this proposal even better!

Would you agree though @chris48s that the original loose schema and transform logic would've been problematic for what we're driving towards on schema guidelines? For example if this had been some other type of field where we couldn't rely on a 3rd party lib like moment

chris48s: I guess if this was a string that describes a completely arbitrary thing (not a datetime) but we wanted to split it up on space and dash, something like that would have been fine. In that case, we'd want to ensure that:
- The characters we're going to split on exist
- If we're going to reference  `myvar[2]` after splitting it, that index is going to exist
- Whatever we end up passing to `formatDate()` (except it would be `formatSomethingElse()` ) isn't going to give the formatting function a problem

I guess this is a situation where it is beneficial to think about readable code and documenting our understanding of the upstream API as well as just the least strict validation possible.

chris48s (approved):
nice :+1: lets get this one merged too

PR #5723, opened 2020-10-16
[Wordpress] Plugin & Theme required php

JoeIzzard:
Related to #5493, this adds the required php badge for both plugin and themes. 

One difference between this and the main PR is that I have now changed one of the tests to a mock, as the live data I was using has changed and no longer works for the test, and I am struggling to find any live data to replace it with.

To help with this, I have refactored some of the test code to help facilitate this.


chris48s (commented):
Nice. Thanks so much for the extra work on splitting these up. Might take a bit to get through them all, but now each review is much less daunting :)


If `requires_php` is `'7.2'` or `'5.6'` or whatever, is that a minimum version i.e: is what this really means the plugin/theme is compatible with PHP `>=7.2` or `>=5.6`?
If so, how about we style it like  - it makes the meaning clearer and will be more consistent with other PHP version badges which express the PHP version requirement as a constraint.

JoeIzzard:
No problem, at least now they are easy to do a quick review of for you guys, so should be quicker than trying to do it all as one! 

Yeah that's right. That's definitely a clearer way to style it, I will get that changed.

chris48s (approved):

PR #5726, opened 2020-10-16
Align `style` prop of `coalesceBadge()` with internal `makeBadge()`

paulmelnikow:
This is an analogous change to #5719 for `’template’` and `’style’`. (See the top comment for more explanation.)

The diff is simpler since there are fewer test changes in this one.


calebcartwright (approved):
I'm good with the change, though are we going to need a new major version bump of the package to accompany? I wasn't quite sure how to take the inline comments quoted `"public"..`

paulmelnikow:
No, unless I've made a mistake here, this is only affecting private/internal APIs.

Issue #4749, opened 2020-03-06
logoColor ignored when passing a base 64 encoded logo


amomchilov:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: Documentation should be clear that the `logoColor` doesn't apply for base64 encoded images.

When using the `?logo=data:image/svg+xml;base64,...` syntax to specify a base64 encoded logo, the logoColor param is ignored. The SVG's `fill` attribute is only changed for simple-icon SVGs, not custom ones passed as base64.

I'm not sure whether that's a desired behaviour, but if it's not, it should be documented more clearly.

Further discussion: https://github.com/simple-icons/simple-icons/issues/2539#issuecomment-595463725

:bulb: **Possible Solution**

1. Update the documentation to make the current behaviour clear
2. Change the behaviour so that `fill` attributes are set affected by `logoColor`, even when logos are base64 encoded.

paulmelnikow:
Hi, yes, this is by design. If sending a custom logo and you want it a different color, it should be colorized in advance.

A PR would be welcome to update the docs on the global query parameters.

I’m not 100% sure, though I think this limitation is already mentioned in the Endpoint docs, which has the same limitation.

PR #5727, opened 2020-10-17
Specifying logoColor does not work with logoSvg

mudlabs:
If your endpoint specifies the `logoSvg` property you can _not_ specify the `logoColor` property. This will result in the 'invalid properties: logoColor' badge.

Related to #4749


calebcartwright (approved):
This isn't exclusive to the endpoint, but more a reflection of the fact that we don't want to be in the business of colorizing custom logos regardless of the badge type. Think it's fine to repeat that fact here though, so thank you!

PR #5748, opened 2020-10-19
Refactor remaining [github] services

Transfusion:
> That means that most/all of the service classes inclusively from GitHubSearch (https://github.com/badges/shields/blob/master/services/github/github-search.service.js) through GitHubWorkFlowStatus probably still need to be refactored if you (or anyone else) wants to work on those

The following services have been refactored:
- GithubSearch
- GithubSize
- GithubStars 
- GithubTag 
- GithubTopLanguage
- GithubWatchers
- GithubWorkflowStatus

#5555


calebcartwright (approved):
Thank you!

PR #5753, opened 2020-10-20
Add support for NuGet badges from [nuget feedz] hosting

jakubfijalkowski:
In this PR I've added the support for the `feedz` badge with package version.

Feedz is a hosted NuGet repository that serves mostly the same purpose as the official NuGet Gallery or (even more) MyGet. It supports a slightly restricted set of features compared to NuGet/MyGet and I wasn't able to use the existing implementation. The unfortunate side effect of this is that it does not support download counts, only a package version is available.

If we want to support downloads, we would need to have a fully featured search endpoint, as this is the only place with `totalDownloads`. Unfortunately, this is where Feedz is limited as it does not support filtering packages. We would need to download all of the available packages in the memory, and do the filtering there. This might result in a substantial memory overhead and I don't think it is worth.

This service differs slightly from `myget` in that the service route is `/feedz/(v|vpre)/(organization)/(repository)/(packageName)` whereas `myget` uses `/(tenant).myget/(feed)/(v|vpre)/(packageName)` route. Apart from the different naming (`tenant` vs `organization`, `feed` vs `repository`), I moved the `v|vpre` segment right after the service name. It should better reflect the current URL conventions, provided that I understand them correctly.


calebcartwright:
Hi @jakubfijalkowski :wave:, thank you for the PR!

> It supports a slightly restricted set of features compared to NuGet/MyGet and I wasn't able to use the existing implementation.

Just to clarify, are you saying that it's not possible to use one of our existing `nuget-v*-service-family` offerings to produce this badge? If so could you expand on why? I'd always prefer to avoid duplication/maximize reuse, so may also be worth seeing if the family constructor function could be expanded a bit to better allow them to be reused even in cases like this where certain data points/badges wouldn't be possible

> We would need to download all of the available packages in the memory, and do the filtering there. This might result in a substantial memory overhead and I don't think it is worth.

Yes, that would not be an acceptable tradeoff :smile:

jakubfijalkowski:
### TL;DR

Feedz does not support filtering in the `SearchQueryService` which requires us to use `RegistrationsBaseUrl`. This works almost the same, except where not (schemas are different and we get slightly different data), and these details makes sharing of the implementations rather difficult.

### Full story

The main reason is that Feedz supports only the most basic NuGet API - they implement the `get` (`RegistrationsBaseUrl`, i.e. get package details) and `list` endpoint (i.e. `SearchQueryService` in v3 and `Packages()` in v2) but don't allow any filtering. The current implementation relies on the list endpoint allowing filtering the same way nuget.org does work (which might not be true as that is not really defined in [the spec](https://docs.microsoft.com/en-us/nuget/api/search-query-service-resource)) - that is, the `?q=packageid:(package name)` (v3)/`?$filter=Id eq (package name)` (v2). Doing so, it is able to download versions and total downloads for package.

Feedz only allows listing all packages, without any filtering (e.g. [v3](https://f.feedz.io/shieldstests/public/nuget/v3/query?q=packageid:Shields.TestPackage) and [v2](https://f.feedz.io/leancode/public/nuget/Packages()?$filter=Id%20eq%20Shields.TestPackage), the older version even explicitly states that filtering is not possible). To overcome that limitation, we might use the registration endpoint to get info about particular package (e.g. [v3](https://f.feedz.io/shieldstests/public/nuget/v3/registration/Shields.TestPackage/index.json)). This gives us a possibility to get available versions, but requires us to do _mostly the same, but slightly different calls_. Not only the endpoint has different name, but also the schema is slightly different and the data is formatted differently (e.g. it returns `404` when package is not found, which throws an error deeper in the shields API - the `SearchQueryService` endpoint just returns empty list which is handles separately).

Feedz also has a different meaning of `tenant` than MyGet (and possibly other hosting services). MyGet treats it like a domain and a separate, paid feature, whereas Feedz requires that for every feed and treats it as part of feed name (so in the path part of URL, not domain).

The only thing that IMO is common between the two is the call flow. All the v3-based implementations basically do this:
1. Build the main URL (the `nuget-v3-service-family` `if`s the small differences between NuGet and MyGet - Feedz differs much more substantially here),
2. Call it to get the real URL for `SearchQueryService` endpoint (`RegistrationsBaseUrl` in Feedz case),
3. Call the `SearchQueryService` to get all versions for specified filter (`RegistrationsBaseUrl` in Feedz case, for info about particular package),
4. Parse the JSON and filter valid versions (this is mostly the same for all the three cases, only that Feedz has different schema).

I think abstracting the call flow is the only way we can share code with this implementation. I might modify the `nuget-v3-service-family` factory function so that, instead of taking all the flags, takes functions that represent the above steps. Might work, but I'm not sure it will benefit much.

I've asked Feedz if they plan to implement this feature in the near future so that we can re-use the existing implementation. I will get back when they send me an answer.

That's it I think. I need to work on my concise writing because this is far too long. Sorry :sweat_smile:

calebcartwright:
> This service differs slightly from `myget` in that the service route is `/feedz/(v|vpre)/(organization)/(repository)/(packageName)` whereas `myget` uses `/(tenant).myget/(feed)/(v|vpre)/(packageName)` route. Apart from the different naming (`tenant` vs `organization`, `feed` vs `repository`), I moved the `v|vpre` segment right after the service name. It should better reflect the current URL conventions, provided that I understand them correctly.

thank you for doing this! moving the `v|vpre` leftwards does indeed position them as the noun portion of the route and better aligns with our [badge url guidelines](https://github.com/badges/shields/blob/master/doc/badge-urls.md)

> Feedz does not support filtering in the SearchQueryService which requires us to use RegistrationsBaseUrl. This works almost the same, except where not (schemas are different and we get slightly different data), and these details makes sharing of the implementations rather difficult.

Fair enough, we may not be able to use it out of the box, but I also want to make sure we don't unnecessarily duplicate the reusable bits either. For example, the `searchQueryServiceUrl`, `stripBuildMetadata`, and `selectVersion` functions added here are highly duplicative of existing functions/blocks in the v3 family class. I'd prefer to extract and encapsulate these in a single place (for example `nuget-helpers`), and then leverage those in the consumers.

Please also double check anything copied from myget or other existing services to ensure they've all been updated. One specific case: the route patterns in the examples here are referring to `:tenant` which I believe is a likely left over from myget. Incorrect patterns i the examples can be confusing for users browsing the listings, and also causes the badge builder window to provide users with invalid/incorrect badge urls.

For example, those patterns are currently causing the builder to produce this url (won't render due to the 404)
https://shields-staging-pr-5753.herokuapp.com/feedz/leancode/mongodb/v/MongoDB.Driver.Core


instead of what I believe the corresponding badge url should be
https://shields-staging-pr-5753.herokuapp.com/feedz/v/leancode/mongodb/MongoDB.Driver.Core

jakubfijalkowski:
> For example, the searchQueryServiceUrl, stripBuildMetadata, and selectVersion functions added here are highly duplicative of existing functions/blocks in the v3 family class. I'd prefer to extract and encapsulate these in a single place (for example nuget-helpers), and then leverage those in the consumers.

You're totally right. Moved these methods to `nuget-helpers`

> Please also double check anything copied from myget or other existing services to ensure they've all been updated.

Ah, yes. I've changed the examples before reworking the service and forgot about them. Fixed that. Also, I've uploaded the `MongoDB.Driver.Core` package to `shieldstests/mongodb` repository so that the examples work as expected (and are mostly the same as in myget):

https://shields-staging-pr-5753.herokuapp.com/feedz/v/shieldstests/mongodb/MongoDB.Driver.Core: https://shields-staging-pr-5753.herokuapp.com/feedz/v/shieldstests/mongodb/MongoDB.Driver.Core
https://shields-staging-pr-5753.herokuapp.com/feedz/vpre/shieldstests/mongodb/MongoDB.Driver.Core: https://shields-staging-pr-5753.herokuapp.com/feedz/vpre/shieldstests/mongodb/MongoDB.Driver.Core

services/feedz/feedz.service.js
```diff
@@ -0,0 +1,109 @@
...
+
+class FeedzVersionService extends BaseJsonService {
+  static category = 'version'
+
+  static route = new RouteBuilder({ base: 'feedz' })
+    .push('(v|vpre)', 'which')
+    .push('([^/]+)', 'organization')
+    .push('([^/]+)', 'repository')
+    .push('(.+?)', 'packageName')
+    .toObject()
```
calebcartwright, line 42: Is there a need to use RouteBuilder here since the route itself is static, unlike in the nuget v3 family which is dynamic?

services/feedz/feedz.service.js
```diff
@@ -0,0 +1,109 @@
...
+      url: `${registrationsBaseUrl}${packageName}/index.json`,
+    })
+
+    if (json.items.length === 1) {
+      return json.items[0].items.map(i =>
+        stripBuildMetadata(i.catalogEntry.version)
+      )
+    } else {
+      throw new NotFound({ prettyMessage: 'not found' })
+    }
```
calebcartwright, line 91: I know this is similar to how the nuget-v3 family was structured, but this is a case where I think we should take advantage of some of the comparatively newer patterns we've been leveraging in our service classes.

For better/cleaner separation of concerns and encapsulation we prefer to use a separate `transform` function that takes the raw response from the upstream provider and does whatever processing/extracting/transforming is needed to get the desired data point (which would be the target `version` in this case).

An added benefit of this approach is that it makes testing logic much easier with standard unit tests, as a `transform` function can be hit directly without having to deal with cumbersome mocking

services/feedz/feedz.service.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  async handle({ which, organization, repository, packageName }) {
+    const includePrereleases = which === 'vpre'
+    const baseUrl = apiUrl({ organization, repository })
+
+    const allVersions = await this.fetch({ baseUrl, packageName })
+    const version = selectVersion(allVersions, includePrereleases)
```
calebcartwright, line 99: Along the same lines as the previous comment, the introduction of a `transform` function would include some minor refactoring here as well. One way to think of the `handle` function is the entry point and orchestrator that typically calls `fetch` to get the raw response, then `transform` to process it, then returning the `render` function

calebcartwright (commented):
Thank you for the updates and refactoring @jakubfijalkowski. Left a few minor items inline for your review, but otherwise this looks just about good to merge!

services/feedz/feedz.service.js
jakubfijalkowski: Yep, I think that I looked too little into other services and missed that. Fixed.

services/feedz/feedz.service.js
jakubfijalkowski: I totally agree. Haven't seen that in most of the services that I looked but the testability is really beneficial. Fixed (and added the tests).

jakubfijalkowski:
I've changed the feedz service according to your fair comments @calebcartwright . The question is - should I also extract the `transform` method in NuGet service or should I leave it as-is for now? I can do so in another PR if you like.

calebcartwright:
> The question is - should I also extract the `transform` method in NuGet service or should I leave it as-is for now? I can do so in another PR if you like

That'd be great! Let's defer any additional changes to another service to a future PR though

> I've changed the feedz service according to your fair comments @calebcartwright .

Thanks for the updates, everything looks good here. One final last ask from me:

Our badge service infrastructure defaults to a `not found` error message in the event of a 404 response code from the upstream provider. Could we differentiate the error messages a bit so there's a distinction between that 404 (which I presume could happen on a bad repo/org param as well as perhaps the package) vs. the package being found successfully but there being no versions/`items` being an empty array?

The default error messages for response codes can be overridden via the `errors` key in the args to `_requestJson`, so could change that to something like:

```js
   return await this._requestJson({
      schema,
      url: `${registrationsBaseUrl}${packageName}/index.json`,
      errors: {
        404: 'package or repo not found',
      }
    })
```

And then add a unit test for the `transform` function that covers _that_ not found scenario (an empty `items` array). We do this type of error throwing testing in a few different ways, with the ScrutinizerCoverage unit tests showing a few. Couple examples below for reference, no real preference on how these scenario as long as we get one, so dealers choice :smile: 
```js
    it('throws NotFound....', function () {
      expect(() =>
        FeedzVersionService.prototype.transform(...)
      )
        .to.throw(NotFound)
        .with.property('prettyMessage', '....')
    })
```

or
```js
    it('throws NotFound....', function () {
      try {
        FeedzVersionService.prototype.transform(...)
        expect.fail('Expected to throw')
      } catch (e) {
        expect(e).to.be.an.instanceof(NotFound)
        expect(e.prettyMessage).to.equal('...')
      }
    })
```

jakubfijalkowski:
I've changed the error messages to be different for these two (three) cases.

I've dug a little bit deeper and found out that `checkErrorResponse` _allows_ to examine error response (I've missed that last time) so I am able to differentiate the cases:
1) package not found
2) org/repo not found (`org/repo` is really a single thing, not two).
It involves much more `if`ology but isn't that complicated in the end.

services/feedz/feedz.service.spec.js
```diff
@@ -0,0 +1,215 @@
...
+        await serviceInstance.tryHandleError({
+          err: error({ statusCode: 404, content: '' }),
+        })
+        expect.fail('Expected to throw')
+      } catch (e) {
+        expect(e).to.be.an.instanceof(NotFound)
+        expect(e.prettyMessage).to.equal('package not found')
+      }
+    })
+  })
```
calebcartwright, line 214: Let's remove these tests.

The service class infra/parent classes are already tested extensively elsewhere. We don't want to re-test parent class members from unit tests in all the extending child classes as that just adds maintenance overhead on our end and makes the test suite ever so slower (especially we started doing this in _every_ child class).

calebcartwright: Sorry, ignore this for now, I completely missed the modifications to the service class that these are testing

services/feedz/feedz.service.js
```diff
@@ -0,0 +1,148 @@
...
+        if (feedNotFoundMessage.test(json.message)) {
+          throw new NotFound({ prettyMessage: 'repository not found' })
+        } else {
+          throw err
+        }
+      }
+    } else {
+      throw err
+    }
+  }
```
calebcartwright, line 105: Thanks for including inline comments on this one. Do you have a feel for how stable or brittle this logic/the error behavior may be?

I think it's nice to be able to provide very contextual error messages to users here, which in most cases will be helping them pinpoint which param value contains a typo. In most other cases though we've eschewed complex error handling logic like this and simply provide a `x or y not found` message.

The tradeoffs for us to consider here is whether the extra logic and behavior is worth maintaining to be able to provide separate `package not found` vs `repository not found` error messages vs. a simple `repository or package not found` message in 404 cases

cc @badges/shields-maintainers for other input

jakubfijalkowski: Yeah, I'm also not really sure whether that is worth it (that's why I first added the tests and then modified everything - we can easily rollback). The gain is really small compared to the complication.

Regarding the stability - I am quite sure that this won't change in the coming future. Feedz hasn't really changed for like couple of years now and I don't think it will change. However, this is a Feedz extension and not something in NuGet Gallery API spec so this might be really just a maintenance burden with no real benefit.

Daniel15: Yeah I don't think we should be looking for particular strings in responses, as that's quite fragile. We should just look for error codes as they're less likely to change, but not all APIs have unique codes per error. Failing that, we should probably just display a generic error message.

jakubfijalkowski:
Considering the complexity, I rolled-back the complicated error handling.

services/feedz/feedz.service.js
PyvesB: Also in agreement with the above. It's good to have specific error messages, but not at the cost of significant added code complexity or parsing that's likely to break.

calebcartwright (approved):
LGTM

Thanks so much for this @jakubfijalkowski, including all the rounds of feedback :smile: If you're still interested in some of the other nuget-related changes we discussed that'd be great too!

jakubfijalkowski:
:tada: 

Awesome! :D Thank you for your help, mentoring and patience! :)

I will try to revisit the nuget service sometime later this week, although I don't want to promise the exact timeline. :)

PR #5754, opened 2020-10-21
Make for-the-badge letter spacing more predictable, and rewrite layout logic

paulmelnikow:
This is a rewrite of the for-the-badge layout logic. The previous logic was very difficult to understand or modify, and also tended to produce inconsistent letter spacing which was hard to match during rasterization.

The goal was to produce consistent letter spacing, such that replacing `textLength` with `letter-spacing` produces a visually identical badge. It's not 100% perfect but it's very close. I suspect it is much more consistent than before. The letter spacing is 1.25px (or in 10x terms, 12.5px).

In the case of a message-only badge, I've removed the extra rect which we used to render behind the logo.

In this rewrite I haven't gone for consistency with the other badge-generation code. Rather, I've tried to produce something which is correct and as clean as possible, and can be used as a model to refactor the others.

Compare here: https://shields-staging-pr-5754.herokuapp.com/dev/styles/

This is a follow-on to #5696


calebcartwright:
And the raster proxy complexity faded away... 

Haven't had a chance to test this yet but excited to do so :smile:

badge-maker/lib/badge-renderers.js
```diff
@@ -576,78 +576,103 @@ function forTheBadge({
   links,
   logo,
   logoWidth,
-  logoPadding,
```
paulmelnikow, line 579: `logoPadding` is not a user-specified parameter, but a value computed here:

https://github.com/badges/shields/blob/f1a5dd9427a5eb9581fb7f0b82f5babe449ca1d6/badge-maker/lib/make-badge.js#L61

I think we probably ought to push this logic into the badge generation code as I've done here in for-the-badge.

badge-maker/lib/badge-renderers.js
```diff
@@ -576,78 +576,110 @@ function forTheBadge({
...
-
-    const text = `<text fill="${textColor}" x="${labelTextX}" y="175" transform="scale(.1)" textLength="${labelTextLength}">${escapedLabel}</text>`
+    const { textColor } = colorsForBackground(outLabelColor)
+    const midX = labelTextMinX + 0.5 * labelTextWidth
+    const text = `
+      <text
+        transform="scale(.1)"
+        x="${10 * midX}"
+        y="175"
+        textLength="${10 * labelTextWidth}"
```
calebcartwright, line 669: would it be worth while adding constants for the scale or is it more readable inline/easier to hypothetically use different scale factor for the label and message/

calebcartwright (commented):
Seems straightforward to me (and a bit easier to grok). Left a few inline questions/comments below

badge-maker/lib/badge-renderers.js
```diff
@@ -576,78 +576,110 @@ function forTheBadge({
...
+      labelRectWidth = 2 * LOGO_MARGIN + logoWidth
     }
+    messageTextMinX = labelRectWidth + TEXT_MARGIN
+    messageRectWidth = 2 * TEXT_MARGIN + messageTextWidth
   } else {
-    labelWidth -= 11
+    if (logo) {
+      messageTextMinX = TEXT_MARGIN + logoWidth + LOGO_TEXT_GUTTER
+      messageRectWidth =
+        2 * TEXT_MARGIN + logoWidth + LOGO_TEXT_GUTTER + messageTextWidth
```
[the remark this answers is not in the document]
paulmelnikow: This line is nested inside the case when there is no label rect (i.e. `!needsLabelRect`), so the logo goes in the message rect, too:



I tried to explain this in the comment above, though maybe you could suggest a way to make it clearer.

badge-maker/lib/badge-renderers.js
paulmelnikow: Sure, I think that's a nice idea.

badge-maker/lib/badge-renderers.js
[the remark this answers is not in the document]
calebcartwright: Ah okay, I think I just misread/misinterpreted that as meaning logo would force a label (somehow :thinking:) on a second reading I'm not really sure how I got there :laughing:

badge-maker/lib/badge-renderers.js
```diff
@@ -576,78 +576,110 @@ function forTheBadge({
...
   labelColor,
 }) {
-  // For the Badge is styled in all caps. Convert to caps here so widths can
-  // be measured using the correct characters.
+  const BADGE_HEIGHT = 28
+  const LOGO_HEIGHT = 14
+  const TEXT_MARGIN = 12
+  const LOGO_MARGIN = 9
+  const LOGO_TEXT_GUTTER = 6
+  const LETTER_SPACING = 1.25
```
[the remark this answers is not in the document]
calebcartwright: oh and just to confirm the same letter spacing value for the bold and non-bold text? I know previously we were discussing a slight difference between the two (though that may matter more on the raster side than here)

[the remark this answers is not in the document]
paulmelnikow: Yep, they're the same now.

From the source code, it doesn't appear they were ever intended to be different. I think it was just due to quirks / bugs in the computation that the effective letter spacing varied given between various different values for label and message.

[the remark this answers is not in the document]
calebcartwright: gotcha. the 1.25 value looks good/correct in the png's too so with this change I think all that will be left on the raster side to finish the FTB style will be to insert the 1.25 letter spacing and then adjust the width values accordingly

calebcartwright (approved):
:+1:

badge-maker/lib/badge-renderers.js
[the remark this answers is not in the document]
paulmelnikow: Amazing!

> gotcha. the 1.25 value looks good/correct in the png's too so with this change I think all that will be left on the raster side to finish the FTB style will be to insert the 1.25 letter spacing and then adjust the width values accordingly

Does it work to just remove the `textLength` attribute after adding `letter-spacing`?

paulmelnikow:
I'll leave this open for a couple days in case @chris48s wants to review.

badge-maker/lib/badge-renderers.js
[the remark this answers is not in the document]
calebcartwright: > Does it work to just remove the textLength attribute after adding letter-spacing?

No, although IIRC the Verdana font is still missing from the Docker image which probably has _some_ impact

For reference here's the badge route `.../badge/foojjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjjj-barfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff-green.png?logo=appveyor&color=orange&style=for-the-badge`


and the corresponding xml after inserting letter-spacing and dropping textLength
```xml
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="946.5" height="28" role="img" aria-label="FOOJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJ: BARFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF">
  <title>FOOJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJ: BARFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF</title>
    <rect width="268.5" height="28" fill="#555" />
    <rect x="268.5" width="678" height="28" fill="#fe7d37" />
  </g>
  <g fill="#fff" text-anchor="middle" font-family="Verdana,Geneva,DejaVu Sans,sans-serif" text-rendering="geometricPrecision" font-size="100">
    <image x="9" y="7" width="14" height="14" xlink:href="data:image/svg+xml;base64,PHN2ZyBmaWxsPSIjMDBCM0UwIiByb2xlPSJpbWciIHZpZXdCb3g9IjAgMCAyNCAyNCIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48dGl0bGU+QXBwVmV5b3IgaWNvbjwvdGl0bGU+PHBhdGggZD0iTSAxMiwwIEMgMTguNiwwIDI0LDUuNCAyNCwxMiAyNCwxOC42IDE4LjYsMjQgMTIsMjQgNS40LDI0IDAsMTguNiAwLDEyIDAsNS40IDUuNCwwIDEyLDAgWiBtIDIuOTQsMTQuMzQgQyAxNi4yNiwxMi42NiAxNi4wOCwxMC4yNiAxNC40LDkgMTIuNzgsNy43NCAxMC4zOCw4LjA0IDksOS43MiA3LjY4LDExLjQgNy44NiwxMy44IDkuNTQsMTUuMDYgYyAxLjY4LDEuMjYgNC4wOCwwLjk2IDUuNCwtMC43MiB6IG0gLTYuNDIsNy44IGMgMC43MiwwLjMgMi4yOCwwLjYgMy4wNiwwLjYgbCA1LjIyLC03LjU2IGMgMS42OCwtMi41MiAxLjI2LC01Ljk0IC0xLjA4LC03LjggLTIuMSwtMS42OCAtNS4wNCwtMS42MiAtNy4xNCwwIGwgLTcuMjYsNS41OCBjIDAuMTgsMS45MiAwLjcyLDIuODggMC43MiwyLjk0IGwgNC4xNCwtNC41IGMgLTAuMywxLjk4IDAuNDIsNC4wMiAyLjEsNS4yOCAxLjQ0LDEuMTQgMy4xOCwxLjQ0IDQuODYsMS4wOCB6Ii8+PC9zdmc+" />
    <text transform="scale(.1)" x="1427.5" y="175" fill="#fff" letter-spacing="12.5">FOOJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJJ</text>
    <text transform="scale(.1)" x="6075" y="175" fill="#fff" font-weight="bold" letter-spacing="12.5">BARFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF</text>
  </g>
</svg>
```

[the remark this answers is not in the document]
paulmelnikow: It looks pretty fine in my browser – yours too?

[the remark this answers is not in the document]
calebcartwright: not so much, Firefox on Ubuntu

image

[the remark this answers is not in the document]
paulmelnikow: What font is that? It doesn't look like Verdana.

This is OS X Chrome:

Screen Shot 2020-10-21 at 10 21 37 PM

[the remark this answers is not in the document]
calebcartwright: Deja Vu and friends :laughing:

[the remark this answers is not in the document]
paulmelnikow: For the benefit of folks reading this, wanted to mention we discussed this a bit offline, and the solution is to have Verdana installed (including on the raster server). The non-working SVG above isn't from this PR.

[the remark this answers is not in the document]
calebcartwright: Posting here for visibility as this is exciting news, the incorporation of the right font on the raster side + these changes and swapping out `textLength` for `letter-spacing` on the squint side before rasterization has solved all the weird spacing issues we had on the raster side!

badge-maker/lib/make-badge.js
```diff
@@ -23,6 +23,10 @@ module.exports = function makeBadge({
   logoWidth,
   links = ['', ''],
 }) {
+  if (!logo && (logoPosition !== undefined || logoWidth !== undefined)) {
+    throw Error('`logoPosition` and `logoWidth` require `logo`')
+  }
+
```
chris48s, line 29: I feel like we've been here before. I can't remember what param it was (probably a different one), but we added some "if you're specifying this param, it has to actually make sense" logic before, loads of stuff broke and we ended up rolling it back the next day. Does that ring a bell.

chris48s (commented):
> I'll leave this open for a couple days in case @chris48s wants to review.

"wants" is a strong word :wink:
I've not really looked at the code in much detail, but there is one thing that immediately jumped out at me. The images on staging look right. I see if I can dig into it more at the weekend

badge-maker/lib/make-badge.js
paulmelnikow: > "wants" is a strong word 😉

😝 

Hmm, no… it doesn't quite ring a bell. Unless it was something in the endpoint badge? I wonder if we added a test. I can do a little digging in the issues.

The reason I added this is so I could remove what appeared to be logic guarding against this inside the render functions.

I think, probably, that if there's a good reason for the server to accept params that don't make sense, that it should strip them out before invoking `makeBadge()`.

badge-maker/lib/badge-renderers.js
```diff
@@ -576,78 +580,111 @@ function forTheBadge({
...
-  labelColor = escapeXml(labelColor)
-
-  let [leftLink, rightLink] = links
-  leftLink = escapeXml(leftLink)
-  rightLink = escapeXml(rightLink)
-  const { hasLeftLink, hasRightLink } = hasLinks({ links })
-
-  const accessibleText = createAccessibleText({ label, message })
+  // Render.
+  function renderLogo() {
```
chris48s, line 654: Do we actually need to define a second `renderLogo()` function here? It looks like we could just call the renderLogo() function we already have with the params

```js
renderLogo({
  logo,
  badgeHeight: BADGE_HEIGHT,
  horizPadding: logoMinX,
  logoWidth,
  logoPadding,
})
```

and it would return the same string?

badge-maker/lib/badge-renderers.js
```diff
@@ -678,24 +726,52 @@ function forTheBadge({
...
+          height="${BADGE_HEIGHT}"
+          fill="${escapeXml(color)}"
+        />
+      `
+    : `
+        <rect
+          width="${messageRectWidth}"
+          height="${BADGE_HEIGHT}"
+          fill="${escapeXml(color)}"
+        />
```
chris48s, line 748: I found this ternary/template spread over ~20 lines a bit hard to parse. How about we wrap the templates so we do something like

```js
const labelRect = `<rect...`
const messageRect = `<rect...`
const renderedBackground = needsLabelRect ? labelRect + messageRect : messageRect
```

or

```js
const renderLabelRect = function(...
const renderMessageRect = function(...
const renderedBackground = needsLabelRect ? renderLabelRect() + renderMessageRect() : renderMessageRect()
```

to keep the actual ternary short?

badge-maker/lib/badge-renderers.js
```diff
@@ -576,78 +580,111 @@ function forTheBadge({
   links,
   logo,
   logoWidth,
-  logoPadding,
   color = '#4c1',
   labelColor,
 }) {
-  // For the Badge is styled in all caps. Convert to caps here so widths can
-  // be measured using the correct characters.
```
chris48s, line 584: I think this comment is still relevant (and helpful), but a few lines further down where we call `.toUpperCase()`

chris48s (commented):
If I compare the source of
https://shields-staging-pr-5754.herokuapp.com/badge/build-passing-brightgreen
to the source of
https://shields-staging-pr-5754.herokuapp.com/badge/build-passing-brightgreen?style=for-the-badge
The first one is compressed, but the second is rendering with whitespace. I can't quite put my finger on why as by the looks of it, they should all be getting processed by `stripXmlWhitespace()` in `makeBadge()`. Any idea why that is happening?

---

I've had this thought before and its not necessarily something we need to do right now in this PR as it applies equally to the other code in `badge-renderers.js` too, but one of the things I wanted to check in this PR was that everything that needs to be escaped is definitely escaped. I'm conscious in attempting to read this is that because we call `escapeXml()` in various different places its hard to trace each variable through the function. I get that's necessary because sometimes we can escape straight away, whereas sometimes we want to (for example) do the text measurements first and then escape, etc..

Sometimes we are doing something like

```js
const [leftLink, rightLink] = links.map(escapeXml)
```

early on, whereas sometimes we're doing it directly in the template string right at the end.

I wonder if it would make sense to adopt a form of hungarian notation just within `badge-renderers.js` where we have to remember to manually call `escapeXml()` before its safe to use certain variables. I'm thinking of the style described in the section "The Real Solution" of this blog post https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/ (from 15 years ago!)

If we say user-supplied vars are always `usVar`, and we always assign them to `sVar` when we escape them or something (and then document that convention), this would hopefully be much more obvious?

badge-maker/lib/make-badge.js
chris48s: I can't remember for the life of me what it is now, but we did something vaguely sensible-souning like like

- `?style=` has to be an actual style
- if you're going to set `?logoColor` you have to set `?logo=`
- `?color=` has to be an actual colour

(it might have been none of those). It seemed completely valid at the time, then we deployed it and realised loads of people had badges out there with spurious values in the URLs and had to roll it back.

Spent about an hour trawling through old issues and PRs to try and track it down - can't find it :(

badge-maker/lib/badge-renderers.js
paulmelnikow: The message rect markup is a bit different for the two cases.

`<rect width="${messageRectWidth}"`

vs

`<rect x="${labelRectWidth}" width="${messageRectWidth}"`

What about replacing the many-line ternary with an `if`/`else`?

```js
let renderedBackground
if (needsLabelRect) {
  renderedBackground = ...
} else {
  renderedBackground = ...
}
```

badge-maker/lib/badge-renderers.js
paulmelnikow: Just took another look at this.

The existing `renderLogo()` function we've got has three return values, and a bunch of parameters which I can't even quite make sense of. The new one I've written here is easier to understand. It's much clearer in for-the-badge, and I suspect will also make the other badge styles clearer.

Swapping in this implementation for the others would require adjusting the rest of the badges, which depend on the spaghetti logic of the old `renderLogo()`. I'd prefer to keep this refactor isolated to for-the-badge, with the idea that it can be used as a model to refactor the others.

How about I rename the existing `renderLogo()` to `renderLogoOld()`, and move the new function alongside it, so that it's ready to use in the other styles?

paulmelnikow:
> If I compare the source of
> https://shields-staging-pr-5754.herokuapp.com/badge/build-passing-brightgreen
> to the source of
> https://shields-staging-pr-5754.herokuapp.com/badge/build-passing-brightgreen?style=for-the-badge
> The first one is compressed, but the second is rendering with whitespace. I can't quite put my finger on why as by the looks of it, they should all be getting processed by `stripXmlWhitespace()` in `makeBadge()`. Any idea why that is happening?

Ah! It looks like `stripXmlWhitespace()` removes whitespace between elements, but not whitespace between _attributes_.

paulmelnikow:
> I've had this thought before and its not necessarily something we need to do right now in this PR as it applies equally to the other code in `badge-renderers.js` too, but one of the things I wanted to check in this PR was that everything that needs to be escaped is definitely escaped. I'm conscious in attempting to read this is that because we call `escapeXml()` in various different places its hard to trace each variable through the function. I get that's necessary because sometimes we can escape straight away, whereas sometimes we want to (for example) do the text measurements first and then escape, etc..
> 
> Sometimes we are doing something like
> 
> ```js
> const [leftLink, rightLink] = links.map(escapeXml)
> ```
> 
> early on, whereas sometimes we're doing it directly in the template string right at the end.
> 
> I wonder if it would make sense to adopt a form of hungarian notation just within `badge-renderers.js` where we have to remember to manually call `escapeXml()` before its safe to use certain variables. I'm thinking of the style described in the section "The Real Solution" of this blog post https://www.joelonsoftware.com/2005/05/11/making-wrong-code-look-wrong/ (from 15 years ago!)
> 
> If we say user-supplied vars are always `usVar`, and we always assign them to `sVar` when we escape them or something (and then document that convention), this would hopefully be much more obvious?

Yea, I thought about this too. The approach I tried to take here was to do all the 10x conversions and the escaping as late as possible, just as the XML was being assembled. The only reason not to put it right into the template literals is when we escape the same text twice. I do think the general gnarliness of this code makes it seem more bug prone than it will as it cleans up, so I'm sort of optimistic that if we do the escaping as late as possible, and either put it right into the XML or into a variable called `escapedLabel`, that it'll be obvious when things are right and when they're wrong. Though I'd also be open to aliasing label, message, etc. to something like `unsafeLabel` on the way in.

badge-maker/lib/make-badge.js
paulmelnikow: I guess the question is what happens when you pass such a URL to the badge server, and does it even get as far as these renderers? How about I move this change to its own draft PR and we can add some tests and investigate. As the code stands now, I don't think it is critical anymore.

chris48s:
> It looks like stripXmlWhitespace() removes whitespace between elements, but not whitespace between attributes

Right, yep should have spotted that. Following on from your suggestion in https://github.com/badges/shields/pull/3637#issuecomment-524078874 (see point 4 in https://github.com/badges/shields/pull/4459 for more info) we dropped the dependency on SVGO. In exchange for removing that library, we have assumed responsibility for manually optimising our outputs. As you note, `stripXmlWhitespace()` is crude, so the way this is dealt with elsewhere in this module is just.. write each XML element on one line, so we can either:

* continue to just write each XML element on one line
* make our own to minification approach more sophisticated
* take on another dependency in the to do it for us package i.e: in https://github.com/badges/shields/blob/d79b0ec46a65b9a269f39f82457f27f3bf8aad3f/badge-maker/package.json#L37-L40
* another (slightly long and rambling) idea which I shall post shortly

badge-maker/lib/badge-renderers.js
chris48s: Yeah sounds good. That would make it clearer :+1:

chris48s:
> Yea, I thought about this too. The approach I tried to take here was to do all the 10x conversions and the escaping as late as possible, just as the XML was being assembled.

OK. As I say, we don't need to do it now. If the way we're going to try and clarify it for now is to escape as late as possible (I think in this case, that probably make sense) are we able to move the escaping of `leftLink` and `rightLink` into the template strings then?

badge-maker/lib/badge-renderers.js
chris48s: Yeah your existing `renderLogo()` does more than one thing. The actual "render a string `<image...`" bit and the params read fairly interchangably in the two functions to me :shrug: . If you want to extract this `renderLogo()` as suggested and call the other one `renderLogoOld()`, you might as well have the "old" one call the new one if possible so you've got something like


```js
function renderLogo(params) {
  return `<image...`
}


function renderLogoOld({
  logo,
  badgeHeight,
  horizPadding,
  logoWidth = 14,
  logoPadding = 0,
}) {
  if (logo) {
    return {
      hasLogo: true,
      totalLogoWidth: logoWidth + logoPadding,
      renderedLogo: renderLogo(params),
    }
  } else {
    return { hasLogo: false, totalLogoWidth: 0, renderedLogo: '' }
  }
}
```

Once you've pulled it out of `forTheBadge()` I think they're going to end up with pretty much the same arg list.

chris48s:
I'm not really sure how to reconcile this review. I kind of drafted the inline comments and then thought about it some more and now I'm actually thinking about this whole module in a different direction so on the whole this review probably comes across like its written by two different people or something. I think I'm just going to paste-dump everything I've got in the interests of moving this on and then we can try to make sense of the soup of comments.



> I'd prefer to keep this refactor isolated to for-the-badge, with the idea that it can be used as a model to refactor the others.

This is going to get a bit long-winded, but if you're thinking of refactoring the whole module, I have a couple of thoughts on this:

1. Broadly speaking, the flat, flat-square and plastic badges are very slight variations on the exact same format/logic. These three lend themselves to sharing as much implementation between them as possible. The social and FTB styles are sufficiently different that they are best thought of as just being two different bespoke things (hence the really different style those two are written). It may be that this style provides a template for social, but I suspect it may not map as well to the flat, flat-square and plastic styles (which are probably starting from a better place anyway tbh).

2. This isn't the _only_ problem, but I think looking back over this module, my biggest regret is probably the poor separation of concerns between calculating values and assembling XML strings. Broadly, this iteration gives us a slightly different configuration of code that has roughly the same underlying issue. Having spent a bit of time chewing this over, I reckon a really big win in this respect would be to factor out a simple `renderXmlElement()` helper function which deals with turning objects into strings. Something like:

```js
function renderXmlElement({
  element, // string
  content, // string
  attrs,   // object
}) {
  let attrs_str = ''
  if (attrs) {
    attrs_str = Object.entries(attrs).map(([k,v]) => `${k}="${v}"`).join(' ')
  }
  if (content) {
    return `<${element} ${attrs_str}>${content}</${element}>`
  }
  return `<${element} ${attrs_str} />`
}
```

That's a rough implementation. Maybe it needs a couple more guard conditions, maybe `content` should actually be an array of strings.. you get the basic idea though. That would allow us to write code like:

```js
const rect = renderXmlElement({
  element: 'rect',
  attrs: {width: 20, height: 15, fill: '#FFF'}
})

const text = renderXmlElement({
  element: 'text',
  content: 'hello'
})

const labelText = renderXmlElement({
  element: 'a',
  content: `${rect}${text}`,
  attrs: {target: '_blank', 'xlink:href': 'https://shields.io/' }
})
```

(intentionally simplified, but again.. you get where I'm going).

This would then abstract out all the string templating so the badge rendering functions are just calculating values and composing function calls, which might simplify things somewhat. This would actually help towards solving two of the things I've mentioned in this review: First we can write the code however we want.

```js
renderXmlElement({
  element: 'rect',
  attrs: {
    width: 20,
    height: 15,
    fill: '#FFF'
  }
})
```

and

```js
renderXmlElement({element: 'rect', attrs: {width: 20, height: 15, fill: '#FFF'}})
```

render the exact same output string because we have seperated presentation and logic, whereas 

```js
`<rect width="20" height="15" fill="#FFF" />`
```

is a different string in the rendered output from

```js
`<rect
  width="20"
  height="15"
  fill="#FFF"
/>`
```

That solves the formatting problem without making `stripXmlWhitespace()` more complex or taking on another dependency, and we can lay our code out however we want.

I think the next place I'd go with that is to see if we can help us simplify the escaping too (but we need to avoid accidentally double-escaping stuff). This may still be a bit half-baked, but I think if we change the line where we assemble the attributes to

```diff
- attrs_str = Object.entries(attrs).map(([k,v]) => `${k}="${v}"`).join(' ')
+ attrs_str = Object.entries(attrs).map(([k,v]) => `${k}="${escapeXml(v)}"`).join(' ')
```

(i.e: we assume the attribute values are always unsafe), we could just completely delegate attribute escaping to `renderXmlElement()` and we can just pass it unsafe vars. This is also sticking with your "escape as late as possible" principle. `content` is more tricky because `content` may be an XML element rather than a value, so maybe we would need to be able to pass a boolean telling us whether the content is safe, or wrap it in a object/type to tell `renderXmlElement()` whether `content` is a string or XML element. e.g:

```js
if (content.type == 'xml') {
  return `<${element} ${attrs_str}>${content.text}</${element}>`
}
return `<${element} ${attrs_str}>${escapeXml(content.text)}</${element}>`
```

I feel like if we can completely delegate calling `escapeXml()` into `renderXmlElement()`, do it in an intuitive way and say "we always make XML strings with `renderXmlElement()`, never manual templating" then this masively helps to make code safe by default and make wrong code look wrong without adopting a wierd naming convention.

What do you think to this as a direction, or element to incorporate in your refactoring?

---

Just to bring this back to the original point.. (which I've strayed from a bit). The core point of this PR is to fix a problem with text width calculation and we don't necessarily need to scope creep and block fixing that bug on a massive refactor of the whole `badge-renderers.js` module (especially as that bugfix is in turn a blocker to switching out the raster proxy), but seeing as that's the direction you seem to be heading in, hopefully some of that is useful. Maybe it would be worth giving a basic `renderXmlElement()` implementation a go just on the FTB code in this PR to solve the whitespace issue though?

chris48s:
@paulmelnikow I am planning to try and revisit this PR over the weekend and see if we can get it finished

calebcartwright:
By the way, https://squint-test.herokuapp.com is still up and configured to fetch the svgs from the review app associated with this PR so if needed you can use that to check the impact of any additional changes here on the raster proxy (note that the current set of changes have resolved all the prior issues on the raster side, recently discovered emoji issues aside)

https://squint-test.herokuapp.com/badge/no-emoji-blue

chris48s:
> By the way, https://squint-test.herokuapp.com is still up and configured to fetch the svgs from the review app associated with this PR

Nice - thanks :+1:

badge-maker/lib/xml.js
```diff
@@ -0,0 +1,82 @@
...
+      .replace(/>/g, '&gt;')
+      .replace(/"/g, '&quot;')
+      .replace(/'/g, '&apos;')
+  }
+}
+
+/**
+ * Representation of an XML element
+ */
+class XmlElement {
```
chris48s, line 29: This `XmlElement` class is the real meat of what I've done here. It is a slight evolution of Chris-from-the-past's suggestion in https://github.com/badges/shields/pull/5754#issuecomment-719817096 . The big advantage of this is it allows us to just construct an `XmlElement` using unescaped/unsafe data, pass it around, and then when we call `.render()` at the end, that will deal with all the escaping for us. It also decouples the string representation of the object from how we lay out the code, solving the whitespace issues.

badge-maker/lib/xml.js
```diff
@@ -0,0 +1,82 @@
...
+        xmlStr += '>'
+      } else {
+        xmlStr += ` ${attrsStr}>`
+      }
+      xmlStr += this.content
+        .map(function (el) {
+          if (el instanceof XmlElement) {
+            return el.render()
+          } else {
+            return escapeXml(el)
```
chris48s, line 71: This recursive approach means `content` can be a mixed array containing strings and other `XmlElement` objects and we won't double-escape when nesting elements.

badge-maker/lib/badge-renderers.js
```diff
@@ -576,126 +569,232 @@ function forTheBadge({
...
-    <g shape-rendering="crispEdges">
-      <rect width="${leftWidth}" height="${height}" fill="${labelColor}"/>
-      <rect x="${leftWidth}" width="${rightWidth}" height="${height}" fill="${color}"/>
-    </g>
-    <g fill="#fff" text-anchor="middle" ${fontFamily} text-rendering="geometricPrecision" font-size="100">
-      ${renderedLogo}
-      ${hasLabel ? renderLabelText() : ''}
-      ${renderMessageText()}
-    </g>`
+    [backgroundGroup.render(), foregroundGroup.render()].join('')
```
chris48s, line 797: I've drawn a line here and I'm just passing strings up to `renderBadge()`, but I think as we work through the other renderers it will ultimately make sense to hand off `XmlElement` objects here and make one final grand call to `.render()` on a `XmlElement({tag: 'svg'...})` object right at the end so the job of calling `escapeXml()` is 100% delegated to `XmlElement.render()`.

badge-maker/lib/badge-renderers.js
```diff
@@ -576,126 +569,232 @@ function forTheBadge({
...
-  const { hasLeftLink, hasRightLink } = hasLinks({ links })
-
-  const accessibleText = createAccessibleText({ label, message })
-
-  function renderLabelText() {
-    const { textColor } = colorsForBackground(labelColor)
-    const labelTextX = ((labelWidth + totalLogoWidth) / 2) * 10
-    const labelTextLength = (labelWidth - (24 + totalLogoWidth)) * 10
-    const escapedLabel = escapeXml(label)
+  const logoElement = new XmlElement({
```
chris48s, line 644: I've ended up leaving the logo rendering inline as we are now making an object not a string, but we can come back to this and extract it later.

chris48s (commented):
I think I built this job up a bit too much in my mind. Once I got into it, there wasn't really as much to do here as I feared. I've stopped pulling the thread with For The Badge in this PR but I think this is going to be a good structure to apply to the other badge renderers moving forward.

badge-maker/lib/xml.js
calebcartwright: I'm still horribly unfamiliar with most of the package code so largely deferring to others. However, did want to ask whether we'd anticipate any additional memory load from the class-based approach, particularly given what we saw recently in production with the got experiment.

chris48s: This will be a slightly heavier implementation than pure string concatenation, but the thing I'm chasing here is developer experience rather than performance. Making it easy for us to "fall into the pit of success" when it comes to escaping has value. Every abstraction has a cost.

I think the difference we will see here will be a lot less than the overhead associated with a full featured HTTP library but it is conjecture and this code is very much on the critical path. I think to quantify it, we need to throw a real workload at it tbh. One thing I did to is A-B it against master using the benchmark script and this seems to add somewhere around 0.02 of a millisecond to a FTB render, which I am not worried about.

One variable here is that although I would like to apply this approach to the other badge styles this PR only changes the rendering of the FTB style. I don't really know what proportion of the traffic we serve is on FTB vs other styles but anecdotally I suspect it is very small so I suspect we wouldn't really start to see a big difference until we apply the approach to the flat style. I'd be surprised if we gather much useful data just from deploying this change.

calebcartwright: Fair enough. Since we've shifted over to Heroku it's _much_ easier to be able to canary deploy these types of changes and quickly revert back out if needed, which makes a huge difference.

This is yet another example of where having the ability to do some load/stress testing against a prod-like env would be useful, so may be worth circling back to this at some point down the road once some other related items (runtime costs/sponsorship, etc.) have concluded.

badge-maker/lib/xml.js
```diff
@@ -0,0 +1,82 @@
...
+    let attrsStr = ''
+    attrsStr = Object.entries(this.attrs)
+      .map(([k, v]) => `${k}="${escapeXml(v)}"`)
+      .join(' ')
+    if (this.content.length > 0) {
+      let xmlStr = `<${this.tag}`
+      if (attrsStr === '') {
+        xmlStr += '>'
+      } else {
+        xmlStr += ` ${attrsStr}>`
```
calebcartwright, line 64: admittedly this may be another comment in the arena of micro-optimization, but would it be easier to just always insert the extra space (the else path), or if we want to avoid having a ` >`, add an inline ternary on the end of the initial `xmlStr` assignment? I got a bit lost in the preceding conversation so while I understand and agree the useless space is indeed undesired, I wasn't clear whether we've already got cases of that today.

Just wondering if it'd be beneficial to be able to avoid the string concatenation since this will be on a hot path with recursion

calebcartwright (commented):
Still feeling somewhat lost in the depth of the conversation but figured I'd take a look at 6571fbf4fb4ad16c4c6730e9ba0f38655cd68cdb

Only had one minor question/comment left inline but seems to me to implement the proposal from https://github.com/badges/shields/pull/5754#issuecomment-719817096 in a readable way :+1: 

I don't really have much to offer around the discussions on the different approaches, and I'm admittedly pretty heavily biased towards a "anything that fixes the FTB style badges" solution so can't help there.

One thing I did find coming back to this after a few months is that the code flow is now a bit easier to follow, but I felt some cognitive complexity was added around visualizing what the end state svgs would look like compared to the approaches that leaned more heavily on the larger template literals.

I think that's just the nature of the tradeoffs with this context but figured I'd share anyway as a perspective from someone largely unfamiliar with this area of the code base

badge-maker/lib/xml.js
chris48s: I suspect [it doesn't matter](https://blog.codinghorror.com/the-sad-tragedy-of-micro-optimization-theater/)

While conceptually this is recursive, for a badge the deepest trees we create are like 5 nodes deep so.. meh

I've pushed a version with fewer template statements, fewer concatenations and zero mutable variables. I won't make any claims about performance, but having revisited it, I reckon this is at least easier to read, so that's something.

badge-maker/lib/xml.js
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Representation of an XML element
+ */
+class XmlElement {
+  /**
+   * Xml Element Constructor
+   *
+   * @param {object} attrs Refer to individual attrs
+   * @param {string} attrs.tag
```
paulmelnikow, line 34: I'd probably call this `name` since these are called element names in XML.

paulmelnikow (commented):
This is looking really nice @chris48s! Thanks for picking this up! (And sorry I've been so swamped! I had a few minutes before our ops meeting and thought 👍🏼-ing this was probably the most impactful thing I could do.)

This approach of generating the nested elements using your XmlElement class leads to very readable code, and feels like a fine approach around which to consolidate the rest of the badge-generation code.

In the future we could consider other alternatives to the XmlElement approach, such as ReactDOMServer or various substitutes. (Does preact have something like that?) That could mean writing the badges in JSX, not sure if that would make them more readable or just the same. Honestly it's a very similar approach, just a different syntax.

Can't approve since I opened this but here is my 👍🏼

calebcartwright (approved):
Sounds like everyone is good with this so happy to give it the :+1: 

Love the improved readability and the dramatically better results these changes allow the raster server to produce :pray:

chris48s:
Thanks :+1:  I'll deploy this shortly.

Next steps:
- This unblocks the work on moving the raster server off vercel and on to heroku. Are you good to pick that up again (time permitting) now? Assuming you can remember anything about what we were doing there.. :laughing:
- Once we've had this deployed for a couple of days I'll do a badge-maker release to NPM with these changes
- As a slightly wider project we can migrate the remaining styles to use `XmlElement` but I don't think there's a big rush there

calebcartwright:
> Are you good to pick that up again (time permitting) now? Assuming you can remember anything about what we were doing there.. laughing

Sure! There will likely be a bit of reacclimating required :laughing:, but the other good news is that the changes upstream here also greatly reduce the original complextity that was needed there before (to work around things here)

PR #5756, opened 2020-10-24
[GithubCommitsSince GithubDownloads GithubRelease] GithubDownloads supports latest release by SemVer

### LINKED ISSUE #4682: Github (pre-)release downloads count latest by SemVer
**Edited by calebcartwright**: for anyone interested in implementing this, detailed steps/guidance can be found at the bottom of the thread, [here](https://github.com/badges/shields/issues/4682#issuecomment-590168537)

<hr />
:clipboard: **Description**

Make it able to get the number of downloads of the latest Github (pre-)release sort by SemVer.

:link: **Data**

Getting the latest (pre-)release sort by semver should be similar to `/github/v/release/:user/:repo?sort=semver` and `/github/v/release/:user/:repo?include_prereleases&sort=semver`.

Getting the download count should be similar to `/github/downloads/:user/:repo/:tag/total` and `/github/downloads-pre/:user/:repo/:tag/total`.

:microphone: **Motivation**

In this way, if a stable version (for example, 0.7.4) is released after a pre-prelease version (0.8.0-rc1), I can get both download counts.

hjagodzinski:
Resolves #4682.


calebcartwright:
Thank you for the PR @hjagodzinski and apologies for the delay in review. Would you mind adding a couple tests to give us some validation of the behavior?

In this case I think using mocked service tests would probably be easiest in this case because:
* test conditions too difficult to find in a real repo, and doesn't seem worth setting up a dummy repo with actual releases to use as the test target
* there has to be so much logic within `handle` and trying to mock everything out with a unit test would be too painful

hjagodzinski:
@calebcartwright I understand completely you guys have other obligations as well and finding time is a challenge at times.

I was felling uncomfortable with this changed not having any test. Having in mind our previous discussion I was not sure how should I approach this and decided I will follow your guidance.

I assume there won't be much more overhead introduced by the service tests than by the unit tests. Provided a service request mocking is inevitable in both cases the service tests seem more appealing as being easier to write (as you already noted). I have covered with new tests all four cases that this change regards.

calebcartwright:
> I was felling uncomfortable with this change not having any test. Having in mind our previous discussion I was not sure how exactly should I approach this and decided I will follow your guidance.

These are among our most popular badges (I believe Downloads/Release are top 10 actually) so the additional tests are great, thanks so much!

> I assume there won't be much more overhead introduced by the service tests than by the unit tests. Provided a service request mocking is inevitable in both cases the service tests seem more appealing as being easier to write (as you already noted). I have covered with new tests all four cases that this change regards.

Understood. It's more of a general strategy and preference for leveraging unit tests when possible whereas historically the only way anything was tested within the class was with complex mocked service tests. Most of the time using unit tests for logic and service tests for high level functional and integration validation works out for the best, but there's definitely still cases like this one where we just really need the mocked service tests :+1:

calebcartwright (approved):
LGTM, thank you!

Issue #5637, opened 2020-10-01
Hacktoberfest Combined Status Badges not updating now Live


msandfor:
Are you experiencing an issue with...

- [X] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [badge-maker NPM package](https://www.npmjs.com/package/badge-maker)

:beetle: **Description**

The badge should display an updated Hacktoberfest Status
![image](https://user-images.githubusercontent.com/5952956/94849572-35223700-0458-11eb-91df-cd10a74ee7c7.png)

but instead it is stuck at one day to go:
![image](https://user-images.githubusercontent.com/5952956/94849672-5d119a80-0458-11eb-916e-a3e87520b532.png)

We are in Day 2 now, so I am expecting to see 3 open issues, and 28 days left etc

:link: **Link to the badge**

https://img.shields.io/github/hacktoberfest/2020/msandfor/10-easy-steps?style=for-the-badge

:bulb: **Possible Solution**

I don't know what's wrong with it :-( Or how to fix

calebcartwright:
> We are in Day 2 now, 

We can't have the rendered badges be based on a locale-specific start time, so UTC is used for start and end.

That being said we do have a minor calc bug from a copy/paste (my fault!). This will resolve itself in a little under 11 hours from the time of this comment, which in all likelihood will happen before we get the bug fixed and deployed anyway.

For anyone interested in working on this (and to prevent the same issue from happening next year), the fix would be here, presumably should be minus 1 instead of plus 1

https://github.com/badges/shields/blob/0c28fe7144be142d3aa8fccc38efa22e3ee8032f/services/github/github-hacktoberfest.service.js#L211

msandfor:
Oh thanks @calebcartwright - I should have realised timezones are the root of all evil [read: confusion] and of course it would be UTC - I'm very happy to know what's up.

PR #5813, opened 2020-11-07
Fix calculation of start date for [GitHubHacktoberfest] badge

calebcartwright:
I was a little reckless with the copy/paste when updating our Hacktoberfest badge to work with the current year (it was previously hardcoded for 2019) and that resulted in a small bug in how the determination of the start date was made (refs #5637).

This PR corrects that by using October 1st (UTC) precisely


PyvesB (approved):
Looks good to me. 👍🏻

PR #5814, opened 2020-11-07
Refactor the [nuget myget] service

jakubfijalkowski:
Hi :wave:!

In this PR I refactored (and tested) the `nuget-v3-service-family`, mainly:
1. Simplified the `fetch` function,
2. Introduced the `transform` functions in each of the base implementations,
3. Tested both.

We talked about the changes with @calebcartwright in my previous PR.

There is also the possibility to unit-test the `buildRoute` and `apiUrl` functions, but I'm not quite sure whether it is worth it. We would need to expose them from the module, but the service tests validate it sufficiently. If you think otherwise then let me know and I will add these.

We could also do the same for `v2` but the `transform` method would be really a one-liner (as we get the necessary data directly from the API) and I don't think it is worth it.

### DIFF
--- a/services/nuget/nuget-v3-service-family.js
+++ b/services/nuget/nuget-v3-service-family.js
@@ -76,7 +76,7 @@ async function fetch(
   serviceInstance,
   { baseUrl, packageName, includePrereleases = false }
 ) {
-  const json = await serviceInstance._requestJson({
+  return await serviceInstance._requestJson({
     schema,
     url: await searchServiceUrl(baseUrl, 'SearchQueryService'),
     options: {
@@ -89,12 +89,6 @@ async function fetch(
       },
     },
   })
-
-  if (json.data.length === 1) {
-    return json.data[0]
-  } else {
-    throw new NotFound({ prettyMessage: 'package not found' })
-  }
 }
 
 /*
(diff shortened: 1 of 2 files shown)


services/nuget/nuget-v3-service-family.js
```diff
@@ -76,7 +76,7 @@ async function fetch(
   serviceInstance,
   { baseUrl, packageName, includePrereleases = false }
 ) {
-  const json = await serviceInstance._requestJson({
+  return await serviceInstance._requestJson({
```
calebcartwright, line 79: No need for the await here since processing of the http response is no longer being done within this async function

```suggestion
  return serviceInstance._requestJson({
```

services/nuget/nuget-v3-service-family.js
```diff
@@ -139,6 +133,21 @@ function createServiceFamily({
       return renderVersionBadge(props)
     }
 
+    /*
+     * Extract version information from the raw package info.
+     */
+    transform({ json, includePrereleases }) {
+      if (json.data.length === 1 && json.data[0].versions.length > 0) {
+        const packageInfo = json.data[0]
+        const versions = packageInfo.versions.map(item =>
```
calebcartwright, line 142: Don't feel obligated to change anything, but just noting for future reference we could utilize destructuring in these cases
e.g.
```suggestion
        const { versions: packageVersions } = json.data[0]
        const versions = packageVersions.map(item =>
```

services/nuget/nuget-v3-service-family.spec.js
```diff
@@ -0,0 +1,102 @@
...
+  return {
+    data: [],
+  }
+}
+
+function tooMuchDataJson() {
+  return {
+    data: [{}, {}],
+  }
+}
```
calebcartwright, line 40: nit but we don't really need these to be functions since they are static values

calebcartwright (commented):
Thanks for the follow up PR @jakubfijalkowski! LGTM overall, few minor comments left inline

services/nuget/nuget-v3-service-family.js
jakubfijalkowski: Ha, I think I don't get used to the `async` quirks of JS. Too much used to C#/Rust.

Fixed!

calebcartwright (approved):
Thanks!

PR #5840, opened 2020-11-17
special case generating baseUrl for production

chris48s:
refs #5014

Turns out this wasn't a big deal. Hopefully this is the last piece of the jigsaw :crossed_fingers:

### DIFF
--- a/frontend/components/customizer/customizer.tsx
+++ b/frontend/components/customizer/customizer.tsx
@@ -11,13 +11,6 @@ import {
   CopiedContentIndicatorHandle,
 } from './copied-content-indicator'
 
-function getBaseUrlFromWindowLocation(): string {
-  // Default to the current hostname for when there is no `BASE_URL` set
-  // at build time (as in most PaaS deploys).
-  const { protocol, hostname } = window.location
-  return `${protocol}//${hostname}`
-}
-
 export default function Customizer({
   baseUrl,
   title,
@@ -50,7 +43,7 @@ export default function Customizer({
 
   function generateBuiltBadgeUrl(): string {
     const suffix = queryString ? `?${queryString}` : ''
-    return `${baseUrl || getBaseUrlFromWindowLocation()}${path}${suffix}`
+    return `${baseUrl}${path}${suffix}`
   }
 
   function renderLivePreview(): JSX.Element {
--- a/frontend/constants.ts
+++ b/frontend/constants.ts
@@ -20,9 +20,14 @@ export function getBaseUrl(): string {
   https://www.shields.io/ to both generate badges with the base
   https://img.shields.io/
   */
-  const { protocol, hostname } = window.location
-  if (['shields.io', 'www.shields.io'].includes(hostname)) {
-    return 'https://img.shields.io'
+  try {
+    const { protocol, hostname } = window.location
+    if (['shields.io', 'www.shields.io'].includes(hostname)) {
+      return 'https://img.shields.io'
+    }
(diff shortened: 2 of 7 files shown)


frontend/components/customizer/customizer.tsx
```diff
@@ -15,6 +15,9 @@ function getBaseUrlFromWindowLocation(): string {
   // Default to the current hostname for when there is no `BASE_URL` set
   // at build time (as in most PaaS deploys).
   const { protocol, hostname } = window.location
```
paulmelnikow, line 17: Could you paste in some of the explanation from https://github.com/badges/shields/issues/5014#issuecomment-729154310 ?

paulmelnikow:
A couple spots this change _doesn't_ affect is the badge URLs used for the preview images in the badge listings, as well as the URL used for the suggest endpoint. I don't think either of those is a problem, but wanted to surface that observation in case you thought they might be.

chris48s:
I'm not particularly worried about the suggest endpoint, but now you've mentioned the preview images, I've realised there are quite a few other places where we're rendering a badge and passing a (potentially empty) baseUrl. Let me have a think about the scope of that.. it would be ideal we "lead by example" when it comes to badge images themselves.

chris48s:
Just thinking about it, would it make more sense to just move this logic out to https://github.com/badges/shields/blob/master/frontend/constants.ts , or are there some cases where we _don't_ want `baseUrl` to be `img.shields.io` in production?

paulmelnikow:
> Just thinking about it, would it make more sense to just move this logic out to https://github.com/badges/shields/blob/master/frontend/constants.ts , or are there some cases where we _don't_ want `baseUrl` to be `img.shields.io` in production?

I think moving it there could work. However, it may not be able to be a constant anymore. Gatsby is isomorphic, so the first time this code runs is at build time, when `window.location` doesn't exist.

chris48s:
Yep. It will need to become a `getBaseUrl()` function :+1:  I'll have a proper go at that tomorrow.

frontend/constants.ts
```diff
@@ -1 +1,33 @@
...
+  */
+  try {
+    const { protocol, hostname } = window.location
+    if (['shields.io', 'www.shields.io'].includes(hostname)) {
+      return 'https://img.shields.io'
+    }
+    return `${protocol}//${hostname}`
+  } catch (e) {
+    // server-side rendering
+    return ''
```
chris48s, line 31: This is not ideal, but I'm not sure there is a better option available given the constraints

paulmelnikow: I think instead of a `try` you can do `if (typeof window === 'undefined')`.

https://www.gatsbyjs.com/docs/debugging-html-builds/#how-to-check-if-window-is-defined

chris48s: :laughing: The thing I was commenting on here was `return ''` as opposed to the use of try/catch

It is neither here nor there whether we use a try/catch or an if/then: It is unfortunate that the server-rendered version of the page will not use `img.shields.io` as the base for badge images, but I don't see another option.

paulmelnikow: Ahh gotcha, I misunderstood.

Yea, that we can't get the right URL for the server rendering is unfortunate. It's fortunate that this path is only used for badge previews.

I think there is an alternative here in the long run, which is that we wire something up, probably in Gatsby, to render the previews at build time and inline them.

paulmelnikow (approved):

PR #5884, opened 2020-11-28
Remove some unused dev dependencies

PyvesB:
Not necessarily an exhaustive cleanup, but I noticed that the following dev dependencies were unused:
* _is-png_ is no longer used since #3644.
* _fs-readfile-promise_ and _tmp_ are no longer used since #5671.
* _cheerio_ was added to fix a warning with _chai-enzyme_ in #4324, but _chai-enzyme_ was removed in #5626.


calebcartwright (approved):
LGTM

PR #5886, opened 2020-11-29
post-Heroku migration cleanup tasks

chris48s:
Refs https://github.com/badges/shields/issues/5014#issuecomment-735405274

A few notes:

- I don't think we're going to hit any issues with this, but maybe we want to hold off merging this for a few days just to make 100% sure we definitely don't need to roll back to gh-pages
- The `add_deployment_status.yml` workflow has served us well, but it is no longer useful
- I'm 99& sure _nothing_ in the `Makefile` is useful any more, but can someone sanity check that and confirm
- I'm also removing the OVH server docs here. Espadrine has already killed the node processes on them. They will be turned off for good tomorrow.
- I had to reformat the entire table in 9be9644 to appease the pre-commit hook gods, but this makes it impossible to see what I have actually changed in the diff. Here's what actually changed: https://github.com/badges/shields/commit/9be9644dedb5e29dfe18fe59268fa3e5ec7d757f?w=1


doc/production-hosting.md
```diff
@@ -95,25 +90,11 @@ files:
 Sitting in front of the three servers is a Cloudflare Free account which
 provides several services:
 
-- Global CDN, caching, and SSL gateway for `img.shields.io`
+- Global CDN, caching, and SSL gateway for `img.shields.io` and `shields.io`
 - Analytics through the Cloudflare dashboard
-- DNS hosting for `shields.io`
```
paulmelnikow, line 100: This bullet seems worth keeping (or clarifying?) The `shields.io` DNS is hosted on Cloudflare.

doc/production-hosting.md
```diff
@@ -123,28 +104,14 @@ hosted on [Zeit Now][]. It's managed in the
 [zeit now]: https://zeit.co/now
 [svg-to-image-proxy]: https://github.com/badges/svg-to-image-proxy
 
-## Deployment
-
-The deployment is done in two stages: the badge server (heroku) and the front-end (gh-pages).
+### Heroku Deployment
 
-### Heroku
+Both the badge server and frontend are served from Heroku.
```
paulmelnikow, line 109: ❤️

paulmelnikow (commented):
This is amazing!

paulmelnikow:
> I'm 99& sure _nothing_ in the `Makefile` is useful any more, but can someone sanity check that and confirm

👍🏼

doc/production-hosting.md
chris48s: Re-reading it, I'm not really sure why I removed that line, even though it was yesterday

paulmelnikow (approved):

PR #5889, opened 2020-11-30
delete eslint-plugin-standard, update eslint-config-standard and eslint; affects [github jira scoop]

chris48s:
Refs #5879

eslint-plugin-standard is now deprecated so we can remove it, but in order to remove it we have to upgrade eslint-config-standard and eslint

The thing that bothers me here is that dependabot should have already bumped eslint-config-standard and eslint :confused: These releases are ~10 days old. I suspect there must be other packages we are behind on too.


services/jira/jira-sprint.service.js
```diff
@@ -106,6 +106,7 @@ module.exports = class JiraSprint extends BaseJsonService {
       if (issue.fields.resolution != null) {
         return issue.fields.resolution.name !== 'Unresolved'
       }
+      return undefined
```
paulmelnikow, line 109: Think this captures the idea better:

```suggestion
      return false
```

paulmelnikow (commented):
Thanks for picking this up!

I have noticed Dependabot gives up when there are conflicts between new versions of dependencies and other dependencies. I wish they surfaced a UI with a list of those, which would be really helpful.

calebcartwright:
> The thing that bothers me here is that dependabot should have already bumped eslint-config-standard and eslint confused These releases are ~10 days old. I suspect there must be other packages we are behind on too.

May be misremembering but didn't we reduce the number of allowed PRs a while back because it was causing review app starvation on the Heroku side? If so I wonder if could also be causing too narrow a funnel for Dependabot to be able to get things to us

paulmelnikow (approved):

chris48s:
> didn't we reduce the number of allowed PRs a while back

Do you know where we did that? It doesn't seem to be in https://github.com/badges/shields/blob/master/.dependabot/config.yml and I don't think we can manually change any settings in dependabot's UI when we're using a config file

calebcartwright:
There's some organization/account level settings that I believe still get applied to repositories that have a config file, and IIUC those limits may span all the enabled repos so that 25 limit would be the aggregate count of PRs in badges/shields, badges/svg-proxy, and badges/daily-test

https://app.dependabot.com/accounts/badges/settings

I've not gone back through the Friday PR rounds yet, but as we've got a sizeable dependency tree and two manifest files (root and badge-maker lib), plus the frequency of gatsby-* PRs, it wouldn't surprise me if Dependabot also has north of 25 bumps for us but can only open a subset?
image

chris48s:
Thanks :+1: 
I'm pretty sure we had less than 25 PRs last weekend, but I've changed it to unlimited for now and lets see how many PRs appear this weekend. I'll set a reminder to review it

calebcartwright:
> I'm pretty sure we had less than 25 PRs last weekend,

Yes I think so too, but 10 of the 11 open PRs in badges/svg-to-image-proxy alone are from Dependabot so that'd reduce the max down to ~15, etc. (not sure if we had any Dependabot PRs open in badges/shields or any other repos on the last Friday bump)

https://github.com/badges/svg-to-image-proxy/pulls

chris48s:
:bulb: That's a good point. I hadn't focussed that those PRs consume some of that allowance, but it is obvious now you say it.

calebcartwright:
We'll find out on Friday one way or the other I suppose. If we get 300 PRs from Dependabot we'll know for sure :laughing:

chris48s:
Judging by the number of PRs open (including some that jump months worth of releases in one e.g: PR https://github.com/badges/shields/pull/5895 ) that seems to have unblocked the dependabot-tubes, but of course now we are using all the circle ci and heroku minutes again :laughing: #jslife

calebcartwright:
Yeah they can stack up really quickly especially with the squash-when-passing label at play. I wish we could get dependabot to open these in two separate batches, perhaps ~8-12 hours apart, so that we still get things regularly but also have a bit of a window to help get some changes through

PR #5949, opened 2020-12-16
use a logo in the README that looks ok in light or dark mode

### LINKED ISSUE #5947: Modify Shields logo to adapt with dark mode.
:clipboard: **Description**

**Improvements** - Since GitHub recently rolled out dark mode, lots of users are switching to it. The current Shields logo is barely visible in dark mode (see below). 
image

Changing color to somewhat greyish or putting a white outline to the text in the image or removing the transparency may be a fix to this.

chris48s:
Closes  #5947

`#555555` already exists in the shields palette and looks ok  using github's light or dark theme

Screenshot at 2020-12-16 20-53-32
Screenshot at 2020-12-16 20-53-20

I suggest we stick with the current (black) image on shields.io itself though


calebcartwright (approved):
works for me :+1:

PR #5966, opened 2020-12-18
increase [David] service test timeouts

calebcartwright:
We've had persistent issues with the David tests (and badges) for a while a now, and an unfortunate side effect of this has been that our daily service test runs have been in a constant state of red/failure for a very long time now. I worry that this constantly failing status of our daily service test runs are starting to defeat the purpose of their existence, as it's human nature to start to ignore the jobs (I confess I rarely check each of the two daily runs anymore because of this).

The root cause of the David failures were upstream issues with the platform that largely seem to have been resolved by some infrastructure/runtime environment changes there, but we're still getting test timeout issues (https://app.circleci.com/pipelines/github/badges/daily-tests/659/workflows/eb63b021-d2a6-4afb-81ea-9ba7b68474e5/jobs/1557)

As such I'm proposing an extensive timeout increase for these David tests. Obviously any badge request that needs 15 seconds will largely be useless in most common badge contexts, but I think it's worth trying it in this case anyway. I'd really like to get our daily test runs back into a passing (and IMO useful) state, and I don't think that it's worth blocking that possibility on these tests given the historical nature of these badges and tests

### DIFF
--- a/services/david/david.service.js
+++ b/services/david/david.service.js
@@ -83,12 +83,12 @@ module.exports = class David extends BaseJsonService {
       options: { qs: { path } },
       errorMessages: {
         /* note:
-        david returns a 500 response for 'not found'
+        david returns a 504 response for 'not found'
         e.g: https://david-dm.org/foo/barbaz/info.json
         not a 404 so we can't handle 'not found' cleanly
         because this might also be some other error.
         */
-        500: 'repo or path not found or david internal error',
+        504: 'repo or path not found or david internal error',
       },
     })
   }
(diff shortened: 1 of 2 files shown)


alanshaw:
🤞 hopefully won't be necessary after https://github.com/badges/shields/pull/5968

PyvesB (approved):
Given that we haven't hear back from #5968, I'm happy to go with this alternative for now. 👍🏻

calebcartwright:
@PyvesB - sorry for the delay, needed to account for the updated 5XX response status code to get the last couple service tests passing. Should be all set now if you can re-:+1: whenever you get a chance

PyvesB (approved):

PR #5968, opened 2020-12-20
refactor: [david] use new status API

alanshaw:
This PR updates the API URL for david to use the new status API. The new API will be more reliable going forwards.

### DIFF
--- a/services/david/david.service.js
+++ b/services/david/david.service.js
@@ -73,14 +73,10 @@ module.exports = class David extends BaseJsonService {
   }
 
   async fetch({ kind, user, repo, path }) {
-    const url = `https://david-dm.org/${user}/${repo}/${
-      kind ? `${kind}-` : ''
-    }info.json`
-
     return this._requestJson({
       schema,
-      url,
-      options: { qs: { path } },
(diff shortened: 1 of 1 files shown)


calebcartwright:
Thank you for the PR! I've modified the PR title so that the corresponding tests for this service are executed, and that's showing that the tests are now failing with this change. Have only take a cursory review, but seems that when the `type` query parameter has a value like `optional` or `dev` that the API is returning empty responses. Is this expected?

https://status.david-dm.org/gh/expressjs/express?type=dev

https://github.com/expressjs/express/blob/508936853a6e311099c9985d4c11a4b1b8f6af07/package.json#L62-L80

alanshaw:
Oh no! Not that's not expected. I'll get it fixed.

PyvesB:
@alanshaw when you have a chance, it would be great if you could pick this up again! We've been holding off on #5966 which in turn is causing regular test failures, I would be keen on getting our David implementation sorted out. 😉

PyvesB:
It's been a while, so I'll go ahead and close this pull request for the time being. Feel free to ping us when and if you get a chance to circle back to it again!

PR #5972, opened 2020-12-21
Fix [Spiget] rating display issue

weihao:
# Issue Description
Spiget rating is too long, because we only need a few decimal places and the rests are pointless.
bugfix

# Issue Cause
Spiget API that shields is using returned a float number, and it was rendered without sanitization.
bugfix2

# Issue Fix
When rendering shields, we round the rating to at most 2 decimal places.

### DIFF
--- a/services/spiget/spiget-rating.service.js
+++ b/services/spiget/spiget-rating.service.js
@@ -58,7 +58,7 @@ module.exports = class SpigetRatings extends BaseSpigetService {
     return this.constructor.render({
       format,
       total: rating.count,
-      average: rating.average,
+      average: Math.round(rating.average * 100) / 100,
     })
   }
 }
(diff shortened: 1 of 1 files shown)


calebcartwright:
Thank you for the PR @weihao. The changes seem fine (although given the nature of the field in the response using `.toFixed(..)` might be a simpler way to achieve the result), but in looking at this I realized we have some minor inconsistencies in precision across our badges and want to have a quick chat with the other maintainers to see if we have any thoughts on whether this should have a precision of 1 or 2

chris48s:
Looks like we normally do `.toFixed(1)` for a rating out of 5 and `.toFixed(0)` for a percentage

weihao:
Originally I thought to use `toFixed(2)`, but I did not like how 3.1 rating could be turned into 3.10 (an extra 0 was added at the end).

But if there is a standard for the shields that let's go with it.

calebcartwright (approved):
Thanks!

Issue #4364, opened 2019-11-25
Badge Request: GitHub Check Runs Status


calebcartwright:
:clipboard: **Description**
Status badge for [check runs](https://developer.github.com/v3/checks/runs/#:~:targetText=The%20check%20runs%20API%20enables,provide%20detailed%20feedback%20on%20commits.) from GitHub. 

- Which service is this badge for e.g: **GitHub**
- What sort of information should this badge show?
  ![](https://img.shields.io/badge/build-passed-brightgreen)

:link: **Data**

- Is there a public API? **Yes**
- Does the API require an API key? Our typical GitHub auth approaches will work. Note that the v3 version of the Checks API requries a custom media type (`application/vnd.github.antiope-preview+json`) in the request header.
- Link to the API documentation. https://developer.github.com/v3/checks/runs/#list-check-runs-for-a-specific-ref

API path would be: `/{{owner}}/{{repo}}/commits/{{branch}}/check-runs?check_name={{name}}`

For example:
https://api.github.com/repos/rust-lang/mdBook/commits/master/check-runs?check_name=Rustfmt

:microphone: **Motivation**

Check Runs (and Check Suites) reflect the status of various GitHub Apps/Integrations/Services/etc. and are most often seen in the `Checks` tab for a PR. See an [example here in our repo](https://github.com/badges/shields/pull/4355/checks) which includes LGTM and Netlify.

The primary use case I'd envision for this badge would be for a specific Job status as part of the GitHub Actions ecosystem. [GitHub Workflows](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#about-workflows) consist of one or more Jobs (and Jobs consist of a sequence of tasks/steps, where each step can run a command or action).  

We already have the Workflow status badge, but a Check Runs status badge would allow for Job-level status badges too (plus it could be used for any other Check Run status)

mbtools:
Furthermore, the GH `check-status` API can return "pending" although the state of the branch/commit is "green". As noted in https://github.community/t/state-is-pending-statuses-are-missing-with-github-api-v3/14386/2, GH support recommend using the `check-runs` or `check-suites` API instead.

I shall attempt a PR for this :-)

SerpentBytes:
Could I get this one assigned to me? Thanks!

mbtools:
There's PR #7759 for it already 😃

PyvesB:
[...]

PR #5973, opened 2020-12-22
[GitHubChecksStatus] Add commit check badge

itsvs:
Related to #4364. Adds a badge for Github's [commit status](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-the-combined-status-for-a-specific-reference) endpoint (not quite the same as the `check-run` endpoints, but similar idea).

### DIFF
--- a/services/github/github-checks-status.service.js
+++ b/services/github/github-checks-status.service.js
@@ -4,9 +4,10 @@ const Joi = require('joi')
 const { NotFound, InvalidParameter } = require('..')
 const { GithubAuthV3Service } = require('./github-auth-service')
 const { documentation, errorMessagesFor } = require('./github-helpers')
+const { isBuildStatus, renderBuildStatusBadge } = require('../build-status')
 
 const schema = Joi.object({
-  state: Joi.equal('failure', 'pending', 'success'),
+  state: isBuildStatus,
 }).required()
 
 module.exports = class GithubChecksStatus extends GithubAuthV3Service {
@@ -24,8 +25,8 @@ module.exports = class GithubChecksStatus extends GithubAuthV3Service {
         repo: 'shields',
         ref: 'master',
       },
-      staticPreview: this.render({
-        state: 'success',
+      staticPreview: renderBuildStatusBadge({
+        status: 'success',
       }),
       keywords: ['status'],
       documentation,
@@ -37,8 +38,8 @@ module.exports = class GithubChecksStatus extends GithubAuthV3Service {
         repo: 'shields',
         ref: '91b108d4b7359b2f8794a4614c11cb1157dc9fff',
       },
-      staticPreview: this.render({
-        state: 'success',
+      staticPreview: renderBuildStatusBadge({
+        status: 'success',
       }),
       keywords: ['status'],
       documentation,
@@ -50,8 +51,8 @@ module.exports = class GithubChecksStatus extends GithubAuthV3Service {
         repo: 'shields',
         ref: '3.3.0',
       },
-      staticPreview: this.render({
-        state: 'success',
+      staticPreview: renderBuildStatusBadge({
+        status: 'success',
       }),
       keywords: ['status'],
       documentation,
@@ -60,40 +61,13 @@ module.exports = class GithubChecksStatus extends GithubAuthV3Service {
 
   static defaultBadgeData = { label: 'checks' }
 
-  static render({ state }) {
-    if (state === 'success') {
-      return {
-        message: `passing`,
-        color: 'brightgreen',
-      }
-    } else if (state === 'pending') {
-      return {
-        message: `pending`,
-        color: 'yellow',
-      }
-    } else {
-      return {
-        message: `failing`,
-        color: 'red',
-      }
-    }
-  }
-
   async handle({ user, repo, ref }) {
-    let state
-    try {
-      ;({ state } = await this._requestJson({
-        url: `/repos/${user}/${repo}/commits/${ref}/status`,
-        errorMessages: errorMessagesFor('ref not found'),
-        schema,
-      }))
-    } catch (e) {
-      if (e instanceof NotFound) {
-        throw new InvalidParameter({ prettyMessage: 'invalid ref' })
-      }
-      throw e
-    }
+    let { state } = await this._requestJson({
+      url: `/repos/${user}/${repo}/commits/${ref}/status`,
+      errorMessages: errorMessagesFor('ref or repo not found'),
+      schema,
+    })
 
-    return this.constructor.render({ state })
+    return renderBuildStatusBadge({ status: state })
   }
 }
--- a/services/github/github-checks-status.service.js
+++ b/services/github/github-checks-status.service.js
@@ -1,9 +1,9 @@
 'use strict'
 
 const Joi = require('joi')
+const { isBuildStatus, renderBuildStatusBadge } = require('../build-status')
 const { GithubAuthV3Service } = require('./github-auth-service')
 const { documentation, errorMessagesFor } = require('./github-helpers')
-const { isBuildStatus, renderBuildStatusBadge } = require('../build-status')
 
 const schema = Joi.object({
   state: isBuildStatus,
@@ -61,7 +61,7 @@ module.exports = class GithubChecksStatus extends GithubAuthV3Service {
   static defaultBadgeData = { label: 'checks' }
 
   async handle({ user, repo, ref }) {
-    let { state } = await this._requestJson({
+    const { state } = await this._requestJson({
       url: `/repos/${user}/${repo}/commits/${ref}/status`,
       errorMessages: errorMessagesFor('ref or repo not found'),
       schema,
--- a/services/github/github-checks-status.service.js
+++ b/services/github/github-checks-status.service.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const Joi = require('joi')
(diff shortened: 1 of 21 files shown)


calebcartwright:
Thank you for the PR @itsvs. Could you please elaborate on the perceived use case for this badge?

I suppose there's technically a brief window where the status will be unknown/dynamic but this largely becomes a static badge very quickly and I don't quite understand the utility of having a commit hash specific status badge.

Typically the build status-type badges are used to dynamically reflect the current status of project/repository/etc. that changes regularly over time, and the same holds true for the check runs data discussed in #4364. However, the status for a single commit doesn't feel like it really fits into that mold

itsvs:
Hey @calebcartwright, thanks for taking the time to look over my PR!

I agree that there's little use for a commit-specific badge, _but_ Github's commit status endpoint actually allows for the `ref` to be a hash, branch, or tag (in the latter two cases it uses the head of the ref, which can change often). The motivation behind this badge is largely similar to that of check-run or CI, but it's for projects that use custom CI and therefore don't have badges available. For example, I'm involved in a couple of projects where we use custom buildservers that report build statuses to Github this way. I was hoping to add a badge to their respective documentations that shows whether the _latest_ commit to master built successfully.

Hope this makes sense!

calebcartwright:
Thanks for the additional context, that's most helpful! In that case I think we'll want to pare back the commit-centric focus to be more explicit that any old ref will suffice. I'd suggest potentially updating the route, the internals like the class and file name, as well as examples that demonstrate non-specific commit uses (e.g. branch). We'd also want to have some non-commit hash tests too.

itsvs:
Makes sense, I'll make those changes later today. Thanks for the input!

services/github/github-checks-status.service.js
```diff
@@ -0,0 +1,99 @@
+'use strict'
+
+const Joi = require('joi')
+const { NotFound, InvalidParameter } = require('..')
+const { GithubAuthV3Service } = require('./github-auth-service')
+const { documentation, errorMessagesFor } = require('./github-helpers')
+
+const schema = Joi.object({
+  state: Joi.equal('failure', 'pending', 'success'),
```
calebcartwright, line 9: Given this is essentially a canonical build-type badge, we should use the standard build badge helpers for the schema and rendering here. That will ensure that this badge is consistent with the others in the category, and also handle any additional states that may pop up if there's any changes upstream

these helpers can be imported from `build-status` located in the services directory (`const { isBuildStatus, renderBuildStatusBadge } = require('../build-status')`

```suggestion
  state: isBuildStatus,
```

calebcartwright, line 9: Given this is essentially a canonical build-type badge, we should use the standard build badge helpers for the schema and rendering here. That will ensure that this badge is consistent with the others in the category, and also handle any additional states that may pop up if there's any changes upstream

these helpers can be imported from `build-status` located in the services directory (`const { isBuildStatus, renderBuildStatusBadge } = require('../build-status')`

```suggestion
  state: isBuildStatus,
```

services/github/github-checks-status.service.js
```diff
@@ -0,0 +1,99 @@
...
+    {
+      title: 'GitHub branch checks state',
+      namedParams: {
+        user: 'badges',
+        repo: 'shields',
+        ref: 'master',
+      },
+      staticPreview: this.render({
+        state: 'success',
+      }),
```
calebcartwright, line 29: helper can abe used here
```suggestion
      staticPreview: renderBuildStatusBadge({
        status: 'success',
      }),
```

services/github/github-checks-status.service.js
```diff
@@ -0,0 +1,99 @@
...
+    {
+      title: 'GitHub commit checks state',
+      namedParams: {
+        user: 'badges',
+        repo: 'shields',
+        ref: '91b108d4b7359b2f8794a4614c11cb1157dc9fff',
+      },
+      staticPreview: this.render({
+        state: 'success',
+      }),
```
calebcartwright, line 42:
```suggestion
      staticPreview: renderBuildStatusBadge({
        status: 'success',
      }),
```

services/github/github-checks-status.service.js
```diff
@@ -0,0 +1,99 @@
...
+    {
+      title: 'GitHub tag checks state',
+      namedParams: {
+        user: 'badges',
+        repo: 'shields',
+        ref: '3.3.0',
+      },
+      staticPreview: this.render({
+        state: 'success',
+      }),
```
calebcartwright, line 55:
```suggestion
      staticPreview: renderBuildStatusBadge({
        status: 'success',
      }),
```

services/github/github-checks-status.service.js
```diff
@@ -0,0 +1,99 @@
...
+        color: 'yellow',
+      }
+    } else {
+      return {
+        message: `failing`,
+        color: 'red',
+      }
+    }
+  }
+
```
calebcartwright, line 81: We can drop this function altogether since rendering is handled by the common util. Note that this changes the messages ever so slightly from what is here right now, but that's a desired change so that these badges are consistent with how we handle build badges

```suggestion
```

services/github/github-checks-status.service.js
```diff
@@ -0,0 +1,99 @@
...
+        schema,
+      }))
+    } catch (e) {
+      if (e instanceof NotFound) {
+        throw new InvalidParameter({ prettyMessage: 'invalid ref' })
+      }
+      throw e
+    }
+
+    return this.constructor.render({ state })
```
calebcartwright, line 97: We can simplify this a bit. The `errorMessagesFor` helper is specifying the error message explicitly for the not found scenario, so we can just specify whatever value we want there instead of putting one value that's caught (and essentially tossed) and then mapped.

In the case of a 404 with the GitHub badges, we don't really have a great way of knowing whether the issues is with the owner/repo params or the noun param (the ref in this cases), so for the not found case the majority of our GitHub badges have an error like `noun or repo not found` and we'll want to do the same here.

We should also utilize the render helper as well:

```suggestion
    let { state } = await this._requestJson({
      url: `/repos/${user}/${repo}/commits/${ref}/status`,
      errorMessages: errorMessagesFor('ref or repo not found'),
      schema,
    })

    return renderBuildStatusBadge({ status: state })
```

services/github/github-checks-status.tester.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
+
+const t = (module.exports = require('../tester').createServiceTester())
```
calebcartwright, line 3: similarly we can use the same helpers in the tests
```suggestion
const t = (module.exports = require('../tester').createServiceTester())
const { isBuildStatus } = require('../build-status')
```

services/github/github-checks-status.tester.js
```diff
@@ -0,0 +1,49 @@
...
+    color: 'brightgreen',
+  })
+
+t.create('branch checks - failing')
+  .get('/badges/shields/now-fix.json')
+  .expectBadge({
+    label: 'checks',
+    message: 'failing',
+    color: 'red',
+  })
```
calebcartwright, line 19: Having fixed tests like this with a live project has a strong potential for flakiness. For service tests we typically utilize the helpers to ensure that we're both (A) still successfully able to integrate with the upstream endpoint and work with the response and (B) produce a badge that's in the expected shape or form (in this case a build status badge).

If we have non-trivial logic handling rendering concerns or data extraction/transformation then we'll typically utilize unit tests to cover those functions directly, or in some cases mock the api response from the upstream provider so that there's a known/fixed response.

We don't need to do either of those things here (the build helpers are both heavily battle tested and extremely well covered in existing tests), so we can drop one of these and use the standard test validator in a single case

```suggestion
t.create('branch checks (branch)')
  .get('/badges/shields/master.json')
  .expectBadge({
    label: 'checks',
    message: isBuildStatus,
  })
```

services/github/github-checks-status.tester.js
```diff
@@ -0,0 +1,49 @@
...
+
+t.create('tag checks - passing').get('/badges/shields/3.3.0.json').expectBadge({
+  label: 'checks',
+  message: 'passing',
+  color: 'brightgreen',
+})
```
calebcartwright, line 41: Same theme/comment as above. I don't think we necessarily need to have tests for each ref type, but we definitely don't need more than one per ref type. I'll defer to you if you'd like to leave two cases to cover a commit and tag ref but be sure to use the `isBuildStatus` helper like the one shown in my suggested change above

services/github/github-checks-status.tester.js
```diff
@@ -0,0 +1,49 @@
...
+
+t.create('checks - nonexistent ref')
+  .get('/badges/shields/this-ref-does-not-exist.json')
+  .expectBadge({
+    label: 'checks',
+    message: 'invalid ref',
```
calebcartwright, line 47: This value will likely need to be tweaked to account for changes in the service test for the 404 scenario

calebcartwright (commented):
Thanks again for the PR @itsvs! Looks to be in fairly good order overall, and have left some inline feedback for your review. It may seem like a lot of comments, but it's largely just places where Shields has some utils/helpers that should be reused instead of duplicating the functionality those cover :+1:

itsvs:
I believe that resolves all your comments, @calebcartwright!

calebcartwright (approved):
Excellent, thank you!

PR #5974, opened 2020-12-22
use new plugin api for [jetbrains]

### LINKED ISSUE #5946: Jetbrains Badge uses old deprecated API that only supports IDEA plugins

- [ ] My own instance

:beetle: **Description**

The implementation to generate Jetbrains badges (https://github.com/badges/shields/blob/master/services/jetbrains/jetbrains-base.js) usees an old deprecated API that only works for IDEA plugins but not for other jetbrains plugins (e.g. TeamCity).
See also Ticket at Jetbrains YouTrack: https://youtrack.jetbrains.com/issue/MP-3341

The new API is ```https://plugins.jetbrains.com/api/plugins/<PLUGIN-ID>``` and supports all plugins in the Jetbrains marketplace.

Works - as long as the deprecated API is operationl (IDEA-Plugin):
:link: https://img.shields.io/jetbrains/plugin/v/15308-import-control 

Does not work (TeamCity-Plugin):
:link: https://img.shields.io/jetbrains/plugin/v/12382-inspection-violation-notification

chris48s:
This PR implements the solution I suggested in https://github.com/badges/shields/issues/5946#issuecomment-748237206 - use the new API if we've got a numeric ID. Fall back to the old (IntelliJ only) API otherwise.

This is a bit messy because the old API is XML-based whereas the new one uses JSON so I've made a base class that inherits from `BaseXmlService` and given it some methods to call/parse/validate JSON. At some point in future when the XML API is deprecated and we drop the string ID format we can just switch everything to inherit from `BaseJsonService` and tidy this up a bit.

closes #5946

### DIFF
--- a/services/jetbrains/jetbrains-base.js
+++ b/services/jetbrains/jetbrains-base.js
@@ -1,11 +1,32 @@
 'use strict'
 
 const { BaseXmlService, NotFound } = require('..')
+const { parseJson } = require('../../core/base-service/json')
 
+/*
+JetBrains is a bit awkward. Sometimes we want to call an XML API
+and sometimes we want to call a JSON API so we need a mongrel base class.
+When the legacy IntelliJ (XML) API is retired we can simplify all this and
+switch JetbrainsDownloads, JetbrainsRating and JetbrainsVersion to just
+inherit from BaseJsonService directly.
+*/
 module.exports = class JetbrainsBase extends BaseXmlService {
+  static _isLegacyPluginId(pluginId) {
+    return !pluginId.match(/^([0-9])+/)
+  }
+
+  static _cleanPluginId(pluginId) {
+    const match = pluginId.match(/^([0-9])+/)
+    if (match) {
+      return match[0]
+    }
+    return pluginId
+  }
+
+  // xml
   static _validate(data, schema) {
     if (data['plugin-repository'] === '') {
-      // Note the 'not found' response from JetBrains Plugins Repository is:
+      // Note the 'not found' response from JetBrains IntelliJ API is:
       // status code = 200,
       // body = <?xml version="1.0" encoding="UTF-8"?><plugin-repository></plugin-repository>
       // which is parsed to object = { 'plugin-repository': '' }
@@ -14,7 +35,7 @@ module.exports = class JetbrainsBase extends BaseXmlService {
     return super._validate(data, schema)
   }
 
-  async fetchPackageData({ pluginId, schema }) {
+  async fetchIntelliJPluginData({ pluginId, schema }) {
     const parserOptions = {
       parseNodeValue: false,
       ignoreAttributes: false,
@@ -24,5 +45,28 @@ module.exports = class JetbrainsBase extends BaseXmlService {
       url: `https://plugins.jetbrains.com/plugins/list?pluginId=${pluginId}`,
       parserOptions,
     })
+  }
+
+  // json
+  _parseJson(buffer) {
+    return parseJson(buffer)
+  }
+
+  static _validateJson(data, schema) {
+    return super._validate(data, schema)
+  }
+
+  async _requestJson({ schema, url, options = {}, errorMessages = {} }) {
+    const mergedOptions = {
+      ...{ headers: { Accept: 'application/json' } },
+      ...options,
+    }
+    const { buffer } = await this._request({
+      url,
+      options: mergedOptions,
+      errorMessages,
+    })
+    const json = this._parseJson(buffer)
+    return this.constructor._validateJson(json, schema)
   }
 }
--- a/services/jetbrains/jetbrains-downloads.tester.js
+++ b/services/jetbrains/jetbrains-downloads.tester.js
@@ -17,25 +17,21 @@ t.create('downloads (user friendly plugin id)')
 
 t.create('downloads')
   .get('/9435.json')
-  .intercept(
-    nock =>
-      nock('https://plugins.jetbrains.com')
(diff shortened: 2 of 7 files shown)


services/jetbrains/jetbrains-base.js
```diff
@@ -1,11 +1,32 @@
...
+    return !pluginId.match(/^([0-9])+/)
+  }
+
+  static _cleanPluginId(pluginId) {
+    const match = pluginId.match(/^([0-9])+/)
+    if (match) {
+      return match[0]
+    }
+    return pluginId
+  }
```
calebcartwright, line 24: Do we know definitively that for any `1234-foo-bar` plugin value the leading numeric portion will always match the underlying plugin id?

I.e. could a valid long form be `1234-foo-bar` but the corresponding plugin id be `4567`?

Based on some admittedly anecdotal testing I've done that does seem to be true, but if there's any conclusive doc/material from JetBrains that'd be helpful (an inline comment may be beneficial one way or the other for posterity)

services/jetbrains/jetbrains-downloads.tester.js
```diff
@@ -17,25 +17,21 @@ t.create('downloads (user friendly plugin id)')
...
-          `<?xml version="1.0" encoding="UTF-8"?>
-            <plugin-repository>
-              <category name="Code editing">
-                <idea-plugin downloads="2" size="13159" date="1485601807000" url=""></idea-plugin>
-              </category>
-            </plugin-repository>`
-        ),
-    {
-      'Content-Type': 'text/xml;charset=UTF-8',
-    }
```
calebcartwright, line 35: Do you think there'd be any benefit in leaving a mocked test that works with the legacy api and the xml response? Know we have the live one above, but wanted to ask since that was the case as well before (with all xml/legacy) and this mocked test was still added anyawy

services/jetbrains/jetbrains-rating.service.js
```diff
@@ -70,9 +74,26 @@ module.exports = class JetbrainsRating extends JetbrainsBase {
...
-    return this.constructor.render({ rating: pluginRating, format })
+    let rating
+    if (this.constructor._isLegacyPluginId(pluginId)) {
+      const intelliJPluginData = await this.fetchIntelliJPluginData({
+        pluginId,
+        schema: intelliJschema,
+      })
+      rating =
+        intelliJPluginData['plugin-repository'].category['idea-plugin'][0]
+          .rating
```
calebcartwright, line 85: Just leaving an unhelpful comment/observation on prettier's formatting here striking me as interesting :thinking:

services/jetbrains/jetbrains-rating.service.js
```diff
@@ -70,9 +74,26 @@ module.exports = class JetbrainsRating extends JetbrainsBase {
...
+    } else {
+      const jetbrainsPluginData = await this._requestJson({
+        schema: jetbrainsSchema,
+        url: `https://plugins.jetbrains.com/api/plugins/${this.constructor._cleanPluginId(
+          pluginId
+        )}/rating`,
+        errorMessages: { 400: 'not found' },
+      })
```
calebcartwright, line 93: Would it be worthwhile to pull any of this up into the base class or helper functions? Seems like there will be a lot of repetition for at least some parts (e.g. the json calls)

services/jetbrains/jetbrains-rating.tester.js
```diff
@@ -33,52 +41,32 @@ t.create('rating stars (number as a plugin id)')
...
-              <category name="User Interface">
-                <idea-plugin downloads="1714" size="208537" date="1586449109000" updatedDate="1586449109000" url="">
-                  <rating>4.5</rating>
-                </idea-plugin>
-              </category>
-            </plugin-repository>`
-        ),
-    {
-      'Content-Type': 'text/xml;charset=UTF-8',
-    }
```
calebcartwright, line 82: Same observation/question as the other tester around whether it'd make sense to maintain a mocked test 

(note I'm not saying that we must keep these, just wondering if it'd be beneficial to do so)

calebcartwright (commented):
Approach and implementation generally LGTM, few inline thoughts below

services/jetbrains/jetbrains-base.js
chris48s: There's nowhere it is documented explicitly, but given that these (for example) do the same thing:
https://plugins.jetbrains.com/plugins/list?pluginId=1347-scala
https://plugins.jetbrains.com/plugins/list?pluginId=1347foobar
I'm pretty confident that the string part is irrelevant.

services/jetbrains/jetbrains-downloads.tester.js
chris48s: done in https://github.com/badges/shields/pull/5974/commits/3f8070d6f337f451849c94c0e350c1b1a37d46ab

services/jetbrains/jetbrains-rating.service.js
calebcartwright: Doesn't matter all that much to me, but seems like the only differences are the value for the schema and an optional/additional trailing path segment on the uri which could both be handled in a single foo/fetch type function (we do something similar for Jenkins with more complex variances on the calls)

services/jetbrains/jetbrains-base.js
calebcartwright: That's so strange on the JetBrains side (accepting any randomness after the id) :laughing:

calebcartwright (approved):

PR #5975, opened 2020-12-22
Test for server running with metrics enabled

platan:
This PR adds the first test that runs the server with metrics enabled. I suggested adding such a test in https://github.com/badges/shields/pull/5962#issuecomment-748503871.


core/server/server.js
```diff
@@ -90,10 +90,10 @@ const publicConfigSchema = Joi.object({
         .integer()
         .min(1)
         .when('enabled', { is: true, then: Joi.required() }),
-      intervalSeconds: Joi.number()
```
platan, line 93: Actually, nothing changes here, but after few commits Prettier formatted it this way.

chris48s:
The code LGTM. I don't know why circle doesn't want to build this branch. Can you try pushing an empty commit or something and see if that kicks off a build

chris48s (approved):

Issue #5948, opened 2020-12-16
New badge: Github generic issue search badge


chris48s:
:clipboard: **Description**

This started as a discussion https://github.com/badges/shields/discussions/5943 but I think we've decided this is a good idea, so I'm moving it to an issue. Here's the plan:

In the github front-end, it is possible to construct fairly nuanced issue/PR queries. For example:
* `is:open is:issue label:core label:bug author:app/sentry-io` - open issues with the labels "core" and "bug" opened by sentry-io bot: https://github.com/badges/shields/issues?q=is%3Aopen+is%3Aissue+label%3Acore+label%3Abug+author%3Aapp%2Fsentry-io
* `is:pr is:closed label:npm-package review:approved author:chris48s` - closed PRs labelled "npm-package" submitted by user "chris48s"  with one or more approving reviews: https://github.com/badges/shields/pulls?q=is%3Apr+is%3Aclosed+label%3Anpm-package+review%3Aapproved+author%3Achris48s

We already have a wide variety of github issue badge variants. Rather than gradually implement that syntax one feature request at a time as people want more nuanced queries, lets implement one badge:

Github Issue (with arbitrary query): `/github/issue-search?q=...`

This would allow us to call for example `/github/issue-search?q=repo%3Abadges%2Fshields%20is%3Aopen%20is%3Aissue%20label%3Acore%20label%3Abug%20author%3Aapp%2Fsentry-io`

which would execute the GraphQL query

```gql
query {
  search(query: "repo:badges/shields is:open is:issue label:core label:bug author:app/sentry-io", type: ISSUE) {
    issueCount
  }
}
```

Then we can basically say most of our existing issue badge variants are a special case of that in one way or another (certainly all the ones that just display a total). We could either:

* Make our existing random mashup of github issue/PR variant badges redirects to preset queries on that general badge **or**
* Continue serving those badges directly without issueing a redirect using a single shared `fetch()` function and simplify the internals

Either way, we say "no more presets" and any arbitrary query you can make through the GH frontend can also be a badge and we can clean up this mess:

![Screenshot at 2020-12-14 20-47-52](https://user-images.githubusercontent.com/6025893/102134380-52a75000-3e4e-11eb-8b9d-48cd549387aa.png)

:link: **Data**

GitHub GraphQL API

:microphone: **Motivation**

* Allow any issue badge for 'number of issues/PRs matching [arbitrary criteria]'
* Tidy up our existing pile of github issue badge variants

[...]

PR #5976, opened 2020-12-23
add generic [GithubIssuesSearch] badges

chris48s:
Refs #5948

Open to feedback on this, but I decided to implement two variants of this badge:

* `/issues-search?query=`
* `/issues-search/:user/:repo?query=`

`/issues-search?query=` is more general. Because it doesn't require a repo to be specified, it allows you to implement queries like "all issues raised by user X in any repo" or "all issues with label Y in organization Z" for example.

I decided to also add a special-case route for `/issues-search/:user/:repo?query=`. This is less flexible, but I think it is worth having a route that allows the query param you paste in to be the exact same query you will construct in the filter box on your issues page (https://github.com/badges/shields/issues ) without having to remember to append `repo:badges/shields` to the query. All our existing issue/PR badges are about querying issues within a repo so I think its fair to say this is the most common case.

`/issues-search/badges/shields?query=is%3Aopen%20label%3Ablocker` and
`/issues-search?query=repo%3Abadges%2Fshields%20is%3Aopen%20label%3Ablocker` do the same thing.

For now, I've decided to just start by adding this badge. I haven't done anything to the existing badges yet. I've left an inline comment explaining why it isn't quite a one-to-one tradeoff to switch from the `repository` API to the `search` API. Either way I think this should be pretty handy and provide an answer for a whole class of feature requests.

### DIFF
--- a/services/github/github-issues-search.service.js
+++ b/services/github/github-issues-search.service.js
@@ -1,0 +1,112 @@
+'use strict'
+
+const gql = require('graphql-tag')
+const Joi = require('joi')
+const { metric } = require('../text-formatters')
+const { nonNegativeInteger } = require('../validators')
+const { GithubAuthV4Service } = require('./github-auth-service')
+const { documentation, transformErrors } = require('./github-helpers')
+
+const issueCountSchema = Joi.object({
+  data: Joi.object({
+    search: Joi.object({
+      issueCount: nonNegativeInteger,
+    }).required(),
+  }).required(),
+}).required()
+
+const queryParamSchema = Joi.object({
+  query: Joi.string().required(),
+}).required()
+
+class BaseGithubIssuesSearch extends GithubAuthV4Service {
+  static category = 'issue-tracking'
+  static defaultBadgeData = { label: 'query', color: 'informational' }
+
+  static render({ issueCount }) {
+    return { message: metric(issueCount) }
+  }
+
+  async fetch({ query }) {
+    const data = await this._requestGraphql({
+      query: gql`
+        query($query: String!) {
+          search(query: $query, type: ISSUE) {
+            issueCount
+          }
+        }
+      `,
+      variables: { query },
+      schema: issueCountSchema,
+      transformErrors,
+    })
+    return data.data.search.issueCount
+  }
+}
+
+class GithubIssuesSearch extends BaseGithubIssuesSearch {
+  static route = {
+    base: 'github',
+    pattern: 'issues-search',
+    queryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'GitHub issue custom search',
+      namedParams: {},
+      queryParams: {
+        query: 'repo:badges/shields is:closed label:bug author:app/sentry-io',
+      },
+      staticPreview: {
+        label: 'query',
+        message: '10',
+        color: 'blue',
+      },
+      documentation,
+    },
+  ]
+
+  async handle(namedParams, { query }) {
+    const issueCount = await this.fetch({ query })
+    return this.constructor.render({ issueCount })
+  }
+}
+
+class GithubRepoIssuesSearch extends BaseGithubIssuesSearch {
+  static route = {
+    base: 'github',
+    pattern: 'issues-search/:user/:repo',
+    queryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'GitHub issue custom search in repo',
+      namedParams: {
+        user: 'badges',
+        repo: 'shields',
+      },
+      queryParams: {
+        query: 'is:closed label:bug author:app/sentry-io',
+      },
+      staticPreview: {
+        label: 'query',
+        message: '10',
+        color: 'blue',
+      },
+      documentation,
+    },
+  ]
+
+  async handle({ user, repo }, { query }) {
+    query = `repo:${user}/${repo} ${query}`
+    const issueCount = await this.fetch({ query })
+    return this.constructor.render({ issueCount })
+  }
+}
+
+module.exports = {
+  GithubIssuesSearch,
+  GithubRepoIssuesSearch,
+}
--- a/services/github/github-issues-search.tester.js
+++ b/services/github/github-issues-search.tester.js
@@ -1,0 +1,50 @@
+'use strict'
+
+const { isMetric } = require('../test-validators')
+const { ServiceTester } = require('../tester')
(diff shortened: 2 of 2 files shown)


services/github/github-issues-search.tester.js
```diff
@@ -0,0 +1,50 @@
...
+    message: 'invalid query parameter: query',
+  })
+
+t.create('GitHub Repo issue search (invalid repo)')
+  .get(
+    '/issues-search/badges/helmets.json?query=is%3Aclosed%20label%3Ablocker%20'
+  )
+  .expectBadge({
+    label: 'query',
+    message: '0',
```
chris48s, line 49: One thing I hadn't focussed on when I originally proposed this is that if we use the `search` API instead of the `repository` API, passing a repo that doesn't exist just returns zero search results (same as passing a username or label that doesn't exist, or a text string that isn't found in any issue). I think that's OK for a custom query endpoint, but I think that changes the tradeoff when it comes to making all our other issue badges a special-case of this one. We'd also have to accept worse error handling, so maybe we just leave them as they are.

calebcartwright:
Admittedly my back-of-the-napkin math may be off here, but do we know what the impact switching to the `search` API will have on our rate limiting and token pool given the [custom (and lower) rate limit that the search API has](https://docs.github.com/en/free-pro-team@latest/rest/reference/search#rate-limit) (30 requests per minute)?

For the most recent 15 minute window, we've had GitHub Issues badges ~3.3k requests so it seems like that would start eating through our tokens pretty quickly. I know we've got a lot of quota available, but do we have enough to sustain that difference/however long it takes for the quota to be reset on the tokens?

chris48s:
The documentation you've linked to there is for the REST (v3) API.

On the GraphQL API, its just a flat 5,000 points per hour https://docs.github.com/en/free-pro-team@latest/graphql/overview/resource-limitations

calebcartwright:
> The documentation you've linked to there is for the REST (v3) API

Good catch, thanks for that.

Another question in a similar vein for the non-repo scoped queries largely for confirmation, the cross-repo queries doesn't cause any additional point hits because it's a flat count that we're returning vs. nodes for each repo, right?

chris48s:
Yes. Just getting `issueCount` is a one point query. How specific or general the query is doesn't impact on the query cost.

```gql
query {
  search(query: "foo", type: ISSUE) {
    issueCount
  }
  rateLimit {
    cost
  }
}
```

```json
{
  "data": {
    "search": {
      "issueCount": 1127897
    },
    "rateLimit": {
      "cost": 1
    }
  }
}
```

---

```gql
query {
  search(query: "repo:badges/shields is:closed label:bug author:app/sentry-io", type: ISSUE) {
    issueCount
  }
  rateLimit {
    cost
  }
}
```

```json
{
  "data": {
    "search": {
      "issueCount": 35
    },
    "rateLimit": {
      "cost": 1
    }
  }
}
```

services/github/github-issues-search.tester.js
calebcartwright: Can't remember which service for the life of me but feel like we had a recent-ish new service that managed to hook into some extra node/field that gave more contextual error information.

Probably not the end of the world from an error message, but if it is problematic not being able to differentiate the repo not found vs found but zero results then we could consider firing off a different call in parallel to check whether the repo exists and use that in the error message (or only do the check in the case of zero results)

services/github/github-issues-search.service.js
```diff
@@ -0,0 +1,112 @@
...
+
+  static render({ issueCount }) {
+    return { message: metric(issueCount) }
+  }
+
+  async fetch({ query }) {
+    const data = await this._requestGraphql({
+      query: gql`
+        query($query: String!) {
+          search(query: $query, type: ISSUE) {
```
calebcartwright, line 34: will the `type: ISSUE` filter out PRs? If so will we need a way to make this conditional to enable our existing badge service classes to be able to leverage?

services/github/github-issues-search.service.js
```diff
@@ -0,0 +1,112 @@
...
+        label: 'query',
+        message: '10',
+        color: 'blue',
+      },
+      documentation,
+    },
+  ]
+
+  async handle(namedParams, { query }) {
+    const issueCount = await this.fetch({ query })
```
calebcartwright, line 71: I don't recall the set of allowed/used characters in a query but any risk of uri encoding/decoding issues with the query?

calebcartwright (commented):
LGTM, couple inline questions

services/github/github-issues-search.service.js
chris48s: `type: ISSUE` includes issues and PRs. If your query includes `is:issue` you just get issues. If your query includes `is:pr` you get just PRs. If it includes neither, you get both.

services/github/github-issues-search.service.js
chris48s: You need to URL encode your query appropriately in the badge URL, but by the time we get to `handle()` it has been decoded elsewhere i.e: if you call `/issues-search/badges/shields?query=is%3Aclosed%20label%3Abug%20author%3Aapp%2Fsentry-io` when we invoke `handle()` the value of `query` is `is:closed label:bug author:app/sentry-io` not `is%3Aclosed%20label%3Abug%20author%3Aapp%2Fsentry-io` so there's nothing extra to do there. If you don't URLEncode the quotes and colons and spaces and things in the URL it won't work but y'know.. that's URLs for you :shrug: nothing special going on here.

chris48s: The other type of escaping worth commenting on is when we pass the query in the GraphQL Body.

Because the query is a GraphQL variable (we're not using string concatenation), we also don't need to escape the quotes, so whereas in https://github.com/badges/shields/discussions/5943#discussioncomment-211217 we need to escape the quotes, with this badge if the query is `repo:TheIOFoundation/ProjectLockdown is:open is:issue label:"Need: Position" label:"Stage: Ready"` we can call `/issues-search?query=repo%3ATheIOFoundation%2FProjectLockdown%20is%3Aopen%20is%3Aissue%20label%3A%22Need%3A%20Position%22%20label%3A%22Stage%3A%20Ready%22` and neither the user nor us need to directly escape the quote characters in that query.

calebcartwright (approved):
LGTM! Definitely agreed that this has the potential to open up a lot of things for users, excited to see it

PR #5979, opened 2020-12-24
Fixed escapes single '_' of badge urls; affects [static website]

SeaHOH:
The second adjacent single underscore in badge urls has been not escaped.

Before:

    '_aa_bb_c_dd__ee_' -> ' aa bb c_dd__ee '

Patched:

    '_aa_bb_c_dd__ee_' -> ' aa bb c dd__ee '


chris48s:
Thanks for having a look at this. I went to have a look at the tests for this function to try and review what the function is supposed to do and realised.. there aren't any :facepalm: 

I guess that makes you the expert at this point :pray:   Would you be able to willing to add a `path-helpers.spec.js` with a really small test suite for this function including the example you've identified and a small selection of other cases using a similar format to this: https://github.com/badges/shields/blob/cd531ffc65c2e833797e34bfd62b748ad7af710e/core/badge-urls/make-badge-url.spec.js#L37-L43 just so we leave this a little better than we found it?

SeaHOH:
I does not use CircleCI, give a review please.

calebcartwright:
Looks like it needs a prettier run to fix the formatting (`npm run prettier` should do the trick)

chris48s (approved):
great first contribution :+1:

PR #5999, opened 2020-12-30
github auth endpoint improvements

chris48s:
I merged #5998 deployed and invalidated the relevant cache in CF. Turned out CF still caches the "Invalid secret" response, so if you hit the endpoint without auth, you then can't log in afterwards 🙃 This PR does a few more things:

- Serve the "Invalid secret" response with an appropriate status code (not 200 OK)
- Explicitly set `cache-control: private` on the "Invalid secret" response
- Fix an outdated comment
- Add some tests for this (note the 'password is missing' test only runs if you invoke with `SLOW=true`)

### DIFF
--- a/services/github/auth/admin.js
+++ b/services/github/auth/admin.js
@@ -16,11 +16,13 @@ function setRoutes({ shieldsSecret }, { apiProvider, server }) {
   // password.
   //
   // e.g.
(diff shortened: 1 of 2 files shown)


calebcartwright (approved):
lgtm

PR #6000, opened 2020-12-31
[PyPI] Fixes Python versions order by using proper Semver comparison

HorlogeSkynet:
Hello over here :wave:

Now that Python 3.10 has been released, the current _naive_ string comparison performed on Python versions proved its limits.  
This patch implements a _proper_ Semver comparison to make (for instance) `3.10` "greater" than previous versions.

Feel free to edit/adapt this as required, I'm not used to JavaScript contributions.

Bye :bow:


calebcartwright:
Thank you for the PR! Could you please add some tests to accompany the changes?

I'd recommend adding a few unit tests that hit the `render` function directly. Fore reference here's an example from [Codecov](https://github.com/badges/shields/blob/master/services/codecov/codecov.spec.js) as well as t[he tests from the Django version sorting helper function](https://github.com/badges/shields/blob/b69f055325377a7711a0918b52916fb3f795f61b/services/pypi/pypi-helpers.spec.js#L74)

HorlogeSkynet:
Thanks for your message @calebcartwright ! Please find a new specification file in the latest `HEAD` commit, implementing some very basic tests for the `render` function. Bye :wave:

calebcartwright (approved):
Thanks!

fbernhart:
Nice work! 👍🏼 I've already addressed the root cause as issue with PyPI here: https://github.com/pypa/warehouse/issues/8843

I guess as soon as https://github.com/pypa/trove-classifiers/pull/57 get's merged, this sorting here might be obsolete. But I'm not 100% sure.

Just wanted to let you guys now. 😊

PR #6009, opened 2021-01-01
add docs on input validation

chris48s:
Took a while to get to this, but I've attempted to summarise the discussion in #5619 into a some documentation.


doc/input-validation.md
```diff
@@ -0,0 +1,50 @@
...
+
+- The schema/validation we choose is informed by the assumptions we're making about the data. e.g:
+
+  - If we're going to use a value, make sure it exists.
+  - If we need to multiply it by something, we check its a number.
+  - If we're going to call `.split()` on it, we make sure its a string.
+  - If we're going to address `foo[0]`, `foo` must be an array.
+  - If we're going to sort a version on the assumption it is a semver, check its a semver
+
+- We don't need to validate characteristics we don't rely on. For example, if we're just going to render a version on a badge, it doesn't matter what format the version number is in.
```
calebcartwright, line 25: Would it be helpful to expand the example a bit? For example:

```suggestion
- We don't need to validate characteristics we don't rely on. For example, if we're just going to render a version on a badge with the same format and value from the API response and do not need to sort or transform the value, then it doesn't matter what format the version number is in. We can use a very relaxed schema to validate ins this case, e.g. `Joi.string().required()`
```

doc/input-validation.md
```diff
@@ -0,0 +1,50 @@
...
+  }).required()
+  ```
+
+  and have both the license and version badges validate the response against that schema.
+
+- For build status badges there is a shared [isBuildStatus](https://github.com/badges/shields/blob/master/services/build-status.js) validator. In most cases build status badges should use `isBuildStatus` or input validation and `renderBuildStatusBadge` should be used for rendering. Any additional status values can be added to the relevant color arrays.
+
+## Identifying problems
+
+- If we know of a real-world example of a package/repo/etc that causes us to render an error value on a badge (e.g: ![](https://img.shields.io/badge/version-null-blue) ![](https://img.shields.io/badge/coverage-NaN%25-red) ![](https://img.shields.io/badge/build-undefined-red) ) our input validation is broken and we should fix it.
```
calebcartwright, line 46: I don't have a suggested alternative ATM, but think we should consider using something other than `error value` to avoid any accidental connotations about a _literal_ error as opposed to a "bad" value

doc/input-validation.md
```diff
@@ -0,0 +1,50 @@
...
+
+- If we know of a real-world example of a package/repo/etc that causes us to thrown an unhandled runtime exception, our input validation is broken and we should fix it.
```
calebcartwright, line 48: minor typo

```suggestion
- If we know of a real-world example of a package/repo/etc that causes us to throw an unhandled runtime exception, our input validation is broken and we should fix it.
```

calebcartwright (commented):
Thanks for putting this together! Overall LGTM, few inline comments below highlighting a couple typos along with a handful of suggested textual tweaks/augments

doc/input-validation.md
chris48s: Do you think "invalid" would be better?

doc/input-validation.md
```diff
@@ -0,0 +1,50 @@
...
+
+When we receive input data from an upstream API, we perform input validation to:
+
+- Ensure we won't throw a runtime error trying to render a badge
+- Ensure we won't render badges with spurious or unexpected output e.g: ![](https://img.shields.io/badge/version-null-blue) ![](https://img.shields.io/badge/coverage-NaN%25-red) ![](https://img.shields.io/badge/build-undefined-red) ![](https://img.shields.io/badge/coverage---10%25-critical) etc
+- Express and document our understanding of the input data
+
+## Writing schemas and validation
+
+- The default validation mechanism should be to use [Joi](https://github.com/sideway/joi) to define a schema for the input data. Validation against Joi schemas is implemented in the base classes every integration inherits from. Sometimes additional manual validation is needed which can't be covered by Joi and plugins in which case we implement it by hand.
```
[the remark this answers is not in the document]
chris48s: The reason why I picked the word "integration" here is because when I'm being careful, I think the correct words are:

* [NPM](https://www.npmjs.com/) is a **service**
* `class NpmVersion` is an **integration**
*  is a **badge**

but also I suspect we're probably not great about using that terminology throughout the docs. I've made a wording change based on your suggestion.

doc/input-validation.md
calebcartwright: Yeah i think that works. The only word that was coming to mind for me was "bogus" which wasn't terribly descriptive nor helpful :laughing:

doc/input-validation.md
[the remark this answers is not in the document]
calebcartwright: > I think the correct words are:

    NPM is a service
    class NpmVersion is an integration
    is a badge

I know somewhere we've an open discussion about terminology so I'll try to find that and post my thoughts over there. Updated text LGTM

calebcartwright (approved):
LGTM! Not sure if sothers want to weigh in, but I feel this accurately reflects what seems to be the overwhelming consensus on our target prescribed guidance for validating inputs

PR #6010, opened 2021-01-01
use new [HexPM] latest_stable_version property

chris48s:
This PR switches us to using the new `latest_stable_version` property in Hex's API. As far as I can tell, if there is no stable version available `latest_stable_version` is the latest unstable (e.g: https://hex.pm/api/packages/aadya ) so we don't need to explicitly code `latest_stable_version || latest_version` or whatever ourselves. This API change means we could also implement a vpre/latest unstable badge with `latest_version` if we wanted, but I haven't done it here.


calebcartwright (approved):
lgtm

PR #6018, opened 2021-01-06
improve [pypi] license parsing from classifiers

### LINKED ISSUE #6013: Different badges depending on how a license is specified in a Python package

- [ ] My own instance

:beetle: **Description**

Shields.io generates different license shields for PyPI dependent on how a license is specified in `setup.py`.

| Specified via...        | PyPI Package | Badge |
| --- | --- | --- |
| Parameter | Coverage | PyPI - License |
| Classifier | pyMetaClasses | PyPI - License |
 

**Specified via parameter**
Example repository: https://github.com/nedbat/coveragepy  
PyPI Package: https://pypi.org/project/coverage/  
Badge: PyPI - License
```python
setup(
# ...
  license='Apache 2.0',
  classifiers=[
    # ...
    "License :: OSI Approved :: Apache Software License",
    # ...
    ]
# ...
)
```
Source: https://github.com/nedbat/coveragepy/blob/master/setup.py?ts=2#L112

**Specified via license classifier**
Example repository: https://github.com/Paebbels/pyMetaClasses
PyPI Package: https://pypi.org/project/pyMetaClasses/  
Badge: PyPI - License
```python
setup(
# ...
  classifiers=[
    # ...
    "License :: OSI Approved :: Apache Software License",
    # ...
  ]
# ...
)
```
Source: https://github.com/Paebbels/pyMetaClasses/blob/master/setup.py?ts=2#L74

The `license` parameter for `setuptools.setup()` is not documented in the user guide (see [setuptools "keywords"](https://setuptools.readthedocs.io/en/latest/userguide/keywords.html)), but in [full "keyword" list](https://setuptools.readthedocs.io/en/latest/references/keywords.html?highlight=license#keywords) and also not [recommended anymore](https://setuptools.readthedocs.io/en/latest/history.html?highlight=license#id232):
> v41.1.0 - Documentation changes: Use license classifiers rather than the license field.

:bulb: **Possible Solution**

Please check why shields.io produces different results based on how a license is specified in a package or how PyPI reports the result. My expectation is an equal result for the old and new style.

chris48s:
Having had a look at this, the existing `parenthesizedAcronymRegex` and `bareAcronymRegex`s are actually doing a pretty good job of converting most of the common licenses (e.g: MIT, BSD, GPLv3, etc) to something we can colour-code but I've added 3 special cases for a few common licenses (Apache-2.0, CC0-1.0, AGPL-3.0) to convert the classifier to a SPDX ID. I've also added a `preserveCase` flag so we can optionally keep the classifier case intact.

I've left us prioritising the `license` key over classifier strings.

Closes #6013


calebcartwright (approved):
Code LGTM. Do we have a feel for the potential impact of this change on existing badges (risk of folks with pitchforks upset about the change), and whether we should do something like a pinned issue to explain? Or can this be classified as a bug fix?

chris48s:
So in terms of the impact, there are 3 cases where this changes something (and only if your project uses classifiers but _doesn't_ declare a license property). The old outputs would have been





The new versions will show a SPDX ID with standard colouring. We can consider those cases as fixes/improvements.

Issue #4655, opened 2020-02-14
migrate away from request


chris48s:
:clipboard: **Description**

Request is no longer maintained:

https://github.com/request/request#deprecated

> As of Feb 11th 2020, request is fully deprecated. No new changes are expected to land. In fact, none have landed for some time.

We need to switch to a different library. This is going to be an absolutely monstrous job.

@calebcartwright you mentioned in https://github.com/badges/shields/pull/4644 that there has been some discussion around this already. I couldn't find an issue anywhere. Do you have any kind of summary of where we've got to on this. I'm aware its somewhat complex as our use of the library is quite involved, there are many other libraries we could potentially migrate to, and none are an obvious like-for-like replacement for request.

calebcartwright:
> @calebcartwright you mentioned in #4644 that there has been some discussion around this already. I couldn't find an issue anywhere.

I believe it was in one of the private channels on Discord, will try to dig it up. IIRC we didn't go into much detail, just that it'd need to be replaced and that we'd have to keep in mind Self-hosting Shields server users which currently benefit from requests ability to auto-detect and honor the standard proxy env vars (something I believe is missing from most request alternatives

paulmelnikow:
This is related to #3368, but yea, I remember @calebcartwright mentioning the limitations in proxy support in other HTTP clients. That was a while ago. I wonder if there is a solution for any of those now.

Should we merge these two issues?

chris48s:
> Should we merge these two issues?

Yeah sorry. They are probably the same thing.

paulmelnikow:
It seems like we should open a new issue which combines the scope of both. Request being unmaintained is making this more urgent than it was before, and deserves high billing, although the first order of business is probably to test the effectiveness of that memory cache.

chris48s:
We talked a bit the other day about some of the mechanics of this (e.g: an adaptor layer that allows our service code to use request-style params with another library), but it seems like a very major decision we need to make is which library we're going to migrate to. While its on my mind, I'll post this link to discussion of [possible alternatives](https://github.com/request/request/issues/3143) as a starting point, but in making this choice we also need to try and work out what our requirements even are (what is the current subset of the request feature set we depend on?)

calebcartwright:
[...]

PR #6037, opened 2021-01-08
Remove the in-memory cache

paulmelnikow:
Shields has long had an in-memory cache with some complex logic for determining when the cached values are used and when they are flushed. At the time this was implemented, squeezing cache performance was helpful since there was no downstream cache. For years now we've used Cloudflare as a CDN, so trying to cache onboard is less useful than before. Furthermore, since the cache is very small and only used in fairly restrictive ways, it probably has very little impact on performance.

This feels like "step one" in addressing #3368 and #4655, as it's disabling most of the complicated behavior in the existing request-handling code.

In addition, #3368 is a dependency for #3328, and given issues like badges/sc#5 (a DOS vulnerability in Shields), ScoutCamp feels like a liability. Getting the train rolling toward #3328 feels useful.

There was some discussion on this in the ops meeting today. This change is slightly experimental (I think there is maybe a 1% chance it causes a serious problem) but it's easy enough to roll back if it does.

Ref #3368


core/base-service/base-non-memory-caching.js
```diff
@@ -1,74 +0,0 @@
...
-//
-// In contrast, The `handle()` function of most other `BaseService`
-// subclasses is wrapped in onboard, in-memory caching. See `lib /request-
-// handler.js` and `BaseService.prototype.register()`.
-//
-// All services, including those extending NonMemoryCachingBaseServices, may
-// be cached _downstream_. This is governed by cache headers, which are
-// configured by the service, the user's request, and the server's default
-// cache length.
-module.exports = class NonMemoryCachingBaseService extends BaseService {
```
paulmelnikow, line 27: This class only existed to dodge the cache behavior which is being removed herewith. The two services using this can now use BaseService directly.

calebcartwright (approved):
> For years now we've used Cloudflare as a CDN, so trying to cache onboard is less useful than before

Also helps that we require traffic to ingress via cloudflare (no more possibility of hitting badge servers directly). 

Changes LGTM, though image the merging and deployment of this one will be done at a predetermined time so the experiment can be executed in isolation for clean observation.

I assume we'll want to keep an eye on the overall request load and uncached requests hitting the backend, with a reference to trending response time to see what, if any, impact this has from the removal.

May be interesting to see if there's any minor drop in memory usage too, though don't expect much on that front

paulmelnikow:
I just deployed the last commit to master and could merge and deploy this one next. We could give it the weekend and maybe Monday morning for observation, and roll back if necessary.

PR #6048, opened 2021-01-12
Use GitHub Sponsors logo instead GitHub logo

outsideris:
As we discussed when I add [GitHub Sponsors badge](https://github.com/badges/shields/pull/5694#issuecomment-707836746), I added [Sponsors logo in simple-icons](https://github.com/simple-icons/simple-icons/pull/3730).

And then it released in [v4.4.0](https://github.com/simple-icons/simple-icons/releases/tag/4.4.0) and shields are now using `simple-icons@4.5.0`. 

So, I changed to GitHub Sponsors logo.

스크린샷 2021-01-13 오전 4 07 06

### DIFF
--- a/services/github/github-sponsors.service.js
+++ b/services/github/github-sponsors.service.js
@@ -36,7 +36,7 @@ module.exports = class GithubSponsors extends GithubAuthV4Service {
 
   static defaultBadgeData = {
     label: 'sponsors',
-    namedLogo: 'github',
+    namedLogo: 'github sponsors',
   }
 
   static render({ count }) {


calebcartwright:
Thank you for following up on this and submitting the PR. However, I think we're unlikely to merge, at least not any time soon. There were subsequently some questions about whether this badge would be a better fit for the `funding` category instead (refs #5846), in which case some of this code would be removed. 

The custom logo can already be used today via the query param, so the new icon can be utilized whether or not we change the default.
https://img.shields.io/github/sponsors/Homebrew?style=social&logo=github-sponsors

outsideris:
I see. Feel free to close this PR if you prefer that users use `logo=` param.

calebcartwright:
> Feel free to close this PR if you prefer that users use logo= param

In cases where a user wants an icon on their badge, they can always use the `logo` param. That's not really the reasoning for the putting the PR on hold. If we decide to keep the badge in the current category, then we can revisit whether to change the default icon used with the social style as proposed by this PR.

However, since there's a decent chance we will change the category, it doesn't really make sense to toggle the default logo because we don't use default logos in non-social styles.

Will leave this open for the time being and can revisit once some of the other questions have been answered

outsideris:
understood! Thank you.

PyvesB:
> However, since there's a decent chance we will change the category, it doesn't really make sense to toggle the default logo because we don't use default logos in non-social styles.

We've indeed changed the category and as you can see in #6512 there no longer is a default logo. Will close this pull request.

Regardless, thanks for your interest in Shields.io @outsideris!

Issue #6016, opened 2021-01-05
Add a badge that shows the date of the newest docker image available


yspreen:
:clipboard: **Description**

This badge would work similarly to the docker version badge. Instead of showing the latest tag by date, it would show the latest date of the newest tag. This shows that a docker image is up to date.

:link: **Data**

The docker version badge already makes use of the date of images. I was thinking about creating a PR myself, but the docker version service looked a bit intimidating so far. As far as I can tell, the same API should also provide the newest date.

:microphone: **Motivation**

Docker hub builds images according to GitHub sources. But there's no way of knowing if a Docker image is actually maintained by looking at the git repo. It should be possible to see at a glance: How old is this image? Is the underlying ubuntu/alpine/debian up to date or is it ancient? Does it make sense to build the image on my own, or can I use this one and trust that it has recent software in it?

You don't have to push code to a git repo to build an image on Docker Hub. In fact, I don't change my repos much but I trigger rebuilds all the time. Someone who looks at a git repo without a commit in 2y might think that its image is outdated but it might be really new.

chris48s:
dockerhub api has a `last_updated`  property on the image object so this should be fairly straightforward (accepting that dockerhub is one of the more slightly complicated services).

yspreen:
Are you suggesting I'd try and give a PR a shot?

chris48s:
I was confirming its a sensible feature request (sometimes people suggest badges we couldn't implement e.g: because there's nowhere we can get the data) and ready to work on. If you want to submit a PR that would be great.. or someone else can work on it if they want.

If you're interested but finding the dockerhub code a bit daunting, here's a few quick pointers

* We have a tutorial aimed at beginners at https://github.com/badges/shields/blob/master/doc/TUTORIAL.md
* The docker version badge is probably a sensible basis to start from
* A really simple example of a version badge would be something like this https://github.com/badges/shields/blob/master/services/bintray/bintray-version.service.js
* The additional complicating factors for docker are:
    * You can optionally specify an architecture (we default to `amd64`)
    * We can order versions by date or semver (you could probably ignore this for your case. It sounds like only date makes sense?)
    * The `/v2/repositories/` endpoint is paginated, so we might need to make >1 api call (hence `getMultiPageData`)

the-maux:
ok

PR #6051, opened 2021-01-14
Add [DockerDate] service

yspreen:
This addresses #6016

Currently, paged data is not used at all. I don't quite get why the paged data is only used in the transform method when tag is undefined, and then only to match digests? Maybe someone can enlighten me here.

### DIFF
--- a/services/docker/docker-date.service.js
+++ b/services/docker/docker-date.service.js
@@ -1,0 +1,138 @@
+'use strict'
+
+const Joi = require('joi')
+const { nonNegativeInteger } = require('../validators')
+const { BaseJsonService, NotFound } = require('..')
+const { formatRelativeDate } = require('../text-formatters')
+const {
+  buildDockerUrl,
+  getDockerHubUser,
+  getMultiPageData,
+} = require('./docker-helpers')
+
+const buildSchema = Joi.object({
+  count: nonNegativeInteger.required(),
+  results: Joi.array().items(
+    Joi.object({
+      last_updated: Joi.date().iso().required(),
+      name: Joi.string().required(),
+      images: Joi.array().items(
+        Joi.object({
+          digest: Joi.string(),
+          architecture: Joi.string().required(),
+        })
+      ),
+    })
+  ),
+}).required()
+
+const queryParamSchema = Joi.object({
+  arch: Joi.string()
+    // Valid architecture values: https://golang.org/doc/install/source#environment (GOARCH)
+    .valid(
+      'amd64',
+      'arm',
+      'arm64',
+      's390x',
+      '386',
+      'ppc64',
+      'ppc64le',
+      'wasm',
+      'mips',
+      'mipsle',
+      'mips64',
+      'mips64le'
+    )
+    .default('amd64'),
+}).required()
+
+module.exports = class DockerDate extends BaseJsonService {
+  static category = 'version'
+  static route = { ...buildDockerUrl('date', true), queryParamSchema }
+  static examples = [
+    {
+      title: 'Docker Image Date (latest)',
+      pattern: ':user/:repo',
+      namedParams: { user: '_', repo: 'alpine' },
+      queryParams: { arch: 'amd64' },
+      staticPreview: this.render({
+        version: new Date(new Date() - 2 * 24 * 60 * 60 * 1000),
+      }),
+    },
+    {
+      title: 'Docker Image Date (tag)',
+      pattern: ':user/:repo/:tag',
+      namedParams: { user: '_', repo: 'alpine', tag: '3.6' },
+      staticPreview: this.render({
+        version: new Date(new Date() - 3 * 24 * 60 * 60 * 1000),
+      }),
+    },
+  ]
+
+  static defaultBadgeData = { label: 'latest docker image', color: 'blue' }
+
+  static render({ version }) {
+    return {
+      label: undefined,
+      message: formatRelativeDate(version),
+      color: 'blue',
+    }
+  }
+
+  async fetch({ user, repo, page }) {
+    page = page ? `&page=${page}` : ''
+    return this._requestJson({
+      schema: buildSchema,
+      url: `https://registry.hub.docker.com/v2/repositories/${getDockerHubUser(
+        user
+      )}/${repo}/tags?page_size=100&ordering=last_updated${page}`,
+      errorMessages: { 404: 'repository or tag not found' },
+    })
+  }
+
+  transform({ tag, data, pagedData, arch = 'amd64' }) {
+    let version
+    if (!tag) {
+      version = data.results[0]
+      return { version: version.last_updated }
+    } else {
+      version = data.find(d => d.name === tag)
+      if (!version) {
+        throw new NotFound({ prettyMessage: 'tag not found' })
+      }
+      return { version: version.last_updated }
+    }
+  }
+
+  async handle({ user, repo, tag }, { arch }) {
+    let data, pagedData
+
+    if (!tag) {
+      data = await this.fetch({ user, repo })
+      if (data.count === 0) {
+        throw new NotFound({ prettyMessage: 'repository not found' })
+      }
+      if (data.results[0].name === 'latest') {
+        pagedData = await getMultiPageData({
+          user,
+          repo,
+          fetch: this.fetch.bind(this),
+        })
+      }
+    } else {
+      data = await getMultiPageData({
+        user,
+        repo,
+        fetch: this.fetch.bind(this),
+      })
+    }
+
+    const { version } = await this.transform({
+      tag,
+      data,
+      pagedData,
+      arch,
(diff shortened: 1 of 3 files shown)


calebcartwright:
Thanks for the PR! Side note, I've tweaked the PR title so that CI will run the corresponding service tests. It's currently failing since no tests have been added yet, but once you do they will get picked up automatically.

> Currently, paged data is not used at all. I don't quite get why the paged data is only used in the transform method when tag is undefined, and then only to match digests? Maybe someone can enlighten me here.

Likely more than you really wanted to know, but to paint a full picture... :smile: 

When the Shields servers receive a request for a badge, they have to interact with upstream services/platforms/etc. to get the underlying raw data which is then used in determining attributes of the badge, like text and color. Most of the time, the Shields code that does this interaction ends up being really concise and simple (for example: https://github.com/badges/shields/blob/master/services/ansible/ansible-quality.service.js) since it's largely just making an API call and snagging some attributes out of the result.

At the same time, there's also a few rigid constraints we have to work with. One is that within GitHub, the place badges are most often used, there's a fixed window with a max of ~3-4 seconds for images to load, otherwise they won't render (so the entire request/response chain between the user, us, and upstream endpoints has to happen quickly). Another is that Shields.io works with a large volume of badge requests, which means that we in turn throw a lot of traffic at upstream services. This means that it's exceptionally important for us to be cognizant of how we interact with those upstream platforms, both to prevent sending unnecessary requests their way, and to minimize risks of our badge servers get throttled/rate limited.

In the case of the some of the existing DockerHub badges (notably size and version), the nature of the available API pushes us on both of those constraints, and that in turn required a tradeoff that increased complexity of the code in order to ensure the badges were returned quickly (even though in many cases we have to make several separate API calls to get the data set) while also avoiding making requests to DockerHub when we could. Additionally, the total size of the data set those badges, and the associated CPU-bound work to process it is significantly higher than the typical case within Shields, so there's additional optimizations that try to short circuit wherever possible.

For example, in the case of a size badge with an explicit tag, we can skip the paginated API calls and processing because we can get what we need with one API call and grab the size directly from the response. Similarly, in the case of a version badge where our user has requested a badge that displays the most recent version by date/calendar, then we're able to skip the multi-page volley of API calls, except in the subcase where the most recently updated tag is literally `latest`, in which case we have to use the corresponding digest to map back to the actual versioned tag (this is so we can show a meaningful badge like  instead of the largely meaningless )

It's not uncommon for version badges to be a bit more complex than the others, and that's often because they have to take into account sort order (most recent version by date/calendar or by some versioning scheme like Semver) as Chris mentioned on the other thread. That complexity is just heavily exacerbated by the nature of the DockerHub API.

My initial instinct is that that complexity could likely be avoided for this badge, since it's literally about dates/calendars. The only wrinkle to consider is whether you'd want to support the tag-based filtering that those other badges support, so that this badge would be able to show the last_updated value for any v3* tag _and_ v4* tag

For example with the version badges:
https://img.shields.io/docker/v/_/alpine/3


https://img.shields.io/docker/v/_/alpine/3.6

yspreen:
Really appreciate the thorough response!

But I'm still puzzled by the fact that out of so many scenarios, only the digest matching for the latest tag seems to use the paged data.

Anyway. For the date service, it's pretty similar. It would only need paging if there's a named tag specified and it hasn't been updated in a long time, while other tags were changed much more recently. I should probably add paged fetches for that specific scenario.

yspreen:
I think it's in a good state now. Only the arch parameter might be a good addition. It's ignored right now. Perhaps it could be used to filter for tags that have an image for the specified arch.

calebcartwright:
> But I'm still puzzled by the fact that out of so many scenarios, only the digest matching for the latest tag seems to use the paged data.

I can understand the confusion when looking at it solely in the context of the `transform` function, but you've got to look back at the full flow, including how the `handle` function calls the `transform` function. In the case of semver sorting, there's no way to shortcircuit, because the upstream API sorts lexically, not via semver. So while in the case of date-based versioning, the transform has to take into account the short circuit and potentially being able to use a single-tag dataset or the entire dataset, with semver it's exclusively the full set.

I suppose this would probably be more clear if `handle` set the multi page results to `pagedData` in the semver case instead of `data` and `transform` used paged data instead of data in that branch. It's not that the paged data isn't being used in other cases, just different args being passed and corresponding param vals used

yspreen:
Yeah, I figured that out now, just took a bit of looking at the code again. Feel free to add some feedback about the newest commits and current implementation 👌

services/docker/docker-date.service.js
```diff
@@ -0,0 +1,129 @@
...
+    } else {
+      version = data.find(d => d.name === tag)
+    }
+    if (!version) {
+      throw new NotFound({ prettyMessage: 'tag not found' })
+    }
+    return { version: version.last_updated }
+  }
+
+  async handle({ user, repo, tag }, { arch }) {
```
chris48s, line 106: As far as I can tell, this architecture param is not implemented. i.e: if I call this with `?arch=amd64` then.. nothing changes.

Just thinking about this param, there are some properties of a dockerhub repo that are architecture independent. For example number of stars. For versions, that's not the case - we need an arch. Last updated seems slightly ambiguous in this respect. We can give "when was the last date something (anything) changed", but also "when was the last date the arm64 image changed" might be different to "when was the last date the mips image changed". Maybe we drop arch, show the last date anything changed and then add an arch param later if people want it (but the default is essentially "all" instead of amd64)? Anyone got any thoughts on what makes the most sense here?

Either way, if we're going to have the param make it actually do something. If it doesn't do anything, bin it.

services/docker/docker-date.service.js
```diff
@@ -0,0 +1,129 @@
...
+      }),
+    },
+  ]
+
+  static defaultBadgeData = { label: 'latest docker image', color: 'blue' }
+
+  static render({ version }) {
+    return {
+      label: undefined,
+      message: formatRelativeDate(version),
```
chris48s, line 77: This variable isn't really a version. Its a date, so its confusing to call the var `version`. How about `last_updated`?

services/docker/docker-date.service.js
```diff
@@ -0,0 +1,129 @@
...
+      'ppc64',
+      'ppc64le',
+      'wasm',
+      'mips',
+      'mipsle',
+      'mips64',
+      'mips64le'
+    )
+    .default('amd64'),
+}).required()
```
chris48s, line 47: It looks like both these schemas are exact copy & pastes from the version service. If we want to use the exact same validation in multiple places, can we move it to `docker-helpers.js` and have `docker-date.js` and `docker-version.js` both import it from there.

Lets work out what we need first though.. if we bin the arch param, that changes things..

services/text-formatters.js
```diff
@@ -119,6 +119,9 @@ function formatDate(d) {
 }
 
 function formatRelativeDate(timestamp) {
+  if (timestamp instanceof Date) {
+    timestamp = timestamp.getTime() / 1000
+  }
```
chris48s, line 124: I think it would actually be clearer/less "magic" if we did something like

```js
function formatRelativeDate(date) {
  return formatRelativeTimestamp(date.getTime() / 1000)
}

function formatRelativeTimestamp(timestamp) {
  return moment()
    .to(moment.unix(parseInt(timestamp, 10)))
    .toLowerCase()
}
```

and switch `date.service.js` (the only other thing using this helper) to use `formatRelativeTimestamp()`

services/docker/docker-date.service.js
```diff
@@ -0,0 +1,129 @@
...
+      staticPreview: this.render({
+        version: new Date(new Date() - 2 * 24 * 60 * 60 * 1000),
+      }),
+    },
+    {
+      title: 'Docker Image Date (tag)',
+      pattern: ':user/:repo/:tag',
+      namedParams: { user: '_', repo: 'alpine', tag: '3.6' },
+      staticPreview: this.render({
+        version: new Date(new Date() - 3 * 24 * 60 * 60 * 1000),
```
chris48s, line 67: Any significance to this?

PyvesB:
@yspreen any chance you could circle back to this and take a look at @chris48s' review comments? It would be great to get this piece of work landed. 😉

yspreen:
Agreed...

It's gotten a bit hectic recently, so I haven't found the time. But I'll add it to my todos. Hopefully I'll get to it soon

PyvesB:

If someone else wants to pick this up, you can simply fetch the commits by running `git fetch origin pull/6051/head:pr-6051`. Address the pending review comments and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

yspreen:
Fair, thanks for the update @PyvesB 

I hope I'll get around to it someday, it's been super busy around here

PR #6052, opened 2021-01-14
Update self-hosting.md documentation for running with Prometheus

Cobertos:
There's two environment variables that need to be set in order to use Prometheus and the /metrics endpoint.


calebcartwright:
(closed/reopened to un-stick CI)

doc/self-hosting.md
```diff
@@ -198,10 +198,11 @@ sudo node server
 ### Prometheus
 
 Shields uses [prom-client](https://github.com/siimon/prom-client) to provide [default metrics](https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors). These metrics are disabled by default.
-You can enable them by `METRICS_PROMETHEUS_ENABLED` environment variable.
+You can enable them by `METRICS_PROMETHEUS_ENABLED` and `METRICS_PROMETHEUS_ENDPOINT_ENABLED` environment variables.
 
 ```bash
-METRICS_PROMETHEUS_ENABLED=true npm start
+METRICS_PROMETHEUS_ENABLED=true
+METRICS_PROMETHEUS_ENDPOINT_ENABLED=true
```
calebcartwright, line 205: Let's keep the script invocation with this, both for clarity and for consistency with other settings above:

```suggestion
METRICS_PROMETHEUS_ENABLED=true METRICS_PROMETHEUS_ENDPOINT_ENABLED=true npm start
```

Cobertos: Ah, didn't realize the intention was as a runnable command. Thought it was just a list of the vars

calebcartwright: > Might be more clear the use the console syntax highlighting and add a $ to denote it as a terminal prompt

Yup that's a fair point. There seems to be quite a variety of identifiers used on the code fences within this doc, but using console (or sh I believe) would be better than using bash. Will defer such changes for now, but if you're interested at all in a follow up PR that'd be most welcome!

calebcartwright (approved):

PR #6067, opened 2021-01-16
remove erroneous clearRequestCache call, run [DynamicJson]

calebcartwright:
This popped up in #6054 and is a follow up to #6037 which removed the in memory cache, but missed this one.

Also observable in the daily tests, where we've had 1915 out of 1926 of the service tests failing consistently for several days now. Just making note of it here to draw attention to the point I'd made in https://github.com/badges/shields/pull/5966#issue-542676424


calebcartwright:
down to some ~25 or so service test failures, which seems to be driven by some of the usual suspects (David, LibrariesIo, etc.) so compared to the 1,915 failing on master currently I think this resolves the issue

image

paulmelnikow (approved):

Issue #6049, opened 2021-01-13
Publish both snapshots and releases to DockerHub


carlspring:
:clipboard: **Description**

At the moment, only "snapshots" are being pushed to DockerHub. It is unclear whether these are nightly builds, whether they're done after a pull request, or manually when decided/needed. The community can clearly benefit from these.

The problem with the currently deployed version is that it's simply labelled as `next` which isn't an actual version and can easily be wiped next time a new version is pushed, making it a silent snapshot release. This is unstable and unreliable for use.

The documentation shows you how you can build your own Docker image from the `Dockerfile`, but it would be much better to have official releases for this. There should be a dedicated page on Docker and how to use the image.

Also, the currently available `shieldsio/shields:next` image, places the code under `/usr/src/app`. This is not really a conventional place to put an appliction and is actually mean for source code for various applications.

calebcartwright:
Thanks for reaching out!

The current versioning approach is deliberate: it always reflects latest from trunk branch, which has all changes validated via our CI and PR/merge processes. The version of the Shields server in the image on DockerHub is basically identical to what's running on Shields.io service (with a couple days lag on average). However, we've been having discussions in the ops team meetings about potentially changing this to something like CalVer.

Are there any specific problems that you've encountered, or are you just offering some general feedback and suggestions?

I think it's important to note for context that we're a small team and are focused on the development and support of the Shields.io service itself. The vast, _vast_ majority of our users are consuming the Shields.io service and comparatively the self-hosted Shields server is not really a core use case nor one that's formally supported in any capacity at this time. That's largely because self-hosting is really only necessary in a handful of edgier cases where a user needs badges for a private service/tool (or private project), or from services/tools that are not internet facing (e.g. private corporate intranet).

We still try to enable that uncommon self-hosting case by having a couple straightforward ways to get up and running simply and quickly, because it tends to be a one-and-done action. For example, a typical self-hoster just wants to get setup so they can have badges for their private GitHub repos or from their Jira server; they neither care about nor need the random assortment of new badges for new services that get added every month.

[...]

Issue #5866, opened 2020-11-23
Automate Calendar Versioning/Changelog for self-hosting users


chris48s:
:clipboard: **Description**

In the last ops meeting we discussed a plan to create tagged releases and produce a changelog for self-hosting users with minimal overhead. Here's the plan:

- We label pull requests affecting self-hosting users (or any changes that we think are worthwhile to flag in a changelog for self-hosting users) with the "self-hosting" label. Maybe we need to put a tiny bit more thought in here but this shouldn't require any substantial change to our current process. We are pretty good at labelling PRs correctly.
- We write a Github Action that runs on a schedule https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events to run once a month. The action:
    - Grabs a list of any PRs since the last "server" tag
    - Filters out any with the self-hosting tag
    - Writes the PR titles and links to a CHANGELOG.md. If there's nothing in the last month, we write a generic message "changes to internals" or something
    - Opens a PR proposing the edits to CHANGELOG.md. The PR is labelled with a "release" label
- A human reviews the PR. This allows us to sanity-check the changelog entries (and edit it manually if we really want to). It also allows us to control when the release happens. For example, if there is a big change affecting self-hosting users in the late stages of review we might choose to hold back a release a few days until after it is merged.
- We have a second action which detects when a PR with the label "release" is merged, tags the commit and pushes the tag. The tag can be something like "server-YYYY-MM-DD" (We need a prefix to differentiate server releases from package releases). We might have to write this ourselves but there may be an off-the-shelf solution in the GH Actions marketplace we can use for this.

This plan would mostly automate the process of creating tags and producing a good-enough changelog (certainly better than what we have now) while affording us a manual review step.

chris48s:
Two bits of follow up to have a look at here:
https://github.com/badges/shields/pull/6132#issuecomment-771076613

PR #6069, opened 2021-01-16
add a github action to create tags from labelled PR

chris48s:
This is one of the bits we need for issue #5866
Also refs #6049

When we merge a PR which is labelled with "release", this will create a tag called `server-YYYY-MM-DD` (for today's date). To show this working (and test it), I set up a demo repo using the action:

- https://github.com/chris48s/release-action-test
- https://github.com/chris48s/release-action-test/pull/1
- https://github.com/chris48s/release-action-test/pull/2
- https://github.com/chris48s/release-action-test/actions
- https://github.com/chris48s/release-action-test/releases

I have also created a release label for use with this: https://github.com/badges/shields/labels?q=release

### DIFF
--- a/.github/workflows/tag-release.yml
+++ b/.github/workflows/tag-release.yml
@@ -1,0 +1,31 @@
+name: Tag Release
+
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  tag-release:
+    if: |
+      github.event_name == 'pull_request' &&
+      github.event.action == 'closed' &&
+      github.event.pull_request.merged == true &&
+      contains(github.event.pull_request.labels.*.name, 'release')
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get current date
+        id: date
+        run: echo "::set-output name=date::$(date --rfc-3339=date)"
+
+      - name: Checkout branch "master"
+        uses: actions/checkout@v2
+        with:
+          ref: 'master'
+
+      - name: Tag Release
+        uses: tvdias/github-tagger@v0.0.2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          tag: server-${{ steps.date.outputs.date }}


.github/workflows/tag-release.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Checkout branch "master"
+        uses: actions/checkout@v2
+        with:
+          ref: 'master'
+
+      - name: Tag Release
+        uses: tvdias/github-tagger@v0.0.2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          tag: server-${{ steps.date.outputs.date }}
```
calebcartwright, line 31: What all capabilities does this action perform? Is it just creating and pushing the tag, does it create a release from the tag (is that even something we'd want)?

Don't have any concerns with this, largely asking out of curiosity, especially thinking about what all we may want to do in part 2

chris48s: It literally just creates a tag, not releases: https://github.com/tvdias/github-tagger#commit-tagger

If we want releases, we could look at something more elaborate like https://github.com/actions/create-release
The reason I have picked this is because I'm working on the assumption that we will just create tags and maintain the changelog as a markdown file (the same workflow as we have for badge-maker).

In general, I think github releases are most useful when:
- You want to associate artefacts with a release (e.g: .exe file, .deb package, etc).
  In our case, we don't need to do this.
- You want to use the release notes as the changelog.
  In our case, the workflow we've discussed is that a bot assembles a proposed changelog and then a human reviews it (and edits it if they want). This lends itself more to maintaining the changelog in a markdown file because the changelog entry can be a PR. There isn't really a way for a bot to "PR" release notes. To use the release notes as the changelog, we'd have to either take whatever is auto-generated, or make it a manual process.

If you're worried about relying on a third-party action for this, I could look at creating the tag just using shell commands instead. Using a third-party action probably isn't really buying us very much here.

calebcartwright: > If you're worried about relying on a third-party action for this,

Nope, don't have any particular concerns there, especially since AIUI it's just using the virtual environment token and not one from any of our accounts.

> The reason I have picked this is because I'm working on the assumption that we will just create tags and maintain the changelog as a markdown file (the same workflow as we have for badge-maker)

Works for me :+1:

calebcartwright (approved):

PR #6073, opened 2021-01-18
Start a changelog, create a tag

chris48s:
Refs #5866 #6069

The reason why I'm submitting this is that for the next part of #5866 I am going to write an action which

- Looks at what changed since our last `server-YY-MM-DD` tag
- Generates a changelog entry

This will be easier to develop if we can assume there is already a changelog file we can append to and the repo already has one `server-YYYY-MM-DD` tag (hence the release label). This PR just creates that baseline to work from.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+Note: this changelog is for the shields.io server. The changelog for the badge-maker NPM package is at https://github.com/badges/shields/blob/master/badge-maker/CHANGELOG.md
+
+---
+
 ## server-2021-01-18
 
 - Gotta start somewhere
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Changelog
+
+## server-2021-01-18
+
+- Gotta start somewhere


calebcartwright:
Do you think there's a chance that having a top level changelog file could confuse a reasonable user, relative to changes between Shields.io vs. tag/Docker vs. package, or will the explicit tag subheadings in the log suffice?

I think it's probably fine (though i suppose it wouldn't hurt to have a section somewhere that states something to the effect of: "this is server, npm package can be found here <<with link>>) but wanted to pose the question anyway

chris48s:
I think we already cover this in the first paragraph of the README but I've added a sentence clarifying.

calebcartwright:
> I think we already cover this in the first paragraph of the README but I've added a sentence clarifying.

Fair enough! I hadn't really thought about the authorship-side of a changelog in a monorepo (or at least multi-artifact repo) before so not really sure what's common practice

calebcartwright (approved):

chris48s:
Well that's gone.. not well
https://github.com/badges/shields/runs/1723610807?check_suite_focus=true
I'll see if I can work out what the difference is between this repo and the test repo

Issue #5940, opened 2020-12-11
Feature Request: Display queried URL and API responses in the pop-up badge preview pages


HariSekhon:
:clipboard: **Description**

Hi,

First off I'd like to say I love your work to date, thank you so much, this project is great! :)

I'd like to request that the shields.io website extend the preview pop-up pages to display the full API URL with parameters that shields.io is querying on the remote service, as well as the full raw returned API output (pretty-printed if JSON 🙏🏼) in order to debug errors or unknown statuses that may be returned by the 3rd party API.

This will make it easier for users to figure out if the badge is broken because they entered incorrect parameters, or whether there is some issue with their build or even a new or unrecognized API result from on the 3rd party system.

This will also make it easier for people to submit patches for API results that aren't currently understood or have changed without having to be experts in the code base or reproducing what it is doing as it'll be obvious from the UI.

Thanks very much

Hari

calebcartwright:
Hello @HariSekhon :wave: Thanks for reaching out with your idea, and for sharing those kind words! 

However, I've got to say that I'm pretty strongly opposed to this idea. The Shields.io website is intended for end user consumption, not as development/troubleshooting portal. Even if there weren't substantial implementation issues, I wouldn't want to make that end user experience noisy/confusing by trying to also do the latter.

> I'd like to request that the shields.io website extend the preview pop-up pages to display the full API URL with parameters that shields.io is querying on the remote service, as well as the full raw returned API output (pretty-printed if JSON 🙏🏼)

This just isn't practical. The frontend site really isn't any different than pasting the badge url into a web browser or a user embedding the badge url into a readme file. It uses the same img.shields.io/.... url to show the badge previews, and has no more insight about the internal implementation of the corresponding backend badge service than any other consumer.

In theory we could massively extend our existing badge server capabilities to expose inner implementation details over separate REST endpoints, but this just feels wrong and unnecessary. It's fairly common in today's world for an API/endpoint to communicate with other endpoints as part of serving a request, but I've never encountered a case of a service providing API endpoints about such internal dependencies and external interactions.

Such a hypothetical feature set would also increase load on our backend, in cases potentially doubling the traffic we have to throw at the upstream services, and would add undo burden and maintenance overhead on the codebase.

We also have services that have to utilize multiple upstream API endpoints in order to fetch all the requisite data required for the badges, and we also have cases where a badge that feels singular to the end user can actually hit entirely different endpoints depending on small changes to the route (like a `sort` query parameter).

[...]

PR #6075, opened 2021-01-19
improve debug output for API calls with query strings

chris48s:
I was trying to debug an issue earlier and spotted this issue - if we're making an API call with a query string, we don't display it properly in the logging output so we have to add debug statements to the code and assemble it by hand. I think I've actually hit this before, but today it annoyed me enough to actually fix it :)

Before:

```
...
 Request  🏹 
 https://repo.magicmq.dev/service/rest/v1/search
 Response status code  🎯 
 200
...
```

After:

```
...
 Request  🏹 
 https://repo.magicmq.dev/service/rest/v1/search?group=dev.magicmq&name=itemapi&sort=version&prerelease=false
{
  "headers": {
    "Accept": "application/json"
  }
}
 Response status code  🎯 
 200
...
```

### DIFF
--- a/core/base-service/base.js
+++ b/core/base-service/base.js
@@ -213,7 +213,18 @@ class BaseService {
 
   async _request({ url, options = {}, errorMessages = {} }) {
     const logTrace = (...args) => trace.logTrace('fetch', ...args)
-    logTrace(emojic.bowAndArrow, 'Request', url, '\n', options)
+    let logUrl = url
+    const logOptions = Object.assign({}, options)
+    if ('qs' in options) {
+      const params = new URLSearchParams(options.qs)
+      logUrl = `${url}?${params.toString()}`
+      delete logOptions.qs
+    }
+    logTrace(
+      emojic.bowAndArrow,
+      'Request',
+      `${logUrl}\n${JSON.stringify(logOptions, null, 2)}`
+    )
     const { res, buffer } = await this._requestFetcher(url, options)
     await this._meterResponse(res, buffer)
     logTrace(emojic.dart, 'Response status code', res.statusCode)
--- a/core/base-service/base.spec.js
+++ b/core/base-service/base.spec.js
@@ -463,9 +463,7 @@ describe('BaseService', function () {
         'fetch',
         sinon.match.string,
         'Request',
-        url,
-        '\n',
-        options
+        `${url}\n${JSON.stringify(options, null, 2)}`
       )
       expect(trace.logTrace).to.be.calledWithMatch(
         'fetch',


calebcartwright:
I feel foolish :laughing: Not because of the code changes, but because I don't remember ever seeing this output anywhere, and think I am about to have a big "TIL" moment...

is the printed options in the updated output from the request or response headers?

calebcartwright:
Trying to think through whether this may have been intentional to avoid logging any sensitive info in either the query (tokens as query params for certain services) and/or in request headers. Think I vaguely remember something along these lines from a while back but unsure

paulmelnikow:
I've run into the same issue a few times, including recently when doing a little paid support to troubleshoot self-hosted services.

It's possible security was a reason for leaving these out. I think it's fine to log the query string, which in most cases is already buried in the output anyway. As far as headers, since those are sometimes sensitive, I wonder if we should place them under a second flag, like TRACE_SERVICES_INSECURE.

Thanks so much for picking this up. It's a good improvement 🍻

chris48s:
> I feel foolish :laughing: Not because of the code changes, but because I don't remember ever seeing this output anywhere, and think I am about to have a big "TIL" moment...

Run `npm run badge https://img.shields.io/appveyor/build/gruntjs/grunt` it will be a voyage of discovery :open_mouth:

> is the printed options in the updated output from the request or response headers?

This is the request headers

> Trying to think through whether this may have been intentional to avoid logging any sensitive info

> It's possible security was a reason for leaving these out

Yeah reasonable question. If you look at the original line of code removed in this PR:

```diff
- logTrace(emojic.bowAndArrow, 'Request', url, '\n', options)
```

my reading of that is we always intended to present the entire `options` object in the log output but we either changed the signature of `logTrace()` at some point or maybe this used to just be a straight `console.log()`. We broke it accidentally. So I don't think we ever _intended_ to not show `options`.

In terms of whether we need to: Where do we actually surface this info? Do we expose it in other places/situations than `npm run badge`? (that's the only place I ever use it). If that is the only place we show it, I think security is a non-issue there: If you can type `npm run badge` then you can type `printenv` or `cat config/local.yml` although we would need to review it if we were to re-purpose the trace outputs for something like https://github.com/badges/shields/issues/5940 - then we would definitely need to revisit that assumption. If we expose it in other contexts, let me know.

If we do decide we need to hide credentials, we could strip or obfuscate `Authorization` and basic auth user/pass headers but leave the rest instead of hiding all headers.

paulmelnikow:
> In terms of whether we need to: Where do we actually surface this info? Do we expose it in other places/situations than `npm run badge`? (that's the only place I ever use it). If that is the only place we show it, I think security is a non-issue there: If you can type `npm run badge` then you can type `printenv` or `cat config/local.yml` although we would need to review it if we were to re-purpose the trace outputs for something like #5940 - then we would definitely need to revisit that assumption. If we expose it in other contexts, let me know.

This makes sense to me. I'm convinced it's fine to print the credentials.

The only other place I could see showing this is if we create a debug endpoint, however if we ever decided to do this we could revisit it at that time.

calebcartwright:
> Where do we actually surface this info? Do we expose it in other places/situations than npm run badge

I wasn't sure if any of this would make it's way into the Heroku logs, though after thinking some more even if it did, it's a pretty rare case where we accept a user's token as a query param and even then we only allow narrowly scoped, read-only tokens

calebcartwright (approved):

chris48s:
> I wasn't sure if any of this would make it's way into the Heroku logs, though after thinking some more even if it did, it's a pretty rare case where we accept a user's token as a query param and even then we only allow narrowly scoped, read-only tokens

I think we need to clarify things a bit as there are two "requests" and two "logs" being discussed and we're mixing and-and-matching.

There's the request to shields e.g: https://img.shields.io/appveyor/build/gruntjs/grunt
And then there's the request (or requests) to the upstream API e.g: https://ci.appveyor.com/api/projects/gruntjs/grunt

The thing we are "logging" in this `logTrace()` statement is the second request - our request to the upstream API. This means that if we send credentials upstream in the request (either query string or headers) we will `console.log()` them here. Those might be from our env vars. Those might be passed via the query string to the first request. Either way, if you have an interactive shell you can read them anyway. These statements don't end up in the heroku logs, but also bear in mind that 100% of the people who can read the heroku logs can also read (and change) our env vars in heroku too.

The thing you see in the heroku logs are the first type of request - the inbound request to shields.io. Any credentials the user passes in that query string do get logged there and retained for a _very_ short amount of time (heroku only lets us see the last 1000 log entries and we process >10 million per day, so that usually covers less than 1 minute of traffic), but those credentials (as you've already mentioned) should also be safe for the user to expose in a (public) README. Either way, This PR doesn't change anything to do with the logging (to the console) or logging (in heroku) of those requests. The second type of request (to upstream APIs) don't appear in the heroku logs.

The other thing that might be confusing the issue here is I gave the example `npm run badge https://img.shields.io/appveyor/build/gruntjs/grunt` further up, but if you run that on your local copy no data is being sent to or received from shields.io. Everything is being done locally - it is actually replacing `https://img.shields.io` with `http://localhost:8080`.
https://github.com/badges/shields/blob/02692fe810ee91ee7da48c996f89fd1185ec7632/scripts/badge-cli.js#L24
That may have also been..non-obvious in retrospect.

calebcartwright:
> I think we need to clarify things a bit as there are two "requests" and two "logs" being discussed and we're mixing and-and-matching.

Good call :+1: Sorry, I had this initial abstract thought bubble of "oh this looks neat, let's do it as long as it doesn't result in anything showing up anywhere it shouldn't/running afoul of no-tracking promise", which was manifested as a couple disjointed posts without actually digging into it.

I agree this is a good improvement, and that this doesn't result in any undesired log/exposure of information/etc. that I had initially been pondering. Thanks for the detailed write up as well!

PR #6076, opened 2021-01-19
[Bitbucket] Server Adding Auth Tokens and Resolving Pull Request api …

danielle:
[Bitbucket Server] Adding Auth Tokens and Resolving Pull Request Api issue.  Added the bitbucket_server_username & bitbucket_server_password variables to the /core/server/server.js file.  Also changed the user variable in the /services/bitbucket/bitbucket/bitbucket-pull-request.service.js the Api documentation [(Atlassian Bitbucket Server api Pull Request Documentation)](https://docs.atlassian.com/bitbucket-server/rest/5.16.0/bitbucket-rest.html#idm8287391664) defines the parameter as requiring the project name that the repository is contained in not the user name.

Thank you @calebcartwright and @paulmelnikow

Please let me know you any changes are needed.


core/server/server.js
```diff
@@ -15,7 +15,10 @@ const GithubConstellation = require('../../services/github/github-constellation'
 const suggest = require('../../services/suggest')
 const { loadServiceClasses } = require('../base-service/loader')
 const { makeSend } = require('../base-service/legacy-result-sender')
-const { handleRequest } = require('../base-service/legacy-request-handler')
+const {
+  handleRequest,
+  clearRequestCache,
+} = require('../base-service/legacy-request-handler')
```
calebcartwright, line 21: Could you do a rebase? Seems like this is missing 2d9ac83bb4fc6e19500aca62c376ed7ceba1a6f0, and we wouldn't want to add these back and re-introduce the issue

core/server/server.js
```diff
@@ -482,6 +487,7 @@ class Server {
   static resetGlobalState() {
     // This state should be migrated to instance state. When possible, do not add new
     // global state.
+    clearRequestCache()
```
calebcartwright, line 490: This would be problematic but will go away after a rebase :+1:

core/server/server.js
calebcartwright: This will also be needed to fix the service tests

services/bitbucket/bitbucket-pull-request.service.js
```diff
@@ -29,23 +29,23 @@ function pullRequestClassGenerator(raw) {
     static category = 'issue-tracking'
     static route = {
       base: `bitbucket/${routePrefix}`,
-      pattern: `:user/:repo`,
+      pattern: `:project/:repo`,
```
calebcartwright, line 32: While I can understand the thinking behind changing these from the Bitbucket Server/DC lens, I'm mildly opposed to changing it now given that we ended up going with a shared class and route to handle both Bitbucket Cloud and Server/DC.

There's other places where there's multiple viable nouns, but we always end up having to pick one (e.g. `:user` vs `:organization` for GitHub). At the end of the day the actual name doesn't matter all that much so long as it's not something outlandish, like `:somethingOutlandish` because our service is targeted at developers/technologists and we generally get to assume that folks can connect the minor dots like these.

It's not so much that I care about the actual value of the name, but want to avoid getting into a place where we have folks pushing to change these back and forth between the cloud/SaaS and self-hosted/on-prem version of tools like these.

calebcartwright (commented):
Thanks for the PR! This will need a rebase to fix some test issues, but the core change looks to be good :+1: 

cc @badges/shields-core in case anyone else has thoughts about toggling the noun names in the route

services/bitbucket/bitbucket-pull-request.service.js
paulmelnikow: > It's not so much that I care about the actual value of the name, but want to avoid getting into a place where we have folks pushing to change these back and forth between the cloud/SaaS and self-hosted/on-prem version of tools like these.

I think the thing that was confusing about the existing setup is that in BB Server, there are users, but users can't own projects, so it's the project that owns the repo.

I agree that we should not just switch from the cloud/SaaS to the self-hosted/on-prem names, though wonder if it would make sense to parameterize the shared code so it can generate different routes for the cloud/SaaS and self-hosted/on-prem, particularly if that can be done without too much hassle, with the goal of making things clearer to users (in the frontend). Though I appreciate what you're saying about hoping folks can connect the minor dots.

paulmelnikow: I guess if you agree @calebcartwright I'd be happy to take a stab at that (or to work with @danielle on same).

core/server/server.js
paulmelnikow: @danielle since we've been chatting about this offline, I'll mention it's fine to use rebase here, or to merge master into this branch. Then add another commit, if necessary, to reset this file.

services/bitbucket/bitbucket-pull-request.service.js
calebcartwright: we're discussing offline now actually, considering splitting the difference with a multiword name (e.g. `:projectOrUser`) since the former would better resonate with Server/DC while the latter more so with Cloud/SaaS.

I can't think of an example off hand but feel like we've done something similar in other routes before

calebcartwright: > I guess if you agree @calebcartwright I'd be happy to take a stab at that (or to work with @danielle on same).

sorry missed this below the other comment. i don't have any objections to providing more contextual names where we can, especially since it's not uncommon for their to be divergence between self-hosted and SaaS tool offerings in cases, just want to avoid having several extra class definitions where we can.

The other prominent case (though not a particularly addressable one) is on GitHub itself where a repo is owned by a user or an organization and we always use `:user`

paulmelnikow: > The other prominent case (though not a particularly addressable one) is on GitHub itself where a repo is owned by a user or an organization and we always use `:user`

The thing I could imagine doing there is adding a description that can appear in the UI in these rare cases.

calebcartwright: Even on Cloud,  Projects seem to be visible these days though I find it a bit puzzling. What we need for our API call is still the owner/user/account and the repo name (AFAICT), not really sure how common it still is for default/unnanmed workspace and projects or the degree of possible confusion for Cloud users :man_shrugging:  

https://img.shields.io/bitbucket/pr-raw/guyzmo/python-bitbucket


https://bitbucket.org/guyzmo/python-bitbucket
image

danielle:
Updated to projectOrUser per my discussion with @calebcartwright

danielle:
@calebcartwright / @paulmelnikow let me know if anything else is needed.

services/bitbucket/bitbucket-pull-request.service.js
paulmelnikow: @danielle did you see this discussion?

danielle: Yes, are you stating that it should not be a compound name such as "projectOrUser" as requested but instead keep it as "project"?  My original pull request included "project" which matched the API documentation but I wasn't aware that there was a separate call for Bitbucket vs Bitbucket Server.

calebcartwright: I'm backtracking on my original thinking now. I feel like anything that indicates "project" as an option would cause issues for Cloud users, as we need the "owner" and the "repo" name, but if a Cloud user saw `:projectOrUser` and tried to use their Cloud project name the badge would not work.

I think the best option, if possible, would be to find a clean way to present different param names for the two routes to account for Server vs Cloud differences as Paul suggested in https://github.com/badges/shields/pull/6076#discussion_r560587197

Otherwise I think we'd be better off to try to find a generic singular term, and then just add help text that shows in the badge builder window

paulmelnikow: So is the idea that we use `project` for Server and `user` for Cloud? (Or is it `owner` for Cloud?)

calebcartwright: In my experience `project` is what will most clearly resonate for `Server`. I've never been a user of BB Cloud so hard to say, though may just be worth sticking with `user` since that's what we have there now

calebcartwright:
@danielle - how about this, let's split out the changes to `server.js` that add the missing Bitbucket keys to the schema from the param/name changes to the Bitbucket classes. The former is a definitive bug that needs to be fixed no matter what, which we could go ahead and land pretty easily, whereas the latter has a lot more nuance and considerations that may take a while.

There's no need to couple these together and hold up the config schema fix, so splitting the proposed changes will be the best way to move forward on that.

danielle:
> @danielle - how about this, let's split out the changes to `server.js` that add the missing Bitbucket keys to the schema from the param/name changes to the Bitbucket classes. The former is a definitive bug that needs to be fixed no matter what, which we could go ahead and land pretty easily, whereas the latter has a lot more nuance and considerations that may take a while.
> 
> There's no need to couple these together and hold up the config schema fix, so splitting the proposed changes will be the best way to move forward on that.

Sure, let me reverse the changes and push only the server.js updates on this pull request

danielle:
@calebcartwright, Done!  

Question:  On the 'core/server/server.js' file, I also don't see the Bitbucket Cloud tokens, should those also be added?

calebcartwright (approved):
Thanks!

calebcartwright:
>  I also don't see the Bitbucket Cloud tokens, should those also be added?

Nope, we don't support that yet. This was observed when the Bitbucket Server support was first added, but the only use case for Cloud would be someone that wanted badges for private repos on Cloud, and we haven't had any requests for that

Issue #6070, opened 2021-01-17
Badge Request: GitHub Total Commits Badge


YashTotale:
:clipboard: **Description**

The badge is for displaying the total amount of commits to a GitHub repository.

Example: ![Example Badge](https://img.shields.io/badge/commits-101-brightgreen)

:link: **Data**

There is no clear way to get the total number of commits to a repository.

One way is to get all the contributors to a repository using the GitHub API and add the `contributions` attribute of each to get the total number of commits. Here is an [example](https://api.github.com/repos/YashTotale/YashTotale.github.io/contributors). This is not a robust solution, however.

This [article](https://blog.notfoss.com/posts/get-total-number-of-commits-for-a-repository-using-the-github-api/) details a more robust solution, however, it is a "hackish" solution.

:microphone: **Motivation**

There are badges for weekly, monthly, and yearly commits, however, there is none for total commits. This badge would be useful in helping users get a better idea of the project's dependability.

A large number of commits generally correlates with a more robust and maintained project, improving its credibility. Conversely, newer projects that are more likely to undergo breaking changes will have a small number of commits (even if they have a high amount of weekly/monthly commits as they are being developed). Therefore, this badge would give users a better idea of a project's dependability than weekly/monthly commit counts.

chris48s:
This can be done with the github V4 API e.g:

```gql
query {
  repository(owner:"badges", name:"shields") {
    object(expression:"master") {
      ... on Commit {
        history {
          totalCount
        }
      }
    }
  }
}
```

I'm not 100% convinced this is a massively useful thing to put on a badge given it is shown pretty prominently on the landing page of a GH repo

![Screenshot at 2021-01-17 13-18-48](https://user-images.githubusercontent.com/6025893/104844212-c5c93980-58c6-11eb-8d91-816533f7ba4e.png)

but if someone wants to work on it, lets go with the graphql query :)

calebcartwright:
Agree with Chris here. I don't have any objections to including the badge, especially since we have other commit related badges already, but I definitely question the use case and rationale in the description.

Specifically, an all-time commit count really should not be viewed as the measure of project health and maintenance status. Case in point, compare the total commit count of the retired/unmaintained [request](https://github.com/request/request) repo against the similar [got](https://github.com/sindresorhus/got) repo. The way repositories manage commits is also highly variant, with many higher profile projects employing strategies that actually reduce count (squash on merge, requiring contributors to squash PR commits before rebasing, etc.).

[...]

PR #6081, opened 2021-01-22
Add [GitHubTotalCommits] Badge

YashTotale:
Adds the GitHub Total Commits Badge discussed in #6070


calebcartwright:
> I'm running into this error every time I try running npm start:

You can ignore the token pool error, or just add a GH_TOKEN to your local env: https://github.com/badges/shields/blob/master/doc/server-secrets.md#github (this is typically a good idea anyway if you're developing badges for GitHub)

As for the rest, looks like you've got a syntax error which is causing errors in Heroku, tests, and presumably your local env as well:
```bash
/tmp/build_0e08ee14_/services/github/github-total-commits.service.js:87

    const json = await this.fetch({ user, repo, branch: branch ?? 'HEAD' })

                                                                ^

SyntaxError: Unexpected token '?'
```

Try just adding 'HEAD' as a default value for the param above on the function signature

YashTotale:
The tests that are failing seem to be failing for the `master` branch as well. The [daily service tests](https://app.circleci.com/pipelines/github/badges/daily-tests) show this. 
Please let me know if I have anything left to add or fix.

calebcartwright:
Have adjusted the title so that only the tests for this service are executed (as opposed to the entire GitHub suite).

From the perspective of the end user experience, I'm not convinced that it makes sense to expose this as a separate badge/separate route. The existing GitHub Commit Activity badge presents users with the option to choose the interval, and I feel like this is just another interval (total/all time), much like the various download badges that allow users to select between the all time download count or count during interval period.

Under the hood, I know the target implementation for this is quite different than the existing commit badges, in no small part because the former is leveraging the newer v4 API option while the latter are still using the older v3 APIs. However, I don't we want those internals to leak through and drive the user experience and add a separate route.

I would rather see the existing commit activity service class refactored to take fully take advantage of the graphql/v4 approach. If it seems like it would be too much to refactor that and add support for total/all time interval in a single PR, then that's alright, but I'd still prefer to see the total interval incorporated on the existing badge route by adding some conditionals to that class to use different fetch/transform logic when the interval is `t` (for total/all time)

PyvesB:
I agree with @calebcartwright 's reasoning. @YashTotale is this something you would be keen on working on? :)

YashTotale:
@PyvesB, I would love to work on this, but probably won't be able to get to it for a while. If anyone else wants to give it a shot, feel free to do so. :)

PyvesB:
Thanks for the update. Will close this for now then. We can always reopen if need-be. 😉

Issue #6042, opened 2021-01-09
"staticPreview.link" is not allowed


rajat19:
Are you experiencing an issue with...

- [ ] [shields.io](https://shields.io/#/)
- [x] My own instance
- [ ] [badge-maker NPM package](https://www.npmjs.com/package/badge-maker)

:beetle: **Description**
While working on https://github.com/badges/shields/issues/2271
I came across an issue 
`[ValidationError]: Example for GithubCommitActivity at index 0 "staticPreview.link" is not allowed`

The issue is generated if someone adds link in response to a service's render function and code for static Preview for that service uses `this.render`
example
```javascript
staticPreview: this.render({ interval: 'm', commitCount: 457, user: 'badges', repo: 'shields' })
```

example for github commit activity, If i change my function to 
```javascript
static render({ user, repo, interval, commitCount }) {
    const intervalLabel = {
      y: '/year',
      m: '/month',
      '4w': '/four weeks',
      w: '/week',
    }[interval]
    const slug = `${encodeURIComponent(user)}/${encodeURIComponent(repo)}`

    return {
      message: `${metric(commitCount)}${intervalLabel}`,
      link: [
        `https://github.com/${slug}`,
        `https://github.com/${slug}/graphs/code-frequency`,
      ]
    }
  }
```

Apparently found an earlier todo for the same https://github.com/badges/shields/blob/7ac57d993a7e776519c37ee8d321ef72653194c9/services/github/github-stars.service.js#L29-L30

Has there been any work on this todo ?
For now should i directly provide object (without link key) in `staticPreview` wherever I would be adding link in render's response ?

PyvesB:
`staticPreview` is no longer a thing in our examples nowadays. Closing as this is no longer relevant.

Issue #2271, opened 2018-11-05
Add links for all github badges in social template


xPaw:
For example the contributors badge in social template does not make the buttons clickable, while the stars badge does.

paulmelnikow:
Just a tip on timing, that it might be a little easier to do this post-refactor.

Once #2253 is merged all the github services will be easy to convert.

rajat19:
I have started working on this issue

As there seems to be a lot of badges for github

I will gradually include remaining badge types, add links for all the respective badges and simultaneously update below

Please update if you find any links that are wrong

- [ ] Activity
- [ ] Build
- [ ] Size
- [ ] Issue Tracking
- [ ] Downloads
- [ ] Other
- [ ] Social
- [ ] Version
- [ ] Analysis
- [ ] License
- [ ] Platform & Version Support
- [ ] Dependencies

Badge (Type: Activity)|Badge path|Link (example for badges/shields)
---|---|---
All Contributors|`/github/all-contributors`|https://github.com/badges/shields/graphs/contributors
Commit Activity|`/github/commit-activity`|https://github.com/badges/shields/graphs/commit-activity
Commit Since Tagged Version|`/github/commit-since`|https://github.com/badges/shields/compare/a0663d8da53fb712472c02665e6ff7547ba945b7...HEAD
Commit Since Tagged Version (Branch)|`/github/commit-since`|https://github.com/badges/shields/compare/a0663d8da53fb712472c02665e6ff7547ba945b7...historical
Commit Since Latest Release (By Date)|`/github/commit-since`|https://github.com/badges/shields/compare/latest...HEAD
Commit Since Latest Release (By Date) for a branch|`/github/commit-since`|https://github.com/badges/shields/compare/latest...historical
Commit Since Latest Release (By Date including pre-releases)|`/github/commit-since`|https://github.com/badges/shields/compare/latest...HEAD
Commit Since Latest Release (By SemVer)|`/github/commit-since`|https://github.com/badges/shields/compare/latest...HEAD
Commit Since Latest Release (By SemVer incl. pre-releases)|`/github/commit-since`|https://github.com/badges/shields/compare/latest...HEAD
Contributors|`/github/contributors`|https://github.com/badges/shields/graphs/contributors
Last Commit|`/github/last-commit`|https://github.com/badges/shields/commits/master
Last Commit (Branch) |`/github/last-commit`|https://github.com/badges/shields/commits/gh-pages
[...]

PR #6096, opened 2021-01-22
Add links for [Github] Activity Badges

rajat19:
Part of solving https://github.com/badges/shields/issues/2271

Currently Covers all badges included in activity section which includes following

Badge (Type: Activity)|Badge path|Link (example for badges/shields)
---|---|---
All Contributors|`/github/all-contributors`|https://github.com/badges/shields/graphs/contributors
Commit Activity|`/github/commit-activity`|https://github.com/badges/shields/graphs/commit-activity
Commit Since Tagged Version|`/github/commit-since`|https://github.com/badges/shields/compare/a0663d8da53fb712472c02665e6ff7547ba945b7...HEAD
Commit Since Tagged Version (Branch)|`/github/commit-since`|https://github.com/badges/shields/compare/a0663d8da53fb712472c02665e6ff7547ba945b7...historical
Commit Since Latest Release (By Date)|`/github/commit-since`|https://github.com/badges/shields/compare/latest...HEAD
Commit Since Latest Release (By Date) for a branch|`/github/commit-since`|https://github.com/badges/shields/compare/latest...historical
Commit Since Latest Release (By Date including pre-releases)|`/github/commit-since`|https://github.com/badges/shields/compare/latest...HEAD
Commit Since Latest Release (By SemVer)|`/github/commit-since`|https://github.com/badges/shields/compare/latest...HEAD
Commit Since Latest Release (By SemVer incl. pre-releases)|`/github/commit-since`|https://github.com/badges/shields/compare/latest...HEAD
Contributors|`/github/contributors`|https://github.com/badges/shields/graphs/contributors
Last Commit|`/github/last-commit`|https://github.com/badges/shields/commits/master
Last Commit (Branch) |`/github/last-commit`|https://github.com/badges/shields/commits/gh-pages
Release Date|`/github/release-date`|https://github.com/badges/shields/releases
Pre-release Date|`/github/release-date-pre`|https://github.com/badges/shields/releases

Got issue https://github.com/badges/shields/issues/6042 while working on this


calebcartwright:
Several of these badges are not in the social category, and as I recall we only permit (by practice) setting links for badges in the social category because the social badge renderer is the only one that will actually do it.

@paulmelnikow @chris48s - you're both far more familiar with the renderers than I, any concerns/issues/etc. with adding default links to non-social badges in the service class definitions?

rajat19:
@calebcartwright I have asked the same question on the issue https://github.com/badges/shields/issues/2271
15 days back but didn't get any reply on that
Because of which I raised PR as it is

chris48s:
Yeah we have tended to only do this for social badges, but we say in the guidelines

> badges can link to a third-party website providing more information, either related to the metadata provided by the badge or about the project

https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md#guidelines

I did a quick search and we do currently have two non-social badges that include a link by default:

* https://github.com/badges/shields/blob/fcf6678a127c9679b0d68284b860181c2580fe26/services/github/github-milestone.service.js#L68
* https://github.com/badges/shields/blob/fcf6678a127c9679b0d68284b860181c2580fe26/services/dependabot/dependabot.service.js#L61

There's no issue with rendering a link on a non-social badge per-se (e.g: https://img.shields.io/dependabot/semver/bundler/puma ) but it is an uncommonly used/not very useful feature because almost everyone embeds badges with an `` tag not an `<object>` tag. I guess that's my concern with adding links by default to loads of other badges. It is a bit of an unfortunate feature in that respect because it seems like it would be useful but doesn't work how most users expect it too and generally creates confusion rather than utility. That said, some really high traffic badges use it so I think we're stuck with it forever.

Even if we do add link attrs by default to a bunch of other badges, I think I'm not in favour of allowing link in the frontend previews on the home page, otherwise the frontpage of https://shields.io/ turns into a really random bag of links out to other disparate stuff.

Sorry. That's not really a decision..

calebcartwright:
> otherwise the frontpage of https://shields.io/ turns into a really random bag of links out to other disparate stuff.

Is this in reference to the badge listings?

calebcartwright:
> Basically, I wouldn't like to see all of these badges be links out to external websites:

I've got some concerns about plugging in all these links too, though aren't all the badge listings configured to just pop up the badge builder? I didn't think any of them would navigate out to an external site instead of launching the builder window (but if they would I definitely share your concern)

I think my two main concerns are:

the side effects of #6042 (I could've sworn we had an existing issue for this, but can't find one. Maybe because it's just a well-known behavior and considered more of an expectation than a bug). If we move to a position where we have entirely separate (or even somewhat disjointed) examples and render functions, as would be required today to move forward with these changes, then I feel like we're going to greatly increase the surface for potential drift between the badge listings and the actual badges, and/or add annoying little maintenance paper cuts (since tweaks to the actual rendering logic would no longer necessarily be automatically reflected in the listings)

and this :point_down: especially since markdown/GitHub are where are badges used most, and the embedded links just don't work
> but it is an uncommonly used/not very useful feature because almost everyone embeds badges with an  tag not an <object> tag. I guess that's my concern with adding links by default to loads of other badges. It is a bit of an unfortunate feature in that respect because it seems like it would be useful but doesn't work how most users expect it too >

chris48s:
> I didn't think any of them would navigate out to an external site instead of launching the builder window

Yes - they currently launch the builder modal. I'm just saying whatever we do re #6042 should not change that - they should continue to launch the builder modal.

calebcartwright:
> I'm just saying whatever we do re #6042 should not change that - they should continue to launch the builder modal.

Gotcha, and agreed :+1:

rajat19:
I totally agree that static previews should not have links in them.

@chris48s @calebcartwright what's your final decision regarding adding links to badges other than the ones present in social  template (not in static previews but otherwise)?
should i continue working on this ? or should i add just the social ones and reject this PR ?

calebcartwright:
> @chris48s @calebcartwright what's your final decision regarding adding links to badges other than the ones present in social template (not in static previews but otherwise)?

Somewhat related aside - I updated the PR title to follow our conventions so that the associated tests would be run, and these changes definitely seem to be causing some non-specious test failures. We'd definitely need to have CI passing for any of the changes to be potentially mergeable.

As for the changes themselves.. I don't think we'll be able to proceed in the immediately foreseeable future, but that doesn't necessarily mean we won't ever accept some or even all of these changes. There's been some potential concerns raised, and there's also a few decisions that we'll need to make first given the potential impacts those decisions could have here.

For the time being I think folks that want clickable badges would want to follow the current recommendation of utilizing the corresponding markdown/markup/etc. syntax based on the context in which they're using the badges.

chris48s:
Does anyone else from the core team have a view on this one? @paulmelnikow ?  @PyvesB ?

PyvesB:
> For the time being I think folks that want clickable badges would want to follow the current recommendation of utilizing the corresponding markdown/markup/etc. syntax based on the context in which they're using the badges.

I would tend to agree with this statement. It would be nice to convey this on the documentation of `link` on the homepage.

As a side note, I'm a bit annoyed that we've ended with somewhat inconsistent practices. Is there really no good reason why we've been adding default links to social badges? I'm guessing we probably have to live with it at this point, but we could add a note to our guidelines and change the two deviant non-social badges.

chris48s:
The twitter and github fork/star/watch badges have had default links for a really really long time (since 2015) and are quite widely used which kind of set that precedent for default links on social badges.

PyvesB:
I updated our [contributing guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md) a few weeks ago to clarify that only badges which have a default style equal to `social` should have links on by default. Accordingly, I'm going to go ahead and close this.

Thanks for opening this @rajat19, even if it didn't get merged, it ultimately resulted in a useful discussion and an improvement of our guidelines. 👍🏻

PR #6098, opened 2021-01-23
Docs.rs badge, run [Docsrs]

### LINKED ISSUE #1382: Badge request: docs.rs
See https://docs.rs/about

coolshaurya:
Resolves #1382


calebcartwright:
Thanks for this! Let us know if you have any questions

services/docsrs/docsrs.service.js
```diff
@@ -0,0 +1,53 @@
+'use strict'
+
+const Joi = require('joi')
+const { BaseJsonService } = require('..')
+
+const schema = Joi.object({
+  build_status: Joi.boolean().required(),
+}).required()
```
calebcartwright, line 8: The API returns an array of objects that contain the `build_status` field, not a single object, so this schema will need to be adjusted accordingly. 

Maybe try something like this?

```suggestion
const schema = Joi.array()
  .items(
    Joi.object({
      build_status: Joi.boolean().required(),
    })
  )
  .min(1)
  .required()
```

services/docsrs/docsrs.service.js
```diff
@@ -0,0 +1,53 @@
...
+        color: 'success',
+      }
+    } else {
+      return {
+        message: `${version} | failed`,
+        color: 'critical',
+        isError: true,
+      }
+    }
+  }
```
calebcartwright, line 37: I think we'll need to go in a different direction in terms of the badge content. One of the core goals of Shields is to provide consistency across our badge offering, and sometimes that means our badges look different than the badges provided directly by the upstream services.

Specifically, the badge message would need to reflect the documentation build status, like our other build status badges do (`passing` or `failing`). If there's a desire to include the crate version as a data point within the badge, then that should go in the label following our standard conventions. I know this will result in a different looking badge than what Docs.rs natively provides, but it would be much more consistent with our badges.

For example, crates download count for a specific version
https://img.shields.io/crates/dv/regex/1.4.1


vs. latest
https://img.shields.io/crates/dv/regex


vs. total
https://img.shields.io/crates/d/regex


@paulmelnikow - you'd suggested some alternative badge content in the linked issue, any thoughts?

services/docsrs/docsrs.service.js
```diff
@@ -0,0 +1,53 @@
...
+    return await this._requestJson({
+      schema,
+      url: `https://docs.rs/crate/${crate}/${version}/builds.json`,
+    })
+  }
+
+  async handle({ crate, version }) {
+    if (version === undefined) {
+      version = 'latest'
+    }
```
calebcartwright, line 49: This could be done a bit more succinctly with a default value on `version`:

```suggestion
  async handle({ crate, version = 'latest' }) {
```

services/docsrs/docsrs.test.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+
```
calebcartwright, line 2: This file should be named `docsrs.tester.js` in order for the test runner to pick it up

services/docsrs/docsrs.service.js
```diff
@@ -0,0 +1,53 @@
...
+    const { build_status: buildStatus } = await this.fetch({ crate, version })
```
calebcartwright, line 50: Note this would need to be updated accordingly to account for the response being an array

calebcartwright (commented):
Thanks again! I think the core of this is looking good, but will likely need to adjust the final badge contents. More detail inline below, along with a few other smaller items

services/docsrs/docsrs.service.js
```diff
@@ -0,0 +1,59 @@
...
+      keywords: ['rust'],
+    },
+  ]
+
+  static defaultBadgeData = {}
+
+  static render({ buildStatus, version }) {
+    if (buildStatus) {
+      return {
+        label: `docs@${version}`,
```
calebcartwright, line 32: I think we'll probably want to eschew this in cases where `version` is `latest`, similar to how we don't have  because the absence of a specific version implicitly conveys latest/current status.

services/docsrs/docsrs.service.js
```diff
@@ -0,0 +1,59 @@
...
+        message: 'passing',
+        color: 'success',
+      }
+    } else {
+      return {
+        label: `docs@${version}`,
+        message: 'failing',
+        color: 'critical',
+        isError: true,
```
calebcartwright, line 41: Let's remove this, as it's used to reflect an error within Shields/processing/etc., not that the upstream build failed.

```suggestion
```

services/docsrs/docsrs.tester.js
```diff
@@ -0,0 +1,11 @@
...
+
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('Passing docs')
+  .get('/tokio/0.3.0.json')
+  .expectBadge({ label: 'docs@0.3.0', message: 'passing' })
+
+t.create('Failing docs')
+  .get('/tensorflow/0.16.1.json')
+  .expectBadge({ label: 'docs@0.16.1', message: 'failing' })
```
calebcartwright, line 11: For these service tests (which can roughly be viewed as high level e2e tests/integration tests with upstream APIs), we want to be sure that we exercise the different routes/inputs for our badge, and that the resultant badge is of the correct "shape"

So here, we'd want to have a test that does not provide an explicit version to provide verification that the default/latest badges are operating correctly.

We also typically use Joi schemas as validators to ensure that the message is _an_ expected value (from a set of known/possible values) as opposed to a literal specific value, especially since in most cases that literal value will be dynamic as things change upstream. In this particular case, I think it's fine to keep these literal tests in here since the doc build states for those specific crate versions is immutable, and we also have to manually directly map an the upstream response value (boolean build_status) to the corresponding message whereas in most other badges we're using the standard helpers (`renderBuildStatusBadge`, `isBuildStatus`, etc.)

However, for the additional test that needs to be added to cover the latest/no-specific-version route, you'll want to use Joi so that the test won't be brittle:

```js
messge: Joi.allow('passing', 'failing'),
```

services/docsrs/docsrs.service.js
```diff
@@ -0,0 +1,59 @@
...
+  static examples = [
+    {
+      title: 'docs.rs',
+      namedParams: { crate: 'regex', version: 'latest' },
+      staticPreview: this.render({ version: 'latest', buildStatus: true }),
+      keywords: ['rust'],
+    },
+  ]
+
+  static defaultBadgeData = {}
```
calebcartwright, line 27: Let's keep the default value in here, as that will still be used in error scenarios (like not found errors on unknown crates/crate versions)

```suggestion
  static defaultBadgeData = { label: 'docs' }
```

calebcartwright (commented):
Thanks for the updates, this is looking good! Few final minor items, then should be good to merge :+1:

calebcartwright (approved):
Awesome, thank you!

PR #6101, opened 2021-01-23
Fix [feedz] service in case the registrations page gets paginated

jakubfijalkowski:
:wave: 

[Last time](https://github.com/badges/shields/pull/5753) I missed one thing - the `items` in the registration response (at the first level) represent _pages_, and if you have enough versions, you will get multiple entries there.

This PR fixes that - it no longer expects a single page and it is now able to get versions from all the pages.

### DIFF
--- a/services/feedz/feedz.service.js
+++ b/services/feedz/feedz.service.js
@@ -22,7 +22,6 @@ const schema = Joi.object({
         ),
       }).required()
     )
-    .max(1)
     .default([]),
 }).required()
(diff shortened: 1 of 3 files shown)


chris48s (approved):

chris48s:
sorry it took a few days for anyone to get round to this one

jakubfijalkowski:
@chris48s No problem, really! :) Thanks for merging!

PR #6102, opened 2021-01-24
add a GH action to generate a changelog and submit a release PR

chris48s:
Refs #5866

This PR adds an action which will

* run once a month (but also allow us to trigger it manually)
* generate a changelog we can manually review
* open a PR tagged with "release" which we can review/merge

Coupled with https://github.com/badges/shields/blob/master/.github/workflows/tag-release.yml merging the PR will create the release tag.

This will generate a PR something like https://github.com/chris48s/release-action-test/pull/5 . The commit messages on this test repo are mostly junk from when I was debugging the action - ours will look like.

```md
- add a github action to create tags from labelled PR (#6069)
- fix [wordpress] 'Can't generate version color for false' (#6072)
- Update self-hosting.md documentation for running with Prometheus (#6052)
- fix: remove erroneous clearRequestCache call (#6067)
```

..or whatever. I'll leave some more notes inline.

### DIFF
--- a/.github/actions/draft-release/entrypoint.sh
+++ b/.github/actions/draft-release/entrypoint.sh
@@ -34,8 +34,7 @@ sed -i "${INSERT_POINT} r temp-changes.txt" CHANGELOG.md
 rm temp-changes.txt
 
 # Run prettier (to ensure the markdown file doesn't fail CI)
-npm ci
-npm run prettier
+npx prettier@$(cat package.json | jq -r .devDependencies.prettier) --write "CHANGELOG.md"
 
 # Generate a unique branch name
 BRANCH_NAME="$RELEASE_NAME"-$(uuidgen | head -c 8)
--- a/.github/actions/draft-release/entrypoint.sh
+++ b/.github/actions/draft-release/entrypoint.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+# Set up a git user
+git config user.name "release[bot]"
+git config user.email "actions@users.noreply.github.com"
+
+# Find last server-YYYY-MM-DD tag
+git fetch --unshallow --tags
+LAST_TAG=$(git tag | grep server | tail -n 1)
+
+# Find the marker in CHANGELOG.md
+INSERT_POINT=$(grep -n "^\-\-\-$" CHANGELOG.md | cut -f1 -d:)
+INSERT_POINT=$((INSERT_POINT+1))
+
+# Generate a release name
+RELEASE_NAME="server-$(date --rfc-3339=date)"
+
+# Assemble changelog entry
+rm -f temp-changes.txt
+touch temp-changes.txt
+{
+    echo "## $RELEASE_NAME"
+    echo ""
+    git log "$LAST_TAG"..HEAD --no-merges --oneline --pretty="format:- %s" --perl-regexp --author='^((?!dependabot).*)$'
+    echo $'\n'
+} >> temp-changes.txt
+
+# Write the changelog
+sed -i "${INSERT_POINT} r temp-changes.txt" CHANGELOG.md
+
+# Cleanup
+rm temp-changes.txt
+
+# Run prettier (to ensure the markdown file doesn't fail CI)
+npm ci
+npm run prettier
+
+# Generate a unique branch name
+BRANCH_NAME="$RELEASE_NAME"-$(uuidgen | head -c 8)
+git checkout -b "$BRANCH_NAME"
+
+# Commit + push changelog
+git add CHANGELOG.md
+git commit -m "Update Changelog"
+git push origin "$BRANCH_NAME"
+
+# Submit a PR
+TITLE="Changelog for Release $RELEASE_NAME"
+PR_RESP=$(curl https://api.github.com/repos/"$REPO_NAME"/pulls \
+    -X POST \
+    -H "Authorization: token $GITHUB_TOKEN" \
+    --data '{"title": "'"$TITLE"'", "body": "'"$TITLE"'", "head": "'"$BRANCH_NAME"'", "base": "master"}')
+
+# Add the 'release' label to the PR
+PR_API_URL=$(echo "$PR_RESP" | jq -r ._links.issue.href)
+curl "$PR_API_URL" \
+    -X POST \
+    -H "Authorization: token $GITHUB_TOKEN" \
+    --data '{"labels":["release"]}'
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -1,0 +1,19 @@
+name: Draft Release
+on:
+  schedule:
+    - cron: '0 1 1 * *'
+    # At 01:00 on the first day of every month
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Draft Release
+        uses: ./.github/actions/draft-release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO_NAME: ${{ github.repository }}
--- a/.github/actions/draft-release/Dockerfile
+++ b/.github/actions/draft-release/Dockerfile
@@ -1,0 +1,8 @@
+FROM node:12-buster
+
+RUN apt-get update
+RUN apt-get install -y jq
+RUN apt-get install -y uuid-runtime
+COPY entrypoint.sh /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
--- a/.github/actions/draft-release/action.yml
+++ b/.github/actions/draft-release/action.yml
@@ -1,0 +1,5 @@
+name: 'draft-release'
+description: 'Generate a changelog and propose a release PR'
+runs:
+  using: 'docker'
+  image: 'Dockerfile'


.github/actions/draft-release/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+# Generate a release name
+RELEASE_NAME="server-$(date --rfc-3339=date)"
+
+# Assemble changelog entry
+rm -f temp-changes.txt
+touch temp-changes.txt
+{
+    echo "## $RELEASE_NAME"
+    echo ""
+    git log "$LAST_TAG"..HEAD --no-merges --oneline --pretty="format:- %s" --perl-regexp --author='^((?!dependabot).*)$'
```
chris48s, line 26: Our original proposal (and the one I wrote in #5866 ) was to grab all the PRs merged since the last release tag with the self-hosting label.
It turns out this is actually a bit of a fiddly thing to try and work out via the GH API (REST or GraphQL) as really the query we are trying to make is something like: "Find all pull requests where the commit we generated by squash-merging the PR is a child of commit X (where commit X is the last tagged commit)". It is certainly not impossible to work it out, but in the interests of keeping things fairly simple I'm going to propose we start off with this, which is every commit since the last release tag except for those where the author is dependabot. Starting with that and removing anything irrelevant at the manual review step seems like a good place to start and this has the advantage that it is easy to generate :) I suggest we try this and once we have manually reviewed a few we will have an idea of whether we want to revise it a bit or try and do something more sophisticated.

.github/workflows/draft-release.yml
```diff
@@ -0,0 +1,19 @@
+name: Draft Release
+on:
+  schedule:
+    - cron: '0 1 1 * *'
+    # At 01:00 on the first day of every month
+  workflow_dispatch:
```
chris48s, line 6: Having written this and thought about how we're going to actually use it, there's a few related issues giving me pause for thought here which are sort of tied together (at least in my mind).

* I think it is useful for the release tag to include the full date.
* We're mostly going to do one release a month. Occasionally (I'm thinking of the situation where there is a security fix) it will be useful to do an out-of-sequence release.
* We generate the title for the changelog entry using the current date when we generate the PR here https://github.com/badges/shields/pull/6102/files#diff-06ec58a42aa9b0bdf423c39dc6b326395f69db72e0127e3e1d3f83b9c87d27f4R18 and then we merge it, we generate the tag name again using the current date here https://github.com/badges/shields/blob/2b3ad63748bf643d61d16721892c9f1795ce0ab6/.github/workflows/tag-release.yml#L20 . If we review/merge the PR the day it is submitted, this is fine, but:
  * If we leave it until another day the title we generated on the changelog entry is now different than the tag name we will generate when we merge it. We need to manually edit it.
  * In fact, if we merge any other PRs after the changelog PR is opened then the suggested changelog entry we generated is now out of date. We may need to manually edit it.

I guess what we ideally want is like a `dependabot rebase` command but in lieu of that I think the solution I've arrived at is this.. the action runs once a month automatically on a schedule, but it can also be triggered manually if we want (workflow_dispatch basically gives us a 'run workflow' button).

Screenshot at 2021-01-24 19-27-22

Then that gives us a reminder every month, but if we ignore it for a bit and need to re-generate the PR (or run the same process mid-month) we can do. Also because we generate a unique branch name we can run the workflow multiple times on the same day if needed. I think that gives us a reasonable balance of those constraints. Anyone else got thoughts on the trade-offs here? Any ideas on how we could deal with it better or make this more fool-proof?

calebcartwright: That's a good observation. My initial reaction to the above comment https://github.com/badges/shields/pull/6102/files#r563348469 was that it sounded perfectly reasonable, though I guess that is part of a minor wrinkle here in that the odds of the log entries vs commits getting out of sync (seems like a much lower chance that there'd be self-hosted labeled PR in between).

I almost wonder if it might be easier to remove the PR and potential human delay from this, and just let the robots do their thing fully automated (permit auto merging/squashing or even push directly). We'd always have the ability to edit the changelog after, and even in the the case mentioned in #5866 about a big pending self-hosted change landing shortly after, we could always kick off manually.

> Any ideas on how we could deal with it better or make this more fool-proof?

I guess another thing we could do is add another action that runs daily, checks for these release PRs, and squash merges if it finds one open and approved, or if there's one open but not approved then closes the PR and re-runs the workflow so that a new one is generated with the latest?

That obviously won't solve any deltas in intra-day commits, but could potentially help prevent the tag and entry from getting out of sync

.github/actions/draft-release/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+} >> temp-changes.txt
+
+# Write the changelog
+sed -i "${INSERT_POINT} r temp-changes.txt" CHANGELOG.md
+
+# Cleanup
+rm temp-changes.txt
+
+# Run prettier (to ensure the markdown file doesn't fail CI)
+npm ci
```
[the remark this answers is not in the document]
chris48s: I want to make sure we're running the same version of prettier that we will run in CI (not whatever the latest version on NPM is), but we can `jq` the version out of `package.json` and so something like: `npx prettier@$(cat package.json | jq -r .devDependencies.prettier)`. I'll apply that change next time I update the PR.

.github/actions/draft-release/entrypoint.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
```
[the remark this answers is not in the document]
chris48s: Yeah it is a good question. Tbh I deliberated a bit over how best to do this and I'm not 100% confident this is the ideal approach, but I'll try to explain the tradeoffs that lead me to here:

One of the real pain points for me with mediumly complicated github actions like this is that I want to work on/test them locally as much as possible. The more I can "change file, run the script locally, see what happens" (as opposed to "change file, push to test repo, wait for the action to spin up, attempt to debug from action output") the better because it speeds up the development process. One of the things I've found with JS actions is they end up very tied to the actions API/remote repo which makes this difficult. One of the coping mechanisms I've tried employing for this (on JS actions) is trying to push as much of the logic out into utility functions which can be tested independently and keep the bits that are tied to the actions API as lean as possible but I haven't really found a workflow I love for this yet. I guess for really big action codebases it becomes viable to TDD it and mock all the external calls but it is more than I really want to do for a release script. Maybe this is partly that I just haven't found the right workflow for it. I am getting into actions but it is still early days. Anyway, meanwhile back at the point.. the nice thing about doing this as a bash script is I could develop and test almost all of this on a local checkout of the repo without having to push anything.. which made the feedback loop quicker. I suspect that will probably be a bit easier to debug in future, but I also accept we sacrifice some readability.

> I'd guess the combinations of interacting with git and not having to pull in additional parsing/process deps for the file processing?

Yes I think there are some things that are easier using shell commands. For example, I think something like `git log "$LAST_TAG"..HEAD --no-merges --oneline --pretty="format:- %s" --perl-regexp --author='^((?!dependabot).*)$'` would be quite a bit more code using the github API and javascript. That said, there are some things here that are clearly more cumbersome in bash, like templating out a JSON body to post with `curl`, so.. we probably win some and lose some. I think we're just about in the realms of sanity though.

Tbh, if we decide we want to look at identifying PRs by label or something I'd probably rewrite it in javascript. I think the pain of trying to assemble GraphQL queries in bash and extract more than a single value from the results would outweigh the pain of having to push to test :D

.github/workflows/draft-release.yml
chris48s: I wouldn't be totally against just letting the bots do their thing (I for one welcome our new release-bot overlords).

One thing I don't really have a handle on yet is how many non-dependabot commits we generate in a typical month and how many of them are just completely irrelevant to self-hosting users (i.e: what's the signal-to-noise ratio of just dumping this to the log with no manual review). I suspect we won't figure this out until we try it for a bit, or maybe I should go and do some research. We have plenty of history to dive into.

While it is true that we can always edit the changelog later, I think as far as possible it is useful if the changelog entry in the tagged ref is final if possible. i.e: if I browse the repo at https://github.com/badges/shields/tree/server-2021-01-18 it would be nice if the changelog file has a useful entry at that commit, rather than being "raw" there and then tidied in a later commit.

calebcartwright: > (I for one welcome our new release-bot overlords).

:laughing: I keep catching myself trying to use them in non-applicable places.. like asking Dependabot to rebase a humans PR or adding a squash label in a project that doesn't have Ranger

> I suspect we won't figure this out until we try it for a bit, or maybe I should go and do some research. We have plenty of history to dive into

Agreed. I've a slight lean towards the bot overlords but no strong opinions. Think it'd be best to just give this a try and learn/adjust as needed

.github/actions/draft-release/entrypoint.sh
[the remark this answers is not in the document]
calebcartwright: Ahhhh yeah that makes sense :+1:

.github/actions/draft-release/entrypoint.sh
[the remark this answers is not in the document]
calebcartwright: > the nice thing about doing this as a bash script is I could develop and test almost all of this on a local checkout of the repo without having to push anything.. which made the feedback loop quicker. I suspect that will probably be a bit easier to debug in future, but I also accept we sacrifice some readability

Makes sense. We're not in a position where the majority of the maintainers are on Windows or anything, and from what I've seen with my work machine, WSL (particularly 2.0) would be more than sufficient in any weird edge cases where someone needed to work on it from Windows world

calebcartwright (approved):

calebcartwright (approved):

chris48s:
I will go ahead and merge this. It should run next on Monday so we will see how it works out. The other thing I've done is I've set up a rule in DockerHub which _should_ create a dockerhub tag when we push a `server-` tag (not tested yet):

Screenshot at 2021-01-28 19-25-45

I think it is probably worth ironing out the process a bit before we do any kind of announcement/comms about releases. Tbh, there is a non-zero chance some part of this rube goldberg machine won't work as expected when we test it all at together for the first time..

PR #6103, opened 2021-01-25
convert approve-bot to local action

chris48s:
When I originally wrote this, I put it in its own repo. Having run it for a bit, I think it is just a bit annoying having this live in a separate repo. The logic is fairly specific to our dependencies/dependabot setup, I'm not using it on any other projects, and it means I have to be the deal-with-automerge-problems-guy. This PR converts it into a local action.


calebcartwright:
> Lots 'o changes. Skipping diff-based checks.

I forgot we did this :laughing: I was unaware that custom actions needed to have the node_modules directory checked in. Is there any downside in including this in-repo, and/or any benefits in having it under the Badges org but in its own repo?

chris48s:
> I was unaware that custom actions needed to have the node_modules directory checked in

The guidance in https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#commit-tag-and-push-your-action-to-github is that you either

- commit `node_modules` or
- pre-compile the action into a `dist/index.js`

but I now wonder if we could `npm install` in the previous step of the workflow before we run it now that it is a local action (I think the reason they advise you to commit `node_modules` is to reduce setup for users if your action will have lots of users). Also one annoyance I didn't spot yesterday is that LGTM is now warning us about some issues in third-party `node_modules` code which I want to ignore. I think I'll have a play with those two alternatives and see if we can find a better way as I think I'd like to find a way round this if possible.

> Is there any downside in including this in-repo

I think the only benefit of it being in a repo of its own is if we want to use it on multiple projects. I think its is probably nicer if when this breaks we can just fix  it here rather than fix it, push a tag and bump the tag

.gitignore
```diff
@@ -66,7 +66,7 @@ bower_components
 build/Release
 
 # Dependency directories
-node_modules/
+/node_modules/
```
calebcartwright, line 69: Think this may pull in deps from the badge-maker package too. Maybe we could use a negation pattern here if we end up keeping the dir checked in?

calebcartwright (commented):
minor item on the ignore file but otherwise lgtm and happy to :+1:. can wait though  if you want to make the `npm install` change mentioned in https://github.com/badges/shields/pull/6103#issuecomment-767827770 (think I'd slightly prefer that if possible though don't feel strongly)

calebcartwright:
The LGTM flags seem minor, though I'm a bit puzzled on it's complaint about the url check within the string since it's a negation :thinking:

calebcartwright (approved):

PR #6127, opened 2021-01-30
Azure Artifacts Badge, run [AzureDevOpsArtifacts]

### LINKED ISSUE #4182: Badge Request: Azure Artifacts packages

devhawk:
fixes #4182

Note, missing tests - not sure what the process is for spinning up sample artifact feed similar to https://dev.azure.com/totodem/Shields.io used for testing azure devops build badges


calebcartwright:
Thanks for the PR!

> not sure what the process is for spinning up sample artifact feed similar to https://dev.azure.com/totodem/Shields.io used for testing azure devops build badges

Ultimately we just need any feed with Public visibility in any Azure DevOps org. The referenced one was added by the original contributor of some AzDO badges, and has been a pretty common pattern.

To my knowledge no one on the maintainer team is a direct user of Azure DevOps Artifact Feeds, so we'd some existing ones to be identified and incorporated as test targets with this proposed addition.

devhawk:
@calebcartwright does the shields.io team own / have admin access to https://dev.azure.com/totodem/Shields.io? I am happy to spin up a azure devops project testing use, but it would seem to make more sense if there was a azure devops org/project owned by shields.io for test purposes.

calebcartwright:
> does the shields.io team own / have admin access to https://dev.azure.com/totodem/Shields.io?

No. As I mentioned in my previous response, this was created by a contributor. 

> but it would seem to make more sense if there was a azure devops org/project owned by shields.io for test purposes.

While we're happy to support this badge, we're not typically in the business of managing all the assets in the myriad tools that Shields integrates with. I won't speak for the other maintains, but I do not have the bandwidth to do the setup and maintain an artifact feed within Azure DevOps that contains various packages and package types.

I would assume that the users that are interested in the addition of this badge have their own feeds already (unless everyone is self-hosting their own private Shields server), and that the tests could simply point to one of those. This is the typical pattern used for the vast majority of our services.

chris48s:
Assuming dev.azure.com is a service where a public project can exist, any public project which we think is likely to continue to exist in the medium-term will do for an integration test.

calebcartwright:
Sorry was short on time earlier, but want to echo what Chris said. We can use any public target, we don't need anything complex/dedicated.

I typically describe our service tests as a combination of high level e2e functional tests (i.e. are we able to produce a badge that's in the right general form) and integration tests with the upstream platform. They serve the purposes of giving us an early alert should something change upstream (new request header required, response body modified, etc.) as well as a degree of confidence that nothing was broken when making any changes, which is super beneficial for us because we provide badges for hundreds of services none of us have ever used before :smile: 

The good news is that we don't need these tests to validate an explicit value, just that it's the right type of badge (e.g. that the result of the test is indeed what our version badges are supposed to look like as opposed to the literal `v1.7.2`).

So here we'd just need to know of a feed (and really, any feed) that was public that had at least one package in it, but we wouldn't need an immutable feed/packages just for us. Most of our tests are pointing to projects/endpoints/targets/etc. that are "real" entities out in the wild, vs. dedicated/managed content for our tests.

I'd imagine there's some relatively well known cases of these types of feeds (maybe even from Microsoft themselves), but we don't have our own Azure DevOps org (nor access to others) where we've set things up for our tests, and we'll have to rely on the community to point out some potential options here :+1:

devhawk:
@clairernovotny who opened the original issue pointed me to https://dev.azure.com/dotnet/NuGetPackageExplorer/_packaging?_a=feed&feed=BuildPackages. I'll add some tests in the next day or so using this feed

calebcartwright:
Have updated the PR title to ensure that the tests get executed by the CI runner. Didn't realize the service class name is `Example` at the moment so whenever you update that in the service file be sure to adjust the name within the `[]` in the PR title first so the tests are picked up

clairernovotny:
It's probably worth verifying that both organization-level and project-level feeds are supported.

devhawk:
@clairernovotny I don't think any of the Azure DevOps badges support auth. Azure Artifacts only supports public access for project level feeds, so as of right now only project-level feeds are supported.

clairernovotny:
Here's an org public feed, no packages in yet yet:

OrgPublicFeed" value="https://pkgs.dev.azure.com/dotnet/_packaging/OrgPublicFeed/nuget/v3/index.json

devhawk:
@clairernovotny I don't have access

``` json
{
"$id": "1",
"innerException": null,
"message": "TF400813: The user 'Windows Live ID\\harrypierson@outlook.com' is not authorized to access this resource.",
"typeName": "Microsoft.TeamFoundation.Framework.Server.UnauthorizedRequestException, Microsoft.TeamFoundation.Framework.Server",
"typeKey": "UnauthorizedRequestException",
"errorCode": 0,
"eventId": 3000
}
```

clairernovotny:
Try in an InPrivate window -- Azure DevOps gets confused if you're already logged into a different instance.

devhawk:
> Try in an InPrivate window -- Azure DevOps gets confused if you're already logged into a different instance.

Asks me to sign in

services/azure-devops/azure-devops-artifacts.service.js
```diff
@@ -0,0 +1,102 @@
...
+const { addv } = require('../text-formatters')
+const {
+  renderVersionBadge: renderNugetVersionBadge,
+} = require('../nuget/nuget-helpers')
+const { NotFound } = require('..')
+const AzureDevOpsBase = require('./azure-devops-base')
+const { keywords } = require('./azure-devops-helpers')
+
+const getPackagesSchema = Joi.object({
+  count: Joi.number().required(),
```
calebcartwright, line 14: As an FYI we have various helpers for schemas too like we do for formatters/renderers, `nonNegativeInteger` from `../validators` would be applicable here

services/azure-devops/azure-devops-artifacts.service.js
```diff
@@ -0,0 +1,102 @@
...
+  value: Joi.array()
+    .items(
+      Joi.object({
+        id: Joi.string().required(),
+        normalizedName: Joi.string().required(),
+        protocolType: Joi.string().required(),
```
calebcartwright, line 20: Given that we're reliant on the literal value of this field for various logic, and that it is presumable a finite set (I assume this maps to the handful of types AzDO supports), I think we should probably use a stricter schema here. For example:

```suggestion
        protocolType: Joi.allow('NuGet', '...').required(),
```

This will give us a louder notice should Microsoft make any changes to this

services/azure-devops/azure-devops-artifacts.service.js
```diff
@@ -0,0 +1,102 @@
...
+}).required()
+
+function renderArtifactBadge(version, protocolType) {
+  return protocolType === 'NuGet'
+    ? renderNugetVersionBadge({ version, feed: `Azure Artifacts NuGet` })
+    : {
+        label: `Azure Artifacts ${protocolType}`,
+        message: addv(version),
+        color: versionColor(version),
+      }
```
calebcartwright, line 41: We'll need to make some adjustments to the label/feed here, though not entire sure precisely what they should be just yet. One of the core tenants of our [specification that drives the badges](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md) is that the badges be non-promotional (i.e. labels should not focus on service names).

This is why the other badges that display data from Azure DevOps have labels like `build`, `deployment`, `tests`, and `coverage` as opposed to things like `Azure Pipelines`.

However, there's plenty of existing version badges, and certainly other nuget-version badges that don't seem to be following this.

I suspect that at a minimum we'll need to:
* use lower case for the feed name in the nuget case
* skip the custom label for the non-nuget case (leave it as the default `version`)

this would at least ensure that these nuget badges are consistent with the other nuget version badges, and non-nuget packages (e.g. npm packages in in AzDO artifact feed) would be more consistent with other non-nuget version badges 

cc @badges/shields-core for other perspectives/in case i'm forgetting anything on our nuget badge labels/etc.

services/azure-devops/azure-devops-artifacts.service.js
```diff
@@ -0,0 +1,102 @@
...
+      throw new NotFound({ prettyMessage: `${packageName} package not found` })
+    }
+
+    const packageObj = json.value[0]
+    if (packageObj.normalizedName !== packageName.toLowerCase()) {
+      throw new NotFound({ prettyMessage: `${packageName} package not found` })
+    }
+
+    const protocolType = packageObj.protocolType
+    const version = packageObj.versions[0].version
```
calebcartwright, line 98: A common pattern followed in many of our service classes is to extract the responsibilities of fetching the data from the upstream service/platform, and then extracting/transforming the data from the response, into separate functions respectively named `fetch` and `transform` that are leveraged by the `handle` function.

These lines would make for a canonical `transform` function, so may be worth considering pulling them out. This is done both for general code organization reasons, but also commonly because it makes testing the logic much easier and doable via simple unit tests

services/azure-devops/azure-devops-artifacts.tester.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
+
+// const Joi = require('joi')
```
calebcartwright, line 3: let's get rid of this if there isn't a need to define any validation schemas 

```suggestion
```

services/azure-devops/azure-devops-artifacts.tester.js
```diff
@@ -0,0 +1,31 @@
...
+
+t.create('prerelease package')
+  .get(
+    '/vpre/dotnet/NuGetPackageExplorer/BuildPackages/Microsoft.Build.NuGetSdkResolver.json'
+  )
+  .expectBadge({
+    label: 'Azure Artifacts NuGet',
+    message: isVPlusDottedVersionNClausesWithOptionalSuffix,
+    color: 'yellow',
+  })
```
calebcartwright, line 31: We'll want to have some tests that exercise the other flows (such as a non-existent package, feed, project, etc.). 

I'm a bit unclear on the path where the requested package is found but the normalized name doesn't match the package name, but we'd ideally have a test for that case too (as noted above moving that logic to a `transform` function would likely make this easier to test, [here's an example of some other unit tests for service class functions](https://github.com/badges/shields/blob/master/services/codecov/codecov.spec.js))

calebcartwright (commented):
Thanks again for this! Things seem to be progressing nicely, and I've left some inline comments below on a few places where we'll want to tweak or extend

services/azure-devops/azure-devops-artifacts.service.js
devhawk: Thanks for the feedback @calebcartwright . I will address these suggestions in the next day or two

PyvesB:
@devhawk any chance you could circle back to this and take a look at @calebcartwright's review comments? It would be great to get this new badge landed. :wink:

devhawk:
I can work on this Friday

PyvesB:
> 
> 
> I can work on this Friday

Doing another pass over some of the open pull requests - did you have a particular Friday in mind? 😄

PyvesB:

If someone else wants to pick this up, you can simply fetch the commits by running `git fetch origin pull/6127/head:pr-6127`. Address the pending review comments and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

PR #6128, opened 2021-01-30
Update Contributing documentation on PRs/service tests & amending history

calebcartwright:
No associated issue, but has been discussed offline in Discord. Given the addition of the PR/merging info, this also includes some minor reshuffling and deduplication of content about PR titles for service tests


.github/pull_request_template.md
```diff
@@ -0,0 +1,7 @@
+<!--
+    Be sure to review our Contributing guidelines to help streamline the merging of your PR!
+
+    * PR title conventions - https://github.com/badges/shields/blob/master/CONTRIBUTING.md#running-service-tests-in-pull-requests
+    * Code formatting - https://github.com/badges/shields/blob/master/CONTRIBUTING.md#prettier
+    * Merge processes and reminders - https://github.com/badges/shields/blob/master/CONTRIBUTING.md#pull-requests
+-->
```
calebcartwright, line 7: I originally had leading newlines before the comment, but forgot prettier doesn't like those. My thinking was to include this as a comment so it'd be presented to folks when the open a PR without adding noise to the PR itself, and that having leading whitespace would avoid any nuisance of having to move the comments around.

Let me know if anyone thinks keeping the leading whitespace would be preferable, and I'll work on setting up an exception for pretteir

chris48s: I think it is fine as it is

CONTRIBUTING.md
```diff
@@ -153,3 +147,30 @@ There is a [High-level code walkthrough](doc/code-walkthrough.md) describing the
 ### Logos
 
 We have [documentation for logo usage](doc/logos.md) which includes [contribution guidance](doc/logos.md#contributing-logos)
+
+## Pull Requests
+
+All code changes are incorporated via pull requests, and pull requests are always squashed into a single commit on merging. Therefore there's no requirement to squash commits within your PR, but feel free to squash or restructure the commits on your PR branch if you think it will be helpful. PRs with well structured commits are always easier to review!
+
+Because all changes are pulled into the main branch via squash merges from PRs, we do **not** support overwriting any aspects of the git history once it hits our main branch. Notably this means we do not support amending commit messages, nor adjusting commit author information once merged.
```
chris48s, line 155: I guess the other case where we might want to (or someone might request that we) rewrite history is where credentials have been exposed. Hopefully we'd catch it in review, but lets say we commit a token or password to the repo. Obviously the ideal course of action is we revoke the token or change the password and make a subsequent commit that removes the credentials. Thinking about that example, can we see a situation in which we would want to break our own rule here?

CONTRIBUTING.md
```diff
@@ -153,3 +147,30 @@ There is a [High-level code walkthrough](doc/code-walkthrough.md) describing the
...
+
+Accordingly, it is the responsibility of contributors to review this type of information and adjust as needed before marking PRs as ready for review and merging.
```
chris48s, line 157: Worth explicitly saying review your `git config` settings/linking to docs?

CONTRIBUTING.md
calebcartwright: Fair question. I think my preference is to set a pretty categorical rule in documentation vs trying to enumerate a full set of exceptions or having a softer policy with vague exceptional cases. We will always have to the option to evaluate the specific scenario should it ever come up again, but for the docs targeted to the audience of contributors, I really want to stress that it's one of their responsibilities.

CONTRIBUTING.md
calebcartwright: Good idea, updated

chris48s (approved):

PR #6131, opened 2021-01-31
add script for adding/removing heroku dynos

chris48s:
We've discussed this a bit on discord, but just to summarise quickly.

Our traffic pattern for shields is "lumpy" in predictable ways. For example:

Screenshot at 2021-01-31 12-57-33

We also serve a lot more traffic during Americans-being-awake hours. It would be helpful if we could run more/less dynos at different times but we can't use heroku's autoscaling features on standard tier dynos :(

Because our traffic pattern is quite predictable an alternative approach would be to scale up/down on a schedule at different times based on our knowledge of regular peaks/troughs, as per https://stackoverflow.com/questions/11949240/schedule-number-of-web-dynos-by-time-of-day/11961899#11961899

This PR implements a script we can use to scale up/down e.g: `npm run heroku:scale 5`. You can test this locally with `HEROKU_API_TOKEN` and `HEROKU_APP_ID` env vars, if you want.

TODO:
- Populate up a `HEROKU_API_TOKEN` env var in heroku (we already have `HEROKU_APP_ID` defined).
- Use [heroku scheduler](https://devcenter.heroku.com/articles/scheduler) to set up some scaling events (there's probably going to be a bit of trial and error to get the times/dyno numbers right)


Procfile
```diff
@@ -1 +1,5 @@
 web: npm run start:server:prod
+scale4: npm run heroku:scale 4
+scale5: npm run heroku:scale 5
+scale6: npm run heroku:scale 6
+scale7: npm run heroku:scale 7
```
chris48s, line 5: As far as I can tell from reading [the docs](https://devcenter.heroku.com/articles/procfile):

1. We need a Procfile entry to schedule a job because the scheduled job runs on a separate dyno. We can't directly schedule `npm run heroku:scale 5`
2. We can use `$VAR` to refer to env vars we've defined, but I don't think there is a way to pass in a variable to a command in the Procfile when we invoke it

so I think the best we can do here is give ourselves some "presets" we can use in schedule jobs.

calebcartwright (approved):
Looks reasonable to me :+1:

chris48s:
Next thing to do is set up some scaling events. If we have an ops meeting tomorrow lets discuss it then

Issue #5994, opened 2020-12-28
Is it possible to connect to a private Jira Installation (on prem)?


danielle:
HI, I have a on prem, installation of Jira, all my servers are local without connectivity outside of our firewall |
how can I use shields ?

I would like to use total number of Jira/Issues open for specific projects.  Stories, Bugs etc....  
Current Sprint, and sprint completion and other details from Jira.

I believe this will require a fix similar to https://github.com/badges/shields/issues/1757 and Pull Request: https://github.com/badges/shields/pull/2598

Thanks for helping

calebcartwright:
I feel there's two parts to this question

The first is related to the Jira badges we currently support. One of the best ways to discover the available badges is to utilize the search feature on the [Shields.io website](https://shields.io) (textbox at the top of the page)

A quick search shows that we already provide a sprint completion progress/status badge, as well as a status badge for a given issue:

![image](https://user-images.githubusercontent.com/13042488/103233447-cd6a7380-4902-11eb-83cf-9758dc6fc200.png)

If you'd like to request additional badges, please utilize the [corresponding issue template](https://github.com/badges/shields/issues/new?labels=service-badge&template=3_Badge_request.md) with the relevant information so that a maintainer or member of our community could potentially work on adding the requested badge.

The second item is about utilizing Shields with private/non-internet facing content. The Shields.io service is the main public deployment of the Shield server which is developed/maintained in this repository, and the Shields.io service is able to provide badges for public tools/projects that are themselves public and internet facing. The Shields.io service cannot provide badges for tools that it cannot reach (non-internet facing) and/or that require project/user scoped authorization.

In order to utilize Shields to get badges for private/non-internet facing tools you'll need to deploy your own Shields server, and that process is documented [here](https://github.com/badges/shields/blob/master/doc/self-hosting.md) (I personally recommend using the Docker image).

Basically, you'll want to deploy your own self-hosted Shields server within your same private network and, if your Jira server requires authentication, configure your Shields deployment with your Jira auth information (https://github.com/badges/shields/blob/master/doc/server-secrets.md#jira)

calebcartwright:
Going to go ahead and close as it feels the question has been answered, though please feel free to follow up with any additional questions!

danielle:
[...]

Issue #5894, opened 2020-12-02
Running shields instance behind reverse proxy breaks page URLs


voodoodror:
Are you experiencing an issue with...

- [ ] [shields.io](https://shields.io/#/)
- [x] My own instance
- [ ] [badge-maker NPM package](https://www.npmjs.com/package/badge-maker)

:beetle: **Description**

As part of effort in securing my shields server using nginx reverse proxy (i.e. with queryparameter validation), it looks like page URLs are set with a slash:
```
href="/.....' 
src="/.....' 
data="/.....' 
```
and as a result I'm unable to use badges nor links. Tried also rewriting the URLs using nginx without any success.

:bulb: **Possible Solution**

Unfortunately, I didn't find any workaround.

paulmelnikow:
Could you explain why the slash is causing a problem?

voodoodror:
@paulmelnikow Sorry, my bad. forgot to mention that I used different path on nginx for reverse proxy, instead of root folder, it redirects to `www.myurl.com/shields/`

paulmelnikow:
You could try setting `GATSBY_BASE_URL` before running the build.

Why do you need to host it from a subfolder?

voodoodror:
@paulmelnikow I will be trying `GATSBY_BASE_URL` and let you know.

Issue #4817, opened 2020-03-26
How to set my 'path' when hosting my own server


forp007:
:question: **Question**

Hi, I am hosting my own shields server,  is it possible to change the default path to be something like: http://myserver.example.com/shields. 

When doing testing,  I would like instead of 'http://localhost:3000'  my server set as http://localhost:3000/shields 

Thanks

paulmelnikow:
There's nothing like this built in, or an easy place where this could be changed in one place. I think the easiest way to accomplish what you're asking would be to place a separate server like [nginx](https://nginx.org/en/) in front of the Node process running Shields.

PR #6156, opened 2021-02-05
add info on releases to self-hosting docs

chris48s:
The main thing going on in this PR is adding docs about tagged releases, but I updated a couple of other things (grouped docs about deploying from source together and docs about deploying with docker together, updated "now" to "vercel").

### DIFF
--- a/doc/self-hosting.md
+++ b/doc/self-hosting.md
@@ -1,6 +1,19 @@
 # Hosting your own Shields server
 
-## Installation
+Some users may wish to host their own instance of shields. This is particularly useful if you want to serve badges for resources that require authentication or are not exposed to the internet (e.g: inside a corporate network). A variety of options are available either by installing from source or using a docker image.
+
+## Releases
+
+- https://github.com/badges/shields is a monorepo and hosts the shields frontend and server code as well as the [badge-maker](https://www.npmjs.com/package/badge-maker) NPM library.
+- The server uses [Calendar Versioning](https://calver.org/). Tags of the form `server-YYYY-MM-DD` are server releases (these are the tags that are relevant to self-hosting users).
+- Badge-maker uses [Semantic Versioning](https://semver.org/). Tags of the form `X.Y.Z` are badge-maker releases.
+- As well as [tags on GitHub](https://github.com/badges/shields/tags), server releases are also pushed to [DockerHub](https://registry.hub.docker.com/r/shieldsio/shields/tags). See the section on [Docker](#Docker) for more details.
+- We publish release notes for server releases in the [CHANGELOG](https://github.com/badges/shields/blob/master/CHANGELOG.md). There may occasionally be non-backwards compatible changes to be aware of.
+- We will normally put out one release per month. If there is a security patch or major bugfix affecting self-hosting users, we may put out an out-of-sequence release.
+- Releases are just a snapshot in time. We advise always tracking the latest release to ensure you are up-to-date with the latest bug fixes and security updates. There are no 'patch' releases - we don't backport fixes to old releases. Tagged versions just provide a convenient way to apply upgrades in a controlled way or roll back to an older version if necessary and communicate about versions.
+- You can stay on the bleeding edge by tracking the `master` branch for source installs or the `next` tag on DockerHub.
+
+## Installing from Source
 
 You will need Node 12 or later, which you can install using a
 [package manager][].
@@ -14,18 +27,19 @@ curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -; sudo apt-get in
 ```sh
 git clone https://github.com/badges/shields.git
 cd shields
+git checkout $(git tag | grep server | tail -n 1)  # checkout the latest tag
 npm ci  # You may need sudo for this.
 ```
 
 [package manager]: https://nodejs.org/en/download/package-manager/
 
-## Build the frontend
+### Build the frontend
 
 ```sh
 npm run build
 ```
 
-## Start the server
+### Start the server
 
 ```sh
 sudo node server
@@ -44,7 +58,7 @@ The root gets redirected to https://shields.io.
 
 For testing purposes, you can go to `http://localhost/`.
 
-## Heroku
+### Deploying to Heroku
 
 Once you have installed the [Heroku CLI][]
 
@@ -57,9 +71,32 @@ heroku open
 
 [heroku cli]: https://devcenter.heroku.com/articles/heroku-cli
 
+### Deploying to Zeit Vercel
+
+To deploy using Zeit Vercel:
+
+```console
+npm run build  # Not sure why, but this needs to be run before deploying.
+vercel
+```
+
 ## Docker
 
-You can build and run the server locally using Docker. First build an image:
+### DockerHub
+
+We publish images to DockerHub at https://registry.hub.docker.com/r/shieldsio/shields
+
+The `next` tag is the latest build from `master`, or tagged releases are available
+https://registry.hub.docker.com/r/shieldsio/shields/tags
+
+```console
+$ docker pull shieldsio/shields:next
+$ docker run shieldsio/shields:next
+```
+
+### Building Docker Image Locally
+
+Alternatively, you can build and run the server locally using Docker. First build an image:
 
 ```console
 $ docker build -t shields .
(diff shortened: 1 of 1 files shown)


doc/self-hosting.md
```diff
@@ -1,6 +1,19 @@
...
+- As well as [tags on GitHub](https://github.com/badges/shields/tags), server releases are also pushed to [DockerHub](https://registry.hub.docker.com/r/shieldsio/shields/tags). See the section on [Docker](#Docker) for more details.
+- We publish release notes for server releases in the [CHANGELOG](https://github.com/badges/shields/blob/master/CHANGELOG.md). There may occasionally be non-backwards compatible changes to be aware of.
+- We will normally put out one release per month. If there is a security patch or major bugfix affecting self-hosting users, we may put out an out-of-sequence release.
+- Releases are just a snapshot in time. We advise always tracking the latest release to ensure you are up-to-date with the latest bug fixes and security updates. There are no 'patch' releases - we don't backport fixes to old releases. Tagged versions just provide a convenient way to apply upgrades in a controlled way or roll back to an older version if necessary and communicate about versions.
+- You can stay on the bleeding edge by tracking the `master` branch for source installs or the `next` tag on DockerHub.
```
chris48s, line 14: @calebcartwright can you confirm this "releases" section covers all the points you think are important

calebcartwright: Apologies in advance as this won't be a succinct, direct answer to your question :smile: I feel like I need to get some clarity on direction before I can really comment on whether I feel we have complete documetation coverage here. Essentially, I'm still unclear if we're making any changes to what I'll call our "official" products, specifically as it relates to the server.

From my perspective, our official/released/supported products have consisted of the Shields.io _service_ and the npm package, but the server/self-hosted hasn't been in that same tier.

Obviously given the open source nature of the repository, everyone has always had access to the source of the server code that powers the Shields.io service, and thus been able to deploy their own, and ~18 months ago [we started building and publishing an image to DockerHub](#3558) instead of only providing the Dockerfile for users to build the image themselves.However, my sense is that these actions were primarily about making it easier for someone to do their own self-hosting, as opposed to making the server a full blown "product" on par with the main service and npm package.

With that context around my current understanding, I have two primary points of concern/hesitation that will directly impact my thoughts on what I think should be documented:

1. Lack of clarity on the server/self-hosted offering, if any, particularly relative to the service and package
2. A reasonable person could interpret/assume things about the server/self-hosted offering that may not be true, particularly given the first bullet and some of the recent enhancements

My strong preference would be that as the core team we make a decision about the nature of our server/self-hosted offering, and then codify what our supported offerings are one way or the other (I think that should likely live outside the how-to-self-host-the-server doc though). Based on past discussions around the topic, I get the sense others my disagree that there's a lack of clarity and/or that there's a need for such an explicit decision. However, I'd really like to avoid ambiguity especially with the recent and ongoing enhancements around the releases of the server.

And to be clear, I'm not suggesting a thesis is needed here; I just think it's important to be explicit about whether the server is developed/maintained/etc. strictly for the Shields.io service with self-hosting/non-service use cases being a best effort and secondary, _or_ if we're actually going to make the server itself as one of our core offerings along side the service and package, and users can expect a similar level of enhancement/support/etc.

Here's a few illustrative cases of why I think it's important to make it clear:
* We've historically rejected new badges that would only be available from self-hosted instances, e.g. those that required user/account-specific auth and would not be offerable via the Shields.io service (I forget the exact issues/cases, but hope others recall these discussions or have better issue-tracker-searching skills than I)
* We've historically declined (or at least de-prioritized) feature requests that are only relevant for self-hosted targets, but which have no utility for the service, for example running behind reverse proxies/serving on non-default paths (#5894, #4817), restricting access to the self-hosted instance at the server itself https://github.com/badges/shields/issues/4862#issuecomment-609600135 (few others I can't find ATM), etc.
* We've historically declined (or at least de-prioritized) deployment options that aren't relevant for the service hosting environment (OVH servers before, and now Heroku), e.g. #1778
* We've historically only offered very minimal support/troubleshooting for self-hosting issues (some of which is forced by the nature that we have no insight/access into self-hosting environments), e.g. #5994, #1753
* We don't do any testing/validation of the released server, and we don't used the released server for the Shields.io deployment artifact (heroku slug vs. docker image, we don't have functional tests for a self-hosted server configured with auth against different versions of tools, like SonarQube, etc.)

My sense is that our posture is really the former strategy from my description above (where we're Shields.io focused, other server releases/use cases are secondary), and all of those above positions and practices make perfect sense with that strategy. However, I'm not sure all of those would necessarily be tenable if our posture is actually the latter (server is a full blown product on par with service and package).

It's already not 100% clear to me as a maintainer what our posture actually is though, and I worry that the introduction of tags/releases in GitHub, versioned tags on published Docker images, change log, etc. without an explicit statement could really muddy the waters for users.

So to close out, and really reiterate, I'm primarily focused on getting clarity on our posture and documenting it accordingly, though if I'm mistaken and we are doing more with the server then I think we'll likely need some follow up discussions offline about the implications.

chris48s: That's a lot of text. I do have some thoughts on various bits of this and I've written some notes on it, but I'm going to be super-annoying and intentionally not reply to any of it here, except for one sentence :smile: 

> I think that should likely live outside the how-to-self-host-the-server doc though

Bingo.
There's some stuff here that is worth talking about, but most of it seems like bigger wider issues and doesn't necessarily fit into a doc explaining how to set your own instance of Shields.

Lets accept the premise that we agree that allowing ourselves and our users to express the concept of "what version do I have" as `server-2021-02-01` rather than `926f1e4` has value and we're going to do it (and given you've reviewed and approved https://github.com/badges/shields/pull/6069 and https://github.com/badges/shields/pull/6102 I think that's a reasonable base assumption).. What out of this do we need to deal with here to improve a document that tells you how to set up your own instance of shields (which is a document that already does exist and will continue to do so), and what can we move to another wider discussion?

calebcartwright: > There's some stuff here that is worth talking about, but most of it seems like bigger wider issues and doesn't necessarily fit into a doc explaining how to set your own instance of Shields.

> What out of this do we need to deal with here to improve a document that tells you how to set up your own instance of shields (which is a document that already does exist and will continue to do so), and what can we move to another wider discussion?

From my POV, if we want to keep this document and the discussion centered around setting up your own instance of Shields, then I don't think any of the above text belongs in this document. I'd note that things like the monorepo nature of repository, the versioning scheme for the npm package, cadence+description of server releases, etc. are not strictly related to the "what specific steps do I take to start my own Shields server" focus that this document currently has (which was why I made the comment about these topics potentially being better suited to a separate doc)

However, if we're going to broaden the scope of this document to also describe various aspects of releases and other information, as I'd suggest these proposed changes do, then I think it's only fair that the corresponding broadened scope be open for discussion as well.

> Lets accept the premise that we agree that allowing ourselves and our users to express the concept of "what version do I have" as server-2021-02-01 rather than 926f1e4 has value and we're going to do it (and given you've reviewed and approved #6069 and #6102 I think that's a reasonable base assumption)

I've been raising these concerns repeatedly in various forums for a while now. I've accepted the collective decision that we're moving forward with some of these changes, and I've reviewed and approved because I felt the "how" (i.e. implementation) was a sound approach for the respective "what" (i.e. the change/enhancement). However, that doesn't mean that my main concern about clearly articulating and documenting the _**why**_ behind the driver for making these changes has gone away.

To be honest, if the addition of documentation text that describes the new and enhanced releases isn't an appropriate forum to discuss/request additional docs about the releases, then I'm rather confused as to what would be an appropriate forum.

> @calebcartwright can you confirm this "releases" section covers all the points you think are important

My position on this is overall theme is still the same:
* no, there's a couple other points about server releases that i think are important
* if we want to describe general, repo-wide release information like this then I think much, if not all, of this should live in a separate doc which this doc can link to

chris48s: The reason I'd suggest explaining some of this here is that step one of setting up your own install is getting some code or an image from somewhere. That does require making a choice so some level of explanation of what release assets exist, where you can find them and what the tradeoffs are is useful in making that decision.

I think I'm just going to close the PR and leave this document how it is for the moment though. Sounds like there's some bigger stuff to work out first (which needs more voices than just you and me). We can circle back to this in that context. Maybe if that drags out I'll do another PR just to pick out the non-controversial fixes (e.g: fix the now/vercel docs etc)

PR #6157, opened 2021-02-06
Improve changelog entries

### LINKED ISSUE #5866: Automate Calendar Versioning/Changelog for self-hosting users

chris48s:
As noted in https://github.com/badges/shields/pull/6132#issuecomment-771076613 after the first run of this I picked out a couple of tweaks I wanted to make to the script

Closes #5866

### DIFF
--- a/.github/actions/draft-release/entrypoint.sh
+++ b/.github/actions/draft-release/entrypoint.sh
@@ -26,6 +26,8 @@ touch temp-changes.txt
     git log "$LAST_TAG"..HEAD --no-merges --oneline --pretty="format:- %s" --perl-regexp --author='^((?!dependabot).*)$'
     echo $'\n- Dependency updates\n'
 } >> temp-changes.txt
+BASE_URL="https:\/\/github.com\/badges\/shields\/issues\/"
+sed -r -i "s/\((\#)([0-9]+)\)$/\[\1\2\]\($BASE_URL\2\)/g" temp-changes.txt
 
 # Write the changelog
 sed -i "${INSERT_POINT} r temp-changes.txt" CHANGELOG.md
--- a/.github/actions/draft-release/entrypoint.sh
+++ b/.github/actions/draft-release/entrypoint.sh
@@ -24,7 +24,7 @@ touch temp-changes.txt
     echo "## $RELEASE_NAME"
     echo ""
     git log "$LAST_TAG"..HEAD --no-merges --oneline --pretty="format:- %s" --perl-regexp --author='^((?!dependabot).*)$'
-    echo $'\n'
+    echo $'\n- Dependency updates\n'
 } >> temp-changes.txt
 
 # Write the changelog


.github/actions/draft-release/entrypoint.sh
```diff
@@ -24,8 +24,10 @@ touch temp-changes.txt
     echo "## $RELEASE_NAME"
     echo ""
     git log "$LAST_TAG"..HEAD --no-merges --oneline --pretty="format:- %s" --perl-regexp --author='^((?!dependabot).*)$'
-    echo $'\n'
+    echo $'\n- Dependency updates\n'
```
calebcartwright, line 27: I was wondering if there were any extreme edge cases where we could have one of these changelogs generated for a delta that didn't actually contain any dependency updates  (e.g. ad-hoc release shortly before scheduled release, preceding another dependabot round). I think there may be some remote possibility, but it seems sufficiently unlikely to happen in reality, and even if it did, there's no substantive impact anyway

calebcartwright (approved):

.github/actions/draft-release/entrypoint.sh
chris48s: If so, it can be removed before merging. Same as we'd remove a commit message about updated documentation or whatever..More often than not, it is going to be useful to not add it by hand

calebcartwright: Completely agreed!

Issue #6158, opened 2021-02-07
Make Twitch a social badge


PyvesB:
:clipboard: **Description**

Our Twitch badges are currently part of the `activity` category. All other badges in that category refer to release dates or contributions, whereas the Twitch badge only really indicates "is this user currently streaming live?", which feels quite unrelated. The `social` category seems much more appropriate, and I suggest we also modify the badge's default style to be social as well.

What do people think?

calebcartwright:
> The `social` category seems much more appropriate, and I suggest we also modify the badge's default style to be social as well.
> 
> What do people think?

I think that's a reasonable category change. I often struggle to place the category for some badges which can seemingly go in multiple, but I can see the social category being a better argument for this one.

Could you elaborate on the change to the default style? Does that mean that existing twitch badges that don't have an explicit `style` query param set would all be changed to the social style?

PyvesB:
> Could you elaborate on the change to the default style? Does that mean that existing twitch badges that don't have an explicit `style` query param set would all be changed to the social style?

Yes, I'm suggesting changing the default style of all Twitch badges without an explicit style parameter. Three thoughts to justify this:
* users can always go back to the previous style by adding the query parameter.
* Twitch badges aren't massively popular so impact would actually be quite limited. In  #4778 and  #5214 we took several weeks before noticing the badge was completely broken, with no users reporting the problems.
* not giving them the social style would mandate that links be removed in  #6159, which is a different potential impact on existing users.

calebcartwright:
Those are sound arguments which make for a logically defensible position :+1: 

Making any change in Shields that results in any impact on any existing badge always gives me great angst, but I can get on board with this as long as we do the typical messaging campaign before actually deploying the change

calebcartwright:
Actually one other point for consideration from recent discussions I'd already forgotten about :laughing: 

We recently removed the default style override to social on the GitHub sponsors badge in #5888 and my recollection is that we generally wanted to try to avoid overriding the default. Do others recall that discussion, and/or is it inapplicable here (or worth overriding in this case anyway)?

I generally don't have a strong opinion on this one way or the other, just trying to recall any recent relevant conversation

PyvesB:
[...]

PR #6159, opened 2021-02-07
Allow links only for social badges, affects [GitHubMilestone Dependabot]

PyvesB:
Related to #6096.

Twitch is also another badge that doesn't follow this, however, I'm suggesting #6158 as a resolution instead.

### DIFF
--- a/services/dependabot/dependabot.service.js
+++ b/services/dependabot/dependabot.service.js
@@ -58,7 +58,6 @@ module.exports = class DependabotSemverCompatibility extends BaseJsonService {
     return {
       color: json.colour,
       message: json.status,
-      link: this._getLink({ packageManager, dependencyName }),
     }
   }
 }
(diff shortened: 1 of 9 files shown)


services/dependabot/dependabot.service.js
```diff
@@ -58,7 +58,6 @@ module.exports = class DependabotSemverCompatibility extends BaseJsonService {
     return {
       color: json.colour,
       message: json.status,
-      link: this._getLink({ packageManager, dependencyName }),
```
chris48s, line 61: Do we still need to keep the `_getLink()` function itself? Seems like we could remove it..

PyvesB: Indeed, had missed it, good catch.

chris48s:
sorry - one more request. Can we add a line to this section of docs

https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines

saying default links are always off for non-social badges

chris48s (approved):

PyvesB:
Deployed to production and communication on Twitter and Discord done.

PR #6162, opened 2021-02-09
Add [Testspace] badges

### LINKED ISSUE #6077: Add Testspace Test Results Badges
:clipboard: **Description**

- Which service is this badge for:  [Testspace](https://www.testspace.com)
- What sort of information should this badge show? Standard test result badges like we have for other services, depending on desired variants (counts of passed/failed/skipped, percentage passed, etc.)
 

Couple existing examples: 

AppVeyor:

Sonar:

:link: **Data**

- Is there a public API? Yes
- Does the API requires an API key? No (data for public projects anonymously accessible)
- Link to the API documentation. https://help.testspace.com/docs/reference/web-api#list-results

The array of results in the response from that API seems to be chronologically ordered, so should suffice to just grab the first element and then extract the test counts from there (`case_counts` field). I tested it out in one of my projects so sharing for reference:

[API Response](https://swellaby.testspace.com/api/projects/swellaby:nyc-config/spaces/133026/results)
[Web Dashboard](https://swellaby.testspace.com/projects/66205)

:microphone: **Motivation**
As I've been migrating more and more of my projects from various CI/pipeline tools to the GitHub Actions ecosystem, one thing I'm finding myself missing is the rich test result metrics and the ability to have my test result badges :disappointed: 

I've been doing a bit of exploring and Testspace has piqued my interest as a potential option. While I suspect that one day that GitHub will provide native test data via Actions/GitHub API, this seems like a worthwhile service to support since users could leverage it from GHA or any other CI platform for that matter.

calebcartwright:
Resolves #6077 

Opted to pull the common bits into a separate base class and define the badge classes in separate file to get shorter files and slightly simpler tests, but otherwise very similar to our other test result related badge friends (Sonar, AppVeyor, Azure DevOps, Jenkins, etc.)


services/testspace/testspace-base.js
```diff
@@ -0,0 +1,61 @@
...
+// Project is generally a repository
+// Space is a container, often a branch
+module.exports = class TestspaceBase extends BaseJsonService {
+  static category = 'build'
+  static defaultBadgeData = { label: 'tests' }
+
+  async fetch({ org, project, space }) {
+    // https://help.testspace.com/docs/reference/web-api#list-results
+    const url = `https://${org}.testspace.com/api/projects/${encodeURIComponent(
+      project
```
calebcartwright, line 35: Project names by default seem to follow the `org-name:repo-name` pattern so think encoding here is necessary

services/testspace/testspace-base.js
```diff
@@ -0,0 +1,61 @@
...
+        403: 'org not found or not authorized',
+        404: 'org, project, or space not found',
+      },
+    })
+  }
+
+  transformCaseCounts(json) {
+    if (json.length === 0) {
+      throw new NotFound({ prettyMessage: 'space not found or results purged' })
+    }
```
calebcartwright, line 50: spaces can either be specified by a branch name, or an integer id. I'm not sure how likely this would be in practice, but I ran into it after I'd saved a link (which default to the integer id) for a PR run

services/testspace/testspace-test-count.service.js
```diff
@@ -0,0 +1,69 @@
+'use strict'
+
+const { metric: metricCount } = require('../text-formatters')
+const TestspaceBase = require('./testspace-base')
+
+module.exports = class TestspaceTestCount extends TestspaceBase {
+  static route = {
+    base: 'testspace',
+    pattern:
+      ':metric(total|passed|failed|skipped|errored)/:org/:project/:space+',
```
calebcartwright, line 10: space will often be specified with a branch name, and of course branches in source repos can have `/`

services/testspace/testspace-test-summary.service.js
```diff
@@ -0,0 +1,104 @@
...
+module.exports = class TestspaceTests extends TestspaceBase {
+  static route = {
+    base: 'testspace/tests',
+    pattern: ':org/:project/:space+',
+    queryParamSchema: testResultQueryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'Testspace tests',
```
chris48s, line 19: If I look at the preview dialog for the non-compact example, the compact option is checked.

services/testspace/testspace-base.js
```diff
@@ -0,0 +1,61 @@
...
+  async fetch({ org, project, space }) {
+    // https://help.testspace.com/docs/reference/web-api#list-results
+    const url = `https://${org}.testspace.com/api/projects/${encodeURIComponent(
+      project
+    )}/spaces/${space}/results`
+    return this._requestJson({
+      schema,
+      url,
+      errorMessages: {
+        403: 'org not found or not authorized',
```
chris48s, line 41: Given we're not offering any way to supply credentials, would it be better to say something like "org not found or private"?

services/testspace/testspace-test-count.tester.js
```diff
@@ -0,0 +1,44 @@
...
+    label: 'skipped tests',
+    message: isMetricAllowZero,
+  })
+
+t.create('Errored')
+  .get('/errored/swellaby/swellaby:testspace-sample/main.json')
+  .expectBadge({
+    label: 'errored tests',
+    message: isMetricAllowZero,
+  })
```
chris48s, line 44: Are we omitting tests for the not found case(s) here because it is handled in the base class and covered by another test file, or is it oversight?

services/testspace/testspace-test-summary.service.js
calebcartwright: Good catch, copy/paste fail

services/testspace/testspace-test-count.tester.js
calebcartwright: Those tests were added to https://github.com/badges/shields/blob/593a6ace605613a95219804fbc06f397ca378cb2/services/testspace/testspace-test-pass-ratio.tester.js 

Some other maintainers in the past have expressed concerns with having multiple not found cases in instances like this where all the service classes are hitting the same upstream endpoint via a base class, so opted to just stick it in another tester that had fewer scenarios

services/testspace/testspace-base.js
calebcartwright: we aren't yet, but there's a non-zero chance imo that we could do so down the road which could of course then bring a bad/invalid auth into the self-hosting picture. i don't have a strong opinion on the text one way or the other though if you think `private` is better

chris48s (approved):

PR #1189, opened 2017-10-20
[Coveralls] bitbucket support #793

piekar294:
- adds support for bitbucket in coveralls
- move to new coveralls api
- add test for coveralls

https://github.com/badges/shields/issues/793

PR #6181, opened 2021-02-14
Make [Coveralls] examples consistent with regards to VCS parameter

PyvesB:
Other services (Codecov, Codefactor, CircleCI, JitePack, Scrutinizer) have the VCS parameter as a dropdown in the example rather than one example for every VCS type. This makes the Coveralls examples more consistent.


chris48s:
For legacy reasons, the coveralls badge has this slightly odd property where the `vcsType` property is optional and if it is missing we assume it is github. At the moment, the examples hide this property and the builder essentially makes the VCS property required, whereas this change exposes it.
I don't have really strong opinions on it as I suspect we'll never drop the legacy route, but its worth pointing out in case you hadn't spotted it. I suspect this might be why this one is a bit different.

PyvesB:
Yes, I did spot that, but hadn't quite clocked that hiding the fact that not specifying the VCS type was considered legacy here. What I can do however is make the VCS type mandatory and instead implement a redirect function for existing badges that don't specify it. What do you think?

chris48s:
> Yes, I did spot that, but hadn't quite clocked that hiding the fact that not specifying the VCS type was considered "legacy"

Really digging into the archives for this one :pick:  https://github.com/badges/shields/pull/1189#discussion_r145990571

> What I can do however is make the VCS type mandatory and instead implement a redirect function for existing badges that don't specify it. What do you think?

Sounds good. That seems like how how we would probably handle this if we added bitbucket today :+1:

PyvesB:
> Sounds good. That seems like how how we would probably handle this if we added bitbucket today 👍

Cool, will update the PR later this week!

chris48s (approved):

PR #6229, opened 2021-02-28
ensure redirect target path is correctly encoded

chris48s:
This fixes a bug that causes `setHeader()` to throw an unhandled `TypeError` if the path of the URL we are redirecting to contains unescaped characters which are invalid for a HTTP header.
Note this is only the url path affected. The querystring was already encoded correctly

Refs https://sentry.io/organizations/shields/issues/1867642393


calebcartwright (approved):
Nice sleuthing 👍

PR #6234, opened 2021-03-01
remove script for adding/removing heroku dynos

chris48s:
This reverts the changes in https://github.com/badges/shields/pull/6131 as it has moved to _the other place_
I've cancelled the scheduled jobs in heroku but I'll keep an eye on the new jobs to make sure they are definitely doing the business as expected before we merge this.


calebcartwright (approved):

chris48s:
seems to be smooth sailing with the new approach :ship:

PR #6256, opened 2021-03-10
send custom user agent when using got

### LINKED ISSUE #6251: Different user agent
This is not particularly a bug, but might also not necessarily be intended "behavior". The `User-Agent` header in requests performed by Shields against custom JSON endpoints used to start with `Shields.io/`, followed by a version number, if I recall correctly. After #6160, the user agent now reads `got (https://github.com/sindresorhus/got)` instead. Is this intended? 

I came across this issue as [wakapi](https://github.com/muety/wakapi), which I am the maintainer of, used to check that header in incoming requests to its Shields.io-compatible for a certain pattern to conditionally let the request succeed or not. Although that check didn't give a lot of benefit anyway, it makes a good example of how the change may cause side effects on other projects.

Just wanted to bring this up in case it was a mistake :)

chris48s:
There was always going to be something that slipped through the cracks.
Closes #6251

### DIFF
--- a/core/base-service/got.js
+++ b/core/base-service/got.js
@@ -3,6 +3,8 @@
 const got = require('got')
 const { Inaccessible, InvalidResponse } = require('./errors')
 
+const userAgent = 'Shields.io/2003a'
+
 function requestOptions2GotOptions(options) {
   const requestOptions = Object.assign({}, options)
   const gotOptions = {}
@@ -49,6 +51,8 @@ async function sendRequest(gotWrapper, url, options) {
   const gotOptions = requestOptions2GotOptions(options)
   gotOptions.throwHttpErrors = false
   gotOptions.retry = 0
+  gotOptions.headers = gotOptions.headers || {}
+  gotOptions.headers['User-Agent'] = userAgent
   try {
     const resp = await gotWrapper(url, gotOptions)
     return { res: resp, buffer: resp.body }
--- a/core/base-service/got.spec.js
+++ b/core/base-service/got.spec.js
@@ -82,6 +82,21 @@ describe('got wrapper', function () {
     )
   })
 
+  it('should pass a custom user agent header', async function () {
+    nock('https://www.google.com', {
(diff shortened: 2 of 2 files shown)


calebcartwright (approved):

core/base-service/got.js
```diff
@@ -3,6 +3,8 @@
 const got = require('got')
 const { Inaccessible, InvalidResponse } = require('./errors')
 
+const userAgent = 'Shields.io/2003a'
```
chris48s, line 6: I have no idea what this number means. I just copied it from https://github.com/badges/shields/blob/2359eb278beb4612ea8e579b062648463dc79813/core/base-service/legacy-request-handler.js#L14

I decided to just duplicate this in `got.js` instead of importing it from legacy-request-handler as I think we can get rid of a lot of what is in there in the near future

chris48s: That.. sort of makes sense for self-hosted installs (unless they track next/master), although we don't pin our own deploys to a version, by which I mean

we tag server-YYYY-MM-DD and deploy
we make some more commits and deploy those too
now the version we're running in production isn't server-YYYY-MM-DD - its a future version that doesn't correspond to any tag

I suppose `Shields.io/server-YYY-MM-DD` would at least mean `server-YYY-MM-DD` (or later)

paulmelnikow: Sure. We could also use `server-YYY-MM-DD` for the Docker releases and includ something like `server-YYY-MM-DD+dev` for production. 

Reading this thread makes me think we should include some extra stuff in the query string on request for the dynamic + endpoint badges, or when a user-specified URL is used, since it allows a service provider to differentiate between our official badge requests, and other requests which are being specified by the user.

chris48s: I've made an issue for this as I suspect we're not going to pick this up immediately and this conversation will disappear into the github ether. https://github.com/badges/shields/issues/6268

PR #6257, opened 2021-03-11
use got to push influx metrics

chris48s:
refs  #4655

Nice bit of low hanging fruit here :green_apple: 
Because this is stand-alone utility code and not part of the services I haven't routed it through `fetchFactory`/`sendRequest`. I've just gone straight from request to got.

### DIFF
--- a/core/server/influx-metrics.js
+++ b/core/server/influx-metrics.js
@@ -1,8 +1,6 @@
 'use strict'
 const os = require('os')
-const { promisify } = require('util')
-const { post } = require('request')
-const postAsync = promisify(post)
+const got = require('got')
 const generateInstanceId = require('./instance-id-generator')
 const { promClientJsonToInfluxV2 } = require('./metrics/format-converters')
 const log = require('./log')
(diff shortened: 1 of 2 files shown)


core/server/influx-metrics.js
```diff
@@ -15,21 +13,19 @@ module.exports = class InfluxMetrics {
...
     const request = {
-      uri: this._config.url,
+      url: this._config.url,
       headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
       body: await this.metrics(),
       timeout: this._config.timeoutMillseconds,
-      auth,
+      username: this._config.username,
+      password: this._config.password,
+      throwHttpErrors: false,
```
calebcartwright, line 23: this just prevents `got` from throwing on some non-successful response status code right?

calebcartwright (approved):
One question for my own edification, but LGTM

core/server/influx-metrics.js
chris48s: Yep. By default got throws an exception if it gets a 404, 500 or whatever. This tells it to return a response. I mainly did this because the existing error handling code assumes this behaviour from request, but I could have rewritten the error handling code here

https://github.com/badges/shields/blob/fd7eddc7bb406b272b8bbf4155315c5f3b198a03/core/server/influx-metrics.js#L33-L45

to handle both cases in the catch block instead. I don't think it makes much difference either way though.

Issue #5883, opened 2020-11-27
Third Party Requests to Google Fonts


samuel-p:
:clipboard: **Description**

At the moment shields does third party requests to Google. CDNs are bad from a privacy and security perspective. So please replace them completely with self-hosted files or at least provide an option to disable it for self-hosted instances. You can download the fonts from [google-webfonts-helper](https://google-webfonts-helper.herokuapp.com/fonts) if you want to self-host it.

![Bildschirmfoto von 2020-11-28 00-16-20](https://user-images.githubusercontent.com/7959516/100488787-90973b00-3110-11eb-8d2f-240b1d4e5986.png)

https://webbkoll.dataskydd.net/en/results?url=http%3A%2F%2Fshields.io%2F#requests

calebcartwright:
Relevant discussion in #4864 and #3026

SISheogorath:
I just added a new version of my previous PR that resolves the fonts using NPM packages instead of manually maintaining the fonts inside the project.

I hope this will help to resolve the situation.

PyvesB:
Aforementioned pull request is now live on our servers, thanks again @SISheogorath  for the contribution and @samuel-p for raising the issue. Do let us know if you have any further concerns by opening a new issue, for now I'll close this one. 😉

PR #6274, opened 2021-03-16
Use NPM packages to provide fonts instead of Google Fonts

SISheogorath:
This patch uses the [fontsource libraries](https://github.com/fontsource/fontsource) instead of using Google Fonts
in order to provide the fonts in the frontend. This way the web frontend
becomes more private, since it no longer shared visitors IPs with Google
and might even loads faster since modern web browsers isolate site
caches from each other and can re-use connections with HTTP/2 instead of
setting up a new connection to Google.[1]

It's basically the next iteration of my previous PR[2], that no longer
relies on a tool to download the fonts separately, but uses a library
instead. It also relates to issues that have been opened before[3] and
should be easily handled and upgraded by the regular NPM package
tooling[4], solving all related problems to that.

1: https://csswizardry.com/2019/05/self-host-your-static-assets/
2: https://github.com/badges/shields/pull/4864
3: https://github.com/badges/shields/issues/5883
4: https://github.com/badges/shields/labels/dependencies

**Note:** I'm not sure what to put into the brackets for the PR title, but I hope it's fine this way.

### DIFF
--- a/frontend/components/meta.tsx
+++ b/frontend/components/meta.tsx
@@ -2,6 +2,8 @@ import React from 'react'
 import { Helmet } from 'react-helmet'
 // @ts-ignore
 import favicon from '../images/favicon.png'
+import '@fontsource/lato'
+import '@fontsource/lekton'
 
 const description = `We serve fast and scalable informational images as badges
 for GitHub, Travis CI, Jenkins, WordPress and many more services. Use them to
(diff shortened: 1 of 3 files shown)


PyvesB (approved):
Thanks for the contribution "SISheogorath! 👍🏻 

@badges/shields-ops briefly discussed this in a meeting, we have no fundamental concerns with the proposed approach.

I've checked things in the Heroku review app, and could not spot any regressions. The [fontsource](https://github.com/fontsource/fontsource) project seems quite well organised and in a good shape. I'm going to go ahead and merge this.

PR #6277, opened 2021-03-16
Revert request --> got migration (for now)

chris48s:
We just talked about this on a call, but for the benefit of our future selves and those following along at home:

We deployed the PR migrating the majority of our HTTP requests from request to got. Since that deploy our baseline memory usage has increased and we've been hitting the memory limit in heroku more frequently (this wasn't _never_ happening before, but the frequency has definitely increased). We think got is probably the cause. To test this, we're going to revert these two commits, go back to request for a few days and monitor the memory usage. If the baseline goes down and we stop hitting the limit so regularly we know we are right.

If that is the case, maybe got isn't right for our needs. The next step would be to investigate if there is another modern/maintained HTTP client suitable for our needs which is more optimised for reducing memory usage. We don't really know the answers to those questions yet.

A few notes..
- I've opted not to bother reverting https://github.com/badges/shields/pull/6257 as this is not on the "critical path" so shouldn't be a deciding factor in performance but if we end up using axios/node-fetch/whatevz we can revisit that and bring it into line.
- I decided to just wholesale revert all of the changes in #6160 and #6256 using `git revert` instead of trying to unpick the changesets at all. I'll deal with unpacking what out of it we might want to keep (or not) when we see what effect this has on performance.
- I'll aim to deploy this change in isolation as far as possible. Hopefully the review will be quick and I can do it tonight or tomorrow, but if it ends up hanging around until the weekend I'll deploy it a few days after the weekly package bump-fest to try and minimise the number of things changing at once.


PyvesB (dismissed):
By reverting wholesale those two commits, got becomes a dev dependency in _package.json_, despite in reality being used by the Influx metrics posting. I don't think it's a problem in our case as got is being transitively pulled in by other runtime dependencies, but I thought I would mention that observation anyway.

chris48s:
No its a good spot and it should be there explicitly rahter than relying on it being a transitive dependency.

Bit of an unrelated tangent, but when I installed it with NPM 7 I noticed I got

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'shields.io@0.0.0',
npm WARN EBADENGINE   required: { node: '^12.18.3', npm: '^6.14.7' },
npm WARN EBADENGINE   current: { node: 'v12.21.0', npm: '7.6.1' }
npm WARN EBADENGINE }
```

so I modified the NPM engine statement

PyvesB (approved):

chris48s:
I'll deploy this shortly and monitor the stats over the next few days

PR #1129, opened 2017-10-04
Update documentation

paulmelnikow:
I wrote a new readme and contributing guidelines, and took a rough pass through the rest of the documentation.

PR #6278, opened 2021-03-17

PyvesB:
This number hadn't been updated since #1129, back in 2017.


paulmelnikow (dismissed):
🎉

calebcartwright:
I'd duplicated this number over in https://github.com/badges/shields/blob/master/doc/releases.md#shieldsio-service so we should probably update there too (or remove I suppose)

AstroOrbis (dismissed):

PyvesB:
> 
> 
> I'd duplicated this number over in https://github.com/badges/shields/blob/master/doc/releases.md#shieldsio-service so we should probably update there too (or remove I suppose)

I've left the more precise number on the homepage but used a slightly more general formulation in the releases doc, let me know what you think.

calebcartwright (approved):

Issue #6276, opened 2021-03-16
YouTube tests intermittently failing


PyvesB:
:clock11: **When did the problem start?**

Unsure on the exact date, the errors have been intermittent for several months as far as I can tell. Some problems were already reported here: #5842

:camera: **Live badge**

![](https://img.shields.io/youtube/likes/abBdk8bSPKU?style=social)

https://img.shields.io/youtube/likes/abBdk8bSPKU?style=social

:wrench: **Is the live badge working?**

The live badge is intermittently failing as well.

:link: **CircleCI link**

https://app.circleci.com/pipelines/github/badges/daily-tests/833/workflows/eeeaf314-50d4-4e5c-83fe-c2f02c5d7197/jobs/1731

:beetle: **Stack trace**

```
AssertionError: label mismatch: expected 'youtube' to equal 'votes'
    at Function._expectField (core/service-test-runner/icedfrisby-shields.js:69:51)
    at IcedFrisbyNock.<anonymous> (core/service-test-runner/icedfrisby-shields.js:53:26)
    at IcedFrisbyNock.<anonymous> (node_modules/icedfrisby/lib/icedfrisby.js:954:10)
    at invokeNextHook (node_modules/icedfrisby/lib/icedfrisby.js:1003:24)
    at /home/circleci/project/shields/node_modules/icedfrisby/lib/icedfrisby.js:1017:7
    at new Promise (<anonymous>)
    at IcedFrisbyNock._runHooks (node_modules/icedfrisby/lib/icedfrisby.js:976:12)
    at IcedFrisbyNock.run (node_modules/icedfrisby/lib/icedfrisby.js:1276:20)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Context.<anonymous> (node_modules/icedfrisby/lib/icedfrisby.js:1348:9)
```

:bulb: **Possible solution**

We seem to be hitting the YouTube API rate limits. According to the API console:
![youtube-errors](https://user-images.githubusercontent.com/10694593/111362154-5f725480-8686-11eb-9929-332aa57ef21e.png)

The badges have gotten quite popular given that there are over 20000 requests per day with badges that were added less than a year ago!

We will basically need to follow the steps I documented here: https://github.com/badges/shields/issues/5101#issuecomment-630330420

PyvesB:
[...]

PR #6304, opened 2021-03-20
Address compliance issues with [YouTube] API Services Terms and Policies

PyvesB:
As part of #6276, I've submitted a request to increase our API quota for YouTube badges. As a response, they've raised three concerns about potential violations of the YouTube terms and services:
* they are unsure whether we're using one or multiple Google projects IDs for our API client. We're only using one, so I'll confirm that with them. This point should be resolved easily.
* they asked us to display a link to their Terms of Service:
  > API Clients must display a link to YouTube's Terms of Service (https://www.youtube.com/t/terms), and they must also state in their own terms of use that, by using those API Clients, users are agreeing to be bound by the YouTube Terms of Service.

  This seems like a fair thing to ask for, so I've done so in the documentation of all four badges.
* they reckon that the "YouTube Video Votes" badge (and only that one) violates the following:
  > API Clients must clearly indicate how user-provided data will be used on YouTube

  You're probably confused at this point, they do clarify things a little, the problem is that we've apparently changed the labels for Likes and Dislikes.  Despite being exactly what they use in their own UI, 👍 and 👎 are not actually the _labels_. Consequently, I've phased out the "Votes" word and clarified in the documentation what 👍 and 👎 map to in the YouTube label world.

  They give an example of non compliant behaviour in their Terms of Service: an API Client with a text box called "feedback" which would end up posting a comment on YouTube without that being clear to the user. This makes perfect sense, but it really doesn't seem relevant to the Shields.io case. In my opinion, they are taking things a little too far and I don't really agree with their interpretation of the ToS, but the proposed changes will hopefully resolve this issue.

### DIFF
--- a/services/youtube/youtube-base.js
+++ b/services/youtube/youtube-base.js
@@ -5,6 +5,10 @@ const { BaseJsonService, NotFound } = require('..')
 const { metric } = require('../text-formatters')
 const { nonNegativeInteger } = require('../validators')
 
+const documentation = `
+<p>By using the YouTube badges provided by Shields.io, you are agreeing to be bound by the YouTube Terms of Service. These can be found here:
+<code>https://www.youtube.com/t/terms</code></p>`
+
 const schema = Joi.object({
   items: Joi.array()
     .items(
@@ -20,7 +24,7 @@ const schema = Joi.object({
     .required(),
 }).required()
 
-module.exports = class YouTubeBase extends BaseJsonService {
+class YouTubeBase extends BaseJsonService {
   static category = 'social'
 
   static auth = {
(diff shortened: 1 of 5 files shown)


calebcartwright:
> This makes perfect sense, but it really doesn't seem relevant to the Shields.io case. In my opinion, they are taking things a little too far and I don't really agree with their interpretation of the ToS, but the proposed changes will hopefully resolve this issue

I agree (assuming I'm following things correctly :sweat_smile:) I could see this being applicable if the badge were like our Twitter followers badge where clicking the badge could result in an action because the embedded badge links to the platform in that way. Here the click just opens the video.

That being said, also agree it's easier to just make the changes; don't think this ends up being detrimental to the badges or anything

calebcartwright (approved):

PyvesB:
Thanks for reviewing promptly. I've deployed these changes to production and followed up with YouTube, hopefully they'll increase our API quota soon and I'll be able to close  #6276. :)

PR #6306, opened 2021-03-22
Fix raster example

GrahamSH-LLK:
This changes the example so that you don't get a 404.


calebcartwright (approved):
Oops, good spot. Thanks!

GrahamSH-LLK:
It looks like everything is passing except for `ci/circleci: package`, which timed out and I don't think is my fault.

PR #6327, opened 2021-03-27
Add [PingPong] Service

bluecap-se:
Adds two new badges for [PingPong Status Page](https://pingpong.one) service:

- Uptime (last 30 days)
  Uptime
- Status
  Uptime


calebcartwright:
hi, thanks for the PR! Could you take a look and address the failing CI checks?

They seem to be linter related, and you should be able to reproduce locally via `npm run lint`

bluecap-se:
Thanks for your feedback @calebcartwright! Please have another look after the fixes.

calebcartwright:
Can you tell us a little more about the API token? I assume that it's absolutely necessary in order to fetch the data?

Can this token have it's permissions/auth. scoped to a read-only mode, or does it inherently provide other permissions as well?

bluecap-se:
Of course. The API token is a public read-only token that's specific for status/uptime badges, making it safe for public sharing. It's required to the fetch the data, otherwise the backend returns 404. It has no other permissions and is separate from the API key with full permissions that's used for the [REST API](https://developers.pingpong.one/).

services/pingpong/pingpong-uptime.service.js
```diff
@@ -0,0 +1,62 @@
+'use strict'
+
```
calebcartwright, line 2: Same general theme for this service class as the status one

services/pingpong/pingpong-uptime.tester.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
+
+const Joi = require('joi')
+const t = (module.exports = require('../tester').createServiceTester())
+
```
calebcartwright, line 5: Same general theme for this tester as the one for the status tester. Another benefit of using the standard/consistent Shields approach for setting badge attributes is that we'll control what the message can be, and can accordingly reuse our test validator schemas

calebcartwright (commented):
Thanks for the PR and additional info! It seems reasonable to incorporate these badges, though I think we've got an opportunity to simplify the implementation somewhat by reusing some standard Shields helpers.

Have left specific details inline

services/pingpong/pingpong-status.service.js
```diff
@@ -0,0 +1,62 @@
...
+const schema = Joi.object({
+  message: Joi.string().required(),
+  color: Joi.string().required(),
+}).required()
+
+const pingpongDocumentation = `
+<p>
+  To see more details about this badge and obtain your api key, visit
+  <a href="https://my.pingpong.one/integrations/badge-status/" target="_blank">https://my.pingpong.one/integrations/badge-status/</a>
+</p>
```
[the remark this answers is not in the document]
bluecap-se: Users are presented with the API key on that page, without further action.

services/pingpong/pingpong-status.service.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  static render({ message, color }) {
+    return {
+      label: 'status',
+      message,
+      color,
+    }
+  }
```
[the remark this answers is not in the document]
bluecap-se: Very good points. In the update, I've referenced [uptimerobot](https://github.com/badges/shields/blob/master/services/uptimerobot/uptimerobot-status.service.js#L25) as it seemed to match this use case; nodeping example doesn't have all status types (`issues`, `maintenance` being the case here) but it pointed me in the right direction - thanks for that! If there's a helper function in Shields that supports all status types, I'd be happy to apply it here.

services/pingpong/pingpong-status.service.js
```diff
@@ -0,0 +1,68 @@
+'use strict'
+
+const Joi = require('joi')
+const { BaseJsonService, InvalidParameter } = require('..')
+
+const schema = Joi.object({
+  status: Joi.string().required(),
+}).required()
```
calebcartwright, line 8: I know that the we've changed to a different schema (`status` vs `message` + `color`) which is good, though my prior point about utilizing the schema for validation still holds true. I'd like to see the `status` value here in the schema restricted to the known/supported statuses, or if we're going to keep this wide open to accept any string then we need to fail loudly (throw an `InvalidResponse`) in the default switch case in the render function.

services/pingpong/pingpong-status.service.js
```diff
@@ -0,0 +1,68 @@
...
+      case 'Operational':
+        return { message: 'up', color: 'brightgreen' }
+      case 'Major issues':
+        return { message: 'issues', color: 'orange' }
+      case 'Critical state':
+        return { message: 'down', color: 'red' }
+      case 'Maintenance mode':
+        return { message: 'maintenance', color: 'lightgrey' }
+      default:
+        return { message: 'unknown', color: 'lightgrey' }
```
calebcartwright, line 52: This pairs with my comment above on the schema

Any occurrence that hits this branch would mean that something's changed upstream that this service class in Shields isn't prepared to handle. That's an error scenario that we want to acknowledge loudly, whereas returning a non-error badge with an `unknown` message could give the user the incorrect impression that PingPong doesn't know the status of their site

services/pingpong/pingpong-uptime.tester.js
```diff
@@ -0,0 +1,27 @@
...
+// Example matches: 87.33%, 100%, 51%
+const isCorrectMessage = Joi.string().pattern(
+  /\b(?<!\.)(?!0+(?:\.0+)?%)(?:\d|[1-9]\d|100)(?:(?<!100)\.\d+)?%/
+)
```
calebcartwright, line 9: I feel like we should be able to replace this with our `isPercentage` helper from `test-validators` and avoid having another  regex to maintain. Could you give that a try, or explain why we need a separate one here?

services/pingpong/pingpong-uptime.tester.js
```diff
@@ -0,0 +1,27 @@
...
+
+t.create('PingPong: Uptime (unexpected response)')
+  .get('/sp_key.json')
+  .intercept(
+    nock =>
+      nock('https://api.pingpong.one')
+        .get('/widget/shields/uptime/sp_key')
+        .reply(200, '{"uptime": "hundred"}') // value should be a number
+  )
+  .expectBadge({ label: 'uptime', message: 'invalid response data' })
```
calebcartwright, line 27: Thanks again for the thoroughness of the tests but please remove this one as it's unnecessary and duplicative

calebcartwright (commented):
Thank you for the changes and apologies for the extended delay on the second review; this one fell off the radar!

This is in much better state now, though with a few comparatively minor things that'll need to be adjusted before merging

services/pingpong/pingpong-status.service.js
bluecap-se: Sounds good. In order to keep the checks DRY, the render function now throws an error.

calebcartwright:
Updates look good thanks! Could you rebase on master? We have the up-to-date branch gate enabled so will need to get this in sync with master before it can be merged, but should be good to go afterwards

PyvesB:
Great to see this work completed, thanks @bluecap-se  and @calebcartwright for your work! :)

calebcartwright:
Oops, looks like another PR got merged inbetween and this is out of date again. Will require another rebase on your end sorry. IIRC you should also be able to check the box to allow maintainers to update the PR which would allow us to sync the branches for you and make it a bit easier to merge

bluecap-se:
@calebcartwright The branch should be updated again. The option to allow maintainers doesn't seem to exist on this PR.

PyvesB (approved):
Strange that the option doesn't exist on this PR. Let's get this merged before it's too late again! :)

PyvesB:
These badges are now live! For example: https://img.shields.io/pingpong/status/sp_2e80bc00b6054faeb2b87e2464be337e 



🍾

bluecap-se:
@calebcartwright @PyvesB Great to see this live, thanks for all your feedback and assistance! 🚀 

Our Twitter handle is [@pingpong_status](https://twitter.com/pingpong_status)

As the creator of PingPong status pages I'd like to present our offering to all open source projects: [free Pro plan](https://pingpong.one/pricing/). If this sounds interesting for Shields, start by [creating a Pro account](https://my.pingpong.one/account/signup/?plan=pro) and then [ping us](mailto:support@pingpong.one?subject=Open-source%20setup) to activate the open source offer.

bluecap-se:
Sorry to see https://status.shields.io still being down, let us at [PingPong](https://pingpong.one) know if we can assist in any way.

Issue #4365, opened 2019-11-27
Badge Request: REUSE compliance


mxmehl:
:clipboard: **Description**

It would be great to have a badge provided by shields.io for showing a project's compliance with the [REUSE best practices](https://reuse.software) that make licensing easier for everyone.

The project already has a service providing a badge but perhaps it's possible to also provide that over shields: https://api.reuse.software. It also offers a JSON endpoint. The overview page of a project looks a little something like this: https://api.reuse.software/info/github.com/fsfe/reuse-tool

The API requires a simple registration of a project.

:link: **Data**

The API is publicly accesisble, but only for repositories that have been registered: https://api.reuse.software/status/github.com/fsfe/reuse-tool

The provided information is faily simple, so no extensive documentation required.

:microphone: **Motivation**

For most developers, software licensing is a pain. There a dozens of suggestions how you should provide licensing and copyright information in a Free Software project. REUSE changes that by offering simple and logical best guidelines that make it easy to tag every file with the required information so it's easy to understand for humans and machines alike.

The API shows on first sight whether the current status is OK. For more detailed checks (e.g. different branches or in CI) there is the REUSE helper tool: https://github.com/fsfe/reuse-tool

calebcartwright:
Thanks for the request @mxmehl! We have a great [tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) if you or anyone else is interested in implementing this badge.

tapanchudasama:
Hey maintainer, If this issue is still up for grabs, I'd like to pick it up.

mxmehl:
@tapanchudasama Sure, would be awesome! I won't be able to work on this in the next months.

chris48s:
Feel free :+1:  Our tutorial for adding new badges is at https://github.com/badges/shields/blob/master/doc/TUTORIAL.md

tapanchudasama:
Hey @chris48s, I have raised a [PR](https://github.com/badges/shields/pull/6330) for this. Please have a look.

mxmehl:
Thanks for working on this! For registered repos, it seems to work really well.

However, what about unregistered? They are labelled as such, but how do people learn about how to register them? Could we add a link to https://api.reuse.software/register ?

chris48s:
[...]

PR #6330, opened 2021-03-28
[REUSE] Add service badges

tapanchudasama:
This PR fixes issue #4365.


services/reuse/reuse-compliance.service.js
```diff
@@ -0,0 +1,59 @@
...
+const { BaseJsonService } = require('..')
+const { renderReuseBadge } = require('./reuse-compliance-helper')
+
+const responseSchema = Joi.object({
+  badge: Joi.string(),
+  hash: Joi.string(),
+  last_access: Joi.date(),
+  lint_code: Joi.number(),
+  lint_output: Joi.string(),
+  status: Joi.string().required(),
```
chris48s, line 12: If we can only deal with 2 values ('compliant' and 'non-compliant') we should validate that status is exactly one of those two values, not just any old string.
Also it looks like the only value we are actually using here is status. We don't need to validate all this other stuff. i.e: if we only need the status value and we get a response back with a valid status but no `url` key we shouldn't fail to render a badge because of it.

services/reuse/reuse-compliance.service.js
```diff
@@ -0,0 +1,59 @@
...
+        user: 'fsfe',
+        repository: 'reuse-tool',
+      },
+      staticPreview: this.render({ status: 'compliant' }),
+      keywords: ['license'],
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'REUSE',
```
chris48s, line 38: Label should be lower-case. See https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines

services/reuse/reuse-compliance.service.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  static render({ status }) {
+    return renderReuseBadge({ status })
+  }
+
+  async fetch({ gitserver, user, repository }) {
+    return await this._requestJson({
+      schema: responseSchema,
+      url: `https://api.reuse.software/status/${gitserver}.com/${user}/${repository}`,
```
chris48s, line 48: What's the deal with this `gitserver` param? Is there a list of compatible services, like this
https://github.com/badges/shields/blob/66a73b9f987685fb0a500bb0607d568d57deefb8/services/coveralls/coveralls.service.js#L22
or can any git remote be used? If so, is .com really a requirement?

services/reuse/reuse-compliance.service.js
```diff
@@ -0,0 +1,59 @@
...
+      errorMessages: {
+        400: 'Not a Git repository',
```
chris48s, line 50: :+1:

chris48s (commented):
thanks for having a look at this

services/reuse/reuse-compliance.service.js
tapanchudasama: The API provided by `REUSE` supports all repositories that use `git`.  And in the API format, the `.com` is necessary.

Here is the example to try: 
https://api.reuse.software/status/github.com/fsfe/reuse-tool 

Reference: https://api.reuse.software/register

services/reuse/reuse-compliance.service.js
chris48s: You can get rid of `badge`, `hash`, `last_access`, `lint_code` and `lint_output` here too. None of them are used by the badge. We only want to validate that values we need to render the badge

services/reuse/reuse-compliance.service.js
chris48s: ok, but if my git remote is not on a `.com` domain (`.org`, `.net` etc e.g: https://gitlab.gnome.org ) then this isn't going to work.

There's another case that may give us a problem which is where your git remote is not hosted at the root of the domain. I don't have an example of this off the top of my head to see how REUSE handles it, but I'll try and find one tomorrow to check (if anyone happens to know one please chime in - this may have come up before). We might have to convert this to a URL param as we have a hard limitation from path-to-regexp that we can't have a capturing group capable of containing a slash unless it is the last capturing group in the route.

services/reuse/reuse-compliance.service.js
mxmehl: FWIW, there are [4 statuses](https://git.fsfe.org/reuse/api/src/branch/master/reuse_api/badges): compliant, non-compliant, checking and unregistered

chris48s: Any (other) valid values should also be

* validated in the schema and
* added to `COLOR_MAP`

services/reuse/reuse-compliance.service.js
chris48s: I had a bit more of a think about this and I think the way that actually makes most sense if any git remote can be used (including something like https://git.eclipse.org/c/4diac/org.eclipse.4diac.ide.git/ where the clone URLs are under a subdir) is to have a single `:remote+` param (which can then contain an arbitrary number of slashes). So the code is something like:

```js
  static route = {
    base: 'reuse/compliance',
    pattern: ':remote+',
  }

  async fetch({ remote }) {
    return await this._requestJson({
      url: `https://api.reuse.software/status/${remote}`,
    })
  }
```

and then the badge URLs will be like:

* `/reuse/compliance/github.com/fsfereuse-tool`
* `/reuse/compliance/gitlab.gnome.org/GNOME/gnome-shell`
* `/reuse/compliance/git.eclipse.org/gitroot/4diac/org.eclipse.4diac.ide`

We can add little HTML fragment explaining the URL format in the `documentation` key of an `example` object.

tapanchudasama: This is a awesome idea!

services/reuse/reuse-compliance.tester.js
```diff
@@ -0,0 +1,55 @@
...
+
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('valid repo -- compliant')
+  .get('/github.com/fsfe/reuse-tool.json')
+  .expectBadge({
+    label: 'reuse',
+    message: 'compliant',
+    color: 'green',
+  })
```
chris48s, line 11: This is looking great now. Only one final thing to address and we are done:

In the service tests the reason we run integration tests against live APIs is because we use so many external services, we want tests that fail if the upstream API goes away or makes some change that we need to respond to. We run them all once a day. We will never notice this stuff otherwise. So for this reason we usually want to accept a range of valid values in the live tests. i.e: if `github.com/fsfe/reuse-tool` becomes non-compliant at some point in futre, it doesn't mean this code is broken - it just means something in this project we are using for testing changed.

Becuase the input values and output values are the same, lets pull out

```js
const isReuseCompliance = Joi.string()
  .valid('compliant', 'non-compliant', 'checking', 'unregistered')
  .required()
```

into `reuse-compliance-helper.js`

and then use it in the schema:

```js
const responseSchema = Joi.object({
  status: isReuseCompliance,
}).required()
```

and to validate the response in the integration test:

```js
t.create('valid repo')
  .get('/github.com/fsfe/reuse-tool.json')
  .expectBadge({
    label: 'reuse',
    message: isReuseCompliance,
  })
```

and then you could either add another mocked service test for the valid case, or you could just write a small unit test for the helper function in a `reuse-compliance-helper.spec.js` instead of doing it with mocked HTTP calls in the service tests. I'll leave it with you...

tapanchudasama: Yeah! That is a nice observation! I'll make these changes.

chris48s (approved):

chris48s:
Thanks for your first contribution :+1:

PR #6331, opened 2021-03-28
Upgrade to gatsby 3

chris48s:
d4a9460 isn;t strictly related to this PR, but it should prevent us flipping between lockfile version 1 and 2 depending on which npm version is used (we seem to have gone back to v1 at some point)


chris48s:
review app at https://shields-staging-pr-6331.herokuapp.com/
as far as I can tell, not actual code changes are needed

calebcartwright:
> [d4a9460](https://github.com/badges/shields/commit/d4a9460081e773b709ad1b61d411d41f200e6df9) isn;t strictly related to this PR, but it should prevent us flipping between lockfile version 1 and 2 depending on which npm version is used (we seem to have gone back to v1 at some point)

well that explains a lot. I'm using node 14 locally (via nvm) with the 6.x npm version that's bundled with it, and the steamroll rebasing I did in #6319 probably drove this.

i think node 15 is the first that ships with npm v7.x, so (assuming that's right) should we update our Contributing docs since we're at Node >=12 but npm >=7?

chris48s:
Hmm. I'm running node 12 with npm 7 (actually I use nvm to switch between node versions depending on project, but I usually develop on shields with node 12 because that is what we run in production). I tend to upgrade npm independently of node. Is that unusual? Npm 7 will run on any node >=10.

calebcartwright:
> I tend to upgrade npm independently of node. Is that unusual? Npm 7 will run on any node >=10

No I don't think so, but I'm trying to look at it with the first-time-contributor hat on, particularly with someone who just grabs the Node 12 installer/not using nvm, and/or is on Windows, etc.

I may have my node/npm installer versions mixed up, but if the node 12 installer _does_ ship with npm 6 then I think we probably want to add a note somewhere around https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#node-npm to say that we're requiring a minimum node 12 and npm 7

chris48s:
Seems fair enough. Worth noting that when you `npm install` or `npm ci` if your npm version is lower than the one in engines it will give you a warning, but I think it does still install. Worth mentioning in the docs though :+1:

calebcartwright (approved):
:+1:

PR #6364, opened 2021-04-04
add setting which allows us to set a timeout on HTTP requests

chris48s:
There are a few reasons why I think this is a useful thing to do:

- There are certain classes of denial of service attack that can cause us to consume a lot of resources e.g: causing us to greedy-evaluate a regex that takes a really long time to evaluate or causing us to parse an XML/JSON payload which takes a really long time to parse. Setting a much lower limit on the amount of time we are willing to spend responding to a single request gives us a crude blanket defence against this class of problem.
- In node 12 the default value for this is 120 seconds. In node 13 and higher, the default will change to 0 (no timeout at all) so it will become more important to explicitly set some limit on this. See
  - https://nodejs.org/docs/latest-v12.x/api/http.html#http_server_timeout
  - https://nodejs.org/docs/latest-v13.x/api/http.html#http_server_timeout
- In situations where we are consuming a lot of resources and every request is taking a long time this should help us to recover more quickly.

I think the value we should pick for this in production should be something more like 10 seconds instead of the 120 we are currently defaulting to, but open to thoughts on this. We can always experiment a bit. We probably want a lower limit in production than in dev/test environments.

Another thing we could do here is instead of just closing the connection with no response we could handle the `timeout` event and return something like a `408 Request Timeout` response. This is probably cleaner than just closing the connection with no response, but I've decided not to do it for now because once we start issueing a response in this case it has an impact on the way we are tracking errors in heroku - it is no longer a "Connection closed without response". I think at least in the first case it is helpful not to change that. Maybe if we start closing the connection after 10 seconds instead of 2 mins, that will just become noise, but lets minimise the number of things we are changing at once.


config/default.yml
```diff
@@ -33,6 +33,8 @@ public:
 
   fetchLimit: '10MB'
 
+  requestTimeoutSeconds: 120
```
chris48s, line 36: I set the default to 120 here so there is no behaviour change for self-hosting users, but we could make this a bit lower if we want as this is also the default for dev/test. We can set our own limit for production using `REQUEST_TIMEOUT_SECONDS`.

core/server/server.js
```diff
@@ -137,12 +138,11 @@ const publicConfigSchema = Joi.object({
     teamcity: defaultService,
     trace: Joi.boolean().required(),
   }).required(),
-  cacheHeaders: {
-    defaultCacheLengthSeconds: Joi.number().integer().required(),
-  },
+  cacheHeaders: { defaultCacheLengthSeconds: nonNegativeInteger },
```
chris48s, line 141: This change is not strictly related to this PR but I noticed it here. I don't think it ever makes sense to say cache something for a negative number of seconds.

calebcartwright (approved):
I agree it's useful to incorporate this feature and the changes LGTM. Also in agreement on having future discussions around what values to use in which environments :+1:

PyvesB (approved):
Change looks good to me, experimenting with a 10 second value sounds sensible. :+1:

chris48s:
Cool :+1:  Having thought about it I think I'm going to deploy this, set it to 20 initially and then crank it down from there

PR #6369, opened 2021-04-08
Support git hash in [nexus] SNAPSHOT version

seglo:
The SNAPSHOT version pre-check wouldn't match a SNAPSHOT version that includes a hash (git hash, a default convention when using [`sbt-dynver`](https://github.com/dwijnand/sbt-dynver)). Incidentally, it would match versions with hashes that end with `0-9`, but not with `a-f`.

There was a test in `nexus.tester.js` that tested a version with a hash that passed because the hash ended with `0-9`.

i.e.)

Match
```
7.0.1+19-8844c122-SNAPSHOT
```

No match
```
7.0.1+19-8844c12a-SNAPSHOT
```

### DIFF
--- a/services/nexus/nexus-version.js
+++ b/services/nexus/nexus-version.js
@@ -1,7 +1,7 @@
 'use strict'
 
 function isSnapshotVersion(version) {
-  const pattern = /(\d+\.)*[0-9a-fA-F]-SNAPSHOT/
+  const pattern = /(\d+\.)*[0-9a-f]-SNAPSHOT/
   return version && version.match(pattern)
 }
 
--- a/services/nexus/nexus-version.js
+++ b/services/nexus/nexus-version.js
@@ -1,7 +1,7 @@
 'use strict'
 
 function isSnapshotVersion(version) {
-  const pattern = /(\d+\.)*\d-SNAPSHOT/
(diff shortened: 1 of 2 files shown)


ignasi35 (dismissed):

services/nexus/nexus-version.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
 function isSnapshotVersion(version) {
-  const pattern = /(\d+\.)*\d-SNAPSHOT/
+  const pattern = /(\d+\.)*[0-9a-fA-F]-SNAPSHOT/
```
PyvesB, line 4: Can the hash really include uppercase letters, i.e. can `A-F` be dropped from the regex?

PyvesB (commented):
Thanks for the contribution @seglo! One small observation, though I would by no means consider it a blocker. 😉

services/nexus/nexus-version.js
seglo: I've never seen it upper case for my usecases (with `sbt-dynver`).  I'll revert that.

PyvesB (approved):

PyvesB:
This improvement has been deployed to production! 🍾

seglo:
Confirmed it works. Thanks!

PR #6368, opened 2021-04-08
Add [Stryker] service

tiagoporto:
Add service for [stryker-mutator](https://stryker-mutator.io/)

Any suggestion for improvement will be very welcome.

Fix #1758

Issue #6307, opened 2021-03-22
New Badge: Upptime


benaltair:
:clipboard: **Description**

[Upptime](https://upptime.js.org/) is a GitHub-powered open-source uptime monitor and status page. It's quite novel, as it runs entirely on Github with actions and pages. It is already set up to provide API endpoints that are compatible with [shields.io](https://shields.io/). It would be great to have it officially implemented on the main website.

:link: **Data**

As [documented](https://upptime.js.org/docs/badges) Upptime already generates API endpoints that look similar to:
```json
{
  "schemaVersion": 1,
  "label": "uptime",
  "message": "100%",
  "color": "brightgreen"
}
```
The current method is to navigate to `https://raw.githubusercontent.com/[USER]/[REPO]/master/api/[SITE]/uptime.json`. For example this URL would work, using their demo:
https://raw.githubusercontent.com/upptime/upptime/master/api/google/uptime.json

Translating into: 
![Uptime](https://img.shields.io/endpoint?url=https%3A%2F%2Fraw.githubusercontent.com%2Fupptime%2Fupptime%2Fmaster%2Fapi%2Fgoogle%2Fuptime.json)

Here is the legend for the three fields that need to be filled (from the above template):
| Variable | Description |
| --- | --- |
| USER | The Github user or organization |
| REPO| The Github repo for the Upptime instance |
| SITE | The slug found in your `/api/` folder within the `master` branch of the Upptime repo |

:microphone: **Motivation**

Right now users have to use the endpoint method, prefixing their URLs with `https://img.shields.io/endpoint?url=`. Perhaps the pattern above could be incorporated into a documented [shields.io](https://shields.io/) schema. Most of the work is already done, and if possible incorporating the service directly would increase accessibility for devs just getting started. It can also help cross-pollinate users across the services. For instance, I first heard about shields.io through Upptime documentation not too long ago.

PyvesB:
Hello @benaltair ! 👋🏻 

Thank you for the suggestion. To be able to implement this on the main Shields.io website, we would need a proper JSON API rather than one that follows the endpoint badge structure and imposes its own fields on generated badges. See https://github.com/badges/shields/pull/5701#discussion_r513048352 and surrounding comments for a bit more context and explanations.

Is such an API already available for the Upptime service?

benaltair:
Hi @PyvesB thanks for the reply. From what I understand, the only API that Upptime provides would be the semi-dynamic (updated every 5 minutes by default) endpoints shown here:
[...]

PR #6370, opened 2021-04-08
Document stance on APIs following the endpoint badge schema

PyvesB:
We've had a few recent contributors wanting to implement native badges using Shields-specific APIs based on our endpoint badge schema (#6368, #6307 and #5701).

This is an attempt at documenting our stance on this matter.

@calebcartwright as you'll notice, I've slightly reused some of your wording here. 😉 

Let me know what you think!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,9 +107,15 @@ There are three places to get help:
 - Badges should not obtain data by scraping web pages - these are likely to break frequently.
   Whereas API publishers are incentivised to maintain a stable platform for their users,
   authors of web pages have no such incentive.
+- Badges should not use a specialised API which is specifically designed to set Shields.io badge
+  attributes. For that purpose, the [endpoint badge][] should be used instead. Where possible, the
+  badge code should be responsible for computing its own attributes, leveraging returned data rather
+  than being imposed all attributes by the API.
 - Badges may require users to specify a token in the badge URL as long it is scoped only to
   fetching information and doesn't expose any sensitive information. Generating a token with the
   correct scope must be clearly documented.
+
+[endpoint badge]: https://shields.io/endpoint
 
 ## Badge URLs
 


chris48s:
I kind of know what this is trying to say, but it seems like an oblique way of approaching it and this doesn't really explain _why_ we are saying this. I don't think we are really saying if an upstream service has no use for an API but adds an JSON endpoint specifically for us surfacing the info we need for an integration (but not using the endpoint schema) we wouldn't use it, are we? I think that's pretty much what you told Localazy they had to do to get a PR accepted.. i.e: I think thing thing we don't want to do is potentially take on maintenance of a wrapper/integration test for every endpoint badge is the only 'value-add' is to give it a nicer URL, right?

It is kind of a hard thing to phrase (I'm also struggling slightly). What if we frame it something more like:

The [JSON endpoint badge](https://shields.io/endpoint) is a allows users to generate bespoke badges with the full power of Shields' badge customization. If a service already provides badges using the JSON endpoint feature, the upstream service and their end users already have access to the full range of shields features. We won't add wrappers for an existing endpoint badge just to give it a 'friendly' URL.

dunno if that is any better?

PyvesB:
Thanks for the feedback. Will give it another go.

calebcartwright:
Thanks for starting this dialog. It's definitely been recurring with enough frequency that IMO it'll be helpful to have something where we can point folks (#6327 may be another example). Have been ruminating on this one a bit so apologies for the lack of brevity :smile: 

I know we're all well aware of this fact, but it's probably also worth noting for posterity that the Shields.io site _has_ turned into a place where developers can discover services that are useful for development/building/analysis/etc. Accordingly there's some understandable motivation for the service maintainers/providers to get native badges included so that their badges are listed and discoverable (I know I came across a couple services I use regularly which I discovered via Shields.io)

I suspect part of the problem is that there's a trend of folks confusing/conflating the Endpoint badge schema as "the" way to get natively integrated in the Shields ecosystem, which isn't really the point of the Endpoint badge.

There's some badge requests that simply aren't feasible to include as native badges, for example those that aren't developer/technologist related services, require user/account-specific authorization details, require page scraping, etc. and which we'd definitely decline/defer to an Endpoint badge. At the same time as Chris noted above, I agree that we wouldn't necessarily want to refuse to add a native badge (at some point) just because the corresponding service's API contract seems based on the Endpoint schema.

I feel like what we want to convey is that we want anything submitted as a native badge to handle the rendering aspects, and only utilize the API response to get the corresponding data point for the particular badge(s) in question (build result, numerical code coverage, etc.) Basically, native badges should utilize the upstream service for the "what", not the "how", if we're dealing with a coverage badge, we want the numerical code coverage value from the upstream source, but Shields doesn't care what the upstream service thinks the color of the badge should be. It doesn't matter to us if the upstream service happens to include aspects of the "how" (badge color, label, etc.) in the response, but those should be ignored. Additionally, we don't want our native badges to have to do extra work to extract the data because of the API response structure is geared towards badge rendering concerns (e.g. for something like test results, we'd strongly prefer a response like `{ "passedTests": 9, "failedTests": 2 }` instead of dealing with something like `{ "message": "9 passed test(s), 2 failed test(s) }`

Maybe it'd be useful to have a doc that's general guidelines for getting a badge incorporated into the Shields ecosystem that's largely targeted at an audience including owners/maintainers of a service as well as an individual contributor that'd like to get a badge for their favorite service/tool. Basically a higher level "How to make a new badge happen" doc that can help folks pick a good path and then guide/point them to the other relevant documentation (i.e. use an Endpoint badge if A, B, or C is true in your case and link out to Endpoint, if native badge then links to guideline, tutorial, API stance, etc.)

I'd think within that type of guide we could codify something relative to our stance on the API structure and contracts, and what the threshold(s) would be for a native badge vs. when an Endpoint (or using our npm package or even just following our badge specification). 

Thoughts?

PyvesB:
I _think_ we're all fundamentally on the same page. The tricky part if how to convey our guidance in a clear and concise manner, which this PR may have failed to do - though I'm still happy to give it another go based on Chris' suggestions. :)

I'm not against a "How to make a new badge happen" doc per say, but in my opinion we should only do that as part as a slightly wider rework of our documentation. Right now our guidance on contributing new badges to Shields.io in scattered all around the place:
* general guidance and aesthetics in the [spec](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md).
* slightly more prescriptive guidance [here](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines).
* conventions on URL structure [here](https://github.com/badges/shields/blob/master/doc/badge-urls.md).
* guidance on schema building [here](https://github.com/badges/shields/blob/master/doc/input-validation.md).
* the main tutorial which points to some of the above but not all, [here](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md).
* the separate service tests guide [here](https://github.com/badges/shields/blob/master/doc/service-tests.md).
* some help to get a dev environment set up in the [README](https://github.com/badges/shields/blob/master/README.md#development).
* other places which I've omitted?

Some of these documents are quite lengthy and it may be a bit of a maze to navigate them for new contributors. I guess what I'm saying is that I wouldn't want yet another separate document, as it would potentially make things harder for new contributors and increase the chances of them TLDRing anyway.

Before we add any more, are there opportunities to better streamline, organise or summarise our existing documentation?

calebcartwright:
> Right now our guidance on contributing new badges to Shields.io in scattered all around the place:

Completely agreed, and this is largely the problem I was suggesting we attempt to solve with a singular entry point. The list you collated shows the breadth of topics that have detailed documentation (which makes sense to me given the quantity of topics an depth of detail), but I feel like we're missing a standard start/entry point we could send everyone to.

The concern I'd have is that in the current state it's quite conceivable for a new contributor to miss something important unless they happened to start with on particular doc or read them all

PyvesB:
I'll close this for now and slowly rework some of the docs, keeping in mind some of the above thoughts.

PR #6371, opened 2021-04-08
return a 408 Request Timeout response for timed out requests

chris48s:
In #6364 I added the ability to time out requests that take longer than some threshold. This is useful, but it generates a lot of noise in the monitoring.
Throwing a `408 Request Timeout` here should clean this up. Requests that are closed by this rule can be filtered with `heroku logs -a <appname> --tail | egrep --line-buffered "status=408"`


core/server/server.js
```diff
@@ -477,6 +477,17 @@ class Server {
     this.registerServices()
 
     camp.timeout = this.config.public.requestTimeoutSeconds * 1000
+    if (this.config.public.requestTimeoutSeconds > 0) {
+      camp.on('timeout', socket => {
```
chris48s, line 481: This is super-low level because the timeout event callback gives us a socket object, not a request/response.
I don't think there is a nice way to deal with this using a higher-level abstraction if we want the timeout to be applied to all requests. If we used `camp.setTimeout()` the callback param would also be a socket.

core/server/server.js
```diff
@@ -477,6 +477,17 @@ class Server {
...
+        socket.write('HTTP/1.1 408 Request Timeout\n')
+        socket.write('Content-Type: text/html; charset=UTF-8\n')
+        socket.write('Content-Encoding: UTF-8\n')
+        socket.write('Cache-Control: max-age=60, s-maxage=60\n')
```
chris48s, line 485: I agonised a bit about what to do here.
On the one hand, it is kind of odd to cache this because repeat requests for any URLs that throw this will return a `Request Timeout` response.. immediately 🙃 That's kind of odd, but requests that tie us up for `requestTimeoutSeconds` seem like exactly the kind of request we don't want to immediately try and process at the origin again. I decided to tell downstream cache it for a bit.

PyvesB: Makes sense, however, a minute does seem slightly high to me. Maybe we could make this configurable?

chris48s: On today's episode of the "naming things is hard" show, I've converted this to a snappily named setting `REQUEST_TIMEOUT_MAX_AGE_SECONDS` and switched the default to 30 seconds.

PyvesB (approved):
Looks good to me! 👍🏻

PR #6386, opened 2021-04-12
Fix Request Timeout responses

chris48s:
It is almost as if there is a good reason we don't usually write raw socket code :thinking: 

Turns out curl, got and firefox were all fine with `\n`. Heroku is not feeling it and throws a 503.

This now returns the correct error from heroku e.g:

```
curl --include "https://shields-staging-pr-6386.herokuapp.com/badge/dynamic/json.svg?label=CDN%20Data&url=https://yacdn.org/global-stats&query=$.cdnData&colorB=blue"
```

(I've set the `shields-staging-pr-6386` app up with a 5 sec timeout)

I think switching the timeouts to return this response was a red herring when it comes to the performance issue we saw today but if we hit a problem, we can roll back.

### DIFF
--- a/core/server/server.js
+++ b/core/server/server.js
@@ -481,11 +481,11 @@ class Server {
     if (this.config.public.requestTimeoutSeconds > 0) {
       camp.on('timeout', socket => {
         const maxAge = this.config.public.requestTimeoutMaxAgeSeconds
-        socket.write('HTTP/1.1 408 Request Timeout\n')
-        socket.write('Content-Type: text/html; charset=UTF-8\n')
-        socket.write('Content-Encoding: UTF-8\n')
-        socket.write(`Cache-Control: max-age=${maxAge}, s-maxage=${maxAge}\n`)
-        socket.write('Connection: close\n\n')
+        socket.write('HTTP/1.1 408 Request Timeout\r\n')
+        socket.write('Content-Type: text/html; charset=UTF-8\r\n')
+        socket.write('Content-Encoding: UTF-8\r\n')
+        socket.write(`Cache-Control: max-age=${maxAge}, s-maxage=${maxAge}\r\n`)
+        socket.write('Connection: close\r\n\r\n')
         socket.write('Request Timeout')
         socket.end()
       })


calebcartwright:
> Turns out curl, got and firefox were all fine with \n. Heroku is not feeling it and throws a 503.

Wow, good spot!

> I think switching the timeouts to return this response was a red herring when it comes to the performance issue we saw today but if we hit a problem, we can roll back.

Can you elaborate on this? Are you saying that you don't think the previous behavior (with the 503 errors) was not the driver of the issues today? If so then I think it's safe to say I disagree, especially given my observations over the course of about 80 minutes and the issues being immediately resolved on rollback.

If we still have doubts though, I'd actually suggest that we redeploy #6371 as-is at some point in the near future and see if the behavior is replicated. I say that because if that really wasn't the driver, then IMO we need to definitively rule it out and try to determine what the root cause was (or at least a reasonable best guess) and think through any changes we may need to make to deal with that type of scenario, given that a 3x capacity scale out had zero impact on ameliorating the outage.

chris48s:
> Are you saying that you don't think the previous behavior (with the 503 errors) was not the driver of the issues today? If so then I think it's safe to say I disagree, especially given my observations over the course of about 80 minutes and the issues being immediately resolved on rollback

I am saying that. The reason I say it is because we deployed that change, nothing happened for ~18 hours and _then_ we suffered a massive performance issue which coincided with a big spike in traffic..

I certainly don't doubt that performing the rollback brought things under control straight away, but rolling back the last thing we deployed will have also had the side-effect of restarting the dynos and resetting any cool-off periods at the same time as rolling back the change. I suspect that was probably what got things under control, rather than rolling back the commit per-se. i.e: I reckon doing `heroku ps:restart -a <our-app-name>` instead of a rollback would have probably had the exact same effect, but obviously I can't prove that.

At the moment we don't really have the data to show either theory right or wrong, so I'm not saying it with a high degree of certainty - I may be wrong. We are both kind of speculating a bit here..

> If we still have doubts though, I'd actually suggest that we redeploy #6371 as-is at some point in the near future and see if the behaviour is replicated.

Yeah that sounds entirely reasonable, but this is what today looked like (with that last change rolled back) :grimacing:

Screenshot at 2021-04-13 20-27-01

I'm not sure how to evaluate changing this under those conditions. We could try going up to 7 dynos in the day for a bit to try and stabilize things and then try changing stuff from a stable baseline then try and crank it down again?

calebcartwright:
Yup sounds reasonable. I'm good with the changes here and will approve, maybe we can discuss in Discord the if/when to run another test

calebcartwright (approved):

Issue #6388, opened 2021-04-16
GitLab tag (latest by date)


guillermin:
:clipboard: **Description**

Exactly the same as the already existing "GitHub tag (latest by date)" badge, but for a GitLab.com repository.
e.g. "tag | v1.3.0"

:link: **Data**

The data can be obtained from the public GitLab API (no key required). The call would be as follows:
`https://gitlab.com/api/v4/projects/{user}%2F{repo}/repository/tags`
(where `{user}` and `{repo}` should be replaced by their respective values)

The response is a JSON array that is by default sorted with the latest tag on top, so the value could be extracted as:
`result[0].name`

The documentation for the call can be found [here](https://docs.gitlab.com/ee/api/tags.html#list-project-repository-tags).

:microphone: **Motivation**

It can be used anywhere people are already using the equivalent GitHub badge, but for GitLab repositories. The main use case is to display the latest version of the app according to the repository.
In my case I want it alongside the F-Droid version badge so that it is clear whether the version that is published to F-Droid is the latest or not.

calebcartwright:
Thanks for reaching out but going to close as a duplicate of #4867

Issue #4867, opened 2020-04-04
GitLab tags and releases


svoop:
Similar to GitHub badges, it would be nice to have GitLab badges for tags and releases as well. Both tags and releases are available for public repos from the GitLab API without authentication, examples:

```
curl -v 'https://gitlab.com/api/v4/projects/gitlab-org%2fgitlab-foss/repository/tags'
curl -v 'https://gitlab.com/api/v4/projects/gitlab-org%2fgitlab-foss/releases'
```

(I might have a go later this year, right now I'm a little overbooked due to COVID-19.)

chris48s:
Thanks for offering to work on this.

It has been a while since anyone looked at GitLab badges, but I think one of the issues we hit in the past was the rate limits on the API. You've noted that a token is not required to access these endpoints for public repos, but presumably there are rate limits on anonymous access?
Probably a good start would be to have a read over https://github.com/badges/shields/issues/541 to understand that history.

The reason we are able to provide badges for GitLab CI pipelines is because we can get the info for them without hitting their API.

svoop:
@chris48s Two quick thoughts: 

* A few days ago, gitlab.com has switched to Cloudflare which could have a positive impact on rate limits. Haven't checked this though.
* For self-hosted GitLab, possible rate limits should be ignored. Either the server admins lift them or the repo users don't used the badges. Either way, admins and users often are part of the same organisation.

guillermin:
I've started by creating a `gitlab-tag` service with `GitLab tag (latest by date)`. This is the one I'm most interested in, but I'll get around to implementing the rest of them soon.

calebcartwright:
Tag support went live earlier today, with the standard by-date/semver and pre-release filters as other similar badges. Release support should get added via #7021 in the hopefully not too distant future

PR #6389, opened 2021-04-16
Added [GitLabTag] service

guillermin:
Added GitLab Tag (latest by date) service, equivalent to GitHub Tag (latest by date) but for gitlab.com instead of github.com.

I created issue #6388 for this, but it is closed as a duplicate of #4867 (which asks to replicate all GitHub Tags and Releases services for GitLab. Right now my priority is this badge, but I promise to to include the rest in the future.

### DIFF
--- a/services/gitlab/gitlab-tag.service.js
+++ b/services/gitlab/gitlab-tag.service.js
@@ -41,6 +41,9 @@ module.exports = class GitlabTag extends BaseJsonService {
     return this._requestJson({
       schema,
       url: `https://gitlab.com/api/v4/projects/${user}%2F${repo}/repository/tags`,
+      errorMessages: {
+        404: 'repo not found',
+      },
     })
   }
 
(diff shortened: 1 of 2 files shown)


PyvesB:
Hello @guillermin! 👋🏻 

Thanks for opening this. Unfortunately, we don't intend on expanding our GitLab badge offering at this point in time, due to API rate limit concerns. Right now, the badge you've implemented would probably fail most of the time on our production servers due to these limitations. See https://github.com/badges/shields/issues/541#issuecomment-640950235 and various other messages for a bit more context. 😉

PyvesB:
Due to aforementioned API rate limiting concerns, I don't see a path forward with this pull request for the time being. Thanks for sharing your code, but I'm afraid I'll have to close it for now.

The next steps for any Gitlab badge using the official API would be to engage in #541 and help us figure out the state of the access token features offered by Gitlab.

PR #6448, opened 2021-05-02
Specification: Add better alt-text

ajs256:
I added more accessible alt text to the sample badges in the `SPECIFICATION` file, so it is more accessible to users who cannot see the images.


PyvesB:
This may sound a little stupid, but I never realised that this was the way to add alt text to an image with markdown. Thanks for helping me join the dots! 😃

PyvesB (approved):

PR #6449, opened 2021-05-05
[GradlePluginPortal] add gradle plugin portal

### LINKED ISSUE #1125: Add badge for gradle plugin portal
It would be cool if we have a badge for the [gradle plugin portal](https://plugins.gradle.org/).

I image it like similar the maven central or the bintray badge.
[Bintray]() (If it's working of course 🙃)
[Maven Central]()
plugin portal | v0.9.95

You can use this URL for example: (don't know how shields works, but I think you need that 😄)
https://plugins.gradle.org/plugin/org.shipkit.java-publish/0.9.95

anatawa12:
Closes #1125


PyvesB (commented):
Thanks for your contribution to Shields.io @anatawa12! I'm kind of in two minds here.

On the one hand it does make building badges for Gradle Plugin Portals slightly simpler, as you don't need to specify the full Maven metadata URL. The resulting badge URL is consequently shorter.

However, it does so at the cost of additional maintenance on our side. In particular, this new implementation is pretty much a copy-paste of https://github.com/badges/shields/tree/master/services/maven-central, with only a few keywords changed. I would personally not be keen on merging this pull request in its current state.

I'm wondering whether adding a specific Gradle Plugin Portal example  to the existing Maven Central implementation would be enough here. This example would effectively act as documenting the solution described in https://github.com/badges/shields/issues/1125#issuecomment-588435609 directly in our homepage.

Thoughts from @badges/shields-maintainers  also welcome!

chris48s:
I don't think I understand the nuances of the java ecosystem well enough to have much of an opinion on this so I'll defer to your judgement @PyvesB  but a couple of things that occur to me:
- If this is essentially a subset of maven (I'm assuming this given you say we could also solve this with documentation), could we implement this as a redirect route to the maven badge?
- If we want to include grade plugin as a different implementation, we could at least pull out the schema and the logic in `handle()` into shared helpers which both gradle and maven share which would make the implementation pretty thin

anatawa12:
> If we want to include grade plugin as a different implementation, we could at least pull out the schema and the logic in handle() into shared helpers which both gradle and maven share which would make the implementation pretty thin

In my opinion, This solution is better because maven-central and gradle plugin portal are same except for url syntax and label. It may also good to put logic for maven-metadata into shared helpers. Both maven-metadata and maven-central have same schema for xml. I don't know why not currently but maven-metadata should also have version prefix so It's good to mix into single logic.

anatawa12:
In my opinion, having gradle logo by default is good. However, by the CONTRIBUTING.md, It looks not allowed. What do you think to have gradle logo by default? Is it allowed to have gradle logo in example?

calebcartwright:
> I'm wondering whether adding a specific Gradle Plugin Portal example to the existing Maven Central implementation would be enough here

I think this would be helpful in the event we're unable to incorporate any other changes. Adding this as an example, with relevant search terms, and any docs for the badge builder window should help folks discover and utilize the badge.

> However, it does so at the cost of additional maintenance on our side. In particular, this new implementation is pretty much a copy-paste of https://github.com/badges/shields/tree/master/services/maven-central, with only a few keywords changed. I would personally not be keen on merging this pull request in its current state.

I agree. I'm not necessarily opposed to having a "native" badge for gradle plugins and a corresponding route, but we'd need an implementation that encapsulated all the common bits (and there indeed seems to be a ton of overlap). The first draft proposed here would have too much code duplication for my liking.

> If we want to include grade plugin as a different implementation, we could at least pull out the schema and the logic in handle() into shared helpers which both gradle and maven share which would make the implementation pretty thin

Agreed :100: 

> Our redirector functionality is generally used for hiding a legacy route, rather than advertising a new one.

Yeah that sounds like it'd be tricky

calebcartwright:
> In my opinion, having gradle logo by default is good. However, by the CONTRIBUTING.md, It looks not allowed. What do you think to have gradle logo by default? Is it allowed to have gradle logo in example?

That's correct @anatawa12, we don't do logos by default and we cannot do so here either. Folks that want to include a logo can do so easily, but this badge would be subject to the same set of rules that govern the other badges.

chris48s:
> do you know how we would go about surfacing an example in the UI?

This is a good point. I don't think we can accommodate that with the current setup but if we wanted I think it would be possible to allow `redirector` to accept `examples` as a param and then set it on the service we return https://github.com/badges/shields/blob/7d452472defa0e0bd71d6443393e522e8457f856/core/base-service/redirector.js#L33-L52

anatawa12:
@PyvesB I'm sorry long time no response. I made gradle-plugin-portal as a redirector. Could you check my code? and Should I rewite maven-central with redirector?

PyvesB:
Quite a lot of my plate at the moment, but I'll try to review this as soon as possible. If you want to refactor maven-central as well, I would suggest doing that in a follow up pull request. 😉

services/gradle-plugin-portal/gradle-plugin-portal.service.js
```diff
@@ -0,0 +1,54 @@
...
+      },
+      staticPreview: {
+        label: 'plugin portal',
+        message: 'v0.14.0',
+        color: 'blue',
+      },
+    },
+    /*
+    {
+      title: 'Gradle Plugin Portal with version prefix filter',
```
PyvesB, line 27: Please remove this commented example for now. When and if we get round to adding a version prefix attribute to maven-metadata, it's enough enough to re-add at that point, without the risk of the commented code having gone stale in the meantime. 😉

PyvesB (changes requested):

services/gradle-plugin-portal/gradle-plugin-portal.service.js
anatawa12: I removed this comment.

PyvesB (approved):
Thanks, looks good to me! 👍🏻

Issue #3222, opened 2019-03-19
Option for unstable Ruby gem version


AlexWayfer:
:clipboard: **Description**

Hello!

I want an option for `gem version` badge about unstable versions.

Example: https://rubygems.org/gems/flame

The current badge has only `4.18.1` value.

I want to display `5.0.0.rc5` value for informing about preparing the new release.

Thanks.

calebcartwright:
Do you know if unstable/rc versions are provided via an API response? 

We're currently using https://guides.rubygems.org/rubygems-org-api/#gem-methods, and the API response for the aforementioned gem lists the version as  `4.18.1`:

```json
{
  "name": "flame",
  "downloads": 61236,
  "version": "4.18.1",
  "version_downloads": 2563,
  "platform": "ruby",
  "authors": "Alexander Popov",
  "info": "Use controller's classes with instance methods as routing actions, mounting its in application 
  class.",
  "licenses": [
    "MIT"
  ],
  "metadata": {},
  "sha": "12533f702a8059a51d6d2ce2bea1858ec8ac2993e8549feedc300af64e310e88",
  "project_uri": "https://rubygems.org/gems/flame",
  "gem_uri": "https://rubygems.org/gems/flame-4.18.1.gem",
  "homepage_uri": "https://github.com/AlexWayfer/flame",
  "wiki_uri": null,
  "documentation_uri": "http://www.rubydoc.info/gems/flame/4.18.1",
  "mailing_list_uri": null,
  "source_code_uri": null,
  "bug_tracker_uri": null,
  "changelog_uri": null,
  "dependencies": {...}
}
```

calebcartwright:
Perhaps this could be used:
https://guides.rubygems.org/rubygems-org-api/#gem-version-methods

AlexWayfer:
> Do you know if unstable/rc versions are provided via an API response?

I didn't, but now I know. :slightly_smiling_face: 

> Perhaps this could be used: https://guides.rubygems.org/rubygems-org-api/#gem-version-methods

Yep, looks appropriate.

> calebcartwright added the good first issue  label

This is wrong if it about issue author, not like "good for the first PR". :smiley: 

https://github.com/badges/shields/issues?utf8=%E2%9C%93&q=author%3AAlexWayfer

calebcartwright:
> This is wrong if it about issue author, not like "good for the first PR".  

The meaning of our labels is documented: 
https://github.com/badges/shields/labels 
[...]

PR #6451, opened 2021-05-06
Add optional query parameter (include_prereleases) to [GemVersion]

profgrammer:
Added an optional query parameter (include_prereleases) to the GemVersion service, with reference to #3222 
- If passed, it will fetch the latest version for the gem, otherwise it will default to the latest stable version


PyvesB (commented):
Hello @profgrammer ! 👋🏻 

Thanks for your contribution to Shields.io! In my opinion, having the new optional parameter as a query parameter rather than an element in the path would be a better fit.

Do we foresee any values other than stable or latest in the near future? If not, we could even make it a "valueless" query parameter, e.g. https://img.shields.io/gem/v/formatador?latest or something appropriately named like that.

profgrammer:
Hi @PyvesB, thank you for your inputs. I don't think we expect values other than `stable` or `latest` in the future, and the valueless query param seems like a good alternative.

PyvesB:
In that case, I can point you to this implementation here: https://github.com/badges/shields/blob/master/services/youtube/youtube-likes.service.js

Note the `withDislikes` query parameter being used.

calebcartwright:
I'm not terribly familiar with the Ruby ecosystem, but would it make sense to use the prerelease type naming/conventions we use in other version badges?

profgrammer:
@calebcartwright I see 2 options for this - 
- create a new route `/gem/vpre` that includes preleases
- renaming the `latest` query param to `include_prereleases`

I think the latter would be a better option, will go ahead and rename the query parameter. Thanks!

services/gem/gem-version.service.js
```diff
@@ -35,8 +57,24 @@ module.exports = class GemVersion extends BaseJsonService {
...
+    return this._requestJson({
+      schema: versionSchema,
+      url: `https://rubygems.org/api/v1/versions/${gem}.json`,
+    })
+  }
+
+  async handle({ gem }, queryParams) {
+    if (queryParams && typeof queryParams.include_prereleases !== 'undefined') {
+      const data = await this.fetchLatest({ gem })
+      if (!Array.isArray(data) || data.length === 0) {
```
PyvesB, line 70: I'm a little confused here: shouldn't the `versionSchema` guarantee that we can't hit this case?

services/gem/gem-version.tester.js
```diff
@@ -11,3 +14,19 @@ t.create('version (valid)').get('/formatador.json').expectBadge({
 t.create('version (not found)')
   .get('/not-a-package.json')
   .expectBadge({ label: 'gem', message: 'not found' })
+
+// this is the same as isVPlusDottedVersionNClausesWithOptionalSuffix from test-validators.js
+// except that it also accepts regexes like 5.0.0.rc5 - the . before the rc5 is not accepted in the original
+const isVPlusDottedVersionNClausesWithOptionalSuffix = withRegex(
```
PyvesB, line 20: I'm wondering whether it would have been better to modify the original regex directly. Admittedly I've not seen many projects use a dot as the separator for the pre-release tag outside the Ruby ecosystem, so what you've done may make more sense. Let's leave it as is for now, but another maintainer will possibly have a stronger view on this one. :)

PyvesB (commented):
Thanks for polishing the implementation promptly @profgrammer, it mostly looks good to me at this point. 👍🏻

services/gem/gem-version.service.js
profgrammer: Yes, that actually makes sense, having this check is redundant. I'll get rid of it, thanks!

PyvesB (approved):
Looks good to me, thanks for the good work! Won't merge just yet in case @calebcartwright has some additional thoughts. :)

services/gem/gem-version.tester.js
calebcartwright: generally agreed but also content with leaving this here for now. we probably do have some minor proliferation of duplicative regexes but I don't have any concerns with having this one here

calebcartwright (approved):
LGTM as well. Thanks for this @profgrammer, nice first contribution to the Shields project!

PR #6508, opened 2021-05-15
[Packagist] Update Packagist service to use v2 api

GeoffSelby:
Packagist deprecated the original `packagist.org/p/username/package` endpoint in favor of v2 `packagist.org/p2/username/package` endpoint. Because of this, new packages aren't being found using v1.

This PR updates the Packagist service to use the new endpoint.

### DIFF
--- a/services/packagist/packagist-base.js
+++ b/services/packagist/packagist-base.js
@@ -2,6 +2,7 @@
 
 const Joi = require('joi')
 const { BaseJsonService } = require('..')
+const { isStable, latest } = require('../php-version')
 
 const packageSchema = Joi.array().items(
   Joi.object({
@@ -35,6 +36,30 @@ class BasePackagistService extends BaseJsonService {
    */
   async fetch({ user, repo, schema, server = 'https://packagist.org' }) {
     const url = `${server}/p2/${user.toLowerCase()}/${repo.toLowerCase()}.json`
+
+    return this._requestJson({
+      schema,
+      url,
+    })
+  }
+
+  /**
+   * Fetch dev releases method.
+   *
+   * This method utilize composer metadata API which
+   * "... is the preferred way to access the data as it is always up to date,
+   * and dumped to static files so it is very efficient on our end." (comment from official documentation).
+   * For more information please refer to https://packagist.org/apidoc#get-package-data.
+   *
+   * @param {object} attrs Refer to individual attrs
+   * @param {string} attrs.user package user
+   * @param {string} attrs.repo package repository
+   * @param {Joi} attrs.schema Joi schema to validate the response transformed to JSON
+   * @param {string} attrs.server URL for the packagist registry server (Optional)
+   * @returns {object} Parsed response
+   */
+  async fetchDev({ user, repo, schema, server = 'https://packagist.org' }) {
+    const url = `${server}/p2/${user.toLowerCase()}/${repo.toLowerCase()}~dev.json`
 
     return this._requestJson({
       schema,
@@ -80,6 +105,43 @@ class BasePackagistService extends BaseJsonService {
 
   getPackageName(user, repo) {
     return `${user.toLowerCase()}/${repo.toLowerCase()}`
+  }
+
+  decompressResponse(json, packageName) {
+    const versions = json.packages[packageName]
+    const expanded = []
+    let expandedVersion = null
+
+    versions.forEach(versionData => {
+      if (!expandedVersion) {
+        expandedVersion = versionData
+        expanded.push(expandedVersion)
+      }
+
+      Object.entries(versionData).forEach(([key, value]) => {
+        if (value === '__unset') {
+          delete expandedVersion[key]
+        } else {
+          expandedVersion[key] = value
+        }
+      })
+
+      expandedVersion = { ...expandedVersion }
+
+      expanded.push(expandedVersion)
+    })
+
+    return expanded
+  }
+
+  findRelease(json, versions = []) {
+    json.forEach(version => {
+      versions.push(version.version)
+    })
+
+    const release = latest(versions.filter(isStable)) || latest(versions)
+
+    return json.filter(version => version.version === release)[0]
   }
 }
 
--- a/services/packagist/packagist-base.js
+++ b/services/packagist/packagist-base.js
@@ -3,11 +3,9 @@
 const Joi = require('joi')
 const { BaseJsonService } = require('..')
 
-const packageSchema = Joi.object()
-  .pattern(
-    /^/,
+const packageSchema = Joi.array()
+  .items(
     Joi.object({
-      'default-branch': Joi.bool(),
       version: Joi.string(),
       require: Joi.object({
         php: Joi.string(),
@@ -38,7 +36,7 @@ class BasePackagistService extends BaseJsonService {
    * @returns {object} Parsed response
    */
   async fetch({ user, repo, schema, server = 'https://packagist.org' }) {
-    const url = `${server}/p/${user.toLowerCase()}/${repo.toLowerCase()}.json`
+    const url = `${server}/p2/${user.toLowerCase()}/${repo.toLowerCase()}.json`
 
     return this._requestJson({
       schema,
--- a/services/packagist/packagist-base.js
+++ b/services/packagist/packagist-base.js
@@ -3,16 +3,14 @@
 const Joi = require('joi')
 const { BaseJsonService } = require('..')
 
-const packageSchema = Joi.array()
-  .items(
-    Joi.object({
-      version: Joi.string(),
-      require: Joi.object({
-        php: Joi.string(),
-      }),
-    }).required()
-  )
-  .required()
+const packageSchema = Joi.array().items(
+  Joi.object({
+    version: Joi.string(),
+    require: Joi.object({
+      php: Joi.string(),
+    }),
+  })
+)
 
 const allVersionsSchema = Joi.object({
   packages: Joi.object().pattern(/^/, packageSchema).required(),
--- a/services/packagist/packagist-license.service.js
+++ b/services/packagist/packagist-license.service.js
@@ -13,6 +13,7 @@ const {
 const packageSchema = Joi.array()
   .items(
     Joi.object({
+      version: Joi.string(),
       license: Joi.array(),
     }).required()
   )
@@ -59,7 +60,11 @@ module.exports = class PackagistLicense extends BasePackagistService {
   transform({ json, user, repo }) {
     const packageName = this.getPackageName(user, repo)
 
-    const license = json.packages[packageName][0].license
+    const decompressed = this.decompressResponse(json, packageName)
+
+    const version = this.findRelease(decompressed)
+
+    const license = version.license
 
     if (!license) {
       throw new NotFound({ prettyMessage: 'license not found' })
--- a/services/packagist/packagist-license.service.js
+++ b/services/packagist/packagist-license.service.js
@@ -12,7 +12,7 @@ const {
 const packageSchema = Joi.array()
   .items(
     Joi.object({
-      license: Joi.array().required(),
+      license: Joi.array(),
     }).required()
   )
   .required()
--- a/services/packagist/packagist-license.service.js
+++ b/services/packagist/packagist-license.service.js
@@ -3,18 +3,15 @@
 const Joi = require('joi')
 const { renderLicenseBadge } = require('../licenses')
 const { optionalUrl } = require('../validators')
-const { NotFound } = require('..')
 const {
   keywords,
   BasePackagistService,
   customServerDocumentationFragment,
 } = require('./packagist-base')
 
-const packageSchema = Joi.object()
-  .pattern(
-    /^/,
+const packageSchema = Joi.array()
+  .items(
     Joi.object({
-      'default-branch': Joi.bool(),
       license: Joi.array().required(),
     }).required()
   )
@@ -59,17 +56,18 @@ module.exports = class PackagistLicense extends BasePackagistService {
   }
 
   transform({ json, user, repo }) {
-    const branch = this.getDefaultBranch(json, user, repo)
-    if (!branch) {
-      throw new NotFound({ prettyMessage: 'default branch not found' })
-    }
-    const { license } = branch
+    const packageName = this.getPackageName(user, repo)
+
+    const license = json.packages[packageName][0].license
+
     return { license }
   }
 
   async handle({ user, repo }, { server }) {
     const json = await this.fetch({ user, repo, schema, server })
+
     const { license } = this.transform({ json, user, repo })
+
     return renderLicenseBadge({ license })
   }
 }
--- a/services/packagist/packagist-license.service.js
+++ b/services/packagist/packagist-license.service.js
@@ -3,6 +3,7 @@
 const Joi = require('joi')
 const { renderLicenseBadge } = require('../licenses')
 const { optionalUrl } = require('../validators')
+const { NotFound } = require('..')
 const {
   keywords,
   BasePackagistService,
@@ -59,6 +60,10 @@ module.exports = class PackagistLicense extends BasePackagistService {
     const packageName = this.getPackageName(user, repo)
 
     const license = json.packages[packageName][0].license
+
+    if (!license) {
+      throw new NotFound({ prettyMessage: 'license not found' })
+    }
 
     return { license }
   }
--- a/services/packagist/packagist-php-version.service.js
+++ b/services/packagist/packagist-php-version.service.js
@@ -68,25 +68,68 @@ module.exports = class PackagistPhpVersion extends BasePackagistService {
     }
   }
 
-  findVersionIndex(json, user, repo, version) {
-    const packageArr = json.packages[this.getPackageName(user, repo)]
-
-    return packageArr.findIndex(v => v.version === version)
+  findVersionIndex(json, version) {
+    return json.findIndex(v => v.version === version)
   }
 
-  transform({ json, user, repo, version = '' }) {
-    const packageVersion =
-      version === ''
-        ? json.packages[this.getPackageName(user, repo)][0]
-        : json.packages[this.getPackageName(user, repo)][
-            this.findVersionIndex(json, user, repo, version)
-          ]
+  async findSpecifiedVersion(json, user, repo, version, server) {
+    let release
+
+    if ((release = json[this.findVersionIndex(json, version)])) {
+      return release
+    } else {
+      try {
+        const allData = await this.fetchDev({
+          user,
+          repo,
+          schema: allVersionsSchema,
+          server,
+        })
+
+        const decompressed = this.decompressResponse(
+          allData,
+          this.getPackageName(user, repo)
+        )
+
+        return decompressed[this.findVersionIndex(decompressed, version)]
+      } catch (e) {
+        return release
+      }
+    }
+  }
+
+  async transform({ json, user, repo, version = '', server }) {
+    let packageVersion
+    const decompressed = this.decompressResponse(
+      json,
+      this.getPackageName(user, repo)
+    )
+
+    if (version === '') {
+      packageVersion = this.findRelease(decompressed)
+    } else {
+      try {
+        packageVersion = await this.findSpecifiedVersion(
+          decompressed,
+          user,
+          repo,
+          version,
+          server
+        )
+      } catch {
+        packageVersion = null
+      }
+    }
 
     if (!packageVersion) {
       throw new NotFound({ prettyMessage: 'invalid version' })
     }
 
-    if (!packageVersion.require || !packageVersion.require.php) {
+    if (
+      !packageVersion.require ||
+      !packageVersion.require.php ||
+      packageVersion.require.php === '__unset'
+    ) {
       throw new NotFound({ prettyMessage: 'version requirement not found' })
     }
 
@@ -100,11 +143,12 @@ module.exports = class PackagistPhpVersion extends BasePackagistService {
       schema: allVersionsSchema,
       server,
     })
-    const { phpVersion } = this.transform({
+    const { phpVersion } = await this.transform({
       json: allData,
       user,
       repo,
       version,
+      server,
     })
     return this.constructor.render({ php: phpVersion })
   }
--- a/services/packagist/packagist-php-version.service.js
+++ b/services/packagist/packagist-php-version.service.js
@@ -68,11 +68,19 @@ module.exports = class PackagistPhpVersion extends BasePackagistService {
     }
   }
 
+  findVersionIndex(json, user, repo, version) {
+    const packageArr = json.packages[this.getPackageName(user, repo)]
+
+    return packageArr.findIndex(v => v.version === version)
+  }
+
   transform({ json, user, repo, version = '' }) {
     const packageVersion =
       version === ''
-        ? this.getDefaultBranch(json, user, repo)
-        : json.packages[this.getPackageName(user, repo)][version]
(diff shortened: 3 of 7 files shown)


calebcartwright:
Thank you for the PR @GeoffSelby! The changes seem to have broken some of the badges, more details can be found in the CI output (e.g. https://app.circleci.com/pipelines/github/badges/shields/6648/workflows/ba687673-7fbf-4cb5-8579-5a699f2b9972/jobs/143021).

Could you please take a look when you get a chance? Let us know if you have any questions

GeoffSelby:
Thanks, @calebcartwright! Looks like it was the validation schema causing problems. Should be good when these tests wrap up.

services/packagist/packagist-license.service.js
```diff
@@ -3,19 +3,16 @@
...
 
-const packageSchema = Joi.object()
-  .pattern(
-    /^/,
+const packageSchema = Joi.array()
+  .items(
     Joi.object({
-      'default-branch': Joi.bool(),
-      license: Joi.array().required(),
+      license: Joi.array(),
```
calebcartwright, line 15: Is there a need to drop `.required()` here? Is this because the new schema may not always define the key? If that's the case then I think in the transform function we need to account for that fact and throw a not found/invalid response error

services/packagist/packagist-license.service.js
```diff
@@ -59,17 +56,18 @@ module.exports = class PackagistLicense extends BasePackagistService {
   }
 
   transform({ json, user, repo }) {
-    const branch = this.getDefaultBranch(json, user, repo)
-    if (!branch) {
-      throw new NotFound({ prettyMessage: 'default branch not found' })
-    }
-    const { license } = branch
```
calebcartwright, line 66: tbh I'm not sure I know what we were doing here before. Do you know what, if any, the chances are of this API change resulting in us finding and rendering a different license because of the branch change?

services/packagist/packagist-php-version.spec.js
```diff
@@ -7,20 +7,20 @@ const PackagistPhpVersion = require('./packagist-php-version.service')
...
-        'dev-main': { require: { php: '^8' }, 'default-branch': true },
-      },
-      'samwise/gardening': {
-        '1.0.x-dev': {},
-        '2.0.x-dev': {},
-      },
-      'pippin/mischief': {
-        '1.0.0': {},
-        'dev-main': { require: {}, 'default-branch': true },
-      },
```
calebcartwright, line 23: it typically concerns me when we have to modify tests as part of some other change. I realize that _some_ updates are inevitable here since the new api has a different response structure, but I feel like there's more going on or at least that I'm missing some context (though I presume this is all branch related)

calebcartwright (commented):
Thanks for the updates! I started going through this but as I'd noted in the inline comments I feel like I could use some more info about the change. Is there some background context, documentation, etc. you can share with us about the nature of the upstream changes and why there's all the cascading impacts relative to the branches?

Shields integrates with so many upstream services that even within the maintainer team we've got quite a few whose APIs we're not terribly familiar with, so any additional info you could share would be most appreciated!

@chris48s for some reason I feel like I recall you have some more hands on PHP experience than others (certainly more so than myself) so do you have any background on what we were doing before vs. the changes with the new API?

services/packagist/packagist-php-version.spec.js
GeoffSelby: This change is solely because of the change in response structure as seen in the diff.

services/packagist/packagist-license.service.js
GeoffSelby: It shouldn't ever find or render the wrong license because we only care about the most recent version which is always the first item in the `json.packages[packageName]` array.

services/packagist/packagist-license.service.js
GeoffSelby: The new api endpoint now only returns the license array with the first item in the response (most recent release). At least that's what I've found in my testing. because of this, the schema validation would fail when the response contains more than 1 version. I do think that throwing a not found error would be good just in case there isn't a license at all for some reason. I'll add that in shortly.

services/packagist/packagist-license.service.js
calebcartwright: Thanks, though my question isn't about "right" or "wrong". We take backwards compatibility pretty seriously, and one of the things we have to be mindful of is whether a change we make to the Shields code has the potential to modify any of our existing badge users without them changing anything in their corresponding project/repo/package/etc.

It's still not clear to me whether that's a possibility here

GeoffSelby: I guess it is technically possible here but there is only one way and it's likely never going to be an issue. The only way it could be a problem is if the default branch has a different license than the most recent tagged release. For example, if a project had an MIT license before (so in the most recent tagged release) but you changed it to Apache License 2.0 in your default branch, main, master, dev, whatever you set your default branch to, and did not tag a new release. The current Shields, with the old api, would return Apache License 2.0 but this PR would return MIT because it's the license of the most recent tagged release.

chris48s:
> @chris48s for some reason I feel like I recall you have some more hands on PHP experience than others (certainly more so than myself) so do you have any background on what we were doing before vs. the changes with the new API?

I'll try and have a proper look over this in the next couple of days and see if I have anything useful to add
Also thanks for picking this up @GeoffSelby :+1:

services/packagist/packagist-license.service.js
calebcartwright: That's helpful clarity, thank you!

Based on a cursory glance of the docs under https://packagist.org/apidoc#getting-package-data it seems like we could maintain access to the full set of dev/branch data if we fetched `packageName-dev.json` instead of `packageName.json`

Do you see any issues with that?

chris48s:
I did start looking at this one tonight and I've made some notes but I still need to spend a bit more time on it before I can post something useful.

services/packagist/packagist-license.service.js
chris48s: There are multiple different issues tied together here. It is a bit messy and complicated..

- The old packagist `/p` API used to return both tagged releases and dev versions (branches) in the same API response. This was always a somewhat esotric behaviour of packagist in comparison to other package registries.
- The new packagist API separates tagged and dev versions into different API endpoints: https://packagist.org/apidoc#get-package-data So https://packagist.org/p2/symfony/symfony.json now returns tagged releases and https://packagist.org/p2/symfony/symfony~dev.json returns dev releases.
- Two of the packagist badges (`/l` and `/php-v`) have long used the default branch rather than the latest version as the reference point for license/php-version. This is one of those things that mostly goes unnoticed (esp for the license case because projects change their license very infrequently), but it is actually a long-standing bug. See https://github.com/badges/shields/issues/2356 It never got fixed, but I'd acknowledge that the existing behaviour is a bug and the correct behaviour is to default to the **latest stable version** not the default branch if `/php-v` or `/l` is called with no explicit version (i.e: a call like https://img.shields.io/packagist/php-v/doctrine/orm rather than https://img.shields.io/packagist/php-v/doctrine/orm/2.8.4 ). Same as we would do for any other package registry. We shouldn't go out of our way to call the `~dev` endpoint to reproduce the existing behaviour IMO. We should use this opportunity to fix it.
- That said, the code here (just using the first item in the array) does not give us the latest stable version. For example, if I call https://packagist.org/p2/symfony/symfony.json the first version in the array is `v5.3.0-RC1` (a pre-release). If we're going to properly fix this we should parse the array for the first stable version and use that, not just pick the first release in the array. It is less of an issue for license, but more important for the supported PHP versions badge.
- Another related issue/knock-on effect is that there are some badge queries which are currently valid which will become invalid with this approach. For example at the moment https://img.shields.io/packagist/php-v/cakephp/cakephp/dev-master is a valid badge but once we switch from `/p` to `/p2` that will become an error because `dev-master` isn't a tagged version. I think (again for the `/php-v` endpoint), if it called with the `:version` param we could:
    - call `/p2/user/repo.json`
    - look for `version` in the tagged releases
    - if we don't find it, make a second call to the `/p2/user/repo~dev.json` endpoint to look for it 
without adding too much extra complexity to keep that variant working as it does now.
- Also mixed up in this is the fact that the new `/p2` API endpoint doesn't return all of the keys for all of the version objects. This is sufficiently unexpected that I raised an issue about it over on the packagist repo: https://github.com/composer/packagist/issues/1194 It turns out this is intended and also adds a bit of extra complexity for us: The recommendation is we port https://github.com/composer/metadata-minifier/blob/c549d23829536f0d0e984aaabbf02af91f443207/src/MetadataMinifier.php#L22-L46 to javascript to 'decompress' the metadata before we use it. Fortunately this doesn't look to complex. So having said all that, I think the correct approach is:

PHP version support (`/php-v`) badge:
- Call `/p2/:user/:repo.json`
- Decompress the response using a port of https://github.com/composer/metadata-minifier/blob/c549d23829536f0d0e984aaabbf02af91f443207/src/MetadataMinifier.php#L22-L46
- If we are called with no `:version` param, identify the latest stable release. If there is none, fall back to the latest pre-release (first item in the array). Report the PHP version support for that release.
- If we are called with a `:version` param, try to find that release. If it does exist, fine. If not, make a second call to `/p2/:user/:repo~dev.json`, decompress that and see if we can find it there. If we find it in either, report the PHP version support for that release.

License badge (`/l`):
- Call `/p2/:user/:repo.json`
- Decompress the response using a port of https://github.com/composer/metadata-minifier/blob/c549d23829536f0d0e984aaabbf02af91f443207/src/MetadataMinifier.php#L22-L46
- Identify the latest stable release. If there is none, fall back to the latest pre-release. Report the PHP license for that release.

Sorry - this is a bit of a can-of-worms for a first contribution :grimacing:

chris48s:
I'm kind of hesitant to bring this one up: I don't want to get too bogged down in the 1% case because I think fixing this for the 99% case (packagist.org) is the priority and that is currently broken or producing unexpected behaviour for some users. That said, this transition is slightly complicated by the fact that the packagist integration allows the user to specify a custom (self-hosted) registry using the `?server=` param.

If we completely drop compatibility with the v1 API and move to only allowing the v2 API to be used, if anyone is currently using this with a self-hosted registry that is running an older version of self-hosted packagist, their badges are going to break.
I suspect this use-case is probably super-rare for shields.io. Maybe some self-hosting users use it too (but we don't support setting credentials for private packagist, so maybe not..) Presumably _someone somewhere_ is using this though: https://github.com/badges/shields/pull/3986

I think with one of the other services (sonar? nexus?) we support multiple API versions and we try the most recent first then fall back to the second (but allow you to explicitly specify the version). This situation is slightly complicated by the fact that it seems like the old API is retained but serves stale data so falling back by default might not be a great idea either :/ I think my view on this is we should just optimise to work for packagist.org and other custom registries that maintain feature parity with packagist.org. I think a fallback is going to further complicate an integration which already has a few pain points to serve a use-case which may turn out to be hypothetical in any case.

services/packagist/packagist-license.service.js
GeoffSelby: Looks like porting the metadata expander is pretty straightforward so I should be able to knock these changes out when I have some time later tonight or tomorrow 👍

calebcartwright: > The new packagist API separates tagged and dev versions into different API endpoints:

Ah okay. That's what I get for not actually looking at the responses :laughing: 

> We shouldn't go out of our way to call the ~dev endpoint to reproduce the existing behaviour IMO. We should use this opportunity to fix it.

Makes sense, and agreed :+1: 

> So having said all that, I think the correct approach is: .....

SGTM

GeoffSelby: I'm taking a look at Composer's MetadataMinifier package and I realized that the expand method doesn't actually do what we need. All it's doing is "expanding" the array of releases with data we already have. In other words, it doesn't do anything useful if we want to get the PHP version requirements or license (or anything really) for the latest stable release if it's not the first item in the array.

Take a simplified version of the response from `/p2/symfony/symfony.json`:
```json
"packages": {
  "symfony/symfony": [
    {
      "version": "v5.3.0-RC1",
      "license": [
        "MIT"
      ],
      "require": {
        "php": ">=7.2.5"
      },
    },
    {
      "version": "v5.3.0-BETA4",
    },
    {
      "version": "v5.3.0-BETA3",
    },
    {
      "version": "v5.3.0-BETA2",
    },
    {
      "version": "v5.3.0-BETA1",
    },
    {
      "version": "v5.2.9",
      "require": {
        "php": ">=7.2.5",
      },
    }
  ]
}
```

After "expanding", `license` for `v5.2.9 (latest stable release) will be whatever `license` is for the first item, `v5.3.0-RC1 in this case. In other words, if Symfony were to change its license in a pre-release,  that would be the license used for the badge regardless. The same could go for PHP version requirements though not in the case of Symfony right now.

The only way to ensure we actually have the right data is to use the standard JSON API like `packagist.org/packages/symfony/symfony.json`. This is a more expensive request because the response is massive, but it does respond with all of the data we would need for every release. This poses some other issues as well. As far as I can tell, it still wouldn't return any branches/dev versions like `dev-master` or `dev-main`. If we needed `dev` we could fall back to `/p2/user/repo~dev.json`. The other issue is that it seems the response is in no particular order. It's not a huge deal but determining the most recent stable release could be a little time-consuming since it would have to go through the entire response. 

I'll set some time aside tomorrow to start tackling this.

Seldaek: This is not the case. Just trust the code instead of trying to infer things :) If the license changed in a prerelease, then the minified json would look like:

```jsonc
"packages": {
  "symfony/symfony": [
    {
      "version": "v5.3.0-RC1",
      "license": [
        "NEW-5.3-LICENSE" // new license here
      ],
      "require": {
        "php": ">=7.2.5"
      },
    },
    {
      "version": "v5.3.0-BETA4",
    },
    {
      "version": "v5.3.0-BETA3",
    },
    {
      "version": "v5.3.0-BETA2",
    },
    {
      "version": "v5.3.0-BETA1", // new license was added in beta1 let's assume, so it is valid until here
    },
    {
      "version": "v5.2.9",
      "require": {
        "php": ">=7.2.5",
      },
      "license": [
        "MIT" // old license now is set as it changed from the versions above
      ],
    }
  ]
}
```

In both cases, after running MetadataMinifier::expand, you would have `license:["MIT"]` present in v5.2.9.

`/p2/user/repo~dev.json` includes *only* dev releases, so if you're looking for latest stable this is entirely useless. Please just use `/p2/user/repo.json` with a MetadataMinifier::expand port.

Seldaek: Also, please everyone take a couple minutes to read through https://packagist.org/apidoc#get-package-data which explains our new v2 api in details. Note the point about caching and `If-Modified-Since`.. I don't know if you can easily integrate this here to save everyone some bandwidth but it would be nice.

services/packagist/packagist-license.service.js
Seldaek: I think ideally you would run the `MetadataMinifier::expand` logic on the response *before* validating the schema, and then you can restore the `required()` I'd say. Note that packagist.org does not enforce the presence of a license key, so it may well be missing if the package author did not include it. I'd not expect they'd display a shields.io license badge if they haven't defined a license though but who knows :)

services/packagist/packagist-license.service.js
GeoffSelby: > This is not the case. Please trust the code instead of trying to infer things :) If the license changed in a prerelease, then the minified json would look like:

@Seldaek so you’re saying that if there is a change between lisences, the minified response would still include the license with the last version to have that license before the change? I guess I just didn’t fully understand what was minified in the response.

services/packagist/packagist-license.service.js
GeoffSelby: @Seldaek i thought about that too but I don’t think it would be possible with the way Shields makes a request since it’s shared with other services. Maybe with some sort of event hook system but that seems like overkill for this one situation.

services/packagist/packagist-license.service.js
Seldaek: If you iterate through versions, the license will always be the last license you saw above. If it isn't anymore, then the license is present in the version object. If an older version lacks a license, the license is set to `"__unset"`. But really IMO the best is to run the `expand` logic and *then* use the objects as you would normally, so you don't need to consider any of this.

GeoffSelby:
@chris48s I have added the changes you mentioned regarding the PHP version support badge and license badge.

chris48s:
@GeoffSelby sorry I've not had a chance to get to this one yet. I've been a bit over-committed this week and I've been slightly putting this review off as I know I need to find a decent block of time to read over it properly but I'm not ignoring it. I will find some time for this over the weekend.

chris48s:
> I think ideally you would run the `MetadataMinifier::expand` logic on the response _before_ validating the schema, and then you can restore the `required()`

> i thought about that too but I don’t think it would be possible with the way Shields makes a request since it’s shared with other services.


I suggest we override `_requestJson()`, like this:

```js
class BasePackagistService extends BaseJsonService {
  async _requestJson({ schema, url, options = {}, errorMessages = {} }) {
    const mergedOptions = {
      ...{ headers: { Accept: 'application/json' } },
      ...options,
    }
    const { buffer } = await this._request({
      url,
      options: mergedOptions,
      errorMessages,
    })
    const compressedJson = this._parseJson(buffer)

    // decompress the response here
    const expandedJson = this.decompressResponse(compressedJson)

    return this.constructor._validateJson(expandedJson, schema)
  }
}
```

then because we apply the schema after decompressing we can make the things we need `.required()`. That will allow us to simplify the code for individual services by removing all the manual error handling/validation code and calls to `decompressResponse()`.

On the subject of validation:

1. Is `version` guaranteed to exist? It looks like we're not handling the case where it is missing at least. If so, lets make it `.required()` in the schema(s)
2. Is it possible for a package on packagist to have zero releases, only dev branches (i.e: `...json` returns no objects but `..~dev.json` returns some). We could get away with assuming that a call to `/p` will give us at least one usable version when they were all mixed together, but maybe now we need to handle this case for `/p2`?

services/packagist/packagist-license.service.js
```diff
@@ -59,17 +58,26 @@ module.exports = class PackagistLicense extends BasePackagistService {
   }
 
   transform({ json, user, repo }) {
-    const branch = this.getDefaultBranch(json, user, repo)
```
chris48s, line 62: We can delete `getDefaultBranch()` from the base class now. It is not being called anywhere.

services/packagist/packagist-php-version.service.js
```diff
@@ -68,17 +68,68 @@ module.exports = class PackagistPhpVersion extends BasePackagistService {
...
+    } else {
+      try {
+        packageVersion = await this.findSpecifiedVersion(
+          decompressed,
+          user,
+          repo,
+          version,
+          server
+        )
+      } catch {
```
chris48s, line 119: stylistically we always use `} catch (e) {`
we should probably have an eslint rule for that

services/packagist/packagist-php-version.service.js
```diff
@@ -68,17 +68,68 @@ module.exports = class PackagistPhpVersion extends BasePackagistService {
...
+        )
+
+        return decompressed[this.findVersionIndex(decompressed, version)]
+      } catch (e) {
+        return release
+      }
+    }
+  }
+
+  async transform({ json, user, repo, version = '', server }) {
```
chris48s, line 101: The logic path for this badge is a bit muddled
Our sequence of execution is fetch --> transform --> render (or sometimes just fetch --> render)

- Fetch gets data from APIs
- Transform re-shapes the data (if necessary)
- Render turns it into a badge

Here we are making an API call in transform, so that's a good indicator we've split the logic up incorrectly or named things poorly.

In this situation, calling `fetch()`, trying to find the latest version in there, falling back to calling `fetchDev()` if necessary is all part of the "fetch" stage.

The "transform" stage (to the extent there is one for this badge) is just extracting `.require.php`

services/packagist/packagist-base.js
```diff
@@ -85,6 +106,43 @@ class BasePackagistService extends BaseJsonService {
...
+      expanded.push(expandedVersion)
+    })
+
+    return expanded
+  }
+
+  findRelease(json, versions = []) {
+    json.forEach(version => {
+      versions.push(version.version)
+    })
```
chris48s, line 140: This function is not that clearly named. How about `findLatestRelease()`?

Also, we should have one method of determining the latest (stable) version that we use consistently - we shouldn't have 2 different methods for different badges. More commentry on this on the `packagist-version.js` file..

I don't think `versions` needs to be a function param here. We never pass anything in when we call it. How about we make this just

```js
const versions = json.map(version => version.version)
```

services/packagist/packagist-version.service.js
```diff
@@ -90,25 +85,27 @@ class PackagistVersion extends BasePackagistService {
 
   transform({ includePrereleases, json, user, repo }) {
```
chris48s, line 86: As noted on `findRelease()`, we should have one method for working out the latest (stable) version. I think most of the logic in this function was there to handle dev branches being mixed in with the releases. Now that they are on a different endpoint `..~dev.json`, I think we can ditch all this can't we (do correct me if I'm wrong)? If so, this simplifies to something like:

```js
const packageSchema = Joi.array().items(
  Joi.object({version: Joi.string().required()})
)

transform({ includePrereleases, json, user, repo }) {
  const versionsData = json.packages[this.getPackageName(user, repo)]
  const versions = versionsData.map( version => version.version)

  if (includePrereleases) {
    return { version: latest(versions) }
  } else {
    const stableVersion = latest(versions.filter(isStable))
    return { version: stableVersion || latest(versions) }
  }
}
```

Then at that point, the logic for identifying the latest release is the same logic as in your `findRelease()` (which we should rename) but with a `includePrereleases` flag.. so we can move this to the base class and call with `includePrereleases=false` when we use it for the license/PHP version badges.

chris48s (commented):
I've left a few more comments but this is coming along nicely. Thanks

Seldaek:
> Is it possible for a package on packagist to have zero releases, only dev branches (i.e: ...json returns no objects but ..~dev.json returns some).

It is possible for either .json to be empty as long as no tag exists. For ~dev.json to be empty is very unlikely and would indicate a broken package most likely, I wouldn't really care about handling this gracefully but it is an eventuality.

If there is a package version object present then yes the version field will always be present.

PyvesB:
@GeoffSelby have you had a chance to read through @chris48s's latest review comments? It would be nice to get this landed. 😉

PyvesB:

If someone else wants to pick this up, you can simply fetch the commits by running `git fetch origin pull/6508/head:pr-6508`. Address the pending review comments and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

services/packagist/packagist-version.service.js
yookoala: Packagist do not have separate tagged version on different branch. Even the main branch is considered a development branch. The `~dev.json` endpoint is useful for accessing information of branch HEAD, regardless if it is tagged or not. The [current version](https://user-images.githubusercontent.com/91274/152666501-b15fcac7-68f5-49ff-a524-a174d4c73249.png) of [documentation](https://packagist.org/apidoc#get-package-metadata-v2) stated:

> The `p2/$vendor/$package.json` file contains only tagged releases. If you want to fetch information about branches (i.e. dev versions) you need to download `p2/$vendor/$package~dev.json`.

Sometimes the tagged pre-release versions (e.g. v0.1.2) of libraries are published and used in production by others. In which case, the latest version would be "not stable" but is not shown on the `~dev.json` endpoint but is still relevant.

PR #6513, opened 2021-05-19
Remove rate limiting functionality

PyvesB:
Our home grown rate limiting feature recently tripped us over in  #6511. When taking a closer look at the class, both @chris48s and I challenged its usefulness.

Its main intent is probably to guard against denial of service attacks, but it doesn't feel particularly robust. We still handle the requests on the same machine, so the rate-limiting itself could get easily overwhelmed if you simply sent a little more requests. Looking at our metrics for the past 30 days, outside the past week where I was experimenting with different numbers of servers, it only kicked in very occasionally to limit by ~5 to 15 requests/second. However, I'm not convinced it did any positive impact.

The rate limiting classes have no unit tests, are not easily configurable and still have comments referencing our old OVH servers. The rate limiting numbers had not been updated in many years, despite us servicing much more traffic on a very different underlying platform. I'm not sure what the intent of rate limiting per badge type is, and the hardcoded Camo IPs feels brittle. To me, this really is technical debt in our server code at this point.

I suggest we delete this functionality, slim down our server code and save a few CPU cycles per request. A CDN seems much better suited for these kind of tasks, and Cloudfare does allow one rule with our free plan. Let's discuss the details of that offline.


paulmelnikow (approved):
It's not completely clear that Cloudflare as currently configured will handle these kind of attacks, however I agree the CDN is a better place to handle these and that removing this from the server is a reasonable step to take.

chris48s (approved):

calebcartwright (approved):
I suspect part of the motivation for this with OVH was the fact that we publicized the server IPs and malicious traffic could be sent directly to a specific server bypassing the CDN.

That's no longer the case with the runtime in Heroku, so I'll join the party and approve too :laughing:

PR #6567, opened 2021-06-01
increase max-age for download and social badges; affects [twitch]

chris48s:
We serve a lot of download and social badges (collectively download and social categories account for ~25% of the traffic we serve from Heroku) and we serve them with the default max-age (2 mins).
In a lot of cases the upstream providers cache these totals and we also often present them as rounded numbers e.g:

* `downloads | 20k`
* `stars | 3k`

I think we can get away with cacheing these for longer and serve more of these badge requests from downstream cache without noticeably downgrading service for most users. Lets try setting them to 15 mins and see how it goes.
I've decided to manually exclude twitch status from this and actually set a shorter max-age on it as "live now" seems like it need to be more immediate to be useful. I think all other download and social badges can use the new default.

### DIFF
--- a/core/base-service/base.js
+++ b/core/base-service/base.js
@@ -143,6 +143,8 @@ class BaseService {
       license: 3600,
       version: 300,
       debug: 60,
+      downloads: 900,
+      social: 900,
     }
     return cacheLengths[this.category]
   }
--- a/services/twitch/twitch.service.js
+++ b/services/twitch/twitch.service.js
@@ -30,6 +30,8 @@ module.exports = class TwitchStatus extends TwitchBase {
     },
   ]
 
+  static _cacheLength = 30
+
   static defaultBadgeData = {
     label: 'twitch',
     namedLogo: 'twitch',


calebcartwright:
I'm onboard for changing the value here but I'm torn about the 15 minute value. It's true in cases with higher numbers that the incremental diffs within a given 15 minutes block would likely be obfuscated by the rounding, but I think that only applies once one gets to badge messages that exceed 1k

Conversely, if a package only has 50 downloads or a user 25 followers, then it's highly unlikely for that number to change dramatically, and be observed, within that cache window.

However, (as I continue debating myself :smile:) it seems surprising at first blush that we'd have a longer cache period for download badges vs. version badges, given that the former seems more dynamic than the latter.

I'm not necessarily opposed to trying out the 15 minute value and won't block if others are content to proceed, but want to ponder on this a bit before approving myself

calebcartwright (approved):
Okay didn't need as long as I thought I might :laughing: 

Figure it's better to try it and see if there's any actual reports or concerns instead of just hypothesizing

chris48s:
Its a good point. In many cases we would expect these values to change more often than version.

I think the reason it is different in my mind is because I see number of downloads/stars/etc as more of an "indicative" statistic. 491 downloads/followers is "about 500" and 508 is also "about 500". Versions are discrete - version 1.2 is a completely different thing to version 1.3.

That said, I'm certainly not married to 15 minutes. Its a number I made up and we can make it smaller if turns out to be a poor choice :)

PR #6570, opened 2021-06-01
run package tests on node 16, drop node 10 testing

chris48s:
We dropped compatibility for node 10 in version 3.0.0 of badge-maker but we've still been running the tests on it. Meanwhile, Node 16 has been generally available for a couple of months. Time to update the CI build.

### DIFF
--- a/scripts/run_package_tests.sh
+++ b/scripts/run_package_tests.sh
@@ -14,22 +14,21 @@ set -euo pipefail
 node --version
 
 # Install the shields.io dependencies.
-if [[ "$NODE_VERSION" == "v10" ]]; then
-    # Avoid a depcheck error.
-    npm ci --ignore-scripts
-else
-    npm ci
-fi
+npm ci
 
 # Run the package tests.
 npm run test:package
 npm run check-types:package
 
-# Delete the shields.io dependencies.
+# Delete the full shields.io dependency tree
 rm -rf node_modules/
+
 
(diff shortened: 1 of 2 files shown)


scripts/run_package_tests.sh
```diff
@@ -14,22 +14,21 @@ set -euo pipefail
...
 
+
 # Run a smoke test (render a badge with the CLI) with only the package
 # dependencies installed.
 cd badge-maker
+
+npm install # install only the package dependencies for this test
 npm link
 badge cactus grown :green @flat
+rm package-lock.json && rm -rf node_modules/ # clean up package dependencies
```
chris48s, line 34: This had me stumped for a bit. I think what was going on here is:
When we were doing the first `npm ci` under node 10, we were actually using an old version of NPM (5.x ?) which created a separate `node_modules/` under the `badge-maker/` dir. After dropping node 10, now the initial `npm ci` happens using a more recent version which _doesn't_ do this (everything goes into the root `node_modules/`) so then when we come to run this "test" we need to explicitly install (and clean up) in the package dir.

calebcartwright: I'm surprised by the need to cleanup here, but so long as we don't have a chance of false positives where deps are missing for the package but present in the server manifest, which seems to still hold true here

calebcartwright (approved):
lgtm

Issue #6590, opened 2021-06-05
🚀 Simple Icons 5


calebcartwright:
Logos on shields.io are provided by SimpleIcons. We've recently upgraded to [SimpleIcons 5](https://github.com/simple-icons/simple-icons/releases/tag/5.0.0) which includes several icons that have been renamed and some that have been removed removed. Please remember that we are just consumers of the SimpleIcons package and have no control over changes nor removals, so please direct any questions about the Simple Icons changes to the Simple Icons team.

### Renamed
These icons have been renamed. If you use any of these,  you will need to update the logo slug reference in your badge URL (e.g. `?logo=dot-net` --> `?logo=dotnet`)

* from `dot-net` to `dotnet`
* from `a-frame` to `aframe`
* from `about-dot-me` to `aboutdotme`
* from `alibaba-dot-com` to `aboutdotme`
* from `at-and-t` to `atandt`
* from `audio-technica` to `audiotechnica`
* from `b-and-rautomation` to `bandrautomation`
* from `battle-dot-net` to `battledotnet`
* from `blockchain-dot-com` to `blockchaindotcom`
* from `chart-dot-js` to `chartdotjs`
* from `co-op` to `coop`
* from `conda-forge` to `condaforge`
* from `counter-srike` to `counterstrike`
* from `d-wavesystems` to `dwavesystems`
* from `d3-dot-js` to `d3dotjs`
* from `dev-dot-to` to `devdotto`
* from `diagrams-dot-net` to `diagramsdotnet`
* from `digi-eyelectronics` to `digikeyelectronics`
* from `draugiem-dot-lv` to `draugiumdotlv`
* from `eclipsevert-dot-x` to `eclipsevertdotx`
* from `ember-dot-js` to `emberdotjs`
* from `f-droid` to `fdroid`
* from `f-secure` to `fsecure`
* from `ferrarin-dot-v-dot` to `ferrarinv`
* from `fraunhofer-gesellschaft` to `fraunhofergesellschaft`
* from `freedesktop-dot-org` to `freedesktopdotorg`
* from `gog-dot-com` to `gogdotcom`
* from `hotels-dot-com` to `hotelsdotcom`
* from `itch-dot-io` to `itchdotio`
* from `ko-fi` to `kofi`
* from `last-dot-fm` to `lastdotfm`
* from `libraries-dot-io` to `librariesdotio`
* from `mail-dot-ru` to `maildotru`
* from `material-ui` to `materialui`
[...]

Issue #4273, opened 2019-10-31
Migrate to simple-icons slugs


paulmelnikow:
Based on our request at simple-icons/simple-icons#1362 and simple-icons/simple-icons#1520, simple-icons is now shipping a canonical slug to go with each icon. As @chris48s put it in https://github.com/badges/shields/pull/3641#issuecomment-520064227:

> This is nice because it means we've got a consistent slug on each object which is discoverable from the simpleicons website/repository/library. Adopting this would be easier for users because our slugs would match the ones used elsewhere.

However there's a downside:

> The downside is that the slugs we've generated in the past are quite different for a lot of icons, so just switching from `key.toLowerCase().replace(/ /g, '-')` to `simpleIcons[key].slug` would break a lot of badges in the wild. This is a rough diff between our slugs and theirs:
>
> This is a rough diff between our slugs and theirs:

```diff
--- oldslugs	2019-08-09 21:39:22.237364058 +0100
+++ newslugs	2019-08-09 21:40:35.632137692 +0100
@@ -3 +2,0 @@
-dev.to
@@ -6,3 +5 @@
-rollup.js
-webcomponents.org
-.net
+dot-net
@@ -10 +7 @@
-about.me
+about-dot-me
@@ -14,12 +11,12 @@
-adobe-acrobat-reader
-adobe-after-effects
-adobe-audition
-adobe-dreamweaver
-adobe-illustrator
-adobe-indesign
-adobe-lightroom-cc
-adobe-lightroom-classic
-adobe-photoshop
-adobe-premiere
-adobe-typekit
-adobe-xd
+adobeacrobatreader
+adobeaftereffects
+adobeaudition
+adobedreamweaver
+adobeillustrator
+adobeindesign
+adobelightroomcc
+adobelightroomclassic
+adobephotoshop
+adobepremiere
+adobetypekit
+adobexd
@@ -27 +24 @@
-allociné
+allocine
@@ -29 +26 @@
-amazon-aws
+amazonaws
@@ -31 +28 @@
-american-express
+americanexpress
@@ -35 +32 @@
-angular-universal
+angularuniversal
@@ -38 +35 @@
-apache-flink
+apacheflink
@@ -40,2 +37,2 @@
-apple-music
-apple-pay
+applemusic
+applepay
@@ -43,2 +40,2 @@
-arch-linux
-archive-of-our-own
+archlinux
+archiveofourown
@@ -49 +46 @@
-at&t
+at-and-t
[...]

PR #3641, opened 2019-07-05
Update simple-icons names

RedSparr0w:
The main purpose of this PR is to avoid future issues like these:
https://github.com/badges/shields/issues/3639
https://github.com/badges/shields/issues/3297
https://github.com/badges/shields/issues/2980

This PR shouldn't break backwards compatibility,
This will just ignore any characters which are not `a-z` or `0-9` _(case insensitive, even though they should already be lowercased)_,

for example `let’s-encrypt` will become `letsencrypt`.

PR #6591, opened 2021-06-05
update simple-icons to v5 with by-name lookup backwards compatibility

calebcartwright:
Implements the second option described in #6580 to update to v5 of simple-icons while maintaining backwards compatibility with how we've been handling named logos that contain spaces in the icon name/title. Note this means that icons can now be referenced via their slug or via their title, and title's containing spaces can be referenced either by escaping the spaces or replacing them with hyphens.

https://shields-staging-pr-6591.herokuapp.com/badge/foo-bar-blue?logo=linux-foundation

https://shields-staging-pr-6591.herokuapp.com/badge/foo-bar-blue?logo=linux%20foundation

https://shields-staging-pr-6591.herokuapp.com/badge/foo-bar-blue?logo=linuxfoundation

(know the icon loading isn't under the core directory but adding the label since it feels like a "core" item)


chris48s:
:books: Background reading:
- https://github.com/badges/shields/issues/4273
- https://github.com/badges/shields/pull/3641

I think if we're going to (finally) enable the proper SI slugs to be used, that should be the only documented option (but the others retained for backwards compatibility). I think our original plan was to literally hard-code old --> new maps and stop generating the old style slugs but maybe we're too far gone for that and the simplest thing to do is just generate more and more of them forever. .Do you have any thoughts on the tradeoffs?

lib/load-simple-icons.js
```diff
@@ -5,18 +5,23 @@ const { svg2base64 } = require('./svg-helpers')
...
-      default: svg2base64(
-        simpleIcons[k].svg.replace('<svg', `<svg fill="#${simpleIcons[k].hex}"`)
-      ),
-      light: svg2base64(
-        simpleIcons[k].svg.replace('<svg', `<svg fill="whitesmoke"`)
-      ),
-      dark: svg2base64(simpleIcons[k].svg.replace('<svg', `<svg fill="#333"`)),
+    const icon = originalSimpleIcons[key]
+    const title = icon.title.toLowerCase()
+    const legacyTitle = title.replace(/ /g, '-')
```
chris48s, line 16: should this be `title.toLowerCase().replace(/ /g, '-')` ?

calebcartwright:
> Background reading:

Thanks for this. Vaguely remember reading those back in the day but the dots hadn't fully connected for me until now.

> I think if we're going to (finally) enable the proper SI slugs to be used, that should be the only documented option (but the others retained for backwards compatibility)

SI have forced our hand on this with the v5 update. We'd have to really bend over backwards to _not_ support the proper slugs, and I can't imagine why we'd do extra work to avoid that. If we intend to continue supporting icon-by-name as well (regardless of fully or partially via hard coded table), then I guess I'd be fine not documenting/publicizing it.

> I think our original plan was to literally hard-code old --> new maps and stop generating the old style slugs but maybe we're too far gone for that and the simplest thing to do is just generate more and more of them forever. . Do you have any thoughts on the tradeoffs?

At this point I'd be opposed to attempting to maintain a hard coded map of old slugs to the current set.

There'd be far too many, and then if/when the slugs are updated in SI we'd have to go update our maps. We'd also have to deal with questions/requests to maintain others (technically if we were to maintain such a mapping we _could_ include entries for all these [updated](https://github.com/badges/shields/issues/6590) to prevent logos from disappearing on existing badges when we deploy a major version update. I know we aren't obligated to do so, just noting I'm not keen on having to explain-to/debate-with users why we're mapping some but not all.

I'd also worry about the impact on the user experience. At this point we certainly have users that have become accustomed to current by-name reference using the `-` or `%20` delimiters, and I think it'd be really confusing as a user if I've always used that pattern, e.g. `?logo=visual-studio-code`, and that pattern works for a bunch of other logos (assuming hard coded mapping), but when I try to apply that same pattern to some new `Foo Bar` logo it doesn't work, since in this hypothetical we wouldn't have mappings for new logos in the old table, and wouldn't directly support the name reference. Just seems it'd be really confusing/magical why the pattern would work for some/most but not all logos.

At the end of the day, I think it's inevitable that we'll start supporting the by-slug references directly as part of the v5 update, so IMO the question is really whether we want to also continue supporting by-name (and if not then if/how on the BC front).

Also, I'm not quite sure what the concerns or issues are with also supporting the by-name pattern given the v5 context. Feels like the reported issues have historically been driven by challenges associating to slug (no longer an issue IMO with the change of the exports in v5), users attempting and failing to use the actual slug in cases where our coercion failed to match the actual slug, couple quirky names (apostrophes/backticks), and our own locally maintained logos. Anything you can share/remind me of relative to why supporting both would be problematic?

lib/load-simple-icons.js
calebcartwright: let me know if i've misunderstood, but `title` would already be lower cased as part of the initial assignment on the preceding line wouldn't it?

developStorm:
I was working on [a project](https://github.com/developStorm/simple-badges) that shows all the Simple Icons with corrensponding Shields IO badge and experenced a hard time to find the correlation between slug/title of Simple Icons with logo option of Shields IO until I saw this PR (looks like I just happened to stuck in a major version update). 

I've successfully integrated the test build of Shields above into my project, and it seems to be working well. Hope this could be deployed soon!

lib/load-simple-icons.js
developStorm: There may be a compatibility problem here - the `title` may have changed in 5.0 as well. I tried before to use 5.0 icons and do the title transformation as what has been done here, but I actually had to do some extra processing to get it to work properly.

An example would be `Node.JS` (actually almost all badges with dot in title have this problem):

https://shields-staging-pr-6591.herokuapp.com/badge/foo-bar-blue?logo=node-dot-js


https://shields.io/badge/foo-bar-blue?logo=node-dot-js



https://shields-staging-pr-6591.herokuapp.com/badge/foo-bar-blue?logo=node.js


https://shields.io/badge/foo-bar-blue?logo=node.js

calebcartwright: @developStorm - you're referencing breaking changes that SimpleIcons made that's separate from the name vs. slug reference (more details in #6590). We don't attempt to maintain backwards compatibility when SI change their own slugs, and that's unrelated to this thread so going to hide these last two comments as off topic to keep the thread focused

chris48s:
> At this point I'd be opposed to attempting to maintain a hard coded map of old slugs to the current set.

All good points :+1: Being ~2 years further down the path we are already on is significant.

> Anything you can share/remind me of relative to why supporting both would be problematic?

I think the only downside with generating two sets of our own additional identifiers forever as well as the SI-provided slugs would be if there is any way we might generate a conflict or collision between an ID we've generated and one of the proper SI slugs. It looks super-unlikely that would happen but its hard to commit to saying it will never happen with any future release of SI.

For belt-and-braces we could throw an exception if we generate `title` or `legacyTitle` and it is in `Object.keys(originalSimpleIcons).filter(slug => slug !== key)`. If it never happens, great. If it does, the tests fail when we upgrade to the release that causes it and we defer working out what to do about it to us-from-the-future.

calebcartwright:
> It looks super-unlikely that would happen but its hard to commit to saying it will never happen with any future release of SI.

Interesting scenario. The only extra slugs we'd be generating with this proposal is the lowercased version of the title, and the lowercased version of the title with the spaces replaced with hyphens. As of v5 SI no longer permit hyphens in the slugs (and never have had whitespace AFAIK), so the only path I could see for this to occur would be something a case where the title of one icon randomly matches the slug of another in the exports, e.g. something like this

```js
{
  dog: { title: 'Cat', slug: 'dog', ... },
  cat: { title: 'Duck', slug: 'cat', ... },
  ...
}
```

Guess it's not 100% impossible even if it does feel like a stretch, but yes we could certainly include a check for this scenario

ericcornelissen:
Hi all :wave: One of the core maintainers of Simple Icons here. I'm sorry for all the trouble we're causing :sweat_smile:

I wanted to chime in to clarify some things regarding the `slug` value. The `slug` value is specifically intended to avoid problems due to brands with the same name or auto-generated slug (for example [we have two icons both for a brand called "Hive"](https://github.com/simple-icons/simple-icons/blob/7cdc8ac6a3ff0f46857ed490661298b6cfe7d90d/_data/simple-icons.json#L3887-L3897)) So, the `slug` is either [derived from the title deterministically](https://github.com/simple-icons/simple-icons/blob/cae25487b37975758a5753e8711681a179110586/scripts/utils.js#L17-L32), indeed without whitespace and now also without hyphens, or it is custom, unique, and hardcoded in our repository in which case it will always contain an `_`.

So, in short, the above scenario would never occur. Supporting our slug value is a must if you want your users to be able to use all our icons; but using the title or hyphenated title should also always work except for the rare case where multiple brands share a name.

calebcartwright:
No worries @ericcornelissen and thanks for the additional context! Long term I think the changes in v5 will actually be better for us and address some long standing problems. Just a matter of whether/how much we want to do to prevent logos disappearing on badges for some of our existing users :smile:

DenverCoder1:
:+1: Version can be bumped to `5.1.0` as well now

calebcartwright:
@chris48s

Have kept the by-name support in there, but updated docs and the site to remove those mentions and point everyone to the slugs. Let me know if you'd still like add some handling for existing functions. The case @ericcornelissen shared with Hive is a great one, as it demonstrates why we wouldn't want throw an exception on encountering a title that maps to an existing key. At the moment, such cases will mean that we process that particular icon twice on loading, but the end result would still be the same correct icon mapping.

That's obviously doing some unnecessary work and I'm tempted to add the key for presence in _our_ map to avoid it, though held off for the moment since this is a one time/loadup not on a hotpath and because I feel like such an addition would need some additional inline commentary to provide context, which would have us trending towards having more lines of comments than javascript :laughing: 

(also went ahead and updated to v5.1)

lib/load-simple-icons.js
```diff
@@ -5,18 +5,27 @@ const { svg2base64 } = require('./svg-helpers')
...
+    icon.base64 = {
+      default: svg2base64(icon.svg.replace('<svg', `<svg fill="#${icon.hex}"`)),
+      light: svg2base64(icon.svg.replace('<svg', `<svg fill="whitesmoke"`)),
+      dark: svg2base64(icon.svg.replace('<svg', `<svg fill="#333"`)),
     }
+    simpleIcons[key] = icon
+    simpleIcons[title] = icon
+    simpleIcons[legacyTitle] = icon
```
chris48s, line 28: Shall we change the order here.
i.e: if we're going to say the SI slug is the canonical ID and the others are fallbacks, then in the case of a collision like "hive" then if we set the SI slug last then that value is guaranteed to "win" the clash here.

calebcartwright: Good shout. Actually needed to incorporate the key check after all as originally proposed (just skipping instead of throwing exception). The order of the assignment within a given iteration doesn't matter so much, but more about needing the conditional checks otherwise the value would be determined by the last iteration that contains a colliding key. 32554a2a5b7471630d2dff5f61dd9e443b437fef

chris48s (approved):
:+1:  Lets ship it :ship:

0f3eacd3:
Hello, is there an estimated time for deploying this on production? 🙏

PR #6592, opened 2021-06-05
[bStats] Create managed servers badge

weihao:
# Note

**Managed Servers** are children servers running behind a proxy server.


calebcartwright:
Thank you for the PR! I haven't had a chance to look at this too closely, but from a cursory glance it seems highly duplicative of the existing bStats server code.

Is there any reason this couldn't be implemented via a single class and route, by simply adjusting the route to remove `servers` from the base, and add a new param that only accepts `servers|managed_servers`? The value of that param could then be used conditionally as part of the fetch/render concerns

weihao:
@calebcartwright thanks for the feedback, and I agree dynamic routing is a brilliant idea.

However, I am not familiar with shields codebase. Could you share some examples where similar functionality is implemented? What we want to achieve is dynamic routing for `managed_servers`, `players`, `servers`, and maybe more data in the future.

chris48s:
By the looks of it you should be able to merge bstats-servers, bstats-players and bstats-managed-servers into one class with a route like this:

```js
static route = { base: 'bstats`, pattern: ':variant(managed_servers|servers|players)/:pluginid' }
```

Then the `variant` param can be used to determine which endpoint to call etc.

PyvesB:
@weihao did you have a chance to look into @chris48s' suggestions? Would be great to get this progressed. 😉

PyvesB:

If someone else wants to pick this up, you can simply fetch the commits by running `git fetch origin pull/6592/head:pr-6592`. Address the pending review comments and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

PR #853, opened 2017-01-10
Adds maven-metadata.xml support.

ArsenArsen:
Adds maven-metadata.xml support using the format `/maven/v/http://example.com/path/to/maven-metadata.xml.format`

This universally supports all Maven artifacts by the current format, in theory.

PR #6594, opened 2021-06-07
Use release/latest fields of [mavenmetadata] instead of last version in the versions, add according query parameter

### LINKED ISSUE #6188: Maven Metadata not showing latest version

- [ ] My own instance

:beetle: **Description**

Badges generated from Maven metadata use the last element of the versions list which may not be the latest version. In this case the latest version is `0.0.12`, but the badge shows version `0.0.9`.

:link: **Link to the badge**

https://img.shields.io/maven-metadata/v.svg?metadataUrl=https%3A%2F%2Fplugins.gradle.org%2Fm2%2Fcom%2Fparmet%2Fbuf-gradle-plugin%2Fmaven-metadata.xml

:bulb: **Possible Solution**

Instead of taking the [last element of the list](https://github.com/badges/shields/blob/fcf6678a127c9679b0d68284b860181c2580fe26/services/maven-metadata/maven-metadata.service.js#L58), use the `metadata.version` or `metadata.versioning.latest` or `metadata.versioning.release` element: https://plugins.gradle.org/m2/com/parmet/buf-gradle-plugin/maven-metadata.xml.

CommanderTvis:
The problem is that it is invalid to just take the last field from Maven metadata. 

This PR solves it by taking `latest` field by default and `release` field if the new query parameter is specified so.

resolves #6188

### DIFF
--- a/services/maven-metadata/maven-metadata.service.js
+++ b/services/maven-metadata/maven-metadata.service.js
@@ -7,14 +7,14 @@ const { BaseXmlService } = require('..')
 
 const queryParamSchema = Joi.object({
   metadataUrl: optionalUrl.required(),
+  latestOrRelease: Joi.string().valid('release', 'latest').optional(),
 }).required()
 
 const schema = Joi.object({
   metadata: Joi.object({
     versioning: Joi.object({
-      versions: Joi.object({
-        version: Joi.array().items(Joi.string().required()).single().required(),
-      }).required(),
+      latest: Joi.string().required(),
+      release: Joi.string().required(),
     }).required(),
   }).required(),
 }).required()
@@ -35,6 +35,7 @@ module.exports = class MavenMetadata extends BaseXmlService {
       queryParams: {
         metadataUrl:
           'https://repo1.maven.org/maven2/com/google/code/gson/gson/maven-metadata.xml',
+        latestOrRelease: 'latest',
       },
       staticPreview: renderVersionBadge({ version: '2.8.5' }),
     },
@@ -52,10 +53,19 @@ module.exports = class MavenMetadata extends BaseXmlService {
     })
   }
 
-  async handle(_namedParams, { metadataUrl }) {
+  async handle(_namedParams, { metadataUrl, latestOrRelease }) {
     const data = await this.fetch({ metadataUrl })
+    let v
+    if (
+      latestOrRelease === 'latest' ||
+      typeof latestOrRelease === 'undefined'
+    ) {
+      v = data.metadata.versioning.latest
+    } else if (latestOrRelease === 'release') {
+      v = data.metadata.versioning.release
+    }
     return renderVersionBadge({
-      version: data.metadata.versioning.versions.version.slice(-1)[0],
+      version: v,
     })
   }
 }
--- a/services/maven-metadata/maven-metadata.tester.js
+++ b/services/maven-metadata/maven-metadata.tester.js
@@ -39,6 +39,34 @@ t.create('version ending with zero')
   )
   .expectBadge({ label: 'maven', message: 'v1.30' })
 
+t.create('release version is handled properly')
+  .get(
+    '/v.json?metadataUrl=https://repo1.maven.org/maven2/mocked-group-id/mocked-artifact-id/maven-metadata.xml&latestOrRelease=release'
+  )
+  .intercept(nock =>
+    nock('https://repo1.maven.org/maven2')
+      .get('/mocked-group-id/mocked-artifact-id/maven-metadata.xml')
+      .reply(
+        200,
+        `
+      <metadata>
+        <groupId>mocked-group-id</groupId>
+        <artifactId>mocked-artifact-id</artifactId>
+        <versioning>
+          <latest>1.30</latest>
+          <release>1.29</release>
+          <versions>
+            <version>1.29</version>
+            <version>1.30</version>
+          </versions>
+          <lastUpdated>20190902002617</lastUpdated>
+        </versioning>
+      </metadata>
+      `
+      )
+  )
+  .expectBadge({ label: 'maven', message: 'v1.29' })
+
 t.create('invalid maven-metadata.xml uri')
   .get(
     '/v.json?metadataUrl=https://repo1.maven.org/maven2/com/google/code/gson/gson/foobar.xml'


calebcartwright:
hi @CommanderTvis :wave: thanks for the PR!

I've updated your PR title to follow our [required conventions](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#running-service-tests-in-pull-requests) so that the tests will be run, and as you'll notice there are failures.

I have a few asks of you:

* Could you please provide some more context on what you are proposing be changed, and why? We don't require PRs have a preceding issue, but PRs without an associated issue should provide sufficient information about the change
* Please utilize the PR description and, if needed, comments to convey the additional information requested. The PR title should be a succinct summary (no need to enumerate every change, nor technical details in the title!)
* Please review the test failure(s) and adjust the proposed changes accordingly. More details on the failure(s) can be found in CircleCI (e.g. https://app.circleci.com/pipelines/github/badges/shields/7087/workflows/cd8b8a44-c769-48a3-a26b-1b3af0392ce1/jobs/145824)

calebcartwright:
Also cc @PyvesB for Java expertise given the Maven touch point

CommanderTvis:
@calebcartwright I've updated the description

calebcartwright:
> @calebcartwright I've updated the description

Thanks, though that still doesn't really answer my question. This is a badge that has been around and in use for a long time, so I'd ask for more information, examples, details about why you think it's invalid vs. just labeling it as such.

CommanderTvis:
> > @calebcartwright I've updated the description
> 
> Thanks, though that still doesn't really answer my question. This is a badge that has been around and in use for a long time, so I'd ask for more information, examples, details about why you think it's invalid vs. just labeling it as such.

Here's my use case: https://img.shields.io/maven-metadata/v?label=Space&metadataUrl=https%3A%2F%2Fmaven.pkg.jetbrains.space%2Fmipt-npm%2Fp%2Fsci%2Fmaven%2Fspace%2Fkscience%2Fkmath-core%2Fmaven-metadata.xml

The metadata looks like this:
```xml
<metadata>
<groupId>space.kscience</groupId>
<artifactId>kmath-core</artifactId>
<versioning>
<latest>0.3.0-dev-12</latest>
<release>0.3.0-dev-12</release>
<versions>
<version>0.3.0-dev-12</version>
<version>0.3.0-dev-10</version>
<version>0.3.0-dev-8</version>
<version>0.3.0-dev-3</version>
<version>0.3.0-dev-2</version>
<version>0.2.1</version>
<version>0.2.0</version>
</versions>
<lastUpdated>20210522171951</lastUpdated>
</versioning>
</metadata>
```

The correct latest version is `0.3.0-dev-12`, but the badge takes `0.2.0`.

CommanderTvis:
The [specification](https://maven.apache.org/ref/3.3.9/maven-repository-metadata/repository-metadata.html) of Maven metadata says nothing about the order of `versions`, but the `latest` is really the latest version.
image

calebcartwright:
Thanks @CommanderTvis! That's exactly what I was looking for.

This looks like another case of #6188 in that JetBrains, like Gradle, aren't sorting the versions like Maven Central. It seems a given that something needs to change, as the status quo seems to be more and more of a "Maven Metadata but really just for Maven Central", but I'm unsure of the potential impact on any existing badge usage (e.g. could this change result in a different/unexpected version value in the badge message for cases using different/older maven versions, different providers/hosts, etc.)

CommanderTvis:
> Thanks @CommanderTvis! That's exactly what I was looking for.
> 
> This looks like another case of #6188 in that JetBrains, like Gradle, aren't sorting the versions like Maven Central. It seems a given that something needs to change, as the status quo seems to be more and more of a "Maven Metadata but really just for Maven Central", but I'm unsure of the potential impact on any existing badge usage (e.g. could this change result in a different/unexpected version value in the badge message for cases using different/older maven versions, different providers/hosts, etc.)

For libraries published with Maven, nothing will change.

calebcartwright:
> For libraries published with Maven, nothing will change.

My point is that while the odds of impact may be minuscule, I don't think it's possible to claim there's a 0% risk of impacting existing badges because this badge can be used with any arbitrary xml file

We take backwards compatibility with our badges seriously, and that includes not breaking badge routes and trying to avoid server-side changes causing existing badges to display differently (different colors, different textual values, etc.). Sometimes such changes are unavoidable, but when they are (or could be) we do need to think through potential impacts and try to communicate the pending changes as proactively as possible.

It's pretty clear that different registries will handle the same maven information differently/contain different file contents for the same package, so I definitely think it's worth taking a moment to see if anyone (maintainer or community) is aware of any potential issues or impacts.

In the interim, please note that you can alternatively use our existing DynamicXML badge today to go ahead and get a working badge, e.g.

https://img.shields.io/badge/dynamic/xml?label=Space&query=//metadata/versioning/latest&url=https%3A%2F%2Fmaven.pkg.jetbrains.space%2Fmipt-npm%2Fp%2Fsci%2Fmaven%2Fspace%2Fkscience%2Fkmath-core%2Fmaven-metadata.xml

CommanderTvis:
> > For libraries published with Maven, nothing will change.
> 
> My point is that while the odds of impact may be minuscule, I don't think it's possible to claim there's a 0% risk of impacting existing badges because this badge can be used with any arbitrary xml file
> 
> We take backwards compatibility with our badges seriously, and that includes not breaking badge routes and trying to avoid server-side changes causing existing badges to display differently (different colors, different textual values, etc.). Sometimes such changes are unavoidable, but when they are (or could be) we do need to think through potential impacts and try to communicate the pending changes as proactively as possible.
> 
> It's pretty clear that different registries will handle the same maven information differently/contain different file contents for the same package, so I definitely think it's worth taking a moment to see if anyone (maintainer or community) is aware of any potential issues or impacts.
> 
> In the interim, please note that you can alternatively use our existing DynamicXML badge today to go ahead and get a working badge, e.g.
> 
> https://img.shields.io/badge/dynamic/xml?label=Space&query=//metadata/versioning/latest&url=https%3A%2F%2Fmaven.pkg.jetbrains.space%2Fmipt-npm%2Fp%2Fsci%2Fmaven%2Fspace%2Fkscience%2Fkmath-core%2Fmaven-metadata.xml
> 

Oh, that's a nice workaround. Thank you. Personally, I don't need it then.

PyvesB:
Whilst I'm on board with the backwards compatibility considerations, in this case one could argue that taking the last item in a list of versions with no defined order is a bug.

In an ideal world, I would want us to collate all the requests for this badge for a day or so, and write a little script that compares the versions results of both approaches. We could then measure the impact and see if it would lead to unintended results in some cases. However, we don't store the requests made to Shields.io, if we wanted to do this we would have to use a much smaller time frame and tail the servers logs in real time.

calebcartwright:
> Whilst I'm on board with the backwards compatibility considerations, in this case one could argue that taking the last item in a list of versions with no defined order is a bug.

Thanks @PyvesB. I was curious whether there would be any well known cases where the `latest` and `release` elements simply wouldn't exist in the XML and if this change would in turn cause `invalid response data` badges. For example packages that had only published snapshots or prereleases, or maybe a system like Bintray or Nexus was known to elide those elements (for whatever odd reason).  If both are always there, and based on the specification (for both v3 and v2) it seems like they should be, I'm struggling to follow why we were using the `versions` collection in the first place :man_shrugging: 

Took a bit of digging but looks like it was added in #853 not too terribly long ago (Paul ended up grabbing the commits on another PR to help the author get it across the finish line, but #853 was the initial implementation). Based on the dialog/details, or lack thereof, I assume this was indeed more likely to be an oversight vs. a deliberate need to avoid `latest` and `release`.

services/maven-metadata/maven-metadata.service.js
```diff
@@ -52,10 +53,19 @@ module.exports = class MavenMetadata extends BaseXmlService {
...
+      typeof latestOrRelease === 'undefined'
+    ) {
+      v = data.metadata.versioning.latest
+    } else if (latestOrRelease === 'release') {
+      v = data.metadata.versioning.release
+    }
     return renderVersionBadge({
-      version: data.metadata.versioning.versions.version.slice(-1)[0],
+      version: v,
     })
```
calebcartwright, line 69: Just FYI, a fairly common pattern in our service classes is to utilize another function `transform` that handles the concern of extracting the target elements from the raw response, and if necessary mapping or transforming that. That's not necessary here due to the simplicity, but sharing for future awareness as it's easier to test and often a bit cleaner too.

We also default our version badges to showing the latest stable/release version (`release` here), with the `include_prereleases` query param available for users that want to display the absolute latest regardless of stable/release vs. prerelease. However, even in the default case if there are no "release" versions, we will still automatically fallback to a prerelease-type version vs. throwing an error.

_If_ both the `latest` and `release` elements are always present and already reflecting that pattern, then we should be fine with a simple conditional. However, if `release` is sometimes absent then we'd need to account for that fallback pattern too

calebcartwright (commented):
Thanks again for this! The gist of the change looks good, though will need to tweak a few things to be more consistent with how Shields handles version badges.

Details left inline, one of the other version badges (e.g. [clojars](https://github.com/badges/shields/blob/master/services/clojars/clojars-version.service.js)) may be a helpful reference as well

CommanderTvis:
Closed the PR because I'm satisfied with the Dynamic XML workaround.

PR #6628, opened 2021-06-12
improve [MavenCentral], [MavenMetadata], and [GradlePluginPortal]

anatawa12:
See https://github.com/badges/shields/pull/6449#issuecomment-853347016


anatawa12:
Question: Is there any way to test `expectBadge` with following redirections.

PyvesB:
> 
> 
> Question: Is there any way to test `expectBadge` with following redirections.

Generally we use `expectRedirect` to check that the redirect is setup correctly, and only test the main path via `expectBadge`.

anatawa12:
@PyvesB Thank you. So, I removed old tests for maven central.

PyvesB:
Thanks, will take a look in the coming days! :)

PyvesB:
I've done a first review pass and test in the Heroku app. For anyone reading through this PR, there's been a bit of scope creep: it's not only the refactor of the redirectors mentioned in https://github.com/badges/shields/pull/6449#issuecomment-853347016, support for version suffix filters was added as well.

Looking at things in the review app, searching for "maven" displays the badges with a huge bit of vertical spacing, due to the long titles. Is anyone else observing the same problem? See screenshot below.

Screenshot 2021-06-22 194119

The only badge that currently supports `versionPrefix` is Maven Central.  However, we've ended up in a inconsistent state with regard to the URL structures in this pull request:
* for Maven Metadata, `versionPrefix` is a query parameter, whereas it's a path element in Gradle and Maven Central.
* for Gradle and Maven Central `versionPrefix` is a path element whereas `versionSuffix` is a query parameter.

Now that we're using redirectors, it should be easy enough to resolve these inconsistencies. Switching everything to query parameters across all three badge types would make most sense to me.

anatawa12:
> The only badge that currently supports `versionPrefix` is Maven Central. However, we've ended up in a inconsistent state with regard to the URL structures in this pull request:
> 
> * for Maven Metadata, `versionPrefix` is a query parameter, whereas it's a path element in Gradle and Maven Central.
> * for Gradle and Maven Central `versionPrefix` is a path element whereas `versionSuffix` is a query parameter.
> 
> Now that we're using redirectors, it should be easy enough to resolve these inconsistencies. Switching everything to query parameters across all three badge types would make most sense to me.

I agree with that so I changed all example to use versionPrefix of queryParams. versionPrefix for gradle-plugin-portal is added by this PR so I removed support with path.

> I've done a first review pass and test in the Heroku app. For anyone reading through this PR, there's been a bit of scope creep: it's not only the refactor of the redirectors mentioned in [#6449 (comment)](https://github.com/badges/shields/pull/6449#issuecomment-853347016), support for version suffix filters was added as well.
> 
> Looking at things in the review app, searching for "maven" displays the badges with a huge bit of vertical spacing, due to the long titles. Is anyone else observing the same problem? See screenshot below.
> 
> Screenshot 2021-06-22 194119

I found that we can generate badges without some query parameters in web UI so, To resolve this problem, I removed examples without versionPrefix or versionSuffix.

services/gradle-plugin-portal/gradle-plugin-portal.service.js
```diff
@@ -7,18 +7,21 @@ module.exports = redirector({
...
+        versionSuffix: '.1',
+        versionPrefix: '0.10',
+      },
       namedParams: {
         pluginId: 'com.gradle.plugin-publish',
       },
       staticPreview: {
-        label: 'plugin portal',
-        message: 'v0.14.0',
+        label: 'maven-central',
```
PyvesB, line 23: Shouldn't this have remained `plugin portal`?

services/maven-metadata/maven-metadata.service.js
```diff
@@ -52,10 +56,25 @@ module.exports = class MavenMetadata extends BaseXmlService {
...
+    }
+    const version = versions[0]
+    // if the filter returned no results, throw a NotFound
+    if (
+      versionPrefix !== undefined &&
+      versionSuffix !== undefined &&
+      version === undefined
+    )
+      throw new NotFound({
+        prettyMessage: 'version prefix or suffix not found',
```
PyvesB, line 76: Could you please add a test for this case?

PyvesB (commented):
Thanks, this is almost ready to go as far as I can tell!

> I found that we can generate badges without some query parameters in web UI so, To resolve this problem, I removed examples without versionPrefix or versionSuffix.

Given that you've removed the examples without the filters, would it be worth being extra clear about them being optional in the remaining examples? You can add a bit of documentation in the UI, you can take a look at our Discord implemention to see how that's done:
https://github.com/badges/shields/blob/5ca4089311192348f06048ccfc2ab01bf2a288ea/services/discord/discord.service.js#L11

Maybe something along the following lines:

> `versionPrefix` and `versionSuffix` allow narrowing down the range of versions the badge will take into account, but they are completely optional.

anatawa12:
@PyvesB I've fixed gradle-plugin-portal and added documentation and tests.

PyvesB (approved):
I noticed a minor bug in our frontend, only the first query parameter has an `e.g.` in front of it, which looks a bit weird:
Screenshot 2021-06-30 081030

However, that's unrelated to your work, just happened to notice it whilst testing your pull request. :)


Anyway, thanks for working on this, it's good to go! 👍🏻

PyvesB:
I was a little slow on the review, so I've tried to be fast on the deployment side: this is already live on our production servers! 🍾

Issue #6629, opened 2021-06-13
npm rating


atian25:
:clipboard: **Description**

- Which service is this badge for e.g: NPM
- What sort of information should this badge show?
  - Rating | 80
  - Quality | 100%  (or convert to star)
  - Popularity | 44%  (or convert to star)
  - Maintenance | 100% (or convert to star)

:link: **Data**

- Is there a public API?  yes
- Does the API requires an API key?  no
- Link to the API documentation: https://api-docs.npms.io/#api-Package-GetPackageInfo

![image](https://user-images.githubusercontent.com/227713/121792287-7d1f6d80-cc25-11eb-80fb-ed605cb3bd8b.png)

:microphone: **Motivation**

- https://packagequality.com/ don't support scope package, such as `@eggjs/router`
- npms.io provide https://npms.io/about
- npm itself use npms.io as well

![image](https://user-images.githubusercontent.com/227713/121792344-2d8d7180-cc26-11eb-9650-327e397f7cc3.png)

PR #6630, opened 2021-06-13
[npmsioscore] Support npm score

atian25:
- How it works: https://npms.io/about

- Routes:
  - npms-io/final-score/vue
  - npms-io/final-score/@vue/cli
  - npms-io/maintenance-score/@vue/cli
  - npms-io/popularity-score/@vue/cli
  - npms-io/quality-score/@vue/cli

close https://github.com/badges/shields/issues/6629


services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
...
+const { BaseJsonService } = require('..')
+
+// https://api-docs.npms.io/#api-Package-GetPackageInfo
+const responseSchema = Joi.object({
+  score: Joi.object({
+    final: Joi.number().required(),
+    detail: Joi.object({
+      maintenance: Joi.number().required(),
+      popularity: Joi.number().required(),
+      quality: Joi.number().required(),
```
calebcartwright, line 14: We do have various schema helpers for common functions, but seems these will be unsigned decimal values, where we often do this to be very explicit:
```suggestion
    final: Joi.number().min(0).required(),
    detail: Joi.object({
      maintenance: Joi.number().min(0).required(),
      popularity: Joi.number().min(0).required(),
      quality: Joi.number().min(0).required(),
```

services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
+'use strict'
```
calebcartwright, line 1: I'm not crazy about the file naming and framing this as npm since the service is npms.io, not npmjs.com. It's important in the badge routes, code structure, etc. that we're very explicit about the corresponding service, and unlike our existing npm badges (the ones under the `services/npm` directory), this one is not from the npm registry.

services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
...
+  }),
+}).required()
+
+module.exports = class NpmRating extends BaseJsonService {
+  static category = 'rating'
+
+  static route = {
+    base: 'npm',
+    pattern:
+      'rating/:type(maintenance|popularity|quality)?/:scope(@.+)?/:packageName',
```
calebcartwright, line 25: Few things on the route:

* We don't need and shouldn't have the `rating` prefix. I imagine this was added given the bundling under `npm`, but since that's changing the base will be something like `npms` or `npmsio` and the extra route segment becomes superfluous. 
* Instead of making the type/kind/variant an optional param and assuming final in the absence, we should instead simply include `final` in the allowed.

These changes ensure this route will follow our [route design guidelines](https://github.com/badges/shields/blob/master/doc/badge-urls.md)

services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  static examples = [
+    {
+      title: 'NPM Rating',
```
calebcartwright, line 30: The title should reflect the same name changes discussed above, and be `npms` or `npms.io` for a bigger visual distinction from the `npm` badges

services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
...
+      namedParams: { packageName: 'egg' },
+      staticPreview: this.render({ score: 0.9711 }),
+      keywords: ['node'],
+    },
+    {
+      title: 'NPM Rating (Popularity)',
+      namedParams: { type: 'popularity', packageName: '@vue/cli' },
```
calebcartwright, line 37: If we're going to have an example package that is scoped let's be sure to use the `scope` and `packageName` params properly :wink:

services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
...
+    {
+      title: 'NPM Rating (Maintenance)',
+      namedParams: { type: 'maintenance', packageName: 'command' },
+      staticPreview: this.render({ type: 'maintenance', score: 0.222 }),
+      keywords: ['node'],
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'Rating',
```
calebcartwright, line 56: we use lower case for label text

```suggestion
    label: 'rating',
```

services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  static render({ type, score }) {
+    return {
+      label: !type ? 'Rating' : type[0].toUpperCase() + type.substring(1),
+      message: starRating(score * 5),
+      color: score > 0.3 ? 'brightgreen' : 'red',
+    }
+  }
```
calebcartwright, line 65: The star rating seems wrong to me for representing a 0-100 percentage, especially as the only display format. Stars are really about a representation of user supplied feedback, whereas all of these numbers are really more of a score or analysis.

I probably wouldn't try to stand in the way of having stars as an opt-in formatting option if other maintainers think that's worthwhile, but I definitely don't think it should be the default. The values seem to be a canonical candidate for percentages to me.

services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
...
+    detail: Joi.object({
+      maintenance: Joi.number().required(),
+      popularity: Joi.number().required(),
+      quality: Joi.number().required(),
+    }),
+  }),
+}).required()
+
+module.exports = class NpmRating extends BaseJsonService {
+  static category = 'rating'
```
calebcartwright, line 20: I don't feel like this is the right category. Rating badges are really intended to reflect the aggregate picture of reviews/ratings provided directly users, whereas npms is really more of a quantitative analysis that takes into account various attributes of a package. This are much more akin to quality and accordingly `analysis`feels a better fit

services/npm/npm-rating.service.js
calebcartwright: The logic could also be performed a bit more clearly and expressively by utilizing local variables, and we also should not be changing the case of the label text.

Finally, I think we probably want to use some kind of color scale. Switching to a percentage based formatting will make this easier since the scale is already provided, but a custom color scale can be defined as well

services/npm/npm-rating.service.js
```diff
@@ -0,0 +1,82 @@
...
+    const url = `https://api.npms.io/v2/package/${encodeURIComponent(slug)}`
+
+    const json = await this._requestJson({
+      schema: responseSchema,
+      url,
+      errorMessages: { 404: 'package not found or too new' },
+    })
+
+    let score = json.score.final
+    if (type) score = json.score.detail[type]
```
calebcartwright, line 78: expect the bots/linters will bark at this, as we don't typically permit this style of conditional

services/npm/npm-rating.tester.js
```diff
@@ -0,0 +1,42 @@
...
+const { isStarRating } = require('../test-validators')
+const t = (module.exports = require('../tester').createServiceTester())
+
+t.create('score of vue').get('/rating/vue.json').expectBadge({
+  label: 'Rating',
+  message: isStarRating,
+  color: 'brightgreen',
+})
+
+t.create('score of @vue/cli').get('/rating/@vue/cli.json').expectBadge({
```
calebcartwright, line 12: I realize that what you've done here has been done elsewhere in other npm tests, but for awareness the goal with the test title is much more about describing the nature of the test/the scenario; not the package target.

i.e. the scenario here is testing that the npms badge will handle a scoped package, doesn't really matter what that package is

calebcartwright (commented):
Thank you for the PR @atian25!

The first pass has the foundations in place, but we'll need to iterate on some of the specifics to get this to a mergable place. Details left inline below

calebcartwright:
Closing and reopening to unstick CI

services/npm/npm-rating.service.js
atian25: In fact, npmjs.com is using the score service and search service which provide by npms.io

and I'm worried about the typo of `npms` vs `npm` by user.

so that's why I put it to `services/npm`,

anyway, I'm open to moving to another directory, what's your options?

image

services/npm/npm-rating.service.js
atian25: - same to the reply above, npms vs npm
- will update to `rating/final`

services/npm/npm-rating.service.js
atian25: little help please.

when I change to `namedParams: { type: 'popularity', scope: '@vue', packageName: 'cli' },`

then got error:

```bash
Error: In example for NpmRating at index 1, expected "scope" to match "@.+", but got "%40vue"
```

how could I unescape it?

services/npm/npm-rating.service.js
atian25: it seems `eslint` is happy when `pre-commit`

services/npm/npm-rating.service.js
calebcartwright: > In fact, npmjs.com is using the score service and search service which provide by npms.io

A few things to keep in mind:

* npms.io is the provider/upstream service of the data from Shields' perspective
* The npm registry is _currently_ a consumer of some, but not all, of that data
* We don't really concern ourselves with who other consumers of a service are, nor what other systems that service integrates with behind the scenes (even if it's well known information)
* npm could always change their provider (or user their own internal implementation), or drop it altogether.
* We have users that leverage self-hosted Shields to be able to generate badges for private packages on the npm registry, and these new badges utilizing npms won't work for private packages, but it would be a potential source of confusion if we made it seem like these new badges were native to the npm registry
    * In a similar vein, folks can self-host their own registry instance too which I also believe npms cannot support 

Users need to go the npms to see the raw data, including things like `final` which AFAIK is not something the npm registry has even incorporated.

> and I'm worried about the typo of npms vs npm by user.

Could you elaborate on your concern?

Badge urls tend to be a one time action for users, not something they type on a regular bases. The only time one would be type this is when using the Shields.io site to search for badges, in which cases typing `npm` would list all the badges (including the new ones proposed in this PR), and a user typing `npms` would just see the new badges. Additionally, the keywords defined on the examples in the service class can be used to add additional search terms, which would eliminate any typo risk in my mind.

Shields.io also supports building and exporting badge urls, but even in the event of a user handcrafting a url and having a typo, there is a loud indicator and an easy fix.

atian25: ok, I will move it to `npms-io`

services/npm/npm-rating.service.js
calebcartwright: Please take a look at the guidelines I linked in my previous response if you haven't already. `rating` should be dropped, both because these aren't really rating badges and because the noun portion will be `maintenance`, `final`, etc.

atian25: certainly, read.

As discussed recently above, will move to `npms`, so I will drop rating.

services/npm/npm-rating.service.js
calebcartwright: Try adding the `pattern` key to the example object

https://github.com/badges/shields/blob/de019c2e529b9b3d20be953f0539563581215112/services/npm/npm-version.service.js#L30-L35

services/npm/npm-rating.service.js
calebcartwright: cool deal. if they're happy i'm happy :smile:

services/npm/npm-rating.service.js
atian25: change to `npms.io`

atian25:
@calebcartwright review again?

only except one: percentage vs star (will modify it later)

calebcartwright:
> @calebcartwright review again?
> 
> only except one: percentage vs star (will modify it later)

Thanks for the dialog and quick updates! I'm signing out for the night but sure another maintainer or myself will pick it up in the near future

services/npm/npm-rating.service.js
atian25: are there any existing examples for the badge style?

atian25: upgrade, now support `msg_type=grade | percentage | star` for user to choose

atian25:
> > @calebcartwright review again?
> > only except one: percentage vs star (will modify it later)
> 
> Thanks for the dialog and quick updates! I'm signing out for the night but sure another maintainer or myself will pick it up in the near future

@calebcartwright now support msg_type=grade | percentage | star for user to choose

atian25:
sorry, any progress on this?

calebcartwright:
Thank you for making the update to be explicit about the service being npms.io!

However, there's some remaining unaddressed items, with the core concern being related to formatting. There are four main things that I'd like to see changed, and below the list I've included the context and rationalization for each.

1. Let's make the nouns more explicit (e.g. `final-score` instead of `rating`, `maintenance-score`)
2. Let's use standard Shields formatting for the percentage option (including color and precision)
3. Let's drop the custom formatting options, as star and grade formatting styles do not feel applicable for these badges
4. Let's drop the "rating" terminology
 
#### 1. More Explicit Nouns
The supported npms-derived data points the proposed badges would provide are the details of the scores as determined by npms' analysis: the three detailed scores for `quality`, `maintenance`, and `popularity`, along with the `final` score which I presume is a combination of those. npms.io provides a lot of data points beyond the scores, however, and we've even had a few previous requests for other data points. By using explicit nouns we avoid any ambiguity or subjective mappings to terminology (like "rating") not used by/within npms.io itself, and it also gives us room to grow should additional badges for the service be needed

#### 2. Percentage formatting
These scores are presented as percentages within the npms.io site, and unsurprisingly the values in the API response are decimals between 0 and 1 clearly representing that percentage (fwiw in the places where the npm.js registry site displays the subset of these npms values it does so as percentages as well).

Accordingly, this feels like a canonical case for using Shields percentage type formatting for the message. This is done by rounding to the nearest whole number and leveraging our `coveragePercentage` color formatter which can be imported from `color-formatters`. Doing this here will ensure that the percentage messages are consistent with how Shields.io does these types of badges, which is important because consistency is a core tenant of the project.

#### 3. Drop star/grade formatting

These two formatting types do not make sense in the context of an npms analysis score percentage and they should be removed.

We support the grade-based formatting because we provide badges for upstream services, like CodeClimate, that define and determine the grade for a project within their system, and that grade is specified in their API response. Shields does not itself internally define a grade scale for the results of some upstream service, nor should we (it's not our job to to qualify their analyses). If npms.io had their own mapping of their analysis scores to a tiered grading scale then we could certainly incorporate that within the badges, but we can't make up a grading scale on their behalf.

The star formatting just doesn't make sense in this context. These npms badges reflect the numerical results of an internal, automated analysis npms.io performs by inspecting myriad attributes of a project including (but not limited to) stats/metadata of the package on the registry, build results and code coverage, issue tracker elements, and social items like stars/forks. This is inherently an analysis and not a rating. If it helps, our ratings badges and the star formatting are for services that allow users to directly rate a project/package, analogous to how you could rate an app in your app store. Technically we _could_ use any formatting option for any badge that worked with numeric types, but we don't do this today (star formatting isn't used anywhere outside true rating badges) and we shouldn't start now IMO.

If you feel strongly about needing to be able to have a star and/or grade format on a badge for an npms score, then that would be best handled by implementing a small custom Endpoint badge after we merge this PR to provide npms support.

#### 4. Drop rating terminology
This has largely been covered in preceding conversation, but these badges are not rating badges within the Shields.io context, so let's drop the existing usages of "rating" (badge label, params, etc.). As an example, a better badge label would be something like `score` as the default and with the `final-score`, and could be set to the detailed variants (e.g. `maintenance` or `maintenance score`) for the others. 

<hr />
<br />
As a final comment, remember that the `keywords` key in the example listings adds searchable terms which aid in discovery of the badges from our site. Feel free to extend this with options beyond the current `node` if you think there are any other worthwhile search terms, and that you could create a single variable to define the terms and reuse that variable in each of the example objects

atian25:
ok, got it, thanks for the long explains, will fix it later.

atian25:
@calebcartwright updated

services/npms-io/score.service.js
```diff
@@ -0,0 +1,85 @@
...
+
+const keywords = ['node', 'npm score']
+
+module.exports = class NpmsIO extends BaseJsonService {
+  static category = 'analysis'
+
+  static route = {
+    base: 'npms-io',
+    pattern:
+      ':type(final|maintenance|popularity|quality)/:scope(@.+)?/:packageName',
```
calebcartwright, line 28: I know we have a couple existing badge routes that are just `/serviceName/quality/:params` but I still think we'd be better served by having more explicit nouns here, e.g.:

```suggestion
      ':type(final|maintenance|popularity|quality)-score/:scope(@.+)?/:packageName',
```

This is partly because some of the types look/sound strange (particularly `npms-io/final/...`, and also because there are other non-score data points under those exact same terms (maintenance, popularity, quality) and we could potentially need to provide badges for in the future (we have open requests for at least one already).

Note the test routes would need to be updated accordingly as well

services/npms-io/score.service.js
```diff
@@ -0,0 +1,85 @@
...
+      maintenance: numberSchema,
+      popularity: numberSchema,
+      quality: numberSchema,
+    }),
+  }),
+}).required()
+
+const keywords = ['node', 'npm score']
+
+module.exports = class NpmsIO extends BaseJsonService {
```
calebcartwright, line 22: We need to make the file name and the service name match up, both for consistency and so that the test filters will work correctly.

Let's rename this file to `npmsio-score.service.js` (or `npms-io-score.servive.js`, your call), update the tester file name accordingly to match, and then rename the class too:

```suggestion
module.exports = class NpmsIOScore extends BaseJsonService {
```

calebcartwright (commented):
Thanks for the updates @atian25, this is looking good! Couple relatively minor items remaining and then it should be good to merge

atian25:
> Thanks for the updates @atian25, this is looking good! Couple relatively minor items remaining and then it should be good to merge

sure, updated

calebcartwright:
> sure, updated

Thanks, that looks like everything but the file renames

atian25:
> > sure, updated
> 
> Thanks, that looks like everything but the file renames

ok, done.

calebcartwright (approved):

PR #6636, opened 2021-06-17
[Sonar] Fix invalid fetch query to sonarqube >=6.6

### LINKED ISSUE #5882: Sonarqube 8.5 quality gate parameter componentKey

- [X] My own instance (docker instance)

:beetle: **Sonarqube 8.5 quality gate parameter componentKey**

When retrieve the metrics from sonarqube instance, url used parameter named componentKey

`GET /api/measures/component?componentKey=org.app%3AmyApp&metricKeys=alert_status`

It seem that the parameter *componentKey* have been renamed to *component* in sonarqube.

So, it should be :
 
`GET /api/measures/component?component=org.app%3AmyApp&metricKeys=alert_status`

:link: **Link to the badge**

:bulb: **Possible Solution**

In services/sonar/sonar-base.js, rename componentKey to component
`qs = {                                                                           
        componentKey: component,
        metricKeys: metricName,                                                                     
      }`

to

`qs = {                                                                           
        component: component,
        metricKeys: metricName,                                                                     
      }`

rafalmaciejewski:
Closes #5882


chris48s:
Thanks for submitting this patch. Do you know of an example of a public sonar instance running >=6.6 that we could use as a test case for this?

calebcartwright:
I don't think we have any existing tests for other instances (all mocked), but I can deploy an instance of this branch and test against such a SonarQube instance on the intranet.

I'm wondering if we may be getting to a point where it'd be worth investigating some kind of sandbox env (preferably container based) where we can run a few instances of some of the self-hosted tools (or spin them up dynamically as needed for tests)

chkpnt:
Just in case you're looking for an instance for an one-time manually test: You can use a project, e.g. [`de.chkpnt:truststorebuilder-gradle-plugin`](https://sonar.chkpnt.de/dashboard?id=de.chkpnt%3Atruststorebuilder-gradle-plugin), in my personal instance at https://sonar.chkpnt.de, which is running SonarQube 8.9.1 and publicly readable.

calebcartwright:
> Just in case you're looking for an instance for an one-time manually test

Beautiful, thank you @chkpnt! 

https://shields-staging-pr-6636.herokuapp.com/sonar/quality_gate/de.chkpnt:truststorebuilder-gradle-plugin?server=https%3A%2F%2Fsonar.chkpnt.de&sonarVersion=6.6


I'm going to assume the choice of "one-time manually" phrasing was very deliberate, but going to ask you a follow up question anyway :smile: 

We run our full suite of integration tests twice daily against (mostly) live targets of upstream services/platforms/etc. Is your SonarQube instance typically internet facing and anonymously accessible (or did you perhaps temporarily make it so for this thread?), and if so, would you be opposed to us using your SonarQube instance for one of the Sonar tests?

That would mean a grand total of two http requests sent to your instance from us per day. The ability to target live instances with our test runs helps us catch issues like this more quickly when vendors change their API contracts, as Shields integrates with far too many services and platforms for our team to be up to date with breaking changes upstream.

services/sonar/sonar-base.js
```diff
@@ -70,8 +70,11 @@ module.exports = class SonarBase extends BaseJsonService {
     } else {
       schema = modernSchema
       url = `${server}/api/measures/component`
+      // componentKey query param was renamed in version 6.6
+      const componentKey =
+        parseFloat(sonarVersion) >= 6.6 ? 'component' : 'componentKey'
       qs = {
-        componentKey: component,
+        [componentKey]: component,
```
calebcartwright, line 77: Long term it'd probably be better to refactor this out and repurpose `isLegacyVersion` to have a single helper that provides the url, schema, and query string. However, this will work for now so happy to proceed to address the immediate blocking issue

calebcartwright (approved):
Thanks!

chkpnt:
@calebcartwright I just phrased it this way to not be blamed for breaking tests ☺️

As you see in the size of my SonarQube and the size of the scanned projects, it's just a hobby instance for my own pet projects, so there is no HA (although it's normally running 24/7). But of course, if you're just looking for a real-world public readable instance, you are free to use it, as there is no difference in using it for the badges in my own READMEs.

Just wanted to throw in another link: I knew the Apache Foundation were hosting it's own SonarQube instance... but it has certainly turned out, that they [sunsetted their service](https://cwiki.apache.org/confluence/display/INFRA/SonarQube+Analysis).

Issue #6365, opened 2021-04-04
Deploy on Node 14


chris48s:
:clipboard: **Description**

While Node 12 is supported for another ~year we should get ourselves deployed on the current LTS before it reaches end-of-life.

| Release | Release Date | End of Life |
|---------|--------------|-------------|
| v12     | 2019-04-23   | 2022-04-30  |
| v14     | 2020-04-21   | 2023-04-30  |
| v16     | 2021-04-20   | 2024-04-30  |

PyvesB:
I read that diagnostic reports are now a stable feature in Node 14. Is this something that would be useful to enable in production?

The new null coalescing operator is amusing.

PR #6651, opened 2021-06-20
Migrate from CommonJS to ESM

### LINKED ISSUE #6528: Change server module to ESM from CommonJS
Several core dependencies within our tree have cut over from CommonJS to ESM which is blocking updates (#6419).

This was discussed at the last ops meeting and consensus seems to be that we'll need to completely move the server over to ESM sooner rather than later

PyvesB:
Fixes #6528.

Our _badge-maker_ package was left untouched, it is still a standard CommonJS module, which means there will be no impact on downstream consumers. The rest of the Shields.io codebase imports it as any other NPM package. The GitHub approve-bot action was left out as well, as it has its own separate _package.json_, it can be migrated in due course.

The automated tools I tried weren't very good, the best I found was Putout convert-commonjs-to-esm. As you'll see, a lot of manual fixing was required afterwards. Some parts of the code needed more changes than others, for example _loader.js_ due to that fact that dynamically importing modules is asynchronous and the structure of imported modules is different. I've tried to break things down in commits describing my thought process, though I doubt reviewing each one individually will be very helpful.

Most test suites seem to be passing locally and the backend loads nicely and is able to service badges.

The frontend is where I'm stuck at the moment. I've got some local changes that should make it compliant once the general setup is working, but I'm struggling to get very far on that bit. The error I'm getting locally is along the following lines:

> [frontend] TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension "" for C:\...\workspace\shields\.cache\tmp-5248-0orlJ0tI72FY

If my understanding is correct the ESM loader is trying to load code generated by Gatsby, but Gatsby generated these files without any extension, which is why they're being rejected. Maybe I'm on the wrong track though, I admittedly really don't have much of an understanding of how our frontend fits together. A lot of folks on the Internet seem to be running into trouble when trying to get Typescript projects running with ESM, it all feels very experimental to me. One solution might be to separate the frontend in a separate CommonJS module with its own _package.json_, but that feels clunky. Help welcome!

As a side note, I'd like to highlight https://github.com/nodejs/modules/issues/471. This issue is annoying for our service classes and service tests, as they are all loaded dynamically. Basically, any syntax error in any of them will be caught by ESM loader, but you'll have no clue where the error is. ESLint was not clever enough to catch all the syntax errors either, the only reliable way I found to help me out was to get the ESM loader to tell me there was some error and then run Prettier in check mode... Whilst this issue is painful when migrating hundreds of files in one go, I suspect it won't be too much of a bother when we're modifying a few files in a normal PR.


PyvesB:
We'll probably need to land #6365 first: `ci/circleci: main` and `ci/circleci: services` are failing on Node 12 due to syntax errors, whereas their Node 14 counterparts seem fine.

The LGTM alert should go away once merged.

chris48s:
Wowzer

Screenshot at 2021-06-20 20-42-02

First off, good effort on all this!


> We'll probably need to land #6365 first

OK. That is as good a reason as any. The only thing we need to do to make this happen is bump https://github.com/badges/shields/blob/8f588179e173d412353de34dda0f3390f38692b4/package.json#L244 and heroku will deploy it on the latest version. https://github.com/badges/shields/pull/5436/files should give a pretty good template for what else needs changing. Lets get that done and drop the node 12 tests this week. We should probably also start running the tests on node 16 but we can punt it to another issue if we want, especially if there are failures we need to fix.

> One solution might be to separate the frontend in a separate CommonJS module with its own package.json, but that feels clunky

I agree it is sub-optimal but I wouldn't be against this as a temporary solution. One upside is it would allow us to split this big job into smaller chunks.

calebcartwright:
I might be misinterpreting, but reading of [this Gatsby issue](https://github.com/gatsbyjs/gatsby/issues/23705) is that they don't support ESM yet, which would be a bummer if true

PyvesB:
> I might be misinterpreting, but reading of [this Gatsby issue](https://github.com/gatsbyjs/gatsby/issues/23705) is that they don't support ESM yet, which would be a bummer if true

Yep, I had noticed that issue. Unfortunately, from what I can tell there doesn't seem to be much enthusiasm around getting Gatsby to fully support ESM.

What is interesting is that trying to build the frontend via `npm run build` leads to the same errors as described in the Gatsby issue, but simply following what the error message suggests and renaming our three Gatsby files to _*.cjs_ seems to solve the immediate errors. I'm still a little unsure of how these _*.mjs_ and _*.cjs_ extension variants really work, the Node.js documentation is quite scarce on the matter and when trying to experiment in other places of our backend code, I wasn't getting consistent results.

The error I listed in the description of this PR is different and comes from running `npm start`. I need to try some more things, but there is a chance I'll manage to get the frontend to build but that incremental code recompilation development feature won't work.

calebcartwright:
> The error I listed in the description of this PR is different and comes from running npm start

Yup understood. I'm just thinking that if Gatsby and its plugin ecosystem just don't support ESM yet then it really strengthens the case IMO for the approach you suggested we consider about (temporarily) converting the frontend to its own commonjs package. Otherwise it seems to me we'll be trying to swim upstream against a pretty swift current, at least for the foreseeable future.

PyvesB:
Yeah, the swimming upstream hasn't been very successful... :)

Currently creating a separate _package.json_ and determining which dependencies belong where.

chris48s:
Presumably the frontend "package" will be a

```
  "version": "0.0.0",
  "private": true,
```

package and we are never going to install the frontend "package" anywhere other than in this codebase. If that is the case, do we actually need to split the dependency tree?

For example, with the badge-maker package we maintain the packages that an actual user needs in `badge-maker/package.json` but we just install all the test dependencies for it in the root because we don't need to redistribute them. If we can do the same with the frontend "package" and just manage the whole dep tree in the root, that would seem easier if poss..

PyvesB:
My reasoning was the following: I'll need to run Gatsby & friends from the child _package.json_, as I'll have to get the CommonJS loader to kick in. So I'll need to have those frontend depencies accessible from that nested directory, and I'll then somehow make scripts from the parent delegate to scripts in the child.

Admittedly, I'm not the most knowledgeable maintainer when it comes to all things _package.json_, let me know if that doesn't make sense.

chris48s:
That makes sense :+1:  It was definitely a question.

I don't have a concrete solution for this, but one issue we have with badge-maker is that it has its own package.json but we have one package-lock.json in the root and a single top-level `node_modules`. When dependabot bumps a package in `badge-maker/package.json` we have to manually update the root package-lock.json. Mercifully this happens quite rarely (I think you did some good work in getting rid of some of the dependencies in badge-maker which helps) so occasionally doing this by hand is not too much of a chore but I suspect doing that manually well become very cumbersome if we need to split the dependency tree in half :( I think we'll have to arrive at a better solution for that.

PyvesB:
> So I'll need to have those frontend dependencies accessible from that nested directory

Actually, my above analysis was wrong. If you launch a script from the parent _package.json_, you don't need to have the dependencies in a nest directory: they're part of the NPM script environment, so you can just do simple tricks like `cd frontend && gatsby build`. You still need to specify the Babel config and the `gatsby` dependency in the child _package.json_ (otherwise Gatsby will tell you it's not a Gatsby project).

Finally managed to get the frontend starting up properly locally. There are still a few problems (e.g. `[frontend] Attempted import error: 'URL' is not exported from 'url' (imported as 'URL'). warn ../core/badge-urls/make-badge-url.js`), and I need to figure why the E2E tests is not working on CircleCI and why it's not deploying to Heroku, but I'm finally seeing light at the end of the tunnel. :)

package.json
```diff
@@ -99,10 +99,10 @@
...
     "depcheck": "check-node-version --node \">= 12.0\"",
     "prebuild": "run-s --silent depcheck",
-    "features": "node scripts/export-supported-features-cli.js > supported-features.json",
-    "defs": "node scripts/export-service-definitions-cli.js > service-definitions.yml",
-    "build": "run-s defs features && gatsby build",
-    "heroku-postbuild": "run-s --silent build",
+    "features": "node scripts/export-supported-features-cli.js > ./frontend/supported-features.json",
+    "defs": "node scripts/export-service-definitions-cli.js > ./frontend/service-definitions.yml",
+    "build": "run-s defs features && cd frontend && gatsby build",
+    "heroku-postbuild": "run-s --silent build && mv frontend/public .",
```
PyvesB, line 105: Neither Gatsby nor Heroku support custom folders, it must be `public` in the directory the command is run for Gatsby and  `public` in the project root for Heroku. Manually moving to align with Heroku's requirements.

https://github.com/heroku/heroku-buildpack-nodejs/issues/385
https://github.com/gatsbyjs/gatsby/discussions/1878

PyvesB:
To the best of my knowledge, this is now functional and is ready for review! 🍾

paulmelnikow:
Hey @PyvesB, thanks for continuing to push this massive rock up the hill! Glad you've reached the top 🙌🏼 

I've read a few of discussions about extensions in imports (including [this one](https://github.com/nodejs/modules/issues/444)). I've not gone nearly as deeply down the rabbit hole as you have, though did have one question: could we avoid the explicit `.js` in all the import statements by renaming our files from `.js` to `.mjs`?

In general, I do think it is okay to merge this when it's working, and refine it later, as it's _such_ a bear of a PR, though if we were to rename all the files, think that probably ought to happen before merging.

paulmelnikow (commented):
I've not read every service file, as once you get past ~ docsrs, the diffs are no longer rendered without clicking each file, but I tried to spot-checked the non-service files. When this is up to date with main, I'd be willing to give this another read locally since it'll let me page through the diff more easily. (I could also do that now if I knew which commit in main to compare to, though I tried and couldn't easily work out which one that is.

doc/TUTORIAL.md
```diff
@@ -171,44 +168,41 @@ This example is based on the [Ruby Gems version](https://github.com/badges/shiel
...
+import Joi from 'joi'
 const schema = Joi.object({
   version: Joi.string().required(),
 }).required()
 
-// (5)
-module.exports = class GemVersion extends BaseJsonService {
-  // (6)
+// (4)
+export default class GemVersion extends BaseJsonService {
```
chris48s, line 182: Really thorough job on updating all the docs in line with the code

frontend/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "badge-frontend",
```
chris48s, line 2: This isn't a blocker for merging this PR, but..

This is a workaround we will probably be stuck with for a while, but presumably at some point down the line we will be able to migrate the frontend from CommonJS to ESM too and then we can remove this.

I think it would be really useful to have an issue with a quick summary of

- How far did you get
- What problems did you encounter
- What are the blockers (is gatsby core the only package we know of with a problem, or do we know of other packages in the frontend tree that would also need an update to unblock us - was tsnode one?)
- If/when we can migrate the frontend to ESM:
  - What hacks can we remove?
  - Roughly what does that process look like?

Even if it is not 100% clear, I guess there is an extent to which that information currently exists in your head (or at least you have a clearer picture of this than anyone else) and it will be useful info at some point in future.

More broadly, I am so relieved you managed to find a way to avoid splitting the dependency tree in two :relieved:

chris48s (commented):
I approached this with a list of things you might have forgotten or overlooked and you've thought of all of them. Beyond that I've had a good read over this and done some testing locally and on staging. I have one request for additional docs which shouldn't block the merge but would be useful to get done this side of your break if poss, but I don't have any requests for changes.
I think once Paul's comments are addressed (which don't look major) I'm on board with this :+1:

frontend/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/badges/shields.git",
+    "directory": "frontend"
+  },
+  "scripts": {
+    "test": "echo 'Run tests from parent dir'; false"
+  },
+  "devDependencies": {
+    "gatsby": "*"
```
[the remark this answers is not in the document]
PyvesB: As I briefly explained in one of my previous messages:
> You still need to specify the Babel config and the gatsby dependency in the child package.json (otherwise Gatsby will tell you it's not a Gatsby project).

Here is the error message you get is Gatsby is not specified:
> gatsby <build> can only be run for a gatsby site.
Either the current working directory does not contain a valid package.json or 'gatsby' is not specified as a dependency

core/badge-urls/make-badge-url.js
```diff
@@ -1,8 +1,6 @@
-'use strict'
-
-const { URL } = require('url')
-const queryString = require('query-string')
-const { compile } = require('path-to-regexp')
+import url from 'url'
```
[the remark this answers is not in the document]
PyvesB: I changed this in 078db820dca8b2b6e58667c35ee853a9279d73ad, we were previously getting the error I mentioned earlier on in this PR, `[frontend] Attempted import error: 'URL' is not exported from 'url' (imported as 'URL'). warn ../core/badge-urls/make-badge-url.js`.

I _think_ the reason why it doesn't work in this specific case is because `make-badge-url` is the only backend file which happens to be imported by the frontend code. I'm guessing that the Typescript loader, which already supported `import` statements, has slightly different behaviours compared to the pure ESM loader.

PyvesB:
> I've read a few of discussions about extensions in imports (including [this one](https://github.com/nodejs/modules/issues/444)). I've not gone nearly as deeply down the rabbit hole as you have, though did have one question: could we avoid the explicit `.js` in all the import statements by renaming our files from `.js` to `.mjs`?

The [Node.js documentation](https://nodejs.org/api/esm.html#esm_mandatory_file_extensions) does state the following:
> A file extension must be provided when using the import keyword to resolve relative or absolute specifiers. Directory indexes (e.g. './startup/index.js') must also be fully specified.

There is an [experimental feature](https://nodejs.org/api/esm.html#esm_customizing_esm_specifier_resolution_algorithm) that allows automatic resolution of extensions, and I believe that's one of the things being talked about in the issue you linked, but if it isn't considered stable, I would not be keen on using it at this stage. :)

frontend/package.json
PyvesB: Good point will prepare a little issue with some of the stuff in my head. :)

paulmelnikow:
> There is an [experimental feature](https://nodejs.org/api/esm.html#esm_customizing_esm_specifier_resolution_algorithm) that allows automatic resolution of extensions, and I believe that's one of the things being talked about in the issue you linked, but if it isn't considered stable, I would not be keen on using it at this stage. :)

I am completely baffled that this "explicit" extension behavior is where Node has landed by default, and can't imagine this is going to be a popular option. Personally, I wouldn't mind using `--experimental-specifier-resolution=node` for a few months (since I think this will probably shake out) but I'm also okay with leaving the explicit extensions for now.

(It's clear that `"type": "module"` can be used in place of `.mjs` extensions.)

core/badge-urls/make-badge-url.js
paulmelnikow, line 1:
```suggestion
// Avoid "Attempted import error: 'URL' is not exported from 'url'" in frontend.
import url from 'url'
```

paulmelnikow (dismissed):

PyvesB:
@paulmelnikow I pushed a last commit: we're no longer using `require-hacker` and I had forgotten to remove it from the dependencies. :)

frontend/package.json
PyvesB: Here we go: #6717

chris48s (approved):

PR #6652, opened 2021-06-21
Run on node 14 in production, drop node 10 support for badge-maker and update test matrix

### LINKED ISSUE #6365: Deploy on Node 14

PyvesB:
Fixes #6365, required by  #6651.

As Node 10 has reached end of life, I've updated the minimum requirements for the badge-maker package, though I can pull that out if people feel this should be done in a future pull request.

Let's see what happens with Node 16, I'm still on 14 locally.


PyvesB:
Changed the required status checks - all other PRs will be marked as pending until this gets merged and they get rebased.

.circleci/config.yml
```diff
@@ -96,23 +96,14 @@ package_steps: &package_steps
...
 
-    - run:
-        environment:
-          mocha_reporter: mocha-junit-reporter
-          MOCHA_FILE: junit/badge-maker/v10/results.xml
-          NODE_VERSION: v10
-          CYPRESS_INSTALL_BINARY: 0
-        name: Run package tests on Node 10
-        command: scripts/run_package_tests.sh
-
```
chris48s, line 115: I am a bit confused about this. We dropped node 10 testing back in https://github.com/badges/shields/pull/6570/files

I'm guessing you must have taken this working branch off an outdated master but I'd expect that the PR status would be showing merge conflicts and it isn't.. Any chance you could have a look at what is going on here?

PyvesB: Hmmm, I took from when I started work on #6651, and indeed  #6570 went through in the meantime.

PyvesB: From #6570:

> We dropped compatibility for node 10 in version 3.0.0 of badge-maker but we've still been running the tests on it.

This is not the case. We dropped support for versions older than Node 10, but at the time of the 3.0.0, Node 10 was still supported. This PR therefore does take care of the actual dropping of Node 10 by updating _package.json_.

chris48s: Ah yes. You're totally right. We dropped node < 10 in 3.0.0.

Agree lets drop it here :+1:  but I think that means the next badge-maker release needs to be a major version bump since it drops a node version.

chris48s (approved):

Issue #6653, opened 2021-06-23
invalid query parameter: flag


bubkoo:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [badge-maker NPM package](https://www.npmjs.com/package/badge-maker)

:beetle: **Description**

badge with codecov's link https://codecov.io/gh/antvis/x6/branch/master/graph/badge.svg?flag=x6-vector
[![codecov](https://codecov.io/gh/antvis/X6/branch/master/graph/badge.svg?token=15CO54WYUV)](https://codecov.io/gh/antvis/X6)

badge generated by shields.io throw error: https://img.shields.io/codecov/c/github/antvis/x6?flag=x6-vector&label=codecov&logo=codecov&token=15CO54WYUV
[![codecov](https://img.shields.io/codecov/c/github/antvis/x6?flag=x6-vector&label=codecov&logo=codecov&token=15CO54WYUV)](https://codecov.io/gh/antvis/X6)

:link: **Link to the badge**

https://img.shields.io/codecov/c/github/antvis/x6?flag=x6-vector&label=codecov&logo=codecov&token=15CO54WYUV

:bulb: **Possible Solution**

PyvesB:
Hello @bubkoo ! 👋🏻 

Well yes, this is expected, Shields.io defines its own set of query parameters, `flag` is not a valid one for our badges. Only the ones indicated in the Shields.io frontend for the Codecov badge will be available. If you'd like new features to be added to the Codecov badge, feel free to submit a pull request or create a separate feature request issue. 😉

bubkoo:
But the url was generated by Shields.io. 

Flag with name "x6-vector" throw errors.
![image](https://user-images.githubusercontent.com/6045824/123046037-3522f680-d42e-11eb-9097-1fd130e593ad.png)

Flag with name "x6_vector" will render the badge.
![image](https://user-images.githubusercontent.com/6045824/123046073-3e13c800-d42e-11eb-89e5-758ce83a7bcf.png)

PyvesB:
Ah okay, sorry for the confusion! I hadn't clocked that we did also support a `flag` parameter for Codecov.

Our code enforces the following constraints for this parameter:
https://github.com/badges/shields/blob/68fb0b7bab93f6cbce9d6176f29b556ac8dc4899/services/codecov/codecov.service.js#L26

[...]

PR #6655, opened 2021-06-23
fix: fix regex to match [codecov]'s flags

bubkoo:
Flags Must consist only of alphanumeric characters, '_', '-', or '.' and not exceed 45 characters. 

Flag names will be validated against the following regex. ^[\w\.\-]{1,45}$

re #6653


calebcartwright (approved):
Thanks so much for both reporting the issue and submitting this to fix it! :heart: 

The changes look good and indeed resolve the issue:
https://shields-staging-pr-6655.herokuapp.com/codecov/c/github/antvis/x6?flag=x6-vector

Issue #4139, opened 2019-10-04
Broken link to tutorial on https://contributing.shields.io/index.html


seetd:
Are you experiencing an issue with...

- [ *] [shields.io] https://contributing.shields.io/index.html

:beetle: **Description**
The link for `tutorial on how to add a badge.` appears to be broken on the site deployed to `contributing.shields.io`. The link is pointing to `https://contributing.shields.io/doc/TUTORIAL.md` which does not exists. The source on `https://github.com/badges/shields/blob/master/README.md` works though and it links to `https://github.com/badges/shields/blob/master/doc/TUTORIAL.md`. It looks like the same content is deployed at `https://contributing.shields.io/tutorial-tutorial`

:bulb: **Possible Solution**

chris48s:
Cheers for raising this. Looking at it, there are actually several places where we've got relative links that do work when you view the docs as a pile of markdown files on Github but then don't work on https://contributing.shields.io/

It would be nice to maintain the property that you can read the docs in either environemnt and the links will all work but I don't have an immediate solution to that.

chris48s:
I've had a bit of an investigate to see if we can achieve this:

JSDoc doesn't give us any way to configure how the tutorials are linked or intercept that process: https://jsdoc.app/about-tutorials.html

I had a look to see if we could use something like [markdown-folder-to-html](https://www.npmjs.com/package/markdown-folder-to-html) to pre-process the markdown into HTML before we pass it to JSDoc to use as a tutorial (we can use a custom template to get round doubling up the menus). markdown-folder-to-html does intercept all the markdown links and change them to HTML links, but then JSDoc changes all the HTML filenames which then breaks the links. Again, this is non-configurable.

We can chuck the HTML files into the JSDoc output dir as [static files](https://jsdoc.app/about-configuring-default-template.html#copying-static-files-to-the-output-directory) but this feature is only really useful for JS, CSS, images etc because if you use it for HTML, JSDoc doesn't add the files to the menus and that is also non-configurable.

The way to get JSDoc to link all the tutorials up correctly is to link them using `{@link}` and [{@tutorial}](https://jsdoc.app/tags-inline-tutorial.html) tags (example: https://github.com/SoftwareBrothers/admin-bro-dev/blob/master/docs-src/tutorials/05-actions.md ). Doing this gets JSDoc to link all your tutorials up nicely once you compile them, but then you lose the internal linking when you view them on GitHub as markdown.

[...]

PR #6676, opened 2021-06-27
docs: broken links and typos

SethFalco:
This is a minor contribution while I was working on making some badges.

Fixes a minor typo, it's "GitHub" not "Github".

In the [API documentation](https://contributing.shields.io/) the README is displayed.
However, relative links are kept relative, so they break as they should be linking to [github.com](https://github.com/), not [contributing.shields.io](https://contributing.shields.io/).

This is resolved by using absolute URLs instead of relative URLs.


calebcartwright:
Thanks @SethFalco! The changes seem reasonable at first blush, though I do recall we had to make some tradeoffs with our documentation links because it wasn't possible to have them work everywhere. I'm unable to find the specifics because my GitHub search skills are failing me at the moment, but @chris48s do you recall the details/know if these proposed changes could have any potential unintended side effects?

chris48s:
https://github.com/badges/shields/issues/4139 still stands but there is nothing here that makes anything worse.
It is on my "if I suddenly have a pile of spare time.." list to have a look at other ways to generate `contributing.shields.io` but that is.. yet to transpire

calebcartwright (approved):
LGTM, thanks!

PR #6677, opened 2021-06-27
Add [Weblate] badges

SethFalco:
This adds [Weblate](https://weblate.org/en/) badges to the repository.
Weblate is a free and open-source crowd translation platform.

Currently, it adds 8 badges:
| Name | Description |
|---|---|
| Weblate project translated | The number of strings of a project that has been translated. |
| Weblate component license | The license that a particular project component is under. |
| Weblate components/projects/users | The total that exist on a Weblate instance. |
| Weblate user translations/suggestions/languages | The number of user has made/done. |

Examples:
image

I might make more badges later, but for now, I wanted to get these right.

### Related
* Depends on https://github.com/badges/shields/pull/6684

### DIFF
--- a/services/weblate/weblate-entity-count.service.js
+++ b/services/weblate/weblate-entity-count.service.js
@@ -1,0 +1,76 @@
+'use strict'
+
+const Joi = require('joi')
+const camelcase = require('camelcase')
+const { BaseJsonService } = require('..')
+const { nonNegativeInteger, optionalUrl } = require('../validators')
+
+const schema = Joi.object({
+  count: nonNegativeInteger,
+}).required()
+
+const queryParamSchema = Joi.object({
+  server: optionalUrl.required(),
+}).required()
+
+class WeblateEntityCountBase extends BaseJsonService {
+  static category = 'other'
+
+  static buildRoute(entityName) {
+    return {
+      base: 'weblate',
+      pattern: entityName,
+      queryParamSchema,
+    }
+  }
+
+  static defaultBadgeData = { label: 'weblate', color: 'brightgreen' }
+
+  async fetch({ entityName, server }) {
+    return this._requestJson({
+      schema,
+      url: `${server}/api/${entityName}/`,
+      errorMessages: {
+        403: 'access denied',
+        429: 'rate limited by remote server',
+      },
+    })
+  }
+}
+
+function WeblateEntityCountFactory({ entityName, exampleValue }) {
+  return class WeblateEntityCountService extends WeblateEntityCountBase {
+    static name = camelcase(`Weblate ${entityName}`, { pascalCase: true })
+    static route = this.buildRoute(entityName)
+
+    static examples = [
+      {
+        title: `Weblate ${entityName}`,
+        namedParams: {},
+        queryParams: { server: 'https://hosted.weblate.org' },
+        staticPreview: this.render({ count: exampleValue }),
+        documentation: `<p>This badge displays the number of ${entityName} on a Weblate instance.</p>`,
+        keywords: ['i18n', 'internationalization'],
+      },
+    ]
+
+    static render({ count }) {
+      return { message: `${count} ${entityName}` }
+    }
+
+    async handle(routeParams, { server }) {
+      const { count } = await this.fetch({ entityName, server })
+      return this.constructor.render({ count })
+    }
+  }
+}
+
+const entityCounts = [
+  { entityName: 'components', exampleValue: 2799 },
+  { entityName: 'languages', exampleValue: 384 },
+  { entityName: 'projects', exampleValue: 533 },
+  { entityName: 'units', exampleValue: 16539728 },
+  { entityName: 'users', exampleValue: 33058 },
+].map(WeblateEntityCountFactory)
+
+module.exports = [...entityCounts]
--- a/services/weblate/weblate-entity-count.service.js
+++ b/services/weblate/weblate-entity-count.service.js
@@ -70,7 +70,6 @@ const entityCounts = [
   { entityName: 'components', exampleValue: 2799 },
   { entityName: 'languages', exampleValue: 384 },
   { entityName: 'projects', exampleValue: 533 },
-  { entityName: 'units', exampleValue: 16539728 },
   { entityName: 'users', exampleValue: 33058 },
 ].map(WeblateEntityCountFactory)
 
--- a/services/weblate/weblate-entity-count.service.js
+++ b/services/weblate/weblate-entity-count.service.js
@@ -25,14 +25,14 @@ class WeblateEntityCountBase extends BaseJsonService {
     }
   }
 
-  static defaultBadgeData = { label: 'weblate', color: 'brightgreen' }
+  static defaultBadgeData = { color: 'informational' }
 
   async fetch({ entityName, server }) {
     return this._requestJson({
       schema,
       url: `${server}/api/${entityName}/`,
       errorMessages: {
-        403: 'access denied',
+        403: 'access denied by remote server',
         429: 'rate limited by remote server',
       },
     })
@@ -50,13 +50,12 @@ function WeblateEntityCountFactory({ entityName, exampleValue }) {
         namedParams: {},
         queryParams: { server: 'https://hosted.weblate.org' },
         staticPreview: this.render({ count: exampleValue }),
-        documentation: `<p>This badge displays the number of ${entityName} on a Weblate instance.</p>`,
         keywords: ['i18n', 'internationalization'],
       },
     ]
 
     static render({ count }) {
-      return { message: `${metric(count)} ${entityName}` }
+      return { label: entityName, message: metric(count) }
     }
 
     async handle(routeParams, { server }) {
@@ -68,7 +67,6 @@ function WeblateEntityCountFactory({ entityName, exampleValue }) {
 
 const entityCounts = [
   { entityName: 'components', exampleValue: 2799 },
-  { entityName: 'languages', exampleValue: 384 },
   { entityName: 'projects', exampleValue: 533 },
   { entityName: 'users', exampleValue: 33058 },
 ].map(WeblateEntityCountFactory)
--- a/services/weblate/weblate-entity-count.service.js
+++ b/services/weblate/weblate-entity-count.service.js
@@ -4,6 +4,7 @@ const Joi = require('joi')
 const camelcase = require('camelcase')
 const { BaseJsonService } = require('..')
 const { nonNegativeInteger, optionalUrl } = require('../validators')
+const { metric } = require('../text-formatters')
 
 const schema = Joi.object({
   count: nonNegativeInteger,
@@ -55,7 +56,7 @@ function WeblateEntityCountFactory({ entityName, exampleValue }) {
     ]
 
     static render({ count }) {
-      return { message: `${count} ${entityName}` }
+      return { message: `${metric(count)} ${entityName}` }
     }
 
     async handle(routeParams, { server }) {
--- a/services/weblate/weblate-project-translated-percentage.service.js
+++ b/services/weblate/weblate-project-translated-percentage.service.js
@@ -1,0 +1,76 @@
+'use strict'
+
+const Joi = require('joi')
+const { BaseJsonService } = require('..')
+const { optionalUrl } = require('../validators')
+
+const schema = Joi.object({
+  translated_percent: Joi.number().required(),
+}).required()
+
+const queryParamSchema = Joi.object({
+  server: optionalUrl.required(),
+}).required()
+
+const documentation = `
+  <p>
+    This badge displays the percentage of strings translated on a project on a
+    Weblate instance.
+  </p>
+`
+
+module.exports = class WeblateProjectTranslatedPercentage extends (
+  BaseJsonService
+) {
+  static category = 'other'
+  static route = { base: 'weblate', pattern: ':project', queryParamSchema }
+
+  static examples = [
+    {
+      title: 'Weblate project translated',
+      namedParams: { project: 'godot-engine' },
+      queryParams: { server: 'https://hosted.weblate.org' },
+      staticPreview: this.render({ translatedPercent: 20.5 }),
+      documentation,
+      keywords: ['i18n', 'translation', 'internationalization'],
+    },
+  ]
+
+  static defaultBadgeData = { label: 'translated' }
+
+  /**
+   * Takes a percentage and maps it to a message and color.
+   *
+   * The colors are determined based on how Weblate does it internally.
+   * {@link https://github.com/WeblateOrg/weblate/blob/main/weblate/trans/widgets.py Weblate on GitHub}
+   *
+   * @param {*} translatedPercent The percentage of translations translated.
+   * @returns {object} Format for the badge.
+   */
+  static render({ translatedPercent }) {
+    if (translatedPercent >= 90)
+      return { message: `${translatedPercent}%`, color: 'success' }
+
+    if (translatedPercent >= 75)
+      return { message: `${translatedPercent}%`, color: 'yellow' }
+
+    return { message: `${translatedPercent}%`, color: 'critical' }
+  }
+
+  async fetch({ project, server }) {
+    return this._requestJson({
+      schema,
+      url: `${server}/api/projects/${project}/statistics/`,
+      errorMessages: {
+        403: 'access denied',
+        404: 'not found',
+        429: 'rate limited by remote server',
+      },
+    })
+  }
+
+  async handle({ project }, { server }) {
+    const { translated_percent } = await this.fetch({ project, server })
+    return this.constructor.render({ translatedPercent: translated_percent })
+  }
+}
--- a/services/weblate/weblate-project-translated-percentage.service.js
+++ b/services/weblate/weblate-project-translated-percentage.service.js
@@ -3,6 +3,7 @@
 const Joi = require('joi')
 const { BaseJsonService } = require('..')
 const { optionalUrl } = require('../validators')
+const { colorScale } = require('../color-formatters')
 
 const schema = Joi.object({
   translated_percent: Joi.number().required(),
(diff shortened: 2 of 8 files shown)


calebcartwright:
Thank you for the PR! Haven't gone through the code in detail yet but a few comments/requests:

* Could you please provide some context around the proposed badges (i.e. what is this badge for, what does it mean, etc.). For example, I'm not really sure what to make of the `components` or `units` badges amongst a few others. We want to make sure that the badges we provide make sense/serve a use case, but we don't need to provide a badge for data points just because those data points exist
* It looks like Webplate **heavily** throttles unauthenticated requests (100/day) so for this to be a viable candidate we'd have to incorporate authentication on our requests (would also be helpful to get an idea of what if any rate limiting they apply on authenticated requests)
* We've got a variety of helper functions for badge formatting as well, and based on the badges shown in the screenshot, we'll definitely at least need to leverage the `metric` formatter on any count/metric type messages
* In cases where we've got a large quantity of badges we just about always end up encapsulating some of the common bits into either reusable functions defined in some common/helper file and/or via a base class, so probably worth seeing if there would be opportunities do to something similar here

SethFalco:
> Could you please provide some context around the proposed badges

* `Weblate project translated` is the main one I wanted, displays how much of a project was translated.
* `Weblate user translations` is how many strings a particular user has translated, I thought it could look nice on a GitHub README for users that translate a lot.
* `Weblate projects/components/users` are how many things are on a particular Weblate instance. I'm not sure if these will be useful, but figured they're notable information for public Weblate hosts.

Admittedly, the rest were just convenient data-points to grab while I was doing the above. If it's preferred, I can remove them?

> I'm not really sure what to make of the components or units badges amongst

I'll define them just so we're on the same page, but I appreciate your point.
* Components - A "Project" might have multiple i18n files, or might split it between a "frontend" and "backend" component, for example.
* Unit - A individual translatable string.

> so for this to be a viable candidate we'd have to incorporate authentication on our requests

That would be possible with the public instance. I think overall it might be hard though since there are many hosts for Weblate. Even I host my own Weblate instance for my projects.

A non-exhaustive list can be found on the [Discover](https://weblate.org/en/discover/) page on Weblate's official website.

> need to leverage the metric formatter on any count/metric type messages

Ahh, thanks for that! Will push this in a moment and update the screenshot.

> large quantity of badges we just about always end up encapsulating some of the common bits

I've already done this a bit, but I can encapsulate more of it, so it's more reusable. I'll do this tonight/tomorrow.

calebcartwright:
> I assume it's fine if I make a new isMetric... test-validator?
>
> We have one called isMetricOpenIssues, I'd like to make one inspired by this which uses custom text instead?
At least I don't see something which does this already, I think something like this might be nice?

Sure! It's a bit of a balancing act of course as we don't want to bloat the helper files with single-use functions for bespoke cases, but we also don't want to reinvent the wheel in the individual service classes/tester files either.

I suspect we've probably got a few cases of such duplication, and if you see opportunities for encapsulating such a helper (regardless of whether that's with these proposed badges or things already merged) definitely feel free to suggest improvements/submit a PR.

I do think that'd probably be best handled via a separate PR which we could land first and this PR could then grab and utilize

calebcartwright:
> That would be possible with the public instance. I think overall it might be hard though since there are many hosts for Weblate. Even I host my own Weblate instance for my projects.

Let me know if I'm off base but so far this sounds similar to some other badges we support, where there's a defacto public instance typically a SaaS-style offering but also comes in a downloadable/self-hostable form too. In these cases we add support to the service class for authenticated requests, configure our production environment with the auth info from an account we own against the main public instance, and then users that need to access an instance other than the main one and/or with their own auth will run their own [self-hosted instance of the Shields server](https://github.com/badges/shields/blob/master/doc/self-hosting.md)

SethFalco:
* I've updated this to use the new `isMetricWithPattern` method test-helper.
* Removed the badge for `units` because the endpoint was incredibly slow and had timeouts often even with 10+ seconds.

> r this sounds similar to some other badges we support
Sounds valid for this one as well. I'll check how other services are configured and update this one to follow suit.

calebcartwright:
>  because the endpoint was incredibly slow and had timeouts often even with 10+ seconds.

Yikes! That's a good call. We have a ceiling of ~3-4 seconds _total_ for the entire round trip to get the badges response back to users, otherwise they timeout and don't render in GitHub

services/weblate/weblate-entity-count.service.js
```diff
@@ -0,0 +1,76 @@
...
+
+  static buildRoute(entityName) {
+    return {
+      base: 'weblate',
+      pattern: entityName,
+      queryParamSchema,
+    }
+  }
+
+  static defaultBadgeData = { label: 'weblate', color: 'brightgreen' }
```
calebcartwright, line 28: Our badge defaults need to adhere to our [specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md) and also be rendered similar to how we handle other badges.

Here, I think that means we should use the entity name for the label, and we should also use the `informational` (blue) color as brightgreen is used to connote various forms of "success" or "passing"

services/weblate/weblate-entity-count.service.js
```diff
@@ -0,0 +1,76 @@
...
+        namedParams: {},
+        queryParams: { server: 'https://hosted.weblate.org' },
+        staticPreview: this.render({ count: exampleValue }),
+        documentation: `<p>This badge displays the number of ${entityName} on a Weblate instance.</p>`,
+        keywords: ['i18n', 'internationalization'],
+      },
+    ]
+
+    static render({ count }) {
+      return { message: `${metric(count)} ${entityName}` }
```
calebcartwright, line 59: Similar comment as the default data. The entity name should be the label with the metric-ified count being the message. Users would always still have the ability to customize the label if they want it to be something more like `weblate components` instead of just `components`

services/weblate/weblate-project-translated-percentage.service.js
```diff
@@ -0,0 +1,72 @@
...
+
+const queryParamSchema = Joi.object({
+  server: optionalUrl.required(),
+}).required()
+
+const documentation = `
+  <p>
+    This badge displays the percentage of strings translated on a project on a
+    Weblate instance.
+  </p>
```
calebcartwright, line 20: Same theme here as the feedback provided in the other class. It's helpful for us as maintainers during initial discussion and implementation to understand what exactly the badges represent, but I feel like we shouldn't need basic descriptions like this in the help text which is what is shown to the end users trying to create a specific instance of a badge (since those are the folks that will be users of the service/know what the badge is).

If there's information like this that you think would be helpful for observers of the Shields _source code_ then I think that'd be best served as an inline comment in the file

services/weblate/weblate-project-translated-percentage.service.js
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Takes a percentage and maps it to a message and color.
+   *
+   * The colors are determined based on how Weblate does it internally.
+   * {@link https://github.com/WeblateOrg/weblate/blob/main/weblate/trans/widgets.py Weblate on GitHub}
+   *
+   * @param {*} translatedPercent The percentage of translations translated.
+   * @returns {object} Format for the badge.
+   */
```
calebcartwright, line 50: hmm, i'm a little torn on this. We don't typically use the full-blown jsdoc on functions except where needed (and do have linter checks for it). I'm not necessarily opposed to this, but also a bit wary of starting a pattern of applying jsdoc to the standard functions that exist in every single one of our service classes. For now I'd prefer just adding an line comment to the `colorScale` call within the function (basically copy/combine lines 45/46 and move that to apply to line 52), but cc @chris48s in case you feel differently since you did all the original jsdoc work

services/weblate/weblate-project-translated-percentage.service.js
```diff
@@ -0,0 +1,72 @@
...
+  static render({ translatedPercent }) {
+    const color = colorScale([75, 90])(translatedPercent)
+    return { message: `${translatedPercent}%`, color }
```
calebcartwright, line 53: We do round all our percentages to the nearest whole number via `.toFixed(0)` and should do this here as well for consistency

services/weblate/weblate-project-translated-percentage.service.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  async fetch({ project, server }) {
+    return this._requestJson({
+      schema,
+      url: `${server}/api/projects/${project}/statistics/`,
+      errorMessages: {
+        403: 'access denied',
+        404: 'not found',
+        429: 'rate limited by remote server',
```
calebcartwright, line 63: Same type of comment as earlier classes relative to the need/text of the error messages

services/weblate/weblate-project-translated-percentage.tester.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const { ServiceTester } = require('../tester')
+const { isPercentage } = require('../test-validators')
+
+const t = (module.exports = new ServiceTester({
+  id: 'weblate',
+  title: 'Weblate',
+}))
```
calebcartwright, line 9: Same comment from the component license tester about using the helper here

services/weblate/weblate-user-statistics.service.js
```diff
@@ -0,0 +1,92 @@
+'use strict'
+
```
calebcartwright, line 2: Won't rehash everything but several of the same feedback items for this file as well

services/weblate/weblate-user-statistics.tester.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+const { ServiceTester } = require('../tester')
+const { isMetricWithPattern } = require('../test-validators')
+
+const t = (module.exports = new ServiceTester({
+  id: 'weblate',
+  title: 'Weblate',
+}))
```
calebcartwright, line 9: Same feedback as the tester for the entity count. Can't use the helper due to the multiclass exports but should be able to give this some better values

services/weblate/weblate-user-statistics.tester.js
```diff
@@ -0,0 +1,39 @@
...
+  .get('/user/nijel/comments.json?server=https://hosted.weblate.org')
+  .expectBadge({ label: 'weblate', message: isMetricWithPattern(/ comments/) })
+
+t.create('Languages')
+  .get('/user/nijel/languages.json?server=https://hosted.weblate.org')
+  .expectBadge({ label: 'weblate', message: isMetricWithPattern(/ languages/) })
+
+t.create('Invalid Protocol')
+  .get('/user/nijel/translations.json?server=ftp://hosted.weblate.org')
+  .expectBadge({ label: 'weblate', message: 'invalid query parameter: server' })
```
calebcartwright, line 39: Let's remove this, as it's only validating the schema which is already tested elsewhere and we really strive to avoid having duplicative tests within the service test suite given the objectives for these types of tests

calebcartwright (commented):
Thanks again @SethFalco! Have taken a first pass through this and left some specific feedback inline below.

I'm not sure where we landed as far as which badges make sense to keep (i.e. which badges users will want/consume), so if some of the files/badges have gone out of scope feel free to remove those files and disregard feedback in those files

SethFalco:
Thank you very much for giving such thorough feedback! I'll be sure to keep this in mind for future pull requests as well.

I believe I've adhered to all feedback. I've resolved the ones that I'm pretty confident I've done exactly as asked.
The ones left unresolved I'm hoping you can review for me, and resolve them if you think they're done, or drop a comment if I've missed something.

> I'm not sure where we landed as far as which badges make sense to keep

The only ones I've pretty adamant on keeping are the following, as I have hopes to use these 2 myself:
* Weblate project translated percentage
* Weblate user translations

I've looked over the badges and removed 3 which, I believe, would have little to no use-case:
* Weblate languages
* Weblate user comments
* Weblate user 

I've left the rest in, as I'd imagine there could be a use-case for them, however you're welcome to suggest otherwise. I have no strong feelings on the matter.

services/weblate/weblate-project-translated-percentage.service.js
chris48s: We have acquired a couple of random services where we have got JSDoc comments like this

https://contributing.shields.io/module-services_dynamic_json-path.html
https://contributing.shields.io/module-services_steam_steam-base-BaseSteamAPI.html

but I think you're probably right that the generated docs are only really useful as a reference for functions/classes which we are going to use in more than one place. I'm not going to use `WeblateProjectTranslatedPercentage.render()` anywhere else so its probably just noise to include it in the docs site (esp if we had hundreds of these). I don't want to discourage contributors from including useful information in comments for service classes though. Maybe we could add en exclude pattern to [jsdoc.json](https://github.com/badges/shields/blob/master/jsdoc.json) to ignore `services/**/*.service.js` files when we build?
Either way, I don't think it has to be a blocker for this PR.

calebcartwright:
> The only ones I'm pretty adamant on keeping are the following, as I have hopes to use these 2 myself:
> I've looked over the badges and removed 3 which, I believe, would have little to no use-case:
> I've left the rest in, as I'd imagine there could be a use-case for them, however you're welcome to suggest otherwise. I have no strong feelings on the matter.

No worries! I genuinely don't know enough about the service/what's meaningful for users so I'm not prescribing anything one way or another, just wasn't sure if we'd done any passes through

> Edit: Sorry, I didn't think to edit the PR title until after I made all the changes. ^-^' Might want to rerun CI.

Have updated and restarted the ci jobs accordingly. The test matcher is smart enough (most of the time) to be able to just use the service name when you want to run all the tests like we do here, but we still give the individual testers more specific names so that we also have the option of running a subset of the tests, i.e. if in the future we make a change to the user translation badge then we can run just those tests vs. having to run the entire weblate suite.

services/weblate/weblate-project-translated-percentage.service.js
calebcartwright: > Either way, I don't think it has to be a blocker for this PR.

Agreed. My concern was really about the hypothetical scenario of this being repeated at scale (every render, handle, etc. function in every service class) and the impact that'd have on code scannability and the docs site.

> Maybe we could add en exclude pattern to jsdoc.json to ignore services/**/*.service.js files when we build?

I like this idea :+1: (and feel free to tackle that one @SethFalco if you're looking for anything else after this!)

calebcartwright (approved):
Excellent work, thank you!

SethFalco:
@calebcartwright Sorry to pester, just a few questions.

I just realized I forgot to add authentication on the PR.
I'll look into that and PR it separately.

Could I just check, should I only add support for an API key with the official instance, or could we try and support other instances?

For example, using the key `WEBLATE_API_KEY` for the official instance.
But supporting API keys for other Weblate instances on a per-instance basis with `WEBLATE_API_KEY_I18N_EXAMPLE_ORG`?

Then we could just take the domain (https://**i18n.example.org**) to get the respective API key if it's defined:
* `_` -> `__`
* `/[!a-z]/g` -> `_`
* `toUpperCase()`

---

I also learned in the PR to Wikiapiary that it's possible to define enums in the URL pattern. 🤔 
Would I be welcome to update the Weblate badge to change all user-statistic badges into a single badge that just uses:
`weblate/user/:user/:statistic(translated|suggestions|languages|uploaded|commented)`

I can also abstract the entity count one similarly:
`weblate/:type(components|projects|users|languages)`

This will reduce the total badge count (only 4 badges in total), and clean up the code/tests. Could be nice to leave in some of those extra options as well, since they'll just be part of the same badge?

calebcartwright:
> Sorry to pester, just a few questions.

No worries at all, don't hesitate to ask any questions!

> Would I be welcome to update the Weblate badge to change all user-statistic badges into a single badge that just uses:

Yes indeed. To be honest we need to adjust a few of these routes anyway to align with our [badge route guidelines](https://github.com/badges/shields/blob/master/doc/badge-urls.md). In this case the badges are so new (and unused) that we don't need to bother with a redirector, so let's just correct the routes to what we want them to be and try to get it completed quickly. By any chance do you think you'll have bandwidth to get this done soon?

SethFalco:
> By any chance do you think you'll have bandwidth to get this done soon?

No problem! Sorry about that.

Tomorrow I'll make a PR with the following changes:
* Reduce the total badges and use the `(...|...)` syntax. (as mentioned above)

* Replace `/license` with `/l`.
  > License: `/l`

* Default to `https://hosted.weblate.org` (it seems I can make server optional based on the docs)
  > The parameter is for an uncommon optional attribute, like an alternate registry URL.

* Review badge paths
  > /SERVICE/NOUN/PARAMETERS?QUERYSTRING e.g: /github/issues/:user/:repo

After that PR is merged, I'll look at authentication again.
Feel free to let me know if there's anything you think I'm missing in that list.

calebcartwright:
> I just realized I forgot to add authentication on the PR.
I'll look into that and PR it separately.

That one slipped my mind as well. That's a lower priority compared to adjusting the routes, but we should try to get this plugged in fairly quickly too given the low rate limit.

> Could I just check, should I only add support for an API key with the official instance, or could we try and support other instances?

I'm not sure I follow what you're asking so will provide a broad answer to see if that covers, but please let me know if you still have any outstanding questions.

There are two core patterns we support for incorporating auth info in the requests we make to the upstream services

* Private service-specific secrets with values set in the server runtime environment (either as an environment variable or in the config file)
* User-provided values that are included in the badge request that we receive via a route or query parameter

The second one is something we tend to prefer to avoid as it's only viable in cases where the upstream service makes it possible to generate a strictly read-only token. Our codecov badge is one example of this where the badge route accepts a `token` query parameter.

The other cases follow that first pattern, where support for a secret is added to the server (currently supported secrets enumerated [here](https://github.com/badges/shields/blob/master/doc/server-secrets.md)) which allows any deployment of the badge server to define a value for that secret. 

For the Shields.io service (which is essentially just the main deployment of the Shields server), we deploy with values provided for those various secrets using accounts that we own and manage on those upstream platforms (more details here: https://github.com/badges/shields/blob/master/doc/production-hosting.md).

Our secrets-setting mechanism doesn't support a variable number of secrets for multiple instances/tenants of a given service; there is a 1:1 mapping between a target service/platform and the corresponding auth secret for that service. This is just a basic reflection of practicality, as it wouldn't be realistic nor necessary for that matter for the main Shields.io instance to be able to authenticate with an open ended number of instances of a given service/tool/platform.

That pattern needs to apply here as well: there should only be one secret defined for weblate. It sounds like there's a de facto public instance, so once the server code is updated to support the new weblate secret then one of us on the Shields maintainer team will create an account and generate the corresponding values, which we'll then plug in to the Shields.io runtime.

PR #147, opened 2014-03-22
Add support for MediaWiki extension count via WikiApiary.

thingles:
Pull request to close badges/shields#146.

PR #6678, opened 2021-06-27
Add Wikiapiary Extension Badge [WikiapiaryInstalls]

### LINKED ISSUE #146: Badge request: MediaWiki extension usage counts from WikiApiary
Allow shields for extension usage count using WikiApiary data. This can pull the data directly from MediaWiki using the Semantic MediaWiki Ask API.

Format: 

`/wikiapiary-ext/ParserFunctions.svg`
`/wikiapiary-ext/Semantic%20MediaWiki.svg`

SethFalco:
* Closes #146
* Related to #147

I thought I could go through the backlog and pick something up.

It seems there are multiple types of extensions for MediaWiki, which can be read about here:
https://www.mediawiki.org/wiki/Manual:Extensions

Wikiapiary labels them as "Extensions" and "Skins".
Technically, this works for "Farms", "Generators", and "Hosts" as well, so it could be worth renaming this? 🤔


services/wikiapiary/wikiapiary-extension.service.js
```diff
@@ -0,0 +1,64 @@
...
+ * This badge displays the total usages of a MediaWiki extensions via
+ * Wikiapiary.
+ *
+ * {@link https://www.mediawiki.org/wiki/Manual:Extensions MediaWiki Extensions Manual}
+ */
+module.exports = class WikiapiaryExtension extends BaseJsonService {
+  static category = 'downloads'
+  static route = {
+    base: 'wikiapiary/extensions',
+    pattern: ':type/:extension',
```
chris48s, line 31: > Technically, this works for "Farms", "Generators", and "Hosts" as well, so it could be worth renaming this?

Yeah lets drop the terms "extension" and "downloads". It seems like what we're actually measuring here is more like install base rather than downloads? (and in principle there could be other metrics). How about making the URLs:

`wikiapiary/:variant(extension|skin|farm|generator|host)/installs/:name`

Wordpress plugins/themes might be a good reference point for the terminology.

chris48s (commented):
There's a couple of cases that are not properly handled here

Firstly there is no handling for the case where the extension (or whatever) does not exist e.g:

- `/wikiapiary/extensions/Extension/doesnotexist`
- `/wikiapiary/extensions/notatype/doesnotexist`


Secondly, there's some errors around case-sensitivity. If I call for example:

- `/wikiapiary/extensions/extension/ParserFunctions` or
- `/wikiapiary/extensions/Extension/parserfunctions`

then an error is thrown because the param case doesn't match the response case.

services/wikiapiary/wikiapiary-extension.service.js
```diff
@@ -0,0 +1,73 @@
...
+
+  static defaultBadgeData = { label: 'installs', color: 'informational' }
+
+  static render({ usage }) {
+    return { message: metric(usage) }
+  }
+
+  static validate({ results }) {
+    if (Array.isArray(results))
+      throw new NotFound({ prettyMessage: 'does not exist' })
```
chris48s, line 54:
```suggestion
      throw new NotFound({ prettyMessage: 'not found' })
```

services/wikiapiary/wikiapiary-extension.service.js
```diff
@@ -0,0 +1,73 @@
...
+    })
+  }
+
+  async handle({ variant, name }) {
+    const response = await this.fetch({ variant, name })
+    const { results } = response.query
+
+    this.constructor.validate({ results })
+
+    const [usage] = results[`${variant}:${name}`].printouts['Has website count']
```
chris48s, line 70: We've still got a case-sensitivity issue here. e.g:

`wikiapiary/Extension/installs/parserFunctions`

It seems fundamentally the difficulty stems from the fact the query is case-insensitive (kind of) but the response isn't.


Would it work if we lowercase the args and lowercase the keys in the results array

```js
results = Object.keys(results).reduce((accum, k) => (accum[k.toLowerCase()] = results[k], accum), {})
```

to do a case-insensitive comparison?


That would also conveniently allow us to make the options for the `:variant` param lowercase. In general, we tend to avoid having uppercase characters in the bit of the URL that is not the repo/package/whatever.

SethFalco: Oh strange. 🤔 
I'll update this PR later.

I tested case-sensitivity and it worked when I tried it, but I used `Parserfunctions`, which returned no result. (So I assumed the extension name was itself case-sensitive.)

Not sure why `parserFunctions` works, but `Parserfunctions` returns no result. 🤔

chris48s: So there is also an issue that the upstream API behaves somewhat esoterically.

Note that

`wikiapiary/Extension/installs/parserFunctions` causes an error (even though the API does return the stats) but `wikiapiary/Extension/installs/pArserfunctions` is just a not found (which obviously we can't do anything about)

SethFalco: Hopefully, this should be resolved now.

I've addressed the issue with case-sensitivity, using an approach similar to what you suggested. (not identical)

I've also added documentation to the badge to clarify the odd behavior.
Unfortunately, I couldn't see documentation upstream to definitively state exactly what is accepted and what isn't.

I just tested a bunch in [Insomnia](https://insomnia.rest/), and attempted to infer a conclusion from there.
Let me know if that's not favorable!

Screenshot:
image

chris48s: > I just don't know if that's a good idea, as I'm not aware of the upstream behaviour.

No - me either. I think the implementation as it stands is fine :+1: 

> One thing you could drop an opinion, on line 97 I throw an error if the key we're after can't be found.

The check should probably be `if (resultKey === undefined)...` rather than `if (!resultKey)...`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#return_value

> I don't know of a way to actually trigger than error because logically, it should never get there.

I'm not exactly sure what you mean here. I think throwing a `NotFound` is probably an OK choice of exception there. If you're not sure how to write a test for it, you could write a mocked test (see the docs: https://github.com/badges/shields/blob/master/doc/service-tests.md#5-mocking-responses ) and then you can make the request return an intentionally malformed object under to test to hit that case. If you mean something else maybve you can give an example?

SethFalco: > I'm not exactly sure what you mean here.

I just meant, it appears that the upstream API would never normally return a response that would trigger such an error.
In other words, I was clarifying that I was just writing [defensive code](https://en.wikipedia.org/wiki/Defensive_programming), rather than covering an actual edge-case.

Since you mention it, I've tried adding a test case anyway, but it's not going well. 🤔 

```js
t.create('Malformed API Response')
  .get('/extension/installs/ParserFunctions.json')
  .intercept(nock =>
    nock('https://wikiapiary.com')
      .get('/w/api.php?action=ask&query=[[extension:ParserFunctions]]|?Has_website_count&format=json')
      .reply(200, { query: { results: { "Extension:Malformed": { printouts: { "Has website count": [ 0 ] } } } } })
  )
  .expectBadge({ label: 'installs', message: 'not found' })
```

I've spent a fair bit of time on this, but I fail to see what's wrong. 🤔 

```log
AssertionError [ERR_ASSERTION]: Mocks not yet satisfied:
GET https://wikiapiary.com:443/w/api.php
```

I can make this work if I use regular expression, but not with a string literal.
Both of the following work:
* `/.+/`
* `/^\/w\/api\.php\?action=ask&query=\[\[extension:ParserFunctions\]\]%7C\?Has_website_count&format=json$/`

Trial and error hell, but I can use a strict regular expression if I replace `|` with the URL encoded equivalent.
However, the string literal equivalent still doesn't pass.

paulmelnikow: Hi Seth! 👋🏼 

I would programmatically URL-encode the weird bit of your query string:

```js
.get(`/w/api.php?action=ask&query=${encodeURIComponent('[[extension:ParserFunctions]]|?Has_website_count')}&format=json`)
```

Alternatively you can use `.query()`:

```js
.get('/w/api.php')
.query({ action: 'ask', query: '...', format: 'json' })
```

If that doesn't work, you can debug the nock match with all sorts of verbose detail by setting `DEBUG=nock.*` when you run the test. (Add `.only()` to the test if necessary.)

SethFalco: > If that doesn't work, you can debug the nock match with all sorts of verbose detail by setting DEBUG=nock.* when you run the test. (Add .only() to the test if necessary.)

Ahh! Thanks for that!
I had already tried URL encoding the parameters before.
Unfortunately, the `.query` method didn't help either.

However... those were my fault.
Thanks to your tip for debugging the nock requests, I found the issue was with the request being made in the service class. The request wasn't being URL encoded in the service, so the interceptor wrongly interpreted/compared it from there.

Actual
```
{ action: 'ask', query: '[[extension:ParserFunctions]]|' }
```

Expected
```
object {
  action: 'ask',
  query: '[[extension:ParserFunctions]]|?Has_website_count',
  format: 'json'
}
```

chris48s (approved):

paulmelnikow:
Hey @SethFalco I'm curious what's gotten your interested in working on Shields! Would you like to have a chat over Zoom sometime? (Face-to-face is great but voice is fine too if you prefer.) No pressure! If you're interested, want to reach out to me on Discord and we can schedule something?

SethFalco:
Hey @paulmelnikow!

There's no special reason, to be honest. I just like contributing to projects I depend on or find interesting. ^-^'
I saw the backlog and figured it'd be fun to close a service-badge issue occasionally.

And sure, I'm up for a chat if you're certain! I have Zoom and Discord already, if you want you can just add me? (Seth#1276)

PR #6679, opened 2021-06-27
add github actions to build/push docker images

### LINKED ISSUE #6634: Build Docker images in CI
:clipboard: **Description**

Docker are disabling autobuilds for free users on 18 June https://www.docker.com/blog/changes-to-docker-hub-autobuilds/ so I think our Dockerhub releases will stop working. Shields does not currently qualify for their Open Source program because CC0 is not an OSI-approved licence https://www.docker.com/blog/expanded-support-for-open-source-software-projects/

Currently the setup we have for Dockerhub is:
- On any push to master, Dockerhub does an autobuild and updates the `next` tag
- When we push a tag matching the regex `/^server\-.+$/`, Dockerhub does an autobuild and tags the image with the same tag

In order to keep things working, we need to switch to building the images ourselves in CI and then pushing them to Dockerhub. We'll need either a github action or Circle CI job (we're currently using a bit of both on this repo) to:
- `docker build`
- `docker login` (we will need to store a token in secrets to use here)
- `docker push`
- `docker tag`
- `docker logout`

in response to the relevant commit/tag events. Note we can continue to use Dockerhub as a registry - we just have to build the images ourselves and puh them rather than having Dockerhub do it for us.

chris48s:
Closes #6634

We made this switch on another repo I work on so I basically already had this knocking about.. these two actions are very slightly adapted from the solution we've used there.

In anticipation, I have set up `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` variables in https://github.com/badges/shields/settings/secrets/actions . I've used a token attached to my dockerhub account (which _should_ have permission to push to `shieldsio/shields` on dockerhub). We could maybe look at setting up a special bot/machine account for this though if we wanted.

Interestingly the DockerHub builds for our repo seem to still be working even though they were theoretically turned off last week. I did have a look to see if Docker have announced a U-turn or extension but the latest word is that autobuilds for free plans should have been disabled on 18th June. We could just keep this PR "in the chamber" and wait to see if they actually break at some point before we make the switch. If we want to just go ahead with this, before we merge this we should probably disable the autobuilds on DockerHub so we don't have two builds trying to push to the same tags.


.github/workflows/publish-docker-snapshot.yml
```diff
@@ -0,0 +1,30 @@
+name: Build and Publish Snapshot Docker Image
+on: create
```
calebcartwright, line 2: I gather GitHub still haven't added a `tag` create filter to this like they have for `release`?

calebcartwright (approved):

calebcartwright:
I'm inclined to go ahead and make the pivot vs. waiting to see. Maybe we could hold off till the July image publish though.

.github/workflows/publish-docker-snapshot.yml
chris48s: Yeah there isn't a specific event for tag creation. `create` is fired when a new ref is created. A ref can either be a tag or a branch, so the way to run an action on tag create is

```yaml
on: create
jobs:
  build:
    if: github.event.ref_type == 'tag' 
```

chris48s:
> Maybe we could hold off till the July image publish though.

Published ok using native dockerhub build:
https://hub.docker.com/layers/shieldsio/shields/server-2021-07-01/images/sha256-967e517c4808936a69a7d80c683be345934d39918ce4e215e58b91fbaa4e8fba

calebcartwright:
> Published ok using native dockerhub build:

Indeed. Think I'm just a bit paranoid about waiting and having the on-demand/latest builds failing without us realizing.

I don't feel strongly about it at all though so we can certainly park it for a while if you'd prefer

chris48s:
Going to just leave this one "in the chamber" while the DockerCloud builds continue to run.
TODO: If we need to resurrect it, we could also run just the `docker build` action (not the push/tag) as a CI check on pull requests.

chris48s:
Looks like DockerHub disabled our builds about a week ago so I am going to merge this.
I'll move `docker build` as a CI check to an issue.

PR #6681, opened 2021-06-27
convert bug report issue template to issue form

chris48s:
GitHub issue forms are now in public beta. This allows you to do stuff like defining dropdown menus, setting required fields, etc in issue templates.
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms
Currently our bug report template (ab)uses a task list which means all our issues say they have X tasks completed e.g:
Screenshot at 2021-06-27 17-21-47
This makes it a good candidate to convert to an issue form, so I decided to have a go. This will change our bug report template to look like this:
https://github.com/chris48s/shields/blob/bug-form/.github/ISSUE_TEMPLATE/1_Bug_report.yml
I haven't looked at any of our other templates, but we could look to converting them too if we think it is useful..

One thing to note is that issue forms do also have a checkbox construct, but there isn't a validator that says "require >=1 boxes to be checked" (maybe there will be at some point). I decided to make the "product" (for want of a better word) choice a dropdown and make it required. This means you can't pick >- things and we also can't use markdown. The other option would be we could make it a checkbox element and include links in the descriptions but that also makes it skippable.. I don't have a super-strong preference either way. I think both are  a slight improvement over what we are doing with markdown now.


calebcartwright (approved):
I'm on board with giving this a whirl. I think the tasklist mapping is as good a reason as any, and wouldn't be surprised if we end up needing/wanting other changes as GitHub start incorporating and rolling out more of the advanced issues/boards type feature.

PR #6684, opened 2021-06-28
test: add generic metric helper [BitbucketIssues BitbucketPullRequest GithubMilestoneDetail AmoDownloads DubDownloads EclipseMarketplaceDownloads GithubCommitActivity hexpm HomebrewDownloads JsDelivrHitsGitHub JsDelivrHitsNPM NpmDownloads packagecontrol PackagistDownloads PypiDownloads Sourceforge StackExchangeMonthlyQuestions wordpress]

SethFalco:
In #6677 I suggested the following:

> ```js
> const isMetricWithText = text => {
>   const pattern = `^([1-9][0-9]*[kMGTPEZY]?|[1-9]\\.[1-9][kMGTPEZY]) ${text}$`
>   const regexp = new RegExp(pattern)
>   return withRegex(regexp)
> }
> 
> const isMetricOpenIssues = isMetricWithText('open')
> ```
> \- https://github.com/badges/shields/pull/6677#issuecomment-869208336

I did not implement the exact suggestion, but this attempts to achieve the same result.
This creates a new helper function called `isMetricWithPattern`.

Usage:
```js
t.create('Translations')
  .get('/user/nijel/translations.json?server=https://hosted.weblate.org')
  .expectBadge({ label: 'weblate', message: isMetricWithPattern(/ translations/) })
```

Matches:
* `246 translations`
* `32k translations`

This also modifies other metric test helpers to build on-top of it to reduce duplication.

I opted to make it use a pattern instead of text because JavaScript does not support the `\Q` and `\Z` (quote) syntax. Rather than handle escaping the input, I thought it'd be better to just accept a pattern. There is a use-case for this anyway (the lines immediately below it) so I think it's a valid approach.

#### Possible Alternatives
* Currently, it takes a RegExp which will just be converted to a string to build a new RegExp anyway. This could be scrapped so it just takes a string parameter. (However, this would reduce compile-time validation.)
* What do you think of this solution, or would you prefer something closer to the initial suggestion and that I not touch the other methods?

### DIFF
--- a/services/test-validators.js
+++ b/services/test-validators.js
@@ -64,16 +64,24 @@ const isStarRating = withRegex(
 // Required to be > 0, because accepting zero masks many problems.
 const isMetric = withRegex(/^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY])$/)
 
-const isMetricOpenIssues = withRegex(
-  /^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) open$/
+/**
+ * @param {RegExp} nestedRegexp Pattern that must appear after the metric.
+ * @returns {Function} A function that returns a RegExp that matches a metric followed by another pattern.
+ */
+const isMetricWithPattern = nestedRegexp => {
+  const pattern = `^([1-9][0-9]*[kMGTPEZY]?|[1-9]\\.[1-9][kMGTPEZY])${nestedRegexp.source}$`
+  const regexp = new RegExp(pattern)
+  return withRegex(regexp)
+}
+
+const isMetricOpenIssues = isMetricWithPattern(/ open/)
+
+const isMetricOverMetric = isMetricWithPattern(
+  /\/([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY])/
 )
 
-const isMetricOverMetric = withRegex(
-  /^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY])\/([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY])$/
-)
-
-const isMetricOverTimePeriod = withRegex(
-  /^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY])\/(year|month|four weeks|quarter|week|day)$/
+const isMetricOverTimePeriod = isMetricWithPattern(
+  /\/(year|month|four weeks|quarter|week|day)/
 )
 
 const isZeroOverTimePeriod = withRegex(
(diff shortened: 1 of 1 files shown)


SethFalco:
Unless I've misunderstood, I don't think the test that's failing is caused by this change. 🤔 
Is it fine if I put an high number of services in PR title?

I believe the following is every service affected by this change:
`[BitbucketIssues BitbucketPullRequest GithubMilestoneDetail AmoDownloads DubDownloads EclipseMarketplaceDownloads GithubCommitActivity hexpm HomebrewDownloads JsDelivrHitsGitHub JsDelivrHitsNPM NpmDownloads packagecontrol PackagistDownloads PypiDownloads Sourceforge StackExchangeMonthlyQuestions wordpress]`

calebcartwright:
> Unless I've misunderstood, I don't think the test that's failing is caused by this change. thinking

You're (almost) certainly correct. The 5 *'s run the entire suite of our service tests whose purpose I typically describe as a blend of integration tests with the upstream services/platforms and a form of top-down functional tests for individual badges. These are inherently brittle in nature, but have been more consistently broken for a while now (another area that could use some love and attention!). We run the full suite of those tests twice a day, so you can always cross reference one of those runs to see if the tests have been failing against the trunk branch and I believe the failures you were seeing before were the usual suspects that are already failing unrelated to these changes

https://app.circleci.com/pipelines/github/badges/daily-tests 

> Is it fine if I put a high number of services at the end of PR title?

Absolutely!

> I was hoping closing/opening it would force GitHub Actions to kick off again. ^-^'

I believe it typically does, but like many CI providers GitHub have put in some additional gates around PR triggers, especially for authors that haven't committed to the repository's default branch, as part of their efforts to combat abuse. I restarted the circle job and it seems to be happier now

calebcartwright (approved):
Thanks! Would be neat if there was a way to simplify the `isMetricOverMetric` even more but this definitely looks like a solid improvement overall

SethFalco:
> Thanks! Would be neat if there was a way to simplify the `isMetricOverMetric` even more but this definitely looks like a solid improvement overall

Technically, we could if we just replace the `/regex/` with a string instead, and kept a constant somewhere?

```js
const metricPattern = '([1-9][0-9]*[kMGTPEZY]?|[1-9]\\.[1-9][kMGTPEZY])'

/ Required to be > 0, because accepting zero masks many problems.
const isMetric = withRegex(new RegExp(`^${metricPattern}$`)

/**
 * @param {string} nestedPattern Pattern that must appear after the metric.
 * @returns {Function} A function that returns a RegExp that matches a metric followed by another pattern.
 */
const isMetricWithPattern = nestedPattern => {
  const pattern = `^${metricPattern}${nestedPattern}$`
  const regexp = new RegExp(pattern)
  return withRegex(regexp)
}

const isMetricOpenIssues = isMetricWithPattern(' open')

const isMetricOverMetric = isMetricWithPattern(`\\/${metricPattern}`)
```

(not tested since I'm just on the GitHub interface right now, but you get the idea)

calebcartwright:
> (not tested since I'm just on the GitHub interface right now, but you get the idea)

indeed! probably not worth the effort right now but maybe down the road

PR #6712, opened 2021-07-04
fix: weblate badges to better fit repo guidelines [Weblate]

SethFalco:
> * Reduce the total badges and use the `(...|...)` syntax. (as mentioned above)
> 
> * Replace `/license` with `/l`.
>   > License: `/l`
> 
> * Default to `https://hosted.weblate.org` (it seems I can make server optional based on the docs)
>   > The parameter is for an uncommon optional attribute, like an alternate registry URL.
> 
> * Review badge paths
>   > /SERVICE/NOUN/PARAMETERS?QUERYSTRING e.g: /github/issues/:user/:repo
>
> \- https://github.com/badges/shields/pull/6677#issuecomment-873433265

Just let me know if there're any other problems!

image

#### Related Issues
* Relates to: https://github.com/badges/shields/pull/6677


services/weblate/weblate-entity-count.service.js
```diff
@@ -11,64 +10,46 @@ const schema = Joi.object({
...
   static category = 'other'
-
-  static buildRoute(entityName) {
-    return {
-      base: 'weblate',
-      pattern: entityName,
-      queryParamSchema,
-    }
+  static route = {
+    base: 'weblate/count',
```
calebcartwright, line 19: I think we should consider using a slightly more descriptive noun here, both because that'd align better with the class and file name, and also due to the terminology overlap with the user statistics.

Is my understanding correct that these data points reflective of the entire server/instance, whereas the user stats are specific to that particular user on the specified server/instance? I guess phrased a different way, what's the difference this `languages` badge vs the user stats `languages` badge?

My sense is that the entity counts would essentially be the aggregate of _all_ user counts, and if that's the case it may make sense to use consistent terminology (i.e.`count` vs. `statistics`). We don't have to structure our routes based on the upstream API route, so if one of those (or even something different) will make more sense as the noun for our consumers we should consider that.

At a minimum, I'd think we'd need something like this given the current routes:

 ```suggestion
    base: 'weblate/entity-count',
```

**Review by calebcartwright — COMMENTED** [review]
Thank you! One minor topic for discussion left inline

File: services/weblate/weblate-entity-count.service.js (hunk 1 above)
Comment (SethFalco) [reply services/weblate/weblate-entity-count.service.js:19]: Sorry about that!

> Is my understanding correct that these data points reflective of the entire server/instance, whereas the user stats are specific to that particular user on the specified server/instance

Correct! Or just to be 100% we're on the same page:

This route is aggregate stats of the entire server instance.

* projects - Is the total number of projects across the entire Weblate instance created by any user.
* languages - Is the total number of languages enabled across all projects. (ie if a server has 2 projects, 1 has `en`, `fr`, `es`, and the other has `de`, `en-GB`, then it'd be 5 in total)
* users - Is the total number of users on the entire Weblate instance. 

Still happy with `entity-count`, or would you think something else would be better?

Comment (calebcartwright) [reply services/weblate/weblate-entity-count.service.js:19]: > Still happy with entity-count, or would you think something else would be better?

Not so much. It's still unclear to me why something like `languages` is an `entity` when in aggregate but a `user` (statistic) when more granular. Ultimately, the noun really is the entity and the fact that the quantity of said noun is displayed is fairly implicit, much in the way that we don't use the noun `download-count` for our download badges.

It really feels like it should be more like:

* `/weblate/:noun`, e.g. `/weblate/languages`
* `/weblate/:noun/:user`, e.g. `/weblate/languages/nijel`

Where the former is inherently the total and the latter inherently specific to/filtered by the provided user.

For the other user stats, we also need to be consistent with the verbiage since they are actually the noun portion (e.g. `commented` -> `comments`, `uploaded` -> `uploads`, etc.). Any requisite mapping to the upstream API values should happen internally within the service class

Comment (SethFalco) [reply services/weblate/weblate-entity-count.service.js:19]: Ahh, I see yeah. 🤔 
Now that I think about it, I'm an idiot.

In the Wikiapiary PR, I was encouraged to use `:variant` as the first part of the path. I guess I can just do that here as well.
I was thinking I shouldn't use a parameter for the noun.

I'll update this now.

**Comment by SethFalco** [comment]
Hoping I fixed everything up now.
Sorry again for the pain! ^-^'

Let me know if there're any other problems, and I'll address them up.

### Weblate entity count
* Changed title/file/class from `Weblate entity count` -> `Weblate entities`
* Changed path from `/weblate/count/{entities}` -> `/weblate/{entities}`

### Weblate user statistic
* Changed file from `weblate-user-statistics` to `weblate-user-statistic` to match class/title. (whoops)
* Changed path from `/weblate/statistic/{user}/{entities}` -> `/weblate/{entities}/{user}`
* Request/badge now uses the appropriate verb everywhere, and uses a switch to translate it to select the correct property from the API response

**Comment by calebcartwright** [comment]
Sorry for the delay, will look at this one later today. Could you rebase/address the merge conflicts when you get a chance?

> Edit: Unfortunately, I'm unable to see why ci/dockercloud-stage failed. thinking (404)

Don't worry about this one, it's not a required check and has a history of being a bit of a flake. Docker also claim to be shutting this service down (refs #6634) thought they apparently either forgot or are running behind, and this could possibly be a symptom that they've finally done it

**Comment by SethFalco** [comment]
Excuse the mess, I messed up while rebasing and did a forced push which changed 20,000 lines. 😆 (https://github.com/badges/shields/commit/cfb70fcc5373e5a9579efeb3f27189335956f38c) 
I fixed it now, just spammed the PR activity a little. ^-^'

File: services/weblate/weblate-user-statistic.service.js [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+        break
+      default:
+        upstreamStatisticName = statistic
+    }
+
+    const data = await this.fetch({ user, server })
+    return this.constructor.render({
+      statistic,
+      count: data[upstreamStatisticName],
+    })
```
calebcartwright, line 79: One very common pattern you'll see in other service classes is the utilization of a `transform` in addition to the usual suspects like `handle`, `fetch`, and `render`. The `transform` function handles the responsibility of extracting the raw data values from the response and performing any translations as necessary.

It's not always used, especially in the really simple cases where a single field access is all that's needed, but this is a decent example of when extracting out the relevant portions currently here into a `transform` function would be worthwhile since the bulk of the `handle` function is now related to transformation duties.

Finally, you might find it helpful to define the route parameter mapping to response key name as an object, and then use that object to retrieve the output. This will both make the mapping a bit more succinct relative to the `switch`, especially since we're guaranteed to be exhaustive due to the bounds set on the route parameter.

E.g. something like this, which would in turn probably make it more reasonable to keep the data extraction in `handle` and not need a separate `transform` function 
```js
const statisticKeyNames = {
  translations: 'translated',
  suggestions: 'suggested',
  uploads: 'uploaded',
  comments: 'commented',
}
....
async handle({ user, statistic }, { server }) {
    const data = await this.fetch({ user, server })
      statistic,
      count: data[statisticKeyNames[statistic]],
    })
}

File: services/weblate/weblate-entities.tester.js [hunk 3]
```diff
@@ -1,10 +1,6 @@
...
 import { isMetric } from '../test-validators.js'
-export const t = new ServiceTester({
-  id: 'WeblateEntity',
-  title: 'Weblate Entity',
-  pathPrefix: '/weblate',
-})
+export const t = await createServiceTester()
 
 t.create('Components')
   .get('/components.json?server=https://hosted.weblate.org')
```
calebcartwright, line 6: In cases where we have an optional query param used to override a default instance, we like to have test cases that validate both the default (i.e. no query param provided in the test used in the route) and at least one test that validates the query param is usable by supplying the param, _preferrably_ with a different instance than the default when possible (though not necessary)

calebcartwright (approved):
Thank you! Left a couple comments inline that could potentially be addressed in another follow up if you're interested, but I'm going to go ahead and merge this given the main goal of correcting the routes before anyone starts using the badge

PR #6749, opened 2021-07-09
Support optionalDependencies in [GithubPackageJson]; also test [npm]

### LINKED ISSUE #6044: Badge request: GitHub NPM package.json - optionalDependencies
:clipboard:  - `optionalDependencies` from NPM package.json at GitHub need be supported like it supported `dependencies` and `devDependencies`. 

Example of view: `dependencyName: version`.

:link:  - similar to the getting of GitHub package.json `devDependencies` and `dependencies`

:microphone:  - **by default badges demonstrates packages that required to use, but optionalDependencies will provide info to user about packages that used, but can be substituted without damage to package**

paulmelnikow:
Closes #6044


calebcartwright (approved):
lgtm, would it be worth adding an example listing? (since we have examples for the other kinds)

paulmelnikow:
Yea, it could be. The existing examples are representative but I don’t think comprehensive.

It would be great if the example system were more closely aligned to the way routes are designed so we could unambiguously communicate the interface a route provides rather than generating representative examples.

calebcartwright:
> Yea, it could be. The existing examples are representative but I don’t think comprehensive.
> 
> It would be great if the example system were more closely aligned to the way routes are designed so we could unambiguously communicate the interface a route provides rather than generating representative examples.

Oh I hope this PR wasn't held up waiting for me to respond to this :smile: 

Agreed on all counts though. I believe the npm dependency version does have a comprehensive set of examples but definitely not needed here

PR #6755, opened 2021-07-11
Re-apply 'Migrate request to got (part 1)'

chris48s:
Now that we're working with a more stable baseline, lets try this again.
Maybe memory usage shoots up and we end up reverting it again, but lets give it another go before I dive into another solution.

I think the failing service tests are all down to other issues

Refs #6160
Refs #6277
Refs #4655


PyvesB (approved):
I've only had a quick look given that this has already been mostly reviewed. Looks ready to go. 👍🏻

chris48s:
Cool. I've just deployed this. I'll monitor the metrics over the next couple of days.

PR #6756, opened 2021-07-11
Delete old deprecated badges

### LINKED ISSUE #6514: Delete deprecated services
We've currently got 22 deprecated service families, some of which containing more than one `deprecatedService` (29 of them in total). Our earliest deprecation, Gratipay, dates all the way back to 2017.

The number of deprecated services will keep growing over time if we do nothing about them. This is not an immediate problem per say, but it would be nice to tidy after ourselves and keep things under control.

Back in https://github.com/badges/shields/pull/1449#issuecomment-370189049, @paulmelnikow had suggested retaining them around for about a year before removing them for good. What do people think?

Whatever removal policy we come up with, let's document it here: https://github.com/badges/shields/blob/master/doc/deprecating-badges.md

PyvesB:
Fixes #6514.


doc/deprecating-badges.md
```diff
@@ -80,6 +80,37 @@ t.create('no longer available (previously number of layers)')
...
+- Issue Stats
+- JitPack Downloads
+- Leanpub
+- Libscore
+- Magnum CI
+- NSP
+- PHP Eye
+- Snap CI
+- VersionEye
+- Waffle
```
chris48s, line 112: A memorial to the fallen soldiers of the great badge wars 🪖

chris48s (approved):

PR #6764, opened 2021-07-13
Delegate discord logo to simple-icons, which matches the current branding

Cisien:
https://discord.com/branding


Cisien:
This still needs more work, the logo changed as well.

calebcartwright:
Thanks for this! There's a bit of a backstory with the handful of logos we actually maintain here locally, but the gist is that when we switched over to Simple Icons for our logos there were a couple we decided to hang on to because they'd other been provided to us directly by the vendor/owner and/or because the version of the logo in Simple Icons looked objectively "bad" at that time.

Discord was one of these logos, which is why we still have a logo definition. Here's what it use to look like:

image

However, the [Simple Icon version has changed markedly since then and has aligned to Discord's branding](https://simpleicons.org/?q=discord) (including their recent changes), so I actually think we should just drop our custom logo and pull in the Discord logo from the Simple Icons package like we do for 99% of the other logos (cc @badges/shields-maintainers in case others feel differently)

I _believe_ our icon loader will handle this properly simply by deleting the corresponding svg file. Is that something you'd be willing to help with @Cisien?

Cisien:
So much for a seemingly simple change (but when is it ever) 😄 

I probably won't have the time to dig into this deeper, sorry.

calebcartwright:
> I probably won't have the time to dig into this deeper, sorry.

Just to be clear, the target direction is to simply delete https://github.com/badges/shields/blob/master/logo/discord.svg from the repo. I'd suggest that's actually a bit easier than modifying the file :wink:

Cisien:
Ahh, got it. I'll update this PR with the change!

calebcartwright:
Thanks! The Simple Icons logo definitely looks much better now. I'd like to move ahead personally but want to wait to see if any of the other maintainers have any thoughts or concerns. The default logo color is their prescribed "blurple" so I don't think there's any going back to the shade of the old logo. Suppose we could change the default to one of their other permitted colors if anyone feels really strongly, but this seems like a good opportunity to me to get rid of the maintenance of another logo.

(Shields.io with custom logo)
https://img.shields.io/badge/discord-custom-blue?logo=discord


(This PR with the Simple Icons logo)
https://shields-staging-pr-6764.herokuapp.com/badge/discord-custom-blue?logo=discord

calebcartwright (approved):

chris48s (approved):
Yeah I'm on board with dumping the custom logo and just using the SI one

paulmelnikow:
> "blurple"

ew!

paulmelnikow (approved):

Issue #6786, opened 2021-07-19
JSON path support for endpoint


simoncozens:
:clipboard: **Description**

The *dynamic* badge allows you to have a static JSON file containing multiple items, and allows you to select a part of that JSON file using the `query` parameter. You can generate multiple badges from the same JSON file, but it only lets you retrieve the message text from a particular JSON path; setting the color, label, logo, etc. must be done in the badges.io URL.

The *endpoint* badge allows you to have a static JSON file containing one item. You can set color, label, logo, and so on, but only for one item.

If the endpoint badge supported the `query` parameter, then a single file like this:

```json
[
 {
    "color": "brightgreen",
    "label": "Outline Correctness Checks",
    "message": "100%",
    "schemaVersion": 1
 },
 {
    "color": "inactive",
    "label": "Shaping Checks",
    "message": "SKIP",
    "schemaVersion": 1
  },
  {
    "color": "green",
    "label": "Universal",
    "message": "85%",
    "schemaVersion": 1
  },
  {
    "color": "green",
    "label": "Fontbakery QA",
    "message": "81%",
    "schemaVersion": 1
  }
]
```

could be used to generate four different badges dynamically.

paulmelnikow:
Hi there, thanks for the question. Can you say more about the use case? Why is it important that these be stored in the same file?

simoncozens:
Thanks - I'm working on a piece of software called [fontbakery](https://github.com/googlefonts/fontbakery) which performs various QA checks on font files, in different areas - variable font support, metadata, outline correctness, etc. We'd like to be able to generate badges for each of those areas. I suppose we could generate one report file per section; it just seemed inconsistent and surprising as a user that shields has the ability to use JSON path filtering in one service-which-consumes-arbitrary-JSON but not another.

calebcartwright:
> shields has the ability to use JSON path filtering in one service-which-consumes-arbitrary-JSON but not another.

I think the core difference here is that the dynamicjson badge _is_ for arbitrary json, whereas the endpoint badge explicitly is not arbitrary.

Though not explicitly opposed to this I am a bit apprehensive about adding json path to the endpoint badge. Seems to me that the actual endpoint should be responsible for any conditionality. I actually don't recall off hand if our badge server forwards along query params to the endpoint (if not perhaps we should), but if we did then the endpoint could use those param values to serve up something conditionally.

[...]

PR #6787, opened 2021-07-19
Add JSON path support for [endpoint] (also check we didn't break [dynamic])

simoncozens:
This implements my feature request from #6786.

- Split jsonPath into two bits
- Have endpoint take a query parameter and filter based on it
- Tests for endpoint plus query

### DIFF
--- a/services/dynamic/dynamic-json.service.js
+++ b/services/dynamic/dynamic-json.service.js
@@ -1,7 +1,7 @@
 import { MetricNames } from '../../core/base-service/metric-helper.js'
 import { BaseJsonService } from '../index.js'
 import { createRoute } from './dynamic-helpers.js'
-import jsonPath from './json-path.js'
+import { jsonPath } from './json-path.js'
 
 export default class DynamicJson extends jsonPath(BaseJsonService) {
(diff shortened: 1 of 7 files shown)


chris48s:
I'm going to close this PR as it just doesn't really make sense as a feature.

@calebcartwright has already touched on it in https://github.com/badges/shields/issues/6786#issuecomment-882673085 but the reason we use JSONPath selectors for the Dynamic JSON badge and no other badges is because the input to that could be any JSON - it is completely arbitrary. We don't know what we will receive so we need the user to tell us what to extract from it and how. Dynamic XML and Yaml follow the same pattern.

For all of our other badges (including endpoint) we know what we are expecting to receive back from the upstream API so we already know how to interpret it. We don't allow the user to pass an arbitrary JSONPath selector to our NPM Version badge, PyPI License badge or Travid Build Status badge (for example) because we know the shape of the response we are expecting. We know what values we need to extract and what we're going to do with them. The endpoint badge is the same: We've defined the requirements of the input we're expecting. We know how to parse it and we know what we're going to do with the values.

PR #6788, opened 2021-07-19
Update simple icons dependency to 5.7.0

dirien:
Update simple icons dependency to 5.7.0

Signed-off-by: Engin Diri <engin.diri@mail.schwarz>


calebcartwright:
Hi, thanks for opening a PR but I'm going to close this.

Our policy and process for incorporating Simple Icons updates is discussed [here](#5369) and we're not going to do ad-hoc ones like this

dirien:
hi @calebcartwright , link is not working…

calebcartwright:
> hi @calebcartwright , link is not working…

#5369

dirien:
Thanks for the link @calebcartwright. Why i cant argue about the content and the procedure behind, i just can say that simple-icons release on a Sunday.

So it may help, to check on a Monday for the dependabot. This could help to reduce "waiting" time 😅 At the end, its just creates a PR, the time to merge it is still on your hand. Just a suggestion, especially when a feature part of the shields functionality relies on simple-icons.

Nevertheles, you folks a the shields.io project to good work and i think everybody appriciate the work you put in! ❤️

calebcartwright:
> Nevertheles, you folks a the shields.io project to good work and i think everybody appriciate the work you put in! heart

Thanks for the kind words!

> Why i cant argue about the content and the procedure behind, i just can say that simple-icons release on a Sunday.

> So it may help, to check on a Monday for the dependabot. This could help to reduce "waiting" time sweat_smile At the end, its just creates a PR, the time to merge it is still on your hand. Just a suggestion, especially when a feature part of the shields functionality relies on simple-icons.

Feels like you may be missing the point here. We are well aware that we don't always have the absolute latest bleeding edge version of every dependency in our dependency tree; that's not a goal of our project. We balance the maintenance/operational overhead of applying dependency upgrades by updating our dependencies consistently on a regular, weekly schedule.

We quite literally just did an update of Simple Icons to v5.6 a couple days ago as part of our regularly scheduled cadence. As noted in the linked discussion, we run our dependency updates, via Dependabot, on Friday. Not Monday. It doesn't matter if one of our dependencies publishes a new release in between our regular updates (outside some emergency security concern) because we'll pull the newer version in during our next dependency update.

We simply aren't going to update the package and deploy it on a daily basis/instantaneously every time Simple Icons (or any of our other dependencies) release a patch update. There's essentially a maximum possible wait of 7 days if SI happen to release another new version immediately after we update on Friday, and that doesn't happen too often.

I appreciate folks' desire to use some brand-new icon just released by SI, but they'll either have to wait for a few days if they want to use that icon with our badges via a logo slug, or if they are unwilling to wait then they can use our custom logo feature with the base 64 encoded version of the SI logo

https://github.com/badges/shields/blob/master/doc/logos.md#custom-logos

PR #6790, opened 2021-07-22
[weblate]: authenticate weblate requests

SethFalco:
Updates the Weblate badges to authenticate on the official instance. (https://hosted.weblate.org)
To keep things a little tidier, also defines `BaseWeblateService` to centralize some Weblate code.


calebcartwright:
Thanks for following up with this. Do you know if there's any kind of read-only API token or other form of scoping? (i forget if we'd discussed this before)

SethFalco:
Unfortunately, not that I'm aware of.
The only read-only API would be the unauthenticated API.

I did just check, and I see nothing in the documentation about it, and on the API key page the only option is "Regenerate API Key" which has no other options, it'll just revoke the old key for a new one.

services/weblate/base-weblate-service.js
```diff
@@ -0,0 +1,21 @@
...
+import { optionalUrl } from '../validators.js'
+
+export default class BaseWeblateService extends BaseJsonService {
+  static queryParamSchema = Joi.object({
+    server: optionalUrl,
+  }).required()
+
+  static auth = {
+    passKey: 'weblate_api_key',
+    authorizedOrigins: ['https://hosted.weblate.org'],
```
calebcartwright, line 12: Will need to remove this. We want Shields server hosters (be it self-hosters or us for Shields.io) to specify which Weblate instances are authorized via the standard configuration options. For the Shields.io runtime we will set this value to `hosted.weblate.org` along with setting our token, but we don't want to strictly lock this in here; otherwise no self-hosters would be able to work with any private Weblate instances.

We only set the value explicitly for things like SaaS-only services

services/weblate/base-weblate-service.js
```diff
@@ -0,0 +1,21 @@
+import Joi from 'joi'
```
calebcartwright, line 1: Let's rename this file. The presence of `service` in the file name can be easily mistaken for an actual service class when browsing contents, so we typically follow a `{upstream service name}-base.js` file naming pattern for base class definitions (e.g. `weblate-base.js`)

calebcartwright (commented):
Thanks! The gist is here but will need to tweak the auth configuration a bit to be able to support working with other Weblate instances

SethFalco:
Sorry about that, hopefully it should be fine now.
I believe I've made all requested changes.

calebcartwright (approved):
LGTM, thanks!

PR #6791, opened 2021-07-22
chore: exclude services from docs

SethFalco:
Related: https://github.com/badges/shields/pull/6677#discussion_r662700176

It was suggested to use a path like `services/**/*.service.js`, however I'm taking a different approach for 2 reasons:
* I'm using `excludePattern`, as it appears to me `exclude` doesn't support wildcards. ([More Info](https://jsdoc.app/about-configuring-jsdoc.html#specifying-input-files))
* It checks for `.js` instead of `.service.js` because the example files in the comment didn't end with `.service.js`, and wouldn't have been excluded otherwise.


calebcartwright:
> it checks for .js instead of .service.js because the example files in the comment didn't end with .service.js, and wouldn't have been excluded otherwise.

I'm going to have to defer to others on this, but my initial reaction is that pattern could be too encompassing. It does make sense to exclude the individual service classes, but there are other files that I could see having docs for being useful for contributors (e.g various helper functions)

SethFalco:
Unfortunate, not that I'm aware of.
The only read-only API would be the unauthenticated API.

I did just check, and I see nothing in the documentation about it, and on the API key page the only option is "Regenerate API Key" which has no other options, it'll just revoke the old key for a new one.

chris48s:
Yeah I think I'd be inclined to go with something like

```js
"excludePattern": ".+\\.service\\.js$"
```

We do have some "base" files under `/services` that contain helpers used by quite a few other badges. We haven't really done it (yet), but I think there is value to documenting those and including them in the docs. I wouldn't want exclude docstrings we add to files like
https://github.com/badges/shields/blob/master/services/npm/npm-base.js or
https://github.com/badges/shields/blob/master/services/github/github-api-provider.js
for example.

On reflection, I think the existing files with docstrings under `/services` do actually match this pattern of being docs for shared helpers not individual services.

SethFalco:
Sure, I've updated it to exclude all `*.service.js` files regardless of parent directory.

---

We currently don't have any docs in `*.spec.js` files or `*.tester.js` files, but could it be worth adding those to ignore as well for future reference?

Something like the following:
`.+\\.(service|spec|tester)\\.js$`

At the very least, it gives build-docs fewer files to process, but performance isn't a significant issue and there aren't  any reasons to use JSDocs on those files, I don't think.

calebcartwright:
> We do have some "base" files under /services that contain helpers used by quite a few other badges. We haven't really done it (yet), but I think there is value to documenting those and including them in the docs

Agreed

> We currently don't have any docs in *.spec.js files or *.tester.js files, but could it be worth adding those to ignore as well for future reference?

Sounds reasonable to me

calebcartwright (approved):
This LGTM, will hold off on merging to see what Chris thinks though

chris48s (approved):

chris48s:
yep - agreed :+1:

Issue #6792, opened 2021-07-22
Fix logos on community page


chris48s:
:clipboard: **Description**

Heroku, NodePing and Sentry logos on https://shields.io/community are dead

Dexters-Hub:
Hey @chris48s can I work on this?

chris48s:
Sure.
The correct images are in https://github.com/badges/shields/tree/master/static/images
I suspect maybe we moved a bunch of stuff around in https://github.com/badges/shields/pull/6651 and forgot to update a path somewhere..

Dexters-Hub:
I have corrected the path of the sponsor images. But it doesn't seems to work.

![Screenshot (686)](https://user-images.githubusercontent.com/47777599/126771721-14cff170-01ba-4272-9cd9-408251b2201e.png)

Dexters-Hub:
I fixed the error by replacing the image tag to a component tag inside the file. @chris48s is this okay?
- [x] Added the image path
- [x] Replaced Image tag with Component tag
- [x] Made adjustments regarding the height of the SVG

![Screenshot (687)](https://user-images.githubusercontent.com/47777599/126779954-bb826afc-4d7c-4b55-9a08-0755b948645b.png)

chris48s:
If you think you have a working solution can you submit a pull request so a maintainer can review it.
I don't think it should be necessary to turn each `<img>` tag into a component to fix this but this conversation will be easier if we can comment on a diff rather than trying to review a screenshot.

PR #6813, opened 2021-07-24
Fix logos on community page

Dexters-Hub:
Issue: #6792

I fixed the error by replacing the image tag to a component tag inside the file.

- [x] Added the image path
- [x] Replaced Image tag with Component tag
- [x] Made adjustments regarding the height of the SVG


chris48s:
I think this approach of turning the SVGs into inline `<svg>` elements is OK but I think suspect there is still an underlying bundling issue here that is not being addressed. I will merge this as a workaround anyway - thanks.

chris48s (approved):

PR #6832, opened 2021-07-26
Add link to static badge collections

developStorm:


calebcartwright:
I appreciate your suggestion but I'm disinclined to move forward with this. A big part of the motivation behind creating a stand alone awesome list was explicitly to not have any promotional relationship between Shields and that collection, and in my opinion having the Shields.io website deep link to specific sections/listings in the external list violates that desired separation (and would re-open doors to corresponding questions/challenges we wanted to avoid).

Let's see if any of the other maintainers feel differently, though note that the CI checks are currently failing and if we were to decide to move forward with this in some fashion then that'd have to be addressed (`npm run prettier` should do the trick)

PR #6834, opened 2021-07-27
update auto close action for dependabot native

chris48s:
update ~approve-bot~ close-bot for new dependabot behaviour

refs discussion in  https://github.com/badges/shields/pull/6823


.github/workflows/auto-close.yml
```diff
@@ -1,16 +1,20 @@
...
         uses: actions/checkout@v2
 
       - name: Install action dependencies
-        run: cd .github/actions/approve-bot && npm ci
+        run: cd .github/actions/close-bot && npm ci
 
-      - uses: ./.github/actions/approve-bot
+      - uses: ./.github/actions/close-bot
         with:
           github-token: '${{ secrets.GITHUB_TOKEN }}'
```
calebcartwright, line 20: I'm struggling to wrap my head around what is insecure/when with GitHub's provided workarounds. If a malicious user were to modify this file in their fork to remove line 10 and then add a step that would exfil the GITHUB_TOKEN somewhere, would this still be subject to the vulnerability they're referring to?

calebcartwright: Or no actually because the env var isn't available in forks regardless. Okay I think I've got it

calebcartwright (approved):

.github/workflows/auto-close.yml
chris48s: So essentially the potential exploit is:
- Dependabot doesn't use forks - it makes a branch of your repo (so does have access to tokens in CI)
- Dependency X on NPM has been compromised by a bad actor
- Dependabot make a PR pulling in Version Y of dependency X
- Version Y of Dependency X has been modified to steal secrets from CI by aforementioned bad actor
- You are exposed to that malicious behaviour even without merging the PR

By scoping the token permissions to only

```
permissions:
  pull_request: write
```

we limit the damange that can be done with that token. The attacker can't just push code to our repo with it, for example.

PR #6836, opened 2021-07-29
use v5 API for [AUR] badges

### LINKED ISSUE #6835: AUR badges use deprecated and undocumented API to retrieve information

shields.io

### 🐞 Description

IN #612 AUR support was added, which currently lives here:

https://github.com/badges/shields/blob/a08b523b4fc55a962aa2562dd4ac9126bbf032ed/services/aur/aur.service.js#L38-L46

Per the AUR documentation, there is supposed to be a version keyword `v=5`, which is missing. This is philosophically mandatory, but not a technical restriction: for backwards compatibility with the original v1 API a missing version is interpreted as the v1 API.

(Side note: the v=5 API was current at the time of that PR, not entirely sure why it ended up using the v=1 API instead.)

This is undocumented and may be dropped at any time. And in fact, we plan to drop it in the near future, and make specifying a version mandatory. (This will simplify the code a lot, and that's beneficial since we're currently porting it from PHP to python.)

According to server logs, 65% or more of unversioned requests are coming from shields.io, so this change would definitely impact you!

### 🔗 Link to the badge

https://img.shields.io/aur/maintainer/google-chrome
AUR maintainer

### 💡 Possible Solution

Please specify the v=5 API and check that it performs as expected.

This will aid us in evaluating the ability to drop old API versions, and future proof shields.io against AUR service changes.

Thanks!

calebcartwright:
Resolves #6835

### DIFF
--- a/services/aur/aur.service.js
+++ b/services/aur/aur.service.js
@@ -3,23 +3,25 @@ import {
   floorCount as floorCountColor,
   age as ageColor,
 } from '../color-formatters.js'
+import { renderLicenseBadge } from '../licenses.js'
 import { addv, metric, formatDate } from '../text-formatters.js'
 import { nonNegativeInteger } from '../validators.js'
 import { BaseJsonService, NotFound, InvalidResponse } from '../index.js'
 
 const aurSchema = Joi.object({
   resultcount: nonNegativeInteger,
-  results: Joi.alternatives(
-    Joi.array().length(0).required(),
-    Joi.object({
-      License: Joi.string().required().allow(null),
-      NumVotes: nonNegativeInteger,
-      Version: Joi.string().required(),
-      OutOfDate: nonNegativeInteger.allow(null),
-      Maintainer: Joi.string().required().allow(null),
-      LastModified: nonNegativeInteger,
-    }).required()
-  ),
+  results: Joi.array()
+    .items(
+      Joi.object({
+        License: Joi.array().items(Joi.string().required()).allow(null),
+        NumVotes: nonNegativeInteger,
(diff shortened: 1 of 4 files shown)


services/aur/aur.service.js
```diff
@@ -54,29 +56,25 @@ class AurLicense extends BaseAurService {
...
-    return { license }
+    return { licenses }
   }
 
   async handle({ packageName }) {
     const json = await this.fetch({ packageName })
-    const { license } = this.transform(json)
-    return this.constructor.render({ license })
+    const { licenses } = this.transform(json)
+    return renderLicenseBadge({ licenses, color: 'blue' })
```
calebcartwright, line 77: I opted to extend our license badge helper to let call sites specify a default color to use instead of the dynamically derived colors based on the license values. This seems a reasonable change in its own right, though opted to do so here since we have historically hardcoded the color for this badge and didn't want to introduce color changes as part of updating to the latest API

services/aur/aur.service.js
```diff
@@ -54,29 +56,25 @@ class AurLicense extends BaseAurService {
     {
       title: 'AUR license',
       namedParams: { packageName: 'android-studio' },
-      staticPreview: this.render({ license: 'Apache' }),
+      staticPreview: renderLicenseBadge({ license: 'Apache' }),
```
chris48s, line 59:
```suggestion
      staticPreview: renderLicenseBadge({ license: 'Apache', color: 'blue' }),
```

calebcartwright: Good spot

chris48s (approved):

Issue #6854, opened 2021-08-06
Redis authentication support?


nearcatch:
:clipboard: **Description**

Please support Redis password protected instances. I went through the documentation and it appears that Shields can only connect to unprotected Redis instances.

calebcartwright:
> it appears that Shields can only connect to unprotected Redis instances.

That's incorrect.

Additionally, it's very rare that a self-hosted Shields server actually needs to utilize a Redis instance for a pool of GitHub tokens as opposed to just using a single token (https://github.com/badges/shields/blob/master/doc/server-secrets.md#github). The only time you'd actually need to use Redis with a self-hosted instance is if you have a single self-hosted Shields server that's hitting GitHub's rate limits (5k requests/hour to GitHub, or 15k/hour for GitHub Enterprise Cloud).

Are you hitting such rate limits using a single GitHub token? If not could you expand on why you need the Redis-backed pool of GitHub tokens?

Regarding the Redis piece itself, for Shields.io we use, and strongly encourage any self-hosters to use, TLS connections (not enabled by default). We utilize Compose.io for Shields.io Redis hosting with TLS connections utilizing the [ioredis](https://github.com/luin/ioredis) package, and we include the auth information within the Redis URL (e.g. `rediss://user:password@....`). We just use this as the means for specifying Redis authentication credentials as opposed to having an additional set of configuration variables.

https://www.compose.com/articles/ssl-connections-arrive-for-redis-on-compose/

nearcatch:
> Are you hitting such rate limits using a single GitHub token? If not could you expand on why you need the Redis-backed pool of GitHub tokens?

No I'm not, I just had a redis server so I thought I might as well connect it.

> Regarding the Redis piece itself, for Shields.io we use, and strongly encourage any self-hosters to use, TLS connections (not enabled by default). We utilize Compose.io for Shields.io Redis hosting with TLS connections utilizing the ioredis package, and we include the auth information within the Redis URL (e.g. rediss://user:password@....). We just use this as the means for specifying Redis authentication credentials as opposed to having an additional set of configuration variables.

I'm sorry I missed this if it was documented somewhere, thanks.

calebcartwright:
> I'm sorry I missed this if it was documented somewhere, thanks.

No worries! I actually don't think it's called out specifically anywhere, though including auth in the url/connection string with Redis is pretty common and well known. It might be helpful to expand the documentation a bit.

> > No I'm not, I just had a redis server so I thought I might as well connect it.

[...]

PR #6877, opened 2021-08-08
Update self hosting docs

chris48s:


doc/self-hosting.md
```diff
@@ -140,11 +140,6 @@ preconfigured raster server.
 [raster server]: https://github.com/badges/svg-to-image-proxy
 [micro]: https://github.com/zeit/micro
 
-## Persistence
-
-To enable Redis-backed GitHub token persistence, point `REDIS_URL` to your
-Redis installation.
```
chris48s, line 146: Refs #6854

I think we can probably just remove this note. As I understand it, in order to need/use this, a user would have to set up their own GitHub OAuth app to collect tokens for their pool. I think this situation is so specialised to our instance that we can just remove this note rather than documenting the process for running a token pool. Does this seem reasonable?
All we say about `GH_CLIENT_ID` and `GH_CLIENT_SECRET` in the server-secrets docs is "These settings are used by shields.io for GitHub OAuth app authorization but will not be necessary for most self-hosted installations"

calebcartwright: > As I understand it, in order to need/use this, a user would have to set up their own GitHub OAuth app to collect tokens for their pool

Isn't the OAuth just a means for soliciting tokens and then filling the pool though? On the off chance a self-hoster _did_ find themselves hitting Hub's 5k hourly limit, I think they'd be able to resolve by getting one or two users (presumably another member of their team) to provide a token which they could stash in Redis

chris48s: It is, but I don't think we provide any alternative way to populate the pool, do we?

calebcartwright: Let me know if my understanding is off, or if I'm just missing something, but I'm trying to make a distinction between the Shields server loading the tokens from the configured Redis instance, separately from how the tokens get into Redis in the first place.

I know our server also provides the means to accept donated tokens and add those into Redis, but my understanding is that the piece that initializes/loads the tokens is actually agnostic of the mechanism used to insert the tokens to the Redis store. I.e., if I were to manually create a new token and use a script/admin console/whatever to add my token to our Redis instance on Compose, that'd be just like all the other tokens and pulled into the Shields runtime where the server wouldn't care how that token originally got into Redis, wouldn't it?

chris48s: Yes. If you wrote your own script to insert tokens into a redis instance, that would work (I've not actually tested it), but we don't provide one. The pooling code doesn't care how the tokens got there.

Do you think that is a scenario we want to explicitly document for self-hosting users?

calebcartwright: > Do you think that is a scenario we want to explicitly document for self-hosting users?

I think I'm okay with the proposed changes as-is since it's highly unlikely anyone would need it and the current text can obviously be a source of confusion for users. I just wanted to get clarity on whether the token pool was tightly coupled to the OAuth app approach

calebcartwright (approved):

PR #6878, opened 2021-08-08
fix DockerHub snapshot build

### LINKED ISSUE #6850: Fix DockerCloud tagged snapshot build
:clipboard: **Description**

- The action to build and push the 'next' tag when we push to master is working OK: https://github.com/badges/shields/actions/workflows/publish-docker-next.yml
- The action to push the snapshot tag when we push a tag is not working: https://github.com/badges/shields/actions/workflows/publish-docker-snapshot.yml (this should have run when we pushed https://github.com/badges/shields/releases/tag/server-2021-08-01 )

Not sure why, but the workflow is not starting at all so I guess there is probably something wrong with this line
https://github.com/badges/shields/blob/2e59a7a4cc82141b78330fdbb7a8238071642749/.github/workflows/publish-docker-snapshot.yml#L7
but I can't see what it is immediately and I haven't got time to debug it right now. I'll come back to it.

chris48s:
Closes #6850

As explained in https://github.com/badges/shields/issues/6850#issuecomment-894480397 I think the reason why this isn't firing is probably not because the condition is wrong but because the tag is being created by another workflow and activities which are performed by an action/workflow don't themselves trigger additional workflows (see https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token for more info)

I think the easiest thing to do here is just to move the code that pushes the image to DockerHub into the same action that creates the GitHub tag.


calebcartwright (approved):
lgtm

PR #6883, opened 2021-08-11
Migrate flat, flat-square, plastic and social to use `XmlElement`

### LINKED ISSUE #6535: Migrate flat, flat-square, plastic and social to use `XmlElement`
:clipboard: **Description**

We introduced this for the FTB style in #5754
Update the rest to match

chris48s:
Closes #6535

This PR applies the work we started in https://github.com/badges/shields/pull/5754 to the other 4 styles. This fully delegates the responsibility for XML escaping and formatting/minification to the `XmlElement` class making it easier to "fall into the pit of success" when it comes to escaping.

One thing to note in review is I've been careful to replicate the existing behaviour such that the outputs exactly match the existing outputs without changing the snapshots at all. This should give us a high level of confidence that these changes don't make any functional changes to the XML output. You can also use http://127.0.0.1:3000/dev/styles/ for a visual side-by-side visual comparison.

I've attempted to rebase the history into something vaguely sensible to try and make the review easier :crossed_fingers:


badge-maker/lib/xml.js
```diff
@@ -73,4 +73,27 @@ class XmlElement {
   }
 }
 
-module.exports = { escapeXml, stripXmlWhitespace, XmlElement }
+class NullElement {
+  // TODO: we can remove this later
+  render() {
+    return ''
+  }
+}
```
chris48s: This was helpful as an intermediate step in the refactoring but it gets deleted in a later commit.

badge-maker/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 4.0.0 [WIP]
+
+- Drop compatibility with Node 10
```
chris48s, line 5: Not strictly related to this PR but the next package release has to be v4 because we've dropped node10. I'm starting the changelog here so we don't lose sight of it.

badge-maker/lib/badge-renderers.js
```diff
@@ -270,25 +178,172 @@ class Badge {
...
+      logoWidth,
+    })
+    this.foregroundGroupElement = this.getForegroundGroupElement()
   }
 
   static render(params) {
     return new this(params).render()
   }
 
+  getTextElement({ leftMargin, content, link, color, textWidth, linkWidth }) {
```
chris48s, line 195: I think there's actually scope to factor out `getTextElement()` into a shared helper (maybe `getBackgroundGroupElement()` and `getForegroundGroupElement()` too), DRY up and have FTB share more logic with flat/flat-square/plastic. At the moment FTB and flat/flat-square/plastic use different attribute order in the XML output though so I think doing this will require tweaking the snapshots. For this reason I think this should be done in a follow-on PR instead of here. The property that we know the code exactly matches the snapshots before and after this PR is probably a valuable property for reviewing plus there is already enough stuff changing at once here.

calebcartwright:
Haven't been able to carve out enough time yet to start going through this in detail but hoping to do so later this week

badge-maker/lib/xml.js
```diff
@@ -73,27 +73,24 @@ class XmlElement {
...
 
   render() {
-    return this.content.reduce((acc, el) => acc + el.render(), '')
+    return this.content.reduce(
+      (acc, el) =>
+        typeof el.render === 'function'
+          ? acc + el.render()
+          : acc + escapeXml(el),
+      ''
+    )
```
calebcartwright: I can't convince myself that I understand this part of the diff :laughing: Would you mind explaining it to me?

calebcartwright (dismissed):
Thanks for tackling this one! Everything LGTM. Have left one inline question for my own edification, and one minor (non-blocking) suggested change.

I also want to double check the raster side of the equation out of an abundance of caution (will do here shortly), but we're good to move ahead with this from my perspective

badge-maker/lib/xml.js
chris48s: Yep so basically `this.content` is a list of things each of which might be:
- an `XmlElement` object
- another nested `ElementList` object
- a string (tbh, this is mostly an empty string when we don't want to render a particular element e.g: https://github.com/badges/shields/blob/b6c5a10548a2ac9471cf0785f18f7b64c5870bdf/badge-maker/lib/badge-renderers.js#L227-L231 )

so if I say 

```js
const list = new ElementList({content: [
  new XmlElement({name: 'foo'}),
  'bar',
  new XmlElement({name: 'baz'}),
]})
```

then I want `render()` to output

```xml
<foo />bar<baz />
```

---

```js
this.content.reduce((acc, el) => acc + el, '')
```

does the same thing as

```js
this.content.join('')
```

so I'm doing that, but then I'm saying in the reduce function:
If this object has a `render()` function (it quacks like an `XmlElement` or an `ElementList`), call `render()` and concat the result. Otherwise (assume it is just a string), escape the string and concat that.

calebcartwright: > tbh, this is mostly an empty string when we don't want to render a particular element e.g:

Ah okay, so even if it's not an element it's not strictly going to be an empty string which requires a more thorough check. makes sense, thanks

badge-maker/lib/badge-renderers.js
```diff
@@ -270,25 +180,188 @@ class Badge {
...
+      attrs: { target: '_blank', 'xlink:href': link },
+    })
+  }
+
+  getLabelElement() {
+    const leftLink = this.links[0]
+    return this.getTextElement({
+      leftMargin: this.labelMargin,
+      content: this.label,
+      link: !shouldWrapBodyWithLink({ links: this.links }) && leftLink,
```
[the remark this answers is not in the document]
chris48s: I accepted this because it looked kinda sensible. Then fortunately a test failed :)
`link` here isn't a boolean - its the link URL. Order matters
so this is behaving more like `link: !shouldWrapBodyWithLink({ links: this.links }) ? leftLink : undefined`
Since we both looked at that and thought it was a bool, maybe writing it like that would be clearer?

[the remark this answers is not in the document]
calebcartwright: It looks like this made the tests fail? It's not at all clear to me why that would be the case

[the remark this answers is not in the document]
calebcartwright: oops, comments crossed 

> Since we both looked at that and thought it was a bool, maybe writing it like that would be clearer?

Yeah that makes sense, it's currently subtle/easy to miss clearly :smile:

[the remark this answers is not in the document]
calebcartwright: will go ahead and re-:+1: since this could certainly be deferred like some of other changes mentioned but happy to circle back and do so again if you want to tack that onto this pr

calebcartwright (dismissed):

calebcartwright (approved):

PR #6886, opened 2021-08-12
enforce strict SSL checking for [coverity]

chris48s:
After posting https://github.com/badges/shields/issues/4655#issuecomment-897061551 yesterday, I started looking at the strictSSL/proxy thing and on digging into it.. it might be a non issue. There are two services where we currently do something with the `strictSSL` param, and I think they're both now pointless. In this PR: coverity

We added this in https://github.com/badges/shields/pull/3336 to work around an upstream issue with coverity's SSL cert. This issue now appears to be fixed upstream: https://www.ssllabs.com/ssltest/analyze.html?d=scan.coverity.com&hideResults=on and the tests pass without it. I think it can just be dropped.

### DIFF
--- a/services/coverity/coverity-scan.service.js
+++ b/services/coverity/coverity-scan.service.js
@@ -48,14 +48,6 @@ export default class CoverityScan extends BaseJsonService {
     const json = await this._requestJson({
       url,
       schema,
-      options: {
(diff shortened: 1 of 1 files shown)


calebcartwright (approved):
SGTM

PR #6887, opened 2021-08-12
remove disableStrictSsl param from [jenkins]

chris48s:
After posting https://github.com/badges/shields/issues/4655#issuecomment-897061551 yesterday, I started looking at the strictSSL/proxy thing and on digging into it.. it might be a non issue. There are two services where we currently do something with the `strictSSL` param, and I think they're both now pointless. In this PR: jenkins

I was looking into how to test this param and I realised this param has actaully been broken for a very long time. Any attempt to use this will throw "strict ssl is required" e.g: https://img.shields.io/jenkins/build.svg?disableStrictSSL&jobUrl=https%3A%2F%2Fjenkins-oss.wixpress.com%2Fjob%2Fmulti-detox-master In order to make it work, you have to remove https://github.com/badges/shields/blob/23c0406bedfc6930735e8f5ea75dfe34faf1f290/core/base-service/auth-helper.js#L118
We broke this over a year ago in https://github.com/badges/shields/pull/4729 and nobody has raised an issue about it. Given this was never documented anywhere and has been broken for over a year without anyone noticing, I reckon we can just quietly drop this param rather than fixing it. Clearly nobody is using it.

### DIFF
--- a/services/jenkins/jenkins-base.js
+++ b/services/jenkins/jenkins-base.js
@@ -12,15 +12,11 @@ export default class JenkinsBase extends BaseJsonService {
     schema,
     qs,
     errorMessages = { 404: 'instance or job not found' },
-    disableStrictSSL,
   }) {
     return this._requestJson(
       this.authHelper.withBasicAuth({
         url,
-        options: {
-          qs,
-          strictSSL: disableStrictSSL === undefined,
-        },
+        options: { qs },
         schema,
         errorMessages,
       })
--- a/services/jenkins/jenkins-build.service.js
+++ b/services/jenkins/jenkins-build.service.js
@@ -68,12 +68,11 @@ export default class JenkinsBuild extends JenkinsBase {
     return { status: colorStatusMap[json.color] }
   }
 
-  async handle(namedParams, { jobUrl, disableStrictSSL }) {
+  async handle(namedParams, { jobUrl }) {
     const json = await this.fetch({
       url: buildUrl({ jobUrl, lastCompletedBuild: false }),
       schema,
       qs: buildTreeParamQueryString('color'),
-      disableStrictSSL,
     })
     const { status } = this.transform({ json })
     return this.constructor.render({ status })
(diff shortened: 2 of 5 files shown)


calebcartwright:
Relevant originating issue for reference was #1956. I still think the self-hosting scenarios and the common self-signed cert usage, particularly within intranets, is something we need to take into consideration.

chris48s:
This (jenkins) has also been broken for self-hosted installs since ~March 2020. As far as I can tell, the only way you'd have a self-hosted install where this works is if you are currently running a pre-March 2020 install or customising by running your own fork. Is there another scenario I'm missing here? (entirely possible)

calebcartwright:
> where this works is if you are currently running a pre-March 2020 install or customising by running your own fork. Is there another scenario I'm missing here? (entirely possible)

That's correct, but not as far fetched as it may sound; we can't assume self-hosters redploy with an updated instance on a regular basis because there's rarely motivation for them to do so.  The concern would be someone subsequently needing an update (security patch, general rehydration, etc.) to discover the issue.

calebcartwright:
And to clarify, I'm not necessarily opposed to this/am undecided, just want to highlight that the absence of reported issues isn't as strong a point of evidence as it typically is IMO due to the self-hosting vector

chris48s:
Cool - just wanted to check there isn't another case I am not seeing here.

calebcartwright:
Having thought on this a bit more, I feel like there are viable options for folks using a self-hosted Shields server. Seems reasonable to assume that in many corporate/intranet scenarios that there would be a corporate-level private CA in use, and even in other cases it wouldn't be that difficult for someone to create their own and add it to both their Jenkins and Shields server.

The Endpoint badge could always be used as an intermediary where they could disable the strict SSL check on the call to Jenkins from the endpoint.

Obviously that story is more tricky for folks trying to use Shields.io (other than the Endpoint badge), but as discussed above the ability to disable the check with Shields.io has been broken for a long time already anyway.

I'm feeling more confident about our ability to move ahead with this :+1:

calebcartwright (approved):
I think I've managed to talk myself into this :smile:

PR #6904, opened 2021-08-16
switch from husky to simple-git-hooks

### LINKED ISSUE #6507: move pre-commit hook from husky to simple-git-hooks
:clipboard: **Description**

https://github.com/toplenboren/simple-git-hooks

chris48s:
closes #6507

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -112,7 +112,6 @@
         "gatsby-plugin-styled-components": "^4.6.0",
         "gatsby-plugin-typescript": "^3.2.0",
         "humanize-string": "^2.1.0",
-        "husky": "^4.3.8",
         "icedfrisby": "4.0.0",
         "icedfrisby-nock": "^2.1.0",
         "is-svg": "^4.3.1",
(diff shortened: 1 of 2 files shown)


calebcartwright:
We should add a line to the Contributing and/or Tutorial docs to let folks know that if they want to opt-in to client side git hooks then they'll need to run `npx simple-git-hooks` (or `./node_modules/bin/simple-git-hooks` for non-`npx` users). Since we have `lint-staged` in our tree I wonder if we need `npx` in the script?

chris48s:
> We should add a line to the Contributing and/or Tutorial docs to let folks know that if they want to opt-in to client side git hooks then they'll need to run `npx simple-git-hooks`

You should find that now when you `npm ci` it registers the hooks in a post-install job. You only need to run `npx simple-git-hooks` locally the first time you add a new hook.

>  I wonder if we need `npx` in the script?

I found I did need `npx`. With just

```json
"simple-git-hooks": {
  "pre-commit": "npx lint-staged"
}
```

it throws `.git/hooks/pre-commit: 2: lint-staged: not found`

calebcartwright:
> You should find that now when you npm ci it registers the hooks in a post-install job. You only need to run npx simple-git-hooks locally the first time you add a new hook

Gotcha, thanks for explaining

> it throws .git/hooks/pre-commit: 2: lint-staged: not found

Ah okay, so the node mods bin directory is no longer on the path during execution context, hence npx. All makes sense now :grin:

calebcartwright (approved):

PR #6914, opened 2021-08-19
Try out node-fetch instead of got; test [*****]

chris48s:
The saga continues
Refs https://github.com/badges/shields/issues/4655

OK, so lets start off with a summary of where we are.:
We migrated the majority of our HTTP traffic to got. There's definitely some noticable increase in memory usage compared to request, although we're now running with a bit more headroom. Before doubling down on got and moving the rest of our traffic, lets try out node-fetch and see if we get better performance with that first. Interesting side note is that _right now_ our resource usage seems to be weirdly low. Maybe everyone is on summer holidays? Once reviewed we might want to hold off deploying this until our usage looks a bit more typical.

I think having got this far with node-fetch, if the resource usage seems about the same I think I'm in favour of got because the higher-level abstraction is just a bit nicer to work with, but if it looks like node-fetch improves our memory footprint I think its a fine solution to move forward with.

For the moment I'm really just interested in seeing what happens to performance with node-fetch vs got so
- I'm going to leave `influx-metrics.js` using got for the moment. Its not on the 'critical path'. Once we've decided on a way forward, we can standardize one one client but while we work it out we're going to have a random mashup of some stuff using node-fetch, some stuff still using request and influx-metrics.js using got.
- There are some rough edges we might want to sand down if we commit to this approach. More on this inline.

@calebcartwright are you able to double-check this works ok through a proxy when you get a chance?

There will be a few failing service tests on the build, but they should all be unrelated (I'll double-check when the build completes)


core/base-service/node-fetch.js
```diff
@@ -0,0 +1,101 @@
...
+    } else if (value === null) {
+      qs[key] = ''
+    } else if (['string', 'number', 'boolean'].includes(typeof value)) {
+      qs[key] = value
+    }
+  }
+  return new URLSearchParams(qs)
+}
+
+function request2NodeFetch({ url, options }) {
```
chris48s, line 21: There's a bit of a separation of concerns issue going on here.
Part of what I am doing here is translating request concepts to node-fetch concepts
That said, if we go with node-fetch I think there are some abstractions here that we would probably want to retain as an abstraction available to services. For example the qs/form abstraction. I think for the purpose of evaluating performance this is fine and we can clean things up if we decide node-fetch is the way forward.

core/base-service/node-fetch.js
```diff
@@ -0,0 +1,101 @@
...
+  if ('auth' in requestOptions) {
+    const user = requestOptions.auth.user || ''
+    const pass = requestOptions.auth.pass || ''
+    const b64authStr = Buffer.from(`${user}:${pass}`).toString('base64')
+    nodeFetchOptions.headers.Authorization = `Basic ${b64authStr}`
+    delete requestOptions.auth
+  }
+
+  if ('form' in requestOptions) {
+    nodeFetchOptions.body = object2URLSearchParams(requestOptions.form)
```
chris48s, line 67: There are 2 ways you can do form encoding: URL Encoded and Multipart I've implemented URL Encoded here. This:
- is simpler to implement
- is easier to mock with nock
- works with all the services we've got that are sending forms

If we wanted to do multipart encoding instead, that implementation would look like

```js
import FormData from 'form-data'

const form = new FormData()
for (const [key, value] of Object.entries(requestOptions.form)) {
    form.append(key, value)
}
nodeFetchOptions.body = form
const formHeaders = form.getHeaders()
nodeFetchOptions.headers['content-type'] = formHeaders['content-type']
delete requestOptions.form
```

One thing I haven't really got my head round is how higher level clients decide or allow the user to specify when to use each encoding type. Given URL Encoding works for everything we're currently doing, I reckon we can leave this how it is and worry about multipart in future if we move forward with node-fetch and find a use-case that needs it.

core/base-service/node-fetch.js
```diff
@@ -0,0 +1,101 @@
...
+    url,
+    options,
+  })
+  nodeFetchOptions.headers['User-Agent'] = userAgent
+  nodeFetchOptions.size = fetchLimitBytes
+  nodeFetchOptions.follow = 10
+  try {
+    const resp = await fetch(nodeFetchUrl, nodeFetchOptions)
+    const body = await resp.text()
+    resp.statusCode = resp.status
```
chris48s, line 89: I've just done this for compatibility/convenience. There's a bunch of code that expects response to have a `statusCode` property. This is the path of least resistance.

calebcartwright:
> @calebcartwright are you able to double-check this works ok through a proxy when you get a chance?

Yup, will do

PyvesB (approved):
I've had a read through the code. Changes look good to me. 👍🏻

calebcartwright:
Will test this in my env this weekend

calebcartwright:
Working fine behind a proxy :+1:

chris48s:
Right. All seems to be in order. I'll plan to deploy this tomorrow evening

PR #6935, opened 2021-08-21
Switch all shields.io links to https

PyvesB:


calebcartwright (dismissed):

services/website/website.tester.js
```diff
@@ -1,8 +1,8 @@
 import { createServiceTester } from '../tester.js'
 export const t = await createServiceTester()
 
-t.create('status of http://shields.io')
-  .get('/website.json?url=http://shields.io')
+t.create('status of https://shields.io')
+  .get('/website.json?url=https://shields.io')
```
chris48s, line 5: This change doesn't make sense.
The intention of this is that should have one test using HTTP and one using HTTPS:

https://github.com/badges/shields/blob/1c36dd2bd0338cc96c2736d9310bf85ad805d648/services/website/website.tester.js#L4-L12

chris48s (changes requested):

services/website/website.tester.js
calebcartwright: whoops! good catch

PyvesB: Yup, good catch, was a bit eager on the find replace.

chris48s (approved):

Issue #6849, opened 2021-08-01
Run `docker build` as a CI check


chris48s:
:clipboard: **Description**

Refs https://github.com/badges/shields/pull/6679#issuecomment-875895440
If a PR introduces changes that will break the docker image, fail the build

PR #6938, opened 2021-08-22
Docker: use multi-stage build to reduce size

soulteary:
Hi, the current container size is too large, the local image size is close to 1GB.

```bash
docker images | grep shields                                                                                                                                                   
shieldsio/shields           server-2021-07-01              c567959e0651   7 weeks ago     848MB
```

And I've test the local build, the size is almost close to 1GB.

```bash
shields                                       latest                              83a611869ad9   7 minutes ago   919MB
```

For images used in the production environment, it is recommended to use docker multi-stage build, which can significantly **reduce the size of the image in actual local use and reduce the size when pushed to dockerhub**.

```bash
shields                                       latest                              c925e98df538   4 seconds ago   255MB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14-alpine
+FROM node:14-alpine AS Builder
 
 RUN mkdir -p /usr/src/app
 RUN mkdir /usr/src/app/private
@@ -16,8 +16,13 @@ RUN npm run build
 RUN npm prune --production
 RUN npm cache clean --force
 
+# Use multi-stage build to reduce size
+FROM node:14-alpine
 # Run the server using production configs.
 ENV NODE_ENV production
+
+WORKDIR /usr/src/app
+COPY --from=Builder /usr/src/app /usr/src/app
 
 CMD node server
 


soulteary:
Using the above code, I made an image submission, and the image size on DockerHub is 70 MB, :D

https://hub.docker.com/r/soulteary/shields/tags?page=1&ordering=last_updated

calebcartwright:
Thank you for this! Admittedly having a compact image size isn't a real goal of the project, but I am generally in favor of leveraging builder images. However, I've not had any luck getting the image to build locally, neither from the PR ref nor from our default branch.

The changes themselves look fine, but I'm mildly apprehensive about moving forward unless one of the other maintainers is able to get it building given that we just recently lost our docker auto builds and have #6849 pending

chris48s:
I tried checking this PR out and running a `docker build`. For me this throws

```
mv: can't rename 'public': Directory not empty
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! shields.io@0.0.0 build: `run-s defs features && cd frontend && gatsby build && mv public ..`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the shields.io@0.0.0 build script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

..but then I checkout out master, ran a build and it throws the same error so that might be a red herring. I'm not sure I've ever built this image locally.
Is that the same error you got @calebcartwright ?

Shall we see if we can get https://github.com/badges/shields/issues/6849 done first and then come back to this?

calebcartwright:
Yup, that's the same thing I was getting. My weekly jobs that build from source are working fine though, including from master. I wonder if there's interference from the output of other scripts/general inner dev loop that don't pop up in the clone-and-build only scenarios

chris48s:
Right yeah I just tried it with a clean checkout and I think it is because of a dirty local working tree, not changes in this PR.
This branch does build and boots a container ok for me :+1: 
While I was doing this I did spot another completely unrelated issue with the docker build though so I'll submit another PR with that fix in a moment..

calebcartwright (approved):
Approving since changes LGTM and we've identified the (high level) cause build issues we were experiencing locally. Going to hold off merging in case you'd still prefer we tackle #6849 first @chris48s

PR #6941, opened 2021-08-24
use the right version of NPM in docker build

chris48s:
Docker build was running using NPM 6. It was probably fine but generating some warnings. This ensures we're running 7 or above and `lockfileVersion: 2` will definitely be parsed correctly.


calebcartwright (approved):
Good catch!

PR #6944, opened 2021-08-25
Add [ClearlyDefined] service

mbtools:
## Description 

This service is for displaying the overall score of repositories as defined by ClearlyDefined. ClearlyDefined is part of the Open Source Initiative, and on a mission to help FOSS projects thrive by being, well, clearly defined. For more details visit
https://clearlydefined.io/about

The badge will display the overall score out of 100 for a given repository. The badge describes how well licenses are defined. Since it's basically an analysis of license metadata and it has been placed under the "Analysis" category. 

Example of a repository on ClearlyDefined:
https://clearlydefined.io/definitions/npm/npmjs/-/jquery/3.4.1
The badge will display the top right number on the screenshot using the "score/100" format. For the given example, the badge will show "Score: 88/100". 

Corresponding API call:
https://api.clearlydefined.io/definitions/npm/npmjs/-/jquery/3.4.1

image

## Data 

The public API is available as Swagger UI and does not require any API key:
https://api.clearlydefined.io/api-docs/

## Motivation 

The badge shall promote the ClearlyDefined initiative and help to bring more clarity around licenses in the open source community.  You can find more details at https://docs.clearlydefined.io/.


services/clearlydefined/clearlydefined-score.service.js
```diff
@@ -0,0 +1,39 @@
...
+      keywords,
+    },
+  ]
+
+  static render({ score }) {
+    score = Math.round(score)
+    return {
+      label: 'ClearlyDefined',
+      message: `${score}/100`,
+      color: floorCountColor(score, 40, 60, 100),
```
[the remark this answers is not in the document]
mbtools: I didn't find it documented. It's probably buried in the website code.

mbtools:
I didn't get any errors on prettier. Maybe it was an incorrect end of line. Anyway, it worked. 

I couldn't figure out if there's a way to see the server response in the Circle CI jobs for the tests. So getting the testing right took some trial and error. There actually IS a npm package called "not-a-real-package"... (https://www.npmjs.com/package/not-a-real-package). So expecting "not found" wasn't going to work. 

Well, I figured it out and now it's all working 😄

calebcartwright:
I'm really confused by the messages used for those last two tests and the fact that they are passing. At least locally I'm getting a 404 response against the upstream api with those params, which is standard/expected, and our service class framework should be returning a `not found` in those cases.

I'll take a closer look at this later today or tomorrow

mbtools:
The second case exists in the ClearlyDefined data. It is rated 0. So we get a valid json response. Kind of the same as first case so we could remove it. 

I was expecting 404 for the third case as well. But it returns a json with an error code.

calebcartwright:
So I grabbed the ref for this PR and am working with it locally, and one of the things that concerns me is the how long the ClearlyDefined API takes to respond, at least on the first call.

The badges our service provides are primarily displayed in places like GitHub readmes which enforce a fairly narrow round trip window in order for things like our badges (and any image/fetched content) to render. If the entire request/response workflow doesn't complete within ~3-4 seconds then GitHub won't display the result. It seems like the initial request for a given package revision can take north of 10 seconds, at least in my anecdotal experience.

Do you know if it's performing the entire analysis on demand when it receives the request? If so then I don't think this badge is going to be viable.

mbtools:
Hmm. My current understanding is that the analysis happens when repos are harvested by ClearlyDefined. The API call shouldn't take long but then who knows how many people have actually tested this. I will request details from the CD folks. Let's put this PR into draft until we have clarity because I agree, it needs to be reasonably fast

mbtools:
The API now respondes in 0.5 and 1.5sec. Looks ok with my tests (from Canada).

calebcartwright:
> The API now respondes in 0.5 and 1.5sec. Looks ok with my tests (from Canada).

Thanks for following up, it does indeed seem much more snappy now. The error behavior is still a bit of concern for me. Here's my observations, let me know if you expect and/or see anything different:

* Invalid/unknown revision: HTTP 200 response with a score of 0
* Invalid/unknown package: HTTP 200 response with a score of 0
* Invalid/unknown namespace: HTTP 200 response with a score of 0
* Invalid/unknown provider: HTTP 500 response
* Invalid/unknown type: HTTP 500 response

That's not necessarily a blocker, but I'd suggest that's a non-ideal response paradigm for an API and complicates our error handling and presentation to users. We could handle the 500 cases and set the message to be something like `unknown type/provider, or upstream issue` (with the latter to account for genuine internal server type errors unrelated to the bad params), but is there anything possible with the other 3 bad input scenarios? Is a score of 0 possible with valid inputs?

mbtools:
No response yet https://github.com/clearlydefined/service/issues/870

calebcartwright:
> No response yet [clearlydefined/service#870](https://github.com/clearlydefined/service/issues/870)

Thanks so much for following up on that! I suggest we hold off for a bit to see what comes out of that, but if there's no news in the near future we can go ahead and proceed.

calebcartwright:
I'm tracking along with the upstream discussion, and while it's good to see an active dialog, I'm also getting the sense that's not anything that's likely to change any time soon.

We definitely wouldn't want to introduce an extra call and sequential call chain for our purposes, so I think we'll be stuck with the existing behavior as-is. What do you think?

mbtools:
It certainly will take a while to sort this on their side, especially since it could be an incompatible change. Let's assume it stays as-is.

The 200 cases are missing certain parts in the JSON. So I can check this and return "not found" for the badge. I'm not sure if I can handle the 500 or if this requires a change somewhere else.

services/clearlydefined/clearlydefined-score.service.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  async fetch({ type, provider, namespace, name, revision }) {
+    return this._requestJson({
+      schema,
+      url: `https://api.clearlydefined.io/definitions/${type}/${provider}/${namespace}/${name}/${revision}`,
+    })
```
calebcartwright, line 50: in the event of a 5xx status code the response handling framework will set the badge message to an according error (it also does so with other codes like 404s, but not necessary here). if you want to map a status code to a specific error that can be done via the `errorMessages` object passed in the options to _requestJson and friends, e.g.

```suggestion
    return this._requestJson({
      schema,
      url: `https://api.clearlydefined.io/definitions/${type}/${provider}/${namespace}/${name}/${revision}`,
      errorMessages: {
        500: 'unknown type, provider, or upstream issue',
      },
    })
```

i think the actual error text could be a bit better so don't feel obligated to use that literally, we just want to account for the fact that we could see a legitimate 500 error response even with a valid type and valid provider

mbtools: The actual error is "An error has occurred" so your suggestion is better :-)

services/clearlydefined/clearlydefined-score.service.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  async handle({ type, provider, namespace, name, revision }) {
+    const data = await this.fetch({ type, provider, namespace, name, revision })
+    // Return score only if definition contains some files,
+    // else it was an incomplete response due to unknown coordinates
+    if (data.described.files > 0) {
+      return this.constructor.render({ score: data.scores.effective })
+    } else {
+      return this.constructor.renderError()
```
calebcartwright, line 77: We actually have a preferred approach these types of scenarios which also hooks into some desirable internal mechanisms in addition to handling the rendering concerns in a consistent manner. If you import `NotFound` from '../index.js` then this line can be replaced with:

```suggestion
      throw new NotFound({ prettyMessage: 'unknown namespace, name, or revision' })
```

which in turn eliminates the need for the separate `renderError` function

calebcartwright (commented):
One last item on the new error handling scenario, then we'll be good to merge! Thanks again for all your and sticking with this even though it had to spawn off some upstream threads

mbtools:
I think we got it. Thanks for your help!

calebcartwright (approved):
Excellent, thank you!

mbtools:
Awesome. I don't see it on https://shields.io/category/analysis yet but I guess there's some follow-up on your end to make it productive.

calebcartwright:
> Awesome. I don't see it on https://shields.io/category/analysis yet but I guess there's some follow-up on your end to make it productive.

Correct. We do automatically deploy merged changes through our staging environment (https://shields-staging.herokuapp.com/), but we don't auto-deploy straight to prod. We typically deploy to prod at least once a week, so odds are pretty high this will be live by Monday if not earlier.

calebcartwright:
https://img.shields.io/clearlydefined/score/npm/npmjs/-/jquery/3.4.1

mbtools:
Woohoo! Works well. 😄
I put one on my repo here (my other repos are still private) 
https://github.com/Marc-Bernard-Tools/Marc-Bernard-Tools-Versions

It was a great learning experience for me. Thanks for your help.

PR #6958, opened 2021-08-27
[freeCodeCamp] Add points badge

SethFalco:
Adds a badge for points a user has on freeCodeCamp. (I'm tempted to say the badge should refer to challenges rather than points, but I'm not 100% sure if completing challenges is the only way to get points.)

I'm unsure about the use of the word "points" in this, as it is indeed "points", but the term is too abstract. Any advice or opinions?

As displayed on a user's profile:
image

As displayed in JSON:
```json
"points": 934,
```


services/freecodecamp/freecodecamp-points.service.js
```diff
@@ -0,0 +1,69 @@
...
+        },
+      },
+    })
+  }
+
+  async handle({ username }) {
+    const response = await this.fetch({ username })
+    this.constructor.validate(response, username)
+
+    const { points } = response.entities.user[username]
```
calebcartwright, line 66: Great idea to separate out the error throwing from `handle`, but let's combine all of this, including the current `validate` function, into a single `transform` function that's called from `handle` That would be much more consistent with our other service classes

calebcartwright (commented):
Thanks! One minor thing but otherwise LGTM. I think using `points` is fine as the default label. Users can always set it to something else if they prefer

calebcartwright (approved):
Thanks!

PR #6959, opened 2021-08-28
refactor [GitHubCommitActivity] to switch to v4/GraphQL API

### LINKED ISSUE #2704: GitHub yearly commit activity badge is flaky

:beetle: **Description**

This badge shows invalid.

:link: **Link to the badge**

- https://img.shields.io/github/commit-activity/y/badges/shields.svg

These work:

- https://img.shields.io/github/commit-activity/4w/badges/shields.svg
- https://img.shields.io/github/commit-activity/4w/eslint/eslint.svg
- https://img.shields.io/github/commit-activity/y/eslint/eslint.svg

:bulb: **Possible Solution**

Not sure! Refactoring the service is probably a good start!

calebcartwright:
Resolves #2704


services/github/github-commit-activity.service.js
```diff
@@ -1,29 +1,35 @@
...
-    pattern: ':interval(y|m|4w|w)/:user/:repo',
+    pattern: ':interval(y|m|4w|w)/:user/:repo/:branch*',
   }
 
   static examples = [
     {
       title: 'GitHub commit activity',
       // Override the pattern to omit the deprecated interval "4w".
-      pattern: ':interval(y|m|w)/:user/:repo',
+      pattern: ':interval(y|m|w)/:user/:repo/',
```
calebcartwright, line 32: was debating whether to add another example that showed the ability to specify a branch/ref but that ties in to a broader question asked below

services/github/github-commit-activity.service.js
```diff
@@ -46,47 +52,67 @@ export default class GithubCommitActivity extends GithubAuthV3Service {
...
+    return this._requestGraphql({
+      query: gql`
+        query (
+          $user: String!
+          $repo: String!
+          $branch: String!
+          $since: GitTimestamp!
+        ) {
+          repository(owner: $user, name: $repo) {
+            object(expression: $branch) {
```
calebcartwright, line 66: Best I can tell this value is used in a call to git's `rev-list` directly or indirectly via `rev-parse` and is a required part of the query. My understanding is that defaulting to this to `FETCH_HEAD` means it'll work with the default branch in the target repo regardless of what that branch is named, and seems like a good idea given we must provide something.

One thing I wasn't clear on was whether the existing v3 API did the same internally or if it spanned across all the branches. At the moment the v3 is still "down" (HTTP 202 - processing) so I can't really compare the results to check for any differences. I think if we go this route with this type of query then it makes sense to surface up the ref that's used here as a route parameter so that users can specify their preference.

services/github/github-commit-activity.service.js
```diff
@@ -46,47 +52,67 @@ export default class GithubCommitActivity extends GithubAuthV3Service {
...
+    return repo.history.totalCount
+  }
+
+  static getIntervalQueryStartDate({ interval }) {
+    const now = new Date()
+
+    if (interval === 'y') {
+      now.setUTCFullYear(now.getUTCFullYear() - 1)
+    } else if (interval === 'm' || interval === '4w') {
+      now.setUTCDate(now.getUTCDate() - 30)
```
calebcartwright, line 105: The classic fun question of "what is a month, really?" :laughing: I know we still have momentjs in our dep tree but I opted to avoid using it to prevent from increasing our dependence on it due to #5573.

IMO showing the commit activity for the last 30 days is a reasonable time period to cover both of these intervals, but suppose we could try to do something more complex. I feel like some of the complexity we had before was more due to the nature of the v3 API response that just gave us total counts in weekly slices, but lmk if folks have any concerns with treating month as the last 30 days

calebcartwright:
Discovered an issue poking around the review app, can't seem to convert the PR back to a draft but working on it

chris48s:
For reference, you should have a button under "reviewers"

Screenshot at 2021-08-28 16-05-21

calebcartwright:
> For reference, you should have a button under "reviewers"

TIL :smile:

calebcartwright:
Comparisons and commentary between the activity shown with this vs. current below. One key difference between the v4 results vs the v3 results is that v4 is currently using an exact date and _time_ window, vs. whole day rounding, so there are slightly different windows at the date boundaries

https://img.shields.io/github/commit-activity/w/badges/shields


https://shields-staging-pr-6959.herokuapp.com/github/commit-activity/w/badges/shields/


These happen to produce the same value (at the moment anyway), but that's largely just a coincidence since our current badge displays the total commits from the _prior_ full week which is something we had to do to account for the structure of the v3 API response to avoid only showing 2-3 days worth of commits on some days earlier in the week.

image

### Monthly
In addition to the start-of-day delta, there's also the 30 day window in this approach vs. the calculated month estimate we had before

#### eslint
https://img.shields.io/github/commit-activity/m/eslint/eslint


https://shields-staging-pr-6959.herokuapp.com/github/commit-activity/m/eslint/eslint



#### shields
https://img.shields.io/github/commit-activity/m/badges/shields


https://shields-staging-pr-6959.herokuapp.com/github/commit-activity/m/badges/shields


Yearly

https://img.shields.io/github/commit-activity/y/eslint/eslint


https://shields-staging-pr-6959.herokuapp.com/github/commit-activity/y/eslint/eslint


https://img.shields.io/github/commit-activity/y/badges/shields


https://shields-staging-pr-6959.herokuapp.com/github/commit-activity/y/badges/shields

services/github/github-commit-activity.service.js
chris48s: The docs are slightly unclear on this, but https://docs.github.com/en/rest/reference/repos#statistics says the stats match the graphs in the front end (e.g: https://github.com/badges/shields/graphs/contributors ). My interpretation is they do only show commits to the default branch. I think this assumption is good

services/github/github-commit-activity.service.js
chris48s: There are other badges where we are making the same assumption e.g:
https://github.com/badges/shields/blob/b02471ff425f1db13d4786f1c7ca933d868f8242/services/packagecontrol/packagecontrol.service.js#L56-L71
https://github.com/badges/shields/blob/b02471ff425f1db13d4786f1c7ca933d868f8242/services/sourceforge/sourceforge.service.js#L22-L24
https://github.com/badges/shields/blob/b02471ff425f1db13d4786f1c7ca933d868f8242/services/homebrew/homebrew-downloads.service.js#L19-L23
https://github.com/badges/shields/blob/b02471ff425f1db13d4786f1c7ca933d868f8242/services/wordpress/wordpress-downloads.service.js#L31-L34
I don't know where it fits in the docs, but I'd be happy to document somewhere that if the upstream api doesn't explicitly provide stats for "month", we assume month = 30 days and formalise this.

chris48s (approved):
I think the core point of this badge is to give a ballpark of how active a repo is rather than to display a precise value. Consistently working with very slightly different assumptions (which might yield slightly different numbers for some users) is better than broken for some users.

services/github/github-commit-activity.service.js
calebcartwright: Cool, glad to be in good company!

calebcartwright:
> I think the core point of this badge is to give a ballpark of how active a repo is rather than to display a precise value. Consistently working with very slightly different assumptions (which might yield slightly different numbers for some users) is better than broken for some users.

Yup completed agreed on this and the documentation front as well. If it starts popping up as a common question/source of confusion then I'll clarify, perhaps as a Discussion topic or somewhere in our markdown docs, as well as inline docs in the badge builder window.

Issue #3200, opened 2019-03-12
Maven badge has "v" prefix


bsideup:
Are you experiencing an issue with...

- [x] [shields.io](https://shields.io/#/)
- [ ] My own instance
- [ ] [gh-badges NPM package](https://www.npmjs.com/package/gh-badges)

:beetle: **Description**
The Maven metadata badges are automatically prefixed with "v" and it is [misleading the users](https://github.com/reactor/BlockHound/issues/9)

:link: **Link to the badge**
[![](https://img.shields.io/maven-metadata/v/https/repo.spring.io/snapshot/io/projectreactor/tools/blockhound/maven-metadata.xml.svg?label=Snapshot)](https://repo.spring.io/snapshot/io/projectreactor/tools/blockhound/)

:bulb: **Possible Solution**
Remove it, or make it configurable

paulmelnikow:
Hi! Thanks for the report.

This is from https://github.com/reactor/BlockHound#getting-it and is probably what is confusing them:

<img width="360" alt="Screen Shot 2019-03-12 at 7 08 33 AM" src="https://user-images.githubusercontent.com/1487036/54195669-afb1ed00-4495-11e9-88bd-49314a7d4c8c.png">

Maybe add something like "(without the 'v')" after these customized badges?

Or if you want to keep those bullet points the way they're written, use a custom XML badge:

https://img.shields.io/badge/dynamic/xml.svg?label=&color=blue&query=%2F%2Fmetadata%2Fversioning%2Fversions%2Fversion&url=https%3A%2F%2Frepo.spring.io%2Fmilestone%2Fio%2Fprojectreactor%2Ftools%2Fblockhound%2Fmaven-metadata.xml
![](https://img.shields.io/badge/dynamic/xml.svg?label=&color=blue&query=%2F%2Fmetadata%2Fversioning%2Fversions%2Fversion&url=https%3A%2F%2Frepo.spring.io%2Fmilestone%2Fio%2Fprojectreactor%2Ftools%2Fblockhound%2Fmaven-metadata.xml)

bsideup:
Hi @paulmelnikow,

Wow, thank you, I didn't know there is such feature! I will use it for now, but would still appreciate the configurable prefix in future ☺️

paulmelnikow:
[...]

Issue #1181, opened 2017-10-18
Leading "v" is forcefully added in github release badges 


tooomm:
Example: ![](https://img.shields.io/github/release/Cockatrice/Cockatrice.svg)
It's a badge for [this](https://github.com/cockatrice/cockatrice/releases/tag/2017-05-05-Release-2.3.17) release on GitHub.
But the release tag there is `2017-05-05-Release-2.3.17` without `v`.

I understand that if you use straight semantic versioning for tags you are mostly ok.
Though, there are different recommendations on the usage of "v1.0" or "1.0".

paulmelnikow:
Yea, that doesn't look great.

Some options:

1. We could also add a flag to control whether the v prefix is there or not.
2. We could leave off the v all the time.
3. We could write a function which detects versions which look strange with a v prefix, e.g. which start with a date.

There was a discussion years ago about 2, and the v seemed slightly preferable; don't feel a need to change that now. Since our aim is to provide consistent badges to users, I like to avoid configuration where possible, so I'd lean toward the last option. If we identify other formats which are less legible with the v prefix, we can easily add them. Option 3 is also easier to implement.

tooomm:
>We could also add a flag to control whether the v prefix is there or not.

I like that a lot.
I understand that more options could lead to less consistency, but anybody could just start using a `v2.3.4` tag over `2.3.4` to archive the same... if you consider the v showing or hiding as inconsistency.
Or do you remove any leading `v` from tags (or check before inserting one) to prevent `release | vv2.3.4`?
As long as there is a solid default without flag it shouldn't be too bad at all?

>We could write a function which detects versions which look strange with a v prefix, e.g. which start with a date.

Also a consideration. As a start, check could be if "tag = valid x.y.z semver" and add a "v" in that case. Otherwise don't?

paulmelnikow:
> As long as there is a solid default without flag it shouldn't be too bad at all?

That's our aim.

I wouldn't want to push projects to change their tags. Since tags with and without `v`'s are both in common use, and supported by many tools, Shields' job is not to encourage one convention or another. Rather we should paper over the inconsistencies between projects.

> As a start, check could be if "tag = valid x.y.z semver" and add a "v" in that case. Otherwise don't?

[...]

PR #6988, opened 2021-09-06
Add API-based support for [GitLab] badges, add new GitLab Tag badge

### LINKED ISSUE #541: Badge request: Gitlab CI and Gitlab

calebcartwright:
Closes #541, refs/implements the tag half of #4867. I've opted to use the close keyword on #541 as I feel that with the authenticated supported added here, and accompanying rate limit increase, we can now move forward with other, specific badge requests as needed following our standard badge intake flow.

Have already setup the account and token, will plug them into the prod env before deploying and will also share the relevant account info with the rest of the ops team.


services/gitlab/gitlab-base.js
```diff
@@ -0,0 +1,19 @@
+import { BaseJsonService } from '../index.js'
+
+export default class GitLabBase extends BaseJsonService {
+  static auth = {
+    passKey: 'gitlab_token',
+    serviceKey: 'gitlab',
```
calebcartwright, line 6: We'll constrain the allowed origins for Shields.io to `gitlab.com`, but need this to be flexible for other self-hosters which is why the key is used here instead of hardcoding the `authorizedOrigins` value to `gitlab.com`

doc/server-secrets.md
```diff
@@ -147,6 +147,15 @@ These settings are used by shields.io for GitHub OAuth app authorization
 but will not be necessary for most self-hosted installations. See
 [production-hosting.md](./production-hosting.md).
 
+### GitLab
+
+- `GITLAB_ORIGINS` (yml: `public.services.drone.authorizedOrigins`)
+- `GITLAB_TOKEN` (yml: `private.drone_token`)
```
chris48s, line 153: Minor copy & paste errors in the docs here (drone)

services/gitlab/gitlab-tag.service.js
```diff
@@ -0,0 +1,62 @@
...
+
+const schema = Joi.array().items(
+  Joi.object({
+    name: Joi.string().required(),
+  })
+)
+
+const queryParamSchema = Joi.object({
+  gitlab_url: optionalUrl,
+}).required()
```
chris48s, line 14: The `gitlab_url` param isn't showing up in the builder/builder modal in the front-end. I think we need a second example for the case where `gitlab_url` is set to enable that.

services/gitlab/gitlab-tag.service.js
```diff
@@ -0,0 +1,62 @@
...
+        404: 'repo not found',
+      },
+    })
+  }
+
+  async handle({ user, repo }, { gitlab_url: baseUrl = 'https://gitlab.com' }) {
+    const tags = await this.fetch({ user, repo, baseUrl })
+    if (tags.length === 0)
+      throw new NotFound({ prettyMessage: 'no tags found' })
+    const { name } = tags[0]
```
chris48s, line 59: This is not the latest tag by date. From the docs https://docs.gitlab.com/ee/api/tags.html

> Get a list of repository tags from a project, sorted by name in reverse alphabetical order

Not checked, but I think this is basically the same situation we've got with the GitHub API so we can maybe copy the sort by date logic from there?

services/gitlab/gitlab-tag.spec.js
```diff
@@ -0,0 +1,47 @@
...
+            authorizedOrigins: ['https://gitlab.com'],
+          },
+        },
+      },
+      private: {
+        gitlab_token: fakeToken,
+      },
+    }
+
+    it('sends the auth information as configured', async function () {
```
chris48s, line 24: My reading of the code here is that we'll send our token to gitlab.com (or whatever is in `GITLAB_ORIGINS`), but you can still specify a custom `gitlab_url` (that is not in `GITLAB_ORIGINS`) and it will work (e.g: https://shields-staging-pr-6988.herokuapp.com/gitlab/tag/GNOME/pango?gitlab_url=https%3A%2F%2Fgitlab.gnome.org does render a badge) but it just won't send the token for those requests. Two follow-ups for that:
1. Can we add a second test case here that explicitly verifies our token is not sent when we specify a custom `gitlab_url` (and the url is not in `GITLAB_ORIGINS`)
2. Do we think the anonymous rate limit becomes a problem for alternate gitlab installs? Presumably self-hosters can configure this differently to the centralised service.. Do you think it is necessary to allow users to also pass a `&token=` param? Maybe we can just ignore this to start with it and deal with if it becomes a problem. I'm just trying to think ahead a bit..

services/gitlab/gitlab-tag.tester.js
```diff
@@ -0,0 +1,15 @@
...
+  .get('/fdroid/fdroidclient.json')
+  .expectBadge({ label: 'tag', message: Joi.string(), color: 'blue' })
+
+t.create('Tag (repo not found)')
+  .get('/fdroid/nonexistant.json')
+  .expectBadge({ label: 'tag', message: 'repo not found' })
+
+t.create('Tag (no tags)')
+  .get('/fdroid/fdroiddata.json')
+  .expectBadge({ label: 'tag', message: 'no tags found' })
```
chris48s, line 15: Can we add another test for the custom `gitlab_url` case here too please.

chris48s (commented):
Nice work on tackling this. One question here. The rest are inline on the diff.

> Have already setup the account and token, will plug them into the prod env before deploying

Do we also need to set a token for staging and/or CI or will the anonymous rate limit be sufficient for those environments?

calebcartwright:
>  Do we also need to set a token for staging and/or CI or will the anonymous rate limit be sufficient for those environments?

I was thinking anonymous will be fine. With a 2k per minute limit I don't anticipate any issues in staging nor CI

services/gitlab/gitlab-tag.service.js
calebcartwright: I haven't actually really looked at any part of the tag implementation, just grabbed the commits from #6389 as had been discussed in #541. Will need take a closer look

chris48s:
> 2k per minute

Right, yeah that is pretty generous and I think that should cover the custom `gitlab_url` case for most users :+1:

services/gitlab/gitlab-tag.spec.js
calebcartwright: > Can we add a second test case here that explicitly verifies our token is not sent when we specify a custom gitlab_url (and the url is not in GITLAB_ORIGINS)

I pulled in this test case because we have one such case for all other authenticated services. I guess I'm not strongly opposed but we don't have it for any of the similar services which I imagine is because the negative case is sufficiently covered in the auth helper tests, such as:

https://github.com/badges/shields/blob/f4bddb9964247a8d83803b3a59c13454e33c9cc6/core/base-service/auth-helper.spec.js#L344-L363

> Do we think the anonymous rate limit becomes a problem for alternate gitlab installs? Presumably self-hosters can configure this differently to the centralised service.. Do you think it is necessary to allow users to also pass a &token= param?

My reading of the GitLab docs on throttling is that it is IP based, so I don't think allowing users to provide a token nor even a hypothetical GitHub-like token pooling solution would be of any help for us on the rate limiting front.

GitLab does support scoped access tokens, including the ability to scope in a strictly read-only way. As such I think allowing users to provide their own token could help with cases of badges for private projects via Shields.io without having to self-host the badge server. However, I suspect that will pose a technical problem with our current auth injection mechanisms, since we'd need to pivot to a "don't worry about the authorized origins if it's a user provided token, but still enforce authorized origins with the configured token if one wasn't provided". Presumably a problem that can be overcome, but probably best done in a separate dedicated thread.

I don't think we'll have the ability to readily support authenticated requests for alternate instances without another type of non-trivial change to our auth story, as we'd need the ability to define multiple mappings of an auth mechanism to a set of authorized origins, whereas today it's strictly a 1 secret/token/creds to many origins.

services/gitlab/gitlab-tag.service.js
calebcartwright: I feel like their documentation is somewhat contradictory :thinking: The textual description of the API shared above is rather unequivocal about the sort being alphabetical by name, but then the param documentation just a couple lines below says:

> order_by | string | no | Return tags ordered by name or updated fields. Default is updated

I'm going to need to setup a test repo to try to vet it out because the handful of real repositories I was trying to spot check with were inconclusive due to both means producing the same ordering

calebcartwright: All signs point to the default sort indeed being by date, despite what the API description says.

https://gitlab.com/api/v4/projects/shields-ops-group%2Ftag-test/repository/tags

<details><summary>API response</summary>

```json
[
  {
    "name": "v2.0.0",
    "message": "",
    "target": "ae4abdc07357a1d26d3c692a9ad3118b6b31af2a",
    "commit": {
      "id": "ae4abdc07357a1d26d3c692a9ad3118b6b31af2a",
      "short_id": "ae4abdc0",
      "created_at": "2021-09-11T11:20:31.000-05:00",
      "parent_ids": [
        "396bf1debee4d196f85ae7f3d00ce2a4a71346a3"
      ],
      "title": "foo",
      "message": "foo\n",
      "author_name": "Caleb Cartwright",
      "author_email": "caleb.cartwright@outlook.com",
      "authored_date": "2021-09-11T11:20:31.000-05:00",
      "committer_name": "Caleb Cartwright",
      "committer_email": "caleb.cartwright@outlook.com",
      "committed_date": "2021-09-11T11:20:31.000-05:00",
      "trailers": {},
      "web_url": "https://gitlab.com/shields-ops-group/tag-test/-/commit/ae4abdc07357a1d26d3c692a9ad3118b6b31af2a"
    },
    "release": null,
    "protected": false
  },
  {
    "name": "v4.0.0",
    "message": "",
    "target": "396bf1debee4d196f85ae7f3d00ce2a4a71346a3",
    "commit": {
      "id": "396bf1debee4d196f85ae7f3d00ce2a4a71346a3",
      "short_id": "396bf1de",
      "created_at": "2021-09-11T11:16:33.000-05:00",
      "parent_ids": [
        "bf047705d78e939fcd7ff5256014c7fcee954a6f"
      ],
      "title": "another update",
      "message": "another update\n",
      "author_name": "Caleb Cartwright",
      "author_email": "caleb.cartwright@outlook.com",
      "authored_date": "2021-09-11T11:16:33.000-05:00",
      "committer_name": "Caleb Cartwright",
      "committer_email": "caleb.cartwright@outlook.com",
      "committed_date": "2021-09-11T11:16:33.000-05:00",
      "trailers": {},
      "web_url": "https://gitlab.com/shields-ops-group/tag-test/-/commit/396bf1debee4d196f85ae7f3d00ce2a4a71346a3"
    },
    "release": null,
    "protected": false
  },
  {
    "name": "v3.0.0",
    "message": "",
    "target": "bf047705d78e939fcd7ff5256014c7fcee954a6f",
    "commit": {
      "id": "bf047705d78e939fcd7ff5256014c7fcee954a6f",
      "short_id": "bf047705",
      "created_at": "2021-09-11T11:11:50.000-05:00",
      "parent_ids": [
        "490f9e3d62ac21d5f64d53b6b8d607d0fd9792d1"
      ],
      "title": "more updates",
      "message": "more updates\n",
      "author_name": "Caleb Cartwright",
      "author_email": "caleb.cartwright@outlook.com",
      "authored_date": "2021-09-11T11:11:50.000-05:00",
      "committer_name": "Caleb Cartwright",
      "committer_email": "caleb.cartwright@outlook.com",
      "committed_date": "2021-09-11T11:11:50.000-05:00",
      "trailers": {},
      "web_url": "https://gitlab.com/shields-ops-group/tag-test/-/commit/bf047705d78e939fcd7ff5256014c7fcee954a6f"
    },
    "release": null,
    "protected": false
  },
  {
    "name": "v1.0.0",
    "message": "",
    "target": "490f9e3d62ac21d5f64d53b6b8d607d0fd9792d1",
    "commit": {
      "id": "490f9e3d62ac21d5f64d53b6b8d607d0fd9792d1",
      "short_id": "490f9e3d",
      "created_at": "2021-09-11T11:10:09.000-05:00",
      "parent_ids": [
        "922ab7ddafd8b7892cfa881f6d67cdde48f59ed1"
      ],
      "title": "update",
      "message": "update\n",
      "author_name": "Caleb Cartwright",
      "author_email": "caleb.cartwright@outlook.com",
      "authored_date": "2021-09-11T11:10:09.000-05:00",
      "committer_name": "Caleb Cartwright",
      "committer_email": "caleb.cartwright@outlook.com",
      "committed_date": "2021-09-11T11:10:09.000-05:00",
      "trailers": {},
      "web_url": "https://gitlab.com/shields-ops-group/tag-test/-/commit/490f9e3d62ac21d5f64d53b6b8d607d0fd9792d1"
    },
    "release": null,
    "protected": false
  }
]
```

</details>

doc/server-secrets.md
calebcartwright: Fixed in ee81adcc8bce2f271d444ed6f144c44f9dff6755

services/gitlab/gitlab-tag.service.js
calebcartwright: Fixed in 8d5cfe16691522fa703bafa16e90d0a5c3cb797f

services/gitlab/gitlab-tag.service.js
calebcartwright: Feel like this has been sufficiently addressed in b96f597e92c1de7a9857cee5f8720c4533904382 so going to resolve, but let me know if you think there's still any outstanding items on this topic

services/gitlab/gitlab-tag.tester.js
calebcartwright: Addressed in b96f597e92c1de7a9857cee5f8720c4533904382

services/gitlab/gitlab-tag.service.js
```diff
@@ -0,0 +1,130 @@
...
+  ]
+
+  static defaultBadgeData = { label: 'tag' }
+
+  static render({ version, sort }) {
+    return {
+      message: addv(version),
+      color: sort === 'semver' ? versionColor(version) : 'blue',
+    }
+  }
```
calebcartwright, line 83: This is mostly copy/pasta from the corresponding GitHub service for consistency, but probably worth a moment of discussion given the change (the introduction of the `v` prefix). 

There's been recurring discussion on this in general, and we certainly want to keep our standard posture (refs #3200, #1181, and #6135). However, we have previously expressed an openness to consider no longer inserting the `v` prefix on the GitHub Tag badges (e.g. https://github.com/badges/shields/issues/3200#issuecomment-535578388).

Accordingly I actually think the message should be adjusted here to just display `version`, sans the `addv` call, but wanted to see what others thought first

services/gitlab/gitlab-tag.spec.js
chris48s: > we don't have it for any of the similar services which I imagine is because the negative case is sufficiently covered in the auth helper tests

OK yeah I think its covered by the tests for the base class.

> I don't think we'll have the ability to readily support authenticated requests for alternate instances without another type of non-trivial change to our auth story, as we'd need the ability to define multiple mappings of an auth mechanism to a set of authorized origins, whereas today it's strictly a 1 secret/token/creds to many origins.

Useful to have thought it through a bit, but I think we can just ignore it for now and cross this bridge if it ever becomes an issue. I think the anon rate limit should cover most custom server cases anyway.

services/gitlab/gitlab-tag.service.js
chris48s: Odd that the docs are wrong, but I agree your assessment of the real world behaviour :+1:

services/gitlab/gitlab-tag.service.js
```diff
@@ -0,0 +1,131 @@
...
+
+  static examples = [
+    {
+      title: 'GitLab tag (latest by date)',
+      namedParams: {
+        user: 'shields-ops-group',
+        repo: 'tag-test',
+      },
+      queryParams: { sort: 'semver' },
```
chris48s, line 37: This example doesn't match the title. This should be `queryParams: { sort: 'date' }` or just nothing

services/gitlab/gitlab-tag.service.js
```diff
@@ -0,0 +1,131 @@
...
+      namedParams: {
+        user: 'GNOME',
+        repo: 'librsvg',
+      },
+      queryParams: {
+        sort: 'semver',
+        include_prereleases: null,
+        gitlab_url: 'https://gitlab.gnome.org',
+      },
+      staticPreview: this.render({ name: 'v2.51.4' }),
```
chris48s, line 72: The objects we're passing to `render()` in the examples here don't include the right keys so all the examples are showing as ` tag | undefined`

services/gitlab/gitlab-tag.service.js
chris48s: I do see where you're coming from and this is a good point to raise.

Our starting point for this is that (pretty much?) all of our badges add the v prefix by default. If we were to decide to allow users to disable it, the pattern is going to be that everything appends v by default but then you can pass `?addv=false` or whatever to disable it.

Given that, I think it probably makes most sense to set GitLab up the same way as everything else to start with (i.e: with `addv()`). If we don't we will either end up saying "all the version badges append v by default and you can pass `?addv=false`, except GitLab which is the other way round" or we have to make a behaviour change for GitLab in future to make it consistent.

calebcartwright: >  If we were to decide to allow users to disable it, the pattern is going to be that everything appends v by default but then you can pass ?addv=false or whatever to disable it.

to clarify, i'm not proposing we expose this level of control to the users via a query param or other means. it's strictly a question of what Shields does internally, either always injecting a leading `v` or not. I don't think this is something we can change our minds on down the road, or at least it's not something I think we should bank on having the liberty to do, so I feel like we should be confident and comfortable with whatever option we select.

in hindsight the tag badge is somewhat of a challenge to incorporate as the example part of what is largely centered around the general auth story for gitlab, given all the associated nuances

services/gitlab/gitlab-tag.service.js
```diff
@@ -0,0 +1,131 @@
...
+  gitlab_url: optionalUrl,
+  include_prereleases: Joi.equal(''),
+  sort: Joi.string().valid('date', 'semver').default('date'),
+}).required()
+
+export default class GitlabTag extends GitLabBase {
+  static category = 'version'
+
+  static route = {
+    base: 'gitlab/v/tag',
```
calebcartwright, line 25: on the recurring theme of consistency, both with github and at large, i've updated the route to match the github tag badge route. however, i realize i'm not completely convinced this aligns with the `service/noun/parameters` badge route pattern enumerated in our docs.

we've also got other github routes that have the `v` and `noun` order reversed, e.g. `github/v/tag/...` vs `github/package-json/v/...`

i think this is probably the best bet to provide the highest degree of consistency, but also feel like maybe the docs could use more clarity if this is a common exception to the main rule?

services/gitlab/gitlab-tag.service.js
calebcartwright: ugh. good catch, sorry for the noise on this one. too many moving parts and keep missing little things

services/gitlab/gitlab-tag.service.js
chris48s: Agree with your decision here :+1:

services/gitlab/gitlab-tag.service.js
chris48s: Yes I think once we pick an approach here we're basically stuck with it. There isn't a single unambiguously correct choice. I think I'm still in favour of `addv()` just because it is what we do on all the other version badges (which this is https://github.com/badges/shields/pull/6988/files#diff-36f70802cce1e0dcc5c595f7913c44e5bce18d58c8357d8567b582399c7fcdb1R22 ) and it is what we'd be doing if we were using `renderVersionBadge()` but if you (or other @badges/shields-maintainers ) reckon there's a strong case for it here I'm not wedded to that opinion. Tags are slightly different and as I say, neither option is unambiguously correct.

chris48s (dismissed):
Pressing the big green tick on this as if we stick with `addv()` it is good to go, but happy to re-:+1: it if we change our minds on https://github.com/badges/shields/pull/6988#discussion_r706641452

services/gitlab/gitlab-tag.service.js
PyvesB: I would tend to agree with the argument "let's be consistent with what the GitHub badges do".

calebcartwright: Fair enough, I can go with that. One notable piece is that at past ops meetings we've discussed the need to brand our new GitLab badges with somewhat of a "beta" tag given the outstanding uncertainty around the auth story once we start to increase the scale of badges we offer and serve at runtime.

I'll post a note back in the main issue to that effect, which could in turn give us some leeway to revisit the message on this badge too.

calebcartwright:
Nuts, should've just let ranger add a merge commit but did a rebase instead. One last call for a :+1: whenever anyone gets a chance :smile:

chris48s (approved):

Issue #5862, opened 2020-11-21
Badge request: OpenBuildService


sp1ritCS:
:clipboard: **Description**

The OpenBuildService is the free Linux packaging platform from openSUSE.
For my usecase, I'd only need the build status from  one package in one repository, but it has a lot more api endpoints than that.

Example badge: https://img.shields.io/badge/openSUSE%20Tumbleweed-Succeeded-success

:link: **Data**

* API Doc: https://build.opensuse.org/apidocs/index
* The public buildservice instance: https://build.opensuse.org/
* Example status request: https://build.opensuse.org/build/hardware:razer/Debian_10/x86_64/openrazer/_status 

:microphone: **Motivation**

My usecase would be to show the build status of a package inside of a readme on github.
And if I click on it, it redirects to the download page of the package (https://software.opensuse.org/download/package?package=openrazer-meta&project=hardware%3Arazer).

chris48s:
All requests to this API must use authentication. The example call you posted is a `401`.
Does authenticating only grant access to your own packages, or can a logged in user read other users packages? If logging in only allows access to your own packages, this probably isn't a service we could work with as credentials are provided via HTTP basic auth so the user would have to expose their login details.

sp1ritCS:
> Does authenticating only grant access to your own packages, or can a logged in user read other users packages? 

no, you can access the result of any public package with any credentials.

But, feel a bit stupid that I havn't tested this, as I've just tried to do requests inside a browser I was already signed in 😅.

chris48s:
> you can access the result of any public package with any credentials

In that case, it could work like bintray, symfony insight etc with a single set of shared credentials

sp1ritCS:
I've gathered a bit more info:

Request via curl: `curl -u sp1rit:mypassword "https://api.opensuse.org/build/hardware:razer/Debian_10/x86_64/openrazer/_status"`

Different fields: `curl -u ${_username}:${_password} "https://${_instance}/build/${project}/${repository}/${arch}/${package}/_status"` (fields with `_` should be set by the environment)

The request returns 
```xml
<status package="openrazer" code="succeeded">
  <details></details>
</status>
```
the only interesting part is the `code` key from `<status>`. It can have different values. Here is the part of the css from the obs webinterface, that lists all build states aswell as their respective color.
```css
.build-state-succeeded {
 color:#008000
}
.build-state-building {
 background-color:#069
}
.build-state-scheduled {
 background-color:#007a7c
}
.build-state-signing,
[...]

PR #6993, opened 2021-09-07
[OBS] add Open Build Service service-badge

sp1ritCS:
this is my attempt of implementing a package status badge, like I suggested in #5862.

However, I did not follow the badge guidelines entirely (repository names are put in the "label" part of the badge, which may conflict with the "not advertise" and lowercase rule of it). I also made, what can be considered a "dirty" hack, by overwriting authHelper._authorizedOrigins in the constructor (due to the nature of authentication, there can only be a OBS single instance that can be used).


sp1ritCS:
Never done this before, so I expected to fail some ci tests.

However I did add tests (`obs.tester.js`), so I don't know why it complains 🙃.

I also don't want to give circleci rw access to all my repositories ( + the likely inbox spam that'd follow if I have them access to my account), so I'd be great if some one could tell me why it's failing so bad :D.

calebcartwright:
> I also don't want to give circleci rw access to all my repositories

Thank you for sharing this perspective. I've been using Circle for so long in one project or another that I didn't even realize that was a requirement. We have had some casual conversations about CI and investigating GitHub Actions, and although I don't anticipate we'll be making changes (at least not in the immediate future), I do think this POV and impact on contributors is something we should keep in mind @badges/shields-ops 

> so I'd be great if some one could tell me why it's failing so bad :D

looks like all the failures stem back to this:

> TypeError: repository.replaceAll is not a function
    at renderBuildStatusBadge (file:///home/circleci/project/services/obs/obs-build-status.js:63:23)
    at Function.render (file:///home/circleci/project/services/obs/obs.service.js:44:12)
    at Function.<static_fields_initializer> (file:///home/circleci/project/services/obs/obs.service.js:33:27)
    at file:///home/circleci/project/services/obs/obs.service.js:11:41

You should be able to test this locally via `npm run test:services -- --only=obs` as well as `npm start` to make sure the server can properly startup

sp1ritCS:
>> TypeError: repository.replaceAll is not a function

it seems like that the replaceAll method of the String type was only introduced in node 15.0.0. That's why the service tester worked flawlessly on my node 16.6.2 machine. (however the main@node-16 test also seems to be failing on ci)

calebcartwright:
> > > TypeError: repository.replaceAll is not a function
> 
> it seems like that the replaceAll method of the String type was only introduced in node 15.0.0 (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility). That's why the service tester worked flawlessly on my node 16.6.2 machine. (however the main@node-16 test also seems to be failing on ci)

Linter is failing too:

> #!/bin/bash -eo pipefail
> npm run lint
>
>
> shields.io@0.0.0 lint
> eslint "**/*.@(js|ts|tsx)"
>
>
> /home/circleci/project/services/obs/obs.service.js
  17:3  error  Expected blank line between class members  lines-between-class-members
>
> ✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
>
>
> Exited with code exit status 1

sp1ritCS:
> Linter is failing too

ah, I didn't run that. Assumed that was part of the pre-commit hook

calebcartwright:
> Only authenticated users are allowed to access the Open Build Service API.

Do we know if this is just the case with the reference server, or is it the open build service product that enforces this? If it's the former, then the implementation could be much simpler, but if it's indeed the latter then I can see why you went the way you did.

sp1ritCS:
> Do we know if this is just the case with the reference server, or is it the open build service product that enforces this?

AFAIK every buildserivce instance enforces it. It seems to be part of [OBS documentation](https://github.com/openSUSE/open-build-service/blob/master/docs/api/api/api.txt) and testing [build.openeuler.org](curl "https://build.openeuler.org/build/openEuler:Mainline/standard_x86_64/x86_64/nginx") and [build.tizen.org](https://build.tizen.org/build/Tizen/standard/x86_64/gcc), both return
```xml
<status code="anonymous_user">
  <summary>Anonymous user is not allowed here - please login</summary>
</status>
```

calebcartwright:
Got it, thanks! After thinking on this one a bit, I actually don't think we need the complexity (or "hack" as you called it :smile:) given that the service is utilizing basic auth as opposed to instance specific tokens. In theory, anyone deploying a Shields server (including our main Shields.io) could setup an account with the same user/pass on multiple OBS instances, and could leverage the current auth config structure with a multi-entry value for `authorizedOrigins`.

That would warrant providing users the ability to specify the instance url via a query param, which is a very common pattern implemented in various other services. I think we'd probably still want to use a default url of the main reference instance, which we could either just bake in as the main reference instance (via default value on said query param), or if we really feel there's a need for self-hosters to be able to specify their own non-reference instance as the default then pivoting a bit on what you have here currently to make it just the _default_ instance, but don't use it to lock down/reset the authorized origins to that single instance.

Hope that makes sense, but let me know if you have any questions

sp1ritCS:
Yeah, I see what you mean. However I'm unsure how that hold's up security wise.

I'd rather have a
```js
obs: Joi.object().pattern(requiredUrl, Joi.object({
	user: Joi.string(),
	pass: Joi.string()
}))
```
in the private config that allows a user/pass pair for each instance.

However I'm unsure how that could work in regards to the auth-helper :)

calebcartwright:
> Yeah, I see what you mean. I like the idea of supporting multiple instances. However I'm unsure how using the same username and password holds up security wise.

Fundamentally I just don't want to introduce the extra code and deployment complexity that's primarily just to rule out such a use case. I understand the thrust of your point on the security, but I don't think general security practices around reusing credentials are much of a factor in this specific context.

For the Shields.io case, having our typical degree of flexibility could allow us to support more than one instance which would be useful especially if there's a couple widely used instances beyond the reference one. We treat the security and protection of the auth information used on Shields.io very seriously, and equal across the many that we have. Even in some worst case scenario, there's no substantial difference in the hypothetical blast radius of the exposure our non-active/real account info for both https://api.opensuse.org and https://build.openeuler.org, as opposed to just one.

For the self-hosting badge server, I don't anticipate the need for multiple instances being very common, but I also don't view it as our job to promote/prevent account strategies on a different tool either. Folks would always have the option of creating their own "badge reader" type of account that could have authorization managed appropriately within the instances. Finally, while admittedly the accounts on OBS seem local (though haven't looked to closely), it's somewhat common for self-hosting Shields users to interact with multiple instances of a tool that works with a backing IdP (e.g. Jira Server/DC) and have the ability to utilize a single, privileged account in the server auth config and the those valid instances being allowed in the `authorizedOrigins` config

> in the private config that allows a user/pass pair for each instance.

Absolutely! It's something we've been talking about for a while and there's plenty of use cases, likely including this one, where it'd be extremely helpful. However, that's not an easy left and is unlikely to be available soon, so we'll need to work with what's possible to do.

sp1ritCS:
Updated. I hope I haven't missed an easier method for authhelper to get the specified authorizedOrigins

services/obs/obs-build-status.js
```diff
@@ -0,0 +1,69 @@
...
+  failed: 'critical',
+  // Gray
+  disabled: 'inactive',
+  blocked: 'inactive',
+  // Orange
+  'scheduled-warning': 'important',
+  unknown: 'lightgray',
+}
+
+const isBuildStatus = Joi.equal(...Object.keys(statusmap))
```
calebcartwright, line 49: I appreciate the desire to try to replicate the design of the native colors/presentation used upstream. However, our goal and design constraints are centered around consistency with _Shields_ and our badge specification, and it's pretty much a non-goal to be consistent with colors/presentation defined in an upstream service.

To that end of providing consistency across our badges, we have both schema helpers (`isBuildStatus`) and render helpers (`renderBuildStatusBadge`) both of which can be imported `build-status.js` and used. It looks like a third of the status values we may see in the OBS response are already accounted for (`succeeded`, `building`, `scheduled`, and `failed`).

`broken` isn't a status we track in the shared helpers yet, but it should definitely be added to our `redStatuses` list because it's an objectively clear/obvious failure indicator. I gather from context that `unresolvable` is a clear error condition as well for this service, but that's something we'll want to handle local to this service class vs. adding to the shared/global lists (it could feasibly mean something different for a different type of build service/platform)

For `disabled`, `blocked`, and `unknown`, we'd want to qualify those as our standard "other" statuses, which we display with the inactive color by default/no color specification needed. The other three (`scheduled-warning`, `signing` and `finished`) are probably fine to treat as orange statuses, though again local to this service.

For the schema validation, let's use the standard `isBuildStatus` validator but as part of an alternative set to include the other 7 (4 already covered in that validator plus `broken` being added as a known/shared red status)

e.g.
```js
Joi.alternatives()
    .try(isBuildStatus, Joi.equal('unknown'))
    .required()
```

Not that's just shared as a reference, the other 7 don't have to be enumerated individually with `Joi.equal`.

For rendering, we also want to rely on the standard/shared `renderBuildStatusBadge` helper as much as possible. There's really only 4 statuses we need to override the color for though (the 3 orange statuses mentioned above plus `unresolvable` as an error color), but for the other 8 we should let `renderBuildStatusBadge` handle everything (it'll take car of setting the color correctly for all the inactive statuses too).

services/obs/obs.service.js
```diff
@@ -0,0 +1,94 @@
...
+      },
+      staticPreview: this.render({
+        repository: 'Debian_11',
+        status: 'succeeded',
+      }),
+      keywords: ['open build service'],
+    },
+  ]
+
+  static defaultBadgeData = { label: 'OBS' }
```
calebcartwright, line 46: All our badge defaults need to adhere to our [specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md) which includes not using the service/platform name in the label

```suggestion
  static defaultBadgeData = { label: 'build' }
```

services/obs/obs-build-status.js
```diff
@@ -0,0 +1,69 @@
...
+  )
+
+  return Joi.equal(...status)
+}
+
+function renderBuildStatusBadge({ repository, status }) {
+  const color = statusmap[status]
+
+  return {
+    label: repository.replace(/_/g, ' '),
```
calebcartwright, line 63: We shouldn't be changing this from the default value need for all build status badges. Users always have the ability to customize the label text via the `?label=foo` query param, but we shouldn't do something like that as the default behavior

```suggestion
```

services/obs/obs-build-status.js
```diff
@@ -0,0 +1,69 @@
...
+    message: status.charAt(0).toUpperCase() + status.slice(1),
```
calebcartwright, line 64: We use all lowercase text on our badges, so if there's any upper case characters in the response then we'd need to ensure the values we add to the textual fields are lower case

```suggestion
    message: status.toLowerCase(),
```

services/obs/obs-build-status.js
```diff
@@ -0,0 +1,69 @@
...
+    color,
+  }
+}
```
calebcartwright, line 67: Also as noted elsewhere we should pivot to relying on the shared helper, and only modify the color of the badge info provided by the `renderBuildStatusBadge` to account for the 1 error/red and 3 orange statuses that will be specific to this service

services/obs/obs.service.js
```diff
@@ -0,0 +1,94 @@
...
+        packageName: 'osc',
+        repository: 'Debian_11',
+        arch: 'x86_64',
+      },
+      staticPreview: this.render({
+        repository: 'Debian_11',
+        status: 'succeeded',
+      }),
+      keywords: ['open build service'],
+    },
```
calebcartwright, line 43: We should also add another example that includes the `queryParams` key so that there's an example listing on the main badge that allows users to leverage the custom instance url in the badge builder window. Note the actual value doesn't matter for `instance` and it's fine to use the reference server, just as long as there's an example with something like:

```js
queryParams: { instance: 'https://api.opensuse.org' },
```

calebcartwright (commented):
Thanks again for the PR! I've left a number of inline comments below in places where we'll need some changes. It may seem like a high quantity at first glance, but it's really just several smaller opportunities for simplification and some more minor updates to be more consistent with our typical build badges.

Let us know if you have any questions!

services/obs/obs.service.js
```diff
@@ -0,0 +1,94 @@
...
+    this.authHelper._authorizedOrigins = origins
+  }
+
+  async fetch({ instance, project, packageName, repository, arch }) {
+    return this._requestXml(
+      this.authHelper.withBasicAuth({
+        schema,
+        url: `${instance}/build/${project}/${repository}/${arch}/${packageName}/_status`,
+        errorMessages: {
+          401: 'Not authorized',
```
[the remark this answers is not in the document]
sp1ritCS: No, this is not the case. If I do make an unauthorized request it'll return HTTP 403.
```
< x-opensuse-errorcode: delete_file_no_permission
[..]
< status: 403 Forbidden
[...]
< 
<status code="delete_file_no_permission">
  <summary>Insufficient permissions to delete file</summary>
</status>
```

I guess it's irrelevant to this anyway as it doesn't do any authorized requests (PUSH, PUT, DELETE)

services/obs/obs.service.js
```diff
@@ -0,0 +1,94 @@
...
+          404: 'Package not found', // extend this to project/package/repository/arch?
```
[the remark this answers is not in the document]
sp1ritCS: it does return error information. can I use that somehow?

```xml
<status code="unknown_project">
  <summary>Project not found: openSUSE:dTools</summary>
</status>

---

<status code="unknown_package">
  <summary>Package not found: openSUSE:Tools/osd</summary>
</status>

---

<status code="404">
  <summary>project 'openSUSE:Tools' has no repository 'dDebian_11'</summary>
  <details>404 project 'openSUSE:Tools' has no repository 'dDebian_11'</details>
</status>

---

<status code="404">
  <summary>repository 'openSUSE:Tools/Debian_11' has no architecture 'x8d6_64'</summary>
  <details>404 repository 'openSUSE:Tools/Debian_11' has no architecture 'x8d6_64'</details>
</status>
```

or just remove it?

services/obs/obs.service.js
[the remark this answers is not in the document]
calebcartwright: I'm thinking that we also wouldn't ever make any unauthenticated requests either, so was wondering if there's actually a chance of seeing a 401 response, but it doesn't matter too much tbh

services/obs/obs.service.js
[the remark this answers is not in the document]
calebcartwright: Great question. At the moment I don't think we can leverage it since our internal request handling logic will bail when the response status code is a 404, and while it does allow service classes to control the message in the 404 case, it doesn't currently support any additional logic/inspection of the response. That could be an interesting future enhancement for us to consider, but for the purposes of this PR let's just use what's available today and either have the generic/encompassing error or remove the mapping and let the default "not found" apply.

If we do keep a custom message here though, let's be sure to make it all lower case

services/obs/obs.service.js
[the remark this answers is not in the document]
sp1ritCS: well, if the shields instance administrator fails to set the right credentials ._.

services/obs/obs.service.js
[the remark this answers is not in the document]
sp1ritCS: I think "not found" is fitting, so I've just the http status mapping block.

services/obs/obs.tester.js
```diff
@@ -0,0 +1,25 @@
...
+    message: isBuildStatus,
+  })
+
+t.create('status (invalid)')
+  .skipWhen(noToken(ObsService))
+  .get('/home:sp1rit/this_package_will_never_exist/repo/arch.json')
+  .expectBadge({
+    label: 'OBS',
+    message: 'Package not found',
+  })
```
calebcartwright, line 25: I've wired up the CI and review app environments with the relevant auth configuration (https://shields-staging-pr-6993.herokuapp.com/obs/openSUSE:Tools/osc/Debian_11/x86_64) so the tests should now be getting executed. Think these will need to be updated in order to pass, to account for some of the label and message changes to the services

sp1ritCS: ah, I forgot to update the invalid test 🤦🏻‍♂️ 

should be fixed now

calebcartwright (approved):
Awesome, thank you!

calebcartwright:
This is up and live now on https://shields.io, but the upstream API/reference server is very sluggish in its responses which means that a lot of the badges will exceed GitHub's timeout and they won't render

e.g.
https://img.shields.io/obs/openSUSE:Factory/aaa_base/standard/x86_64?instance=https%3A%2F%2Fapi.opensuse.org


https://img.shields.io/obs/openSUSE:Tools/osc/Debian_11/x86_64?instance=https%3A%2F%2Fapi.opensuse.org&label=debian


There's not much utility in us providing badges that aren't practically usable, so we'll need to keep an eye out and potentially need to revisit.

sp1ritCS:
hmm, it usually isn't, I also never had issues with it during testing.

However it seems like the reference instance has some issues right now, most of the times my browser times out before it loads; [status.opensuse.org](https://status.opensuse.org/) however tells "Operational". 

I'll ask on [buildservice@lists.opensuse.org](https://lists.opensuse.org/archives/list/buildservice@lists.opensuse.org/) if this is a known issue, however I don't expect an answer during the weekend.

sp1ritCS:
should work again. The badges load far faster for me now (almost instant :)) and I can access the webui again.

According to  Henne Vogelsang the issue was, that: 
> We had multiple spiders (bing, yandex, unicom, sprinthost.ru) crawling 
> 'expensive' routes like
>
> /package/view_file/...
>
> Every once in a while their frequencies align and then it becomes a bit 
> too much. But in general we want them to index things of course.
>
> But it wasn't *that* bad, 95% of all requests still where answered 
> faster than 256 milliseconds.

I guess you concern regarding
> There's not much utility in us providing badges that aren't practically usable

is therefore resolved?

calebcartwright:
> is therefore resolved?

Yes indeed! Thanks for following up. Obviously windows like this happen from time to time (including with our own service), but wanted to make sure that was an exception/issue driven behavior as opposed to the standard response times we could expect.

PR #7012, opened 2021-09-11
update branch in [githublernajson] test

chris48s:


calebcartwright (approved):
Seems like Facebook have been steadily making the switch from master to main across their repo portfolio

PyvesB (approved):

PR #7013, opened 2021-09-12
Stop attempting to override Accept header in [GitHub] API provider

### LINKED ISSUE #6960: GitHub Accept header is not set as intended

shields.io

### 🐞 Description

We attempt to set the `Accept` header for GitHub API calls here:
https://github.com/badges/shields/blob/be3fa207b5883eddfde1ffb5518bba806c6d3670/services/github/github-api-provider.js#L181

However, two lines below that line of code, we merge options that were passed in the `request` function call:
https://github.com/badges/shields/blob/be3fa207b5883eddfde1ffb5518bba806c6d3670/services/github/github-api-provider.js#L183

The passed in headers will contain the default of the base class (`application/json` for both `BaseGraphqlService` and `BaseJsonService`). In other words, `application/vnd.github.v3+json` is systematically overwritten.

For the GraphQL API, this bug is not a bad thing, as we wouldn't want to be setting an `application/vnd.github.v3+json` header anyway. For the other JSON endpoints, we're slightly deviating from GitHub's recommendations, and there's a theoretical chance that things could break if they changed the default version of the API in the future. Unsure how likely that is though, especially as they seem to be pushing the GraphQL API nowadays.

### 🔗 Link to the badge

_No response_

### 💡 Possible Solution

Definitely nothing major going on here, but probably best to avoid some head scratching in the future due to this confusing state of things.

We could either:
* override the defaults from `BaseJsonService` here: https://github.com/badges/shields/blob/master/services/github/github-auth-service.js. However, having them in _github-api-provider.js_ makes more sense than _github-auth-service.js_.
* remove the `application/vnd.github.v3+json` line entirely as it won't change from today's behaviour anyway. We can always revisit in the future if we have specific `Accept` header needs in the future. This may well be the pragmatic thing to do.

PyvesB:
Fixes  #6960.

@chris48s, is this the kind of test you had in mind?


PyvesB:
The test failure will be fixed by  #7012.

chris48s (approved):
Works for me. Tbh, I was thinking this would be more important in the case where we try to fix sending `application/vnd.github.v3+json`. It matters less if we just drop the `Accept` header but having a test for this will do us no harm :)

PR #7016, opened 2021-09-13
[freecodecamp]: allow + symbol in username

SethFalco:
### Related
* Fixes https://github.com/badges/shields/issues/6998

### Problem
The pattern I used in the `schema` instance in `freecodecamp-points.service.js` was wrong.

I used `\w` because I thought usernames could only have letters, numbers, hyphens and underscores. I wasn't aware they could have plus until now. Then `allowAndStripUnknownKeys` kicked off and removed the properties that didn't match the pattern defined, including usernames with a `+` in it.

### Solution
I stole the pattern used to validate usernames from the official freeCodeCamp repository this time, so it should match this time.

> ```js
> const validCharsRE = /^[a-zA-Z0-9\-_+]*$/;
> ```
> 
> — https://github.com/freeCodeCamp/freeCodeCamp/blob/main/utils/validate.js#L14

Now it handles `+` symbols correctly.

### DIFF
--- a/services/freecodecamp/freecodecamp-points.service.js
+++ b/services/freecodecamp/freecodecamp-points.service.js
@@ -2,11 +2,17 @@ import Joi from 'joi'
 import { metric } from '../text-formatters.js'
 import { BaseJsonService, InvalidResponse, NotFound } from '../index.js'
 
+/**
+ * Validates that the schema response is what we're expecting.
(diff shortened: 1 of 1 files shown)


calebcartwright (approved):
Was going to say it might be helpful to add a tests for this but I think the comment covers it sufficiently.

Thank you!

PR #7018, opened 2021-09-14
fix [SpigetDownloadSize] service test

calebcartwright:
Forgot to open an issue, but this test has been failing for a while. Something changed upstream with the target we were using for the non-externally hosted resource scenario, and that target now is externally hosted (don't know enough about the service to know what that actually means, but was a straightforward root cause analysis).

### DIFF
--- a/services/test-validators.js
+++ b/services/test-validators.js
@@ -93,7 +93,9 @@ const isPercentage = Joi.alternatives().try(
   isDecimalPercentage
 )
 
-const isFileSize = withRegex(/^[0-9]*[.]?[0-9]+\s(B|kB|MB|GB|TB|PB|EB|ZB|YB)$/)
+const isFileSize = withRegex(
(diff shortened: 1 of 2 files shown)


services/test-validators.js
```diff
@@ -93,7 +93,9 @@ const isPercentage = Joi.alternatives().try(
   isDecimalPercentage
 )
 
-const isFileSize = withRegex(/^[0-9]*[.]?[0-9]+\s(B|kB|MB|GB|TB|PB|EB|ZB|YB)$/)
+const isFileSize = withRegex(
+  /^[0-9]*[.]?[0-9]+\s(B|kB|KB|MB|GB|TB|PB|EB|ZB|YB)$/
+)
```
calebcartwright, line 98: had to account for some pedantry here since the random target I grabbed had a file size in the kilobytes, but Spiget deals with base 2 as opposed to base 10, and in base 2 the leading `K` for kilobytes is capitalized

chris48s (approved):

PR #7021, opened 2021-09-15
add GitLab Release badge, run all [GitLab]

### LINKED ISSUE #4867: GitLab tags and releases

calebcartwright:
Resolves #4867

### DIFF
--- a/services/gitlab/gitlab-base.js
+++ b/services/gitlab/gitlab-base.js
@@ -16,4 +16,50 @@ export default class GitLabBase extends BaseJsonService {
       })
     )
   }
+
+  async fetchPage({ page, requestParams, schema }) {
+    const { res, buffer } = await this._request({
+      ...requestParams,
+      ...{ options: { qs: { page } } },
+    })
+
+    const json = this._parseJson(buffer)
+    const data = this.constructor._validate(json, schema)
+    return { res, data }
+  }
+
+  async fetchPaginatedArrayData({
+    url,
+    options,
+    schema,
+    errorMessages,
+    firstPageOnly = false,
+  }) {
+    const requestParams = this.authHelper.withBasicAuth({
+      url,
+      options: {
+        headers: { Accept: 'application/json' },
+        qs: { per_page: 100 },
+        ...options,
+      },
+      errorMessages,
+    })
+
+    const {
+      res: { headers },
+      data,
+    } = await this.fetchPage({ page: 1, requestParams, schema })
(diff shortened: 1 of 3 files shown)


services/gitlab/gitlab-base.js
```diff
@@ -16,4 +16,50 @@ export default class GitLabBase extends BaseJsonService {
...
+      return data
+    }
+
+    const pageData = await Promise.all(
+      [...Array(numberOfPages - 1).keys()].map((_, i) =>
+        this.fetchPage({ page: ++i + 1, requestParams, schema })
+      )
+    )
+    return [...data].concat(...pageData)
+  }
```
calebcartwright, line 64: GitLab's max per page value is 100, and I saw that exceeded in several cases just with the releases on the handful of projects I tested with. A cursory glance through the GitLab API docs makes it seem a very common response structure will be pages of arrays with the page keys in the response headers.

Opted to go ahead and pull that into the base class in anticipation of other APIs. Added it with relevant short circuiting logic (since GitLab doesn't seem to have a separate "latest" API endpoint that returns a single object like GitHub does in some cases), and also requesting all the pages in parallel since it's both possible to do so and preferable to 1 by 1

services/gitlab/gitlab-release.service.js
```diff
@@ -0,0 +1,136 @@
...
+    name: Joi.string().required(),
+    tag_name: Joi.string().required(),
+  })
+)
+
+const queryParamSchema = Joi.object({
+  gitlab_url: optionalUrl,
+  include_prereleases: Joi.equal(''),
+  sort: Joi.string().valid('date', 'semver').default('date'),
+  display_name: Joi.string().valid('tag', 'release').default('tag'),
```
calebcartwright, line 18: Figured it made sense to add this from day 0 given where we ended up with this on GitHub. I've opted to set the default to `tag` to be consistent with what GitHub currently is but don't have strong opinions on this one way or another.

calebcartwright:
Blocked on a decision/resolution of #7167

PR #7023, opened 2021-09-16
support gitlab token via env var

calebcartwright:
Was banging my head against the desk trying to figure out why the Shields.io prod environment wasn't sending the auth info with gitlab.com bound requests... only to discover I left out this crucial piece :man_facepalming: 

I've been meaning to suggest some documentation for adding new server secrets, as I think it'd be a helpful reference for folks, plus I've clearly forgotten it myself on more than one occasion!


chris48s (approved):

chris48s:
Yeah docs on this are a good shout :+1:  I think a new config var usually needs adding to

- config/custom-environment-variables.yml
- config/default.yml
- core/server/server.js (schema)
- relevant docs (usually doc/server-secrets.md )

so there's definitely some scope to mess that up

PR #7024, opened 2021-09-16
Doc self hosting correction

cnico:
Hi, 
I have tried to run shields following the self-hosting documentation but it was not working for configuration values (like the http port or other). After having retry many times, I finally find the reason in this discussion : https://github.com/badges/shields/discussions/5773

I think, the documentation did not follow the code since the doc was still referencing shields.env file which does not work any more...

So this pull request is to improve the documentation by explaining the 2 methods that I confirm work successfully.

I let you review this and merge if you think it is correctly explained. 
Sorry for my English that is not perfect (I am a french guy :) ).

### DIFF
--- a/doc/self-hosting.md
+++ b/doc/self-hosting.md
@@ -94,20 +94,27 @@ Sending build context to Docker daemon 3.923 MB
 Successfully built 4471b442c220
 ```
 
-Optionally, create a file called `shields.env` that contains the needed
-configuration. See [server-secrets.md](server-secrets.md) and [config/custom-environment-variables.yml](/config/custom-environment-variables.yml) for examples.
+Optionally, alter the default values for configuration via one of the two methods : through a configuration file (ex : `local.yml`) or through environnement variables.
+See [server-secrets.md](server-secrets.md) and [config/custom-environment-variables.yml](/config/custom-environment-variables.yml) for possible values.
+In [config/custom-environment-variables.yml](/config/custom-environment-variables.yml), environnement variable keys are defined in the quotes.
+
+```docker
+FROM shieldsio/shields:next
+COPY local.yml ./config/local.yml
+```
 
 Then run the container:
 
 ```console
-$ docker run --rm -p 8080:80 --name shields shields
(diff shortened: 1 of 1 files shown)


calebcartwright:
Haven't had a chance to go over this in detail, but I certainly agree with you that some changes are needed here to reflect the .env piece. Unfortunately we sometimes miss things in docs when we make changes :smile: 

It looks like there's a formatting issue in the updated files, could you run the formatter (`npm run prettier`) and push the update to fix CI whenever you get a chance?

cnico:
@calebcartwright, I did not look at the CI result considering there where no check for documentation.
I corrected it : one empty line has been removed in the commit.

I let you merge this and continue your work on shields which is a good product.

calebcartwright:
Apologies for the delay on this. As I mentioned before I definitely agree with you that something should change here, but I don't want to go with the actual proposals you've suggested here.

When users are going the Docker route for self-hosting, I would prefer to encourage the following options for specifying config values:

- Use Docker's ability to set environment variables in the container context, either via the individual `-e ...` args or if a file is preferred via `--env-file`
- Use Docker compose
- Use a volume mount

Obviously there's more considerations if an orchestrator like Kubernetes will be utilized, but that's really outside the scope of Shields and not something I want to cover in our documentation.

I generally dislike hardcoding any sensitive information into the Docker image because there's always too great a risk someone forgets down the road and that image is handed around exposing sensitive information. As such, I wouldn't want our documentation to encourage users to create a docker image with hardcoded secrets in config/local.yml, regardless of whether that's done by building the image directly from source control, or using our published images as the base as you've initially suggested here.

If you're interested and willing in updating this PR to change the text accordingly that would be great, but if not that's perfectly okay too, it was definitely helpful that you flagged this issue in the docs!

cnico:
Hi @calebcartwright,

OK, I corrected the proposal not to include the solution with local config yml file and leave only env configurations which actually works with docker and with kubernetes (this is my final usage).

I let you merge this if it is OK for you.

PR #7037, opened 2021-09-19
support GitHub OAuth token scopes, run all [GitHub]

calebcartwright:
This would bring us to a resolution on #4169 (intentionally not using any close keywords just yet).

Two minor things to note on the PR, _tried_ to structure the commits in a way to facilitate review, and the diff is deceptively large. The actual app code change is ~70 lines, but I opted to add a lot of tests given the nature of the change.

I had originally intended to just take the first minor step of updating our requested scopes, but quickly remembered the associated challenge that we'd discussed elsewhere around having a pool of tokens with unknown scopes. As such I opted to go ahead and step through the looking glass and propose a full implementation. While I fully anticipate some dialog/additional changes needed/etc., I do think this has the potential to be a viable path, but at a minimum hope this at least gets the ball rolling.

As to the implementation :smile: Here's some things I considered, but decided against along with my thinking...

There's a high level binary decision at the top of whether or not we should _persist_ the corresponding scopes with each token vs. dynamically checking scopes on the fly. I decided against trying to dynamically check scopes because that'd require we either:

* Check scopes on the respective current token prior to using it for an API request
* Check scopes on all tokens at startup

Both of those options seem unacceptably excessive, as in the latter case every time a dyno started each dyno would have to fire off 20k+ requests, and the former would inherently require a sequential set of API calls to satisfy every request.

Accordingly, I'm of the opinion we do need to persist the associated scopes along with each token.

Our current persistence strategy utilizes Redis with a single key that has a [Set-type value](https://redis.io/topics/data-types#sets) with that value containing all the tokens. One option would've been introducing new key(s)/set(s) pairs in Redis, where we'd have something like:

currentKey = (Set of unscoped tokens)
newKey = (Set of read:packages scoped tokens)

That would seemingly have provided a simpler story, at least on the persistence side, but would make for a notably more complicated world if (and almost certainly when) GitHub introduces some new feature that requires some new token scope. E.g. would we need and additional `newerKey` for that next scope, how would we persist tokens that have multiple scopes, etc.

As such, I believe it'll be easier and more maintainable/flexible if our persisted tokens carry their associated scope directly vs. trying to use collection-based keys to imply scopes.

The next question was really about _how_ to represent the respective scopes on each token in Redis. Redis doesn't really have a first class recognition of "objects", so presuming we don't want to experiment with radically different data types (none of which would offer any real utility for our context IMO) then it's a matter of incorporating the scope information on each token element in the set. One approach could have been more object-based representations in code of the raw persisted token material which is then json-stringified for persistence, however, I believe that adds unnecessary noise and verbosity to the set, and also introduces serialization concerns.

Instead, I'd propose we simply append the scopes to the end of the token string for each element in the set, using a delimiter. I've opted for a `.scopes.` delimiter, producing something like:

`abc123UnscopedTokenExample  another_unscoped_token   defscoped_example.scopes.read:packages  another_scoped.scopes.read:packages%20read:user`

The actor which handles loading up the tokens and dealing with new/invalidated ones (GithubConstellation) then incorporates the responsibility of taking those persisted forms and mapping them to the form used with our `Token` and `TokenPool` classes internally, leveraging the existing (but unused) `data` property on the `Token` to capture scopes. Internally, the constellation also tracks a dictionary of tokens for simplicity/to avoid having to do full scans of all tokens in all pools in order to handle re-authorized/re-scoped tokens.

The final piece is the separate pooling of scoped and standard unscoped tokens in the GithubApiProvider, with a mechanism for the actual Github service classes to define any scopes they required from their constructor, which in turn will ensure that the provider selects the next token from the correct token pool. Because the tokens are truly batched into separate pools now, I've set a cap on the portion of the total token pool that get funneled into the scoped pool. My thinking is that for the foreseeable future the vast majority of our ~21k will remain unscoped, and as such we'll probably need to ensure that the subset of tokens that _are_ scoped are only used for requests requiring scopes (so we don't eat through them when unnecessary). However, we'll conceivably/hopefully start to see a tide shift some ways down the road as the ratio of scoped tokens increases, and will also need to prevent pool starvation for the unscoped requests. I think the distinct set of pools with a slideable ratio will give us the flexibiltiy we need to handle both of those concerns.

Last note/observation is that I've added this in a way such that if a token is re-authorized the pools are adjusted accordingly, at least on that dyno, so that corresponding token can be used correctly without that dyno needing to restart/reload from Redis. I realize that might be overkill, especially since our dynos don't have any form of token sharing anymore and our dynos are restarted at least once a day anyway. This seemed like a sane/useful thing to do, but there's probably a few statements that could be removed if we decide that's unnecessary

### DIFF
--- a/services/github/github-api-provider.js
+++ b/services/github/github-api-provider.js
@@ -38,6 +38,7 @@ class GithubApiProvider {
     onTokenInvalidated = tokenString => {},
     globalToken,
     reserveFraction = 0.25,
+    tokenScopeNames = {},
   }) {
     Object.assign(this, {
       baseUrl,
@@ -45,12 +46,14 @@ class GithubApiProvider {
       onTokenInvalidated,
       globalToken,
       reserveFraction,
+      tokenScopeNames,
     })
 
     if (this.withPooling) {
       this.standardTokens = new TokenPool({ batchSize: 25 })
       this.searchTokens = new TokenPool({ batchSize: 5 })
       this.graphqlTokens = new TokenPool({ batchSize: 25 })
+      this.packageScopedTokens = new TokenPool({ batchSize: 25 })
     }
   }
 
@@ -60,17 +63,41 @@ class GithubApiProvider {
         standardTokens: this.standardTokens.serializeDebugInfo({ sanitize }),
         searchTokens: this.searchTokens.serializeDebugInfo({ sanitize }),
         graphqlTokens: this.graphqlTokens.serializeDebugInfo({ sanitize }),
+        packageScopedTokens: this.packageScopedTokens.serializeDebugInfo({
+          sanitize,
+        }),
       }
     } else {
       return {}
     }
   }
 
-  addToken(tokenString) {
+  numReservedScopedTokens() {
+    return this.packageScopedTokens.count()
+  }
+
+  addReservedScopedToken(tokenString, data) {
+    if (!this.withPooling) {
+      throw Error('When not using a token pool, do not provide tokens')
+    }
+
+    const { scopes } = data
+    if (!scopes) {
+      throw new Error('Cannot add unscoped token to reserved token pools')
+    }
+
+    scopes.split('%20').forEach(scope => {
+      if (scope === this.tokenScopeNames.readPackages) {
+        this.packageScopedTokens.add(tokenString, data)
+      }
+    })
+  }
+
+  addToken(tokenString, data) {
     if (this.withPooling) {
-      this.standardTokens.add(tokenString)
-      this.searchTokens.add(tokenString)
-      this.graphqlTokens.add(tokenString)
+      this.standardTokens.add(tokenString, data)
+      this.searchTokens.add(tokenString, data)
+      this.graphqlTokens.add(tokenString, data)
     } else {
       throw Error('When not using a token pool, do not provide tokens')
     }
@@ -141,7 +168,11 @@ class GithubApiProvider {
     this.onTokenInvalidated(token.id)
   }
 
-  tokenForUrl(url) {
+  tokenForUrl(url, { needsPackageScope }) {
+    if (needsPackageScope) {
+      return this.packageScopedTokens.next()
+    }
+
     if (url.startsWith('/search')) {
       return this.searchTokens.next()
     } else if (url.startsWith('/graphql')) {
@@ -154,14 +185,14 @@ class GithubApiProvider {
   // Act like request(), but tweak headers and query to avoid hitting a rate
   // limit. Inject `request` so we can pass in `cachingRequest` from
   // `request-handler.js`.
-  request(request, url, options = {}, callback) {
+  request({ request, url, options = {}, neededScopes = {}, callback }) {
     const { baseUrl } = this
 
     let token
     let tokenString
     if (this.withPooling) {
       try {
-        token = this.tokenForUrl(url)
+        token = this.tokenForUrl(url, neededScopes)
       } catch (e) {
         callback(e)
         return
@@ -198,14 +229,20 @@ class GithubApiProvider {
     })
   }
 
-  requestAsPromise(request, url, options) {
+  requestAsPromise({ request, url, options, neededScopes }) {
     return new Promise((resolve, reject) => {
-      this.request(request, url, options, (err, res, buffer) => {
-        if (err) {
-          reject(err)
-        } else {
-          resolve({ res, buffer })
-        }
+      this.request({
+        request,
+        url,
+        options,
+        neededScopes,
+        callback: (err, res, buffer) => {
+          if (err) {
+            reject(err)
+          } else {
+            resolve({ res, buffer })
+          }
+        },
       })
     })
   }
--- a/services/github/github-auth-service.js
+++ b/services/github/github-auth-service.js
@@ -2,21 +2,22 @@ import gql from 'graphql-tag'
 import { mergeQueries } from '../../core/base-service/graphql.js'
 import { BaseGraphqlService, BaseJsonService } from '../index.js'
 
-function createRequestFetcher(context, config) {
+function createRequestFetcher(context, config, neededScopes) {
   const { sendAndCacheRequestWithCallbacks, githubApiProvider } = context
 
   return async (url, options) =>
-    githubApiProvider.requestAsPromise(
-      sendAndCacheRequestWithCallbacks,
+    githubApiProvider.requestAsPromise({
+      request: sendAndCacheRequestWithCallbacks,
       url,
-      options
-    )
+      options,
+      neededScopes,
+    })
 }
 
 class GithubAuthV3Service extends BaseJsonService {
-  constructor(context, config) {
+  constructor(context, config, neededScopes) {
     super(context, config)
-    this._requestFetcher = createRequestFetcher(context, config)
+    this._requestFetcher = createRequestFetcher(context, config, neededScopes)
     this.staticAuthConfigured = true
   }
 }
@@ -27,10 +28,10 @@ class GithubAuthV3Service extends BaseJsonService {
 // useful when consuming GitHub endpoints which are not rate-limited: it
 // avoids wasting API quota on them in production.
 class ConditionalGithubAuthV3Service extends BaseJsonService {
-  constructor(context, config) {
+  constructor(context, config, neededScopes) {
     super(context, config)
     if (context.githubApiProvider.globalToken) {
-      this._requestFetcher = createRequestFetcher(context, config)
+      this._requestFetcher = createRequestFetcher(context, config, neededScopes)
       this.staticAuthConfigured = true
     } else {
       this.staticAuthConfigured = false
@@ -39,9 +40,9 @@ class ConditionalGithubAuthV3Service extends BaseJsonService {
 }
 
 class GithubAuthV4Service extends BaseGraphqlService {
-  constructor(context, config) {
+  constructor(context, config, neededScopes) {
     super(context, config)
-    this._requestFetcher = createRequestFetcher(context, config)
+    this._requestFetcher = createRequestFetcher(context, config, neededScopes)
     this.staticAuthConfigured = true
   }
 
--- a/services/github/github-auth-service.spec.js
+++ b/services/github/github-auth-service.spec.js
@@ -25,17 +25,32 @@ describe('GithubAuthV3Service', function () {
     }
   }
 
-  it('forwards custom Accept header', async function () {
-    const sendAndCacheRequestWithCallbacks = sinon.stub().returns(
+  class ScopedDummyGithubAuthV3Service extends DummyGithubAuthV3Service {
+    constructor(context, config) {
+      super(context, config, { needsPackageScope: true })
+    }
+  }
+
+  let sendAndCacheRequestWithCallbacks, mockToken
+  const githubApiProvider = new GithubApiProvider({
+    baseUrl: 'https://github-api.example.com',
+  })
+
+  beforeEach(function () {
+    sendAndCacheRequestWithCallbacks = sinon.stub().returns(
       Promise.resolve({
         buffer: '{"requiredString": "some-string"}',
         res: { statusCode: 200 },
       })
     )
-    const githubApiProvider = new GithubApiProvider({
-      baseUrl: 'https://github-api.example.com',
-    })
-    const mockToken = { update: sinon.mock(), invalidate: sinon.mock() }
+    mockToken = { id: 'abc123', update: sinon.mock(), invalidate: sinon.mock() }
+  })
+
+  afterEach(function () {
+    sinon.restore()
+  })
+
+  it('forwards custom Accept header', async function () {
     sinon.stub(githubApiProvider.standardTokens, 'next').returns(mockToken)
 
     DummyGithubAuthV3Service.invoke({
@@ -47,7 +62,26 @@ describe('GithubAuthV3Service', function () {
       headers: {
         'User-Agent': 'Shields.io/2003a',
         Accept: 'application/vnd.github.antiope-preview+json',
-        Authorization: 'token undefined',
+        Authorization: 'token abc123',
+      },
+      url: 'https://github-api.example.com/repos/badges/shields/check-runs',
+      baseUrl: 'https://github-api.example.com',
+    })
+  })
+
+  it('uses token with correct read scope', function () {
+    sinon.stub(githubApiProvider.packageScopedTokens, 'next').returns(mockToken)
+
+    ScopedDummyGithubAuthV3Service.invoke({
+      sendAndCacheRequestWithCallbacks,
+      githubApiProvider,
+    })
+
+    expect(sendAndCacheRequestWithCallbacks).to.have.been.calledOnceWith({
+      headers: {
+        'User-Agent': 'Shields.io/2003a',
+        Accept: 'application/vnd.github.antiope-preview+json',
+        Authorization: 'token abc123',
       },
       url: 'https://github-api.example.com/repos/badges/shields/check-runs',
       baseUrl: 'https://github-api.example.com',
--- a/services/github/github-constellation.js
+++ b/services/github/github-constellation.js
@@ -5,6 +5,11 @@ import GithubApiProvider from './github-api-provider.js'
 import { setRoutes as setAdminRoutes } from './auth/admin.js'
 import { setRoutes as setAcceptorRoutes } from './auth/acceptor.js'
 
+const readPackagesScope = 'read:packages'
+// Multiple scopes need to be uri-encoded space delimited
+const tokenScopes = `${readPackagesScope}`
+const persistenceScopeDelimiter = '.scopes.'
+
 // Convenience class with all the stuff related to the Github API and its
 // authorization tokens, to simplify server initialization.
 class GithubConstellation {
@@ -24,6 +29,8 @@ class GithubConstellation {
     this._debugEnabled = config.service.debug.enabled
     this._debugIntervalSeconds = config.service.debug.intervalSeconds
     this.shieldsSecret = config.private.shields_secret
+    this._tokenScopes = {}
+    this._maxNumReservedScopedTokens = 0
 
     const { redis_url: redisUrl, gh_token: globalToken } = config.private
     if (redisUrl) {
@@ -38,6 +45,9 @@ class GithubConstellation {
       baseUrl: process.env.GITHUB_URL || 'https://api.github.com',
       globalToken,
       withPooling: !globalToken,
+      tokenScopeNames: {
+        readPackages: readPackagesScope,
+      },
       onTokenInvalidated: tokenString => this.onTokenInvalidated(tokenString),
     })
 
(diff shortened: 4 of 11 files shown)


services/github/github-api-provider.js
```diff
@@ -154,14 +185,14 @@ class GithubApiProvider {
   // Act like request(), but tweak headers and query to avoid hitting a rate
   // limit. Inject `request` so we can pass in `cachingRequest` from
   // `request-handler.js`.
-  request(request, url, options = {}, callback) {
+  request({ request, url, options = {}, neededScopes = {}, callback }) {
```
calebcartwright, line 188: Perhaps somewhat a subjective/stylistic preference, but I get uneasy having to deal with positional based parameters once you have more than 2 or 3, so opted to switch the signature to an objected/named params which of course introduces diffs at respective call sites (which is the source of most of the diff in the test files in this commit as well as the update to the suggestion service)

services/github/github-auth-service.js
```diff
@@ -2,21 +2,22 @@ import gql from 'graphql-tag'
...
-      options
-    )
+      options,
+      neededScopes,
+    })
 }
 
 class GithubAuthV3Service extends BaseJsonService {
-  constructor(context, config) {
+  constructor(context, config, neededScopes) {
```
calebcartwright, line 18: This felt like the best way to allow service classes to enumerate their scope needs. Just like what's shown in the unit tests below, all that'd be needed for such classes (e.g. #4184) is an explicit constructor with a "hey i need this scope" flag. We could even extend the inheritance hierarchy with another set of class levels, e.g. `GitHubPackagesAuthV4 extends GithubAuthV4Service` with common scope needs so that individual service classes could extend and not have to worry about explicitly providing them

services/github/github-constellation.js
```diff
@@ -5,6 +5,11 @@ import GithubApiProvider from './github-api-provider.js'
 import { setRoutes as setAdminRoutes } from './auth/admin.js'
 import { setRoutes as setAcceptorRoutes } from './auth/acceptor.js'
 
+const readPackagesScope = 'read:packages'
+// Multiple scopes need to be uri-encoded space delimited
+const tokenScopes = `${readPackagesScope}`
+const persistenceScopeDelimiter = '.scopes.'
```
calebcartwright, line 11: realize that defining these up here has some impact on readability in places where they are used, but i opted to encapsulate and hoist them up here after twice failing to type the trailing `.` when using the raw delimiter value in each site

services/github/github-constellation.js
```diff
@@ -70,8 +80,21 @@ class GithubConstellation {
       log.error(e)
     }
 
+    // Reserve a subset of scoped tokens from the total set
+    // to be used for queries which require an explicit scope,
+    // while leaving a sufficient amount of tokens (scoped or unscoped)
+    // for the bulk of our requests which don't care about scopes.
+    this._maxNumReservedScopedTokens = Math.floor(tokens.length * 0.15)
```
calebcartwright, line 87: If we opt to go with the approach proposed in this PR, then I could envision a point down the line where we allow this ratio to be set from an environment variable so that we can adjust via a simple heroku config update vs code change/PR

services/github/github-constellation.js
```diff
@@ -70,8 +80,21 @@ class GithubConstellation {
...
     tokens.forEach(tokenString => {
-      this.apiProvider.addToken(tokenString)
+      const [token, scopes] = tokenString.split(persistenceScopeDelimiter)
+      this._tokenScopes[token] = scopes || null
```
calebcartwright, line 90: i opted to make this explicitly `null` vs the existing `undefined` as I found that made troubleshooting and debugging easier. kept it in here as i don't think null vs. undefined as any runtime impact given the code, since i can see it being beneficial for any future debugging needs

chris48s:
Fair play for biting the bullet and tackling this one. I have got some rough notes on this PR sitting in a text file but I haven't had a chance to fully digest this or type the notes I have up properly yet. I will try and get them posted this week though so we don't lose the momentum on this.

chris48s:
> There's a high level binary decision at the top of whether or not we should persist the corresponding scopes with each token vs. dynamically checking scopes on the fly. I decided against trying to dynamically check scopes because that'd require we either:
> 
> - Check scopes on the respective current token prior to using it for an API request
> - Check scopes on all tokens at startup
> 
> Both of those options seem unacceptably excessive, as in the latter case every time a dyno started each dyno would have to fire off 20k+ requests, and the former would inherently require a sequential set of API calls to satisfy every request.
> 
> Accordingly, I'm of the opinion we do need to persist the associated scopes along with each token.

:+1: Fully agree with this

---

> As such, I believe it'll be easier and more maintainable/flexible if our persisted tokens carry their associated scope directly vs. trying to use collection-based keys to imply scopes.

:100: Also this

---

> One approach could have been more object-based representations in code of the raw persisted token material which is then json-stringified for persistence, however, I believe that adds unnecessary noise and verbosity to the set, and also introduces serialization concerns.

I'd like to at least question this assumption, or drill down into the thinking more.

I think I'd more naturally gravitate to serializing a JSON string in the redis key. i.e:

- `'{"token":"abc123","scopes":["read:packages","read:user"]}'` rather than
- `'abc13.scopes.read:packages%20read:user'`

I don't have super deep redis knowledge but from https://redis.com/redis-best-practices/data-storage-patterns/json-storage/

> This, while seemingly simple, has some very real drawbacks:
> 
> - The serialization takes client computational resources during read and write.
> - The JSON format adds size to the data
> - Redis has only indirect ways to manipulate the actual JSON data.
> 
> The first couple of points may be negligible on small amounts of data, but can definitely add up. Especially with multi-megabyte JSON stores. The third point is more critical, however.

We're not looking at I think for our case, size is probably a non-issue - the difference between `'abc13.scopes.read:packages%20read:user'` and `'{"token":"abc123","scopes":["read:packages","read:user"]}'` is a few bytes. Given we're looking at a JSON object with a couple of keys and we'd deserialize all the objects once at server boot time (its off the critical path for badge rendering) I think we don't need to worry too much about the performance difference and the third point is the same if we invent our own (potentially janky) serialization format anyway.

I suppose the only meaningful upside of this I can think of is we can dispense with the need to write a data migration to read in every token from the pool, convert `abc123` to `'{"token":"abc123","scopes":[]}'` and write them back.

I'd be interested if there's other factors you're thinking about here.

---

> Last note/observation is that I've added this in a way such that if a token is re-authorized the pools are adjusted accordingly, at least on that dyno, so that corresponding token can be used correctly without that dyno needing to restart/reload from Redis

I can't remember what we said on this yesterday, but did we agree that the possible scenarios here are:

- token gets new scopes added
- token becomes invalid

There isn't a case where an existing token gets scopes removed, is that right?

---

I have not tested any of this code locally. If I wanted to try running this on my local copy with multiple tokens, how would I do it? I'm not sure I've ever done it before or I might have done and then forgotten.

services/github/github-constellation.js
```diff
@@ -5,6 +5,11 @@ import GithubApiProvider from './github-api-provider.js'
 import { setRoutes as setAdminRoutes } from './auth/admin.js'
 import { setRoutes as setAcceptorRoutes } from './auth/acceptor.js'
 
+const readPackagesScope = 'read:packages'
+// Multiple scopes need to be uri-encoded space delimited
+const tokenScopes = `${readPackagesScope}`
```
chris48s, line 10: I think even if we don't have any integrations that use them yet, it probably makes sense to start _requesting_ more scopes sooner rather than later. i.e: even if we only add `packageScopedTokens` to `GithubApiProvider` now, if we think we've got a good case for needing `public_repo`, `read:user`, `read:org`, etc. I think it makes sense to have a think and make some decisions on that now and implement stuff which needs them later. The reason I say this is because the process of trying to convince new users to donate a token or existing users to re-authorize is in many ways the most cumbersome part of this process and GitHub does not provide any nice way of doing this.. or at least it is the part of the process that is an exercise in cat herding. We have to convince people to do a thing for us - we can't solve it with code, even if its really good code. There's also an element of user 'fatigue' to consider here. If we ask users for 3 new scopes once I suspect we'll get more total takeup than asking for one new scope 3 times over the next 6 months for example.

I think its inevitable that we will have to go through that process more than once over time, but if we can think about it a bit now and reduce the total number of times we need to do it, that is going to be very advantageous.

Just to start collating our pile of possible scopes we might (or might not) want to ask for aside from `package:read`, I think the others I know of are:

- `public_repo`
- `read:user`
- `read:org`

Are there any scopes you know of that would be needed for any other open feature requests that are blocked on this?

services/github/github-auth-service.spec.js
```diff
@@ -25,17 +25,32 @@ describe('GithubAuthV3Service', function () {
     }
   }
 
-  it('forwards custom Accept header', async function () {
-    const sendAndCacheRequestWithCallbacks = sinon.stub().returns(
+  class ScopedDummyGithubAuthV3Service extends DummyGithubAuthV3Service {
+    constructor(context, config) {
+      super(context, config, { needsPackageScope: true })
```
chris48s, line 30: This partly relates to the conversation we were having yesterday about integrations potentially needing >1 scope. I'd tend to think of this as being `scopes: ['read:package']`. I think it also probably makes sense to keep the terminology at the service level (i.e: code a casual contributor is going to write) using the documented scope names https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes

calebcartwright:
We discussed this at our Operations team meeting earlier this week and wanted to recap the discussion here for visibility and to vet my own recollection :smile: 

From my perspective the main takeaway is that we've got a general consensus on the approach and core aspects (need to persist scopes, don't want/need a different DB engine for persistence, etc.), with the biggest outstanding item to work through really being the exact format/structure of scope persistence.

I believe there's 3 proposals for persistence so I think a visual will help.

Let's start with a scenario where we have 10 tokens with varying scopes:

| token | scopes |
| ------- | ---------- |
| abc   | |
| def | read:packages |
| ghi | read:packages, read:user |
| jkl | read:packages, read:user |
| mno |  |
| pqr | read:packages |
| stu | read:packages |
| vwx | |
| yza | read:packages |
| 123 | read:packages, read:user |

And then mapping those to the three different options we discussed (no significance implied to ordering of the options) to their representation in Redis. NB - this is just for illustrative purposes so let's not drill into the specifics at this point (e.g. key names)

Option 1 - single key, token set elements have scopes appended (initial proposal in this PR)

| key | value |
| ------- | ---------- |
| githubTokens | ( abc, def.scopes.read:packages, ghi.scopes.read:packages%20read:user, jkl.scopes.read:packages%20read:user, mno, pqr.scopes.read:packages, stu.scopes.read:packages, vwx, yza.scopes.read:packages, 123.scopes.read:packages%20read:user )  |

Option 2 - single key, token set elements are serialized/stringified objects with scope property

| key | value |
| ------- | ---------- |
| githubTokens | ( '{"token":"abc","scopes":[]}', '{"token":"def","scopes":["read:packages"]}', '{"token":"ghi","scopes":["read:packages","read:user"]}', '{"token":"jkl","scopes":["read:packages","read:user"]}', '{"token":"mno","scopes":[]}', '{"token":"pqr","scopes":["read:packages"]}', '{"token":"stu","scopes":["read:packages"]}', '{"token":"vwx","scopes":[]}', '{"token":"yza","scopes":["read:packages"]}', '{"token":"123","scopes":["read:packages","read:user"]}' )  |

Option 3 - multiple, per-single-scope key, tokens added to sets for each key the token has the corresponding scope for, with the original/current key continuing to contain all tokens (regardless of scope)

| key | value |
| ------- | ---------- |
| githubTokens | ( abc, def, ghi, jkl, mno, pqr, stu, vwx, yza, 123 ) |
| readPackages | ( def, ghi, jkl, pqr, stu, yza, 123 ) |
| readUser | ( ghi, jkl, 123 ) |

@paulmelnikow  - Option 3 was your proposal on the call, so wanted to confirm what I've captured here accurately reflects your proposal and preference?

services/github/github-constellation.js
calebcartwright: > I think even if we don't have any integrations that use them yet, it probably makes sense to start requesting more scopes sooner rather than later

Agreed. We do know that it's impossible to completely future-proof ourselves from _ever_ having to go through this again, but we should absolutely strive to cover everything known any time we need to request more scopes, and I'd like for us to identify the full set of scopes we want for this pass before merging the PR.

The balance of course is not overbidding on scopes in a way that will turn people off/reduce the population of those willing to donate. For example, I've paused/avoided authorizing apps that request the `read:user` scope because it grants full access to all private profile information (including contact info).

Obviously we store donated tokens securely and have no intention of ever using them for anything beyond GitHub API calls to fetch data needed for badges. However, I do think it's important to think about it from the concerns/perspective of users considering donating/granting scope increases.

> Are there any scopes you know of that would be needed for any other open feature requests that are blocked on this?

The only scenario I recall that also hit the scope wall was #6104, which would've needed `public_repo` for repository-level projects, and `read:org` for organization level projects. I don't think that first one is going to be viable regardless, as `public_repo` also includes write permissions (https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps)

> Limits access to public repositories. **That includes read/write access to code**, commit statuses, repository projects, collaborators, and deployment statuses for public repositories and organizations. Also required for starring public repositories.

I personally wouldn't authorize a token with that level of scope to anyone, so I wouldn't feel right asking Shields users to do this. In my opinion that feels like a use case best handled via a custom endpoint badge where users could individually wire up their endpoint with an appropriately scoped token they keep/maintain.

`read:org` _might_ be more viable, though it's unclear to me whether that's limited to just exposing the organization/team mapping for that specific user (i.e. what organizations and what org teams is the owner of the token a member of), or if it can be used to see _all_ org and org team membership for any org/team the token owner is a member of. The former would be okay, but the latter would give me pause to say the least as I believe GitHub allows users to choose whether to convey their membership publicly for a reason and such a token would allow token-donators to theoretically expose that info of _other_ org/team members without their consent.

At least at this point in time, I'd lean towards directing all GitHub Project-related use cases to the Custom Endpoint path, and we could even consider implementing an example endpoint and linking out to it (e.g. from an issue/discussion in this repo and/or our awesome list)

services/github/github-auth-service.spec.js
calebcartwright: I originally started out doing what you're suggesting, but here's the main reasons I decided to pivot to what's here now:

- using named boolean properties doesn't prevent the ability for a service class to request multiple scopes, it's just a different "how" (e.g. `{ needsPackageScope: true, needsReadOrgScope: true }`), plus it also gives us the ability to define our own grouping scopings that can be mapped directly to a corresponding pool (e.g. `{ needsPackageAndUserScope: true }`)
- we're working with an entirely constrained, and small universe of possibilities which are all known at build time. As such there's no reason to have to iterate over collections and do string comparisons on every request at runtime

I acknowledge and appreciate that there's a lower barrier for someone _writing_ a service class in isolation to be able to just copy in the literal scope value from the GitHub docs into a `scopes` array, and that using named boolean properties introduces a "Shields-ism" that's not immediately obvious to new contributors.

However, I don't view that as a significant problem nor hurdle. New GitHub badges are not exactly a common occurrence, and I think we can cover this pretty easily via the review process and documentation, especially since a net-new badge that requires a net new scope/scope combination is going to require touching some core code (at least in the api provider) anyway.

calebcartwright:
> I'd like to at least question this assumption, or drill down into the thinking more.
> I think I'd more naturally gravitate to serializing a JSON string in the redis key. i.e
> We're not looking at I think for our case

Good note. Context is crucial, and the Shields.io production usage context for Redis is **very** different than the canonical hot cache usage pattern for which Redis is often utilized. We have _one_ read at startup, and very rarely, a couple non-latency sensitive writes at runtime.

I'm not a Redis veteran/expert by any stretch, but the commentary on json you shared definitely aligns with my own knowledge and experience, and did influence my thinking.

> I'd be interested if there's other factors you're thinking about here.

The other factor I was worried about is having to do functional replacement of set elements with a single-key based store (options 1 and 2 in https://github.com/badges/shields/pull/7037#issuecomment-932775668) to account for tokens being granted additional scopes. May not be a huge risk, but the json formatting does introduce a larger radius for specious match failures and key duplication (e.g. potential issues with ordering of scopes in a collection, whitespace differences in json, etc.). I felt like the scopes-as-appendages route in option 1 helped reduce some of those risks since the value that would be used for an update is comparatively closer to an immutable literal that could only get botched if the delimiter literal (`.scopes.`) or concatenation was broken

> I can't remember what we said on this yesterday, but did we agree that the possible scenarios here are:
>
>    token gets new scopes added
>    token becomes invalid
>

Correct

> There isn't a case where an existing token gets scopes removed, is that right?

That's my understanding as well, and our collective consensus from the call. 

> I have not tested any of this code locally. If I wanted to try running this on my local copy with multiple tokens, how would I do it? I'm not sure I've ever done it before or I might have done and then forgotten.

I dunno if we have an official/preferred way of doing this, but a few paths depending on how thoroughly you want to test, all of which require you don't have the `gh_token` config value set locally:

* have redis running locally and get your corresponding connection string plugged into your `redis_url` config value, and within your redis instance add at least tokens to the set value for the `githubUserTokens` key (PATs will work fine for this, so could use our bot account to generate a token for another user, and/or I can share one with you offline)
* configure your local to point to the prod redis endpoint. would advise against this pretty strongly for a variety of reasons, but since this would be in a read-only mode (no token scope changes) i suppose it's technically feasible
* with local redis configured, use scripts (similar to https://github.com/badges/shields/blob/master/scripts/redis-connectivity-test.js) to programmatically invoke the add/remove methods
* if you just want to test the token utilization pieces without having to deal with redis, tweak the code to use/return a hardcoded collection of tokens, e.g. in either of these places:
    * https://github.com/badges/shields/blob/master/core/token-pooling/redis-token-persistence.js#L26
    * https://github.com/badges/shields/blob/master/services/github/github-constellation.js#L66 

For the last mile of testing I suspect we'd want a service badge that actually hits the new pools, e.g. a #4184 or #4101. I'd originally wanted to keep the addition of new badges separate, but open to grabbing and adjusting commits to include a new badge if that helps

services/github/github-constellation.js
chris48s: Yes you're right. We shouldn't ask for public_repo (or any other scope that grants write access). Good shout.

calebcartwright:
Want to bump this because I'll confess that my motivation to get this over the finish line is likely inversely proportionally to the number of merge conflicts/amount of time it sits in limbo :rofl:

calebcartwright:
I think at this point it's probably not worth attempting to shave the yak that a rebase/conflict resolution has become, so I'm going to close this. I still think the nuts and bolts of this PR lay out a reasonable roadmap for us to finally tackle this piece, but think that'll be most easily handled in a separate PR.

One thing that I do feel more strongly about now then when I started is that it's most likely better to invert the order of implementation we originally had in mind, to make the programmatic/storage/etc. updates first _before_ changing the requested scopes. As noted above this isn't the easiest thing to test, so I have that perspective because I think it'd be best to validate our existing token pool with whatever new storage strategy/code flows and then augment it with updated scopes (doing so also avoids a scenario of us having tokens with unknown scopes and/or having to attempt to retroactively process stored tokens)

Issue #4474, opened 2019-12-27
[david] Support selecting a branch


umarcor:
david-dm allows to specify a branch as an additional item in the URL. This can be combined with the `path` query parameter. For example: https://david-dm.org/umarcor/hwstudio/actions?path=app Unfortunately, this feature seems not to be supported in this repo: `/david/:kind?/:user/:repo?path=packages%2Fbabel-core`. I tried https://img.shields.io/david/umarcor/hwstudio/actions.svg?path=app&longCache=true&style=flat-square&label=app%20deps.

calebcartwright:
Do you know what the request to David would be to get the data? 

The typical upstream endpoint Shields uses to get the data is:
https://david-dm.org/umarcor/hwstudio/info.json?path=app

Unfortunately David also returns an http 500 error for the "not found" use case (vs. the typical 404), and I'm getting that same http 500 error on the link you provided in the issue description:
https://david-dm.org/umarcor/hwstudio/actions?path=app

Is `actions` the branch you are targetting, and if so, does that branch actually exist in your repo?

umarcor:
I'm so sorry. Actually, I renamed that branch and I forgot to update it here. These do exist:

- https://david-dm.org/umarcor/hwstudio/develop?path=app
- https://david-dm.org/umarcor/hwstudio/webpack?path=app

And these endpoints return content:

- https://david-dm.org/umarcor/hwstudio/develop/info.json?path=app
- https://david-dm.org/umarcor/hwstudio/webpack/info.json?path=app

calebcartwright:
That explains it then, thanks @umarcor! To address this we'd need to update the route for our David service to accept an optional branch parameter.

At the moment I'm not sure whether it'd be best to do so via a route param or query param. Generally Shields routes accept a `branch` value as an optional trailing route param (`/:user/:repo/:branch?`) but David has a leading optional `kind` route param as well that could make things a bit trickier than usual

alanshaw:
FYI, in the new david status API the branch is a query string parameter called `ref` https://github.com/alanshaw/david-status-api#status-api

umarcor:
@alanshaw, I just tried `https://david-dm.org/umarcor/hwstudio?ref=vue/threejs/main` and it's returning:

```
FetchError: invalid json response body at https://project.david-dm.org/gh/umarcor/hwstudio reason: Unexpected token u in JSON at position 0
    at /home/node/app/node_modules/node-fetch/lib/body.js:48:31
[...]

PR #7070, opened 2021-09-25
switch to [David] status API and update error handling

calebcartwright:
Service tests have been failing for a while on account of the upstream service issuing a wider mix of 50x status responses for basic not founds so this both switches to the new status API (refs #5968) which is noticeably more performant, and also extends the error handling to account for those additional status codes


calebcartwright:
Adding blocker label since we'll need this to land for #4474 to be able to proceed

chris48s (approved):

services/david/david.service.js
```diff
@@ -71,22 +71,19 @@ export default class David extends BaseJsonService {
...
-        /* note:
-        david returns a 504 response for 'not found'
-        e.g: https://david-dm.org/foo/barbaz/info.json
-        not a 404 so we can't handle 'not found' cleanly
-        because this might also be some other error.
-        */
-        504: 'repo or path not found or david internal error',
+        502: notFoundError,
+        503: notFoundError,
+        504: notFoundError,
```
calebcartwright, line 86: And yes, I've encountered all of these (though in fairness I think the 503 could have been on the prior endpoint). These have been so persistently painful on the daily service test runs that I'd rather just play it safe and cover our 50x bases

PR #7071, opened 2021-09-25
Add [Revolt] badge

Jan0660:
Adds Revolt server invite badge and logo.


nerrixDE (approved):
looks good, great work!

calebcartwright:
Thank you for the PR! I've not had a chance to go through this in detail, but a few things I do want to call out after a cursory pass:

- There's an issue with the tests that's causing CI to fail. Would advise taking a look at the [service test guide](https://github.com/badges/shields/blob/master/doc/service-tests.md), at least the first two sections
- We don't really accept custom logos that we have to maintain, at least not until all other options have been exhausted. Please take a look at our [guide on logos](https://github.com/badges/shields/blob/master/doc/logos.md#contributing-logos) for more information, including preferred routes for new logos
- We have a [badge specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md) that codifies what our badges should look like, at least in the default cases. There's definitely some items in the initial proposal that go against the spec and would need to be adjusted (e.g. the default badge label should be something like `chat` instead of branding, which is the same label as the Discord badges by the way)

PyvesB:
What's the status on this one? Reading through the diff and @calebcartwright's comment, my understanding is that the custom logo and the non-default badge example are the blockers here.

PyvesB:

If someone else wants to pick this up, you can simply fetch the commits by running `git fetch origin pull/7071/head:pr-7071`. Address the pending review comments and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

PR #7072, opened 2021-09-26
add documentation guide for adding new secrets

calebcartwright:
Refs #7023


chris48s (dismissed):
I've pressed approve on this because what's here isn't bad or wrong. That said, the process for adding new secrets is a very slight subset of the process for adding settings more generally. For example if you think about a commit like this which adds a public/config setting https://github.com/badges/shields/commit/b1fc4925928c061234e9492f3794c0797467e123 the process/gotchas are the roughly the same. I think we could generalise this doc very slightly to cover all settings and note that secrets go in the private settings whereas configuration goes in public and make this more broadly useful.

calebcartwright:
Yup that sounds reasonable, generalized a bit in 2d97faf4ff286a0159f9ab047028fecb50773f3b. Do you think this covers them both well enough or is there more we could do?

chris48s (approved):
Looks good :+1:  If we do think of anything else, we've got a place to add it

PR #7074, opened 2021-09-26
handle null licenses in crates.io response schema, run [crates]

### LINKED ISSUE #7073: Error badge display "invalid response data" on https://img.shields.io/crates/v/stun.svg

shields.io

### 🐞 Description

https://img.shields.io/crates/v/stun.svg shows "invalid response data", instead of crates.io version.

### 🔗 Link to the badge

https://img.shields.io/crates/v/stun.svg

### 💡 Possible Solution

N/A

calebcartwright:
Fixes #7073

### DIFF
--- a/services/crates/crates-license.spec.js
+++ b/services/crates/crates-license.spec.js
@@ -1,0 +1,27 @@
+import { expect } from 'chai'
+import { InvalidResponse } from '../index.js'
+import CratesLicense from './crates-license.service.js'
+
+describe('CratesLicense', function () {
+  it('throws InvalidResponse on error response', function () {
+    expect(() =>
+      CratesLicense.transform({ errors: [{ detail: 'invalid semver' }] })
+    )
+      .to.throw(InvalidResponse)
+      .with.property('prettyMessage', 'invalid semver')
+  })
+
+  it('throws InvalidResponse on null license with specific version', function () {
+    expect(() =>
+      CratesLicense.transform({ version: { num: '1.2.3', license: null } })
+    )
+      .to.throw(InvalidResponse)
+      .with.property('prettyMessage', 'invalid null license')
+  })
+
+  it('throws InvalidResponse on null license with latest version', function () {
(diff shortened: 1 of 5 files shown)


services/crates/crates-license.spec.js
```diff
@@ -0,0 +1,27 @@
...
+      .to.throw(InvalidResponse)
+      .with.property('prettyMessage', 'invalid null license')
+  })
+
+  it('throws InvalidResponse on null license with latest version', function () {
+    expect(() => CratesLicense.transform({ versions: [{ license: null }] }))
+      .to.throw(InvalidResponse)
+      .with.property('prettyMessage', 'invalid null license')
+  })
+})
```
chris48s, line 27: Should we have an explicit test for the valid case here, or have you left it because its covered by the service tests?

calebcartwright: To be honest I hadn't really thought about it. I started out with the unit tests, then decided to add the service tests to account for the schema, and did so without realizing the service tests make these unit tests largely duplicative.

Do you think it's worth even keeping these?

chris48s: I think yes - we're covering a wider range of cases here. Also its more explicit what we're really testing here. In fact...

services/crates/crates-license.tester.js
```diff
@@ -16,4 +11,12 @@ t.create('license (with version)')
 
 t.create('license (not found)')
   .get('/not-a-real-package.json')
-  .expectBadge({ label: 'crates.io', message: 'not found' })
+  .expectBadge({ label: 'license', message: 'not found' })
+
+t.create('license (null licenses in history)')
+  .get('/stun.json')
```
chris48s, line 17: This is an example of such a specific condition, lets link to https://github.com/badges/shields/issues/7073 in a comment here for future reference

chris48s (approved):

PR #7076, opened 2021-09-26
refactor and simplify [AzureDevOpsTests]

calebcartwright:
Subsequent to the addition of this badge, we developed some helper functions for both service classes and testers to deal with the common bits of providing and testing badges relating to displaying test results.

This updates the tester for the AzureDevOpsTests badge to take advantage of those helpers, and also does some minor cleanup/refactoring of the service class to extract transformation concerns out of the `handle` function


chris48s:
As well as the refactoring and using shared helpers (which seems fine) we seem to also be dropping quite a lot of test cases in this PR and its not really explained in the top post. What's your thinking on this?

calebcartwright:
> As well as the refactoring and using shared helpers (which seems fine) we seem to also be dropping quite a lot of test cases in this PR and its not really explained in the top post. What's your thinking on this?

Yeah I guess it was more subtle than I realized. We currently have duplicated tests for most scenarios, one that's mocked and another that uses real data. There's no point/value in having both, especially since I've switched the test targets to an Azure DevOps project that I own and can ensure the live pipelines have their runs retained.

There were also tests that were checking for the [invalid response data](https://github.com/badges/shields/pull/7076/files#diff-6a995efb90aa51b5f1784b148ac8f9bac524ae45499410bfdc834fa730a873e3L154-L163) message in the event schema validation fails, which is also not something we do/duplicate in service test runs.

I believe what's here fully covers both the test-badge aspects as well as the real responses encountered with the API call chain for all scenarios. But please let me know if you think there's anything significant that's getting dropped and I'll take a closer look

PR #7080, opened 2021-09-29
Add authentication for Libraries.io-based badges, run [Libraries Bower]

### LINKED ISSUE #6179: Libraries.io Dependencies Not Working

- [ ] My own instance

:beetle: **Description**

Whenever I use the Libraries.io dependency status badge, results are inconsistent, and I get "invalid" as a status for most packages I try. Sometimes they will work until I add a query param (logo or badge style for ex.), then they will say invalid. Also it seems like the info is straight up wrong sometimes?

TypeScript example: Libraries.io dependency status for latest release

:link: **Link to the badge**

https://img.shields.io/librariesio/release/npm/@ebth/design-system

libraries.io entry: https://libraries.io/npm/@ebth%2Fdesign-system

As you can see, there are many dependencies that should be out of date. The Sourcerank endpoint also does not work for me.

:bulb: **Possible Solution**

I'm guessing that zero dependencies out of date is equivalent to "invalid". Would be nice if this just said zero instead? Many common packages are showing invalid.

**side note**
I know you can't get Github info for private repos, but it would be nice if that info (stars, number of releases, latest release date, etc.) from npm based endpoints like Libraries.io since those can be public with a private repo.

calebcartwright:
Resolves #6179 - would advise reviewing by commit

Our Lirbaries.io-based badges have been problematic for a long time. In the past it was advantageous for us to _not_ authenticate requests as we were somehow actually getting a higher rate limit with the anonymous requests. However, that changed some time back which is why the service tests have been failing for months and the badges have been flaky to say the least.

According to the response headers on anonymous requests, the rate limit is 30/minute (compared to 60/minute with authentication). However, each anonymous request seems to use at least 5 points, so in practice it ends up being more like only a handful of requests per minute.

This PR adds an API provider type pattern similar to, but much simpler than, what we do with our GitHub requests, wherein a single token can be specified or multiple tokens can be used to fill a rotating pool. We have had past discussions about just doing the simple/standard auth route, however, we did have concerns that a single token would be insufficient during heavier traffic periods with our existing load patterns for these badges. Additionally, I could envision some growth/traffic increase for these badges once we get them stabilized (e.g. some folks have left Shields over the issue with these badges, and could return/future users could stick once we're able to offer these consistently).

As discussed on today's ops meeting, this implementation allows multiple tokens to be provided via the standard config instead of having them persisted in/pulled from Redis. 

I've already wired up the CI environment with a token (using a bot account I own), and myself and a couple other maintainers will donate our Libraries.io tokens in our prod environment.


core/token-pooling/token-pool.js
```diff
@@ -80,6 +80,10 @@ class Token {
     return this.usesRemaining <= 0 && !this.hasReset
   }
 
+  get decrementedUsesRemaining() {
+    return this._usesRemaining - 1
+  }
+
```
calebcartwright, line 86: The rationale/necessity for this is explained in the inline comments in the `getRateLimitFromHeaders` function in the new `librariesio-api-provider` class

services/librariesio/librariesio-api-provider.js
```diff
@@ -0,0 +1,106 @@
...
+// Provides an interface to the Libraries.io API.
+export default class LibrariesIoApiProvider {
+  constructor({ baseUrl, tokens = [], defaultRateLimit = 60 }) {
+    const withPooling = tokens.length > 1
+    Object.assign(this, {
+      baseUrl,
+      withPooling,
+      globalToken: tokens[0],
+      defaultRateLimit,
+    })
```
calebcartwright, line 14: Opted to stick with simplicity in the event of a single token, to avoid dealing with all the pooling when there isn't really a pool

services/librariesio/librariesio-api-provider.js
```diff
@@ -0,0 +1,106 @@
...
+
+    if (this.withPooling) {
+      this.standardTokens = new TokenPool({ batchSize: 45 })
+      tokens.forEach(t => this.standardTokens.add(t, {}, defaultRateLimit))
+    }
+  }
```
calebcartwright, line 20: Arbitrary number selected for the batch size. I don't think it matters all that much what the actual value is, so don't feel too strongly about it.

services/librariesio/librariesio-api-provider.js
```diff
@@ -0,0 +1,106 @@
...
+    // advance the value to _something_
+    const retryAfter = headers['retry-after']
+    const nextReset = Date.now() + (retryAfter ? retryAfter * 1000 : 0)
+
+    return {
+      rateLimit,
+      totalUsesRemaining,
+      nextReset,
+    }
+  }
```
calebcartwright, line 55: I hope the inline comments provide sufficient explanation for what's happening here. I have opened an upstream issue in the Libraries.io repo (link included in the comments) to see if we can get some consistent response header behavior, but otherwise felt like we had to deal with the reality of the present.

services/librariesio/librariesio-api-provider.js
```diff
@@ -0,0 +1,106 @@
...
+
+  updateToken({ token, res }) {
+    const { totalUsesRemaining, nextReset } = this.getRateLimitFromHeaders({
+      headers: res.headers,
+      token,
+    })
+    token.update(totalUsesRemaining, nextReset)
+  }
+
+  async requestAsPromise(request, url, options = {}) {
```
calebcartwright, line 65: I kept the same name as the corresponding function in the GitHub API Provider, but this is using the same fetcher (got based) as the standard request routes, and is async/await based. Open to bikeshedding on the name, I left it as-is for lazy copy/paste reasons, don't have any real opinions on the name

services/librariesio/librariesio-api-provider.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+    const response = await request(`${baseUrl}${url}`, mergedOptions)
+    if (this.withPooling) {
+      if (response.res.statusCode === 401) {
+        this.invalidateToken(token)
+      } else if (response.res.statusCode < 500) {
+        this.updateToken({ token, url, res: response.res })
+      }
+    }
+    return response
```
calebcartwright, line 104: My reading of how we've set up the actual http request issuing code with got is that this will only reject/throw in the event of major issue (no connectivity) and not on standard http response errors. In such a major issue type of scenario, I don't think it would be safe/valid to attempt to update the token pooling info, so I've not wrapped this in any kind of try catch. I _think_ the call making infrastructure already handles anything that would be caught here, but let me know if a catch is needed with something like an explicit error conversion

config/custom-environment-variables.yml
```diff
@@ -86,6 +86,7 @@ private:
   jenkins_pass: 'JENKINS_PASS'
   jira_user: 'JIRA_USER'
   jira_pass: 'JIRA_PASS'
+  librariesio_tokens: 'LIBRARIESIO_TOKENS'
```
calebcartwright, line 89: opted to just stick an `s` on the end to, along with the docs, convey this supports accepting multiple tokens. I'm not sure we have any prior art with this type of value beyond the various `*origins` vars though, so let me know if folks have any preferred alternatives

services/librariesio/librariesio-base.js
```diff
@@ -0,0 +1,39 @@
...
+  async fetchProject({ platform, scope, packageName }) {
+    return this._requestJson({
+      schema: projectSchema,
+      url: `/${encodeURIComponent(platform)}/${
+        scope ? encodeURIComponent(`${scope}/`) : ''
+      }${encodeURIComponent(packageName)}`,
+      errorMessages: { 404: 'package not found' },
+    })
+  }
+}
```
calebcartwright, line 39: I personally tend to prefer the reusable helper/fetcher type functions over extending the inheritance hierarchy, but it felt cleaner/more straightforward to me to use a base class for setting up the request fetcher on construction in a base class.

services/bower/bower-license.tester.js
```diff
@@ -6,15 +6,6 @@ t.create('licence')
...
 
-t.create('license not declared')
-  .get('/bootstrap.json')
-  .intercept(nock =>
-    nock('https://libraries.io')
-      .get('/api/bower/bootstrap')
-      .reply(200, { normalized_licenses: [] })
-  )
-  .expectBadge({ label: 'license', message: 'missing' })
-
```
calebcartwright, line 17: Our Bower badges had a few such mocked requests. With the hard requirement that we have the API Keys enabled in CI (we **always** exceed the rate limit with our ~20+ tests without auth), trying to mock these becomes increasingly difficult.

We've used some nock tricks elsewhere (e.g. `skipWhen`) that I considered, but I actually found these mocked tests were either duplicative/unnecessary tests of a render helper (as was the case here) and/or the test scenario would be better covered with unit tests.

services/bower/bower-version.service.js
```diff
@@ -38,6 +36,14 @@ class BowerVersion extends BaseBowerService {
...
 
+    return version
+  }
+
+  async handle({ packageName }, queryParams) {
+    const data = await this.fetch({ packageName })
+    const includePrereleases = queryParams.include_prereleases !== undefined
+    const version = this.constructor.transform(data, includePrereleases)
+
     return renderVersionBadge({ version })
```
calebcartwright, line 47: transformation logic separated out here to make it easier to use standard unit tests, because we were using mocked service tests to validate this transformation logic before.

services/bower/bower-version.tester.js
```diff
@@ -34,24 +34,6 @@ t.create('Pre Version for Invalid Package')
...
-
-t.create('Version label should be `no releases` if no pre-release')
-  .get('/v/bootstrap.json?include_prereleases')
-  .intercept(nock =>
-    nock('https://libraries.io')
-      .get('/api/bower/bootstrap')
-      .reply(200, { normalized_licenses: [], latest_release_number: null })
-  )
-  .expectBadge({ label: 'bower', message: 'no releases' })
-
```
calebcartwright, line 54: Covered with new unit tests, mocked tests removed for the aforementioned reasons

services/bower/bower-version.spec.js
```diff
@@ -0,0 +1,92 @@
...
+    const librariesIoApiProvider = new LibrariesIoApiProvider({
+      baseUrl: 'https://libraries.io/api',
+      tokens: [fakeApiKey],
+    })
+
+    it('sends the auth information as configured', async function () {
+      const scope = nock('https://libraries.io/api')
+        // This ensures that the expected credentials are actually being sent with the HTTP request.
+        // Without this the request wouldn't match and the test would fail.
+        .get(`/bower/bootstrap?api_key=${fakeApiKey}`)
```
calebcartwright, line 65: For whatever reason, I was unable to get nock to intercept this request when I using the options arg (e.g. `get('....', { qs: ... })`. As is often the case for me with nock it was probably a pebkac issue, but I didn't feel like trying to battle it

PR #7086, opened 2021-10-01
add 2021 support to [GitHubHacktoberfest]

### LINKED ISSUE #7083: Badge Request: Hacktoberfest 2021! 🎃
:clipboard: **Description**

Yep, being that guy. But I'm not _that_ early - it's already technically October 1st in some parts of the world - I'm looking at you, Christmas Island 😉

I may just be over spamming and this functionality is already built into shields.io to simply appear for each year, but I don't see it working for 2021 yet. Just trying to prepare my repository today and add the shield before it all starts tomorrow.

:link: **Data**

Simply add `/github/hacktoberfest/2021/user/repo` - data source and functionality would remain the same as previous years.

:microphone: **Motivation**

Because Hacktoberfest is awesome!

For maintainers like me who want to put the badge on their GitHub repo 😄

calebcartwright:
Resolves #7083

### DIFF
--- a/services/github/github-hacktoberfest.service.js
+++ b/services/github/github-hacktoberfest.service.js
@@ -60,7 +60,7 @@ export default class GithubHacktoberfestCombinedStatus extends GithubAuthV4Servi
   static category = 'issue-tracking'
   static route = {
     base: 'github/hacktoberfest',
-    pattern: ':year(2019|2020)/:user/:repo',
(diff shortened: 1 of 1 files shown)


chris48s (approved):
Do you reckon we should just change this to allow `[0-9]{4}` or something so we don't have to bump it every year?

PR #7101, opened 2021-10-01
Add files via upload

snxx-lppxx:


calebcartwright:
Closing, both because it directly contradicts our [documented procedures on Logos](https://github.com/badges/shields/blob/master/doc/logos.md) without any description/explanation nor associated issue, and because it seems like Hacktoberfest spam to be honest.

If you are genuinely interested in contributing to the Shields project (which would be great!), we've got a wide range of [issues marked on our backlog that are good for first time contributors](https://github.com/badges/shields/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22). Additionally, while we [don't require](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#suggesting-improvements) issues be created/referenced for a PR, we do still need contributors to explain the what/why of a suggested change _especially_ in the absence of an associated issue.

PR #7105, opened 2021-10-04
remove [github] admin routes

chris48s:
As discussed in last ops meeting, we don't need this anymore. At this point it just represents risk.

TODO: remove `SHIELDS_SECRET` var in heroku after we deploy.


core/server/secret-is-valid.js
```diff
@@ -1,18 +0,0 @@
-function constEq(a, b) {
```
paulmelnikow, line 1: Good riddance to this function!

paulmelnikow (approved):
🙌🏼

PR #7160, opened 2021-10-16
Switch callbacks to `React.useCallback`

chris48s:
Tbh, I couldn't fully explain this but it seems this is what appeases the lint gods, which in turn unblocks upgrading gatsby. So that's good, I guess.
My instinct based on reading https://github.com/yannickcr/eslint-plugin-react/issues/3060 is that this is more desirable for us than disabling `allowFunctions` completely and less messy than peppering the code with `ignore` comments, but equally I'm no expert on React.

Refs https://github.com/badges/shields/pull/7088
Refs https://github.com/badges/shields/pull/6982


paulmelnikow:
Hmm, this change seems sort of opinionated and weird to me, and is definitively less readable, but time will tell…

These are a couple of articles which express views on this:
- Supportive: https://attardi.org/why-we-memo-all-the-things/
- Mixed? https://kentcdodds.com/blog/usememo-and-usecallback

Certainly fine with keeping it in.

calebcartwright:
More than happy to rely on your expertise around what you think will be best Paul. I think the only opinion the rest of us really had is that we couldn't wait indefinitely in the hopes that eslint-plugin-react would change their mind and revert to the old behavior.

paulmelnikow:
Yep, all makes sense. Though I dislike the loss of readability, I'm not really concerned about it. I'm not enough of a React expert to have a personal opinion; all I can do is read various articles and sort through which arguments I find the most convincing. So happy to leave it in, and glad you are pushing things forward 👍🏼

calebcartwright:
Sounds good :+1: My sense is that with the type of UI we have, it probably doesn't make a difference one way or another from a user experience and performance perspective, so I'm good with prioritizing whatever will be best for code readability and maintenance

PR #7170, opened 2021-10-18
fix [github] release/tag/download schema

### LINKED ISSUE #7169: GitHub release (latest by date) returns invalid response data

shields.io

### 🐞 Description

GitHub release (latest by date) bagage doesn't work. Invalid response data instead of release version.

### 🔗 Link to the badge

https://img.shields.io/github/v/release/truemail-rb/truemail-rack

### 💡 Possible Solution

_No response_

chris48s:
closes #7169
by default `Joi.string()` doesn't allow `''`

### DIFF
--- a/services/github/github-common-release.js
+++ b/services/github/github-common-release.js
@@ -12,7 +12,7 @@ const releaseInfoSchema = Joi.object({
     })
     .required(),
   tag_name: Joi.string().required(),
-  name: Joi.string().allow(null),
+  name: Joi.string().allow(null).allow(''),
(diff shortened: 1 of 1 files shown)


chris48s:
The failing service test is because the test is looking at the number of downloads for a release which currently has 0 downloads :laughing:  so https://img.shields.io/github/downloads-pre/photonstorm/phaser/latest/total is failing the `isMetric` validator. I just downloaded the release a few times so it should probably pass if we restart it in a bit and the counter updates

PR #7163, opened 2021-10-16
Add a render helper for downloads badges, run [amo ansible apm chromewebstore clojars conda crates docker dub eclipse gem githubdownloads]

calebcartwright:
We've got a quite a large number of download count related badges, and as I'm sure folks can imagine (and see in the diff), we've got a good deal of duplication around rendering concerns because we don't have a rendering helper like we do for others (e.g. version badges). In my opinion there's a lot of opportunity for encapsulation and reuse, both with our existing badges and to make things simpler for any future such badges

This PR both creates the render helper for download count badges, and includes the updates to the first batch of service classes to take advantage. Once the PR review app is up I'll post a comment that includes before/after badges between the review app and what we do today as an additional step to ensure there's no changes introduced.

PR #7210, opened 2021-10-29
use downloads renderer in next batch of classes, run [hexpm homebrewdownloads jenkinsplugininstalls jetbrainsdownloads jsdelivr modrinth]]

calebcartwright:
Continuation from #7163

### DIFF
--- a/services/downloads.js
+++ b/services/downloads.js
@@ -18,6 +18,8 @@ import { metric } from './text-formatters.js'
  *    value instead of the color being based on the count of downloads
  * @param {string} [attrs.messageSuffixOverride] If provided then the badge message will
  *    will have this value added to the download count, separated with a space
+ * @param {string} [attrs.versionedLabelPrefix] If provided then the badge label will use
+ *    this value as the prefix for versioned badges, e.g. `foobar@v1.23`. Defaults to 'downloads'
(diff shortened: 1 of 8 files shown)


services/downloads.js
```diff
@@ -27,6 +29,7 @@ function renderDownloadsBadge({
   labelOverride,
   colorOverride,
   messageSuffixOverride,
+  versionedLabelPrefix = 'downloads',
```
calebcartwright, line 32: Added to support use cases where classes want to support tag/version specific download counts but are using a label other than `downloads`

services/jenkins/jenkins-plugin-installs.service.js
```diff
@@ -23,22 +22,6 @@ const schemaInstallationsPerVersion = Joi.object()
...
-    }
-  }
-
-  static _getLabel(version) {
-    if (version) {
-      return `installs@${version}`
-    } else {
-      return 'installs'
-    }
-  }
```
calebcartwright, line 40: Trying to keep the diff focused on the refactoring to utilize the new render helper, but felt inclined to update this one with our more canonical fetch/transform/render separation of concerns and applied some broader refactoring to that end

chris48s (approved):

Issue #6268, opened 2021-03-12
Send better user-agent values


chris48s:
:clipboard: **Description**

At the moment we send a hard-coded user-agent header.

It would be nice if:
- Shields.io itself sent a different value than self-hosted instances send by default (obviously customisation/spoofing is possible)
- The user agent contained some kind of meaningful indication of the server version

One useful thing we can do is write the version to a file we can read from when we prep a release.
We would probably also want to set a default user agent and then make it configurable via an env var we can set in production.

A bit more discussion on this: https://github.com/badges/shields/pull/6256#discussion_r591810269

PR #7212, opened 2021-10-31
migrate github badges to use got instead of request; affects [github librariesio]

chris48s:
Refs #4655

This PR deals with migrating request --> got in:

- https://github.com/badges/shields/blob/master/core/base-service/base.js
- https://github.com/badges/shields/blob/master/services/github/github-api-provider.integration.js
- https://github.com/badges/shields/blob/master/services/suggest.js

A lot of what is changing here is just updating tests

I've also made a couple of tweaks to libraries.io in this Pr. These changes are not _strictly_ related but I spotted a couple of minor things there while using that code as a base for porting the github token pooling from callbacks to async/await

---

Remaining uses of request not addressed in this PR:

- https://github.com/badges/shields/blob/master/services/github/auth/acceptor.js
- https://github.com/badges/shields/blob/master/services/php-version.js
- https://github.com/badges/shields/blob/master/core/legacy/regular-update.js
- https://github.com/badges/shields/blob/master/core/base-service/legacy-request-handler.js

### DIFF
--- a/core/base-service/base.js
+++ b/core/base-service/base.js
@@ -450,8 +450,7 @@ class BaseService {
           const namedParams = namedParamsForMatch(captureNames, match, this)
           const serviceData = await this.invoke(
             {
-              sendAndCacheRequest: fetcher,
-              sendAndCacheRequestWithCallbacks: request,
+              sendAndCacheRequest: fetcher, // TODO: rename sendAndCacheRequest
               githubApiProvider,
               librariesIoApiProvider,
               metricHelper,
--- a/core/base-service/got.js
+++ b/core/base-service/got.js
@@ -64,7 +64,8 @@ async function sendRequest(gotWrapper, url, options) {
   }
 }
 
-function fetchFactory(fetchLimitBytes) {
+const TEN_MB = 10485760
+function fetchFactory(fetchLimitBytes = TEN_MB) {
   const gotWithLimit = got.extend({
     handlers: [
       (options, next) => {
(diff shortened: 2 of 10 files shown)


core/base-service/base.js
```diff
@@ -450,8 +450,7 @@ class BaseService {
           const namedParams = namedParamsForMatch(captureNames, match, this)
           const serviceData = await this.invoke(
             {
-              sendAndCacheRequest: fetcher,
-              sendAndCacheRequestWithCallbacks: request,
+              sendAndCacheRequest: fetcher, // TODO: rename sendAndCacheRequest
```
chris48s, line 453: Note for futureChris. This needs doing, but not now..

core/base-service/got.js
```diff
@@ -64,7 +64,8 @@ async function sendRequest(gotWrapper, url, options) {
   }
 }
 
-function fetchFactory(fetchLimitBytes) {
+const TEN_MB = 10485760
+function fetchFactory(fetchLimitBytes = TEN_MB) {
```
chris48s, line 68: There are a couple of places in this PR where we want to call this outside of the context of the `Server` object. This means we can't use the config/settings because we only parse them when we construct a `Server` object. For now I have decided to just give this param a default value and deal with it that way, but slightly more widely I'm thinking of looking at how we could decouple settings parsing from bootstrapping the server at some point. It is fringe, but another issue where it would be useful to read the config settings outside the context of the server object would be #6268 . Another option would be to say HTTP calls made outside the server are 'out of scope' for these settings. It is something for another day though. I'll continue to chew it over as I progress #4655

calebcartwright (approved):
LGTM

#7037 is really starting to drift/getting set up for some fun conflicts, but there's still some big TBDs over there and no sense blocking on it

PR #7215, opened 2021-11-02
Continue request --> got migration; affects [travis wordpress myget jenkins-plugin node nuget]

### LINKED ISSUE #7214: TypeError: githubApiProvider.request is not a function
Sentry Issue: [SHIELDS-73](https://sentry.io/organizations/shields/issues/2761525544/?referrer=github_integration)

```
TypeError: githubApiProvider.request is not a function
  File "/app/services/php-version.js", line 237, in request
    githubApiProvider.request(request, url, {}, cb),
  File "/app/core/legacy/regular-update.js", line 44, in regularUpdate
    request(url, options, (err, res, buffer) => {
  ?, in new Promise
  File "/app/services/php-version.js", line 221, in getPhpReleases
    return promisify(regularUpdate)({
  File "/app/services/travis/travis-php-version.service.js", line 83, in TravisPhpVersion.transform
    await getPhpReleases(this._githubApiProvider)
...
(5 additional frame(s) were not displayed)
```

chris48s:
Closes #7214
Refs #4655

This PR migrates `regularUpdate()` from request and callbacks to got and async/await and all the knock-on effects of that.

### DIFF
--- a/core/legacy/regular-update.js
+++ b/core/legacy/regular-update.js
@@ -1,93 +1,62 @@
-import requestModule from 'request'
-import { Inaccessible, InvalidResponse } from '../base-service/errors.js'
+/**
+ * @module
+ */
+
+import { InvalidResponse } from '../base-service/errors.js'
+import { fetchFactory } from '../../core/base-service/got.js'
+import checkErrorResponse from '../../core/base-service/check-error-response.js'
+const fetcher = fetchFactory()
 
 // Map from URL to { timestamp: last fetch time, data: data }.
 let regularUpdateCache = Object.create(null)
 
-// url: a string, scraper: a function that takes string data at that URL.
-// interval: number in milliseconds.
-// cb: a callback function that takes an error and data returned by the scraper.
-//
-// To use this from a service:
-//
-// import { promisify } from 'util'
-// import { regularUpdate } from '../../core/legacy/regular-update.js'
-//
-// function getThing() {
-//   return promisify(regularUpdate)({
-//     url: ...,
-//     ...
-//   })
-// }
-//
-// in handle():
-//
-// const thing = await getThing()
-
-function regularUpdate(
-  {
-    url,
-    intervalMillis,
-    json = true,
-    scraper = buffer => buffer,
-    options = {},
-    request = requestModule,
-  },
-  cb
-) {
+/**
+ * Make a HTTP request using an in-memory cache
+ *
+ * @param {object} attrs Refer to individual attrs
+ * @param {string} attrs.url URL to request
+ * @param {number} attrs.intervalMillis Number of milliseconds to keep cached value for
+ * @param {boolean} [attrs.json=true] True if we expect to parse the response as JSON
+ * @param {Function} [attrs.scraper=buffer => buffer] Function to extract value from the response
+ * @param {object} [attrs.options={}] Options to pass to got
+ * @param {Function} [attrs.requestFetcher=fetcher] Custom fetch function
+ * @returns {*} Parsed response
+ */
+async function regularUpdate({
+  url,
+  intervalMillis,
+  json = true,
+  scraper = buffer => buffer,
+  options = {},
+  requestFetcher = fetcher,
+}) {
   const timestamp = Date.now()
   const cached = regularUpdateCache[url]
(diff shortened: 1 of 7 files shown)


core/legacy/regular-update.js
```diff
@@ -1,93 +1,62 @@
...
-    }
 
-    if (res.statusCode < 200 || res.statusCode >= 300) {
-      cb(
-        new InvalidResponse({
-          prettyMessage: 'intermediate resource inaccessible',
-        })
-      )
-    }
+  const { buffer } = await checkErrorResponse({})(
```
chris48s, line 39: I decided to replace the bespoke error handling here with just `checkErrorResponse()`. I don't think it was really buying us anything

core/legacy/regular-update.js
```diff
@@ -1,93 +1,62 @@
-import requestModule from 'request'
```
chris48s, line 1: This file is the last file left in `core/legacy/` and it has no unit tests although it is not _completely_ untested because the service tests for all the services in the title exercise this code.

My instinct is that having migrated this to got and async/await we should move this out of `core/legacy/` and write a proper test suite for it. Before I go ahead with that, it would be useful to understand why this is in legacy. Maybe @paulmelnikow  can chime in on this one. Does this sufficiently modernise it, or is there another reason we want to ditch this?

core/legacy/regular-update.js
```diff
@@ -1,93 +1,62 @@
...
+      })
     }
+  } else {
+    reqData = buffer
+  }
 
-    regularUpdateCache[url] = { timestamp, data }
-    cb(null, data)
-  })
+  const data = scraper(reqData)
```
calebcartwright, line 57: it's a small diff but admittedly my eyes are having difficulty following it, though curious if there's any potential behavioral concerns since the `scraper` call site is no longer happening within a `try` block?

core/legacy/regular-update.js
calebcartwright: i think that's reasonable, assuming we don't have a large bucket somewhere of these intermediate errors that are providing helpful context (I don't think I've ever seen one)

calebcartwright (commented):
couple small questions, otherwise lgtm :+1:

core/legacy/regular-update.js
chris48s: Yeah this took me a while to unpick too.

You'll notice that the old `regularUpdate` function takes a callback function as a param (`cb`) in the sig, but when we're invoking `regularUpdate()` we are not actually passing it a callback. This is because we're wrapping it in [`promisify()`](https://github.com/nodejs/node/blob/657fb9a77ca36f729da484f55899dad7a13759b0/lib/internal/util.js#L317-L365) from the standard library. One of the things this does is it passes our function a callback that [rejects the promise if we pass it an error](https://github.com/nodejs/node/blob/657fb9a77ca36f729da484f55899dad7a13759b0/lib/internal/util.js#L341-L343)

So

```js
function regularUpdate(params, cb) {
  try {
    // do a thing
  } catch (e) {
    cb(e)
    return
  }
}
```

is there to catch any exception we throw and pass it into the callback function, but its not actually handling it pre-se - its just allowing us to catch it when we resolve the promise (maybe "bubbling" it is the right word for this?). i.e: it allows us to do:

```js
try:
  await promisify(regularUpdate)(params)
} catch (e) {
  // either handle the error cleanly, or ignore it and leave it sentry
}
```

Now that `regularUpdate` is an async function, we can safely just throw in an async function, so

```js
function regularUpdate(params, cb) {
  try {
    // do a thing
  } catch (e) {
    cb(e)
    return
  }
}
```

in callback-land can just be replaced with

```js
async function regularUpdate(params) {
  // do a thing
}
```

in async/await-land because:

```js
try:
  await regularUpdate(params)
} catch (e) {
  // either handle the error cleanly, or ignore it and leave it sentry
}
```

is already a thing.

calebcartwright: got it, thanks for the explanation!

core/legacy/regular-update.js
chris48s: I'm going to go ahead and merge this so we can get #7214 resolved, but it would be useful to get some input on this anyway and I will follow up

PR #7233, opened 2021-11-05
remove request from legacy request handler; affects [librariesio github]

chris48s:
refs #4655 
delete a bunch of stuff we don't need any more


calebcartwright (approved):
Was the utilization of some of these (caching in particular) eliminated as part of the more holistic got migration, or was it already dead code?

chris48s:
Yeah its a good question. Now I've done this, I think this was actually dead code before https://github.com/badges/shields/pull/6160 although I haven't pinpointed the exact commit where it became unreachable. Working on https://github.com/badges/shields/issues/4655 has just forced me to actually read and understand this function a bit :)

PR #7248, opened 2021-11-11
migrate request --> got in [github] auth acceptor

chris48s:
Refs #4655

This PR moves our last remaining module that calls request to use got instead. That said, I have quite a few [mop-up tasks](https://github.com/badges/shields/issues/4655#issuecomment-958152213) to fully "complete" the job. We're making progress though :)

A lot of this diff is just indenting changes due to callback --> async/await. Also improved the tests slightly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "prom-client": "^14.0.1",
     "qs": "^6.10.1",
     "query-string": "^7.0.1",
-    "request": "~2.88.2",
     "semver": "~7.3.5",
     "simple-icons": "5.21.1",
     "webextension-store-meta": "^1.0.4",
--- a/services/github/auth/acceptor.js
+++ b/services/github/auth/acceptor.js
@@ -1,7 +1,8 @@
 import queryString from 'query-string'
-import request from 'request'
-import { userAgent } from '../../../core/base-service/got.js'
+import { fetchFactory } from '../../../core/base-service/got.js'
 import log from '../../../core/server/log.js'
+
+const requestFetcher = fetchFactory()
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -60,7 +60,6 @@
     "prom-client": "^14.0.1",
     "qs": "^6.10.1",
     "query-string": "^7.0.1",
-    "request": "~2.88.2",
```
chris48s, line 63: :tada:

services/github/auth/acceptor.js
```diff
@@ -23,61 +24,64 @@ function setRoutes({ server, authHelper, onTokenAccepted }) {
...
+      form: {
         // TODO The `_user` and `_pass` properties bypass security checks in
         // AuthHelper (e.g: enforceStrictSsl and shouldAuthenticateRequest).
         // Do not use them elsewhere. It would be better to clean
         // this up so it's not setting a bad example.
         client_id: authHelper._user,
         client_secret: authHelper._pass,
         code: data.code,
-      }),
+      },
```
calebcartwright, line 46: guess `got` will handle the query param mappings for us?

calebcartwright (approved):
:+1:

services/github/auth/acceptor.js
chris48s: Yep. By default got will use URL encoding for the `form` object. It will also set the content-type header to `application/x-www-form-urlencoded`, but I left the custom header with the `;charset=UTF-8` in place.

The "a code is provided" test case decodes the request body with `queryString.parse()` to make assertions about it, so that behaviour is covered under test.

PR #7270, opened 2021-11-13
Remove requestOptions2GotOptions compatibility layer

chris48s:
Refs #4655

This PR is a bit of a beast but I've broken it down into smaller chunks as far as possible to make it easier to follow. Unfortunately 49476e30 is still really big but it was unavoidable.


chris48s:
Service test failures covered by known issues

core/base-service/auth-helper.js
```diff
@@ -74,7 +74,7 @@ class AuthHelper {
   }
 
   static _isInsecureSslRequest({ options = {} }) {
-    const { strictSSL = true } = options
+    const strictSSL = options?.https?.rejectUnauthorized ?? true
```
calebcartwright, line 77: TIL Node supports an optional chaining operator

core/base-service/got.js
```diff
@@ -27,13 +28,6 @@ function requestOptions2GotOptions(options) {
     delete requestOptions.qs
   }
 
-  if ('strictSSL' in requestOptions) {
-    gotOptions.https = {
-      rejectUnauthorized: requestOptions.strictSSL,
-    }
-    delete requestOptions.strictSSL
-  }
-
```
calebcartwright, line 34: if the overall mapping was going to be long-lived i'd wonder if we should keep some kind of check given the object vs. boolean type of the top level key, but given that we're moving away from needing this I suspect that doesn't matter

core/base-service/base.js
```diff
@@ -108,11 +108,14 @@ class BaseService {
    *
    * See also the config schema in `./server.js` and `doc/server-secrets.md`.
    *
-   * To use the configured auth in the handler or fetch method, pass the
-   * credentials to the request. For example:
-   * - `{ options: { auth: this.authHelper.basicAuth } }`
-   * - `{ options: { headers: this.authHelper.bearerAuthHeader } }`
-   * - `{ options: { searchParams: { token: this.authHelper._pass } } }`
```
calebcartwright: Were these inaccurate already/currently, independent of the got migration? Interesting if so, seems like we've discovered a couple inaccuracies along the way as part of this migration.

calebcartwright (dismissed):
Couple inline comments, but looks good and ready to go to me

calebcartwright:
I think most of our authenticated requests get skipped in CI since we don't typically have the relevant auth. I don't think this would be necessary based on the code diff, but do you think there's any authenticated-style tests we should run locally or in staging?

core/base-service/auth-helper.js
chris48s: Yeah it was introduced in Node 14. This might be the first time we've used it.

core/base-service/base.js
chris48s: Yes. I updated this from
`{ options: { qs: { token: this.authHelper._pass } } }`
to
`{ options: { searchParams: { token: this.authHelper._pass } } }`
in 
49476e3

Then I was going to update the `auth` key and realised that it just didn't really make sense.

chris48s:
There's probably 3 categories of services that use auth:

- Services that require auth to work at all
- Services that allow low volume anonymous usage but need a key to get a higher rate limit
- Services that allow access to public instances/resources with no auth but require auth to access private instances/resources

We've got credentials set for

- GitHub
- Libraries.io
- OBS
- SL Insight
- Twitch
- Wheelmap
- Youtube

in CI. I think that actually covers all the services that require auth so we shouldn't be skipping any services in CI as far as I'm aware :crossed_fingers:  but if there are lets identify them and set a key in CI. There's another case where we only test against public resource but in principle you can also set credentials to access a private resource which isn't tested.

calebcartwright (approved):

PR #7271, opened 2021-11-14
upgrade to node 16

### LINKED ISSUE #7130: Deploy on Node 16
:clipboard: **Description**

Node 16 has been generally available for ~6 months now and includes a number of performance improvements we may benefit from. We've had the full test suite consistently passing for some time and we're now starting to see dependency upgrades which require us to make the jump. I see no reason not to :rocket: in production in the near future.

chris48s:
Closes #7130


.circleci/config.yml
```diff
@@ -137,33 +137,33 @@ package_steps: &package_steps
...
-      - image: circleci/node:14
+      - image: circleci/node:16
 
     <<: *main_steps
 
-  main@node-16:
+  main@node-17:
     docker:
-      - image: circleci/node:16
+      - image: circleci/node:17
```
chris48s, line 146: I don't think we should actually aim to deploy on Node 17 at any point. There won't be a node 18 build available until April 2022 but I reckon me might as well keep a 'next' build in play and then switch it to 18 next year.

chris48s:
Assuming this gets approved, we'll need to update the branch protection rules to merge it

calebcartwright (approved):

PR #7273, opened 2021-11-14
fix [spigetdownloadsize] service test

calebcartwright:
Thought this would require a bunch of digging in an unfamiliar service to find an appropriate test target, but fortunately the example listing we had was also transitioned from the direct/hosted kind to the externally hosted kind, so a simple swap suffices for both the test and the example


chris48s (approved):

services/spiget/spiget-download-size.tester.js
```diff
@@ -6,8 +6,8 @@ t.create('EssentialsX (hosted resource)')
   .get('/771.json')
   .expectBadge({ label: 'size', message: isFileSize })
 
-t.create('Pet Master (external resource)').get('/15904.json').expectBadge({
-  lavel: 'size',
```
chris48s, line 10: after seeing this I raised #7276

PR #7275, opened 2021-11-15
increase timeout for [tokei] gitlab tests

calebcartwright:
These have been failing repeatedly, albeit somewhat intermittently with occasional passes, for a while. The test target has more than 1 million lines of code so I'm thinking this _may_ be timeout related (able to reproduce a couple timeouts locally). Want to try increasing the limit for a while to see if it helps stabilize these tests

https://app.circleci.com/pipelines/github/badges/daily-tests/1334/workflows/f5af6f9a-364f-4413-91c5-ca786916d963/jobs/2236/tests#failed-test-9


calebcartwright:
or not... think the divergent results between the two service test jobs is fairly indicative of the transient nature of the issue. i still feel like there's a pattern where the initial request fails and subsequent ones succeed, presumably due to caching or some other benefit from that initial request :thinking:

calebcartwright:
TBH while I was able to reproduce the failures (intermittently) with the current test target, I was unable to determine a root cause. My partially informed suspicion is that there's something going on with the particular repo we are targeting given that it's got well north of 1 million lines of code, and changing the test target as proposed here has, so far, appeared to provide a consistently passing test.

I realize that changing the test target is likely a form of a bandaid, and that if we've got a root issue somewhere on our side then that's certainly still going to be present even if we go ahead with this proposed change of test target. However, I've already spent too much time on this and can't allocate any more bandwidth to it, so I'd prefer we knock out the test issue and then circle back to it if it in fact turns out to be a real problem users are experiencing with the badge (which doesn't seem to be the case given the lack of reports)

chris48s:
> if we've got a root issue somewhere on our side

I think its more like a root issue with the upstream service tbh. Checking out a whole GH repo and counting the lines of code in it takes a long time. This is based on observation from the outside, but if the service has the result cached upstream, you get a result fast. If it doesn't, you get a result slow. I accept this is worse for really big repos. As an extreme example, if you request https://tokei.rs/b1/github/torvalds/linux when it is cached you'll get a response back instantly. If it is not cached you're going to be waiting a looooong time for that SVG (you might actually get a few `408 Request Timeout` first). If we think changing the test target to a smaller repo makes the tests fail less frequently I'm happy to :+1: it but I think pass/fail probably has more to do with cached/not cached. Probably if we want to make the test work consistently we'd get more mileage from a retry/backoff as there's always going to be times when the result is not served from cache.

chris48s (approved):
either way, merging this probably won't make things any worse

calebcartwright:
Agreed, and had the same general suspicion. Felt like I hit a brick wall though being unable to produce anything other than a 200 or 400 response code from the upstream service. A 408 would certainly explain the `invalid` response we're seeing though so that makes me feel better 👍

PR #7281, opened 2021-11-17
Support "github.io" group id for jitpack.io version badge

pweisenburger:
Currently, the jitpack.io version badge only supports the "com.github" group id for GitHub projects. This pull request adds support for "io.github" as group id. On the jitpack.io side, support was added in jitpack/jitpack.io#1102. The version badge now uses the "io.github" variant if selected. If just "github" is selected, "com.github" is used as the group id (which is compatible with existing badges).

The pull request also enables options that contain dots in the drop-down list of the web frontend.

### DIFF
--- a/frontend/lib/pattern-helpers.ts
+++ b/frontend/lib/pattern-helpers.ts
@@ -3,9 +3,9 @@ import { parse } from 'path-to-regexp'
 // Given a patternToRegex `pattern` with multiple-choice options like
 // `foo|bar|baz`, return an array with the options. If it can't be described
 // as multiple-choice options, return `undefined`.
-const basicChars = /^[A-za-z0-9-]+$/
+const basicChars = /^[A-Za-z0-9-.]+$/
 export function patternToOptions(pattern: string): string[] | undefined {
-  const split = pattern.split('|')
+  const split = pattern.replace(/\\(.)/g, '$1').split('|')
   if (split.some(part => !part.match(basicChars))) {
     return undefined
   } else {
--- a/services/jitpack/jitpack-version.service.js
+++ b/services/jitpack/jitpack-version.service.js
@@ -12,7 +12,7 @@ export default class JitPackVersion extends BaseJsonService {
 
   static route = {
     base: 'jitpack/v',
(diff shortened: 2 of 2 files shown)


calebcartwright:
Thank you for submitting a PR! However, I think I'd prefer to see a different approach that doesn't require modification of the patterns, much in the same way that we didn't do so for the `com.github` group id. I think adding a new option to the route param like `githubio`, `iogithub`, `github-io`, etc. should suffice, which we can then map accordingly.

Also, we should add a new test that has the io variant of the group id

pweisenburger:
Thanks! However, it's unclear to me what a new option (like `githubio`) would mean when the `vcs` part of the pattern is set to, say, `bitbucket`. It seems to me that both the new option and vcs are not orthogonal concepts but describe, in fact, the exact same parameter. But maybe I got the suggestion wrong.

calebcartwright:
> It seems to me that both the new option and vcs are not orthogonal concepts but describe, in fact, the exact same parameter. But maybe I got the suggestion wrong.

Yes, I think you've misunderstood my suggestion of adding a new option/variant to the _existing_ route parameter as meaning a _new_ route parameter should be added. I was suggesting the former, and we definitely do not want to do the latter.

The more I've thought about this though, the less inclined I am to have us continue to surface the group id via a "version control system" parameter which we then attempt to map to an actual group id. All that does is save the user from typing a couple keys and creating blocking problems for new group ids as in the case you reported here.

Would you have any concerns if we pivoted to allowing the user to just provide the group id directly via a single route parameter?

This would be an adjustment to the badge route, but we have a widely used mechanism to allow us to adjust routes without breaking existing badges via our `redirector`'s, one of which is already in place for our JitPack badge so it would be pretty easy to add another entry

PyvesB:
@pweisenburger have you had a chance to read through @calebcartwright's latest comment? It would be great to get this landed. 😉

PyvesB:

If someone else wants to pick this up, you can simply fetch the commits by running `git fetch origin pull/7281/head:pr-7281`. Address the pending review comments and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

PR #7309, opened 2021-11-22
Send better user-agent values (and got config changes)

### LINKED ISSUE #6268: Send better user-agent values

chris48s:
Refs #4655
Closes #6268

There is not loads of changed code in this PR, but there's a couple of things going on in here which are kind of linked and kind of not.

- One is a bit of cleanup. The request --> got work left us with this slightly clunky thing where `got.js` exports a `fetchFactory` which takes `fetchLimitBytes` and returns a function. That allows us to inject `fetchLimitBytes` from config in the server but then sometimes it is useful to construct it outside the context of the server object.
- We've also had this long-standing issue #6268 which I had a look at before and was also a bit clunky/fiddly for the same reason.

I did have a look at completely decoupling config parsing from the server, but what I realised was that because we can set some of the args at runtime (see https://github.com/badges/shields/blob/95a439a1cc9a912371ae0bce0cf0f7ed8862f719/server.js#L28-L33 ) this didn't seem like the right approach as the whole config object doesn't depend exclusively on the yml/env vars and we'd have to double validate the config at server start time.
The approach I settled on was making a subset of config available outside the context of the server object, which seemed like the right tradeoff.

The other bit of this is adding a feature. Basically before this PR, every instance of shields (our own, self-hosted, dev copies) sends the exact same user agent value (`Shields.io/2003a`, for..reasons).
As of this PR:
- The default user agent is `shields (self-hosted)/dev`
- We (or users who self-host their own instance) can override the base value (the part before the slash)
- If either `HEROKU_SLUG_COMMIT` (we have this set in production) or `DOCKER_SHIELDS_VERSION` (this PR adds it to the image) are set, that will be used for the second part of the UA string (after the slash)
- If neither of those vars are set it will be `/dev`

This should mean in most cases the userAgent string will be reasonably meaningful.


core/base-service/got.js
```diff
@@ -52,4 +55,6 @@ function fetchFactory(fetchLimitBytes = TEN_MB) {
   return sendRequest.bind(sendRequest, gotWithLimit)
 }
 
-export { fetchFactory, userAgent }
+const fetch = _fetchFactory()
+
+export { fetch, _fetchFactory }
```
chris48s, line 60: This is not my favourite solution ever. Testing this with different values for `fetchLimitBytes` is tricky, but I do prefer the ergonomics of exporting `fetch` "fully formed" as it were. Exporting an underscore prefixed function we can test from the module seems like a reasonable tradeoff but I'm not wedded to it.. we could go back.

core/base-service/got-config.js
```diff
@@ -0,0 +1,26 @@
...
+const schema = Joi.object({
+  fetchLimit: fileSize,
+  userAgentBase: Joi.string().required(),
+}).required()
+const config = configModule.util.toObject()
+const publicConfig = Joi.attempt(config.public, schema, { allowUnknown: true })
+
+const fetchLimitBytes = bytes(publicConfig.fetchLimit)
+
+function getUserAgent(userAgentBase = publicConfig.userAgentBase) {
```
chris48s, line 15: We will need to set `USER_AGENT_BASE=Shields.io` in prod before we deploy

core/base-service/got.js
calebcartwright: > Exporting an underscore prefixed function we can test from the module seems like a reasonable tradeoff but I'm not wedded to it.. we could go back.

We do the same in plenty of places elsewhere so I've no concerns with it

services/validators.js
```diff
@@ -17,3 +17,6 @@ export const optionalDottedVersionNClausesWithOptionalSuffix =
 // TODO This accepts URLs with query strings and fragments, which for some
 // purposes should be rejected.
 export const optionalUrl = Joi.string().uri({ scheme: ['http', 'https'] })
+export const fileSize = Joi.string()
+  .regex(/^[0-9]+(b|kb|mb|gb|tb)$/i)
+  .required()
```
calebcartwright, line 22: I guess we can't really consolidate this pattern with the one we already use for test validation against our badges that display file size, but perhaps worth something like an inline comment to indicate that this is really to reuse the pattern we use for the Shields server config as opposed to API response or badge content? I could easily picture myself reviewing a hypothetical future PR that adds a file size badge for some new service Foo, seeing `fileSize` used in the tests and thinking all was well.

https://github.com/badges/shields/blob/d2bde17fd6c1d50156e3c5f1f328940af420fc16/services/test-validators.js#L96-L98

calebcartwright (dismissed):
I'm good with the changes, both code and strategic, and approving accordingly so you can move forward if you'd like. One minor inline item that could be ignored or handled in a follow up, but happy to re-:+1: if you decide to make any other changes

calebcartwright (approved):

chris48s:
I've set the env var, deployed and confirmed it is working in production by using the endpoint badge to make a request to an endpoint where I can see the header.
Hopefully this is all fine and I think it is a good change, but there is this slight worry at the back of my mind that somewhere out there in the world is a rate limit or deny list with a specific exception for `Shields.io/2003a` hard-coded which is going to break something now that we're sending a different suffix each time. We will have to see..

calebcartwright:
> but there is this slight worry at the back of my mind that somewhere out there in the world is a rate limit or deny list with a specific exception for Shields.io/2003a hard-coded which is going to break something now that we're sending a different suffix each time

This was in the back of my mind as well, though I actually think this change is perhaps the best way forward to surface any potential issues. I don't think we can (or should) perpetually pin a user agent, and it feels like changing it as we've done here may be the only way for us to truly discover any such cases, which will then allow us to figure out next steps

PR #7319, opened 2021-11-28
Clean up cache module; affects [feedz jenkinsplugin myget node nuget packagist travis wordpress]

### LINKED ISSUE #4655: migrate away from request

chris48s:
Closes #4655
Refs https://github.com/badges/shields/pull/7215#discussion_r741433020

This PR finishes up the last bit of tangential follow up from the refactoring work done under #4655

I opted not to bother taking on another cache dependency for this. I've tidied things up a bit, made the terminology a bit more normal and added some tests but I've pretty much left the implementation alone.

### DIFF
--- a/core/base-service/resource-cache.js
+++ b/core/base-service/resource-cache.js
@@ -6,6 +6,8 @@ import { InvalidResponse } from './errors.js'
 import { fetch } from './got.js'
 import checkErrorResponse from './check-error-response.js'
 
+const oneDay = 24 * 3600 * 1000 // 1 day in milliseconds
+
 // Map from URL to { timestamp: last fetch time, data: data }.
 let resourceCache = Object.create(null)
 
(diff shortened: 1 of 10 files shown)


calebcartwright (approved):
lgtm

Issue #7161, opened 2021-10-16
Finalize migration to squint


chris48s:
We've been running this four a couple of weeks and all seems to be well so far. @calebcartwright  - I had a few points swirling around in my head so I wanted to get them noted down somewhere. Can you edit this post to add any stuff that I haven't thought of to the list.

- [x] Shut down vercel app, ensure our vercel costs are zero
- [x] Update references to svg-to-image-proxy in docs:
    - https://github.com/badges/shields/blob/4a8c1d1d0fa01e5c043e172fa24614e180c7a1c3/doc/production-hosting.md
    - https://github.com/badges/shields/blob/7d452472defa0e0bd71d6443393e522e8457f856/SECURITY.md
    - https://github.com/badges/shields/blob/f4781804f612f91256f512502d9b920663d52eec/doc/self-hosting.md
- [x] Deprecate svg-to-image-proxy NPM package https://www.npmjs.com/package/svg-to-image-proxy
- [x] Archive svg-to-image-proxy repo https://github.com/badges/svg-to-image-proxy
- [x] Review/update DNS settings in Cloudflare (e.g. http -> https, caching, etc.)
- [x] Turn off the Vercel app on the svg-to-image-proxy repo (and anywhere else we've got it turned on)

calebcartwright:
Probably need to review the DNS settings too as I think we had the Clourflare configuration largely hands off since Vercel/Now handled some of those concerns internally.

Is there a need to deprecate the package necessarily? I know we haven't exactly been keeping up with the maintenance which is arguably a good reason in and of itself; someone could always fork if they wanted to keep it going.

calebcartwright:
Actually, after some reflection I think archiving the repo and deprecating the package is a good call. It's a small simple project that didn't need much in the way of maintenance, but we weren't even keeping the dependencies up to date so I think the archival route is the right move to reflect the fact we aren't maintaining anymore.

chris48s:
### Cloudflare

At the moment we've got raster.shields.io set to "DNS only" in Cloudflare (i.e: no CDN). Any reason why we shouldn't switch this to "Proxied" (which would turn on downstream cache)? I guess this isn't a huge issue as we serve so little traffic but if it serves the right headers, we may as well enable it.

### Vercel

Vercel is still charging us $2.36 per month for.. something, despite the fact we're not serving any traffic from it.

[...]

PR #7348, opened 2021-12-04
update raster proxy docs

chris48s:
Refs https://github.com/badges/shields/issues/7161


calebcartwright (approved):
Technically squint runs as a native binary, but don't think it's worth mentioning here that folks could manually compile for any supported platform

PR #7352, opened 2021-12-07
Add config for testing status code in [website] service

OllieJC:
This change adds two optional parameters for the `website` service to specify the expected status code(s) of responses:
- `status_code_less_than` default = 310
- `status_code_more_than` default = 0

Added an additional test using the new parameters.

### DIFF
--- a/services/website-status.js
+++ b/services/website-status.js
@@ -5,6 +5,8 @@ const queryParamSchema = Joi.object({
   down_message: Joi.string(),
   up_color: Joi.alternatives(Joi.string(), Joi.number()),
   down_color: Joi.alternatives(Joi.string(), Joi.number()),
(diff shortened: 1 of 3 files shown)


calebcartwright:
Could you elaborate a bit on the "why" for this change, beyond the description of what's been changed? We don't require an associated issue and will certainly consider such PRs, but it's particularly useful for such a PR to provide the backing rationale in the absence of a plan/discussion spawning out of an issue.

calebcartwright:
I'm going to close this. In practice I can't see a scenario where a <310 response code for a HEAD request would need to be treated as an error/down, nor why any higher 3xx, 4xx, 5xx response should ever be treated as a success.

However, please feel free to respond to the previous question if you'd like us to reconsider

PR #7354, opened 2021-12-09
throw on unexpected keys in `expectBadge()`

### LINKED ISSUE #7276: `expectBadge()` should throw on unexpected keys
:clipboard: **Description**

This, for example, should fail:

https://github.com/badges/shields/blob/be013e175b0048f796817c431773d5a069a9a31d/services/spiget/spiget-download-size.tester.js#L10

chris48s:
Closes #7276


services/docsrs/docsrs.tester.js
```diff
@@ -14,5 +14,5 @@ t.create('Getting latest version works')
   .get('/rand/latest.json')
   .expectBadge({
     label: 'docs',
-    messsage: Joi.allow('passing', 'failing'),
+    message: Joi.allow('passing', 'failing'),
```
chris48s, line 17: :smile:

chris48s:
Test failures are unrelated but we need to do another triage of consistently failing tests

services/docsrs/docsrs.tester.js
calebcartwright: Ha. Spelling common words is surprisingly difficult sometimes

calebcartwright (approved):
Nice :+1:

services/docsrs/docsrs.tester.js
chris48s: tbh I was slightly surprised this only kicked up the one

PR #7374, opened 2021-12-15
change closed GitHub issue color to purple, run [GitHubIssueDetail]

### LINKED ISSUE #7372: GitHubIssueDetail Closed issues message color
GitHub's experience tiers have switched from displaying closed issues as red to purple (e.g. see our own closed issues https://github.com/badges/shields/issues?q=is%3Aissue+is%3Aclosed), ostensibly to avoid categorically connoting the failed/rejected/etc. theme of red.

I'm wondering if we should do the same for our issue status badge (PRs will remain the same since they already differentiate between merged/rejected) and curious what others think. I confess I'm still on the fence myself, but think I'm leaning towards changing to purple

calebcartwright:
Resolves #7372 

We handle the coloring of non-open PRs directly within the service's render function, so this change will only apply to closed issues


services/github/github-helpers.js
```diff
@@ -11,7 +11,7 @@ const documentation = `
 `
 
 function stateColor(s) {
-  return { open: '2cbe4e', closed: 'cb2431', merged: '6f42c1' }[s]
+  return { open: '2cbe4e', closed: '6f42c1', merged: '6f42c1' }[s]
```
chris48s, line 14: A closed (but not merged) PR is still presented as red. I think we need to split this into 2 functions

```js
function issueStateColor(s) {
  return { open: '2cbe4e', closed: '6f42c1' }[s]
}

function pullRequestStateColor(s) {
    return { open: '2cbe4e', closed: 'cb2431', merged: '6f42c1' }[s]
}
```

calebcartwright: good point, will update

chris48s (approved):

PR #7403, opened 2021-12-21
check engine and peer deps before installing

chris48s:
Not much code here, but it does require a bit of explanation :)

In our `package.json` we declare an engines key with our node/npm compatibility:
https://github.com/badges/shields/blob/24355a0773c4d4333399cbfcc4dec52cba62344f/package.json#L242-L245
By default if NPM installs a package which declares itself incompatible with our engines statement, NPM will just install anyway but emit an `EBADENGINE` warning. In general we tend to catch this because the tests fail but if the tests pass we either have to identify this ourselves from the changelog or the CI logs (dunno about anyone else, but I never read the install log for dependency bump PRs) or in some cases we might actually merge a package update which officially drops compatibility/testing with our target node version (probably not for a while now we are on 16 in prod, but point still stands).

NPM has a useful flag [`--engine-strict`](https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict) which will cause NPM to error (and fail the build) in this situation instead of emitting a warning, which seems a lot safer.

Another super-fun default NPM has is it will allow installation of packages with conflicting `peerDependencies`. Again, the default is install anyway and emit a warning. Again, there is a flag [`--strict-peer-deps`](https://docs.npmjs.com/cli/v8/using-npm/config#strict-peer-deps) which tells it to treat this as a failure instead of a warning.

So far so good. Unfortunately both of these flags work with `npm install` but [don't work with `npm ci`](https://github.com/npm/cli/issues/1219), so if we want to use `npm ci` and those flags we can use the `--dry-run` flag annd do `npm install --engine-strict --strict-peer-deps --dry-run` first (which will do the checks but not install anything) and then `npm ci` after if it passes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,9 @@ main_steps: &main_steps
 
     - run:
         name: Install dependencies
-        command: npm ci
+        command: |
+          npm install --engine-strict --strict-peer-deps --dry-run
+          npm ci
         environment:
           # https://docs.cypress.io/guides/getting-started/installing-cypress.html#Skipping-installation
           # We don't need to install the Cypress binary in jobs that aren't actually running Cypress.
@@ -47,7 +49,9 @@ integration_steps: &integration_steps
 
     - run:
         name: Install dependencies
-        command: npm ci
+        command: |
+          npm install --engine-strict --strict-peer-deps --dry-run
+          npm ci
         environment:
           CYPRESS_INSTALL_BINARY: 0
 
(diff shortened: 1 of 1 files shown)


chris48s:
hmm. Ran this locally and it passed. Hadn't counted on the node 17 builds failing, although that does make sense given our node requirement on this repo is ^16. I'm not super keen to make that >=16 as I think that will mean heroku deploys us on 17. Probably need to be a bit smarter :)
Going to convert this to draft for now and have a think

chris48s:
setting the options with env vars seems to be the right approach

calebcartwright (approved):
Thanks for looking at this one :+1: I feel like there was even a case fairly recently this would've caught/alarmed on and saved us some dep headaches, so I'm all for making this change

PR #7410, opened 2021-12-22
[YouTube] Drop support for removed dislikes

### LINKED ISSUE #7373: YouTube video view count: "invalid response data"

shields.io

### 🐞 Description

All YouTube video view counts shows `invalid response data` in a field of number of views.

### 🔗 Link to the badge

- https://github.com/Symbian9/SolveSpace-Daily-Engineering#showcases
  - https://img.shields.io/youtube/views/1uZgUpjTg18?style=social&logo=youtube
  - https://img.shields.io/youtube/views/KY95-xrbSNo?style=social&logo=youtube
  - https://img.shields.io/youtube/views/XrbDdYCB_Hs?style=social&logo=youtube
  - https://img.shields.io/youtube/views/-DfILpX7v-4?style=social&logo=youtube
  - https://img.shields.io/youtube/views/8kiYJYwhXHc?style=social&logo=youtube
  - https://img.shields.io/youtube/views/rYYxQSBI2c8?style=social&logo=youtube

### 💡 Possible Solution

_No response_

pzygielo:
- fixes #7373

### DIFF
--- a/services/youtube/youtube-base.js
+++ b/services/youtube/youtube-base.js
@@ -18,18 +18,11 @@ const schema = Joi.object({
         Joi.object({
           viewCount: nonNegativeInteger,
           likeCount: nonNegativeInteger,
-          dislikeCount: nonNegativeInteger,
           commentCount: nonNegativeInteger,
         }),
         Joi.object({
           viewCount: nonNegativeInteger,
           subscriberCount: nonNegativeInteger,
-        }),
-        Joi.object({
-          viewCount: nonNegativeInteger,
-          likeCount: nonNegativeInteger,
-          commentCount: nonNegativeInteger,
-          favoriteCount: nonNegativeInteger,
         })
       ),
     })
--- a/services/youtube/youtube-base.js
+++ b/services/youtube/youtube-base.js
@@ -24,6 +24,12 @@ const schema = Joi.object({
         Joi.object({
           viewCount: nonNegativeInteger,
           subscriberCount: nonNegativeInteger,
+        }),
+        Joi.object({
+          viewCount: nonNegativeInteger,
+          likeCount: nonNegativeInteger,
+          commentCount: nonNegativeInteger,
+          favoriteCount: nonNegativeInteger,
         })
       ),
     })
--- a/services/youtube/youtube-likes.service.js
+++ b/services/youtube/youtube-likes.service.js
@@ -62,11 +62,14 @@ export default class YouTubeLikes extends YouTubeVideoBase {
       id,
     })
     if (queryParams && typeof queryParams.withDislikes !== 'undefined') {
+      const likes = `${metric(statistics.likeCount)} 👍`
+      const dislikes =
+        statistics.dislikeCount !== undefined
+          ? `${metric(statistics.dislikeCount)} 👎`
+          : ''
       renderedBadge = {
         ...renderedBadge,
-        message: `${metric(statistics.likeCount)} 👍 ${metric(
-          statistics.dislikeCount
-        )} 👎`,
+        message: `${likes} ${dislikes}`,
       }
     }
     return renderedBadge
--- a/services/youtube/youtube-likes.service.js
+++ b/services/youtube/youtube-likes.service.js
@@ -1,24 +1,9 @@
-import Joi from 'joi'
-import { metric } from '../text-formatters.js'
 import { documentation, YouTubeVideoBase } from './youtube-base.js'
-
-const documentationWithDislikes = `
-  ${documentation}
-  <p>
-    When enabling the <code>withDislikes</code> option, 👍 corresponds to the number
-    of likes of a given video, 👎 corresponds to the number of dislikes.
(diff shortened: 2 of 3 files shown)


services/youtube/youtube-base.js
```diff
@@ -24,6 +24,12 @@ const schema = Joi.object({
         Joi.object({
           viewCount: nonNegativeInteger,
           subscriberCount: nonNegativeInteger,
+        }),
+        Joi.object({
+          viewCount: nonNegativeInteger,
+          likeCount: nonNegativeInteger,
+          commentCount: nonNegativeInteger,
+          favoriteCount: nonNegativeInteger,
```
calebcartwright, line 32: As far as I know we don't actually need the `favoriteCount` value for any of our badges, so instead of adding what is essentially a duplicate schema for `statistics`, let's instead make the `dislikeCount` field optional by changing from `nonNegativeInteger` to `optionalNonNegativeInteger` (which can be imported from the same validators module)

services/youtube/youtube-likes.service.js
```diff
@@ -62,11 +62,14 @@ export default class YouTubeLikes extends YouTubeVideoBase {
...
+      const dislikes =
+        statistics.dislikeCount !== undefined
+          ? `${metric(statistics.dislikeCount)} 👎`
+          : ''
       renderedBadge = {
         ...renderedBadge,
-        message: `${metric(statistics.likeCount)} 👍 ${metric(
-          statistics.dislikeCount
-        )} 👎`,
+        message: `${likes} ${dislikes}`,
```
calebcartwright, line 72: We already have the `withDislikes` query parameter that allows the user to control whether or not the dislikes are included on their badge. If we hit this block then the user has explicitly asked for the dislike portion to be included as well, and as such we should instead display `0 👎`.

Perhaps `${metric(statistics.dislikeCount || 0)} 👎` ?

services/youtube/youtube-likes.tester.js
```diff
@@ -22,7 +22,7 @@ t.create('video vote count')
   .expectBadge({
     label: 'likes',
     message: Joi.string().regex(
-      /^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) 👍 ([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) 👎$/
+      /^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) 👍$/
```
calebcartwright, line 25: Thanks for looking at the tests but as noted above we need to change the behavior a bit.

Additionally, we should instead add a new test for the scenario described in #7373. Given the specific use case of a missing `dislikeCount` field, it would probably be better for that to be a mocked test (https://github.com/badges/shields/blob/master/doc/service-tests.md#5-mocking-responses) given the likelihood for the upstream target to change/receive dislikes

calebcartwright (commented):
Thanks for working on this! Feedback left inline below

services/youtube/youtube-likes.service.js
pzygielo: But `0 👎` looks like there is no 👎, while in fact this is unknown. I was choosing between not presenting this piece or using `? 👎`. I'm against using `0` for the case.

calebcartwright: > while in fact this is unknown.

What is your evidence for this? Do you have some documentation to the API or an example target that includes a `"dislikeCount": 0` in the response?

calebcartwright: Actually, https://techcrunch.com/2021/11/10/youtube-is-removing-the-dislike-count-on-all-videos-across-its-platform/

Youtube _has_ dropped the dislike count from public view altogether, and that's not something we'll ever be able to see. I think we should actually remove dislikes across the board

pzygielo: I'd rather like this PR to only restore YT stats.
I think total dislikes removal could be the subject for the next one.

calebcartwright: We have a feature that was based on an attribute in the API response that no longer exists, which in turn has broken badges. I don't understand what you're suggesting be split, could you elaborate?

We need a resolution to the actual issue, i.e. our badge needs to account for the upstream change. Splitting that into multiple PRs with ~5 line diffs is unnecessary overhead for us

pzygielo: > Splitting that into multiple PRs with ~5 line diffs is unnecessary overhead for us

Understood.

Removed code IMO related to the dislikes.
Locally tested that the old links, with (now unused) parameter `withDislikes` render the no-parameterized version fine.

calebcartwright: Awesome thanks so much! Will give this a final pass here shortly

calebcartwright (approved):
LGTM, thanks for the updates and getting this one resolved!

PR #7411, opened 2021-12-22
[HackerNews] Show User Karma 

tapanchudasama:


services/hackernews/hackernews-user-karma.tester.js
```diff
@@ -0,0 +1,16 @@
...
+import { isMetric } from '../test-validators.js'
+
+export const t = await createServiceTester()
+
+t.create('valid repo -- live').get('/pg.json').expectBadge({
+  label: 'pg',
+  message: isMetric,
+})
+
+t.create('invalid user -- live')
```
chris48s, line 11: We don't need to add "-- live" to the tests. The service test runner will automatically prefix tests with `[live]` or `[mocked]` as applicable.

services/hackernews/hackernews-user-karma.service.js
```diff
@@ -0,0 +1,70 @@
...
+  static defaultBadgeData = {
+    label: 'HackerNews User Karma',
+    namedLogo: 'ycombinator',
+  }
+
+  static render({ karma, userid }) {
+    return {
+      label: userid,
+      message: metric(karma),
+      color: 'red',
```
chris48s, line 38: Why always red here? This gets used if you request the non-social version. Maybe `brightgreen` for positive, `red` for negative and `orange` for zero would make sense?

services/hackernews/hackernews-user-karma.service.js
```diff
@@ -0,0 +1,70 @@
...
+
+  static defaultBadgeData = {
+    label: 'HackerNews User Karma',
+    namedLogo: 'ycombinator',
+  }
+
+  static render({ karma, userid }) {
+    return {
+      label: userid,
+      message: metric(karma),
```
chris48s, line 37: I think the deal with karma is it can be either positive or negative, right? (Hence validating with `anyInteger`)
If so, `metric()` currently only works properly for positive numbers. I think it probably wouldn't be hard to get `metric()` to format negatives properly too though. Relevant code and tests are at:
https://github.com/badges/shields/blob/23c0406bedfc6930735e8f5ea75dfe34faf1f290/services/text-formatters.js#L53
https://github.com/badges/shields/blob/23c0406bedfc6930735e8f5ea75dfe34faf1f290/services/text-formatters.spec.js#L40
In terms of the validator, in almost all cases I'd want to validate that a metric badge is positive (if we've got a badge that reports number of stars, number of downloads, etc and its giving us a negative value I'd want that to fail - something is probably wrong) but I think changing `isMetric` to `isPositiveMetric` is going to be a pain because it is used in _all the places_ so maybe we make a `isMetricAllowNegative` validator for the edge case where we want this (probably just this one for now). Test validators are in:
https://github.com/badges/shields/blob/8a1c69ead63c312284cec6615e7de350a101fc0f/services/test-validators.js

chris48s (commented):
Thanks for making your first contribution. I've left a few comments but its looking in pretty good shape already

tapanchudasama:
> Thanks for making your first contribution. I've left a few comments but its looking in pretty good shape already

Actually its my second. 😄  First was [this](https://github.com/badges/shields/pull/6330). But this is a new account so it might seem new.

services/hackernews/hackernews-user-karma.service.js
tapanchudasama: Made the changes. Please have a look.

services/hackernews/hackernews-user-karma.service.js
```diff
@@ -0,0 +1,71 @@
...
+  static examples = [
+    {
+      title: 'HackerNews User Karma',
+      namedParams: { userid: 'drumstick' },
+      staticPreview: this.render({ userid: 'drumstick', karma: 15536 }),
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'HackerNews User Karma',
```
calebcartwright, line 30: This is a bit of a mouthful for a label, and I'd argue, going against the grain of the [branding/promotional clause in our badge specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md#example). There are a few cases where our default badge label has the service name included because we felt it was necessary to qualify/describe the value in the message.

However, I'm not sure that's the case here. Consider in reference our other social-style badges, including the Reddit Karma badge which is `combined karma` (I'm not sure what "combined" means in that context, but the main point is that the label doesn't include "Reddit").

https://shields.io/category/social

I think we should change this to:

```suggestion
    label: 'User Karma',
```

or

```suggestion
    label: 'Karma',
```

services/hackernews/hackernews-user-karma.service.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  async handle({ userid }) {
+    const json = await this.fetch({ userid })
+    const { karma, id } = this.transform({ json })
+    return this.constructor.render({
+      karma,
+      userid: id,
+    })
+  }
```
calebcartwright, line 70: Is there a difference in the `id` value in the response vs the `userid` value we already have via the route parameter?

If not then perhaps consider doing the null check inline with the same early throw/bail, and then calling the render function (with or without the deconstructing).

It's great to see the application of the common `transform` pattern, especially when we have complex logic that's much more easily tested directly via a unit test. However, in this case the canonical not found scenario can be easily exercised already and we aren't really doing anything else but deconstructing and then building a new object.

Just food for thought, don't feel obligated to make any changes

services/text-formatters.js
```diff
@@ -46,27 +46,31 @@ function ordinalNumber(n) {
   return n + (s[(v - 20) % 10] || s[v] || s[0])
 }
 
-// Given a number, string with appropriate unit in the metric system, SI.
+// Given a number (positive or negative), string with appropriate unit in the metric system, SI.
```
calebcartwright, line 49: My initial thought was that we'd introduce a new function, and thus formatter, that would support positive/negative while leveraging `metric` internally. In hindsight though I'm questioning myself and whether that'd actually be helpful, and think I'm convinced by having a proverbial one-stop-shop as you've done here.

However, I do think we should increase the unit test suite (can be found in `text-formatters.spec.js`) to account for the newly supported inputs

calebcartwright (commented):
Few thoughts from my end, inline below

services/text-formatters.js
chris48s: I thought about this too. The conclusion I came to before writing https://github.com/badges/shields/pull/7411#discussion_r774163739 is I don't see any downside to `metric()` just working for both positive and negative as long as we've got separate test validators for metric (positive only) and metric (including zero/negative).

We do need some tests in https://github.com/badges/shields/blob/23c0406bedfc6930735e8f5ea75dfe34faf1f290/services/text-formatters.spec.js#L40 for the negative cases though :+1:

services/test-validators.js
```diff
@@ -62,6 +62,11 @@ const isStarRating = withRegex(
 // Required to be > 0, because accepting zero masks many problems.
 const isMetric = withRegex(/^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY])$/)
 
+// Same as isMetric, but tests for negative numbers also.
+const isMetricAllowNegative = withRegex(
+  /^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]?|-[1-9][0-9]*[kMGTPEZY]?|-[1-9]\.[1-9][kMGTPEZY]?)$/
+)
+
```
chris48s, line 69: I think we could simplify this to
`withRegex(/^(-?[0-9]*[kMGTPEZY]?|-?[0-9]\.[0-9][kMGTPEZY])$/)`
i.e: in this case zero is valid.

services/text-formatters.js
calebcartwright: Gotcha, sorry not sure how I missed that. I blame the eggnog and/or holidays in general :laughing: :gift:

services/hackernews/hackernews-user-karma.service.js
tapanchudasama: Yep, this makes sense. `transform` does not do much work apart from throwing Error. If it was doing some object manipulation, then it would make sense to use it. 

I think we can remove it.

services/hackernews/hackernews-user-karma.service.js
```diff
@@ -0,0 +1,64 @@
...
+
+  static route = {
+    base: 'hackernews/user-karma',
+    pattern: ':id',
+  }
+
+  static examples = [
+    {
+      title: 'HackerNews User Karma',
+      namedParams: { id: 'drumstick' },
```
chris48s, line 23: This example give me 'user not found'. Lets replace it with one that does

```suggestion
      namedParams: { id: 'pg' },
```

services/text-formatters.spec.js
```diff
@@ -60,6 +60,27 @@ describe('Text formatters', function () {
     given(1100000000000000000000).expect('1.1Z')
     given(2222222222222222222222222).expect('2.2Y')
     given(22222222222222222222222222).expect('22Y')
+    given(-999).expect('-999')
```
chris48s, line 63: lets add a test for zero in here too

```suggestion
    given(22222222222222222222222222).expect('22Y')
    given(0).expect('0')
    given(-999).expect('-999')
```

chris48s (commented):
couple of v minor things and this is done :+1:

chris48s (approved):

chris48s:
nice - thanks for your contribution

tedmiston:
Question here: I assumed since this has been merged, that I'd be able to use it on shields.io.  But when I search for "hn" or "hacker news", I get no results.

Am I missing something?

chris48s:
It does exist. Example badge https://img.shields.io/hackernews/user-karma/pg?style=social call

If you search "hackernews", or just "hacker" it shows up. That is.. not amazing, but if you wanted to submit a PR to add a few extra keywords, feel free. Here's an example of a service that defines some extra keywords for the examples:

https://github.com/badges/shields/blob/fbb2ff3619764d7343d8dcad8acd719ce67caa1c/services/piwheels/piwheels-version.service.js#L23-L46

PR #7413, opened 2021-12-22
remove old email from package.json

calebcartwright:
Just noticed we still had this in here. Obviously the value is wrong, but have opted to drop it altogether as (a) I don't think we want bug reports going over email regardless, and even if we do, (b) I don't think we've got a good address for it (afaik we only have `ops`, `security`, and a few other addresses for CoC)

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
   ],
   "homepage": "https://shields.io",
   "bugs": {
-    "url": "https://github.com/badges/shields/issues",
-    "email": "thaddee.tyl@gmail.com"
+    "url": "https://github.com/badges/shields/issues"
   },
   "license": "CC0-1.0",
   "author": "Thaddée Tyl <thaddee.tyl@gmail.com>",


chris48s:
We do have `team@shields.io` if we want to put something in there, but I agree we can leave it blank.

Shall we also set
https://github.com/badges/shields/blob/31b4aaac22119c29eee9db2c2af7d8ffe14558bc/package.json#L19
and
https://github.com/badges/shields/blob/31b4aaac22119c29eee9db2c2af7d8ffe14558bc/badge-maker/package.json#L19
to something generic like "Shields Contributors" given several hundred people have had a hand in this thing over the years

calebcartwright:
> to something generic like "Shields Contributors" given several hundred people have had a hand in this thing over the years

tbh i'm rather fuzzy on how these are supposed to be handled. we had similar discussion in https://github.com/badges/squint/pull/1#discussion_r453223601 when first setting up squint. it's just a little unclear to me if/how original authorship can be extended or modified later on :man_shrugging:

chris48s (approved):
ok, lets not block on that

PR #7430, opened 2021-12-24
ignore @types/node updates (for now)

calebcartwright:
I don't see our circumstances changing to need/be ready for v17 any time soon, and the package is updated frequently enough that I think we need a global ignore so that we don't have to do the weekly routine (e.g. #7417)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,7 @@ updates:
       time: '12:00'
     open-pull-requests-limit: 99
     ignore:
+      # https://github.com/badges/shields/pull/7288#issuecomment-974699240
       - dependency-name: '@types/node'
 
   # badge-maker package dependencies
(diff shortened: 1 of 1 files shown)


chris48s:
We've been ignoring this for a sufficiently long time I've now forgotten what the blocker is :)
Can we bung a comment on this ignore explaining why we're ignoring it/ what is going to allow us to upgrade/stop ignoring it

calebcartwright:
> We've been ignoring this for a sufficiently long time I've now forgotten what the blocker is :)

https://github.com/badges/shields/pull/7288#issuecomment-974699240

If anyone wants to have a go at trying to get it working then please be my guest, it's possible things have advanced far enough to work now. I just hit a point of not wanting to waste any more time on it as it seemed to me the best bet might be to wait for the update to v18 with fingers crossed

chris48s (approved):

PR #7436, opened 2021-12-28
minor [reddit] improvements

chris48s:
There are a couple of things from https://github.com/badges/shields/pull/7411 that we should apply here too


calebcartwright (approved):
Sounds reasonable to me. Probably worth keeping an eye out for other badges with similar rendering needs, in case it makes sense at some point to extract a new render helper

Issue #7447, opened 2022-01-04
shields.io doesn't work at all on Safari 15.2, macOS Monterey


willhains:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Safari: Version 15.2 (17612.3.6.1.6)  
macOS: Monterey 12.1

- Clicking on any links does nothing.
- Clicking on badges does nothing.
- Badge creator UI doesn't display.

I tried disabling all browser extensions in Safari, to no effect.
I tried restarting Safari, to no effect.

The web inspector console shows the following errors & warnings.
```
[Error] SyntaxError: Invalid regular expression: invalid group specifier name
	(anonymous function) (a5f660104e61081cc75f27105d31d879c7b5c92d-4b0c973ec07360cf14b8.js:2)
[Warning] The resource https://shields.io/page-data/category/size/page-data.json was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it wasn't preloaded for nothing.
[Warning] The resource https://shields.io/page-data/app-data.json was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it wasn't preloaded for nothing.
```

chris48s:
This looks like it is related to, or possibly a duplicate of https://github.com/badges/shields/issues/7324

I don't use a mac so I can't confirm or reproduce this. I'm going to leave it tagged "question" for the moment pending a confirmation but I'm marking it "good first issue" as it would be really useful to get some help from the wider community on these Apple-specific issues.

willhains:
@chris48s I don't know how much I can help with the fix, but I can certainly gather more info (debug console messages, for example) if needed.

As a starting point, can you tell me which platforms and browsers shields.io is tested on?

chris48s:
We run our frontend tests using cypress which is Chrome-based.
I also usually check frontend dependency bumps in Firefox on an ad-hoc basis.

My theory is that this issue is caused by https://github.com/sindresorhus/decamelize/issues/36 after upgrading decamelize in https://github.com/badges/shields/pull/7098

I've submitted a PR over at #7449 to revert decamelize to version 5. I have no way of testing this myself but is someone who has an i-Device of some description able to check the staging deploy at https://shields-staging-pr-7449.herokuapp.com/ and see if that fixes it in Safari? The staging deploy is on a dyno that goes to sleep so it might take a while to load

paulmelnikow:
[...]

Issue #7324, opened 2021-12-02
Static badge `Make Badge` button always disabled on Safari


LitoMore:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

**Operating system:** macOS Monetary
**Browser version:** Safari 15.1

#### How to reproduce

1. Open Safari
2. Open https://shields.io
3. Fill inputs for `Make Badge`

<img  width="370" src="https://user-images.githubusercontent.com/8186898/144363935-12d1b0a4-b59c-4622-9a82-7406f738a1fc.png">

#### Relevant issue

I cannot open `http://localhost:3000` on my local. It always shows the error view. This issue can only be reproduced on Safari.

<img width="1259" alt="CleanShot 2021-12-02 at 13 43 51@2x" src="https://user-images.githubusercontent.com/8186898/144364777-1d68e1bf-d92a-4a95-ac39-1fbd086e7ea6.png">

~~I guess it's a `gatsby`-related issue.~~

paulmelnikow:
Confirmed on Safari 14.1.2. Thanks for the report.

<img width="549" alt="Screen Shot 2021-12-02 at 9 12 35 AM" src="https://user-images.githubusercontent.com/1487036/144438506-345145f3-158a-4e96-91a4-4daae3a451a2.png">

I see that exception in the JavaScript console too though I am not sure if it is related to the button activation.

Pull requests welcome!

chris48s:
This is an issue where it would be great to get a community contribution for the fix, or a bit of time from you @paulmelnikow if nobody steps in. I don't use a mac so can't reproduce/debug this and I'm pretty sure Caleb/P-Y don't use macs either

calebcartwright:
Correct, I've no way of testing this. I was curious whether any automated testing (with Cypress or otherwise) would be capable of detecting/confirming/preventing this sort of thing, though I suppose we'd at least need a Mac build/CI environment to test on

LitoMore:
OK. I can give a check.

chris48s:
My theory is that this issue is caused by https://github.com/sindresorhus/decamelize/issues/36 after upgrading decamelize in https://github.com/badges/shields/pull/7098

I've submitted a PR over at #7449 to revert decamelize to version 5. I have no way of testing this myself but is someone who has an i-Device of some description able to check the staging deploy at https://shields-staging-pr-7449.herokuapp.com/ and see if that fixes it in Safari? The staging deploy is on a dyno that goes to sleep so it might take a while to load

[...]

PR #7449, opened 2022-01-05
revert decamelize and humanize-string to old versions

chris48s:
closes https://github.com/badges/shields/issues/7447
closes https://github.com/badges/shields/issues/7324


calebcartwright (dismissed):
Sounds like this may not address some specific issues, but given the linked upstream issue it seems like we may want to downgrade this regardless so providing an approval accordingly

package.json
```diff
@@ -33,7 +33,7 @@
     "cloudflare-middleware": "^1.0.4",
     "config": "^3.3.6",
     "cross-env": "^7.0.3",
-    "decamelize": "^6.0.0",
+    "decamelize": "^5.0.1",
```
LitoMore, line 36: You should pin it at [version 3](https://github.com/sindresorhus/decamelize/releases/tag/v4.0.0). The `humanize-string` needs to upgraded as well.

```diff
# dependencies
-    "decamelize": "^6.0.0",
+    "decamelinze": "3.2.0",

# devDependencies
- "humanize-string": "^3.0.0",
+ "humanize-string": "2.0.0",
```

LitoMore (changes requested):

LitoMore:
The title of the PR needs to be updated.

LitoMore (changes requested):
Let's remove the `^` mark to make sure dependabot will not update these dependencies by accident.

And please run `npm ci` after these suggestions are applied to ensure that the lockfile is up to date.

chris48s:
Dependabot bumps dependencies even if they're pinned in package.json e.g: https://github.com/badges/shields/pull/7431/files
Assuming we get a confirmation on https://github.com/badges/shields/issues/7447 I'll add entries to https://github.com/badges/shields/blob/4fef1335d238a4f89a778e857d265b43273dd184/.github/dependabot.yml#L11 to tell dependabot to ignore them

calebcartwright (approved):
:heart:

calebcartwright:
Let's go ahead and merge+deploy

PR #7462, opened 2022-01-07
[Hackernews] Show Story Score

tapanchudasama:


calebcartwright:
Thanks for continuing to work on these, and the fast update on the class names! I feel like there's some commonalities/repeated bits, perhaps relative to the existing karma badge, but certainly to each other.

In cases like this we often like to either create a separate mod with functions that encapsulate the common fetching/rendering/whatever logic, or insert a service-specific base class within the inheritance hierarchy that contains those common functions.

Do you think something like that would be feasible here, and if so would you be interested in going that route, perhaps even combining the two PRs?

tapanchudasama:
> separate mod with functions that encapsulate the common

@calebcartwright your mean creating a sort of `helper.js` file which would contain the render/fetch/handle logic?

calebcartwright:
> > separate mod with functions that encapsulate the common
> 
> @calebcartwright your mean creating a sort of `helper.js` file which would contain the render/fetch/handle logic?

Yup, you've got it! There's a good number of other services that leverage one (or in a few cases both) of those patterns to avoid having duplicate function definitions across classes in the same service family. None are immediately coming to mind for me but I can try to find one if that helps

tapanchudasama:
@calebcartwright I'll make these changes, and the label changes in the comments PR itself. I hope that's okay.

PyvesB:

If others think that repository level projects would be an option and want to pick this up, you can simply fetch the commits by running git fetch origin pull/7462/head:pr-7462. Change the implementation to target repository level projects and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

PR #7468, opened 2022-01-08
Fix c8 coverage

PyvesB:
Code coverage using c8 is no longer functional and daily tests have been failing since #7439. It seems that we now need to explicitly specify `@istanbuljs/schema` as a dev dependency.


calebcartwright:
Odd. Coverage works for me locally as-is, including with a clean clone/install. Wonder if it's an `npm i` vs `npm ci` issue, but either way, I see no harm in adding the dep.

Will also note that the daily tests have been consistently failing for ages due to service test failures, so this is just another indicator that we're all largely ignoring them because of the constant failure :sweat_smile:

calebcartwright (approved):

PR #7475, opened 2022-01-11
[Depfu] Add support for Gitlab

### LINKED ISSUE #7474: Depfu for Gitlab repos
:clipboard: **Description**

The [Depfu](https://github.com/badges/shields/pull/2204/files#diff-07f982763d8a47feb189c97cf5053d5fb2d7251a649340d3b7d3c0706fe66692R13) service is using the `/github/shields` endpoint of Depfu, which means it can only query the status of Github repos. It would be nice to be able to query Gitlab repos.

:link: **Data**

Depfu also has a `https://depfu.com/gitlab/shields/${userRepo}` endpoint, which acts the same as the `/github/shields` endpoint, but for Gitlab repos.

:microphone: **Motivation**

While Depfu provides built-in badges, their style cannot be customized (e.g., `?style=flat-square`), so it would be still useful to have this as a `shields.io` badge.

kris7t:
Fixes #7474

Following the recommendation https://github.com/badges/shields/issues/7474#issuecomment-1009239968 of @chris48s, this PR adds a `:vcsType` parameter to depfu URLs to select between Github and Gitlab. Badges without a VCS type are redirected to the Github route via a legacy route.


services/depfu/depfu.tester.js
```diff
@@ -11,12 +11,16 @@ const isDependencyStatus = Joi.string().valid(
...
   .expectBadge({
     label: 'dependencies',
     message: isDependencyStatus,
   })
 
 t.create('depfu dependencies (repo not found)')
-  .get('/pyvesb/emptyrepo.json')
+  .get('/github/pyvesb/emptyrepo.json')
   .expectBadge({ label: 'dependencies', message: 'not found' })
+
```
calebcartwright, line 23: Could we add a test that hits a gitlab repo too?

calebcartwright (commented):
Thanks! Think we could use another test but otherwise LGTM

@chris48s maybe not worth discussing as part of this PR since the route change is both consistent with what we had before as well as other dependency management/update services, but it seems like this class of badge routes are all missing the NOUN portion from our canonical badge url pattern

services/depfu/depfu.tester.js
kris7t: Sure! I just pushed some tests. For an existing repo, Depfu provides an example repo on Gitlab, too. For a nonexistent repo, I used `fdroid/nonexistant` from https://github.com/badges/shields/blob/50c0be0029a5ebfe9b387419c10495e04790d033/services/gitlab/gitlab-release.tester.js#L44

services/depfu/depfu.service.js
```diff
@@ -1,18 +1,22 @@
...
   colorscheme: Joi.string().required(),
 }).required()
 
-export default class Depfu extends BaseJsonService {
+class Depfu extends BaseJsonService {
   static category = 'dependencies'
-  static route = { base: 'depfu', pattern: ':user/:repo' }
+  static route = {
+    base: 'depfu',
+    pattern: ':vcsType(github|gitlab)/:user/:repo',
```
chris48s, line 13: With GitLab we can't just use `:user/:repo`, repos can be nested in an aribtrary number of nested subgroups e.g:
https://gitlab.com/megabyte-labs/docker/ci-pipeline/ansible-lint so for projects that work with gitlab we would usually make the route something like `:vcsType(github|gitlab)/:project+` (which allows an arbitrary number of slashes) and then call `https://depfu.com/${vcsType}/shields/${project}` assuming depfu inherits that design descension and exposes it in the same way. Do we have an example of this?

chris48s (commented):
Nice. Thanks for having a look at this and providing tests.

chris48s:
> maybe not worth discussing as part of this PR since the route change is both consistent with what we had before as well as other dependency management/update services, but it seems like this class of badge routes are all missing the NOUN portion from our canonical badge url pattern

@calebcartwright  I think if we want to change it to `/depfu/dependencies/(github|gitlab)/:project+` we might as well do it now in this PR. Then we only need to add one redirect from `/depfu/:user/:repo` to `/depfu/dependencies/github/:project+`. If we ship it with `/depfu/:vcsType(github|gitlab)/:project+` and then change it in another PR we need another redirect to avoid a BC break. Lets not do that if we don't need to. That said, we're doing the same thing on requires.io and libraries.io so I can see a case for just ignoring it.

kris7t:
I just created an example project to test this. It might be a bit trickier, because Depfu seems to require urlencoded subgroup separators: https://depfu.com/gitlab/shields/shields-example-group%252Fsubgroup/example-nodejs (replacing `%252F` with `/` doesn't work).

I guess the best approach would be to parse the `:project+` argument, and replace all but the last `/` characters with `%252F`?

chris48s:
Lets use `encodeURIComponent()` instead of manually replacing.

For the Gitlab case we can do something like

```js
encodeURIComponent(project.substr(0, project.lastIndexOf('/')))
project.substr(project.lastIndexOf('/'))
```

We'll also want to make sure project has at least one / in it before doing that, so something like

```js
if (!project.includes('/')) {
  throw InvalidParameter()
}
```

https://github.com/badges/shields/blob/23c0406bedfc6930735e8f5ea75dfe34faf1f290/core/base-service/errors.js#L157

For the Github case, we can just pass project as-is

kris7t:
I updated this PR according to the review comments:

* The endpoint was moved to `/depfu/dependencies`.
* If there is no `/` in the `project`, we throw an `InvalidParameter` error. I left it with a default message, because I couldn't come up with any specific `prettyMessage` that wasn't way too long.
* I added the URI encoding regardless the `vcsType` to simplify logic (it doesn't do anything for Github, but it won't hurt either).
* For testing, I use the example group, subgroup, and repo that I created on my Gitlab account (the test nested repo has to have Depfu enabled, but I couldn't find any existing example repo that was nested in a subgroup).

chris48s (approved):
Great - thanks for finishing this up :+1: Sorry it took a few days to get back to it

PR #7476, opened 2022-01-12
Support [Feedz] response with multiple pages without items

jakubfijalkowski:
This PR fixes issue #7446 .

Basically, when package has more than 128 versions, Feedz stops sending the _catalog entries_ with the root page and we need to fetch all the individual pages to get the entries.

This isn't ideal but Feedz still lacks support of `packageid` when searching packages (this is used with nuget.org and MyGet). I've asked their support if they plan to add it and I will adjust this service if they do plan so.


calebcartwright:
Thanks for working on this! While it does make the code a bit gnarlier, I think it would be better to fire off all the requests for the subsequent pages asynchronously so they don't get made in sequence (which would rarely render in practice due to network latency compounding).

There's a few examples of this for reference, e.g .https://github.com/badges/shields/blob/94a8d945c31e9365acbc7652deceef5103e638df/services/gitlab/gitlab-base.js#L48-L63

jakubfijalkowski:
Thanks for the comment, fixed that. Also, the code is, I think, even simpler now (since we don't really need to map the fetches at all).

services/feedz/feedz.service.js
```diff
@@ -23,6 +24,18 @@ const schema = Joi.object({
...
+const singlePageSchema = Joi.object({
+  items: Joi.array()
+    .items(
+      Joi.object({
+        catalogEntry: Joi.object({
+          version: Joi.string().required(),
+        }).required(),
+      })
+    )
+    .default([]),
```
calebcartwright, line 36: What would you think about creating a separate schema for this, e.g. `itemsSchema` and then reusing it in the main two schemas? That should help make things a bit more succinct and encapsulate the definition in one place

services/feedz/feedz.service.js
```diff
@@ -80,6 +93,25 @@ class FeedzVersionService extends BaseJsonService {
...
+            errorMessages: {
+              404: 'repository or package not found',
+            },
+          })
+        )
+      )
+      return { items }
+    } else {
+      return json
+    }
```
calebcartwright, line 112: Bit of a nit since it probably won't have much of a runtime impact, but could we invert the conditional with an early return on the iterator to hit the return path?

E.g. would something like this work? (hand typed in browser so might be botched but hopefully still conveys the thought)

```suggestion
    if (json.items.length === 0 || json.items.any(i => i.catalogEntry)) {
      return json
    } else {
      const items = await Promise.all(
        json.items.map(i =>
          this._requestJson({
            schema: singlePageSchema,
            url: i['@id'],
            errorMessages: {
              404: 'repository or package not found',
            },
          })
        )
      )
      return { items }
    }
```

services/feedz/feedz.tester.js
```diff
@@ -46,6 +48,14 @@ t.create('multi-page')
     color: 'blue',
   })
 
+t.create('multi-page-no-items')
+  .get('/feedz/v/shieldstests/public/Shields.MultiPageNoItems.json')
+  .expectBadge({
+    label: 'feedz',
+    message: 'v1.0.0',
+    color: 'blue',
+  })
```
calebcartwright, line 57: I gather from the target that this has been setup specifically for our needs to reproduce the scenario, and we don't have to worry about this changing upstream/invalidating the test, correct?

calebcartwright (commented):
Thanks for the updates! It's looking good, few smaller items left inline for your review

services/feedz/feedz.service.js
jakubfijalkowski: Yes, you're right. I can also go a little bit further and re-use the `singlePageSchema` as that is the type of object inside the first `items` array. Missed that fact earlier.

services/feedz/feedz.service.js
jakubfijalkowski: Yeah, that should make the "no need to download anything" path a little bit faster.

services/feedz/feedz.tester.js
jakubfijalkowski: Correct, I've prepared these repos just for this integration. The code is public, although I haven't written the scripts necessary to upload the packages (but I can write them down if there is a need). I can also transfer the repo to you/shields org (although that _might_ pose some problems, as Feedz lacks in some areas).

calebcartwright: No worries, what you've done is fine; just wanted to confirm!

calebcartwright (approved):
LGTM, thanks again for the quick fix!

Issue #4498, opened 2020-01-08
Badge request: Curseforge


LoneWanderer-GH:
:clipboard: **Description**

Badges for [CurseForge](https://www.curseforge.com/) addons.

- What sort of information should this badge show?
  - Repository name
  - Download count
  - Managed game versions

Simple example:
![](https://img.shields.io/badge/dynamic/json?label=Curse%20downloads&color=orange&query=downloadCount&url=https://addons-ecs.forgesvc.net/api/v2/addon/356973/)
One could also give a shot to the very nice project https://github.com/way2muchnoise/CurseForgeBadge  that does part of the job but may not be as popular a shileds

:link: **Data**

Where can we get the data from?

- Is there a public API? https://addons-ecs.forgesvc.net/api/v2/addon/<project number>  see example above
- Does the API requires an API key? seems not !
- Link to the API documentation **broken links** : https://www.curseforge.com/docs/api https://github.com/curseforge/api

:microphone: **Motivation**

Please explain why this feature should be implemented and how it would be used.
- What is the specific use case?
Most addons developper have a custom/own repo (say github) and might want to show some status of their addon on CurseForge. CurseForge is a repo of ready-to-use addons for video games and has addon managing tools (such as Twitch app)

Example of request result:

```
{
  "id": 356973,
  "name": "CoolLine (classic updated)",
  "authors": [
    {
      "name": "lonewanderer_twitch",
      "url": "https://www.curseforge.com/members/436037-lonewanderer_twitch?username=lonewanderer_twitch",
      "projectId": 356973,
      "id": 262075,
      "projectTitleId": null,
      "projectTitleTitle": null,
      "userId": 436037,
      "twitchId": 65711671
    }
  ],
  "attachments": [
    {
      "id": 242739,
      "projectId": 356973,
      "description": "",
      "isDefault": true,
      "thumbnailUrl": "https://media.forgecdn.net/avatars/thumbnails/242/739/256/256/637136000695779179.png",
      "title": "637136000695779179.png",
      "url": "https://media.forgecdn.net/avatars/242/739/637136000695779179.png",
      "status": 1
    }
  ],
  "websiteUrl": "https://www.curseforge.com/wow/addons/coolline-classic-updated",
  "gameId": 1,
  "summary": "CoolLine for retail and classic",
  "defaultFileId": 2854263,
  "downloadCount": 44,
  "latestFiles": [
    {
      "id": 2854263,
      "displayName": "8.0.008",
[...]

PR #7499, opened 2022-01-16
add [curseforge] downloads badge

Minecraftschurli:
this PR adds the curseforge downloads badge requested in #4498

it includes:
- auth helpers
- documentation for the api token
- the curseforge downloads badge

it sadly requires an api token to work for which the process for acquiring can be found [here](https://support.curseforge.com/en/support/solutions/articles/9000208346-about-the-curseforge-api-and-how-to-apply-for-a-key)

### DIFF
--- a/config/custom-environment-variables.yml
+++ b/config/custom-environment-variables.yml
@@ -77,6 +77,7 @@ private:
   bitbucket_password: 'BITBUCKET_PASS'
   bitbucket_server_username: 'BITBUCKET_SERVER_USER'
   bitbucket_server_password: 'BITBUCKET_SERVER_PASS'
+  curseforge_api_token: 'CURSEFORGE_API_TOKEN'
   discord_bot_token: 'DISCORD_BOT_TOKEN'
   drone_token: 'DRONE_TOKEN'
   gh_client_id: 'GH_CLIENT_ID'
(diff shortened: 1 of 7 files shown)


calebcartwright:
Thanks for the PR!

> it sadly requires an api token to work for which the process for acquiring can be found here

Do you know how the authorization is scoped? I.e. will a request that authenticates with a valid token be allowed to pull data for _any_ target, or does the token have to be explicitly authorized/scoped to the target in order to work?

Minecraftschurli:
it is an api token auth which is normally not scoped (in this case it is unscoped)

Minecraftschurli:
I have tested it with my api token and it worked (I didn't have to set any scopes)

calebcartwright:
The process for requesting an api key looks notably more involved and complicated than what we typically have to go through. What was your experience like?

Minecraftschurli:
I used the game developer api key which just requires a google account login

calebcartwright:
> I used the game developer api key which just requires a google account login

Gotcha. We're going to need some time to talk this over amongst the core team, as well as likely needing to have a dialog with the Curseforge folks to determine if/when this might be viable to proceed.

curio-sitas:
Any news ?

starboyceleste:
?

PyvesB:
This work was effectively merged via #9252. Thanks for the contribution and the patience!

PR #7542, opened 2022-01-28
WIP: fix package build

chris48s:


chris48s:
Hmm. Whatever it was that caused all the builds to fail earlier appears to be transient. Re-running the builds now has the tests passing.

PR #7570, opened 2022-02-05
[AUR] Fix RPC URL

ouuan:
Before this fix, the AUR badges were showing "not found".

See https://github.com/Jguer/yay/issues/1684


ouuan:
It seems that https://gitlab.archlinux.org/kevr/aurweb/-/commit/101de8e7b18916dd84ae52d39fb853924ea63a42 was released just now. So the badges are temporarily working fine.

PyvesB (approved):
Thanks for fixing this @ouuan , and linking to that issue and commit. Let's get this merged! 👍🏻

PR #7582, opened 2022-02-08
fix docker image user agent

### LINKED ISSUE #7573: Docker Image serves incorrect `User-Agent`

shields.io

### 🐞 Description

The requests coming from shields.io currently have `Shields.io/${GITHUB_SHA::7}` as user agent. That's obviously unintended, and it might lead to those requests being blocked by web application firewalls.

### 🔗 Link to the badge

_No response_

### 💡 Possible Solution

_No response_

chris48s:
Closes #7573

`${GITHUB_SHA::7}` is a bash subscript syntax, so we need to run it with bash and write it to a var

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -12,6 +12,9 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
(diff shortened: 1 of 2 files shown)


calebcartwright (approved):
Can't think of any easy way to readily test this, but it looks like the right change to me anyway

PR #7586, opened 2022-02-10
Update [amo] to use v4 API, set custom `cacheLength`s

### LINKED ISSUE #7584: Excessive API usage for Mozilla Firefox Add-ons (and deprecated API usage)

shields.io

### 🐞 Description

Hi,

The addons.mozilla.org team is looking at removing its v3 API as planned per [deprecation announcement](https://blog.mozilla.org/addons/2021/02/01/addons-mozilla-org-api-v3-deprecation/) (We're in fact a bit late, we had announced we would shut it down on December 31st 2021).

When looking at this, we noticed a few problems with Shields.io:
- It is using the deprecated v3 version instead of the stable v4.
- It's making requests without the expected trailing slash, causing each request to follow a redirect to the URL with the trailing slash (doubling requests)
- The usage seems excessive, totaling over 2 million requests per month.

Since we are going to shut down v3, you should start using v4 instead, perhaps adjusting your code depending on what this would break. Furthermore, we'd strongly urge you to find ways to avoid calling our API so much. Fixing that trailing slash issue will halve the number of requests, but please consider caching more aggressively. We can and will tweak rate limiting on our end, but not having to deal with the requests in the first place would be preferable.

### 🔗 Link to the badge

_No response_

### 💡 Possible Solution

_No response_

chris48s:
Closes #7584

### DIFF
--- a/services/amo/amo-users.service.js
+++ b/services/amo/amo-users.service.js
@@ -14,6 +14,8 @@ export default class AmoUsers extends BaseAmoService {
     },
   ]
 
+  static _cacheLength = 21600
+
   static defaultBadgeData = { label: 'users' }
 
   static render({ users: downloads }) {
--- a/services/amo/amo-downloads.service.js
+++ b/services/amo/amo-downloads.service.js
@@ -24,6 +24,8 @@ class AmoWeeklyDownloads extends BaseAmoService {
     },
   ]
 
+  static _cacheLength = 21600
+
   static defaultBadgeData = { label: 'downloads' }
 
   static render({ downloads }) {
--- a/services/amo/amo-rating.service.js
+++ b/services/amo/amo-rating.service.js
@@ -23,6 +23,8 @@ export default class AmoRating extends BaseAmoService {
     },
   ]
 
+  static _cacheLength = 7200
+
   static render({ format, rating }) {
     rating = Math.round(rating)
     return {


services/amo/amo-base.js
```diff
@@ -21,7 +21,7 @@ class BaseAmoService extends BaseJsonService {
   async fetch({ addonId }) {
     return this._requestJson({
       schema,
-      url: `https://addons.mozilla.org/api/v3/addons/addon/${addonId}/`,
+      url: `https://addons.mozilla.org/api/v4/addons/addon/${addonId}/`,
```
chris48s, line 24: There were no breaking changes from v3 to v4 affecting any of the values we use on any of these badges

services/amo/amo-users.service.js
```diff
@@ -14,6 +14,8 @@ export default class AmoUsers extends BaseAmoService {
     },
   ]
 
+  static _cacheLength = 21600
```
chris48s, line 17: @calebcartwright based on your comment in https://github.com/badges/shields/issues/7584#issuecomment-1034296124 I've set downloads/users to `21600` seconds (6 hours). Feel free to tweak it if you want.

PyvesB (approved):
Thanks for picking this up, looks good to me. 👍🏻

PR #7606, opened 2022-02-12
Remove no download handling and add API warning to [Wordpress] badges

### LINKED ISSUE #7356: wordpress downloads service tests failing

5 days ago

:camera: **Live badge**

this looks sensible until you realise the crucial detail: is not a real wordpress plugin

Kinda. It definitely returns _some numbers_ for a valid plugin, but also returns some numbers for plugins that definitely don't exist so who knows if they are right. Seemingly whatever slug I give this API it gives us some numbers back

https://api.wordpress.org/stats/plugin/1.0/downloads.php?slug=100
https://api.wordpress.org/stats/plugin/1.0/downloads.php?slug=definitely-not-a-real-plugin
https://api.wordpress.org/stats/plugin/1.0/downloads.php?slug=sadikovtdjoiybvjtdoivgrdjashc

These are definitely not real wordpress plugin names.

:link: **CircleCI link**

example:
https://app.circleci.com/pipelines/github/badges/daily-tests/1381/workflows/efbc35ec-6526-4fe6-9e68-9740a51293a3/jobs/2283/tests#failed-test-17

:lady_beetle: **Stack trace**

```
AssertionError: message mismatch: expected '35/week' to equal 'plugin not found or too new'
    at Function._expectField (file:///home/circleci/project/shields/core/service-test-runner/icedfrisby-shields.js:70:51)
    at IcedFrisbyNock.<anonymous> (file:///home/circleci/project/shields/core/service-test-runner/icedfrisby-shields.js:55:26)
    at IcedFrisbyNock.<anonymous> (node_modules/icedfrisby/lib/icedfrisby.js:954:10)
    at invokeNextHook (node_modules/icedfrisby/lib/icedfrisby.js:1003:24)
    at /home/circleci/project/shields/node_modules/icedfrisby/lib/icedfrisby.js:1017:7
    at new Promise (<anonymous>)
    at IcedFrisbyNock._runHooks (node_modules/icedfrisby/lib/icedfrisby.js:976:12)
    at IcedFrisbyNock.run (node_modules/icedfrisby/lib/icedfrisby.js:1276:20)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Context.<anonymous> (node_modules/icedfrisby/lib/icedfrisby.js:1348:9)
```

:bulb: **Possible solution**

I can't begin to fathom what manner of terrible thing has happened here, but I'm pretty sure the problem is is upstream and there's not much we can do about it.

PyvesB:
Fixes #7356.

### DIFF
--- a/services/wordpress/wordpress-base.js
+++ b/services/wordpress/wordpress-base.js
@@ -41,7 +41,7 @@ const notFoundSchema = Joi.object()
 const pluginSchemas = Joi.alternatives(pluginSchema, notFoundSchema)
 const themeSchemas = Joi.alternatives(themeSchema, notFoundSchema)
 
-export default class BaseWordpress extends BaseJsonService {
+export class BaseWordpress extends BaseJsonService {
   async fetch({ extensionType, slug }) {
     const url = `https://api.wordpress.org/${extensionType}s/info/1.2/`
     let schemas
@@ -84,3 +84,12 @@ export default class BaseWordpress extends BaseJsonService {
     return json
   }
 }
+
+export const documentation = `
+<p>
+  These badges rely on an API that is no longer supported by Wordpress. You are
+  still free to use them, simply bear in mind that Shields.io cannot guarantee
+  that they'll keep on working in the future. Please also double-check the
+  provided slug, as an incorrect value may lead to unexpected results.
+</p>
+`
--- a/services/wordpress/wordpress-downloads.service.js
+++ b/services/wordpress/wordpress-downloads.service.js
@@ -1,7 +1,6 @@
 import Joi from 'joi'
 import { renderDownloadsBadge } from '../downloads.js'
-import { NotFound } from '../index.js'
-import BaseWordpress from './wordpress-base.js'
+import { documentation, BaseWordpress } from './wordpress-base.js'
 
 const dateSchema = Joi.object()
   .pattern(Joi.date().iso(), Joi.number().integer())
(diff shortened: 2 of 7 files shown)


calebcartwright (approved):
lgtm, thanks!

chris48s:
Based on the comment in https://meta.trac.wordpress.org/ticket/6061#comment:3 and https://meta.trac.wordpress.org/ticket/6061#comment:5 do we want to apply this documentation to all wordpress badges (including the ones calling the `/info` endpoint, or just the ones calling the `/stats` endpoint)? It seems like the ones calling `/info` are more reliable.

PyvesB:
> Based on the comment in https://meta.trac.wordpress.org/ticket/6061#comment:3 and https://meta.trac.wordpress.org/ticket/6061#comment:5 do we want to apply this documentation to all wordpress badges (including the ones calling the `/info` endpoint, or just the ones calling the `/stats` endpoint)? It seems like the ones calling `/info` are more reliable.

Yeah, I did hesitate, and was leaning slightly towards applying the documentation everywhere for three reasons:
* they state that `/info` is "not intended to be used outside of WordPress.org/WordPress but we don't prevent it", which still feels like "unsupported for external users".
* their doc page will apparently soon be removed due to the Codex shutdown. All Wordpress badges will effectively start breaching our guideline of "should not obtain data from undocumented or reverse-engineered API endpoints". I thought it'd be good to somehow reflect "we know this is not right and not an example to follow for new badges".
* the downloads badge actually uses both `/info` and `/stats`, the former for the `dt` variant, the latter for the `dy`, `dm`, `dw`, and `dd` variants. They're part of the same example, to be precise we'd have to highlight the difference in the documentation text, making it a little less straightforward to grasp for users.

I don't feel too strongly though, so happy to go down the other route if you think it's best.

chris48s (approved):
sounds reasonable

PR #7612, opened 2022-02-14
Support for Test at Scale Badge, run [TAS]

nevilm-lt:
## Added
- Added a badge support for [Test At Scale](https://www.lambdatest.com/test-at-scale)

## Usage
- URL: http://localhost:8080/tas/tests/:PROVIDER/:ORGNAME/:REPONAME
- Example Usage: http://localhost:8080/tas/tests/github/test-at-scale/demo-repo

## Badge
tests


calebcartwright:
Thanks for the PR! Did one very quick glance over this and a few initial thoughts:

- Please generalize the default label name to something like `tests`. That will ensure it's consistent with our [other badges in the same category](https://shields.io/category/test-results), and also ensure we aren't violating the promotional/branding aspects of our [badge specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md#guidelines)
- We've got a number of utility functions for test-related badges that I'd encourage checking out (should make things easier!) For reference:
  - https://github.com/badges/shields/blob/master/services/appveyor/appveyor-tests.service.js
  - https://github.com/badges/shields/blob/master/services/sonar/sonar-tests.service.js

nevilm-lt:
Hey @calebcartwright, Thanks for the suggestions.
- As pointed out - I've updated the label to be more generic.
- I've gone though the utility functions. But looks like our use case is more specific which can not be catered by functions already available. Please let me know if I'm missing something here I'll be happy to make those changes.

Thanks.

calebcartwright:
Could you share some more info about the API surface? It gives me pause when I see "badge" as part of the path as it implies it's a specialized endpoint with minimal data for a single badge purpose. That's not _necessarily_ a problem, but as I see it, part of the reason why you can't incorporate our existing functions (which provide end-user features around customization) is because the API response is so minimal.

I would think/hope there could be an endpoint that would provide some of the standard details around passed/failed/skipped/etc. that we see with the various other platforms we integrate with.

nevilm-lt:
@calebcartwright On test-at-scale [TAS] we are not running all test cases for each build. We try to execute minimum number of test cases that are affected by the latest PR changes.
So the metric that we trying to display is the number of tests executed in "x" duration.

Here's the api example.
https://api.tas.lambdatest.com/repo/badge?git_provider=github&org=nevilm-lt&repo=cheerio
`{"badge":{"status":"passed","duration":"27s","total_tests":"0"}}`

calebcartwright:
Thank you for the reply @nevilm-lt, but I'm not sure you've really responded to the point/topic.

We don't want Shields to be locked into an API from a platform that's trying to dictate a specific type of badge down to basic rendering concerns. We want to consume standard APIs that provide data points relevant to general users of the platform, such that we are then able to offer our users a range of options for badges as opposed to assuming/pushing a _single_ use case. 

We're absolutely on board with being able to provide our users with _a_ TAS badge that displays the test count + duration, but I don't want us to be coupled to an API endpoint that only allows that badge because that really shouldn't be the only one.

nevilm-lt:
Hi @calebcartwright,

We have updated the badge to show standard information as other badges provides. i.e. Passed, Failed, Skipped etc. Also we have used standard utility functions to render the badge. Please feel free to let us know if there are any other changes required.

Thanks.

palashmarantas:
> Hi @calebcartwright,
> 
> We have updated the badge to show standard information as other badges provides. i.e. Passed, Failed, Skipped etc. Also we have used standard utility functions to render the badge. Please feel free to let us know if there are any other changes required.
> 
> Current Badge: tests
> 
> Thanks.

Any updates on this @calebcartwright ?

calebcartwright:
Awesome, thanks for the updates! To clarify a bit on the prior discussion, if you'd like to be able to show a badge that includes the duration in **addition** to all the other variants that's perfectly alright as well.

Things like okay here based on a quick glance, so I can either give this a more thorough review as-is if you're fine with the existing badges, or I can hold off/suggest some next steps if you'd like to get a count+duration badge variant too.

palashmarantas:
> Awesome, thanks for the updates! To clarify a bit on the prior discussion, if you'd like to be able to show a badge that includes the duration in **addition** to all the other variants that's perfectly alright as well.
> 
> Things like okay here based on a quick glance, so I can either give this a more thorough review as-is if you're fine with the existing badges, or I can hold off/suggest some next steps if you'd like to get a count+duration badge variant too.

Hi @calebcartwright, 
    Thanks for response. Please provide thorough review for existing batch PR. Let's merge this PR first. Please share you points for count+duration will create separate PR for new batch.

services/tas/tas-tests.tester.js
```diff
@@ -0,0 +1,8 @@
+import Joi from 'joi'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('tas tests').get('/github/tasdemo/axios.json').expectBadge({
+  label: 'tests',
+  message: Joi.string(),
```
chris48s, line 7: Can we make this more specific. This isn't really a useful test as an error message will match `Joi.string()` too. There are some validators in `services/test-validators.js` for testing this type of test summary badge: https://github.com/badges/shields/blob/6e803367e0f3c8e0cc4196a700af37fac1629f4d/services/test-validators.js#L135-L154

Also can we have a couple of tests for the behaviour in the 401 and 404 cases.

nevilm-lt:
Hey @chris48s, We have added the test cases for multiple behaviours which mocks specific use cases. Please let us know if there are anything else we can do.

Thanks.

services/tas/tas-tests.tester.js
```diff
@@ -0,0 +1,89 @@
...
+  .intercept(nock =>
+    nock(tasBaseUri)
+      .get('?git_provider=github&org=tasdemo&repo=private')
+      .reply(401, {})
+  )
+  .expectBadge({
+    label: 'tests',
+    message: 'private project not supported',
+    color: 'lightgrey',
+  })
```
calebcartwright, line 89: We'd want to make these live tests; no need to mock

services/tas/tas-tests.service.js
```diff
@@ -0,0 +1,110 @@
...
+  },
+}
+
+const schema = Joi.object({
+  badge: Joi.object({
+    passed: Joi.number().required(),
+    failed: Joi.number().required(),
+    skipped: Joi.number().required(),
+    total_tests: Joi.number().required(),
+    status: Joi.string().required(),
```
calebcartwright, line 28: Not necessary to make any changes here, but sharing for awareness/future reference that we do have some schema helpers as well (`nonNegativeInteger` from `validators.js` would have been useful)

calebcartwright (commented):
Overall looks good to me! Believe we're in the home stretch now, couple final minor items

services/tas/tas-tests.tester.js
nevilm-lt: updated the test cases to make them live instead of mock.

services/tas/tas-tests.service.js
nevilm-lt: thanks for the nit. I've updated this in the latest changes.

services/tas/tas-tests.tester.js
nevilm-lt: added.

services/tas/tas-tests.tester.js
calebcartwright: thanks!

services/tas/tas-tests.service.js
```diff
@@ -0,0 +1,111 @@
...
+  renderTestResultBadge,
+} from '../test-results.js'
+import { nonNegativeInteger } from '../../services/validators.js'
+
+const commonAttrs = {
+  namedParams: {
+    provider: 'github',
+    org: 'test-at-scale',
+    repo: 'badge-demo',
+  },
```
calebcartwright, line 14: Sorry I missed this before but is this a valid, public project? I'm not seeing it on GitHub and the badge is unsurprisingly showing a not found. We really like to have these examples be valid projects so that users can actually see/work with them in the modal window.

If it's something you all are planning on setting up in the near future that's alright, but otherwise it would probably be better to swap in a real target 

https://shields-staging-pr-7612.herokuapp.com/tas/tests/github/test-at-scale/badge-demo?compact_message


Using the same target as the tests would suffice too:

https://shields-staging-pr-7612.herokuapp.com/tas/tests/github/tasdemo/axios

nevilm-lt: @calebcartwright Thanks for pointing out. I've updated the example project.

calebcartwright (approved):
Awesome, thanks so much!

PR #7613, opened 2022-02-16
[Docsrs] Use the latest build status for multiple builds

Nemo157:
When documentation for a crate is rebuilt the `builds.json` can return multiple results, ordered latest first, so we must take the first element of the array rather than the last.

As an example currently <https://img.shields.io/docsrs/bevy_tweening/0.3.1> is showing failing, despite the latest build in <https://docs.rs/crate/bevy_tweening/0.3.1/builds.json> being successful.


services/docsrs/docsrs.service.js
```diff
@@ -54,7 +54,7 @@ export default class DocsRs extends BaseJsonService {
   async handle({ crate, version = 'latest' }) {
     const { build_status: buildStatus } = (
       await this.fetch({ crate, version })
-    ).pop()
+    ).shift()
```
calebcartwright, line 57: Would you mind changing this to just destructure the first item in the array? I realize we were already mutating it and that in practice this will typically only have one element, but there's really no need to mutate the collection the first place

calebcartwright (commented):
Thank you! Changes look good for me, though it made me realize there's another small change worth making too if you're up for it :pray:

calebcartwright (approved):

PR #7627, opened 2022-02-19
fix [githubpipenv] service test

calebcartwright:


chris48s:
Just from knowing the upstream project, that dependency is only pinned to that fork until a PR gets merged upstream so will break again soon.
I reckon we should update this to look at another project. Something like https://github.com/GSS-Cogs/databaker-docker/blob/master/Pipfile might be a good candidate

calebcartwright:
Good looking out, updated accordingly

chris48s (approved):

PR #7633, opened 2022-02-24
Add [galaxytoolshed] services

guillaume-gricourt:
Please considerate this PR to make badge from this API [Galaxy Toolshed](https://galaxyproject.org/toolshed/api/).
This PR lets to get the version of the tools, but if needed, this service may be extends to other features (downloads, ...).
Thanks


calebcartwright:
Thank you for the PR! I've not had a chance to go through the code, but did take a quick look at the review app and there was a couple things I wanted to go ahead and put on your radar.

https://shields-staging-pr-7633.herokuapp.com/

https://shields-staging-pr-7633.herokuapp.com:/galaxy-toolshed/v/sra_tools/iuc/fastq_dump


https://shields-staging-pr-7633.herokuapp.com:/galaxy-toolshed/vr/sra_tools/iuc/fastq_dump



(screenshot also provided since the review apps go to sleep fairly quickly and get deleted after a few days)

image

<br />
<br />

Specifically:

* Our badge defaults are based on our [Specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md), which includes things like not including branding/promotional material of the service/platform name. So in this case the label should be `version` instead of `galaxy-toolshed`
* We have common helpers for things like schema validation, transforming/extracting values from responses, and rendering that help simplify the implementation and incorporate spec-compliant defaults out of the box. Some more info can be found in our [tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) and by looking at other similar badges. I'm not sure whether those were used here but want to make sure you were aware in case not. e.g. https://github.com/badges/shields/blob/b23ea32088d4649c18f69219fda7a76ad5bd269e/services/apm/apm.service.js#L71
* We've similarly got conventions for our [badge routes](https://github.com/badges/shields/blob/master/doc/badge-urls.md). The `v` you've used for noun is perfect for the version badge, but the `vr` noun for the other is a little confusing. It seems a bit like a platform support or dependency version type of badge (e.g. like some of our existing badges in https://shields-staging-pr-7633.herokuapp.com/category/platform-support), so perhaps a more descriptive name might help?
* The label on the example for the second badge (the requirement-named one) looks off to me, both in general in terms of how we structure our labels, as well as compared to the resultant badge that's produced with the same parameters

image

chris48s:
> So in this case the label should be version instead of galaxy-toolshed

For version badges we normally use the name of the registry or platform as the badge label

Screenshot at 2022-02-27 22-11-14

calebcartwright:
> For version badges we normally use the name of the registry or platform as the badge label

Nuts that's right. @guillaume-gricourt you may disregard that portion from my prior comment

Chris - apologies if we've discussed this before (I feel like we have but maybe just déjà vu) but want to understand our plans for new badges going forward, perhaps an AOB item for next meeting. Definitely see + agree with the majority being the service/brand name, but also a significant subset that are just `version` or noun-equivalent (e.g. `tag` or `release`), even more so when the more platform-support type badges are ignored (e.g. Go or R version)

guillaume-gricourt:
Hi,
Thanks for your comment @calebcartwright and  @chris48s . In an attempt to provide a point-by-point answer:

- It's a good thing to mention your [Specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md), I missed this document.
- I used `versionText()` and `versionColor()` to manage label format
- This is the schema of the API. We have a repository (identified by a name and an owner) with one or several tools (with a name and a version). Each tool could have one or several requirements and each has a name and a version. It would be useful to grab version of a tool available in a repository and the version of a requirement corresponding to a tool. So, taking account of your request, I changed the route format. 
- Lastly, I agree, but I put the label format options in parameters.

chris48s:
Yeah I also think this has come up before, but I can't find the relevant PR and I'm not sure if we've clearly documented what we do here. Happy to get into it in more detail, but we can move it to a different issue/discussion.

guillaume-gricourt:
Ok, let me explain the use case. During our continuous deployment, we have some packages in:
- github: version badge available
- conda: version badge available
- galaxy-toolshed: version badge not available. This is the goal of this PR.

I've respected, as mush as possible, your [guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md)
@chris48s What do you mean about : 


> I'm not sure if we've clearly documented what we do here.


Could you precise what was happened before ?

calebcartwright:
> @chris48s What do you mean about :
> 
> > I'm not sure if we've clearly documented what we do here.
> 
> Could you precise what was happened before ?

This was in response to my question in the prior comment, and relates to some past conversations amongst the maintainer team. There's not anything actionable needed from you on that front, we've just identified an off-topic item that we as maintainers want to discuss offline.

guillaume-gricourt:
> 

Ok, thanks,

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
```diff
@@ -0,0 +1,70 @@
...
+import { formatDate } from '../text-formatters.js'
+import { age as ageColor } from '../color-formatters.js'
+import BaseGalaxyToolshedService from './galaxy-toolshed-base.js'
+
+const repositoryRevisionInstallInfoSchema = Joi.array().items(
+  Joi.object({
+    create_time: Joi.string().required(),
+  }).required(),
+  Joi.object({}),
+  Joi.object({})
```
calebcartwright, line 11: This is a very odd looking schema definition. For the sake of facilitating discussion on the PR could you please post a sample API response that illustrates the necessity of this particular schema structure?

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
```diff
@@ -0,0 +1,70 @@
...
+)
+
+export default class GalaxyToolshedActivity extends BaseGalaxyToolshedService {
```
calebcartwright, line 14: Unless we anticipate additional metrics/badges being surfaced via this class, let's rename both the file and class name to be more explicit about the badge as opposed to the category, e.g. `GalaxyToolshedReleaseDate` or `GalaxyToolshedCreateDate`

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
```diff
@@ -0,0 +1,70 @@
...
+      },
+      staticPreview: this.render({
+        createDate: new Date(0).setUTCSeconds(1538288239),
+      }),
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'create date',
+  }
```
calebcartwright, line 36: I realize that the underlying field name in the response is `create_time`, however, we're not obliged to mirror the wording of upstream providers. In this case I find `create date` to sound odd, and would suggest we use `release date` instead, which will be far more consistent with the other similar types of badges we provide for other services.

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
```diff
@@ -0,0 +1,70 @@
...
+export default class GalaxyToolshedActivity extends BaseGalaxyToolshedService {
+  static category = 'activity'
+  static route = {
+    base: 'galaxy-toolshed/create-date',
+    pattern: ':repositoryName/:owner',
+  }
+
+  static examples = [
+    {
+      title: 'Galaxy Toolshed - Create Time',
```
calebcartwright, line 23: Similarly here, we've stuck with the response field name of `Time` even though we're displaying a date. This should be updated to be consistent with whatever term we go with, see below comment

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  static transform(response) {
+    const metadata = response
+      .filter(function (x) {
+        return Object.keys(x).length > 0
+      })
+      .shift()
+    return metadata.create_time
+  }
```
calebcartwright, line 63: Perhaps there's a subtle distinction I've overlooked, but this entire block seems to be duplicated across all of these classes with the only differentiator being the key name used in the last line of the `transform` function. This seems like a prime candidate for simplification and consolidation by extracting and encapsulating into the base class or some other shared function definition with a signature that accepts any additional parameters (e.g. metadataFieldName) necessary to handle the minor differences

services/galaxy-toolshed/galaxy-toolshed-activity.tester.js
```diff
@@ -0,0 +1,19 @@
+import { isFormattedDate } from '../test-validators.js'
+import { ServiceTester } from '../tester.js'
+
+export const t = new ServiceTester({
+  id: 'galaxy-toolshed',
+  title: 'GalaxyToolshed',
+})
```
calebcartwright, line 7: Please use the helper functions for the boilerplate items, more details in our [guide on writing service tests](https://github.com/badges/shields/blob/master/doc/service-tests.md#1-boilerplate)

services/galaxy-toolshed/galaxy-toolshed-base.js
```diff
@@ -0,0 +1,30 @@
...
+
+export default class BaseGalaxyToolshedService extends BaseJsonService {
+  static defaultBadgeData = { label: 'galaxy-toolshed' }
+  static baseUrl = 'https://toolshed.g2.bx.psu.edu'
+
+  async fetchOrderedInstallableRevisionsSchema({ repositoryName, owner }) {
+    return this._requestJson({
+      schema: orderedInstallableRevisionsSchema,
+      url: `${this.constructor.baseUrl}/api/repositories/get_ordered_installable_revisions?name=${repositoryName}&owner=${owner}`,
+    })
```
calebcartwright, line 16: Can the `repositoryName` or `owner` params ever have values that would need to be uri encoded? This should depend on what all may or may not be allowed upstream on the galaxy toolshed service

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
```diff
@@ -0,0 +1,70 @@
...
+        repositoryName,
+        owner,
+      })
+    return this.fetchRepositoryRevisionInstallInfoSchema({
+      schema: repositoryRevisionInstallInfoSchema,
+      repositoryName,
+      owner,
+      changesetRevision: changesetRevisions.shift(),
+    })
+  }
```
calebcartwright, line 54: Having to make two sequential API calls like this is generally frowned upon. The reason is that we have a very limited window in which to complete the entire badge request/response flow in order to get the badge back to the user in time before the renderers in common places like GitHub timeout and refuse to show the badge.

I gather you're taking this sequential approach to first get the latest version and then fetch the stats for that specific version, but is there a different endpoint that would return the stats for all versions? It's a lot easier for us to do a tad more processing (and we've a ton of utility functions for this) as opposed to dealing with multiple networking round trips just to fetch data from the provider

services/galaxy-toolshed/galaxy-toolshed-downloads.service.js
```diff
@@ -0,0 +1,67 @@
...
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'downloads',
+  }
+
+  static render({ downloads }) {
+    return renderDownloadsBadge({ downloads })
+  }
```
calebcartwright, line 37: Nit but you can skip this extra function definition and associated call, just invoke `renderDownloadsBadge` directly

services/galaxy-toolshed/galaxy-toolshed-version.service.js
```diff
@@ -0,0 +1,163 @@
...
+      .required(),
+  }).required(),
+  Joi.object({})
+)
+
+export default class GalaxyToolshedVersion extends BaseGalaxyToolshedService {
+  static category = 'version'
+  static route = {
+    base: 'galaxy-toolshed/v',
+    pattern: ':repositoryName/:owner/:toolId?/:requirementName?',
```
calebcartwright, line 36: This is just too much to try to cram into a single badge, both in terms of the user facing surface of the badge url and in terms of the code. It also ends up accepting invalid paths (e.g. toolId is strictly required for showing the version of a tool's requirement). 

There are 3 very distinct badges here, and we need to reflect our code as such

services/galaxy-toolshed/galaxy-toolshed-version.service.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+        return [...prev]
+      }, [])
+    },
+    [])
+    if (typeof versions === 'undefined' || !versions.length) {
+      throw new NotFound({ prettyMessage: 'requirement not found' })
+    }
+    return { label: requirementName, version: versions.shift() }
+  }
```
calebcartwright, line 143: Similar thread as above, there's too much going and this needs to be split and cleaned up.

calebcartwright (commented):
Thanks again for the PR! I've done an initial pass and while the bones are there I've left a number of inline comments, some with questions and some with things that need to be changed.

There's also a lot of comments that are applicable across services and testers which I didn't repeat in each one, but please review those relevant comments against all.

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
guillaume-gricourt: Yes, maybe a file is more appropriated: [get_repository_revision_install_info.sra_tools.iuc.b723c120161a.txt](https://github.com/badges/shields/files/8409503/get_repository_revision_install_info.sra_tools.iuc.b723c120161a.txt)

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
guillaume-gricourt: The API grabs only the first date of the release, not the last. Should I change it anyway ?

services/galaxy-toolshed/galaxy-toolshed-activity.tester.js
guillaume-gricourt: I tried but the character "-" in the service name raise an error. So, it's inspired from `f-froid` service

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
guillaume-gricourt: You are right, I will update the code

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
guillaume-gricourt: You are right, I will update the code

services/galaxy-toolshed/galaxy-toolshed-downloads.service.js
guillaume-gricourt: Ok, I will update the code

services/galaxy-toolshed/galaxy-toolshed-base.js
guillaume-gricourt: From the documentation:
- owner: *Public names must be at least three characters in length and contain only lower-case letters, numbers, dots, underscores, and dashes ('.', '_', '-')*
- repositoryName: *should contain only [a-z0-9_-]* and *Multiple words should be separated by underscore or dashes*

services/galaxy-toolshed/galaxy-toolshed-version.service.js
guillaume-gricourt: Ok, I will manage to have 3 different routes

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
calebcartwright: That was my understanding as well, and that's why I suggested `release date` as that's what is used for other badges (like the VS Marketplace). An alternative might to simply use `created`

services/galaxy-toolshed/galaxy-toolshed-activity.tester.js
calebcartwright: F-Droid is a bad point of comparison because the hyphen is actually part of the service name (the service name is literally "F-Droid"), not a hyphen we introduce into the file name convention.

That's not the case here as I understand it; Galaxy and Toolshed are separate, non-hyphenated words. As such we should not have a `-` in the class name, which in turn won't cause any issues with the service registration. If you follow the standard prescriptions for naming the file and the service class (e.g. `GalaxyToolshedDownloads`) things should just work

services/galaxy-toolshed/galaxy-toolshed-activity.service.js
calebcartwright: Fair enough. Once you get a chance to rebase we should be able to update the review app and can start getting a ballpark feel for the performance

guillaume-gricourt:
I believe the name written with the hyphen character creates some troubles. I changed the name "galaxy-toolshed" to "galaxytoolshed". Maybe reload the tests could avoid this behavior ?

andreped:
> I changed the name "galaxy-toolshed" to "galaxytoolshed"

I hope it is okay if I make a comment here. I have the exact same issue when catching dowloads/month in a shield when a package name has a hyphen in it. Is there a way around it, which avoids the removing it altogether, @guillaume-gricourt? I have tried replacing it with "--".

guillaume-gricourt:
No, I didn't find a suitable way.

andreped:
> No, I didn't find a suitable way.

No worries. Perhaps I should add a separate issue about this. Thanks, for the fast reply!

guillaume-gricourt:
Hello,
What can I do to improve the PR ?

calebcartwright:
> Hello, What can I do to improve the PR ?

No this just fell off the radar, apologies. It might be worth doing a rebase though

calebcartwright:
I keep trying to come back to this but keep remembering that it needs more time than I'm typically able to allocate in one sitting. I think part of the challenge is that this PR is on the larger side because it's adding an entire family of badges in one fell swoop.

I think the best way to move forward with this would be to break it up into more manageable chunks that can be addressed iteratively. I'd recommend we start with the downloads badge, and if you'd like to include the base class as part of that initial PR that's fine too. Once we get that landed then you can open separate PRs for the other individual badges so that they, and their implementations, can be reviewed and discussed independently.

I believe that if we go that route then we should be able to get the first one addressed in a much more straightforward and hopefully expeditious fashion.

PR #7655, opened 2022-02-27
generate links without trailing : if port not set

chris48s:
This is a minor thing, but at the moment our staging app and review apps generate links like this:
https://shields-staging.herokuapp.com:/appveyor/build/gruntjs/grunt
https://shields-staging-pr-7633.herokuapp.com:/galaxy-toolshed/v/sra_tools/iuc/fastq_dump
(with a trailing `:` character after the `.com`). They work, but I find it a minor irritation


PyvesB (approved):
No more minor irritations!

PR #7656, opened 2022-02-27
Add [piwheels] version badge

### LINKED ISSUE #6506: Badge request: piwheels package version
:clipboard: **Description**

Show the current version of a Python package on piwheels.org:

```
piwheels | 0.3.0
```

:link: **Data**

piwheels has a JSON API, documented here: https://www.piwheels.org/json.html

:microphone: **Motivation**

piwheels.org is a mirror of PyPI, the Python Package Index, with packages rebuilt on Raspberry Pi OS for ARM CPUs. I'd like to include a "piwheels" badge to show support for piwheels and also to allow me to spot if piwheels has fallen behind PyPI (because a package is not building successfully for Raspberry PI).

chris48s:
Closes #6506

I don't often do PRs adding a new badge these days, but I'd like this one to exist :)
The transform method is fairly comprehensively unit tested, so I've kept the live tests fairly light-touch.

### DIFF
--- a/services/piwheels/piwheels-version.service.js
+++ b/services/piwheels/piwheels-version.service.js
@@ -1,0 +1,94 @@
+import Joi from 'joi'
+import { BaseJsonService, InvalidResponse } from '../index.js'
+import { renderVersionBadge } from '../version.js'
+
+const schema = Joi.object({
+  releases: Joi.object()
+    .pattern(
+      Joi.string(),
+      Joi.object({
+        prerelease: Joi.boolean().required(),
+        yanked: Joi.boolean().required(),
+      })
+    )
+    .required(),
+}).required()
+
+const queryParamSchema = Joi.object({
+  include_prereleases: Joi.equal(''),
+}).required()
+
+const keywords = ['python', 'arm', 'raspberry pi']
+
+export default class PiWheelsVersion extends BaseJsonService {
+  static category = 'version'
+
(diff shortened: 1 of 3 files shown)


services/piwheels/piwheels-version.service.js
```diff
@@ -0,0 +1,94 @@
...
+      )
+      .filter(release => !release.yanked) // exclude any yanked releases
+
+    if (allReleases.length === 0) {
+      throw new InvalidResponse({ prettyMessage: 'no versions found' })
+    }
+
+    if (includePrereleases) {
+      return allReleases[0].version
+    }
```
calebcartwright, line 79: I didn't see anything in the docs indicating a sort order, so figured I'd double check to confirm they are?

calebcartwright (approved):
One inline question but also approving as I assume it's a simple answer/confirmation

services/piwheels/piwheels-version.service.js
chris48s: Yeah its a good question. I did look over the source for piwheels. The code that generates the HTML pages and the JSON API does parse all the versions to PEP440 version objects and then sorts them. https://github.com/piwheels/piwheels/blob/fd6f3d12d9a33390ccb235a2995edb24a8fa782d/piwheels/master/the_scribe.py#L525-L542 I'm happy to rely on the ordering, even though it is not explicitly called out in the API docs.

Issue #7020, opened 2021-09-14
Packagist badge cannot handle composer v2-only packages


SteveDesmond-ca:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Packagist (the PHP dependency management SaaS) deprecated their v1 API some time ago, so new packages are only being added to their v2 API.

The shields site uses v1 of the API, so it is not able to retrieve version numbers for newer packages.

### 🔗 Link to the badge

New package, not working: ![ecoapm/libyear](https://img.shields.io/packagist/v/ecoapm/libyear)

Old package, works fine: ![symfony/symfony](https://img.shields.io/packagist/v/guzzlehttp/guzzle)

### 💡 Possible Solution

There are a couple possible options:
1. update `BasePackagistService.fetch()` to use the new API and modify `transform()` accordingly
1. update `PackagistVersion.handle()` to fallback to the JSON API

I would think the first would be preferrable, from a future maintenance perspective, so I'll get a PR ready for that tomorrow.

SteveDesmond-ca:
I started diving in already, it looks like the v2 API has some pretty gnarly breaking changes so maybe holding off to have a discussion on the best way forward is better...

chris48s:
If you're interested in working on this, have a look over https://github.com/badges/shields/pull/6508

Nobody has done anything on it for a while and but from memory there was not too much work left to do after the last round of comments to get it landed. Unfortunately the original author stopped responding (tbh it is one of those PRs that seemed simple but turned into a bit of a con of worms).

What's there is probably a pretty decent starting point if you wanted to try and get it over the line.

SteveDesmond-ca:
Logging in this morning, the deprecation warning has been removed from my package and it now shows up via the v1 API (and therefore shields.io badge, see OP) so from an end-user perspective, if newly-created packages just need to wait until the next day to show up, I don't think that's a huge deal, but I can see from a technical perspective why we wouldn't want to be using a deprecated endpoint.

kdambekalns:
As explained on https://blog.packagist.com/deprecating-composer-1-support/

> This means new packages and old unused packages will only be visible to Composer 2 until someone installs them, at which point they will become visible to Composer 1 as well (within 24h as this process happens on a nightly basis).

So, yeah, not that urgent. Just a bit confusing.

PR #7575, opened 2022-02-05
[packagist] api v2 support

yookoala:
Rebased #6508 on the current master branch. An attempt to fix #7020.

I have a packagist v2 only package ([this one](https://packagist.org/packages/monogatari-php/remote-storage)). The v1 API doesn't really show any information about my package. I'd want to fix this so badges for my package works normally. Please let me know what I need to do to get this merged.

---

Update: Seems the badge for my package was affected by the [reduced update rate of v1 Packagist API](https://blog.packagist.com/deprecating-composer-1-support/#reduced-v1-metadata-api-update-rate-starting-in-may-2021). It works for me now. But I'd still like to contribute and fix this for long run.

PR #7681, opened 2022-03-04
[packagist] api v2 support

chris48s:
Closes https://github.com/badges/shields/issues/7020.
As discussed in https://github.com/badges/shields/pull/7575#issuecomment-1058497818 , lets merge this bit and use it as a baseline for further work
@calebcartwright  - are you good to give this one a quick :+1: ?


calebcartwright:
> @calebcartwright - are you good to give this one a quick +1 ?

Yeah will give it a quick pass after work today

chris48s:
Yeah no worries.

Just to clarify, these commits take us up to the point in the previous PR where you left the comments

>  Was going to ask if it'd be worth adding some unit tests for this function since the logic, though not terribly complex, isn't exactly trivial either.
> I see that we're indirectly testing it with tests that targetting some of the child classes. That works for me, at least as long as we don't end up duplicating tests down the road

and

> I believe my mental context is fresh enough at the moment to follow this, but if anyone is receptive and willing to do so in a follow up PR, I think a brief inline comment summarizing the what/why would be helpful for future readers (likely myself included)

last time.

calebcartwright (approved):

PR #5592, opened 2020-09-22
[PyPI] When Python version classifiers are absent, fall back to `requires_python`

paulmelnikow:
Ref #5551
Alternative to #5576

Let's keep the discussion in #5551 until we agree on how to move forward.

PR #7682, opened 2022-03-04
Improve python version formatting 🐍 affects [pypi piwheels github]

chris48s:
After implementing the piwheels badge the other day I realised that I had just used `renderVersionBadge()` but our standard version formatter
https://github.com/badges/shields/blob/76a64a7896893bac4a1b52e683a0a542541b3bdb/services/color-formatters.js#L16-L20
isn't great for python packages because it will format something like `2.0.0rc1` or `2.0.0b2` as "stable".

Then I started pulling the thread and realised this problem actually also applies to some other python ecosystem badges.

so in this PR I have:

- Added a `pep440VersionColor()` formatter function which uses https://github.com/renovatebot/pep440 to parse versions
- Added the ability to pass `renderVersionBadge()` a custom formatter function as an optional param - this might be handy if we want to use a custom formatter for other package ecosystems.
- Used this in the badges for pypi, piwheels and github-pipenv

### DIFF
--- a/services/github/github-pipenv.service.js
+++ b/services/github/github-pipenv.service.js
@@ -1,3 +1,4 @@
+import { pep440VersionColor } from '../color-formatters.js'
 import { renderVersionBadge } from '../version.js'
 import { isLockfile, getDependencyVersion } from '../pipenv-helpers.js'
 import { addv } from '../text-formatters.js'
@@ -80,6 +81,7 @@ class GithubPipenvLockedPythonVersion extends ConditionalGithubAuthV3Service {
       version,
       tag: branch,
       defaultLabel: 'python',
+      versionFormatter: pep440VersionColor,
     })
   }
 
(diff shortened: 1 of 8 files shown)


services/github/github-pipenv.service.js
```diff
@@ -147,7 +149,7 @@ class GithubPipenvLockedDependencyVersion extends ConditionalGithubAuthV3Service
     return {
       label: dependency,
       message: version ? addv(version) : ref,
-      color: 'blue',
+      color: version ? pep440VersionColor(version) : 'blue',
     }
```
PyvesB, line 153: Could we use `renderVersionBadge` here, either with a conditional or ternary operator? For example:
````
return version ? renderVersionBadge({ version, versionFormatter: pep440VersionColor }) 
               : { label: dependency, message: ref, color: 'blue' }
````

PyvesB (dismissed):
Looks good to me. 👍🏻 Interestingly , @paulmelnikow  intended on using pep440 in  #5592 as well.

services/github/github-pipenv.service.js
chris48s: We'd have to extend `renderVersionBadge()` to also take a custom label. For this badge the label is the name of the locked dependency. Tbh I'm inclined to leave it

chris48s:
I've rebased this to resolve the conflicts on package.json and package-lock.json as it was easier than doing it as a merge. No functional changes

PyvesB (approved):

PR #7686, opened 2022-03-05
Fix [SnykVulnerability] test

PyvesB:
The fix from #7629 didn't last for long, the newly targeted repository has also been made private or otherwise removed. Targeting a file that hasn't moved since 2015 in the main Snyk repository should hopefully last a little longer.


calebcartwright (approved):
fwiw the resolve-deps-fixtures repo had been there since~2015/2016 too. I'm not going to hold my breath just yet as it seems like Snyk are overhauling their public repo surface/example projects, but should be as good a target as any

Issue #5941, opened 2020-12-12
Badge request: (standardised) open-source project Criticality Score


sadielbartholomew:
:clipboard: **Description**

* This is a badge for GitHub.
* The badge would give the [Criticality Score metric (a float between 0 and 1) as calculated from the standard (non-customised) algorithm](https://github.com/ossf/criticality_score#criticality-score) so as to be fairly comparable across projects.
* There could be incremental colour grading based on the value (1 as the ideal best to 0 as the worst score) to imply the standard of the criticality at a glance. 
* The format that seems natural to me for the badge is `criticality score: N` where N is the score between 0 and 1, noting that strictly the name Criticality Score is capitalised (going by the usage of the creators) but for consistency with the shields style it should be in lower case.

:link: **Data**

* API is provided via the public [ossf/criticality_score](https://github.com/ossf/criticality_score) repository.
* A GitHub access token is required.
* There is basic documentation provided in [the repository README](https://github.com/ossf/criticality_score/blob/main/README.md) but it looks like it could easily be sufficient.

:microphone: **Motivation**

The Open Source Project Criticality Score is a recently-announced metric created by Google's OpenSSF, as detailed [in this dedicated blog post](https://opensource.googleblog.com/2020/12/finding-critical-open-source-projects.html). Having this metric in a badge:
* would be a very useful way for someone looking at a public project on GitHub to gauge immediately that is objectively rated as being influential and important (or otherwise), with the 0 to 1 nature being very simple to comprehend;
* would allow projects rated as critical to advertise that they are deemed critical and gain further esteem based on that;
* would facilitate inter-comparison of the criticality of projects.

calebcartwright:
Thanks for the suggestion @sadielbartholomew. While it does seem like a bade that would be good to support, I'm not sure it's really a viable candidate in its current state.

In order for us to be able to provide a badge we need something like a rest endpoint that quickly provides machine readable quickly which we then use to construct the badges. Our badge servicing process is primarily i/o bound on the fetching of the target data points from the systems of record, as we don't produce or calculate any such raw data points ourselves nor do we really run other tools/binaries on our production environment which do.

AFAICT what exists today for the criticality score is a command line tool which makes multiple requests to GitHub to fetch various data points that are then used for calculating the value, but that's not really something we can utilize for getting the criticality score for badges.

[...]

Issue #5506, opened 2020-09-09
Badge request: Flathub downloads


tooomm:
:clipboard: **Description**

- Which service is this badge for e.g: GitHub, Travis CI
Flathub: https://flathub.org
- What sort of information should this badge show?
Download count of a particular package. Currently only a version badge is supported.

- Provide an example in plain text e.g: "version | v1.01" or as a static badge
(static badge generator can be found at https://shields.io)
<kbd>Flathub Downloads | 123.456</kbd>

:link: **Data**

Where can we get the data from?

There is all the data at https://flathub.org/stats/ in various json files.
This page uses the data already: https://klausenbusk.github.io/flathub-stats/

:microphone: **Motivation**

- What is the specific use case?

Showing useful download data like all time, recent month etc. as with other services.

chris48s:
I think the data in the structure at https://flathub.org/stats/ isn't workable for us: In order to query that on-the-fly in the form the data is provided (1 json file per day) we would have to make many network requests to get the stats for a single package over a period of time (i.e: to do downloads for 1 month we'd have to make ~30 requests).

It looks like https://klausenbusk.github.io/flathub-stats/ relies on a GH actions ETL job which runs on a schedule, harvests it all and writes it all back out to github in one-file-per-package format rather than one-file-per-day format: https://github.com/klausenbusk/flathub-stats/tree/gh-pages/data

JakobDev:
There is now a API for that: https://flathub.org/api/v2/stats/{appid}. e.g to query the Stats of Firefox: https://flathub.org/api/v2/stats/org.mozilla.firefox

JakobDev:
I created a Issue at Flathub to ask if it is safe to use. If not, we should take a look at https://klausenbusk.github.io/flathub-stats/ which linked above. We also may have a Problem with the Flathub version Badge, which uses Version 1 of the API.

tooomm:
Thanks for adding this in #7724 @JakobDev! It works nicely. ❤️ 

I'm just wondering why you label it `downloads` if it's "only" `installs`. Downloads is misleading here IMHO, users don't know that flathub differentiates here further.
I think the badge would be more clear and helpful, if installs+updates stats would be combined, or the label states which subsection we display. Looks like in the flathub website repo and related issues (https://github.com/flathub/website/pull/67) it got mentioned as well... and they finally went with `Installs` as seen in the beta, e.g: https://beta.flathub.org/en/apps/details/org.mozilla.firefox
[...]

PR #7687, opened 2022-03-06
[OSSFScorecard] Create scorecard badge service

rohankh532:
Badge service to display scores from the [OpenSSF Security Scorecard.](https://github.com/ossf/scorecard)


PyvesB:
Hello @rohankh532 ! 👋🏻 

It would have been helpful to add some context about the Scorecard service, what these badges are trying to achieve, how they would be used, etc.

I've not looked at your implementation in details, but I did notice your `score (valid)` test which points to `github/org/repo`. The https://github.com/org/repo repository clearly isn't valid. The API seems to return sane-looking data regardless of what parameters are passed to it: https://api.securityscorecards.dev/projects/github/dqshsjqkhdqsjdhqsghd44/does-not-exist.json. I've also tried a bunch of repositories listed in https://github.com/ossf/scorecard/blob/main/cron/data/projects.csv, I've not yet managed to get any response that didn't have a score of 1, including for the Scorecard repository itself.

Is the API actually working? Is it documented anywhere?

calebcartwright:
Believe #5941 was a request along the same lines.

For what it's worth, I do think many of OSSF-based metrics would make for good badges, but also agree that we need to make sure the APIs we'd be depending on are in good order as P-Y noted

rohankh532:
> Hello @rohankh532 ! 👋🏻
> 
> It would have been helpful to add some context about the Scorecard service, what these badges are trying to achieve, how they would be used, etc.
> 
> I've not looked at your implementation in details, but I did notice your `score (valid)` test which points to `github/org/repo`. The https://github.com/org/repo repository clearly isn't valid. The API seems to return sane-looking data regardless of what parameters are passed to it: https://api.securityscorecards.dev/projects/github/dqshsjqkhdqsjdhqsghd44/does-not-exist.json. I've also tried a bunch of repositories listed in https://github.com/ossf/scorecard/blob/main/cron/data/projects.csv, I've not yet managed to get any response that didn't have a score of 1, including for the Scorecard repository itself.
> 
> Is the API actually working? Is it documented anywhere?

Scorecards is a tool that assesses a repository with several security checks and provides a score for different categories. It is trigger via a GitHub workflow, which would run every time the project was updated. We are currently incorporating signatures on scorecard results so that their authenticity can be proved. This way, if someone "claims" their project has a certain score, it can actually be verified.

We are using badges to allow users to display their scorecard score so that other users can get a sense for how vulnerable the project is. The API currently just returns 1 for every repository as we are still working on a method for computing an aggregate score. So for now it is just a placeholder, but the format will still remain the same.

PyvesB:
Thanks for sharing the extra details! If the API is still work in progress, I suggest we put this PR on hold for the time being, and circle back to it once things are in a more final state. How does that sound?

rohankh532:
> Thanks for sharing the extra details! If the API is still work in progress, I suggest we put this PR on hold for the time being, and circle back to it once things are in a more final state. How does that sound?

Sounds good. I'll change this PR to a draft for now.

rohankh532:
The API is now functional, here are 3 example endpoints for which the scorecard results are given in response:
1. https://api.securityscorecards.dev/projects/github.com/rohankh532/scorecard-OIDC-test
2. https://api.securityscorecards.dev/projects/github.com/rohankh532/shields
3. https://api.securityscorecards.dev/projects/github.com/rohankh532/org-workflow-add

And the corresponding badges:
image
image
image

calebcartwright:
Great news! Could you please rebase and make sure the CI jobs are passing?

rohankh532:
> Great news! Could you please rebase and make sure the CI jobs are passing?

Thanks! Should be fixed now.

rohankh532:
@calebcartwright Is there any update on this?

PyvesB:
I'll aim at reviewing the PR this weekend. We're a small maintainer team, who dedicate their limited spare time to Shields for free. We do our best, but response times can be a little inconsistent. 😉

rohankh532:
> I'll aim at reviewing the PR this weekend. We're a small maintainer team, who dedicate their limited spare time to Shields for free. We do our best, but response times can be a little inconsistent. 😉

No problem at all, thank you!

services/scorecard/scorecard.service.js
```diff
@@ -0,0 +1,41 @@
+import Joi from 'joi'
```
calebcartwright, line 1: We should use a more explicit, descriptive name that represents the actual upstream service. "Scorecard" is very generic and we can't just assume/assign that generic term to mean OSSF's score card.

Please rename the directory, files, and class accordingly

services/scorecard/scorecard.service.js
```diff
@@ -0,0 +1,41 @@
...
+import { BaseJsonService } from '../index.js'
+
+const schema = Joi.object({
+  score: Joi.required(),
+}).required()
+
+export default class ScorecardScore extends BaseJsonService {
+  static category = 'analysis'
+
+  static route = { base: 'scorecard', pattern: ':host/:orgName/:repoName' }
```
calebcartwright, line 11: Does the upstream service support branch-specific scans or is it only at the repo level?

Is there a fixed set of providers it integrates with (e.g. GitHub, GitLab, Bitbucket, etc.) or is it truly open ended?

services/scorecard/scorecard.service.js
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+      clr = 'green'
+    }
+
+    return {
+      label: 'scorecard score',
+      message: score,
+      color: clr,
+    }
+  }
```
calebcartwright, line 40: This is all related to rendering concerns and we have a very specific pattern we follow for this.

I'd encourage you to read through our [tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) on adding new badges which covers this as well as another piece you are missing which are the example listings that make the badge show up on Shields.io

services/scorecard/scorecard.service.js
```diff
@@ -0,0 +1,41 @@
...
+
+    if (score <= 3) {
+      clr = 'red'
+    } else if (score <= 5) {
+      clr = 'orange'
+    } else if (score <= 7) {
+      clr = 'yellow'
+    } else {
+      clr = 'green'
+    }
```
calebcartwright, line 33: The scaled mapping of a numeric value to a range of colors is, as you might imagine, fairly common within Shields and we've a number of reusable helpers for such circumstances.

In this instance, I'd suggest taking a look at `colorScale` which can be imported from `color-formatters.js`, and [our code climate analysis class](https://github.com/badges/shields/blob/master/services/codeclimate/codeclimate-analysis.service.js) may serve as a helpful reference

services/scorecard/scorecard.service.js
```diff
@@ -0,0 +1,41 @@
...
+
+    return {
+      label: 'scorecard score',
```
calebcartwright, line 36: `score` should be more than sufficient here, though note that static attributes of the final badge can and should be set in the `defaultBadgeData` function (also covered in more detail in the aforementioned Tutorial)

services/scorecard/scorecard.tester.js
```diff
@@ -0,0 +1,10 @@
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('score (valid)')
+  .get('/github.com/rohankh532/org-workflow-add.json')
+  .expectBadge({
+    label: 'scorecard score',
+    message: '4.2',
+    color: 'orange',
+  })
```
calebcartwright, line 10: It's good to have a live test (see also our [guide on writing service tests](https://github.com/badges/shields/blob/master/doc/service-tests.md)), but we don't typically want to pin a live test to a specific value because that can make the test brittle. Instead, we follow a similar pattern as we do for validating response schemas and define the _type_ of result we'd expect to see, typically with another schema.

In this case I think `message: Joi.number().min(0)` would suffice instead of the explicit `4.2` value

calebcartwright (commented):
Thanks for the PR! I've left feedback inline below with guidance on changes needed for next steps

services/scorecard/scorecard.service.js
Rohan6789: The scans are branch specific. Currently, it will only be integrated with GitHub, but there are future plans to expand it to GitLab and BitBucket

services/scorecard/scorecard.service.js
Rohan6789: Fixed this using color-formatters.js, thanks for the reference

services/scorecard/scorecard.tester.js
Rohan6789: Fixed this, along with assuming the color value

services/scorecard/scorecard.service.js
rohankh532: The scans are branch specific. Currently, it will only be integrated with GitHub, but there are future plans to expand it to GitLab and BitBucket

services/scorecard/scorecard.service.js
rohankh532: Fixed this using color-formatters.js, thanks for the reference

services/scorecard/scorecard.tester.js
rohankh532: Fixed this, along with assuming the color value

rohankh532:
> Thanks for the PR! I've left feedback inline below with guidance on changes needed for next steps. Also, is there any publicly available documentation for the API endpoint?

Thanks for the feedback! The endpoint is publicly defined [here](https://api.securityscorecards.dev/) for now. More detailed documentation is WIP.

Also, I'm having issues with the circleci package test. I haven't changed any dependencies and it's erroring on Node 14. Can't seem to fix this.

calebcartwright:
> Also, I'm having issues with the circleci package test. I haven't changed any dependencies and it's erroring on Node 14. Can't seem to fix this.

Don't worry about this for now as it's definitely not related to your changes (more info [here](https://github.com/badges/shields/pull/7788))

calebcartwright:
> The endpoint is publicly defined [here](https://api.securityscorecards.dev/) for now. More detailed documentation is WIP.

Thanks but the endpoint being internet accessible is not the same thing as the endpoint being documented or made officially available. That is something that we would need before being able to move forward, or at least get some kind of official/authoritative blessing from the OSSF project. There's a number of reasons why we require a level of "official-ness" for the upstream APIs we consume, with many of them well summarized [here](https://github.com/badges/shields/issues/5506#issuecomment-1067129626) for reference

azeemshaikh38:
> Thanks but the endpoint being internet accessible is not the same thing as the endpoint being documented or made officially available. That is something that we would need before being able to move forward, or at least get some kind of official/authoritative blessing from the OSSF project. There's a number of reasons why we require a level of "official-ness" for the upstream APIs we consume, with many of them well summarized [here](https://github.com/badges/shields/issues/5506#issuecomment-1067129626) for reference

Hi @calebcartwright, I'm one of the admins on the Scorecard project and part of the OSSF WGs. We plan to officially announce both the Scorecard API and the Scorecard badges to the public soon and I can vouch for the officialness of this project both from the Scorecard and OSSF-side. We would like to do a pre-release of these features to some OSSF repositories first and gather feedback before announcing to the larger public. So getting this PR merged would unblock us from doing that. Let me know if this helps.

calebcartwright:
Thanks @azeemshaikh38! That should provide us sufficient cover for consuming the endpoint. 

> We would like to do a pre-release of these features to some OSSF repositories first and gather feedback before announcing to the larger public. So getting this PR merged would unblock us from doing that.

I'm not entirely sure I understood this part. Are you suggesting that our consumption of your API as part of our delivery of badges to your users is something that would be beneficial for you as part of testing/validation/feedback of your API?

calebcartwright:
@rohankh532 you should be able to rebase now to pull in the fix for the CI issue

rohankh532:
> @rohankh532 you should be able to rebase now to pull in the fix for the CI issue

Yup its fixed now, thanks!

azeemshaikh38:
> I'm not entirely sure I understood this part. Are you suggesting that our consumption of your API as part of our delivery of badges to your users is something that would be beneficial for you as part of testing/validation/feedback of your API?

We are looking to test/feedback both - API and badge. They are both new features rolling out to the public. The way this works is - when you "enable" Scorecard on a repository, you get an API which provides latest Scorecard result and the ability to add a Scorecard badge to your repo. So we'll be collecting feedback on user journey, badge design etc.

calebcartwright:
> The way this works is - when you "enable" Scorecard on a repository, you get an API which provides latest Scorecard result and the ability to add a Scorecard badge to your repo. So we'll be collecting feedback on user journey, badge design etc.

I guess that's why this is returning an http 500 response

> $ curl -i  https://api.securityscorecards.dev/projects/github.com/badges/shields

As an early piece of feedback, I'll share that's semi-problematic for us, though not necessarily blocking. We typically try to map the http errors we receive when trying to fetch data from the upstream provider to something more contextual to our badge users. However, because this is returning a 500 response we have no way of distinguishing between whether there's a legitimate upstream service issue or the user hasn't enabled their repository or if they just had a typo in the badge route parameter or something else.

azeemshaikh38:
> As an early piece of feedback, I'll share that's semi-problematic for us, though not necessarily blocking. We typically try to map the http errors we receive when trying to fetch data from the upstream provider to something more contextual to our badge users. However, because this is returning a 500 response we have no way of distinguishing between whether there's a legitimate upstream service issue or the user hasn't enabled their repository or if they just had a typo in the badge route parameter or something else.

Ah, that's a good point. I think returning `500` is a bug really. We should return a `404` in this case instead to indicate the resource was not found. Let me know what you think. 

@rohankh532 could you update the API to ensure that we don't return a `500` error status here? If the project URL is valid and we don't have any data for it, we should return `404`.

rohankh532:
> > As an early piece of feedback, I'll share that's semi-problematic for us, though not necessarily blocking. We typically try to map the http errors we receive when trying to fetch data from the upstream provider to something more contextual to our badge users. However, because this is returning a 500 response we have no way of distinguishing between whether there's a legitimate upstream service issue or the user hasn't enabled their repository or if they just had a typo in the badge route parameter or something else.
> 
> Ah, that's a good point. I think returning `500` is a bug really. We should return a `404` in this case instead to indicate the resource was not found. Let me know what you think.
> 
> @rohankh532 could you update the API to ensure that we don't return a `500` error status here? If the project URL is valid and we don't have any data for it, we should return `404`.

Fixed this to differentiate between `404` and `500` errors.
Ex: https://shields-staging-pr-7687.herokuapp.com/ossf-scorecard/github.com/invalid/invalid

services/ossf-scorecard/ossf-scorecard.service.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  async fetch({ host, orgName, repoName }) {
+    return this._requestJson({
+      schema,
+      url: `https://api.securityscorecards.dev/projects/${host}/${orgName}/${repoName}`,
+      errorMessages: {
+        404: 'invalid repo path',
+        500: 'upstream service error',
```
calebcartwright, line 46: Our request handling infrastructure actually covers these cases already. We obviously allow for individual classes to apply their own message overriding for specific response codes, but it's not necessary to do so. In this case since we're not doing anything particularly unique/adding additional context I'd recommend that we remove all of these overrides, but at a minimum do so for the 500 (in order to keep it consistent and make things a tad easier/more obvious for us from an operations/support perspective)

services/ossf-scorecard/ossf-scorecard.service.js
```diff
@@ -0,0 +1,59 @@
...
+      },
+    })
+  }
+
+  async handle({ host, orgName, repoName }) {
+    console.log('fetching...')
```
calebcartwright, line 52: Please remove all of these

services/ossf-scorecard/ossf-scorecard.service.js
```diff
@@ -0,0 +1,59 @@
...
+export default class OSSFScorecard extends BaseJsonService {
+  static category = 'analysis'
+
+  static route = { base: 'ossf-scorecard', pattern: ':host/:orgName/:repoName' }
+
+  static examples = [
+    {
+      title: 'OSSF-Scorecard Score',
+      namedParams: {
+        host: 'github.com',
```
calebcartwright, line 23: Is your intent to support integrating with server/self-hosted instances of the product(s) you'll support? I ask because we've a very large number of badges that are based on the big source code platforms, but we typically don't require our users to include the TLD for those platforms.

If the OSSF service does/will support any arbitrary git host then it obviously becomes necessary to incorporate more of the uri, although if that's the case then we'd need to rethink the route a bit to account for scenarios where people have opted to host their self-hosted instance at some non-root level path, e.g. my-awesome-github-enterprise-server.com/some-path/my-org/my-repo

services/ossf-scorecard/ossf-scorecard.service.js
```diff
@@ -0,0 +1,59 @@
+import Joi from 'joi'
+import { BaseJsonService } from '../index.js'
+import { colorScale } from '../color-formatters.js'
+
+const schema = Joi.object({
+  score: Joi.required(),
```
calebcartwright, line 6: This is a numeric, >=0 value field in the raw response correct? If so let's reflect that in that in the schema

services/ossf-scorecard/ossf-scorecard.tester.js
```diff
@@ -0,0 +1,17 @@
...
+    label: 'score',
+    message: Joi.number().min(0),
+    color: Joi.string().allow(
+      'red',
+      'yellow',
+      'yellowgreen',
+      'green',
+      'brightgreen'
+    ),
+  })
```
calebcartwright, line 17: Please add one more case that hits the 404 path, presumably by specifying some non-existent repo

services/ossf-scorecard/ossf-scorecard.service.js
rohankh532: I removed the 500 and kept the 404

services/ossf-scorecard/ossf-scorecard.tester.js
rohankh532: Added a test for this

services/ossf-scorecard/ossf-scorecard.service.js
azeemshaikh38: As of now, the primary audience for Scorecard badges is public repositories hosted on GitHub. Although the Scorecard team has plans of supporting non-GitHub VCS and non-public (say enterprise servers), we aren't there yet. So I don't want to overcomplicate the design here by trying to incorporate a generic uri. What if we update the uri to: `/github/:host/:repo`? If we plan to extend this in the future for other platforms, we can consider a different badge service. Would that work? @calebcartwright

calebcartwright: > If we plan to extend this in the future for other platforms, we can consider a different badge service. Would that work? @calebcartwright

No it would not. We don't want to be in the business of having duplicative routes/codes/etc. within Shields based on what a given provider may or may not do.

Let's just leave this as-is for now, and if/when you all make relevant changes upstream we can adjust the parameter in the route to support any optional extra paths

calebcartwright (approved):
Thanks for the dialog and collaboration! This looks to be in good order so going to merge :rocket:

azeemshaikh38:
> Thanks for the dialog and collaboration! This looks to be in good order so going to merge 🚀

Thanks a lot for all your help here @calebcartwright!

rohankh532:
> Thanks for the dialog and collaboration! This looks to be in good order so going to merge 🚀

That's great--thank you so much for your help!

PR #7690, opened 2022-03-06
Clarify GitHub All Contributors badge

PyvesB:
I was wondering what the difference was between these two badges and had to go digging through the code to figure things out:
Screenshot 2022-03-06 121939

Changing the title slightly should help clarify this.

### DIFF
--- a/services/github/github-all-contributors.service.js
+++ b/services/github/github-all-contributors.service.js
@@ -17,7 +17,7 @@ export default class GithubAllContributorsService extends ConditionalGithubAuthV
 
   static examples = [
     {
-      title: 'Github All Contributors',
+      title: 'GitHub contributors (via allcontributors.org)',
       namedParams: {
         repo: 'all-contributors',
         user: 'all-contributors',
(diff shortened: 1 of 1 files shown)


calebcartwright (approved):
This one has tripped me up on multiple occasions too so agreed a change is needed. I wonder if it might also be helpful to extend the help text to provide more context about All Contributors too

PyvesB:
> extend the help text to provide more context about All Contributors too

Hmm, I'm not sure. What specifically did you have in mind? I see the UI documentation as a place where we surface caveats or what steps need to be taken to get the badge working. Adding context about individual services feels out of scope, the allcontributors.org will probably do a better job than any description we come up with. In that sense, I'd rather fully delegate to it by just surfacing the hostname in the title.

calebcartwright:
To be clear, I'm not suggesting help text as a mutually exclusive alternative to the example title. 

I think an extra one liner in the modal window help text that includes a hyperlink to the service and mentions that All Contributors defines/counts contributors very differently than GitHub is useful context to a casual observer. Yes, folks can also go and read the details (which is where i think a navigable hyperlink is helpful), but IMO it's also useful to give users a quick inline blurb within the context of Shields to convey they are actually different metrics (which is why they are different badges). From my perspective that hits the definition of a caveat given how identical the two look/sound at first blush. 

My prior approval was intended as an implicit :+1: on this as-is; consider the idea about help text as a "yes, and" notion that's not a blocking point for me whatsoever.

PyvesB:
I've given it a go, let me know what you think.

calebcartwright (approved):
lgtm!

PR #7691, opened 2022-03-06
Fix GitHub Deployments example color

PyvesB:
The example color currently appears as grey on the homepage:
Screenshot 2022-03-06 124250

`render` expects the state to be passed in as uppercase.


calebcartwright (approved):
good catch

Issue #7714, opened 2022-03-11
add a 'publisher' badge for Dart packages


devoncarew:
:clipboard: **Description**

This request is for a badge to display the publisher information for Dart packages. There already exists a 'last published version' badge for Dart packages - implementation here: https://github.com/badges/shields/blob/master/services/pub/pub.service.js.

The badge might look something like `[publisher | my.publisher.io]`; or, thuswise: ![publisher my.publisher.io](https://img.shields.io/badge/publisher-my.publisher.io-blue).

:link: **Data**

There is an API for this data - it's available as a simple http get and returns JSON. It's `https://pub.dev/api/packages/<package-name>/publisher`. So, for package path, it's https://pub.dev/api/packages/path/publisher, and for `usage`, it's https://pub.dev/api/packages/usage/publisher.

The returned json looks like `{"publisherId":"tools.dart.dev"}` - you just want to extract the `publisherId` out of the map.

:microphone: **Motivation**

This is one of the more important pieces of data related to Dart packages (the last published version being the other one). The publisher info is useful for anyone looking at a readme for a package - it can given the potential consumer of that package a rough indication of the quality of the package and potentially an expected SLO. 

The use case for this badge is to add it to the top of a github readme file - along with the github actions status and publisher version badges. For high quality packages, the publisher badge can act as an indication of the expected SLO.

ℹ️ **also**

I see the contributing info here: https://github.com/badges/shields/blob/master/doc/TUTORIAL.md. I'm willing to take a stab at an implementation here, but don't have a lot of background in JS development.

calebcartwright:
Thank you for sharing your suggestion.

Is the publisher id of a published package a mutable value?

devoncarew:
> Is the publisher id of a published package a mutable value?

Yes - it is possible for it to change. Typically that doesn't happen though - generally a package won't change publishers, and for those that do, they'd likely only do that once or twice over the package's lifetime.

Is your question related to a caching lifetime?

calebcartwright:
It's more about trying to determine whether we're dealing with a static or dynamic data point.

We do provide the static badge builder (which you leveraged for the example in the description) that's best suited for static content

https://img.shields.io/badge/publisher-my.publisher.io-blue

[...]

PR #7715, opened 2022-03-11
Add a [pub] publisher badge

devoncarew:
This PR adds a new badge for the pub package publisher information. More details on the proposed badge is available at https://github.com/badges/shields/issues/7714. Also, there's an existing pub version badge at https://github.com/badges/shields/tree/master/services/pub. Thanks!

- add a new badge for pub package publisher
- fix https://github.com/badges/shields/issues/7714


services/pub/pub-publisher.service.js
```diff
@@ -0,0 +1,49 @@
...
+
+  static route = {
+    base: 'pub/publisher',
+    pattern: ':packageName',
+  }
+
+  static examples = [
+    {
+      title: 'Pub Publisher',
+      namedParams: { packageName: 'example' },
```
chris48s, line 19: Normally we use a real package for the examples on the home page. Can we replace this with a real package that has a verified publisher

chris48s (commented):
One minor thing. Aside from that, looks good

services/pub/pub-publisher.service.js
devoncarew: Gotcha - I'll update this to `dart.dev`.

devoncarew:
Thanks for the review - updated.

services/pub/pub-publisher.service.js
```diff
@@ -0,0 +1,49 @@
...
+    const publisher = data.publisherId
+    return {
+      label: 'publisher',
+      message: publisher == null ? 'unverified' : publisher,
+      color: publisher == null ? 'red' : 'blue',
+    }
+  }
+}
+
+export { PubPublisher }
```
calebcartwright, line 49: Just a heads up since I imagine this was mirrored off the module for the other pub badge (and admittedly a nit), but we typically do default class exports unless we have something like a redirector defined in the same module and need multiple exports

services/pub/pub-publisher.tester.js
```diff
@@ -0,0 +1,26 @@
+import { ServiceTester } from '../tester.js'
+export const t = new ServiceTester({
+  id: 'PubPublisher',
+  title: 'Pub Publisher',
+  pathPrefix: '/pub',
+})
```
calebcartwright, line 6: Similarly as the class export, we prefer and encourage the model documented in our [tutorial on service tests](https://github.com/badges/shields/blob/master/doc/service-tests.md#1-boilerplate) and only fall back to this more verbose model when strictly necessary, such as when the corresponding *.service.js file has multiple exports

```suggestion
import { createServiceTester } from '../tester.js'
export const t = await createServiceTester()
```

services/pub/pub-publisher.service.js
```diff
@@ -0,0 +1,49 @@
...
+    pattern: ':packageName',
+  }
+
+  static examples = [
+    {
+      title: 'Pub Publisher',
+      namedParams: { packageName: 'dart.dev' },
+      staticPreview: {
+        label: 'publisher',
+        message: 'dart.dev',
```
calebcartwright, line 22: Is this a valid target? :thinking: 
https://shields-staging-pr-7715.herokuapp.com/pub/publisher/dart.dev


image

I see it's being used for the tests and passing but something still looks off

calebcartwright (commented):
Few thoughts from me as well though agree with the general sentiment about this looking good :+1:

services/pub/pub-publisher.service.js
devoncarew: Hmm, so I don't need this explicit export line?

services/pub/pub-publisher.tester.js
devoncarew: Gotcha, thanks - will update this.

devoncarew (commented):
Thanks for the review! Comments in-line; I'll make code changes and ping this PR again once they're uploaded.

services/pub/pub-publisher.service.js
chris48s: Its the `packageName` that is off. dart.dev is not a package: https://pub.dev/api/packages/dart.dev
How about we use `path` as the example package - that's the example you're using in the tests

services/pub/pub-publisher.service.js
devoncarew: Moved the `export` to be in front of the PubPublisher declaration.

services/pub/pub-publisher.tester.js
devoncarew: Hmm - when I make this change, my tests fail - `404`'ing against the server - and it doesn't look like the service itself is registered.

devoncarew (commented):
The PR's now updated based on the review comments.

Note that I wasn't able to make one requested change - the feedback around test registration.

Here is the current look of the badges:

Screen Shot 2022-03-16 at 12 43 51 PM

Screen Shot 2022-03-16 at 12 44 38 PM

Screen Shot 2022-03-16 at 12 43 43 PM

services/pub/pub-publisher.tester.js
calebcartwright: Sorry, I forgot to mention that as part of this you can and will need to remove the `/publisher` path in your tests too since the test registration helper handles the full path automatically

services/pub/pub-publisher.tester.js
```diff
@@ -0,0 +1,26 @@
...
+
+t.create('package publisher').get('/publisher/path.json').expectBadge({
```
calebcartwright, line 8: This accompanies the import/registration change

```suggestion
t.create('package publisher').get('/path.json').expectBadge({
```

services/pub/pub-publisher.tester.js
```diff
@@ -0,0 +1,26 @@
...
+  label: 'publisher',
+  message: 'dart.dev',
+})
+
+t.create('package not verified publisher')
+  .get('/publisher/utf.json')
```
calebcartwright, line 14:
```suggestion
  .get('/utf.json')
```

services/pub/pub-publisher.tester.js
```diff
@@ -0,0 +1,26 @@
...
+  .expectBadge({
+    label: 'publisher',
+    message: 'unverified',
+    color: 'lightgrey',
+  })
+
+t.create('package not found')
+  .get('/publisher/does-not-exist.json')
```
calebcartwright, line 22:
```suggestion
  .get('/does-not-exist.json')
```

services/pub/pub-publisher.service.js
```diff
@@ -0,0 +1,49 @@
...
+        color: 'blue',
+      },
```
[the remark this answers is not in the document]
calebcartwright: This looks like the last remaining piece. Please update the example's `staticPreview` key to invoke the `render` function instead of keeping it hardcoded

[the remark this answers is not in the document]
devoncarew: Ah - I missed this; it's now been updated to call `render()`.

calebcartwright (approved):
Thank you!

PR #7718, opened 2022-03-12
Delete [TeamCityCoverage] and [BowerVersion] redirectors

PyvesB:
These two redirectors date back from 2019 and have not been used in the past 6 months: https://metrics.shields.io/d/aESRBSjmz/services?orgId=1&from=now-6M&to=now&viewPanel=19. If they're not useful to anyone, let's take this opportunity to get rid of the corresponding code.


calebcartwright (approved):
I'm good with dropping these given the utter lack of use.

Perhaps for the future it might be worth discussing/selecting a standard time period (be it 1 month or 1 year) and documenting it somewhere

PyvesB:
> Perhaps for the future it might be worth discussing/selecting a standard time period (be it 1 month or 1 year) and documenting it somewhere

For deprecated badges, our one year policy makes sense as users are explicitly warned, and the added value of the deprecated badge is quite slim compared to falling back to the not found badge.

For redirectors, there's a backwards compatibility aspect, and we risk impacting a lot of users (unless there aren't any users, as is the case here). Additionally, we're not explicitly telling users they're using a legacy path, they could legitimately blame us for not warning them, and breaking their badges. If we wanted to enforce a time-based policy, I'd suggest we at least write up a document listing all legacy paths, link and surface it in the Readme, and communicate widely ahead of any user-impacting removal. Does that sound about right?

calebcartwright:
Yes I'm familiar with the differences between the two and the respective implications of their removal. I'm specifically talking about redirectors here given the context and what I believe is a novel case of dropping a redirector. 

It's certainly plausible that at some point in the future we could have redirectors that have gone unused for some X amount of time, and it's that X period I am speaking to.

We could either keep this entirely ad-hoc and subjective, wherein if someone (technically two people given the need for approval) decides that it has been "long enough" then they can drop a redirector, or, we as a core team could try to quantify this a bit to define a value for X that says "if a redirector has gone unused for <<insert time period agreed upon>> then the core team may remove it".

I don't have an opinion yet personally, but wanted to pose the question

PR #7723, opened 2022-03-13

PyvesB:
We've not updated this number for a year. According to the latest Cloudfare statistics:
Screenshot 2022-03-12 192205

We've effectively grown by ~13% since #6278. Hopefully we'll reach a billion by this time next year. 🚀


chris48s (approved):
> Hopefully we'll reach a billion by this time next year

I think my hope is more centred around finding a funding/hosting model that can sustain that level of growth on an ongoing basis

PR #7781, opened 2022-03-26
allow `requireCloudflare` setting to work when hosted on fly.io

chris48s:
y'know - just in case anyone wants to do that


paulmelnikow (approved):
:+1:

PR #7782, opened 2022-03-27
two fixes for [packagist] schemas

### LINKED ISSUE #7779: Packagist invalid response data for package version, but only on one repo

shields.io

### 🐞 Description

I have two packages on packagist that I am using version shields for. I first thought this was some caching issue, but it's been a few days and it's still the same.

On one of the packages it works perfectly, but the second badge gives the invalid response data error.

image
image

Any ideas on what's wrong and how to fix?

### 🔗 Link to the badge

https://img.shields.io/packagist/v/hyde/framework

### 💡 Possible Solution
Edit: Managed to get it to work by adding the flag `?include_prereleases`. Not sure why it works well on the other one without including prereleases.

chris48s:
closes #7779

### DIFF
--- a/services/packagist/packagist-base.js
+++ b/services/packagist/packagist-base.js
@@ -5,9 +5,12 @@ import { isStable, latest } from '../php-version.js'
 const packageSchema = Joi.array().items(
   Joi.object({
     version: Joi.string().required(),
-    require: Joi.object({
-      php: Joi.string(),
-    }),
+    require: Joi.alternatives(
+      Joi.object({
+        php: Joi.string(),
+      }).required(),
+      Joi.string().valid('__unset')
+    ),
   })
 )
 
--- a/services/packagist/packagist-version.service.js
+++ b/services/packagist/packagist-version.service.js
@@ -3,7 +3,6 @@ import { renderVersionBadge } from '../version.js'
 import { optionalUrl } from '../validators.js'
 import { redirector } from '../index.js'
 import {
-  allVersionsSchema,
   keywords,
   BasePackagistService,
   customServerDocumentationFragment,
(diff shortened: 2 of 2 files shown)


services/packagist/packagist-base.js
```diff
@@ -5,9 +5,12 @@ import { isStable, latest } from '../php-version.js'
...
   Joi.object({
     version: Joi.string().required(),
-    require: Joi.object({
-      php: Joi.string(),
-    }),
+    require: Joi.alternatives(
+      Joi.object({
+        php: Joi.string(),
+      }).required(),
+      Joi.string().valid('__unset')
```
calebcartwright, line 12: I'm not terribly familiar with the PHP ecosystem, but are there any other well known values we could potentially see here or is `__unset` essentially it based on what we know today?

calebcartwright: I gather adding a live test would probably be overkill, but would it be worth leaving an inline comment with a link to the issue?

calebcartwright (approved):
lgtm, no objections to merging as is so have approved though left a few inline questions too

services/packagist/packagist-base.js
chris48s: Right - having looked at it, we are running the validation on the uncompressed response which is why the `__unset` values are still in there at the point we validate. I misremembered that we had done something like I suggested in https://github.com/badges/shields/pull/6508#issuecomment-855295658 but we never did.
I think I'd like to revisit this code as I don't think what we are doing now is the best solution but I don't think I want to shave this specific yak tonight :cow2: . I'm going to merge this and punt that to an issue.

chris48s: I've written it up in #7783

PR #7816, opened 2022-04-07
Try out dependency-review-action

chris48s:
There have been a number of high-profile supply chain attacks or issues in the NPM ecosystem, either due to packages being compromised by an attacker or sabotaged by their maintainer e.g:

- [node-ipc](https://security.snyk.io/vuln/SNYK-JS-NODEIPC-2426370)
- [faker/colors](https://github.com/advisories/GHSA-gh88-3pxp-6fm8)
- [coa](https://github.com/advisories/GHSA-73qr-pfmq-6rp8)
- [rc](https://github.com/advisories/GHSA-g2q5-5433-rhrf)
- [ua-parser-js](https://github.com/advisories/GHSA-pjwm-rvh2-c87w)
- [eslint-config-eslint/eslint-scope](https://eslint.org/blog/2018/07/postmortem-for-malicious-package-publishes)
- [event-stream](https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream/)
- etc

Shields has mercifully not been directly affected by any of these, but this is more through luck than judgement. Like most javascript projects with a large dependency tree we simply don't have enough hours in the day to fully review every dependency bump. One potentially exciting project in this field is Socket.dev, but its early days for that.

In a similar nitch, but attacking the problem from a different angle, Github have recently released this more fully formed action aimed at flagging dependency bumps with known issues.

Further reading:
- https://github.blog/2022-04-06-prevent-introduction-known-vulnerabilities-into-your-code/
- https://github.com/actions/dependency-review-action
- https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review

I have not played with this at all yet, but I'm going to suggest we enable it ahead of this weeks barrage of dependency bumps (but lets not make it a requirement for merging just yet) and try it out.

### DIFF
--- a/.github/workflows/enforce-dependency-review.yml
+++ b/.github/workflows/enforce-dependency-review.yml
@@ -1,0 +1,14 @@
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@v1


calebcartwright:
I certainly don't think it hurts, and am happy to pull it in. I do, however, question the efficacy of GitHub's solution in the presence of regular dependency updates. I suppose it could still help catch some issue stemming from older transitive deps?

I.e. I assume that 99% of the time during dependency updates that any issues haven't actually been discovered and reported yet

calebcartwright (approved):

calebcartwright:
Going to go ahead and merge as well, and then let Dependabot rebase this weeks bumps to pull it in

chris48s:
Yeah I kind of agree with you. I think there's more mileage in the socket approach long-term which is aiming to be proactive rather than reactive. The one thing we have going for us is we update weekly rather than daily, so if something like this does happen we may be just far enough behind the curve to benefit from this in some cases.

I feel like the sweet spot for package updates is to be updating fast enough that you're ahead of the attackers trying to exploit vulnerabilities but slow enough that if there's a supply chain attack, someone else notices it before you do :laughing:

PR #7862, opened 2022-04-17
add labels to Dockerfile

### LINKED ISSUE #7838: write git commit hash to docker meta-data
:clipboard: **Description**

We should be able to do this using a docker object label
https://docs.docker.com/config/labels-custom-metadata/

chris48s:
Closes #7838

I've added 2 labels here.
- One is `fly.version`. I _think_ this is what `flyctl` shows when you run `flyctl image show` :crossed_fingers: but this is mainly based on looking at their own docker images e.g: https://github.com/fly-apps/postgres-standalone/blob/ddbaaacf1d946dcd68ab0d532995906d13e3df6f/Dockerfile#L20 rather than documentation - we'll have to see
- The other is `version`, which just seems like something sensible to set for wider consumers.


calebcartwright (approved):
:+1:

PR #7877, opened 2022-04-22
Migrate to Express [*****]

### LINKED ISSUE #3328: Modernizing the server

### LINKED ISSUE #3368: Modernizing the request code

### LINKED ISSUE #4948: Move JSON output out of badge-maker and into server

paulmelnikow:
There's surely a bit of cleanup to do here, but thought it worth getting some 👀 on this.

I'd welcome help updating and improving the documentation, though that could also be done in a follow-on PR.

An area I've not yet addressed is query parameters. Scoutcamp used to turn the params into numbers, which I imagine Express does not do. It means we probably can simplify places where we've got `Joi.alternatives().try(Joi.string(), Joi.number())` in the query param schemas.

There are surely bugs in this, and edge cases. I removed a few tests in `legacy-request-handler.spec.js` which we may want to rewrite elsewhere.

We should test this out (maybe even on Fly), in particular things like suggest and token acceptance.

I would be surprised if performance were a problem, though we may want to experiment with this under load to make sure.

Closes #3328 
Closes #3368
Closes #4948


badge-maker/lib/make-badge.spec.js
```diff
@@ -1,143 +1,48 @@
...
-        label: 'cactus',
-        value: 'grown',
-        message: 'grown',
-        link: ['https://example.com/', 'https://other.example.com/'],
-      })
+      expectBadgeToMatchSnapshot({ label: 'cactus', message: 'grown' })
     })
 
     it('should replace undefined svg badge style with "flat"', function () {
-      const jsonBadgeWithUnknownStyle = makeBadge({
```
paulmelnikow, line 117: Inline

badge-maker/lib/make-badge.spec.js
```diff
@@ -1,143 +1,48 @@
...
-      expectBadgeToMatchSnapshot({
-        label: 'cactus',
-        message: 'grown',
-        format: 'svg',
-      })
-    })
-  })
-
-  describe('JSON', function () {
-    it('should produce the expected JSON', function () {
```
paulmelnikow, line 100: Remove JSON tests

badge-maker/lib/make-badge.spec.js
```diff
@@ -557,19 +427,17 @@ describe('The badge generator', function () {
         makeBadge({
           label: '',
           message: 'some-value',
-          format: 'json',
           style: 'social',
         })
       )
-        .to.include('""')
+        .to.include('></text>')
```
paulmelnikow, line 433: SVG equivalent

core/base-service/base-static.js
```diff
@@ -1,58 +1,29 @@
...
+      (req, res, next) => {
+        if (serverHasBeenUpSinceResourceCached(req)) {
+          // Send Not Modified.
+          res.status(304)
+          res.end()
+        } else {
+          next()
+        }
+      },
+      this.makeExpressHandler(serviceContext, serviceConfig)
```
paulmelnikow, line 26: Handle the 304's using a middleware, then delegate to the base service request handler, overriding the cache headers via `_applyCacheHeaders()`. Way DRYer and I think a big improvement, though I could imagine making this a little cleaner in the future.

core/base-service/legacy-request-handler.spec.js
```diff
@@ -1,251 +0,0 @@
...
-          handleRequest(
-            cacheHeaderConfig,
-            (queryParams, match, sendBadge, request) => {
-              fakeHandler(queryParams, match, sendBadge, request)
-            }
-          )
-        )
-      }
-
-      it('should set the expires header to current time + defaultCacheLengthSeconds', async function () {
```
paulmelnikow, line 127: As I mentioned in the top comment it may make sense to comb through these behaviors and see if we should add test coverage elsewhere.

core/base-service/base.spec.js
```diff
@@ -321,63 +327,24 @@ describe('BaseService', function () {
...
-      }
-      mockHandleRequest = sinon.spy()
-      DummyService.register(
-        { camp: mockCamp, handleRequest: mockHandleRequest },
-        defaultConfig
-      )
+  describe('Express integration', function () {
+    let harness
+    beforeEach(async function () {
+      harness = new ExpressTestHarness()
```
paulmelnikow, line 333: Have added this helper class to reduce boilerplate in setting up, starting, and issuing requests to, an empty express app.

core/base-service/route.js
```diff
@@ -44,23 +44,29 @@ function prepareRoute({ base, pattern, format, capture, withPng }) {
   return { regex, captureNames }
 }
 
-function namedParamsForMatch(captureNames = [], match, ServiceClass) {
-  // Assume the last match is the format, and drop match[0], which is the
-  // entire match.
-  const captures = match.slice(1, -1)
-
-  if (captureNames.length !== captures.length) {
+function paramsForReq(captureNames = [], req, ServiceClass) {
```
paulmelnikow, line 47: - Rewrite for express, where `req.params` is an object like `{1: 'foo', 2: 'bar'}`.
- Unpack `namedParams` and `format` in one place, which feels like better encapsulation (and is DRYer too).

core/base-service/route.spec.js
```diff
@@ -83,13 +112,13 @@ describe('Route helpers', function () {
...
     })
 
-    expect(() =>
-      namedParamsForMatch(captureNames, regex.exec('/foo/bar/baz.svg'), {
-        name: 'MyService',
-      })
-    ).to.throw(
-      'Service MyService declares incorrect number of named params (expected 2, got 1)'
-    )
+    it('Throws the expected error', function () {
```
paulmelnikow, line 115: This test was triggering inside of setup; needed to be wrapped in an `it`.

core/server/server.js
```diff
@@ -306,87 +297,78 @@ class Server {
 
   // See https://www.viget.com/articles/heroku-cloudflare-the-right-way/
   requireCloudflare() {
-    // Set `req.ip`, which is expected by `cloudflareMiddleware()`. This is set
-    // by Express but not Scoutcamp.
```
paulmelnikow, line 310: So, `cloudflareMiddleware()` is supposed to handle this out of the box, though as I re-read this, I imagine that only supports `x-forwarded-for`, and that we may need to restore the `isOnFly()` part of this behavior.

core/server/server.js
```diff
@@ -494,65 +487,65 @@ class Server {
...
+      })
+    )
     await this.registerServices()
+    if (registerExtras) {
+      registerExtras(app)
+    }
+    this.registerNotFoundHandlers()
 
-    camp.timeout = this.config.public.requestTimeoutSeconds * 1000
-    if (this.config.public.requestTimeoutSeconds > 0) {
```
paulmelnikow, line 536: Replace with Express' built-in timeout support. It doesn't return a 408; I think it just hangs up the socket.

core/server/server.spec.js
```diff
@@ -98,7 +96,7 @@ describe('The server', function () {
         `${baseUrl}:fruit-apple-green.svg`
       )
       expect(statusCode).to.equal(200)
-      expect(headers['content-type']).to.equal('image/svg+xml;charset=utf-8')
+      expect(headers['content-type']).to.equal('image/svg+xml; charset=utf-8')
```
paulmelnikow, line 99: This space here seems to be placed by Express.

core/server/server.spec.js
```diff
@@ -112,7 +110,9 @@ describe('The server', function () {
         `${baseUrl}:fruit-apple-green.json`
       )
       expect(statusCode).to.equal(200)
-      expect(headers['content-type']).to.equal('application/json')
+      expect(headers['content-type']).to.equal(
+        'application/json; charset=utf-8'
```
paulmelnikow, line 114: Charset added by Express

core/server/server.spec.js
```diff
@@ -200,19 +200,12 @@ describe('The server', function () {
...
-      // TODO It would be nice if this were 404 or 410.
-      expect(statusCode).to.equal(200)
+      expect(statusCode).to.equal(410)
       expect(body)
         .to.satisfy(isSvg)
         .and.to.include('410')
         .and.to.include('jpg no longer available')
     })
-
-    it('should return cors header for the request', async function () {
```
paulmelnikow, line 211: This duplicates a test a bit higher up in this context

doc/code-walkthrough.md
```diff
@@ -80,29 +80,22 @@ test this kind of logic through unit tests (e.g. of `render()` and
...
 2.  The Server, which is defined in
-    [`core/server/server.js`][core/server/server], is based on the web
-    framework [Scoutcamp][]. It creates an http server, sets up helpers for
-    token persistence and monitoring. Then it loads all the services,
-    injecting dependencies as it asks each one to register its route
-    with Scoutcamp.
+    [`core/server/server.js`][core/server/server], is based on [Express][].
+    It creates an http server, sets up helpers for token persistence and
+    monitoring. Then it loads all the services, injecting dependencies as it
+    asks each one to register its route with the Express app.
```
paulmelnikow, line 86: Made a rough effort to update this, though it may make sense for someone to take a pass through later. This whole process is much more direct than before.

package.json
```diff
@@ -25,7 +25,6 @@
     "@fontsource/lekton": "^4.5.6",
     "@renovate/pep440": "^1.0.0",
     "@sentry/node": "^6.19.6",
-    "@shields_io/camp": "^18.1.1",
```
paulmelnikow, line 28: 💁🏼 🎩

services/github/auth/acceptor.js
```diff
@@ -15,56 +16,64 @@ function setRoutes({ server, authHelper, onTokenAccepted }) {
...
+    res.end()
   })
 
-  server.route(/^\/github-auth\/done$/, async (data, match, end, ask) => {
-    if (!data.code) {
-      log.log(`GitHub OAuth data: ${JSON.stringify(data)}`)
-      return end('GitHub OAuth authentication failed to provide a code.')
-    }
+  app.post('/github-auth/done', multer().none(), async (req, res) => {
+    const code = (req.body ?? {}).code
```
paulmelnikow, line 27: This hasn't been tested yet

paulmelnikow:
In the [first service test run](https://app.circleci.com/pipelines/github/badges/shields/12550/workflows/5a382250-bf1d-4fd5-8ea4-bce023a0bfa1/jobs/180536) there are some interesting failures.

One issue is that the JSON badge no longer strips leading and trailing whitespace from `label` and `message`. That could be added to `makeJsonBadge`… or maybe we should just fix the services.

core/base-service/route.spec.js
```diff
@@ -46,33 +67,41 @@ describe('Route helpers', function () {
...
+        format: 'svg',
+      })
     })
   })
 
   context('No named params are declared', function () {
     const { regex, captureNames } = prepareRoute({
       base: 'foo',
-      format: '(?:[^/]+)',
+      format: '(?:[^/]+?)',
```
paulmelnikow, line 97: To correctly detect `.svg` and `.json` as intended, this regex needs to be lazy.

services/suggest.js
```diff
@@ -146,54 +146,59 @@ async function findSuggestions(githubApiProvider, url) {
 //        - link: target as a string URL
 //    - preview: object (optional)
 //      - style: string
-function setRoutes(allowedOrigin, githubApiProvider, server) {
-  server.ajax.on('suggest/v1', (data, end, ask) => {
+function setRoutes(allowedOrigin, githubApiProvider, app) {
+  app.get('/[$]suggest/v1', (req, res) => {
```
paulmelnikow, line 150: The Scoutcamp AJAX routes implicitly started with $. For some reason the dollar needs to be escaped for Express.

chris48s:
Thanks for taking this on. This aspect of the codebase and the reliance on ScoutCamp is one of the biggest pain points of this codebase so it is really good to be making some progress on this :)

I haven't done a full review of this, but I have checked it out locally and skimmed the diff.

The main issue I hit is if I run `npm start` on a checkout of this branch, it throws 

```
Failed to load plugin 'graphql' declared in 'BaseConfig': Cannot find module 'ws'
```

and won't start the frontend, so I've only been able to test the badge server. Any idea what is happening there?

---

There were a few bits I wanted to specifically check as possible edge cases:

- Badges with regex routes e.g:
    - `services/travis/travis-build.service.js`
    - `services/wercker/wercker.service.js`
    - `services/requires/requires.service.js`
    - `services/static-badge/static-badge.service.js`
    Those all seem to work fine 👍

- .png redirect to raster proxy. Also works 👍

- `/endpoint` route conflict - not tested yet as I couldn't start the frontend.

---

I'll try and have a more detailed look at this next week.

core/server/server.js
```diff
@@ -494,65 +488,65 @@ class Server {
...
     this.registerRedirects()
+    app.use(
+      express.static(this.config.public.documentRoot, {
+        // Since express's `maxAge` parameter sets `Cache-Control: public`, set
+        // the headers manually insetad.
+        cacheControl: false,
+        setHeaders: res =>
+          res.setHeader('Cache-Control', 'max-age=300, s-maxage=300'),
+      })
+    )
```
chris48s, line 527: Is this replacing `staticMaxAge`?
https://github.com/badges/sc/commit/dc89a3a1861b1ff3e5d6ce44ae51aca375128e63

core/base-service/legacy-request-handler.js
```diff
@@ -1,143 +0,0 @@
...
-      const badgeData = coalesceBadge(
-        filteredQueryParams,
-        { label: 'vendor', message: 'unresponsive' },
-        {}
-      )
-      const svg = makeBadge(badgeData)
-      const extension = (match.slice(-1)[0] || '.svg').replace(/^\./, '')
-      setCacheHeadersOnResponse(ask.res)
-      makeSend(extension, ask.res, end)(svg)
-    }, 25000)
```
chris48s, line 115: I guess we've just totally dropped this in favour of express `this.server.setTimeout()` ?

badge-maker/lib/make-badge.js
```diff
@@ -24,22 +23,6 @@ module.exports = function makeBadge({
   label = `${label}`.trim()
   message = `${message}`.trim()
 
-  // This ought to be the responsibility of the server, not `makeBadge`.
-  if (format === 'json') {
```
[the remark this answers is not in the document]
chris48s: I haven't read over the code for this bit at all yet. I agree with the objective. Has anything significant has changed since https://github.com/badges/shields/pull/4980 in terms of the implementation, specifically https://github.com/badges/shields/pull/4980#issuecomment-623119680 ? I will have a look through it. If a slightly different configuration of not eating our own dogfood is necessary to get off scoutcamp, this shouldn't be a blocker but until we do #4947 and #4949 I suspect all we can really do is move from one undesirable implementation to another. I do accept #4947 and #4949 have been open for 2 years and none of us are actively working on them.
As a note, we may want to consider documenting it as a non-BC change for badge-maker, even though `format='json'` is undocumented. The next badge-maker will be a major release anyway due to dropping node 10.

core/base-service/redirector.js
```diff
@@ -65,30 +66,30 @@ export default function redirector(attrs) {
...
         )
         trace.logTrace('inbound', emojic.ticket, 'Named params', namedParams)
-        trace.logTrace('inbound', emojic.crayon, 'Query params', queryParams)
+        trace.logTrace('inbound', emojic.crayon, 'Query params', req.query)
 
         const targetPath = encodeURI(transformPath(namedParams))
         trace.logTrace('validate', emojic.dart, 'Target', targetPath)
 
-        let urlSuffix = ask.uri.search || ''
+        let urlSuffix = url.parse(req.url).search ?? '' // eslint-disable-line node/no-deprecated-api
```
chris48s, line 92: Can we do this with `let urlSuffix = new url.URL(req.url).search` instead of introducing a deprecated call here? I think that should be the same.

core/server/server.spec.js
```diff
@@ -200,19 +200,12 @@ describe('The server', function () {
       const { statusCode, body } = await got(`${baseUrl}npm/v/express.jpg`, {
         throwHttpErrors: false,
       })
-      // TODO It would be nice if this were 404 or 410.
-      expect(statusCode).to.equal(200)
+      expect(statusCode).to.equal(410)
```
chris48s, line 203: :+1:

core/server/server.js
```diff
@@ -494,65 +488,65 @@ class Server {
...
-      })
+    if (secure) {
+      this.server = https.createServer({ hostname, cert, key }, app)
+    } else {
+      this.server = http.createServer({ hostname }, app)
     }
-    camp.listenAsConfigured()
 
-    await new Promise(resolve => camp.on('listening', () => resolve()))
+    this.server.setTimeout(this.config.public.requestTimeoutSeconds * 1000)
```
chris48s, line 540: 1. If we're going to move to just dropping the connection on timeout, I think we can just get rid of the `requestTimeoutMaxAgeSeconds` setting as it won't do anything
2. `requestTimeoutSeconds` is an optional setting (for self hosting users). We should handle the case where it is not set.

chris48s (commented):
Did another slightly more detailed pass over this and left some more comments

paulmelnikow:
Thanks for the comments. Will nudge this forward when I can!

PyvesB:

If others think that repository level projects would be an option and want to pick this up, you can simply fetch the commits by running git fetch origin pull/7877/head:pr-7877. Change the implementation to target repository level projects and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

chris48s:
It is still my ambition to pick this up and finish it at some point, but agree that there is limited value in keeping the PR hanging around

PR #7885, opened 2022-04-25
move all public production config into yaml

chris48s:
- Refs https://github.com/badges/shields-ops/issues/16
- Refs https://github.com/badges/shields-ops/pull/22

### DIFF
--- a/config/shields-io-production.yml
+++ b/config/shields-io-production.yml
@@ -6,13 +6,20 @@ public:
       enabled: true
       url: https://metrics.shields.io/telegraf
       instanceIdFrom: env-var
-      instanceIdEnvVarName: HEROKU_DYNO_ID
+      instanceIdEnvVarName: FLY_ALLOC_ID
       envLabel: shields-production
 
   ssl:
-    isSecure: true
+    isSecure: false
 
   cors:
     allowedOrigin: ['http://shields.io', 'https://shields.io']
 
+  services:
+    gitlab:
+      authorizedOrigins: 'https://gitlab.com'
+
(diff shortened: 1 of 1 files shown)


config/shields-io-production.yml
```diff
@@ -6,13 +6,20 @@ public:
...
   ssl:
-    isSecure: true
+    isSecure: false
 
   cors:
     allowedOrigin: ['http://shields.io', 'https://shields.io']
 
+  services:
+    gitlab:
+      authorizedOrigins: 'https://gitlab.com'
```
calebcartwright, line 20: I'm trying to remember why I needed to set this as an env var in the first place when we introduced support for GitLab API based data acquisition, but it's escaping my memory :shrug:

calebcartwright (approved):
wfm :+1:

config/shields-io-production.yml
chris48s: I reckon what ended up as yaml config and what ended up as env vars was just us being a bit hit and miss over time tbh.

If you mean why we specifically needed to set this for GitLab, GitLab is the only service where we've got credentials set in production and there is no default `authorizedOrigins`. Everything else that takes an `authorizedOrigins` either has a default on `authorizedOrigins` or we allow credentials to be set for self-hosting purposes but don't do it ourselves.

calebcartwright: > If you mean why we specifically needed to set this for GitLab,

Sorry I forgot to resolve this thread last night. This was the question I had in mind but remembered that was the case

PR #7888, opened 2022-04-29
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -2,6 +2,9 @@ name: Build Docker Image
 on:
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -3,8 +3,13 @@ on:
   push:
     branches:
       - master
+permissions:
+  contents: read
+
 jobs:
   build-and-deploy:
+    permissions:
+      contents: write  # for JamesIves/github-pages-deploy-action to push changes in repo
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
--- a/.github/workflows/publish-docker-next.yml
+++ b/.github/workflows/publish-docker-next.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest


calebcartwright:
Thank you for the PR!

While I'm not explicitly opposed to the proposed changes, they also strike me as purely ceremonial with no value add. Any security mechanisms based on user-modifiable content in the repo are inherently self-defeating. The correct way to manage these settings is via repository settings (something GitHub's docs note in many places) where they can't be modified by anyone other than repository admins.

Additionally, GitHub doesn't propagate any repository secretes, including GITHUB_TOKEN, to runs triggered from a fork. That in turn really only left the supply chain vulnerability potential via Dependabot-style updates that were opened from an in-repo branch. However, GitHub also internally handles that risk already as noted in some of the related posts and documentation to the ones you included in the PR description.

In short, we're familiar with the material you've shared, and it was already reviewed/considered/addressed. I'd personally prefer to primarily rely on repo-setting management instead of establishing a practice of inserting duplicative, easily manipulated, keys into our various yaml files.

Will defer to the other maintainers though as to whether we should move forward with this anyway.

naveensrinivasan:
> Thank you for the PR!
> 
> While I'm not explicitly opposed to the proposed changes, they also strike me as purely ceremonial with no value add. Any security mechanisms based on user-modifiable content in the repo are inherently self-defeating. The correct way to manage these settings is via repository settings (something GitHub's docs note in many places) where they can't be modified by anyone other than repository admins.
> 
> Additionally, GitHub doesn't propagate any repository secretes, including GITHUB_TOKEN, to runs triggered from a fork. That in turn really only left the supply chain vulnerability potential via Dependabot-style updates that were opened from an in-repo branch. However, GitHub also internally handles that risk already as noted in some of the related posts and documentation to the ones you included in the PR description.
> 
> In short, we're familiar with the material you've shared, and it was already reviewed/considered/addressed. I'd personally prefer to primarily rely on repo-setting management instead of establishing a practice of inserting duplicative, easily manipulated, keys into our various yaml files.
> 
> Will defer to the other maintainers though as to whether we should move forward with this anyway.

Ok. I understand. Thanks

chris48s:
I'll try and have a proper look at this next week.

I think one of the things going on here is that GitHub have recently made a change to the default permissions granted to `GITHUB_TOKEN` but it only applies to mew repos. i.e: if I create a new repo, `GITHUB_TOKEN` has "Read repository contents permission" out of the gate unless you elevate it, whereas for an older repo like ours we retain the previous default, which is that `GITHUB_TOKEN` has "Read and write permissions" by default. There may be some mileage in us switching our setting to the new default and elevating where necessary.

@naveensrinivasan is this PR related to that GH policy change, or independent of it?

naveensrinivasan:
> I'll try and have a proper look at this next week.
> 
> I think one of the things going on here is that GitHub have recently made a change to the default permissions granted to `GITHUB_TOKEN` but it only applies to mew repos. i.e: if I create a new repo, `GITHUB_TOKEN` has "Read repository contents permission" out of the gate unless you elevate it, whereas for an older repo like ours we retain the previous default, which is that `GITHUB_TOKEN` has "Read and write permissions" by default. There may be some mileage in us switching our setting to the new default and elevating where necessary.
> 
> @naveensrinivasan is this PR related to that GH policy change, or independent of it?

It is independent of it.

chris48s:
merged d6df4fb in preference to this. Thanks :+1:

PR #7921, opened 2022-05-01
fix [pub] service tests

chris48s:
These not found case tests have been failing because pub has changed the input validation upstream. Calling with `does-not-exist` now returns a `400 Bad Request` with body

```json
{
  "error": {
    "code": "InvalidInput",
    "message": "\"package\" must match RegExp: pattern=^[a-zA-Z_][a-zA-Z0-9_]*$ flags="
  },
  "code": "InvalidInput",
  "message": "\"package\" must match RegExp: pattern=^[a-zA-Z_][a-zA-Z0-9_]*$ flags="
}
```

instead of a 404. Using a package name that matches the regex in the tests still gives us a 404 so I've updated the tests.

We could also consider calling `fetch()` with

```js
errorMessages: {
  400: 'not found',
},
```

or we could leave that as an `invalid`. I've left it for now. Thoughts?


G1Joshi (approved):

calebcartwright (approved):
lgtm, though i think it would be worth adding the fetch mapping for the 400 response code at some point too

chris48s:
Going to merge this for now. Thinking about it

```js
errorMessages: {
  400: 'invalid package name',
},
```

probably makes more sense, but I reckon I'll come back to this one after @G1Joshi 's PRs just to avoid blocking the reviews

PR #7927, opened 2022-05-05
update actions checkout step version

calebcartwright:
Have been meaning to submit this for a while but it kept slipping. The v2 version of this action is pinned to Node 12, which just went end of life a few weeks ago. The new v3 version uses Node 16, with the major version bump being applied out of an abundance of caution for GitHub Enterprise Server scenarios

https://github.com/actions/checkout#whats-new

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.3.1
+        uses: actions/checkout@v3.3.1
         with:
           persist-credentials: false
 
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.3.1
+        uses: actions/checkout@v3.3.1
```
chris48s, line 11:
```suggestion
        uses: actions/checkout@v3
```

I don't think `v3.3.1` is a real version :smile:  Tbh, we don't need to be pinned to a specific minor version here anyway.

chris48s (approved):

PR #7928, opened 2022-05-05
Set permissions for GitHub actions

chris48s:
Following on from @naveensrinivasan 's suggestion in #7888 I agree with minimising the possible harm from token exposure but I think the approach we should actually adopt here is:

- Change our repo settings in https://github.com/badges/shields/settings/actions so that actions get "Read repository contents permission" by default (i.e: all actions get `contents: read` unless otherwise specified). This would bring us into line with GitHub's default for new repos. I haven't done this yet.
- Explicitly elevate permissions for the workflows that need more access (which I _think_ I've done correctly in this PR)

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     types: [closed]
 
+permissions:
+  contents: write
+
 jobs:
   create-release:
     if: |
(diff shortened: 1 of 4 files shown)


naveensrinivasan:
Thanks!

calebcartwright (approved):
Agreed, and essentially what I had in mind/was attempting to articulate on the other thread :+1:

PR #7956, opened 2022-05-11
Update node builds/images

chris48s:


chris48s:
Well, you can't blame me for being optimistic. I'll have a bash at debugging this locally and circle back to this PR.

Another thing I realise having I pushed this is that

a) Now that we're off Heroku buildpacks, the `engines` statement in `package.json` no longer controls what node version we're deployed on: its our Dockerfile that does that. This means we can set `engines.node: ">=16"` and run with `NPM_CONFIG_ENGINE_STRICT` on all node versions
b) `npm ci` may now respect the strict engine/peer dependency flags

so there's probably potential to simplify some of that too. I'll revisit that too..

chris48s:
Did a bit of playing with this locally.

- Core tests all pass under node 18, as long as you install the packages under node 16 and switch to 18
- `npm ci` also fails locally on node 16
- I think the thing that is breaking in the install is lmdb, which is a transitive dependency of gatsby (and some other gatsby-related things). See https://github.com/gatsbyjs/gatsby/issues/35576 and https://github.com/gatsbyjs/gatsby/pull/35585
- Good news: there should be a release of gatsby either already released or soon to be released that is installable under node 18
- Bad news: we can't upgrade to it because of graphql peer dependency issues

so I think that means the gatsby/graphql mess is going to be a blocker to being able to upgrade to node 18. We don't have an immediately pressing need to do this. Node 18 is brand new and Node 16 will continue to get security fixes until 2024 but its worth establishing where the blocker is. I think we've probably reached the point with this where "just assume it will get sorted out upstream" is probably not going to work out and we need to look at moving away from one or other of the packages that are causing us problems here. I need to go back and do some reading to try and remember what we have already established there.

In terms of this PR, there are some other semi-related bits that I'm going to extract to another PR or two as a passing build on node 18 itself is probably not happening any time soon.

calebcartwright:
> I need to go back and do some reading to try and remember what we have already established there.

Same, but will post some links to relevant discussion for broader awareness

https://github.com/badges/shields/pull/7624
https://github.com/badges/shields/pull/7230
https://github.com/badges/shields/pull/7428
https://github.com/badges/shields/pull/7795

chris48s:
binning this in favour of https://github.com/badges/shields/pull/9291

PR #7973, opened 2022-05-14
simplify NPM strict checks

chris48s:
There's a few things going on here:

1. https://github.com/npm/cli/issues/1219 is resolved upstream, so we don't need to do the `npm install --dry-run && npm ci` workaround in order to do these checks. We can just run `npm ci`. I'm not 100% sure what the lowest version this is fixed in, but I tested it with NPM 8.0.0 and `npm ci` respects the settings so I've bumped us to NPM >= 8 everywhere.
2. Now that we're off Heroku buildpacks, the `engines` statement in `package.json` no longer controls what node version we're deployed on in production - our Dockerfile that does that. This means we can set `engines.node: ">=16"` and run with `engine-strict` on all node versions (previously we had this check disabled on the node 17 builds) without that meaning we're deployed on node 17 in prod.
3. Given we don't need to disable these settings for node 17 (either in CI, or for contributors) I've moved them to a `.npmrc` instead of env vars and I've just turned off `engine-strict` for the node 12/14 package builds with env vars. This keeps everything a bit simpler.


calebcartwright (approved):
lgtm. we should probably reflect the same in the tutorial setup docs, but can be handled in a follow up 

https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#node-npm

calebcartwright (approved):

PR #7976, opened 2022-05-16
update docs on production hosting

chris48s:
There was actually less to update here than I thought.

There are a few bits of heroku-specific code knocking about like:

- https://github.com/badges/shields/blob/master/Procfile
- https://github.com/badges/shields/blob/6dff32138b7316228dbde6feffbf23610ec14570/package.json#L106
- https://github.com/badges/shields/blob/8821ff85fcd494ebb802bf9bb26ddcba11d21ec8/core/server/server.js#L312-L317
- https://github.com/badges/shields/blob/99bffd3a86cb85f4ca20a22599e64be6abf7efcb/core/base-service/got-config.js#L20-L22

I have decided to just leave them all where they are because we have documented heroku buildpack-based (as opposed to docker-based) deployment as a self-hosting option for a long time

https://github.com/badges/shields/blob/6dff32138b7316228dbde6feffbf23610ec14570/doc/self-hosting.md?plain=1#L50-L61

and I don't see a strong reason to break backwards compatibility for the sake of removing a few lines.

### DIFF
--- a/doc/production-hosting.md
+++ b/doc/production-hosting.md
@@ -16,9 +16,8 @@ Production hosting is managed by the Shields ops team:
 
 | Component                     | Subcomponent                    | People with access                                              |
 | ----------------------------- | ------------------------------- | --------------------------------------------------------------- |
-| shields-production-us         | Account owner                   | @calebcartwright, @paulmelnikow                                 |
-| shields-production-us         | Full access                     | @calebcartwright, @chris48s, @paulmelnikow, @pyvesb             |
-| shields-production-us         | Access management               | @calebcartwright, @chris48s, @paulmelnikow, @pyvesb             |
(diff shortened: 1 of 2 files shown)


doc/production-hosting.md
```diff
@@ -95,33 +94,26 @@ The raster server `raster.shields.io` (a.k.a. the rasterizing proxy) is
...
 
-## Logs
-
-Logs can be retrieved [from heroku](https://devcenter.heroku.com/articles/logging#log-retrieval).
-
 ## Error reporting
 
 [Error reporting][sentry] is one of the most useful tools we have for monitoring
 the server. It's generously donated by [Sentry][sentry home]. We bundle
-[`raven`][raven] into the application, and the Sentry DSN is configured via
```
calebcartwright, line 120: wow, good catch

calebcartwright (approved):

PR #7978, opened 2022-05-17
Upgrade Gatsby

chris48s:
I'm not necessarily proposing we should merge this now. This PR shows how we _can_ upgrade Gatsby and get the tests passing if we need to, but we might choose to leave it for now.

The issue we've got with gatsby/express-graphql/graphql is basically around type checking/incompatibility with tsc. Using [skipLibCheck](https://www.typescriptlang.org/tsconfig#skipLibCheck) disables all type checking in `.d.ts` files. See https://bobbyhadz.com/blog/typescript-disable-type-checking-node-modules . There is a tradeoff here. We do lose some type safety not only within `node_modules` but also to a lesser extent within our own codebase: It does also disable https://github.com/badges/shields/blob/master/frontend/types/assets.d.ts and https://github.com/badges/shields/blob/master/frontend/types/mapbox__react-click-to-select/index.d.ts . I did have a look to see if it was possible to do this conditionally using `exclude` but I wasn't able to get it to work. I got this workaround from https://github.com/gatsbyjs/gatsby/issues/35212 which I am now watching for updates (but also my level of faith that it will be solved soon is quite low).

One option would be we could merge this and accept this tradeoff in exchange for being back on the gatsby release train. We could also leave things as they are and stay pinned to old gatsby. However it is useful to know we've got this "in our back pocket" if we _have_ to upgrade gatsby because it has become a blocker to other updates (e.g: node 18) or there is a security fix we need to apply.

This wouldn't allow us to upgrade graphql - we're still stuck pinned to graphql 15 due to `peerDependencies` but it would allow us to upgrade gatsby.


chris48s:
Hmm. I've run the e2e tests locally and they are passing on my machine, but I've tried restarting the CI build a couple of times and it is still failing :angry: . I'll have to come back to this one another time.

paulmelnikow:
Hey, just my 2c that I think the type-checking trade off here is fine. We could even move the types upstream for click-to-select (I think I’m a maintainer) or into DT.

Am eager to replace Gatsby but that’s a separate discussion. GraphQL is clearly not at all necessary for what we are doing and as much as I like it, I think it is fairly overkill as a batteries-included technology for static-site generation.

chris48s:
Yeah it is an annoyance that gatsby stuff (where we're not using any graphql stuff) is preventing us from upgrading graphql and graphql-tag on the backend (where we are actually using for services that extend `BaseGraphqlService`). Straying a bit off topic here, but do you have any thoughts on what a good replacement for Gatsby is? We were using next.js before and decided we didn't like that. I don't really know what the other options are in that space.

Either way, I need to work out what is going on with these e2e tests. I managed to intermittently reproduce these failures locally by running `GATSBY_BASE_URL=http://localhost:8080 npm run build && npm run e2e-on-build` instead of just `npm run e2e` and then "fixed" it by upping the timeout but I tried increasing the timeout on this PR to like 20 seconds and it was still failing in CI. It totally does work interactively in a browser though.

chris48s:
Not sure what that tells us yet, but its... something

paulmelnikow:
Hmm… for troubleshooting I might suggest zipping up the build and saving it as a build artifact, and then running that locally, just to make sure it both renders okay and passes the tests.

If that works fine then we definitely have a Cypress-in-CI problem…

calebcartwright:
Probably won't have an impact on CI failures, but would it be worth turning off video recording? (`"video": false` in cypress config).

> I tried increasing the timeout on this PR to like 20 seconds and it was still failing in CI.

Which setting/env var/etc. does Cypress use? I probably just missed one of the prior runs but current CI failures are complaining about timing out after 4,000ms and I know locally I was seeing a few (including the ones CI is complaining about) take north of 10s, especially when video recording was enabled

chris48s:
> Which setting/env var/etc. does Cypress use/honor for timeouts?

I've force-pushed this branch a few times but I was using `defaultCommandTimeout`  https://github.com/badges/shields/commit/41b0771fd246f52bb7abedda694529c4eed7b7b6

LekoArts:
Have you tried just updating `graphql` e.g. via https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides and see if it breaks?

chris48s:
I've tried it with `npm upgrade graphql --force`. I've not tried it with overrides. We'd have to drop `strict-peer-deps=true` but I'll try it with overrides in the next few days.

chris48s:
I tried 2 variations of this. One was

```json
  "overrides": {
    "gatsby": {
      "graphql": "^16.5.0"
    }
  }
```

That one allows us to upgrade graphql for the backend stuff where we are actually using it but we have to disable `strict-peer-deps=true`.

The other was

```json
  "overrides": {
    "gatsby": {
      "graphql-compose": "9.0.4"
    },
    "gatsby-plugin-manifest": {
      "graphql-compose": "9.0.4"
    },
    "gatsby-plugin-utils": {
      "graphql-compose": "9.0.4"
    }
  }
```

which makes everything use a version that wants graphql <= 16

Both of those generate a working application (as does this PR), but also both of those solutions deliver us another different set of typescript errors on the frontend build. This time round it is

```
node_modules/@types/readable-stream/index.d.ts:19:15 - error TS2417: Class static side 'typeof _Readable' incorrectly extends base class static side 'typeof Readable'.
  The types of 'Stream.Readable.Duplex' are incompatible between these types.
    Property 'isDisturbed' is missing in type 'typeof _Readable.Duplex' but required in type 'typeof import("stream").Duplex'.

19 declare class _Readable extends stream.Readable {
                 ~~~~~~~~~

  node_modules/@types/node/stream.d.ts:59:20
    59             static isDisturbed(stream: Readable | NodeJS.ReadableStream): boolean;
                          ~~~~~~~~~~~
    'isDisturbed' is declared here.

node_modules/@types/readable-stream/index.d.ts:68:11 - error TS2720: Class 'Duplex' incorrectly implements class '_Readable'. Did you mean to extend '_Readable' and inherit its members as a subclass?
  Type 'Duplex' is missing the following properties from type '_Readable': readableAborted, readableDidRead

68     class Duplex extends Writable implements /*extends*/_Readable, stream.Duplex {
             ~~~~~~

node_modules/@types/readable-stream/index.d.ts:68:11 - error TS2720: Class '_Readable.Duplex' incorrectly implements class 'import("stream").Duplex'. Did you mean to extend 'import("stream").Duplex' and inherit its members as a subclass?
  Type 'Duplex' is missing the following properties from type 'Duplex': readableAborted, readableDidRead

68     class Duplex extends Writable implements /*extends*/_Readable, stream.Duplex {
             ~~~~~~

node_modules/@types/readable-stream/index.d.ts:111:11 - error TS2720: Class '_Readable.PassThrough' incorrectly implements class 'import("stream").PassThrough'. Did you mean to extend 'import("stream").PassThrough' and inherit its members as a subclass?
  Type 'PassThrough' is missing the following properties from type 'Transform': readableAborted, readableDidRead

111     class PassThrough extends Transform implements stream.PassThrough {
              ~~~~~~~~~~~

node_modules/@types/readable-stream/index.d.ts:173:11 - error TS2720: Class '_Readable.Transform' incorrectly implements class 'import("stream").Transform'. Did you mean to extend 'import("stream").Transform' and inherit its members as a subclass?
  Type 'Transform' is missing the following properties from type 'Transform': readableAborted, readableDidRead

173     class Transform extends Duplex implements stream.Transform {
              ~~~~~~~~~


Found 5 errors in the same file, starting at: node_modules/@types/readable-stream/index.d.ts:19
```

which is similar to the errors we are looking at in

- https://app.circleci.com/pipelines/github/badges/shields/13238/workflows/9acc6569-f79d-43c2-9b8f-8cc0f3dc18f8/jobs/185012 and 
- https://app.circleci.com/pipelines/github/badges/shields/13248/workflows/0ef54416-c723-45b9-9b4a-f342e9394354/jobs/185080

so we'd also need `"skipLibCheck": true` if we wanted to use either of those solutions so.. meh

I'm starting to think `"skipLibCheck": true` may be the way forward given it would probably also unblock some other upgrade PRs too.

chris48s:
revived this in https://github.com/badges/shields/pull/8241 not going to bother updating this PR

PR #8003, opened 2022-05-25
Docstrings for contributor-count and deprecation-helpers services

pr7prashant:
- Added docstrings for `contributor-count` service
- Deleted `deprecation-helpers` service as it is not being used anywhere


services/deprecation-helpers.js
```diff
@@ -1,5 +1,19 @@
...
 
+/**
+ * Checks if the effective date is deprecated or not by comparing it with the current time.
+ * A date is considered to be deprecated if its timestamp is less than or equal to the current timestamp.
+ * Deprecated Error will be thrown if the effective date is deprecated.
+ *
+ * @param {object} effectiveDate Date object
+ * @throws {Deprecated} Deprecated Error
+ */
 function enforceDeprecation(effectiveDate) {
```
chris48s, line 17: This is an interesting function in that.. we're not actually Calling it anywhere :D

I think we must have been using this in the past but it has fallen out of use. I think we can just delete `deprecation-helpers.js` and `deprecation-helpers.spec.js` rather than documenting it.

pr7prashant: sure :)

calebcartwright (approved):
lgtm, thanks!

Issue #7371, opened 2021-12-11
Badge request: GitLab License


DaRubyMiner360:
It would be useful to have the ability to see the license from the README of a GitLab repository. You can already do it with GitHub, after all.

Example:
![License](https://img.shields.io/badge/license-MIT-green)

There is a public API with an API token and it's already used for other GitLab badges.

calebcartwright:
Thanks for the suggestion!

> You can already do it with GitHub, after all.

Want to note that just because a badge exists for service Foo doesn't make a similar badge automatically come into existence for some other service Bar :wink:

In order for a badge to be added someone has to actually add the code for it, and that doesn't typically happen unless and until someone asks for it. There's also a rather lengthy history with added GitLab badges that was concluded in https://github.com/badges/shields/issues/541#issuecomment-919580720.

> There is a public API with an API token and it's already used for other GitLab badges.

Please do the work to gather the details/link on the API as requested in the issue template, and then either update your issue description with that info or include it in a follow up comment. The more information and actionable you make the issue the greater the chance that someone will work on it.

sunny0826:
Maybe I can contribute this part of the code. @calebcartwright Can you assign this Issue to me?

chris48s:
bit late closing the issue, but this was added in #8024

PR #8024, opened 2022-05-31
Add [gitlablicense] services

sunny0826:
## Overview
This service is for displaying the license of porject in GitLab.

re #7371


calebcartwright:
Updated title so that service tests will run, and closing/reopening since CI doesn't seem to be executing

services/gitlab/gitlab-license.service.js
```diff
@@ -0,0 +1,73 @@
...
+  documentation,
+}
+
+export default class GitlabLicense extends GitLabBase {
+  static category = 'license'
+
+  static route = {
+    base: 'gitlab/v/license',
+    pattern: ':project+',
+  }
```
calebcartwright, line 29: I think we should include the standard query parameter that allows users to specify the GitLab instance to use. (It looks like you structured the `handle` function to accept this parameter already, but we still need to wire up the route properly to ensure the query param is actually available)

You can see an example of this in the GitLab release badge:
https://github.com/badges/shields/blob/fd13db83bab6450e38245cafb9c81cb6c5291225/services/gitlab/gitlab-release.service.js#L14-L15

https://github.com/badges/shields/blob/fd13db83bab6450e38245cafb9c81cb6c5291225/services/gitlab/gitlab-release.service.js#L39-L43

https://github.com/badges/shields/blob/fd13db83bab6450e38245cafb9c81cb6c5291225/services/gitlab/gitlab-release.service.js#L132-L135

services/gitlab/gitlab-license.service.js
```diff
@@ -0,0 +1,73 @@
...
+  static examples = [
+    {
+      title: 'GitLab',
+      ...commonProps,
+      staticPreview: {
+        label: 'license',
+        message: 'MIT',
+        color: 'green',
+      },
+    },
```
calebcartwright, line 40: Will probably need to add an example that includes a custom instance usage too

services/gitlab/gitlab-license.tester.js
```diff
@@ -0,0 +1,57 @@
...
+          html_url: 'http://choosealicense.com/licenses/apache-2.0/',
+          source_url: '',
+        },
+      })
+  )
+  .expectBadge({
+    label: 'license',
+    message: 'Apache License 2.0',
+    color: unknownLicenseColor,
+  })
```
calebcartwright, line 57: I don't think we necessarily have to remove this test (in part because I realize it bears similarity to the corresponding github tests), but did want to note that it's not necessary for purposes of service/integration tests.

The license renderer is already extensively tested so we're typically content not trying to retest that logic in service tests

calebcartwright (commented):
Thank you! Few minor things inline but looking great so far

services/gitlab/gitlab-license.service.js
```diff
@@ -0,0 +1,73 @@
...
+  ]
+
+  static defaultBadgeData = { label: 'license' }
+
+  static render({ license }) {
+    if (license) {
+      return renderLicenseBadge({ license })
+    } else {
+      return { message: 'not specified' }
+    }
```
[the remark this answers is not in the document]
sunny0826: I refer to the Github licence for this part.

https://github.com/badges/shields/blob/dc06b445c9a44849ff35676a6305808b6020344b/services/github/github-license.service.js#L29-L37

services/gitlab/gitlab-license.tester.js
sunny0826: Yep, I wrote this test more to show the data structure of `licence`, so that the developer can understand what fields are in the structure directly through this test case, without having to actually call it.

services/gitlab/gitlab-license.service.js
sunny0826: done

services/gitlab/gitlab-license.service.js
sunny0826: done

services/gitlab/gitlab-license.service.js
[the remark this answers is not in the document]
calebcartwright: Interesting, thanks for sharing. I'd be curious to know what other maintainers think, if there's more backstory on why GitHub is handled this way. Given the goals of the project I'd be hesitant to introduce this another case of inconsistent behavior in the no license scenario, unless there's good reason

services/gitlab/gitlab-license.tester.js
calebcartwright: Understood, and again don't feel like you need to change anything, but my point is that we don't really do that especially not with our service tests. The schema we define in the service class that is used for validating response objects from the upstream APIs is how we typically "document" the structure and relevant fields, along with trying to include links to the upstream documentation.

Compared to other types of tests our service tests are slow and expensive and we typically use them to provide validation against the upstream services we integrate with/to catch when vendors/platforms make breaking changes to their APIs. They can be used to test/do other things, it's just that we typically try to refrain from doing so because of the associated drawbacks.

This is just an fyi/food for thought in case you decide you'd like to work on other badges in the future, GitLab or otherwise

sunny0826:
@calebcartwright Can this PR be merged, and I hope to contribute a series of badges about GitLab afterwards.

calebcartwright:
> @calebcartwright Can this PR be merged, and I hope to contribute a series of badges about GitLab afterwards.

https://github.com/badges/shields/pull/8024/files#r890712363 was still pending in my mind. However, I think my fellow maintainers have been quite busy of late and haven't had much bandwidth for Shields, and I suspect that will continue to be the case for the foreseeable future.

I'll do one more pass and proceed with merging if nothing else jumps out

calebcartwright (approved):
Will go ahead and approve but also note a few things that may/will require follow up:

* The behavior discussed in https://github.com/badges/shields/pull/8024#discussion_r890712363 should be revisited, at least to confirm this is desired. I don't have any concerns with proceeding though as we should be able to change this semi-error path later on.
* We strive to have our sample parameters used in defaults be valid/usable, so if a user utilizes them to test the badge will work. However, the example listing with the custom server url (https://jihulab.com) ends up being invalid because it is reusing the shared/common props, and there is no gitlab-org/gitlab-foss project on that server. We should update the server url or use explicit project args for the existing server in order for that example to be valid

sunny0826:
> Will go ahead and approve but also note a few things that may/will require follow up:
> 
> * The behavior discussed in [#8024 (comment)](https://github.com/badges/shields/pull/8024#discussion_r890712363) should be revisited, at least to confirm this is desired. I don't have any concerns with proceeding though as we should be able to change this semi-error path later on.
> * We strive to have our sample parameters used in defaults be valid/usable, so if a user utilizes them to test the badge will work. However, the example listing with the custom server url (https://jihulab.com) ends up being invalid because it is reusing the shared/common props, and there is no gitlab-org/gitlab-foss project on that server. We should update the server url or use explicit project args for the existing server in order for that example to be valid

jihulab.com is GitLab's SaaS platform in China, and it is always available. Maybe change the project to `gitlab-cn/gitlab`, which is public and is used to collaborate with GitLab.com.

calebcartwright:
> Maybe change the project to gitlab-cn/gitlab, which is public and is used to collaborate with GitLab.com.

Either way, doesn't really matter. Just want to target a working/valid combination of server url and project reference

PR #8027, opened 2022-06-01
Docstrings for dynamic-common service

pr7prashant:
- Added docstrings for `dynamic-common` service


services/dynamic-common.js
```diff
@@ -1,22 +1,54 @@
...
+ * @type {object}
+ */
 const compoundValueSchema = Joi.alternatives().try(
   individualValueSchema,
   Joi.array().items(individualValueSchema).required(),
   Joi.array().length(0)
 )
 
+/**
+ * Transforms and validates the data against compoundValueSchema.
```
chris48s, line 44: I think a better description of this function for the docs would be something like


"Extract value from data object by key and validate value against compoundValueSchema."

or

"Look up value in data object by key and validate value against compoundValueSchema."

chris48s (commented):
Thanks for continuing to document these - its a big help.

services/dynamic-common.js
pr7prashant: This looks better. I'll update the docs. Thanks

pr7prashant: > Thanks for continuing to document these - its a big help.

No worries, hopefully, we can finish the entire documentation this year :)

chris48s (approved):

PR #8044, opened 2022-06-07
🌱 Update OpenSSF Scorecard badge text

azeemshaikh38:
Current badge label for OpenSSF Scorecard only shows `score` which is very vague. So updating this to `openssf scorecard` (similar to other openssf badge, Best Practices - https://bestpractices.coreinfrastructure.org/projects/5621/badge)

Some lines were auto changed due to linters but do not affect functionality.

### DIFF
--- a/services/ossf-scorecard/ossf-scorecard.service.js
+++ b/services/ossf-scorecard/ossf-scorecard.service.js
@@ -21,7 +21,7 @@ export default class OSSFScorecard extends BaseJsonService {
 
   static examples = [
     {
-      title: 'OSSF-Scorecard Score',
+      title: 'OpenSSF Scorecard Score',
       namedParams: {
         host: 'github.com',
         orgName: 'rohankh532',
--- a/services/ossf-scorecard/ossf-scorecard.service.js
+++ b/services/ossf-scorecard/ossf-scorecard.service.js
@@ -31,7 +31,7 @@ export default class OSSFScorecard extends BaseJsonService {
     },
   ]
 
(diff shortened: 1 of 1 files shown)


calebcartwright:
Thank you for the PR, but I am going to close because the thrust of this change around updating the default label text is something we do not want.

Our badge defaults are guided by our [Specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md) which is explicit about being non-promotional and non-branding, which is why our default badge labels describe the metric/value as opposed to the service or platform that provides those values.

Users that want a different label can always utilize our customization options, namely `?label=foo` to override the default value to their preference.

If you'd like to pursue any of the other changes I'd suggest opening a new PR to maintain clean separation from this one, and if you do open a new one, please also be sure follow the [PR title naming conventions](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#running-service-tests-in-pull-requests) to ensure that the proper test suites are executed

azeemshaikh38:
Thanks for the explanation @calebcartwright. Using `?label=foo` should provide what I'm looking and I can do that on my end. 

Would you be ok if I updated the default label text to `security score`? That would be less vague than just `score`. If so, I'll send out another PR for it (with correct naming conventions).

calebcartwright:
> Would you be ok if I updated the default label text to security score? That would be less vague than just score. If so, I'll send out another PR for it (with correct naming conventions).

Acknowledge seeing this @azeemshaikh38 but still thinking on it. My hesitation is that it would potentially necessitate, or at least open the door, for extending the label on all the other badges and I'm just not convinced that's necessary yet.

PR #8065, opened 2022-06-10
Remove invalid await

iamphduc:
There is no async function to use await correctly


iamphduc:
Thanks for your reviewing! Silly me that I didn't know top-level await existed.

I've learned a new thing today xD

calebcartwright:
No worries!

PR #8073, opened 2022-06-13
increase [reuse] service test timeout

calebcartwright:
This has been failing consistently for a while, as apparently the not found scenarios take more than the default 5 seconds (hovering right around 7 seconds on my local machine)


calebcartwright:
Closing in favor of #8519

Issue #5505, opened 2020-09-08
Badge request : openuserjs.org, greasyfork.org


NatoBoram:
:clipboard: **Description**

Services : openuserjs.org, greasyfork.org

Information : Install count, rating, version code

Examples : 

![OpenUserJS](https://img.shields.io/badge/OpenUserJS-3%20installs-%232c3e50)
![GreasyFork](https://img.shields.io/badge/GreasyFork-3%20installs-%23990000)

:link: **Data**

* No public API
* No key needed

Data can be directly accessed from the HTML file.

:microphone: **Motivation**

This feature can help assert a certain "reputation" based on install count for public scripts.

More specifically, I want to be able to show off that people have installed my userscript.

calebcartwright:
Thanks for the request @NatoBoram. However, we can't do html scraping to retrieve the requisite data for the badges. If these tools/services do not provide an endpoint to retrieve the data, then we can't provide native badges for them, but you could leverage our [Endpoint badge](https://shields.io/endpoint) instead.

NatoBoram:
Found the API for GreasyFork!

* `https://greasyfork.org/scripts/411035.json`
* ![GreasyFork](https://img.shields.io/badge/dynamic/json?color=%23990000&label=GreasyFork&query=total_installs&suffix=%20installs&url=https%3A%2F%2Fgreasyfork.org%2Fscripts%2F411035.json)
* `https://img.shields.io/badge/dynamic/json?color=%23990000&label=GreasyFork&query=total_installs&suffix=%20installs&url=https%3A%2F%2Fgreasyfork.org%2Fscripts%2F411035.json`

PyvesB:
@NatoBoram is the API documented anywhere? It would be quite useful!

jerone:
Related issue: #829

NatoBoram:
I don't think there's a documentation anywhere. I found out in https://github.com/JasonBarnabe/greasyfork/issues/247 that you can add `.json` to some web pages and it'll return a JSON response. 

Their example is https://greasyfork.org/users/173559.json, so I tried on my script and https://greasyfork.org/scripts/411035.json works.

I asked in https://github.com/JasonBarnabe/greasyfork/issues/787.

Martii:
[...]

Issue #829, opened 2016-11-06
[FEATURE] Support for userscript version


jerone:
Since the dead of Userscripts.org a few years ago, a lot of script-writers moved their userscripts to GitHub. It would be great if there was a shield showing the `@version` [<sup>1</sup>](https://wiki.greasespot.net/Metadata_Block#.40version) [<sup>2</sup>](http://tampermonkey.net/documentation.php?ext=dhdg#_version) of an userscript. Other metadata would be great too. I also noticed that userscripts mostly don't live on the repos root, so navigating to the `.user.js` file should be part of the shield url.

paulmelnikow:
Hi, thanks for the suggestion. Could you link to some examples?

jerone:
Sure thing. 

For example the following repo: https://github.com/jerone/UserScripts
With the following userscript file: [/Github_Reply_Comments/Github_Reply_Comments.user.js](https://github.com/jerone/UserScripts/blob/master/Github_Reply_Comments/Github_Reply_Comments.user.js)
With the `@version` property on [line 14](https://github.com/jerone/UserScripts/blob/master/Github_Reply_Comments/Github_Reply_Comments.user.js#L14).

There are [more userscripts repos](https://github.com/search?q=topic%3Auserscripts&type=Repositories)  on GitHub and all have at least an `@version` property.

paulmelnikow:
Thanks!

We have [a tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md). Would you like to open a pull request?

Issue #5157, opened 2020-06-01
Add support for Greasy Fork and UserScriptJS


mysticflute:
:clipboard: **Description**

It would be cool to show badges from user script installs or version.

mysticflute:
I added support for this using RunKit. https://github.com/mysticflute/user-script-badges-endpoint

PR #8080, opened 2022-06-13
[greasyfork] Add Greasy Fork service badges

### LINKED ISSUE #8074: [Badge Request] Greasy Fork
:clipboard: **Description**

Greasy Fork stats and info

### Examples

**Version**

image

**Downloads**

image
image

**Rating**

image
image

**License**

image

:link: https://greasyfork.org

API:  `https://greasyfork.org/scripts/<id>.json` where `<id>` is replaced by a script ID, e.g. https://greasyfork.org/scripts/407466.json

An easy way to get this URL is by adding `.json` at the end your script's listing page.
https://greasyfork.org/en/scripts/407466-unedit-for-reddit -> https://greasyfork.org/en/scripts/407466-unedit-for-reddit.json
The ID is only important part; the URL will still work no matter what words if any are following the script ID.

No API key is required.

The existence of the API is documented at: https://github.com/JasonBarnabe/greasyfork/wiki/API

Example of data returned:

```json
{
  "id": 407466,
  "created_at": "2020-07-20T19:52:40.000Z",
  "daily_installs": 18,
  "total_installs": 3412,
  "code_updated_at": "2022-06-13T01:05:17.000Z",
  "support_url": null,
  "fan_score": "61.5",
  "namespace": "http://tampermonkey.net/",
  "contribution_url": null,
  "contribution_amount": null,
  "good_ratings": 17,
  "ok_ratings": 2,
  "bad_ratings": 3,
  "users": [
    {
      "id": 668659,
      "name": "DenverCoder1",
      "url": "https://greasyfork.org/users/668659-denvercoder1"
    }
  ],
  "name": "Unedit and Undelete for Reddit",
  "description": "Creates the option next to edited and deleted Reddit comments/posts to show the original comment from before it was edited",
  "url": "https://greasyfork.org/scripts/407466-unedit-and-undelete-for-reddit",
  "code_url": "https://greasyfork.org/scripts/407466-unedit-and-undelete-for-reddit/code/Unedit%20and%20Undelete%20for%20Reddit.user.js",
  "license": null,
  "version": "3.9.3",
  "locale": "en",
  "deleted": false
}
```

:microphone: **Motivation**

It would be useful for GitHub repositories hosting userscript code.

For example, similar to Chrome Web Store and Mozilla Add-ons, people will want to display the current version number and user statistics.

DenverCoder1:
Closes #8074
Partially resolved issues: #5505, #829

image

<details><summary>Reverted rating badges</summary>

image

</details>


calebcartwright:
#5157 too

calebcartwright:
Could you help me understand the difference between daily and total installs, and why the total installs field is represented as users?

DenverCoder1:
Total installs = total number of installs
Daily installs = average installs per day

I suppose total installs could be under "downloads" instead of users

calebcartwright:
> I think I will change it to dt (consistent with Eclipse Marketplace)

Yeah let's do that, `dd` and `dt` variants would be consistent with how we handle downloads available over different intervals https://github.com/badges/shields/blob/master/doc/badge-urls.md#badge-url-conventions

DenverCoder1:
I have updated the code with the discussed change

services/greasyfork/greasyfork-downloads.service.js
```diff
@@ -0,0 +1,58 @@
...
+    return renderDownloadsBadge({ downloads })
+  }
+
+  async handle({ scriptId }) {
+    const data = await this.fetch({ scriptId })
+    return this.constructor.render({ downloads: data.total_installs })
+  }
+}
+
+export { GreasyForkDailyInstalls, GreasyForkTotalInstalls }
```
calebcartwright, line 58: This can and should be combined into a single class, both because it creates opportunities for a richer user experience on Shields.io similar to other services with multiple variants, and to avoid code duplication.

The only variable parts are the field in the response and the `interval` value to pass to the render helper which can be handled via inline logic. Having a single class with a variant-type route parameter will handle exposing both options to end users and be more consistent with what we do for other classes. Some examples for reference

https://github.com/badges/shields/blob/2d7d8c5b784cbcb7f2728a818d7c36a73dc71030/services/gem/gem-downloads.service.js#L28

https://github.com/badges/shields/blob/2d7d8c5b784cbcb7f2728a818d7c36a73dc71030/services/crates/crates-downloads.service.js#L9

services/greasyfork/greasyfork-downloads.service.js
```diff
@@ -0,0 +1,58 @@
...
+
+  static examples = [
+    {
+      title: 'Greasy Fork',
+      namedParams: { scriptId: '407466' },
+      staticPreview: this.render({ downloads: 3420 }),
+    },
+  ]
+
+  static _cacheLength = 21600
```
calebcartwright, line 46: Could you share your thinking on why you're overriding this?

The default value we use for download/install type badges is 900 which strikes a good balance between having fresh/accurate data and reducing spam against the upstream server. We typically don't increase it unless the latter becomes an issue

services/greasyfork/greasyfork-downloads.tester.js
```diff
@@ -0,0 +1,24 @@
+import { ServiceTester } from '../tester.js'
+import { isMetric, isMetricOverTimePeriod } from '../test-validators.js'
+
+export const t = new ServiceTester({
+  id: 'GreasyForkInstalls',
+  title: 'GreasyFork Installs',
+  pathPrefix: '/greasyfork',
+})
```
calebcartwright, line 8: Note that once these are combined into a single class with a single export then this can be simplified to use the standard `createServiceTester` helper (more info in https://github.com/badges/shields/blob/master/doc/service-tests.md#1-boilerplate)

services/greasyfork/greasyfork-license.service.js
```diff
@@ -0,0 +1,42 @@
...
+      namedParams: { scriptId: '407466' },
+      staticPreview: this.render({ licenses: ['MIT'] }),
+    },
+  ]
+
+  static defaultBadgeData = { label: 'license' }
+
+  static render({ licenses }) {
+    return renderLicenseBadge({ licenses })
+  }
```
calebcartwright, line 25: In cases where we're just utilizing a common helper like this there's no need to have a separate class-level render function, and it's fine to just call the helper directly instead at the call sites where `render` is being used

services/greasyfork/greasyfork-license.service.js
```diff
@@ -0,0 +1,42 @@
...
+    if (data.license === null) {
+      throw new InvalidResponse({
+        prettyMessage: 'version not found',
+      })
+    }
+    // remove suffix " License" from data.license
+    const license = data.license.replace(/ License$/, '')
+    return this.constructor.render({
+      licenses: [license],
+    })
```
calebcartwright, line 38: This is fine and don't feel obligated to make any changes, but for awareness, a common pattern we use in our service classes is to incorporate a `transform` function which takes the raw response and does any such extraction and/or transformation. We do this both for cleaner separation of concerns and because it's much easier to test with standard unit tests vs. trying to use nock with mocked requests for validating code-level statements

services/greasyfork/greasyfork-license.tester.js
```diff
@@ -0,0 +1,16 @@
+import { ServiceTester } from '../tester.js'
+
+export const t = new ServiceTester({
+  id: 'GreasyForkLicense',
+  title: 'Greasy Fork License',
+  pathPrefix: '/greasyfork',
+})
```
calebcartwright, line 7: This should use the helper model too, similar to comment above in previous tester

services/greasyfork/greasyfork-rating.service.js
```diff
@@ -0,0 +1,132 @@
+import { floorCount as floorCountColor } from '../color-formatters.js'
```
calebcartwright, line 1: (Speaking to this entire class, not specific to this one line)

I'm not sure how much value there is in a set of badges that just display the counts, but perhaps I'm just not familiar enough with the service.

What comes to mind for me as far as existing badges we provide that have some level of similarity are ones like VS/Open VSX rating and pkgreview where we display the rating (star count), optionally with the quantity of reviews as a suffix. Would something like that even be feasible with Greasy Fork or are the actual rating values/aggregated rating values not available through the API?

calebcartwright (commented):
Thanks for this! This is looking good overall, with a few inline questions and requests for changes below

services/greasyfork/greasyfork-downloads.service.js
DenverCoder1: It was copied from AMO as I used that as a starting point, I can remove it

services/greasyfork/greasyfork-rating.service.js
DenverCoder1: Greasy Fork doesn't really have an aggregated score or star system like most platforms.

This is how ratings are displayed on their website, just counts for Good, OK, and Bad ratings. When rating scripts, you select one of these 3 categories.

image

services/greasyfork/greasyfork-license.service.js
DenverCoder1: I seem to get the error `render() function not implemented for GreasyForkInstalls` when removing `render`, do you know how to get around that?

services/greasyfork/greasyfork-rating.service.js
calebcartwright: Thanks for sharing, very helpful. I think it would be best to at least pull out the rating badges from this PR so we can go ahead and get the rest of them merged, and then circle back to if/when/how on the ratings badges

calebcartwright:
Everything else looks good to me, so happy to move forward once we get the ratings badges pulled out to a separate PR or issue for discussion

DenverCoder1:
Thanks, I have removed them from this PR

calebcartwright (dismissed):
Thanks a ton! This has been requested a few times over the years so will be nice to get this to users

Martii:
@DenverCoder1 

As I'm not on GF usually you may want to double check multiple licenses and any other key that normally supports more than a single JSON key value pair just like was done on OUJS. As I mentioned on the OUJS PR all key value pairs in all .user.js engines are required to be LIFO. I don't know if GF reverses the arrays for multiple key or not like they should be.

OUJS Admin

DenverCoder1:
@Martii 

> As I'm not on GF usually you may want to double check multiple licenses and any other key that normally supports more than a single JSON key value pair just like was done on OUJS. As I mentioned on the OUJS PR all key value pairs in all .user.js engines are required to be LIFO. I don't know if GF reverses the arrays for multiple key or not like they should be.

I did test it and found that GF does not seem to do anything about multiple licenses. If the key appears multiple times, only the *first* license appears in the JSON.

As the value appears to only ever be a string, the solution I've done here seems to be the only option unless they change it on their end.

(eg: see https://greasyfork.org/en/scripts/446487?version=1060911 and https://greasyfork.org/en/scripts/446487.json?version=1060911)

Martii:
> As the value appears to only ever be a string, the solution I've done here seems to be the only option unless they change it on their end.

Agreed. When he adopted our SPDX code implementation I think he may have missed that long time .user.js standard... perhaps even before that adoption.

services/greasyfork/greasyfork-license.service.js
```diff
@@ -0,0 +1,34 @@
...
+      staticPreview: renderLicenseBadge({ licenses: ['MIT'] }),
+    },
+  ]
+
+  static defaultBadgeData = { label: 'license' }
+
+  transform({ data }) {
+    if (data.license === null) {
+      throw new InvalidResponse({
+        prettyMessage: 'version not found',
```
calebcartwright, line 22: Should this be license instead?

```suggestion
        prettyMessage: 'license not found',
```

calebcartwright:
TBH I'm not 100% sure I followed all the recent commentary, but it sounds like it's sorted?

@DenverCoder1 - I spotted one last minor thing, and if you could also rebase/merge to get your branch updated with the latest, or allow the maintainers to edit your PR (so our bots can do their thing) that'd be great. We have the branch protection policy enabled that requires branches be up to date, so that will be required for merging

DenverCoder1:
Yeah, everything should be good.

I'm not sure I see an option to allow maintainer edits and I'm not sure why it would be disabled, but I've gone ahead and merged master.

calebcartwright:
> Yeah, everything should be good.
> 
> I'm not sure I see an option to allow maintainer edits and I'm not sure why it would be disabled, but I've gone ahead and merged master.

This article may help, although there could just be something funky going on with GitHub

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

DenverCoder1:
> This article may help, although there could just be something funky going on with GitHub
> 
> [docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

Yeah, I've looked before and I'm still not seeing that option. Possibly it could have to do with contributing from an organization account? I'm not sure.

calebcartwright:
> Possibly it could have to do with contributing from an organization account? I'm not sure.

Yup that's it. Didn't realize your account was setup as an org. Will just be something to keep an eye on for current open PRs and any potential future ones

PR #8082, opened 2022-06-13
feat: Dynamic badge text formatters [Dynamic]

### LINKED ISSUE #8075: [Feature] Ability to format numbers in short form when using dynamic JSON badges
:clipboard: **Description**

If for example a JSON API endpoint you want to use has a number, eg:

```json
{
    "count": 1234
}
```

When using this endpoint on a dynamic badge, it would be nice to be able to make use of a number formatter to make its appearance more similar to other badges from shields.io.

Eg, using the JSON above, you could make a badge:

```md
https://img.shields.io/badge/dynamic/json?formatter=metric&color=088&label=count&query=%24.count&url=https%3A%2F%2Fexample.com
```

Which would display like:

image

Instead of:

image

In this example, I have used `formatter=metric` to show a possible way this could be requested in the URL, this is open to alternative suggestions.

### Text formatters

To add to this, possibly more formatters from https://github.com/badges/shields/blob/master/services/text-formatters.js can be supported.

- none
- starRating
- ordinalNumber
- metric
- omitv
- addv
- formatDate
- formatRelativeDate

DenverCoder1:
Fixes #8075 

If, for example, a JSON/XML/YAML API endpoint you want to use has a number or date, eg:

```json
{
    "count": 1234
}
```

When using this endpoint on a dynamic badge, it would be nice to be able to make use of the text formatters to make the appearance more similar to other badges from shields.io.

Eg, using the JSON above, you could make a badge:

```md
https://img.shields.io/badge/dynamic/json?formatter=metric&color=088&label=count&query=%24.count&url=https%3A%2F%2Fexample.com
```

Which would display like:

image

Instead of:

image

### Text formatters

The following values are supported for the `formatter` parameter:

- metric
image image
- starRating
image image
- ordinalNumber
image image
- omitv
image image
- addv
image image
- formatDate
image image
- formatRelativeDate
image image

### DIFF
--- a/core/badge-urls/make-badge-url.d.ts
+++ b/core/badge-urls/make-badge-url.d.ts
@@ -92,6 +92,7 @@ export function dynamicBadgeUrl({
   suffix,
   color,
   style,
+  formatter,
   format,
 }: {
   baseUrl?: string
@@ -103,6 +104,7 @@ export function dynamicBadgeUrl({
   suffix: string
   color?: string
   style?: string
+  formatter?: string
   format?: string
 }): string
 
--- a/core/badge-urls/make-badge-url.js
+++ b/core/badge-urls/make-badge-url.js
@@ -118,6 +118,7 @@ function dynamicBadgeUrl({
   suffix,
   color,
   style,
+  formatter,
   format = '',
 }) {
   const outExt = format.length ? `.${format}` : ''
@@ -137,6 +138,9 @@ function dynamicBadgeUrl({
   }
   if (suffix) {
     queryParams.suffix = suffix
+  }
+  if (formatter) {
+    queryParams.formatter = formatter
   }
 
   const outQueryString = queryString.stringify(queryParams)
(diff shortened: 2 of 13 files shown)


calebcartwright:
Closing per https://github.com/badges/shields/issues/8075#issuecomment-1154559716

DenverCoder1:
@calebcartwright Thanks for your assistance with explaining endpoint badges.

After a lot of creative planning, I managed to get my idea to work using only Runkit and the Shields.io endpoint badges.

I apologize for arguing the issue longer than necessary and I hope you find [my new solution](https://github.com/DenverCoder1/dynamic-badge-formatter) more acceptable.

calebcartwright:
No worries, and thanks so much for both working on this and for following up @DenverCoder1!

It looks great and I think it will certainly be useful for users and provide a viable, directly consumable solution we can point to for at least a chunk of the customization requests we receive for the dynamic badge :rocket: 

What I'd like to do next is to get this highlighted under Awesome Badges, so if you're up for it please feel free to submit a PR over there: https://github.com/badges/awesome-badges 

Personally, I think we should probably move towards creating a dedicated section for Endpoints with broader appeal/usability, but feel free to suggest any structure that you think makes sense.

Thanks again!

PR #8083, opened 2022-06-14
fix: gitlab licence service docs and example

sunny0826:
re https://github.com/badges/shields/pull/8024#issuecomment-1153331982

Fixed the documentation and example of custom server, now they are displayed properly.

/assign @calebcartwright

### DIFF
--- a/services/gitlab/gitlab-license.service.js
+++ b/services/gitlab/gitlab-license.service.js
@@ -21,7 +21,7 @@ const documentation = `
 
 const customDocumentation = `
 <p>
-  Note that only network-accessible jihulab.com and other self-hosted GitLab instances are supported.
+  Note that only internet-accessible GitLab instances are supported, for example https://jihulab.com, https://gitlab.gnome.org, or https://gitlab.com/.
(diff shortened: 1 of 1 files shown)


services/gitlab/gitlab-license.service.js
```diff
@@ -15,15 +15,16 @@ const queryParamSchema = Joi.object({
...
 const documentation = `
 <p>
-  You may use your GitLab Project Id (e.g. 13083) or your Project Path (e.g. gitlab-org/gitlab-foss )
+  You may use your GitLab Project Id (e.g. 278964) or your Project Path (e.g. gitlab-org/gitlab )
+</p>
+`
+
+const customDocumentation = `
+<p>
+  Note that only network-accessible jihulab.com and other self-hosted GitLab instances are supported.
```
calebcartwright, line 24: Let's adjust this a bit:

```suggestion
  Note that only internet-accessible GitLab instances are supported, for example https://jihulab.com, https://gitlab.gnome.org, or https://gitlab.com/.
```

calebcartwright (approved):

PR #8088, opened 2022-06-16
[greasyfork] Greasy Fork daily downloads minor example fix

DenverCoder1:
Noticing now that the daily downloads example is missing the "/day". The actual badge itself displays correctly, it is just an issue with the example.

image

image


DenverCoder1:
This is just fixing an example, so the tests are not affected.

PR #8105, opened 2022-06-18
style: unified self-managed [gitlab] instance name

sunny0826:
At present, there is no unified naming for the self-managed instance fields in GItLab Service, and the names that appear are: `self-host`, `custom server`, `custom instance`, in order to reduce the burden on the user's mind, the naming is unified here.

The name has been standardized to `self-managed`, a reference to the official GItLab naming scheme.

reference: https://docs.gitlab.com/ee/subscriptions/self_managed/index.html


chris48s (approved):
good change - thanks :+1:

Issue #8077, opened 2022-06-13
GitLab Service Badge tracking


sunny0826:
## Purpose

Provide more GitLab badges and track the progress of these tasks. If you have other needs, please write them in the comments.

## Existing badges

- coverage
- license
- pipeline-status
- release
- tag

## Services Tracking list

- [ ] contributors
- [ ] code size
- [ ] repo size
- [ ] issue number(opened\closed\all)
- [ ] merger request number(opened\closed\all)
- [ ] milestone
- [ ] labels
- [ ] forks
- [ ] stars
- [ ] language
- [ ] last commit
- [ ] GitLab (`go.mod` \ `package.json` \ etc.) version

## References

- https://github.com/badges/shields/tree/master/services/gitlab

/cc @calebcartwright

calebcartwright:
Thanks for writing these up!

sunny0826:
### TBD

- [ ] Encapsulate `document` texts in one place

Jieiku:
I have the badges working for gitlab: stars, contributors, last commit.

What I have not figured out is Open issues and closed issues...
for github it starts like this:

```
https://img.shields.io/github/issues-raw/
https://img.shields.io/github/issues-closed-raw/
```

What is the proper way to make those badges for gitlab? or is the feature not live yet?

sunny0826:
> I have the badges working for gitlab: stars, contributors, last commit.
> 
> What I have not figured out is Open issues and closed issues... for github it starts like this:
> 
> ```
> https://img.shields.io/github/issues-raw/
> https://img.shields.io/github/issues-closed-raw/
> ```
> 
> What is the proper way to make those badges for gitlab? or is the feature not live yet?

This feature is available. You can customize your badge in [shields.io](https://shields.io/).

just like
```
https://img.shields.io/gitlab/stars/gitlab-org/gitlab
```

Jieiku:
Ah, thanks for the link to the customizer, got it working by looking over this page: https://shields.io/category/issue-tracking

Gabbar-v7:
The repo-size requires to have role of Reporter, Developer, Maintainer, or Owner on the repo. It isn't publicly available. Is there any work around?
https://docs.gitlab.com/api/projects/#retrieve-a-project

End point: https://gitlab.com/api/v4/projects/project?statistics=true

Gabbar-v7:
Addition to the tracking list: 
- [ ] Labels with counts

PR #8108, opened 2022-06-19
feat: add [gitlabissues] service

sunny0826:
re #8077


calebcartwright:
Will try to get to this soon. It's a bit on the larger side so will need a bit more time to be carved out for a review :smile:

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+import { optionalUrl } from '../validators.js'
+import { metric } from '../text-formatters.js'
+import GitLabBase from './gitlab-base.js'
+
+const schema = Joi.object({
+  statistics: Joi.object({
+    counts: Joi.object({
+      all: Joi.number().required(),
+      closed: Joi.number().required(),
+      opened: Joi.number().required(),
```
calebcartwright, line 11: These could all take advantage of our `nonNegativeInteger` helper that can be imported from the `validators.js` module located in the root `services` directory

```suggestion
      all: nonNegativeInteger,
      closed: nonNegativeInteger,
      opened: nonNegativeInteger,
```

Also I assume these fields will always exist even in a repository that's never had any issues?

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+const customDocumentation = `
+<p>
+  Note that only internet-accessible GitLab instances are supported, for example https://jihulab.com, https://gitlab.gnome.org, or https://gitlab.com/.
+  You may use your GitLab Project Id (e.g. 13953) or your Project Path (e.g. gitlab-cn/gitlab ) in <a href="https://jihulab.com">https://jihulab.com</a>
+</p>
+`
+
+const labelDocumentation = `
+<p>
+  Support Multi-label and <a href="https://docs.gitlab.com/ee/user/project/labels.html#scoped-labels">Scoped labels</a>, Multi-lable please use <code>,</code> to separate.
```
calebcartwright, line 36: I think our scope should be limited to describing our feature to our users/what they'd need to do. In that spirit, I'm not sure we need much here, particularly given that the listing is named "issues by-label", and I don't think we need to educate our users on GitLab features to accomplish that.

I think we can probably drop this extra documentation altogether, but if you feel strongly that something is needed, I'd ask that we reduce it to simply note the need for comma delimiters, e.g. something like

```suggestion
  If you want to use multiple labels then please use commas (<code>,</code>) to separate them, e.g. <code>foo,bar</code>.
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+</p>
+`
+
+export default class GitlabIssues extends GitLabBase {
+  static category = 'issue-tracking'
+
+  static route = {
+    base: 'gitlab/v/issues',
```
calebcartwright, line 44: We shouldn't have the `v` route member, which is only applicable for version-related badges

```suggestion
    base: 'gitlab/issues',
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+    const renderLabel = `${labelPrefix}${labelText}issues`
+
+    return {
+      label: renderLabel,
+      message: `${metric(issueCount)}${
+        messageSuffix ? ' ' : ''
+      }${messageSuffix}`,
+      color: issueCount > 0 ? 'yellow' : 'brightgreen',
+    }
+  }
```
calebcartwright, line 419: I appreciate that you're (presumably) trying to model this, but we can deviate in the implementation to account for the specific nature of the upstream service and response structure. Particularly here I feel like we're conflating the response processing and rendering concerns in a manner that's not necessary and adds complexity.

For example, the entire process deriving the count of issues should be moved to a `transform` function, with the result simply being passed as an argument to this function

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+  Support Multi-label and <a href="https://docs.gitlab.com/ee/user/project/labels.html#scoped-labels">Scoped labels</a>, Multi-lable please use <code>,</code> to separate.
+</p>
+`
+
+export default class GitlabIssues extends GitLabBase {
+  static category = 'issue-tracking'
+
+  static route = {
+    base: 'gitlab/v/issues',
+    pattern: ':variant(all|opened|closed):raw(-raw)?/:project+',
```
calebcartwright, line 45: We don't have to make our user-facing naming conventions directly map to the underlying fields in the upstream responses, and more often than not, we don't want to.

In this case, I believe `open` would be a better variant name, and would also eliminate some of the re-mapping you are currently performing in the rendering function.

However, I am operating under the assumption that the GitLab issue specifics are showing the number of issues currently open (optionally with the specified labels applied), as opposed to every issue that was ever opened, including those which may be close. I make that assumption because if GitLab didn't, then I feel like the `all` and `opened` fields in the API response would be synonymous, and duplicative. As such the GitLab field name feels like a bit of a confusing misnomer, and we should correct that in our route variant.

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+      },
+      staticPreview: {
+        label: 'open test,failure::new issues',
+        message: '16',
+        color: 'yellow',
+      },
+      documentation: documentation + labelDocumentation,
+    },
+    {
+      title: 'GitLab closed issue',
```
calebcartwright, line 109:
```suggestion
      title: 'GitLab closed issues',
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+
+  static route = {
+    base: 'gitlab/v/issues',
+    pattern: ':variant(all|opened|closed):raw(-raw)?/:project+',
+    queryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'GitLab issue',
```
calebcartwright, line 51:
```suggestion
      title: 'GitLab issues',
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+      },
+      staticPreview: {
+        label: 'issues',
+        message: '44k open',
+        color: 'yellow',
+      },
+      documentation,
+    },
+    {
+      title: 'GitLab issue',
```
calebcartwright, line 64:
```suggestion
      title: 'GitLab issues',
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+      },
+      staticPreview: {
+        label: 'issues',
+        message: '72k closed',
+        color: 'yellow',
+      },
+      documentation,
+    },
+    {
+      title: 'GitLab closed issue',
```
calebcartwright, line 122:
```suggestion
      title: 'GitLab closed issues',
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+      },
+      staticPreview: {
+        label: 'closed test,failure::new issues',
+        message: '4',
+        color: 'yellow',
+      },
+      documentation: documentation + labelDocumentation,
+    },
+    {
+      title: 'GitLab all issue',
```
calebcartwright, line 167:
```suggestion
      title: 'GitLab all issues',
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+      },
+      staticPreview: {
+        label: 'issues',
+        message: '115k all',
+        color: 'yellow',
+      },
+      documentation,
+    },
+    {
+      title: 'GitLab all issue',
```
calebcartwright, line 180:
```suggestion
      title: 'GitLab all issues',
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+      staticPreview: {
+        label: 'all test,failure::new issues',
+        message: '20',
+        color: 'yellow',
+      },
+      documentation: documentation + labelDocumentation,
+    },
+
+    {
+      title: 'GitLab (self-managed) issue',
```
calebcartwright, line 210:
```suggestion
      title: 'GitLab (self-managed) issues',
```

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,446 @@
...
+        gitlab_url: 'https://jihulab.com',
+        labels: 'backend,feature::enhancement',
+      },
+      staticPreview: {
+        label: 'all backend,feature::enhancement issues',
+        message: '84',
+        color: 'yellow',
+      },
+      documentation: customDocumentation + labelDocumentation,
+    },
```
calebcartwright, line 377: I feel like we should try to cut back on the instances of example listings. Would it be possible to cut this list in half, and just add the instance url parameter on every example listing with an appropriate default, instead of having 2 of each scenario?

calebcartwright (commented):
Thank you for the continued work on the GitLab badges! I've done a pass through most of this and have left inline feedback below for your review

services/gitlab/gitlab-issues.service.js
sunny0826: > Also I assume these fields will always exist even in a repository that's never had any issues?

yep

services/gitlab/gitlab-issues.service.js
sunny0826: Indeed

services/gitlab/gitlab-issues.service.js
sunny0826: done

sunny0826:
@chris48s @calebcartwright 
Feedback on Circle CI not running issue: According to this https://github.com/badges/shields/issues/8109 , Circle CI can be triggered automatically based on push after I finish registration. Perhaps that note could be added to the contributor documentation.

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,291 @@
...
+      messageSuffix ? ' ' : ''
+    }${messageSuffix}`
+    const renderColor = issueCount > 0 ? 'yellow' : 'brightgreen'
+
+    return {
+      renderLabel,
+      renderMsg,
+      renderColor,
+    }
+  }
```
calebcartwright, line 275: Appreciate you making the updates, but you've actually swung too far in the opposite direction. There's some more detail in [this section](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#43-querying-an-api) of our tutorial, but our service classes all follow a very a standard flow (they are all essentially doing the same thing after all!) and accordingly utilize a standard set of functions. We do this for several reasons, including readability, testability, and separation of concerns amongst others.

Some classes are simple enough that they don't need to utilize every function from that set of standard functions, but the set of functions includes:

- `handle`: entry point and orchestrator
- `fetch`: retrieve raw response from upstream endpoint
- `transform`: process raw response and extract/transform needed data point(s)
- `render`: specify the final badge attributes, like label, message, and color.

The line between `transform` and `render` _can_ get a little blurry in some extreme cases, but generally speaking, the `render` function shouldn't be processing the raw response and it should strictly be focused on setting the label, etc.. Similarly, the `transform` function should only be processing the raw response and getting the result(s), which will then be passed along to the `render` function because it will use that value to determine what to specify for the badge message, color, etc.

In this case, I think a `transform` function is probably a good idea to avoid having to put too much work into the `handle` function. Such a `transform` function should be responsible for determining the issue count (based on the variant) and it also needs to account for the case where the `statistics` field is `null` in the response. That's all it should do, and it should simple return back the count of issues to the caller.

services/gitlab/gitlab-issues.service.js
```diff
@@ -0,0 +1,291 @@
...
+  ) {
+    const { statistics } = await this.fetch({
+      project,
+      baseUrl,
+      labels,
+    })
+    return this.constructor.render({
+      issues: this.constructor.transform({ variant, raw, statistics, labels }),
+    })
+    // return this.constructor.render({ variant, raw, statistics, labels })
```
calebcartwright, line 289: Please remember to remove commented out code too :smile:

sunny0826: ops...

services/gitlab/gitlab-issues.service.js
sunny0826: Thanks for your guidance. I'm actually not very good at JS, but your explanation and official documentation helped me a lot.

calebcartwright: Glad it helped. And I never would've known you didn't have much experience with JS if you hadn't mentioned it, so perhaps you're better than you think!

calebcartwright (approved):

Issue #7863, opened 2022-04-18
AUR badges sometimes provide invalid value


ReenigneArcher:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Sometimes the badges provide invalid as the value. I originally thought this is depending on the badge type. However now I believe it may be some sort of rate limit.

![image](https://user-images.githubusercontent.com/42013603/163853090-c15d6d02-027b-4563-8f1a-61c90d1f7f71.png)

![image](https://user-images.githubusercontent.com/42013603/163853142-a144bd2e-3a62-4de6-8f8a-d4a69921bbfc.png)

### 🔗 Link to the badge

https://img.shields.io/aur/maintainer/sunshine?style=for-the-badge

![AUR maintainer](https://img.shields.io/aur/maintainer/sunshine?style=for-the-badge)

![AUR maintainer](https://img.shields.io/aur/maintainer/sunshine?style=flat-square)

chris48s:
After a bit of investigation I think there is either an intermittent issue with the upstream API or we've been rate limited (possibly some of the servers behind our load balancer have been banned and some of them haven't). I'm not sure there is much we can do about this.

ReenigneArcher:
Can you cache responses from AUR for 24 hours or so, to reduce the number of hits to their API?

calebcartwright:
The [upstream API is rate limited at 4k requests per _day_](https://wiki.archlinux.org/title/Aurweb_RPC_interface#Limitations) which is a rather strict amount to begin with. For reference, we've received >18k requests for AUR badges in the last 24 hours.

There's just a fundamental conflict between the volume we're dealing with and what the upstream service is willing to support.

We do cache the badges, and I suppose we could try increasing the ttl on those to see if it helps. However, I'm not convinced that'll really solve the problem. Additionally, there have been a number of discussions over the years around caching in the responses from the upstream API response and on paper that would be beneficial in this case given that there's a family of badges all using the same endpoint. However, there's good reasons why we haven't expanded our solution architecture to incorporate that; there's a lot more complexity than just "adding cache".

I'm just not sure caching is going to be a silver bullet here, regardless of which layer(s) caching is applied/extended. I worry that given our volume we'd have to cache so heavily that the badges would be displaying information that's far too outdated in my opinion.

[...]

PR #8110, opened 2022-06-20
increase cache length on AUR version badge, run [AUR]

calebcartwright:
Refs #7863

### DIFF
--- a/services/aur/aur.service.js
+++ b/services/aur/aur.service.js
@@ -108,7 +108,7 @@ class AurVotes extends BaseAurService {
 
 class AurVersion extends BaseAurService {
   static category = 'version'
-
+  static _cacheLength = 3600
   static route = { base: 'aur/version', pattern: ':packageName' }
 
   static examples = [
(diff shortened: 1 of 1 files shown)


services/aur/aur.service.js
```diff
@@ -108,7 +108,7 @@ class AurVotes extends BaseAurService {
 
 class AurVersion extends BaseAurService {
   static category = 'version'
-
+  static _cacheLength = 3600
```
calebcartwright, line 111: Open to alternative suggestions, but I'm very intentionally swinging fairly big here. Of the >18k AUR requests we've had over the last 24 hours, nearly 16k of those were for the version badges.

I realize that version badges tend to be more sensitive to recency needs, but I think we'll need to be aggressive here if we're to have a chance of staying within AURs 4k/day limit

chris48s (approved):
might as well give it a go..

calebcartwright:
Is circle not working on in-repo PRs now too??

chris48s:
It is the first time I've seen it happen on one that is not from a fork, but apparently so..

chris48s (approved):

Issue #7660, opened 2022-03-03
Badge request: galaxy-toolshed


guillaume-gricourt:
:clipboard: **Description**

This badge is dedicated to [Galaxy Toolshed](https://toolshed.g2.bx.psu.edu/).
People could have the version of their tools in their repositories.
Like: galaxy-toolshed|v1.0.0
In repository of Galaxy Toolshed, you have a repository (identified by a name and an owner) with one or several tools (with a name and a version). Each tool could have one or several requirements and each has a name and a version.
It would be useful to grab version of a tool available in a repository and the version of a requirement corresponding to a tool.

:link: **Data**

- Is there a public API? yes
- Does the API requires an API key? no
- Link to the API documentation.  [API](https://galaxyproject.org/toolshed/api/)

:microphone: **Motivation**

The specific use case: show the badge in Github repository.
During a continuous deployment, you want to check the version of a tool in:
- github: version badge available
- conda: version badge available
- galaxy-toolshed: version badge not available <- the need

Suggest PR #7633

calebcartwright:
Shifting discussion over here vs. the PR for now.

Could you please elaborate on the specific badges you are proposing, and do so individually? The motivation field in our issue template is primarily about conveying information about the service and the underlying data points, which is especially important given the hundreds of platforms/services Shields integrates with (the majority of which are not tools we as maintainers actually use/are necessarily familiar with). We're already very aware that folks can and do put our badges in places like readme files :wink:

In particular, it would be really helpful if you could provide an overview of what Galaxy Toolshed is, how it's used, who the primary audience/users are. It would then be helpful to enumerate the badges that you'd like to see, which we can then use to have a more structured conversation around implementation in your PR

guillaume-gricourt:
Yes, let me talk about the [Galaxy Project](https://galaxyproject.org):
>Galaxy is an open-source platform for [FAIR](https://www.go-fair.org/fair-principles/) data analysis that enables users to:
>- use tools from various domains (that can be plugged into workflows) through its graphical web interface.
>- run code in interactive environments (RStudio, Jupyter...) along with other tools or workflows.
>- manage data by sharing and publishing results, workflows, and visualizations.
>- ensure reproducibility by capturing the necessary information to repeat and understand data analyses.
>The Galaxy Community is actively involved in helping the ecosystem improve and sharing scientific discoveries.

So, one of Galaxy's features runs tools through a web interface. From 2005 to 2021, 11686 scientific papers have cited the Galaxy project.
[...]

PR #8114, opened 2022-06-23
Add [galaxytoolshed] services

guillaume-gricourt:
This PR refers to the PR #7633 and the issue #7660 suggesting one feature, get the number of downloads for a repository, for a new service `galaxytoolshed`.
Many thanks


services/galaxytoolshed/galaxy-toolshed-base.js
```diff
@@ -0,0 +1,56 @@
+import Joi from 'joi'
+import { BaseJsonService } from '../index.js'
+
+const orderedInstallableRevisionsSchema = Joi.array()
+  .items(Joi.string())
+  .required()
```
calebcartwright, line 6: Is it ever possible for this to be an empty array? I suppose this could be analogous to a claimed package name that has no versions/revisions?

If it is possible, then we'll need to account for that empty array scenario in the code that handles the processing of the response. If it's not possible (or at least if we think it should be impossible), then we should update the schema:

```suggestion
const orderedInstallableRevisionsSchema = Joi.array()
  .items(Joi.string().required())
  .required()
```

services/galaxytoolshed/galaxy-toolshed-base.js
```diff
@@ -0,0 +1,56 @@
...
+    repository,
+    owner,
+    schema,
+    changesetRevision,
+  }) {
+    return this._requestJson({
+      schema,
+      url: `${this.constructor.baseUrl}/api/repositories/get_repository_revision_install_info?name=${repository}&owner=${owner}&changeset_revision=${changesetRevision}`,
+    })
+  }
```
calebcartwright, line 29: Could you explain your thinking around why `schema` is an input parameter?

Something feels off here because there's typically a strict 1:1 relationship between an upstream endpoint we are calling, and the schema we use to validate/make assertions about the structure of the response provided by that upstream endpoint.

Additionally, I assume that because you'd placed this into the base class that it's a response we'll need for other, non-download/non-install related badges?

services/galaxytoolshed/galaxy-toolshed-base.js
```diff
@@ -0,0 +1,56 @@
...
+    const changesetRevisions =
+      await this.fetchOrderedInstallableRevisionsSchema({
+        repository,
+        owner,
+      })
+    return this.fetchRepositoryRevisionInstallInfoSchema({
+      repository,
+      owner,
+      schema,
+      changesetRevision: changesetRevisions.shift(),
```
calebcartwright, line 45: My understanding is that we just need to get the revision id/value from the first element in the array (which I'm also assuming will always have at least one element as per above discussion on the schema).

If that's indeed the case, we should just grab the first element instead of grabbing it and modifying the entire collection, i.e.:

```suggestion
      changesetRevision: changesetRevisions[0],
```

the `shift()` function does result in the first element of the array being returned, but it does so by popping that element off the top and then rearranging all the other elements which would be unnecessary and wasteful work (even if it would only have minimal impact in practice)

services/galaxytoolshed/galaxy-toolshed-base.js
```diff
@@ -0,0 +1,56 @@
...
+    })
+  }
+
+  static filterRepositoryRevisionInstallInfo({ response }) {
+    return response
+      .filter(function (x) {
+        return Object.keys(x).length > 0
+      })
+      .shift()
+  }
```
calebcartwright, line 55: I think we can likely get rid of this function entirely, and I'll explain part of why that's the case here as well as on the schema that defines this object which is (currently, but probably shouldn't be) defined in one of the child classes.

Fundamentally I feel like you had this before to aid in some processing which was required based on how you were defining/utilizing the schemas. However, as alluded to in the above comment on the `fetchRepositoryRevisionInstallInfoSchema` definition, we most likely need to change that anyway which should obviate the need for this function.

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
```diff
@@ -0,0 +1,55 @@
+import Joi from 'joi'
+import { renderDownloadsBadge } from '../downloads.js'
+import BaseGalaxyToolshedService from './galaxy-toolshed-base.js'
+
+const repositoryRevisionInstallInfoSchema = Joi.array().items(
+  Joi.object({
+    times_downloaded: Joi.number().required(),
+  }).required(),
+  Joi.object({}),
+  Joi.object({})
```
calebcartwright, line 10: As per comments on the base class, this schema definition almost certainly needs to be moved to the base class where the specifics of the associated API are encapsulated.

My impression is that you were trying to use the schemas in more of a class-specific or field-specific manner but that's not really what we want to use the schemas for, and it also adds some unnecessary processing complexity.

Just to reiterate, the schemas define what the response structure should look like for a singular endpoint, and that's still true even in cases like this one where we have multiple badges in a badge family that use different fields from the same response.

What I think we need to do here is to move this schema into the base class to maintain a single definition and encapsulate it alongside its accompanying api call. We can then expand it as needed with additional fields to support additional badges.

Given that the response is structured as a fixed size array with 3 different objects, we need to define our schema to match. At the moment/purely for download count data, that would look something like this:

```suggestion
const repositoryRevisionInstallInfoSchema = Joi.array().ordered(
  Joi.object({
    times_downloaded: Joi.number().required(),
  }).required(),
).items(Joi.any())
```

Note the `.ordered()` call [which guarantees that important ordering](https://joi.dev/api/?v=17.6.0#arrayorderedtype), and the trailing `.items()` which covers the fact that there are other, differently structured elements we don't currently care about. If/when we add more badges that need elements from the second and/or third element from the response, then we can update this schema to match (happy to provide more details when/if we get to that point, but it's basically just adding corresponding entries to the `.ordered()` section

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  static examples = [
+    {
+      title: 'Galaxy Toolshed - Downloads',
+      namedParams: {
+        repository: 'sra_tools',
+        owner: 'iuc',
+      },
+      staticPreview: this.render({ downloads: 10000 }),
```
calebcartwright, line 27: Should call `renderDownloadsBadge` directly:

```suggestion
      staticPreview: renderDownloadsBadge({ downloads: 10000 }),
```

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
```diff
@@ -0,0 +1,55 @@
...
+  ]
+
+  static defaultBadgeData = {
+    label: 'downloads',
+  }
+
+  static render({ downloads }) {
+    return renderDownloadsBadge({ downloads })
+  }
+
```
calebcartwright, line 38: We don't need a dedicated `render` function if we're just passing through directly to a render helper like this 

```suggestion
```

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
```diff
@@ -0,0 +1,55 @@
...
+  static transform(response) {
+    const data = this.filterRepositoryRevisionInstallInfo({
+      response,
+    })
+    return data.times_downloaded
+  }
+
```
calebcartwright, line 45: Not strictly necessary, but we can get rid of this too since we just need to grab one field from a fixed, well known position in the response. I'll provide the specifics down below in the `handle` function but if you prefer you could keep this `transform` function and replace this function's body with the suggested line below for extracting the download count

```suggestion
```

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
```diff
@@ -0,0 +1,55 @@
...
+  async handle({ repository, owner }) {
+    const response = await this.fetchLastOrderedInstallableRevisionsSchema({
+      repository,
+      owner,
+      schema: repositoryRevisionInstallInfoSchema,
```
calebcartwright, line 50: This should go away along with moving the schema to the base class

```suggestion
```

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
```diff
@@ -0,0 +1,55 @@
...
+    })
+    const downloads = this.constructor.transform(response)
```
calebcartwright, line 52: With the aforementioned changes to the schema, we're guaranteed that `response` will be an array with at least 1 element where the first element will contain the `times_downloaded` field, so we can simply just grab it (this is a large part of why we do response schema validation in the first place, as it greatly simplifies working with/processing the response object)

We can simply destructure that object, and rename the destructured value to `downloads`

```suggestion
    const { times_downloaded: downloads } = response[0]
```

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
```diff
@@ -0,0 +1,55 @@
...
+    return this.constructor.render({ downloads })
```
calebcartwright, line 53: Same theme as the `render` function, can call helper directly:

```suggestion
    return renderDownloadsBadge({ downloads })
```

calebcartwright (commented):
Thank you for your patience, persistence, and breaking up the original PR as this is much easier to review and discuss!

I think the bones of this are in place, but there's a few changes I'd like to see and a couple questions I've left inline below

services/galaxytoolshed/galaxy-toolshed-base.js
guillaume-gricourt: Yes you're right. This array could be empty due to undefined problems.

services/galaxytoolshed/galaxy-toolshed-base.js
guillaume-gricourt: Exactly, this function will be use by others utilities: 
* releases
* activity
* version
* ...

services/galaxytoolshed/galaxy-toolshed-base.js
guillaume-gricourt: Yes, I forgot this stuff. This is managed by the schema definition. 7a37735b6

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
guillaume-gricourt: Ok, I will update this schema for the other utilities. a6c20a5b3

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
guillaume-gricourt: Ok, it's modified 059ab9d18

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
guillaume-gricourt: Yes, updated into a6c20a5b3

services/galaxytoolshed/galaxy-toolshed-downloads.service.js
guillaume-gricourt: Updated into 238d9607c

calebcartwright:
Puzzled by the CI failure so changing the title to test some things

calebcartwright:
CI should be fixed. Can you rebase your branch?

services/galaxytoolshed/galaxytoolshed-downloads.service.js
```diff
@@ -0,0 +1,44 @@
...
+import BaseGalaxyToolshedService from './galaxytoolshed-base.js'
+
+const repositoryRevisionInstallInfoSchema = Joi.array().items(
+  Joi.object({
+    times_downloaded: Joi.number().required(),
+  }).required(),
+  Joi.object({}),
+  Joi.object({})
+)
+
```
calebcartwright, line 12: This is no longer needed

```suggestion
```

services/galaxytoolshed/galaxytoolshed-base.js
```diff
@@ -0,0 +1,53 @@
...
+import { NotFound, BaseJsonService } from '../index.js'
+
+const orderedInstallableRevisionsSchema = Joi.array()
+  .items(Joi.string())
+  .required()
+
+const repositoryRevisionInstallInfoSchema = Joi.array()
+  .ordered(
+    Joi.object({
+      times_downloaded: Joi.number().required(),
```
calebcartwright, line 11: Just want to both confirm that this field will always be present, even when the download count is zero?

Also, we typically wouldn't require this to be changed, but since there's one last round of cleanup needed in the downloads class, this could be changed to utilized our `nonNegativeInteger` helper from the `validators` module (located in the `services` directory)

calebcartwright (commented):
Couple minor cleanup items, but happy to approve and merge after they are taken care of!

services/galaxytoolshed/galaxytoolshed-downloads.service.js
```diff
@@ -0,0 +1,44 @@
...
+  static defaultBadgeData = {
+    label: 'downloads',
+  }
+
+  async handle({ repository, owner }) {
+    const response = await this.fetchLastOrderedInstallableRevisionsSchema({
+      repository,
+      owner,
+      schema: repositoryRevisionInstallInfoSchema,
```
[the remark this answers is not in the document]
guillaume-gricourt: Remove into 03231b2

services/galaxytoolshed/galaxytoolshed-downloads.service.js
guillaume-gricourt: Applied into 03231b2

services/galaxytoolshed/galaxytoolshed-base.js
guillaume-gricourt: Right, the schema is updated with `nonNegativeInteger` 480cc22

calebcartwright:
It's going to be tricky getting this merged because it's coming from an Organization-owned fork. We have a branch protection policy in place that requires the source branch be up to date with the target before PRs can be merged.

We utilize bots and automation which make this seamless for most contributors, but GitHub doesn't allow PRs from Org forks to enable write access to the upstream maintainers (and their bots). This means we'll likely have to try to coordinate a squash/merge update and approval to happen at the same time, as any time any other PR is merged in this repo it'll reset yours back to an unmergable, out-of-date state

guillaume-gricourt:
If I open a new PR, the same as this PR, with a fork coming from my personal account, can be helpful ?

calebcartwright:
If you can go ahead and update this one shortly then we can probably get it resolved. For future PRs though yes it would definitely be worth considering submitting them from a personal fork

calebcartwright (approved):

guillaume-gricourt:
Thank you so much !

PR #8135, opened 2022-06-28
ci: support new circle PR url variable

### LINKED ISSUE #8134: Service test jobs failing due to issue in PR title check
Not sure what the root cause is, but have seen this with multiple PRs including #8114 and follow up test in #8133

calebcartwright:
Fixes #8134

### DIFF
--- a/core/service-test-runner/infer-pull-request.js
+++ b/core/service-test-runner/infer-pull-request.js
@@ -59,7 +59,9 @@ function _inferPullRequestFromTravisEnv(env) {
 }
 
 function _inferPullRequestFromCircleEnv(env) {
-  return parseGithubPullRequestUrl(env.CI_PULL_REQUEST)
+  return parseGithubPullRequestUrl(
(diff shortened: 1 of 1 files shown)


calebcartwright:
We were getting an error about the url being `undefined` which led me to consider the environment variable. I don't see the one we were previously using, `CI_PULL_REQUEST`, anywhere in https://circleci.com/docs/2.0/env-vars#built-in-environment-variables so I suspect Circle have made a semirecent change that dropped that var.

I tried checking through their doc history to see when that might've occurred but their process makes it hard to trace through git blame. I've opted to leave the original in place for now in case this impacts others and they end up reverting the change, while also adding in the new, explicitly mentioned variable that seems to solve the problem.

Can circle back (no pun intended) to consider removing the older one at a later date, but would like to get this fixed asap as the current issue is a blocker for all PRs

chris48s (approved):

PR #8141, opened 2022-06-29
add docs on nouns in route

chris48s:
Writing #8140 made me realise there is a rule for how we decide on the correct route for a badge which we follow (mostly? :grimacing: ) but is not expressed in the documentation.


doc/badge-urls.md
```diff
@@ -3,6 +3,10 @@
 - The format of new badges should be of the form `/SERVICE/NOUN/PARAMETERS?QUERYSTRING` e.g:
   `/github/issues/:user/:repo`. The service is github, the
   badge is for issues, and the parameters are `:user/:repo`.
+- The `NOUN` part of the route is:
+  - singular if there is only one of the thing (e.g: `/license`, `/rating`)
```
calebcartwright, line 7: bit of a nit, but there are packages/projects/etc. that are dual+ licensed and our badges will reflect the same. as such, should we actually prefer the plural `/licenses`, or should we reword this a bit to reflect the typical scenario, e.g.

```suggestion
  - singular if there is exclusively, at least typically, one of the thing (e.g: `/license`, `/rating`)
```

calebcartwright (approved):
probably good to go as-is, but one comment/suggestion inline for consideration

doc/badge-urls.md
chris48s: This is a good point. Looking back at it, license specifically is actually a bad example here because we don't use `/license` we use `/l`, although your point still stands.

Just thinking about replacing this with something like `/build`, there is a category here where there are many of the thing, but we're only telling you about one of them. So there are many **builds**, but we're the badge is reporting the status of the latest **build**. Your package has many **versions** but the thing on the badge is the latest **version**.

I'm not exactly sure how to succinctly explain that.

calebcartwright: Would it be easier if we combined the two singular lines?

```suggestion
  - singular if the badge message represents a single entity, such as the current status of a build (e.g: `/build`), or a more abstract or aggregate representation of the thing (e.g.: `/coverage`, `/quality`)
```

chris48s: seems reasonable :+1:

calebcartwright (approved):

Issue #8143, opened 2022-06-30
Gitlab Badges not working for Private Repositories


CanisHelix:
### Are you experiencing an issue with...

My own instance of shields

### 🐞 Description

On a self-hosted solution (Docker), with the GITLAB_ORIGINS and GITLAB_TOKEN defined as environment variables. Badges for private repositories on Gitlab.com are returning 'project not found'.

At request on Discord, enabling TRACE_SERVICES reveals that the gitlab password is not being picked up.

```
        url: URL {
          href: 'https://gitlab.com/api/v4/projects/<project id>/repository/contributors?page=1&per_page=1',
          origin: 'https://gitlab.com',
          protocol: 'https:',
          username: '',
          password: '',
          host: 'gitlab.com',
          hostname: 'gitlab.com',
          port: '',
          pathname: '/api/v4/projects/<project id>/repository/contributors',
          search: '?page=1&per_page=1',
          searchParams: URLSearchParams { 'page' => '1', 'per_page' => '1' },
          hash: ''
        },
```

Mounting a local.yml into the container also produces the same. So neither solution picks up a password for Gitlab.com at the moment.

chris48s:
Thanks for moving the conversation over here from Discord.

Working out that the credentials aren't getting set is a step forward.

You said you are running this using docker. Are you running one of the tagged releases https://registry.hub.docker.com/r/shieldsio/shields/tags?page=1&ordering=last_updated if so which one? If you're just pulling from the next tag or building your image locally are you able to give us the SHA? I'd like to try and reduce this to a minimal repro.

calebcartwright:
Echo the sentiment about shifting to an issue, thanks!

Think we may want to change the title to reduce the scope.

I've confirmed that the auth is being sent for all other GitLab badges (license, tags, releases) but isn't for the Contributors badge and that's due to the approach taken with that particular badge being different from the others in terms of how the request is made and how the request is processed.

Note that it's not utilizing the `fetch` function defined in the GitLab base class which is what wires up the respective GitLab auth

https://github.com/badges/shields/blob/0ceb558bb4262b22f9e089d5935c3f926a4f28c8/services/gitlab/gitlab-contributors.service.js#L61-L70

[...]

PR #8145, opened 2022-07-01
fix [gitlab] auth

calebcartwright:
Fixes the overwhelming majority, but not totality, of #8143.

I also added a dummy license to our private test repo under the shields-ops-group

### DIFF
--- a/services/gitlab/gitlab-base.js
+++ b/services/gitlab/gitlab-base.js
@@ -8,7 +8,7 @@ export default class GitLabBase extends BaseJsonService {
 
   async fetch({ url, options, schema, errorMessages }) {
     return this._requestJson(
-      this.authHelper.withBasicAuth({
+      this.authHelper.withBearerAuthHeader({
         schema,
         url,
         options,
--- a/services/gitlab/gitlab-license.tester.js
+++ b/services/gitlab/gitlab-license.tester.js
@@ -1,6 +1,9 @@
 import { licenseToColor } from '../licenses.js'
 import { createServiceTester } from '../tester.js'
+import { noToken } from '../test-helpers.js'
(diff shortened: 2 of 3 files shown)


calebcartwright:
Forgot to explicitly note, but the outstanding piece is the original circumstance that identified this issue. I provided more detail in https://github.com/badges/shields/issues/8143#issuecomment-1171759388 but in short the Contributors badge doesn't utilize auth at all and will need a separate adjustment.

cc @sunny0826 in case you'd like to work on getting the Contributors badge adjusted to utilize the auth

chris48s (approved):

chris48s:
Just to confirm, passing the token in the password field using basic auth just flat-out doesn't work anymore - right?

calebcartwright:
> Just to confirm, passing the token in the password field using basic auth just flat-out doesn't work anymore - right?

Correct. I know I tested this back when I first added it so can only assume it worked at the time, but perhaps that was just legacy behavior GL had been planning on dropping for a while and that finally happened at some point over the last year.

Or perhaps my brain just mistakenly interpreted 404s as 200s :man_shrugging:

sunny0826:
> Just to confirm, passing the token in the password field using basic auth just flat-out doesn't work anymore - right?

GitLab supports  [multiple authentication methods](https://docs.gitlab.com/ee/api/index.html#authentication), but not `user: "", pass: ""`, which was probably deprecated before I joined GItLab... I even I didn't even find it in the documentation for the old version

calebcartwright:
To be honest I'm not sure the past really matters all that much at this point unless anyone has an operable time machine and use cases that predate the one surfaced this week.

Whether or not basic auth with a pat only ever worked with git doesn't really make a difference going forward; using the pat in the oauth header does, for now, and that's what we're now using

PR #8164, opened 2022-07-04
Add [galaxytoolshed] Activity

guillaume-gricourt:
Hi,
This PR adds a feature to the `galaxytoolshed` service. Referring to #7660.
I am at your disposal for correction,
Many thanks,


guillaume-gricourt:
> Does this represent the date the item was first created/published/released, or is it the date of the most recent update/release?

Yes, I agree. It's represents the created date of the repository in the galaxy toolshed.

services/galaxytoolshed/galaxytoolshed-activity.service.js
```diff
@@ -0,0 +1,41 @@
...
+      }),
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'created date',
+  }
+
+  static render({ date }) {
+    return { message: formatDate(date), color: age(date) }
```
calebcartwright, line 30: I think for the color we should just `blue` since this is an informational point. The `age` color helper applies a color scale that's useful for updates because it can reflect activity (i.e. this package is regularly updated/maintained vs. it hasn't been updated in 10 years).

In this case though, we're just displaying the static date when the package was first created so using this color scale would essentially penalize older/more established packages with a red color.

Let me know if I've misunderstood, but if the above is correct then I'd suggest we do this instead, along with an extra line up above in defaultBadgeData

```suggestion
    return { message: formatDate(date) }
```

services/galaxytoolshed/galaxytoolshed-activity.service.js
```diff
@@ -0,0 +1,41 @@
...
+        owner: 'iuc',
+      },
+      staticPreview: this.render({
+        date: this.render({ date: '2022-01-01' }),
+      }),
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'created date',
```
calebcartwright, line 26:
```suggestion
    label: 'created date',
    color: 'blue',
```

services/galaxytoolshed/galaxytoolshed-activity.service.js
```diff
@@ -0,0 +1,41 @@
...
+export default class GalaxyToolshedCreatedDate extends BaseGalaxyToolshedService {
+  static category = 'activity'
+  static route = {
+    base: 'galaxytoolshed/created-date',
+    pattern: ':repository/:owner',
+  }
+
+  static examples = [
+    {
+      title: 'Galaxy Toolshed (created time)',
```
calebcartwright, line 14: Apologies bit of a nit but since we're using 'date' everywhere else let's do the same here:

```suggestion
      title: 'Galaxy Toolshed (created date)',
```

calebcartwright (commented):
Looks good! Couple small things then will be ready to merge

services/galaxytoolshed/galaxytoolshed-activity.service.js
guillaume-gricourt: Sorry for the delay,
It's true, the modifications are done in 353338867

services/galaxytoolshed/galaxytoolshed-activity.service.js
guillaume-gricourt: Done in 353338867

services/galaxytoolshed/galaxytoolshed-activity.service.js
guillaume-gricourt: Done in 353338867

calebcartwright (approved):
thanks!

PR #8166, opened 2022-07-05
feat: add [gitlabmergerequests] service

sunny0826:
re #8077


calebcartwright:
Have only done a quick pass through this but a couple early notes:

- given the 10k ceiling, that seems like a case we should handle explicitly in our rendering, even if it requires a special case handling
- consider if/how the handle/fetch/transform function paradigm could be applied with this class

calebcartwright:
@sunny0826 just want to make sure you saw this :point_right: https://github.com/badges/shields/pull/8166#issuecomment-1175232531

calebcartwright:
> > @sunny0826 just want to make sure you saw this point_right [#8166 (comment)](https://github.com/badges/shields/pull/8166#issuecomment-1175232531)
> 
> I saw this, but wasn't sure what I needed to do.

Ah okay, no worries! In the future don't hesitate to ask questions in these types of cases, especially if something is unclear. I'll follow up tomorrow with some elaboration on my prior comment so we can figure out next steps

sunny0826:
> After looking through a little more closely, I see that you are already applying special case handling to the >10k scenario which is good 👍 (I was just emphasizing that we should in my prior comment).
> 
> The second part of my comment was really just related to the length of some of those methods, and a suggestion to consider using the `transform` function to encapsulate some concerns and hopefully shorten some other functions. However, I'm not sure how much that will really buy given the nature of what's going on in those two other functions. I'll try to leave some inline feedback over the weekend as I think there's some other changes that can make those a little shorter, and we can reconsider if a separate `transform` function is still worthwhile afterwards.
> 
> Finally, thanks again for all your work on these GitLab badges, it's very much appreciated!

Thank you for your feedback.

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -0,0 +1,359 @@
+import Joi from 'joi'
+import { optionalUrl, nonNegativeInteger } from '../validators.js'
+import { metric } from '../text-formatters.js'
+import GitLabBase from './gitlab-base.js'
+
+// check schema for the response of the GitLab API
```
calebcartwright, line 6: I think it would be better to use the more detailed comment from down below up here:

```suggestion
// The total number of MR is in the `x-total` field in the headers.
// https://docs.gitlab.com/ee/api/index.html#other-pagination-headers
```

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -0,0 +1,359 @@
...
+    }
+    const message = `${mergeRequestCount > 10000 ? 'more than ' : ''}${metric(
+      mergeRequestCount
+    )}${messageSuffix ? ' ' : ''}${messageSuffix}`
+    return {
+      label: `${labelPrefix}${labelText}merge requests`,
+      message,
+      color: mergeRequestCount > 0 ? 'brightgreen' : 'yellow',
+    }
+  }
```
calebcartwright, line 295: This function is fairly complex, largely because it contains a number of inline conditionals that leverage the ternary operator. I think it might help improve readability to break those up a bit, even if it does make things a tad more verbose.

Additionally, I don't think we should be applying custom colorization here. This badge strikes me as a purely informational data point and as such I feel like our `blue`/`informational` color would be most appropriate. I'm aware that the analog badges for GitHub utilize the yellow coloring, but as I think we may have discussed before I think those badges are colored that way for reasons that aren't necessarily applicable for GitLab.

Finally, given the breadth of values this function could produce, I think it would be best to do some unit testing of this functions to cover all those values and paths. Let me know if you're unfamiliar with how we do unit tests (as opposed to the service tests), and I'll be happy to share some additional info!

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -0,0 +1,359 @@
...
+      labels,
+    })
+    return this.constructor.render({
+      variant,
+      raw,
+      labels,
+      mergeRequestCount,
+    })
+  }
+}
```
calebcartwright, line 359: Somewhat conversely to the `render` function, I feel like the `fetch` function is unnecessarily verbose and could be cleaned up a bit. Also note for future reference that there's no need to check for truthiness for url query params, you can just toss the value along and the http handling code farther up the pipe will handle that automatically.

I started out with a suggested change to just `fetch` but ended up wanting to suggest a broader refactoring that I think reads a bit cleaner. It may seem like a large diff, but it's basically just small tweaks/clean up on the `fetch` function, pulling out a separate `render` function, and not wrapping results in an object and forcing destructuring at the call sites.

```suggestion
  static transform(data) {
    if (data['x-total'] !== undefined) {
      return data['x-total']
    } else {
      // https://docs.gitlab.com/ee/api/index.html#pagination-response-headers
      // For performance reasons, if a query returns more than 10,000 records, GitLab doesn’t return `x-total` header.
      // Displayed on the page as "more than 10k".
      return 10001
    }
  }

  async fetch({ project, baseUrl, variant, labels }) {
    // https://docs.gitlab.com/ee/api/merge_requests.html#list-project-merge-requests
    const { res } = await this._request(
      this.authHelper.withBearerAuthHeader({
        url: `${baseUrl}/api/v4/projects/${encodeURIComponent(
          project
        )}/merge_requests`,
        options: {
          searchParams: {
            state: variant == 'open' ? 'opened' : variant,
            page: '1',
            per_page: '1',
            labels,
          },
        },
        errorMessages: {
          404: 'project not found',
        },
      })
    )
    return this.constructor._validate(res.headers, schema)
  }

  async handle(
    { variant, raw, project },
    { gitlab_url: baseUrl = 'https://gitlab.com', labels }
  ) {
    const data = await this.fetch({
      project,
      baseUrl,
      variant,
      labels,
    })
    const mergeRequestCount = this.constructor.transform(data)
    return this.constructor.render({
      variant,
      raw,
      labels,
      mergeRequestCount,
    })
  }
}
```

services/gitlab/gitlab-merge-requests.tester.js
```diff
@@ -0,0 +1,169 @@
...
+    message: isMetric,
+  })
+
+t.create('Opened merge requests by Scoped labels')
+  .get('/open/gitlab-org%2Fgitlab.json?labels=test,failure::new')
+  .expectBadge({
+    label: 'test,failure::new merge requests',
+    message: Joi.string().regex(
+      /^([0-9]+[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) open$/
+    ),
```
calebcartwright, line 60: I know the potential for these live tests to get a "0" result makes the validator tricky, but I personally prefer to use something like this:

```suggestion
    message: Joi.alternatives(isMetricOpenIssues, Joi.equal('0 open')),
```

services/gitlab/gitlab-merge-requests.tester.js
```diff
@@ -0,0 +1,169 @@
...
+/**
+ *  Closed issue number case
+ */
+t.create('Closed merge requests')
+  .get('/closed/guoxudong.io/shields-test/issue-test.json')
+  .expectBadge({
+    label: 'merge requests',
+    message: Joi.string().regex(
+      /^([0-9]+[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) closed$/
+    ),
```
calebcartwright, line 72: Think we should be able to use the `isMetricClosedIssues` validator here:

```suggestion
    message: isMetricClosedIssues,
```

services/gitlab/gitlab-merge-requests.tester.js
```diff
@@ -0,0 +1,169 @@
...
+    message: isMetric,
+  })
+
+t.create('Closed merge requests by label is > zero')
+  .get('/closed/guoxudong.io/shields-test/issue-test.json?labels=bug')
+  .expectBadge({
+    label: 'bug merge requests',
+    message: Joi.string().regex(
+      /^([0-9]+[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) closed$/
+    ),
```
calebcartwright, line 88: Same comment as above, and if necessary we can use the alternatives format:

```suggestion
    message: Joi.alternatives(isMetricClosedIssues, Joi.equal('0 closed')),
```

services/gitlab/gitlab-merge-requests.tester.js
```diff
@@ -0,0 +1,169 @@
...
+  })
+
+t.create('Closed merge requests by  multi-word label is > zero')
+  .get('/closed/guoxudong.io/shields-test/issue-test.json?labels=bug,critical')
+  .expectBadge({
+    label: 'bug,critical merge requests',
+    message: Joi.string().regex(
+      /^([0-9]+[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) closed$/
+    ),
```
calebcartwright, line 97: same here

calebcartwright (commented):
Thanks and apologies for the delay! I think the core of this is in place, just have a few suggestions/requested changes detailed below

services/gitlab/gitlab-merge-requests.service.js
sunny0826: Indeed, I've written a lot of unit tests for Go or Python, but never tried to write unit tests for JS functions, thanks for your help.

calebcartwright: Absolutely! I've found Go unit tests to be a bit odd compared to other languages, but I think you'll find the JS tests will feel fairly familiar to Python (albeit with some different conventions and apis).

For a pretty basic example I'd suggest taking a look at the unit tests for the codecov service:

https://github.com/badges/shields/blob/fa3839cdab3a7fc7b15c1cd35b4da175d4e28b07/services/codecov/codecov.spec.js#L1-L23

One difference with JS is that the actual "test" functions are typically implemented as callback functions that are passed as a call arg to the hooks/functions exported by the framework (as opposed to the free function defs, or test classes you'd define for tests in python), and inside that callback is where you'll see the typical arrange/act/assert paradigm.

For our unit tests we leverage a tool called [sazerac](https://github.com/sazeracjs/sazerac) which takes care of a lot of the boilerplate code for writing tests and provides a data-driven type of interface. With this we use a `describe` block (this just helps organize tests and makes it easier to run these subset of tests) and then we define a number of `test` calls for each respective function we want to test on the service class, almost always for a `render` and/or `transform` type of function.

The first arg to the `test` call should be the service class function, and the second is the callback that contains the test. The body of that callback will contain one or more `given(...)` calls, where the arg passed to `given` is the arg that will be passed to the service class function as part of the test. The chained `.expect()` call is where you'd specify the expected return value from calling that service class function with the arg(s) you specified for the associated `given` call.

I'm not sure if this will be applicable here, but in cases where passing multiple different inputs to the function under test should produce the same result, you can use the `forCases` function which just avoids the need to have multiple `.expect()` calls.

You can run these tests with `npm run test:core -- --grep=FooBar` where FooBar would be whatever value you name the suite in the `describe(...)` call, e.g. `npm run test:core -- --grep=Codecov` (note this is case sensitive, I forget if there's a case insensitive filter)

sunny0826: awesome!

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -290,20 +291,11 @@ export default class GitlabMergeRequests extends GitLabBase {
     return {
       label: `${labelPrefix}${labelText}merge requests`,
       message,
-      color: mergeRequestCount > 0 ? 'brightgreen' : 'yellow',
+      color: mergeRequestCount > 0 ? 'brightgreen' : 'blue',
```
calebcartwright, line 294: Sorry if I was unclear previously, but what I was saying is that the color should _always_ be blue and not having a conditional color.

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -0,0 +1,350 @@
...
+const queryParamSchema = Joi.object({
+  labels: Joi.string(),
+  gitlab_url: optionalUrl,
+}).required()
+
+const documentation = `
+<p>
+  You may use your GitLab Project Id (e.g. 278964) or your Project Path (e.g. gitlab-org/gitlab ).
+  Note that only internet-accessible GitLab instances are supported, for example https://jihulab.com, https://gitlab.gnome.org, or https://gitlab.com/.
+</p>
```
calebcartwright, line 22: Just a note for a potential follow up PR, but we may want to encapsulate this doc text in one place and then reuse it, similar to what we've done with the GitHub class tree (it seems like we've got duplicate copies of this on multiple GitLab classes nowadays)

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -0,0 +1,350 @@
...
+`
```
calebcartwright, line 23: Might also be worth mentioning the >=10k behavior here since that may not be obvious to users the badge message will be capped at 10k because the GitLab API caps the count at 10k

services/gitlab/gitlab-merge-requests.spec.js
```diff
@@ -0,0 +1,36 @@
...
+      label: 'merge requests',
+      message: '0 open',
+      color: 'blue',
+    })
+    given({ variant: 'open', mergeRequestCount: 10001 }).expect({
+      label: 'merge requests',
+      message: 'more than 10k open',
+      color: 'brightgreen',
+    })
+  })
```
calebcartwright, line 35: Thanks for these! Think it would be good to add one that has a raw variant and labels too

services/gitlab/gitlab-merge-requests.spec.js
```diff
@@ -0,0 +1,36 @@
...
+})
```
calebcartwright, line 36: Let's also add a test that validates the authentication integration (especially since we recently found we'd missed a couple classes). I think you should be able to copy in the the test from one of the other GitLab services like Tag

https://github.com/badges/shields/blob/master/services/gitlab/gitlab-tag.spec.js

and just change the import class to be the MergeRequest service class instead

services/gitlab/gitlab-merge-requests.tester.js
```diff
@@ -0,0 +1,165 @@
...
+  })
+
+t.create('locked merge requests')
+  .get('/locked/gitlab-org%2Fgitlab.json')
+  .expectBadge({
+    label: 'merge requests',
+    message: Joi.string().regex(
+      /^([0-9]+[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) locked$/
+    ),
+  })
```
calebcartwright, line 165: Could you also add a test that has a custom `gitlab_url` set?

calebcartwright (commented):
Excellent, thank you for the updates. Few additional asks but think we're on the home stretch.

These issue/PR/MR types of badges can get quite involved so thanks for sticking with it!

services/gitlab/gitlab-merge-requests.tester.js
sunny0826: done

services/gitlab/gitlab-merge-requests.service.js
sunny0826: TODO, note at https://github.com/badges/shields/issues/8077#issuecomment-1207597218

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -19,6 +19,7 @@ const documentation = `
 <p>
   You may use your GitLab Project Id (e.g. 278964) or your Project Path (e.g. gitlab-org/gitlab ).
   Note that only internet-accessible GitLab instances are supported, for example https://jihulab.com, https://gitlab.gnome.org, or https://gitlab.com/.
+  This badge message will be capped at 10k because the <a href="https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers">GitLab API caps the count at 10k</a>.
```
calebcartwright, line 22: What about something like this instead:

```suggestion
  <a href="https://docs.gitlab.com/ee/user/gitlab_com/index.html#pagination-response-headers">GitLab's API </a> only reports up to 10k Merge Requests, so badges for projects that have more than 10k will not have an exact count.
```

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -291,7 +292,6 @@ export default class GitlabMergeRequests extends GitLabBase {
     return {
       label: `${labelPrefix}${labelText}merge requests`,
       message,
-      color: mergeRequestCount > 0 ? 'brightgreen' : 'blue',
```
calebcartwright: We should either set the color to blue here, or we can set the color up in `defaultBadgeData`

services/gitlab/gitlab-merge-requests.spec.js
```diff
@@ -1,17 +1,19 @@
 import { test, given } from 'sazerac'
+import nock from 'nock'
+import { expect } from 'chai'
+import { cleanUpNockAfterEach, defaultContext } from '../test-helpers.js'
 import GitlabMergeRequests from './gitlab-merge-requests.service.js'
+import GitLabTag from './gitlab-tag.service.js'
```
calebcartwright, line 6: We shouldn't need/be using this here because this file contains unit tests for the MergeRequests class

services/gitlab/gitlab-merge-requests.spec.js
```diff
@@ -20,17 +22,63 @@ describe('GitlabMergeRequests', function () {
...
+        },
+      },
+      private: {
+        gitlab_token: fakeToken,
+      },
+    }
+
+    it('sends the auth information as configured', async function () {
+      const scope = nock('https://gitlab.com/')
+        .get('/api/v4/projects/foo%2Fbar/repository/tags?order_by=updated')
```
calebcartwright, line 63: This should be updated to reflect the internal API call we make to get the merge request count

services/gitlab/gitlab-merge-requests.spec.js
```diff
@@ -20,17 +22,63 @@ describe('GitlabMergeRequests', function () {
...
+        // This ensures that the expected credentials are actually being sent with the HTTP request.
+        // Without this the request wouldn't match and the test would fail.
+        .matchHeader('Authorization', `Bearer ${fakeToken}`)
+        .reply(200, [{ name: '1.9' }])
```
calebcartwright, line 67: This should be updated so that this mock response includes the headers we look for to get the merge request count, e.g.:

```suggestion
        .reply(200, {}, { 'Header-Name': 'Header-Value' })
```

services/gitlab/gitlab-merge-requests.spec.js
```diff
@@ -20,17 +22,63 @@ describe('GitlabMergeRequests', function () {
...
+
+      expect(
+        await GitLabTag.invoke(
+          defaultContext,
+          config,
+          { project: 'foo/bar' },
+          {}
+        )
```
calebcartwright, line 75: We should be using the merge requests class here:

```suggestion
        await GitlabMergeRequests.invoke(
          defaultContext,
          config,
          { project: 'foo/bar' },
          {}
        )
```

services/gitlab/gitlab-merge-requests.spec.js
```diff
@@ -20,17 +22,63 @@ describe('GitlabMergeRequests', function () {
...
+      ).to.deep.equal({
+        message: 'v1.9',
```
calebcartwright, line 77: And finally this would be updated with whatever the expected merge request count would be based on the mocked response you define above

calebcartwright (commented):
Thanks for the updates and apologies for the delay in review. Few final items then think we'll be set

services/gitlab/gitlab-merge-requests.spec.js
sunny0826: ops

services/gitlab/gitlab-merge-requests.spec.js
sunny0826: I'm very sorry about that. I simply copied the code and didn't look at the code specifically. Thanks for your time.

services/gitlab/gitlab-merge-requests.service.js
```diff
@@ -0,0 +1,351 @@
...
+      title: 'GitLab merge requests',
+      pattern: 'open/:project+',
+      namedParams: {
+        project: 'gitlab-org/gitlab',
+      },
+      queryParams: { gitlab_url: 'https://gitlab.com' },
+      staticPreview: {
+        label: 'merge requests',
+        message: '1.4k open',
+        color: 'brightgreen',
```
calebcartwright, line 52: Sorry to do this again, but I missed this. Can you change all the examples to, preferably, use the `render` function, or at least ensure that the color in the examples is set to blue too?

calebcartwright (commented):
Sorry I missed one final thing, noted inline below

services/gitlab/gitlab-merge-requests.service.js
sunny0826: ops...It was an oversight on my part.

calebcartwright (approved):

PR #8167, opened 2022-07-05
fix npm tagged version examples, run [npmversion]

calebcartwright:
Noticed while browsing the badge listings that a few of the npm version badges had a literal 'undefined' in the label:

image

After poking around it seems this occurs because the examples are using the classes render function, which grabs the label value from the default badge data when rendering a tagged version badge. However, `defaultBadgeData` is still an empty object when the render function is invoked for the examples. Not 100% sure why this is the case, but it seems to be dependent on the order in which the class members are defined. Moving the `defaultBadgeData` property above the `examples` property also resolved the issue, but that violates the member order we have specified in the lint config (side note, I wonder if it would be worth revisiting this ordering to move defaults above examples?)

### DIFF
--- a/services/npm/npm-version.service.js
+++ b/services/npm/npm-version.service.js
@@ -4,6 +4,7 @@ import { NotFound } from '../index.js'
 import NpmBase from './npm-base.js'
 
 const keywords = ['node']
+const defaultLabel = 'npm'
 
 // Joi.string should be a semver.
 const schema = Joi.object()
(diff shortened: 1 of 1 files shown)


chris48s:
Would we be able to still use `this.render()` if we moved the declaration but used an `eslint-disable` comment e.g:

https://github.com/badges/shields/blob/90694594352034b8c56c7ececd2dcf708080c1d3/services/text-formatters.spec.js#L41-L42

If so, I think that would probably be preferable.

---

More generally, if this ordering is leading to bugs in some cases (albeit limited), I think we probably should change the ordering in https://github.com/badges/shields/blob/7ea9e01194f672130808ab88bc0bb770a36a3ff5/.eslintrc.yml#L83-L105
It looks like the order we've got there was mostly to aid readability ( https://github.com/badges/shields/pull/3353 ). We would need to do a commit that touches basically every `.service.js` file to do it but I _think_ ESLint should be able to autofix it for us if we go ahead with it.

calebcartwright:
Agreed, and I'd prefer to update the ordering we're enforcing. There's a number of active PRs open at the moment which are adding new service classes, so to avoid pushing churn on those, I'd suggest we abandon this for now and once those (or most) are through I'll open one that adjusts that ordering across all classes which will fix the original item from here as well.

PR #8169, opened 2022-07-06
Add [ROS] version service

jtbandes:
Adds version service at `/ros/v/:distro/:packageName` for [ROS packages](https://index.ros.org/packages/). Version information is loaded directly from the [ros/rosdistro repo](https://github.com/ros/rosdistro) on GitHub, using tags to determine the latest available distribution, and then parsing the distribution.yaml file from that distribution.

I was hoping to make the `:distro` show up as the badge `label` in error cases, but it seems like when an error is thrown, only the defaultBadgeLabel (or `'version'`) is used as the label. And I would've happily used the `raw.githubusercontent.com` method of loading the distribution.yaml file, but it didn't seem like one service would be able to make both GraphQL requests and REST requests (when I tried to make a REST request, it still prepended the GraphQL endpoint's base URL). Any advice on these would be appreciated!


calebcartwright:
Thanks for opening a PR!

To the best of my knowledge, we don't have any open issues requesting this badge with the relevant details, so before we proceed it would be helpful for us if you could speak to some of the questions we ask in the issue template for new badges so that we've a sufficient amount of context (I've not heard of ROS myself)

I think the first several are implicitly covered already, so I've taken a pass at filling them in, but certainly feel free to correct. 

- **Which service is this badge for e.g:** ROS (but more info/context about the nature of this service would be most appreciated)
- **What sort of information should this badge show?** standard version badge
- **Is there a public API?** No. The data is retrieved through a set of sequential API calls to GitHub to retrieve a yaml file which is then parsed and processed
- **Does the API requires an API key?** Not in the traditional sense
- **Link to the API documentation.** (anything that helps codify the schema or would otherwise be relevant)
- Who are the users of the service/target audience for the badge?

jtbandes:
- **Which service is this badge for:** [ROS](https://ros.org) is an ecosystem of robotics software and tools, organized into [packages](https://wiki.ros.org/Packages). It has wide adoption across both education and industry. See [ROS community metrics](http://download.ros.org/downloads/metrics/metrics-report-2021-07.pdf), [ROSCon](https://roscon.ros.org/2022/). There are two major versions of ROS (ROS 1 and ROS 2) which are actively maintained (the last distro of ROS 1 will be EOL in 2025).
- **Link to the API documentation.** Some info about the rosdistro repo contents/structure is here:
  - https://github.com/ros-infrastructure/rosindex/blob/aedbdddaec60a66eb5be100909c14d941fff8027/about/design.md#official-index
  - https://ros.org/reps/rep-0141.html
- **What is the specific use case? Who are the users of the service/target audience for the badge?** I'm imagining that developers of ROS packages would use the version badges in their repo READMEs just like npm package developers do with npm version badges today. The version service is segmented by distro, because multiple ros distros are being maintained at the same time ([ROS 1](https://wiki.ros.org/Distributions), [ROS 2](https://docs.ros.org/en/rolling/Releases.html)), and a different package version can be released in each distro. The service supports both ROS 1 & 2 because all of the distribution info is stored in the same repository.

calebcartwright:
> Hi @calebcartwright, just to follow up on this, is there any more info I can provide?

No, the info you've provided is great thanks! Someone should get around to reviewing this in the not too distant future

services/ros/ros-version-helpers.js
```diff
@@ -0,0 +1,29 @@
...
+    })
+  }
+
+  if (!packageInfo) {
+    throw new NotFound({ prettyMessage: `package not found: ${packageName}` })
+  }
+
+  const version = packageInfo.release?.version
+  if (typeof version !== 'string') {
+    throw new NotFound({
```
chris48s, line 22: Instead of manually checking for the existance and type of properties here, I'd prefer us to define another Joi schema here and validate the JS object we got from parsing the yaml against the Joi schema. If you define this method on the `RosVersion` class, you can use the `static _validate()` method from `BaseService` to do the validation e.g:



```js
const schema = //TODO Joi...

try {
  parsed = yaml.load(buffer.toString(), encoding)
} catch (err) {
  throw new InvalidResponse({
    prettyMessage: 'unparseable yaml response',
    underlyingError: err,
  })
}
return this.constructor._validate(parsed, schema)
```

services/ros/ros-version.service.js
```diff
@@ -0,0 +1,111 @@
...
+        }
+      `,
+      variables: { refPrefix: `refs/tags/${distro}/` },
+      schema: tagsSchema,
+    })
+
+    // Filter for tags that look like dates: humble/2022-06-10
+    const tags = tagsJson.data.repository.refs.edges
+      .map(edge => edge.node.name)
+      .filter(tag => /^\d+-\d+-\d+$/.test(tag))
```
chris48s, line 76: Is what we actually ideally want here

- the tag corresponding to the most recent commit date (where the tag name looks like a date), or
- the tag with the most recent date in the tag name.

In general, I'd hope those will be the same, but making the assumption we want to code defensively here.

services/ros/ros-version.service.js
```diff
@@ -0,0 +1,111 @@
...
+
+    const ref = tags[0] ? `refs/tags/${distro}/${tags[0]}` : 'refs/heads/master'
+    const prettyRef = tags[0] ? `${distro}/${tags[0]}` : 'master'
```
chris48s, line 79: Question: If the user specified a particular distro and we can't find any corresponding tag, does it actually make sense to fall back to `master` here, or should we really just throw an exception saying we can't find any tags matching the requested distro here?

services/ros/ros-version.service.js
```diff
@@ -0,0 +1,111 @@
...
+      throw new NotFound({
+        prettyMessage: `distribution.yaml not found: ${distro}@${prettyRef}`,
+      })
+    }
+    const version = parseReleaseVersionFromDistro(
+      contentJson.data.repository.object.text,
+      packageName
+    )
+
+    return { ...renderVersionBadge({ version }), label: distro }
```
chris48s, line 109: For version badges, the label is usually the name of the package registry, as opposed to the package.

I think we do want to include ros in the label. With the conda badges, we include the channel in the label separated by a pipe. How about we do similar here - `ros|distro` e.g:

chris48s (commented):
Thanks for submitting this and including tests. It is looking in pretty good shape to start with and its clear you've worked through the documentation and contributing guidelines :+1:

chris48s:
> I was hoping to make the `:distro` show up as the badge `label` in error cases, but it seems like when an error is thrown, only the defaultBadgeLabel (or `'version'`) is used as the label

This is normal - I wouldn't worry about it

> I would've happily used the raw.githubusercontent.com method of loading the distribution.yaml file, but it didn't seem like one service would be able to make both GraphQL requests and REST requests

We could find a way to do this, but tbh I think the approach you've already used of doing both queries using GraphQL is fine and maybe the neatest solution anyway. I don't think we need to change it

services/ros/ros-version.service.js
jtbandes: It seems like the tagging scheme is relatively new, and only the currently active (not end-of-life) distributions have tags for releases. Older ones (e.g. `eloquent`) that have not been published in the recent past don't have tags, but the master branch still contains all the version info of packages from the old distros (the distro is also part of the file path that we read). I guess it's less likely that people will care to display versions for old distros, but it seemed like a reasonable fallback to me if there is not a current version tag available

services/ros/ros-version.service.js
jtbandes: Ok, makes sense. What do you think about including spaces around the pipe? Like this:

chris48s: Yeah agreed :+1:  Change it on this PR and I will do another PR to update the conda one.

jtbandes: done!

PR #8189, opened 2022-07-12
add spaces round pipe in [conda] version badge

chris48s:
As discussed in https://github.com/badges/shields/pull/8169#discussion_r917443156 - this is a bit clearer


calebcartwright (approved):
Technically a change, but completely agreed with both the reasoning and the aesthetic improvement of the end result (plus this makes the label more consistent with how we handle this elsewhere, albeit in the message of other badges)

PR #8192, opened 2022-07-14
Replace moment with dayjs [github node sourceforge wordpress aur eclipse steam vaadin visualstudio]

### LINKED ISSUE #5573: Moment.js Maintenance Mode Announcement

DenverCoder1:
Fixes #5573

`moment` has been mostly deprecated for a couple years now. [`dayjs`](https://github.com/iamkun/dayjs) is a lighter and faster, actively maintained alternative.

`dayjs` has nearly the same API as `moment` so it is very simple to migrate.

The text and color formatters have been tested and are basically identical to moment after importing the necessary dayjs plugins.


calebcartwright (approved):
lgtm, thanks!

PR #8195, opened 2022-07-15
document service integration information

calebcartwright:
As discussed at the last ops meeting. Opening early in draft mode to start the dialog and to pose a few questions

### DIFF
--- a/doc/service-integrations.md
+++ b/doc/service-integrations.md
@@ -1,0 +1,75 @@
+# Integration with upstream services
+
+## Overview
+
+In a nutshell, the Shields Badge Server handles the responsibilities of accepting requests for badges, and then serving those badges back to users.
+
+A grossly oversimplified visualization would probably look something like this:
+
+```mermaid
+sequenceDiagram
+    actor User
+    participant B as Badge Server
+    participant P as Data Provider
+    User->>+B: I'd like a badge for my project please!
+    B->>+P: Get data for project
+    P-->>-B: Data
+    B->>B: Make badge with data
+    B-->>-User: Here's your badge!
+```
+
+Shields is not a system of record (we're not the package registry, pipeline tool, etc.) so when Shields receives a request for a badge, the badge server will first have to reach out to the system of record in order to get the data points it needs to create your badge.
+
+For example, if you ask Shields for a build status badge for your CircleCI pipeline, then Shields has to reach out to CircleCI to figure out what the status of your pipeline is (CircleCI would be the "Data Provider" actor in the above diagram). Similarly if you want a badge that shows the count of downloads of your npm package, Shields has to reach out to the npm.js registry.
+
+That covers the gist, but the actual story is a bit more involved and complicated than that of course. There's a number of other components along the way, ranging from the browser on your local machine to extra services and actors we have deployed as part of the Shields.io runtime ecosystem which help ensure we can provide a stable and reliable deployment of the badge server-as-a-service. Additionally, badges rendered in GitHub have some additional factors at play that impose some additional constraints, detailed in the next section.
+
+### GitHub Badge Rendering
+
+A common usage pattern for badges is to embed them in your project's README files so relevant information is conveyed to the project's users. This means badges are often utilized and rendered in source control management platforms, like GitHub.
+
+[GitHub utilizes a proxy service, called camo][camo], for handling the images that you see when you browse project pages on github.com, and this is utilized for badges too (both svg and png formats). GitHub does this for a number of reasons, including to anonymize requests and protect your privacy. However, this also requires the upstream images (including badges) to be returned quickly in order for those images to show up on your screen, with a rough ceiling of 3-4 seconds. If the upstream image provider is too slow to respond, then camo will timeout and the image won't be displayed.
+
+This imposes constraints on Shields, as we need to ensure that the badge server completes the entire request/response workflow and returns the final badge within a few seconds.
+
+[camo]: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-anonymized-urls
+
+## Interaction patterns
+
+Because of the time limits around the full badge flow discussed above, the badge server is somewhat constrained in terms of patterns it can employ to fetch data from the upstream provider. When we incorporate new badges we typically need to ensure when we receive a badge request and need to get data to serve that badge request, that we can get that data by making a single call to the upstream provider to get the data points needed for the badge.
+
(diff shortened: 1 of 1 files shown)


doc/service-integrations.md
```diff
@@ -0,0 +1,70 @@
...
+sequenceDiagram
+    actor User
+    participant B as Badge Server
+    participant P as Data Provider
+    User->>+B: I'd like a badge for my project please!
+    B->>+P: Get data for project
+    P-->>-B: Data
+    B->>B: Make badge with data
+    B-->>-User: Here's your badge!
+```
```
calebcartwright, line 19: I've been looking for an excuse to play with this ever since GitHub [announced embedded support](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/), rendered version can be found [here](https://github.com/badges/shields/blob/d24e055ec821cef365e279b1ca2b1c82afb48aec/doc/service-integrations.md#overview).

As is often the case with diagrams as code, they aren't quite as pretty and pixel perfect as human drawn ones would be, but i think something simple and embedded here is worthwhile to have a visual

calebcartwright:
As I started pulling more of my notes together, I felt like there's potentially a few audiences for this material (as well as the rest of the material not yet included):

- an audience of a person/user that's just interested in more info without having to come through the tens of thousands of lines of code
- an audience of an upstream vendor/maintainer that's interested in understanding the how/why/etc. of the touchpoints with their service
- an audience of an upstream vendor/maintainer that just wants something akin to a quick cheat sheet with bullets of what they need to know/do

The different motivations between those groups have me wondering if a single doc is best, or how to best split out the material if not. I'm currently leaning towards at least 2 separate pages, one that contains more of the overview content (largely the first couple sections) which can then be referenced by a second that's more of a succinct version tailored to the vendor/maintainer of upstream providers

chris48s:
I think its fine for this to be one document rather than several with different audiences. I suspect for the most part the value of this is going to be so that we can say "the way this works is documented at [some-markdown-file.md#header]" or whatever in response to questions rather than having to explain the same thing multiple times. Its probably a bit optimistic to expect adding some docs will stop the questions being asked in the first place.

doc/service-integrations.md
chris48s: :art:

doc/service-integrations.md
```diff
@@ -0,0 +1,75 @@
...
+
+The nature of the call Shields makes to upstream data providers is obviously dependent upon the nature of the upstream endpoints. The overwhelming majority of the time this occurs via a GET request, though there are a few occasions where the upstream endpoint requires a POST in order for us to retrieve data, and a few others where we issue a HEAD request because the data points we need for the badge actually reside in the response headers.
+
+There are a couple other exceptions, but as a general pattern we strive to integrate with services via a stateless, single call manner.
+
+### Authentication
+
+Shields typically integrates with upstream data providers anonymously, largely because the data targets we need are anonymously available (open source packages, repositories, pipelines, etc.)
+
+The badge server can be configured to make authenticated requests for certain supported services. This exists so that users who are [self-hosting] their own instance of the badge server can get badges for their private content, and also as part of agreements we've made with certain upstream data providers for the main Shields.io deployment.
```
chris48s, line 49:
```suggestion
The badge server can be configured to make authenticated requests for certain services. This exists so that users who are [self-hosting] their own instance of the badge server can get badges for their private content, and also as part of agreements we've made with certain upstream data providers for the main Shields.io deployment.
```

doc/service-integrations.md
```diff
@@ -0,0 +1,75 @@
...
+
+### GitHub Badge Rendering
+
+A common usage pattern for badges is to embed them in your project's README files so relevant information is conveyed to the project's users. This means badges are often utilized and rendered in source control management platforms, like GitHub.
+
+[GitHub utilizes a proxy service, called camo][camo], for handling the images that you see when you browse project pages on github.com, and this is utilized for badges too (both svg and png formats). GitHub does this for a number of reasons, including to anonymize requests and protect your privacy. However, this also requires the upstream images (including badges) to be returned quickly in order for those images to show up on your screen, with a rough ceiling of 3-4 seconds. If the upstream image provider is too slow to respond, then camo will timeout and the image won't be displayed.
```
chris48s, line 31: Worth mentioning here that also some other places badges are commonly viewed do this too. For example PyPI also runs a camo proxy for images and it is one of our biggest sources of traffic

Also it feels like the concept of other people cacheing our badges (with Camo) and us cacheing our badges (with CloudFlare) kind of go together in that they happen in basically the same phase of the "flow" described by the diagram. So maybe it makes sense to talk about those things in the same section or adjacent sections

doc/service-integrations.md
chris48s: Also we sometimes use credentials for services that just flat-out require auth (I think Symfony Insights and Twitch might be example of this). Sometimes we allow users to pass their own token in a querystring too. Maybe linking out to https://github.com/badges/shields/blob/master/doc/server-secrets.md for more details would be good here.

doc/service-integrations.md
```diff
@@ -0,0 +1,75 @@
...
+
+Many upstream data providers employ common techniques to protect the availability and integrity of their service, and one common technique is [rate limiting].
+
+Typically, when clients/consumers of rate limited endpoints will employ client-side techniques to avoid running afoul of those limits and potentially getting blocked or having their requests throttled. Unfortunately, those techniques aren't really viable for the Shields.io environment due to the workflow and constraints discussed above.
+
+As such, we instead try to ensure that Shields.io never makes more calls to an upstream provider than their rate limits allow.
+
+In cases where Shields.io may run close to or exceed those limits, we typically consider:
+
+- increasing the cache periods we set (to reduce the number of badge requests we receive)
```
chris48s, line 63: Lets clarify the cache is downstream rather than upstream here

chris48s (commented):
I've left some minor comments here but this is looking good and I agree with the main points/general thrust of the doc

calebcartwright:
Haven't forgotten about this, just haven't gotten back around to working on it yet

PyvesB:
Let's close this one for the time being, we can always reopen later. :)

calebcartwright:
I don't particularly mind this being closed for now, though there's a number of PRs of similar age that but this seems to be the only one that was closed and I'm curious as to why?

PyvesB:
Every now and again, I do a pass over old PRs. In terms of organisation and sending a signal to contributors, I don't think it's great to have a long list lingering about. There is one older one, I'm working on getting it through the line, and will slowly continue upwards. :)

PR #8204, opened 2022-07-15
Fix missing conversion to `dayjs`, run [stackexchange]

chxseh:
Should fix `Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'moment' imported from /usr/src/app/services/stackexchange/stackexchange-monthlyquestions.service.js` from one file being missed in the PR #8192.


calebcartwright (approved):
good catch, thank you!

Issue #6275, opened 2021-03-16
Support filtering by prefix in GitHub latest Release and Tag Badges


xingao267:
:clipboard: **Description**

Please see context and discussions at https://github.com/badges/shields/discussions/6253

DAmNRelentless:
I'd love to see this as well, my reason is mentioned here #7082

mvndaai:
I want this for having multiple go packages with their own `go.mod` files within the same repo. Go allows versioning the subpackages by tagging like `subfolder/v0.0.1`. I would like to show badges for each subpackage in a table on the readme. That is why I made that PR ^.

The only issue I have in the PR is that the code already limits to getting 100 tags when sorting so if one subpackage is never updated and others have more that 100 releases then the filter might not work.

razonyang:
Hi, is there any updates? Need this for Go modules in same repo.

ricardojdsilva87:
Hello, this feature would be really helpfull.
On the README we have we are checking for releases for an helm chart for example for argoCD.
https://github.com/argoproj/argo-helm/releases
In this case we have several different helm releases for the different argocd components, like events, workflows, etc.
It would be usefull if we can have some kind of regex filter to show for example per date the release for argo-events-* or argo-cd-*.
This way we could monitor different components in the same repo.
Keep up the good work!
Thanks

chris48s:
Hello. We have just shipped a new feature which allows you to pass an arbitrary filter to the github tag and release badges. For example

The `filter` param can be used to apply a filter to the project's tag or release names before selecting the latest from the list. Two constructs are available: `*` is a wildcard matching zero or more characters, and if the pattern starts with a `!`, the whole pattern is negated.

e.g:
https://img.shields.io/github/v/tag/badges/shields
https://img.shields.io/github/v/tag/badges/shields?filter=%21server-%2A

This feature should be powerful and flexible enough to enable this use-case (and a bunch of others).

razonyang:
Hi @chris48s, is there an option to transform the tag name, in some context, the prefix (`modules/socials`) is redundant.

![Image](https://img.shields.io/github/v/tag/hbstack/header?filter=modules/socials%2A)

chris48s:
No the tag badge will always just display the name of the tag.
For releases you can pick either the tag name or the release name.

Issue #6766, opened 2021-07-14
Total downloads does not count all releases


liamg:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

The number in the total downloads badge for our repo has been fluctuating lately. We believe the issue is the badge is only counting downloads for the most recent 1000 releases, instead of paging over all API result pages for every release.

### 🔗 Link to the badge

https://img.shields.io/github/downloads/aquasecurity/tfsec/total

### 💡 Possible Solution

Iterate over each page of the GitHub API call results using `per_page=1000` and `page=N`.

chris48s:
I think the max value for `per_page` on this endpoint is 100, not 1000 (or 500)

https://api.github.com/repos/aquasecurity/tfsec/releases?per_page=500

returns the first 100 items

calebcartwright:
I'm not sure how common this will be in practice, but we may need something similar to the approach we had to take for the Docker badges of firing off the requests for the pages in parallel. If we have to go through the 5+ page sequence synchronously then I wouldn't be surprised if we run into timeout issues

calebcartwright:
The other challenge is that while we could theoretically have a max ceiling of two sequential calls with the v3 API (since we can calculate the calls for each page and fire them in parallel). However, I'm not sure the same is viable with graphql, as the latter will require us to actually iterate through each page one by one. This is unfortunate as it's tough to rationalize increased investment in v3 at this point.

AumGupta:
@liamg hi sorry for reinitialising a 5y old thread. But, I am facing the same issue now, was this ever resolved?

IvanJosipovic:
Also hitting this issue.

PR #8205, opened 2022-07-16
[GithubRelease],[GithubTag] filter by prefix

mvndaai:
This change allows:
* filtering github releases/tags by a prefix as requested in https://github.com/badges/shields/issues/6275
* trimming tag prefix before '/' using graphql `refPrefix` query


calebcartwright:
Thanks for the PR @mvndaai!

> One possible complication/bug in the code, especially for the tests, is that it requests the most recent 100 tags so if a prefix is further than 100 tags away it will ignore the prefix.

This is true but unrelated to the addition of this PR, and more just another impact of the fact that we don't support walking the pages for releases/tags yet (refs #6766); I wouldn't worry about this

services/github/github-release.tester.js
```diff
@@ -38,6 +38,10 @@ t.create('Release (repo not found)')
   .get('/v/release/badges/helmets.json')
   .expectBadge({ label: 'release', message: 'no releases or repo not found' })
 
+t.create('Release (filter by prefix)')
+  .get('/v/release/expressjs/express.json?prefix=4.16')
+  .expectBadge({ label: 'release', message: 'v4.16.4' })
```
calebcartwright, line 43: Thank you for adding a live test, but I worry this may prove to be too brittle (and yes I realize the latest version of express at this moment is 2 minor versions ahead).

It's beneficial to have a live test here to ensure the new query param is supported, but the matching value should be the schema based `isSemver`.

I would like to include a precise test as part of this but that should be done with unit tests. Consider adding one in `github-common-release.spec.js`

services/github/github-tag.service.js
```diff
@@ -73,8 +83,8 @@ class GithubTag extends GithubAuthV4Service {
     }
   }
 
-  async fetch({ user, repo, sort }) {
-    const limit = sort === 'semver' ? 100 : 1
+  async fetch({ user, repo, sort, prefix }) {
+    const limit = sort === 'semver' || prefix ? 100 : 1
```
calebcartwright, line 87: I can't add the comment down below but, is there any reason we couldn't incorporate the prefix (when one is provided) as part of the query itself?

services/github/github-tag.tester.js
```diff
@@ -28,6 +28,10 @@ t.create('Tag (repo not found)')
   .get('/v/tag/badges/helmets.json')
   .expectBadge({ label: 'tag', message: 'repo not found' })
 
+t.create('Tag (filter by prefix')
+  .get('/v/tag/expressjs/express.json?prefix=4.16')
+  .expectBadge({ label: 'tag', message: 'v4.16.4' })
```
calebcartwright, line 33: Same comment about this test as the release tester

calebcartwright (commented):
Thanks again, few items left inline for your review

services/github/github-tag.service.js
mvndaai: I tried to update the graphql but a prefix is not supported. I can add a `query` parameter but that matches by the [github query syntax ](https://docs.github.com/en/search-github/getting-started-with-searching-on-github/understanding-the-search-syntax) which might be a nice way to search but prefix is not supported there either.

services/github/github-release.tester.js
mvndaai: I agree that since expressjs can change their tags this test is brittle. I removed the live tests and added unit tests instead.

If you want live tests I think I have an old repo that I never plan on touching. I can add a few releases but then my username is all over your repos.

services/github/github-tag.tester.js
mvndaai: Same as above. I removed this and added a unit test

services/github/github-tag.service.js
chris48s: I haven't fully read this PR, but we recently added another badge which uses the graphql API to query tags by prefix using `refPrefix`:

https://github.com/badges/shields/blob/37a36f5feac3c2143ea06fb9b6528ff5979f2dd2/services/ros/ros-version.service.js#L62-L82

does that help?

mvndaai: @chris48s thanks for reviewing!

When working on this change I tried editing the `refPrefix` line and I don't know what is accepted there. Putting a prefix of the tag as that `${distro}` variable errors and doesn't return anything. Maybe I am just using it wrong but I wasn't able to find documentation on what would go there.

Also, I don't see any variable in the shield.io UI that allows a prefix when you search for `github tag` under Versions. 

If this is already done I would be happy to just use that I just didn't see a way.

calebcartwright: Tangentially related, but you may find GitHub's Explorer useful in the future when fiddling around with their GraphQL queries: https://docs.github.com/en/graphql/overview/explorer

What I was alluding to, and I believe Chris as well, is that it seems like we'd be able to utilize the value of the `prefix` parameter of this function directly as part of the GraphQL query we are issuing which would allow GitHub to do the work for us.

Specifically, I think this would entail adding a param on line 80/90 (old/new line):

```js
       query ($user: String!, $repo: String!, $limit: Int!, $prefix: String!) {
```

and then incorporate that param down on line 83/93:
```js
refPrefix: "refs/tags/$prefix"
```

finally updating the variables on line 96/106

```js
variables: { user, repo, limit, prefix },
```

It's not clear to me what issue you were encountering or if this was even something you'd already tried, but even if the above doesn't work literally as-is I'd still think something like this would be possible. E.g. if the string interpolation for the ref prefix) then we could set the refPrefix value outside the query and just pass that value in as the variable (`refPrefix: $prefix`)

services/github/github-tag.service.js
```diff
@@ -99,7 +109,13 @@ class GithubTag extends GithubAuthV4Service {
...
   }
 
-  static getLatestTag({ tags, sort, includePrereleases }) {
+  static getLatestTag({ tags, sort, includePrereleases, prefix }) {
+    if (prefix) {
+      const tagsWithPrefix = tags.filter(tag => tag.startsWith(prefix))
+      if (tagsWithPrefix.length > 0) {
+        tags = tagsWithPrefix
+      }
+    }
```
calebcartwright, line 118: My guess is that you did this given what you said about the 100 tag limit given the context of the current query. However, I believe if we can get the query right and allow GitHub to only return results that match that prefix, then we can handle things more directly here. I.e. if we ask GitHub to give us only the tags that are prefixed with `foo-` and GitHub tells us there are none, then we can and should throw an error here (a `NotFound`) which we could do safely

services/github/github-tag.service.js
mvndaai: @calebcartwright thanks for sharing that explorer, it was super helpful! I guess it comes down to what you want the prefix to do. The  `refPrefix` requires that a tag have a prefix that ends with `/`

I want this prefix for Golang sub packages and their requirement is to have a `folder/` prefix but the ticket that I used for this PR wanted to have any generic prefix.

For example my package `mvndaai/ctxerrhelper` currently has 4 tags `["logrus/v0.0.2", "logrus/v0.0.1", "example/v0.0.2",  "example/v0.0.1"]`
If I search for the prefix `example/` then I get back ~`["v0.0.2", "v0.0.1"]`. The only allowed prefixes are `example` and `logrus` which totally works for me but doesn't work for the other generic ask of something like `ex`.

Here is an example query that you can use if you want to test
```graphql
query {
    repository(owner: "mvndaai", name: "ctxerrhelper") {
        refs(
            refPrefix: "refs/tags/example/"
            first: 30
            orderBy: { field: TAG_COMMIT_DATE, direction: DESC }
        ) {
            edges { node { name } }
        }
    }
}
```
--------
I guess the questions are:
- Is it okay to require a prefix to end with `/`?
- Since only tags uses graphql, are you happy if tags and releases work differently?

mvndaai: I decided to simplify this PR only change only tags, no releases, and use the graphql refPrefix.

chris48s: I had not realised that the `refPrefix` param requires the prefix to end with a slash. Sorry if I've muddied the waters there. I can see it will be useful to filter by any prefix. For example, if you think about the tags on this repo https://github.com/badges/shields/tags?after=server-2021-07-01 I can see you might want to filter by `server-`. Maybe the original implementation would be better after all. Sorry if I've sent us on a goose chase here.

mvndaai: I think both ways have their place. Maybe I will update this tonight to handle both ways with a different param for the prefix that requires a '/'

services/github/github-tag.service.js
```diff
@@ -73,14 +91,26 @@ class GithubTag extends GithubAuthV4Service {
     }
   }
 
-  async fetch({ user, repo, sort }) {
-    const limit = sort === 'semver' ? 100 : 1
+  async fetch({ user, repo, sort, prefix, subpackage }) {
```
chris48s, line 94: I'm confused about what we're trying to achieve here. There's 2 variables here: `prefix` and `subpackage` but `prefix` doesn't seem to actually get used and 'subpackage' is an odd choice of concept/variable name for a generic github badge. You've talked about go subpackages, but this is a generic github tag badge. We shouldn't be introducing a concept which is specific to some particular language ecosystem here.

It seems like we just want a single generic "prefix" filter var, which may or may not end with a / and could be a go subpackage, or any other prefix. If you want to use `refPrefix` where possible (due to the 100 objects per page limit), maybe we just check for `prefix.endsWith('/')` ?

services/github/github-common-release.js
```diff
@@ -68,6 +76,8 @@ function getLatestRelease({ releases, sort, includePrereleases }) {
 const queryParamSchema = Joi.object({
   include_prereleases: Joi.equal(''),
   sort: Joi.string().valid('date', 'semver').default('date'),
+  prefix: Joi.any(),
+  subpackage: Joi.any(),
```
chris48s, line 80: Once we've sorted out what the params are, we shouldn't use `Joi.any()`. They need a type. Probably should be `Joi.string().required()` but if we actually need to accept multiple types, we can use https://joi.dev/api/?v=17.6.0#alternatives

services/github/github-tag.service.js
chris48s: Agreed :+1:  Lets throw a `NotFound()` if there are 0 matching tags

services/github/github-tag.tester.js
```diff
@@ -28,6 +28,10 @@ t.create('Tag (repo not found)')
   .get('/v/tag/badges/helmets.json')
   .expectBadge({ label: 'tag', message: 'repo not found' })
 
+t.create('Tag (filter by prefix')
+  .get('/v/tag/ros/rosdistro.json?subpackage=galactic')
+  .expectBadge({ label: 'tag', message: isKebabYearMonthDay })
```
chris48s, line 33: You don't need to define a custom regex validator for this. `Joi.date().required()` will do what you're trying to do here.

services/github/github-tag.service.js
mvndaai: The tags returned by `refPrefix`do not include the prefix so if you had a tag `example/v0.0.1` the subpackage route returns `v0.0.1` and the prefix would return `example/v0.0.1` Since those are different I didn't want to one param to auto switch between them because it would make for a bad experience.

`prefix` here is just used to get 100 tags instead of just the newest which is the same experience as `sort === 'semver'`. Filtering by prefix happens in the function `getLatestTag `.

I agree that you don't want language specific things so do you have a better name for a prefix that end with `/`? Maybe `folder` or `pathPrefix`

services/github/github-common-release.js
mvndaai: I used `Joi.any()` over `Joi.string()` because when I had something that it thought was a number it would return undefined. So a prefix of `0.1` Showed up as undefined. I will look into accepting multiple.

services/github/github-tag.service.js
mvndaai: Done, thanks!

services/github/github-common-release.js
mvndaai: Fixed. Thanks!

services/github/github-tag.tester.js
mvndaai: That worked. Thanks!

services/github/github-tag.service.js
chris48s: OK, and is the reason we are talking about a variant that strips the prefix because we're looking for a way to use `refPrefix`, or because that was the original intent? As I said earlier, I think suggesting the use of `refPrefix` was not super-useful on my part and may have sent us down the wrong road - I hadn't realised the limitations when I suggested it.

I think if we do want both a variant that includes the prefix and a variant that excludes it (because that is actually useful), then the way to express this would be with a flag param called `exclude_prefix` or maybe `strip_prefix` which works regardless of whether `prefix` ends with a `/` or not.

i.e: if your list of tags is

https://github.com/badges/shields/tags

- `?prefix=server-` ==> latest tag is `server-2022-08-01` (this should be the default, because that is the actual tag name)
- `?prefix=server-&exclude_prefix` ==> latest tag (excluding prefix) is `2022-08-01`

We express these flags in the `queryParamSchema` using `exclude_prefix: Joi.equal('')`.

How does that sound?

mvndaai: I can remove the `subpackage` stuff if you want but I actually like the two separate flows. In my use case for Go the `subpackage` flow is the better fix. 

Since the subpackage flow uses `refPrefix` it doesn't have the issue of needing to be within the last 100 tags. Also in the case that you want to filter the subpackage tags by a `prefix` that would work as well. 

I personally don't think excluding the prefix is often useful but `refPrefix` does that automatically. I don't want to have some code checking for a prefix and adding or removing it depending on if the prefix ends in `/` or not. That just seems too complicated. I just think the subpackage flow needs a better name.

calebcartwright: At a minimum `subpackage` is going to have to be renamed as the user-facing query parameter. It's crucial that we separate inner details, like behavior of a GraphQL query parameter, from the user interface surface. The parameters, and their respective names, need to clearly indicate to any and every user how that parameter will behave and reflect in the final badge.

It shouldn't be driven by one particular usage pattern (e.g. `subpackage` may make sense for tags from a repo that's Go-based, but would be puzzling in say a JavaScript repository).

At a high level, I feel like the end state in terms of badge scenarios is pretty straightforward, wherein there's a singular repository that's a monorepo of sorts or which at least contains multiple components that are versioned and tagged independently, and folks in those scenarios want to be able to get github repo tag-based version badges for those respective components.

`prefix` makes sense as a query param in this context, as we're allowing users to filter down to a subset of the tags. `subpackage` as the name does not immediately scream "the selected tag version value based on the prefix, but with the prefix stripped".

We can do whatever we need to internally, and certainly still support both having the message be prefix-ful and prefix-less to support different user needs, but we just need to find a way to surface up those two separate cases in a more intuitive manner.

I'd echo Chris' prior suggestion for how to handle this because it's very explicit and clear. Yes, that would require 2 query params instead of just 1 in the badge url, but always important to remember that badge urls are copied/written once and then never looked at again (not something regularly typed by a human). If you'd still prefer trying to support both paths with a single query parameter, then we'd have to rename both params to make each explicit (e.g. `filterAndStripPrefix`)

mvndaai: Thanks for that feedback! I think this final change makes everyone happy.

I removed `subpackage` and added a bool of `trimPrefixAtSlash` that splits by the last slash into a prefix used in a filter and a prefix used in graphql. When it is not checked the whole prefix is used in the filter.

services/github/github-tag.tester.js
```diff
@@ -28,6 +28,10 @@ t.create('Tag (repo not found)')
   .get('/v/tag/badges/helmets.json')
   .expectBadge({ label: 'tag', message: 'repo not found' })
 
+t.create('Tag (filter by prefix + trim prefix at slash)')
+  .get('/v/tag/ros/rosdistro.json?subpackage=galactic/&trimPrefixAtSlash')
```
calebcartwright, line 32: As I was reading through the diff again I was starting to feel like I'd be more comfortable with a greater number of tests, and I definitely feel that way after getting to this test, which is using the no-longer valid/existent query parameter name.

I think it would be helpful to try to move more of the logic out of the `handle` function(s) (which are more difficult to test) into standalone functions (either class members or free functions) that can be directly unit tested and consumed by `handle` and/or `fetch` functions as needed

mvndaai: I fixed that test. I had assumed that `Joi.date().required()` meant that the string had to be only a date, not just include one. Since all that repo's tags are something like `galactic/2022-10-04` it matched on all tags, which is disappointing because originally I had made a regex to just match the date but switched to `Joi.date` instead on feedback. Now It checks for a date and a prefix of `20`.

I just reviewed the code and I don't know how much more I can break up. I could maybe make a function of splitting the prefix but when you want to trim but I don't see how much more the code could be split. If you have some pointers I would gladly take them.

Thanks for all the time you have spent reviewing this!

chris48s:
Hello. We have just shipped a new feature which allows you to pass an arbitrary filter to the github tag and release badges. For example

The `filter` param can be used to apply a filter to the project's tag or release names before selecting the latest from the list. Two constructs are available: `*` is a wildcard matching zero or more characters, and if the pattern starts with a `!`, the whole pattern is negated.

e.g:
https://img.shields.io/github/v/tag/badges/shields
https://img.shields.io/github/v/tag/badges/shields?filter=%21server-%2A

This feature should be powerful and flexible enough to enable this use-case (and a bunch of others).

PR #8208, opened 2022-07-20
feat: add [GitlabForks] server

sunny0826:


services/gitlab/gitlab-forks.service.js
```diff
@@ -0,0 +1,74 @@
...
+    },
+  ]
+
+  static defaultBadgeData = { label: 'forks', namedLogo: 'gitlab' }
+
+  static render({ baseUrl, project, forkCount }) {
+    return {
+      message: metric(forkCount),
+      color: 'blue',
+      link: [`${baseUrl}/${project}/-/forks/new`],
```
chris48s, line 52: The equivalent links for what we have on the GitHub badge would be

```js
[
    `${baseUrl}/${project}/-/forks/new`,
    `${baseUrl}/${project}/-/forks`
]
```

how about we do the same here?

services/gitlab/gitlab-forks.tester.js
```diff
@@ -0,0 +1,20 @@
...
+    color: 'blue',
+    link: ['https://gitlab.com/gitlab-org/gitlab/-/forks/new'],
+  })
+
+t.create('Forks (project not found)')
+  .get('/user1/gitlab-does-not-have-this-repo.json')
+  .expectBadge({
+    label: 'forks',
+    message: 'project not found',
+  })
```
chris48s, line 20: Could we have a test for the private repo case? Also, we've overlooked it in the other reviews but none of these GitLab badges we have added recently include a test for the case where we pass a custom `baseUrl`. I think I've noticed it here because the links include `baseUrl`. We don't need to go back and deal with them all in this PR but it would be nice to have test cases for this so we don't accidentally break it in a future refactor. https://gitlab.gnome.org/ is a good example of a self-hosted GitLab server with lots of public projects. We've used it in some of the other gitlab test modules

sunny0826: This was an oversight on my part, I added the `gitlab_url` parameter to the example, but did not provide a test case for the custom instance.

services/gitlab/gitlab-forks.service.js
sunny0826: I refer to the GitHub Service code at https://github.com/badges/shields/blob/c64bb7e1f35c71d7679a415748b093c5e06badae/services/github/github-stars.service.js#L49-L50

chris48s: so the corresponding code for the equivalent github badge would be

https://github.com/badges/shields/blob/3f63990049b0f5967811aaced4c7db91a8bb2e51/services/github/github-forks.service.js#L49-L52

that makes the left side of the bade a link to for the repo and the right side a link to view forks. I'm suggesting we do the same for gitlab

sunny0826: got it

chris48s (approved):
Cheer.s I've pushed e3c45e9 which just updates the link on the message side to the forks page

PR #8209, opened 2022-07-20
feat: add [GitlabStars] service

sunny0826:


services/gitlab/gitlab-stars.service.js
```diff
@@ -0,0 +1,74 @@
...
+    },
+  ]
+
+  static defaultBadgeData = { label: 'stars', namedLogo: 'gitlab' }
+
+  static render({ baseUrl, project, starCount }) {
+    return {
+      message: metric(starCount),
+      color: 'blue',
+      link: [`${baseUrl}/${project}`],
```
chris48s, line 52: The equivalent links for what we have on the GitHub badge would be

```js
[
    `${baseUrl}/${project}`,
    `${baseUrl}/${project}/-/starrers`
]
```

how about we do the same here?

services/gitlab/gitlab-stars.tester.js
```diff
@@ -0,0 +1,20 @@
...
+    color: 'blue',
+    link: ['https://gitlab.com/gitlab-org/gitlab'],
+  })
+
+t.create('Stars (project not found)')
+  .get('/user1/gitlab-does-not-have-this-repo.json')
+  .expectBadge({
+    label: 'stars',
+    message: 'project not found',
+  })
```
chris48s, line 20: Similar comment to https://github.com/badges/shields/pull/8208#discussion_r927078542

It would be great if we had tests for the private repo and custom `baseUrl` cases.

services/gitlab/gitlab-stars.service.js
calebcartwright: Unresolving as we should include the `starrers` link on the right too, e.g. https://gitlab.com/gitlab-org/gitlab/-/starrers

sunny0826: ops, I will fix this.

sunny0826: done

chris48s (approved):

PR #8211, opened 2022-07-20
Test review apps

chris48s:


calebcartwright:
I want to say i had to manually kick off deploys last time, but I wasn't sure if that was just because there were failures that weren't showing up or whether the auto triggers were just off

chris48s:
Do you have a button to manually kick one off? I don't

Screenshot at 2022-07-20 19-22-52

PR #8213, opened 2022-07-20
Fix heroku review apps (sort of)

chris48s:
Right. This is probably worth merging as it is better than a kick in the teeth, but we have some problems still to solve.

1. We've got review apps set up to auto-create review apps for pull requests

    Screenshot at 2022-07-20 20-22-05

    but as far as I can tell, its only creating them if the PR is submitted from a branch on this repo - it won't auto-create the review app if the PR comes from a fork. This would at least enable review apps doe dependabot PRs which would help.

2. If heroku auto creates the review app, I can see the logs etc but I don't have a button for "create review app" (but I think maybe @calebcartwright  does?) This is what I have

    Screenshot at 2022-07-20 20-37-15

4. Previously, we did have review apps set up so that they had access to a few tokens. The most important one being `GH_TOKEN` but there were a few other services where the review apps had a token. We seem to have lost all those somewhere along the line but I can't remember how we set them. If you have a look at https://dashboard.heroku.com/apps/shields-fix-review-apps-uhq3yh/settings (the corresponding settings for https://shields-fix-review-apps-uhq3yh.herokuapp.com/ ) we've only got the ones that have an explicit value in `app.json` or the ones heroku automatically sets.

5. The review apps use a heroku nodejs buildpack but then we deploy using our Dockerfile. Heroku does allow docker builds so we should probably switch to using a docker image for the review apps so that our environments resemble each other more closely. In some ways this is a longer term big picture thing but there's also a part of me that is thinking maybe we should deal with this one first because we might find we solve all the above problems then switch the buildpack and have to solve them all again.


calebcartwright:
> If heroku auto creates the review app, I can see the logs etc but I don't have a button for "create review app" (but I think maybe @calebcartwright does?) This is what I have

I see the same as you when logged in with my personal account, but had the deploy buttons when logged in with the admin account.

> We seem to have lost all those somewhere along the line but I can't remember how we set them

I noticed that large chunks of the environment had been wiped out when I first got logged back in with the admin account, including the removal of all our collaborator access. I added all of us back (although something still seems up there), and I think we can set the vars for the review ups under the review app settings here: https://dashboard.heroku.com/pipelines/d8fd42e8-8e42-466e-98fc-f5f50edbdd8f/settings. However, similarly to deploy the "Add" button is disabled for me

> The review apps use a heroku nodejs buildpack but then we deploy using our Dockerfile. Heroku does allow docker builds so we should probably switch to using a docker image for the review apps so that our environments resemble each other more closely.

fwiw squint is setup this way. however, my recollection is that if you wanted auto deployed review apps then you need to let heroku build and push the image to its own internal registry, or at least there were some complications with trying to use an image hosted on an external registry

calebcartwright (approved):
i'm good with moving ahead just to give us _something_ and avoid the need to locally check out remote PR branches as part of reviews. don't feel too strongly though if you want to hold off

chris48s:
> you need to let heroku build and push the image to its own internal registry, or at least there were some complications with trying to use an image hosted on an external registry

I think it is a non-issue in this case. We don't push a build to dockerhub for every open PR (and I wouldn't want to) so there is no external registry we could use anyway.

> i'm good with moving ahead just to give us something and avoid the need to locally check out remote PR branches as part of reviews. don't feel too strongly though if you want to hold off

I'm going to merge this on the basis that I'm unlikely to make it better this evening and having review apps in this state will make reviewing the dependabot PRs quicker tomorrow - those come from branches on this repo not a fork so they should work. This isn't a solved problem though - there is clearly more to be done.

PR #8241, opened 2022-07-24
try upgrading gatsby again

chris48s:
> The definition of a fool is someone who does the same thing over and over again expecting different results


chris48s:
possibly solvable https://github.com/badges/shields/pull/8168

not so much
- https://github.com/oliver-moran/jimp/pull/1092
- https://github.com/oliver-moran/jimp/pull/1087

chris48s:
Right. Looks like https://github.com/badges/shields/pull/8265 solves one of these. Still need to circle back to the rest of it..

chris48s:
replaced by #8334

Issue #8246, opened 2022-07-25
PyPI Framework Version Badge


ericof:
:clipboard: **Description**

- Plone version badge (Implemented via PyPI API)
- Example: "plone version: 5.2|6.0"

:link: **Data**

- Implemented via PyPI API, in a similar way it is done with _PyPI - Django Version:_

:microphone: **Motivation**

Plone is one of the most popular open-source CMS, and many of our add-ons already use shields.io to display PyPi and GitHub information.

chris48s:
I don't disagree with this as a feature request, and I appreciate we do currently have a specific badge for django versions (fun fact: the django versions badge was my first contribution to shields many years ago), but I wonder if it would actually make sense to approach this in a slightly more general way.

PyPI supports a whole bunch of these classifiers https://github.com/pypa/trove-classifiers/blob/d4ad70c1b4c79ef2e0649c4a3b3184432f19a108/src/trove_classifiers/__init__.py#L71-L209 and a badge listing compatible Wagtail or Django CMS versions is as valid as Django or Plone.

I think we could probably generalise this to a framework classifiers versions badge and convert the existing djversions route to a redirect.

I think the most sensible route for this would probably be `/pypi/frameworkversions/:frameworkName/:packageName` e.g: `/pypi/frameworkversions/django%20cms/djangocms-history`.

ericof:
Hello @chris48s!

I just reviewed the [Trove Classifiers](https://pypi.org/classifiers/), and for `Frameworks` I came with the following table of possible options:

| Framework | Regex 1 | Regex Alt |
| --- | --- | --- |
| AWS CDK | \d+  |  |
| Django | \d+.\d+ | \d+ |
| Django CMS | \d+.\d+ |  |
| JupyterLab | \d+ |  |
| Odoo | \d+.\d+ |  |
| Plone | \d+.\d+ |  |
| Wagtail | \d+ |  |
| Zope | \d+ |  |

I could prepare a PR to implement that with a route like `/pypi/frameworkversions/:frameworkName/:packageName` but I would consider the following restrictions:

- Only Frameworks on the table above would be valid
- `frameworkName` should be case sensitive (to avoid "guessing the correct spelling")

And I have no idea - yet - how to implement the redirect ;-)

What do you think?

PR #8247, opened 2022-07-25
[pypi] Implement Plone versions badge

ericof:
This PR implements #8246 using the PyPI api.

### DIFF
--- a/services/pypi/pypi-base.js
+++ b/services/pypi/pypi-base.js
@@ -9,18 +9,16 @@ const schema = Joi.object({
     license: Joi.string().allow('').allow(null),
     classifiers: Joi.array().items(Joi.string()).required(),
   }).required(),
-  releases: Joi.object()
-    .pattern(
-      Joi.string(),
-      Joi.array()
-        .items(
-          Joi.object({
-            packagetype: Joi.string().required(),
-          })
-        )
-        .required()
-    )
-    .required(),
+  releases: Joi.object().pattern(
+    Joi.string(),
+    Joi.array()
+      .items(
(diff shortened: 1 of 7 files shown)


ericof:
Hello there, any idea why the tests failed?

calebcartwright:
> Hello there, any idea why the tests failed?

Not from looking at the CI output (which is admittedly and suspiciously empty). Are the tests passing for you locally? 

```
npm run test:services -- --only=pypi
```

ericof:
@calebcartwright I just re-run the tests locally and noticed it was flaky due to a change on the [PyPI API](https://warehouse.pypa.io/api-reference/json.html).
Hopefully, this was the cause of the error.

ericof:
New error, but not sure what it means

calebcartwright:
I've still not looked at this in detail, but I wouldn't have expected the addition of this badge to require modification to unrelated unit and service tests for other badges.

> noticed it was flaky due to a change on the [PyPI API](https://warehouse.pypa.io/api-reference/json.html).

It does seem like we've got some recurring test failures with `invalid response data` being quite prevalent which does indeed typically indicate an upstream change. However, I'm a bit surprised we haven't had any bug reports about this yet :thinking: 

If you feel like there's an existing bug (and I suspect something may be afoot too), it would be best to address that in a separate standalone PR so we can push that through independent of adding a new badge. Would you be up for doing that?

ericof:
@calebcartwright I could break this PR into 2: One for the PyPI issue and another one for the new badge.

The issue for the PyPI API change is #8250

I will close this PR for now

PR #8249, opened 2022-07-26
Add [galaxytoolshed] Version

guillaume-gricourt:
Hi,
This PR adds a feature to the galaxytoolshed service. Referring to https://github.com/badges/shields/issues/7660.
The schema is updated, I hope there is no mistake about its construction.
Many thanks

### DIFF
--- a/services/galaxytoolshed/galaxytoolshed-version.service.js
+++ b/services/galaxytoolshed/galaxytoolshed-version.service.js
@@ -1,0 +1,85 @@
+import { NotFound } from '../index.js'
+import { renderVersionBadge } from '../version.js'
+import GalaxyToolshedService from './galaxytoolshed-base.js'
+
+export class GalaxyToolshedVersion extends GalaxyToolshedService {
+  static category = 'version'
+  static route = {
+    base: 'galaxytoolshed/v',
+    pattern: ':repository/:owner/:tool?/:requirement?',
+  }
+
+  static examples = [
+    {
+      title: 'Galaxy Toolshed (repository)',
+      pattern: ':repository/:owner',
+      namedParams: {
+        repository: 'sra_tools',
+        owner: 'iuc',
+      },
+      staticPreview: renderVersionBadge({ version: '1.2.5' }),
+    },
+    {
+      title: 'Galaxy Toolshed (tool)',
+      pattern: ':repository/:owner/:tool',
+      namedParams: {
+        repository: 'sra_tools',
+        owner: 'iuc',
+        tool: 'fastq_dump',
+      },
+      staticPreview: renderVersionBadge({ version: '1.2.5' }),
+    },
+    {
+      title: 'Galaxy Toolshed (requirement)',
+      pattern: ':repository/:owner/:tool/:requirement',
+      namedParams: {
+        repository: 'sra_tools',
+        owner: 'iuc',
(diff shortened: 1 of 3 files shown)


calebcartwright:
Thank you for the PR! I recognize that we probably discussed this previously on other PRs, but I think it would be helpful if you could articulate what these badges are/what they look like here inline for greater visibility

calebcartwright:
Thanks, though I was actually thinking more along the lines of the types of things we ask for/discuss when new badges are requested (e.g. https://github.com/badges/shields/issues/new?assignees=&labels=service-badge&template=3_Badge_request.md)

Some of those items are already inherent, but it would be good if you could demonstrate the different variants of badges that would be displayed along with an associated "why" for each one

services/galaxytoolshed/galaxytoolshed-version.service.js
```diff
@@ -0,0 +1,85 @@
...
+        tool: 'fastq_dump',
+        requirement: 'perl',
+      },
+      staticPreview: renderVersionBadge({ version: '5.18.1' }),
+    },
+  ]
+
+  static transform({ response, tool, requirement }) {
+    if (tool !== undefined) {
+      const dataTool = response[1].valid_tools.filter(function (x) {
```
PyvesB, line 47: We generally use the less verbose `x => x.id === tool` anonymous form for these functions, could you please update them?

services/galaxytoolshed/galaxytoolshed-version.service.js
```diff
@@ -0,0 +1,85 @@
...
+import GalaxyToolshedService from './galaxytoolshed-base.js'
+
+export class GalaxyToolshedVersion extends GalaxyToolshedService {
+  static category = 'version'
+  static route = {
+    base: 'galaxytoolshed/v',
+    pattern: ':repository/:owner/:tool?/:requirement?',
+  }
+
+  static examples = [
```
PyvesB, line 12: We in the process of migrating our examples to the OpenAPI format (see #9285). Could. I please get you to update your PR accordingly? https://github.com/badges/shields/blob/master/services/cirrus/cirrus.service.js is a simple example using the new format.

PyvesB (commented):
Sorry this has been left on the sideway for so long @guillaume-gricourt. I'm gonna pick the review up and try to get it landed - hopefully not too late from your perspective :)

services/galaxytoolshed/galaxytoolshed-version.service.js
guillaume-gricourt: `examples` were migrated to `openApi`

services/galaxytoolshed/galaxytoolshed-version.service.js
guillaume-gricourt: Simplifications of filtering and conditional tests were done in `transform()`

PyvesB:
We're almost there! When searching for the new badges, it's not obvious that they are version badges:
Screenshot 2024-01-04 211604

I suggest adding "Version" to the title of the new badges, for example `Galaxy Toolshed - Tool Version`.

PyvesB (approved):
Thanks for the contribution and the patience, this looks great! Will merge now and deploy this weekend if no one does before then :)

Issue #7876, opened 2022-04-22
GitHub file size for a specific branch


mattrossman:
:clipboard: **Description**

A version of the [GitHub file size service](https://shields.io/category/size) that lets you specify a branch (or other ref value) to read from.

The output would look the same: `size | 20 kB`

:link: **Data**

This data is available from the same endpoint as the existing service in [`github-size.service.js`](https://github.com/badges/shields/blob/560d26784494a39965fe0bcb0ab658bb37245349/services/github/github-size.service.js). The branch/tag/commit can be specified via a `ref` query parameter.

Example: https://api.github.com/repos/badges/shields/contents/index.html?ref=gh-pages

Documentation: https://docs.github.com/en/rest/repos/contents#get-repository-content--parameters

:microphone: **Motivation**

I deploy build artifacts to a `gh-pages` branch for hosting on GitHub Pages as part of my CI/CD pipeline. I would like a badge that shows the size of these build artifacts. The current size service only uses the default branch.

calebcartwright:
Thanks for sharing your idea and including relevant details! I've not read through the specifics of the documentation but sounds straightforward enough and beneficial.

This should be a relatively easy addition for anyone interested in working on it, with the one implementation note being that our badge would need to accept the ref parameter as a query parameter since the route currently ends with a parameter aligning to path to the file within the repo (and both paths and branch-refs can contain slashes)

PaulaBarszcz:
I would like to work on this issue.

@calebcartwright - thanks for your comment, I will make sure that the ref parameter is accepted as a query parameter (as to avoid confusion between the ref and the path to the file) :)

calebcartwright:
> I would like to work on this issue.
> 
> @calebcartwright - thanks for your comment, I will make sure that the ref parameter is accepted as a query parameter (as to avoid confusion between the branch/tag/commit reference and the path to the file) :)

Awesome! Let us know if you have any questions along the way (feel free to open a draft PR to discuss code level things, and/or more general chat, and/or reach out on Discord)

PaulaBarszcz:
This issue can be closed thanks to the https://github.com/badges/shields/pull/8262 having already been merged to master. Thanks :)

PR #8262, opened 2022-07-29
[GitHub] GitHub file size for a specific branch

PaulaBarszcz:
Related issue: #7876 

It is my first contribution to Shields - all comments and suggestions are welcome! 😉 

Tested locally, gives expected results, some screenshots for example:

image
**commit**: 345da16
"size": 3572,
https://api.github.com/repos/badges/shields/contents/services/text-formatters.js?ref=345da16
-
image
**branch** `integration-docs`
 "size": 3442,
https://api.github.com/repos/badges/shields/contents/services/text-formatters.js?ref=integration-docs
-

image
**tag** `server-2021-10-04`
"size": 3250,
https://api.github.com/repos/badges/shields/contents/services/text-formatters.js?ref=server-2021-10-04
-
image
**tag** `2.0.0-beta1`
our file is not present in this tag
https://api.github.com/repos/badges/shields/contents/services/text-formatters.js?ref=2.0.0-beta1 
-
If we don't specify the branch/commit/tag, it searches the master branch by default.

On the Shields website locally, it looks like this:

image

---
The previous endpoint for GitHub file size still works correctly.


services/github/github-size.service.js
```diff
@@ -41,16 +60,20 @@ export default class GithubSize extends GithubAuthV3Service {
...
     })
   }
 
-  async handle({ user, repo, path }) {
-    const body = await this.fetch({ user, repo, path })
+  async handle({ user, repo, path }, queryParams) {
+    let branch = queryParams.branch
+    if (!branch) {
+      branch = 'master'
+    }
```
chris48s, line 75: The default branch for a repo is configurable and most repos now use `main`. Instead of setting `branch` explicitly here if it is not supplied, pass a `?ref=` param in our API call only if `branch` is supplied. If the user doesn't supply the param we can leave github to work out the default.

PaulaBarszcz: Hi @chris48s,
Good point, updated, thanks.

PaulaBarszcz:
Hi @chris48s,

I see that the test suite failed on my feature branch. Is it something that I can fix? :

image

https://app.circleci.com/pipelines/github/badges/shields/14245/workflows/19814146-9649-439a-ab11-7b6899bb2f07/jobs/192075

calebcartwright:
> Hi @chris48s,
> 
> I see that the test suite failed on my feature branch. Is it something that I can fix?:
> 
> ```
> GitlabContributors: no credentials configured, tests for this service will be skipped. Add credentials in local.yml to run them.
> GitlabLicense: no credentials configured, tests for this service will be skipped. Add credentials in local.yml to run them.
> ```
> 
> from: https://app.circleci.com/pipelines/github/badges/shields/14245/workflows/19814146-9649-439a-ab11-7b6899bb2f07/jobs/192075
> 
> image

Seems it was specious. I reran the failed jobs and they passed

PaulaBarszcz:
> > Hi @chris48s,
> > I see that the test suite failed on my feature branch. Is it something that I can fix?:
> > ```
> > GitlabContributors: no credentials configured, tests for this service will be skipped. Add credentials in local.yml to run them.
> > GitlabLicense: no credentials configured, tests for this service will be skipped. Add credentials in local.yml to run them.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > from: https://app.circleci.com/pipelines/github/badges/shields/14245/workflows/19814146-9649-439a-ab11-7b6899bb2f07/jobs/192075
> > image
> 
> Seems it was specious. I reran the failed jobs and they passed

Ah, ok, thanks :)

I am currently not working on anything on this branch. If there is something that should be changed before a merge to the master branch, I am open to suggestions :)

chris48s (approved):

chris48s:
Ta - merged. Running the full test suite for github is a sufficiently large number of integration tests that sometimes we'll hit a flaky one. You can actually filter this down further if you want. For example, if you run `npm run test:services -- --only="githubsize"` locally that will run just the github size tests, and you can also use `[githubsize]` in a PR title.

services/github/github-size.service.js
```diff
@@ -41,16 +60,25 @@ export default class GithubSize extends GithubAuthV3Service {
...
+        schema,
+        errorMessages: errorMessagesFor('repo, branch or file not found'),
+      })
+    } else {
+      return this._requestJson({
+        url: `/repos/${user}/${repo}/contents/${path}`,
+        schema,
+        errorMessages: errorMessagesFor('repo or file not found'),
+      })
+    }
```
calebcartwright, line 76: This is fine, but one thing to be aware of for future reference is that we typically like to use the `options` key on the arg which in turn allows our http client library to handle the conditionality of the query search parameters for us. I gather there's also a need to have slightly different error texts depending on the branch, though that could also be handled with a single call too, e.g.:

```js
    return this._requestJson({
      url: `/repos/${user}/${repo}/contents/${path}`,
      schema,
      options: { searchParams: { ref: branch } },
      errorMessages: errorMessagesFor(
        `repo${branch ? ', branch' : ''} or file not found`
      ),
    })
```

PaulaBarszcz: That’s neat, I will try to remember about this syntax - thanks ☺️

PaulaBarszcz:
> Ta - merged. Running the full test suite for github is a sufficiently large number of integration tests that sometimes we'll hit a flaky one. You can actually filter this down further if you want. For example, if you run `npm run test:services -- --only="githubsize"` locally that will run just the github size tests, and you can also use `[githubsize]` in a PR title.

Cool, thanks! I was not sure if it was better to run the whole GitHub test suite or a narrower one, now I know that it would be fine also with a narrower version 😌

Issue #8248, opened 2022-07-25
NPM logo color won't change


gxlpes:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

I have some shields.io badges in my GitHub profile that I like to be kept with a white color, but the NPM won't accept the white color or any other color. 

The `logoColor= `method is not working at all.

My line of code inside of a GitHub README looks like this:
`![NPM](https://img.shields.io/badge/NPM-2D3136.svg?style=for-the-badge&logo=npm&logoColor=white)`

### 🔗 Link to the badge

https://img.shields.io/badge/NPM-2D3136.svg?style=for-the-badge&logo=npm&logoColor=white

PaulaBarszcz:
Hi @gxlpes,

I only just started looking into Shields but I think that I might be able to help :)

You're right, the `logoColor` doesn't seem to work with npm logo. It is probably because for npm the custom logo is used in Shields.
Related issue:
https://github.com/badges/shields/issues/6208#issuecomment-782718510


You can use a white npm logo via base64-encoded custom svg:

https://img.shields.io/npm/v/npm.svg?logo=data:image/svg%2bxml;base64,PHN2ZyByb2xlPSJpbWciIHZpZXdCb3g9IjAgMCAyNCAyNCIgZmlsbD0id2hpdGUiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyI+PHRpdGxlPm5wbTwvdGl0bGU+PHBhdGggZD0iTTEuNzYzIDBDLjc4NiAwIDAgLjc4NiAwIDEuNzYzdjIwLjQ3NEMwIDIzLjIxNC43ODYgMjQgMS43NjMgMjRoMjAuNDc0Yy45NzcgMCAxLjc2My0uNzg2IDEuNzYzLTEuNzYzVjEuNzYzQzI0IC43ODYgMjMuMjE0IDAgMjIuMjM3IDB6TTUuMTMgNS4zMjNsMTMuODM3LjAxOS0uMDA5IDEzLjgzNmgtMy40NjRsLjAxLTEwLjM4MmgtMy40NTZMMTIuMDQgMTkuMTdINS4xMTN6Ii8+PC9zdmc+

[...]

Issue #8139, opened 2022-06-29
Gitlab badge. not respecting logoColor


davidmpaz:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Hi,

When specifying **logoColor** for the icon, this is not taken into account. Notice please the difference with the one for Github:

Github
https://img.shields.io/badge/CICD-Github-informational?style=flat&logo=github&logoColor=white&color=FFFFFF
![CICD][1]

Gitlab
https://img.shields.io/badge/CICD-Gitlab-informational?style=flat&logo=gitlab&logoColor=white&color=FFFFFF
![CICD][2]

Many thanks in advance for this project.

Cheers,
David

[1]: https://img.shields.io/badge/CICD-Github-informational?style=flat&logo=github&logoColor=white&color=FFFFFF 
[2]: https://img.shields.io/badge/CICD-Gitlab-informational?style=flat&logo=gitlab&logoColor=white&color=FFFFFF 

### 🔗 Link to the badge

https://img.shields.io/badge/CICD-Gitlab-informational?style=flat&logo=gitlab&logoColor=white&color=FFFFFF

chris48s:
update: https://github.com/badges/shields/issues/6208#issuecomment-1328051956

Issue #7684, opened 2022-03-05
Simple Icons or custom Shields logos?


PyvesB:
As you may already be aware, Shields.io supports adding logos to badges by leveraging the [Simple-Icons project](https://github.com/simple-icons/simple-icons). However, it still maintains a small set of custom icons, which can be found in [this directory](https://github.com/badges/shields/tree/6e803367e0f3c8e0cc4196a700af37fac1629f4d/logo). We decided to keep these as they were deemed nicer than the Simple Icons versions at the time (see #2510 for the full history).

In #6208, we've been discussing whether we should get rid of the remaining icons, and fully rely on Simple Icons. Reasons for considering a removal include the following:
* many Simple Icons have improved. Their 2022 versions arguably look better than the ones that had led us to prefer our custom logos back in 2018.
* we're keen on simplifying things on the Shields.io side, in terms of the code on the startup path that handles logo loading and in terms of codebase maintenance. My estimate is that we'd be able to shred a few dozen lines of bespoke code.
* we want to make things easier and reduce confusion for users. In particular, one caveat is that our custom icons do not work in conjunction with the `logoColor` query parameter. Users occasionally open issues about this (e.g. #6208, #7576).
* users can still fall back to the old custom logos by using our `logo` parameter with a base64-encoded image.

However:
* some of our custom logos still arguably look better than the 2022 Simple Icons ones.
* retaining the ability to add new custom logos that are different or don't exist in Simple Icons might come in handy at some point in the future.
* our custom Travis icon is too large and cannot be used as is as a base64-encoded image.

The main point of debate is aesthetics, and it's a highly subjective matter. If you have badges that use one of our custom logos listed below, we're keen on hearing what you think! For each logo that I'll be posting in follow up messages, either add 🎉 if you prefer the custom Shields.io version, a 🚀 if you prefer the Simple Icons version, or 👀 if you're undecided. If you don't use or intend on ever using some of the listed logos, you can simply skip voting on those.

A logo with a majority of votes in favour of the Simple-Icons version will be considered for removal from the Shields.io codebase. If there is a significant drive towards Simple-Icons for several logos, we'll discuss whether visual concerns for the remaining ones outweigh the gains that could be achieved by a full removal. We'll examine the votes carefully to inform our decisions, but the maintainer team will be responsible for making the final call.

PyvesB:
| Logo                 | Shields.io 🎉        | Simple-Icons 🚀  |
| -----------------|:--------------------|:--------------------|
[...]

Issue #7576, opened 2022-02-05
'Telegram' logo does not respect logoColor value


Welding-Torch:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

So basically, while I was trying to fancy up my GitHub readme with some badges, I added Instagram first and then I added Telegram. Now I realised that I could use the `logoColor` property on Instagram (for example `https://img.shields.io/badge/-Instagram-red?color=DD2A7B&logo=instagram&logoColor=white` to make the Instagram logo become white. But when I tried to make the Telegram logo white, I realised that **the Telegram logo remains blue regardless of logoColor**. 

This is a bit of an annoying problem seeing as how I'm trying to give the badges in my ReadMe a consistent look, and the Telegram logo sticks out like a **sore thumb**.

### 🔗 Link to the badge

https://simpleicons.org/icons/telegram.svg

And for reference help: https://simpleicons.org/icons/instagram.svg

Images for visual help:
![image](https://user-images.githubusercontent.com/46340124/152658530-81bf7415-e00d-4461-9081-b4e7235a05be.png)

![image](https://user-images.githubusercontent.com/46340124/152658772-1167df3e-01f5-4252-8423-0318d2606939.png)

The code for the image directly above is:
```
![Telegram](https://img.shields.io/badge/-telegram-red?color=white&logo=telegram&logoColor=black)
![Instagram](https://img.shields.io/badge/-Instagram-red?color=white&logo=instagram&logoColor=black)
```
---
**As you can see, while the logo color for the Instagram logo changes, the logo color for Telegram does not.** 

### 💡 Possible Solution

Possible solution: Talk to the people who run the Simple Icons API and let them know about this problem and ask them to fix it.

calebcartwright:
That's by design as we use a custom logo for telegram that takes preference over the SI reference. I believe the underlying calculus, which led us to keep telegram as one of the handful of custom logos we maintain, remains unchanged so I don't anticipate changing anything on our side.

In this case I think your best bet is to get the logo you want (via SI or otherwise) and adjust the color as needed, and then leverage our [custom logo](https://github.com/badges/shields/blob/master/doc/logos.md#custom-logos) feature to embed that version in your badge 

https://github.com/badges/shields/blob/master/doc/logos.md
https://github.com/badges/shields/issues/2510
https://github.com/badges/shields/pull/2880

PyvesB:
This is by and large a duplicate of #6208, shall we continue the discussion there?

[...]

PR #8263, opened 2022-07-29
fix: support logoColor to shield icons.

### LINKED ISSUE #6208: logoColor is ignored for specific icons

- [ ] My own instance

:beetle: **Description**
Some specific logos/icons seem to ignore the `logoColor` override when applied.
In my case does this happen with the `paypal` logo.

The SVG icon provided by SimpleIcons seems to be a single-coloured SVG.

- https://github.com/simple-icons/simple-icons/blob/develop/icons/paypal.svg
- https://simple-icons.github.io/simple-icons-website/icons/paypal.svg

I can't tell if the icon actually is single-coloured tho, or if either shields.io or SimpleIcons override this default SVG colouring when requested.

:link: **Link to the badge**

Example:
https://img.shields.io/badge/Donate%20with-PayPal-blue?logo=paypal&logoColor=white

Result:
badge

:bulb: **Possible Solution**

If any override is used to alter the colour, or if the PayPal icon is actually a different one (Not from SimpleIcons), should it either respect the logoColor option, or not have the override at all.

If this is a SimpleIcons bug, let me know and I'll report it to them.

regseb:
This pull request fixes #6208 (and its duplicates: #7576, #8139, #8248) and finds a compromise for [_Simple Icons or custom Shields logos?_ #7684](https://github.com/badges/shields/issues/7684).

- Without `logoColor` query parameter: display [custom Shields logo](https://github.com/badges/shields/tree/22380c5e9435e2a59e236d2a58ff55d1db7c8f1e/logo) multiple colors;
- With `logoColor` query parameter: display [Simple Icon](https://github.com/simple-icons/simple-icons) by changing the color.

If you want:

- multiple custom Shields logo multiple colors: `https://img.shields.io/badge/shields-gitlab-blue?logo=gitlab`
- orange Simple Icon with the default color (you must specify color): `https://img.shields.io/badge/shields-gitlab-blue?logo=gitlab&logoColor=FC6D26`
- white Simple Icon with specific color: `https://img.shields.io/badge/shields-gitlab-blue?logo=gitlab&logoColor=white`

There is one exception for the [Dependabot logo](https://github.com/badges/shields/blob/cf7b76d5a68f3edf0d5c4c929b3fceed0f046c8e/logo/dependabot.svg). The custom Shields logo version is monochrome and its color is customizable. With `logoColor` query parameter, it is the custom Shields logo that is displayed by changing its color.


chris48s:
I haven't been massively following all of the discussion in https://github.com/badges/shields/issues/7684 but I think I'd be fine with this as a solution: Seems like a reasonable way to reduce our support burden without taking on much additional complexity (which are really the 2 things I think we want to optimise for here).

The one issue is `travis` (because in simpleicons the slug is `travis-ci`).

calebcartwright:
> I haven't been massively following all of the discussion in #7684 but I think I'd be fine with this as a solution: Seems like a reasonable way to reduce our support burden without taking on much additional complexity (which are really the 2 things I think we want to optimise for here).
> 
> The one issue is `travis` (because in simpleicons the slug is `travis-ci`).

IMO this isn't _directly_ related to #7684 and the fundamental question of whether users still want the custom logos. I view this as simply adding new functionality to Shields to give a minor assist to users that want to use the SI version instead of one of the corresponding ten custom logos we have, without having to base64 encode the SI icon themselves to be used with `?customLogo`. I realize we could use this to decide to settle the matter since both camps could rely on us to do their logo work for them, but just want to underscore that we could still drop the custom logos before or even after potentially making a change like this.

There's also the potential for a small slice of our existing users to be impacted by a change like this, if they happen to have a `logoColor` param in their badge urls. That's most likely a fairly small portion given that param is a functional no-op today, but it's still the case that the same badge url would produce a different badge and potentially require unhappy users to update their badge urls. Obviously that doesn't need to be treated as a hard blocker, but it's something to be cognizant of, and transparent about, if we were to move forward.

chris48s:
> IMO this isn't directly related to https://github.com/badges/shields/issues/7684

Yeah - sorry. I think @regseb 's description:

> This pull request fixes https://github.com/badges/shields/issues/6208 (and its duplicates: https://github.com/badges/shields/issues/7576, https://github.com/badges/shields/issues/8139, https://github.com/badges/shields/issues/8248) and finds a compromise for https://github.com/badges/shields/issues/7684.

is better.

> but I just wanted to also underscore that we could still drop the custom logos before or even after potentially making a change like this

Agreed. We could. Tbh though, as much as I'd like to bin them all and simplify things, I think #7684 tells us there isn't really a clear consensus for it.

---

Given the point we're starting from, It seems like merging some variation of this solution (accepting it probably needs an additional special-case for the travis logo) would:

- reduce support requests/confusion like in the linked issues
- not add much additional complexity to the code
- keep users who like the custom logos happy
- keep users who want to use `logoColor=` happy
- maybe require some existing users to remove an unwanted/spurious `logoColor` param from some existing badge URLs
- not make anything else worse than it already is (I think?)

which seems like a pretty decent set of tradeoffs to me, all things considered.

calebcartwright:
I think I largely agree, though do you think it'd be better to shift the discussion over to #6208 or to at least recap? Both so that those involved/subscribed to that convo can see this articulated position and for posterity so that thread will contain all relevant parts?

regseb:
@chris48s @calebcartwright Is there anything I should do to advance this pull request?

chris48s:
I think the thing that stalled this is not "the code is bad and it needs changes", its more "do we agree this idea is the right solution?". Personally, I think yes it is.

@calebcartwright - I'm going to invoke your suggestion about making decisions here and say: If you want to say "no its not", object in the next week.

@regseb - In terms of the code, the one thing we'd need to do is add a special case to fix travis. This one breaks because our custom logo is `travis` and in simpleicons the slug is `travis-ci` (`?logo=travis&logoColor=red` show no logo at all). If you want to do it straight away, fine but if you want to hold off on making that change then that's reasonable. I don't like to see people spend more time than needed on a PR that doesn't get merged.

@SnoozeThis wait 7 days

SnoozeThis:
(https://snoozeth.is/bRTxKqHPz6o) I will wait until Tue, 29 Nov 2022 16:50:08 UTC and then add a comment.

regseb:
I handled the Travis case.

calebcartwright:
> @calebcartwright - I'm going to invoke your suggestion about making decisions here and say: If you want to say "no its not", object in the next week.

Agreed. Hopefully already evident from the above thread, but I'm not personally going to object. I do think the code will need another pass through (and I can't promise I'll have a chance to do so particularly given the holiday this week in the US), and would also need some accompanying items e.g. solid documentation, both in readme docs and on the site, and a communication plan

chris48s:
The code is fine so far and I'm happy to review any further commits.

Documentation is a good shout - thanks :+1:

At the moment we don't mention anything about `logoColor` in https://github.com/badges/shields/blob/master/doc/logos.md so I am fine with leaving that as it is.

We do need to update

https://github.com/badges/shields/blob/c5476b4dd506e4b205311c4a3e1cbc0d10932e9f/frontend/components/usage.tsx#L362-L372

and

https://github.com/badges/shields/blob/c5476b4dd506e4b205311c4a3e1cbc0d10932e9f/frontend/pages/endpoint.tsx#L210-L214

Also the comment in

https://github.com/badges/shields/blob/c5476b4dd506e4b205311c4a3e1cbc0d10932e9f/core/base-service/coalesce-badge.js#L9-L34

could do with an update.

In terms of communicating the change, I think posting a comment in the various linked issues when we merge this and close them will be fine. https://github.com/badges/shields/issues/7684 is already pinned and it can stay pinned. I've not drafted a post yet, but I'm happy to cover that.

regseb:
I have updated the documentation. English is not my first language, so I'm open to any changes.

chris48s (approved):
Going to merge this with a 1 word tweak to the docs. I'll update the issue threads once I've deployed

SnoozeThis:
Resolved.

PR #8270, opened 2022-08-02
Add OSSF Scorecard badge

azeemshaikh38:
This PR adds the OSSF Scorecard badge for badges/shields. The `ossf/scorecard-action` GitHub Action is needed to enable this badge since the result published by this action is used to update the badge result in a timely manner.

`badges/shields` repo is rated quite high in the [OSSF criticality_score](https://github.com/ossf/criticality_score) (criticality rate of 0.7 out of 1). So it would be great to have this repo enabled and running Scorecard.

Note to reviewers: the GitHub action steps which upload the SARIF results are totally optional and not needed for enabling the badge. It'll be useful to enable them because Scorecard will provide code scanning alerts for any issues it finds on the repository. 

PS: the Scorecard badge is powered by shields.io :)

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -3,10 +3,10 @@ on:
   push:
     # Only the default branch is supported.
     branches:
-    - master
+      - master
   schedule:
     # Weekly on Saturdays.
-    - cron:  '30 1 * * 6'
+    - cron: '30 1 * * 6'
 
 permissions: read-all
 
@@ -18,28 +18,28 @@ jobs:
       security-events: write
       id-token: write
     steps:
-      - name: "Checkout code"
+      - name: 'Checkout code'
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
-      - name: "Run analysis"
+      - name: 'Run analysis'
         uses: ossf/scorecard-action@3155d134e59d8f47261b1ae9d143034c69572227 # v2.0.0-beta.1
         with:
           results_file: results.sarif
           results_format: sarif
           repo_token: ${{ secrets.GITHUB_TOKEN }}
           # Publish the results for public repositories to enable scorecard badges. For more details, see
-          # https://github.com/ossf/scorecard-action#publishing-results. 
-          # For private repositories, `publish_results` will automatically be set to `false`, regardless 
+          # https://github.com/ossf/scorecard-action#publishing-results.
+          # For private repositories, `publish_results` will automatically be set to `false`, regardless
           # of the value entered here.
           publish_results: true
       # https://docs.github.com/en/actions/advanced-guides/storing-workflow-data-as-artifacts
       # Optional.
-      - name: "Upload artifact"
+      - name: 'Upload artifact'
         uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v2
         with:
           name: SARIF file
           path: results.sarif
           retention-days: 5
-      - name: "Upload SARIF results"
+      - name: 'Upload SARIF results'
         uses: github/codeql-action/upload-sarif@f5d822707ee6e8fb81b04a5c0040b736da22e587 # v1
         with:
           sarif_file: results.sarif
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -1,0 +1,45 @@
+name: OSSF Scorecard
+on:
+  push:
+    # Only the default branch is supported.
+    branches:
+    - master
+  schedule:
+    # Weekly on Saturdays.
+    - cron:  '30 1 * * 6'
+
+permissions: read-all
+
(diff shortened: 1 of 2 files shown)


chris48s:
Personally I don't think we should merge this, but I'll leave it open in case anyone strongly wants it. If anything we've already got too many badges in our README and I don't think this adds much.

My slightly spicy take :hot_pepper: on the OSSF criticality score (that nobody asked for :laughing: )  is: The fact shields scores quite highly is a good indicator that this is not a massively useful number. I think it is probably placing too much weight on "popularity" and "activity" metrics and not enough weight on the project's position in the supply chain or and size/nature of the install base. This is at least partially a function of what is easy/hard to measure :slightly_smiling_face: 

I do however wish you luck with the project and it is great that you have been able to use shields (either as a library or service - slightly unclear which) to provide users with a badge.

azeemshaikh38:
> Personally I don't think we should merge this, but I'll leave it open in case anyone strongly wants it. If anything we've already got too many badges in our README and I don't think this adds much.
> 
> My slightly spicy take 🌶️ on the OSSF criticality score (that nobody asked for 😆 ) is: The fact shields scores quite highly is a good indicator that this is not a massively useful number. I think it is probably placing too much weight on "popularity" and "activity" metrics and not enough weight on the project's position in the supply chain or and size/nature of the install base. This is at least partially a function of what is easy/hard to measure 🙂
> 
> I do however wish you luck with the project and it is great that you have been able to use shields (either as a library or service - slightly unclear which) to provide users with a badge.

Thanks for your input @chris48s. I agree that OSSF criticality_score places more weight on "popularity" instead of the project's position in supply chain, but I'd argue that `badges/shields` is actually quite critical to OSS :)  Seeing a popular project like `badges/shields` adopt some of these security practices is a great way to encourage other OSS developers to do the same. That's kind of what this PR is hoping to achieve. It may feel like adding the badge to the README doesn't add much, but I strongly believe this will give us (OSSF Scorecard) a lot of visibility within OSS community and hopefully drive further adoption. I'll leave it to the maintainers to either merge or close this :) 

PS: we used shields as a service.

calebcartwright:
I agree with Chris' conclusion that we shouldn't add this badge, though my reasoning is somewhat different. Additionally since the door has been opened, I'll share my take on the criticality score too :grin: 

First and foremost, my rationale for not adding this badge is simply that it's not a service that we directly utilize for the assets developed/maintained in this project. There's a large number of badges we could theoretically add to display some analysis result of our npm package, website, etc., for example CodeClimate, SonarCloud, npms.io, Mozilla Observatory, etc., but we don't add such badges to our own repo solely for the purposes of demonstrating or promoting a service. If we were to start down that path we would either end up having to add an absurd number of badges or we have to pick and choose favorites; neither of which is viable.

As to OSSF and the criticality score more generally, I'm highly appreciate of all the OSSF efforts underway to identify important projects and to help support/secure/ensure those projects in some shape or form. However, I think a disconnect comes into play because all of the OSSF announcements and marketing really stress the security and supply chain efforts which are of course crucial, but only really applicable to projects that are directly consumed as part of a dependency graph or installed/deployed (e.g. kubernetes).

I would absolutely agree that that Shields.io has a massive reach and is a highly impactful service, but I don't personally view it as critical to a functioning internet nor secure development supply chain. I'd also agree that status badges have essentially become ubiquitous in development projects these days, but it's also true that in practice the worst thing that happens if we go offline is that we have lots of fussy devs who have broken images in project readme pages.

If OSSF's intent is to _also_ identify open source projects/services that play a large role in modern development practices and experiences then that's great too, but tossing such projects (like Shields) under the security/supply chain umbrella is a bit of a head scratcher

azeemshaikh38:
> First and foremost, my rationale for not adding this badge is simply that it's not a service that we directly utilize for the assets developed/maintained in this project. There's a large number of badges we could theoretically add to display some analysis result of our npm package, website, etc., for example CodeClimate, SonarCloud, npms.io, Mozilla Observatory, etc., but we don't add such badges to our own repo solely for the purposes of demonstrating or promoting a service. If we were to start down that path we would either end up having to add an absurd number of badges or we have to pick and choose favorites; neither of which is viable.

That's completely fair and I understand if you choose not to go forward with this.

> If OSSF's intent is to _also_ identify open source projects/services that play a large role in modern development practices and experiences then that's great too, but tossing such projects (like Shields) under the security/supply chain umbrella is a bit of a head scratcher

To clarify, I wasn't justifying criticality_score itself - its well known within OSSF circles that we rely too much upon popularity metrics for criticality_score and there are efforts underway to improve this (feedback or ideas very much welcome and appreciated :) ). My point simply being that shields is a highly popular project (something we all seem to agree) and its helpful for OSSF as an org to work/collaborate with maintainers of such projects.

calebcartwright:
Understood. Thanks for the suggestion and dialog @azeemshaikh38 but think it makes sense to close.

PR #8271, opened 2022-08-02
Fix [ROS] support for repos with multiple packages

jtbandes:
Follow up to #8169 to fix an issue reported in https://discourse.ros.org/t/shields-io-badge-service/26379/3. Some repositories have multiple packages, in which case a `packages` array is present. We need to search these arrays for the requested package name, not just the keys under `repositories`.

cc @ijnek @cst0


ijnek:
I'm not familiar enough with JS and this project to do a proper review, but the changes look like they address the exact isssue, thanks @jtbandes!

calebcartwright:
I realize there's an unusual level of complexity here given the subsequent schema validation, but I'm having trouble following what the expected behavior is relative to the schema shared in the tests.

From what I can tell if a repository has multiple packages, all the packages will have the same version as dictated by the `release` key on the repository?

Are package names guaranteed unique across the system or is there a namescoping function of sorts wherein the package names are only unique within a given repository?

cst0:
What it mainly comes down to is that packages and repositories are different things in the ROS ecosystem. It's very common to see one repository hosting multiple interconnected packages: for example, the standard robot navigation stack is made up of several packages (global path planning, local path planning, costmaps, recovery mechanisms) which each share a single repository ([ros-planning/navigation](https://github.com/ros-planning/navigation)). It's also common to see an entire robot stack sharing a repository (e.g., the [Fetch](https://github.com/fetchrobotics/fetch_ros.git)) to allow installation of single packages while keeping the build-from-source overhead low. 

> From what I can tell if a repository has multiple packages, all the packages will have the same version as dictated by the release key on the repository?

Releases of a package are determined by the `package.xml` found within each package. I've only ever seen that correspond to a git tag across the repository, but the tag isn't what makes a version and you could probably produce a repository with multiple package versions. I've never tried that myself though.

However, what is very common is different versions targeting different ROS distributions. For example, I have a repository with one package, released for ROS kinetic and ROS melodic. Kinetic is EOL so it's no longer allowing updates, but Melodic is actively accepting contributions. So these will have different version numbers (despite being the same repository). So that would be a case where there are different versions despite one relevant tag. This case is already covered in the badge though, since it takes a ROS distribution as a parameter.

> Are package names guaranteed unique across the system

Yes. Packages are made available via the OSRF's apt repository, so duplicate names would be rejected. However, #8169 takes a string pointing to a repository: if I create my own "navigation" repository with zero package name overlap, it wouldn't break anything ROS-side. What would the shield behavior be?

My question would be: is a badge intended to represent a package, or a repository? Badges in #8169 take in a package name, but they really look for a repository which happens to have the same name as the package. Provide an actual package and the badge breaks. This PR appears to fix that problem by now looking explicitly at a package. The alternative solution would be to point at a repository by looking for a full URL (note that not all ROS packages are on GitHub). But, this is nonideal partly for the previously-mentioned cases.

calebcartwright:
Thanks @cst0

> Releases of a package are determined by the package.xml found within each package

If there's a desire to provide package-level version badges, then this would be the data source we would want to go after, and I still don't see that happening.

As previously submitted in #8169, we're driving everything based off the contents of a yaml file retrieved using a tag, and afaict that yaml (or at least the parts of it that have been shared here with us non-ROS users) only includes versions for a repository object. The adjustment proposed in this PR seems to simply be allowing the _parameter_ to be a package name instead of the repository name, but the result is still the version of the repository.

Assuming that's correct, I'd suggest we repurpose the existing badge (even if it's just name changes) to be explicit that it's a repository version badge, if such a thing is valuable for your community. If there's then a desire for package version badges, then we'd need to have a separate discussion around how Shields would be able to go about getting the version data from the correct, authoritative location (wherever the package.xml files are)

cst0:
The yaml file you're referring to is already informed by the `package.xml` files within a package. When a package is released in the ROS ecosystem, the final step is to list it as available-for-installation by making a pull request in the relevant yaml file. This is all an automated process (via the `bloom` tool).

calebcartwright:
Perhaps I'm missing something.

If there's any documentation that covers all of this that would be most helpful, but I'd never heard of your service until #8169, so that PR plus this PR is the totality of information I have to work with. We integrate with hundreds of services and don't really need to understand much about them beyond their API surface that provide the data point(s) we need to retrieve the values we plug into the badge text.

> My question would be: is a badge intended to represent a package, or a repository? 

I'm not sure who this question was directed to, but I hope it isn't me/the Shields team. We can't answer this question,and it's the type of thing we ask when people want us to provide badges for some service/platform they use.

From what I can tell, the only version data point that's available in the data source that's been shared with us is the version key for a repository release in the yaml file. For example, the test added here as a demonstration of getting a version for a package:

https://github.com/badges/shields/pull/8271/files#diff-6b1a75d06bae90b956da0e6dafbd6d99f53158fb9bbfd4be8f2d657fb94eb195R33-R52

```yaml
  navigation_2d:
    doc:
      type: git
      url: https://github.com/skasperski/navigation_2d.git
      version: noetic
    release:
      packages:
      - nav2d
      - nav2d_exploration
      - nav2d_karto
      - nav2d_localizer
      - nav2d_msgs
      - nav2d_navigator
      - nav2d_operator
      - nav2d_remote
      - nav2d_tutorials
      tags:
        release: release/noetic/{package}/{version}
      url: https://github.com/skasperski/navigation_2d-release.git
      version: 0.4.3-1
    source:
      type: git
      url: https://github.com/skasperski/navigation_2d.git
      version: noetic
    status: maintained
```

where, as defined in that test yaml snippet, `navigation_2d` is the repository. It's `release` key contains multiple packages, but there's just one `version`

The point I'm driving at is that we only see that singular `version` key so that's the only data point we can use for a badge. It's still unclear to me whether individual packages are versioned independently, but regardless, we can only work with the data that's available and the yaml file is the only one that's been shared with us so far.

As such, it seems to me that it's only possible for us to display a repository version badge. Was the goal/expectation of these changes simply to be able to use a package name instead of the repository name in the badge url, but still end up with the exact same badge (the `0.4.3` version from the example above)?

If so then I'd most likely be opposed to these changes. We've already got a fairly complex and time sensitive process (sequential network calls with processing) just to get to the correct yaml file, and these changes would only add more work we have to perform as part of each and every request. It also adds even more complexity to the codebase, and the only benefit I could see is to provide a minor convenience to users for a one time action (i.e. create badge url, copy/paste into readme). Those tradeoffs aren't worth it in IMO, and I'd much rather just rebrand things on our badge to make it clear it's working with repositories, not packages.

However, if I've gotten something wrong and packages can/do have versions that are divergent from the aforementioned key, and you/ROS users would like such a badge, then we will need more information about where that data can be found.

jtbandes:
> I'd much rather just rebrand things on our badge to make it clear it's working with repositories, not packages.

This would make sense to me. I originally used the package terminology because packages are what I’m used to working with, but after following this discussion and considering how the badges are meant to be used on repos, and packages in a repo share the same version, I think it’s probably fine to keep the current behavior. @cst0 and @ijnek, do you think it makes sense for the badges to represent a whole repository as they do today, if we just change the wording to correct my misuse of “package”?

ijnek:
> do you think it makes sense for the badges to represent a whole repository as they do today, if we just change the wording to correct my misuse of “package”?

Yes, that makes sense.

jtbandes:
> The alternative solution would be to point at a repository by looking for a full URL (note that not all ROS packages are on GitHub). But, this is nonideal partly for the previously-mentioned cases.

@cst0 could you clarify why a full URL would be needed, rather than just a repository name as used by the current implementation?

cst0:
> could you clarify why a full URL would be needed, rather than just a repository name

Sorry if I'm introducing confusion, I'm not familiar enough with this project, language, or implementation to provide clear insight into that side of things. What I was referring to was how ROS doesn't restrict to github vs. any other version control host, so specifying (for example) "ros-perception/[slam_gmapping](https://github.com/ros-perception/slam_gmapping)" wouldn't be suitable.

So whatever final solution is chosen would need to account for that, but it seems like you already have a solution to that:

> do you think it makes sense for the badges to represent a whole repository as they do today, if we just change the wording to correct my misuse of “package”?

As far as I can tell, yeah.

jtbandes:
But since the data source we are using is the rosdistro yaml file, isn’t the name `slam_gmapping` alone enough?

jtbandes:
Closing in favor of https://github.com/badges/shields/pull/8292

Issue #7884, opened 2022-04-25
GitHub Gist "last commit" Badge


smeinecke:
:clipboard: **Description**

A badge for the latest update on a GitHub gist would be nice.
Right now it's possible to add a custom badge that shows the full date, but that does not change color based on freshness or shows a short DateTime.

Custom Bade example: 
![GitHub last update](https://img.shields.io/badge/dynamic/json?color=lightgray&style=flat&label=last%20update&query=%24.updated_at&url=https%3A%2F%2Fapi.github.com%2Fgists%2F4401361&cacheSeconds=86400)

Normal "Latest commit" badge:
![GitHub last commit](https://img.shields.io/github/last-commit/disposable/static-disposable-lists)

:link: **Data**

Github API: https://docs.github.com/en/rest/gists/gists#get-a-gist
`updated_at` field.

:microphone: **Motivation**

Would like to present the "freshness" of a list as it gets updated every now and then:

https://github.com/disposable/disposable#external-sources

PaulaBarszcz:
It seems that it already works:

`![1.3.0](https://img.shields.io/github/last-commit/badges/shields/1.3.0)`
![1.3.0](https://img.shields.io/github/last-commit/badges/shields/1.3.0)

--
`![master](https://img.shields.io/github/last-commit/badges/shields/master)`
![master](https://img.shields.io/github/last-commit/badges/shields/master)

--
`![integration-docs](https://img.shields.io/github/last-commit/badges/shields/integration-docs)`
![integration-docs](https://img.shields.io/github/last-commit/badges/shields/integration-docs)

--
`![server-2021-10-04](https://img.shields.io/github/last-commit/badges/shields/server-2021-10-04)`
![server-2021-10-04](https://img.shields.io/github/last-commit/badges/shields/server-2021-10-04)


Print screen for posteriority:
![image](https://user-images.githubusercontent.com/31891675/181460591-45c5cc4c-cacc-4015-9b2c-aaf4b7459352.png)


@chris48s  Can this issue be closed? :)

chris48s:
The examples you've posted there are all last commit for a _repo_ . This feature request is last commit for a _gist_ (which is a different thing)

PaulaBarszcz:
True @chris48s, thanks for pointing it out!
I would like to work on this issue :)

chris48s:
Feel free. All the info needed is in the top post. Thanks

PaulaBarszcz:
[...]

PR #8272, opened 2022-08-02
[GithubGistLastCommit] GitHub gist last commit

PaulaBarszcz:
Related issue: #7884 

Locally looks like this (updated with the new path, `{ base: 'github-gist/last-commit', pattern: ':gistId' }`):
image

Print screens of example gists generated today (3rd of August 2022):
https://api.github.com/gists/8710649 
gistId: 8710649
 "updated_at": "2022-08-03T02:58:04Z",
image

https://api.github.com/gists/870071abadfd66a28bf539677332f12b
gistId: 870071abadfd66a28bf539677332f12b
  "updated_at": "2020-10-01T13:50:24Z",
  image

https://api.github.com/gists/7e188c35fd5ca754c970e3a1caf045ef
gistId: 7e188c35fd5ca754c970e3a1caf045ef
  "updated_at": "2020-07-27T17:28:45Z",
image

https://api.github.com/gists/871064 
gistId: 871064
 "updated_at": "2015-09-25T05:28:06Z",
2015 Sept

https://api.github.com/gists/55555555555555 
gistId: 55555555555555  
"message": "Not Found",
image

---
I confirmed locally that it still works after the changes following the review:

image


PaulaBarszcz:
The github-gist-last-commit service development is done- but I have issues with tests.

If I add `.json` extension to the call’s url, the call fails.
If I don’t add it, I receive an SVG file in a response.

calebcartwright:
@PaulaBarszcz - the one uncommented test you have works fine with when the `.json` extension is added (the `last commit in gist (ancient)` test). The issues you mentioned offline with the others is probably because there is an unnecessary leading `/gists` in them. The way our `createServiceTester` works is that it handles determining and incorporating the base part of the badge route (the part you've currently got set to `github/gists` in the service classes) automatically, so the only parts that are necessary to specify in the test route are the paramters

PaulaBarszcz:
> @PaulaBarszcz - the one uncommented test you have works fine with when the `.json` extension is added (the `last commit in gist (ancient)` test). The issues you mentioned offline with the others is probably because there is an unnecessary leading `/gists` in them. The way our `createServiceTester` works is that it handles determining and incorporating the base part of the badge route (the part you've currently got set to `github/gists` in the service classes) automatically, so the only parts that are necessary to specify in the test route are the paramters

Thanks, the fixed tests are pushed  :)

Now I'm trying to figure out where can I check the available names for the test suites to put in the PR's title.
In this PR should I put `[GitHub]`, `[GitHubActivity]`, `[GitHubCommitActivity]` or something else? :)

services/github/github-gist-last-commit.service.js
```diff
@@ -0,0 +1,47 @@
...
+import { GithubAuthV3Service } from './github-auth-service.js'
+import { documentation, errorMessagesFor } from './github-helpers.js'
+
+const schema = Joi.object({
+  updated_at: Joi.string().required(),
+}).required()
+
+export default class GithubGistLastCommit extends GithubAuthV3Service {
+  static category = 'activity'
+  static route = { base: 'github/gists', pattern: ':gistId' }
```
calebcartwright, line 13: We'll need to adjust the route here though I'm not entirely sure to what just yet.

We've already got a few more requests for gist-related badges, and our [docs on badge urls](https://github.com/badges/shields/blob/master/doc/badge-urls.md) prescribe a pattern we'll need to follow.

I suspect we'll probably want to do something like:

```suggestion
  static route = { base: 'github-gist/last-commit', pattern: ':gistId' }
```

Though I suppose we _could_ also consider doing a secondary noun, e.g.

```suggestion
  static route = { base: 'github/gist/last-commit', pattern: ':gistId' }
```

I'd suggest the former for now, but we can see what the other maintainers think since whatever we choose here will likely end up setting the standard for gist-related badges

services/github/github-gist-last-commit.service.js
```diff
@@ -0,0 +1,47 @@
...
+    return this._requestJson({
+      url: `/gists/${gistId}`,
+      schema,
+      errorMessages: errorMessagesFor('gist not found'),
+    })
+  }
+
+  async handle({ gistId }) {
+    const body = await this.fetch({ gistId })
+    return this.constructor.render({ commitDate: body.updated_at })
```
calebcartwright, line 45: This is perfectly fine but just note for future reference that destructuring may be an easier way of extracting+renaming, e.g.

```suggestion
    const { updated_at: commitDate } = await this.fetch({ gistId })
    return this.constructor.render({ commitDate })
```

Thought it's mostly stylistic

services/github/github-gist-last-commit.tester.js
```diff
@@ -0,0 +1,36 @@
...
+  .get('/7e188c35fd5ca754c970e3a1caf045ef.json')
+  .intercept(nock =>
+    nock('https://api.github.com')
+      .get('/gists/7e188c35fd5ca754c970e3a1caf045ef')
+      .reply(200, {
+        updated_at: dayjs(),
+      })
+  )
+  .expectBadge({ label: 'last commit', message: 'today', color: 'brightgreen' })
+
```
calebcartwright, line 15: I hate to mention this since I'm sure you had fun battling nock :smile: but I think we should drop this test.

We mostly use our service tests as an integration and e2e style test suite to both give us some high level assurances about the service and routes, but particularly the touch points with the hundreds of backend APIs we integrate with (and which can change without notice).

In cases where we have service class with non trivial `render` or `transform` functions we will add more standard unit tests to validate those, but we typically refrain from doing so when those are simple functions or just using some of our standard helpers like this one is (those helpers are extensively tested elsewhere).

calebcartwright (commented):
Thank you! This is looking pretty good, few minor things left inline for your review

services/github/github-gist-last-commit.tester.js
PaulaBarszcz: > I hate to mention this since I'm sure you had fun battling nock 😄 but I think we should drop this test.

Haha, a little bit :D

Sure, removed this test :)

services/github/github-gist-last-commit.service.js
PaulaBarszcz: Cool, I applied the first suggestion.
If you and the other maintainers decide to go with the second option, I will update it accordingly :)

services/github/github-gist-last-commit.service.js
PaulaBarszcz: I didn't know this way of destructuring and concurrent renaming, very neat - thanks, updated ;)

PaulaBarszcz:
Since we're changing the route to be `github-gist/last-commit`,
should we also update the file names?

Currently we have:

image

PaulaBarszcz:
Implementation of the changes is done, and this PR is ready for a re-review :)

calebcartwright:
> Now I'm trying to figure out where can I check the available names for the test suites to put in the PR's title.
In this PR should I put [GitHub], [GitHubActivity], [GitHubCommitActivity] or something else?

You typically only want to run the tests relevant for the code being added or modified. The more generic/higher level the filter the more tests get executed, and the greater the potential for failures/brittleness in tests that weren't actually related to proposed changes. In this case we'd actually want to use `GithubGistLastCommit` to run the tests here instead of the tests for the repo commit badge.

I've updated the title and re-executed the tests accordingly

calebcartwright:
> should we also update the file names?

No, these are fine as-is :+1:

PaulaBarszcz:
> > should we also update the file names?
> 
> No, these are fine as-is 👍

Cool, thanks.

> > Now I'm trying to figure out where can I check the available names for the test suites to put in the PR's title.
> > In this PR should I put [GitHub], [GitHubActivity], [GitHubCommitActivity] or something else?
> 
> You typically only want to run the tests relevant for the code being added or modified. The more generic/higher level the filter the more tests get executed, and the greater the potential for failures/brittleness in tests that weren't actually related to proposed changes. In this case we'd actually want to use `GithubGistLastCommit` to run the tests here instead of the tests for the repo commit badge.
> 
> I've updated the title and re-executed the tests accordingly

Great, thank you.

Should anything else be changed/added in this PR? :)

calebcartwright:
Nope all set. I was just waiting for feedback from others offline relative to the route pattern for github gist badges going forward but we're good to go!

calebcartwright (approved):

Issue #8109, opened 2022-06-19
Circle CI not triggering builds on some pull requests


chris48s:
:clipboard: **Description**

This is absolutely doing my head in and we seem to be seeing it more frequently.

I don't know what the problem is but I have spent the time to write up an issue on the Circle CI support forum. Lets see where it goes.. https://discuss.circleci.com/t/circle-ci-builds-failing-to-trigger-on-some-pull-requests/44472

chris48s:
I raised this as a support request. Here's the response I got

> Thank you for reaching out to CircleCI Support. From the information you've provided, we've determined that this could be hitting a known issue related to permissions on open projects at this time. This scenario has open work in flight and we've linked your case to this internal tracker for further updates in case this does fit after further investigation.
> 
> In the meantime, will you please ask the user to sign in to the CircleCI application and authorize the app so that it appears in the personal account settings in case this helps: https://github.com/settings/applications
> 
> Can you let us know if that particular user has this under those settings and if not, did that help?
> 
> Another scenario where this could happen is where the user is following the project on the fork and not on the source of the project. Has this user's builds ever run? See https://circleci.com/docs/2.0/oss under "Note" or [Why aren't pull requests triggering jobs on my organization?](https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization-?source=search)

I can't find the PR but having read this, I seem to remember someone submitting a PR and saying they had to create an account/log in to CircleCI to get the build to start.. does that ring a bell?

@sunny0826 there might be some advice above or in https://support.circleci.com/hc/en-us/articles/360008097173-Why-aren-t-pull-requests-triggering-jobs-on-my-organization- that might be worth a try next time you open a PR (although you're not the only contributor we've seen have this issue).

There is a part of me also wondering if the time has come to migrate everything off CircleCI and on to Github Actions. We're already using GHA for quite a lot of things. It might improve everyone's contributor experience if we moved the CI builds there too.

calebcartwright:
We'd seen this before with people not being able to view the Circle logs if they didn't have an account, but nothing like what we've seen recently. There was a non-brief period when the jobs weren't triggering against a PR I opened, in this-repo, so at a minimum I think there are (or at least have been) multiple issues at play.

> There is a part of me also wondering if the time has come to migrate everything off CircleCI and on to Github Actions. We're already using GHA for quite a lot of things. It might improve everyone's contributor experience if we moved the CI builds there too.

[...]

PR #8275, opened 2022-08-05
Fix invalid `rst` format when `alt` or `target` is present

abitrolly:


calebcartwright:
Could you elaborate on the problem you're trying to solve with these proposed changes? If you have an example that reproduces the behavior that'd be great

abitrolly:
@calebcartwright I've looked for a test and found none, but I guess I could at least post the output, indeed. 

The badge generated for https://github.com/fschulze/sqlalchemy_schemadisplay/pull/31 was
```
.. image:: https://img.shields.io/pypi/v/sqlalchemy_schemadisplay   :alt: PyPI
```
And that is rendered as a missing image, because GitHub probably translates it to URL. The correct rST markup.
```
.. image:: https://img.shields.io/pypi/v/sqlalchemy_schemadisplay
   :alt: PyPI
```

chris48s:
@calebcartwright This PR does fix a bug. See https://docutils.sourceforge.io/docs/ref/rst/directives.html#image for the correct RST syntax.

@abitrolly The tests for this are in https://github.com/badges/shields/blob/7d452472defa0e0bd71d6443393e522e8457f856/frontend/lib/generate-image-markup.spec.ts#L49-L63 and will need updating

I've closed and re-opened this to get the CI to run. We do have an issue where sometimes the Circle CI builds don't run for particular PRs - there is some info on this at https://github.com/badges/shields/issues/8109 I dunno if any of the info in https://github.com/badges/shields/issues/8109#issuecomment-1169152338 are actionable for you. If not, we'll have to do the same again.

abitrolly:
The `.spec.ts` extension didn't trigger me that it is a file with tests. Looked like some alien typescript thing to me. Or maybe I was browsing my fork, which was too old to include this file.

Anyway, I think I've fixed tests.

abitrolly:
Regarding unblocking `CircleCI`, I don't see this as acceptable for my repos.

image

chris48s (approved):
Merging this having pushed 5260a08 - Lets just merge the fix than do another round of back-and-forth over whitespace. Thanks

abitrolly:
Forgot `\n` at the ends :facepalm: but seems that wouldn't satisfy the linter anyways.

Thanks.

Issue #8238, opened 2022-07-23
Support for multi-arch docker images on size badge


sh-cho:
:clipboard: **Description**

```
![Docker Image Size (latest by date)](https://img.shields.io/docker/image-size/_/mysql/latest)
```
![Docker Image Size (latest by date)](https://img.shields.io/docker/image-size/_/mysql/latest)

For example, `mysql:latest` image supports `["linux/amd64", "linux/arm64/v8"]` os/architecture, but looks like I can't choose which one to use in the badge.

Is there any way to specify docker image's platform?

I can see docker version badge is available to select architecture with `&arch=..` parameter.

chris48s:
For the docker image size badge, we don't have an `arch` param and we also don't explicitly make an assumption about architecture. We just report the value of `full_size`, so if we look at 

https://registry.hub.docker.com/v2/repositories/library/mysql/tags/latest

```json
{
  "creator": 7,
  "id": 20021,
  "images": [
    {
      "architecture": "amd64",
      "features": "",
      "variant": null,
      "digest": "sha256:44f98f4dd825a945d2a6a4b7b2f14127b5d07c5aaa07d9d232c2b58936fb76dc",
      "os": "linux",
      "os_features": "",
      "os_version": null,
      "size": 131789857,
      "status": "active",
      "last_pulled": "2022-07-23T14:07:03.54267Z",
      "last_pushed": "2022-07-13T07:05:12.182358Z"
    },
    {
      "architecture": "arm64",
      "features": "",
      "variant": "v8",
      "digest": "sha256:12cf01a51f803d0ad49ee0dbbb3025a6eef3341e24757c2ed8150b6654c3fb07",
      "os": "linux",
      "os_features": "",
      "os_version": null,
      "size": 138716495,
      "status": "active",
      "last_pulled": "2022-07-23T08:20:42.387497Z",
      "last_pushed": "2022-07-13T08:05:31.345335Z"
    }
  ],
  "last_updated": "2022-07-13T08:05:52.466381Z",
  "last_updater": 1156886,
  "last_updater_username": "doijanky",
  "name": "latest",
  "repository": 21179,
  "full_size": 131789857,
  "v2": true,
  "tag_status": "active",
  "tag_last_pulled": "2022-07-23T14:07:03.54267Z",
  "tag_last_pushed": "2022-07-13T08:05:52.466381Z"
}
```

[...]

PR #8290, opened 2022-08-06
[DockerSize] Docker image size multi arch

PaulaBarszcz:
Related issue: https://github.com/badges/shields/issues/8238

All of my changes are pushed to the feature branch (`docker-imageSize-multiArch`).

The `full_size` is used only if no architecture was specified.

If the user supplied the arch param:
if the image with the given arch was found, we give its size,
if the image with the given arch was not found, we throw an `architecture not found` error.

Some examples below:

## with tag:

https://registry.hub.docker.com/v2/repositories/library/mysql/tags/latest

* incorrect repo name:

http://localhost:8080/docker/image-size/_/mysqppppl

Zrzut ekranu 2022-08-11 o 18 20 23

* incorrect tag:

http://localhost:8080/docker/image-size/_/mysql/nonexistentTag

Zrzut ekranu 2022-08-11 o 18 15 17

* correct arch param supplied:

http://localhost:8080/docker/image-size/_/mysql/latest?arch=amd64

Zrzut ekranu 2022-08-11 o 18 03 00

* incorrect arch param supplied, that is not one of the possible architectures (https://golang.org/doc/install/source#environment , GOARCH):

http://localhost:8080/docker/image-size/_/mysql/latest?arch=amd64555

Zrzut ekranu 2022-08-11 o 18 10 30

## without tag:

https://registry.hub.docker.com/v2/repositories/library/mysql/tags?page_size=100&ordering=last_updated 

* correct arch param supplied:

http://localhost:8080/docker/image-size/_/mysql?arch=arm64

Zrzut ekranu 2022-08-11 o 18 08 24

* incorrect arch param supplied, that is not one of the possible architectures (https://golang.org/doc/install/source#environment , GOARCH):

http://localhost:8080/docker/image-size/_/mysql?arch=444444

Zrzut ekranu 2022-08-11 o 18 11 46

* correct arch param supplied, that is not present in the fetched docker images response:

http://localhost:8080/docker/image-size/_/mysql?arch=386

Zrzut ekranu 2022-08-11 o 18 26 39

* correct arch param supplied (that is present in the fetched docker images response), with `semver` as a sorting method:

http://localhost:8080/docker/image-size/_/mysql?sort=semver&arch=arm64

Zrzut ekranu 2022-08-11 o 18 27 46


services/docker/docker-size.service.js
```diff
@@ -12,6 +12,7 @@ import {
 const buildSchema = Joi.object({
   name: Joi.string().required(),
   full_size: nonNegativeInteger.required(),
+  images: Joi.any().required(),
```
chris48s, line 15: We shouldn't be using `any()`. The point of defining schemas is to make sure all the values we need are in our input object and handle it cleanly if they are not. `any()` will pass any old value which could lead to us throwing unhandled runtime exceptions later in the code. There are more docs on how and why we validate at https://github.com/badges/shields/blob/master/doc/input-validation.md

services/docker/docker-size.service.js
```diff
@@ -73,30 +84,77 @@ export default class DockerSize extends BaseJsonService {
...
+        // If no tag is specified, and sorting is by date, check if any of the returned images has an architecture matching the arch parameter supplied by the user.
+        // If yes, return the size of the image with this arch.
+        // If not, return the value of the `full_size` in the latestEntry from the response.
+        // For details see: https://github.com/badges/shields/issues/8238
+
+        Object.values(latestEntry.images).forEach(img => {
+          if (img.architecture === arch) {
+            sizeFromSpecifiedArchitecture = img.size
+          }
+        })
```
chris48s, line 105: There's 3 places in this file where we're doing basically the same thing with slightly different variables. You could pull this out to a helper function and then call it in 3 places to DRY up:


```js
function getImageForArch(images, arch)
  Object.values(images).forEach(img => {
    if (img.architecture === arch) {
      return img
    }
  })
  // TODO: if we get here throw a `NotFound()`
}
```

Also notice the early return as soon as we find the result. We don't need to keep iterating the array after we've found what we want. We could have also done this with a `break;`.

services/docker/docker-size.service.js
```diff
@@ -73,30 +84,77 @@ export default class DockerSize extends BaseJsonService {
...
+          })
+        }
+      })
+
+      return { size: sizeFromSpecifiedArchitecture || matches[version] }
     } else {
-      return { size: data.full_size }
+      // If the tag is specified, check if any of the returned images has an architecture matching the arch parameter supplied by the user.
+      // If yes, return the size of the image with this arch.
+      // If not, return the value of the `full_size` from the response (the image with the `latest` tag).
```
chris48s, line 143: I think rather than falling back to `full_size`, we should throw a `NotFound()` if the `arch` is not found. We should only use `full_size` if no arch is specified. We can also make sure the `arch` value itself is valid and throw a sensible error before we start using this approach https://github.com/badges/shields/blob/3aae8ced784c6b3ab5c6490597749396346e3db0/services/docker/docker-version.service.js#L32-L43 (but we shouldn't default to `amd64` given we don't actually need an `arch` for this badge).

Imagine you are a user trying to make a badge and specifying the `arch` param wrong, an informative error is way more useful than just reporting a number that isn't the number you wanted.

chris48s (commented):
Tbh this will be easier to review once there are some tests, but it looks like a reasonable start. I'd suggest given the type of changes you're making here, you're going to want most (or possibly all) of your tests to be unit tests in https://github.com/badges/shields/blob/master/services/docker/docker-size.spec.js rather than service tests. If writing the tests is a bit cumbersome because this `transform()` function is doing a lot of things, it might also make things easier if we split out the 3 main code paths in `transform()` (sort by date, sort by semver, specified tag) down to smaller functions.

I've not really done a proper review but I've had a quick look and left a few comments. We can do a proper review once it is finished.

services/docker/docker-size.service.js
PaulaBarszcz: Good point, thanks, updated.

services/docker/docker-size.service.js
PaulaBarszcz: Very good point, thanks!

Updated to look like this:

```js
function getImageForArch(images, arch) {
  const imgWithArch = Object.values(images).find(
    img => img.architecture === arch
  )

  if (!imgWithArch) {
    throw new NotFound({ prettyMessage: 'architecture not found' })
  }
  return imgWithArch.size
}

```

Since the `find()` method returns the first element in the provided array that satisfies the provided testing function, no unnecessary iteration would be done.

services/docker/docker-size.service.js
PaulaBarszcz: Good point, thanks.

I adjusted the code so that now the full_size is used only if no architecture was specified.
If the user supplied the arch param:
if the image with the given arch was found, we give its size,
if the image with the given arch was not found, we throw an `architecture not found` error.

I moved validArchitectures to docker helper file,
and I added two more possible architectures - `386` as a number and `riscv64` (info from https://golang.org/doc/install/source#environment).

PaulaBarszcz:
Hi @chris48s,
Thank you for all your suggestions, I finished this batch of corrections and updated tests.

This PR is ready to be reviewed.

services/docker/docker-helpers.js
```diff
@@ -2,6 +2,24 @@
...
+  '386',
+  'ppc64',
+  'ppc64le',
+  'wasm',
+  'mips',
+  'mipsle',
+  'mips64',
+  'mips64le',
+  'riscv64',
+  386,
```
chris48s, line 20: Do we actually need `386` as a number as well as a string here?
If I look at an image which publishes a 386 version, for example:
https://registry.hub.docker.com/v2/repositories/library/debian/tags/latest
it appears in the JSON as a string:

```
"architecture": "386",
```

Do we have an example where the architecture is a number?

services/docker/docker-size.service.js
```diff
@@ -20,14 +27,40 @@ const pagedSchema = Joi.object({
...
     })
   ),
 }).required()
 
 const queryParamSchema = Joi.object({
   sort: Joi.string().valid('date', 'semver').default('date'),
+  arch: Joi.alternatives([
+    Joi.string().valid(...validArchitectures),
+    Joi.number().valid(...validArchitectures),
+  ]),
```
chris48s, line 45: I've put a comment in elsewhere saying I don't think we need to allow `architecture` to be a number at all. Assuming we can ditch it, this can just be `arch: Joi.string().valid(...validArchitectures)`. If not, lets come back to it.

services/docker/docker-size.service.js
```diff
@@ -20,14 +27,40 @@ const pagedSchema = Joi.object({
...
+// For details see: https://github.com/badges/shields/issues/8238
+function getImageForArch(images, arch) {
+  const imgWithArch = Object.values(images).find(
+    img => img.architecture === arch
+  )
+
+  if (!imgWithArch) {
+    throw new NotFound({ prettyMessage: 'architecture not found' })
+  }
+  return imgWithArch.size
```
chris48s, line 61: The thing you're actually getting/returning here is a size, not an image
This should either be called `getImageForArch` and return an image object
Or it should be called `getImageSizeForArch` and return a size

services/docker/docker-size.service.js
```diff
@@ -73,30 +114,76 @@ export default class DockerSize extends BaseJsonService {
...
+    let sizeOfImgWithArch
+
+    if (arch) {
+      Object.keys(images).forEach(ver => {
+        if (ver === version) {
+          sizeOfImgWithArch = getImageForArch(images[ver], arch)
+          return { size: sizeOfImgWithArch }
+        }
+      })
+      return { size: sizeOfImgWithArch }
```
chris48s, line 158: If we have iterated through every image and not returned anything inside the `forEach`, then this will `return { size: undefined }`. We don't want to do that. Instead, this line should throw an exception. Probably a `NotFound()`, or maybe an `InvalidResponse()` if we never expect to end up here.

services/docker/docker-size.spec.js
```diff
@@ -3,40 +3,105 @@ import DockerSize from './docker-size.service.js'
...
+    given(
+      {
+        count: 152,
+        results: [
+          {
+            full_size: 300000000,
+            name: 'next',
+            images: [{ architecture: 'amd64', size: 219939484 }],
+          },
+        ],
```
chris48s, line 29: These tests in this file are a bit confusing to read because the test data isn't logically consistent. Can we make the `count` values match the number of items in the array i.e: if there's 1 thing in the array, count should be 1 (not 152), if we're relying on matching `count === 0` in the test data, the array should actually have 0 things in it (not 1), etc.

services/docker/docker-size.service.js
```diff
@@ -73,30 +114,76 @@ export default class DockerSize extends BaseJsonService {
...
 
-  async handle({ user, repo, tag }, { sort }) {
+  transform({ tag, sort, data, arch }) {
+    if (!tag && sort === 'date') {
+      return this.noTagWithDateSortTransform(data, arch)
+    } else if (!tag && sort === 'semver') {
+      return this.noTagWithSemverSortTransform(data, arch)
+    } else {
+      return this.yesTagTransform(data, arch)
+    }
```
chris48s, line 183: These 3 functions are all a bit strangely named. In general, the name of a function should describe _what the function does_ not _when we call it_. So if you think about the fragment:

```js
if (!tag && sort === 'date') {
  return this.noTagWithDateSortTransform(data, arch)
```

this is essentially writing the same information that is in the `if` condition (which we can already read - `not tag`, `sort === 'date'`) again as the function name. It doesn't really describe what action is being performed by the function. Naming things is hard, but I'd suggest something like

`noTagWithDateSortTransform()` --> `getSizeFromImageByLatestDate()`
`noTagWithSemverSortTransform()` --> `getSizeFromImageByLatestSemver()`
`yesTagTransform()` --> `getSizeFromTag()`

might be an improvement.

services/docker/docker-size.service.js
PaulaBarszcz: This is connected to my comment https://github.com/badges/shields/pull/8290#discussion_r951697982 
If needed, I will update this chunk of code :)

services/docker/docker-size.service.js
PaulaBarszcz: Good point, updated, thanks

services/docker/docker-size.service.js
PaulaBarszcz: I modified it, now instead of 
```js 
return { size: sizeOfImgWithArch }
```
we have here:
```js
 if (sizeOfImgWithArch) {
        return { size: sizeOfImgWithArch }
      } else {
        throw new NotFound({ prettyMessage: 'architecture not found' })
      }
```

services/docker/docker-size.service.js
PaulaBarszcz: Thanks, done.

services/docker/docker-size.spec.js
PaulaBarszcz: You're right; updated

PaulaBarszcz:
Hi @chris48s,
Thanks for your review; I applied your suggestions. This PR is ready to be re-reviewed.

Who closes discussions in PRs - the author of the comment/suggestion,
or the author of the PR (once the changes are implemented)?

services/docker/docker-helpers.js
chris48s: OK. thanks for explaining. I can see what we're trying to achieve now. Here's my suggestion.

Lets define a shared Joi schema in `docker-helpers.js` that looks like this:

```js
const archSchema = Joi.alternatives([
  Joi.string().valid(
    'amd64',
    'arm',
    'arm64',
    's390x',
    '386',
    'ppc64',
    'ppc64le',
    'wasm',
    'mips',
    'mipsle',
    'mips64',
    'mips64le',
    'riscv64',
  ),
  Joi.number().valid(386).cast('string'),
])
```

That will accept all the normal architecture values and `386` as a number (but will cast it to a string so it will match the API data when we compare with `===` ).

Then in both `docker-size.service.js` and `docker-version.service.js` we can use that schema in our `queryParamSchema`, and in `docker-version.service.js` you can chain `.default('amd64')`

How does that sound?

PaulaBarszcz: Good idea.
I checked for both `docker-size.service.js` and `docker-version.service.js`, and they still work correctly after the changes.

`docker-version.service.js` :

http://localhost:8080/docker/v/_/alpine?sort=date&arch=386
image

`docker-version.service.js` :
http://localhost:8080/docker/v/_/alpine?sort=date&arch=riscv64
image

http://localhost:8080/docker/v/_/alpine?sort=date&arch=notExistingArch
image

---
`docker-size.service.js` :

http://localhost:8080/docker/image-size/_/mysql?arch=amd64
image


http://localhost:8080/docker/image-size/_/mysql?arch=386
image

http://localhost:8080/docker/image-size/_/mysql?arch=notExistingArch
image

services/docker/docker-size.service.js
PaulaBarszcz: Updated as per https://github.com/badges/shields/pull/8290#discussion_r954208573

PaulaBarszcz:
Hi @chris48s,
Thanks for your suggestion regarding valid architectures. The changes are already pushed to the feature branch.

This PR is ready to be re-reviewed.

chris48s (approved):
nice :+1:  thanks for working on this one

PR #8292, opened 2022-08-08
Fix variable naming for [ROS] version service

jtbandes:
The version information that this service is fetching is the version for a whole repository of ROS packages which all share the same version. A repository may contain only a single package with the same name, but it may also contain multiple packages, and this badge service fetches version information based on the repo name, not the package name.

### DIFF
--- a/services/ros/ros-version.service.js
+++ b/services/ros/ros-version.service.js
@@ -40,6 +40,17 @@ const repoSchema = Joi.object({
   }).required(),
 })
 
+const documentation = `
+<p>
+  To use this badge, specify the ROS <a href="http://docs.ros.org">distribution</a>
+  (e.g. <code>galactic</code> or <code>humble</code>) and the package repository name
+  (in the case of single-package repos, this may be the same as the package name).
+  This badge determines which versions are part of an official ROS distribution by
+  fetching from the <a href="https://github.com/ros/rosdistro">rosdistro</a> YAML files,
+  at the tag corresponding to the latest release.
+</p>
+`
+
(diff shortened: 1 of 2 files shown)


services/ros/ros-version.service.js
```diff
@@ -43,12 +43,12 @@ const packageSchema = Joi.object({
 export default class RosVersion extends GithubAuthV4Service {
   static category = 'version'
 
-  static route = { base: 'ros/v', pattern: ':distro/:packageName' }
+  static route = { base: 'ros/v', pattern: ':distro/:repoName' }
 
   static examples = [
     {
       title: 'ROS Package Index',
```
chris48s, line 50: Given the confusion so far, as well as changing the internal terminology (var/test names, etc) shall we attempt to clarify this a bit more in the user-facing text? We could consider changing the `title` here to clarify that the 'unit of versioning' (for want of a better term) here is the repo not the package. The other thing we sometimes do is include a documentation key in the example (e.g: https://github.com/badges/shields/blob/23c0406bedfc6930735e8f5ea75dfe34faf1f290/services/osslifecycle/osslifecycle.service.js#L3-L12 ) with some extra info that is shown to the user in the builder modal. We could use this to explain what the `repoName` param corresponds to in ROS-land or link to relevant documentation. I'm not 100% sure what to suggest as the platform specifics of ROS are out of my wheelhouse, but perhaps we can do a bit more to explain.

jtbandes: Added some documentation:

image

chris48s (approved):
nice - this looks sensible to me

PR #8295, opened 2022-08-09
feat: add [gitlab]-helper for encapsulate docs and err

sunny0826:
re https://github.com/badges/shields/issues/8077#issuecomment-1207597218 https://github.com/badges/shields/pull/8166#discussion_r939465622


calebcartwright:
Thank you for following up on this! Would you mind rebasing and ensuring this gets applied to some of the newer classes too? (e.g. merge requests)

sunny0826:
> Thank you for following up on this! Would you mind rebasing and ensuring this gets applied to some of the newer classes too? (e.g. merge requests)

Of course.

calebcartwright:
One thing that's a little tricky with this is that we have some example badge listings that support the custom instance query parameter, and some that do not. We currently have the docs for those examples structured in such a way to reflect that (i.e. the custom instance doc info is conditionally presented), but as currently structured, that would go away and the custom instance docs would be mentioned for all badge examples, regardless of whether that example permitted a custom instance.

I think we need to find a way to address this, either by adding the custom instance query parameter on all examples, or by maintaining the current conditional behavior.

Also, I think we can clean up some of these by defining locals in one place/globally and then reusing. That should help avoid a lot of the repeated concatenation we currently have, e.g. `documentation: documentation + more,`

sunny0826:
> One thing that's a little tricky with this is that we have some example badge listings that support the custom instance query parameter, and some that do not. We currently have the docs for those examples structured in such a way to reflect that (i.e. the custom instance doc info is conditionally presented), but as currently structured, that would go away and the custom instance docs would be mentioned for all badge examples, regardless of whether that example permitted a custom instance.
> 
> I think we need to find a way to address this, either by adding the custom instance query parameter on all examples, or by maintaining the current conditional behavior.
> 
> Also, I think we can clean up some of these by defining locals in one place/globally and then reusing. That should help avoid a lot of the repeated concatenation we currently have, e.g. `documentation: documentation + more,`

True, but there are already a lot of examples, and if we add the custom instance example, the number will double.

Also, Do you construct such a function in `gitlab-helper.js`? I don't feel like this is an elegant way to write, any better suggestions?
```
function docsMessageFor(more=''){
  return documentation + more
}
```

calebcartwright:
> > One thing that's a little tricky with this is that we have some example badge listings that support the custom instance query parameter, and some that do not. We currently have the docs for those examples structured in such a way to reflect that (i.e. the custom instance doc info is conditionally presented), but as currently structured, that would go away and the custom instance docs would be mentioned for all badge examples, regardless of whether that example permitted a custom instance.
> > I think we need to find a way to address this, either by adding the custom instance query parameter on all examples, or by maintaining the current conditional behavior.
> > Also, I think we can clean up some of these by defining locals in one place/globally and then reusing. That should help avoid a lot of the repeated concatenation we currently have, e.g. `documentation: documentation + more,`
> 
> True, but there are already a lot of examples, and if we add the custom instance example, the number will double.
> 
> Also, Do you construct such a function in `gitlab-helper.js`? I don't feel like this is an elegant way to write, any better suggestions?
> 
> ```
> function docsMessageFor(more=''){
>   return documentation + more
> }
> ```

Sorry but I'm not following. All I'm suggesting is that you create a constant global (or multiple globals as needed) that has the combined docs, and then just reuse that global in all the relevant examples.

I.e.
```js
const foo = bar + barfoo

....
{
  docs: foo,
},
...
{
  docs: foo,
}
```

instead of having that `bar + barfoo` concatenation in multiple places

calebcartwright (approved):
Thanks! TBH I think this could still benefit from more cleanup/elimination of duplicative string concatenation, but going to go ahead and merge since it is an improvement over current state

Issue #8291, opened 2022-08-08
PyPI version badge doesn't include a link


abitrolly:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Generated markup for PyPI version badge doesn't include a link to package repo. 
```
![PyPI](https://img.shields.io/pypi/v/hy)
```
https://github.com/badges/shields/blob/master/doc/TUTORIAL.md doesn't make it immediately obvious how to pass `target` param from example template to markdown. Maybe add some mermaid diagrams to it?

### 🔗 Link to the badge

/pypi/v

chris48s:
The markdown we generate doesn't usually wrap the image in a link for any service (its not just pypi - its the same for all badges).

> doesn't make it immediately obvious how to pass target param from example template to markdown

At the moment we only use this when the badge is a result from the "suggest badges" feature - in that case, the generated markdown (or rst or whatever) will link to the project URL from the 'suggest' search. I always forget this exists until I dig in to the frontend code.

abitrolly:
> The markdown we generate doesn't usually wrap the image in a link for any service.

Why not?

Where to see  the "suggested badges"? It is not present on the front page.

chris48s:
So there's 2 questions here..

> Why not?

You're not the first person to suggest this as a feature. This is a dupe of other issues like https://github.com/badges/shields/issues/7438 , https://github.com/badges/shields/issues/4726 (and others) but although someone occasionally pops up with a query, nobody has really wanted to push it forward.

In principle it would be possible to add a property or method on each badge class that allows us to generate a link based on the badge params to wrap the image (obviously the url we assemble to make an API call for data is a different thing from the URL you might want a user to visit). We've never done it. I don't think I've ever really articulated my thoughts on it, but..

There are some reasons why would be a reasonable thing to do, if someone really wanted to champion it:

- Users often choose to wrap these things with a link "in the wild" (although predictably, not everyone makes the exact same choice)
- GitHub's default behaviour (which is our primary source of traffic) is to make all images a link to the (proxied) image itself so you can't have a badge in a GH readme which is just a link to nothing.

There would be a number of considerations for us of maintainers of varying levels of importance:

- Its another detail to debate in every PR
- There is probably some murky 'promotional vs informational' ground to resolve here
- For some services there isn't a natural thing to link to, or there are several possible options. In general we try not to be too 'editorial' about these things
- For some services there may not be a natural thing to link to for all variants of the badge
[...]

Issue #7815, opened 2022-04-05
User Experience Design


kohasummons:
:clipboard: **Description**

I popped into shields.io recently to make a badge and at first glance, I had zero idea what I was looking at. It was my first time using this service. While the site is technical and straight to the waffles, I felt there could be an improvement on the frontend.

🏄🏽‍♀️ I scanned through past issues (research, I believe) to see if previous concerns have been earlier raised. I was able to identify a few discussions that hinge on enhancing the current interface. 

Let me list a tiny number of them here:
- https://github.com/badges/shields/pull/1163#issue-265025516
- https://github.com/badges/shields/issues/843#issue-196938870
- https://github.com/badges/shields/issues/2496#issue-389114635
- https://github.com/badges/shields/issues/777#issue-172880880
- https://github.com/badges/shields/issues/7176#issue-1030683469
- https://github.com/badges/shields/discussions/7129#discussion-3616401
- https://github.com/badges/shields/issues/1899#issue-349718225
- https://github.com/badges/shields/pull/1070#issue-255100790

## User Research
I would love to contribute to shields by improving the user experience. This will involve:
1. Usability Audit
2. User research
3. Ultimately, a redesign of the interface

calebcartwright:
Glad to hear you're interested in helping, that's fantastic news! As you've probably gathered from discussions and issues you've read, it's a widely shared opinion and goal amongst the maintainer team and all signs point to the community feeling the same. The biggest roadblock to making significant progress really has been the dearth of extra bandwidth amongst ourselves, and a lack of volunteers (so far) from the community to really push on the bigger pieces, so any assistance is greatly appreciated.

While I do absolutely agree that a refresh of the site is needed, I also want to stress the importance of keeping things framed in terms of our goals and audience.

At the end of the day, our service is functionally a static content generator/provider/host that's targeted explicitly at developers/technologists. The Shields.io site has largely, in my opinion, simply provided a lightweight catalog to our service designed to help our users figure out the url path for their desired static content so that they can then embed it somewhere (e.g. a markdown file).

I mention this because IMO it's very important for our site/catalog to remain focused on that user persona with a clean, simple experience. I wouldn't want to see a Shields.io site try to cater to other personas and get too noisy and busy; I don't think we want to be in the markdown-101 business, so I don't think our site needs to explain basics like adding images to markdown files (even though we do get those types of questions from time to time).

[...]

PR #8311, opened 2022-08-14
Remove 'suggest badges' feature

chris48s:
As noted in #8275 , #8291 and probably other places over the years: The 'suggest badges' feature is somewhat hidden/not discoverable and even if you do find it, the value of the suggestions is pretty limited. We don't really maintain this feature at all as we add services (it basically makes an incomplete suggestion for github and gitlab URLs and does nothing useful for anything else) and it mainly just adds complexity to the frontend.

I propose removing this feature to clean up the frontend code a bit for ourselves and reduce confusion for users.


core/server/server.js
```diff
@@ -112,9 +111,6 @@ const publicConfigSchema = Joi.object({
   },
   redirectUrl: optionalUrl,
   rasterUrl: optionalUrl,
-  cors: {
-    allowedOrigin: Joi.array().items(optionalUrl).required(),
-  },
```
chris48s, line 117: I've completely removed the `allowedOrigin`/`ALLOWED_ORIGIN` setting here as the 'suggest' endpoint was the only thing using it. There's a couple of things to consider here.

We could just leave this here as a setting that does nothing in case we want it again in future. I can see CORS allowed origins being a thing we could use again at some point and removing the setting then adding it in again might be kind of annoying for self-hosting users?

We could consider not removing the setting and issueing some kind of deprecation warning for a bit first. The effect of just removing the setting is that if you configure using env vars the `ALLOWED_ORIGIN` will just be ignored. If you have a yaml config file with `allowedOrigin` in it, Joi will fail to validate your config on startup. I feel like we've previously discussed how we deprecate/remove settings in the past and decided that just removing them like this is OK for that reason, but I can't find it. Anyone remember?

abitrolly:
Discoverability of the feature is actually a task of https://github.com/badges/shields/issues/7815

I would actually find it useful, if it worked.

calebcartwright:
I think the real question should be centered around what, if any, utility this provides in its current state. The premise of badge suggestions provided some input certainly has value in concept, but I'd suggest very little of that value has actually been realized/can be realized within the current contexts.

I'd be in favor of moving forward with the proposal and dropping this in its current form, and then as part of the frontend redo trying to identify a more visible and powerful way to provide the similar types of capabilities

abitrolly:
If maintaining React JS for connecting services with badges is a big burden, maybe devise static JSON format for describing badge and URL? Then different services interested in having a badge could place the description at https://example.com/badge.json And CI/CD could scrape that.

calebcartwright:
> If maintaining React JS for connecting services with badges is a big burden, maybe devise static JSON format for describing badge and URL? Then different services interested in having a badge could place the description at https://example.com/badge.json And CI/CD could scrape that.

I'm not entirely sure I follow what you are proposing, but just want to reiterate that we don't really want to use this PR as the forum to ideate on the future of badge suggestion features/implementations. From my perspective, the only conversation to be had here is whether we keep the feature as-is in its current form, or whether we remove the current implementation and feature.

abitrolly:
As I understand it, you do not want any more development, because the maintenance burden is already overwhelming, and after the redesign this feature became unusable, and need more coding, so the only viable solution for you is to kill it. That's why I proposed an alternative how it could work.

I personally find the feature to automatically generating a badge given URL and service name very useful from UX experience. Than typing the badge variables by hand, but maybe it just should be a browser extension.

I any case, it is your code, so I am +0.5 on removing the code that has little chances to being useful.

calebcartwright:
@abitrolly - my fundamental point is that this is the wrong place to talk about alternative implementations for a feature.

It's purely a binary decision, we either leave it precisely as it is or remove it. Regardless of which option we choose, we can still discuss if/how the feature can be incorporated in the new frontend. However, that future-looking discussion is best had elsewhere both because it would get lost/forgotten here, and because it has no impact on what we do relative to that aforementioned binary decision.

Put differently, this code is going to get deleted/completely overhauled at some point no matter what. It's really just a question of when, and the complete overhaul isn't going to happen today, and isn't going to happen until a number of large prerequisite work items are addressed.

chris48s:
I'll sort out rebasing this at the weekend :+1:

chris48s:
I've updated to resolve the conflicts, before we merge, any thoughts on https://github.com/badges/shields/pull/8311#discussion_r945337530 ?

core/server/server.js
calebcartwright: I'd suggest we leave it in for the moment, and then open an issue to cover removing. I feel like that allows us to move forward, ensure we have something to keep track of the removal, and gives us more time to find out what our posture was/is on these types of things (I don't remember either :smile:)

chris48s: OK cool. I've pushed a6165e7 adding that back in (plus a comment explaining)

calebcartwright (approved):

PR #8316, opened 2022-08-17
fix [dockerstars] service

### LINKED ISSUE #8314: :bug:  Docker stars broken (invalid) :star:

shields.io

### 🐞 Description

Docker stars shield is broken : 

image

While docker repo is ok : 

https://hub.docker.com/r/optnc/api-transitaires

### 🔗 Link to the badge

Can be reproduced on any docker image with `/docker/stars/`

### 💡 Possible Solution

Maybe point to this API `https://hub.docker.com/v2/repositories/optnc` which makes it possible to get the stars with `star_count` attribute

chris48s:
closes #8314

### DIFF
--- a/services/docker/docker-stars.service.js
+++ b/services/docker/docker-stars.service.js
@@ -1,13 +1,18 @@
+import Joi from 'joi'
 import { metric } from '../text-formatters.js'
 import { nonNegativeInteger } from '../validators.js'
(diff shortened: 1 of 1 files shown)


calebcartwright (approved):
I'm still unclear, and mildly concerned, about what happened upstream, but definitely agreed with the direction here :+1:

PR #8332, opened 2022-08-20
migrate 'main' tests to GH actions

chris48s:
Refs #8109

This process begins the process of migrating the Circle CI tests to GitHub Actions. I've started off with eslint/prettier, main and main@node17. There are some bits of tooling here that will help make the rest of the job easier.

Assuming we merge this, we'll want to update the branch protection rules in the repo settings to require these to pass instead of the Circle builds.


.github/workflows/test-lint.yml
```diff
@@ -0,0 +1,26 @@
+name: Lint
```
chris48s, line 1: I decided to split out the linters into a different build from the core tests

.github/actions/core-tests/action.yml
```diff
@@ -0,0 +1,21 @@
...
+  using: 'composite'
+  steps:
+    - name: Core tests
+      if: always()
+      run: npm run test:core -- --reporter json --reporter-option 'output=reports/core.json'
+      shell: bash
+
+    - name: Entrypoint tests
+      if: always()
+      run: npm run test:entrypoint -- --reporter json --reporter-option 'output=reports/entrypoint.json'
```
chris48s, line 13: I can't really remember why `:entrypoint` (which is only one test) needs to be a different test suite than `:core`. I think I'm going to leave them how they are for now and open that worm-can another day. It is not a huge deal.

scripts/mocha2md.js
```diff
@@ -0,0 +1,36 @@
+import fs from 'fs'
```
chris48s, line 1: This little script allows us to generate a [markdown job summary](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) from a mocha JSON report. We should be able to re-use it for the other mocha-based test runs: integration, package, services, etc
You can see an examples of the output here:

- https://github.com/badges/shields/actions/runs/2895774184
- https://github.com/badges/shields/actions/runs/2895785327

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,25 @@
+name: 'Set up project'
```
chris48s, line 1: This re-usable action should allow us to DRY up the setup slightly for the other builds

calebcartwright (approved):
lgtm, and curious to see what the markdown summary looks like in action (no pun intended)

chris48s:
> curious to see what the markdown summary looks like in action

Just before we merge this, can I check you can see the markdown summaries on

- https://github.com/badges/shields/actions/runs/2895774184 (passing)
- https://github.com/badges/shields/actions/runs/2895785327 (failing)

They should look like

https://github.com/badges/shields/actions/runs/2895774184

Screenshot at 2022-08-24 19-35-08

and

https://github.com/badges/shields/actions/runs/2895785327

Screenshot at 2022-08-24 19-34-52

calebcartwright:
> Just before we merge this, can I check you can see the markdown summaries on

Yeah I see them with the direct links, but they aren't showing up in the view one is navigated to after clicking on the check statuses above (though perhaps I keep clicking on the wrong ones)

chris48s:
I see. There is a summary link

Screenshot at 2022-08-25 17-46-58

but its interesting that isn't always the first page you land on.

I think I'm going to go ahead and merge this so I've got the building blocks to start moving other builds over to GHA, but if we think the summaries aren't obvious enough here's another idea: If the build fails, we could use the markdown summary to post a comment the markdown on the PR too. (It will be too much noise if every build comments the output if it succeeds but pushing failed the summary to a PR comment for failed comments would make it really obvious for new users). I might have a look at doing that as part of this process.

PR #8334, opened 2022-08-22
upgrade gatsby

chris48s:
boom :boom:

- conflicts resolved
- no new security vulns
- tests passing

I've labelled this blocker because I'd quite like to merge this before dependabot pops up on Friday and we get a bunch of merge conflicts on `package-lock.json`, although if I do need to do it again, I can.

### DIFF
--- a/cypress/e2e/main-page.cy.js
+++ b/cypress/e2e/main-page.cy.js
@@ -1,3 +1,7 @@
+import { registerCommand } from 'cypress-wait-for-stable-dom'
+
+registerCommand()
+
 describe('Main page', function () {
   const backendUrl = Cypress.env('backend_url')
   const SEARCH_INPUT = 'input[placeholder="search / project URL"]'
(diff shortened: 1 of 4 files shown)


calebcartwright (approved):
I think the dictionary needs an updated definition for "persistence" to include this :laughing:

calebcartwright:
Also, I think I get the gist of the `skipLibCheck` tsconfig option, but would you mind giving me a quick recap of the impact and implications for us here?

chris48s:
The issue we've got with gatsby/express-graphql/graphql is basically around type checking/incompatibility with tsc. Using [skipLibCheck](https://www.typescriptlang.org/tsconfig#skipLibCheck) disables all type checking in `.d.ts` files. See https://bobbyhadz.com/blog/typescript-disable-type-checking-node-modules . There is a tradeoff here. We do lose some type safety not only within `node_modules` but also to a lesser extent within our own codebase: It does also disable https://github.com/badges/shields/blob/master/frontend/types/assets.d.ts and https://github.com/badges/shields/blob/master/frontend/types/mapbox__react-click-to-select/index.d.ts .

calebcartwright:
Thanks for that. Agreed the tradeoffs are worth it here, we've been in dependency upgrade hell for too long

PR #8351, opened 2022-08-28
migrate package tests to GHA

chris48s:
Again, I'll need to update the branch protection rules


.github/workflows/test-package-lib.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - node: '14'
+            engine-strict: 'false'
+          - node: '16'
+            engine-strict: 'true'
+          - node: '18'
+            engine-strict: 'false'
```
chris48s, line 19: While I was working on this I took the opportunity to drop node 12 testing (which is now out of LTS), add node 18 and update the package changelog and package.json

badge-maker/package.json
```diff
@@ -26,7 +26,7 @@
     "badge": "lib/badge-cli.js"
   },
   "engines": {
-    "node": ">= 12",
+    "node": ">= 14",
```
calebcartwright, line 29: I always forget this, but what if any posture do we have relative to engine/dependency update impacts on the major version of our package itself? I.e. this doesn't require a major version bump of the package does it?

chris48s: Yep. The next package version will be 4.0.0. We decided that when we dropped testing on node 10. We've just not actually changed anything that matters since then so we've not done a 4.0.0 release yet.

https://github.com/badges/shields/pull/8351/files#diff-aea71380a717e7f6c60bfb50b18e320f64350f96a62fd3fb805d1aad351bca46

calebcartwright (approved):

Issue #6963, opened 2021-08-31
Packagist dependency version support


Gummibeer:
:clipboard: **Description**

A badge for Packagist to display the version constraint of a dependency. This should be similar/same to the already existing `packagist/php-v` badge but instead of showing hardcoded PHP version constraint it should return the version constrained for a dynamic dependency.
I'm not sure what the best URL structure would be as we would have to pass two package names to the shields.io service. First idea would be something like:
```
packagist/dependency-v/:user/:repo/:dependencyVendor/:dependencyRepo
```

:link: **Data**

The data is already available in the https://github.com/badges/shields/blob/master/services/packagist/packagist-php-version.service.js - but instead of `packageVersion.require.php` it should use something like `packageVersion.require[vendor+'/'+name]`.

:microphone: **Motivation**

Some repositories want to showcase version constraints for important dependencies - for example the supported Symfony or Laravel versions.

PaulaBarszcz:
I would like to work on this one

chris48s:
feel free :+1:

Issue #7783, opened 2022-03-28
Packagist service improvements


chris48s:
:clipboard: **Description**

Refs https://github.com/badges/shields/pull/7782#discussion_r835976765

At the moment we are validating the response before applying the decompression. This means we have to accept missing values and `__unset` in the validation. A better approach would be to do something like I suggested in https://github.com/badges/shields/pull/6508#issuecomment-855295658 - decompress the response first then validate it. This would have two advantages:

- It would allow us to clean up the schemas/validation - we can `.require()` the things that are required
- It would make it easier to "fall into the pit of success" when working on this code. At the moment if we add a new packagist badge it is really non-obvious that we need to call `expandPackageVersions()` on the response and if the latest release is stable you probably won't notice anything off. This should be handled transparently in the base class (both for `fetch()` and `fetchDev()`).

There may also be some test cases that need updating for this

Additionally at the moment there is a single `BasePackagistService` class for both the v2 API and the cached API but really these are two different APIs. If we're going to go down the route of overriding `_requestJson` we'll need to split this into two base classes as the cached API doesn't return compressed responses, but maybe this would be clearer what is going on to split into two base classes anyway.

PR #8371, opened 2022-09-02
[Packagist] dependency version

PaulaBarszcz:
[Updated description]

Related to: https://github.com/badges/shields/issues/6963

We can check the dependencies' versions against the `require` section of https://packagist.org/p2/symfony/symfony.json

Below are some badge screenshots for a couple of scenarios.

1. dependency's name contains only one part (no slash needed), e.g. `php`.
For the php case, I added a redirect from an old endpoint (`packagist-php-version.service.js`), as @chris48s advised:

http://localhost:8080/packagist/php-v/symfony/symfony/v3.2.8?server=https://packagist.org => 
http://localhost:8080/packagist/dependency-v/symfony/symfony/php.svg?server=https%3A%2F%2Fpackagist.org&version=v3.2.8 

image

2. dependency's name contains two parts, e.g. `twig/twig`. 

- both given correctly

http://localhost:8080/packagist/dependency-v/symfony/symfony/twig/twig

image

- one of them given incorrectly

http://localhost:8080/packagist/dependency-v/symfony/symfony/twiiiiiig/twig 

image

3. when dependency's name is the same as the package name 

http://localhost:8080/packagist/dependency-v/symfony/symfony/symfony/contracts

image

4. name of the dependency contains a hyphen

http://localhost:8080/packagist/dependency-v/symfony/symfony/ext-xml

image

5. with specified version of the package

- with correct version

http://localhost:8080/packagist/dependency-v/symfony/symfony/twig/twig?version=v3.2.8

image

- with incorrect version

http://localhost:8080/packagist/dependency-v/symfony/symfony/twig/twig?version=v3.2.8000

image

6. with custom server given (plus version given)

- with a correct server

http://localhost:8080/packagist/dependency-v/symfony/symfony/twig/twig?version=v3.2.8&server=https://packagist.org

image

- with an incorrect server

http://localhost:8080/packagist/dependency-v/symfony/symfony/twig/twig?version=v3.2.8&server=https://packagiiiist.org

image

7. not specified dependency's name

http://localhost:8080/packagist/dependency-v/symfony/symfony/

image


services/packagist/packagist-base.js
```diff
@@ -6,9 +6,7 @@ const packageSchema = Joi.array().items(
   Joi.object({
     version: Joi.string().required(),
     require: Joi.alternatives(
-      Joi.object({
-        php: Joi.string(),
-      }).required(),
+      Joi.object().pattern(Joi.string(), Joi.string()).required(),
```
PaulaBarszcz, line 9: This change was needed so that we accept different dependencies' names for the package from the response,
for example:
```js
  "require": {
          "php": ">=8.1",
          "composer-runtime-api": ">=2.1",
          "ext-xml": "*",
          "friendsofphp/proxy-manager-lts": "^1.0.2",
          "doctrine/event-manager": "~1.0",
          "doctrine/persistence": "^2|^3",
          "twig/twig": "^2.13|^3.0.4",
          "psr/cache": "^2.0|^3.0",
          "psr/container": "^1.1|^2.0",
          "psr/event-dispatcher": "^1.0",
          "psr/link": "^1.1|^2.0",
          "psr/log": "^1|^2|^3",
          "symfony/contracts": "^2.5|^3.0",
          "symfony/polyfill-ctype": "~1.8",
          "symfony/polyfill-intl-grapheme": "~1.0",
          "symfony/polyfill-intl-icu": "~1.0",
          "symfony/polyfill-intl-idn": "^1.10",
          "symfony/polyfill-intl-normalizer": "~1.0",
          "symfony/polyfill-mbstring": "~1.0",
          "symfony/polyfill-uuid": "^1.15"
        },
```
from https://packagist.org/packages/symfony/symfony.json

services/packagist/packagist-dependency-version.service.js
```diff
@@ -0,0 +1,197 @@
...
+      },
+      staticPreview: this.render({
+        dependencyNameForLabel: 'twig/twig',
+        dependencyVersion: '2.13|^3.0.4',
+      }),
+      documentation: customServerDocumentationFragment,
+    },
+  ]
+
+  static defaultBadgeData = {
```
PaulaBarszcz, line 79: If we throw an error, the badge's label will show `dependency version` (instead of the dependency's name)

PaulaBarszcz:
This PR is ready to be reviewed :)

chris48s:
Thanks for picking this one up. I'm just having a look through this, and it seems to me like given the `require` key for a packagist API response is something like:

```json
"require": {
    "php": ">=8.1",
    "composer-runtime-api": ">=2.1",
    "ext-xml": "*",
    "friendsofphp/proxy-manager-lts": "^1.0.2",
    "doctrine/event-manager": "~1.0",
    "doctrine/persistence": "^2|^3",
    "twig/twig": "^2.13|^3.0.4",
    "psr/cache": "^2.0|^3.0",
    "psr/container": "^1.1|^2.0",
    "psr/event-dispatcher": "^1.0",
    "psr/link": "^1.1|^2.0",
    "psr/log": "^1|^2|^3",
    "symfony/contracts": "^2.5|^3.0",
    "symfony/polyfill-ctype": "~1.8",
    "symfony/polyfill-intl-grapheme": "~1.0",
    "symfony/polyfill-intl-icu": "~1.0",
    "symfony/polyfill-intl-idn": "^1.10",
    "symfony/polyfill-intl-normalizer": "~1.0",
    "symfony/polyfill-mbstring": "~1.0",
    "symfony/polyfill-uuid": "^1.15"
}
```

the distinction between a "packagist PHP version" badge ( `/packagist/php-v` ) and a "packagist generic dependency version" badge ( `packagist/dependency-v` ) is somewhat artificial - right?

i.e: if we've got a badge where we can use that data structure to show the version of `twig/twig` or `composer-runtime-api` then using that data structure to show the version of `php` should just be that same badge, but passing it `php` - right? ..or is there something subtly different we need to deal with if the thing we're reporting the version of PHP specifically?

I can also see that in that case things are a bit fiddly because of maintaining compatibility with the current URL structure, but I think I can see where I think we should go with this to streamline things a bit. I just want to check that assumption before I start making suggestions - you've probably spent more time thinking about this lately than I have.

PaulaBarszcz:
> Thanks for picking this one up. I'm just having a look through this, and it seems to me like given the `require` key for a packagist API response is something like:
> 
> ```json
> "require": {
>     "php": ">=8.1",
>     "composer-runtime-api": ">=2.1",
>     "ext-xml": "*",
>     "friendsofphp/proxy-manager-lts": "^1.0.2",
>     "doctrine/event-manager": "~1.0",
>     "doctrine/persistence": "^2|^3",
>     "twig/twig": "^2.13|^3.0.4",
>     "psr/cache": "^2.0|^3.0",
>     "psr/container": "^1.1|^2.0",
>     "psr/event-dispatcher": "^1.0",
>     "psr/link": "^1.1|^2.0",
>     "psr/log": "^1|^2|^3",
>     "symfony/contracts": "^2.5|^3.0",
>     "symfony/polyfill-ctype": "~1.8",
>     "symfony/polyfill-intl-grapheme": "~1.0",
>     "symfony/polyfill-intl-icu": "~1.0",
>     "symfony/polyfill-intl-idn": "^1.10",
>     "symfony/polyfill-intl-normalizer": "~1.0",
>     "symfony/polyfill-mbstring": "~1.0",
>     "symfony/polyfill-uuid": "^1.15"
> }
> ```
> 
> the distinction between a "packagist PHP version" badge ( `/packagist/php-v` ) and a "packagist generic dependency version" badge ( `packagist/dependency-v` ) is somewhat artificial - right?
> 
> i.e: if we've got a badge where we can use that data structure to show the version of `twig/twig` or `composer-runtime-api` then using that data structure to show the version of `php` should just be that same badge, but passing it `php` - right? ..or is there something subtly different we need to deal with if the thing we're reporting the version of PHP specifically?
> 
> I can also see that in that case things are a bit fiddly because of maintaining compatibility with the current URL structure, but I think I can see where I think we should go with this to streamline things a bit. I just want to check that assumption before I start making suggestions - you've probably spent more time thinking about this lately than I have.

Yes, `packagist/dependency-v` is nearly a copy of  `packagist/php-v`. I did not want to change php-v endpoint for the sake of backwards compatibility. But if you think that we should follow a different approach than what I wrote in this PR, I am definitely open to suggestions :)

PaulaBarszcz (commented):
I modified a regex in `services/test-validators` for checking the dependecy’s version so that it would accept also 2 options present in the response from packagist’s API.
An asterisk and 2 alternative versions (separated by a single `|`):
B59F9460-3A23-499D-BC66-0A11CEAD26F5


I checked with https://regexr.com/ to confirm that the previously accepted strings are still valid.

chris48s:
> Yes, `packagist/dependency-v` is nearly a copy of `packagist/php-v`. I did not want to change php-v endpoint for the sake of backwards compatibility. But if you think that we should follow a different approach than what I wrote in this PR, I am definitely open to suggestions :)

OK cool. I think we can usefully tidy up and reduce some duplication here then :)

In terms of the URL schema for the new packagist dependency version badge we've got a few requirements:

- In general, params which are required to render a badge should be positional params in the URL rather than passed in the query string. There are some docs on badge URLs in https://github.com/badges/shields/blob/master/doc/badge-urls.md
- We would like the ability for the 'dependency' to either inlcude a / character or not include a / character. i.e: `php` and `twig/twig` should both be valid
- There is also this optional `:version` param (for specifying we want the to query dependencies for a particular package version instead of the latest) that currently exists on the `php-v` badge which is uncommonly used and not required but we need to maintain compatibility with it.

so here's my suggestion:

- Lets make the URL pattern for the new badge `/packagist/dependency-v/:user/:repo/:dependency+`. This will make the `:dependency` param required and also allow it to contain slash characters. Then the value of `dependency` can be either `php` or `twig/twig` and we can just look for `dependency` in `require` and use it as the badge label. That should simplify a lot of the `dependencyVendor`/`dependencyRepo` stuff you are doing here.
- Using `+` does mean that we then can't make the version param positional in the URL, so lets convert that to a query param instead, like `server`.
- We completely get rid of the existing `PackagistPhpVersion` class, make sure any test cases are covered in the tests for the new badge and we convert the `php-v` route to a redirect:
  - `/packagist/php-v/symfony/symfony` should redirect to `/packagist/dependency-v/symfony/symfony/php`
  - `/packagist/php-v/symfony/symfony/v2.8.0` should redirect to `/packagist/dependency-v/symfony/symfony/php?version=v2.8.0`
  - We can have a specific example on the class for `/packagist/dependency-v` showing PHP version as it is a reasonably important special-case

There are lots of places where we have replaced an old route with a redirect to a new one you can refer to, but here is one example. https://github.com/badges/shields/blob/master/services/pypi/pypi-django-versions.service.js

services/packagist/packagist-base.js
chris48s: Yep - this is fine. As long as we're throwing an exception if the dependency we're looking for isn't found in `require` loosening the validation here is fine.

services/packagist/packagist-php-version.spec.js
```diff
@@ -1,5 +1,8 @@
-import { expect } from 'chai'
+import chai from 'chai'
+import chaiAsPromised from 'chai-as-promised'
 import PackagistPhpVersion from './packagist-php-version.service.js'
+const { expect } = chai
+chai.use(chaiAsPromised)
```
chris48s, line 5: Do we actually need to add this here?

PaulaBarszcz: If we don’t, these test suite did not pass when I checked ;)

PaulaBarszcz:
> > Yes, `packagist/dependency-v` is nearly a copy of `packagist/php-v`. I did not want to change php-v endpoint for the sake of backwards compatibility. But if you think that we should follow a different approach than what I wrote in this PR, I am definitely open to suggestions :)
> 
> OK cool. I think we can usefully tidy up and reduce some duplication here then :)
> 
> In terms of the URL schema for the new packagist dependency version badge we've got a few requirements:
> 
> * In general, params which are required to render a badge should be positional params in the URL rather than passed in the query string. There are some docs on badge URLs in https://github.com/badges/shields/blob/master/doc/badge-urls.md
> * We would like the ability for the 'dependency' to either inlcude a / character or not include a / character. i.e: `php` and `twig/twig` should both be valid
> * There is also this optional `:version` param (for specifying we want the to query dependencies for a particular package version instead of the latest) that currently exists on the `php-v` badge which is uncommonly used and not required but we need to maintain compatibility with it.
> 
> so here's my suggestion:
> 
> * Lets make the URL pattern for the new badge `/packagist/dependency-v/:user/:repo/:dependency+`. This will make the `:dependency` param required and also allow it to contain slash characters. Then the value of `dependency` can be either `php` or `twig/twig` and we can just look for `dependency` in `require` and use it as the badge label. That should simplify a lot of the `dependencyVendor`/`dependencyRepo` stuff you are doing here.
> * Using `+` does mean that we then can't make the version param positional in the URL, so lets convert that to a query param instead, like `server`.
> * We completely get rid of the existing `PackagistPhpVersion` class, make sure any test cases are covered in the tests for the new badge and we convert the `php-v` route to a redirect:
>   
>   * `/packagist/php-v/symfony/symfony` should redirect to `/packagist/dependency-v/symfony/symfony/php`
>   * `/packagist/php-v/symfony/symfony/v2.8.0` should redirect to `/packagist/dependency-v/symfony/symfony/php?version=v2.8.0`
>   * We can have a specific example on the class for `/packagist/dependency-v` showing PHP version as it is a reasonably important special-case
> 
> There are lots of places where we have replaced an old route with a redirect to a new one you can refer to, but here is one example. https://github.com/badges/shields/blob/master/services/pypi/pypi-django-versions.service.js

Cool, I’ll change the code. And I’ll learn the badge’s redirection, yay! 😁

PaulaBarszcz:
I added `toLowerCase()` to prevent the below case:

http://localhost:8080/packagist/dependency-v/symfony/symfony/PHP
image

Now we're good both with `PHP` and `php` as a dependency's name:
http://localhost:8080/packagist/dependency-v/symfony/symfony/php
image

http://localhost:8080/packagist/dependency-v/symfony/symfony/PHP
image

services/packagist/packagist-dependency-version.service.js
```diff
@@ -0,0 +1,204 @@
...
+      } catch (e) {
+        packageVersion = null
+      }
+    }
+
+    if (!dependency) {
+      throw new NotFound({
+        prettyMessage: 'dependency vendor or repo not specified',
+      })
+    }
```
chris48s, line 133: Do we need this check? I think `:dependency+` should enforce this constraint at the routing layer

services/packagist/packagist-dependency-version.service.js
```diff
@@ -0,0 +1,204 @@
...
+
+    if (!packageVersion) {
+      throw new NotFound({ prettyMessage: 'invalid version' })
+    }
+
+    if (!packageVersion.require) {
+      throw new NotFound({ prettyMessage: 'version requirement not found' })
+    }
+
+    const depLowerCase = dependency.toLowerCase()
```
chris48s, line 143: Are we sure no packagist vendor name, package name or extension name can contain an upper-case character?

services/packagist/packagist-dependency-version.service.js
```diff
@@ -0,0 +1,204 @@
...
+      title: 'Packagist PHP Version Support (custom server)',
+      namedParams: {
+        user: 'symfony',
+        repo: 'symfony',
+        dependency: 'php',
+      },
+      queryParams: {
+        server: 'https://packagist.org',
+        version: 'v2.8.0',
+      },
```
chris48s, line 78: I think I'd probably lose the params (server and version on this one)

services/packagist/packagist-dependency-version.service.js
```diff
@@ -0,0 +1,204 @@
...
+
+  static route = {
+    base: 'packagist/dependency-v',
+    pattern: ':user/:repo/:dependency+',
+    queryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'Packagist Dependency Version Support',
```
chris48s, line 26: I know the titles for the old PHP version badge included the word "support", but looking over the titles for everything else in the "Platform & Version Support" category, I think we should drop the word "support" on all 4 titles.

services/packagist/packagist-dependency-version.tester.js
```diff
@@ -0,0 +1,88 @@
+import { isComposerVersion } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('gets the package version of symfony 5.2.3 - without query params')
+  .get('/symfony/symfony/v5.2.3.json')
+  .expectBadge({
+    label: 'dependency version',
+    message: 'version requirement not found',
+  })
```
chris48s, line 10:
```suggestion
```

This test is a bit muddled and can be deleted.
What `/symfony/symfony/v5.2.3` is really doing is looking for a package that symfony requires which is called "v5.2.3" and then failing to find it, but the description and code don't really make that clear. You've got another test further down called "gets the package version of symfony 5.2.3 - incorrectly with missing part of dependency's name" which is testing the same thing really.

services/packagist/packagist-dependency-version.tester.js
```diff
@@ -0,0 +1,88 @@
...
+  'gets the package version of symfony 5.2.3 - name of the dependency contains hyphen'
+)
+  .get('/symfony/symfony/ext-xml.json?version=v5.2.3')
+  .expectBadge({ label: 'ext-xml', message: isComposerVersion })
+
+t.create(
+  'gets the package version of symfony 5.2.3 - correctly when the dependency is the same as the name of the package'
+)
+  .get('/symfony/symfony/symfony/contracts.json')
+  .expectBadge({ label: 'symfony/contracts', message: isComposerVersion })
```
chris48s, line 74:
```suggestion
```

I think this test is another duplicate of the one further up called
"gets the package version of symfony - with correct dependency"
They're both just testing the valid case: we're looking for a dependency, we find it, and we report the version constraint.

services/packagist/packagist-dependency-version.tester.js
```diff
@@ -0,0 +1,88 @@
...
+  .expectBadge({
+    label: 'dependency version',
+    message: 'version requirement not found',
+  })
+
+t.create(
+  'gets the package version of symfony 5.2.3 - name of the dependency contains hyphen'
+)
+  .get('/symfony/symfony/ext-xml.json?version=v5.2.3')
+  .expectBadge({ label: 'ext-xml', message: isComposerVersion })
```
chris48s, line 68: We don't really need to explicitly test `-` is a valid character in a URL at the individual service level. It is conventional that any `:param` allows a hyphen in it.
The interesting thing about this test is that what we're getting is the version of a PHP extension: `twig/twig` (`vendor/package`) is a package, `ext-xml` (no vendor) is a language extension.
Also, version=v5.2.3 is probably noise for this test

services/packagist/packagist-dependency-version.tester.js
```diff
@@ -0,0 +1,88 @@
...
+  })
+
+t.create(
+  `gets the package version of symfony 5.2.3 - incorrectly with missing part of dependency's name`
+)
+  .get('/symfony/symfony/twig.json')
+  .expectBadge({
+    label: 'dependency version',
+    message: 'version requirement not found',
+  })
```
chris48s, line 62: In general, the naming of tests in this file is a bit jumbled. For example, this one is called `gets the package version of symfony 5.2.3...` but you're not passing `?version=5.2.3`. There's another one further down with the same problem that I've suggested deleting as a duplicate. Then you've got another one further up where you're actually explicitly passing `?version=v3.2.8` in the URL but the test name is just "gets the package version of symfony...", and so on.

Can you have a look over the tests and make sure:

- you're passing the `?version=` param when that is actually part of the test, and leave it out if it doesn't make any difference to the test
- if you want to explicitly describe the test params in the name of the test, make sure the test names and the params in the test body match up consistently.

That said, I know the number of cases we need to cover is now a bit larger and naming things is hard, but if you look at the test names that are deleted elsewhere in this PR (e.g: "invalid package name", "invalid version", "custom server", "invalid custom server", etc) those test names are actually pretty good. They explain what we are actually trying to test (i.e: what behaviour are we trying to trigger by providing these inputs, and what are we making an assertion about) rather than describing all of the inputs in each test. You might get more mileage out of just simplifying the test names a bit. I'll leave it with you to have a think about it.

chris48s (commented):
The bulk of this PR is good to go. I've commented some suggestions, but its all minor tweaks now.

services/packagist/packagist-php-version.spec.js
chris48s: Interesting.. This was needed to get the test file to run in isolation, right?

I guess the reason this was passing when the file is run as part of the whole suite is that some other test has registered `chai.use(chaiAsPromised)` but it once the plugin is registered _somewhere_ its registered for the full duration of the test run - it doesn't get cleaned up after each test module.

services/packagist/packagist-dependency-version.service.js
PaulaBarszcz: Good question - I don’t think we can be sure of that.
Should i remove the above line?
It would then have issues with querries containing capital letters 
(https://github.com/badges/shields/pull/8371#issuecomment-1241246084)

We can do one of the following:
1) first check if there is a position in the response in ‚require’ section that contains directly the dependecy’s name (with capital letters)
and if not, transform the queried dependency’s name to lowercase and repeat the comparison
2) transform every dependency’s name from the reeponse to lowercase - and compare with lowercase depemdency’s name from the query.

What would be your preference?

services/packagist/packagist-dependency-version.service.js
PaulaBarszcz: You’re right, thanks, I’ll remove this check.

services/packagist/packagist-php-version.spec.js
PaulaBarszcz: > Interesting.. This was needed to get the test file to run in isolation, right?
> 
Yes.

Your explanation sounds plausible :)

services/packagist/packagist-dependency-version.service.js
PaulaBarszcz: I modified it, now it looks like this - is it fine?

```js

    // All dependencies' names in the 'require' section from the response should be lowercase,
    // so that we can compare lowercase name of the dependency given via url by the user.
    Object.keys(packageVersion.require).forEach(dependency => {
      packageVersion.require[dependency.toLowerCase()] =
        packageVersion.require[dependency]
    })

    const depLowerCase = dependency.toLowerCase()

    if (!packageVersion.require[depLowerCase]) {
      throw new NotFound({ prettyMessage: 'version requirement not found' })
    }

    return { dependencyVersion: packageVersion.require[depLowerCase] }
```

services/packagist/packagist-dependency-version.service.js
PaulaBarszcz: ok, done

services/packagist/packagist-dependency-version.tester.js
PaulaBarszcz: deleted

services/packagist/packagist-dependency-version.tester.js
PaulaBarszcz: deleted

services/packagist/packagist-dependency-version.tester.js
PaulaBarszcz: Modified the test to say:
```js
t.create('gets the package version - dependency does not need the vendor')
  .get('/symfony/symfony/ext-xml.json')
  .expectBadge({ label: 'ext-xml', message: isComposerVersion })
```

services/packagist/packagist-dependency-version.tester.js
PaulaBarszcz: I removed the `version` query param, where it was not needed.
I modified the tests' names, I hope that now they are ok.

PaulaBarszcz:
Suggested changes were implemented. This PR is ready to be re-reviewed.

chris48s:
I'm just having a look over the tests in `services/packagist/packagist-dependency-version.spec.js` and it seems like we have dropped some important test cases that exist in https://github.com/badges/shields/blob/master/services/packagist/packagist-php-version.spec.js

There's the case where the `require` key is missing:

https://github.com/badges/shields/blob/e3c938b4d7b5fac0d468d4e463410c904a0b1db9/services/packagist/packagist-php-version.spec.js#L36-L52

and the case where the `require` key is the string `'__unset'`:

https://github.com/badges/shields/blob/e3c938b4d7b5fac0d468d4e463410c904a0b1db9/services/packagist/packagist-php-version.spec.js#L63-L80

https://github.com/badges/shields/issues/7779 gives a good example of a response including both the missing and `__unset` case: https://packagist.org/p2/hyde/framework.json

The packagist API has some surprising behaviours and edge cases. Most of this is handled by calling `expandPackageVersions()` to 'decompress' the API response. I'd like to ensure we are still covering these cases in the tests. It is quite easy to accidentally break the packagist code due to these gotchas (I actually have an open issue about refactoring this to make it a bit easier at https://github.com/badges/shields/issues/7783 but lets leave that oen for another day).

PaulaBarszcz:
> There's the case where the `require` key is missing:
>[...]
> and the case where the `require` key is the string `'__unset'`:
> [...]
>
> 
> #7779 gives a good example of a response including both the missing and `__unset` case: https://packagist.org/p2/hyde/framework.json
> 
> The packagist API has some surprising behaviours and edge cases. Most of this is handled by calling `expandPackageVersions()` to 'decompress' the API response. I'd like to ensure we are still covering these cases in the tests. It is quite easy to accidentally break the packagist code due to these gotchas (I actually have an open issue about refactoring this to make it a bit easier at #7783 but lets leave that one for another day).

You're right:)
I added the missing cases to unit tests (missing require key and require key with the value of the `__unset` string).

services/packagist/packagist-dependency-version.spec.js
```diff
@@ -0,0 +1,93 @@
...
+    ).to.be.rejectedWith('invalid version')
+  })
+
+  it('should throw NotFound when `require` section is missing in the response', async function () {
+    await expect(
+      PackagistDependencyVersion.prototype.getDependencyVersion({
+        json: fullPackagistJson,
+        user: 'frodo',
+        repo: 'the-one-package',
+        version: 'v2.5.0',
```
chris48s, line 51:
```suggestion
        version: 'v2.4.0',
```

services/packagist/packagist-dependency-version.spec.js
```diff
@@ -0,0 +1,93 @@
...
+    ).to.be.rejectedWith('version requirement not found')
+  })
+
+  it('should throw NotFound when `require` section in the response has the value of __unset (thank you, Packagist API :p)', async function () {
+    await expect(
+      PackagistDependencyVersion.prototype.getDependencyVersion({
+        json: fullPackagistJson,
+        user: 'frodo',
+        repo: 'the-one-package',
+        version: 'v2.4.0',
```
chris48s, line 62:
```suggestion
        version: 'v2.5.0',
```

services/packagist/packagist-dependency-version.spec.js
```diff
@@ -0,0 +1,93 @@
...
+          version: 'v3.0.0',
+          require: { php: '^7.4 || 8', 'twig/twig': '~1.28|~2.0' },
+        },
+        {
+          version: 'v2.5.0',
+          require: '__unset',
+        },
+        {
+          version: 'v2.4.0',
+        },
```
chris48s, line 21: I was initially confused by this, but I see what you did with the ordering to fit all 3 cases into a single input.

chris48s (commented):
Nearly done. There is one final thing to get this over the line: In a6fd477 the captions and version are the wrong way round: `v2.5.0` is the `__unset` case and `v2.4.0` is the missing case. I've suggested 2 corrections to make the inputs match the descriptions.

PaulaBarszcz:
> Nearly done. There is one final thing to get this over the line: In [a6fd477](https://github.com/badges/shields/commit/a6fd477f7c23a73f18523dbcfd65907302913c9b) the captions and version are the wrong way round: `v2.5.0` is the `__unset` case and `v2.4.0` is the missing case. I've suggested 2 corrections to make the inputs match the descriptions.

Thanks for spotting this, you’re right 😅 Committed.

services/packagist/packagist-dependency-version.spec.js
PaulaBarszcz: 👍☺️

chris48s (approved):
Got there in the end. Packagist is definitely one of the trickier service families to work on.

:rocket:

PR #8373, opened 2022-09-03
Docstrings for package json helpers

pr7prashant:
- Added docstrings for package json helpers

### DIFF
--- a/services/package-json-helpers.js
+++ b/services/package-json-helpers.js
@@ -10,7 +10,7 @@ import { InvalidParameter } from './index.js'
 /**
  * Joi schema for validating dependency map.
  *
- * @type {object}
+ * @type {Joi}
  */
 const isDependencyMap = Joi.object()
   .pattern(
@@ -24,7 +24,7 @@ const isDependencyMap = Joi.object()
  * Joi schema for validating package json object.
  * Checks if the object has all the dependency types and the dependency types are valid.
  *
- * @type {object}
+ * @type {Joi}
  */
 const isPackageJsonWithDependencies = Joi.object({
   dependencies: isDependencyMap,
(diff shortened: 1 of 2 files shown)


services/package-json-helpers.js
```diff
@@ -9,13 +20,33 @@ const isDependencyMap = Joi.object()
   )
   .default({})
 
+/**
+ * Joi schema for validating package json object.
+ * Checks if the object has all the dependency types and the dependency types are valid.
+ *
+ * @type {object}
+ */
```
chris48s, line 28: Because Joi schemas are so commonly used, we've actually set up jsdoc so that we can use `Joi` as a "type" for schemas

https://github.com/badges/shields/blob/daca76726cc01184b3f851e6436309db90248d3b/.eslintrc.yml#L170-L171

e.g:

https://github.com/badges/shields/blob/daca76726cc01184b3f851e6436309db90248d3b/core/base-service/base-json.js#L29

I guess we've probably merged a few of these PRs describing the schemas as `object` but it lets use `@type {Joi}` for Joi schemas moving forwards.

services/package-json-helpers.js
```diff
@@ -9,13 +20,33 @@ const isDependencyMap = Joi.object()
...
+ * @param {object} attrs - Refer to individual attributes
+ * @param {string} attrs.kind - Wanted dependency type, defaults to prod
+ * @param {string} attrs.wantedDependency - Name of the wanted dependency
+ * @param {object} attrs.dependencies - Map of dependencies
+ * @param {object} attrs.devDependencies - Map of dev dependencies
+ * @param {object} attrs.peerDependencies - Map of peer dependencies
+ * @param {object} attrs.optionalDependencies - Map of optional dependencies
+ * @throws {string} - Error message if unknown dependency type provided
+ * @throws {InvalidParameter} - Error if wanted dependency is not present
+ * @returns {object} Object containing wanted dependency version
```
chris48s, line 48: This function does currently return an object, but it doesn't really need to. I expect this is an artefact of some past refactoring. A nice cleanup we could do here would be to update this so it just returns a string and update the handful of places it is consumed accordingly. Also, the return value (`range`) might be an exact version, but it might also be a semver range i.e: this value could be something like "~2.1.6" or ">=3.0.0,<4.0.0" rather than a precise version number so lets explain that in the docstring.

chris48s (commented):
thanks for continuing to work on these

services/package-json-helpers.js
pr7prashant: Noted. Joi type does make more sense here. Will remember this moving forward.

services/package-json-helpers.js
pr7prashant: Sure, will update this description. I'll raise a new PR for the cleanup to keep it simple.

chris48s (approved):

PR #8377, opened 2022-09-05
feat: add [GitlabLanguageCount] service

sunny0826:


services/gitlab/gitlab-languages-count.service.js
```diff
@@ -0,0 +1,69 @@
...
+      queryParams: { gitlab_url: 'https://gitlab.com' },
+      staticPreview: {
+        label: 'languages',
+        message: '5',
+      },
+      documentation,
+    },
+  ]
+
+  static defaultBadgeData = { label: 'languages', namedLogo: 'gitlab' }
```
calebcartwright, line 41: We shouldn't have a default logo on this one, so let's drop it:

```suggestion
  static defaultBadgeData = { label: 'languages' }
```

services/gitlab/gitlab-languages-count.service.js
```diff
@@ -0,0 +1,69 @@
...
+
+  static render({ languagesCount }) {
+    return {
+      label: 'languages',
```
calebcartwright, line 45: The label is already set in the defaults so we don't need to specify this again

```suggestion
```

calebcartwright (commented):
Thanks! Couple minor items inline below

services/gitlab/gitlab-languages-count.service.js
sunny0826: done

services/gitlab/gitlab-languages-count.service.js
sunny0826: done

calebcartwright (approved):
LGTM, thanks!

Issue #6901, opened 2021-08-16
Number of commits between two GitHub commits/tags (1-indexed)


ghost:
:clipboard: **Description**

- **Which service is this badge for?** GitHub
- **What sort of information should this badge show?**
  [![Commits](https://img.shields.io/badge/dynamic/json?style=flat-square&label=Commits&logo=git&query=%24.total_commits&logoColor=ffffff&labelColor=212121&color=0093ed&url=https%3A%2F%2Fapi.github.com%2Frepos%2FSNDST00M%2Fmaterial-dynmap%2Fcompare%2Fv0.5.0...v0.6.0)](https://img.shields.io/badge/dynamic/json?style=flat-square&label=Commits&logo=git&query=%24.total_commits&logoColor=ffffff&labelColor=212121&color=0093ed&url=https%3A%2F%2Fapi.github.com%2Frepos%2FSNDST00M%2Fmaterial-dynmap%2Fcompare%2Fv0.5.0...v0.6.0)

:link: **Data**

- **Is there a public API?** Yes
- **Does the API requires an API key?** No
- **Link to the API documentation.**
  https://api.github.com/repos/SNDST00M/material-dynmap/compare/v0.5.0...v0.6.0 (total_commits)
  

:microphone: **Motivation**

Counting commits in release descriptions, with a link to the actual commits.

chris48s:
The issue with doing this via the v3/REST API is that if there are many commits between Tag A and Tag B the output will be paginated and we might have to make many HTTP requests to count them. Another approach would be to use the v4/GraphQL API and make a request like this:

```gql
query {
  first: repository(owner:"SNDST00M", name:"material-dynmap") {
    object(expression:"v0.5.0") {
      ... on Commit {
        history {
          totalCount
        }
      }
    }
  }
  second: repository(owner:"SNDST00M", name:"material-dynmap") {
    object(expression:"v0.6.0") {
      ... on Commit {
        history {
          totalCount
        }
      }
    }
  }
}
```

That would return an object like

```json
{
  "data": {
    "first": {
      "object": {
        "history": {
          "totalCount": 93
        }
      }
    },
    "second": {
      "object": {
        "history": {
          "totalCount": 106
        }
      }
    }
  }
}
```

and we could use the difference between the two totals to give us the difference with only one request regardless of the number of commits. There is a `GithubAuthV4Service` for working with the GraphQL API if anyone is interesting in working on this.

SNDST00M:
[...]

PR #8394, opened 2022-09-11
[GitHub] Number of commits between branches/tags/commits

PaulaBarszcz:
Related to: #6901 

It returns the number of commits between given branches/tags/commits.

We call under the hood:
https://api.github.com/repos/microsoft/vscode/compare/1.60.0...82f2db7

Example branches names taken from `stale branches` from:
https://github.com/microsoft/vscode/branches/stale

correct - branches
http://localhost:8080/github/commits-difference/microsoft/vscode?branchA=standalone/0.1.x&branchB=release/1.21
image

correct - tag...commit
http://localhost:8080/github/commits-difference/microsoft/vscode?branchA=82f2db7&branchB=1.60.0
image

incorrect - commits
http://localhost:8080/github/commits-difference/microsoft/vscode?branchA=82f2db7&branchB=fffffff
image

incorrect - missing required query param
http://localhost:8080/github/commits-difference/microsoft/vscode?branchA=82f2db7
image


PaulaBarszcz:
Following a discussion with @calebcartwright  on Discord
I’ll change separating  two branches names with …
to 1 branch name in the url and the other in the query params, as they both can contain slashes.

PaulaBarszcz:
This PR is ready to be reviewed.

PaulaBarszcz:
I don’t think that the recent pipeline failed because of the changes introduced in this PR (?) 

https://github.com/badges/shields/actions/runs/3055809500/jobs/4929284513

chris48s:
> I don’t think that the recent pipeline failed because of the changes introduced in this PR

Yeah - looks like a transient issue with NPM. They should pass next time you push a commit

chris48s:
> Following a discussion with @calebcartwright  on Discord.....

I think the approach of having 2 query params is fine, and we don't need to change it. That said, while I was reading this it occurred to me that another way we could have approached this would have been to take a single string param `/:diff+`, leave it to the user to call `/github/commits-difference/myuser/myrepo/main...feature` and just pass the whole `diff` string to the GH API. Was that the solution you were thinking of to start with?

services/github/github-commits-difference.service.js
```diff
@@ -0,0 +1,60 @@
...
+  static examples = [
+    {
+      title: 'GitHub commits difference between two branches/tags/commits',
+      namedParams: {
+        user: 'microsoft',
+        repo: 'vscode',
+      },
+      queryParams: {
+        branchA: '1.60.0',
+        branchB: '82f2db7',
```
chris48s, line 31: As is often the case, the hardest jobs is giving things good names :)

As demonstrated by the fact that the 2 examples you've used here are a commit hash and a tag, `branchA` and `branchB` don't actually have to be branches. Fortunately in git there is a word for this identifier that could be a branch name, commit hash or tag name, which is "ref" (you'll also notice you can use the "special" ref names just fine here like `HEAD~1`, etc if you try them).

The other thing here is that A and B aren't very descriptive either. It isn't really clear to me which one of this is expected to be the ref that is further ahead/behind. The Github API calls these params `{base}...{head}` which isn't going to be completely obvious to everyone, but I think those terms are more descriptive than A and B and at least have the advantage of maintaining consistency with the terms GitHub uses. Another option is GitLab uses `{from}...{to}` for these concepts..

So we could go with

- `base` / `head`
- `base_ref` / `head_ref`
- `from` / `to`
- `from_ref` / `to_ref`

(we use underscores in query params)

Personally, I have a weak preference for `base` and `head` just because they're the same words github uses in the API docs, but if you have a strong preference that one of the other 3 options there is more meaningful we can go with any of those. And lets make the error message "could not establish commit difference between **refs**"

chris48s (commented):
OK. This one is a bit easier than the packagist PR :relieved: Nearly all of this is already ready to go.

PaulaBarszcz:
> > Following a discussion with @calebcartwright  on Discord.....
> 
> I think the approach of having 2 query params is fine, and we don't need to change it. That said, while I was reading this it occurred to me that another way we could have approached this would have been to take a single string param `/:diff+`, leave it to the user to call `/github/commits-difference/myuser/myrepo/main...feature` and just pass the whole `diff` string to the GH API. Was that the solution you were thinking of to start with?

Yes, at first I wrote the solution with `/:diff+` and directly passing the string with `...` but Caleb mentioned that trying to use token separators in a single param could get tricky, so I switched to 2 separate query params ;)

services/github/github-commits-difference.service.js
PaulaBarszcz: > Personally, I have a weak preference for `base` and `head` just because they're the same words github uses in the API docs, but if you have a strong preference that one of the other 3 options there is more meaningful we can go with any of those. And lets make the error message "could not establish commit difference between **refs**"

Why do we use underscores in query params (and not hyphens or camelCase)?

Sure, I'll rename `branchA` to `base`, and `branchB` to `head`.

The error message will read `"could not establish commit difference between refs"`.

Yeah, I was also not fond of putting `branch` in the URL (since users can use a couple of other options) - you're right that `ref` is a great term to encompass all of the possibilities. Thanks;)

chris48s (approved):

chris48s:
This is a good question.

I can tell you that at some point we seem to have (mostly?) consistently settled on underscore for query params e.g:

- https://github.com/badges/shields/blob/93fa955dde5bc2f8d50139c519816043a7bd7a79/services/npm/npm-base.js#L33-L35
- https://github.com/badges/shields/blob/93fa955dde5bc2f8d50139c519816043a7bd7a79/services/gitlab/gitlab-forks.service.js#L10-L12
- https://github.com/badges/shields/blob/93fa955dde5bc2f8d50139c519816043a7bd7a79/services/github/github-release.service.js#L12-L14
- https://github.com/badges/shields/blob/93fa955dde5bc2f8d50139c519816043a7bd7a79/services/matrix/matrix.service.js#L4-L6
- and so on

Given the param names get converted to variables, it would be inconvenient to use kebab-case as `-` isn't valid in a variable name. I'm not really sure exactly how we arrived at snake_case over camelCase for the query params. I guess in general we try to keep the URLs the end-user calls lowercase characters only. I'm not sure I have a 100% satisfying answer though. This may be a pattern that had already emerged before I started contributing to the project.

PR #8395, opened 2022-09-11
conform to standard js quote rules

chris48s:
I'm not really sure where this crept in (quite a while ago by the looks of it), but we should be following the standard js rules for quotes: https://github.com/standard/eslint-config-standard/blob/bcd019906cd379be04ef8e6347e108bad4fc448d/.eslintrc.json#L201
..so using backtick quotes should be allowed if there is a variable substitution but not otherwise. i.e:

```js
const foo = `bar ${baz}` // OK

const foo = `bar baz`    // lint error
```

I guess some other plugin we are using is overriding the rule. Not 100% sure.

Anyway, this PR explicitly re-enables the rule in our eslint config and re-formats all the places where we are using backtick strings with no templating. In general, I don't like to spend a lot of time bikeshedding lint config but I think this is an improvement.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -144,6 +144,8 @@ rules:
   func-style: ['error', 'declaration', { 'allowArrowFunctions': true }]
   new-cap: ['error', { 'capIsNew': true }]
   import/order: ['error', { 'newlines-between': 'never' }]
+  quotes:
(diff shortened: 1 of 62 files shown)


paulmelnikow (approved):

calebcartwright (approved):
Agreed with this. Also surprised to see it as I feel like I've been bitten by the linter when making this mistake in the past :man_shrugging:

PR #8396, opened 2022-09-12
Return string instead of object from getDependencyVersion function; affects [githubpackagejson npmdependencyversion]

pr7prashant:
Cleanup for https://github.com/badges/shields/pull/8373#discussion_r962304386
> This function does currently return an object, but it doesn't really need to. I expect this is an artefact of some past refactoring. A nice cleanup we could do here would be to update this so it just returns a string and update the handful of places it is called accordingly.


chris48s (approved):

chris48s:
nice - thanks :+1:

PR #8421, opened 2022-09-19
migrate service tests to GH actions

chris48s:
Finally closes https://github.com/badges/shields/issues/8109
(although we still have the daily-tests repo to do and a bit more work on service tests that require credentials :grimacing: )

The core purpose of this is moving the services build to run on GitHub Actions instead of Circle CI, but there are some intricacies.


chris48s:
This is 90% done, but marking it as 'on hold' for the moment while we sort out the other 10%

chris48s:
Bleh. Unexpectedly, it looks like the token we get in CI can't call the API for gists or file contents. Need to think about how to deal with that. Also (more predictably) running ~250 github tests on 2 node versions lots of times is enough to exceed the rate limit.

chris48s:
TODO: use a PAT if the PR is from a local branch (like the integration tests)

.circleci/config.yml
```diff
@@ -1,77 +0,0 @@
-version: 2
```
chris48s, line 1: Deleting the remaining circle config causes the Circle CI build to fail with "No `.circleci/config.yml` was found in your project. Add a configuration file to your repository or refer to documentation.". When we get to merging this I will disable Circle from trying to run on the repository but for now I am going to leave it while there are still other open PRs running the service tests on Circle. Also I will update the branch protection rules, delete all the env vars from Circle, etc.

.github/workflows/test-services.yml
```diff
@@ -0,0 +1,38 @@
...
+          twitch-client-id: '${{ secrets.SERVICETESTS_TWITCH_CLIENT_ID }}'
+          twitch-client-secret: '${{ secrets.SERVICETESTS_TWITCH_CLIENT_SECRET }}'
+          wheelmap-token: '${{ secrets.SERVICETESTS_WHEELMAP_TOKEN }}'
+          youtube-api-key: '${{ secrets.SERVICETESTS_YOUTUBE_API_KEY }}'
+
+      - name: Service tests (triggered from fork)
+        if: github.event.pull_request.head.repo.full_name != github.repository
+        uses: ./.github/actions/service-tests
+        with:
+          github-token: '${{ secrets.GITHUB_TOKEN }}'
```
chris48s, line 38: If we submit a PR from a branch on this repo (Anyone with push access can do this. Also importantly, this applies to dependabot PRs - they're subimtted from a local branch, not a fork), all of the service tests can run using credentials which I have configured as secrets.

If we recieve a PR from a fork (like this one) tests for the following services will fail due to lack of credentials, or be flaky due to reduced rate limit:
- Libraries.io
- OBS
- Symfony Insight
- Twitch
- Wheelmap
- Youtube
- GitHub gists only, but not the rest of GitHub (more on this further down)

All other service tests that do not require auth will run fine in either case.
As discussed on the call, this means this are a small handful of services where if an external contributor wants to submit a PR changing them, we'll have to verify the tests locally and then edit the PR title before merging.
Seperately, I will have a look at finding a way to run tests for these services in another context.

You can see an example of service tests running with the required credentials on a PR from a local branch at https://github.com/badges/shields/actions/runs/3923463559/jobs/6707091330

services/github/gist/github-gist-last-commit-redirect.tester.js
```diff
@@ -1,6 +1,7 @@
 import { ServiceTester } from '../../tester.js'
+
 export const t = new ServiceTester({
-  id: 'GithubGistLastCommitRedirect',
+  id: 'GistLastCommitRedirect',
```
chris48s, line 4: Annoyingly, the built-in `GITHUB_TOKEN` secret has enough permissions to run _neeeeearly_ all of the GitHub test suite. The one exception to this is Gists. Also it is not possible to grant a token a `gist: read` scope: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token This is a limitation which does apply to github actions/workflow tokens but doesn't exist for PATs.

The way I've proposed to deal with this is we rename the `GithubGist` service to just `Gist`. This will mean if I submit a PR

- "do a thing to [github] services" then all the tests should run and pass, even if the PR is submitted from a fork
- "do a thing to [gist] services" then all the tests should run and pass from a local branch but will fail from a fork

The downside of that is we have to remember that `GithubGist` has a different service tag from `Github`. The other thing we could do is leave it and accept that a handful of tests will fail if we try to run the full github suite from a forked PR.

.github/workflows/test-services-17.yml
```diff
@@ -0,0 +1,40 @@
...
+          github-token: '${{ secrets.GH_PAT }}'
+          librariesio-tokens: '${{ secrets.SERVICETESTS_LIBRARIESIO_TOKENS }}'
+          obs-user: '${{ secrets.SERVICETESTS_OBS_USER }}'
+          obs-pass: '${{ secrets.SERVICETESTS_OBS_PASS }}'
+          sl-insight-user-uuid: '${{ secrets.SERVICETESTS_SL_INSIGHT_USER_UUID }}'
+          sl-insight-api-token: '${{ secrets.SERVICETESTS_SL_INSIGHT_API_TOKEN }}'
+          twitch-client-id: '${{ secrets.SERVICETESTS_TWITCH_CLIENT_ID }}'
+          twitch-client-secret: '${{ secrets.SERVICETESTS_TWITCH_CLIENT_SECRET }}'
+          wheelmap-token: '${{ secrets.SERVICETESTS_WHEELMAP_TOKEN }}'
+          youtube-api-key: '${{ secrets.SERVICETESTS_YOUTUBE_API_KEY }}'
```
chris48s, line 34: I've prefixed these just to help organise things a bit.

.github/workflows/test-services.yml
chris48s: here's an equivalent build trying to run the same service tests from a fork
https://github.com/badges/shields/actions/runs/3923735407

chris48s:
@calebcartwright - this one is now ready for a review

core/service-test-runner/pull-request-services-cli.js
```diff
@@ -8,54 +8,26 @@
...
-  } else {
-    console.error(
-      `Services: (${services.length} found) ${services.join(', ')}\n`
-    )
-    console.log(services.join('\n'))
+try {
+  if (process.argv.length < 3) {
+    throw new Error()
   }
+  title = process.argv[2]
```
calebcartwright, line 20: I know this is right because it works, but I'm having a little trouble convincing myself I've got the correct mental model of the flow. Is it just that this is expecting that the `...:prepare` npm script was invoked with the title arg in the first true arg position?

calebcartwright (approved):
Works for me, and definitely aligns to the plan we came up with at the last team meeting. Understand the need and agreed with the approach for special-case naming the GH gist badges as well. I'm fine with the name as-is, but suppose one thing worth considering is using some kind of additional (non-`Github`) identifier to reduce the risk of any name collisions down the road, e.g. `GHGist` or something like that.

Happy to proceed as-is, happy to re-:+1: if you decide to make any additional tweaks

core/service-test-runner/pull-request-services-cli.js
chris48s: Even though we invoke it with `npm run...`,
`argv[0]` is the full path to `node`
`argv[1]` is the full path to `pull-request-services-cli.js`
and then `argv[2]` is the first positional argument.

chris48s: and we call it here
https://github.com/badges/shields/blob/00fe69418eeeb963ca94f817438a1a744bc3b355/.github/actions/service-tests/action.yml#L47-L57

PR #8423, opened 2022-09-22
migrate integration tests to GH actions

chris48s:
and so it continues..


.github/workflows/test-integration-17.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: ./.github/actions/setup
+        with:
+          node-version: 17
+        env:
+          NPM_CONFIG_ENGINE_STRICT: 'false'
+
+      - name: Integration Tests
+        uses: ./.github/actions/integration-tests
+        with:
+          github-token: '${{ secrets.GITHUB_TOKEN }}'
```
chris48s, line 37: Even when the PR is submitted from a fork, the build is assigned a short-lived read-only token which is enough for running the tests.

.github/workflows/test-integration-17.yml
```diff
@@ -0,0 +1,37 @@
...
+    services:
+      redis:
+        image: redis
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 6379:6379
```
calebcartwright, line 21: this seems mildly annoying to have to setup ourselves. did the circle image have all this baked in already?

.github/actions/integration-tests/action.yml
```diff
@@ -0,0 +1,20 @@
+name: 'Integration tests'
+description: 'Run integration tests'
+inputs:
+  github-token:
+    description: 'The GITHUB_TOKEN secret'
+    required: true
```
calebcartwright, line 6: Is it a requirement and/or are there benefits of requiring this as an input vs. using the system var directly? asking as it seems like both "call sites" are just passing the same secret

.github/actions/integration-tests/action.yml
```diff
@@ -0,0 +1,20 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - name: Integration Tests
+      if: always()
+      run: npm run test:integration -- --reporter json --reporter-option 'output=reports/integration-tests.json'
```
calebcartwright, line 12: I think you spoke to this already on a prior PR (apologies if so), but were we seeing issues with the env var route for these mocha values or do you just think it's better with inline args?

I've no preference either way and am content with whatever works, more just curious

calebcartwright (approved):
Couple non-blocking questions left inline below, but lgtm as-is if you want to merge

.github/actions/integration-tests/action.yml
chris48s: An action doesn't automatically inherit secrets/vars from the workflow that calls it. You have to explicitly pass them in as params. This makes a lot of sense from a security pov if you are running third-party actions in your workflow and you want to send them some variables but not everything.

.github/actions/integration-tests/action.yml
chris48s: Don't think it has come up in any of the previous PRs. Not sure whether we pass the options as args or env vars makes a huge amount of difference either way tbh.

.github/workflows/test-integration-17.yml
chris48s: I think both Circle and GHA are just using the `redis` image from DockerHub.

We definitely do need to explicitly map the port on GitHub Actions - if you remove that it doesn't work. Not sure why we didn't have to do that on Circle.

With the health checks, I put them in because all the examples on https://docs.github.com/en/actions/using-containerized-services/creating-redis-service-containers have it but the tutorial doesn't really explain it beyond `# Set health checks to wait until redis has started`. I could try deleting it and see if it still works.

PR #8434, opened 2022-09-24
fix version sorting for ruby [gem] s

### LINKED ISSUE #8425: By default Ruby gem version is beta even when there is a stable release

shields.io

### 🐞 Description

There is a gem: https://rubygems.org/gems/filewatcher

As you can see, it had a few beta-versions (of 2.0).

By default, `gem` (or Bundler) doesn't recognize them (until specific lock version).

But… there is 2.0.0 stable release. And it's newer, and it's stable.

But the badge still "dangerous" about beta-versions: https://img.shields.io/gem/v/filewatcher?include_prereleases&style=flat-square

image

Now I see there is `include_prereleases` param, but I believe it should count released versions.

### 🔗 Link to the badge

https://img.shields.io/gem/v/filewatcher?include_prereleases&style=flat-square

### 💡 Possible Solution

Reject any `.beta`, `.rc`, etc. versions if there is a stable version with the same digits.

chris48s:
closes #8425

As noted in #8425 our behaviour when sorting ruby gems with pre-releases is incorrect.
This is because we're attempting to use semver sorting, which assumes the pre-release identifier is separated from the main part of the version number by a `-`, rather than a `.`. This PR switches us to using https://www.npmjs.com/package/@renovatebot/ruby-semver to implement ruby ecosystem-specific version sort/color logic instead of the generic version functions.
Related docs: https://guides.rubygems.org/patterns/#prerelease-gems
I'm really surprised this has never come up before given how long these gem badges have been around, but lets get it fixed now.


services/gem/gem-helpers.js
```diff
@@ -0,0 +1,25 @@
+import { valid, maxSatisfying, prerelease } from '@renovatebot/ruby-semver'
+
+function latest(versions) {
+  // latest Ruby Gems version, including pre-releases
+  return maxSatisfying(versions, '>0')
+}
```
chris48s, line 6: I haven't implemented including/excluding pre-releases here - for stable version only we just report the `.version` param from the API so we only actually need the 'including pre-releases' case.

services/gem/gem-helpers.spec.js
```diff
@@ -0,0 +1,17 @@
+import { test, given } from 'sazerac'
+import { latest, versionColor } from './gem-helpers.js'
+
+describe('Gem helpers', function () {
```
chris48s, line 4: I haven't written loads of test cases here. For the most part I think we can rely on the tests for the underlying library in https://github.com/renovatebot/ruby-semver/blob/main/test/index.test.ts I have covered a handful of cases, especially specific to the sorting of pre/stable releases here

services/gem/gem-helpers.js
calebcartwright: Do you know offhand if this will support or typical behavior of falling back to a prerelease version in cases where the user has requested stable-only versions but all the versions are prerelease?

chris48s: As I say, I haven't implemented `includePre` here because we don't need to use this function if the user wants latest stable.
If we want the latest stable, we just return the value of `.version` here: https://github.com/badges/shields/blob/d8c3dd12e2eb8b1afd9543783fb6d894e198ed61/services/gem/gem-version.service.js#L71-L73
`maxSatisfying()` doesn't care about stable/pre-releases. It just returns the latest version that matches the constraint, so if we wanted to implement a version that returns the latest stable, we'd do it something like (not actually tested, but hopefully you get the point):

```js
function latest(versions, includePre) {
  const stableVersions = versions.filter((v) => !prerelease(v) && v[0] !== '0')
  if (stableVersions.length >0 && !includePre) {
   return maxSatisfying(stableVersions, '>0') 
  }
  return maxSatisfying(versions, '>0')
}
```

but we don't need it
so I haven't bothered

calebcartwright: got it. apologies it wasn't clear to me on first read what you meant in the initial comment but i see it now :+1:

calebcartwright (approved):

services/gem/gem-helpers.js
chris48s: no probs

PR #8435, opened 2022-09-24
Add `rc` to pre-release identifiers

chris48s:


calebcartwright (approved):
Suppose there's technically a non-zero chance this will tweak some existing badges, but it feels like the correct thing to do regardless

PR #8472, opened 2022-10-03
allow/validate bitbucket_username / bitbucket_password in private config schema

### LINKED ISSUE #8445: BITBUCKET_USER and BITBUCKET_PASS can't be passed as environment vars

My own instance of shields

### 🐞 Description

I try to use the official docker image of shields.io on-premises.
When invoking docker run with env params runs fine, eg:

`docker run -p 16880:16880 -e PORT=16880 shieldsio/shields:next`

However when passing our bitbucket credentials to reach private repos it fails

`docker run -p 16880:16880 -e BITBUCKET_USER=xxxx -e BITBUCKET_PASS=xxxx shieldsio/shields:next`

```
Configuration:
{
  bind: { address: '0.0.0.0' },
  metrics: {
    prometheus: { enabled: false, endpointEnabled: false },
    influx: { enabled: false, timeoutMilliseconds: 1000, intervalSeconds: 15 }
  },
  ssl: { isSecure: false },
  cors: { allowedOrigin: [] },
  services: {
    github: {
      baseUri: 'https://api.github.com/',
      debug: { enabled: false, intervalSeconds: 200 }
    },
    obs: { authorizedOrigins: 'https://api.opensuse.org' },
    weblate: { authorizedOrigins: 'https://hosted.weblate.org' },
    trace: false
  },
  cacheHeaders: { defaultCacheLengthSeconds: 120 },
  handleInternalErrors: true,
  fetchLimit: '10MB',
  userAgentBase: 'shields (self-hosted)',
  requestTimeoutSeconds: 120,
  requestTimeoutMaxAgeSeconds: 30,
  requireCloudflare: false
}
file:///usr/src/app/core/server/server.js:277
      throw Error(
            ^

Error: Private configuration is invalid. Check these paths: bitbucket_username
    at Server.validatePrivateConfig (file:///usr/src/app/core/server/server.js:277:13)
    at new Server (file:///usr/src/app/core/server/server.js:230:32)
    at file:///usr/src/app/server.js:56:23
    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:533:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)

```

### 🔗 Link to the badge

_No response_

### 💡 Possible Solution

Not a node developer, however taking a look at the code, here

https://github.com/badges/shields/blob/8821ff85fcd494ebb802bf9bb26ddcba11d21ec8/core/server/server.js

It looks like `privateConfigSchema` does not include `bitbucket_username` nor `bitbucket_password`, only `bitbucket_server_username `and `bitbucket_server_password`

https://github.com/badges/shields/blob/8821ff85fcd494ebb802bf9bb26ddcba11d21ec8/core/server/server.js#L160

chris48s:
closes #8445


ferarias (approved):
looks good

calebcartwright (approved):
I have a very mild case of paranoia because I feel like there was something we did intentionally once upon a time relative to Bitbucket cloud, but I can't find any supporting evidence and don't recall any specifics so I do not have a high level of confidence.

This seems like a simple and reasonable enough thing to do, so I'll chalk my paranoia up to a weird dream :shrug:

chris48s:
I think its just oversight. The one thing I can think of where setting credentials for this could cause a problem would be if we weren't setting `authorizedOrigins` properly, but we're hard-coding an appropriate value in https://github.com/badges/shields/blob/ae018cdddc34b1638868edb6c05defa19844d691/services/bitbucket/bitbucket-pull-request.service.js#L66-L73 so I am not worried about this.

I think in general, the BitBucket badges are a bit of a patchwork: Not all of them work with both BB server and BB cloud. Not all of them support auth if you have credentials set. There's definitely some tidy up needed, but I don't think this causes any problem.

Issue #8399, opened 2022-09-14
Badge request: OSSF Best Practices


miguelnietoa:
:clipboard: **Description**

[OpenSSF Best Practices Badge](https://bestpractices.coreinfrastructure.org/en) is used to show the level of security practices implemented in your open-source project.

> The [Open Source Security Foundation (OpenSSF)](https://openssf.org/) Best Practices badge is a way for Free/Libre and Open Source Software (FLOSS) projects to show that they follow best practices. Projects can voluntarily self-certify, at no cost, by using this web application to explain how they follow each best practice. The OpenSSF Best Practices Badge is inspired by the many badges available to projects on GitHub. Consumers of the badge can quickly assess which FLOSS projects are following best practices and as a result are more likely to produce higher-quality secure software.

- **Which service is this badge for:** 
  [OpenSSF Best Practices](https://bestpractices.coreinfrastructure.org/en)

- **What sort of information should this badge show?**
Examples in plain text:
    - `openssf best practices | in progress 19%`
    - `openssf best practices | passing`
    - `openssf best practices | silver`
    - `openssf best practices | gold`

:link: **Data**
Data is get from an API request like this:
`https://bestpractices.coreinfrastructure.org/projects/:projectId/badge.json`

https://bestpractices.coreinfrastructure.org/projects/6367/badge.json

```json
{
  "id": 6367,
  "name": "Elixir Stellar SDK",
  "updated_at": "2022-08-16T19:44:26.876Z",
  "badge_level": "passing",
  "tiered_percentage": 109
}
```

- **Is there a public API?** Yes.
- **Does the API require an API key?** No.
- **Link to the API documentation.** https://github.com/coreinfrastructure/best-practices-badge/blob/main/doc/api.md

:microphone: **Motivation**

This feature should be implemented since lately it is becoming popular to add these kinds of badges related to security practices and supply chain to the README of open source projects.

I have noticed that the badge service already exists for OSSF Scorecard, but not yet for OSSF Best Practices, so I think is nice to have this badge available too!

miguelnietoa:
I would like to work on implementing this! 😄

chris48s:
Are there any docs for this API. Is there any rate limit on it we need to be aware of?

miguelnietoa:
Yes, @chris48s 

**API documentation:**
https://github.com/coreinfrastructure/best-practices-badge/blob/main/doc/api.md

**About rate limits:**
[...]

PR #8473, opened 2022-10-03
fix display/search of [cii] badge examples

chris48s:
Refs https://github.com/badges/shields/issues/8399
These badges weren't visible or searchable through the frontend because we were not registering any examples


calebcartwright (approved):
Who was the joker that couldn't spell this corr... oh wait, his name is me :disappointed:

lol

Issue #8146, opened 2022-07-01
Not valid Github status badge for broken workflow


dmvict:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

I noticed that `shields.io` provide status badge `passing` for broken workflow

![image](https://user-images.githubusercontent.com/47529590/176834948-c0569144-a9a5-49ab-8f6f-917962ef0236.png)

The real results is : 

![image](https://user-images.githubusercontent.com/47529590/176835200-5a1494dd-f0ca-4f44-8c8f-a9734fbf568b.png)

[The link to the workflow run](https://github.com/Wandalen/wTools/actions/runs/2591315400) on the screenshot. [Repository](https://github.com/Wandalen/wTools) where shields.io status badges are used. 

The badge works well with valid workflow files. And this workflow is fixed now. Please, try to reproduce and fix the bug.

### 🔗 Link to the badge

https://img.shields.io/github/workflow/status/Wandalen/wTools/mod_interface?label=

chris48s:
Our badges for github workflows are basically just "scrapers" for github's own badges so when you call https://img.shields.io/github/workflow/status/Wandalen/wTools/mod_interface/alpha it calls https://github.com/Wandalen/wTools/workflows/mod_interface/badge.svg?branch=alpha and then reports the value from that. It is essentially just a "wrapper" that allows you to use the full range of shields styles. Looks like your builds are now fixed so I now can't reproduce it but if you can break one again, check the actual github badges and see what value you've got.

dmvict:
> 

I missed this screenshot in the issue description
![image](https://user-images.githubusercontent.com/47529590/176968383-194f5282-b35b-4ec1-b979-af6de4173c40.png)

The screenshot was maiden just before the issue was opened. So, the bug exists, and the working case for the bug is a broken workflow.

If I will have time to test this scenario, then I will report the results. 
I will close the issue after approval of the correct badge work for the described case.

calebcartwright:
Unfortunately I suspect this is reflective of some upstream changes GitHub have rolled out, relative to #7102

Here's what GitHub shows for some of your branches (screen grab at the bottom in case statuses change)
https://github.com/Wandalen/wTools/actions/workflows/ModuleModInterfacePush.yml/badge.svg?branch=master
![](https://github.com/Wandalen/wTools/actions/workflows/ModuleModInterfacePush.yml/badge.svg?branch=master)
[...]

Issue #4587, opened 2020-01-28
Switch GitHub Workflow Status Badge to API


calebcartwright:
There's now GitHub Actions/Workflow specific APIs that are available, and we'll probably want to switch over to use them (we're currently svg scraping since APIs weren't readily available/limited when we first rolled out support)

Details on the v3 API linked below, though we should probably see if the v4 endpoints are available first.

https://developer.github.com/v3/actions/
https://github.com/badges/shields/pull/3898#issuecomment-579006954

calebcartwright:
One of the things that's going to be interesting with this is that the Actions API only supports specifying the desired Workflow by id or by the yaml file name where the Workflow is defined

https://developer.github.com/v3/actions/workflow_runs/#list-workflow-runs

However.. our current Workflow Status badge (which is svg scraping the native badge because the API wasn't readily available back then) is based off a third identifier, the Workflow _name_. 

On order to maintain BC for those existing badges, I see two paths forward:
1) We could keep the svg scraping class that will maintain BC for workflow-name based badges, and create a new one that leverages the API and accepts the workflow file or id 
2) We maintain BC for workflow-name param by making an extra API call ([list repository workflows](https://developer.github.com/v3/actions/workflows/#list-repository-workflows)) to extract the Workflow id, and then make the same API call.

Would there be any quota hit concerns with that second option? My sense is that we've got plenty of quota with the existing GH token pool, though I'm not sure if folks would have concerns with making multiple GH calls to render one badge

vitr:
Have you got any updates on this? 
Please, correct me if I'm wrong, but without the switch, it's impossible to display private repo badges even on a self-hosted instance.

simonhyll:
Ran into this issue while researching why my self hosted instance could create badges for things like commit activity but not for workflows. I assume this still isn't fixed.

What's the current issue that needs to be resolved? No one just hasn't bothered making it yet? Should I take a look at it?

> On order to maintain BC for those existing badges, I see two paths forward:

Can't you just do it as 2 queries, one for a complete list of all workflows in the repo, then match against the workflow name to get the id? Or maybe add a flag to change the behavior from workflow name to workflow id / yaml file? Personally I'd prefer specifying the yaml file since the name can result in a more awkward looking url.

chris48s:
This is a useful bump in the context of:

- https://github.com/badges/shields/issues/8146
- https://github.com/badges/shields/pull/8475

[...]

Issue #8398, opened 2022-09-14
Github actions badge not working on private repo


danielgroen:
### Are you experiencing an issue with...

My own instance of shields

### 🐞 Description

I have a private repo on github.

Also i have my own hosted shields server:
https://githubshieldsio.herokuapp.com

This is all set with a PAT token so it should read all the info from the repo.

Now the badge of the workflow can't find the repo.
When i set the repo to public, then the badge does work!

### 🔗 Link to the badge

Example of a badge reading the same repo (but does work)
https://githubshieldsio.herokuapp.com/github/package-json/v/danielgroen/affiliate-program

The badge that doesn't work.
https://githubshieldsio.herokuapp.com/github/workflow/status/danielgroen/affiliate-program/prod

LitoMore:
See https://github.com/badges/shields/discussions/5676.

chris48s:
Unlike most of our github badges which call an API endpoint, this one is basically a scraper or wrapper for github's own badge that allows you to apply the shields styles/icons/etc to it.
You will find that if you call https://github.com/danielgroen/affiliate-program/workflows/prod/badge.svg when you're logged in to github, it works. If you call it when you're not logged in, its a 404. With all our other badges that use the GitHub API, providing a token allows us access to private resources. This badge is an outlier in this respect. If it is possible to use a token in the request (I've not checked), we'd happily take a PR for this but at the moment this specific badge won't work for private repos.

calebcartwright:
Going to close this as a duplicate of #4681 in favor of keeping things consolidated under #4587

PR #8475, opened 2022-10-03
switch from github workflows to github actions workflows; test [githubactionsworkflowstatus githubworkflowstatus]

chris48s:
Refs #8146
Refs #4587
This PR implements the solution suggested in https://github.com/badges/shields/issues/8146#issuecomment-1262572065 - deprecate the old route and make a new one.


services/github/github-actions-workflow-status.service.js
```diff
@@ -0,0 +1,85 @@
...
+}).required()
+
+const keywords = ['action', 'actions']
+
+export default class GithubActionsWorkflowStatus extends BaseSvgScrapingService {
+  static category = 'build'
+
+  static route = {
+    base: 'github/actions/workflow/status',
+    pattern: ':user/:repo/:workflow+',
```
chris48s, line 24: A workflow file can be in a subdirectory. This means `:workflow` can contain a slash, so I've made this `:workflow+` and converted branch to a query param.

services/github/github-workflow-status.service.js
```diff
@@ -1,100 +1,11 @@
...
-  event: Joi.string(),
-}).required()
-
-const keywords = ['action', 'actions']
-
-export default class GithubWorkflowStatus extends BaseSvgScrapingService {
-  static category = 'build'
-
-  static route = {
+export default deprecatedService({
```
chris48s, line 3: Normally when we deprecate a badge we are just getting rid of it completely. Often the upstream service has been retired, or whatever. If we can provide a migration path from old to new, we use a redirect. In this case, we can't do either of those things because the `:workflow` param is not compatible between the two badges. Additionally, the GitHub workflow status badge is one of our most popular badges :grimacing: I think deprecating the old route and making people switch is the right thing to do because the behaviour of the existing badge is unexpected but its going to generate some support requests/anger/confusion/etc. Before we merge this I definitely want to write up a pinned issue (I've not written it yet, but I will) explaining what is going on and how to migrate.

Because we've never had this situation where we're asking people to switch and we can't provide a redirect, we've never done this before, but I wonder if instead of serving `build | no longer available` we make the issue and serve `build | github.com/badges/shields/issues/9999` (also use the `link` property for any `<object>` embeds) to help direct people towards upgrade advice. I'm somewhat conflicted as it is mainly info for maintainers and most of the people who will see it are going to be end users who can't do anything about it.

calebcartwright:
Acknowledge seeing this one, processing the proposed options

chris48s:
In light of #8398 and #4587 I've pushed 7ea7cf5 which updates the implementation of this to use the GitHub API instead of scraping the badge. This has the advantage it will allow self-hosting users to use it for private repos.

services/github/github-actions-workflow-status.service.js
```diff
@@ -0,0 +1,101 @@
...
+    return await this._requestJson({
+      schema,
+      url: `/repos/${user}/${repo}/actions/workflows/${workflow}/runs`,
+      options: {
+        searchParams: {
+          branch,
+          event,
+          page: '1',
+          per_page: '1',
+          exclude_pull_requests: 'true',
```
chris48s, line 87: These params are here for performance reasons. There's huge execution time difference between this query with `exclude_pull_requests=true` (fast) vs `exclude_pull_requests=false` (slow)

services/github/github-actions-workflow-status.service.js
```diff
@@ -0,0 +1,101 @@
...
+      })
+    )
+    .required()
+    .min(0)
+    .max(1),
+}).required()
+
+const queryParamSchema = Joi.object({
+  event: Joi.string(),
+  branch: Joi.string().required(),
```
chris48s, line 23: I've made branch a required param for this badge. Not my favourite solution, but workflow can contain a `/` and `branch=HEAD` doesn't work so we can't make the default "whatever the default branch is" in one request, so required query param it is. Lets lay that marker down while we're making a breaking change.

I don't think "latest workflow status across all branches" is a sensible default.

calebcartwright: Do you know off hand if this has to explicitly be a branch or will the upstream endpoint accept any valid ref?

services/github/github-actions-workflow-status.service.js
calebcartwright: Interesting. Definitely understand the performance angle, but curious if there's any functional impacts? Am I correctly interpreting this as meaning that runs triggered by a pull request would be excluded?

services/github/github-workflow-status.service.js
calebcartwright: > but I wonder if instead of serving build | no longer available we make the issue and serve build | github.com/badges/shields/issues/9999 (also use the link property for any <object> embeds) to help direct people towards upgrade advice. I'm somewhat conflicted as it is mainly info for maintainers and most of the people who will see it are going to be end users who can't do anything about it.

I'd recommend going with your issue link suggestion. Completely agreed that 99% of those that see the badge won't be able to do anything about it, but having a `no longer available` message instead doesn't really improve their experience one way or another and I think having the issue link only increases awareness; some subset of those users that see the issue-link badge in someone else's project will have their own such badges in their respective repos that they _do_ need to take action on but may still be unaware

services/github/github-actions-workflow-status.service.js
chris48s: The docs say

> `branch` string
>
> Returns workflow runs associated with a branch. Use the name of the branch of the `push`.

Anecdotally, I tried some calls with a tag and a commit hash (which I expected would return a valid result if this could accept any ref). They returned 0 results.

services/github/github-actions-workflow-status.service.js
chris48s: Nope. Passing this means the `pull_requests` array in the JSON response is not populated, but we aren't using any values from it so it doesn't really matter here. We can still get the details of a workflow that is triggered by a PR. For example, if I call https://api.github.com/repos/badges/shields/actions/workflows/enforce-dependency-review.yml/runs?branch=frontend-gha&exclude_pull_requests=true (the `enforce-dependency-review.yml` workflow on our repo only runs on PR triggers), we get a response with all the info we need for a badge. Just the `pull_requests` array is empty.

services/github/github-workflow-status.service.js
chris48s: OK. I have made a placeholder issue https://github.com/badges/shields/issues/8671 so we have a known URL we can link to, and I've done that in de39617 . This should allow us to approve the code changes. I will work on writing up the issue explaining why we are making the change, what users need to do about it, etc before I deploy it.

services/github/github-workflow-status.service.js
```diff
@@ -1,100 +1,28 @@
...
+  async handle() {
+    return {
+      label: 'build',
+      message: 'https://github.com/badges/shields/issues/8671',
+      /*
+      This is a 'special' deprecation because we are making a breaking change
+      We've implemented it as a custom class instead of a normal
+      deprecatedService so that we can include link.
+      */
+      link: ['https://github.com/badges/shields/issues/8671'],
```
chris48s, line 24: Because `deprecatedService()` throws an exception, we can't customise it beyond the label/message so I've done this as a custom class. I think the number of cases where this will be actually clickable will be low but it is worth doing IMO. This is a sufficiently exceptional edge-case that I don't mind this being a bit special and wierd and custom.

chris48s:
Here is a draft of the issue text go to in https://github.com/badges/shields/issues/8671


---


# 🥱 TL;DR

If you have a github workflow `test.yml`

```yaml
name: Run Tests
...
```

You need to update your badge URL from

`https://img.shields.io/github/workflow/status/<user>/<repo>/Run%20Tests`
to
`http://img.shields.io/github/actions/workflow/status/<user>/<repo>/test.yml?branch=main`

There are two changes here:

- The workflow parameter is now the workflow filename, not the `name:` defined in yaml
- Branch is now a required query param

# ❓ Why are you making this change?

This change resolves issue the problem reported in https://github.com/badges/shields/issues/8146 After consulting with GitHub support, it transpired that identifying workflows using the name in the yaml file can be ambiguous in some situations and defaults to finding workflows on _any_ branch when a branch param is not specified, rather than the _default_ branch. Switching to identifying workflows by the workflow filename and making branch a required parameter resolves these problems. Two workflows can have the same name string in yaml, but filename is unique on a given branch.

We decided that dropping the old badges (using the workflow name) and requiring users to switch to the new URLs (rather than keeping the old ones "working") was the right choice for our users because in some cases the old badges were serving data which was inaccurate or unexpected in subtle ways. Serving accurate data is the most important thing for us to prioritise.

# 😭 You made a breaking change! Waa

We know. We try never to do this and it is not a decision we took lightly, especially given this is one of our highest traffic badges. In general, our philosophy is: If you embedded a shields badge in your README in 2013 and the upstream service is still online, that badge should still work today. We achieve this with almost no exceptions.

Occasionally we do need to change badge routes for various reasons, but usually we keep old URLs working using redirects. In this case, it was not possible to provide a transition because the new badge route uses a different parameter.

One of the reasons we try to never do this is because we don't really have a great way to communicate a breaking change to our users or provide warning before it happens. We have no idea who uses our service. We don't ask anyone to sign up, we don't track you. Hopefully given that constraint we have done a reasonable job of this.

# 📚 Further Reading

- https://github.com/badges/shields/issues/8146
- https://github.com/badges/shields/issues/4587
- https://github.com/badges/shields/pull/8475

calebcartwright:
Couple thoughts on the text from https://github.com/badges/shields/pull/8475#issuecomment-1336418529

> The workflow parameter is now the workflow filename, not the name: defined in yaml

Since the file can be placed in a subdirectory, I think it would be worth including an extra line in the issue to (hopefully) cover that preemptively. Something simple along the lines of the below should help IMO:

```
If your workflow file resides in a subdirectory, then be sure to include the subdirectory path along with the file name, e.g.

http://img.shields.io/github/actions/workflow/status/<user>/<repo>/sub_directory/test.yml?branch=main
```

> question Why are you making this change?

I'd recommend we open with a summary of #8146 since most people won't read and digest. At the root of all this was that GitHub essentially silently abandoned the endpoint we were using which semi-broke our badge that relied on that endpoint. 

It might then be worth underscoring this in the final section too; our hand was forced and we were left with no real alternative options

calebcartwright (approved):

chris48s:
Thanks. I'll do another draft of the issue text and deploy this in the next few days.

PR #8496, opened 2022-10-07
fix [wordpress] php version not set test

chris48s:
This plugin now does declare a PHP version requirement. Rather than trying to find another live package to test against I've just switched this to a mocked test.


calebcartwright (approved):
wfm

PR #8498, opened 2022-10-07
fix [pub] points badge test and example

chris48s:
Pub points is now a score out of 140.
I've decided not to change the `floorCount` thresholds for formatting, but maybe if this changes again we review it.


calebcartwright (approved):
sounds reasonable to me

PR #8500, opened 2022-10-07
fix [flathub] version error handling

chris48s:
Not Found behaviour has changed (for the worse) upstream


services/flathub/flathub.service.js
```diff
@@ -26,6 +31,13 @@ export default class Flathub extends BaseJsonService {
       schema,
       url: `https://flathub.org/api/v1/apps/${encodeURIComponent(packageName)}`,
     })
+
+    // the upstream API indicates "not found"
+    // by returning a 200 OK with a null body
+    if (data === null) {
+      throw new NotFound()
+    }
```
calebcartwright, line 39: :facepalm:

calebcartwright (approved):

Issue #1844, opened 2018-08-03
Show only examples/explanations for dynamic badges type which is selected


tooomm:
Currently all three are shown together, would be better to only show the selected and relevant ones:
![dynamic examples](https://user-images.githubusercontent.com/9874850/43668776-eeae6ffc-977e-11e8-95f9-ab04d81c3439.png)

tooomm:
bump :)

MonliH:
@tooomm by "selected and relevant ones," do you mean the one selected in the dropdown above? I could implement this if others think it provides a better UX.

tooomm:
@MonliH Exactly, once a user decides for a format, the other examples and query links are only disturbing and not helpful.

To be in line with the rest of the page, it makes sense to always show a single more general example that's always visible:
`https://img.shields.io/badge/dynamic/<TYPE>?url=<URL>&label=<LABEL>&query=<QUERY>&color=<COLOR>&prefix=<PREFIX>&suffix=<SUFFIX>`

And change that for the specific one once a data type is selected in the dropdown?

chris48s:
We have just launched a completely redesigned frontend https://shields.io/

The new version of the site has separate pages for the dynamic json/xml/yaml bages

- https://shields.io/badges/dynamic-json-badge
- https://shields.io/badges/dynamic-xml-badge
- https://shields.io/badges/dynamic-yaml-badge

so I'm going to close this issue

PR #8505, opened 2022-10-08
Only show example format for the dynamic badge that is selected in the frontend

MonliH:
This PR implements the suggestion in #1844. That is, it shows the example format for the current selected data type:
image

When no datatype is selected, this implementation shows a more general form, `https://img.shields.io/badge/dynamic/<DATA_TYPE>?url=<DATA_URL>&label=<LABEL>&query=<QUERY>&color=<COLOR>&prefix=<PREFIX>&suffix=<SUFFIX>`


calebcartwright:
Really appreciate the PR, but I'm probably a :-1: on this one for a few reasons.

First, we've consistently gotten reports from users that the existing UI is unclear/doesn't necessarily convey relevant features. I can understand where someone deeply familiar with using the dynamic badge builder in that section of the page would view the other examples as visual noise, but I think having it present (with zero user action) for new users is more important than reducing that visual noise.

Secondly, we're in need/planning on overhauling the frontend entirely, and I don't think it's worth making any minor changes which aren't objectively, non-controversial improvements to the existing ui (refs #1899 and #7815).

I move we close this, but will wait to hear what @chris48s thinks

MonliH:
Ah I understand. That's actually what I thought after investigating the redesign efforts (very exciting stuff!). Fair enough to not want to make minor changes like this when most of the UI will be overhauled anyway.

calebcartwright:
Thanks again @MonliH but going to go ahead and close given the above discussion

PR #8508, opened 2022-10-09
feat: add [GitlabLastCommit] service

sunny0826:


services/gitlab/gitlab-last-commit.service.js
```diff
@@ -0,0 +1,78 @@
...
+      errorMessages: errorMessagesFor('project not found'),
+    })
+  }
+
+  async handle(
+    { project },
+    { gitlab_url: baseUrl = 'https://gitlab.com', ref }
+  ) {
+    const data = await this.fetch({ project, baseUrl, ref })
+    return this.constructor.render({ commitDate: data[0].committed_date })
```
chris48s, line 76: Sorry it has taken a while to get to this one. One small suggestion: if we're going to reference `data[0]` here we want to make sure data has at least one element. Lets make the schema

```js
Joi.array()
  .items(
    //...
  )
  .required()
  .min(1)
```

Saying this has also made me realise we don't have this on the github version of this badge. I'll make a quick PR fixing that.

Aside from this one minor point, this looks good to merge :+1:

chris48s (approved):
nice work - thanks :+1:

PR #8509, opened 2022-10-09
Update README.md

simonhyll:
Added an example for using shields.io as a service in a Github Actions workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,47 @@ There is documentation about [hosting your own server][self-hosting].
 
 [self-hosting]: https://github.com/badges/shields/blob/master/doc/self-hosting.md
 
+## Github Actions
+
+You can use [shields.io](https://shields.io) as a service inside a Github workflow. This can be useful for private repositories, organizations that do not wish to rely on the availability of our hosted service, or in conjunction with custom scripts to act as a render service for custom badges.
+
+```yaml
+name: Shields.io
+on:
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    services:
+      shields:
+        image: shieldsio/shields:next
+        ports:
+          - 80:80
+        options: >-
+          --health-cmd "wget --no-verbose --tries=1 --spider http://localhost:80 || exit 1"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Create badge
+        run: curl http://localhost/badge/hello-world-green > hello-world.svg
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          branch: bot/hello-world-badge
+          title: "[Bot] Update to hello-world.svg"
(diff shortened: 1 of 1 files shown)


calebcartwright:
Thank you for the PR!

I'd be curious to understand the use case and motivation though. Spinning up an entire badge server all just to get an svg on the local file system seems like a lot of unnecessary work. Using the [badge-maker](https://www.npmjs.com/package/badge-maker) package (which we produce and which is utilized by the server) seems like it would be a much simpler and more straightforward approach

calebcartwright:
> And I disagree that badge-maker is more straightforward mostly because badge-maker requires more manual labour to create a lot of the badges that the full shields.io experience offers. Getting your commit activity for example would require you to write the necessary source code in Node.js to get the data and then use badge-maker to create the badge. With the service approach all that work is done for you.

I'm working off the text of your PR, and the example you've proposed is using a static badge, not a dynamic badge like a GitHub commit activity.

So while I take your point that your model is easier for such an alternative scenario, it doesn't really match the text of what you initially submitted and the associated context in which my comment was made.

calebcartwright:
I view this as documenting two items:

* explicitly, how to run the Shields docker image in a CI/pipeline tool (GitHub Actions)
* implicitly, some process/workflow for using svg files ostensibly checked into a repository that are dynamically updated 

I'm not vehemently opposed to incorporating docs for the pipeline contexts, but I do feel fairly strongly that (a) we cannot have multiple copies/versions of the material (refs #8516) and (b) it doesn't belong in the main readme. I do have reservations about including the secondary action for creating a PR (I believe I get the gist of what it's going for, but doubt that will be inherently obvious for our typical user).

@chris48s thoughts? I think most of my ambivalence around whether to pull this in stems from consideration around any ongoing overhead (GHA syntax changes, etc.), any additional pipeline tools people would want to add, etc.

simonhyll:
I made two PR's for it in different places partly so you could pick which place it belongs in, and also because it wasn't until after I had already made the one to the main README.md that I realized it's probably better to put it in the self hosting document instead.

chris48s:
My reaction to this is it is a creative use of shields, and perhaps an interesting blog post for you. If you did write it up, it is the sort of thing we'd link to from https://github.com/badges/awesome-badges , but probably not something we'd position as a core use-case.

Tangent: The concept of using the badge-maker lib to generate badges in a GitHub action and then write the image back to the repo (as an alternative to self-hosting, a way to address use-cases for which a centralized service can't serve well and a way to write badges that rely on libraries/tooling that don't exist in the JS ecosystem) is something I've been casually noodling about with under the https://github.com/action-badges org. I've not really made any attempt to promote it, but if you want to have a nose about https://github.com/action-badges/core is a good places to start.

Another thing I've seen people do is GitHub actions that write a JSON file to a public repo using the endpoint badge schema, then they use shields.io to render it. https://upptime.js.org/docs/badges/ is the slickest example of that, but there are others doing similar things.

There's some cool stuff going on with making badges as an output of GitHub actions/workflows, and I do think there is mileage in it. I've not really seen or arrived at a pattern I'd solidly recommend/support for an audience the size of the one we have for this repo though.

I'm :-1: on merging this or https://github.com/badges/shields/pull/8516 but it is always cool to see when people in the community are finding interesting things to do with the project.

simonhyll:
Guess I finally have a reason to start a developer blog then 😅

PR #8510, opened 2022-10-09
tidy up [GitHubGist] routes

chris48s:
We've merged a couple of PRs recently adding badges for GitHub gists.
Both of these made quite different decisions about what route to use. In particular, the stars one could have caused a conflict. Fortunately, you can't create a github username or team called "gists" :relieved: . Bit of sloppy code review on my part there.

In this PR, I propose that the route we should use for github gist badges is

`/github/gist/NOUN/PARAMETERS` with `gist` being singular

and then I've updated the 2 existing badges + added redirects.

I'm open to feedback on that. If we think that `/github-gist`, `/github-gists` or `/github/gists` would be better as a base than `/github/gist` I am happy to update the PR. Whatever we decide, lets: pick one, use it consistently and continue to do so moving forwards.

The other thing I've done in this PR is moved the gist badges to a subdir. The `/github` dir already has a lot of files in it, so I think it is useful to compartmentalise any further gist badges.


calebcartwright:
Unfortunately we didn't have much discussion on this during the original PR, but I'd attempted to tee up the same topic in https://github.com/badges/shields/pull/8272#discussion_r939464149

I don't personally have strong opinions either way, though I know in the past others have expressed more rigorous preferences for having a singular noun in the route. The most important element for me is consistency within the "family" (gists in this case), but also trying to be cognizant of consistency across the service offering.

chris48s:
Yeah agreed. I care more about establishing the pattern and using it moving forwards than exactly which variant we go with.

Just thinking it through a bit more:

- Nouns in routes are singular if there is one of the thing: There's only one gist
- GitHub brand the service as "GitHub Gist" (not "GitHub Gists")

Screenshot at 2022-10-12 20-45-18

which strengthens the case for singular.

The first part of the route should be the service. Gists are kind of an odd feature of GitHub, but I think I'd fundamentally characterise gists as a feature of GitHub, rather than a separate service. For example, we get the data from the GitHub API. I think that leads me back to `/github/gist` as the base.

PaulaBarszcz:
This PR seems helpful and complete - can we merge it or does it still need something before being ready for a merge? :)

calebcartwright (approved):

PR #8514, opened 2022-10-09
fix [tas] private project test

chris48s:
This service test has been failing for a while.

There are a couple of things that might have happened here:
1. The project we're testing against has gone away
2. TAS have changed their API to just issue a 404 for private projects so that they don't expose the existence/otherwise of private projects (this is quite common)

I can't really answer which has happened and we're not going to be able to find another example of a private project so I'm going to provisionally suggest just mocking this case.

That said, maybe @nevilm-lt or @palashmarantas could weigh in on this one? Presumably one of them owns (or owned) `tasdemo/nexe-private`? If the project is still around, we can probably assume a behaviour change from TAS and remove the special-case for private projects.


chris48s:
no word on this one, so I'm just going to suggest we go ahead with mocking it

calebcartwright (approved):
wfm

PR #8519, opened 2022-10-10
increase timeout for [reuse] not found test

chris48s:
The 'not found' case on this badge seems to consistently take > 5 seconds for this service. Not much we can do about the upstream, but I'd like to get rid of the noise from the service tests.


calebcartwright (approved):
Think this is largely a duplicate of #8073 but fine moving forward with the higher timeout limit regardless

chris48s:
Bah I'm not sure how I missed #8073 - sorry.
FWIW, I tried increased it to 10 seconds initially and it still failed in CI so I bumped it up again in a force-push.

PR #8538, opened 2022-10-16
fixes for integration test rate limit issues

### LINKED ISSUE #8535: failing integration tests
:clipboard: **Description**

sigh

5 days ago we merged https://github.com/badges/shields/pull/8423
First dependabot barrage comes along, we've got failing builds with "API rate limit exceeded", "Error: Token pool is exhausted", etc
The token that GitHub allocates for workflow builds has a rate limit of 1,000 requests per hour. That's less than a PAT, but still quite a lot. I'm surprised we are managing to hit it, but we are.
I pushed a build to a branch that calls https://api.github.com/rate_limit just to check that the token definitely has a rate limit of 1,000 and we're definitely maxing it.

<details>
    <summary>We are</summary>
<pre>
{
  "resources": {
    "core": {
      "limit": 1000,
      "used": 1000,
      "remaining": 0,
      "reset": 1665865803
    },
    "search": {
      "limit": 30,
      "used": 0,
      "remaining": 30,
      "reset": 1665864671
    },
    "graphql": {
      "limit": 1000,
      "used": 0,
      "remaining": 1000,
      "reset": 1665868211
    },
    "integration_manifest": {
      "limit": 5000,
      "used": 0,
      "remaining": 5000,
      "reset": 1665868211
    },
    "source_import": {
      "limit": 100,
      "used": 0,
      "remaining": 100,
      "reset": 1665864671
    },
    "code_scanning_upload": {
      "limit": 1000,
      "used": 0,
      "remaining": 1000,
      "reset": 1665868211
    },
    "actions_runner_registration": {
      "limit": 10000,
      "used": 0,
      "remaining": 10000,
      "reset": 1665868211
    },
    "scim": {
      "limit": 15000,
      "used": 0,
      "remaining": 15000,
      "reset": 1665868211
    },
    "dependency_snapshots": {
      "limit": 100,
      "used": 0,
      "remaining": 100,
      "reset": 1665864671
    }
  },
  "rate": {
    "limit": 1000,
    "used": 1000,
    "remaining": 0,
    "reset": 1665865803
  }
}
</pre>
</details>

sooo.. I need to work out what we're going to do about that.

One obvious problem is that the integration build is running on both push and pull-request. Dependabot pushes to this repo, rather than a fork which means the builds are doubling up so we can probably half the usage by fixing that. Looking at the number of failures we have, that will not be a complete solution.

This will also get worse once we've got danger and the service tests sharing the same token.

chris48s:
closes #8535

Right. A few things going on here:

First 2 commits should be pretty clear. Just trying to run fewer pointless builds really. Regardless of if the builds are consuming rate limit or not, this is good. Less noise, faster builds, etc

The crucial commit is actually the third one though. What I've done is set a Personal Access Token as both a [repo secret](https://github.com/badges/shields/settings/secrets/actions) and a [dependabot secret](https://github.com/badges/shields/settings/secrets/dependabot). Then in the build, we use the PAT if it is available, or fall back to the standard action token (which has the lower rate limit of 1000/hour) otherwise. This will mean PRs submitted from forks can use the standard token without us having to expose a PAT while pushes to master, dependabot PRs, etc can use the PAT with the higher rate limit. That should stop dependabot/repo ranger from completely smashing our rate limit.

### DIFF
--- a/.github/workflows/auto-close.yml
+++ b/.github/workflows/auto-close.yml
@@ -1,5 +1,7 @@
 name: Auto close
-on: pull_request_target
+on:
+  pull_request_target:
+    types: [opened]
 
 permissions:
   pull-requests: write
--- a/.github/workflows/enforce-dependency-review.yml
+++ b/.github/workflows/enforce-dependency-review.yml
@@ -1,5 +1,7 @@
 name: 'Dependency Review'
-on: [pull_request]
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize]
 
 jobs:
   enforce-dependency-review:
--- a/.github/workflows/test-integration-17.yml
+++ b/.github/workflows/test-integration-17.yml
@@ -1,6 +1,7 @@
 name: Integration@node 17
 on:
   pull_request:
+    types: [opened, edited, reopened, synchronize]
   push:
     branches-ignore:
       - 'gh-pages'
--- a/.github/workflows/test-integration-17.yml
+++ b/.github/workflows/test-integration-17.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches-ignore:
       - 'gh-pages'
+      - 'dependabot/**'
 
 jobs:
   test-integration-17:
--- a/.github/workflows/test-integration-17.yml
+++ b/.github/workflows/test-integration-17.yml
@@ -9,6 +9,8 @@ on:
 jobs:
   test-integration-17:
     runs-on: ubuntu-latest
+    env:
+      PAT_EXISTS: ${{ secrets.GH_PAT != '' }}
 
     services:
       redis:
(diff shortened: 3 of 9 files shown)


chris48s:
@calebcartwright - if you get a chance, it would be really useful to get a :+1: on this one before the dependabot PRs come in for the week. In the current configuration we can only merge about 2 or 3 PRs per hour, so merging this one would be super high impact.

calebcartwright:
Actually have a question on f0ad7b6b75b72c5347fc6e52aae1059a38753b62 relative to disabling the trigger for pushes to dependabot branches (assuming IUC). What's the potential impact of this in cases where we need dependabot to rebase/recreate, as those involve force pushing to the branch associated with the already created PR, and then the more rare cases where we have to manually push some commits to a dependabot branch ourselves?

chris48s:
As I understand it, a dependabot rebase or recreate should trigger the unhelpfully named `pull_request: synchronize` event

> `synchronize`: Triggered when a pull request's head branch is updated. For example, when the head branch is updated from the base branch, when new commits are pushed to the head branch, or when the base branch is changed.

but it is possible this might need another round of tweaks. If we find there are some pull request events which are not triggering a rebuild, we could try inverting the logic: disable workflows for pull requests where the base branch is `dependabot/**` and enable them for pushes.

calebcartwright (approved):
Sounds reasonable, let's give it a whirl

PR #8539, opened 2022-10-17
allow passing key to [stackexchange]

### LINKED ISSUE #3591: Production servers are hitting Stack Overflow access quota

- [ ] My own instance

:beetle: **Description**

Badge for StackOverflow fails to provide the correct questions intermittently

:link: **Link to the badge**

https://img.shields.io/stackexchange/stackoverflow/t/point-cloud-library.svg?logo=stackoverflow

Live image:
SO badge

Discussion: https://github.com/PointCloudLibrary/pcl/issues/3175

:bulb: **Possible Solution**
This might not be the cause, but caching SO results might prevent a non-response from their end (if that's the cause).

Manually calling the SO API that shields.io is supposed to call results in a correct response which led to this hypothesis.

### LINKED ISSUE #5415: Problems with using StackOverflow reputation badge
you experiencing an issue with...

- [ ] My own instance

:beetle: **Description**

When I use _StackOverflow reputation_ badge without any parameters everything works okay:

But when I add parameters in order to change badge color or shape, it says there's `invalid parameters`

Also, if I want to change title from `stackoverflow reputation` to `StackOverflow reputation`:

:link: **[Link to the badge](https://img.shields.io/stackexchange/stackoverflow/r/11589183)**

:bulb: **Possible Solution**

¯\_(ツ)_/¯

chris48s:
closes #5415
closes #3591

This is the code bit. It allows us to optionally pass a key when calling the StackExchange API. Once deployed, I'll also need to set a `stackapps_api_key` in prod. I've tested locally with a key and can confirm that without auth our quota is 300 per day and passing a key raises our quota from 300 requests per day to 10,000. This should cover our current usage (although these have been unreliable for a long time, so I suspect that contributes to the low traffic on these badges).

If we ever outgrow that:

- I think there is a mechanism to request an increased quota if we need it
- We could look at increasing the cache lengths on these


doc/server-secrets.md
```diff
@@ -244,6 +244,17 @@ Create an account, sign in and obtain a uuid and token from your
 to give your self-hosted Shields installation access to a
 private SonarQube instance or private project on a public instance.
 
+### StackApps
```
calebcartwright, line 247: Something I've often wondered about is how cognitively transferable one Stack* is to another for the average dev. I.e. is someone looking at this from a purely StackOverflow-centric frame of mind going to know to look/search for StackApps or StackExchange?

services/stackexchange/stackexchange-base.js
```diff
@@ -0,0 +1,37 @@
...
+    color: floorCountColor(numValue, 1000, 10000, 20000),
+  }
+}
+
+export class StackExchangeBase extends BaseJsonService {
+  static category = 'chat'
+
+  static auth = {
+    passKey: 'stackapps_api_key',
+    authorizedOrigins: ['https://api.stackexchange.com'],
```
calebcartwright, line 24: Seems reasonable. Worth noting for future consideration that there is an enterprise/self-hosted options available for the Stack universe. Not sure how widespread the adoption is yet, and we've obviously not had any requests so it's def not something that needs to be tackled as part of this.

calebcartwright (dismissed):
lgtm. couple inline thoughts but nothing that needs to block

doc/server-secrets.md
cdhunt: `StackApps` is where we keep track of API registration. It's not really meaningful to a typical user.  Stack Overflow or Stack Exchange Network would be more recognizable names.

services/stackexchange/stackexchange-base.js
cdhunt: As far as I know, non-Admin users on private instances can't provision API keys and there's no unauthenticated access. If a user got their site admin to provision a key, the functionality should work the same.

doc/server-secrets.md
chris48s: I've tweaked the header to include all 3

chris48s:
Modifying the docs has dismissed the review. If you can give me a quick re-:+1: on this one I'll sort out getting it deployed with a key over the weekend.

calebcartwright (approved):

PR #8540, opened 2022-10-17
fix debug logging of undefined query params

chris48s:
If we pass `{ someparam: undefined }` to got, `someparam` will not be passed at all in the querystring. When we log the params to console (for example when running `npm run badge`), this currently appears as `?someparam=undefined`. This PR switches us to dropping these params completely, in line with the behaviour of got.


calebcartwright:
Not necessarily opposed to this, but curious as to how feasible/preferable it might be to adjust this in our logging (less common) vs. what I imagine is essentially duplicative processing of got's internal behavior that's happening at runtime for every request

chris48s:
We could define a custom version of `logTrace` just for when we log the request URL.

It has been a while since I looked at this but I think the reason why I thought stripping the `undefined` values and then passing the modified `params` object to both `got` and `logTrace` is this way we are passing the exact same object to both, rather than logging a slightly different thing than what we passed got. There's not a lot in it though.

calebcartwright (approved):
Perhaps a bit overly cautious from my end but I'd be inclined to deploy this in isolation

chris48s:
Yeah good call. I'll leave this one to mature for a bit :cheese:

chris48s:
I've just pushed this to prod. I'll keep an eye on sentry/issue tracker over the next few days and plan to hold off doing another deploy until the weekend

PR #8544, opened 2022-10-18
Fix [Testspace] with new "untested" value in case_counts array

Aldaviva:
Hello,

I noticed that all Testspace badges started returning "invalid response data" recently. The cause was a recent change in the response object of the Testspace web API: the `case_counts` array in the [Results object](https://help.testspace.com/reference/web-api#results) now has length 5 instead of 4. A new integer was appended to the end which represents the number of [untested](https://help.testspace.com/manual/execution-session#start) tests, which are [manual](https://help.testspace.com/manual/overview) tests that have not been run yet.

|Live example|Screenshot for posterity|
|---|---|
|Testspace tests|tests: invalid response data|

The API docs haven't been updated yet as of 2022-10-17, but I [contacted the Testspace developers and figured out what the new value means](https://github.com/testspace-com/testspace.support/discussions/29).

To fix the invalid error, this change updates the schema for the Testspace API response with the new array length. It also adds a new `metric` parameter value (`untested`) to the `TestspaceTestCount` service and includes it in the internal object returned by `TestspaceBase.transformCaseCounts()`, as well adding it to the `total` value in that object. The documentation URLs have also changed.

I didn't add `untested` to the `TestspacePassRatio` or `TestspaceTests` services, since those already didn't use `skipped` and `errored`, respectively, and it didn't seem like it would make sense to start including untested manual cases there.

I think I've covered all the requirements for contributing, but let me know if I missed anything or if any other changes need to be made.

Thanks!

### DIFF
--- a/services/testspace/testspace-base.js
+++ b/services/testspace/testspace-base.js
@@ -2,8 +2,8 @@ import Joi from 'joi'
 import { nonNegativeInteger } from '../validators.js'
 import { BaseJsonService, NotFound } from '../index.js'
 
-// https://help.testspace.com/docs/reference/web-api#list-results
-// case_counts|array|The contained cases [passed, failed, na, errored]|counters of result
+// https://help.testspace.com/reference/web-api#list-results
+// case_counts|array|The contained cases [passed, failed, na, errored, untested]|counters of result
 // session_* fields are for manual runs
 // There are instances where the api returns a 200 status code with an empty array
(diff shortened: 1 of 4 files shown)


chris48s (approved):
Cheers. I'd spotted this was failing on the nightly service tests runs but I hadn't had a chance to dig into it yet.

PR #8547, opened 2022-10-20
remove failing [lgtm] test case

chris48s:
The upstream project has disappeared and unfortunately I'm not able to find another public project which is hosted on bitbucket and using LGTM. Given that the response we expect from LGTM is identical regardless of where the source is hosted and there's no difference in the logic there is no value in replacing with a mocked test IMO. I suggest we can just drop it.


calebcartwright (approved):
Fair enough

Issue #8207, opened 2022-07-20
Add support of Visual Studio Code prerelease extensions to the version badge


lextm:
:clipboard: **Description**

Visual Studio Code started to support prerelease versions of extensions for a while, but the version badge (`/visual-studio-marketplace/v/:extensionId`) does not support switching between the stable/prerelease channels,

https://code.visualstudio.com/updates/v1_63#_pre-release-extensions

For end users, they might wonder why a version (prerelease) showed in this badge is not available to install on their machines.

chris48s:
Most of our version badges have 2 variants:

- `/v/packageName`
- `/v/packageName?include_prereleases`

It would be nice if we could do the same here. Looking at the API response we get back from the VS Marketplace API I'm not sure if we get a pre-release flag. Do you have an example of an extension with a mix of pre and stable releases?

lextm:
You might use my extension as example,

https://marketplace.visualstudio.com/items?itemName=lextudio.restructuredtext

All old versions (189.x.x and below) are stable releases, while new versions (190.x.x) are all pre releases. Thus,

* A stable version badge should show 189.1.0.
* A prerelease version badge should show 190.1.4.

chris48s:
OK. Thanks for the example.
It looks like if we call

```
curl -X POST "https://marketplace.visualstudio.com/_apis/public/gallery/extensionquery/" --data "{\"filters\":[{\"criteria\":[{\"filterType\":7,\"value\":\"lextudio.restructuredtext\"}]}],\"flags\":0x10}" --header "Accept: application/json;api-version=3.0-preview.1" --header "Content-Type: application/json"
```

(note `"flags":0x10` in the POST body)

there is a

```js
{
  "key": "Microsoft.VisualStudio.Code.PreRelease",
  "value": "true"
}
```

value we can look for in the version `properties` array to identify pre-releases

OronW:
@chris48s 
Hi, 
I'm a first time contributor and would like to try and work on this issue :) 
I've read the relevant docs, cloned the repo and started digging in the code. 

I'd appreciate some general guidance in terms of the files to look at. From what I've read so far, it seems the relevant place would be the base class at `visual-studio-marketplace-base.js`, 
the service at `visual-studio-marketplace-version.service.js` 
and the tester. 

Am I correct?
Anything else that's important?

Please let me know if this could be assigned to me.
Thanks.

chris48s:
[...]

PR #8561, opened 2022-10-22
[VisualStudioMarketplace] Add support to prerelease extensions version (Issue #8207)

OronW:
Added pre-release version support to VScode marketplace service.
Updated response schema.
Updated Tests to include pre-release version.

PR fixing issue #8207


services/visual-studio-marketplace/visual-studio-marketplace-version.service.js
```diff
@@ -27,15 +41,22 @@ export default class VisualStudioMarketplaceVersion extends VisualStudioMarketpl
     return renderVersionBadge({ version })
   }
 
-  transform({ json }) {
+  transform({ json }, includePrereleases) {
     const { extension } = this.transformExtension({ json })
-    const version = extension.versions[0].version
+    const preRelease = 'Microsoft.VisualStudio.Code.PreRelease'
```
chris48s, line 46: I think the intent here would be a bit clearer if we called this var `preReleaseKey` or `preReleaseTag`

services/visual-studio-marketplace/visual-studio-marketplace-version.service.js
```diff
@@ -27,15 +41,22 @@ export default class VisualStudioMarketplaceVersion extends VisualStudioMarketpl
...
+    const version = includePrereleases
+      ? extension.versions[0].version
+      : extension.versions.find(
+          obj => !obj.properties.find(({ key }) => key === preRelease)
+        ).version
+
```
chris48s, line 52: There are a couple of problems here.

1. If there are only pre-releases and no stable releases this will fail because `.find()` will just return `undefined`. For this case, we should fall back to showing the latest pre-release if there are no stable releases at all (this is what we normally do for other badges). 
2. This one is a bit more of an edge case: It looks like the deal is pre-releases are indicated by `{"key": "Microsoft.VisualStudio.Code.PreRelease", "value": "true"}` and stable releases are indicated by complete absence of that object, but I think if we were to encounter a release with `{"key": "Microsoft.VisualStudio.Code.PreRelease", "value": "false"}` we should also consider that stable (whether or not we currently think this is likely). This code would classify that as a pre-release.

chris48s (commented):
Thanks for submitting this and including tests. It is looking in pretty good shape to start with and its clear you've worked through the documentation and contributing guidelines. :+1:

chris48s:
Minor point: I've edited the PR title so it will run the tests for the whole VSCode Marketplace service family on the next push. This PR modifies the `VisualStudioMarketplaceBase` class which impacts on all the badges that inherit from it.

services/visual-studio-marketplace/visual-studio-marketplace-version.service.js
OronW: I agree. Fixed.

OronW:
@chris48s 
Thank you for the review and all your comments!

I've updated my code accordingly, and I've added 2 more tests - 
1. For the case where there are only prereleases but the include_prereleases flag is false (fallback to the latest version)
2. For the case where we might get a "false" value in the prerelease object.

I've also run the entire VisualStudioMarketplace tests locally without issues.

Please let me know if there's anything else :)

chris48s (approved):
great first contribution :+1:

PR #8562, opened 2022-10-24
migrate e2e tests to GHA

chris48s:
This PR migrates the E2E tests from Circle to GitHub Actions

I haven't bothered with making any markdown summary for this (yet). The script I wrote for generating a markdown summary from mocha JSON output doesn't work for this because its cypress running the tests not mocha so it would need a whole new script/parser. I think it would be nice to improve this, but I feel like its not super common for contributions from someone outside the core team to break/require debugging this handful of end-to-end tests so this is much less urgent than making sure the outputs of the core tests, service tests, etc are exposed nicely.


calebcartwright (approved):
wfm

PR #8578, opened 2022-10-28
[Ansible] Add collection badge

s3lph:
Add a badge for Ansible collections, almost identical to those for Ansible roles.

Unfortuntately, the `/api/v2/collections` endpoint (as opposed to `/api/v1/roles`) of Ansible Galaxy exposes neither download numbers nor quality scores, so there's only the name badge for now.

### DIFF
--- a/services/ansible/ansible-collection.service.js
+++ b/services/ansible/ansible-collection.service.js
@@ -8,17 +8,7 @@ const ansibleCollectionSchema = Joi.object({
   }),
 }).required()
 
-class AnsibleGalaxyCollection extends BaseJsonService {
-  async fetch({ collectionId }) {
-    const url = `https://galaxy.ansible.com/api/v2/collections/${collectionId}/`
-    return this._requestJson({
-      url,
-      schema: ansibleCollectionSchema,
-    })
-  }
-}
-
-class AnsibleGalaxyCollectionName extends AnsibleGalaxyCollection {
+class AnsibleGalaxyCollectionName extends BaseJsonService {
   static category = 'other'
   static route = { base: 'ansible/collection', pattern: ':collectionId' }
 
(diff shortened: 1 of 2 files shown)


calebcartwright:
Thank you for the PR. A few things:

- it would be helpful if you could articulate the what+why for this badge along the same lines of the questions asked in our [issue template]() for new badges
- the tests you've added are failing, please take a look and get them passing
- given that there's a single class defined in one file, there's no need for an extra base class in the inheritance hierarchy; let's drop that base and just have the new class extend the basejson class directly

s3lph:
> * it would be helpful if you could articulate the what+why for this badge along the same lines of the questions asked in our issue template for new badges

:clipboard: **Description**

- This new badge is for **Ansible Galaxy**
- It shows the **name of an Ansible Collection**, as it is available on Ansible Galaxy
- Example: [Ansible Galaxy](https://galaxy.ansible.com/community/general)

:link: **Data**

- The data is retrieved from the public, unauthenticated Collections API of Ansible Galaxy
- https://galaxy.ansible.com/api/v2/collections/


:microphone: **Motivation**

- Collections are the "new" (if we can still consider it "new") packaging format for sharing content on Ansible Galaxy, allowing users to not only distribute single Ansible roles, but e.g. Python modules, plugins, ...
- Before that, it was only possible to share single Ansible roles on Galaxy.
- Similar to the already existing "Ansibl Role" badge, this badge shows the name of an Ansible Collection.

> * the tests you've added are failing, please take a look and get them passing

For some reason the tests of some Gitlab service are failing, even though I used the `[Ansible]` tag in the PR. Got any insights why this might happen?

> * given that there's a single class defined in one file, there's no need for an extra base class in the inheritance hierarchy; let's drop that base and just have the new class extend the basejson class directly

I copied this 1:1 from the Ansible Role service, where there are two additional services. Unfortunately, the data for these services are not (yet) exposed in the collections API. I'd propose to keep the base class around so that they can be added later on. Any thoughts?

chris48s (commented):
Your service tests are failing with

```
AssertionError: label mismatch: expected 'collection' to equal 'role'
```

You need to switch `label: 'role'` to `label: 'collection'` in your service tests. That will get the build passing.

---

> I copied this 1:1 from the Ansible Role service, where there are two additional services. Unfortunately, the data for these services are not (yet) exposed in the collections API. I'd propose to keep the base class around so that they can be added later on. Any thoughts?

I think while we just have the one badge lets drop the base class and move `fetch()` into `AnsibleGalaxyCollectionName`. We can factor it out later if we add more badges in future.

chris48s (approved):
nice work - thanks

PR #8584, opened 2022-10-31
[Coveralls] for GitLab

### LINKED ISSUE #8132: Badges for Coveralls.io code coverage do not support GitLab repositories

shields.io

### 🐞 Description

The Coveralls.io badges for code coverage support github and bitbucket only. 
However, Coveralls.io also integrates with GitLab. 
There is no option to select a GitLab repository and adding 'gitlab' to the url manually fails.

### 🔗 Link to the badge

_No response_

### 💡 Possible Solution

_No response_

PaulaBarszcz:
closes #8132 

Add GitLab option for Coveralls + move `branch` to query param.

coverage from: https://coveralls.io/gitlab/selcouth/wsrouter

image

http://localhost:8080/coverallsCoverage/gitlab/selcouth/wsrouter

---------------------
EDIT: the below is no longer valid, but leaving it here for posterity (in case someone runs into a similar problem and is curious to know what fixed it in this case).

I have a problem with an old redirect.
For some reason it allows multiple slashes in the `user` param and the `branch` param remains undefined.

I am not sure how this pattern from the Coveralls redirector is interpreted:
```
   route: {
      base: 'coveralls',
      pattern: ':user((?!github|bitbucket|gitlab).*)/:repo/:branch*',
    },
```

This: `(?!github|bitbucket|gitlab)` is probably a negative lookahead (https://stackoverflow.com/a/38893082).
This: `.*` looks a bit like an unnamed parameter in the `path-to-regexp` package (https://github.com/pillarjs/path-to-regexp#unnamed-parameters).

```js
npm run test:services:trace -- --only='coveralls'
```
image

image

image

When for a test I did:
image
it was interpreted as 
```
 Named params  🎫
 { user: 'jekyll/jekyll/master', repo: 'lol', branch: undefined }
```
image


PaulaBarszcz:
If someone could shed light on my redirector-issue from the description, that would be great 😃

chris48s:
Probably one thing worth stating up-front is: Fundamentally the point of these redirects is to keep existing badges working as we make routing changes, not to add multiple ways to address new badges. I've not had a chance to check this code out locally, but lets think through what we are actually trying to make happen with these redirects..

- `/coverallsCoverage/github/badges/shields` : shouldn't redirect - it was a valid URL before, its a valid URL now
- `/coverallsCoverage/bitbucket/myuser/myrepo` : shouldn't redirect - it was a valid URL before, its a valid URL now
- `/coverallsCoverage/github/badges/shields/somebranch` : should redirect to `/coverallsCoverage/github/badges/shields?branch=somebranch`
- `/coverallsCoverage/bitbucket/myuser/myrepo/somebranch` : should redirect to `/coverallsCoverage/bitbucket/myuser/myrepo?branch=somebranch`
- `/coverallsCoverage/gitlab/anything/anything/anything/anything...` : shouldn't redirect - this is a new route, so we're not trying to preserve backwards compatibility (which is the point of the redirect). There are zero badges in the wild today calling this with `/gitlab`

So I think one problem with your redirects is that you're including `|gitlab` in the capturing group.

I think another is probably that you're trying to make branch optional in the redirect route (`/branch*`) and that makes the routing ambiguous. We only want the new redirect to be triggering in the case where `vcsType` is `github` or `bitbucket` and there are 3 or more further params (`/branch+`). Doing that should make the redirect route specific enough that it works as intended. https://github.com/badges/shields/blob/55d711c39d5782eb10f9620c59245cc04f0ed7c2/services/gitlab/gitlab-pipeline-status.service.js#L114-L124 is probably a good example to crib from here.

With the old route (from `/coveralls` to `/coverallsCoverage`), I'd imagine you probably don't actually need to change anything? (again, we just want to keep existing badges working), but the explanation of what that is doing is here: https://github.com/badges/shields/pull/6181 but it is basically matching the first param as `:user` only if it is not equal to `github` or `bitbucket` (otherwise the first param is `:vcsType`).

You will probably find it useful to write some tests with `.expectRedirect()` to make sure this is all doing what you expect.

chris48s (approved):
Nice solution :)
Getting the redirects right on this one turned out to be a bit tricker than we first thought.

PR #8587, opened 2022-11-01
add 404 error message for dart's [pub]

### LINKED ISSUE #7922: Add a specific error message for 'invalid package name' on dart pub badges
:clipboard: **Description**

Refs https://github.com/badges/shields/pull/7921#issuecomment-1116129257
+tests

PaulaBarszcz:
closes #7922 

@chris48s In #7921 you mentioned that it would be good to add the error message for 400:
```js
errorMessages: {
  400: 'invalid package name',
},
```

Now these 2 pub endpoints return a `404` when given an incorrect package name,
so I added this instead:
```js
errorMessages: {
  404: 'invalid package name',
},
```

ok: https://pub.dev/api/packages/path/publisher
404: https://pub.dev/api/packages/pat/publisher
 
http://localhost:8080/pub/publisher/path
http://localhost:8080/pub/publisher/pat 
 
ok: https://pub.dartlang.org/packages/path.json
404: https://pub.dartlang.org/packages/pat.json
 
http://localhost:8080/pub/v/path 
http://localhost:8080/pub/v/pat
 
 
services/pub/pub-publisher.service.js
services/pub/pub.service.js

a
b
c
dart4
dart5
dart6

### DIFF
--- a/services/pub/pub-publisher.service.js
+++ b/services/pub/pub-publisher.service.js
@@ -38,6 +38,9 @@ export class PubPublisher extends BaseJsonService {
     return this._requestJson({
       schema,
       url: `https://pub.dev/api/packages/${packageName}/publisher`,
+      errorMessages: {
+        404: 'invalid package name',
+      },
     })
   }
 
--- a/services/pub/pub-publisher.tester.js
+++ b/services/pub/pub-publisher.tester.js
@@ -12,7 +12,7 @@ t.create('package not verified publisher').get('/utf.json').expectBadge({
   color: 'lightgrey',
 })
 
-t.create('package not found').get('/doesnotexist.json').expectBadge({
+t.create('invalid package name').get('/doesnotexist.json').expectBadge({
   label: 'publisher',
-  message: 'not found',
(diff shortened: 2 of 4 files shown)


chris48s:
I think the original intent of that was: If we're getting
- 404 to indicate not found (but theoretically valid package name)
- 400 to indicate non valid package name

then we could consider saying..

```js
errorMessages: {
  404: 'not found',  // implicit
  400: 'invalid package name',
},
```

If we're now getting 404 for both cases, I think sticking with the existing not found messaging makes most sense.

PaulaBarszcz:
> I think the original intent of that was: If we're getting
> 
> * 404 to indicate not found (but theoretically valid package name)
> * 400 to indicate non valid package name
> 
> then we could consider saying..
> 
> ```js
> errorMessages: {
>   404: 'not found',  // implicit
>   400: 'invalid package name',
> },
> ```
> 
> If we're now getting 404 for both cases, I think sticking with the existing not found messaging makes most sense.

Can we close this PR and the related issue, if no change is needed?:)

PR #8600, opened 2022-11-08
Update the Modrinth API to v2

Insprill:
Updates the usage of the Modrinth API to v2.
The relevant documentation can be found [here](https://docs.modrinth.com/api-spec/#tag/projects/operation/getProject).


chris48s (approved):
:+1:

PR #8603, opened 2022-11-08
handle missing properties array in [VisualStudioMarketplaceVersion]

chris48s:
Refs https://sentry.io/organizations/shields/issues/3725784089

Fixes `TypeError: Cannot read properties of undefined (reading 'find')` when `properties` array is not present.


calebcartwright (approved):
afaict the only access to this array currently does so safely even in the empty array case, though for any future service class additions or modifications we may want to try to remember to double check

PR #8604, opened 2022-11-08
Add [modrinth] version

Insprill:
Adds a badge for the version of a Modrinth project.

Since there are a few Modrinth badges now, I've also moved the schemas and the fetch methods into a base class so they can be reused.

A simple documentation line has also been added to all badges explaining where to find your project ID.


services/modrinth/modrinth-version.tester.js
```diff
@@ -0,0 +1,12 @@
+import { createServiceTester } from '../tester.js'
+import { isMetric } from '../test-validators.js'
+
+export const t = await createServiceTester()
+
+t.create('Version')
+  .get('/AANobbMI.json')
+  .expectBadge({ label: 'version', message: isMetric })
```
chris48s, line 8: The test for this is failing with

```
ValidationError: message mismatch: "value" with value "mc1.19.2-0.4.4" fails to match the required pattern: /^([1-9][0-9]*[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY])$/
```

We do have a selection of test validators in https://github.com/badges/shields/blob/master/services/test-validators.js for validating version numbers, but I'm not sure `mc1.19.2-0.4.4` will match any of them. We might need a special validator for this. Are there any docs on what the format of a modrinth version should be?

Insprill: >Are there any docs on what the format of a modrinth version should be?

There aren't, it's up to the author to choose a version. I could choose another Modrinth project to run the test against but then if they change their versioning it will fail. Perhaps we could do a simpler match like `.*\d+\.\d+(\.d+)?.*`? It's not as strict, but less likely to break.

chris48s: Yeah something like that seems like a reasonable tradeoff.

services/modrinth/modrinth-base.js
```diff
@@ -0,0 +1,38 @@
...
+  followers: nonNegativeInteger,
+}).required()
+
+const versionSchema = Joi.array()
+  .items(
+    Joi.object({
+      version_number: Joi.string(),
+    })
+  )
+  .required()
```
chris48s, line 16: The schema should be

```suggestion
const versionSchema = Joi.array()
  .items(
    Joi.object({
      version_number: Joi.string().required(),
    }).required()
  )
  .required()
```

We want to make sure that there is at least one item in the array and the `version_number` peoperty is required.

services/modrinth/modrinth-version.service.js
```diff
@@ -0,0 +1,28 @@
...
+      namedParams: { projectId: 'AANobbMI' },
+      staticPreview: renderVersionBadge({ version: '0.4.4' }),
+      documentation,
+    },
+  ]
+
+  static defaultBadgeData = { label: 'version' }
+
+  async handle({ projectId }) {
+    const { 0: latest } = await this.fetchVersions({ projectId })
```
chris48s, line 24: Do we definitely know that the first item in the array will be the latest version? The docs are not explicit on this.

chris48s (commented):
Thanks for getting the build passing. I've had a proper look over this now. A couple of comments.

services/modrinth/modrinth-version.service.js
```diff
@@ -0,0 +1,28 @@
+import { renderVersionBadge } from '../version.js'
+import { BaseModrinthService, documentation } from './modrinth-base.js'
+
+export default class ModrinthVersion extends BaseModrinthService {
+  static category = 'version'
+
+  static route = {
+    base: 'modrinth/version',
```
chris48s, line 8: Normally we use `/v` as the route for version badges - see https://github.com/badges/shields/blob/master/doc/badge-urls.md

services/modrinth/modrinth-version.service.js
Insprill: I don't see a reference to this on the docs, however, I asked them in their Discord server and they confirmed this is the case for the v2 API.
image

chris48s (approved):
Sweet. Cheers

PR #8623, opened 2022-11-12
Add [Coincap] Cryptocurrency badges

alexandresanlim:
Added

- Symbol
- Rank
- Price
- Change Percent on 24hr


services/coincap/coincap-symbol.service.js
```diff
@@ -0,0 +1,35 @@
+import Joi from 'joi'
```
chris48s, line 1: Given this isn't a value that is going to change over time, is the symbol badge actually providing any value?. I'd be inclined to say if people really want this, you could use https://img.shields.io/badge/bitcoin-BTC-green without us having to call an API to get the value, especially given there is a rate limit to consider. I think we should just ditch this one.

services/coincap/coincap-changepercent24hr.service.js
```diff
@@ -0,0 +1,42 @@
+import Joi from 'joi'
+import { floorCount } from '../color-formatters.js'
+import BaseCoincapService from './coincap-base.js'
+
+const schema = Joi.object({
+  data: Joi.object({
+    changePercent24Hr: Joi.string().required(),
```
chris48s, line 7: If we're going to treat this as a number in `percentFormat()`, we should validate that it is a number in the schema.

services/coincap/coincap-priceusd.service.js
```diff
@@ -0,0 +1,43 @@
+import Joi from 'joi'
+import BaseCoincapService from './coincap-base.js'
+
+const schema = Joi.object({
+  data: Joi.object({
+    priceUsd: Joi.string().required(),
```
chris48s, line 6: Again, if we're going to manipulate this as a number in `priceFormat()` we should validate more strictly here. Any old string won't do.

services/coincap/coincap-base.js
```diff
@@ -0,0 +1,19 @@
...
+
+export default class BaseCoincapService extends BaseJsonService {
+  static category = 'other'
+
+  static defaultBadgeData = { label: 'bitcoin' }
+
+  async fetch({ assetId, schema }) {
+    return this._requestJson({
+      schema,
+      url: `https://api.coincap.io/v2/assets/${assetId}`,
```
chris48s, line 11: A note here.
There is a rate limit on this API. From https://docs.coincap.io/

> Free Tier (No API Key)
> 200 requests per minute

> Free Tier (API Key)
> 500 requests per minute

I'm not _super_ concerned by this. 200 requests per minute is pretty generous - it would allow this to become quite a popular service (not convinced this is super likely) and there is scope to either request an API key to get a higher limit or tweak the cache lengths to reduce the number of requests we need to make but we should be aware of this.

services/coincap/coincap-priceusd.service.js
```diff
@@ -0,0 +1,43 @@
...
+    return `$${parseFloat(price)
+      .toFixed(2)
+      .replace(/\d(?=(\d{3})+\.)/g, '$&,')}`
+  }
+
+  static render({ asset }) {
+    return {
+      label: asset.name,
+      message: this.priceFormat(asset.priceUsd),
+      color: 'green',
```
chris48s, line 35: `color` should be blue for informational badges (i.e: we're not using colour to convey any meaning)

services/coincap/coincap-rank.service.js
```diff
@@ -0,0 +1,35 @@
...
+      staticPreview: this.render({ asset: { rank: '1' } }),
+      keywords: ['bitcoin', 'crypto', 'cryptocurrency'],
+    },
+  ]
+
+  static render({ asset }) {
+    return {
+      label: asset.name,
+      message: asset.rank,
+      color: 'green',
```
chris48s, line 27: `color` should be blue for informational badges (i.e: we're not using colour to convey any meaning)

services/coincap/coincap-changepercent24hr.service.js
```diff
@@ -0,0 +1,42 @@
...
+
+  static percentFormat(changePercent24Hr) {
+    return `${parseInt(changePercent24Hr).toFixed(2)}%`
+  }
+
+  static render({ asset }) {
+    return {
+      label: asset.name,
+      message: this.percentFormat(asset.changePercent24Hr),
+      color: floorCount(asset.changePercent24Hr),
```
chris48s, line 34: What are we trying to achieve by using `floorCount()` with no params for formatting here?
Tbh, I'd be inclined to make this an informational blue badge too rather than try to use colour to convey meaning here. It seems like it would be hard to definitively even say whether stability or volatility is 'good' or 'bad' in all cases.

services/coincap/coincap-rank.tester.js
```diff
@@ -0,0 +1,13 @@
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('rank').get('/bitcoin.json').expectBadge({
+  label: 'Bitcoin',
+  message: '1',
```
chris48s, line 6: For the tests that call a live service, we sholdn't hard-code a response in the test. Lets validate that this as `Joi.number().integer().min(1)`

chris48s (commented):
This is in pretty good shape. It is clear you have read the contributing docs and thanks for including tests.

services/coincap/coincap-base.js
alexandresanlim: I didn't pay attention to that, I referenced the documentation in the code.

services/coincap/coincap-changepercent24hr.service.js
alexandresanlim: It is a convention in finance, market information with negative percentage variation be red and positive green. Including coincap.

services/coincap/coincap-rank.tester.js
alexandresanlim: The .json return a string even being number, i had to validated it with regex.

services/coincap/coincap-changepercent24hr.service.js
chris48s: Right yes I see where you are going with this now :smile:

services/coincap/coincap-rank.tester.js
chris48s: You can still use `Joi.number()` to picture-check a string.

Try this as a minimal example:

```js
import Joi from 'joi'

const schema = Joi.number().integer().min(1).required()

Joi.attempt('1', schema); // passes
Joi.attempt('0', schema); // fails
```

Same goes for the changepercent24hr and priceusd schemas - they can use `Joi.number()` instead of a regex which should make things a bit neater.

services/coincap/coincap-changepercent24hr.tester.js
```diff
@@ -0,0 +1,43 @@
...
+  )
+  .expectBadge({
+    label: 'Bitcoin',
+    message: '-1.00%',
+    color: 'red',
+  })
+
+t.create('change percent 24hr').get('/bitcoin.json').expectBadge({
+  label: 'Bitcoin',
+  message: isPercentage,
```
chris48s, line 37: Looking at the build for this branch it is failing with
`ValidationError: message mismatch: "value" does not match any of the allowed types`
because at the time the tests ran the change was negative and the validator isn't expecting a leading `-`.
We'll need to adjust the validator to allow negative values.

services/coincap/coincap-base.js
```diff
@@ -0,0 +1,22 @@
+import { BaseJsonService } from '../index.js'
+
+export default class BaseCoincapService extends BaseJsonService {
+  static category = 'other'
+
+  static defaultBadgeData = { label: 'Bitcoin' }
```
chris48s, line 6: Sorry I should have spotted this the first time round. There's a couple of things here: Labels should always be lowercase, so we should lowercase `asset.name` when we set it. Also, the default label should really be something generic. i.e: if you pass 'ethereum' as your param, get an error and the error is `bitcoin | something went wrong`, that would be odd. I think `coincap` should be the default label as that is the service name.

services/coincap/coincap-priceusd.service.js
```diff
@@ -0,0 +1,45 @@
...
+  data: Joi.object({
+    priceUsd: Joi.string()
+      .pattern(/[0-9]*\.[0-9]+/i)
+      .required(),
+    name: Joi.string().required(),
+  }).required(),
+}).required()
+
+export default class CoincapPriceUsd extends BaseCoincapService {
+  static route = { base: 'coincap/priceUsd', pattern: ':assetId' }
```
chris48s, line 14: Sorry. This is another one I didn't spot the first time. In general the hard-coded parts of the route are snake-case all lowercase. So `price-usd` here and `change-percent-24-hr` for the other one.
The params can stay as they are because they get turned into variables.

services/coincap/coincap-changepercent24hr.tester.js
alexandresanlim: I have added negative possibilities on `test-validators.js` ok?

services/coincap/coincap-rank.tester.js
alexandresanlim: Nice! but to changepercent24hr and priceusd `Joi.number().required()` doesn't worked.

chris48s (approved):
I spotted one more issue on the last check. The examples were rendering with label undefined. Merging with a fix for that

PR #8643, opened 2022-11-18
don't call external services in core tests

chris48s:
There are some tests in `core/server/server.spec.js` which make unmocked calls to NPM badges.

This is mostly not a big issue because it is a pretty reliable service and if NPM is working well enough for us to install the dependencies then the badges are probably going to work too. That said, I've been seeing some PRs failing because of a timeout on these tests.

I did consider mocking the calls, but having thought about it, I don't think there is actually any solid reason why these tests have to call a dynamic badge. As such, I've just replaced them all with calls to static badges instead. I don't think we lose anything by doing this and it reduces the scope for the core tests to be flaky.


core/server/server.spec.js
```diff
@@ -197,22 +202,19 @@ describe('The server', function () {
...
         .to.satisfy(isSvg)
         .and.to.include('410')
         .and.to.include('jpg no longer available')
     })
-
-    it('should return cors header for the request', async function () {
-      const { statusCode, headers } = await got(`${baseUrl}npm/v/express.svg`)
-      expect(statusCode).to.equal(200)
-      expect(headers['access-control-allow-origin']).to.equal('*')
-    })
```
calebcartwright, line 215: I am a tad worried that in dropping this test we're potentially losing validation of header behavior. Would you mind elaborating on your thinking behind this?

calebcartwright (approved):
One question inline below but will go ahead and :+1: in case it's easily answered so that merging won't be blocked

core/server/server.spec.js
chris48s: Sorry. I hoped this was clear from the commit message on 202805d but I should have left a comment on the diff.

Before this PR we had 2 copies of the same test in this file:
https://github.com/badges/shields/blob/3efe143edc17a3bd16d07b532ee0021c3485ff35/core/server/server.spec.js#L67-L71
https://github.com/badges/shields/blob/3efe143edc17a3bd16d07b532ee0021c3485ff35/core/server/server.spec.js#L211-L215
I noticed it while I was doing this so I removed the duplicate instead of updating both copies

Issue #8666, opened 2022-11-30
Option or default to using `max_stable_version` when getting crate info from https://crates.io/api/v1/crates/:crate_name


zeeshanlakhani:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

A Rust project I've contributed to, called [ucan](https://crates.io/crates/ucan) (and [ucan-key-support](https://crates.io/crates/ucan-key-support)), recently transitioned from (higher-versioned) alpha releases to a stable 0.1.0 release. However, the badge is not getting updated (https://img.shields.io/crates/v/ucan.svg), as it uses the max_version, per this line of code: https://github.com/badges/shields/blob/master/services/crates/crates-version.service.js#L22. In viewing the crate information via https://crates.io/api/v1/crates/ucan, is there a way to use the max_stable_version if it exists, defaulting to max_version otherwise? Here's what the crate data returns: 

```json
{
   "crate":{
      "badges":[],
      "categories":[
         "web-programming",
         "cryptography",
         "encoding"
      ],
      "created_at":"2022-03-11T05:26:14.291552+00:00",
      "description":"Implement UCAN-based authorization with conciseness and ease!",
      "documentation":"https://docs.rs/ucan",
      "downloads":1875,
      "exact_match":false,
      "homepage":"https://github.com/cdata/rs-ucan",
      "id":"ucan",
      "keywords":[
         "jwt",
         "ucan",
         "pki",
         "authz"
      ],
      "links":{
         "owner_team":"/api/v1/crates/ucan/owner_team",
         "owner_user":"/api/v1/crates/ucan/owner_user",
         "owners":"/api/v1/crates/ucan/owners",
         "reverse_dependencies":"/api/v1/crates/ucan/reverse_dependencies",
         "version_downloads":"/api/v1/crates/ucan/downloads",
         "versions":null
      },
      "max_stable_version":"0.1.0",
      "max_version":"0.7.0-alpha.1",
      "name":"ucan",
      "newest_version":"0.1.0",
      "recent_downloads":1032,
      "repository":"https://github.com/cdata/rs-ucan/",
      "updated_at":"2022-11-29T18:39:54.791532+00:00",
      "versions":[
         634470,
         552070,
         525823,
         525094,
         520718,
         519720,
         517394,
         517384,
         514425,
         514412,
         673114
      ]
   }
}
```

### 🔗 Link to the badge

https://img.shields.io/crates/v/ucan.svg
[...]

PR #8667, opened 2022-11-30
fix: for crates.io versions, use max_stable_version if it exists [CRATES]

zeeshanlakhani:
Proposal to fix/handle #8666.


chris48s:
Thanks for submitting this and the extra info in https://github.com/badges/shields/issues/8666 . There are couple of problems here:

1. CircleCI isn't running the service tests from your fork. This is a [known issue](https://github.com/badges/shields/issues/8109) affecting some contributors. We're working on migrating away from it (hence the current random mashup of GH actions and Circle CI)
2. Looking at this code has made me realise there is another issue here.

I've moved this change to PR https://github.com/badges/shields/pull/8687 but I've left you as the commit (co-)author

PR #8668, opened 2022-11-30
docs/issue template improvements for badge requests

chris48s:
We've been seeing a fair few new badge requests like "add a badge for java" recently. I'd like to try and make it clearer up-front in the issue template that users can already self-serve for this type of request so I've re-vamped the "Badge Request" issue template a bit.


.github/ISSUE_TEMPLATE/3_Badge_request.yml
```diff
@@ -0,0 +1,55 @@
+name: '💡 Badge Request'
+description: Ideas for new badges
+labels: ['service-badge']
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## Ideas for new badges
+
+        This issue template is for suggesting new badges which **fetch and display data from an upstream service**. If your suggestion is for a static badge (which shows the same information every time it is requested), it is [already possible to make these](https://github.com/badges/shields/blob/master/doc/static-badges.md). We don't add specific routes for badges which only show static information.
```
chris48s, line 10: `static-badges.md` is added in this PR so the link will be dead till we merge it

doc/static-badges.md
```diff
@@ -0,0 +1,12 @@
+# Static Badges
```
chris48s, line 1: I'd love there to be a really good place on shields.io we could link to here. I don't think we can completely blame users for not being able to self-serve these requests when we don't really explain the feature super well. Re-vamping the site is an whole other issue (which I have also been noodling with) but adding this seems like a 'quick fix' to give us something to link to from the issue form.

chris48s:
@PaulaBarszcz - would you be up for giving me a review on this one?

.github/ISSUE_TEMPLATE/3_Badge_request.yml
```diff
@@ -0,0 +1,55 @@
...
+  - type: textarea
+    id: data
+    attributes:
+      label: '🔗 Data'
+      description: |
+        Where can we get the data from?
+
+        Please consider and cover details like:
+        - Is there a public API?
+        - Does the API require authentication or an API key?
```
PaulaBarszcz, line 35: I would add info that if a private key is required, we will not be able to add this badge. This way, the issue's author can abstain from submitting it, and we save time :)

.github/ISSUE_TEMPLATE/3_Badge_request.yml
PaulaBarszcz: To be coherent with the other long lines, let's break down this line into shorter lines that fit within the screen.

doc/static-badges.md
```diff
@@ -0,0 +1,12 @@
...
+
+It is possible to use shields.io to make a wide variety of badges displaying static text and/or logos. For example:
+
+- ![any text you like](https://img.shields.io/badge/any%20text-you%20like-blue) - https://img.shields.io/badge/any%20text-you%20like-blue
+- ![just the messgae](https://img.shields.io/badge/just%20the%20message-8A2BE2) - https://img.shields.io/badge/just%20the%20message-8A2BE2
```
PaulaBarszcz, line 6: Typo in `message`

doc/static-badges.md
```diff
@@ -0,0 +1,12 @@
...
+- ![for the badge style](https://img.shields.io/badge/for%20the%20badge%20style-20B2AA?style=for-the-badge) - https://img.shields.io/badge/for%20the%20badge%20style-20B2AA?style=for-the-badge
```
PaulaBarszcz, line 7: Maybe `'for the badge' style`?

doc/static-badges.md
```diff
@@ -0,0 +1,12 @@
...
+- ![with a logo](https://img.shields.io/badge/with%20a%20logo-grey?style=for-the-badge&logo=javascript) - https://img.shields.io/badge/with%20a%20logo-grey?style=for-the-badge&logo=javascript
+
+Full documentation of styles and parameters: https://shields.io/#styles
+
+More documentation on logos: https://github.com/badges/shields/blob/master/doc/logos.md
```
PaulaBarszcz, line 12: To https://github.com/badges/shields/blob/master/doc/logos.md we can add info about the change in the `logoColor` param introduced in this PR (https://github.com/badges/shields/pull/8263) and how to use it.

PaulaBarszcz:
> @PaulaBarszcz - would you be up for giving me a review on this one?

Sure, gladly :)

doc/static-badges.md
```diff
@@ -0,0 +1,12 @@
+# Static Badges
+
+It is possible to use shields.io to make a wide variety of badges displaying static text and/or logos. For example:
+
+- ![any text you like](https://img.shields.io/badge/any%20text-you%20like-blue) - https://img.shields.io/badge/any%20text-you%20like-blue
+- ![just the messgae](https://img.shields.io/badge/just%20the%20message-8A2BE2) - https://img.shields.io/badge/just%20the%20message-8A2BE2
+- ![for the badge style](https://img.shields.io/badge/for%20the%20badge%20style-20B2AA?style=for-the-badge) - https://img.shields.io/badge/for%20the%20badge%20style-20B2AA?style=for-the-badge
+- ![with a logo](https://img.shields.io/badge/with%20a%20logo-grey?style=for-the-badge&logo=javascript) - https://img.shields.io/badge/with%20a%20logo-grey?style=for-the-badge&logo=javascript
+
```
PaulaBarszcz, line 9: Very good idea to give these badge examples here.
It might be helpful to add also some more badge examples (for now in this file, but also on the re-vamped version of the Shields website):

- how the logo can be added (in different colours), 
- how to add specific URLs from which data can be fetched, and how can the dynamic badges be used (examples related to `query=<[$.DATA.SUBDATA](https://jsonpath.com/)>`).


We can also link this here: https://shields.io/endpoint
`Create badges from your own JSON endpoint.` 
This option is powerful, and I would highlight it, for example:


Screenshot_4
https://shields.redsparr0w.com/2473/monday 
```js
// 20221201214359
// https://shields.redsparr0w.com/2473/monday

{
  "schemaVersion": 1,
  "label": "is it monday",
  "message": "no",
  "color": "orange"
}
```


Maybe this option would already meet the need of the issue's author :)

Sidenote: Our preview badge on https://shields.io/endpoint doesn't work.
Screenshot_3

.github/ISSUE_TEMPLATE/3_Badge_request.yml
chris48s: Just to explain 9bd932a

The reason I originally wrote this on a single line is because what we are actually writing at this point is Github-Flavoured markdown embedded inside a multi-line yaml string (which then generates an issue form :smile: ). If you write

```yml
value: |
    A line
    Another line
```

that's going to render as 2 paragraphs, and I wanted this whole thing to render as one paragraph. The other makrdown strings in this file are bullet lists, so they split across multiple lines naturally.

I think the solution to this is to switch from a _literal_ multi-line yaml string `|` to a _folded_ multi-line yaml string `>` (see https://yaml-multiline.info/ ) which should allow us to write this paragraph on multiple lines in the yaml but then have it render as a single paragraph when someone raises an issue. I'm going to have to double check this once we merge it though and might need to tweak it. I'm more concerned about the user seeing the right thing in the issue form than the how the raw yaml looks.

doc/static-badges.md
chris48s: 7b9a8b3

doc/static-badges.md
chris48s: a4d3fe6

doc/static-badges.md
chris48s: d5e2bd3

doc/static-badges.md
chris48s: Logos and `logoColor` are covered in https://github.com/badges/shields/blob/master/doc/logos.md which is linked in this doc, so I think that is fine.

The other stuff you've mentioned here is all about dynamic badges, rather than static. I agree they're not as well documented as they could be, but I think the main thing I'm trying to head off here is people raising issues about static badges. Basically since everyone started doing this

Screenshot at 2022-12-04 12-03-12

sort of thing on their profile READMEs, we've had an uptick in these kind of requests so I want to be a bit clearer about it.

On dynamic badges, I'm inclined not to shave this yak any further in this specific PR.

> Sidenote: Our preview badge on https://shields.io/endpoint doesn't work.

The preview for the endpoint badge works, but you need to give it an endpoint to preview.. I think there's a bit of an annoying historical issue here where we can't serve the 'some parameters missing' error when `?url=` is not specified (like with all the other badge routes) because the base URL `/endpoint` doubles up as the docs. We made some bad choices which bit us later. I think the relevant history is:
https://github.com/badges/shields/issues/5014
https://github.com/badges/shields/pull/5137

.github/ISSUE_TEMPLATE/3_Badge_request.yml
```diff
@@ -0,0 +1,62 @@
...
+          If so, please review our documentation on [Badges Requiring Authentication](https://github.com/badges/shields/blob/master/doc/authentication.md)
```
chris48s, line 42: This line is a bit long, but I think in this case, the string as a whole is going to be more readable using a literal multi-line string (`|`) and accepting this line is a bit long. Again, this link will be broken till we merge this.

.github/ISSUE_TEMPLATE/3_Badge_request.yml
PaulaBarszcz: Cool, thanks, I didn't know about the distinction between a literal multi-line YAML string (`|`) and a folded multi-line YAML string (`>`). 
And you're right that it is more important how the result will look for the user vs what the raw YAML will look like :)

.github/ISSUE_TEMPLATE/3_Badge_request.yml
PaulaBarszcz: P e r f e c t - this is exactly what I was hoping to see here! Thanks :)

doc/static-badges.md
PaulaBarszcz: Ok, thanks for the clarification.

You're right, my comments were about dynamic badges, and this PR's goal is to tackle static ones.

Abandoning the subject 🏃 ;)

PaulaBarszcz (approved):
Thanks for the changes, @chris48s - in my opinion this PR is ready to be merged 👍

Issue #8665, opened 2022-11-29
Version GitHub API requests


chris48s:
:clipboard: **Description**

Yesterday GitHub announced that they are going to adopt Calendar Versioning for v3 API requests. API version will be specified via a `X-GitHub-Api-Version` header. GraphQL requests are unaffected.

:books:  Blog post: https://github.blog/2022-11-28-to-infinity-and-beyond-enabling-the-future-of-githubs-rest-api-with-api-versioning/

I think what we need to do is start sending `X-GitHub-Api-Version: 2022-11-28` with all our v3 requests and then work out an approach for keeping up-to-date with changes. This will involve:

- Bumping the date as new releases are available (or at least as old ones are retired)
- Accounting for any breaking changes when we bump it

I'll have a more detailed read over the blog post and work out a plan. I think the vast majority of our request to GitHub's API go through one function so sending the header with all requests shouldn't be a big deal. A process for keeping on top of the changes is probably the hard bit.

PR #8669, opened 2022-11-30
Send `X-GitHub-Api-Version` when calling [GitHub] v3 API

chris48s:
Closes https://github.com/badges/shields/issues/8665

This PR does three things:

1. Sends the `X-GitHub-Api-Version` header when calling the GitHub v3 API
2. Adds a scheduled GitHub workflow to:
    - Call https://api.github.com/versions
    - Find the latest version by date
    - Read in `config/default.yml`
    - If `$date != public.services.github.restApiVersion`:
        - sub in the new date + write `config/default.yml`
        - push to a branch
        - submit a PR `update [github] api version`
3. Tangent: Fixes a long-standing bit of oddness: `github-constellation.js` has been ignoring the `config.service.baseUri` setting and reading straight from the env var. I guess most people either don't configure this or configure with env vars anyway, so it has gone unnoticed.


calebcartwright:
> Fixes a long-standing bit of oddness: github-constellation.js has been ignoring the config.service.baseUri setting and reading straight from the env var.

Though a fix, imagine there's a non-zero chance of this impacting a self-hoster? May warrant a special call out in the next server relnotes

chris48s:
If you are currently setting this via `GITHUB_URL` then it will carry on working. `services.github.baseUri` is set from https://github.com/badges/shields/blob/41c53c73c26379bd826a142ce0aabb794f6bb5ea/config/custom-environment-variables.yml#L39

I think the only case where this might break would be if someone is hosting (using a source install, rather than docker), they edited `services.github.baseUri` in the yaml config (which would have no effect) and then they update to this release and suddenly `services.github.baseUri` starts actually doing something.

Is that the case you were thinking about?

calebcartwright:
> Is that the case you were thinking about?

Precisely. Someone who has a value in there that they forgot about which currently is a no-op but which would become operational once they pulled in this next version.

Obviously not saying we shouldn't do it, but it's the type of thing I'd put at/near the top of the release notes to call it out specially

chris48s:
Yep OK. I'll note this when I assemble the changelog for the January release. We need to call out the change to the GH workflows route there too.

calebcartwright (approved):

PR #8673, opened 2022-12-02
Add [modrinth] game versions

Insprill:
Adds a badge for Modrinth game versions.

The base routes for similar badges seemed to be all over the place so it's currently `modrinth/game-versions`. If it should be changed to something else please let me know.


PaulaBarszcz (approved):
👍

triphora (dismissed):

services/modrinth/modrinth-game-versions.service.js
```diff
@@ -0,0 +1,41 @@
...
+      return {
+        message: versions.join(' | '),
+        color: 'blue',
+      }
+    } else {
+      return {
+        message: 'missing',
+        color: 'red',
+      }
+    }
```
chris48s, line 33: Instead of manually returning an 'error' response here, lets require that the `game_versions` array has at least 1 element in the Joi schema using

```js
Joi.array().items(Joi.string()).min(1).required() // note min(1)
```

If we do this, an error will be thrown at the point we validate the response and we can just drop the length check here.

chris48s (commented):
Sorry it has taken so long to get to this.

I agree the existing routes for platform support badges are a bit of a mess. Lets go with your suggestion. If we decide to tidy these up in future, we can always use a redirect.

chris48s (approved):
Cheers. I'll get this merged/deployed soon :+1:

PR #8686, opened 2022-12-04
Add [Infracost] cloud cost badge

hugorut:
Adds support for an Infracost cloud cost shield that can be broken down by repo or project.

### DIFF
--- a/services/infracost/infracost.service.js
+++ b/services/infracost/infracost.service.js
@@ -1,0 +1,104 @@
+import Joi from 'joi'
+import BaseJsonService from '../../core/base-service/base-json.js'
+
+const queryParamSchema = Joi.object({
+  badgeToken: Joi.string().required(),
+  projectName: Joi.string(),
+}).required()
+
+const schema = Joi.object({
+  cost: Joi.string().required(),
+})
+
+const documentation = `
+  <p>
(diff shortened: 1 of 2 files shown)


calebcartwright:
Thank you for the PR. Though we do not strictly require PRs have an associated issue, in cases like this where a new badge is being proposed absent an issue we'd still like to get the details and answers to the questions we pose as part of fielding new badge requests.

Sharing them inline below to streamline things, please follow up with responses whenever you get a chance!

### Description

* Which service is this badge for e.g: GitHub, Travis CI
* What sort of information should this badge show?
* Provide an example in plain text e.g: "version | v1.01" or as a static badge (static badge generator can be found at https://shields.io/#your-badge )


### Where can we get the data from?

Please consider and cover details like:

* Is there a public API?
* Does the API require authentication or an API key? (If so, please review our documentation on [Badges Requiring Authentication](https://github.com/badges/shields/blob/master/doc/authentication.md))
* Link to the API documentation.

PyvesB:
@hugorut have been open for a while, are you planning on circling back to it in the foreseeable future or shall we close it for the time being?

hugorut:
@PyvesB I'll close, we're using a different strategy to display the shields. Thanks

PR #8687, opened 2022-12-04
use `max_stable_version` in rust [crates] badge

chris48s:
Refs #8666
Refs #8667

This PR switches us to using `max_stable_version` if present, or `max_version` otherwise, as suggested in #8666

While I was looking at this, I realised returning `json.version.num` makes no sense here and we should never take this branch. It is a hangover from a really really old implementation https://github.com/badges/shields/blob/302c8606ff4c975b6310fe2cdfe58f92cede290b/server.js#L459-L528 and we can get rid of it. We only ever call the `/api/v1/crates/:crate` endpoint when we render this badge - never the `/api/v1/crates/:crate/:version` endpoint, so we don't need to handle that response.


zeeshanlakhani:
@chris48s +1.

zeeshanlakhani (approved):

PaulaBarszcz (approved):
👍

zeeshanlakhani:
@chris48s just wanted to check in on this?

calebcartwright:
> @chris48s just wanted to check in on this?

This is pending on me, and through a mixture of forgetfulness and not having adequate bandwidth, I haven't gotten round to it yet.

Will try to soon though :crossed_fingers:

calebcartwright (approved):
Confirmed this is the route we want to go, apologies for this being held up waiting for action from my end!

PR #8690, opened 2022-12-07
fall back to classifiers if [pypi] license text is really long

### LINKED ISSUE #8689: PyPI: Very long license badge

shields.io

### 🐞 Description

Per PEP 621 a Python package can be distributed with a [`license`](https://peps.python.org/pep-0621/#license) metadata field set to either `file: link_to_license_file` or `text: some_license_text`.

Packages that set `file` instead of `text` can end up with very long content. This seems to be valid, as demonstrated by the long version content you can see on [Bokeh's PyPI page](https://pypi.org/project/bokeh/), this project [setting `file` indeed](https://github.com/bokeh/bokeh/blob/0326ae1c0a52213bbe62a4624f3621af3e03affe/pyproject.toml#L26).

### 🔗 Link to the badge

This leads to a very long badge :)
PyPI - License
image

### 💡 Possible Solution

Shouldn't the badges be built from the *License* classifier instead of the core `license` metadata?

chris48s:
Closes #8689


services/pypi/pypi-helpers.js
```diff
@@ -50,7 +50,7 @@ function getLicenses(packageData) {
   const {
     info: { license },
   } = packageData
-  if (license) {
+  if (license && license.length < 40) {
```
chris48s, line 53: As noted in https://github.com/badges/shields/issues/8689#issuecomment-1341526309 we're going to have to make some kind of assumption here. The longest SPDX identifier in https://raw.githubusercontent.com/jslicense/spdx-license-ids/main/index.json is 36 characters so this seems like a reasonable-ish threshold.

My hunch is that the license meta-data field is more commonly used than the classifiers, so I think switching to prefer the classifiers in all cases is probably going to break more badges out in the wild than doing this.

There isn't really a perfect solution here.

paulmelnikow (approved):

services/pypi/pypi-helpers.js
calebcartwright: Agreed with the approach. I do think it might be worth adding an inline comment explaining this though so that `40` doesn't appear like a magic number to readers of the code.

I recognize that one could use git blame info to backtrack to this PR and explanation, but I'm a fan of having things be as self-explanatory as possible within the code

PR #8717, opened 2022-12-15
handle workflow runs that do not have a conclusion in [githubactionsworkflowstatus]

### LINKED ISSUE #8716: [GitHub Actions] "invalid response data" for pending jobs

shields.io

### 🐞 Description

After switching to the new route for GitHub Actions workflows, as described in https://github.com/badges/shields/issues/8671 , the badge shows "invalid response data" when the workflow is running, rather than "running" or "pending" or a more informative status.

### 🔗 Link to the badge

https://img.shields.io/github/actions/workflow/status/kevinoid/example-slow-workflow/ci.yml?branch=main

### 💡 Possible Solution

_No response_

chris48s:
Closes #8716

It was always too much to ask for this to go _completely_ to plan :laughing: 

When a build is not completed yet, it has a null value for `.conclusion` e.g:

```js
status: 'in_progress',
conclusion: null,
```

This PR allows `.conclusion` to be `null` and requires a non-null value for `status` which we can fall back to.

### DIFF
--- a/services/github/github-actions-workflow-status.service.js
+++ b/services/github/github-actions-workflow-status.service.js
@@ -8,8 +8,23 @@ const schema = Joi.object({
   workflow_runs: Joi.array()
     .items(
       Joi.object({
+        status: Joi.equal(
+          'completed',
+          'action_required',
+          'cancelled',
+          'failure',
+          'neutral',
+          'skipped',
+          'stale',
+          'success',
+          'timed_out',
+          'in_progress',
+          'queued',
+          'requested',
+          'waiting'
+        ).required(),
         conclusion: Joi.alternatives()
-          .try(isBuildStatus, Joi.equal('no status'))
+          .try(isBuildStatus, Joi.equal('no status'), null)
           .required(),
       })
     )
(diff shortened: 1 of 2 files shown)


services/github/github-actions-workflow-status.service.js
```diff
@@ -8,8 +8,23 @@ const schema = Joi.object({
...
+          'neutral',
+          'skipped',
+          'stale',
+          'success',
+          'timed_out',
+          'in_progress',
+          'queued',
+          'requested',
+          'waiting'
+        ).required(),
```
chris48s, line 25: Values for status are taken from

https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run

Screenshot at 2022-12-15 20-09-48

I was in two minds about whether to add all the status values to `build-status.js` and attempt to classify them. The solution I've settled on is to leave that as-is and if we fall back to a status value and we don't have a formatting rule for it, just use default lightgrey for formatting.

I was also in two minds about whether to use a closed list here or allow any string (but make it required). Open to feedback on it..

calebcartwright (approved):
Schema proposal works for me :+1: In general I'd have been more inclined to grab a transform function and unit test, but no objections to this since you've already battled and defeated the nock intercepts :smile:

calebcartwright:
Going to go ahead and let the bots merge this and then will run a deploy later tonight

Issue #8720, opened 2022-12-16
invalid query parameter: branch


Yurunsoft:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

invalid query parameter: branch

### 🔗 Link to the badge

https://img.shields.io/github/actions/workflow/status/imiphp/imi/ci.yml?branch=2.1

chris48s:
We've just merged https://github.com/badges/shields/pull/8734 and deployed to production, which fixes this bug.

PR #8734, opened 2022-12-16
allow all numeric branch names in [githubactionsworkflowstatus]

chris48s:
Closes https://github.com/badges/shields/issues/8720

Lets get a nice easy one out of the way first.

Tbh, I suspect this might be an issue that exists in other places, but I'm not going to shave that yak right this second. Squeaky wheel gets the grease today.

### DIFF
--- a/services/github/github-actions-workflow-status.service.js
+++ b/services/github/github-actions-workflow-status.service.js
@@ -11,7 +11,7 @@ const schema = Joi.object({
 
 const queryParamSchema = Joi.object({
   event: Joi.string(),
-  branch: Joi.string(),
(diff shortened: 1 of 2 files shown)


calebcartwright (approved):

chris48s:
https://github.com/badges/shields/pull/8738 makes this one irrelevant

chris48s:
no it doesn't. I need to come back to this one.. :facepalm:

chris48s:
rebased this to resolve the conflict and make branch optional, merging

dunno why I thought making the param optional resolved the issue with number-only branch names, aside from: I need to take a few days away from this

Issue #8736, opened 2022-12-16
Github Actions Workflow badge not working for "branchless" push


chris48s:
### Discussed in https://github.com/badges/shields/discussions/8719

<div type='discussions-op-text'>

<sup>Originally posted by **StormFX** December 16, 2022</sup>
I didn't want to create an issue for what simply be user error on my end, but I can't seem to get my build status badges to work since the change to the badge routes.

I have, for example, a repo at `SFX-WoW/Masque`. My workflows are in `.github/workflows`, thus, the URL for a status badge _should_ be `https://img.shields.io/github/actions/workflow/status/SFX-WoW/Masque/build-release.yml?branch=main` according to #8671, But it doesn't work. It says, "repo or workflow not found". If I remove the `workflows/` it changes to, "branch or event not found". Any ideas?

Edit: I also tried the URL generator and it doesn't work, either.</div>

chris48s:
examples:

https://github.com/SFX-WoW/Masque/blob/8853f25d46303e6c260afc395e09a57356c13871/.github/workflows/build-release.yml#L5-L12

https://github.com/pftf/RPi4/blob/cbdf29976acc3cff38093e61adcab6ab618ecc1b/.github/workflows/linux_edk2.yml#L6-L9

StormFX:
Mind you, I don't know that much about Shields.io's API, but it seems like the changes brought it more in parity with GitHub's API, at least as far as functionality. As I mentioned in #8671, GitHub's badges work with no branch filter in the workflow if no branch parameter is specified in the URL. Iit _seems_ like the parameters (or lack thereof) target only workflows with explicit filters (or no parameter-specific filters). Therefore, it seems the logical solution would be for Shields API to allow no parameters and in that case, target workflows with no relevant filters.

chris48s:
Having done a bit of work on this, I think there are 2 possible options:

[...]

Issue #8733, opened 2022-12-16
Github workflow badge showing `failure` when is not the case


mondeja:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Hi,

After updating Github workflow from the old URL:

```
https://img.shields.io/github/workflow/status/simple-icons/simple-icons/Verify/develop?logo=github&label=tests
```

to the new API

```
https://img.shields.io/github/actions/workflow/status/simple-icons/simple-icons/verify.yml?branch=develop&logo=github&label=tests
```

The badge is showing that the tests are failing, which is not the case as [the latest workflow for `develop` branch is passing the CI](https://github.com/simple-icons/simple-icons/actions/runs/3672019831).

### 🔗 Link to the badge

https://img.shields.io/github/actions/workflow/status/simple-icons/simple-icons/verify.yml?branch=develop&logo=github&label=tests

chris48s:
Thanks for reporting. My suggestion for now is to use https://img.shields.io/github/actions/workflow/status/simple-icons/simple-icons/verify.yml?branch=develop&logo=github&label=tests&event=push but I need to dig into this one more.

chris48s:
OK. The issue here is that `branch=develop` is also bringing back builds from pull requests where the branch is called `develop`, so we're reporting the first build status shown here, rather than the third one (which is what you'd expect).

![Screenshot at 2022-12-16 20-47-53](https://user-images.githubusercontent.com/6025893/208186772-fd0c8857-7487-4c50-81ab-365a392d4f58.png)

I'll have a look at a fix for this one tomorrow

chris48s:
Of the various issues that have come out of this, this is the one that is worrying me the most as we are serving the wrong data in some situations and I can't see a completely obvious fix.

As far as I can tell, there is no way to filter to only branches called `develop` that are in the target repo. I was hoping something like
- https://api.github.com/repos/simple-icons/simple-icons/actions/workflows/verify.yml/runs?branch=simple-icons%2Fsimple-icons%3Adevelop&page=1&exclude_pull_requests=true or
- https://api.github.com/repos/simple-icons/simple-icons/actions/workflows/verify.yml/runs?branch=refs%2Fheads%3Adevelop&page=1&exclude_pull_requests=true
might do the job, but no dice. This is unexpected, but as far as I can see there is not much we can do about it. I think I am going to raise an issue with GitHub support, but I suspect the answer is going to be "you need to pass `event=push`"

[...]

PR #8738, opened 2022-12-18
switch [GithubActionsWorkflowStatus] to SVG scraper

chris48s:
This PR implements the suggestion in #8733 and #8736 : Go back to the badge just being a 'scraper' for GitHub's native badge. I've not heard anything back from GitHub support yet. I'm putting this here for now so we can easily just merge it assuming the API is a dead end.


services/github/github-actions-workflow-status.service.js
```diff
@@ -1,46 +1,22 @@
...
-    .max(1),
+  message: Joi.alternatives()
+    .try(isBuildStatus, Joi.equal('no status'))
+    .required(),
 }).required()
 
 const queryParamSchema = Joi.object({
   event: Joi.string(),
-  branch: Joi.string().required(),
+  branch: Joi.string(),
```
chris48s, line 14: Two things here:
1. As long as it isn't `.required()`, this works for an entirely numeric branch name. Refs #8734 and #8720
2. This puts us back to a situation where `branch` is optional. If we think we might want to do this as a temporary solution, we could:
    - keep branch required
    - if the user passes `*`, we request `/badge.svg` without a `?branch=` param

    and then that way, our options are slightly more open. If we make `branch` optional again now, we're probably committed to this implementation forever.

StormFX: > * if the user passes `*`, we request `/badge.svg` without a `?branch=` param

In the event that the above solution _does_ end up being temporary, would it not be more prudent to request `/badge.svg` if the branch parameter is _empty_ rather than requiring a specific branch value? Unless I'm missing something.

chris48s:
@calebcartwright  I went back to change this following our conversation.

When I was working on updating the code/tests, I thought through it properly and re-read https://github.com/badges/shields/issues/8146#issuecomment-1262572065 and I realised there is a subtle difference here in that calling
`https://github.com/:user/:repo/actions/workflows/:workflow/badge.svg`
returns latest build on default branch if there is one, or falls back to latest build on any branch if there isn't one.

Our proposed implementation for `branch=*` (if we went back to an API-based implementation) would be latest build on any branch.

As such, if we make `branch=*` equivalent to `https://github.com/:user/:repo/actions/workflows/:workflow/badge.svg` then
a) we're not really delivering the documented or 'advertised' behaviour of latest build on any branch
b) implementing it as latest build on any branch in future becomes a subtle behaviour change for users who are using `branch=*` and have one or more builds on the default branch.

I think given that, I'm going to say that mapping `branch=*` (or however you represent it "opt in to latest status on any branch") to calling `https://github.com/:user/:repo/actions/workflows/:workflow/badge.svg` actually _doesn't_ make sense as a stop gap.

I reckon we just go with this as it stands (branch param becomes optional) as the fix for #8733 and #8736 and if at some point in future we got the API feature requests of our wildest dreams and revisit doing this with the API for the private repo/self-hosting case, it would have to be a different route given the history.

chris48s:
If you're happy to :+1: this, I'm happy to take deploying/update the various open issues tomorrow

calebcartwright:
> @calebcartwright I went back to change this following our conversation.

I started looking at the code before I saw this and was about to ask if that part was still pending before reading https://github.com/badges/shields/pull/8738#issuecomment-1360241431.

I think the change is fair as well, and will deliver the most consistency with the upstream provider behavior which will hopefully reduce some of the historical confusion there's been around our badge.

I had another thought about potentially handling the private repo case in the future too so feel better about this change not necessarily being mutually exclusive with that one.

calebcartwright (approved):

PR #8755, opened 2022-12-27
add [VpmVersion] badge

### LINKED ISSUE #8752: Add version badge for VPM Repository
### 📋 Description

> - Which service is this badge for?

The version badge for VPM, VRChat Package Manager.

> - What sort of information should this badge show?

As other version badges show, version name should be shown.

For `com.anatawa12.animator-controller-as-a-code` at https://vpm.anatawa12.com/vpm.json, badge should be .

### 🔗 Data

> - Where can we get the data from?
> - Is there a public API?

We can get from VPM Repository JSON.
As there's no official repository to publish user packages, I think it's good to request users for URL to VPM Repository.

> - Does the API require authentication or an API key?

AFAIK, we cannot have access restrictions for VPM repositories. We should just HTTP Get request to the json.

> - Link to the API documentation.

Here's documentation for format of VPM repository: https://vcc.docs.vrchat.com/vpm/repos

### 🎤 Motivation

> What is the specific use case?

Same as endpoints for other package version badges, I wait to put badge on the README.

anatawa12:
Closes #8752

### DIFF
--- a/services/vpm/vpm-version.service.js
+++ b/services/vpm/vpm-version.service.js
@@ -5,6 +5,7 @@ import { BaseJsonService, NotFound } from '../index.js'
 
 const queryParamSchema = Joi.object({
   repository_url: optionalUrl.required(),
+  include_prereleases: Joi.equal(''),
 }).required()
 
 const schema = Joi.object({
@@ -51,13 +52,16 @@ export default class VpmVersion extends BaseJsonService {
     })
   }
 
-  async handle({ packageId }, { repository_url: repositoryUrl }) {
+  async handle(
+    { packageId },
+    { repository_url: repositoryUrl, include_prereleases: prereleases }
+  ) {
     const data = await this.fetch({ repositoryUrl })
     const pkg = data.packages[packageId]
     if (pkg === undefined)
       throw new NotFound({ prettyMessage: 'package not found' })
     const versions = Object.keys(pkg.versions)
-    const version = latest(versions)
+    const version = latest(versions, { pre: prereleases !== undefined })
 
     return renderVersionBadge({ version })
   }
--- a/services/vpm/vpm-version.service.js
+++ b/services/vpm/vpm-version.service.js
@@ -39,6 +39,17 @@ export default class VpmVersion extends BaseJsonService {
       },
       staticPreview: renderVersionBadge({ version: '1.1.6' }),
     },
+    {
+      title: 'VPM Package Version (including prereleases)',
+      namedParams: {
+        packageId: 'com.vrchat.udonsharp',
+      },
+      queryParams: {
+        repository_url: 'https://packages.vrchat.com/curated?download',
+        include_prereleases: null,
+      },
+      staticPreview: renderVersionBadge({ version: '1.1.6' }),
+    },
   ]
 
   static defaultBadgeData = {
--- a/services/vpm/vpm-version.service.js
+++ b/services/vpm/vpm-version.service.js
@@ -4,7 +4,7 @@ import { latest, renderVersionBadge } from '../version.js'
 import { BaseJsonService, NotFound } from '../index.js'
 
 const queryParamSchema = Joi.object({
-  repositoryUrl: optionalUrl.required(),
+  repository_url: optionalUrl.required(),
 }).required()
 
 const schema = Joi.object({
@@ -34,7 +34,7 @@ export default class VpmVersion extends BaseJsonService {
         packageId: 'com.vrchat.udonsharp',
       },
       queryParams: {
-        repositoryUrl: 'https://packages.vrchat.com/curated?download',
+        repository_url: 'https://packages.vrchat.com/curated?download',
       },
       staticPreview: renderVersionBadge({ version: '1.1.6' }),
     },
@@ -51,7 +51,7 @@ export default class VpmVersion extends BaseJsonService {
     })
   }
 
-  async handle({ packageId }, { repositoryUrl }) {
+  async handle({ packageId }, { repository_url: repositoryUrl }) {
     const data = await this.fetch({ repositoryUrl })
     const pkg = data.packages[packageId]
     if (pkg === undefined)
--- a/services/vpm/vpm-version.service.js
+++ b/services/vpm/vpm-version.service.js
@@ -13,7 +13,7 @@ const schema = Joi.object({
     .pattern(
       /./,
       Joi.object({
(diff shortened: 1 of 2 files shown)


services/vpm/vpm-version.service.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  static examples = [
+    {
+      title: 'VPM Package Version',
+      namedParams: {
+        packageId: 'com.vrchat.udonsharp',
+      },
+      queryParams: {
+        repositoryUrl: 'https://packages.vrchat.com/curated?download',
```
chris48s, line 38: I realise that this is a user-supplied parameter so the user can pass in whatever they want, but for a version badge it seems like you could pass in https://packages.vrchat.com/curated (without `?download`) here and works just the same. For the purposes of the example/test, do we need this?

services/vpm/vpm-version.service.js
```diff
@@ -0,0 +1,81 @@
...
+    const pkg = data.packages[packageId]
+    if (pkg === undefined)
+      throw new NotFound({ prettyMessage: 'package not found' })
+    const versions = Object.keys(pkg.versions)
+    let version
+    if (sort === 'defined') {
+      version = versions.reverse()[0]
+    } else if (sort === 'semver') {
+      version = latest(versions)
+    }
```
chris48s, line 77: Can you expand a bit on the reason for allowing multiple sort methods here? I don't really know the norms of this community, but for package registries we generally want the concept of 'latest' to align with that of the upstream registry (What version would I get if I run `vpm install`, or whatever?). We do have this for some badges like github tags/releases where the platform is unaware of the language ecosystem norms or version schema (a github release could be anything - an NPM package, an android app, whatever..). For package registries (NPM, Ruby gems, Rust crates, PyPI, etc) we don't generally do this.

chris48s: Also, normally for version badges, `/v` would be the latest version (excluding pre-releases) and we'd have an `include_prereleases` flag for latest version (including pre-releases). Does VPM have any concept of stable/pre-releases?

services/vpm/vpm-version.service.js
```diff
@@ -0,0 +1,81 @@
+import Joi from 'joi'
+import { optionalUrl } from '../validators.js'
+import { latest, renderVersionBadge } from '../version.js'
+import { BaseJsonService, NotFound } from '../index.js'
+
+const queryParamSchema = Joi.object({
+  repositoryUrl: optionalUrl.required(),
```
chris48s, line 7: Can we make the query param `repository_url`. To make the variable name `repositoryUrl` internally, you can change the function signature for `handle()` to

```js
async handle({ packageId }, { repository_url: repositoryUrl, sort }) {
```

(you'll need to update the examples to use `repository_url`)

services/vpm/vpm-version.service.js
```diff
@@ -0,0 +1,81 @@
...
+  sort: Joi.string().valid('defined', 'semver').default('defined'),
+}).required()
+
+const schema = Joi.object({
+  packages: Joi.object()
+    .pattern(
+      /./,
+      Joi.object({
+        versions: Joi.object().pattern(/./, Joi.object()).required(),
```
chris48s, line 16: We should make this

```js
versions: Joi.object().pattern(/./, Joi.object()).min(1).required(),
```

This will require version to have at least one key. At the moment, `Object.keys(pkg.versions)` could be `[]`.

services/vpm/vpm-version.tester.js
```diff
@@ -0,0 +1,45 @@
+import { isSemver } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
```
chris48s, line 3: The service tests didn't run in CI. This is a [known issue](https://github.com/badges/shields/issues/8109) - we're working on it, but we still have one foot in GitHub actions and one foot in Circle CI. Lets not get too bogged down in that, but I ran the service tests locally and they all failed. Can you have a look at getting the tests passing. You'll need to run them locally.

services/vpm/vpm-version.service.js
anatawa12: The reason why I use url with `? download` os it's the url vpm uses.

services/vpm/vpm-version.service.js
anatawa12: Because official repository template is made based on GitHub releases, in most case, its ordered by release date. however, it's not documented & not a rule for the repository so we can remove this sort.

I think this sort of useful in case of releasing multiple major releases but it may not be a enough reason.

> Does VPM have any concept of stable/pre-releases?
Yes. VPM is based on semver. I forget to check prereleases in `defined` sort.

services/vpm/vpm-version.tester.js
anatawa12: Sorry, i did ran `npm run test` but not ran `npn run test:services`. I fixed tests.

services/vpm/vpm-version.service.js
```diff
@@ -0,0 +1,68 @@
...
+    {
+      title: 'VPM Package Version',
+      namedParams: {
+        packageId: 'com.vrchat.udonsharp',
+      },
+      queryParams: {
+        repository_url: 'https://packages.vrchat.com/curated?download',
+      },
+      staticPreview: renderVersionBadge({ version: '1.1.6' }),
+    },
```
chris48s, line 41: Can we add a second example with the include_prereleases flag?

chris48s (commented):
Nice. This latest round of changes looks great. One more thing and we are good to go.

services/vpm/vpm-version.service.js
```diff
@@ -0,0 +1,79 @@
...
+    {
+      title: 'VPM Package Version (including prereleases)',
+      namedParams: {
+        packageId: 'com.vrchat.udonsharp',
+      },
+      queryParams: {
+        repository_url: 'https://packages.vrchat.com/curated?download',
+        include_prereleases: null,
+      },
+      staticPreview: renderVersionBadge({ version: '1.1.6' }),
```
anatawa12, line 51: should I use fake prerelease version like `1.1.7-beta.1` for static preview?

chris48s: I think it is fine as it stands

chris48s (approved):
great work on this :+1:

PR #8766, opened 2023-01-01
Docstrings for php-version service

pr7prashant:
- Added docstrings for `php-version` service
- Updated docstring for `resource-cache` service


services/php-version.js
```diff
@@ -2,14 +2,23 @@
...
 import { omitv } from './text-formatters.js'
 
-// Return a negative value if v1 < v2,
-// zero if v1 = v2, a positive value otherwise.
+/**
+ * Return a negative value if v1 < v2,
+ * zero if v1 = v2, a positive value otherwise.
+ *
+ * @param {number} v1 - First version for comparison
+ * @param {number} v2 - Second version for comparison
```
chris48s, line 19: I think `v1` and `v2` are actually strings (they can be something like "1.0.0", for example)

pr7prashant: You are right. I've updated the PR now. Thanks

Issue #8437, opened 2022-09-25
Error running `test:services:trace`,  "Cannot find any files matching pattern "$*"" on Windows


alexandresanlim:
:clock11: **When did the problem start?**

When start [this sample](https://github.com/badges/shields/blob/master/doc/service-tests.md#3-running-the-tests) to run my first test

:lady_beetle: **Stack trace**

```
<!-- Provide the complete stack trace from the CircleCI test summary. -->

PS C:\Workspace\shieldsale\shields> npm run test:services:trace -- --only=wercker

> shields.io@0.0.0 test:services:trace
> cross-env NODE_CONFIG_ENV=test TRACE_SERVICES=true npm run test:services -- $*


> shields.io@0.0.0 test:services
> cross-env NODE_CONFIG_ENV=test mocha core/service-test-runner/cli.js

Warning: Cannot find any files matching pattern "$*"
Running tests for 1 services: wercker.


Error: Unknown services: wercker
    at Runner.only (file:///C:/Workspace/shieldsale/shields/core/service-test-runner/runner.js:64:13)
    at file:///C:/Workspace/shieldsale/shields/core/service-test-runner/cli.js:127:10

```

calebcartwright:
The output from this is just.. odd. Can you please confirm you are working a local copy of this repository (as opposed to a fork) and share which commit you are on?

alexandresanlim:
I'm working on this fork:
https://github.com/alexandresanlim/shields

chris48s:
If you just run `npm run test:services -- --only="wercker"`, does that work?

alexandresanlim:
> If you just run `npm run test:services -- --only="wercker"`, does that work?

Yes, I just ran this command according to the first line

chris48s:
OK. Thanks I guess the issue is that the `$*` in

https://github.com/badges/shields/blob/7330da1038904c0be5b21c80ca440bebc07335ed/package.json#L93

doesn't work on Windows. Not sure how we work round that in a cross-platform way.

alexandresanlim:
@chris48s I have new changes to open a PR, can i open it without tests or waiting fix?

chris48s:
If you just want to run the tests for your new service, you can do it with `npm run test:services -- only="your-service-name"`.

`npm run test:services:trace` is useful for debugging failing tests, but you should be able to submit a working PR without this issue being fixed.

alexandresanlim:
`npm run test:services -- only="wercker"` results in:
```
[...]

PR #8350, opened 2022-08-28
get the tests passing on windows

chris48s:
closes #8289

Now that we are running on GHA we can pretty easily just run the tests on Windows and we've got tests covering the functionality in #8289

PR #8786, opened 2023-01-07
fix service tests on windows; run [crates] (to prove service tests still run)

chris48s:
Refs #8437

This doesn't fix the `:trace` variant but it does fix being able to run `npm run test:services -- only="myservice"` on Windows.

The core fix here is actually:

```diff
diff --git a/core/base-service/loader.js b/core/base-service/loader.js
index 9bbf0c948..5fd87cd87 100644
--- a/core/base-service/loader.js
+++ b/core/base-service/loader.js
@@ -103,7 +103,7 @@ async function collectDefinitions() {
 async function loadTesters() {
   return Promise.all(
     glob
-      .sync(path.join(serviceDir, '**', '*.tester.js'))
+      .sync(toUnixPath(path.join(serviceDir, '**', '*.tester.js')))
       .map(async path => await import(`file://${path}`))
   )
 }
```

(basically in https://github.com/badges/shields/pull/8350 we missed the tests loader).

My original plan was to do that and directly write tests for `loadTesters()` and `loadServiceClasses()` but loading all the tester classes with ```.map(async path => await import(`file://${path}`))``` inside an `expect()` was causing some strange side effects with chai, so I decided to separate discovering all the files from loading them and write a test for just making sure we discover the files, which is why there is a bit more refactoring going on in this PR.

Side note: This doesn't have any bearing on anything, but I am unreasonably annoyed with myself, windows users, and the world as a whole that I had to install Windows in a VirtualBox in order to work this out when it basically amounted to applying a fix I had already worked out previously _without_ having to do that.

### DIFF
--- a/core/base-service/loader.js
+++ b/core/base-service/loader.js
@@ -27,11 +27,13 @@ class InvalidService extends Error {
   }
 }
 
+function getServicePaths(pattern) {
+  return glob.sync(toUnixPath(path.join(serviceDir, '**', pattern)))
+}
+
 async function loadServiceClasses(servicePaths) {
   if (!servicePaths) {
-    servicePaths = glob.sync(
-      toUnixPath(path.join(serviceDir, '**', '*.service.js'))
-    )
+    servicePaths = getServicePaths('*.service.js')
   }
 
   const serviceClasses = []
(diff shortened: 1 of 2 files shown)


calebcartwright (approved):
LGTM. Might be worth setting up a recurring CI check with Actions too

PR #8352, opened 2022-08-30
Add various [Polymart] badges

MherZaqaryan:
Added:

- Polymart `Downloads` badge
- Polymart `Latest Version` badge
- Polymart `Rating` badge
- Polymart `Star Rating` badge

Closes #7429

PR #8811, opened 2023-01-15
Add various [Polymart] badges

### LINKED ISSUE #7429: Badge Request: Polymart
:clipboard: **Description**
[Polymart](https://polymart.org/) this is a website similar to [Spigot](https://www.spigotmc.org/) (Spiget) which badge's also available at the moment.
 (https://img.shields.io/spiget/download-size/1234)

:link: **Link**
Official website: https://polymart.org
Public API: https://api.polymart.org
API Documentation: https://polymart.org/wiki/api

:microphone: **Motivation**
This should help Polymart resources look better and also will help some sellers to make their resources look good.
Good Luck!

MherZaqaryan:
Added:

- Polymart `Downloads` badge
- Polymart `Latest Version` badge
- Polymart `Rating` badge
- Polymart `Star Rating` badge

Closes #7429


MherZaqaryan:
> 

@chris48s Hey, I made a new pull request, everything else you said in the previous pull request is now fixed. Also, the specific service to run on Windows is currently working. 

One more thing, could you please help me with this Danger thing? I don't understand what's wrong.

chris48s:
Thanks! I'm aiming to have another look at this tomorrow.
You can ignore the issue with danger. I need to fix it, but it is unrelated to the changes in this PR.

MherZaqaryan:
> Thanks! I'm aiming to have another look at this tomorrow. You can ignore the issue with danger. I need to fix it, but it is unrelated to the changes in this PR.

Alright! Thanks for letting me know!

services/polymart/polymart-downloads.service.js
```diff
@@ -0,0 +1,36 @@
...
+      },
+      staticPreview: renderDownloadsBadge({ downloads: 655 }),
+      documentation,
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'downloads',
+    color: 'green',
+  }
```
chris48s, line 27:
```suggestion
  static defaultBadgeData = {
    label: 'downloads'
  }
```

I don't think we need to set a default colour here. Either `renderDownloadsBadge()` or throwing a `NotFound()` will set one for us.

services/polymart/polymart-latest-version.service.js
```diff
@@ -0,0 +1,45 @@
...
+      },
+      staticPreview: this.render({
+        version: '1.2.9',
+      }),
+      documentation,
+    },
+  ]
+
+  static defaultBadgeData = {
+    label: 'version',
```
chris48s, line 26: Usually our version badges use the name of the service as the label. I suggest we make the label for this one 'polymart' (and update the tests)

services/polymart/polymart-latest-version.service.js
```diff
@@ -0,0 +1,45 @@
...
+      message: `${version}`,
+    }
+  }
+
+  async handle({ resourceId }) {
+    const { response } = await this.fetch({ resourceId })
+    if (!response.resource) {
+      throw new NotFound()
+    }
+    return this.constructor.render({
```
chris48s, line 41: Same comment as https://github.com/badges/shields/pull/8352#discussion_r959829384
Can we use `renderVersionBadge()` here?

services/polymart/polymart-latest-version.tester.js
```diff
@@ -0,0 +1,15 @@
+import { withRegex } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('Polymart Plugin (id 323)')
+  .get('/323.json')
+  .expectBadge({
+    label: 'version',
+    message: withRegex(/^(?!not found$)/),
```
chris48s, line 9: This would pass even if the badge was something like 'inaccessible' or some other error. We are going to need to use a more specific validator here. I don't know what the requirements are for a polymart version. Are there any docs? If it is pretty loose, maybe something like `isVPlusDottedVersionNClauses` would be a good choice? https://github.com/badges/shields/blob/2fb98897429c9a5bfa2960a85461a2df66654222/services/test-validators.js#L17-L18

services/polymart/polymart-latest-version.service.js
chris48s: Once we're doing that, we could also ditch the default colour in `defaultBadgeData` for this class.

chris48s (commented):
Thanks for picking this up again.

services/polymart/polymart-latest-version.tester.js
MherZaqaryan: The version for it doesn't have any specific form, you can check out the documentation at https://polymart.org/wiki/api.

chris48s:
We don't need to copy and paste all the code from version.js into your file. We can import `renderVersionBadge` using `import { renderVersionBadge } from '../version.js'`.

Maybe looking at something like https://github.com/badges/shields/blob/master/services/cdnjs/cdnjs.service.js as an example would help?

MherZaqaryan:
> We don't need to copy and paste all the code from version.js into your file. We can import `renderVersionBadge` using `import { renderVersionBadge } from '../version.js'`.
> 
> Maybe looking at something like https://github.com/badges/shields/blob/master/services/cdnjs/cdnjs.service.js as an example would help?

Yep, thank you, I just fixed it, I actually didn't get a response from my question in the last pull request, and thought that I have to add it manually https://github.com/badges/shields/pull/8352#discussion_r959916503

chris48s (approved):
Thanks. Latest changes look good :+1: 
I'll get this deployed in the next few days

PR #8813, opened 2023-01-16
fix build according to badge/shields#8671

samkenxstream:


samkenxstream (commented):
It is merged clean

chris48s:
Did you submit this PR to the wrong repo?

calebcartwright:
I'm going to close given the inexplicable nature of the change and the dearth of any accompanying context or explanation, similar to e.g. https://github.com/nodejs/node/pull/46201

I don't think we need to block the user per se at this point though as this could have been a simple mistake

PR #8815, opened 2023-01-16
remove danger-plugin-no-test-shortcuts

chris48s:
Turns out this is the thing that was trying to read from the local checkout:
https://github.com/macklinu/danger-plugin-no-test-shortcuts/blob/ebd8c91e333d96a7c6d6e5e6fa1411d0fb460a40/src/index.ts#L49

I think the short term solution is to just ditch this check to get danger working again.

Longer term, I'll make an issue to find another solution to this. One option would be to re-implement the check in danger but based on the diff instead of the local files (like we do with some of the other checks). The other option would be to try and write it as a lint/pre-commit check (we're basically checking the same thing for mocha tests via eslint: https://github.com/badges/shields/blob/2fb98897429c9a5bfa2960a85461a2df66654222/.eslintrc.yml#L115 ).

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -11,7 +11,6 @@
 // DANGER_GITHUB_API_TOKEN=your-github-api-token npm run danger -- pr https://github.com/badges/shields/pull/2665
 
 const { danger, fail, message, warn } = require('danger')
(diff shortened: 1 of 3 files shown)


calebcartwright (approved):
Agreed. Better to get this check working again holistically since we can guard against this case via reviews in the interim

Issue #8819, opened 2023-01-17
Implement check for `.only()` in service tests


chris48s:
:clipboard: **Description**

In https://github.com/badges/shields/pull/8815 I removed danger-plugin-no-test-shortcuts because it wasn't playing nice with the `pull_request_target` workflow trigger
As noted in the PR, we should re-implement this check, either using danger but based off the diff or as some kind of lint check.

PR #8839, opened 2023-01-22
Add ESLint checks for IcedFrisby exclusive/skipped tests

chris48s:
Closes https://github.com/badges/shields/issues/8819

We had previously been using danger for this, mainly because there was a plugin we could easily use to do it but really this should have ideally been an ESLint rule, much like `mocha/no-exclusive-tests` (which we already use).

Also I quite fancied learning how to make an ESLint plugin, so I made https://github.com/chris48s/eslint-plugin-icedfrisby

This PR adds `eslint-plugin-icedfrisby` to our ESLint setup. Also I've enabled `mocha/no-skipped-tests` which we were not using previously, but probably should have been.

Obviously there is not much going on in this PR as all the actual ESLint code lives somewhere else, but do feel free to have a nose about in https://github.com/chris48s/eslint-plugin-icedfrisby as part of the review. I suspect it is imperfect, but also an improvement over what we previously had with the danger plugin we were using. Also, if there are any other custom rules we would like to add for working with IcedFrisby in our service tests, that plugin repo gives us a place for them to live.


calebcartwright (approved):
Sgtm, will take a look at the upstream code at some point as well

PR #8886, opened 2023-02-04
Add [ChessCom] rating and social for chess game

alexandresanlim:
Added

- Badge to user rating
- Badge to user followers

### DIFF
--- a/services/chesscom/chesscom-base.js
+++ b/services/chesscom/chesscom-base.js
@@ -1,6 +1,6 @@
 import { BaseJsonService } from '../index.js'
 
-class BaseStatsChessDotComService extends BaseJsonService {
+class BaseChessComStatsService extends BaseJsonService {
   static category = 'rating'
 
   static defaultBadgeData = { label: 'rating' }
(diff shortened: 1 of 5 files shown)


calebcartwright:
Thank your contribution!

However, I don't think this is something that makes sense for us to proceed with. The Shields service is, first and foremost, targeted at developers and technologists. One of the considerations we have for new badge requests is whether the upstream service/platform is strictly, or at least primarily, used by developers/technologists. There are a few badges we offer with native routes, e.g. Twitter, that are admittedly a bit fuzzy around this line, though we can rationalize those since they are services heavily used by our user base.

Unfortunately, I don't feel like Chess.com meets any of those categories, as an online Chess gaming platform, and so I'm going to close this.

Users may still utilize our more open ended features, such as the dynamic badge (https://img.shields.io/badge/dynamic/json?label=rating&query=%24.chess_daily.last.rating&url=https%3A%2F%2Fapi.chess.com%2Fpub%2Fplayer%2Fhikaru%2Fstats)


to get badges, or even the [Custom Endpoint](https://shields.io/endpoint) for those that want to do more advanced customization

Issue #7982, opened 2022-05-18
Choose a new static-site generator


paulmelnikow:
Gatsby isn't the best fit for us. I think these are the key pain points (though other maintainers may describe them better!)

1. It's laborious to upgrade. The monorepo publishing mechanism issues lots of empty version bumps. We've worked around theis.
2. The project has been slow to implement key features such as gatsbyjs/gatsby/discussions/1878 (which we've worked around) and gatsbyjs/gatsby#31599
3. The GraphQL functionality feels like bloat: a feature we don't need, which complicates our usage of GraphQL in the server. Probably could be worked around with a different repo organization.
4. Big dependency tree complicates installation in Windows (see https://github.com/badges/shields/pull/7974#issuecomment-1128292215)

From https://github.com/badges/shields/pull/7978#issuecomment-1130411676:

> Straying a bit off topic here, but do you have any thoughts on what a good replacement for Gatsby is? We were using next.js before and decided we didn't like that. I don't really know what the other options are in that space.

I think the ideal criteria are:

- Easy to programmatically generate static pages which contain dynamic React content
- Works well with React and TypeScript
- Well documented

For single-page apps, I use [Create React App](https://create-react-app.dev/), though it doesn't generate static pages. I've been meaning to check out [Vite](https://vitejs.dev/), which feels like a fresh look at the problem, though I think it may have the same limitation in the sense that static site generation is not what it does.

As a starting point I would suggest we evaluate these:

- [Astro](https://astro.build/)
- [Remix](https://remix.run/)
- [Docusaurus](https://docusaurus.io/)

There are a few other static site generators for React listed here: https://jamstack.org/generators/

<img width="401" alt="Screen Shot 2022-05-18 at 3 02 04 PM" src="https://user-images.githubusercontent.com/1487036/169145886-128a89ba-0915-45c4-9a0f-035b704ad3a3.png">

And… I'm curious if other folks might have ideas of options to consider here! (cc @Daniel15)

Daniel15:
>  I'm curious if other folks might have ideas of options to consider here! 

Docusaurus is pretty great for mostly static sites - I'd recommend trying it out. It does let you embed React components too 😄 

> For single-page apps, I use [Create React App](https://create-react-app.dev/), though it doesn't generate static pages

You can use something like `react-snap` to generate static pages - that's what I do for the home page on https://dnstools.ws/.

chris48s:
[...]

PR #8887, opened 2023-02-05
documentation html fixes

chris48s:
In connection with https://github.com/badges/shields/issues/7982 I have started noodling on producing an OpenAPI spec for our endpoints to be consumed by docusaurus-openapi.

One of the side effects of this is rendering our html documentation fragments using JSX throws a hard error on invalid HTML and also has some indenting requirements.

This PR contains 2 categories of changes:
1. Fixing invalid HTML fragments
2. Minor whitespace changes that have no effect on how the help renders in our current site but will help it the fragments render right in docusaurus-openapi.

This might not cover 100% of the html help fragments, but its covers the stuff I've git so far.


calebcartwright (approved):
Good with the changes proposed here, though curious/mildly concerned about how we prevent such changes from being necessary in the future

chris48s:
I think in the immediate term, there is not much we can do to prevent more of these issues creeping in and I'll probably have to do another sweep. Assuming we do go down the docusaurus  route, `npm run build` will fail on invalid HTML (e.g: opening `<tag>` with no closing `</tag>`) so that's a pretty easy CI check for that. The more spacing-related issues are probably going to come down to checking "does the page look right" at review time.

Issue #8906, opened 2023-02-14
File UIDs/GIDs in docker image messed up since server-2022-12-01


rroesch1:
### Are you experiencing an issue with...

My own instance of shields

### 🐞 Description

The UIDs/GIDs of files in the docker image are messed up since `shieldsio/shields:server-2022-12-01`

```bash
$ docker run -it shieldsio/shields:server-2022-12-01 /bin/sh -c 'ls -lah ./node_modules/*/package.json | egrep -v "root\\s+root"'
-rw-r--r--    1 node     node         551 Dec  1 19:03 ./node_modules/after/package.json
-rw-r--r--    1 501      dialout      413 Dec  1 19:03 ./node_modules/arraybuffer.slice/package.json
-rw-r--r--    1 501      dialout      282 Dec  1 19:03 ./node_modules/backo2/package.json
-rw-r--r--    1 501      dialout     1.0K Dec  1 19:03 ./node_modules/balanced-match/package.json
-rw-r--r--    1 node     node         971 Dec  1 19:03 ./node_modules/base64-arraybuffer/package.json
-rw-r--r--    1 501      dialout      639 Dec  1 19:03 ./node_modules/bintrees/package.json
-rw-r--r--    1 node     node        2.3K Dec  1 19:03 ./node_modules/check-error/package.json
-rw-r--r--    1 501      dialout      381 Dec  1 19:03 ./node_modules/component-bind/package.json
-rw-r--r--    1 502      dialout      356 Dec  1 19:03 ./node_modules/component-inherit/package.json
-rw-r--r--    1 501      dialout      651 Dec  1 19:03 ./node_modules/core-util-is/package.json
-rw-r--r--    1 501      dialout     1.1K Dec  1 19:03 ./node_modules/debug/package.json
-rw-r--r--    1 501      dialout      751 Dec  1 19:03 ./node_modules/decode-uri-component/package.json
-rw-r--r--    1 501      dialout     1.0K Dec  1 19:03 ./node_modules/deep-is/package.json
-rw-r--r--    1 501      dialout     2.7K Dec  1 19:03 ./node_modules/esprima/package.json
-rw-r--r--    1 501      dialout     1.1K Dec  1 19:03 ./node_modules/esutils/package.json
-rw-r--r--    1 501      dialout      998 Dec  1 19:03 ./node_modules/fast-levenshtein/package.json
-rw-r--r--    1 501      dialout      639 Dec  1 19:03 ./node_modules/filter-obj/package.json
[...]

PR #8908, opened 2023-02-15
fix: Set uid/gid in docker image to 0

rroesch1:
* UIDs/GIDs in the docker image are messed up due to strange UIDs/GIDs in tarballs from the npm registry
* This change overrides the UID/GID of all files in `/usr/src/app`

fixes: badges/shields#8906


chris48s:
Cheers. I tested this locally and it doesn't appear to introduce any problems

chris48s (approved):

Issue #8921, opened 2023-02-17
Proposal: migrate GitHub token pool to fly.io postgres


chris48s:
:clipboard: **Description**

There are several known issues with our existing approach to storing all of our tokens in a redis set.
1. The number of tokens we now have is quite large. The size of the data has somewhat outgrown the approach.
2. Storing everything in a single redis key has been kind of a blocker to https://github.com/badges/shields/issues/4169 . We ended up abandoning https://github.com/badges/shields/pull/7037 mainly because we couldn't agree on how best to smoosh more structured data into it, rather than because the idea was fundamentally bad.
3. ComposeDB is kinda expensive for what we actually use it for. In a typical month we spend ~$70 with fly.io, $7 with Heroku and $18.50 with ComposeDB.

I've been mostly ignoring this because it is _not on fire right now_ and the token pool code is a giant PITA, but ComposeDB is shutting down in 2 weeks and finding a drop-in replacement that retains all the same things we had from ComposeDB has been not impossible but also not trivial (conversation about that on Discord). So now this is _on fire right now_ 🔥.

I think the next place we should head to is storing our tokens in a SQL database. In the past running a whole managed SQL database for the one table we need would have been kinda overkill, but fly kinda changes the game. On fly.io we can run a single node, 1x shared CPU, 256MB RAM, 1GB disk [semi-managed](https://fly.io/docs/postgres/getting-started/what-you-should-know/) postgres server for $1.94/mo, which should be plenty tbh (but untested). We'd be responsible for our own upgrades, but backups are included. Having everything inside fly.io (rather than straddling multiple clouds) is also convenient. Moving to a table with one row per token also makes it natural to start adding columns to store permissions against tokens so we can do stuff like `SELECT FROM tokens WHERE read_packages=True` when we need a token with the `read:packages` scope, for example.

There are a few things that occur to me as possible considerations:

1. If you want to work on the token pool, you'll need to run postgres locally. This is already the case for redis - at the moment if you want to work on the token pool, you need to run redis locally. We mostly abstract it from 99% of contributors with redis and we can continue to do so with postgres. Not a big deal IMO.
2. We'll need to run postgres in CI to run the integration tests. This also isn't a big deal. We can run postgres as a service in a GH workflow in much the same way we do with redis.
3. We'll need to migrate our data from redis to postgres. Again, I don't think this is a huge deal. We'll just need to write a one-off script to move the data from A to B.
[...]

PR #8922, opened 2023-02-18
migrate token pooling to postgres

chris48s:
This PR implements the suggestion from issue #8921 allowing us to migrate the token pool from redis to fly.io postgres.


core/token-pooling/sql-token-persistence.integration.js
```diff
@@ -0,0 +1,103 @@
...
+    const initialTokens = ['a', 'b', 'c'].map(char => char.repeat(40))
+
+    beforeEach(async function () {
+      initialTokens.forEach(async token => {
+        await pool.query(
+          `INSERT INTO pg_temp.${tableName} (token) VALUES ($1::text);`,
+          [token]
+        )
+      })
+    })
```
chris48s, line 63: This isn't the most efficient way to insert this data, but given we are only dealing with 3 records.. meh :shrug:

package.json
```diff
@@ -54,8 +54,10 @@
     "lodash.groupby": "^4.6.0",
     "lodash.times": "^4.3.2",
     "node-env-flag": "^0.1.0",
+    "node-pg-migrate": "^6.2.2",
```
chris48s, line 57: Even though we don't expect to change our DB schema super often, I think its worth having a migration package to make it easier to evolve this over time. Ideally we should also look at running `npm run migrate up` on deploy. For the first deploy I think it can live with doing it manually.

migrations/1676731511125_initialize.cjs
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable camelcase */
```
chris48s, line 1: This needs to be a CommonJS file even though our project is ESM, which is why it uses the `.cjs` extension

core/token-pooling/sql-token-persistence.js
```diff
@@ -0,0 +1,55 @@
...
+    return result.rows.map(row => row.token)
+  }
+
+  async stop() {
+    await this.pool.end()
+  }
+
+  async onTokenAdded(token) {
+    return await this.pool.query(
+      `INSERT INTO ${this.table} (token) VALUES ($1::text) ON CONFLICT (token) DO NOTHING;`,
```
chris48s, line 28: `ON CONFLICT (token) DO NOTHING` effectively makes this an "upsert" which closely mirrors the behaviour of a redis `sadd()`

package-lock.json
```diff
@@ -52756,14 +53150,12 @@
         "ansi-regex": {
           "version": "5.0.0",
```
chris48s, line 53151: enforce-dependency-review is complaining about this. I figure we're going to merge a bunch of dependabot stuff ahead of this and I'm going to need to resolve a conflict on `package-lock.json` anyway. Lets ignore it for now and circle back when we're ready to merge it. I've noted it.

core/token-pooling/sql-token-persistence.integration.js
```diff
@@ -0,0 +1,103 @@
...
+  })
+
+  beforeEach('Create temporary table', async function () {
+    await pool.query(
+      `CREATE TEMPORARY TABLE ${tableName} (LIKE github_user_tokens INCLUDING ALL);`
+    )
+  })
+  afterEach('Drop temporary table', async function () {
+    await pool.query(`DROP TABLE IF EXISTS pg_temp.${tableName};`)
+  })
```
chris48s, line 44: This copies our table to a temp schema for the purposes of running the tests then clears it up again after. This means we can run the tests against a local DB with real data in it and not break our local data. We don't need a separate test DB. It also means the tests will keep step with schema changes from the migrations. We do need to `migrate up` before running the tests though.

package.json
```diff
@@ -117,7 +119,8 @@
     "e2e": "start-server-and-test start http://localhost:3000 test:e2e",
     "e2e-on-build": "cross-env CYPRESS_baseUrl=http://localhost:8080 start-server-and-test start:server:e2e-on-build http://localhost:8080 test:e2e",
     "badge": "cross-env NODE_CONFIG_ENV=test TRACE_SERVICES=true node scripts/badge-cli.js",
-    "build-docs": "rimraf api-docs/ && jsdoc --pedantic -c ./jsdoc.json . && echo 'contributing.shields.io' > api-docs/CNAME"
+    "build-docs": "rimraf api-docs/ && jsdoc --pedantic -c ./jsdoc.json . && echo 'contributing.shields.io' > api-docs/CNAME",
+    "migrate": "node scripts/write-migrations-config.js > migrations-config.json && node-pg-migrate --config-file=migrations-config.json"
```
chris48s, line 123: This is a bit hacky. I want the DB config to be specify-able in yaml or env vars like all our other setting but `node-pg-migrate` has its own ideas about how to handle config https://salsita.github.io/node-pg-migrate/#/cli . This little script reads the postgres connection string (however it has been set - yaml or env vat) and writes it to a config file for `node-pg-migrate`. I did consider writing a script that outputs the connection string to stdout and then setting an env var with `DATABASE_URL=$(node /scripts/get-dataabse-url.js)` but I picked this way because it should be cross-platform for Windows users.

server.js
```diff
@@ -42,6 +42,12 @@ if (fs.existsSync('.env')) {
   process.exit(1)
 }
 
+if (config.private.redis_url != null) {
+  console.warning(
+    'RedisTokenPersistence is deprecated for token pooling and will be removed in a future release. Migrate to SqlTokenPersistence'
+  )
+}
+
```
chris48s, line 50: I decided that it is useful to keep both backends around for some length of time because it allows us to switch between them just by changing an env var - no code changes needed. Doing this means we can:

- deploy this code with `REDIS_URL` set (so we're still using the redis backend)
- migrate our data to postgres with `scripts/migrate-token-pool.js`
- set `POSTGRES_URL` to switch to the postgres backend
- if it all goes wrong, we can easily just unset `POSTGRES_URL` to go back to redis
- unset `REDIS_URL` once we're happy everything is fine
- delete the redis code at a later date

This both makes the migration super low-risk/easy to back out of and means we can provide a migration path for self-hosting users with basically no effort.

scripts/migrate-token-pool.js
```diff
@@ -0,0 +1,42 @@
+import pg from 'pg'
+import RedisTokenPersistence from '../core/token-pooling/redis-token-persistence.js'
+
+let redisUrl, postgresUrl
+
+try {
+  redisUrl = process.argv[2]
+  postgresUrl = process.argv[3]
```
chris48s, line 8: For this script it is useful to pass the connection strings in on the console instead of using the application config because it allows us to copy the data over before we set `POSTGRES_URL` to switch the pool over.

Also I'm not too worried about achieving perfection in this specific script. It is a file we are going to use once and then delete it.

core/token-pooling/sql-token-persistence.integration.js
```diff
@@ -0,0 +1,103 @@
...
+        const toRemove = expected.pop()
+
+        await persistence.initialize(pool)
+        await persistence.noteTokenRemoved(toRemove)
+
+        const result = await pool.query(
+          `SELECT token FROM pg_temp.${tableName};`
+        )
+        const savedTokens = result.rows.map(row => row.token)
+        expect(savedTokens.sort()).to.deep.equal(expected)
```
chris48s, line 99: The test cases in this file are the exact same ones we have for redis.

server.js
calebcartwright: Might be worth extending the warning message to include a link to https://github.com/badges/shields/issues/8921

core/token-pooling/sql-token-persistence.integration.js
calebcartwright: > This means we can run the tests against a local DB with real data in it and not break our local data. We don't need a separate test DB. It also means the tests will keep step with schema changes from the migrations. We do need to migrate up before running the tests though

No objections to this in the local/inner dev loop (nor CI) context, and agree the benefits outweigh the tradeoffs. Would be nervous abut this in the context of the real DB though (whether connected to deliberately or accidentally)

migrations/1676731511125_initialize.cjs
```diff
@@ -0,0 +1,14 @@
...
+
+exports.shorthands = undefined
```
calebcartwright, line 3: What's this?

package.json
calebcartwright: Does this need to be a prod dep or can it still be utilized as a dev dep?

calebcartwright:
LGTM, happy to :+1: once merge conflicts are sorted

package.json
chris48s: In order to create the DB (in the first instance) or modify it (later) we need to be able to run `npm run migrate up` in the production environment so it has to be a prod dependency.

core/token-pooling/sql-token-persistence.integration.js
chris48s: In principle it should be safe but I agree - I wouldn't want to run the tests while connected to the prod DB. Tbh that would be fairly unusual thing to do for any application. I think the fact that the dev dependencies (e.g: mocha, etc) aren't installed in the prod container is enough of a guard rail to stop this from happening.

Personally I'd tend to avoid connecting my dev environment up to a prod database for any application.

server.js
chris48s: Yes. I'll call this out at the top of the changelog for the next server release and write it with a bit more of a focus on self-hosting users than the discussion in https://github.com/badges/shields/issues/8921 (which is quite shields.io-specific). I'm going to set myself a reminder to ping me so I remember to do this when I do the changelog.

@SnoozeThis wait until Feb 28 2023

migrations/1676731511125_initialize.cjs
chris48s: https://github.com/salsita/node-pg-migrate/blob/9331f6fda98795e3f3733461f9b71345168b99d8/docs/migrations.md?plain=1#L3-L19

chris48s:
@SnoozeThis wait until Feb 28 2023

SnoozeThis:
(https://snoozeth.is/0ky67I-_i6g) I will wait until Tue, 28 Feb 2023 00:00:00 UTC and then add a comment.

chris48s:
Should be good to go, I think.

I can look at setting up a postgres instance on fly and getting the actual data migration done in the week.

calebcartwright (approved):

SnoozeThis:
Resolved.

PR #8931, opened 2023-02-23
remove migrate-token-pool.js

chris48s:
I'll leave this PR sitting open for a few days just in case but this script has basically done its job and can be deleted.


calebcartwright (approved):
:+1: whenever you're ready

Issue #8932, opened 2023-02-23
Remove `RedisTokenPersistence`


chris48s:
:clipboard: **Description**

Lets leave it in for the March and April releases then after we've done the April release, remove it so it is gone in the May release.

@SnoozeThis wait until April 2 2023

SnoozeThis:
(https://snoozeth.is/XjxEWFli9B4) I will wait until Sun, 02 Apr 2023 00:00:00 UTC and then add a comment.

chris48s:
also remember `REDIS_URL`, associated documentation, integration tests, etc

SnoozeThis:
Resolved.

PR #8933, opened 2023-02-23
update docs following token pool changes

chris48s:
Did a quick skim over all the docs relating to token pooling/redis/postgres and made some updates to reflect the current state of play. Tbh I probably should have done some of this in #8922 but I didn't.. so I'm doing it now.


doc/production-hosting.md
```diff
@@ -14,49 +14,41 @@ Production hosting is managed by the Shields ops team:
 [operations issues]: https://github.com/badges/shields/issues?q=is%3Aissue+is%3Aopen+label%3Aoperations
 [ops discord]: https://discordapp.com/channels/308323056592486420/480747695879749633
 
-| Component                     | Subcomponent                    | People with access                                              |
-| ----------------------------- | ------------------------------- | --------------------------------------------------------------- |
-| shields-io-production         | Full access                     | @calebcartwright, @chris48s, @paulmelnikow                      |
-| shields-io-production         | Access management               | @calebcartwright, @chris48s, @paulmelnikow                      |
-| Compose.io Redis              | Account owner                   | @paulmelnikow                                                   |
-| Compose.io Redis              | Account access                  | @paulmelnikow                                                   |
-| Compose.io Redis              | Database connection credentials | @calebcartwright, @chris48s, @paulmelnikow, @pyvesb             |
```
chris48s, line 23: Just to clarify what changed here (prettier reformatted the whole table so it is hard to tell), the only change here is I removed the 3 rows about compose.io.

There might be some other things in this table that could do with review but that can be another PR.

CONTRIBUTING.md
```diff
@@ -134,9 +134,20 @@ Prettier before a commit by default.
 When adding or changing a service [please write tests][service-tests], and ensure the [title of your Pull Requests follows the required conventions](#running-service-tests-in-pull-requests) to ensure your tests are executed.
 When changing other code, please add unit tests.
 
-To run the integration tests, you must have Redis installed and in your PATH.
-Use `brew install redis`, `yum install redis`, etc. The test runner will
-start the server automatically.
+The integration tests are not run by default. For most contributions it is OK to skip these unless you're working directly on the code for storing the GitHub token pool in posgres/redis.
```
calebcartwright, line 137: I think the `t` is still used even in the shorthand 

```suggestion
The integration tests are not run by default. For most contributions it is OK to skip these unless you're working directly on the code for storing the GitHub token pool in postgres/redis.
```

CONTRIBUTING.md
```diff
@@ -134,9 +134,20 @@ Prettier before a commit by default.
...
+
+To run the integration tests:
+
+- You must have Redis installed and in your PATH. Use `brew install redis`, `apt-get install redis`, etc. The test runner will start the server automatically.
```
calebcartwright, line 141: Do you think it would be worth going ahead and nuking this? Understand it's accurate atm, but I think that's likely to change fairly quickly (guessing your plan is to delete this when we drop the redis portion of the codebase?)

calebcartwright (approved):
Only minor items noted inline so will go ahead and approve in case you want to press on. Happy to re-:+1: though if you want to make any adjustments

CONTRIBUTING.md
chris48s: yeah this is just me being bad at typing/checking things :)

CONTRIBUTING.md
chris48s: I think it can be removed when I do https://github.com/badges/shields/issues/8932

For the moment, if you try to run the integration tests without redis they won't work so I think it is worth keeping it in the docs for the moment

calebcartwright (approved):

Issue #8892, opened 2023-02-08
Investigate GitHub merge queues


chris48s:
:clipboard: **Description**

Launched today: https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/

At the moment we are using repo ranger for this. Try out GH's native implementation on a test repo. Would this give us the same functionality we currently get from repo ranger?

chris48s:
We now have the beta enabled on our repo

![Screenshot at 2023-02-10 14-31-58](https://user-images.githubusercontent.com/6025893/218117322-7567f933-ae5b-4810-8b02-21faf7369404.png)

chris48s:
OK. I enabled this, got it working ( https://github.com/badges/shields/pull/8943 ) and merged this week's dependabot PRs using the GH merge queue instead of repo ranger.

Initial thoughts:

- **Pro**: If we queue a bunch of PRs to merge, this gets all the PRs merged much faster than repo ranger (while maintaining the "Require branches to be up to date before merging" property)
- **Pro**: If we queue a bunch of PRs to merge, this performs fewer builds than repo ranger, reducing the scope for us to hit rate limits
- **Pro**: Because the merge queue doesn't update the source branch, we won't have the issue that `@dependabot rebase` won't work because another bot has committed to the branch
- **Con**: We will run service tests on the initial PR, but once they're in the queue, the service test builds say "No services found. Nothing to do.". This is because the selection of services to run are normally picked up from the PR title and once a commit is in the queue, its not a PR anymore. It is on a `gh-readonly-queue/**` branch. There is probably no workaround for this.
- **Unknown**: I haven't really got a handle on failure modes. I am not clear what happens if a  build fails. What happens to the commit it failed on, and then what happens to the other commits behind it in the queue?
- **Unknown**: I haven't really got a handle on merge conflict handling, either

Based on my experience so far, I think I'm of the opinion that a GH merge queue has enough advantages over Repo Ranger that we should go with this and remove Repo Ranger, but we will continue to evaluate.

@calebcartwright  @PyvesB  - would be interesting to get any feedback from you as well.

Now that the queue is enabled, once we've approved a PR, "merge when ready" will add it to the queue.

![Screenshot at 2023-07-14 18-12-18](https://github.com/badges/shields/assets/6025893/610a64b1-8643-498f-b9e9-10ff5032d4c3)

https://github.blog/2023-07-12-github-merge-queue-is-generally-available/

calebcartwright:
[...]

PR #8943, opened 2023-02-24
run service/docker builds on merge queue builds

chris48s:
refs #8892

I gave merge queues a quick spin. First problem I hit was that these builds are set to only trigger on pull requests so the queue doesn't try to run them, but then it never merges the thing at the head of the queue because its waiting for these builds (which haven't triggered) to run, so it just waits forever :upside_down_face: 

I suspect this won't be the only thing we need to change, but doing this should unblock finding out what the next one is..

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -1,6 +1,9 @@
 name: Build Docker Image
 on:
   pull_request:
+  push:
+    branches:
+      - 'gh-readonly-queue/**'
 
 jobs:
   build-docker-image:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-docker-image.yml
```diff
@@ -1,6 +1,9 @@
 name: Build Docker Image
 on:
   pull_request:
+  push:
+    branches:
+      - 'gh-readonly-queue/**'
```
calebcartwright, line 6: Is the double * necessary/significantly different than a single *?

calebcartwright (approved):
I don't fully understand the implications of this, but trust you that it's something we need :+1:

.github/workflows/build-docker-image.yml
chris48s: Its a glob pattern, so * matches any character other than a slash. ** matches slashes too,

chris48s:
Really lost sight of this one. I think I'm going to give merge queues another spin.

PR #8923, opened 2023-02-19
Add [Vcpkg] version service

njakob:
Adds a new version service for [Vcpkg](https://vcpkg.io) packages and fixes #5816.

All the data regarding Vcpkg can be found in its public repository [microsoft/vcpkg](https://github.com/microsoft/vcpkg) and therefore does not require any extra authentication than GitHub. Besides, you can find more information about its structure in the following [documentation](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json).

For example, `/vcpkg/v/entt` would extract the following [data](https://github.com/microsoft/vcpkg/blob/master/ports/entt/vcpkg.json) and you can find a visual representation on https://vcpkg.link/ports/entt (community project built by myself).

entt

PR #8945, opened 2023-02-25
Fix [Vcpkg] version service for different version fields

njakob:
This PR fixes the missing cases for other version fields that can be present in Vcpkg manifest which were not considered in the initial implementation (#8923).

This includes the following cases (see [documentation](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json#version)):

- `version-date`: [`opengl`](https://github.com/microsoft/vcpkg/blob/223d33be7d3ec6c3d64381ca37f501b8c87dda6a/ports/opengl/vcpkg.json#L3) 

   opengl
- `version-semver`: [`nlohmann-json`](https://github.com/microsoft/vcpkg/blob/223d33be7d3ec6c3d64381ca37f501b8c87dda6a/ports/nlohmann-json/vcpkg.json#L3)
   
   nlohmann-json
- `version-string`: [`7zip`](https://github.com/microsoft/vcpkg/blob/223d33be7d3ec6c3d64381ca37f501b8c87dda6a/ports/7zip/vcpkg.json#L3)

   7zip
- `version`: [`entt`](https://github.com/microsoft/vcpkg/blob/223d33be7d3ec6c3d64381ca37f501b8c87dda6a/ports/entt/vcpkg.json#L3)

   entt


services/vcpkg/vcpkg-version.tester.js
```diff
@@ -1,11 +1,23 @@
-import { isSemver } from '../test-validators.js'
+import Joi from 'joi'
 import { createServiceTester } from '../tester.js'
 
 export const t = await createServiceTester()
 
-t.create('gets the port version of entt')
+t.create('gets the port version of entt using `version` field')
```
njakob, line 6: **note**: These tests aim to validate the 4 different fields that can be included in Vcpkg manifests. However, if these specific ports use suddenly a different field, the tests won't cover all of them anymore. If this is a problem, we could imagine having a different collection of tests against static data.

services/vcpkg/vcpkg-version.tester.js
```diff
@@ -1,11 +1,23 @@
...
   .get('/entt.json')
-  .expectBadge({ label: 'vcpkg', message: isSemver })
+  .expectBadge({ label: 'vcpkg', color: 'blue', message: Joi.string() })
```
njakob, line 8: **note**: Related to the [other comment](https://github.com/badges/shields/pull/8945/files#r1117987743), in order to avoid having tests failing if the version pattern changes (e.g. date format is changed to semantic versioning), we only test if the version has been properly extracted by looking at the color of the badge.

services/vcpkg/vcpkg-version.service.js
```diff
@@ -4,9 +4,20 @@ import { fetchJsonFromRepo } from '../github/github-common-fetch.js'
 import { renderVersionBadge } from '../version.js'
 import { NotFound } from '../index.js'
 
-const vcpkgManifestSchema = Joi.object({
-  version: Joi.string().required(),
-}).required()
```
chris48s, line 9: Can add a comment here linking to https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json?source=recommendations#version to make it a bit more obvious what is going on here

services/vcpkg/vcpkg-version.service.js
```diff
@@ -4,9 +4,20 @@ import { fetchJsonFromRepo } from '../github/github-common-fetch.js'
...
+const vcpkgManifestSchema = Joi.alternatives().try(
```
chris48s, line 7: If the input file contains more than one of these, it is an invalid manifest and we won't know which one to pick. Lets use Joi's 'one' match mode https://joi.dev/api/?v=17.8.1#alternativesmatchmode to fail if more than one exists.

services/vcpkg/vcpkg-version.tester.js
chris48s: How about we keep one "live" test and test this level of detail by either:

- splitting out the logic for choosing a version field into its own function and unit testing it (have a look for `*.spec.js` files in `services/` for examples) or
- using mocked tests for each detail case (have a look for `*.tester.js` that include `.intercept` files in `services/` for examples)

If we pick a project that uses the `version-semver` field and validate with `isSemver` for the integration test, that should be stable enough.

services/vcpkg/vcpkg-version.service.js
njakob: That's a good idea to directly link the documentation! Add this in  https://github.com/badges/shields/pull/8945/commits/79d48e64e05ca0e70a8a2ac3b305b54178b5989e

services/vcpkg/vcpkg-version.tester.js
njakob: I went for the second approach and mocked the other test cases for the different version fields in https://github.com/badges/shields/pull/8945/commits/23a169acbe210b80c06a5f62945b7d1707a60f45. Let me know if this makes sense!

njakob:
I'm not completely sure why some checks are failing after the latest changes as it seems unrelated. So you have an idea @chris48s?

chris48s:
The docker and package test failures were transient. I restarted them. The service test failures are legit. Are you able to see the summary report on https://github.com/badges/shields/actions/runs/4286928999 ?

"Mocks not yet satisfied" normally means there is an issue with the mock setup and the outbound API request we're making is not matching any mocked request.

You should see the same failures locally if you run `npm run test:services -- --only="vcpkg"`

services/vcpkg/vcpkg-version.tester.js
njakob: Since I ran into some issues mocking these tests, I went for the approach of extracting the version parsing helper.

The remaining live test is checking a port that currently exposes its version through the `version-semver` as you suggested.

services/vcpkg/vcpkg-version-helpers.js
```diff
@@ -0,0 +1,17 @@
+import { InvalidResponse } from '../index.js'
+
+export function parseVersionFromVcpkgManifest(manifest) {
+  if (manifest['version-date']) {
```
njakob, line 4: **thought**: It would have been better to also validate the schema directly inside the helper. However, this would require working around the current validation API in place through the service class and directly using [`validate`](https://github.com/njakob/shields/blob/23c0406bedfc6930735e8f5ea75dfe34faf1f290/core/base-service/validate.js#L5) or the Joi schema validation.

However, looking at other usages of `fetchJsonFromRepo`, I left the schema validation outside of this parsing helper.

njakob:
I went into a slightly different direction of unit testing the parsing helper @chris48s, now all the (new) tests are passing!

chris48s (approved):

PR #8946, opened 2023-02-26
misc minor fixes to [githubsize node pypi]

chris48s:
This is a random grab bag of things I've spotted as I've been plugging away at the docusaurus site.

1. I found a couple of badges with errors in the route, which I have fixed.
2. On the examples for the node badges, we've framed the route as `/@:scope/:packageName` rather than `/:scope/:packageName` (which is what we do with all the other npm badges). The change I've made is backwards-compatible. Any existing badges will include the @ in the route - it just changes how it is presented in the frontend. This brings the way we present it into line with the other NPM and package.json badges.
3. I spotted some more HTML that is giving us issues in a similar vein to https://github.com/badges/shields/pull/8887 which I've fixed


chris48s:
some sonar tests were failing, but unrelated to these changes

calebcartwright (approved):

PR #8948, opened 2023-02-27
adjust route for [npmsio] badges

chris48s:
This is another "get everything to play nice with openapi" PR.
In openapi, a route "segment" (the bit in-between 2 slashes basically) can either be a constant value or a variable parameter, but not a mix of both.
Shields almost universally sticks to this (its a fairly sensible rule), except for this one badge. Path-to-regexp lets us do this, but it doesn't translate well. This PR adjusts the npms-io route and adds a redirect for compatibility.

### DIFF
--- a/services/npms-io/npms-io-score.service.js
+++ b/services/npms-io/npms-io-score.service.js
@@ -23,7 +23,7 @@ export default class NpmsIOScore extends BaseJsonService {
   static route = {
     base: 'npms-io',
     pattern:
-      ':type(final|maintenance|popularity|quality)-score/:scope(@.+)?/:packageName',
+      'score/:type(final|maintenance|popularity|quality)/:scope(@.+)?/:packageName',
   }
 
   static examples = [
--- a/services/npms-io/npms-io-score-redirect.service.js
+++ b/services/npms-io/npms-io-score-redirect.service.js
@@ -1,0 +1,27 @@
+import { redirector } from '../index.js'
+
+export default [
+  redirector({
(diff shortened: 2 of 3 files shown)


services/npms-io/npms-io-score.service.js
```diff
@@ -23,7 +23,7 @@ export default class NpmsIOScore extends BaseJsonService {
   static route = {
     base: 'npms-io',
     pattern:
-      ':type(final|maintenance|popularity|quality)-score/:scope(@.+)?/:packageName',
+      'score/:type(final|maintenance|popularity|quality)/:scope(@.+)?/:packageName',
```
calebcartwright, line 26: Suppose this becomes a question of end user experience on the badge builder vs. processing of a request, but did you consider keeping the route the same and add `-score` duplicatively on each type variant? Recognize that'd be a bit verbose inline here and add a dash of noise in the modal, but it would obviate the need for a redirect, and arguably adheres closer to that single-segment noun under our badge route guidelines

calebcartwright (approved):
Don't feel too strongly about this one so will go ahead and approve if you'd prefer to continue with this route. Have added inline question below about an alternative option for consideration/brief discussion too though

services/npms-io/npms-io-score.service.js
chris48s: Weirdly, I actually hadn't thought of that, although it seems obvious now. In the interests of not introducing more redirects where we don't need them, I think I do prefer that option.
I've submitted https://github.com/badges/shields/pull/8967 which implements that suggestion.

PR #8966, opened 2023-03-06
Export OpenAPI definitions from service examples; affects [dynamic endpoint static]

chris48s:
Refs #7982

This PR changes lots of things, and also nothing 🙃

One of the things we need to do in order to migrate the frontend to docusaurus-openapi is to produce an OpenAPI spec for each service/endpoint we expose.

This PR does a few main things:

- Allows service classes to optionally define an `openApi` property directly
- Adds some code that makes a decent (but imperfect in some cases) attempt to derive one from the existing `examples` array
- Adds a script which we can use to export OpenAPI definitions that can then be consumed by a frontend that can read them
- Adds OpenAPI definitions to a handful of services. I've started off by adding these to the dynamic badges, endpoint badge and static badges because in the new frontend I want to document these the same as any other endpoint but in the existing frontend they have custom "documentation" (if you can call it that) and no `examples`, so we can do this without having any impact on the existing site.

In the interests of keeping changesets reasonably small and easy-ish to review, I'm not doing anything about adding that frontend here - just generating the OpenAPI specs. Also, merging this PR will not change anything about the existing frontend code. Merging this would give us the next piece of the jigsaw to start working on this but we can deploy this without affecting anything.

Next steps from here (may not be exhaustive) are something like:

- work on switching out the existing gatsby-based frontend for docusaurus-openapi using the best-effort OpenAPI definitions generated from examples[]
- once we switch the frontends, (auto-)convert all the existing `examples` arrays on service classes to `openApi` objects
- look for ways we can reduce the duplication between the service definition and OpenAPI spec
- make `openApi` required/validated
- delete all `examples` properties + remove any validation for it
- tweak and improve from there (e.g: fixing required query params, removing unnecessary duplicates for query param variants, adding param descriptions) - this could be a gradual process
- delete all the code for bludgeoning `examples` arrays into an OpenAPI spec (this will clean things up a lot)
- update contributing docs for adding new services to reflect new frontend

The commits in this PR may not be especially helpful. I kind of meandered about a bit. It might be best to just look at the whole diff to review.

Hopefully this all makes sense as a direction.

~There are a couple of failing service tests which are not related to this PR. They are also failing on the daily-tests repo. I'm going to ignore them for now.~


scripts/export-openapi-cli.js
```diff
@@ -0,0 +1,349 @@
...
+          description:
+            'Specify what clicking on the left/right of a badge should do. Note that this only works when integrating your badge in an `<object>` HTML tag, but not an `<img>` tag or a markup language.',
+          style: 'form',
+          explode: true,
+          schema: {
+            type: 'array',
+            maxItems: 2,
+            items: {
+              type: 'string',
+            },
```
chris48s, line 298: At the moment, this doesn't actually render right in docusaurus-openapi. I've got a PR open over at
https://github.com/cloud-annotations/docusaurus-openapi/pull/241
but it hasn't been reviewed yet

services/static-badge/static-badge.service.js
```diff
@@ -3,13 +3,62 @@ import { BaseStaticService } from '../index.js'
...
   static route = {
     base: '',
     format: '(?::|badge/)((?:[^-]|--)*?)-?((?:[^-]|--)*)-((?:[^-.]|--)+)',
     capture: ['label', 'message', 'color'],
   }
 
+  static examples = [
+    {
+      title: 'Static Badge',
+      namedParams: { badgeContent: 'build-passing-brightgreen' },
```
chris48s, line 15: This param can't be called `:label-message-color`.
I went for `:badgeContent`, but it could be something else. `:labelMessageColor`? `:allTheThings`?

scripts/export-openapi-cli.js
```diff
@@ -0,0 +1,349 @@
...
+    outParam.required = true
+  } else {
+    /*
+    Occasionally we do have required query params,
+    but at the moment we can't pick this up.
+
+    TODO: Add something to the example format which allows
+    us to populate this based on the queryParamSchema.
+    */
+    outParam.required = false
```
chris48s, line 73: There's some stuff we just can't really populate from the existing data structure. Rather than try to extend it, move to allowing services to declare openapi path directly.

core/base-service/base.js
```diff
@@ -140,6 +140,15 @@ class BaseService {
...
   static examples = []
 
+  /**
+   * Optional: an OpenAPI Paths Object describing this service's
+   * route or routes in OpenAPI format.
+   *
+   * @see https://swagger.io/specification/#paths-object
+   * @abstract
+   */
+  static openApi = undefined
```
chris48s, line 150: TODO: Moar/better docs, typedef

core/base-service/service-definitions.js
```diff
@@ -46,6 +46,28 @@ const serviceDefinition = Joi.object({
       })
     )
     .default([]),
+  openApi: Joi.object().pattern(
```
chris48s, line 49: This is not indented to be a full validation of every allowable case in the spec - just a subset that is relevant for shields.

This might need a bit of tweaking as we add more cases

services/dynamic/dynamic-xml.service.js
```diff
@@ -15,6 +15,53 @@ export default class DynamicXml extends BaseService {
...
+          The Dynamic XML Badge allows you to extract an arbitrary value from any
+          XML Document using an XPath selector and show it on a badge.
+        </p>`,
+        parameters: [
+          {
+            name: 'url',
+            description: 'The URL to a XML document',
+            in: 'query',
+            required: true,
+            schema: { type: 'string' },
```
chris48s, line 32: In most cases, this is duplicating information which exists in `queryParamSchema`. It would be nice to infer one from the other, but I think I'm not going to get bogged down in that just yet.

core/base-service/openapi.js
```diff
@@ -0,0 +1,335 @@
...
+      }
+    } else {
+      // ...otherwise do our best to build one from examples[]
+      for (const [key, value] of Object.entries(
+        examples2openapi(service.examples)
+      )) {
+        // allow conflicting routes for legacy examples
+        paths[key] = value
+      }
+    }
```
chris48s, line 216: Once we get to a stage where every service defines an `openApi` property and get rid of `examples`, we can ditch this path, which in turn allows us to ditch most of the messier logic in this file.

scripts/export-openapi-cli.js
```diff
@@ -0,0 +1,49 @@
...
+      service => service.category === 'static' && !service.isDeprecated
+    )
+  )
+  coreServices = coreServices.concat(
+    definitions.services.filter(
+      service => service.category === 'dynamic' && !service.isDeprecated
+    )
+  )
+  writeSpec(
+    path.join(specsPath, '1core.yaml'),
```
chris48s, line 46: Bit of a hack: I've called this `1core.yaml` so it will sort as the first filename alphabetically.

scripts/export-openapi-cli.js
```diff
@@ -0,0 +1,349 @@
+import fs from 'fs'
+import path from 'path'
+import yaml from 'js-yaml'
+import { collectDefinitions } from '../core/base-service/loader.js'
+
+const baseUrl = process.env.BASE_URL || 'https://img.shields.io'
+const specsPath = path.join('frontend', 'categories')
```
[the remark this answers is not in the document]
chris48s: Having started the next PR, this is the right place.

chris48s:
Merged in master. Service tests now passing

scripts/export-openapi-cli.js
calebcartwright, line 1: My first reaction when browsing through this (albeit as part of a commit-by-commit review) is that it's not overtly obvious what this file is doing and why. I've got enough context from the PR and past conversations to be able to put it together, but if I put my pretending-to-be-a-new-future-contributor hat on I could see this being unclear and somewhat magical. 

Do you think it would be worth adding an inline comment (perhaps at the top of the file) summarizing the what+why, or potentially even jsdoc commentary on the individual functions? Note this isn't a blocker in my mind, and could easily be done as a follow up

scripts/export-openapi-cli.js
```diff
@@ -0,0 +1,349 @@
...
+  const re = new RegExp(`${paramName}\\(([A-Za-z0-9_\\-|]+)\\)`)
+  const match = pattern.match(re)
+  if (match === null) {
+    return undefined
+  }
+  if (!match[1].includes('|')) {
+    return undefined
+  }
+  return match[1].split('|')
+}
```
calebcartwright: Will readily confess my brain quickly bailed on trying to mentally process these. I know these function as expected based on the work you've done so far, but overall I feel like we've got quite a bit of logic in this file and I'm wondering if, also likely in a follow up, it'd be worth trying to extract out some of the logic functions and incorporate tests, or if we're sufficiently confident that if/when an issue ever pops up that it'll be easy enough to troubleshoot and resolve?

scripts/export-openapi-cli.js
calebcartwright: Could you elaborate on this for me? I'm not confident I understand the logic and comments of the if/else block and the conditionality of `required`

scripts/export-openapi-cli.js
```diff
@@ -0,0 +1,349 @@
...
+            type: 'string',
+          },
+          example: 'flat',
+        },
+        logo: {
+          name: 'logo',
+          in: 'query',
+          required: false,
+          description:
+            'Insert one of the named logos (bitcoin, dependabot, gitlab, npm, paypal, serverfault, stackexchange, superuser, telegram, travis) or <a href="https://simpleicons.org/" rel="noopener noreferrer" target="_blank">simple-icons</a>. Simple-icons are referenced using icon slugs which can be found on the simple-icons site or in the <a href="https://github.com/simple-icons/simple-icons/blob/develop/slugs.md" rel="noopener noreferrer" target="_blank">slugs.md</a> file in the simple-icons repository.',
```
calebcartwright: Perhaps we should go ahead and change this to use the SI master branch too given #9054

```suggestion
            'Insert one of the named logos (bitcoin, dependabot, gitlab, npm, paypal, serverfault, stackexchange, superuser, telegram, travis) or <a href="https://simpleicons.org/" rel="noopener noreferrer" target="_blank">simple-icons</a>. Simple-icons are referenced using icon slugs which can be found on the simple-icons site or in the <a href="https://github.com/simple-icons/simple-icons/blob/master/slugs.md" rel="noopener noreferrer" target="_blank">slugs.md</a> file in the simple-icons repository.',
```

scripts/export-openapi-cli.js
```diff
@@ -0,0 +1,349 @@
...
+            type: 'string',
+          },
+          example: 'healthiness',
+        },
+        labelColor: {
+          name: 'labelColor',
+          in: 'query',
+          required: false,
+          description:
+            'Set background of the left part (hex, rgb, rgba, hsl, hsla and css named colors supported). The legacy name "colorA" is also supported.',
```
calebcartwright: Potentially worth dropping the explicit mention of `colorA` at this point as part of the frontend overhaul?

scripts/export-openapi-cli.js
```diff
@@ -0,0 +1,349 @@
...
+            type: 'string',
+          },
+          example: 'abcdef',
+        },
+        color: {
+          name: 'color',
+          in: 'query',
+          required: false,
+          description:
+            'Set background of the right part (hex, rgb, rgba, hsl, hsla and css named colors supported). The legacy name "colorB" is also supported.',
```
calebcartwright: Same food for thought as the label color comment

services/static-badge/static-badge.service.js
calebcartwright: I think `badgeContent` is fine, `labelMessageColor` or `badgeComponents` would also be fine with me

services/dynamic/dynamic-yaml.service.js
```diff
@@ -6,29 +6,53 @@ import jsonPath from './json-path.js'
...
-        <ul>
-          <li><code>url</code> is the URL to a YAML document</li>
-          <li><code>query</code> is a <a href="https://jsonpath.com/">JSONPath</a> expression that will be used to query the document</li>
-        </ul>
-        Also an optional <code>prefix</code> and <code>suffix</code> may be supplied.
-      </p>`,
-      staticPreview: {
-        message: '2',
+  static openApi = {
+    '/badge/dynamic/yaml': {
```
calebcartwright, line 10: Probably not worth it, at least not as part of this PR, but intuition is that we should either try to extract this from the `route` variable that's assigned a value above at the `createRoute` call site, or perhaps we should just make the route explicit inline here; It just gives me pause that we've got a hardcoded value here two lines down from where this is set dynamically in a function call, even though I know we're not changing the actual route values ever

core/base-service/openapi.js
calebcartwright: Some of the subsequent commits and comments are starting to obviate some of the comments I left on the earlier commits, so glad to see we're thinking about this in similar ways :+1:

scripts/export-openapi-cli.js
calebcartwright: Case in point, I see this comment was a bit premature given how things evolved over subsequent commits

scripts/export-openapi-cli.js
calebcartwright: Ah the bane of commit-by-commit reviews :laughing:

calebcartwright (dismissed):
> The commits in this PR may not be especially helpful. I kind of meandered about a bit. It might be best to just look at the whole diff to review.

I remember reading this back in March, but I'd definitely forgotten it by the time I got around to reviewing. I left a few inline comments/questions/suggestions, several of which were unnecessary and stemmed from reviewing by commit so I closed them out. A couple of other minor thought comments were left open, but none are blocking, especially given how long this has been open and how it is a blocker itself to other work.

The only concern I have is that it feels like there's a lot of what I'll call "domain specifics" around OpenAPI that are definitely unfamiliar to me, and some non-trivial logic in place to reshape some of our existing structures into those domain specifics. I don't think it's great to only have one person understand those well, but I also don't think that's an adequate enough justification to not do anything.

The frontend needs a refresh, that "domain specifics" problem would exist with any solution, and this solution seems as reasonable as any a path forward

scripts/export-openapi-cli.js
chris48s: I know you resolved this comment, but I think it is worth explaining this one in more detail.

Basically, in our existing `examples` array that we attach to service classes, we don't say if a param is required or optional.
For path (or route) params, this doesn't matter. All path params are required in OpenAPI.
For query params, this does matter because in OpenAPI they can be required or optional, and we do have a few services where there are query params that are required at the schema level.
For the moment I'm just setting them all to optional, which is right in most cases but there will be a handful of cases where we are documenting a required query param as optional in the frontend but then requiring it at the schema level. These will need improving later, but the way I want to do this is by moving to declaring an `openApi` spec object directly on the service classes, rather than extending the existing `examples` object (which already requires some bashing with a hammer to get it into the right format).

services/dynamic/dynamic-yaml.service.js
chris48s: Agreed. This comes a few steps further on in the migration plan I laid out.
This is exactly the sort of thing I have in mind when I say "look for ways we can reduce the duplication between the service definition and OpenAPI spec" in the top post - there is definitely information we are specifying twice.
I think I want to have a slightly more holistic view of it all first to arrive at the right abstractions.

chris48s:
> I remember reading this back in March, but I'd definitely forgotten it by the time I got around to reviewing. I left a few inline comments/questions/suggestions, several of which were unnecessary and stemmed from reviewing by commit so I closed them out

Yes. I think there are a couple of challenges with this PR (and the next one). One of them is that it was one of those PRs where I had a vague idea where I was going, but didn't really know _exactly_ what the endgame looked like until I was into the problem (hence the commits kind of wandered around a bit).

The second one is that there is a tension between

- Attempt to arrive at a perfect solution in one go and boil the ocean in enormous PRs that change everything all at once
- Tackle the problem in smaller chunks (which hopefully makes it easier to review/make changes incrementally), at the expense of introducing some sub-optimal or non-obvious intermediate steps

I'm trying to tackle this in smaller chunks, but the tradeoff of that there are clear improvements that can be made and some of the code I've written here is stop-gap code I hope to delete in the near future. I am also trying to communicate where I envisage we go next with this. I appreciate that it is much clearer to me where we should go next with this, whereas for the person reviewing it you can only really see clearly where I've gone so far.

> The only concern I have is that it feels like there's a lot of what I'll call "domain specifics" around OpenAPI that are definitely unfamiliar to me

Yes. This is a good point :+1:  Once we drop the `examples` array and every service class exports an `openApi` object, two of the things we really need to do are:

a) More clearly document the subset of OpenAPI that we (and contributors) need to declare on a service class and how that subset specifically applies to shields (this is what I'm getting at with the point "update contributing docs for adding new services to reflect new frontend")
b) Implement some automated checks (be that Joi schema, CI checks, whatever) that help to prevent common issues/make it easy to "fall into the pit of success"

While we are still declaring an examples array on the service class for 99% of services and haven't quite nailed down how much of the openapi bit needs to be handwritten vs generated, we're not quite in that place yet.

I think this will also get a lot easier once there is a decent body of examples to crib from too.

In general, I think OpenAPI is probably a good direction of travel though. If you pick a random developer, there is more chance they've some familiarity reading or writing an OpenAPI spec (a widely used standard) than a shields `examples` array (a thing we made up).

services/static-badge/static-badge.service.js
chris48s: I reckon let stick with `badgeContent`

Issue #8026, opened 2022-06-01
Keeps on showing 'invalid response data' for VS Code theme extension


sjns19:
### Are you experiencing an issue with...

My own instance of shields

### 🐞 Description

I really have no idea why it's happening but this is very strange.

So, I have two themes for VS Code:

https://marketplace.visualstudio.com/items?itemName=SujanShrestha.everest-night-theme

https://marketplace.visualstudio.com/items?itemName=SujanShrestha.chrome-console-dark

The latter one was published just today and the rating and downloads badges for it keep on giving the invalid response data error. What's strange is that, it seems to be working sometimes but not always. I republished the theme like 10 times but nothing happened so I let it like that thinking it must have been some cache related issue and it's still updating. 

After an hour, it did show the downloads as 2 and "no ratings" text but now the same errors are back. Another strange thing is, it's showing perfectly fine on its github repo but not on the VS Code Markeplace.

The way the package.json for both of the themes are structured, is exactly the same. Only differences are the names and versions but the badges work fine for the other theme but not for the "Chrome Console Dark" one. 

I really have no idea if it's the issue from my side or the shields itself so I opened this issue.

### 🔗 Link to the badge

 https://img.shields.io/visual-studio-marketplace/d/SujanShrestha.chrome-console-dark?color=%23f19767&style=for-the-badge

https://img.shields.io/visual-studio-marketplace/stars/SujanShrestha.chrome-console-dark?color=%23f19767&style=for-the-badge

### 💡 Possible Solution

No idea...

calebcartwright:
Could you please update the issue description with all badge links, and the embedded badges so that they will render inline?

Additionally, since you say the behavior is intermittent, it would be helpful if you could capture a screenshot of the badge for reference. All we can do is speculate until we're able to reproduce the issue, but as to some of the other items:

The "invalid response data" means that the upstream API (the Marketplace in this case) returned an unexpected HTTP response structure that diverges from the typical, expected response. When this happens consistently, it's typically due to the upstream provider changing their API contract (such as removing a field we were using). If it's an intermittent occurrence though then it's possible the upstream API is having some issues and returning some kind of error-ish response.

It's also possible that there's a temporary issue that occurs shortly after publishing an updated version to the marketplace.

[...]

PR #8985, opened 2023-03-11
handle missing statistics array in [VisualStudioMarketplace] badges

chris48s:
refs https://github.com/badges/shields/issues/8026


calebcartwright (approved):
lgtm

PR #9009, opened 2023-03-17
push images to both GHCR and DockerHub (2)

chris48s:
https://github.com/badges/shields/pull/9008 did not work first time :smile:


chris48s:
second time lucky
https://github.com/badges/shields/actions/runs/4451364940/jobs/7817929982

Issue #3654, opened 2019-07-06
Review Shields descriptions


PyvesB:
We describe Shields in different ways in different places:
* on GitHub:
  > Concise, consistent, and legible badges in SVG and raster format
* on Twitter:
  > Quality metadata badges for open source projects.
* on OpenCollective:
  > Concise, consistent, and legible badges… in a readme near you
* ...

Is there room for improvement or making things more consistent? Or are we happy with the current state of these headlines? Either way, I think there's value in having an informal discussion about how we communicate about what Shields is.

paulmelnikow:
It'd be a great idea to land on a single tagline and use it everywhere.

Two more we use:

- the website title:
  > Quality metadata badges for open source projects
- the website feature top line:
  > Pixel-perfect,   Retina-ready,   Fast,   Consistent,   Hackable,   No tracking

I thought I'd start things off by mentioning some of the things that I think are important about Shields. 

1. We make readmes look consistent
2. Shields is easy to use. Our URLs are friendly, and we provide good documentation and great error messages.
3. Shields is a community project. We have 300+ contributors, 2k forks, and merge a lot of PRs. We have a good development environment that makes it easy to add badges for new services. We make an environment that's friendly for people to ask questions and make contributions.
4. It's really powerful. You can make your own static badges and make your own badges using custom endpoints.
5. We don't track you or your users.
6. It's fast and reliable. We have tests of all our badges and the daily tests catch problems before users report them.
7. While it's a service to the open-source community, it's also startup- and corporate-friendly. You can host your own instances that have secure access to private repos, and you can run it on a private network.

The tagline should be way shorter than that, though maybe it draws from some of these things!

PyvesB:
Suggestion:

> Metadata taken to the next level with quality badges generated by a powerful open-source project

paulmelnikow:
I wonder, instead of "metadata" if we could use a phrase like "status" or "status at a glance."

PyvesB:
I find the status/metadata words somewhat restrictive. Badges such as Twitter followers or Wheelmap don't naturally fall under these. Maybe:

> Data at a glance with quality badges generated by a powerful open-source project.

Or:

> Summarize the world with quality badges generated by a powerful open-source project.

paulmelnikow:
> Data at a glance

I like this ^^

It's nice not to describe it in a way that's overly restrictive, though I think we should probably convey that it's for developers.

Some ideas:

> Community-powered informational badges for software developers

[...]

PR #9014, opened 2023-03-20
migrate frontend to docusaurus

chris48s:
Refs https://github.com/badges/shields/issues/7982

This PR is now ready for review.

As with https://github.com/badges/shields/pull/8966 the commit history for this isn't very useful. Sorry. If it is just too much of a mess, let me know and I will try to do a massive rebase/squash and clean it up a bit.

I've set up a review app at https://pr-9014-badges-shields.fly.dev/ which won't delete itself for the moment because we haven't merged https://github.com/badges/shields/pull/9069 yet.

I've been though all the comments. Anything that is resolved or hidden is now irrelevant. Anything still unresolved is still relevant. Note that I've spent so much time chatting away to myself in this PR that Github has started to hide bits of it.

Screenshot at 2023-04-15 19-52-29

Once we get this merged/deployed, I'll start writing up the next steps from this PR and #8966 into issues which we can hopefully tackle a bit more incrementally.

I'll need to remove test-frontend from the required checks in settings when we merge this.

I've attempted to give us some basically sensible content as an opening gambit, but we can always fiddle with content later. I'm mainly trying to get the tech bit right in this PR.


frontend/src/components/homepage-features.js
```diff
@@ -0,0 +1,92 @@
...
+  {
+    title: 'Host your own instance',
+    description: (
+      <>
+        Host a shields instance behind your firewall with our{' '}
+        <a href="https://registry.hub.docker.com/r/shieldsio/shields/">
+          docker image
+        </a>
+        <br />
+        <code>docker pull shieldsio/shields</code>
```
chris48s, line 53: This will not age well.

package.json
```diff
@@ -97,15 +95,13 @@
...
+    "pretest": "run-s --silent defs",
+    "test": "run-s --silent --continue-on-error lint test:package test:core test:entrypoint check-types:package prettier:check",
     "check-types:package": "tsd badge-maker",
-    "check-types:frontend": "tsc --noEmit --project .",
     "depcheck": "check-node-version --node \">= 16.0\"",
     "prebuild": "run-s --silent depcheck",
-    "features": "node scripts/export-supported-features-cli.js > ./frontend/supported-features.json",
-    "defs": "node scripts/export-service-definitions-cli.js > ./frontend/service-definitions.yml",
-    "build": "run-s defs features && cd frontend && gatsby build && mv public ..",
+    "defs": "echo 'TODO: https://github.com/badges/shields/pull/8966'",
```
chris48s, line 103: Blocked on https://github.com/badges/shields/pull/8966

core/server/server.js
```diff
@@ -412,7 +412,7 @@ class Server {
 
     if (rasterUrl) {
       // Redirect to the raster server for raster versions of modern badges.
-      camp.route(/\.png$/, (queryParams, match, end, ask) => {
+      camp.route(/^\/((?!img\/)).*\.png$/, (queryParams, match, end, ask) => {
```
chris48s, line 415: This is to allow `/img/logo.png` to work without trying to redirect to squint.
This does bring up a slightly wider question though: The way I've done this, I have put the frontend in the root so the compiled website structure (`assets`, `badges`, `community`, `search`, etc) lives in the same namespace as the badge URLs. This is pretty much what we did with gatsby too but because every badge has its own URL for the builder, we could consider putting the frontend in `/docs` or whatever and make `/` a redirect to `/docs`. I think we should really decide this now because once we launch this version of the site there will be a bunch of links out there in the wild which will break if we change our minds. Thoughts?

package.json
```diff
@@ -71,22 +71,20 @@
...
+    "coverage:test": "run-s --silent --continue-on-error coverage:test:core coverage:test:package coverage:test:entrypoint coverage:test:integration",
     "coverage:report:generate": "c8 report",
     "coverage:report:open": "open-cli coverage/lcov-report/index.html",
     "coverage:report": "run-s --silent coverage:report:generate coverage:report:open",
     "lint": "eslint \"**/*.@(js|ts|tsx)\"",
     "prettier": "prettier --write \"**/*.@(js|ts|tsx|md|json|yml)\"",
     "prettier:check": "prettier --check \"**/*.@(js|ts|tsx|md|json|yml)\"",
     "danger": "danger",
-    "test:frontend": "cross-env NODE_ENV=test ts-mocha --config .mocharc-frontend.yml \"frontend/**/*.spec.@(js|ts|tsx)\"",
     "test:e2e": "cypress run",
```
chris48s, line 88: TODO: write some new e2e tests that actually pass

package.json
```diff
@@ -243,7 +242,14 @@
...
     "typescript": "^5.0.2",
-    "url": "^0.11.0"
+    "url": "^0.11.0",
+
+    "@docusaurus/core": "^2.0.0",
+    "@easyops-cn/docusaurus-search-local": "^0.33.6",
+    "@mdx-js/react": "^1.6.21",
+    "clsx": "^1.1.1",
+    "docusaurus-preset-openapi": "0.6.4",
+    "prism-react-renderer": "^1.2.1"
```
chris48s, line 252: I've not done it yet, but one of the bits of this I am looking forward to most is really going to town deleting all the dependencies for gatsby, typescript, etc. I'm going to leave this until the end though just to minimise the need to solve conflicts. Lets get to a stage where we're happy with the code changes first. Then I'll do that and clean this up. I also need to go through the prod dependencies - there's some stuff like `@fontsource/lato` that wants deleting from there too.

package.json
```diff
@@ -97,15 +95,13 @@
     "test:services:pr:prepare": "node core/service-test-runner/pull-request-services-cli.js > pull-request-services.log",
     "test:services:pr:run": "cross-env NODE_CONFIG_ENV=test mocha core/service-test-runner/cli.js --stdin < pull-request-services.log",
     "test:services:pr": "run-s --silent test:services:pr:prepare test:services:pr:run",
-    "pretest": "run-s --silent defs features",
-    "test": "run-s --silent --continue-on-error lint test:frontend test:package test:core test:entrypoint check-types:package check-types:frontend prettier:check",
+    "pretest": "run-s --silent defs",
+    "test": "run-s --silent --continue-on-error lint test:package test:core test:entrypoint check-types:package prettier:check",
     "check-types:package": "tsd badge-maker",
-    "check-types:frontend": "tsc --noEmit --project .",
```
chris48s, line 103: Given how little of our own frontend code we are going to be left with, I think the value of typescript or unit tests for the frontend is limited. I think I'm going to write a few e2e tests and leave it at that.

frontend/docusaurus.config.cjs
chris48s, line 1: Docusaurus still doesn't support the config file being an ESModule so there's a few files of CommonJS here and `/frontend` still has its own special `package.json` to allow it to be a CJS module

package.json
```diff
@@ -114,13 +110,16 @@
     "debug:server": "cross-env NODE_CONFIG_ENV=development nodemon --inspect server.js 8080",
     "profile:server": "cross-env NODE_CONFIG_ENV=development node --prof server 8080",
     "benchmark:badge": "cross-env NODE_CONFIG_ENV=test node scripts/benchmark-performance.js --iterations 10100 | node scripts/capture-timings.js --warmup-iterations 100",
-    "prestart": "run-s --silent depcheck defs features",
-    "start": "concurrently --names server,frontend \"npm run start:server\" \"cd frontend && cross-env GATSBY_BASE_URL=http://localhost:8080 gatsby develop --port 3000\"",
+    "prestart": "run-s --silent depcheck defs",
+    "start": "concurrently --names server,frontend \"npm run start:server\" \"npm run docusaurus:start\"",
```
chris48s, line 114: When you run the built version of the docs (i.e: `npm run build` and serve it from scoutcamp), the website performs really well because it is just serving static HTML files.

The dev server (with hot reload when you change the files and all that jazz) is a bit of a resource hog, especially the first time you start it. It does seem to get a bit better once docusaurus has cached some stuff, but your first experience of running `npm start` is likely to be a bit sluggish. I think the fact we have over 600 pages doesn't help but in general this is an acknowledged issue: https://github.com/facebook/docusaurus/issues/4765 It is not great dev experience, but I can live with it. Its not an issue that will affect production. This was kinda true of gatsby too, but docusaurus is way worse. If it turns into a big pain point, we could consider having 2 variants of the dev server:
- A "I want to work on the website" command (that runs `docusaurus start` so you get hot reloads for the frontend) and
- A "I want to work on the badge server" command (that runs `npm run build` and serves it from scoutcamp)

For now I am going to ignore this issue.

chris48s:
These lists probably are not complete. Also needs a general sweep. Grep for `frontend`, `gatsby`, etc

TODOs that need to happen in this PR:

- Inline comments once https://github.com/badges/shields/pull/8966 is in
- E2E tests
- Update docs
- Review ESLint config again. Any config specific to the old frontend we can ditch?


TODOs that I will punt to later:

- Flag URL conflicts between badge server and frontend
- CI automation to warn about swizzled components when upgrading `docusaurus-theme-openapi`
- Bring back the blog
- Triage https://github.com/badges/shields/issues?q=is%3Aopen+is%3Aissue+label%3Afrontend and close most of them

cypress/e2e/main-page.cy.js
```diff
@@ -26,36 +19,38 @@ describe('Main page', function () {
     cy.contains('PyPI - License')
```
chris48s, line 19: TODO: This test should pass in CI but fails locally because search only works with a compiled frontend. Think I need to bin off the ability to run cypress against a dev build

frontend/src/components/homepage-features.js
chris48s: ef74cbb

frontend/src/components/homepage-features.js
```diff
@@ -0,0 +1,92 @@
...
+  {
+    title: 'Host your own instance',
+    description: (
+      <>
+        Host a shields instance behind your firewall with our{' '}
+        <a href="https://github.com/badges/shields/pkgs/container/shields">
+          docker image
+        </a>
+        <br />
+        <code>docker pull ghcr.io/badges/shields</code>
```
chris48s, line 53: Maybe roll back https://github.com/badges/shields/pull/9014/commits/ef74cbb26b1c24ce987a8975e60313682f9161f8

https://www.docker.com/blog/no-longer-sunsetting-the-free-team-plan/

services/github/auth/acceptor.js
```diff
@@ -3,6 +3,7 @@ import { fetch } from '../../../core/base-service/got.js'
 import log from '../../../core/server/log.js'
 
 function setRoutes({ server, authHelper, onTokenAccepted }) {
+  // TODO
   const baseUrl = process.env.GATSBY_BASE_URL || 'https://img.shields.io'
```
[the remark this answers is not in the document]
chris48s: `GATSBY_BASE_URL` was never available to the backend server anyway f1243e3

package.json
chris48s: Just thinking this through a bit more..

If we think the performance of the webpack devserver is terrible, I think changing the default behaviour of `npm start` in dev to just compile the frontend and serve it with no watch on the frontend (but keep nodemon doing hot reloads for the backend) does not actually lose us very much. I've deleted almost all of the frontend code so there is less to watch :) I think it would be useful to get another opinion on the dev experience here.

README.md
```diff
@@ -107,7 +107,7 @@ You can read a [tutorial on how to add a badge][tutorial].
 
 When server source files change, the badge server should automatically restart
 itself (using [nodemon][]). When the frontend files change, the frontend dev
-server (`gatsby dev`) should also automatically reload. However the badge
+server (`docusaurus start`) should also automatically reload. However the badge
 definitions are built only before the server first starts. To regenerate those,
 either run `npm run defs` or manually restart the server.
```
chris48s, line 112: Depending on what we decide when it comes to https://github.com/badges/shields/pull/9014#discussion_r1142384738 we might need to update this and also https://github.com/badges/shields/blob/master/doc/TUTORIAL.md?plain=1#L305

package.json
chris48s: The other check we can perform on the frontend is ensuring `npm run build` completes without errors. I haven't added n explicit CI build for this. It is covered implicitly because it happens in both the E2E tests and Dockerfile builds.

chris48s:
Socket flags a couple of possible issues to look into with dependencies introduced in this PR which are packages that use postinstall scripts. Having reviewed the scripts, I'm fine with them.

@SocketSecurity ignore core-js@3.24.1

@SocketSecurity ignore postman-code-generators@1.7.1

The 2x danger warnings are false positives.

.gitignore
```diff
@@ -104,11 +100,6 @@ frontend/categories/*.yaml
 # Template for the local runtime configuration.
 !/config/local*.template.yml
 
-# Gatsby
-/frontend/.cache
-/frontend/public
-/public
```
chris48s, line 110: If you already have a checkout of shields, these removed paths will become unstaged files in your checkout after pulling this. Other than "just live with it", is there any good pattern for dealing with this?

core/server/server.js
calebcartwright: do you have a feel for which may give us more flexibility over the long term? Also, I gather that `/docs` was just an example but given that we have an existing `doc` directory I think we'd want to use something with a more differentiated name for the generated site output

frontend/docusaurus.config.cjs
```diff
@@ -0,0 +1,124 @@
+const lightCodeTheme = require("prism-react-renderer/themes/github");
+const darkCodeTheme = require("prism-react-renderer/themes/dracula");
+
+/** @type {import('@docusaurus/types').Config} */
+const config = {
+  title: "Shields.io",
+  tagline: "Concise, consistent, and legible badges",
```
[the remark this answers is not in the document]
calebcartwright: I don't have a strong position on this but will spend some time contemplating. My initial inclination is something along the lines of...

> yeah, we probably don't need to go out of our way to push them, but, we do still provide them because they are helpful in a few contexts and it's helpful for new users to know that we _do_ offer them

though i'm trying to challenge myself to gauge whether i'm just being biased given my prior work on squint

package.json
calebcartwright: i agree with the sentiment, but want to get a feel for the scope of TS left before fully forming my opinion (i'm starting my review by perusing your remaining comments like this one, and haven't really dug in to the meat of the diff yet)

core/base-service/legacy-request-handler.js
```diff
@@ -65,7 +65,7 @@ function handleRequest(cacheHeaderConfig, handlerOptions) {
     */
     if (match[0] === '/endpoint' && Object.keys(queryParams).length === 0) {
       ask.res.statusCode = 301
-      ask.res.setHeader('Location', '/endpoint/')
+      ask.res.setHeader('Location', '/badges/endpoint-badge')
```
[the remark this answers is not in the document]
calebcartwright: may be misunderstanding but this leads me to two questions:

* does this mean we can no longer directly link to a page that enumerates badges under a specific category?
* if that's wrong (and we can still do so), does this mean we'd break any existing links? (note that's not the end of th world imo, just want to understand implications)

package.json
calebcartwright: i don't think i'll have time today but i'll pull your branch down locally and give it a test drive soon

.gitignore
calebcartwright: not that i know of. it's something i've experienced occasionally in a few projects, and always just dealt with via local cleanups (either via git or just standard file system management tooling).

i do not think this is worth the effort, especially given our current quantity of regular contributors, but the only potential automated/assistive route i can think of would be seeing if there's any git client hooks (post-merge maybe?) we could hook into (pun very much intended) that would deal with any such known files/directories for our project

core/server/server.js
chris48s: So if we followed that suggestion, the site frontend would live at https://shields.io/docs and https://shields.io/ would redirect to https://shields.io/docs

The code (or output) would not need to live in https://github.com/badges/shields/tree/master/docs in the repo though.

Basically the advantage of that is it would get all the frontend routes out of the same namespace we use for the badge routes but at the moment we don't have any conflicts to worry about.

frontend/docusaurus.config.cjs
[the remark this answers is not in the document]
chris48s: The reason I said that specifically here is we currently [sometimes](https://github.com/badges/shields/issues/3654) use the tag line "Concise, consistent, and legible badges in SVG and raster format". I've dropped "in SVG and raster format" from that. I don't think it needs to be in the first line someone reads about the project. We don't necessarily need to completely stop documenting that png exists as an option though.

I can't remember if raster made its ways into this anywhere, but let me take an action to make sure it is in there somewhere.

package.json
chris48s: This PR drops almost all typescript code and tooling. The exceptions to that are:

- badge-maker types:
    - `./badge-maker/index.d.ts`
    - `./badge-maker/index.test-d.ts`
- We are still using `@typescript-eslint/parser` as out parser for ESLint. This is because the standard ESLint parser doesn't implement support for any node js feature that is not stable yet. This includes some features we use (one of them is top-level `await`, but there are others). As well as typescript stuff, `@typescript-eslint/parser` also implements compatibility with a broader subset of node features. So that is why we're using a typescript parser, despite mostly not using typescript. Upgrading from ESLint 7 to ESLint 8 might allow us to drop `@typescript-eslint/parser` and switch to using the standard ESLint parser but I have not tried it. It is another issue for anoter day.

That's it.

This isn't necessarily some crusade against type safety. It just happens that the only place in the codebase where we were using typescript was the frontend and we're moving to an approach where we are mostly relying on packages and there is very little custom frontend code.

core/base-service/legacy-request-handler.js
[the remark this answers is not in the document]
chris48s: Yep. Every badge gets its own page in the docs now, but there are no pages for the categories. The categories are groupings in the menu but not pages in their own right.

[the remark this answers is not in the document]
calebcartwright: that's a minor setback imo. i've found it helpful to be able to give folks a link to the category pages in the past. to be clear i don't think that's a blocker, just unfortunate

package.json
calebcartwright: Understood, and there were never any implications of crusades. FWIW I'm a big fan of TypeScript, but there's a point of metaphorical "critical mass" that has to be reached in order for me to feel the benefits justify the overhead (extra tooling/dependencies, build process, etc.). I think we were/are _arguably_ past that point previously/currently (at least one could reasonably make the case we were), so I was just noting that I intended on making that same personal assessment with these proposed changes.

I don't have any predispositions one way or the other, though based on the way you've described the changes I thoroughly anticipate I'd reach the same conclusion as you in that removal of TS is the right approach, and that the overhead of keeping TS around would outweigh any benefits in the new context.

core/server/server.js
calebcartwright: I hate trying to mentally parse regexes, but last, semi-pathological question on this.. would this potentially match any cases that _should_ get redirected to squint?

e.g. a repo named `img`, perhaps like https://img.shields.io/github/v/tag/pathological-username/img/some-branch.png

calebcartwright: If that's not the case and just some bad mental regex parsing and hypothesizing then great, if we think that's an extremely fringe edge case hypothetical that's not worth the gymnastics to try to work around then I'm also good with that; just wanted to think through possibilities.

core/base-service/openapi.js
calebcartwright, line 1: Just trying to grok these changes in isolation within this file, why is it we no longer need `servers`/ a base url defined?

cypress/e2e/main-page.cy.js
```diff
@@ -26,36 +19,37 @@ describe('Main page', function () {
...
+    cy.contains('button', 'rSt').should('have.class', 'api-code-tab')
+    cy.contains('button', 'AsciiDoc').should('have.class', 'api-code-tab')
+    cy.contains('button', 'HTML').should('have.class', 'api-code-tab')
+  })
 
-    cy.get('input[name="user"]').type('badges')
-    cy.get('input[name="repo"]').type('shields')
-    cy.get('table input[name="color"]').type('orange')
+  it('Build a badge', function () {
+    visitAndWait('/badges/git-hub-issues')
```
calebcartwright, line 43: Suspect this has always been present within the docusaurus ecosystem, but first time I'm noticing the hyphen-ization on the paths. Bit unfortunate that it's splitting "GitHub" this way :thinking:, assume that's unavoidable barring widespread changes to something like all the service class names?

frontend/src/components/HomepageFeatures.module.css
calebcartwright, line 1: I tend to not like mixing in proper case nor camel case into file names in order to be friendlier on case-sensitive platforms and file systems. Is this something we have autonomy to change, or does Docusaurus or something else need this exact name?

frontend/src/theme/ApiDemoPanel/Curl/index.js
```diff
@@ -0,0 +1,300 @@
...
+    variant: 'requests',
+    options: {
+      followRedirect: true,
+      trimRequestBody: true,
+    },
+  },
+]
+const languageTheme = {
+  plain: {
+    color: 'var(--ifm-code-color)',
```
calebcartwright, line 57: I'm guessing we've pulled/copied this from somewhere, and no concerns with that, but did want to note that the specifics of the args being passed to these function calls (I presume) are a complete black box to me :shrug:

core/base-service/openapi.js
calebcartwright: this is answered below via diff in a different file

frontend/src/theme/ApiDemoPanel/Curl/index.js
```diff
@@ -0,0 +1,300 @@
...
+  https://staging.shields.io/ to generate badges with the base
+  https://staging.shields.io/
+  (and we want similar behaviour for users hosting their own instance)
+
+  When we promote to production we want https://shields.io/ and
+  https://www.shields.io/ to both generate badges with the base
+  https://img.shields.io/
+
+  For local dev, we can deal with setting the api and front-end
+  being on different ports using the BASE_URL env var
```
calebcartwright, line 123: I know we already employ this so consider this a tangential future topic, but with our current hosting setup, I wonder if we've got access to do the www -> non-www redirect at some proxy/traffic manager level (e.g. if Fly has something like nginx or haproxy in there in a way that we could configure/use to add the redirect

frontend/src/theme/ApiDemoPanel/Curl/index.js
```diff
@@ -0,0 +1,300 @@
...
+
+function getServer() {
+  return {
+    url: getBaseUrl(),
+    variables: {},
+  }
+}
+
+function Curl({ postman, codeSamples }) {
+  // TODO: match theme for vscode.
```
calebcartwright, line 143: Is this a real todo for us or a copy pasta from somewhere else?

calebcartwright:
Have finally done a full pass through this and think it's in good shape. I don't have any blocking concerns nor issues, just a few thoughts and questions left inline.

There's only two things I'd like to give a spin/glance before moving ahead. I anticipate having time to do both this week, so I'd suggest we leave the remainder of the week for that, and then once that's done or if Sunday comes, I'd be good with resolving the conflicts and moving ahead.

Two things I'd still like to do:

* Seeing the frontend live with these changes (the review app has been down for me ever since I first looked at this PR. that being said, I should be able to rebuild the app or fix that myself)
* Pulling things down locally and launching the server to get a feel for the inner dev loop experience discussed above.

Finally, thanks again for taking this on. It's been a long standing desired and need, and was obviously a major, time consuming haul.

core/server/server.js
chris48s: Yeah good question.

`img` would have to be the first "clause", so it would clash if we tried to add a _service_ called `img`, but it is fine if there is a package or repo called `img`, as in your example.

The first part of the regex is:

- `^` - beginning of the string
- `\/` - escaped slash character
- `((?!img\/))` - anything other than `img`

cypress/e2e/main-page.cy.js
chris48s: Yes I agree this is annoying.

I did have a look to see if there is a way we can customise the slug.

The slugs for these pages are generated by docusaurus-plugin-openapi here https://github.com/cloud-annotations/docusaurus-openapi/blob/0beb11b248e34e2b88f171406fc4fb04a390f065/packages/docusaurus-plugin-openapi/src/openapi/openapi.ts#L105-L114

They're derived by slugging the endpoint title through [lodash.kebabCase](https://lodash.com/docs#kebabCase).

So there would be no way to display "GitHub Issues" on the page but have it slug to `github-issues`.

Although this is suboptimal, I think we just roll with it.

frontend/src/components/HomepageFeatures.module.css
chris48s: Nope. That is change-able. Done in https://github.com/badges/shields/pull/9014/commits/b2ff593a75d3918cf7814087fccb08ad4980263e

frontend/src/theme/ApiDemoPanel/Curl/index.js
chris48s: Yep. So this is one of the components from `docusaurus-theme-openapi` I have customised.

Its basically a copy of
https://github.com/cloud-annotations/docusaurus-openapi/blob/0beb11b248e34e2b88f171406fc4fb04a390f065/packages/docusaurus-theme-openapi/src/theme/ApiDemoPanel/Curl/index.tsx with some customisations.

I think this is mostly covered in
https://github.com/badges/shields/pull/9014#discussion_r1142363731

frontend/src/theme/ApiDemoPanel/Curl/index.js
chris48s: Nope it is from the upstream
https://github.com/cloud-annotations/docusaurus-openapi/blob/0beb11b248e34e2b88f171406fc4fb04a390f065/packages/docusaurus-theme-openapi/src/theme/ApiDemoPanel/Curl/index.tsx#L126

In https://github.com/badges/shields/pull/9014#discussion_r1142363731
I've touched on the topic of why I've tried to keep the changes I've made minimal in the section starting "There are 2 competing objectives here..."
Essentially: if a new version of the upstream package is released, I want it to be reasonably easy to diff our customised version against the package version to see if there are any changes we need to make (this is something I'd like to make some CI tooling for, but.. later).

frontend/src/theme/ApiDemoPanel/Curl/index.js
chris48s: I think the current answer to this is "no"  https://community.fly.io/t/how-to-redirect-from-non-www-to-www/5795

chris48s:
This review is getting pretty disjointed, so just to try and make it super-clear. Here's a list of links to the comments I've left in this pass:

- https://github.com/badges/shields/pull/9014#discussion_r1225450812
- https://github.com/badges/shields/pull/9014#discussion_r1225451196
- https://github.com/badges/shields/pull/9014#discussion_r1225451360
- https://github.com/badges/shields/pull/9014#discussion_r1225451667
- https://github.com/badges/shields/pull/9014#discussion_r1225452261
- https://github.com/badges/shields/pull/9014#discussion_r1225451909

chris48s:
> I'd like to give a spin/glance before moving ahead.

To help out with this, I've merged all this week's package bumps and resolved all the conflicts. There is a review app up at https://pr-9014-badges-shields.fly.dev/

If it has deleted itself, you can re-deploy it


Screenshot at 2023-06-10 17-32-12

I think I'd ideally like to get this merged and deployed next week.

If you're able to find a few mins to have a poke about in the review app on Sunday that would be absolutely ideal :pray:

calebcartwright:
I'm on pretty pedestrian hardware and although my first `npm start` wasn't what I'd describe as snappy, I also don't think it was all that noticeably worse than my experience with gatsby.

I don't view impact on the inner dev loop as a deal breaker, whatsoever

calebcartwright:
Not sure if this is a quirk, bug, or pebcak, but any ideas what's happening with these seemingly duplicative Azure DevOps badge listings in the sidebar?

image


As compared to what we have today:

image


I feel like I may have seen this elsewhere too but didn't register it at the time while browsing

chris48s:
> any ideas what's happening with these seemingly duplicative Azure DevOps badge listings

Yep, so this is something that we've touched on before

Definitely in https://github.com/badges/shields/issues/7982#issuecomment-1346871332
and probably somewhere in the review for https://github.com/badges/shields/pull/8966

but worth a recap.


In OpenAPI, query string params can be optional but there is no such thing as an optional path (positional) parameter, so you can't do

```js
{
  "name": "branch",
  "in": "path",
  "required": false,  // nope
  "description": "Branch",
  "schema": { "type": "string" }
}
```

This means we need to generate a separate page for every with/without optional param variant.
If we haven't already got 2 separate examples for each with/without optional route param variant, I'm creating the variants using https://github.com/badges/shields/blob/2651e5fe87d92e0b344e3f0ca41dae0b3f9a9502/core/base-service/openapi.js#L91-L119 but I'm just doubling up the existing title for now. It is not perfect, but it is workable to get this part of the migration done.

So the difference between

- https://pr-9014-badges-shields.fly.dev/badges/azure-dev-ops-builds-job
- https://pr-9014-badges-shields.fly.dev/badges/azure-dev-ops-builds-job-1

is one has a `branch` param, and one doesn't.

Also, separately from that, we've also got a bunch of places where we already have duplicated titles in the existing examples e.g:

Screenshot at 2023-06-12 10-52-45

which we also replicate on the new site.

I am very aware that having multiple menu items with the same title is unhelpful for discovery and navigation now that the only thing the user can see is the title, but one of the ways I'm trying to avoid boiling the whole ocean at once is by separating platform and content a bit. As an opening gambit to get the "platform" bit of the migration done, I'm comfortable if every bit of content is not perfect to start with.

Once we move to defining the open api definitions on the service classes directly, there's a lot of content improvements I'd like to make (once we've got the right structure in place, it will be possible to do this in lots of smaller simpler "improve docs for service X" PRs rather than mixing it in with the platform migration) and sorting out duplicate titles is going to be one of those jobs.

Maybe once we get to a point where each one has a unique title, we can add a check somewhere (maybe the service loader) to ensure all the titles are unique.

calebcartwright:
> one of the ways I'm trying to avoid boiling the whole ocean at once is by separating platform and content a bit. As an opening gambit to get the "platform" bit of the migration done, I'm comfortable if every bit of content is not perfect to start with.

Believe I follow the prior portion of the response explaining the presence of the duplicative titles, and at least AIUI, the same underlying reasoning also explains the absence of some titles (and corresponding different badges).

I think the platform -> content sequencing is a reasonable approach, although along this aspect :point_down: 

> Yep, so this is something that we've touched on before

I'm certainly having trouble maintaining a mental model of all these aspects, though admittedly the size of changes and the delta between the windows I'm actively thinking about it are both certainly aggravating factors.

Do we have anything like a high level tracking issue with a succinct enumeration of some of these content-related items we're aware of and deferring to follow up work? (and apologies if we do, my lack of awareness of such an issue speaks back to that failing mental model :brain:).

If not, do you think it would be reasonable to pull something like that together? I know you've probably got such a list in your head and I'm sure the items that would constitute such a list are covered within various Issue and PR discussions, but I feel like having things captured in one place would make them more easily consumable and trackable

chris48s:
Yep so the top post of https://github.com/badges/shields/pull/8966 starting from

> Next steps from here (may not be exhaustive) are something like:
> [bullets points]

was my last attempt to do that and is still broadly accurate. Obviously this PR is the first thing in that list.

There are probably some other things that do exist in my head or as bits where I've said "This is a bit crap - we should make it better" somewhere in a diff comment on this PR or on https://github.com/badges/shields/pull/8966 at the moment that need extracting into issues.

One of the next things I want to do after we merge this is go through https://github.com/badges/shields/issues?q=is%3Aopen+is%3Aissue+label%3Afrontend work out what (if any of it) is still relevant, but probably close all/most of them. Then write up a bunch of new frontend issues for the next steps on this with some proper details. If you want though I can try to pull all that together into a rough bullet list here tomorrow evening before we merge this.

chris48s:
today is the day

PR #9016, opened 2023-03-23
Update to [CTAN] API version 2.0

muzimuzhi:
also use https protocol for CTAN tests

see https://www.ctan.org/help/json/2.0/pkg


chris48s:
Thanks for this, but your mocked tests are failing in CI. I think it is because you've forgotten to update the intercept URLs to use `/2.0` so the mock doesn't match the API call the service code is making. Any chance you could revisit?

muzimuzhi:
> Thanks for this, but your mocked tests are failing in CI. I think it is because you've forgotten to update the intercept URLs to use `/2.0` so the mock doesn't match the API call the service code is making.

Thank you for pointing this out to me!  I didn't notice there're more helpful error messages in the workflow run summary at the time.

With the updated intercept URLs and a new regex/validator to match version `v1.81a`, now all tests have passed.

chris48s:
Thanks. LGTM

chris48s (approved):

Issue #6308, opened 2021-03-23
Add Minecraft server stats (online players, server online status)


YHDiamond:
:clipboard: **Description**

A badge to show the amount of players on a Minecraft server at one time and other stats about a server

- This is a service for Minecraft servers
- Can show online server status (https://img.shields.io/badge/Minecraft%20server-online-brightgreen or https://img.shields.io/badge/Minecraft%20server-offline-red)
- Can show number of online players (https://img.shields.io/badge/Online%20Players-126-blue)
- Can show server software (https://img.shields.io/badge/Server%20Software-Paper-blue)
- Can show server version (https://img.shields.io/badge/Server%20Version-1.16.5-green)
- Online players over player slots (https://img.shields.io/badge/Online%20players-12%2F20-yellow)

:link: **Data**

https://api.mcsrvstat.us/ this is an API that would probably work for most (if not all) of these things.

:microphone: **Motivation**

This would be most useful for Minecraft projects with a server or test server (GeyserMC and Mineweb are just 2 that I can name off the top of my head, I am sure there are more).

PyvesB:
Hello @YHDiamond ! 👋🏻 

> This would be most useful for Minecraft projects with a server or test server (GeyserMC and Mineweb are just 2 that I can name off the top of my head, I am sure there are more).

Being somewhat involved in the Minecraft community myself, I can also confirm that some projects do have test servers as you've described.

However, going back to our [contributing guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md):

> Shields.io hosts integrations for services which are primarily used by developers or which are widely used by developers.

I feel that saying that a Minecraft server is a service primarily/widely used by developers is somewhat stretching things. I'm not against this badge per say, but I think it would be useful for other maintainers to share their views as well. 😉

calebcartwright:
> I feel that saying that a Minecraft server is a service primarily/widely used by developers is somewhat stretching things. I'm not against this badge per say, but I think it would be useful for other maintainers to share their views as well.

Yeah I had the same reaction. I think this would be best suited to the endpoint badge.

PyvesB:
Going to close this for now. We can always reopen if a use case that truly fits with the core Shields.io mission comes to light in the future. For reference, documentation about the endpoint badge mentioned by @calebcartwright can be found here: https://shields.io/endpoint

PR #9020, opened 2023-03-24
[Hangar] Add service

BrycensRanch:
# Hangar, a new Minecraft marketplace
Hangar allows you to find the best Paper, Velocity, or Waterfall plugins for your Minecraft server. While still in closed beta, they have a extensive API that shields can use to provide day one support for this new, emerging, service that has been worked on 24/7 by PaperMC, the organization behind Paper, one of the most popular Minecraft server software and many contributions to the Minecraft community. Due to the Hangar closed beta, it's robots.txt doesn't allow SEO bots, but, believe me, it will be optimized for SEO. This PR features support for everything authors would ever need and is meant to handle possible API changes. The validation schema for the responses are well-made and specific for known response keys. 

## Tests
All tests have been added. I haven't had a chance to run them, but they're rather simple. Trust me bro.

## Acknowledgements
I have prettier


BrycensRanch:
..? why are tests failing? the error msg doesn't seem very useful

calebcartwright:
Failure seemed potentially specious. Also appeared execution was too old to re-run so closed/re-opened the PR to restart the jobs. Let's see whether the failure repeats

calebcartwright:
Test failures seem consistent, not sure offhand what the cause may be as I've not really looked at the details of proposed changes and tests.

However, while I appreciate the PR, I think there's some higher level challenges with incorporating this, and as such I'm going to close with apologies for the likely resultant disappointment.

For context, we previously rejected a request for Minecraft related badges back in #6308, primarily because of not having great alignment with the project scope:

> Shields.io hosts integrations for services which are primarily used by developers or which are widely used by developers.

I don't think the calculus has really changed on this front, so I don't think these Minecraft-related badges are a good fit either.

Furthermore, we also don't typically bring on native badges (those with direct routes) for nascent services for a variety of reasons, but mostly because from our end it doesn't make sense to add on an integration with a service until it reaches a certain level of maturity and adoption.

Finally, even if/once the above challenges were to be obviated, there's some implementation considerations that would have to be tackled as well (e.g. we aren't accepting new custom logos, we need explicit, public facing documentation for the APIs and support/issue reporting process, etc.).

If you (or anyone else) would like to add support for these via the Custom Endpoint model we suggested in #6308 then we'd be happy to get that listed over in https://github.com/badges/awesome-badges for greater visibility and adoption

PR #9036, opened 2023-03-27
[CTAN] fallback to date if version is empty

### LINKED ISSUE #9017: Some version badges for CTAN packages are broken

shields.io

### 🐞 Description

Some CTAN packages like `l3kernel`, `l3build` and `zhspacing` (deliberately) use date as version,  their queried json look like
```bash
$ curl https://www.ctan.org/json/2.0/pkg/l3kernel | python3 -m json.tool | grep -A 3 '"version"'
    "version": {
        "number": "",
        "date": "2023-02-22"
    },
```
Hence their version badges are broken.

See a real use case of `zhspacing` in https://github.com/CTeX-org/ctex-kit#our-packagescollections-on-ctan.

### 🔗 Link to the badge

https://img.shields.io/ctan/v/zhspacing.svg

### 💡 Possible Solution

If `json.version.number` is empty, try `json.version.date` instead.
https://github.com/badges/shields/blob/19ec4a3c92229a2c9e10cb55eb206c9e1dfaec9a/services/ctan/ctan.service.js#L68-L72

From CTAN doc https://www.ctan.org/help/json/2.0/pkg, 
> At least one of number or date have to be given. Otherwise the tag is suppressed.

I'm not sure if `number: ""` represents a not-given version number.  If so, perhaps `date` should be added to the `schema`.
https://github.com/badges/shields/blob/19ec4a3c92229a2c9e10cb55eb206c9e1dfaec9a/services/ctan/ctan.service.js#L6-L11

muzimuzhi:
I'm new to node.js, hence this PR is more or less an imitation of existing code in the same repo.

Do you suggest to make `version.date` in `schema` also required?

Fixes #9017

### DIFF
--- a/services/ctan/ctan.service.js
+++ b/services/ctan/ctan.service.js
@@ -70,7 +70,7 @@ class CtanVersion extends BaseCtanService {
     const json = await this.fetch({ library })
     const version = json.version.number
     if (version.length > 0) {
-      return renderVersionBadge({ version: version })
+      return renderVersionBadge({ version })
     } else {
       const date = json.version.date
       if (date.length > 0) {
--- a/services/ctan/ctan.service.js
+++ b/services/ctan/ctan.service.js
@@ -1,8 +1,7 @@
 import Joi from 'joi'
 import { renderLicenseBadge } from '../licenses.js'
 import { renderVersionBadge } from '../version.js'
-import { BaseJsonService } from '../index.js'
-import { InvalidResponse } from '../error.js'
+import { BaseJsonService, InvalidResponse } from '../index.js'
 
 const schema = Joi.object({
   license: Joi.array().items(Joi.string()).single(),
--- a/services/ctan/ctan.service.js
+++ b/services/ctan/ctan.service.js
@@ -7,7 +7,7 @@ const schema = Joi.object({
   license: Joi.array().items(Joi.string()).single(),
   version: Joi.object({
     number: Joi.string().allow('').required(),
-    date: Joi.string().allow('')
+    date: Joi.string().allow(''),
   }).required(),
 }).required()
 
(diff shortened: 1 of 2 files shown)


services/ctan/ctan.service.js
```diff
@@ -1,12 +1,13 @@
...
 import { renderVersionBadge } from '../version.js'
-import { BaseJsonService } from '../index.js'
+import { BaseJsonService, InvalidResponse } from '../index.js'
 
 const schema = Joi.object({
   license: Joi.array().items(Joi.string()).single(),
   version: Joi.object({
-    number: Joi.string().required(),
+    number: Joi.string().allow('').required(),
+    date: Joi.string().allow(''),
```
chris48s, line 10: if we expect this to be present in the response but set to empty string to indicate "not set" then yes it should be `.required()`

services/ctan/ctan.tester.js
```diff
@@ -8,6 +8,10 @@ const isVPlusDottedVersionAtLeastOneWithOptionalAlphabetLetter = withRegex(
   /^v\d+(\.\d+)?(\.\d+)?[a-z]?$/
 )
 
+// a loose pattern matches an ISO 8601 date
+// e.g.: 2023-01-01
+const isIsoDate = withRegex(/^\d{4}-\d{2}-\d{2}$/)
```
chris48s, line 13: I don't think we need to make a custom validator here. We should be able to use `Joi.date().iso()`

chris48s (commented):
Thanks for picking this up :) Couple of minor comments

services/ctan/ctan.service.js
muzimuzhi: In https://github.com/muzimuzhi/shields/commit/80c1fe94c9106659ec010d6b5247edb103f8a377 I added `.required()` to `date` without adding `date` entry to tests. Surprisingly, all checks passed.  Is this the expected result?

services/ctan/ctan.tester.js
muzimuzhi: Ah, I just don't know `Joi.date().iso()` can also be used here.

chris48s (approved):
nice LGTM :+1:

PR #9054, opened 2023-04-05
Update the hint of simple-icons

LitoMore:
### Changes

- Update the hint of simple-icons. (https://github.com/simple-icons/simple-icons-website/pull/281)
- Fix the link of `slug.md`. We should use the `slug.md` file on the master branch instead.


chris48s:
The reason I linked to `develop` is because that's the default branch on the simple-icons repo. I can't see any docs in https://github.com/simple-icons/simple-icons that explain the branching model. Does the tip of `master` correspond to the latest released version?

LitoMore:
Simple Icons has a weekly script to create a pull request to merge develop to master like this: https://github.com/simple-icons/simple-icons/pull/8501.

And the `npm publish` workflow only triggers when the master branch change. See https://github.com/simple-icons/simple-icons/blob/cea8a4365098fd27076ed53b3fa4678644089976/.github/workflows/publish.yml.

chris48s:
Thanks. Makes sense.
I've made some minor edits to the text but I'll merge this.
More importantly I'll update https://github.com/badges/shields/pull/9014 with this. Hopefully we are close to the final days of this frontend :crossed_fingers:

chris48s (approved):

Issue #8266, opened 2022-07-31
Badges for bundlejs


okikio:
:clipboard: **Description**

- This badge is for bundlejs.com. It would allow people to bundle, treeshake and compress their code, and then view the compressed size in multiple different compression formats, e.g. brotli, lz4, gzip
- The badge will show the treeshaken & compressed size of packages in a format similar to this `{size} ({compresssion-format}) | {bundlejs-logo}`

:link: **Data**

- We can get the data from bundlejs' official public API (https://bundlejs.com/api?q=react) or we can bundle the code live.
- The API docs are still being finalized but here are the incomplete API docs for bundlejs https://blog.okikio.dev/documenting-an-online-bundler-bundlejs#heading-sharing-bundle-sessions

:microphone: **Motivation**

- The specific use case is for more accurate, treeshaken, bundle sizes. The biggest motivating factor is that users of bundlejs want a badge, and I thought other might want one as well, https://github.com/okikio/bundlejs/issues/30

calebcartwright:
I think this would be a reasonable badge to add, though it sounds like we might need to wait to incorporate until the api is viewed as production ready, etc.

jNullj:
Closing as PR adding the badge seems to be merged and live for a long time now.

PR #9055, opened 2023-04-06
[bundlejs] add badge for the npm package size

crimx:
View [esbuild](https://github.com/evanw/esbuild) minified and [denoflate](https://github.com/hazae41/denoflate) gzipped size of a package or selected exports, via [bundlejs](https://bundlejs.com/).

Comparing to the existing bundlephobia service, bundlejs is more accurate in terms of treeshaking. It also supports displaying only the size of the selected exports.

cc @okikio

### DIFF
--- a/services/bundlejs/bundlejs-package.service.js
+++ b/services/bundlejs/bundlejs-package.service.js
@@ -126,7 +126,7 @@ export default class BundlejsPackage extends BaseJsonService {
         ETIMEDOUT: { prettyMessage: 'timeout', cacheSeconds: 10 },
       },
       httpErrors: {
-        404: 'timeout',
+        404: 'package or version not found',
       },
     })
   }
--- a/services/bundlejs/bundlejs-package.service.js
+++ b/services/bundlejs/bundlejs-package.service.js
@@ -1,0 +1,122 @@
+import Joi from 'joi'
+import { BaseJsonService } from '../index.js'
+
+const schema = Joi.object({
+  size: Joi.object({
+    compressedSize: Joi.string(),
+  }).required(),
+}).required()
+
+const keywords = ['node', 'bundlejs']
+
+const esbuild =
+  '<a href="https://github.com/evanw/esbuild" target="_blank" rel="noopener">esbuild</a>'
+const denoflate =
+  '<a href="https://github.com/hazae41/denoflate" target="_blank" rel="noopener">denoflate</a>'
+const bundlejs =
+  '<a href="https://bundlejs.com/" target="_blank" rel="noopener">bundlejs</a>'
+
+const documentation = `
+<p>
+ View ${esbuild} minified and ${denoflate} gzipped size of a package or selected exports, via ${bundlejs}.
+</p>
+`
+
+export default class BundlejsPackage extends BaseJsonService {
+  static category = 'size'
+
+  static route = {
+    base: 'bundlejs/package',
+    pattern: ':scope(@[^/]+)?/:packageName/:exports?',
+  }
+
+  static examples = [
+    {
+      title: 'npm package minimized gzipped size',
+      pattern: ':packageName',
+      namedParams: {
+        packageName: 'react',
+      },
+      staticPreview: this.render({ size: '2.94 kB' }),
+      keywords,
+      documentation,
+    },
+    {
+      title: 'npm package minimized gzipped size (version)',
+      pattern: ':packageName',
+      namedParams: {
+        packageName: 'react@18.2.0',
+      },
+      staticPreview: this.render({ size: '2.94 kB' }),
+      keywords,
+      documentation,
+    },
+    {
+      title: 'npm package minimized gzipped size (scoped)',
+      pattern: ':scope/:packageName',
+      namedParams: {
+        scope: '@cycle',
+        packageName: 'rx-run',
+      },
+      staticPreview: this.render({ size: '32.3 kB' }),
+      keywords,
+      documentation,
+    },
+    {
+      title: 'npm package minimized gzipped size (select exports)',
+      pattern: ':packageName/:exports',
+      namedParams: {
+        packageName: 'value-enhancer',
+        exports: 'isVal,val',
+      },
+      staticPreview: this.render({ size: '823 B' }),
+      keywords,
+      documentation,
+    },
+    {
+      title:
+        'npm package minimized gzipped size (scoped version select exports)',
+      pattern: ':scope/:packageName/:exports',
+      namedParams: {
+        scope: '@ngneat',
+        packageName: 'falso@6.4.0',
+        exports: 'randEmail,randFullName',
+      },
+      staticPreview: this.render({ size: '17.8 kB' }),
+      keywords,
+      documentation,
+    },
+  ]
+
+  static defaultBadgeData = { label: 'bundlejs', color: 'informational' }
+
+  static render({ size }) {
+    return {
+      label: 'minified size (gzip)',
+      message: size,
+    }
+  }
+
+  async fetch({ scope, packageName, exports }) {
+    const searchParams = {
+      q: `${scope ? `${scope}/` : ''}${packageName}`,
+    }
+    if (exports) {
+      searchParams.treeshake = `[{${exports}}]`
+    }
+    return this._requestJson({
+      schema,
+      url: 'https://deno.bundlejs.com',
+      options: { searchParams },
+      errorMessages: {
+        404: 'package or version not found',
+      },
+    })
+  }
+
+  async handle({ scope, packageName, exports }) {
+    const json = await this.fetch({ scope, packageName, exports })
+    const size = json.size.compressedSize
+    return this.constructor.render({ size })
+  }
+}
--- a/services/bundlejs/bundlejs-package.service.js
+++ b/services/bundlejs/bundlejs-package.service.js
@@ -107,7 +107,15 @@ export default class BundlejsPackage extends BaseJsonService {
     return this._requestJson({
       schema,
       url: 'https://deno.bundlejs.com',
-      options: { searchParams },
+      options: {
+        searchParams,
+        timeout: {
+          request: 3000,
+        },
+      },
+      systemErrors: {
+        ETIMEDOUT: { prettyMessage: 'bundling', cacheSeconds: 10 },
+      },
       errorMessages: {
         404: 'package or version not found',
       },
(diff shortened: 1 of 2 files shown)


okikio:
@crimx Thanks

services/bundlejs/bundlejs-package.tester.js
```diff
@@ -0,0 +1,33 @@
+import { isFileSize } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('bundlejs/package (packageName)')
+  .timeout(20000)
```
chris48s, line 6: Before reviewing this in any detail, seeing these timeouts made me want to look into API peformance.

For shields we really need the upstream API call to usually return in under 4 seconds for a badge to be viable. GitHub (which is the place where badges are most commonly viewed) has a hard limit of 4 seconds on an image request so if it takes us >4 seconds to render a badge, they just don't get displayed.

Having tried making a few API calls, it seems sometimes the request does take a really long time (23 seconds was the longest I observed), but then the second call to the same URL is quick. I guess the API is slow if the request is a cache miss but then fast once the data is cached.

@okikio - are you also to provide any info on this? Also there any rate limits on using this API, etc?

okikio: That's due to the bundlejs API bundling the package 📦. 

By default bundlejs caches the bundle on request but we set the TTL to 1 day to allow for new versions to propagate quickly.

chris48s: This feels like a really similar issue to what we had with socket.dev in https://github.com/badges/shields/pull/7727

For a really popular package its probably not a huge issue if one request in 24 hours is really slow. For smaller less popular packages, its probably more of an issue. I don't think I have a really clear answer right now, but it might be a good one for us to discuss next time we catch up @calebcartwright

okikio: For smaller packages it's usually rather fast, it the larger packages that are the largest slow down

crimx: Theoretically bundlejs should be faster than bundlephobia who is using webpack. @okikio Have you looked into the performance bottleneck of the service?

okikio: Bundlephobia is bundling all the packages and storing them permanently, I'm currently not doing that because bundlejs allows for custom code so you can get treeshaken bundle sizes, and I just hadn't gotten around to setting up a permanent database for storing all the bundle size data for simple straight forward bundles, e.g. react@18.0.0

I'll get on it as soon as I can

crimx: Bundlephobia also caches tree-shaking size of each export. They are limited and inaccurate most of the time due to the webpack config.

okikio: Yeah, in the future I hope to also do treeshake bundle analysis for all the exports in a package, right now you have to manually give it a the functions and methods you'd like tteeshaken

chris48s: Hello. After some discussion in https://github.com/badges/shields/issues/9174 we've settled on a pattern for handling services that work in this way and implemented it in https://github.com/badges/shields/pull/9233

I appreciate this has take a while, but would you be able to rebase this PR on master and update it use the approach shown in code example in https://github.com/badges/shields/pull/9233 ?

crimx: Sure.

okikio: Yeah, don't use `no-cache` if you can avoid it

crimx: > Yeah, don't use `no-cache` if you can avoid it

Any ETA for the fix? Last time I checked the package size was still incorrect. Maybe cache with `q` plus other search params?

calebcartwright:
Just a bit of a triage ping. Is there still an intent to overhaul this to incorporate some of the subsequent changes that landed on our default branch, including the new pattern Chris mentioned in https://github.com/badges/shields/pull/9055#discussion_r1228651419 that's specifically geared for upstream services/platforms that exhibit this type of behavior?

If not perhaps we could close this and shift discussion back to #8266 in case someone else is willing/interested in working on it?

services/bundlejs/bundlejs-package.tester.js
okikio: I forgot to leave a message here this has now been fixed

crimx:
I was waiting for the [`options.timeout.request` API](https://github.com/badges/shields/issues/9174#issuecomment-1650384154). [@okikio](https://github.com/okikio) just metioned that the bundlejs results are permanently cached now. I can also make other workarounds for testing.

crimx:
@calebcartwright ready to merge

chris48s:
Thanks for updating. I can have a look at reviewing this one over the weekend.

services/bundlejs/bundlejs-package.service.js
```diff
@@ -0,0 +1,130 @@
...
+    {
+      title: 'npm package minimized gzipped size',
+      pattern: ':packageName',
+      namedParams: {
+        packageName: 'react',
+      },
+      staticPreview: this.render({ size: '2.94 kB' }),
+      keywords,
+      documentation,
+    },
```
calebcartwright, line 43: Will defer to Chris on whether we want to let this through as-is and refactor with the others later, or try to apply the new model now.

(we've changed the way we set some of these as part of an overhaul to the frontend site)

services/bundlejs/bundlejs-package.tester.js
calebcartwright: > bundlejs now permanently caches the bundle size for packages

Does this mean that the timeout behavior will only ever occur for first-time packages, and that e.g. the size for jquery is permanently stored? If so then we could remove these timeouts from the code (it'll never get that far anyway given the timeout short circuit in the request fetching)

calebcartwright (commented):
Couple comments left below

services/bundlejs/bundlejs-package.service.js
```diff
@@ -0,0 +1,130 @@
+import Joi from 'joi'
+import { BaseJsonService } from '../index.js'
+
+const schema = Joi.object({
+  size: Joi.object({
+    compressedSize: Joi.string(),
```
chris48s, line 6:
```suggestion
    compressedSize: Joi.string().required(),
```

If we're going to use `compressedSize`, it needs to be `.required()`

services/bundlejs/bundlejs-package.service.js
```diff
@@ -0,0 +1,130 @@
...
+</p>
+`
+
+export default class BundlejsPackage extends BaseJsonService {
+  static category = 'size'
+
+  static route = {
+    base: 'bundlejs/package',
+    pattern: ':scope(@[^/]+)?/:packageName/:exports?',
+  }
```
chris48s, line 31: Based on our URL conventions https://github.com/badges/shields/blob/master/doc/badge-urls.md , I'd say:

- the base should be `bundlejs/size` - the thing we're actually displaying on the badge is a size, not a package
- `exports` should be a query param, not part of the route. It is optional and I'd expect it to be relatively infrequently used

services/bundlejs/bundlejs-package.tester.js
```diff
@@ -0,0 +1,33 @@
+import { isFileSize } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
```
chris48s, line 3: Could we add one test cases that uses `.replyWithError()` https://github.com/nock/nock#replying-with-errors to test the `ETIMEDOUT` case

chris48s, line 3: Just for clarity, is the strategy we're relying on for the tests here just picking really popular packages that are likely to be already cached at any given time?

services/bundlejs/bundlejs-package.service.js
chris48s: I'm not really worried about this. Given the documentation still says to use `examples`, I think it is unreasonable to expect contributors to do something else as this point.

services/bundlejs/bundlejs-package.tester.js
chris48s: Yeah I think given this badge will always return a response of some description (even if that is `bundling`) within 3 seconds, we should be able to remove all these `.timeout(20000)`s .

services/bundlejs/bundlejs-package.tester.js
crimx: Bundlejs now [permanently caches the result](https://github.com/badges/shields/pull/9055#discussion_r1292840083).

crimx: Don't know how to test it if the result is permanently cached. Maybe to fake one but I doubt if it's useful.

crimx: Found a way to test it.

crimx: Nope. Still unable to test this.

crimx:
@chris48s Moved exports to query params and removed test timeouts.

When requesting non-existing packages, bundlejs responses with 404 after unpkg timeout. The timing is unpredictable so it may also trigger the ETIMEDOUT. I changed the badge message to "timeout" for both.

chris48s:
So with the 404 case, we're actually hitting the original behaviour. We hit the timeout first time round, but the second request we get the 404 immediately.

I think what we should actually do here is give the user a real "not found" error in the 404 case (if it comes in under the timeout) and test the 404 and timeout cases using mocks, like this:

```js
t.create('bundlejs/package (not found)')
  .get('/react@18.2.0.json')
  .intercept(nock =>
    nock('https://deno.bundlejs.com')
      .get(/./)
      .query({ q: 'react@18.2.0' })
      .reply(404),
  )
  .expectBadge({ label: 'bundlejs', message: 'package or version not found' })

t.create('bundlejs/package (timeout)')
  .get('/react@18.2.0.json')
  .intercept(nock =>
    nock('https://deno.bundlejs.com')
      .get(/./)
      .query({ q: 'react@18.2.0' })
      .replyWithError({ code: 'ETIMEDOUT' }),
  )
  .expectBadge({ label: 'bundlejs', message: 'timeout' })
```

crimx:
Updated. Thanks for the help.

chris48s (approved):
Sorted. Thanks for getting this one over the line!

okikio:
OH, wow, this is amazing, great work @crimx and @chris48s

PR #9064, opened 2023-04-07
[RokuChannelStoreRating] Add Roku Channel Store rating badges

iBicha:
Adding a few badges for the Roku Channel Store, specifically for rating.


iBicha:
I don't know who to tag for review, maybe @chris48s ?

calebcartwright:
There's no need to tag anyone explicitly. We're rather short staffed, so to speak, and we'll get to it when we can. To be fully transparent though reviews for net-new badges aren't as high on our priority list as some other core/maintenance related work that needs to get done, so it may (potentially) be a while before we get back to you.

Thanks for submitting a PR though, appreciate your willingness to contribute!

services/roku-channel-store/roku-channel-store-base.js
```diff
@@ -0,0 +1,30 @@
...
+
+const schema = Joi.object({
+  rating: Joi.number().min(0),
+  starRating: Joi.number().min(0),
+  starRatingCount: nonNegativeInteger,
+}).required()
+
+export default class RokuChannelStoreBase extends BaseJsonService {
+  async fetch({ channelId }) {
+    const url = `https://channelstore.roku.com/api/v6/channels/${channelId}`
```
chris48s, line 13: Is there any documentation for this API?

Also can you explain what is sold in the Roku Channel store? It looks like it is more like TV and radio content than software.

iBicha: > Is there any documentation for this API?

Not that I know of

> Also can you explain what is sold in the Roku Channel store? It looks like it is more like TV and radio content than software.

They are tailored towards TV (since Roku is a TV OS) but it's still apps.
Here's a couple of example repos that could display the badges
https://github.com/iBicha/playlet (store link https://channelstore.roku.com/en-ca/details/840aec36f51bfe6d96cf6db9055a372a/playlet)
https://github.com/jellyfin/jellyfin-roku (store link https://channelstore.roku.com/en-ca/details/cc5e559d08d9ec87c5f30dcebdeebc12/jellyfin)

chris48s: OK thanks. The reason we say

> Badges should not obtain data from undocumented or reverse-engineered API endpoints.

in [our guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines)

is because we had some situations where a contributor worked out that a project had an API by inspecting the frontend code for their website or whatever, added a badge that calls it, and then we ended up relying on an API that wasn't really intended for public use. This then either annoyed the maintainers of the company/project behind it or we had no reasonable expectation of stability/compatibility over time (because devs maintaining the the API were making the assumption the only thing consuming it was their own frontend code) and stuff kept breaking.

That's looks like the situation we've got here so I am going to close this PR, but we could revisit in future if there is a reliable public source for the data.

PR #9065, opened 2023-04-07
remove redis token persistence backend

### LINKED ISSUE #8932: Remove `RedisTokenPersistence`

chris48s:
closes #8932

When I assemble the changelog for the May release, I'll call this out at the top.


server.js
```diff
@@ -42,15 +42,6 @@ if (fs.existsSync('.env')) {
   process.exit(1)
 }
 
-if (config.private.redis_url != null) {
```
chris48s, line 45: The decision I have taken here is just to remove `redis_url` from the config schema.
The other thing we could do would be leave it in the schema, keep this line here and throw an exception if `redis_url` is set.

calebcartwright: Long term I don't think it's necessary to have that, though I find myself somewhat ambivalent about what to do in the short term. I could see value in having the exception explicitly thrown for another monthly release or two

calebcartwright (approved):
lgtm. i'm of two minds about whether to throw the exception if the var is set, so will approve in case you decide to stick with this but happy to re-:+1: if you go the other route instead

chris48s:
belt and braces..

chris48s:
I'm going to go ahead and merge this one

Issue #8987, opened 2023-03-11
Full service test run `[*****]` doesn't produce a summary


chris48s:
:clipboard: **Description**

At the moment, a full service test run `[*****]` doesn't produce a summary.

Based on https://github.com/badges/shields/actions/runs/4392173004 my theory is that this is because the 2 GitLab private repo tests are still pending and we're not handling that correctly in some way when we generate the summary.

If we want to test the GitLab token theory, we could set a GL token:

- Generate a GitLab API token that can read the private repos we test against
- Set it in https://github.com/badges/shields/settings/secrets/actions and https://github.com/badges/shields/settings/secrets/dependabot
- Consume it in https://github.com/badges/shields/blob/master/.github/actions/service-tests/action.yml
- Pass it through in https://github.com/badges/shields/blob/master/.github/workflows/test-services.yml and https://github.com/badges/shields/blob/master/.github/workflows/test-services-17.yml

That might not be a bad thing to do anyway.

It might also be useful to see if we can reproduce this locally with the mocha2md script and see if we can get it to produce a summary even if tests are skipped. This would be useful for PRs from forks anyway as we can only use the secrets for tests from local branches.

PR #9068, opened 2023-04-09
fix markdown summary if there are pending tests; run [amo youtube]

chris48s:
closes https://github.com/badges/shields/issues/8987

Not 100% sure why I wrote this to exit if there are pending tests, but this fixes it. Included services are just so we can view an example output, not because those services are affected by this change.

### DIFF
--- a/scripts/mocha2md.js
+++ b/scripts/mocha2md.js
@@ -20,21 +20,28 @@ if (data.stats.passes > 0) {
   process.stdout.write(`✔ ${data.stats.passes} passed\n`)
 }
 if (data.stats.failures > 0) {
-  process.stdout.write(`✖ ${data.stats.failures} failed\n\n`)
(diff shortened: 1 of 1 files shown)


chris48s:
> :rocket: Updated review app: https://pr-9068-badges-shields.fly.dev/

Not strictly related to this PR - I'm just using this to test another thing. PR incoming for that soon. Its all coming together...

calebcartwright:
> rocket Updated review app: https://pr-9068-badges-shields.fly.dev

Just want to re-confirm that this came from you manually re-running the deploy workflow after an update right?

calebcartwright (approved):

chris48s:
Yes. There is a tradeoff here between adding information and generating noise. Given we need to manually trigger updating the review app, I figure it is useful to post in the PR each time we update the review app so its easier to keep track of whether commits have been made since the review app was updated.

Issue #8337, opened 2022-08-25
Heroku removing free product plans


chris48s:
> Dear Customer,

oh hai Heroku :wave: 

> 
> Thank you for being a Heroku user. Starting November 28, 2022, free Heroku Dynos, free Heroku Postgres, and free Heroku Data for Redis® will no longer be available. You can learn more about these and other important changes from our GM, Bob Wise, on the [Heroku blog](https://blog.heroku.com/next-chapter).
> 
> Existing free dynos and Heroku data add-ons will be impacted, so action by you is required. To prevent any disruption to your apps or data using free plans, you will need to upgrade from a free plan to a paid plan before November 28, 2022.
> 
> For instructions on how to upgrade and for other questions, please [visit our FAQ](https://help.heroku.com/RSBRUH58/removal-of-heroku-free-product-plans-faq).
> 
> Thank you,
> 
> The Heroku Team

I think for squint this probably doesn't make a difference. It is running on a hobby-dev dyno and will continue to do so.

Neither of the linked articles talk about review apps specifically, but I assume that our review apps will continue to work but bill by the minute running as a hobby-dev dyno instead of a free dyno. I don't think this is an immediate worry, but its something to take into account as we think about fixing review apps (which are half-working, kinda).

I think those are the only things we need to think about, but is there anything else we need to consider here?

Issue #7924, opened 2022-05-03
Restore or replace Heroku Review Apps


chris48s:
:clipboard: **Description**

Until https://status.heroku.com/incidents/2413 happened we were using Heroku review apps to build a preview site for each PR. This was a big help for reviewing contributions and dependency bumps.

The latest update from Heroku is

> For the protection of our customers, we will not be reconnecting to GitHub until we are certain that we can do so safely, which may take some time. We recommend that customers use alternate methods rather than waiting for us to restore this integration.
> 
> The simplest approach to deploy Heroku apps is directly using [git push heroku main](https://devcenter.heroku.com/articles/git). Alternatively, you can refer to one of the add-ons listed in the Continuous Integration and Delivery section of the [Heroku Ecosystem Marketplace](https://elements.heroku.com/addons#ci-cd) or other third party CI/CD products that offer Heroku support, such as [Circle CI](https://circleci.com/integrations/heroku/). You can also use community-maintained integrations such as [this GitHub Action for deploying to Heroku](https://github.com/AkhileshNS/heroku-deploy).
> 
> We sincerely regret any inconvenience you may have experienced because of this issue and appreciate your trust in us as we continue to make your success our top priority. Please continue to visit [status.heroku.com](https://status.heroku.com/incidents/2413) for the latest updates.

The key phrase there for me is:

> **We recommend that customers use alternate methods rather than waiting for us to restore this integration**

Investigate how we could re-enable review apps, either using Heroku via some other avenue or by migrating away from Heroku. One possible thing we could look at is https://github.com/superfly/fly-pr-review-apps

calebcartwright:
One of the challenges I think we'd run into with the Fly review apps model is that it's predicated on having a GitHub Actions secret available, but those are disabled and unavailable on PRs coming from forks for security reasons

calebcartwright:
https://blog.heroku.com/github-integration-update (some progress but still not resolved)

chris48s:
Good find and good point about the fly review apps. I think in the short term re-enabling Heroku review apps when they enable github deploys again is probably the play.

@SnoozeThis wait until May 26 2022

SnoozeThis:
(https://snoozeth.is/X6ov_cwqaA8) I will wait until Thu, 26 May 2022 00:00:00 UTC and then add a comment.

SnoozeThis:
Resolved.

chris48s:
https://blog.heroku.com/github-integration-update

[...]

PR #9069, opened 2023-04-09
add fly.io PR review apps

chris48s:
closes https://github.com/badges/shields/issues/7924
closes https://github.com/badges/shields/issues/8337

We lost access to Heroku review apps, but they were very useful :sob: 

I had a look into https://github.com/superfly/fly-pr-review-apps but this only works on PRs from branches on the source repo (not PRs from forks) because a fly token must be exposed to the workflow. Even if we could expose a fly.io token to PRs from forks, we really don't want to do that.

This PR is based on the code in https://github.com/superfly/fly-pr-review-apps but adapted for our needs. Merging this PR will give us a workflow which we can trigger manually. We give it a PR number when we trigger it and that will create or update a review app for us using fly.io. Some notes:

- It is up to us to manually create/update the review apps. Review apps will not be auto-created or auto-update when the PR is updated after creation. That said, I think this should be an improvement over having to check PRs out locally to review them.
- It is up to us as maintainers to check to make sure PRs we deploy are not going to exflitrate all our secrets or whatever before we click deploy. Any code changes are going to get pulled into a context with a fly.io token in the environment.
- I've set a token in the repo secrets.
- One thing we do need to consider is that review apps will have some cost attached to running them, but we have the budget available in opencollective and I think the convenience is worth paying for.


.github/scripts/deploy-review-app.sh
```diff
@@ -0,0 +1,32 @@
...
+region="ewr"
+org="shields-io"
+
+# Get PR JSON from the API
+# This will fail if $PR_NUMBER is not a valid PR
+pr_json=$(curl --fail "https://api.github.com/repos/badges/shields/pulls/$PR_NUMBER")
+
+# Attempt to apply the PR diff to the target branch
+# This will fail if it does not merge cleanly
+curl --location "$(echo "$pr_json" | jq .diff_url -r)" | git apply
```
chris48s, line 15: Just to clarify: This will fail to deploy a review app if there is a merge conflict.

.github/workflows/deploy-review-app.yml
```diff
@@ -0,0 +1,41 @@
...
+          sudo apt-get -qq update
+          sudo apt-get install -y jq
+
+      - run: .github/scripts/deploy-review-app.sh
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+          PR_NUMBER: ${{ github.event.inputs.pr_number }}
+          # credentials to set when we create the review app
+          SECRETS: |
```
chris48s, line 31: This will give our review apps to the same credentials we use for the service tests for testing services that need them.

.github/workflows/cleanup-review-apps.yml
```diff
@@ -0,0 +1,24 @@
+name: Cleanup Review Apps
+on:
+  schedule:
+    - cron: '0 7 * * *'
+    # At 07:00, daily
```
chris48s, line 5: Apps we have created will exist until we delete them and running them will cost some non-zero amount of money. We need some way to delete inactive apps.

The way I've done this, I've just made a scheduled action that runs once a day and deletes any review apps that exist. This means in some cases, we will delete review apps for in-progress PRs so we'll have to re-create the review app. I think this is OK. It only takes a few minutes to run the workflow create a review app, even from scratch. There's a number of ways we could tweak this: We could run it less frequently. We could only delete review apps for closed PRs. I reckon probably run this for a month and see what the bill looks like then review if we want to tweak it.

.github/scripts/deploy-review-app.sh
```diff
@@ -0,0 +1,32 @@
...
+# Deploy
+flyctl deploy --app "$app" --region "$region"
+
+# Post a comment on the PR
+app_url=$(flyctl status --app "$app" --json | jq -r .Hostname)
+comment_url=$(echo "$pr_json" | jq .comments_url -r)
+curl "$comment_url" \
+    -X POST \
+    -H "Authorization: token $GITHUB_TOKEN" \
+    --data "{\"body\":\"🚀 Updated review app: https://$app_url\"}"
```
chris48s, line 32: We can see an example of this here https://github.com/badges/shields/pull/9068#issuecomment-1501198114

.github/scripts/deploy-review-app.sh
calebcartwright: I'm having a bit of deja vu so apologies if this has been asked and answered before, but is there any reason we wouldn't want to check out the PR ref directly? I.e.

```terminal
git fetch pull/$PR_NUMBER/head:pr-$PR_NUMBER && git checkout pr-$PR_NUMBER 
```

fly.toml
calebcartwright, line 1: Does the creation of this file have the potential to impact any of other workflows/processes? (e.g. the staging/prod apps)?

calebcartwright (dismissed):
Couple questions but otherwise lgtm

.github/scripts/deploy-review-app.sh
chris48s: I don't think we've discussed it before.

At this stage, there is a very pragmatic reason why I'm doing it this way: There is no open PR which has these files in the source branch yet. So if I check out the source branch for any existing PR, `fly.toml` won't be on it. Merging the PR source branch into the working branch allowed me to test this workflow on some real PRs.

Once we get to the stage where all the open PRs are based off a master branch that has these files on it, it might make more sense to check out the branch directly. That would also allow us to deploy review apps for PRs where there is a merge conflict.

I think in the first case, being able to use this on existing open PRs is useful.

fly.toml
chris48s: Our existing `fly.toml`s live elsewhere and just says "grab an image from DockerHub and deploy it". Its `shieldsio/shields` but it could be any old image on DockerHub so what's in this repo only really impacts anything to the extent that the code in it is in the docker image we deploy.

Adding this here does mean if you were to `flyctl launch` from the root of this repo, that would create an app called `shields-io-review-apps` (which doesn't currently exist in our account). If we don't want it in the root, I think we could also move it somewhere else (maybe under `.github`) and specify the path to the config file in `deploy-review-app.sh` with the `--config` arg.

You have made me think this doesn't need to go in the docker image though, so I've added it to `.dockerignore`

.github/scripts/deploy-review-app.sh
chris48s: One thing I realised when I tried to make a preview app for https://github.com/badges/shields/pull/9014/ was that applying the diff doesn't work if there are binary files in it.

I've switched to merging the source branch (which solves that, but will still fail if there is a conflict), but still sticking with merging for now for the reason above

chris48s:
I am clearly still not quite finished with this

calebcartwright (approved):

PR #9072, opened 2023-04-10
remove pull_request:edited trigger from most workflows

chris48s:
When we have a lot of dependabot PRs to review, sometimes we use all our rate limit up and stuff starts failing.

I think one of the issues causing this is that because we trigger a build on `pull_request:edited`, as soon as dependabot edits the top post of a PR to say "dependabot is rebasing this PR", all the builds unnecessarily re-trigger. Then they retrigger again on `pull_request:synchronize` when it force-pushes the branch.

This PR removes the `pull_request:edited` trigger from most of the workflows. I've left it on the service tests because its quite useful that editing the PR title re-runs those.

Hopefully this should reduce the number of builds we run a bit and make it a bit easier to stay under the rate limit.

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,7 +1,7 @@
 name: Danger
 on:
   pull_request_target:
-    types: [opened, edited, reopened, synchronize]
+    types: [opened, reopened, synchronize]
 
 permissions:
   checks: write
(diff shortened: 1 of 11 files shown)


chris48s:
going to merge this one and see if it helps as I bash through this week's dependabot PRs

PR #9076, opened 2023-04-12
[GithubTag] Add ability to exclude GitHub tags containing string

slaclau:


chris48s:
Hello. We have just shipped a new feature which allows you to pass an arbitrary filter to the github tag and release badges. For example

The `filter` param can be used to apply a filter to the project's tag or release names before selecting the latest from the list. Two constructs are available: `*` is a wildcard matching zero or more characters, and if the pattern starts with a `!`, the whole pattern is negated.

e.g:
https://img.shields.io/github/v/tag/badges/shields
https://img.shields.io/github/v/tag/badges/shields?filter=%21server-%2A

This feature should be powerful and flexible enough to enable this use-case (and a bunch of others).

PR #9081, opened 2023-04-14
[crates] Use `?include=` to reduce crates.io backend load

Turbo87:
By default the `/crates/:crate_name` API endpoint returns the categories and keywords related to the crates. Unfortunately this causes two additional database queries on the crates.io backend.

This commit change the request URL to use `?include=` to only include the additional metadata that is needed for the requested badges.


chris48s:
works for me :+1:

chris48s (approved):

PR #9078, opened 2023-04-12
[SourceForge] Added badges for SourceForge

Av32000:
Hi !
In order to answer issue #8741 I added badges for the SourceForge service:
  - `last-commit`
  - `commit-count`
  - `languages`
  - `contributors`
  - `translations`
  - `platform`
  
 Close #8741

PR #9101, opened 2023-04-20
[SourceForge] Add the old API back

wh201906:
PR #9078 breaks the existing URLs for downloads like this
`https://img.shields.io/sourceforge/dt/sevenzip.svg`
broken img
The API is changed to something like this
`https://img.shields.io/sourceforge/downloads/dt/sevenzip.svg`
current api

This commit add the support for the old URLs by adding `sourceforge.service.js` and `sourceforge.tester.js` back with some modifications to adapt the existing `sourceforge-downloads.service.js`.

### DIFF
--- a/services/sourceforge/sourceforge.service.js
+++ b/services/sourceforge/sourceforge.service.js
@@ -1,0 +1,104 @@
+import Joi from 'joi'
+import dayjs from 'dayjs'
+import { renderDownloadsBadge } from '../downloads.js'
+import { nonNegativeInteger } from '../validators.js'
+import { BaseJsonService } from '../index.js'
+
+const schema = Joi.object({
+  total: nonNegativeInteger,
+}).required()
+
+const intervalMap = {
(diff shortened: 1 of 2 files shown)


wh201906:
You can check the diff result of 8d060fd4192acf8bb580092998fc2b93094fbb13(parent of c2998a2dc960c854dd3b571d5b54fd50f4d87b5e, which breaks the existing URLs) and this PR to see what was changed
https://github.com/badges/shields/compare/8d060fd4192acf8bb580092998fc2b93094fbb13...wh201906:old_api?expand=1#diff-8d0f2396969ff76e7735d4e5069f3bf11b83880dd3383191c0b3783a40c69027

chris48s:
Agree with the objective - we broke this by mistake - but we shouldn't duplicate all this code. It only needs a one line change to fix. Closing in favour of https://github.com/badges/shields/pull/9102

PR #9122, opened 2023-04-27
[Hangar] Add badges for Hangar plugin repository

Sytm:
This PR adds badges for https://hangar.papermc.io
API-docs: https://hangar.papermc.io/api-docs

Hangar itself is based on Ore, for which also badges exist, but aren't entirely a 1to1 match
 
Multiple things I am not really sure about:
- For the version badge, Hangar provides a default release channel, but other ones can have custom names. Should it be a url pattern parameter like it is done at the moment or with query parameters similar to `include_prereleases`
- The way I added the notice to the documentation in hangar-version, is that something that is doable, because it only applies to that one service
- Can I use others public projects to test against or should I create some specifically for testing? (I already did for the invalid test, md5lukas is also a name I use)?
- Should I use :slug as the parameter name, because that is the way it is named in the API-docs or should I prefer the more user-friendly :name?

### DIFF
--- a/services/hangar/hangar-base.js
+++ b/services/hangar/hangar-base.js
@@ -1,0 +1,53 @@
+import Joi from 'joi'
+import { BaseJsonService } from '../../core/base-service/index.js'
+
+const projectSchema = Joi.object({
+  stats: Joi.object({
+    stars: Joi.number().required(),
+    downloads: Joi.number().required(),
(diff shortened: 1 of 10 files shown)


Sytm:
I swear I checked thrice, but it looks like someone already made a PR

Issue #9152, opened 2023-05-07
Make a govulncheck badge


ldemailly:
for a CI passing https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck

e.g 

https://github.com/fortio/proxy/actions/runs/4909112322/jobs/8765310589
```
govulncheck is an experimental tool. Share feedback at https://go.dev/s/govulncheck-feedback.

Using go1.20.4 and govulncheck@v0.0.0 with
vulnerability data from https://vuln.go.dev/ (last modified 2023-05-05 21:10:24 +0000 UTC).

Scanning your code and 176 packages across [14](https://github.com/fortio/proxy/actions/runs/4909112322/jobs/8765310589#step:4:15) dependent modules for known vulnerabilities...
No vulnerabilities found.
```

calebcartwright:
Thanks for reaching out, but I'm going to close. If you'd like to request a new badge please use our [corresponding issue form](https://github.com/badges/shields/issues/new?assignees=&labels=service-badge&template=3_Badge_request.yml) for new badges and include all the relevant data so that this can be more fully considered, and potentially actionable.

ldemailly:
@calebcartwright no problem, but consider fixing the link at the bottom of https://shields.io/

because it goes to:

https://github.com/badges/shields/issues/new?labels=service-badge&template=3_Badge_request.md

which is how this (incomplete) Issue got created (template changed maybe?)

calebcartwright:
Thanks for pointing that out! Indeed a broken link issue that we should get fixed

ldemailly:
@calebcartwright here: https://github.com/badges/shields/pull/9153/files (did a search, found 2 occurences)

ldemailly:
I sort of gave up and put some hardcoded "green" badge but... I need to check the doc on how to just tie the badge color on some CI output, because I do run a github action for govulncheck and I expect there is a semi easy way to tie a job to the results, all within github+shields urls?

chris48s:
If you have a github actions workflow that runs govulncheck and fails if there is >1 vulnerability, it sounds like you could use a github workflow badge for this.

PR #9153, opened 2023-05-07
Correct issue template in footer (new badge requests url is incorrect)

ldemailly:
https://github.com/badges/shields/issues/9152#issuecomment-1537534258


calebcartwright (approved):
ty :heart:

Issue #9155, opened 2023-05-09
Uptime Robot badges randomly say "invalid"


tdulcet:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Some of the Uptime Robot badges will randomly say "invalid" instead of showing the uptime percentage as expected:

![image](https://github.com/badges/shields/assets/15305150/be5ef2b3-67fe-451f-a491-addab2030995)

The ones that say invalid are not consistent, as refreshing the page after a while just causes different ones to say invalid.

### 🔗 Link to the badge

Here is the link for the README that the above screenshot was taken of: https://github.com/tdulcet/send-instances/blob/master/README.md#instances

chris48s:
Hello. I've had a look at one of the ones that was failing and the reason they're failing is because of a 429 `Rate limit exceeded, retry in 20 seconds`.
With uptime robot, the API key used is user-supplied so I guess you are exceeding the rate limit for your plan.

tdulcet:
@chris48s - Thanks for the information. I can confirm that the rate limit is being exceeded:
```
x-ratelimit-limit: 10
x-ratelimit-remaining: 9
x-ratelimit-reset: 1683707896
```

However, this "invalid" message implies that the API key is invalid, which is incorrect. It should say something like "rate" or "limit" instead to make this clearer or maybe not even load at all like the Dynamic badges do when the server for `url` parameter is down.

Ideally, Shields.io should be caching these Uptime Robot API responses and then just returning those cached values whenever this rate limit is reached. Another solution would be for the Shields.io server to just wait out the 20 seconds and then retry the API request until it succeeds. Most browser response timeouts are much more than 20 seconds, so while this obviously would not be ideal for the user, it should at least work. I just loaded our README again and now 16 of the 17 Uptime Robot badges say "invalid", so they are completely unusable in their current state. Please consider reopening this issue.

chris48s:
I think its reasonable for us to add a custom error for the 429 response to make this clearer.

Shields is basically stateless and we don't cache any upstream API responses. We do cache the badges themselves downstream though. Uptime robot badges are cached for 2 minutes by default, but you can explicitly ask us to cache them for longer using the `cacheSeconds` query param.

[...]

PR #9159, opened 2023-05-10
set a custom error on 429; affects [matrix twitch weblate]

chris48s:
This PR is a follow up to #9155 but it is also the starting point of a slightly wider conversation.

I think I'm fairly convinced that this is a good idea for services where the user supplies an API key and we should actually do this on other services following this pattern. One thing I would quite like to do though is see if we can come up with a way to express this that clarifies that it is not us who has applied the rate limit and succinctly communicate that the rate limit which has been exceeded is that of the upstream service/user's API key, not shields.io.

In principle, we could set a custom error like this one level up (so it applies to all badges). 429 always means "too many requests". I'm slightly in 2 minds about this though. For services where the api key is supplied by the user this feels like more useful information/more likely to be actionable in some way. For a service that relies on anon access or a key we hold "centrally", its not really an actionable error. That said, its probably at least more informative. Any thoughts on this?


chris48s:
Had a quick chat about this today.

Decided we are going to make it global (even if its our api key) and message should be "rate limited by upstream service"

Also consider impact on..
https://github.com/search?q=repo%3Abadges%2Fshields%20429&type=code

calebcartwright:
> Also consider impact on..
https://github.com/search?q=repo%3Abadges%2Fshields%20429&type=code

Interesting. The Twitch one raised my eye :eyes:

calebcartwright (approved):

Issue #9189, opened 2023-05-21
Support GitHub workflow badge for private repos


agboom:
:clipboard: **Description**

The GitHub workflow badge does not support private repos, even with a PAT with access to them.

If I understand the code correctly, the current badge service uses the public url for [Workflow Status badges](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge) is fetched in the `GithubActionsWorkflowStatus` and then converted to a Shields badge. That explains why the status of private repos cannot be fetched.
As said in the GitHub docs:
> Workflow badges in a private repository are not accessible externally, so you won't be able to embed them or link to them from an external site.

To support this I think the GitHub API needs to be used instead.

chris48s:
The history behind why this badge doesn't use the API is long and spread across many issues, but we did at one point migrate this to use the API instead of scraping GitHub's native badge.

Unfortunately, this introduced several bugs that we weren't able to solve using the data/endpoints exposed by the API so we switched back to scraping the native badge. The key issue is https://github.com/badges/shields/issues/8733 (this was one of the bugs introduced by switching to the API as the data source) and to solve it we picked option 3 from https://github.com/badges/shields/issues/8733#issuecomment-1356274451 . We accepted that the tradeoff of that solution is that users who self host would not be able to use it for private repos, but it is a known limitation rather than an unexpected bug.

If you want to read the whole saga that lead to this, the relevant issues and PRs are:

- https://github.com/badges/shields/issues/4587
- https://github.com/badges/shields/issues/8146
- https://github.com/badges/shields/pull/8475
- https://github.com/badges/shields/issues/8671
- https://github.com/badges/shields/issues/8736
- https://github.com/badges/shields/issues/8733
- https://github.com/badges/shields/pull/8738

I haven't really looked at https://github.com/badges/shields/pull/9190 but I suspect you're probably introducing the same bugs I introduced in https://github.com/badges/shields/pull/8475 when I tried to migrate this before and then reverted in https://github.com/badges/shields/pull/8738.

agboom:
Looking at the issues more closely I understand that using the API has some problems outside of our control that affect the performance required for this badge. In that case I think we indeed have no choice but to use the SVG scraper.

chris48s:
> I should have searched more thoroughly before I created this issue

No worries. As I said, the history of this is spread across multiple discussions so there isn't really a single search result that would have given all the background on this.

[...]

PR #9190, opened 2023-05-21
[GithubActionsWorkflowStatus] Support private repos

agboom:
Implements #9189 

The badge is fully backwards compatible with the tests of the current implementation, but instead of the `badge.svg` URL it fetches the workflow status using the GitHub API, allowing for access to private repos using a PAT.

As a bonus, the `:workflow` path parameter is now optional. If empty, it takes the status of the latest run of all workflows in a repository.
Depending on whether the workflow name is passed, it uses either the [List runs for a repo API](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-repository) or [List runs for a workflow API](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow).

I made a few small changes outside of the service code. I may be able to implement this in other ways if it doesn't fit the overall design. Please let me know what you think:
- In `build-status.js` I added a `readableStatus`, replacing `_` with ` ` for readability. This will show a badge message `in progress` instead of `in_progress`, which I think is nicer to read.
- In `validate.js`, I added a way to pass options to the `_validate` method via `_requestJson` to be able to override the `prettyErrorMessage`. I did this to make sure the `no build` message is set if any validation error occurs (e.g. if there are no runs found). Another way would be to loosen the schema of the request and check the length somewhere else.

Please let me know any feedback you have. I'm not a first-time contributor to shields, but it's been a long time so I might be rusty :see_no_evil:

### DIFF
--- a/core/base-service/base-json.js
+++ b/core/base-service/base-json.js
@@ -34,10 +34,18 @@ class BaseJsonService extends BaseService {
    *    and custom error messages e.g: `{ 404: 'package not found' }`.
    *    This can be used to extend or override the
    *    [default](https://github.com/badges/shields/blob/master/core/base-service/check-error-response.js#L5)
+   * @param {object} [attrs.validateArgs={}] Arguments to pass to `_validate`.
+   *    For example to override the default `prettyErrorMessage`.
    * @returns {object} Parsed response
    * @see https://github.com/sindresorhus/got/blob/main/documentation/2-options.md
    */
-  async _requestJson({ schema, url, options = {}, errorMessages = {} }) {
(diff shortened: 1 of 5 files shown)


chris48s:
I suspect we're going to close this but I'm going to mark it "on hold" for the moment while we continue the conversation over in https://github.com/badges/shields/issues/9189#issuecomment-1556265808

agboom:
https://github.com/badges/shields/issues/9189#issuecomment-1556265808

Issue #8372, opened 2022-09-02
Add auto-size support for Simple Icons logos


LitoMore:
:clipboard: **Description**

The `logo` feature only renders in the `viewBox="0 0 24 24"`. For some logos, we may not get a nice view. For example:

![](https://shields.io/badge/Amazon_Prime-too_small-eee?logo=amazonprime)
![](https://shields.io/badge/shields.io-too_small-eee?logo=shieldsdotio)

We could calculate the aspect ratio of the icon by using [svg-path-bbox](https://github.com/mondeja/svg-path-bbox):

```js
import svgPathBbox from 'svg-path-bbox'

const [x0, y0, x1, y1] = svgPathBbox(icon.path)
const width = x1 - x0
const height = y1 - y0

// We may need to deal with `width > height` for badges
if (width > height) {
  // Resize the logo part
}
```

We could handle the resizing logic in `load-simple-icons.js` or other places.

Other  `logoWidth` and `logoHeight` related logics may also need to be updated.

chris48s:
Just to summarise: At the moment, the logo portion of the badge is a fixed width. Your suggestion is that the logo portion of the badge becomes variable width - is that right?

LitoMore:
> Just to summarise: At the moment, the logo portion of the badge is a fixed width. Your suggestion is that the logo portion of the badge becomes variable width - is that right?

Yes.

chris48s:
Got it.

So I guess the tradeoff we have now is there is some subset of "wide but short" logos that render unhelpfully small, whereas one of the nice things about the way we do things right now is that logos which don't take up 100% of the viewport vertically but look perfectly fine are all the same width. e.g: These 3 logos all have slightly different non-blank vertical height, but they all render the same width:

- ![](https://img.shields.io/badge/some-text-blue?logo=3m)
- ![](https://img.shields.io/badge/some-text-blue?logo=adafruit)
- ![](https://img.shields.io/badge/some-text-blue?logo=alfred)

Going to a variable horizontal width for logos would make all of those different widths.

Then thinking about the most annoying edge case I can generate in the other direction: Imagine a "logo" which is a 1 pixel wide horizontal line. That would cause us to render a really really wide badge :) Obviously that's a deliberately adversarial input, but worth thinking about.

I'm on the fence about this - it might be good to mock up some example images to help decide. Maybe if we accepted a PR for this it would make sense to set some kind of upper and lower bound on this. For example:

[...]

PR #9191, opened 2023-05-22
feat(logos): support auto-size mode

LitoMore:
Resolves https://github.com/badges/shields/issues/8372

I will add test and documentation once we finalize the idea.

### Principles

This `path` was initially wrapped in a `viewBox="0 0 24 24"` and centered. We need to reset the position to `{x: 0, y: 0}` of the path before updating its viewBox.

### Notes

- ~~The parameter name `logoSize` and its value name `cover` may not be the best. Just let me know what you prefer.~~
- Both `load-simple-icons.js` and `svg-helper.js` called `loadSimpleIcons()`, we could reorganize the code to improve its performance.

### New params

- `logoSize`
- `logoHeight`

### Previews

```
http://localhost:8080/badge/gaming-chip-ED1C24?logo=amd&logoColor=white&logoSize=auto
http://localhost:8080/badge/gaming-ED1C24?logo=amd&labelColor=gray&logoSize=auto
http://localhost:8080/badge/gaming-ED1C24?logo=amd&labelColor=555&logoSize=auto
http://localhost:8080/badge/gaming-ED1C24?logo=amd&labelColor=555&logoSize=auto&logoHeight=10
http://localhost:8080/badge/gaming-ED1C24?logo=amd&labelColor=555&logoSize=auto&logoWidth=20
http://localhost:8080/badge/gaming-eee?logo=amd&logoColor=black&labelColor=ED1C24&logoSize=auto
http://localhost:8080/badge/gaming-555?logo=amd&logoColor=black&labelColor=ED1C24&logoSize=auto
```


chris48s:
Thanks for having a look at this.

@caleb - in terms of the concept (i.e: "do we think this feature is a good idea?", as opposed to "is this good code?"), what are your views on this one?

There is some more detail/discussion in https://github.com/badges/shields/issues/8372

calebcartwright:
> @calebcartwright - in terms of the concept (i.e: "do we think this feature is a good idea?", as opposed to "is this good code?"), what are your views on this one?

I agree that the auto-sizing feature is a good idea. I haven't looked at the implementation at all so not sure what the current state on this topic is, but my only preference is that this behavior be an explicit opt-in (e.g. via a query param). I  say that because it feels like this is relevant in more niche use cases and I don't want to apply it by default as I suspect that would be accompanied with a large number of unknowns

chris48s:
OK, sounds reasonable. I'll put this one on my list to review the code then

chris48s:
I've had a look over this. Broadly, the implementation looks fine. I'll leave a few comments inline.

We would need to add some documentation for this param. In the new frontend, globally available params are documented in

- https://github.com/badges/shields/blob/e7737ce050b72e8e24739da53e0389954b5b2541/core/base-service/openapi.js#L2
- https://github.com/badges/shields/blob/e7737ce050b72e8e24739da53e0389954b5b2541/core/base-service/openapi.js#L233

Other admin: there are failing tests and merge conflicts that would need to be sorted.

chris48s:
In terms of naming, I'm slightly unsure what the best name for this param is either, but I'm keen to get this right as once it is out there, we're basically stuck with it forever.

Some other viable suggestions are `?logoSize=auto` or `?logoSize=fit`. Naming things is hard.

Any other maintainers have a view on this?

core/base-service/coalesce-badge.js
```diff
@@ -149,9 +150,16 @@ export default function coalesceBadge(
     logoPosition = coalesce(overrideLogoPosition, serviceLogoPosition)
   }
   if (namedLogo) {
```
chris48s, line 152: This means this works for simple-icons, the small handful of shields named logos (which are all pretty close to 'square'), but not custom user-supplied base64 logos.
Is there any particular reason why it would be harder or not make sense to make this work for all logos?

core/base-service/coalesce-badge.js
```diff
@@ -149,9 +150,16 @@ export default function coalesceBadge(
...
+    const iconSize = getIconSize(namedLogo.toLowerCase())
+
+    if (!logoWidth && iconSize && overrideLogoSize === 'cover') {
+      logoWidth = (iconSize.width / iconSize.height) * 14
```
chris48s, line 156: What is the significance of 14? Could this be a named constant?

calebcartwright: definitely want to avoid any new [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) :mage:

calebcartwright:
> Any other maintainers have a view on this?

I've a mild preference for `?logoSize=auto`

core/base-service/coalesce-badge.js
LitoMore: User-supplied base64 logos might be non-SVG images. We can figure out its aspect ratio with SVG libraries.

This might need another image process library like https://github.com/FFmpeg/FFmpeg. But this will consume more server resources.

chris48s: OK. I think that's a valid reason to not do it. I'd forgotten base64 logo is not even necessarily a SVG.

`logoWidth` is.. awkward. It already exists as an undocumented param. Also it doesn't do what you would expect: https://img.shields.io/badge/build-passing-brightgreen?logo=javascript&logoWidth=100 it is kind of more like `logoHorizontalPadding` than `logoWidth`. Mostly unrelated to this PR - I think we should probably deprecate/remove this at some point, but of course that will break something for someone. This is its own can of worms, but I think I probably don't want to overload `logoWidth`.

I think my preference would be that we stick with the implementation that only applies to named logos for now. Lets deal with custom logos as another issue (if at all). Sorry we are kinda going back and forth on this one, but I don't want to hold up the part of this that we are fairly settled on with this bit of messy scope creep.

chris48s: Right, yes. Interesting, so this would allow you to take

 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amg&logoSize=auto
 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amd&logoSize=auto
 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=coinbase&logoSize=auto

and make it

 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amg&logoSize=auto&logoWidth=50
 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amd&logoSize=auto&logoWidth=50
 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=coinbase&logoSize=auto&logoWidth=50

which is potentially useful :thinking:

LitoMore: Do you need to update these two options?

- `logoWidth` -> `logoBoundingWidth`
- `logoHeight` -> `logoBoundingHeight`

This will be a breaking change. My suggestion is to write more detailed description for `logoWidth` and `logoHeight`.

```diff
logoWidth
-Default: 14. Same meaning as the query string. Can be overridden by the query string.
+Default: 14. This parameter is used to change the width of the bounding box for the logo.

logoHeight
-Default: 14. Same meaning as the query string. Can be overridden by the query string.
+Default: 14. This parameter is used to change the height of the bounding box for the logo.
```

chris48s: OK. So I've had a bit of play with this, and a bit of a think. I know we're moving around a bit on the implementation, but being able to play with this on a review app is helping and I appreciate the work you're doing on it. I just want to zoom out on this slightly and think about some of the things we're trying to optimise for as a project. It is important to get this right as once we ship some new query params we're basically stuck with them forever.

- **Consistency**: Thinking about why shields was started in the first place, one of the goals was consistency. https://olivierlacan.com/posts/an-open-source-rage-diamond/ In general, as a project, we value providing a small number of good options over loads of configuration options.
- **Good defaults**: We try to give the user good defaults, rather than put the onus on users to configure. A good example of this is the work we did to try and automatically pick a sensible-ish background colour for logos: https://github.com/badges/shields/pull/2833 Make it easy for the user to "fall into the pit of success".
- **Minimal support burden**: We are a small team of volunteers supporting a large userbase. Adding configuration which is complicated or requires a lot of explanation generates support requests and we have limited time to answer them. Even if we provide good docs, lots of people don't read it.

Thinking through what we've got at the moment, I think the implementation is not that intuitive, even if we rename the params. My instinct is that being able to use a combination of `logoSize=auto` and some fixed witdth is useful but I'm not convinced that allowing one user to have `logoSize=auto&logoWidth=50`, another to have `logoSize=auto&logoWidth=60` and another to have `logoSize=auto&logoWidth=70` is that valuable, and it goes against the goal of consistency. This also allows the user to do wierd or undesirable things like https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amg&logoSize=auto&logoHeight=400

So just zooming back out to the original user need..
The problem we are trying to solve here is really: Some logos are wide and thin, and they look bad in the default viewbox. The original problem we were trying to solve was not so much an appetite for limitless customisation.

Given that, I think my suggestion is that we try and limit this to some "presets". Maybe

- nothing/default
- `logoSize=auto`
- `logoSize=wide` (a fixed-width that looks good with wide logos, probably something like `logoSize=auto&logoWidth=40` in the current implementation)

but remove the ability for the user to explicitly set the viewbox numerically, and don't worry too much about user-supplied logos for now.

Thoughts?

LitoMore: I think we can limit it's max-width to the full-filled-width.

So the https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amg&logoSize=auto&logoHeight=400 will still be: https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amg&logoSize=auto

So the code will be:

```diff
-logoWidth = (iconSize.width / iconSize.height) * (logoHeight || DEFAULT_LOGO_HEIGHT)
+const autoLogoWidth = (iconSize.width / iconSize.height) * (logoHeight || DEFAULT_LOGO_HEIGHT)
+logoWidth = logoWidth > autoLogoWidth ? autoLogoWidth : logoWidth
```

What do you think?

chris48s: That constraint would help with the 'wierd or undesirable' outputs, but I think the other points still stand.

LitoMore: The `logoWidth: 'auto' | 'wide'` doesn't display icons well in all scenarios.

The most wide icon is `comsol`. The logo still looks so small even we set width to 40. So it's hard to find out some preset values for icons.

I think it is still necessary to choose the appropriate width according to their icons and the context of the icon and text content.

 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=cosmol&logoSize=auto&logoWidth=40
 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amg&logoSize=auto&logoWidth=40
 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amg&logoSize=auto&logoWidth=40

The `logoHeight` is also useful, you can understand the icon with 14 zoom levels. The `1` for the minimum size, `7` for the middle size, and `14` for the max size.

 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=cosmol&logoSize=auto&logoHeight=7
 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amg&logoSize=auto&logoHeihgt=7
 https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=amd&logoSize=auto&logoHeihgt=7

This function is convenient for many people to personalize his badge wall, and it is not a complex function for our maintainers to maintain.

chris48s: For anyone confused by the above links https://pr-9191-badges-shields.fly.dev/badge/foo-bar-blue?logo=comsol&logoSize=auto&logoWidth=40 is the right link :smile: 

> Personally I think it is not a complex function to maintain.

I agree it isn't a difficult feature to maintain in the sense that I don't think we would be constantly updating this code over time. I'm sure you'll agree replying to support requests is a form of maintenance though so I'm keen that we find a solution that is intuitive.

---

@calebcartwright  @PyvesB  Either of you got any thoughts on this one?

If the review app has been cleaned up, you can re-deploy it:

Screenshot at 2023-06-23 19-30-19

PyvesB: I've not read through all the details in this PR, and the review app is failing to deploy again. I'll try to circle back to this tomorrow.

Once thing I want to bring back to our attention is https://github.com/badges/shields/issues/3635. There was quite a lot of heated debate at the time about how we had allowed for inconsistent logo positioning and sizing with the popout style. The resolution plan also included removal of the `logoPosition` code at the time, but we apparently didn't do it. We need to be careful if we want to reintroduce new options related to logo appearance.

chris48s: Yes I am feeling echoes of the popout style here and keen find a way to solve this without going down the same route.
I think the fact that `logoWidth` still exists is a remnant of that.

The reason why the review app is not deploying is because of the merge conflicts (side note: I'm going to raise an issue about that). In order to get this deploying again, we'll need to resolve the conflicts. Are you able to have a look at that @LitoMore ? :pray:

LitoMore: I have rebased with the latest master. It should be working now. I have no permission to trigger workflows.

LitoMore: I also created a PR to adjust the `deploy-review-app` script.

- https://github.com/badges/shields/pull/9317

LitoMore: I can revert the exposed `logoWidth` `logoHeight` codes and wait for other users' feedback if we still have reservations about this.

calebcartwright: > @calebcartwright @PyvesB Either of you got any thoughts on this one?

I confess I probably need to go back and do a thorough read through the motivating request/issue in #8372 and the full commentary. However, my current impression is that we've started drifting from the "how" (the implementation) and are circling back to the "what" and "why"

The way you described this @chris48s above:

> So just zooming back out to the original user need..
The problem we are trying to solve here is really: Some logos are wide and thin, and they look bad in the default viewbox. The original problem we were trying to solve was not so much an appetite for limitless customisation.

Still resonates with me, and that seems like a valid thing we should consider whether we want to address (and if so, be able to articulate why).

As many will likely remember, I'm quite adamantly opposed to infinite customization. However, I don't inherently currently view attempts to address that user need as being contradictory with the goals of the project, nor as falling into that "too much customization" bucket (although I recognize certain implementations/solutions _could_).

I.e. I suspect there's a universe in which we could potentially address the user need that would still align with our goals and self-imposed constraints around where we draw the customization line

chris48s: > I can revert the exposed logoWidth logoHeight codes and wait for other users' feedback if we still have reservations about this.

Yes. An incremental approach is one way forward. We could add the `logoSize=auto` param as a first step and then see where we think we should go next based on user feedback. Adding that doesn't prevent us adding more params later. Once we have added params or settings, removing or changing them is harder: We can't really version a service like we can with a library.

calebcartwright:
Recognize this has some conflicts that would be blockers, but in general, where did we land as far as our target strategic direction with this one?

chris48s:
> where did we land as far as our target strategic direction with this one?

My last comment was

> add the `logoSize=auto` param as a first step and then see where we think we should go next based on user feedback. Adding that doesn't prevent us adding more params later. Once we have added params or settings, removing or changing them is harder: We can't really version a service like we can with a library.

and P-Y :+1:d it You good with that?

calebcartwright:
> > where did we land as far as our target strategic direction with this one?
> 
> My last comment was
> 
> > add the `logoSize=auto` param as a first step and then see where we think we should go next based on user feedback. Adding that doesn't prevent us adding more params later. Once we have added params or settings, removing or changing them is harder: We can't really version a service like we can with a library.
> 
> and P-Y +1d it You good with that?

Ah yes, it's coming back to me now. This was one you and I discussed synchronously on the team call and I'd intended to comment indicating my :+1: but forgot to do so. Apologies for the noise, but yes indeed, I'm :+1: with that

LitoMore:
I've resolved conflicts and rebased with the latest code.

calebcartwright:
Seems CI is still rather unhappy. One is just the linter, and a quick `npm run prettier` should fix, but the others failing jobs have test failures, e.g.:

https://github.com/badges/shields/actions/runs/6156362420?pr=9191

> AssertionError: expected { openapi: '3.0.0', …(3) } to deeply equal { openapi: '3.0.0', …(3) }
    at Context.<anonymous> (file:///home/runner/work/shields/shields/core/base-service/openapi.spec.js:411:15)
    at process.processImmediate (node:internal/timers:476:21)

LitoMore:
@calebcartwright CI passed now. Please review.

chris48s:
I checked this out locally. I'm seeing that `logoHeight` is still exposed as a URL param
http://localhost:8080/badge/gaming-ED1C24?logo=amd&labelColor=555&logoSize=auto&logoHeight=1
http://localhost:8080/badge/gaming-ED1C24?logo=amd&labelColor=555&logoSize=auto&logoHeight=100

LitoMore:
@chris48s The exposed `logoHeight` should be removed now.

lib/logos.js
```diff
@@ -88,25 +88,38 @@ function getSimpleIconStyle({ icon, style }) {
   return 'default'
 }
 
-function getSimpleIcon({ name, color, style }) {
+function getSimpleIcon({ name, color, style, size = 'contain' }) {
```
chris48s, line 91: It is odd that the default value here is "contain", but then we never reference "contain" anywhere (because it basically means "don't change anything").

First, I wonder if something like "square" might be a better name for the default (open to suggestions on this), or we could just make the default `null`?

Second, can we add a comment in the function explaining what the 2 possible values are and what they mean.

core/base-service/openapi.js
```diff
@@ -270,6 +272,27 @@ function category2openapi(category, services) {
...
+        logoWidth: {
+          name: 'logoWidth',
+          in: 'query',
+          required: false,
+          description: 'The width of the logo, default to `14`',
+          schema: {
+            type: 'string',
+          },
+          example: '20',
+        },
```
chris48s, line 295: There has been some discussion in this PR about the `logoWidth` param, which already exists (as a hangover from the "popout" style) but is not documented. For this PR, lets maintain the current situation and not add it to the docs. Lets leave what we're doing about `logoWidth` for another day and remove this. Thanks.

services/endpoint/endpoint.service.js
```diff
@@ -103,6 +103,21 @@ const description = `<p>
...
+          It's useful for some wider logos like <code>amd</code> and <code>amg</code>.
+        </td>
+      </tr>
+      <tr>
+        <td><code>logoWidth</code></td>
+        <td>
+          Default: 14. Same meaning as the query string. Can be overridden by
+          the query string.
+        </td>
+      </tr>
```
chris48s, line 120: same here - lets remove `logoWidth` for this PR

services/endpoint/endpoint.service.js
```diff
@@ -103,6 +103,21 @@ const description = `<p>
           the query string.
         </td>
       </tr>
+      <tr>
+        <td><code>logoSize</code></td>
```
chris48s, line 107: We've added this to the endpoint service docs, but it won't actually work. The `render()` method has not been updated to do anything with it.

core/base-service/openapi.js
```diff
@@ -270,6 +272,27 @@ function category2openapi(category, services) {
           },
           example: 'violet',
         },
+        logoSize: {
+          name: 'logoSize',
+          in: 'query',
+          required: false,
+          description:
+            "Make icons adaptively resize by setting `auto`. It's useful for some wider logos like `amd` and `amg`.",
```
chris48s, line 280: Can we include in the docs that this only works for simple-icons logos.

lib/logos.js
```diff
@@ -88,25 +88,38 @@ function getSimpleIconStyle({ icon, style }) {
...
+        .replace('viewBox="0 0 24 24"', `viewBox="0 0 24 ${iconHeight}"`)
+        .replace(/<path d=".*"\/>/, `<path d="${path}"/>`)
+    }
   }
+
+  return svg2base64(iconSvg)
 }
 
-function prepareNamedLogo({ name, color, style }) {
+function prepareNamedLogo({ name, color, style, size }) {
```
chris48s, line 122: Now that we're pretty settled on the implementation, could we add a test case for the `auto` style to https://github.com/badges/shields/blob/2a417804d54626938d6fbcbeb9659315a2f98ce2/lib/logos.spec.js#L33

chris48s (commented):
Thanks. I think we've now got the behaviour to expose settled.

I've now had a proper look over the code and flagged some points we'll need to get settled before we can merge. It is mostly quite minor stuff now though.

chris48s (approved):
Thanks for following up - latest changes LGTM. I've pushed another commit just resolving the merge conflicts on package.json and package-lock.json

chris48s:
Quick FYI.
I deployed this to production today, but hit a bit of an issue. Its unrelated to this PR, but basically this happened https://github.com/badges/shields/issues/10077 but in production this time.

There's nothing you can do to help - its an issue with fly.io - but in order to get things stable I had to roll back to the previous release. I'll try deploying again later in the week, but just wanted to clarify what happened because this was deployed for a bit and then rolled back.

Issue #9173, opened 2023-05-16
Ability to apply a filter when we are picking a string from a list of strings


chris48s:
There are a couple of open PRs relating to filtering github tag/release badges.

One is about filtering github tag and release badges to only include tags that start with a prefix https://github.com/badges/shields/pull/8205
One is about excluding tags that contain some substring https://github.com/badges/shields/pull/9076

There is also an open issue relating to filtering dockerhub tags by a prefix https://github.com/badges/shields/issues/8789

We've also got some badges where we've implemented some kind of prefix or suffix filter that may or may not be working well https://github.com/badges/shields/issues/8714

See also https://github.com/badges/shields/issues/6275

In general, it seems like there is enough demand for this kind of thing that it is worth us thinking about, but I'm not overly keen to merge either of the currently open PRs. Maybe we add the ability to filter only github tags that start with a prefix, or that don't include a substring. Then someone else comes along and wants to only filter releases that end with a suffix, or we want to apply it to other badges. Feels like we could be heading down a messy road.

I'd rather find a more general solution that allows us the user to say "all the releases that end with the suffix `-ubuntu`" by specifying some kind of filter query like `?filter=%2A-ubuntu` which we can apply to an array of strings anywhere where the task boils down to "pick a string from an array of strings and shove it on a badge". Then that gives us a flexible pattern we can apply in multiple places as this request comes up.

What I **don't** want to do is:

- Accept an arbitrary user-supplied regex. Evaluating a user-supplied regex opens us up to a ReDos attack. It is fairly easy for a user to supply a regex which causes us to either spend a long time evaluating it or consume a large amount of memory evaluating it. Regex is powerful, but it is overkill for this task.
- Invent our own query language/parser. It should be possible to adopt some kind of existing solution to this problem.

So ideally we want to find some kind of "string with wildcards or placeholders" that satisfies the following criteria:

- At least expressive enough to allow a user to convey stuff like:
  - Does/does not include substring
  - Does/does not start with prefix
  - Does/does not end with suffix
- Less capable than regex - won't open us up to ReDos or RCE vulns by evaluating user-supplied filters
- Has an existing javascript implementation

One option I think is worth evaluating is format strings ( https://www.npmjs.com/package/scanf seems like a reasonable implementation)

Glob patterns also spring to mind as a possibility, but they are specifically for paths, and we're mostly not going to be dealing with paths. It is not quite the right fit.

There are also some very widely used wildcard matching libraries:

[...]

Issue #9080, opened 2023-04-14
Allow GitHub tags matching a pattern to be excluded from latest tag


slaclau:
I have submitted a pull request for this - comments welcome and appreciated.

chris48s:
Hello. We have just shipped a new feature which allows you to pass an arbitrary filter to the github tag and release badges. For example

The `filter` param can be used to apply a filter to the project's tag or release names before selecting the latest from the list. Two constructs are available: `*` is a wildcard matching zero or more characters, and if the pattern starts with a `!`, the whole pattern is negated.

e.g:
https://img.shields.io/github/v/tag/badges/shields
https://img.shields.io/github/v/tag/badges/shields?filter=%21server-%2A

This feature should be powerful and flexible enough to enable this use-case (and a bunch of others).

PR #9193, opened 2023-05-22
Allow user to supply an arbitrary filter param to [GithubTag] and [GithubRelease] badges

chris48s:
This PR implements the solution discussed in #9173 and applies it to the GitHub Tag and GitHub Release badges, although we could apply the same approach to others in future.

We've talked through the reasons and use-cases for this, but this provides a fairly flexible solution which enables users to do things like `/github/v/tag/badges/shields?filter=%21server-%2A` in monorepos and other situations where repos contain a mix of different tags/releases.

Closes https://github.com/badges/shields/discussions/6253
Closes https://github.com/badges/shields/issues/6275
Closes https://github.com/badges/shields/pull/8205
Closes https://github.com/badges/shields/pull/9076
Closes https://github.com/badges/shields/issues/9080
Closes https://github.com/badges/shields/issues/9173

@calebcartwright  - Although it would be nice to get this merged, if you've got time to review stuff, I am more keen to get https://github.com/badges/shields/pull/9014 over the line tbh.


services/github/github-tag.spec.js
```diff
@@ -53,4 +53,24 @@ describe('GithubTag', function () {
...
+
+  test(GithubTag.applyFilter, () => {
+    const tags = ['v1.1.0', 'v1.2.0', 'server-2022-01-01']
+    given({ tags, filter: undefined }).expect(tags)
+    given({ tags, filter: '' }).expect(tags)
+    given({ tags, filter: '*' }).expect(tags)
+    given({ tags, filter: '!*' }).expect([])
+    given({ tags, filter: 'foo' }).expect([])
+    given({ tags, filter: 'server-*' }).expect(['server-2022-01-01'])
+    given({ tags, filter: '!server-*' }).expect(['v1.1.0', 'v1.2.0'])
```
chris48s, line 74: I haven't really gone to town on testing different filters here. `matcher` has its own test suite.

I also haven't added any additional tests at the service layer. Just unit tests.

services/github/github-release.tester.js
```diff
@@ -32,7 +32,7 @@ t.create('Release (No releases)')
 
 t.create('Prerelease (No releases)')
   .get('/v/release/badges/daily-tests.json?include_prereleases')
-  .expectBadge({ label: 'release', message: 'no releases' })
+  .expectBadge({ label: 'release', message: 'no releases found' })
```
chris48s, line 35: I've taken the opportunity here to align with the tag error message.

services/github/github-release.service.js
```diff
@@ -85,6 +86,21 @@ class GithubRelease extends GithubAuthV3Service {
...
+        sort: 'date',
+        display_name: 'release',
+        filter: '*Open Beta',
+      },
+      staticPreview: this.render({
+        version: 'Release v6.0.2 - Open Beta',
+        sort: 'date',
+        isPrerelease: false,
+      }),
+      documentation: documentation + filterDocs,
```
chris48s, line 102: The move to OpenAPI in #9014 and follow-on PRs is going to make it way easier to document services that have lots of optional query params and provide documentation relevant to individual parameters.

services/github/github-common-release.js
```diff
@@ -68,8 +69,39 @@ function getLatestRelease({ releases, sort, includePrereleases }) {
...
+  or more characters, and if the pattern starts with a <code>!</code>,
+  the whole pattern is negated.
+  </p>
+</div>`
+
+function applyFilter({ releases, filter, displayName }) {
+  if (!filter) {
+    return releases
+  }
+  if (displayName === 'tag') {
```
chris48s, line 89: Here we apply the filter to either the release names or tag names depending on which is going to get displayed on the badge

chris48s:
The socket.dev issues are false positive. Turns out there are a lot of similarly named packages called lodash.something :)

services/github/github-tag.spec.js
calebcartwright: I agree with the approach, we don't need to re-test the internals of a 3rd party library, just a couple of integration-like tests that give us confidence we're using that lib correctly.

Think you've fully cleared that threshold here :+1:

calebcartwright (approved):
lgtm, will defer to you on when to merge as I know there's some other PRs that you may want to juggle (e.g. ones touching the lockfile, ones implementing similar end results, etc.)

PR #9197, opened 2023-05-26
Add docstrings for validators service

pr7prashant:
- Added docstrings for `validators` service


services/validators.js
```diff
@@ -1,25 +1,79 @@
...
 export const optionalDottedVersionNClausesWithOptionalSuffix =
   Joi.string().regex(/^\d+(\.\d+)*([-+].*)?$/)
-// TODO This accepts URLs with query strings and fragments, which for some
-// purposes should be rejected.
+
+/**
+ * TODO: Joi validator that accepts URLs with query strings and fragments, which for some purposes should be rejected.
+ *
+ * @type {Joi}
+ */
```
chris48s, line 68:
```suggestion
/**
 * Joi validator that checks if a value is a URL
 *
 * TODO: This accepts URLs with query strings and fragments, which for some purposes should be rejected.
 *
 * @type {Joi}
 */
```

chris48s (approved):
Cheers! Merging with one edit

PR #9233, opened 2023-06-02
WIP Implement a pattern for dealing with upstream APIs which are slow on the first hit; run [*****]

chris48s:

### DIFF
--- a/core/base-service/base-json.js
+++ b/core/base-service/base-json.js
@@ -34,10 +34,22 @@ class BaseJsonService extends BaseService {
    *    and custom error messages e.g: `{ 404: 'package not found' }`.
    *    This can be used to extend or override the
    *    [default](https://github.com/badges/shields/blob/master/core/base-service/check-error-response.js#L5)
+   * @param {object} [attrs.customExceptions={}] Key-value map of got network exception codes
+   *    and an object of params to pass when we construct an Inaccessible exception object
+   *    e.g: `{ ECONNRESET: { prettyMessage: 'connection reset' } }`.
+   *    See {@link https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#errorcodes got error codes}
+   *    for allowed keys
+   *    and {@link module:core/base-service/errors~RuntimeErrorProps} for allowed values
    * @returns {object} Parsed response
    * @see https://github.com/sindresorhus/got/blob/main/documentation/2-options.md
    */
-  async _requestJson({ schema, url, options = {}, errorMessages = {} }) {
+  async _requestJson({
+    schema,
+    url,
+    options = {},
+    errorMessages = {},
+    customExceptions = {},
+  }) {
     const mergedOptions = {
       ...{ headers: { Accept: 'application/json' } },
       ...options,
@@ -46,6 +58,7 @@ class BaseJsonService extends BaseService {
       url,
       options: mergedOptions,
       errorMessages,
+      customExceptions,
     })
     const json = this._parseJson(buffer)
     return this.constructor._validate(json, schema)
--- a/core/base-service/cache-headers.js
+++ b/core/base-service/cache-headers.js
@@ -39,6 +39,7 @@ function coalesceCacheLength({
   assert(defaultCacheLengthSeconds !== undefined)
 
   const cacheLength = coalesce(
+    serviceOverrideCacheLengthSeconds,
     serviceDefaultCacheLengthSeconds,
     defaultCacheLengthSeconds
   )
@@ -46,7 +47,6 @@ function coalesceCacheLength({
   // Overrides can apply _more_ caching, but not less. Query param overriding
   // can request more overriding than service override, but not less.
   const candidateOverrides = [
-    serviceOverrideCacheLengthSeconds,
     overrideCacheLengthFromQueryParams(queryParams),
   ].filter(x => x !== undefined)
 
--- a/core/base-service/errors.js
+++ b/core/base-service/errors.js
@@ -42,6 +42,7 @@ class ShieldsRuntimeError extends Error {
     if (props.underlyingError) {
       this.stack = props.underlyingError.stack
     }
+    this.cacheSeconds = props.cacheSeconds
   }
 }
 
@@ -206,6 +207,9 @@ class Deprecated extends ShieldsRuntimeError {
  * @property {string} prettyMessage User-facing error message to override the
  * value of `defaultPrettyMessage()`. This is the text that will appear on the
  * badge when we catch and render the exception (Optional)
+ * @property {number} cacheSeconds Length of time to cache this error response
+ * for. Defaults to the cacheLength of the service class throwing the error
+ * (Optional)
  */
 
 export {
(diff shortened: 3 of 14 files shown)


core/base-service/base-json.js
```diff
@@ -34,10 +34,22 @@ class BaseJsonService extends BaseService {
...
    * @returns {object} Parsed response
    * @see https://github.com/sindresorhus/got/blob/main/documentation/2-options.md
    */
-  async _requestJson({ schema, url, options = {}, errorMessages = {} }) {
+  async _requestJson({
+    schema,
+    url,
+    options = {},
+    errorMessages = {},
+    customExceptions = {},
```
chris48s, line 51: `errorMessages` and `customExceptions` do somewhat similar but importantly different things. Naming things is hard.

core/base-service/errors.js
```diff
@@ -42,6 +42,7 @@ class ShieldsRuntimeError extends Error {
     if (props.underlyingError) {
       this.stack = props.underlyingError.stack
     }
+    this.cacheSeconds = props.cacheSeconds
```
chris48s, line 45: I've done this here so any `ShieldsRuntimeError` can have a custom `cacheSeconds` but at the moment we're only using it for `Inaccessible`

core/base-service/got.js
```diff
@@ -22,6 +27,12 @@ async function sendRequest(gotWrapper, url, options) {
         underlyingError: new Error('Maximum response size exceeded'),
       })
     }
+    if (err.code in customExceptions) {
+      throw new Inaccessible({
+        ...customExceptions[err.code],
+        underlyingError: err,
+      })
```
chris48s, line 34: Spreading in this order means you can't manually override `underlyingError`. I don't think it would ever make sense to do that.

services/endpoint/endpoint.service.js
```diff
@@ -178,7 +178,10 @@ export default class Endpoint extends BaseJsonService {
       logoWidth,
       logoPosition,
       style,
-      cacheSeconds,
+      // don't allow the user to set cacheSeconds any shorter than this._cacheLength
+      cacheSeconds: Math.max(
+        ...[this._cacheLength, cacheSeconds].filter(x => x !== undefined)
+      ),
```
chris48s, line 184: I can't leave a line comment on it because I've now changed it, but there is an existing endpoint service test covering this.

core/base-service/cache-headers.js
```diff
@@ -39,14 +39,14 @@ function coalesceCacheLength({
...
+    serviceOverrideCacheLengthSeconds,
     serviceDefaultCacheLengthSeconds,
     defaultCacheLengthSeconds
   )
 
   // Overrides can apply _more_ caching, but not less. Query param overriding
   // can request more overriding than service override, but not less.
   const candidateOverrides = [
-    serviceOverrideCacheLengthSeconds,
     overrideCacheLengthFromQueryParams(queryParams),
```
chris48s, line 50: For the history on this, see https://github.com/badges/shields/pull/2755

Basically, the only thing relying on the logic as it stood here was the endpoint badge (we don't want endpoint badge users to be able to set a lower `cacheSeconds` value than the service default). I've moved this logic so it gets applied in `endpoint.service.js` so that we can now lower `cacheSeconds` with a custom exception property if we want. If we left this as it was, we'd only be able to raise `cacheSeconds` with a custom exception property.

core/base-service/legacy-request-handler.spec.js
```diff
@@ -185,7 +185,7 @@ describe('The request handler', function () {
         )
 
         const { headers } = await got(`${baseUrl}/testing/123.json`)
-        expect(headers['cache-control']).to.equal('max-age=300, s-maxage=300')
+        expect(headers['cache-control']).to.equal('max-age=200, s-maxage=200')
```
calebcartwright, line 188: this made me do a double take on first pass, and I think it's because I was reading the `createFakeHandlerWithCacheLength` function name a _little_ too literally resulting in me thinking this showed that a user could set a _lower_ value with `cacheLength` (i.e. I think I just misinterpreted what the `200` up on line 177 was being used for

core/base-service/base-json.js
calebcartwright: ack yeah I see that being tricky down the road since both of these are (ultimately) used in the setting of error messages.

I feel like in a perfect world (one where no refactoring would be required), I'd be inclined to keep a single top level object for errors and maybe have it have two properties, one that would point to an object that's our standard http response status code to error message dictionary we have today, and the other would be the new I feel like a single top level object that had two properties, one being new dictionary that maps nodejs error codes to a desired error message.

Thinking on a bit more, I suppose having those two mappings separate as you do here is fine; I think it's just the first having squatted on the highly generic `errorMessages` name that feels problematic.

What if we come up with a more descriptive name for this one for now, e.g. (overly verbosely) `nodeErrorCodeToShieldsExceptions` and then we could potentially consider renaming `errorMessages` to something similarly more descriptive down the road?

calebcartwright (approved):
One thought left inline below wrt the names with one suggested change that could accompany this PR and another (potential) follow up. Don't feel terribly strongly though and wouldn't object to this moving ahead as-is with both of those items punted till later, or even if we never do either

core/base-service/base-json.js
chris48s: Yeah broadly agreed. Just thinking about the ergonomics of consuming this at the service layer...

Mapping HTTP error codes to error messages is really common - we do stuff like

```js
errorMessages: { 404: 'package not found' }
```

all over the place, so lets keep that easy to do in a convenient way.

I've designed this feature (provisionally lets call it `customExceptions`) to be relatively generic, but I'd expect us to actually use this quite rarely.

I think given that, I'd prefer to keep a flatter structure and avoid making `errorMessages` a nested object. In the vast majority of cases we would only care about one of the keys. If we can come up with 2 good names we would need to change the code in loads of places anyway, so lets make those two good names the top-level names rather than introduce another layer of nesting.

i.e: If we need to go and change a load of existing service classes, I'd rather do a big find and replace changing

```js
errorMessages: { 404: 'package not found' }
```

to

```js
httpErrors: { 404: 'package not found' }
```

than

```js
errorMessages: { httpErrors: { 404: 'package not found' } }
```

or whatever.

If we ignore the fact that there is loads of existing code using `errorMessages` and we were naming these two concepts from scratch today, what do you think would be the ideal names/API for these things? I guess the things we currently call `errorMessages` are really HTTP Status Codes. The thing I've called `customExceptions` are really System Error Codes. How about `httpErrors` and `systemErrors` ?

calebcartwright: > How about httpErrors and systemErrors ?

Yup, love it. Definitely clears up the naming overlap and is much more succinct than the ones I threw out there

calebcartwright (approved):
fwiw this is a case where I'm glad we took a moment to bikeshed on names; I really like the new names and think they provide an additional level of clarity that's useful above and beyond this particular new pattern

chris48s:
Agreed :+1:

Issue #9236, opened 2023-06-04
jsDelivr shield falsely shows 0/month


adamlui:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

At times (like the time of this writing) the jsDelivr shield falsely shows 0 hits/month:
![image](https://github.com/badges/shields/assets/10906554/f2d5a6b0-9643-429b-a91f-bc820e186a3c)

### 🔗 Link to the badge

https://img.shields.io/jsdelivr/gh/hm/chatgptjs/chatgpt.js?color=ff6427&label=jsDelivr+Requests

### 💡 Possible Solution

Show 'unavailable' if reponse = 0

adamlui:
I just noticed sometime after passing midnight PST 16 mins ago, the counter works again, so it seems like a rate limit issue

chris48s:
This sounds more like a transient issue with the upstream API tbh. If we were getting rate limited, I would expect the upstream API to return a `429 Too Many Requests` (which would cause us to show an error badge, rather than report zero). We'd only render a badge saying `0/month` if we get back a `200 OK` response saying `{ "total": 0 }`.

Also the jsDelivr API docs say

> The API is free to use and imposes no rate limits


I wonder if @MartinKolarik can shed any light on this one?

MartinKolarik:
Yes, there was an issue with the data availability earlier today, which resulted in zero totals being reported.

adamlui:
@MartinKolarik is it an issue with jsDelivr? If so is it common? (Because weeks ago I had tested this shield on shields.io but it showed 0 so I thought it doesn't work with GitHub libs)

adamlui:
@chris48s i just noticed your gravatar looks like Heisenberg 😄

MartinKolarik:
It seems to happen occasionally around midnight, I opened a task to fix it. Availability of the library is not affected, and the stats are always collected as well, it's just an issue with generating reports for the API.

chris48s:
I'm going to close this issue as there is a fix in progress with the upstream service. Thanks!

adamlui:
@MartinKolarik fyi it happened a few mins ago

MartinKolarik:
Yes, I expect a fix in about two weeks.

MartinKolarik:
We've made some improvements that should prevent this from happening and are further monitoring the stats to confirm 100% availability.

adamlui:
[...]

Issue #8605, opened 2022-11-09
Youtube badge fails intermittently


Jetpack-Cat:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

My youtube badge suddenly says invalid, it appears youtube changed something. Making a new one still shows invalid. It may have something to do with the new youtube handles for channels. I believe it was working before I accepted the @handle youtube suggested

### 🔗 Link to the badge

You'll see it 
https://www.curseforge.com/minecraft/modpacks/jetpack-cat

### 💡 Possible Solution

I believe youtube changed something and made this stop working, because nothing has changed on that page for months and suddenly it's invalid

PaulaBarszcz:
I cannot see the yt badges on the link that you provided (https://www.curseforge.com/minecraft/modpacks/jetpack-cat), you probably removed them if they stopped working :P 

Which badge(s) were you using for youtube, that are not working now?
![image](https://user-images.githubusercontent.com/31891675/201084986-ea38baa6-64fe-493a-8f8f-77f3eab97c69.png)
https://shields.io/

DenverCoder1:
It appears to working now

DenverCoder1:
> I cannot see the yt badges on the link that you provided

They have this badge still at that link

![](https://img.shields.io/youtube/channel/subscribers/UCpbZjb0o2Nx9MTrIOjeuL_w?labelColor=030B14&color=F7F3F5&logo=youtube&logoColor=white&style=for-the-badge)

It was showing all grey and as "invalid" yesterday but it appears to be working now.

Same happened with the ones on my profile.

chris48s:
I wasn't able to catch this and see what the API response from youtibe was when it wasn't working. I'm going to close this for now on the assumption it was a transient issue but if it happens again, re-open. Cheers.

DenverCoder1:
It seems like the issue is back now.

Jetpack-Cat:
It happens intermittently.

DenverCoder1:
Yeah, seems to be working again now

chris48s:
https://github.com/badges/shields/issues/8758

programmingwithalex:
I am encountering the same issue showing invalid. I get the invalid response on https://shields.io/category/social and on my GitHub profile page.

kassbohm:
Same issue here. Using https://shields.io/category/social leads to **invalid** for any youtube video I tested...

benderillo:
[...]

PR #9238, opened 2023-06-04
increase default cache on youtube badges

chris48s:
refs #9236
refs #9237

I've still not managed to catch this happening. If we think it is rate limit-related, lets bump the max age up on these.


calebcartwright:
Good move updating this. I think I managed to catch it once or twice, and feel pretty confidently it's rate-limit driven based on some digging I did around the time I posted https://github.com/badges/shields/issues/8605#issuecomment-1455163387

PR #9251, opened 2023-06-10
feat: Add author filter option for [GithubCommitActivity]

### LINKED ISSUE #9215: New filter option for CommitActivity
:clipboard: **Description**

I would like to show my commit activity in a specific repo on my github profile readme using a shield.

To do that i want to modify the comit activity shield to enable filtering by user.

I suggest the label to read "commits by [user]" for total amount and "commit activity by [user]" for anything else (only if a user filter is selected).

I think that adding the filtered username to the path will be confusing as commit activity already includes the repo's owner in the path. I suggest adding it in an argument.

Here i present my vision with examples:
"https://img.shields.io/github/commit-activity/m/eslint/eslint?userFilter=exampleUser"
image
"https://img.shields.io/github/commit-activity/t/eslint/eslint?userFilter=exampleUser"
image

i suggest to modify the UI as follows:

image

I recently added total commits to the commitActivity shield and i don't mind adding this as well if you will have me.

I opened an issue to discuess those changes first please let me know if this looks alright and aligns with the project's goals.

jNullj:
Add a new filter option to [GithubCommitActivity], allowing users to filter the commit activity by a specific author.

To make the filter more explicit, The label display "commits by [author]" for the total amount of commits and "commit activity by [author]" for other intervals when an author filter is selected.

To maintain a clear and organized code structure, The filtered author is added as an argument and not to the shield path.

The request to find the number of commits by the author is made using the REST api rather then the GraphQL api to make it in 1 request rather then 2.

Resolves #9215


jNullj:
I am ready for a review! Hope i got some quality on first attempt.

jNullj:
Just to make it a bit nicer to see behavior before running this, here is a short video of how it looks like on my testing env. (note, i used node 20)


https://github.com/badges/shields/assets/15849761/8714e544-e9a3-4f1d-95f5-79c9a1c87bcd

services/github/github-commit-activity.service.js
```diff
@@ -18,11 +23,16 @@ const schema = Joi.object({
...
+
 export default class GitHubCommitActivity extends GithubAuthV4Service {
   static category = 'activity'
   static route = {
     base: 'github/commit-activity',
     pattern: ':interval(t|y|m|4w|w)/:user/:repo/:branch*',
+    queryParamSchema,
   }
 
   static examples = [
```
chris48s, line 38: The example badges are showing "undefined" as the label. Are you able to have a look at this?

services/github/github-commit-activity.service.js
```diff
@@ -31,6 +41,7 @@ export default class GitHubCommitActivity extends GithubAuthV4Service {
       // Override the pattern to omit the deprecated interval "4w".
       pattern: ':interval(t|y|m|w)/:user/:repo',
       namedParams: { interval: 'm', user: 'eslint', repo: 'eslint' },
+      queryParams: { authorFilter: 'chris48s' },
```
chris48s, line 44: Just to make the example "make sense", can we pick a username of someone who has made one or more commits to the example repos. e.g: for eslint/eslint, `nzakas` would make more sense as an example.

services/github/github-commit-activity.service.js
```diff
@@ -115,6 +152,20 @@ export default class GitHubCommitActivity extends GithubAuthV4Service {
     return repo.history.totalCount
   }
 
+  static transformAuthorFilter({ res, buffer }) {
+    if (buffer.message === 'Not Found') {
+      throw new InvalidResponse({ prettyMessage: 'invalid branch' })
+    }
```
chris48s, line 158: Do we ever hit this line?
If I call something like  `/github/commit-activity/t/badges/shields/does-not-exist?authorFilter=chris48s` it tells me "repo not found". If we can't distinguish between branch not found and repo not found then we can just make the other error "repo or branch not found" (this is quite common).

While I was testing this, I also noticed that the branch not found case doesn't seem to be hanled correctly. It is not new in this PR (already exists in prod https://shields.io/github/commit-activity/t/badges/shields/does-not-exist ), but if you have a chance to look at it while you are looking at this, that would be amazing.

Given neither of these work but the spec tests pass, I'd suggest we move the "branch not found" test cases out to the service tests layer.

services/github/github-commit-activity.service.js
jNullj: I will pick one of the users for the example, no problem

services/github/github-commit-activity.service.js
jNullj: After years of debugging i figured the issue.
defaultBadgeData is defined *after* examples..., becouse of that, the value of the parent calss is used (BaseService) which is empty object {}.
Moving the static defenition of defaultBadgeData  before examples solves the issue...
But if i change the order eslint is mad at me for breaking the expected order of static properties *sigh*.
Should i change the order in the base service class? is there a reason for this order?

chris48s: Just to avoid this becoming a gigantic refactoring job that changes every service to add one badge, I think the path of least resistance to getting this merged is to change one line in render from

```
const label = interval === 't' ? 'commits' : this.defaultBadgeData.label
```

to

```
const label = interval === 't' ? 'commits' : 'commit activity'
```

We can live with a duplicated string here.

There's some frontend work in progress in https://github.com/badges/shields/pull/9014 which will make a lot of this irrelevant soon anyway.

services/github/github-commit-activity.service.js
chris48s: Hmm. I am not following. What do you mean when you say

> You are correct as well about the `interval='t'` not being used

?

This behaves the same way if I call https://shields.io/github/commit-activity/m/badges/shields/does-not-exist

services/github/github-commit-activity.service.js
```diff
@@ -103,6 +116,30 @@ export default class GitHubCommitActivity extends GithubAuthV4Service {
...
+      url: `/repos/${user}/${repo}/commits`,
+      options: {
+        searchParams: {
+          sha: branch,
+          author: authorFilter,
+          per_page: '1',
+          since,
+        },
+      },
+      errorMessages: errorMessagesFor('repo not found'),
```
chris48s, line 139: Sorry. Another request. We don't often make changes to the interface of `BaseService`, but we've just merged https://github.com/badges/shields/pull/9233 which means this will need to become `httpErrors: httpErrorsFor('repo not found')`. Could you merge in master and update that too (sorry).

services/github/github-commit-activity.service.js
jNullj: My bad, there is no diffrence, no idea how i mixed this in my reply. I edited the reply as i was debugging.
Should have read it twice before posting.

services/github/github-commit-activity.service.js
chris48s: Thanks. Could we do the same on the other example (pick an author who has contributed to the repo)

services/github/github-commit-activity.service.js
chris48s: Could we add a couple of test cases to the service tests file (`.tester.js`) covering invalid branch (for both the GraphQL and Rest paths).

services/github/github-commit-activity.service.js
jNullj: Sure

services/github/github-commit-activity.service.js
jNullj: I separated the GraphQL test and fixed the error handling in PR #9258 as this is not related directly to this change, i think its better for history.
I will add another test for the REST path in this PR

chris48s:
I've merged #9258 Can you sort out the merge conflict.

I suspect once that conflict is resolved, that will be the last thing and this will be ready to merge too, but I'll give it another quick read over tomorrow.

Cheers

chris48s (approved):
LGTM :+1: thanks

PR #9254, opened 2023-06-11
Delete old deprecated services

PyvesB:
Deletes a bunch of old deprecated services, as per https://github.com/badges/shields/blob/master/doc/deprecating-badges.md#what-happens-next. All of these were deprecated in 2021 or early 2022.


doc/deprecating-badges.md
```diff
@@ -90,13 +90,17 @@ Past that point, all related code will be deleted, and a not found error will be
...
 - Cauditor
 - CocoaPods Apps
 - CocoaPods Downloads
 - Codetally
 - continuousphp
 - Coverity
+- David
+- dependabot
 - Dockbit
 - Dotnet Status
```
calebcartwright, line 105: It's a bit tricky for me to determine whether I think it's worth mentioning the Docker build service here :thinking: 

I believe what you've got here is probably for the best, since it could be more confusing than helpful to have it listed (i.e. we still have a "Docker Build" badge from the end user pov, the deprecated one just had the nuances of some legacy with the platform and an associated badge route)

calebcartwright (approved):
One inline thought, but lgtm either way. Happy to re-:+1: if you decide you want to make any changes

doc/deprecating-badges.md
PyvesB: I figured the potential for confusion probably outweighed the benefits, but anyway I'm not convinced this list is read through very often. Let's merge as is :)

PR #9280, opened 2023-06-17
add a couple missing spaces in docs

calebcartwright:
Was poking around the new frontend and noticed these. Knew that if I didn't fix them now I'd permanently forget about them

### DIFF
--- a/services/amo/amo-downloads.service.js
+++ b/services/amo/amo-downloads.service.js
@@ -4,7 +4,7 @@ import { BaseAmoService, keywords } from './amo-base.js'
 
 const documentation = `
 <p>
-  Previously <code>amo/d</code> provided a &ldquo;total downloads&rdquo; badge. However,
+  Previously <code>amo/d</code> provided a &ldquo;total downloads&rdquo; badge. However,&nbsp;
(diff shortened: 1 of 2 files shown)


chris48s:
I'm currently working on this exact issue. There are a number of them dotted about. Slowly working through them :)

I think inserting non-breaking spaces probably isn't the right solution for these because we do actually want a breaking space.

I think the way I'm going to handle these is to convert the problematic `documentation` strings to markdown instead of HTML (which doesn't have this issue where a space isn't inserted if we start a tag on the next line).

calebcartwright:
> I think inserting non-breaking spaces probably isn't the right solution for these because we do actually want a breaking space.

Yeah that's fair

> I think the way I'm going to handle these is to convert the problematic documentation strings to markdown instead of HTML (which doesn't have this issue where a space isn't inserted if we start a tag on the next line).

I'm not sure how big a haul that'll be, but sounds like a reasonable approach. I think the changes proposed here would still be an improvement over current state, but at the same time I don't think it's a severe enough issue to matter all that much.

Will defer to you on what to do with this; feel free to close if you'd rather just focus on the markdown conversion path

chris48s:
I reckon I'm on course to have the PR open in the next hour or so

PR #9283, opened 2023-06-18
set a sensible meta description

chris48s:
This is shown when a link to shields.io is "unfurled"


chris48s:
shoot first and ask questions later

PR #9171, opened 2023-05-13
deprecate [travis].org badges

chris48s:
closes https://github.com/badges/shields/issues/9071

This PR deprecates the travis .org build badges and travis PHP version but keeps the .com build badges

PR #9284, opened 2023-06-18
remove obsolete [travis] .org examples

chris48s:
We deprecated these in https://github.com/badges/shields/pull/9171 but I forgot to remove the .org examples (the class uses to cover both the .com and .org cases) so they are still (wrongly) shown in the frontend.


PyvesB (approved):
Thanks! 👍🏻

PR #9292, opened 2023-06-18
run core/service tests on node 18

chris48s:
We've been blocked on this for a looooong time. Ditching gatsby has unblocked us :tada:

We are currently deploying on Node 16, which will reach EOL and stop getting security patches in ~2 months on 11 Sep 2023
https://nodejs.org/en/blog/announcements/nodejs16-eol
Node 18 will get patches for another ~2 years
The current LTS release is Node 20.

I'm going to suggest we:

- Merge this, update the branch protection rules (to account for the changed workflow names) and continue to run the tests on Node 16/18
- Continue to deploy on Node 16 in the immediate term
- Start running on Node 18 in local dev
- Switch the docker image over to Node 18 some time in the next few weeks
- Try running the tests on Node 20 once we're prod is happy on Node 18

We don't have a completely working service test suite as a baseline for this (the daily tests are in a bit of a sorry state at the moment :grimacing: ) but the number of failures was the same as last night's daily test run, so that's a good sign :)


calebcartwright (approved):
sounds reasonable to me. going to defer merging this myself in case there's some sequencing with other prs/setting changes you have in mind

PR #9295, opened 2023-06-18
add simple-icons to homepage features

LitoMore:
### Preview

CleanShot 2023-06-19 at 01 15 52@2x

### Changes

- Add `Include [ICONS COUNT] brand icons` section
- Add `target="_blank"` to all links on homepage features

### DIFF
--- a/frontend/src/components/homepage-features.js
+++ b/frontend/src/components/homepage-features.js
@@ -1,6 +1,9 @@
 import React from 'react'
 import clsx from 'clsx'
+import * as icons from 'simple-icons'
 import styles from './homepage-features.module.css'
+
+const iconsCount = Object.keys(icons).length
 
 const FeatureList = [
   {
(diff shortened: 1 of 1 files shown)


chris48s:
Personally I am :-1: on promoting Simple Icons on the home page. There is already a link to it on every badge "detail" page in the `logo` param docs (which means there are over 1,200 links to simpleicons.org on the site as it is. I think that is plenty.

I'd rather keep the landing page for our own projects/links.

I'll leave this open for now in case another maintainer feels strongly this is something we want.

Changing the other links to `target="_blank"` seems reasonable. Feel free to open another PR for that.

calebcartwright:
I'd agree. My preference would be to keep the landing page as clean and concise as possible.

The logo aspect feels like more of a "how to" with Simple Icons being an implementation detail. I would in theory be receptive to having Awesome Badges listed there, but I like that in its current position (linked towards the bottom).

Never say never, but as such I can't think of anything else that I'd lean towards promoting/emphasizing in the main center pane

LitoMore:
Good point. Thank you!

PR #9300, opened 2023-06-19
chore(frontend): open homepage feature links in new tab

LitoMore:
Related to https://github.com/badges/shields/pull/9295#issuecomment-1596255289


chris48s (approved):
thanks :+1:

Issue #9278, opened 2023-06-17
Embed badge URL section overflows on smaller screens


uncenter:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

On smaller screens, the tab names overflow and overlap with the page content:
<img width="480" alt="image" src="https://github.com/badges/shields/assets/47499684/a15ecab5-43f4-436d-98ab-3f2155bb50cc">

### 🔗 Link to the badge

Visible on all badge pages.

calebcartwright:
Thanks for letting us know! This is the new frontend just rolled out today so I suspect we'll probably see a few things like this, and very helpful to get these reports

uncenter:
> Thanks for letting us know! This is the new frontend just rolled out today so I suspect we'll probably see a few things like this, and very helpful to get these reports

Yes! I've been watching from the sidelines for a while now as the PR was being updated and really glad to see it finally get merged this morning! Looking great so far.

chris48s:
I've had a quick look at this.

I think rather than override the styles in our repo I am going to try and contribute a patch upstream to https://github.com/cloud-annotations/docusaurus-openapi/

I need to spend a bit more time on it, but (mainly for my own reference), I think the patch is going to look something like:

```diff
diff --git a/packages/docusaurus-theme-openapi/src/theme/ApiDemoPanel/Curl/styles.module.css b/packages/docusaurus-theme-openapi/src/theme/ApiDemoPanel/Curl/styles.module.css
index af3c9e8..f868110 100644
--- a/packages/docusaurus-theme-openapi/src/theme/ApiDemoPanel/Curl/styles.module.css
+++ b/packages/docusaurus-theme-openapi/src/theme/ApiDemoPanel/Curl/styles.module.css
@@ -10,7 +10,8 @@
   border-radius: var(--openapi-card-border-radius)
     var(--openapi-card-border-radius) 2px 2px;
   display: flex;
-  justify-content: flex-end;
+  justify-content: center;
+  flex-wrap: wrap;
 }
 
 .buttonGroup button {
@@ -59,7 +60,6 @@
   /* border-radius: 0.25rem; */
 
   display: block;
-  width: 100%;
 
   /* margin: 2px; */
   margin: var(--margin);
```

uncenter:
What about this instead?

```diff
.buttonGroup {
  background: var(--openapi-card-background-color);
  color: var(--ifm-pre-color);
  font-family: var(--ifm-font-family-monospace);
[...]

PR #9302, opened 2023-06-22
temp fix: wrap code examples tabs in narrow browser windows

chris48s:
closes https://github.com/badges/shields/issues/9278

I've contributed a fix for this upstream at https://github.com/cloud-annotations/docusaurus-openapi/pull/250

Its not in a package release yet though, so while we wait for that I'm going to apply the fix here (by winning the selector specificity war). We can roll it back when docusaurus-theme-openapi@0.6.5 is available.


uncenter:
Yay!

Issue #9285, opened 2023-06-18
Convert all the existing `examples` arrays on service classes to `openApi` objects


chris48s:
:clipboard: **Description**

Now that we have launched the new frontend, the current situation we are in is:

- Service classes may either define an `openApi` spec directly, or a legacy `examples` array
- There are a small handful of service classes which define an `openApi` spec directly, for example https://github.com/badges/shields/blob/master/services/dynamic/dynamic-xml.service.js#L18
- Most services still define a legacy `examples` array
- For services defining a legacy `examples` array, we make some assumptions and [bash it](https://github.com/badges/shields/blob/30377f6f33f7506364911aa034bb6d9ad188111a/core/base-service/openapi.js) into an Open API spec at frontend build time as best we can (accepting there are some limitations)

We want to move to a situation where every service class defines an `openApi` object, so step one is to auto-convert the legacy `examples` array to an `openApi` object as best we can. Then we can start tweaking/improving from there.

This is going to be a big PR touching basically every .service.js file in the codebase.

chris48s:
I'm not going to raise individual issues for these tasks yet as they are all blocked on this, but the next steps after this will be:

- Delete all `examples` properties + remove any validation for it
- Make `openApi` property required/validated
- Delete all the code for bludgeoning legacy `examples` objects to OpenAPI objects
- Correctly document required query params
- Fix duplicated titles
- Remove unnecessary duplicates for query param variants
- Start adding param descriptions
- Update contributing docs for adding new services to reflect new frontend + documenting with `openApi` property instead of `examples`
- Look for ways we can reduce duplication between the service definition/schema and OpenAPI spec
- Implement some automated checks (be that Joi schema, CI checks - TBC) that help to prevent common issues

Some of those might end up getting done in the first PR. I'll split the rest out into issues once they are unblocked.

calebcartwright:
I want to be clear up front that I'm **not** arguing _against_ doing this. However, I would like to know what, if any, level of coupling risk exists here.

I.e. if we convert everything over to openapi and then decide we need to e.g. go back to old UI or something different, would we be able to easily continue using the existing openapi-based example listings defined in each service class, and/or centrally "convert" them at the frontend build time like we're doing today (a openapi -> foo translation vs. the current legacyExamples -> openapi)

chris48s:
[...]

PR #9320, opened 2023-06-26
Convert `examples` arrays to `openApi` objects (part 1)

chris48s:
I've been umming and aahing about how best to start approaching #9285

My first thought was that I would try and write a script to auto-convert everything, and then tweak from there, but I think that is going to end up with making several enormous PRs that touch ~400 files. I have written some very rough scripts to help with this process and give me a starting point. That said, having got this deep into it, I've decided that I'm just going to tackle this in small to medium-sized batches of services (auto-converting first where possible, then manually reviewing and tweaking from there) rather than multiple passes across the whole codebase. I will also look at adding appropriate tooling and abstractions to handle various cases as I go along and meet them.

This is batch one of... many. This one also establishes some of the functions we are going to need moving forward.


core/base-service/openapi.js
```diff
@@ -332,4 +332,34 @@ function category2openapi(category, services) {
...
+  return { name, in: 'path', required: true, schema, example, description }
+}
+
+function queryParam({
+  name,
+  example,
+  schema = { type: 'string' },
+  required = false,
+  description,
+}) {
```
chris48s, line 350: I think these two functions actually handle a lot of the de-duplication heavy lifting for us.

core/base-service/openapi.js
```diff
@@ -332,4 +332,34 @@ function category2openapi(category, services) {
...
+      name,
+      in: 'query',
+      required,
+      schema,
+      allowEmptyValue: true,
+      example,
+      description,
+    }
+  }
+  return { name, in: 'query', required, schema, example, description }
```
calebcartwright, line 362: does the order of the fields matter for consumers of this function? i know it's a huge nit, but if not, it seems like this could be deduped a bit since it's just `allowEmptyValue` being conditionally added to the returned object

services/amo/amo-downloads.service.js
```diff
@@ -12,15 +12,15 @@ class AmoWeeklyDownloads extends BaseAmoService {
...
-      title: 'Mozilla Add-on',
-      namedParams: { addonId: 'dustman' },
-      staticPreview: this.render({ downloads: 120 }),
-      keywords,
-      documentation,
+  static openApi = {
+    '/amo/dw/{addonId}': {
+      get: {
+        summary: 'Mozilla Add-on Downloads',
+        description: documentation,
```
calebcartwright, line 19: Another mega-nit, but do you think we should rename the `documentation` vars as part of these conversion efforts? I'm not partial to one name over the other, but I feel like the main, perhaps only, reason we chose `documentation` as the var name in the various service classes was because that's what the example objects needed

services/amo/amo-rating.service.js
```diff
@@ -1,27 +1,26 @@
...
 
-  static examples = [
-    {
-      title: 'Mozilla Add-on',
-      pattern: 'rating/:addonId',
-      namedParams: { addonId: 'dustman' },
-      staticPreview: this.render({ format: 'rating', rating: 4 }),
-      keywords,
+  static openApi = {
+    '/amo/rating/{addonId}': {
```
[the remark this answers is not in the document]
calebcartwright: I'm currently ambivalent on it. In cases like this where we had an enum-like route param in the pattern it doesn't seem like much of a change from what we were doing before.

However, in cases where we have examples relying on the previously defined route it does feel like we're introducing duplication. I think I can live with that though, so long as we've still got something automated that can be used to validate the validity of these routes, and be easy to check both locally and CI.

I.e. If we ever have a small PR that updates a route path, I don't want to have to rely on human reviewers to remember to check the spec (which may or may not be part of the diff)

services/ansible/ansible-role.service.js
```diff
@@ -47,15 +48,14 @@ class AnsibleGalaxyRoleName extends AnsibleGalaxyRole {
...
 
-  static examples = [
-    {
-      title: 'Ansible Role',
-      namedParams: { roleId: '3078' },
-      staticPreview: this.render({
-        name: 'ansible-roles.sublimetext3_packagecontrol',
-      }),
+  static openApi = {
+    '/ansible/role/d/{roleId}': {
```
calebcartwright, line 52: Is this right? Based on the route I wouldn't have expected the `/d/` portion here

services/appveyor/appveyor-build.service.js
```diff
@@ -1,23 +1,31 @@
...
-      namedParams: { user: 'gruntjs', repo: 'grunt', branch: 'master' },
-      staticPreview: this.render({ status: 'success' }),
+    '/appveyor/build/{user}/{repo}/{branch}': {
+      get: {
+        summary: 'AppVeyor (with branch)',
+        parameters: [
+          pathParam({ name: 'user', example: 'gruntjs' }),
+          pathParam({ name: 'repo', example: 'grunt' }),
+          pathParam({ name: 'branch', example: 'master' }),
+        ],
```
calebcartwright, line 25: This has me wondering if it'd be worth either adding another wrapper function around `pathParam` that takes an array, e.g. `pathParams` that'll iterate over the entry args and contain the call sites to `pathParam`.

That in turn would allow this to be transformed into something like:

```suggestion
       parameters: pathParams([
          { name: 'user', example: 'gruntjs' },
          { name: 'repo', example: 'grunt' },
          { name: 'branch', example: 'master' },
        ]),
````

My thinking is that might help minimize some of the rightward drift (fewer characters/columns after all), and the potential for prettier to want to do some multi-line wrapping to account for the width.

Thoughts?

**Review by calebcartwright — COMMENTED** [review]
Mostly lgtm, few inline thoughts and one question where I suspect a potential copy/pasta issue

**Comment by chris48s** [comment]
Thanks for providing feedback on this. There's some useful points here. I'll try to get into them in the next week or so.

File: services/appveyor/appveyor-build.service.js (hunk 6 above)
Comment (chris48s) [reply services/appveyor/appveyor-build.service.js:25]: Yes. Love it!

I'll add these for both `pathParams()` and `queryParams()`.

Having got into this...

One implementation of this would be that the plural variants take an array as you've suggested.
So the implementation would look like:

```js
function pathParams(params) {
  return params.map(param => pathParam(param))
}
```

..and then calling that would look like:

```js
pathParams([{ name: 'onlyParam', example: 'onlyValue' }])

pathParams([
  { name: 'param1', example: 'value1' },
  { name: 'param2', example: 'value2' },
])
```

The other option would be to have the plural variants take a spread as input instead of an array.
So then the implementation would look like:

```js
function pathParams(...params) {
  return params.map(param => pathParam(param))
}
```

..and calling that would look like:

```js
pathParams({ name: 'onlyParam', example: 'onlyValue' })

pathParams(
  { name: 'param1', example: 'value1' },
  { name: 'param2', example: 'value2' },
)
```

Having thought about it a bit, I prefer the second variant myself. What do you reckon?

Note to self: When you finish this off, these should also have DocStrings!

core/base-service/openapi.js
chris48s: I've pushed 5a5ceb4 to tidy this up

services/amo/amo-downloads.service.js
chris48s: Yeah agreed. Just calling the var `description` would be better. I'll probably do that next time I push more commits to the branch, but as you say, it is a style point, not a huge issue.

chris48s:
Hmm. For some reason builds do not appear to be running on this PR. Maybe merge conflicts. I have more commits to push to finish this so I will sort it then.

services/amo/amo-downloads.service.js
chris48s: Done in https://github.com/badges/shields/pull/9320/commits/b1de47f7cb85a4138cd6967af08a4de42189b06d

chris48s:
This one has been sitting around for a while, and I'd like to get some momentum back into this project. So I have made some decisions and pushed some commits:

- On https://github.com/badges/shields/pull/9320#discussion_r1258429498 I have decided to go with the version that uses a spread instead of the array - https://github.com/badges/shields/pull/9320/commits/8c669dd13ea14a680d27737c08903aa2fae80c15
- On https://github.com/badges/shields/pull/9320#discussion_r1258426044 I have decided to do nothing for the moment, but we can revisit it

I have also decided to scope creep this a little:
In response to https://github.com/badges/shields/issues/9393 another task I want to do as I go through and convert all these is to make sure every route/documentation page has a unique title to help the search results make sense. In https://github.com/badges/shields/pull/9320/commits/2d407709be2901f3e1791d3599477e8d6771950f I have also added a check to make sure that this gets done and stays done as we convert the examples.

For the moment I am going to leave this open, but if I've got more time to work on this in the next few days, I might just go rogue and merge it so I can make some more progress.

calebcartwright:
Sorry for the delay on a follow up review and response to discussion. It's on my list for this weekend but I've got some #day-job stuff that's likely going to eat up a lot of my free time this weekend unfortunately

services/amo/amo-rating.service.js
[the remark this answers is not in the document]
calebcartwright: Do you feel like the latter option might make it easier for us to more have a single example in the badge listing (e.g. a single listing for a downloads badge with the interval param vs. a listing per interval) if/when we want?

Or, do you anticipate we'd like to maintain individual listings per variant and then keep that component of the route static in each of the corresponding builders?

services/amo/amo-downloads.service.js
calebcartwright: I should've noted before this wasn't a blocker by any stretch, kinda a half question/suggestion, though I suspect it'll end up being a worthwhile change overall

services/appveyor/appveyor-build.service.js
calebcartwright: Sure, second approach is my preference too :+1:

calebcartwright:
> For the moment I am going to leave this open, but if I've got more time to work on this again in the next few days, I might just go rogue and merge it so I can make some more progress.

I think it's good to go as-is tbh. The only semi-open thing from my POV would be https://github.com/badges/shields/pull/9320#discussion_r1242334091 but as you noted it's more of a stylistic thing, and one which I think we could discuss elsewhere/in parallel to decide if/when it's something we want to do (I don't think this and all the other conversions need to be blocked on that)

services/amo/amo-rating.service.js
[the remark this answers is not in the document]
chris48s: I think in _most_ cases, probably optimising for fewer different pages/menu items in the menu on the left with more options to customise is going to be the better option.

Issue #9315, opened 2023-06-25
github actions workflow badge returns "unparseable svg response" for (some?) workflows containing slash


chris48s:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Calling a URL like https://img.shields.io/github/actions/workflow/status/msr8/tictactoe/pages/pages-build-deployment returns "unparseable svg response"

This is because when we call the upstream, we request

- https://github.com/msr8/tictactoe/actions/workflows/pages%2Fpages-build-deployment/badge.svg (which doesn't work) instead of
- https://github.com/msr8/tictactoe/actions/workflows/pages/pages-build-deployment/badge.svg (which does)

`pages/pages-build-deployment` is a "special" workflow name: It is the name github pages uses for its deploy workflow, but it doesn't actually correspond to a real workflow file.

I am surprised we have not seen this before with workflows in subdirectories, but now I look, I am struggling to find an example of a repo that has the workflows split into subdirectories. I might have to set up an example to test this.

### 🔗 Link to the badge

![](https://img.shields.io/github/actions/workflow/status/msr8/tictactoe/pages/pages-build-deployment)

https://img.shields.io/github/actions/workflow/status/msr8/tictactoe/pages/pages-build-deployment

### 💡 Possible Solution

Preserve slashes when passing the URL

PR #9322, opened 2023-06-26
don't URL encode slash in [githubactionsworkflow] badge

chris48s:
closes https://github.com/badges/shields/issues/9315

Interestingly, in setting up a repo with some test workflows, I discovered that putting workflows in a subdirectory under `.github/workflows` doesn't work. So that's probably why we've never seen this before :bulb:


calebcartwright (approved):
No objections to doing this categorically :+1:

I suppose it could be conditionally special cased (only when the path starts with `pages`), but I don't think there's any remotely meaningful/observable differences between the two approaches at execution time

Issue #9075, opened 2023-04-11
generating lockfile with npm 9 breaks node 14 package build


chris48s:
:clipboard: **Description**

Hit this while working on https://github.com/badges/shields/pull/9014

Somewhere along the line I updated to npm 9. In the process of resolving the merge conflict merging master into https://github.com/badges/shields/pull/9014 I reg-generated the lockfile with npm 9. This upgraded the lockfile format to version 3, which in turn broke the node 14 package build.

I think we need to either:

- Set `engines.npm` to `^8.0.0` instead of `>=8.0.0` in `package.json` if we want to keep testing badge-maker on node 14
- Upgrade the lockfile format to v3 and drop node 14 compatibility for badge-maker (which we need to do anyway. I feel like there was a issue/PR where this came up recently, but I can't find it)

I'll probably do the second one, but I don't want to get completely bogged down in it right now.

chris48s:
I think the other PR I was thinking about was https://github.com/badges/shields/pull/8959 - that one was about migrating badge-maker to ESM and dropping CJS, which is a slightly different issue than dropping node 14

chris48s:
Try to come back to this one once https://github.com/badges/shields/pull/9290 and https://github.com/badges/shields/pull/9289 are merged but before we get to next week's dependabot PRs.

PR #9323, opened 2023-06-26
upgrade to npm 9

chris48s:
Closes https://github.com/badges/shields/issues/9075

- update lockfile to version 3
- require npm 9
- stop testing package on node 14

:spiral_notepad:  Notes:
- We'll need to update the branch protection rules to remove the node 14 package builds
- If you need to rebuild the lockfile again due to merge conflicts, `npm i --lockfile-version 3 --package-lock-only` is the command to force a `lockfileVersion` 2 --> 3 migration without actually installing/removing anything.


package.json
```diff
@@ -212,7 +212,7 @@
   },
   "engines": {
     "node": "^16.13.0",
-    "npm": ">=8.0.0"
+    "npm": "^9.0.0"
```
chris48s, line 215: I've switched this to a caret. As we can see from this Pr, jumping a major NPM version tends to require some intervention/review from us

calebcartwright (dismissed):

PR #9326, opened 2023-06-28
[SourceForge] Added more badges for SourceForge 

### LINKED ISSUE #9301: [REQ] More useful badges for Sourcefoge
:clipboard: **Description**

Need more useful badges for SF like ([JSON here](https://sourceforge.net/rest/p/scorchcrafter)):

- License [id]: https://img.shields.io/sourceforge/license:1/scorchcrafter -> 
- Language [id]: https://img.shields.io/sourceforge/language:0/scorchcrafter -> 
- Status: https://img.shields.io/sourceforge/status/scorchcrafter -> 
- etc.

Hope that inspires !

Av32000:
Hi !

In order to answer issue #9301 I added more badges for the SourceForge service:

| Data Field 	| Build URL                                       	| Badge 	|
|------------	|-------------------------------------------------	|-------	|
| License (index)   	| /sourceforge/license/0/mingw?style=flat-square  	| license |
| Language (index)   	| /sourceforge/language/2/mingw?style=flat-square 	| language  	|
| Status     	| /sourceforge/status/mingw?style=flat-square     	| status |

New license and language badges allow you to choose the item to be displayed on the badge according to its index in the array (because sourceforge lists all languages and licenses for each project).
  
Close #9301

### DIFF
--- a/services/sourceforge/sourceforge-language.service.js
+++ b/services/sourceforge/sourceforge-language.service.js
@@ -1,0 +1,45 @@
+import Joi from 'joi'
+import BaseSourceForgeService from './sourceforge-base.js'
+
+const schema = Joi.object({
+  categories: Joi.object({
+    language: Joi.array().required(),
+  }).required(),
+}).required()
+
+export default class SourceforgeLanguage extends BaseSourceForgeService {
+  static category = 'analysis'
+
+  static route = {
+    base: 'sourceforge/language',
+    pattern: ':index/:project',
+  }
+
+  static examples = [
+    {
+      title: 'SourceForge language',
+      namedParams: {
+        project: 'mingw',
+        index: '0',
+      },
+      staticPreview: this.render('Fortran'),
+    },
+  ]
+
+  static defaultBadgeData = { label: 'language' }
+
+  static render(language) {
+    return {
+      message: language,
+      color: 'blue',
+    }
+  }
+
+  async handle({ project, index }) {
+    const body = await this.fetch({ project, schema })
+    return this.constructor.render(
+      body.categories.language[parseInt(index)]?.fullname ||
+        body.categories.language[0]?.fullname
+    )
+  }
+}
--- a/services/sourceforge/sourceforge-status.service.js
+++ b/services/sourceforge/sourceforge-status.service.js
@@ -1,0 +1,39 @@
+import Joi from 'joi'
+import BaseSourceForgeService from './sourceforge-base.js'
+
+const schema = Joi.object({
+  status: Joi.string().required(),
+}).required()
+
+export default class SourceforgeStatus extends BaseSourceForgeService {
+  static category = 'monitoring'
+
+  static route = {
+    base: 'sourceforge/status',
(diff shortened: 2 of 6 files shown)


services/sourceforge/sourceforge-language.service.js
```diff
@@ -0,0 +1,45 @@
...
+    language: Joi.array().required(),
+  }).required(),
+}).required()
+
+export default class SourceforgeLanguage extends BaseSourceForgeService {
+  static category = 'analysis'
+
+  static route = {
+    base: 'sourceforge/language',
+    pattern: ':index/:project',
```
chris48s, line 15: This pattern of the user supplying an array index isn't really something we would usually do. I think if you wanted to show just a specific language or license then a static badge would probably be a more reliable way to do that than selecting based on an array index.

For licenses specifically, we can already handle projects that are released under >1 license. `renderLicenseBadge()` does have the ability to accept an array of licenses so lets go with that for license and implement something similar for language.

services/sourceforge/sourceforge-language.service.js
```diff
@@ -0,0 +1,45 @@
...
+    return {
+      message: language,
+      color: 'blue',
+    }
+  }
+
+  async handle({ project, index }) {
+    const body = await this.fetch({ project, schema })
+    return this.constructor.render(
+      body.categories.language[parseInt(index)]?.fullname ||
```
chris48s, line 41: If we rely on the `fullname` property being present, we need to validate it exists in the schema
Here's an example of how to validate an array is an array of objects containing a given property
https://github.com/badges/shields/blob/1543d0f363f8d00dca27bac486b8b3cb112dbd81/services/pypi/pypi-base.js#L12-L18

---

Also, some of these names are a bit too wordy for a badge. For example "GNU General Public License version 2.0 (GPLv2)".

Do SourceForge publish anywhere a list of their identifiers? On the PyPI badge, we manually aliased some of the longer descriptions to SPDX identifiers https://github.com/badges/shields/blob/1543d0f363f8d00dca27bac486b8b3cb112dbd81/services/pypi/pypi-helpers.js#L69-L71 We might need to do the same here.

services/sourceforge/sourceforge-status.service.js
```diff
@@ -0,0 +1,39 @@
+import Joi from 'joi'
+import BaseSourceForgeService from './sourceforge-base.js'
+
+const schema = Joi.object({
+  status: Joi.string().required(),
```
chris48s, line 5: What does this status value mean? What values could it have other than "active"?

Av32000: I don't know because SourceForge doesn't publish full documentation of their api. This status system is not documented anywhere.  More, I did not find any project with a status other than ''active''

chris48s: I think if we're going to have a badge for it, we need to know what it is.

For example, how do we know "monitoring" is the right category for this if we don't really know what the value means?

Issue #9345, opened 2023-07-02
opencollective members counts are imprecise


xxchan:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

e.g., `npm run badge -- opencollective/backers/logseq`

request `https://opencollective.com/logseq/members/users.json` returns 990 members. But https://opencollective.com/logseq has more

### 🔗 Link to the badge

https://img.shields.io/opencollective/backers/logseq

### 💡 Possible Solution

Maybe paging is needed?

There's another API https://opencollective.com/logseq.json, which contains `"backersCount":5260`, but it includes both individual backers and organizations (sponsors).

xxchan:
Oh, there are new graphQL API.

chris48s:
This is still an issue, but there is a PR at https://github.com/badges/shields/pull/9346 which takes some steps towards solving it.

In order for that to be viable, we also need to:

- Implement authentication
- Increase the length of time we cache the badges for

Pointers:

- There is a helper for Query String Auth in in https://github.com/badges/shields/blob/master/core/base-service/auth-helper.js and examples of other services using it https://github.com/search?q=repo%3Abadges%2Fshields%20withQueryStringAuth&type=code
- There is documentation for adding new config variables in https://github.com/badges/shields/blob/master/doc/adding-new-config-values.md
- We can sort out an API key to use in production. The anon rate limit should be ok under test and on staging.
- We set the max-age for badges using the  `_cacheLength` property e.g: https://github.com/badges/shields/pull/9343 I'd suggest increasing this to 900 seconds (15 minutes) for the OpenCollective GraphQL badges as a starting point.

PR #9346, opened 2023-07-02
[OpenCollective] update opencollective to api v2

xxchan:
fix https://github.com/badges/shields/issues/9345


services/opencollective/opencollective-base.js
```diff
@@ -38,45 +42,49 @@ export default class OpencollectiveBase extends BaseJsonService {
...
-      url: `https://opencollective.com/${collective}/members/${
-        userType || 'all'
-      }.json${tierId ? `?TierId=${tierId}` : ''}`,
-      httpErrors: {
-        404: 'collective not found',
+  getCnt(data) {
+    const {
+      data: {
+        account: {
+          members: { totalCount: cnt },
```
chris48s, line 83: Can we just use `count`/`getCount()`. Count is not a long word and readability matters. Lots of our contributors are not first language English speakers.

services/opencollective/opencollective-base.js
```diff
@@ -1,26 +1,30 @@
...
-// https://developer.opencollective.com/#/api/collectives?id=get-info
-const collectiveDetailsSchema = Joi.object().keys({
-  slug: Joi.string().required(),
-  backersCount: nonNegativeInteger,
-})
+const schema = Joi.object({
+  data: Joi.object({
+    account: Joi.object({
+      name: Joi.string().required(),
+      slug: Joi.string().required(),
```
chris48s, line 11: Looks like there is some stuff defined in the schema that we are not actually using. For example, we've made `name` and `slug` required fields in the response, but I don't think we're actually using them anywhere? Can we make sure we only fail if fields we rely on are missing. There's some docs on writing schemas at https://github.com/badges/shields/blob/master/doc/input-validation.md

services/opencollective/opencollective-by-tier.service.js
```diff
@@ -21,12 +21,31 @@ export default class OpencollectiveByTier extends OpencollectiveBase {
...
-    if (result.tier) {
-      if (!result.tier.endsWith('s')) result.tier += 's'
-    } else result.tier = 'new tier'
-    return this.constructor.render(result.backersCount, result.tier)
+
+    if (!members) {
+      return this.constructor.render(0, 'open collective')
+    }
+
+    // FIXME: should use pagination to get all members
```
chris48s, line 39: Would it be possible to do this properly if `OpencollectiveByTier` made its own GraphQL query rather than sharing a query with the other services, or does the API not support the relevant params to do the queries we would need here?

services/opencollective/opencollective-base.js
```diff
@@ -38,45 +42,49 @@ export default class OpencollectiveBase extends BaseJsonService {
...
+          }
+        }
+      `,
+      variables: {
+        slug: collective,
+        accountType,
+      },
+      httpErrorMessages: {
         404: 'collective not found',
       },
```
chris48s, line 72: This looks like a hangover for the old Rest API code. It looks like the standard GraphQl error handling catches the not found case and returns an `InvalidResponse` with message `"No collective found with slug..."`. I think you can ditch this (and the mocked test in `opencollective-all.tester.js`).

chris48s (commented):
Hi. Thanks for working on this. It is nice when people report an issue and then submit a good PR working on their own issue :) I've had a look over this and submitted a first pass of comments.

chris48s:
I think the biggest issue I'd like to flag with this is we are slightly mis-counting some things here, or there are some discrepancies between the APIs.

For example, if I request
https://img.shields.io/opencollective/all/shields
that tells us we have 175 backers and sponsors. I think that is the right number
This implementation is telling us 170

If I request
https://img.shields.io/opencollective/tier/shields/2988
that tells us we have 39 users at tier ID 2988
This implementation is telling us 22

If the only issue was that we were under-counting for collectives with large numbers when using the Rest API, I'd expect those totals to be the same.

If there is legitimate discrepancy between the counts from the Rest and GraphQL APIs I'd like us to understand that and be able to explain exactly why they are different or not comparable..

services/opencollective/opencollective-by-tier.service.js
xxchan: I don't find an API to get the count by tier.

xxchan: https://graphql-docs-v2.opencollective.com/types/Account

image

xxchan:
> For example, if I request
[img.shields.io/opencollective/all/shields](https://img.shields.io/opencollective/all/shields)
that tells us we have 175 backers and sponsors. I think that is the right number
This implementation is telling us 170

It turns out there are other type of accounts like `COLLECTIVE`. Fixed. Now it's 175 again.

> If I request
[img.shields.io/opencollective/tier/shields/2988](https://img.shields.io/opencollective/tier/shields/2988)
that tells us we have 39 users at tier ID 2988
This implementation is telling us 22

I guess this is due to Rest API's batch limit is 1000, but GraphQL API's limit is 100.

services/opencollective/opencollective-by-tier.service.js
xxchan: I don't want to fix it now in this PR. Do you think it's better to fallback to old rest api for /tier (i.e., keep it as-is)? But I think sooner or later we need to use new api. Besides, opencollective calls it `legacyId` :)

xxchan: I copied the old JsonBase to `services/opencollective/opencollective-by-tier.service.js` and reverted the changes there.

chris48s: I agree for the moment what you've done makes sense. At least with the Rest API we get a larger page size in a single request so we'll be giving a correct result in more cases.

Lets punt where we go next with this one to another issue. I will write something up.

chris48s (approved):
Sorry it has taken a few days to circle back to this one

services/opencollective/opencollective-by-tier.service.js
chris48s: https://github.com/badges/shields/issues/9361

chris48s:
Hit a bit of a snag here.

I merged this PR, deployed it, and all the OpenCollective badges stopped working.

Here's what happened:

The OpenCollective GraphQL API has got much more restrictive rate limits than the Rest API: https://graphql-docs-v2.opencollective.com/access
For anonymous users, you get 10 requests/minute
With auth, you get 100 requests/minute

Merging and deploying this PR put us onto the 10 requests/minute tier, which is way too low and we just instantly hit the rate limit and most of the badges just started saying "rate limited by upstream service".

This was sloppy on my part: I should have picked this up in review. Usually rate limits are one of the first things we look at when adding a new service, but I didn't think of it because OpenCollective is a service we've been happily doing a high volume of traffic with for years.

For now, I have reverted this change in https://github.com/badges/shields/pull/9362 and deployed. That rolls us back to using the Rest API. This mostly fixes things, but it does re-introduce this bug https://github.com/badges/shields/issues/9345 :sob:

Implementing authentication will give us a higher rate limit. While 100 requests/minute is reasonably generous and will cover our usage at most times, looking at the metrics for production, I think we could still be trying to make more than 100 requests/minute at peak times. We will also need to cache the badges for longer if we are going to make sure we stay under that threshold (although at the moment we only cache these for 2 mins - I'd be happy to crank the max-age up quite a bit on these, so I think it is do-able to get under that threshold).

I think in order to move this forward, we're going to need to either:

- Stick with GraphQL but
  - Implement authentication
  - Increase the length of time we cache the badges for
- Find a way to solve that issue without migrating to the GraphQL API

@xxchan - You seem pretty engaged with this issue so far.. Would you be happy to work on this? If so:

- There is a helper for Query String Auth in in https://github.com/badges/shields/blob/master/core/base-service/auth-helper.js and examples of other services using it https://github.com/search?q=repo%3Abadges%2Fshields%20withQueryStringAuth&type=code
- There is documentation for adding new config variables in https://github.com/badges/shields/blob/master/doc/adding-new-config-values.md
- We can sort out an API key to use in production. The anon rate limit should be ok under test and on staging.
- We set the max-age for badges using the  `_cacheLength` property e.g: https://github.com/badges/shields/pull/9343 I'd suggest increasing this to 900 seconds (15 minutes) for the OpenCollective GraphQL badges as a starting point.

xxchan:
Wow I didn't expected that as I'm not very familiar with API dev 🥲... Thanks for the detailed writeup. I'm interested in fixing it, but maybe not recently. We can reopen https://github.com/badges/shields/issues/9345 first, and I can fix it in the future if no one takes it.

Issue #9214, opened 2023-05-29
🚀 SimpleIcons 9


chris48s:
Logos on Shields.io are provided by SimpleIcons. We've recently upgraded to [SimpleIcons 9](https://github.com/simple-icons/simple-icons/releases/tag/9.0.0). There are 4 icons that have been removed or renamed in this release. Please remember that we are just consumers of SimpleIcons. Decisions about changes and removals are made by the [SimpleIcons](https://github.com/simple-icons/simple-icons) project.

## Removed Icons

<ul>
<li>Alitalia (<a href="https://redirect.github.com/simple-icons/simple-icons/issues/8310">#8310</a>)</li>
<li>Linode (<a href="https://redirect.github.com/simple-icons/simple-icons/issues/8549">#8549</a>)</li>
</ul>

## Renamed Icons

<ul>
<li>MonoGames (<a href="https://redirect.github.com/simple-icons/simple-icons/issues/8614">#8614</a>) (renamed to <code>monogame</code>)</li>
<li>PyTorch Lightning (<a href="https://redirect.github.com/simple-icons/simple-icons/issues/8415">#8415</a>) (renamed to <code>lightning</code>)</li> 
</ul>

calebcartwright:
Think we can close and unpin this one 😅

PR #9347, opened 2023-07-03
move some docs to the website

### LINKED ISSUE #9286: Move some more docs to the website
:clipboard: **Description**

There are a few bits of documentation (like https://github.com/badges/shields/blob/master/doc/static-badges.md and https://github.com/badges/shields/blob/master/doc/logos.md ) which are user-facing (how to use shields.io) rather than developer-facing that ended up being markdown files in the repo, just because the old website made it difficult to put them anywhere sensible.

It is now much easier to make a website page based on a markdown file and link it up in a menu. Lets review this documentation and move it on to shields.io where appropriate.

chris48s:
Closes #9286

There's actually a bit of a mixed-bag of things going on here.

1. Move some "how to use the website" docs from the repo to the website. Technical and contributing docs stay where they are
2. Update various links in documentation and issue templates. Some of it is linking to the new things, but also there are a few places where we still have stuff like https://shields.io/#your-badge knocking about that needed updating
3. Start a blog. More comments on this inline.

It is probably easiest to have a look at this one in a review app.


frontend/docusaurus.config.cjs
```diff
@@ -60,7 +60,13 @@ const config = {
         },
         items: [
           { to: '/badges', label: 'Badges', position: 'left' },
+          {
+            to: '/docs',
+            label: 'Documentation',
```
chris48s, line 65: "Documentation" might not be the best title for this menu item, but I don't have a better idea. Very open to suggestions on this.

frontend/blog/2023-07-03-new-frontend.md
```diff
@@ -0,0 +1,21 @@
+---
+slug: new-frontend
+title: We launched a new frontend
```
chris48s, line 3: We joked about the silly first blog post I put in the proof-of-concept site I made to evaluate this tech.

I actually think a blog on shields.io is quite a good place to put things like:

- We're making some kind of unavoidable breaking change to shields.io (e.g: simple-icons major update with removed icons)
- We're deprecating thing X for self hosting users (e.g: removal of redis token pool)
- We just put out a new package version - here's the release notes

i.e: the kind of thing might usually put in pinned issues. I don't think that means we stop using pinned issues, but having somewhere on shields.io to put those "announcements" is useful IMO.

I do have other ideas for some things we might post (if we have the time/inclination). Some possible topics or post formats might include:

- Behind the scenes: Donations + Spending. What we get, how we spend it. How you can help.
- Here's a cool thing someone did with shields that we like (e.g: https://upptime.js.org/docs/badges/ )
- We have an awesome list: https://github.com/badges/awesome-badges submit your stuff to it
- Say thanks to one of our high value sponsors

..but I see giving us a place for those "announcement" posts as being the primary thing this is useful for.

I've put this here as a first post. Tbh, it is mainly just here to "seed" the blog with something. Gotta start somewhere. If we don't like this, maybe we could just put https://github.com/badges/shields/issues/9214 here as a back-dated post?

frontend/docusaurus.config.cjs
PyvesB: No brilliant alternatives in mind. We can change it later if anything comes up :)

PyvesB (dismissed):
Thanks for working on this, looks good to me 👍🏻

chris48s:
Thanks for having a look at this.

Just having a quick re-read over this PR, I realised that there are a bunch of places where we have linked to https://github.com/badges/shields/blob/master/doc/static-badges.md and https://github.com/badges/shields/blob/master/doc/logos.md in the past. I've just pushed another small commit linking to the website from those old locations to try and help people following links in old issues or whatever.

PR #9348, opened 2023-07-03
add a bit of border to select boxes

chris48s:
In https://github.com/badges/shields/pull/9324 I added a bit of border to text boxes.
I forgot to add it to select boxes.

**Before:**

Screenshot at 2023-07-03 14-59-32

**After:**

Screenshot at 2023-07-03 14-59-54


frontend/src/css/custom.css
```diff
@@ -31,7 +31,7 @@ html[data-theme="dark"] .docusaurus-highlight-code-line {
   color-scheme: initial;
 }
 
-input[type="text"] {
+input[type="text"], :not(#fakeID#fakeId#fakeID) select {
```
chris48s, line 34: This is not great CSS. Every select box has a couple of classes attached to it with long auto-generated names that I don't really want to use directly. They seem like they may not be reliable over time.
In order to make a more specific selector, I've used this hack https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#increasing_specificity_by_duplicating_selector

I might have a look at opening a PR to implement this and https://github.com/badges/shields/pull/9324 upstream at the component layer so we clean this up a bit.

calebcartwright (approved):

PR #9349, opened 2023-07-03
deprecate [snyk] badges

### LINKED ISSUE #9325: Snyk vulnerabilities shield shows 'Invalid response data'

shields.io

### 🐞 Description

Instead of count it shows 'Invalid response data'

image

### 🔗 Link to the badge

https://img.shields.io/snyk/vulnerabilities/github/adamlui/chatgpt-infinity?label=Vulnerabilities&labelColor=464646&color=gold

### 💡 Possible Solution

I wanted to open an issue in Snyk's GitHub but couldn't find an appropriate repo

chris48s:
Closes #9325

As far as I am aware, there is no way we can re-implement the old functionality.
Nobody has suggested anything better.
I suggest we deprecate them.


calebcartwright (approved):
Seems like the right path forward to me too. Can always reintroduce them should something change upstream with the API to make it possible to provide badges again

PR #9281, opened 2023-06-17
fix a load of spacing issues in frontend content

chris48s:
I'm not going to claim I have caught every single `documentation` string with a spacing issue, but I reckon I've got >90% of them. This also lays down a pattern for fixing them if we find any more I have missed.

I have mostly chosen to fix these by converting them to markdown (which we can now use instead of HTML, if we want). In one or two cases, I have moved ht HTML around a bit. In general, we might want to consider converting all of these to markdown, but that is a bigger job.

I won't go through every single page with a spacing issue, but here's a few before/after links

| before | after |
| ------ | ----- |
| https://shields.io/badges/git-hub-contributors-via-allcontributors-org | https://pr-9281-badges-shields.fly.dev/badges/git-hub-contributors-via-allcontributors-org |
| https://shields.io/badges/you-tube-channel-views | https://pr-9281-badges-shields.fly.dev/badges/you-tube-channel-views |
| https://shields.io/badges/ping-pong-status | https://pr-9281-badges-shields.fly.dev/badges/ping-pong-status |
| https://shields.io/badges/git-lab-all-merge-requests | https://pr-9281-badges-shields.fly.dev/badges/git-lab-all-merge-requests |

Refs https://github.com/badges/shields/pull/9280

PR #9360, opened 2023-07-07
improve docs for github issues search badge

chris48s:


calebcartwright (approved):
idk if I was previously aware we could use markdown in the docs :shrug:

chris48s:
With the old frontend, it had to be HTML.

Markdown support is one of the new things we get from docusaurus-openapi. I switched some of our existing documentation to start using it in https://github.com/badges/shields/pull/9281 so solve some spacing issues.

It is worth noting that (like most markdown implementations) the MDX parser it gives us does have some quirks https://github.com/cloud-annotations/docusaurus-openapi/issues/142

PR #9368, opened 2023-07-11
Add support for [Lemmy]

### LINKED ISSUE #9367: Lemmy service badge
### 📋 Description

Lemmy - A link aggregator for the fediverse.

For a community, it may be helpful to show the number of subscribers, active users in a particular timeframe, etc.

Example: https://img.shields.io/badge/community-123%20subscribers-4c1

### 🔗 Data

Yes, the API docs are public at: https://join-lemmy.org/docs/contributors/04-api.html and https://join-lemmy.org/api/

For example, to get the subscribers count for `!piped@feddit.rocks`:
A request is made to https://feddit.rocks/api/v3/community?name=piped
The subscriber count can be gotten using this JSON path: `community_view.counts.subscribers`

### 🎤 Motivation

Lemmy is a good alternative to Reddit, built with the Fediverse in mind.

FireMasterK:
Closes #9367 

I'm a bit stuck when it comes to writing tests for this currently.

### DIFF
--- a/services/lemmy/lemmy.service.js
+++ b/services/lemmy/lemmy.service.js
@@ -1,0 +1,94 @@
+import Joi from 'joi'
+import { metric } from '../text-formatters.js'
+import { BaseJsonService, InvalidParameter } from '../index.js'
+
+const queryParamSchema = Joi.object({
+  server_fqdn: Joi.string().hostname(),
+}).required()
+
+const lemmyCommunitySchema = Joi.object({
+  community_view: Joi.object({
+    counts: Joi.object({
+      subscribers: Joi.number().required(),
+      posts: Joi.number().required(),
+      comments: Joi.number().required(),
+    }).required(),
+  }).required(),
+}).required()
+
+export default class Lemmy extends BaseJsonService {
+  static category = 'social'
+
+  static route = {
+    base: 'lemmy',
+    pattern: ':community',
+    queryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'Lemmy',
+      namedParams: { community: 'asklemmy@lemmy.ml' },
+      staticPreview: this.render({
+        community: 'asklemmy@lemmy.ml',
+        members: 42,
+      }),
+    },
+    {
+      title: 'Lemmy',
+      namedParams: { community: 'asklemmy@lemmy.ml' },
+      queryParams: { server_fqdn: 'lemmy.world' },
+      staticPreview: this.render({
+        community: 'asklemmy@lemmy.ml',
+        members: 42,
+      }),
+    },
+  ]
+
+  static _cacheLength = 30
+
+  static defaultBadgeData = { label: 'community' }
+
+  static render({ community, members }) {
+    return {
+      label: `subscribe to ${community}`,
+      message: `${metric(members)} subscribers`,
+      color: 'brightgreen',
+    }
+  }
+
+  async fetch({ community, serverFQDN }) {
+    let host
+    if (serverFQDN === undefined) {
+      const splitAlias = community.split('@')
+      // The community will be in the format of `community@server`
+      if (splitAlias.length !== 2) {
+        throw new InvalidParameter({
+          prettyMessage: 'invalid community',
+        })
+      }
+
+      host = splitAlias[1]
+    } else {
+      host = serverFQDN
+    }
+    const data = await this._requestJson({
+      url: `https://${host}/api/v3/community`,
+      schema: lemmyCommunitySchema,
+      options: {
+        searchParams: {
+          name: community,
+        },
+      },
+      httpErrors: {
+        404: 'community not found',
+      },
+    })
+    return data.community_view.counts.subscribers
+  }
+
+  async handle({ community }, { server_fqdn: serverFQDN }) {
+    const members = await this.fetch({ community, serverFQDN })
+    return this.constructor.render({ community, members })
+  }
+}
--- a/services/lemmy/lemmy.service.js
+++ b/services/lemmy/lemmy.service.js
@@ -2,16 +2,10 @@ import Joi from 'joi'
 import { metric } from '../text-formatters.js'
 import { BaseJsonService, InvalidParameter } from '../index.js'
 
-const queryParamSchema = Joi.object({
-  server_fqdn: Joi.string().hostname(),
-}).required()
-
 const lemmyCommunitySchema = Joi.object({
   community_view: Joi.object({
     counts: Joi.object({
       subscribers: Joi.number().required(),
-      posts: Joi.number().required(),
-      comments: Joi.number().required(),
     }).required(),
   }).required(),
 }).required()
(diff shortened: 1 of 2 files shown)


chris48s:
There are some docs on writing tests at https://github.com/badges/shields/blob/master/doc/service-tests.md and plenty of examples to crib from.

For centralised services, it is good to have tests that call the live service (even if they can be a bit flaky in some cases) because it means if something changes upstream (e.g: a package registry make a change to their API, or an endpoint gets deprecated, or whatever) then we know about it because our tests start failing. With services like that are self-host only and there is no central service, I think I would be happy to just test against mocks because we don't really get the payoff we normally exchange for the live tests being flaky.

There are some docs specifically on using mocks in service tests at https://github.com/badges/shields/blob/master/doc/service-tests.md#5-mocking-responses and again, plenty of examples to crib from.

Another option would be if there is a server that acts as the "reference implementation" (in that way that mastodon.social is for the mastodon community), we could test against that for "live" tests.

FireMasterK:
I added a few tests, I hope they are good enough!

services/lemmy/lemmy.service.js
```diff
@@ -0,0 +1,94 @@
...
+      // The community will be in the format of `community@server`
+      if (splitAlias.length !== 2) {
+        throw new InvalidParameter({
+          prettyMessage: 'invalid community',
+        })
+      }
+
+      host = splitAlias[1]
+    } else {
+      host = serverFQDN
```
chris48s, line 73: Can you explain what's going on here with being able to extract the URL we're going to query from either a param or the community name?

Are there situations where the community name is `something@thisurl.com` but we want to query `thaturl.com` for the data?

FireMasterK: We can technically lookup a community on the community's instance, or use another instance to lookup a community.

This may be helpful for adding kbin communities for example.

chris48s: I've heard of lemmy, but I don't know what kbin is. I'm not familiar enough with it to connect the dots of why this would be more useful or necessary in that context.

Can you give an example?

FireMasterK: Kbin is another federated platform like lemmy, compatible with Lemmy. Users on lemmy and kbin can interact with them. Kbin has it's own API, but with this change, we can use Lemmy's API to view data from kbin.

FireMasterK: For example, see https://lemmy.ml/c/gnome@kbin.social

chris48s: OK. Got it. So

- https://lemmy.ml/api/v3/community?name=asklemmy%40lemmy.ml is fine (and tell me the same number of subscribers as https://lemmy.ml/c/asklemmy )
- https://kbin.social/api/v3/community?name=gnome%40kbin.social doesn't work

but as a workaround, you could do

- https://lemmy.ml/api/v3/community?name=gnome%40kbin.social or
- https://lemmy.world/api/v3/community?name=gnome%40kbin.social

A question about that:

- If I look at https://lemmy.ml/api/v3/community?name=gnome%40kbin.social it tells me `gnome@kbin.social` has got 2 subscribers.
- If I look at https://lemmy.world/api/v3/community?name=gnome%40kbin.social it says `gnome@kbin.social` has got 29 subscribers.
- ..but if I look at https://kbin.social/m/GNOME it says 245 subscribers.

Any idea why the counts are so wildly different? I guess each instance only knows about some subset of `gnome@kbin.social` it is federated with (maybe lemmy.ml only knows about subscribers to `gnome@kbin.social` whose home instance is lemmy.ml?) Given that constraint, is this actually a useful feature? i.e: If you look up the stats for a community on some other non-home instance, you're always going to be getting way less than the full picture.

Feels like this feature (looking up on a non-home instance) could be something that technically works, but may not provide a metric anyone actually wants to display.

FireMasterK: I think the counts are different since it likely shows the number of subscribers from that instance, which does make it a bit useless. Should I remove this feature then?

FireMasterK: I've just removed the ability to specify FQDNs to fetch from in the latest commit.

services/lemmy/lemmy.service.js
```diff
@@ -0,0 +1,90 @@
+import Joi from 'joi'
+import { metric } from '../text-formatters.js'
+import { BaseJsonService, InvalidParameter } from '../index.js'
+
+const queryParamSchema = Joi.object({
```
chris48s, line 5: We can delete this schema now, and remove it from the route declaration

services/lemmy/lemmy.service.js
```diff
@@ -0,0 +1,90 @@
...
+    },
+    {
+      title: 'Lemmy',
+      namedParams: { community: 'asklemmy@lemmy.ml' },
+      queryParams: { server_fqdn: 'lemmy.world' },
+      staticPreview: this.render({
+        community: 'asklemmy@lemmy.ml',
+        members: 42,
+      }),
+    },
```
chris48s, line 45: We can ditch this second example

services/lemmy/lemmy.service.js
```diff
@@ -0,0 +1,90 @@
...
+  ]
+
+  static _cacheLength = 30
```
chris48s, line 48: Why cache so short? We'd normally only drop this low for stuff that is really time-sensitive. I think we can delete this and go with the default for social badges (2 mins).

services/lemmy/lemmy.service.js
```diff
@@ -0,0 +1,90 @@
...
+
+  static defaultBadgeData = { label: 'community' }
+
+  static render({ community, members }) {
+    return {
+      label: `subscribe to ${community}`,
+      message: `${metric(members)} subscribers`,
+      color: 'brightgreen',
```
chris48s, line 56: Normally the label (left side of the badge) should describe what the number in the message (right side means). Optionally with the social badges, this can be a call to action. Lets ditch the " subscribers" suffix from the message here.

services/lemmy/lemmy.service.js
```diff
@@ -0,0 +1,90 @@
...
+const queryParamSchema = Joi.object({
+  server_fqdn: Joi.string().hostname(),
+}).required()
+
+const lemmyCommunitySchema = Joi.object({
+  community_view: Joi.object({
+    counts: Joi.object({
+      subscribers: Joi.number().required(),
+      posts: Joi.number().required(),
+      comments: Joi.number().required(),
```
chris48s, line 14: If the only value we're going to use is subscribers we can drop posts/comments from the schema

services/lemmy/lemmy.service.js
chris48s: OK. I think this makes sense. We can always revisit it as a feature if there is a use-case for it.

chris48s:
You'll need to update the tests to remove the " subscribers" suffix too.

chris48s (approved):

PR #9372, opened 2023-07-14
Go back to default YouTube cache

PyvesB:
YouTube has finally granted the API quota increase.

I'm still trying to figure out what the new quota is, as the YouTube API team unhelpfully stated:
> We have increased your Quota by 200,000. The total quota for xxxx is now  200,000 units per day

The request form also insisted on how much _**additional**_ quota we wanted, our previous quota being 50,000, the new quota may actually be 250,000. Either way, we should have more than enough for many months to come. Let's set the cache back to its default value.

### DIFF
--- a/services/youtube/youtube-base.js
+++ b/services/youtube/youtube-base.js
@@ -40,8 +40,6 @@ class YouTubeBase extends BaseJsonService {
     isRequired: true,
   }
 
-  static _cacheLength = 7200
-
   static defaultBadgeData = {
     label: 'youtube',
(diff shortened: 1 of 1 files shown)


chris48s:
In https://console.cloud.google.com/apis/api/youtube.googleapis.com/quotas it says our limit is 200,000 queries per day.

We might actually want to settle on something a bit longer than the default but shorter than this. Lets see what happens to our usage when we deploy this anyway.

chris48s (approved):

Issue #9297, opened 2023-06-18
Deploy on Node 18


chris48s:
:clipboard: **Description**

From #9292

We are currently deploying on Node 16, which will reach EOL and stop getting security patches in ~2 months on 11 Sep 2023
https://nodejs.org/en/blog/announcements/nodejs16-eol
Node 18 will get patches for another ~2 years
The current LTS release is Node 20.

I'm going to suggest we:

- ~Merge this, update the branch protection rules (to account for the changed workflow names) and continue to run the tests on Node 16/18~
- Continue to deploy on Node 16 in the immediate term
- Start running on Node 18 in local dev
- Switch the docker image over to Node 18 some time in the next few weeks
- Try running the tests on Node 20 once we're prod is happy on Node 18

chris48s:
@SnoozeThis wait until July 16 2023

SnoozeThis:
(https://snoozeth.is/StSfHlZUDFI) I will wait until Sun, 16 Jul 2023 00:00:00 UTC and then add a comment.

SnoozeThis:
Resolved.

PR #9385, opened 2023-07-15
deploy on node 18

chris48s:
Refs https://github.com/badges/shields/issues/9297
Refs https://github.com/badges/shields/pull/9323

Node 16 will reach EOL and stop getting security patches on 11 Sep 2023. I've been running my local dev copy with node 18 on and off recently and not noticed any problems. We've also been running the tests on node 18 in CI for ~a month 

~I would like to get https://github.com/badges/shields/pull/9323 reviewed and merged ahead of this one because once we upgrade to node 18, the default npm version is npm 9. Requiring contributors to use node 18 but requiring an old npm version and providing a v2 lockfile is a bit of a crap contributor experience.~ done + updated

Alternatively, I could smoosh those changes into this PR too.

TODO: update branch protection rules when we merge
- 16 + 18 package lib tests have changed due to engine-strict true/false
- remove main-18, integration-18, services-18
- add main-20, integration-20, services-20


.github/workflows/test-integration-20.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Setup
         uses: ./.github/actions/setup
         with:
-          node-version: 18
+          node-version: 20
```
PyvesB, line 38: Do other values in this file need to be updated as well? There still seem to be a couple of references to Node 18.

PyvesB (commented):
This still uses Node 16, shall we update it as well whilst we're at it?
https://github.com/badges/shields/blob/7d966ab6bd340ce358a733c9db40900f3650a573/.github/actions/close-bot/action.yml#L11

jNullj (commented):
Shouldn't [badge-maker/package.json](https://github.com/badges/shields/blob/7d966ab6bd340ce358a733c9db40900f3650a573/badge-maker/package.json#L29) also get updated to node 18?

.github/workflows/test-integration-20.yml
chris48s: Good catch in both cases :+1:

chris48s:
> This still uses Node 16, shall we update it as well whilst we're at it?

`node16` is the highest supported runner for javascript actions. The next jump will be to 20. See https://github.com/actions/runner/discussions/2704

Fortunately, keeping this in sync with the rest of the codebase doesn't really matter as none of the application code is run by this action. Only the files in https://github.com/badges/shields/tree/master/.github/actions/close-bot


> Shouldn't badge-maker/package.json also get updated to node 18?

Badge-maker is a package which is used by other applications as well as shields.io, so we try to test against a fairly wide range of node versions (that range needs to include whatever major node version we deploy shields.io on, but ideally it is as wide as we can make it without jumping through too many hoops). At the moment, badge-maker is tested on Node 16, 18 and 20. There's no need to drop compatibility with node 16 for now so `>=16` is the right range.

PyvesB (approved):
Looks good to me! 👍🏻

PR #9386, opened 2023-07-15
devops: automate bug report tests

### LINKED ISSUE #9351: automation to help with bug reports
:clipboard: **Description**

Usually if someone files a bug report, the first thing I do is use `npm run badge` locally to get some additional information.

It would be nice if we had a GitHub actions workflow that fires when a new issue is opened and does the following:

If the user filled in the "bug report" template **AND** filled in "shields.io" in the "Are you experiencing an issue with..." field **AND** filled in a URL in the "Link to the badge" field then...

- Extract the value from "Link to the badge" to `$URL`
- Spin up an environment
- Run `npm run badge $URL`
- Capture the output and post it as an issue comment in a code fence (make it clear this was posted by a bot)

It might be that we need to tweak the issue template yaml a bit to make this easier?

jNullj:
Devs run `npm run badge` often when getting a bug report for shields.io service. This PR automates this tests so the maintainer can get test results automaticly when getting into a new ticket. Add test-bug-run-badge.yml workflow for github actions automation.

Will only run if bug has the 'qeustion' label added by bug reporting template. And will only setup enviorment and perform the test if the link provided in the issue is valid and the issue is related to shields.io. Link to job results is sent as a new comment on the issue.

Resolves #9351

Image of the bot in action
image

### DIFF
--- a/.github/workflows/test-bug-run-badge.yml
+++ b/.github/workflows/test-bug-run-badge.yml
@@ -41,7 +41,7 @@ jobs:
           node-version: 16
           cypress: false
 
-      - name: Run tests
+      - name: Output debug info
         id: run_tests
         env:
           TEST_BADGE_LINK: '${{ needs.extract-bug-badge-url.outputs.link }}'
--- a/.github/workflows/test-bug-run-badge.yml
+++ b/.github/workflows/test-bug-run-badge.yml
@@ -42,7 +42,6 @@ jobs:
           cypress: false
 
       - name: Output debug info
-        id: run_tests
         env:
           TEST_BADGE_LINK: '${{ needs.extract-bug-badge-url.outputs.link }}'
         run: npm run badge $TEST_BADGE_LINK
--- a/.github/workflows/test-bug-run-badge.yml
+++ b/.github/workflows/test-bug-run-badge.yml
@@ -1,0 +1,78 @@
+name: Test new bug report badge
+run-name: Test bug report on issue ${{ github.event.issue.number }}
+on:
+  issues:
+    types: [opened]
+jobs:
+  extract-bug-badge-url:
+    if: ${{ contains(github.event.issue.labels.*.name, 'question') }}
+    runs-on: ubuntu-latest
+    outputs:
+      runBadgeTest: ${{ steps.testCondition.outputs.runNext }}
+      link: ${{ steps.testCondition.outputs.link }}
+    steps:
+      - name: Test badge test run conditions
+        id: testCondition
+        run: |
+          product=$(echo "${{ github.event.issue.body }}" | grep -A2 "Are you experiencing an issue with.*" | tail -n 1)
+          link=$(echo "${{ github.event.issue.body }}" | grep -A2 "Link to the badge.*" | tail -n 1)
+
+          if [[ "$product" == "shields.io" && "$link" == "https://img.shields.io"* ]]; then
+            echo "runNext=true" >> "$GITHUB_OUTPUT"
+            echo "link=$link" >> "$GITHUB_OUTPUT"
+          else
+            echo "Conditions not met. Skipping the workflow..."
+            echo "runNext=false" >> "$GITHUB_OUTPUT"
+          fi
+
+  run-bug-badge-url-test:
+    needs: extract-bug-badge-url
+    if: needs.extract-bug-badge-url.outputs.runBadgeTest == 'true'
+    permissions:
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup
+        uses: ./.github/actions/setup
+        with:
+          node-version: 16
+          cypress: false
+
+      - name: Run tests
+        id: run_tests
+        env:
+          TEST_BADGE_LINK: '${{ needs.extract-bug-badge-url.outputs.link }}'
+          GH_TOKEN: '${{ secrets.GH_PAT }}'
+          LIBRARIESIO_TOKENS: '${{ secrets.SERVICETESTS_LIBRARIESIO_TOKENS }}'
+          OBS_USER: '${{ secrets.SERVICETESTS_OBS_USER }}'
+          OBS_PASS: '${{ secrets.SERVICETESTS_OBS_PASS }}'
+          SL_INSIGHT_USER_UUID: '${{ secrets.SERVICETESTS_SL_INSIGHT_USER_UUID }}'
+          SL_INSIGHT_API_TOKEN: '${{ secrets.SERVICETESTS_SL_INSIGHT_API_TOKEN }}'
+          TWITCH_CLIENT_ID: '${{ secrets.SERVICETESTS_TWITCH_CLIENT_ID }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/test-bug-run-badge.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Setup
+        uses: ./.github/actions/setup
+        with:
+          node-version: 16
+          cypress: false
+
+      - name: Run tests
+        id: run_tests
+        env:
```
chris48s, line 46: We shouldn't pass this job any secrets.

There is nothing in `npm run badge` to scrub or obfuscate credentials so any credentials we pass in are going to be echoed out in the request details. GH actions does have some built in masking functionality, but it won't cover basic auth credentials for example. They get base64 encoded. Obviously this means there are 9 service (out of ~200) where this will fail but I think we can live with that.

If we want to improve that, I would want to implement something at the logging layer that optionally prevents credentials from being included in the log output at all. Although we would want it to be an optional flag - outputting the credentials is really useful locally. I used it literally yesterday when I was working on https://github.com/badges/shields/pull/9387 to make sure the auth was being passed properly.

.github/workflows/test-bug-run-badge.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup
+        uses: ./.github/actions/setup
+        with:
+          node-version: 16
+          cypress: false
+
+      - name: Run tests
```
chris48s, line 44: This isn't really a test. The purpose of this step is to output debugging information. How about "Output debug info" for this step?

.github/workflows/test-bug-run-badge.yml
```diff
@@ -0,0 +1,78 @@
...
+        id: run_tests
```
chris48s, line 45: Do we use this id for anything?

.github/workflows/test-bug-run-badge.yml
```diff
@@ -0,0 +1,78 @@
...
+          LIBRARIESIO_TOKENS: '${{ secrets.SERVICETESTS_LIBRARIESIO_TOKENS }}'
+          OBS_USER: '${{ secrets.SERVICETESTS_OBS_USER }}'
+          OBS_PASS: '${{ secrets.SERVICETESTS_OBS_PASS }}'
+          SL_INSIGHT_USER_UUID: '${{ secrets.SERVICETESTS_SL_INSIGHT_USER_UUID }}'
+          SL_INSIGHT_API_TOKEN: '${{ secrets.SERVICETESTS_SL_INSIGHT_API_TOKEN }}'
+          TWITCH_CLIENT_ID: '${{ secrets.SERVICETESTS_TWITCH_CLIENT_ID }}'
+          TWITCH_CLIENT_SECRET: '${{ secrets.SERVICETESTS_TWITCH_CLIENT_SECRET }}'
+          WHEELMAP_TOKEN: '${{ secrets.SERVICETESTS_WHEELMAP_TOKEN }}'
+          YOUTUBE_API_KEY: '${{ secrets.SERVICETESTS_YOUTUBE_API_KEY }}'
+        run: npm run badge $TEST_BADGE_LINK
```
chris48s, line 58: I've not tested this but I suspect we probably need to quote this.

For example, `&` is a special character.
Have a look at what happens if you run

`npm run badge https://img.shields.io/jira/issue/KAFKA-2896?style=plastic&baseUrl=https%3A%2F%2Fissues.apache.org%2Fjira`
vs
`npm run badge "https://img.shields.io/jira/issue/KAFKA-2896?style=plastic&baseUrl=https%3A%2F%2Fissues.apache.org%2Fjira"`
in your terminal.

Can you have a look at that and check that case is handled correctly?

chris48s (commented):
This looks like a great start. Thanks. I've done a first pass of review comments. I have not tested this out yet. I guess I need to try your workflow out on a forked copy. Any chance you could look at these comments first and then I will give it a spin. Cheers

.github/workflows/test-bug-run-badge.yml
```diff
@@ -0,0 +1,78 @@
...
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup
+        uses: ./.github/actions/setup
+        with:
+          node-version: 16
```
chris48s, line 41: For the moment, this needs to stay as 16 due to `strict-englines`, but I'm going to quickly note here that when we merge https://github.com/badges/shields/pull/9385 we will need to make this 18.

.github/workflows/test-bug-run-badge.yml
jNullj: I will remove the secrets (will commit soon), but i am not sure that only 9 services will be affected.
For example, the github token is there to avoid rate limit with a large number of services.

Regarding avoiding secrets from leaking, seems a bit risky. I think we could open a new issue to get ideas about how to tackle that.

.github/workflows/test-bug-run-badge.yml
jNullj: Sounds good to me, will add in commit soon

.github/workflows/test-bug-run-badge.yml
jNullj: no, i will remove it, i used it when i did debug to output the results into a comment

.github/workflows/test-bug-run-badge.yml
jNullj: almost same output, this did not effect the string to badge-cli

.github/workflows/test-bug-run-badge.yml
jNullj: cool, whatever gets merged first should update the later

chris48s: :+1: I'll update https://github.com/badges/shields/pull/9385 after I merge this

.github/workflows/test-bug-run-badge.yml
chris48s: GitHub is a "service"
GitHub Issues or GitHub License is a "badge"
So it is one "service", but your core point is completely right: GitHub is a service that has a lot of badges and they are really widely used so it would be very nice if that one worked.

Fortunately, if we're only calling one badge URL we don't really _need_ a token. We need one under test because we're making lots of API calls. Anonymous usage would be fine here because we're not going to hit the late limit by making only one request. Allowing GH badges to work with no GitHub credentials at all would be a great solution. This has actually been a really long-standing issue affecting local development too: https://github.com/badges/shields/issues/2754 I don't think we need to fix that for this workflow to be useful, but if you were looking for a next issue to help with, https://github.com/badges/shields/issues/2754 would be a great shout.

We can merge this as it stands and treat that as a follow up.

chris48s (approved):
Thanks for working on this. This was a slightly unusual but very welcome contribution. Most contributors who are new to the project focus on adding more badges. A PR adding automation that will help the core team to maintain the project more effectively more rare, but much appreciated.

PR #9387, opened 2023-07-15
Switch [OpenCollective] badges to use GraphQL and auth

chris48s:
Closes https://github.com/badges/shields/issues/9345
Refs https://github.com/badges/shields/issues/9345#issuecomment-1627787603
Refs https://github.com/badges/shields/pull/9346#issuecomment-1627233063

This PR takes the code changes from https://github.com/badges/shields/pull/9346 and makes them viable for production traffic by adding auth and increasing the cache.

I've already reviewed c1a2593 in https://github.com/badges/shields/issues/9345 so it is only really the 3 additional commits that need a look here.

As things stand, we're currently using a more than 100 reqs/minute at busy times, but also we only cache these badges for 2 mins at the moment. Lets bump that up to 15 mins. That should keep us in line.
Potentially we could exceed this in future if OpenCollective badges become more popular. However I feel like if we need it, OpenCollective is a bit more of a human-scale organisation to ask for an increase than Youtube or DockerHub, for example.

I've made an OpenCollective account attached to the team AT shields.io email that has access to nothing and generated an API token with no scopes that we can use in production
I think for tests, review apps and staging, the anonymous rate limit will be OK

**TODO:** Before deploying this, add the token to the production env vars and shared credentials store


doc/server-secrets.md
```diff
@@ -218,7 +218,7 @@ installation access to private npm packages
 
 [npm token]: https://docs.npmjs.com/getting-started/working_with_tokens
 
-## Open Build Service
+### Open Build Service
```
chris48s, line 221: This isn't really related to this PR. I just noticed this heading level was wrong while I was editing the file.

services/opencollective/opencollective-base.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  static route = this.buildRoute('dummy')
+
+  async handle({ collective }) {
+    const data = await this.fetchCollectiveInfo({
+      collective,
+      accountType: [],
+    })
+    return this.constructor.render(this.getCount(data))
+  }
+}
```
calebcartwright, line 16: No concerns with using a dummy child class, though it does remind of a scenario we ran into with some of the GitLab classes a ways back where we forgot to wire up the auth for one or two of them. As a thought/potential future exercise, I wonder if there's a clean way we could consistently test "does every class that's supposed to use the auth" without having to write all the boilerplate for every single class :thinking:

calebcartwright (approved):
lgtm, one somewhat tangential inline thought below

services/opencollective/opencollective-base.spec.js
chris48s: Yeah it is a fair point, but the pattern here is not not unique to this service or PR. We use the same approach in other places e.g: https://github.com/badges/shields/blob/master/services/stackexchange/stackexchange-base.spec.js

I've raised https://github.com/badges/shields/issues/9493 as I agree it is worth improving

chris48s:
I'm going to hold off merging this just yet as I don't think I've set the OpenCollective tokens in staging/prod yet. Will sort that out and get a deploy done soon though.

PR #9408, opened 2023-07-23
[OpenVSX] Fix OpenVSX API call for unversioned package URLs

empathicqubit:
The unversioned package URL doesn't like having a trailing slash, so I made that slash part of the optional version number, instead of always adding it.


chris48s (approved):

chris48s:
thanks

PR #9411, opened 2023-07-27
Add docstrings for version service

pr7prashant:
- Added docstrings for `version` service

### DIFF
--- a/services/version.js
+++ b/services/version.js
@@ -2,13 +2,24 @@
  * Utilities relating to generating badges relating to version numbers. Includes
  * comparing versions to determine the latest, and determining the color to use
  * for the badge based on whether the version is a stable release.
- *
  * For utilities specific to PHP version ranges, see php-version.js.
+ *
+ * @module
  */
+
 import semver from 'semver'
 import { addv } from './text-formatters.js'
 import { version as versionColor } from './color-formatters.js'
 
+/**
+ * Compares two arrays of numbers lexicographically and returns an integer value.
+ * For example, listCompare([1, 2, 3], [1, 2, 4]) would return -1 because
+ * the third element of the first array is smaller than the third element of the second array.
+ *
+ * @param {number[]} a - The first array to compare
+ * @param {number[]} b - The second array to compare
+ * @returns {number} -1 if a is smaller than b, 1 if a is larger than b, 0 if a and b are equal
+ */
 function listCompare(a, b) {
   const alen = a.length
   const blen = b.length
@@ -22,8 +33,15 @@ function listCompare(a, b) {
   return alen - blen
 }
 
-// Take string versions.
-// -1 if v1 < v2, 1 if v1 > v2, 0 otherwise.
+/**
+ * Compares two strings representing version numbers lexicographically and returns an integer value.
+ * For example, compareDottedVersion(“1.2.3”, “1.2.4”) would return -1 because
+ * the numeric part of the first version is smaller than the numeric part of the second version.
+ *
+ * @param {string} v1 - The first version to compare
+ * @param {string} v2 - The second version to compare
+ * @returns {number} -1 if v1 is smaller than v2, 1 if v1 is larger than v2, 0 if v1 and v2 are equal
+ */
 function compareDottedVersion(v1, v2) {
   const parts1 = /([0-9.]+)(.*)$/.exec(v1)
   const parts2 = /([0-9.]+)(.*)$/.exec(v2)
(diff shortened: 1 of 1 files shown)


services/version.js
```diff
@@ -64,6 +88,17 @@ function latestDottedVersion(versions) {
   return version
 }
 
+/**
+ * Finds the largest version number lexicographically or semantically from an array of strings representing version numbers and returns it as a string.
+ * For example, latestMaybeSemVer([“1.2.3”, “1.2.4”, “1.3”, “2.0”], false) would return “2.0” because it is the largest version number and pre-release versions are excluded.
+ * latestMaybeSemVer([“1.2.3”, “1.2.4”, “1.3”, “2.0”], true) would also return “2.0” because pre-release versions are included but none of them are present in the array.
+ * latestMaybeSemVer([“1.2.3”, “1.2.4”, “1.3-alpha”, “2.0-beta”], false) would return “1.2.4” because pre-release versions are excluded and it is the largest version number among the remaining ones.
+ * latestMaybeSemVer([“1.2.3”, “1.2.4”, “1.3-alpha”, “2.0-beta”], true) would return “2.0-beta” because pre-release versions are included and it is the largest version number.
```
chris48s, line 96: Can we replace all the curly quote characters in the code examples with standard quote characters so the code samples could be copy & pasted out. `var a = “1.2.3”` is a `SyntaxError: Invalid or unexpected token`. Also, if you put the code samples in backticks, they'll render as a `<code></code>` in the docs. I've commented on it here, but most comments in this file have the curly quote characters.

services/version.js
```diff
@@ -147,11 +200,31 @@ function slice(v, releaseType) {
   }
 }
 
+/**
+ * Returns the lowest possible version that satisfies a given semver range expression.
+ * For example, rangeStart(“^1.2.3”) would return “1.2.3” because it is the version string of the first comparator in the first set of the range object, which is >=1.2.3 and <2.0.0.
+ *
+ * @param {string} v - The semver range to evaluate
+ * @returns {string|null} The version string of the first comparator in the first set of the range object, or null if the range is not valid
```
chris48s, line 208: When does this return `null`? I'd have thought an invalid semver range would throw `TypeError: Invalid SemVer Range` when we try to construct it.

services/version.js
```diff
@@ -147,11 +200,31 @@ function slice(v, releaseType) {
...
 function rangeStart(v) {
   const range = new semver.Range(v, /* loose */ true)
   return range.set[0][0].semver.version
 }
 
+/**
+ * Creates a badge object that displays information about a version number.
+ * For example, renderVersionBadge({version: “1.2.3”, tag: “alpha”, defaultLabel: “npm”}) would return {label: “npm@alpha”, message: “v1.2.3”, color: “orange”} because
+ * it uses the tag and the defaultLabel to create the label, the addv function to add a “v” prefix to the version in message,
+ * and the versionColor function to assign an orange color based on the version.
```
chris48s, line 219: Can we flag here that `renderVersionBadge` is the function that should usually be used to output a version badge.

services/version.js
```diff
@@ -86,8 +121,18 @@ function latestMaybeSemVer(versions, pre) {
   return version
 }
 
-// Given a list of versions (as strings), return the latest version.
-// Return undefined if no version could be found.
+/**
+ * Finds the largest version number lexicographically or semantically from an array of strings representing version numbers and returns it as a string.
```
chris48s, line 125: Can we differentiate `latest()` and `latestMaybeSemVer()` a bit more. It isn't really clear from the docstrings what the difference is.

`latestMaybeSemVer()` is used for versions that match some kind of dotted version pattern. `latest()` is looser and will attempt to make sense of anything, falling back to alphabetic sorting.

Also worth noting that `latest()` is the one we should normally use when adding version badges.

chris48s (commented):
thanks for picking this up

pr7prashant:
Hey @chris48s, I've made all the suggested changes but I couldn't get the backticks to render as `<code></code>` in the docs so I used `@example` tag instead to make the code more discrete.

services/version.js
chris48s: this one is still outstanding

chris48s:
Cheers. There's one still to follow up on, I think: https://github.com/badges/shields/pull/9411#discussion_r1278585596

> I've made all the suggested changes but I couldn't get the backticks to render as <code></code> in the docs so I used @example tag instead to make the code more discrete.

Using `@example` works fine :+1: but here's a working example using backticks and the rendered output:

- https://github.com/badges/shields/blob/7d966ab6bd340ce358a733c9db40900f3650a573/core/base-service/openapi.js#L357-L380
- http://contributing.shields.io/module-core_base-service_openapi.html#~pathParams

services/version.js
pr7prashant: done

pr7prashant:
> 

Oh my bad, I was using single backtick. It renders a code block in description but this does not render code block inside the `@example` tag. I hope the `@example` tag is fine. Please let me know if we need to change. 

Here is an example how it looks in description vs `@example` with backticks:
image

services/version.js
pr7prashant: updated

chris48s (approved):

PR #9422, opened 2023-07-30
Migrate from [DocsRs]' builds API to status API

Nemo157:
To simplify refactoring how our builds are processed I want to remove the `/builds.json` endpoint on docs.rs. As a replacement [I added](https://github.com/rust-lang/docs.rs/pull/2147) a new endpoint simply giving the status that sites like shields.io wants: are there docs for this version available to link to. This also fixes some small issues with the old endpoint (some edgecases meant the first build was successful but there were no docs, or there were docs but the first build was marked unsuccessful).

### DIFF
--- a/services/docsrs/docsrs.service.js
+++ b/services/docsrs/docsrs.service.js
@@ -1,14 +1,9 @@
 import Joi from 'joi'
 import { BaseJsonService } from '../index.js'
 
-const schema = Joi.array()
-  .items(
-    Joi.object({
-      build_status: Joi.boolean().required(),
-    }),
-  )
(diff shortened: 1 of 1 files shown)


chris48s (approved):

chris48s:
Thanks. Merging :+1: 

Just a quick comment on the endpoint: It is not a problem for us because we just look at the status code and discard the body for the not found cases, but it would be nicer if the 404 responses like https://docs.rs/crate/not-a-crate/latest/status.json and https://docs.rs/crate/regex/not-a-version/status.json returned a JSON response rather than HTML.

PR #9424, opened 2023-07-30
add a blog post about GH Tag/Release filter param

chris48s:


frontend/blog/2023-07-29-tag-filter.md
```diff
@@ -0,0 +1,19 @@
...
+title: Applying filters to GitHub Tag and Release badges
+authors:
+  name: chris48s
+  title: Shields.io Core Team
+  url: https://github.com/chris48s
+  image_url: https://avatars.githubusercontent.com/u/6025893
+tags: []
+---
+
+We recently shipped a feature which allows you to pass an arbitrary filter to the github tag and release badges. The `filter` param can be used to apply a filter to the project's tag or release names before selecting the latest from the list. Two constructs are available: `*` is a wildcard matching zero or more characters, and if the pattern starts with a `!`, the whole pattern is negated.
```
calebcartwright, line 12: might be worth proper casing GitHub in this context since it seems to be more of a brand/product reference than a route path, and would also be more consistent with subsequent usage later in the post that is capitalized

```suggestion
We recently shipped a feature which allows you to pass an arbitrary filter to the GitHub tag and release badges. The `filter` param can be used to apply a filter to the project's tag or release names before selecting the latest from the list. Two constructs are available: `*` is a wildcard matching zero or more characters, and if the pattern starts with a `!`, the whole pattern is negated.
```

calebcartwright (approved):
Super minor comment, no strong feelings either way though

PR #9425, opened 2023-07-31
fix [github] service tests

chris48s:

### DIFF
--- a/services/github/github-downloads.service.js
+++ b/services/github/github-downloads.service.js
@@ -247,7 +247,7 @@ export default class GithubDownloads extends GithubAuthV3Service {
     }
 
     if (releases.length === 0) {
-      throw new NotFound({ prettyMessage: 'no releases' })
+      throw new NotFound({ prettyMessage: 'no releases found' })
     }
 
     const { downloads } = this.constructor.transform({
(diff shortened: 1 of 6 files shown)


services/github/github-downloads.service.js
```diff
@@ -247,7 +247,7 @@ export default class GithubDownloads extends GithubAuthV3Service {
     }
 
     if (releases.length === 0) {
-      throw new NotFound({ prettyMessage: 'no releases' })
+      throw new NotFound({ prettyMessage: 'no releases found' })
```
chris48s, line 250: The issue here is that `fetchLatestRelease()` (which we also call in this service) throws "no releases found". I've brought the message for both services into line.

services/github/github-search.tester.js
```diff
@@ -8,4 +8,4 @@ t.create('hit counter')
 
 t.create('hit counter for nonexistent repo')
   .get('/badges/puppets/async%20handle.json')
-  .expectBadge({ label: 'counter', message: 'repo not found' })
+  .expectBadge({ label: 'async handle counter', message: '0' })
```
chris48s, line 11: This query no longer throws a 404 for repo doesn't exist - it just says zero results

PyvesB: Well, this is unfortunate.

PyvesB (approved):
Thanks for updating these!

PR #9426, opened 2023-07-31
allow calling [github] without auth

chris48s:
Moved to https://github.com/badges/shields/pull/9427

Its the same PR, but on a local branch. That allows us to use a PAT. The combination of this and https://github.com/badges/shields/pull/9425 hit the rate limits

### DIFF
--- a/services/github/github-api-provider.integration.js
+++ b/services/github/github-api-provider.integration.js
@@ -17,11 +17,32 @@ describe('Github API provider', function () {
 
   let githubApiProvider
 
-  context('without token pool', function () {
+  context('with no auth', function () {
     before(function () {
       githubApiProvider = new GithubApiProvider({
         baseUrl,
-        withPooling: false,
+        authType: GithubApiProvider.AUTH_TYPES.NO_AUTH,
+      })
+    })
+
+    it('should be able to run 10 requests', async function () {
+      this.timeout('20s')
+      for (let i = 0; i < 10; ++i) {
+        const { res } = await githubApiProvider.fetch(
+          fetch,
+          '/repos/rust-lang/rust',
+          {},
+        )
+        expect(res.statusCode).to.equal(200)
+      }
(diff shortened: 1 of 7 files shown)


services/github/github-auth-service.js
```diff
@@ -63,7 +63,14 @@ class GithubAuthV4Service extends BaseGraphqlService {
       `,
     )
 
-    return super._requestGraphql({ ...attrs, ...{ url, query } })
+    return super._requestGraphql({
+      ...attrs,
+      ...{
+        url,
+        query,
+        httpErrorMessages: { 401: 'auth required for graphql api' },
```
chris48s, line 71: The v4/GraphQL API doesn't allow anonymous access - a token is required

services/github/github-search.service.js
```diff
@@ -49,7 +49,11 @@ export default class GithubSearch extends GithubAuthV3Service {
         },
       },
       schema,
-      httpErrors: httpErrorsFor('repo not found'),
+      httpErrors: {
+        401: 'auth required for search api',
```
chris48s, line 53: The search API doesn't allow anonymous access - a token is required

services/github/github-api-provider.integration.js
```diff
@@ -30,7 +51,12 @@ describe('Github API provider', function () {
     it('should be able to run 10 requests', async function () {
       this.timeout('20s')
       for (let i = 0; i < 10; ++i) {
-        await githubApiProvider.fetch(fetch, '/repos/rust-lang/rust', {})
+        const { res } = await githubApiProvider.fetch(
+          fetch,
+          '/repos/rust-lang/rust',
+          {},
+        )
+        expect(res.statusCode).to.equal(200)
```
chris48s, line 59: When I was working on this, I realised this test never actually asserted the calls were successful! :grimacing:

PR #9427, opened 2023-07-31
allow calling [github] without auth

chris48s:
Closes https://github.com/badges/shields/issues/2754
Refs https://github.com/badges/shields/pull/9386#discussion_r1264705287

This PR resolves a long-standing issue that even though GitHub allows anonymous access to the v3 API with a reasonable rate limit, you have to set a key locally to do any development on any GitHub badges. The changes in https://github.com/badges/shields/pull/9386 make it extra desirable to be able to call the GH API without auth if possible.

I've set this PR to run the github test suite. Five of them will fail, but I submitted another PR to fix the tests at https://github.com/badges/shields/pull/9425 so we could merge that first then rebase this onto it, although I think the integration tests are actually the critical ones here.

### DIFF
--- a/services/github/github-api-provider.integration.js
+++ b/services/github/github-api-provider.integration.js
@@ -17,11 +17,32 @@ describe('Github API provider', function () {
 
   let githubApiProvider
 
-  context('without token pool', function () {
+  context('with no auth', function () {
     before(function () {
       githubApiProvider = new GithubApiProvider({
         baseUrl,
-        withPooling: false,
+        authType: GithubApiProvider.AUTH_TYPES.NO_AUTH,
+      })
+    })
+
+    it('should be able to run 10 requests', async function () {
+      this.timeout('20s')
+      for (let i = 0; i < 10; ++i) {
+        const { res } = await githubApiProvider.fetch(
+          fetch,
+          '/repos/rust-lang/rust',
+          {},
+        )
+        expect(res.statusCode).to.equal(200)
+      }
+    })
+  })
+
+  context('with global token', function () {
+    before(function () {
+      githubApiProvider = new GithubApiProvider({
+        baseUrl,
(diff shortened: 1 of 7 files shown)


services/github/github-auth-service.js
```diff
@@ -63,7 +63,14 @@ class GithubAuthV4Service extends BaseGraphqlService {
       `,
     )
 
-    return super._requestGraphql({ ...attrs, ...{ url, query } })
+    return super._requestGraphql({
+      ...attrs,
+      ...{
+        url,
+        query,
+        httpErrorMessages: { 401: 'auth required for graphql api' },
```
chris48s, line 71: The v4/GraphQL API doesn't allow anonymous access - a token is required

services/github/github-search.service.js
```diff
@@ -49,7 +49,11 @@ export default class GithubSearch extends GithubAuthV3Service {
         },
       },
       schema,
-      httpErrors: httpErrorsFor('repo not found'),
+      httpErrors: {
+        401: 'auth required for search api',
```
chris48s, line 53: The search API doesn't allow anonymous access - a token is required

services/github/github-api-provider.integration.js
```diff
@@ -30,7 +51,12 @@ describe('Github API provider', function () {
     it('should be able to run 10 requests', async function () {
       this.timeout('20s')
       for (let i = 0; i < 10; ++i) {
-        await githubApiProvider.fetch(fetch, '/repos/rust-lang/rust', {})
+        const { res } = await githubApiProvider.fetch(
+          fetch,
+          '/repos/rust-lang/rust',
+          {},
+        )
+        expect(res.statusCode).to.equal(200)
```
chris48s, line 59: When I was working on this, I realised this test never actually asserted the calls were successful! :grimacing:

PyvesB (approved):
Changes look good to me 👍🏻

PR #9429, opened 2023-07-31
migrate `examples` to `openApi` part 3; affects [conan cookbook coverity cpan debian docker docsrs dub eclipse]

chris48s:
Refs #9285

- only path params
- no documentation string


services/depfu/depfu.service.js
```diff
@@ -13,16 +18,23 @@ class Depfu extends BaseJsonService {
     pattern: ':vcsType(github|gitlab)/:project+',
```
chris48s, line 18: TODO: enum

chris48s: I've just reverted this for now

calebcartwright:
I just had a thought that I wish I'd had sooner. I know we've had some discussions about potential options to create helper functions that can return the objects needed for the open api spec, as well as some of the tradeoffs around helper functions/abstractions vs. explicit spec definitions.

However, I feel like we've been discussing that option mostly in the context of routes, and I cant' remember whether we've discussed something similar for the route parameters. Specifically, should we consider adjusting the path/route param helper functions to accept the same form of key/value pairs we've used historically as opposed to each class having to explicitly designate `name` and `example` fields with the respective values?

I don't feel terribly strongly, but it seems like the type of thing we should decide for at least the near to mid future before we get through too many of these conversions

calebcartwright:
E.g. instead of:

```js
pathParams({
  name: 'packageName',
  example: 'boost',
})
```

perhaps:
```js
pathParams({ packageName: 'boost' })
```

chris48s:
So that works in the case where we only care about `name` and `example`, but there are other keys in this object: `schema` and `description`.

In https://github.com/badges/shields/pull/9437 (probably the next one to review tbh) we begin using `schema`.

Currently we don't use `description` because documentation for individual params was not a feature we previously had so all of this is just jammed in the top-level `documentation`/`description`. I'd like to start using it to move documentation about individual params out of the top-level `documentation`/`description` key and onto the relevant params as we go though.

I did use this feature when I set up the new docs for the dynamic badges though. For example https://shields.io/badges/dynamic-json-badge

Screenshot at 2023-08-16 21-22-35

calebcartwright:
Fair enough, it feels like it's more verbose than it could be, but I can also live with it given the schema being broader than what I'd seen/anticipated

calebcartwright (approved):

PR #9430, opened 2023-07-31
migrate `examples` to `openApi` part 4; affects [ecologi elm flathub gem gitter GithubTotalDiscussions greasyfork hackage hackernews homebrew]

chris48s:
Refs #9285

- only path params
- no documentation string


services/homebrew/homebrew-downloads.service.js
```diff
@@ -38,13 +38,23 @@ export default class HomebrewDownloads extends BaseJsonService {
     pattern: 'installs/:interval(dm|dq|dy)/:formula',
```
chris48s, line 38: TODO: enum

chris48s: I've just reverted this for now

uncenter (approved):

PyvesB (approved):
Looks good to me 👍🏻

PR #9431, opened 2023-07-31
migrate `examples` to `openApi` part5; affects [itunes jetbrains jitpack keybase lemmy luarocks maintenance openvsx]

chris48s:
Refs #9285

- only path params
- no documentation string


services/jsdelivr/jsdelivr-hits-github.service.js
```diff
@@ -6,17 +7,27 @@ export default class JsDelivrHitsGitHub extends BaseJsDelivrService {
     pattern: ':period(hd|hw|hm|hy)/:user/:repo',
```
chris48s, line 7: TODO: enum

chris48s: _sigh_ and also work out why the liberapay tests are failing (they pass locally)

chris48s: I've just reverted this for now

chris48s:
The tests for liberapay pass locally.
When I try to run them in GitHub actions, they fail.
It seems that when we call liberapay from GH actions, instead of a JSON response we get a 403 error with a javascript challenge (designed to prevent bots/scrapers/etc). You can see the output of `npm run badge https://img.shields.io/liberapay/gives/Changaco` invoked from GitHub actions here https://github.com/badges/shields/actions/runs/5730805389/job/15530375550

I'm not going to attempt to address that here. I've just removed it from the list of services we're testing against.

services/jetbrains/jetbrains-downloads.service.js
```diff
@@ -29,15 +30,17 @@ export default class JetbrainsDownloads extends JetbrainsBase {
...
 
-  static examples = [
-    {
-      title: 'JetBrains plugins',
-      namedParams: {
-        pluginId: '1347',
+  static openApi = {
+    '/jetbrains/plugin/d/{pluginId}': {
+      get: {
+        summary: 'JetBrains plugin Downloads',
```
PyvesB, line 36:
```suggestion
        summary: 'JetBrains Plugin Downloads',
```

For consistency with the other JetBrains badge :)

PyvesB (approved):
Only a small nit, feel free to merge whenever 👍🏻

PR #9437, opened 2023-08-01
WIP migrate `examples` to `openApi` part N: enums

chris48s:
WIP TODO

### DIFF
--- a/services/codefactor/codefactor-grade.service.js
+++ b/services/codefactor/codefactor-grade.service.js
@@ -1,5 +1,5 @@
 import Joi from 'joi'
-import { BaseSvgScrapingService } from '../index.js'
+import { BaseSvgScrapingService, pathParams } from '../index.js'
 import { isValidGrade, gradeColor } from './codefactor-helpers.js'
 
 const schema = Joi.object({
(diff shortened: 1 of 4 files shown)


core/base-service/base.js
```diff
@@ -102,6 +103,26 @@ class BaseService {
...
   }
 
+  /**
+   * Extract an array of allowed values from this service's route pattern
+   * for a given route parameter
+   *
+   * @param {string} param The name of a param in this service's route pattern
+   * @returns {string[]} Array of allowed values for this param
+   */
+  static getEnum(param) {
```
chris48s, line 113: I was in two minds about where to put this.

On the one hand:
- We need to reference `this.route` so doing this in a method on `BaseService` kinda makes sense
- What we're doing here is basically extracting an array `['this', 'that']` from a string `':pattern(this|that)`. It isn't really OpenAPI specific

On the other hand:
- We need to import `pathParam`, `pathParams`, `queryParam`, `queryParams` from `openapi.js`, so why is the `getEnum` method on `BaseService`?
- An alternative implementation would either be to:
    - move all those existing helpers to `BaseService` as well, so everything lives on the base class
    - move this to a standalone helper in `openapi.js` so that we call `getEnum(this.route, 'paramName')` instead of `this.getEnum('paramName')`

I don't have really strong preferences on this but I think it is worth thinking about this now while the number of things we have to update is still relatively small. Also consider once I move on to more complicated examples, we will probably add some more helpers which look at `this.route` (and may be more OpenAPI specific).

PyvesB: No super strong opinion, I think the way you've done it is fine.

services/github/github-issue-detail.service.js
```diff
@@ -179,35 +179,37 @@ export default class GithubIssueDetail extends GithubAuthV3Service {
...
+        summary: 'GitHub issue/pull request detail',
+        description: documentation,
+        parameters: pathParams(
+          {
+            name: 'issueKind',
+            example: 'issues',
+            schema: { type: 'string', enum: this.getEnum('issueKind') },
+          },
+          {
+            name: 'property',
```
PyvesB, line 194: Shouldn't this one be an enum as well?

chris48s: Yes - well spotted. Thanks

PyvesB (approved):

Issue #9410, opened 2023-07-25
Python versions from pyproject.toml


HealthyPear:
### 📋 Description

- service-independent
- should show required range of python versions same as the one for PyPI

### 🔗 Data

any `pyproject.toml` file, from the `requires-python` field

### 🎤 Motivation

should be even easier than the one on PyPI and it will work as a dynamic badge for people who are working on packages that are either public or private, but not a public package

chris48s:
Couple of notes:

This would be for a PEP 621-compliant `pyproject.toml` Not all packageing tools follow the spec. For example, poetry stores this in `tool.poetry.dependencies.python`, rather than `.requires-python`

Currently shields has base classes for dealing with requests that return JSON, XML and Yaml. We don't have one for Toml so part of the job would be to add a toml parser and base class for it. This is not a blocker and would be useful to have anyway, but we don't currently have it as a building block.

jNullj:
Hey, i could help.
I guess there is no point in writing a parser from scratch just like we use parsers from other modules for yaml, xml etc...
The toml site refers to a 3 packages that are complaint with tomlv1.0.0 (latest):

| name |
|--------|
| [smol-toml](https://www.npmjs.com/package/smol-toml) | 
| [toml-nodejs](https://www.npmjs.com/package/toml-nodejs) |
| [@ltd/j-toml](https://www.npmjs.com/package/@ltd/j-toml) |

smol-toml is relatively new but seems easy to use, boasts big performance and appears to be easier to implement.

All of those are hardly in use thats why i also took a look at packages that don't comply with latest toml spec.
There is also [toml](https://www.npmjs.com/package/toml) which is way more in use (900k weekly downloads, 240k github projects) but it only supports toml v0.4.0. could this be an issue? it seems to be very old, i fear newer nodejs versions might conflict with it in the future.

Should i got ahead with smol-toml, This would add a new dependency to the project.

chris48s:
Yeah we should definitely use a package and not write our own parser.

---

> There is also toml which is way more in use (900k weekly downloads, 240k github projects) but it only supports toml v0.4.0. could this be an issue? it seems to be very old, i fear newer nodejs versions might conflict with it in the future.

Yes. This from the readme does not give me high confidence this lib is being actively maintained :smile: 

> toml-node is tested on Travis CI and is tested against:
>
> - Node 0.10
> - Node 0.12
> - Latest stable io.js

Lets pass on this one.

---

Of the other packages you've suggested:

[...]

PR #9438, opened 2023-08-04
Add TOML support with BaseTomlService

jNullj:
Add base toml service to enable fetch of toml files
Add spec file for the new toml service for automated testing
This was added to allow a new way to retrive python version from pyproject.toml as described in issue #9410 

Note: this adds a new dependency smol-toml which requires nodejs 18 or newer. i was testing this in nodejs 20 on my machine.
Note: when parsed, toml dates are a new extended date object (TomlDate) due to limitations in native js date, read more [here](https://github.com/squirrelchat/smol-toml/tree/mistress#dates).


jNullj:
Seems to fail due to node16 in github workflow.
I will keep this draft until #9385 is merged

chris48s:
Thanks. I've now merged the node 18 PR so you can update this branch to finish it off.

I've removed `[BaseTomlService]` from the PR title. The base classes are just covered in the core tests - there are no service tests to run.

I'll have a proper look over this at the weekend

core/base-service/base-toml.js
```diff
@@ -0,0 +1,82 @@
...
+    options = {},
+    httpErrors = {},
+    systemErrors = {},
+  }) {
+    const logTrace = (...args) => trace.logTrace('fetch', ...args)
+    const mergedOptions = {
+      ...{
+        headers: {
+          Accept:
+            // the officeal header should be application/toml - see https://toml.io/en/v1.0.0#mime-type
```
chris48s, line 51:
```suggestion
            // the official header should be application/toml - see https://toml.io/en/v1.0.0#mime-type
```

core/base-service/base-toml.spec.js
```diff
@@ -0,0 +1,150 @@
...
+# example toml - unexpected val
+unexpectedKey = "some-string"
+`
+
+const invalidToml = `
+# example invalid toml
+missing= "space"
+colonsCantBeUsed: 42
+missing "assignment"
+`
```
chris48s, line 38: I find the use of "expected" and "unexpected" a bit unclear here. Particularly given that in the context of unit tests, we'd also talk about the "expected" value as the thing we want to compare the observed value to.

I also appreciate that _naming things is hard_ here because we've got two very different "invalid" cases. There is:

- "invalid toml" - legal toml syntax that doesn't validate against the schema and
- "invalid toml" - illegal toml syntax that can't be parsed

I think we could probably express this more clearly by calling these 3

- `validToml`
- `invalidToml`
- `malformedToml`

or

- `validToml`
- `invalidToml`
- `unparseableToml`

chris48s (commented):
Couple of comments, but this is mostly there. Thanks

core/base-service/base-toml.spec.js
jNullj: Solved with https://github.com/badges/shields/pull/9438/commits/f762a3d22600870bcd592bf2007456ec4219d038
I changed a bit the names, hope those make it as clear as it can get

`invalidSchemaToml` . `invalidTomlSyntax`

chris48s: Yes. Awesome - thanks

chris48s (approved):

PR #9452, opened 2023-08-08
Fix puppetforge module feedback badges

smortex:
The forge has stopped to collect user feedback a long time ago. However, it gained the capability to now run static analysis on modules and give it a score.  This score has been work in progress for some time will probably change again in the future, and might include back user feedback in the future, but as this is moving quite slow, rather than waiting to unbreak this badge, we can switch to the current API endpoint that provide this scoring as a drop-in replacement for user feedback.

Since the badge is already labelled "score", this seems a reasonable move.

### DIFF
--- a/services/puppetforge/puppetforge-base.js
+++ b/services/puppetforge/puppetforge-base.js
@@ -24,6 +24,8 @@ const modulesSchema = Joi.object({
   ),
 }).required()
 
+const moduleValidationsSchema = Joi.array().required()
+
 class BasePuppetForgeUsersService extends BaseJsonService {
   async fetch({ user }) {
     return this._requestJson({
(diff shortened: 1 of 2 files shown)


chris48s:
When you say, they stopped collecting user feedback long ago, if I call
https://img.shields.io/puppetforge/f/camptocamp/openssl
for example, I still get a valid score back. So that is not "broken" to me. I guess that feedback is historical - no new data is being added to it, but it is still represents a true data point from the past.

I'd would say that "Puppet Forge Static Analysis Score" is just a completely new/different badge route. We can totally add that - feel free to modify the PR to just add that as a new badge.

It is a different thing from user feedback though. Particularly given that collecting feedback may be re-introduced at a later date, we're painting ourselves into a corner if we've re-used the route for a different concept.

PyvesB:
@smortex any chance you could address Chris's feedback in the coming weeks? :)

smortex:
@PyvesB d'oh, I forgot about this.  I opened #10201 to address it. Closing this one.

PR #9454, opened 2023-08-08
Use correct latest crates.io version

### LINKED ISSUE #9453: Incorrect version shown for crates.io

shields.io

### 🐞 Description

The [martin crate](https://crates.io/crates/martin), currently at v0.8.7, at one point published a v1.0.0-alpha version that was later yanked. The crates.io API returns versions in semver-DESC order, which means the yanked alpha version is accessed as the most recent one by this code (license access):

https://github.com/badges/shields/blob/7d966ab6bd340ce358a733c9db40900f3650a573/services/crates/crates-license.service.js#L38

Relevant portion of the API response for https://crates.io/api/v1/crates/martin

```json
{
  "versions": [
    {
      "created_at": "2021-10-18T13:32:02.600041+00:00",
      "num": "1.0.0-alpha.0",
      "updated_at": "2022-11-21T03:56:25.244956+00:00",
      "yanked": true
    },
    {
      "created_at": "2023-06-23T04:05:38.815265+00:00",
      "num": "0.8.7",
      "updated_at": "2023-06-23T04:05:38.815265+00:00",
      "yanked": false
    }
  ]
}
```

### 🔗 Link to the badge

https://img.shields.io/crates/l/martin.svg

### 💡 Possible Solution

I believe it would be better to pick the first non-yanked version from the array instead, plus probably parse the version value and ignore all the pre-released too.

nyurik:
For download count and license, get the correct latest version from the versions array.

Fix #9453


calebcartwright:
I think it would probably be best to first take the simplest route with the existing problem on the yanked version (first non-yanked version), and then have a follow up discussion/potential PR for any logic changes.

calebcartwright:
See also #8666, as it may be worth overhauling the selection logic and encapsulating it somewhere for usage across the relevant classes

nyurik:
@calebcartwright thx for the link, it does confirm the same type of logic I used as used by the crates.io itself [to compute](https://github.com/rust-lang/crates.io/blob/f9c7a4d9d409065133fcc6a66d5a9d6f1ea50e03/src/models/version.rs#L83C35) latest stable vs latest, etc.

I do not think filtering out yanked versions is the right approach.  I think everyone's expectation would be that shields show the same version as the default page on crates.io for all non-version-specific requests.  The version shield already follows that logic, but the download count and license ones do not.  Unfortunately I am not as familiar with ts as I would like, so I might not be the best person for overall code restructuring.

Here's the [current logic](https://github.com/rust-lang/crates.io/blob/20bbac9f5521cb4888ef63f8464fddb28fd973f5/app/models/crate.js#L34-L41) of crates.io:

```js
  /**
   * This is the default version that will be shown when visiting the crate
   * details page. Note that this can be `undefined` if all versions of the crate
   * have been yanked.
   * @return {string}
   */
  get defaultVersion() {
    if (this.max_stable_version) {
      return this.max_stable_version;
    }
    if (this.max_version && this.max_version !== '0.0.0') {
      return this.max_version;
    }
  }
```

I think shields should implement the same function in the base class, plus there should be another function `defaultVersionObject` which finds the `.versions` array entry that corresponds  to the one returned by above `defaultVersion()`.

Then all 3 classes would call it and act accordingly.

nyurik:
oops, i think there was a in-flight issue - i tried to address the comments

nyurik:
After a few back and forth, I think i got the more consolidated version - where the base class handles both the `defaultVersion() -> string` and `defaultVersionObject() -> obj`, and the 3 derived classes just use those functions directly. Seems far cleaner, and more consistent with the expectations

nyurik:
@calebcartwright  hi, are there any blockers for this PR? Anything I can improve?

nyurik:
CC: @chris48s - not sure if that's the right person though, my apologies if not

chris48s:
I don't mind jumping in on this one if I need to, but given @calebcartwright

- has mad all the comment on this PR and https://github.com/badges/shields/issues/9453 so far
- is actively involved in the rust community

I'm assuming he will do the final review on this one.

calebcartwright:
Thanks @nyurik. We're a rather smaller maintainer team, so it's basically an "any maintainer, as and when they have time" review strategy vs. any specific person.

I haven't had a chance to go through the updated code yet, but will also note in response to expectations around versions is that one thing we have to balance is that Shields also has its own behavior around our version badges and we strive for consistency in our badges/badge behavior across the various services and ecosystem.

To be clear, I'm not pushing back on anything nor prescribing anything specific at this point, just noting that we have other factors and constraints our badge defaults work within separate from the target service we fetch data from.

nyurik:
@calebcartwright thx for the response. I am not too certain I understood it though - shields already uses crate.io-provided "current version" as the default for one of the shields (get current version). What changes are you saying this PR should make in order to be merged?  How would this be different from other shields?  Thx!

PyvesB:
@calebcartwright any chance you could take another peak at this one? :)

PyvesB:
This PR has stalled for a while, sorry about this @nyurik. I've not got the full context, but I believe PRs such #9883, #9949, and perhaps others, have in the meantime addressed both some of the complexity concerns and the behavioural change around yanked versions.

PR #9463, opened 2023-08-11
migrate examples to openApi part 8; affects [ubuntu uptimerobot vaadin vcpkg visualstudiomarketplace wheelmap]

chris48s:
Refs #9285

- only path params
- no documentation string


PyvesB (approved):
Looks good to me 👍🏻

PR #9464, opened 2023-08-11
migrate examples to openApi part 7; affects [reuse sourceforge sourcegraph spack stackexchange testspace treeware twitch]

chris48s:
Refs #9285

- only path params
- no documentation string

### DIFF
--- a/services/sourceforge/sourceforge-commit-count.service.js
+++ b/services/sourceforge/sourceforge-commit-count.service.js
@@ -17,7 +17,7 @@ export default class SourceforgeCommitCount extends BaseJsonService {
   static openApi = {
     '/sourceforge/commit-count/{project}': {
       get: {
-        summary: 'SourceForge commit count',
+        summary: 'SourceForge Commit Count',
         parameters: pathParams({
           name: 'project',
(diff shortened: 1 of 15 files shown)


services/sourceforge/sourceforge-commit-count.service.js
```diff
@@ -14,17 +14,17 @@ export default class SourceforgeCommitCount extends BaseJsonService {
...
 
-  static examples = [
-    {
-      title: 'SourceForge commit count',
-      namedParams: {
-        project: 'guitarix',
+  static openApi = {
+    '/sourceforge/commit-count/{project}': {
+      get: {
+        summary: 'SourceForge commit count',
```
PyvesB, line 20: Nit: we seem to prefer capitalising all words within the badge title, even though we're not consistent across the board. We should probably agree on one or the other. In this particular instance, there's inconsistency within different SourceForge badges themselves, so maybe we could at least fix these? Feel free to dismiss if you don't want to do this now, after all this inconsistency was already present before.

PyvesB (approved):
Looks good to me!

PR #9466, opened 2023-08-11
migrate examples to openApi part 10

chris48s:
Refs #9285

This one doesn't have passing service tests and I don't have the examples to hand to fix them, so I'm going to push a commit that does nothing, let the tests fail, then push a commit that does stuff so we can at least see the failures aren't new.

In general, this whole process probably will require modifying more code that doesn't have (reliably) passing service tests.


PyvesB (approved):
Looks good to me, we can circle back to fixinf failures later :+1:

Issue #9287, opened 2023-06-18
CI automation to help upgrading docusaurus-theme-openapi


chris48s:
:clipboard: **Description**

There are two React components from docusaurus-theme-openapi which we have [swizzled](https://docusaurus.io/docs/swizzling) to customise them for our needs.

This means that when we upgrade the docusaurus-theme-openapi, if either of the components we have customised (`Curl` and `Response`) have changed in the upstream package, we need to review our custom components to see if any changes or updates are needed.

I want to write a GitHub action that can flag dependabot PRs bumping docusaurus-theme-openapi that require manual intervention to review or update these components. Something like close-bot is probably a reasonable starting point for what this should look like.

calebcartwright:
> I want to write a GitHub action that can flag dependabot PRs bumping docusaurus-theme-openapi that require manual intervention to review or update these components. Something like close-bot is probably a reasonable starting point for what this should look like.

Is this proposed (I presume first step) flagging in the sense of essentially just posting some "hey Shields maintainer, pay close attention to this one" comment on the dependabot PR?

chris48s:
I think the easiest thing (so probably the first step) is to just do something like (as you say)

> if packageName == 'docusaurus-theme-openapi' then post a comment on the PR saying "hey Shields maintainer, pay close attention to this one"

But I think we could look at more elaborate stuff. One option would be to post a diff of the files from the package against the ones from our repo, or diff the specific files in the old/new package versions and say if they have changed or not.

jNullj:
Could give a hand with this implanting, assign me if you want.

chris48s:
If you wanted to pick this one up, that would be awesome. I think for a first implementation, just adding a warning to the PR saying:

- This package contains components we've overridden
- We need to watch out for changes to the `Curl` and `Response` components

is a great start.

https://github.com/badges/shields/tree/master/.github/actions/close-bot is probably a good point of reference for a similar-ish action we have written.

jNullj:
So here is what i came up with (The old and new are mixed due to the way i am testing, it should not be an issue)
![image](https://github.com/badges/shields/assets/15849761/d73145ef-cf72-44c9-8892-85d19be1cc2a)
It will inform which components are changed, the list on the last row is dynamic, the script tests what files are changed in the diff of the tags for the two versions.

[...]

PR #9467, opened 2023-08-11
devops: Add ci warning for swizzled docusaurus components

jNullj:
Add new action and workflow that runs on new dependabot PR.
The action checks changes in components of docusaurus-theme-openapi  that are swizzled and warns maintainers to review those changes.
For more info see https://github.com/badges/shields/issues/9287

image

Solves https://github.com/badges/shields/issues/9287


.github/actions/docusaurus-swizzled-warning/action.yml
```diff
@@ -0,0 +1,12 @@
+name: 'Docusaurus swizzled component changes warning'
```
chris48s, line 1:
```suggestion
name: 'docusaurus-theme-openapi swizzled component changes warning'
```

.github/actions/docusaurus-swizzled-warning/action.yml
```diff
@@ -0,0 +1,12 @@
...
+description: 'Check for changes in Docusaurus components which are swizzled and prints out a warning'
```
chris48s, line 2:
```suggestion
description: 'Check for changes in docusaurus-theme-openapi components which are swizzled and prints out a warning'
```

.github/actions/docusaurus-swizzled-warning/index.js
```diff
@@ -0,0 +1,135 @@
...
+    const packageName = 'docusaurus-theme-openapi'
+    const packageParentName = 'docusaurus-preset-openapi'
+    const overideComponents = ['Curl', 'Response']
+    const messageTemplate = `<table><thead><tr><th>
+      ⚠️ This PR contains changes to components of ${packageName} we've overridden
+    </th></tr></thead>
+    <tbody><tr><th>
+      We need to watch out for changes to the ${overideComponents.join(
+        ', ',
+      )}components
```
chris48s, line 31:
```suggestion
      )} components
```

.github/actions/docusaurus-swizzled-warning/index.js
```diff
@@ -0,0 +1,135 @@
...
+        if (file.filename !== 'package-lock.json') {
+          continue
+        }
+
+        const pkgLockNewJson = await (await fetch(file.raw_url)).json()
+        const pkgLockOldJson = await (
+          await fetch(
+            `https://raw.githubusercontent.com/${github.context.repo.owner}/${github.context.repo.repo}/master/${file.filename}`,
+          )
+        ).json()
```
chris48s, line 55: I don't think we need to pull in node-fetch as a dependency and request these URLs. We've got an OctoKit instance in scope, so I think we should be able to use:

```
octokit.rest.repos.getContent({
  owner,
  repo,
  path,
  ref?,
});
```

to replace both these calls.

.github/actions/docusaurus-swizzled-warning/index.js
```diff
@@ -0,0 +1,135 @@
...
+        client,
+        github.context.repo.owner,
+        github.context.repo.repo,
+        pr.number,
+      )
+
+      for (const file of files) {
+        if (file.filename !== 'package-lock.json') {
+          continue
+        }
```
chris48s, line 48: In general, it is harder to read large blocks of code that are deeply nested.

We can often work round nesting by taking some code like

```js
if (condition) {
  // do stuff
}
```

and re-writing it as

```js
if (!condition) {
  return
}

// do stuff
```

I think we could make this whole `run` function easier to read by getting rid of a couple of layers of nesting here. For example:

```js
if (!['dependabot[bot]', 'dependabot-preview[bot]'].includes(pr.user.login)) {
  return
}

const file = files.filter(f => f.filename === 'package-lock.json')[0]
if (file === undefined) {
   return
}

// do something with file
```

This would also make it clearer that we are not expecting to process more than one `package-lock.json`.

.github/actions/docusaurus-swizzled-warning/index.js
```diff
@@ -0,0 +1,135 @@
...
+    const { pull_request: pr } = github.context.payload
+    if (!pr) {
+      throw new Error('Event payload missing `pull_request`')
+    }
+
+    const client = github.getOctokit(token)
+    const packageName = 'docusaurus-theme-openapi'
+    const packageParentName = 'docusaurus-preset-openapi'
+    const overideComponents = ['Curl', 'Response']
+    const messageTemplate = `<table><thead><tr><th>
```
chris48s, line 25: Minor points, but:

1. If we have a look at your screenshot, we can see the table looks wierd because not all of the table rows have the same number of columns. Making these headers <th colspan="2"> will fix that.

Screenshot at 2023-08-12 19-19-02

2. We're using `<th>` for every cell in this table, but `<th>` is only for headers. The "normal" cells should be `<td>`s.

.github/actions/docusaurus-swizzled-warning/package.json
```diff
@@ -0,0 +1,17 @@
...
+    "main": "index.js",
+    "private": true,
+    "scripts": {
+        "test": "echo \"Error: no test specified\" && exit 1"
+    },
+    "author": "jNullj",
+    "license": "CC0",
+    "dependencies": {
+        "@actions/core": "^1.10.0",
+        "@actions/github": "^5.1.1",
```
chris48s, line 14: If we are adding a package.json with dependencies here, we should register it withe dependabot, like this:

https://github.com/badges/shields/blob/e9b3b505b324df5cbd72740a7376d09e66e5a4a6/.github/dependabot.yml#L33-L41

chris48s (commented):
Thanks for working on this. This looks like it will be really useful.

I've had a read over the code and left you a first pass of comments. I've not actually run the code yet though. JS actions are very powerful, but one of the big issues is that it is quite hard to just check code like this out and run it locally to make sure it works.

How have you been testing this? Did you do it all in a test repo, or were you able to run bits of the code locally?

.github/actions/docusaurus-swizzled-warning/index.js
jNullj: I am new to octokit, I will look into this and add this if applicable

jNullj:
> Thanks for working on this. This looks like it will be really useful.
> 
> I've had a read over the code and left you a first pass of comments. I've not actually run the code yet though. JS actions are very powerful, but one of the big issues is that it is quite hard to just check code like this out and run it locally to make sure it works.
> 
> How have you been testing this? Did you do it all in a test repo, or were you able to run bits of the code locally?

Thank you for the code review and the useful insights.

Regarding testing - I used a forked repo on github for testing, I think I might be able to run it locally if i look for some testing tools or try to simulate github with a crafted payload.

.github/actions/docusaurus-swizzled-warning/index.js
jNullj: Updating this, not sure what you mean at 1, but i used `colspan="2"` to fix the visual issue.

jNullj: After testing
image

.github/actions/docusaurus-swizzled-warning/package.json
jNullj: Solved with https://github.com/badges/shields/pull/9467/commits/c77af690195f3e2eb05ee04c828d1de204e6319b

.github/actions/docusaurus-swizzled-warning/index.js
jNullj: solved with https://github.com/badges/shields/pull/9467/commits/db2e03c72368ba629098e9cb1474566b8dfd15ed
notice that files are > 1MB therefor an getContent won't return content (tried even with raw)
I had to use the sha and make a new blob request.

.github/actions/docusaurus-swizzled-warning/index.js
jNullj: I improved code readability with https://github.com/badges/shields/pull/9467/commits/81f020f69639a63f0c1a4a03b31c4add3b2db0a6
then fixed a type with https://github.com/badges/shields/pull/9467/commits/9e1c2e43b7f4be4f361c316b132359fa9bfb4439

.github/actions/docusaurus-swizzled-warning/index.js
```diff
@@ -0,0 +1,146 @@
...
+        repo: github.context.repo.repo,
+        issue_number: pr.number,
+        body,
+      })
+
+      core.debug('Found changes and posted comment, done.')
+      return
+    }
+
+    core.debug('No changes found, done.')
```
chris48s, line 140: Can we make these log statements use `core.info()` instead of `core.debug()` so they always get logged.

.github/actions/docusaurus-swizzled-warning/index.js
```diff
@@ -0,0 +1,146 @@
...
+
+    // if parent dependency is higher version then existing
+    // npm install will retrive the newer version from the parent dependency
+    if (oldVersionParent > oldVersion) {
+      oldVersion = oldVersionParent
+    }
+    if (newVersionParent > newVersion) {
+      newVersion = newVersionParent
+    }
+
```
chris48s, line 104: It would be useful here if we could write a log message here (again, `core.info()`) with `oldVersion` and `newVersion`.

.github/workflows/docusaurus-swizzled-warning.yml
```diff
@@ -0,0 +1,22 @@
+name: Docusaurus swizzled component changes warning
+on:
+  pull_request_target:
```
chris48s, line 3: This workflow doesn't actually need to run on `pull_request_target`. It can just run on `pull_request`. I'm aware this is a copy & paste from close-bot, which probably could have also just used `pull_request`.

chris48s (commented):
OK. Latest changes look great - thanks. I've done a bit of testing on this in another repo. I found it a bit tricky to test this, but my experience is that when it comes to these GitHub actions, there is a large extent to which we can "set it and forget it". We don't tend to be constantly changing them. So I'm not too worried about this being a bit of a pain to test.

I've left a few final comments inline, but this is 99% done.

.github/workflows/docusaurus-swizzled-warning.yml
jNullj: Oh yea, my bad, didn't notice this when I started with close-bot as a ref for the file.
Fixed with 640567d89defc418918a8e2614289b971dc9f7b6

.github/actions/docusaurus-swizzled-warning/index.js
jNullj: Sure. This would be easier to understand.
Added with 78551fffbe5c6884663a5309ffb18b98359d78b3

.github/actions/docusaurus-swizzled-warning/index.js
jNullj: Added with 5311d0326573056d08cce2639cec430063a57ad3

chris48s (approved):
Sweeeet. Thanks for working on this one

calebcartwright:
Nicely done @jNullj!

PR #9468, opened 2023-08-11
delete close-bot

chris48s:
We originally made this to deal with pointless version bumps in gatsby and typescript-eslint monorepo packages. We don't use either of those things any more, so this has served its useful purpose. It can be deleted.


calebcartwright (approved):
guess we can always bring this back in the event we introduce some new dependency with useless bumps

Issue #9096, opened 2023-04-16
Bitbucket badge returns Invalid Response Data


HariSekhon:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

This call results in a badge saying Invalid Response Data

https://img.shields.io/bitbucket/pipelines/harisekhon/diagrams-as-code/master?logo=bitbucket&label=BitBucket%20CI

![](https://img.shields.io/bitbucket/pipelines/harisekhon/diagrams-as-code/master?logo=bitbucket&label=BitBucket%20CI)

I've compared this to much older well established repos I have which used to work and the same thing has started happening there although I'm not sure when it started but these definitely used to work, for example this one:

https://img.shields.io/bitbucket/pipelines/harisekhon/devops-bash-tools/master?logo=bitbucket&label=BitBucket%20CI

which you can see here hasn't been changed in 3 years:

https://github.com/HariSekhon/DevOps-Bash-tools/blame/master/README.md#L82

### 🔗 Link to the badge

https://img.shields.io/bitbucket/pipelines/harisekhon/diagrams-as-code/master?logo=bitbucket&label=BitBucket%20CI

calebcartwright:
> Any ideas how to debug what the actual response data is and why it is breaking?

Two common approaches:

1. Look at the Shields source code to determine the underlying API call we're making to the upstream provider (Bitbucket Cloud ostensibly in this case), then try to inspect that endpoint via your own calls/client (e.g. `curl ...`). It's possible the endpoint has changed their contract so that the response body structure no longer matches what we're expecting
2. Run Shields locally with `TRACE_SERVICES=true` environment variable set, and then make the same Bitbucket badge request to your local instance (instead of `img.shields.io`). That should, amongst various other log output content, show the response received from the API call

HariSekhon:
I think that's because my account is on the free plan which has run out of build minutes.

This must be quite a common occurrence and would probably be better handled with a grey HALTED status rather than throwing an error?

chris48s:
Yeah I think that's a reasonable way to handle it. Would you be up for submitting a PR?

reeceappling:
Put together a small PR that should resolve this. Let me know if any changes need to be made.

PR #9472, opened 2023-08-12
[bitbucketpipelines] resolve "invalid response data" for halted pipelines

reeceappling:
Should resolve issue #9096 
Bitbucket responses for pipelines in progress will now fall back to utilizing the stage name instead of being invalid.

### DIFF
--- a/services/bitbucket/bitbucket-pipelines.service.js
+++ b/services/bitbucket/bitbucket-pipelines.service.js
@@ -78,7 +78,7 @@ class BitbucketPipelines extends BaseJsonService {
       value =>
         value.state &&
         (value.state.name === 'COMPLETED' ||
-         value.state.name === 'IN_PROGRESS'),
+          value.state.name === 'IN_PROGRESS'),
     )
     if (values.length > 0) {
       if (Object.prototype.hasOwnProperty.call(values[0].state, 'result')) {
--- a/services/bitbucket/bitbucket-pipelines.service.js
+++ b/services/bitbucket/bitbucket-pipelines.service.js
@@ -21,7 +21,9 @@ const bitbucketPipelinesSchema = Joi.object({
             name: Joi.string().required(),
             type: Joi.string().optional(),
           }),
-        }).required().or('result','stage'),
+        })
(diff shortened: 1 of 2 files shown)


reeceappling:
It appears to be failing due to some missing credentials. Not sure what needs to be done about that...

calebcartwright:
Thanks for the PR!

> It appears to be failing due to some missing credentials. Not sure what needs to be done about that...

in CI we funnel the test output to a different area where it's easier to consume, the console output about the credentials is unrelated. You can find the test failures in the summary for the PR at https://github.com/badges/shields/actions/runs/5843476444/attempts/1#summary-15845658997, and imo from a quick glance, the failures do seem highly likely related to the proposed changes.

reeceappling:
Sweet, thanks for the explaination, I'll have time to check it out tomorrow.

chris48s:
I think this PR does introduce a failure.

But also unfortunately Bitbucket does have some failing tests anyway, so we are not starting from a point where everything passes with this one.

I've updated the PR title to only run the tests for BB pipelines. Lets take failing issues tests out of the equation as they're not being touched here.

The baseline for this service is

- 9 passing
- 2 failing

with the failures on

- Bitbucket Pipelines [live] branch build result (valid)
- Bitbucket Pipelines [live] branch build result (never built)

because the upstream repo we were testing against has been deleted. To fix those, we need to find or make another repo for those tests, or switch them to mocked tests.

so I think for this PR if we can keep everything else passing that is OK for review.

PyvesB:
@reeceappling I spent some time fixing all Bitbucket tests in #9881, it should help you out if you were to revive this PR. :)

PyvesB:

If others think that repository level projects would be an option and want to pick this up, you can simply fetch the commits by running git fetch origin pull/9472/head:pr-9472. Change the implementation to target repository level projects and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

PR #9499, opened 2023-08-21
migrate examples to openApi part 13; affects [curseforge date fedora hsts modrinth ore]

chris48s:
Refs #9285

I'm not going to repeat the same blurb. In this one I am still tackling easier cases (all path params) but I'm now moving on to the ones with a description/documentation string of some kind.

### DIFF
--- a/services/date/date.service.js
+++ b/services/date/date.service.js
@@ -1,7 +1,7 @@
 import { formatRelativeDate } from '../text-formatters.js'
-import { BaseService } from '../index.js'
+import { BaseService, pathParams } from '../index.js'
 
-const documentation = `
+const description = `
 <p>
   Supply a unix timestamp in seconds to display the relative time from/to now
 </p>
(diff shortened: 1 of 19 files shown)


services/date/date.service.js
```diff
@@ -11,16 +11,18 @@ export default class Date extends BaseService {
...
-  static examples = [
-    {
-      title: 'Relative date',
-      pattern: ':timestamp',
-      namedParams: { timestamp: '1540814400' },
-      staticPreview: this.render({ relativeDateString: '2 days ago' }),
-      keywords: ['time', 'countdown', 'countup', 'moment'],
-      documentation,
+  static openApi = {
+    '/date/{timestamp}': {
```
calebcartwright, line 15: I'm just now realizing (or perhaps have just forgotten) that I'm not sure if/how we're doing any parameter validation on the frontend badge builder in the new ui :thinking: 

More of a reminder/edification question, definitely non-blocking

calebcartwright (approved):

services/date/date.service.js
chris48s: I think this is the relevant bit of validation:

https://github.com/badges/shields/blob/9118ba8d24d8abc0de723a3e8a5cb36020248fbc/core/base-service/base.js#L214-L229

Added in #9320 in response to https://github.com/badges/shields/pull/9320#discussion_r1249935938

PR #9500, opened 2023-08-21
migrate examples to openApi part 12; affects [GitHub Gist]

chris48s:
Refs #9285

I'm not going to repeat the same blurb. In this one I am still tackling easier cases (all path params) but I'm now moving on to the ones with a description/documentation string of some kind.

### DIFF
--- a/services/github/gist/github-gist-last-commit.service.js
+++ b/services/github/gist/github-gist-last-commit.service.js
@@ -16,7 +16,7 @@ export default class GistLastCommit extends GithubAuthV3Service {
     '/github/gist/last-commit/{gistId}': {
       get: {
         summary: 'GitHub Gist last commit',
-        description: documentation,
+        description: `Shows the latest commit to a GitHub Gist.\n${documentation}`,
         parameters: pathParams({
           name: 'gistId',
           example: '8710649',
--- a/services/github/gist/github-gist-last-commit.service.js
+++ b/services/github/gist/github-gist-last-commit.service.js
@@ -1,4 +1,5 @@
 import Joi from 'joi'
+import { pathParams } from '../../index.js'
 import { formatDate } from '../../text-formatters.js'
 import { age as ageColor } from '../../color-formatters.js'
 import { GithubAuthV3Service } from '../github-auth-service.js'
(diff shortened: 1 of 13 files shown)


services/github/gist/github-gist-stars.service.js
```diff
@@ -32,18 +32,18 @@ export default class GistStars extends GithubAuthV4Service {
...
   }
 
-  static examples = [
-    {
-      title: 'Github Gist stars',
-      namedParams: { gistId: '47a4d00457a92aa426dbd48a18776322' },
-      staticPreview: {
-        label: this.defaultBadgeData.label,
-        message: metric(29),
-        style: 'social',
```
PyvesB, line 42: Is leaving the style out intentional?

services/github/gist/github-gist-last-commit.service.js
```diff
@@ -11,17 +12,18 @@ const schema = Joi.object({
...
+      get: {
+        summary: 'GitHub Gist last commit',
+        description: documentation,
+        parameters: pathParams({
+          name: 'gistId',
+          example: '8710649',
+        }),
       },
-      staticPreview: this.render({ commitDate: '2022-07-29T20:01:41Z' }),
-      keywords: ['latest'],
```
PyvesB, line 21: I may have missed something, we're dropping keywords in this example and a few others. Are these not supported anymore?

chris48s: Good question.

With our old search, it could only search some limited fields.

The OpenAPI format doesn't have "keywords", but the search plugin we use now does give us full text search over all text on the page (so that includes the summary, description and all parameter fields, including param names, descriptions, etc). So yes we are losing `keywords` but we have gained more powerful search. For example https://shields.io/search?q=vcsType shows me all badges that take a param called `vcsType`, even though none of them have it in their title or description. We didn't have this before.

I think in the grand scheme of things, the best way to deal with this is to try and make sure that if we're removing keywords that would help users search for things and they're not covered by other text, we make sure we're adding a description that includes them. This is something I've overlooked in some cases. For example, https://shields.io/search?q=ruby doesn't return results for any of our ruby gems badges because they're all just called "gem [something]". I think in those cases, we can usefully add a description that also includes the word "ruby" to help with that.

I'm happy to take an action to review any we've already merged or got open for this.

PyvesB (dismissed):

chris48s:
There are a couple of failing github tests on the most recent run, but they're unrelated to the changes on this branch

PR #9502, opened 2023-08-23
Use defaultBadgeData as object instead of function in tutorial docs

### LINKED ISSUE #9501: defaultBadgeData is referred as a function in Tutorial but is an object
I was going through the documentation for creating a new badge and found a mismatch.

Here, `defaultBadgeData` is defined as an object.
https://github.com/badges/shields/blob/531d2acf5cf899ce26be459300845e2fef5c4cab/doc/TUTORIAL.md?plain=1#L190

In the description, `defaultBadgeData()` is mentioned as a callable.
https://github.com/badges/shields/blob/531d2acf5cf899ce26be459300845e2fef5c4cab/doc/TUTORIAL.md?plain=1#L226

**Expected Behaviour :**
`defaultBadgeData()` should be `defaultBadgeData` in the description.

pr7prashant:
Closes #9501 

**Solution Description :** 
Update the [Tutorial documentation](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) to use `defaultBadgeData` as an object instead of function.


chris48s (approved):
thanks

PR #9505, opened 2023-08-24
log an error to sentry if an upstream service responds with `429 Too Many Requests`

chris48s:
I would quite like to get a notification in Sentry when an upstream service gives us a 429. Seems like hitting rate limits is something we would like to know about proactively rather than waiting for users to hit errors and let us know. This might turn out to be too noisy, but lets give it a try.

### DIFF
--- a/core/base-service/check-error-response.js
+++ b/core/base-service/check-error-response.js
@@ -1,3 +1,4 @@
+import log from '../server/log.js'
 import { NotFound, InvalidResponse, Inaccessible } from './errors.js'
 
 const defaultErrorMessages = {
@@ -25,6 +26,13 @@ export default function checkErrorResponse(httpErrors = {}) {
         error = new InvalidResponse(props)
       }
     }
+
+    if (res.statusCode === 429) {
+      log.error(
+        new Error(`429 Too Many Requests calling ${res.requestUrl.origin}`),
+      )
+    }
+
     if (error) {
       error.response = res
       error.buffer = buffer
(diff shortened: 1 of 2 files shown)


core/base-service/check-error-response.js
```diff
@@ -25,6 +26,13 @@ export default function checkErrorResponse(httpErrors = {}) {
         error = new InvalidResponse(props)
       }
     }
+
+    if (res.statusCode === 429) {
+      log.error(
+        new Error(`429 Too Many Requests calling ${res.requestUrl.origin}`),
```
chris48s, line 32: Using `res.requestUrl.origin` instead of `res.requestUrl.toString()` here will only log the domain part of the request. This is useful for two reasons:

1. It will mean Sentry considers every "429 Too Many Requests calling https://example.com" to be the same issue for de-duplication purposes - we'll log them all as many events attached to one issue. If we used the whole URL then "429 Too Many Requests calling https://example.com/foo" and "429 Too Many Requests calling https://example.com/bar" would be considered different issues, leading to a lot of noise.
2. We won't send any URL parameters (which could include an API key or token for services that use query string auth) to Sentry

Just the domain should be enough information to tell us which upstream service is applying a rate limit though.

PyvesB (approved):
Agreed, this sounds like a useful thing to add. 👍🏻

chris48s:
Follow up: https://github.com/badges/shields/pull/9523

PR #9506, opened 2023-08-24
Replace localhost with domain in example link.

tomheaton:
Replace localhost with domain in example link.


chris48s (approved):
good catch - thanks

Issue #5343, opened 2020-07-18
move osslifecycle and nycrc into /gtihub


chris48s:
As noted in https://github.com/badges/shields/pull/5234 these two anomolies should be moved into the `/github` dir and URL namespace, like all the other "get a file from a github repo and extract a vlue" badges. We should set up redirects for compatibility.

calebcartwright:
I know those two badges only support looking for their respective files in github today, but is there any reason that has to be the case? For example I could just easily use Bitbucket and nyc and want the badge. I know we haven't had any asks for those just yet, but perhaps worth considering.

chris48s:
Hmm.. that's broadly also true of all the other "get a file from a github repo and extract a value" badges. i.e:

- github-go-mod
- github-all-contributors
- github-lerna-json
- github-pipenv
- github-manifest
- github-package-json

Whichever way we look at it, we should treat osslifecycle and nycrc the same as all of those, but maybe there is a case that none of those should be github-specific

calebcartwright:
Excellent point!

Maybe in a perfect world we'd have a "get a file from (VCS provider) and extract a value" framework/reusable functions or base classes.

calebcartwright:
And just to clarify, I don't have any objections to moving these services under the github directory, regardless of what we may or may not do down the line. At this point it's really more about easily knowing where all the service classes are that use the GitHub APIs to do their thing.

chris48s:
Another relevant consideration here
https://github.com/badges/shields/pull/9516
adds a badge that just takes an arbitrary file (not hard-coded to GitHub). One of the issues here is: How these fit into the URL schema?

PR #9516, opened 2023-08-25
Add [PythonVersionFromToml] shield

### LINKED ISSUE #9410: Python versions from pyproject.toml

jNullj:
Added new shield PythonVersionFromToml
The shield display required python versions for packages based on pyproject.toml
Example:
For URL `https://raw.githubusercontent.com/numpy/numpy/main/pyproject.toml`
Shield URL is: `http://localhost:8080/python/required-version-toml?tomlFilePath=https%3A%2F%2Fraw.githubusercontent.com%2Fnumpy%2Fnumpy%2Fmain%2Fpyproject.toml`
And the result: 
image

Tester is also added in this PR.

Fixes #9410


services/python/python-version-from-toml.service.js
```diff
@@ -0,0 +1,65 @@
...
+
+  static route = {
+    base: '',
+    pattern: 'python/required-version-toml',
+    queryParamSchema,
+  }
+
+  static examples = [
+    {
+      title: 'Python Required Version TOML',
```
chris48s, line 33: Can we clarify in the title and description that this is for PEP 621-compliant `pyproject.toml`

Python packaging is..fun :grimacing:  and lots of projects have a pyproject.toml that doesn't use this format e.g: https://github.com/chris48s/arcgis2geojson/blob/master/pyproject.toml#L12

services/python/python-version-from-toml.tester.js
```diff
@@ -0,0 +1,26 @@
...
+export const t = await createServiceTester()
+
+const isCommaSeparatedPythonVersions = Joi.string().regex(
+  // This should test for PEP440
+  // Accepted values are one or more Version specifiers as defined at https://peps.python.org/pep-0440/#version-specifiers
+  // Some strings might include spaces, those are valid, values are comma seperated
+  // Versions should fit the version scheme https://peps.python.org/pep-0440/#version-scheme
+  // This is based on the example in PEP440 at https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions
+  /^\s*(?:(?:===|!=|<=|>=|<|>)\s*)?((?:(?:\d+!)?(?:\d+(?:\.\d+)*))(?:(?:[abc]|rc)\d+)?(?:\.post\d+)?(?:\.dev\d+)?)(?:\s*,\s*(?:(?:===|!=|<=|>=|<|>)\s*)?((?:(?:\d+!)?(?:\d+(?:\.\d+)*))(?:(?:[abc]|rc)\d+)?(?:\.post\d+)?(?:\.dev\d+)?))*\s*$/,
+)
```
chris48s, line 12: Rather than copy/pasing this monster regex (I hope you didn't write this from scratch! :smile: ), I'd suggest we use `Joi.custom()` https://joi.dev/api/#anycustommethod-description here to make a validator using the `validRange()` function from `@renovate/pep440` (which is a package we already use).

```js
import pep440 from '@renovate/pep440'
pep440.validRange(">=3.6, <4.0") // true
```

services/python/python-version-from-toml.service.js
jNullj: Fixed with https://github.com/badges/shields/pull/9516/commits/e2ecbf5ff2a83f419307f752c453460be345b027

services/python/python-version-from-toml.service.js
```diff
@@ -0,0 +1,65 @@
...
+And not https://github.com/numpy/numpy/blob/main/pyproject.toml
+`
+
+class PythonVersionFromToml extends BaseTomlService {
+  static category = 'platform-support'
+
+  static route = {
+    base: '',
+    pattern: 'python/required-version-toml',
+    queryParamSchema,
```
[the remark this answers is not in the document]
jNullj: I was not sure myself, i used SecurityHeaders as reference and saw it used an empty base route.
What should we do with that?

services/python/python-version-from-toml.tester.js
jNullj: Should be fixed with https://github.com/badges/shields/pull/9516/commits/b3de30f1859274749dc0e513be77ecd2b63ddfcd
I still have issue testing so i feel like mountain climbing where i try to get it first try, lets see the  CI results.

services/python/python-version-from-toml.service.js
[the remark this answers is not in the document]
chris48s: So it isn't really about the empty base. It is more that our URL schema assumes the first part of the route is a service, but there's no service here.
Given nobody else has chimed in on this, I think I'm going to merge this but note this as a consideration in https://github.com/badges/shields/issues/5343

chris48s (approved):

PR #9517, opened 2023-08-25
Add dynamic TOML support via [DynamicToml] Service

cptpiepmatz:
This adds a dynamic service for TOML files.

I kept everything as close to the other dynamic services as possible. To be able to use the `BaseTomlService` from #9438, I exported in the core where the other base services are exported.
For the tests and default values I used the [`toml-spec-example.toml`](https://github.com/squirrelchat/smol-toml/blob/mistress/bench/testfiles/toml-spec-example.toml) from [`smol-toml`](https://github.com/squirrelchat/smol-toml) because `smol-toml` is used in here and it is the same example TOML file that [toml.io](https://toml.io) shows on their homepage, sadly they do not host this toml themselves.

I mostly ported over the test from the dynamic YAML service.

### DIFF
--- a/services/dynamic/dynamic-toml.tester.js
+++ b/services/dynamic/dynamic-toml.tester.js
@@ -39,11 +39,11 @@ t.create('TOML from url | caching with new query params')
   .get(
     '.json?url=https://raw.githubusercontent.com/squirrelchat/smol-toml/mistress/bench/testfiles/toml-spec-example.toml&query=$.owner.name',
   )
-  .expectBadge({ label: 'Tom Preston-Werner', message: '0.8.0' })
+  .expectBadge({ label: 'custom badge', message: 'Tom Preston-Werner' })
 
 t.create('TOML from url | with prefix & suffix & label')
   .get(
-    '.json?url=https://raw.githubusercontent.com/squirrelchat/smol-toml/mistress/bench/testfiles/toml-spec-example.toml&query=$.database.temp_targets.cpu&prefix=+&suffix=°C&label=CPU Temp Target',
+    '.json?url=https://raw.githubusercontent.com/squirrelchat/smol-toml/mistress/bench/testfiles/toml-spec-example.toml&query=$.database.temp_targets.cpu&prefix=%2B&suffix=°C&label=CPU Temp Target',
   )
   .expectBadge({ label: 'CPU Temp Target', message: '+79.5°C' })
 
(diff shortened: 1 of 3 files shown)


chris48s (approved):
LGTM. Nice first contribution

services/dynamic/dynamic-toml.tester.js
```diff
@@ -0,0 +1,107 @@
...
+  .get('.json?query=$.name&label=Package Name')
+  .expectBadge({
+    label: 'Package Name',
+    message: 'invalid query parameter: url',
+    color: 'red',
+  })
+
+t.create('No query specified')
+  .get(
+    '.json?url=https://raw.githubusercontent.com/squirrelchat/smol-toml/mistress/bench/testfiles/toml-spec-example.toml&label=Package Name',
```
Mulugruntz, line 14: I'm not sure I understand...
All these tests use the json endpoint?

Mulugruntz (commented):
Nice feature, I'm looking forward to using it!

(do we know when it'll be available on shields.io?)

services/dynamic/dynamic-toml.tester.js
cptpiepmatz: The yml and xml do it too, I don't really know why

chris48s: Just to explain what is going on here:

You can call any shields badge URL with a `.json` extension to request a JSON response instead of SVG. So for example https://img.shields.io/github/v/tag/badges/shields (SVG badge) https://img.shields.io/github/v/tag/badges/shields.json (same thing as JSON)

In the service tests we use this because it is easier to make assertions about the JSON response than the SVG response. The logic for rendering that object into a SVG is tested separately.

`createServiceTester()` makes a test object with a base of `/badge/dynamic/toml` so every url we `.get()` is relative to `/badge/dynamic/toml`

The dynamic badges don't take any positional parameters - only query string parameters.

So when we call 

```
.get(
  '.json?url=https://raw.githubusercontent.com/squirrelchat/smol-toml/mistress/bench/testfiles/toml-spec-example.toml&label=Package Name',
)
```

we're really requesting

`/badge/dynamic/toml.json?url=https%3A//raw.githubusercontent.com/squirrelchat/smol-toml/mistress/bench/testfiles/toml-spec-example.toml&label=Package%20Name`

which returns a response like

https://img.shields.io/badge/dynamic/toml.json?url=https%3A//raw.githubusercontent.com/squirrelchat/smol-toml/mistress/bench/testfiles/toml-spec-example.toml&label=Package%20Name

..and then we're making assertions about it.

cptpiepmatz: Thanks!

PR #9520, opened 2023-08-29
(Minor) Fix broken URLs

DJStompZone:
Changed to full url to fix broken hyperlinks

### DIFF
--- a/frontend/docs/static-badges.md
+++ b/frontend/docs/static-badges.md
@@ -9,5 +9,5 @@ It is possible to use shields.io to make a wide variety of badges displaying sta
 
 For more info, see:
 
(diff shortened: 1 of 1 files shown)


chris48s:
Can you explain this?

The links on https://shields.io/docs/static-badges work for me. If they don't work for you, can you provide more info?

In any case, these links should be relative. i.e:

- If I spin up a local dev copy, "Logos" should link to http://localhost:3000/docs/logos not https://shields.io/docs/logos
- If someone self hosts their own instance, "Logos" should link to https://theirdomain.com/docs/logos not https://shields.io/docs/logos
- etc

calebcartwright:
Going to close as this seems to have been resolved with a "no action required" disposition

PR #9523, opened 2023-08-29
Stop logging 429s to sentry (for now)

chris48s:
Well.. that was exciting, wasn't it? :sweat_smile: 

Doing this has been useful. Just having this turned on for a couple of hours has revealed that there are:

1. Problems we need to solve
2. Categories of noise we need to silence in order for this to be useful

In its current form, I want to turn this off for now, otherwise we're just going to smash our sentry allowance for the month on ingesting events which are mostly duplicating things we now already know.

This will probably come back, but I'm going to go and do some work on the above points before we re-enable it it.


calebcartwright (approved):
agreed on all counts :+1:

Issue #9515, opened 2023-08-25
The Testspace Test Results Badges is inaccessible


henry-hub:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

The Testspace Test Results Badges is inaccessible

### 🔗 Link to the badge

https://img.shields.io/testspace/tests/swellaby/swellaby/main
![Testspace tests with custom labels](https://img.shields.io/testspace/tests/swellaby/swellaby/main)

jNullj:
Hello @henry-hub and thank you for using shields and taking the time to open an issue.

While looking at the shield link you provided I noticed you might be missing the `metric` path parameter, as this is a required parameter this might be the cause of your issue.

Please take a look at the docs for this [shield here](https://shields.io/badges/testspace-tests)
Notice how the url format is `https://img.shields.io/testspace/:metric/:org/:project/:space` where metric should be one of the following: `[total, passed, failed, skipped, errored, untested]`

The log posted by the github actions bot also shows that:
```
 Response status code  🎯 
 404
 Handled error  🙅 
 NotFound: Not Found: org, project, or space not found
```

I also think that the project you specified, swellaby does not exist in the swellaby space. Please pick a project that exists at https://swellaby.testspace.com/ like: `swellaby:testspace-sample`, `swellaby:rusty-hook`, `swellaby:nyc-config` or `swellaby:dotfiles`

I will note that while testing this shield with the default example it worked for me in my local build but did not work on shields.io, So this might be related to an issue with the main deployment or the connection from it to testspace.com.
@chris48s Who can debug the shields.io deployment?

henry-hub:
> Hello @henry-hub and thank you for using shields and taking the time to open an issue.
> 
> I think that the project you specified, swellaby does not exist in the swellaby space. Please pick a project that exists at https://swellaby.testspace.com/ like: , , or `swellaby:testspace-sample``swellaby:rusty-hook``swellaby:nyc-config``swellaby:dotfiles`
> 
> The log posted by the github actions bot also shows that:
> 
> ```
>  Response status code  🎯 
>  404
>  Handled error  🙅 
>  NotFound: Not Found: org, project, or space not found
> ```
> 
> I will note that while testing this shield with the default example it worked for me in my local build but did not work on shields.io, So this might be related to an issue with the main deployment or the connection from it to testspace.com. @chris48s Who can debug the shields.io deployment? Also the github workflow gets the correct response while the shields.io deployment gets inaccessible which does not seems to be right.

[...]

PR #9525, opened 2023-08-29
Fix [testspace] badges

chris48s:
Closes https://github.com/badges/shields/issues/9515


PyvesB (approved):
Great investigation in the linked issue!

Issue #9524, opened 2023-08-29
Triage 429s in Sentry


chris48s:
:clipboard: **Description**

There's a bit of a mixed bag of stuff that came out of doing https://github.com/badges/shields/pull/9505

https://shields.sentry.io/issues/?project=251614&query=is%3Aunresolved+429&referrer=issue-list&statsPeriod=90d

chris48s:
OK.. quick notes on this.

We should implement a way to disable logging 429s to Sentry at the service class level and then use it to disable them from:

- endpoint
- dynamic json/yaml/toml/xml

There's not really anything we can do when we don't control the service we're calling

- uptimerobot (more details https://github.com/badges/shields/issues/9155 ). There's not really anything we can do about this. It is not our key that is getting rate limited
- weblate (if not calling default server). We aren't getting 429s from https://hosted.weblate.org - they are coming from people's self-hosted instances

In terms of the services where we actually need to fix stuff:

- pypistats - Independently of this I submitted https://github.com/badges/shields/pull/9522 which should solve it. Probably no further action needed
- Discord - Lets up the cache on this, but I'd also like to do a bit more digging into what the rate limits on our key are
- OpenCollective - Up the cache. I'm also going to write an issue and see if we can ask for an increased rate limit
- GitHub - This isn't a token pool issue. All the errors are coming from `GithubActionsWorkflowStatus`, which doesn't use the API, or a key. I think the only thing we can do is increase the max-age really.

There are probably a few other cases, but that will deal with the notifications that are happening in bulk. The others are low enough volume I'm happy to calm the notifications down a bit and then pick at them once we re-enable the sentry events. Some of them are known issues like https://github.com/badges/shields/issues/9256

I'll try and have a go at this over the weekend.

PR #9546, opened 2023-09-04
log 429s to sentry (attempt 2); affects [dynamic endpoint uptimerobot weblate opencollective discord github]

chris48s:
Refs https://github.com/badges/shields/pull/9505
Refs https://github.com/badges/shields/pull/9523
Closes https://github.com/badges/shields/issues/9524

### DIFF
--- a/services/opencollective/opencollective-all.service.js
+++ b/services/opencollective/opencollective-all.service.js
@@ -16,7 +16,7 @@ export default class OpencollectiveAll extends OpencollectiveBase {
     },
   }
 
-  static _cacheLength = 900
+  static _cacheLength = 1800
 
   static defaultBadgeData = {
     label: 'backers and sponsors',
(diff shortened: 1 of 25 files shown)


services/uptimerobot/uptimerobot-base.js
```diff
@@ -74,6 +74,7 @@ export default class UptimeRobotBase extends BaseJsonService {
           ...opts,
         },
       },
+      logErrors: [],
```
chris48s, line 77: See https://github.com/badges/shields/issues/9155 for details.

The rate limits we're hitting here are on the user-supplied tokens. There's not really anything we can do about this.

services/opencollective/opencollective-all.service.js
```diff
@@ -16,7 +16,7 @@ export default class OpencollectiveAll extends OpencollectiveBase {
     },
   }
 
-  static _cacheLength = 900
+  static _cacheLength = 1800
```
chris48s, line 19: Lets see if this gets us consistently under the limits. If not, I'll raise an issue with OC and see if they'd be willing to grant us a higher limit.

PyvesB (approved):
Changes look reasonable to me. Best to filter out errors we can't act on, otherwise it's just noise. Hopefully cache increase on the others will do the job.

Issue #7569, opened 2022-02-05
Lines of code is showing `0` for several repositories that used to count lines OK


sgbaird:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

See https://github.com/sparks-baird/mat_discover and https://github.com/sgbaird-5DOF/interp for example. If it's not showing "lines 0" when you open this, then I'll assume this was fixed. Not sure if the issue is on the shields.io side or the `tokei` side.

### 🔗 Link to the badge

https://img.shields.io/tokei/lines/github/sparks-baird/mat_discover
![](https://img.shields.io/tokei/lines/github/sparks-baird/mat_discover)

PyvesB:
Hello @sgbaird ! :wave:

Thanks for reporting this! Seems like our [Tokei tests have started failing](https://app.circleci.com/pipelines/github/badges/daily-tests/1497/workflows/79b757da-4b9d-4553-ae04-3197a6428d15/jobs/2399/tests#failed-test-11) as well.

It does seem like a problem on the Tokei side, hitting the API for the [sparks-baird/mat_discover](https://img.shields.io/tokei/lines/github/sparks-baird/mat_discover), I get the following response:
````
{"blanks":0,"code":0,"comments":0,"lines":0,"stats":[],"inaccurate":false}
````

Other users have also reported this problematic behaviour here: https://github.com/XAMPPRocky/tokei/issues/881

calebcartwright:
Things appear to have been sorted upstream so going to close. If there's any recurrence in the near term please feel free to post here and we can reopen.

If anyone significantly in the future sees similar symptoms, it's probably better to open a new issue

calebcartwright:
Reopening as the upstream issue has recurred

librity:
I'm having the same problem. Examples:
- https://github.com/librity/ft_pipex
- https://github.com/librity/ft_libft
- https://github.com/librity/ft_get_next_line
- https://github.com/librity/ft_fractol

igorskyflyer:
+1, everything was working fine until like 2-3 days ago 👀

calebcartwright:
Folks, please refrain from posting the "me too" types of comments as they just add unhelpful noise. The above thread already contains relevant information explaining that the issue upstream.

[...]

PR #9560, opened 2023-09-10
migrate examples to openApi part 14; affects [azuredevops discord netlify opencollective pub ros visualstudio]

chris48s:

The tokei tests are known to be failing https://github.com/badges/shields/issues/7569 so I've excluded them form the list of service tests to run. For now, lets just abstract that and refactor it anyway


PyvesB (approved):
Looks good to me 👍🏻

PR #9567, opened 2023-09-11
fix [bower] version badge

chris48s:
Our bower badges use the libraries.io API. Also bower is a low traffic badge/mostly abandoned service.

The version badge (which should be latest stable version) consistently returns "no releases" for pretty much any package.

This is because if you look at the API response for a bunch of packages e.g:

- https://libraries.io/api/bower/bootstrap
- https://libraries.io/api/bower/angular
- https://libraries.io/api/bower/backbone
- https://libraries.io/api/bower/Font-Awesome
- https://libraries.io/api/bower/grunt
- https://libraries.io/api/bower/gulp
- https://libraries.io/api/bower/jQuery

they all return `latest_stable_release_number: null`, although it is worth noting that all of them do have a proper version number against `latest_release_number`, so the `include_prereleases` variant works in all those cases.

Given the `latest_release_number` number doesn't seem to be populated for any package (or at least I can't find an example where it works), I suggest we just switch this so that the version badge returns `latest_release_number` whether you call it with `include_prereleases` or not.

### DIFF
--- a/services/bower/bower-version.service.js
+++ b/services/bower/bower-version.service.js
@@ -1,15 +1,10 @@
-import Joi from 'joi'
 import { renderVersionBadge } from '../version.js'
 import { InvalidResponse } from '../index.js'
 import BaseBowerService from './bower-base.js'
 
-const queryParamSchema = Joi.object({
-  include_prereleases: Joi.equal(''),
-}).required()
-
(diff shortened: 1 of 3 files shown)


calebcartwright (approved):
sounds reasonable to me

Issue #9569, opened 2023-09-14
Lines of code just reports invalid data in the badge


ascopes:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Seems the lines of code badge just gives an error. The example suggests GitHub is valid.

### 🔗 Link to the badge

https://img.shields.io/tokei/lines/github/ascopes/java-compiler-testing

![](https://img.shields.io/tokei/lines/github/ascopes/java-compiler-testing)

![Screenshot_2023-09-14-07-49-58-84_40deb401b9ffe8e1df2f1cc5ba480b12](https://github.com/badges/shields/assets/73482956/57b11e3d-1199-4e7c-bccc-c142794cdef1)

calebcartwright:
Closing as a duplicate of #9389 and #7569

Those issues have links to the upstream issues in the Tokei service (which is where the underlying issue resides and needs to be resolved)

Issue #9389, opened 2023-07-16
"invalid response data" in total lines


dishb:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

I'm using the [`total lines`](https://shields.io/badges/lines-of-code) badge, and I'm getting an "invalid response data" error. Here is a screenshot:
![Screen Shot 2023-07-15 at 7 20 57 PM](https://github.com/badges/shields/assets/130259005/0a67336c-c3e6-4af0-9cb6-0d59e58274ff)

I've gotten this error before, so it's not like I had a bad day today 😅

As for the repository I'm trying to access, it can be viewed at [dishb/hive](https://github.com/dishb/hive). So the repository does indeed exist.

### 🔗 Link to the badge

Here is the badge in markdown:
![Lines of code](https://img.shields.io/tokei/lines/github/dishb/hive)

The code:
`![Lines of code](https://img.shields.io/tokei/lines/github/dishb/hive)`

The URL:
https://img.shields.io/tokei/lines/github/dishb/hive

chris48s:
The reason this is failing is because the response from tokei does not have a `lines` property. The detailed error is `ValidationError: "lines" is required`

I am also seeing this on the responses for other repos. I'm going to raise an issue for this one upstream.

dishb:
Okay, thank you.

chris48s:
follow https://github.com/XAMPPRocky/tokei/issues/1017 for updates

dishb:
Just checked the [API docs](https://docs.rs/tokei/latest/tokei/struct.CodeStats.html#structfield.code), maybe the correct way to access the lines of code from the response is through `code`?

ashuthe1:
Hi! I am new to this community and would love to work on this.

dishb:
@ashuthe1 - You'd probably need to ask to be assigned to [this issue.](https://github.com/XAMPPRocky/tokei_rs/issues/233). If you noticed, the `needs-upstream-help` means this isn't necessarily an issue with shields but with one of its service providers.

calebcartwright:
Yeah appreciate the offer @ashuthe1 but there's nothing we can do here and are blocked on the upstream issue. 

If you're just interested in getting involved with Shields more generally there's a number of other issues you could potentially look at, and I'd suggest starting with these: https://github.com/badges/shields/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

ashuthe1:
[...]

Issue #8796, opened 2023-01-12
total lines (tokei) don't work


CloudlyBot:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

The total lines bage don't show the number of lines anymore.

### 🔗 Link to the badge

https://img.shields.io/tokei/lines/github/Argantiu/minecraft-manager

### 💡 Possible Solution

It would be cool if this will be fixed. :)

calebcartwright:
Closing as a duplicate of #7569

Issue #7889, opened 2022-04-29
Total Lines of Code gives HTTP 408 periodically


ascopes:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

The lines of code badge has been giving a 408 intermittently for the past few days.

When this occurs, the badge does not render and shows a "missing image" icon from the browser, rather than providing an error badge.

### 🔗 Link to the badge

https://img.shields.io/tokei/lines/github/ascopes/java-compiler-testing

calebcartwright:
The Shields.io service provides badges which are based on data we fetch from some external source (we don't provide CI, package registry, code counting, etc. capabilities ourselves). As such we're directly impacted by upstream issues and latency in those external sources and there's not really anything we can do about it.

There's been recurrent issues in the Tokei service for the last several months, and I don't think there's really anything new to discuss nor investigate here. Refs: https://github.com/badges/shields/issues/7569 https://github.com/XAMPPRocky/tokei/issues/881

calebcartwright:
No, I do not think we should categorically issue a 503 nor 500 response status code to our users when we're stuck waiting on the tokei service to respond to us so that we can then actually build and return the badge to our client.

calebcartwright:
I'm not sure what your objective is here anymore, so let's take a step back.

Was your goal to try to get an understanding of the issues with the Tokei badge, or do you want to have a more holistic discussion about specific response codes orthogonal to/independent of the tokei badge essentially being broken due upstream issues?

In general we're receptive to the latter types of discussions, even if they trend towards pedantry and away from pragmatism. I'm open to considering alternative behavior, but it's also not clear to me what the practical benefits would be; these badges aren't going to render.

I'll also note it would be particularly helpful if you could capture the behavior you're describing at the point you're seeing it, as it's not reproducible currently. Additionally, worth keeping in mind that with the 408 code in particular there's plenty of literature that notes it can be sent by the server in cases beyond a client/request specific issue, despite that being one of the typical differentiators between 4xx and 5xx codes.

ascopes:
Right, I follow now. I was getting confused as to whether this was something being triggered by the request I was sending that I needed to change (going by that status code), or whether it was outside my control and just a general issue.

I don't have any concrete way of reproducing this as it seems to occur randomly. What would you suggest would be a good course of action? Is it best to stop using that badge?

paulmelnikow:
[...]

PR #9581, opened 2023-09-17
deprecate [tokei] service

### LINKED ISSUE #7569: Lines of code is showing `0` for several repositories that used to count lines OK

chris48s:
This might be a bit controversial, but..

This service has been out of action for some time. We've tried raising issues upstream with no response. Even when it has been "working", this has been an unreliable service, frequently producing unexpected results for users or not returning a response within a timeframe that allows us to render a useful response. This is unsurprising given it is essentially fetching an upstream repository (which might be very large) and counting the lines of code on-demand. If I pass https://github.com/torvalds/linux as the input, you're just not going to do that in 4 seconds or less, even if you did write the code in rust.

For example:

- https://github.com/badges/shields/issues/6637
- https://github.com/badges/shields/issues/6833
- https://github.com/badges/shields/issues/7889
- https://github.com/badges/shields/issues/8796
- https://github.com/badges/shields/issues/9389
- https://github.com/badges/shields/issues/9569

I think it is time to give up on this one.

Closes #7569


calebcartwright:
Let's hold off on this for a _bit_ as the Tokei maintainer has commented on relevant issues earlier today.

In general I'd be in favor of deprecating this service, but only because:

* There's been service availability issues, the maintainer has acknowledged it's a side/passion project they can't spend much time on, and some of the issues that pop up ultimately stem from cost considerations
* This service doesn't necessarily meet our guidelines wrt documentation (the fact that you can get a json response from their badge url via the content header is an undocumented feature), support/maintainer availability, etc.

However, I have to disagree with the performance characteristics portion of the reasoning. IMO this is another canonical example of the "good badge for us, but can be quite slow to respond on the first call" scenario, which is precisely the reasoning behind #9233.

The tokei.rs web service does have caching enabled, so even if the first call for counting all the lines of the linux kernel is slow, the subsequent calls should return well within the window we'd need to satiate camo's timeout limit.

AFAICT the service is working just fine and has been for a while, it's just that the `lines` field mysteriously disappeared. I've bumped your upstream issue with the maintainer, and if we can get a timely resolution on that (even if it's something to the effect of "that was intentional, just add the `code`, `blanks`, and `comments` fields to get the total moving forward) then I'd be in favor of retaining this badge.

But I'm :+1: with deprecating this if we can't get an answer on that and/or see a recurrence of the availability issues in the future

chris48s:
Yeah - fair shout. If the other issues can be sorted, we could add an `ETIMEDOUT` handler for the performance issue :+1:

Naegolus:
We removed all tokei badges from our repositories because they weren't working most of the time.

PyvesB (approved):
I'd be in favour of removing this badge at this point, it's been broken for months now, latest issue being SSL handshakes with the API.

PR #9584, opened 2023-09-18
migrate examples to openApi part 17; affects [buildkite cii codeship crates jsdelivr npms] and bundlephobia

chris48s:

I have touched fewer files in this PR as I'm starting to get into some more slightly more involved services now.

I'll be honest. 17 PRs deep into this project and knowing how many more there are to do, my enthusiasm for this endeavour is starting to decrease a bit. Staying strong though.

I've skipped the bundlephobia tests on this PR. They seem to be having some kind of outage at the moment affecting some packages but not others e.g:

https://bundlephobia.com/api/size?package=react (works)
https://bundlephobia.com/api/size?package=preact (error)

lets keep an eye on that one..

### DIFF
--- a/services/crates/crates-downloads.service.js
+++ b/services/crates/crates-downloads.service.js
@@ -1,6 +1,6 @@
 import { renderDownloadsBadge } from '../downloads.js'
-import { InvalidParameter, NotFound } from '../index.js'
-import { BaseCratesService, keywords } from './crates-base.js'
+import { InvalidParameter, NotFound, pathParams } from '../index.js'
+import { BaseCratesService } from './crates-base.js'
 
 export default class CratesDownloads extends BaseCratesService {
   static category = 'downloads'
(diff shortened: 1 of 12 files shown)


services/crates/crates-downloads.service.js
```diff
@@ -9,49 +9,54 @@ export default class CratesDownloads extends BaseCratesService {
     pattern: ':variant(d|dv|dr)/:crate/:version?',
```
PyvesB, line 9: For the jsdelivr badge you leveraged a period enum, whereas here you broke down wvery example. Any specific reason why?

chris48s: Yes. I think the variants here need more explanation.

I don't think I would know

d = total downloads
dv = downloads for version
dr = recent downloads

so I think it makes sense to give each one an individual title.

In the jsdelivr example, it was more straightforward to just explain this in the param description: `description: 'Hits per day, week, month or year',`.

PyvesB: Thanks for explaining, makes sense to me.

PyvesB (approved):
Looks good!

PR #9608, opened 2023-09-24
sync docusaurus plugin versions

chris48s:
Yesterday I merged https://github.com/badges/shields/pull/9600

When I was working on shields today, I realised that docusaurus was outputting

```
[ERROR] Invalid docusaurus-plugin-content-docs version 2.4.1.
All official @docusaurus/* packages should have the exact same version as @docusaurus/core (2.4.3).
Maybe you want to check, or regenerate your yarn.lock or package-lock.json file?
[ERROR] Invalid docusaurus-plugin-content-blog version 2.4.1.
All official @docusaurus/* packages should have the exact same version as @docusaurus/core (2.4.3).
Maybe you want to check, or regenerate your yarn.lock or package-lock.json file?
[ERROR] Invalid docusaurus-plugin-content-pages version 2.4.1.
All official @docusaurus/* packages should have the exact same version as @docusaurus/core (2.4.3).
Maybe you want to check, or regenerate your yarn.lock or package-lock.json file?
[ERROR] Invalid docusaurus-plugin-sitemap version 2.4.1.
All official @docusaurus/* packages should have the exact same version as @docusaurus/core (2.4.3).
Maybe you want to check, or regenerate your yarn.lock or package-lock.json file?
[ERROR] Invalid docusaurus-theme-classic version 2.4.1.
All official @docusaurus/* packages should have the exact same version as @docusaurus/core (2.4.3).
Maybe you want to check, or regenerate your yarn.lock or package-lock.json file?
```

This didn't cause anything to fail in CI - it just logged the errors to the console, which is why I didn't notice it.

The place where this is checked is https://github.com/facebook/docusaurus/blob/v2.4.3/packages/docusaurus/src/server/siteMetadata.ts#L75-L92

There doesn't appear to be any command or switch that can make this a hard error but I would really like this mismatch to cause a CI fail so I don't forget this next time. I am going to raise an issue about this upstream, but in the meantime I have cobbled together a crude bash script we can use. It is not very efficient because we have to build the whole frontend, but it is better than a kick in the teeth.
If you follow the commit history, you can see the build fails on the commit where I've added the check and then passes on the commit where I have fixed it.

FWIW, this particular mismatch (2.4.1 / 2.4.3) doesn't appear to actually be causing any issues, but I'd like to get this one reviewed/merged sooner rather than later.

### DIFF
--- a/.github/scripts/check-docusaurus-versions.sh
+++ b/.github/scripts/check-docusaurus-versions.sh
@@ -9,6 +9,7 @@ set -euo pipefail
 
 if ( npm run build 2>&1 | grep '\[ERROR\]' )
 then
+    echo "You probably need to run: npm update @docusaurus/preset-classic"
     exit 1
 else
     exit 0
--- a/.github/scripts/check-docusaurus-versions.sh
+++ b/.github/scripts/check-docusaurus-versions.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Docusaurus outputs some important errors as log messages
+# but doesn't actually fail
+# https://github.com/facebook/docusaurus/blob/v2.4.3/packages/docusaurus/src/server/siteMetadata.ts#L75-L92
+# this script runs `docusaurus build`. If it outputs any [ERROR]s, we exit with 1
+
+if ( npm run build 2>&1 | grep '\[ERROR\]' )
+then
+    exit 1
+else
+    exit 0
+fi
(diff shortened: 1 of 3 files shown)


chris48s:
raised https://github.com/facebook/docusaurus/issues/9342 - we'll see where it goes..

calebcartwright:
Thanks for finding, explaining, and fixing. It's the type of thing that would've had me in a helpless :shrug: had I seen myself

calebcartwright (approved):

PR #9610, opened 2023-09-25
migrate examples to openApi part 19; affects [azuredevops codacy greasyfork jenkinsplugininstalls jetbrains osslifecycle travis visualstudio] and librariesio, uptimerobot

chris48s:

I think this is nearly all the services that use path params only done, except for a few particularly tricky cases.

https://github.com/badges/shields/pull/9565 is still the priority though so I can crack on with the documentation and more services using query params


chris48s:
Disabling a couple of the flakier service tests. Have had them passing locally

PyvesB (approved):
Looks good to me!

PR #9614, opened 2023-09-28
move from @renovate/pep440 to @renovatebot/pep440; affects [pypi piwheels python]

chris48s:
At some point this package moved namespace and we didn't notice.
We're already using the right copy of `@renovatebot/ruby-semver`


chris48s:
This is also a very uncontroversial change. Going to just get this one merged too.

Issue #9611, opened 2023-09-26
[Greasy Fork Badge] Due to API changes, scripts with adult content will display as "not found"


hymbz:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

<https://greasyfork.org/discussions/greasyfork/203565>

### 🔗 Link to the badge

![Greasy Fork](https://img.shields.io/greasyfork/dd/374903)

### 💡 Possible Solution

After receiving a 404 error from Greasyfork, let's try the API of Sleazyfork.

PR #9632, opened 2023-10-02
fix [GreasyFork] 404 bug

hymbz:
refs #9611

### DIFF
--- a/services/greasyfork/greasyfork-base.js
+++ b/services/greasyfork/greasyfork-base.js
@@ -1,6 +1,6 @@
 import Joi from 'joi'
 import { nonNegativeInteger } from '../validators.js'
-import { BaseJsonService } from '../index.js'
+import { BaseJsonService, NotFound } from '../index.js'
 
 const schema = Joi.object({
   daily_installs: nonNegativeInteger,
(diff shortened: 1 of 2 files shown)


chris48s:
Can you explain the relationship between greasyfork and sleazyfork for someone who doesn't use either?

It is clear they are closely related services, but my instinct is that it would make more sense to add sleazyfork as its own service, rather than treating it as a fallback for greasyfork.

hymbz:
You're right, I have made the necessary changes to the code.

services/greasyfork/greasyfork-base.js
```diff
@@ -16,9 +16,17 @@ export default class BaseGreasyForkService extends BaseJsonService {
...
-      schema,
-      url: `https://greasyfork.org/scripts/${scriptId}.json`,
-    })
+    try {
+      return await this._requestJson({
+        schema,
+        url: `https://greasyfork.org/scripts/${scriptId}.json`,
+      })
+    } catch (e) {
+      if (e.message !== 'Not Found') throw e
```
chris48s, line 25: Could we check for the exception type using `instanceof` so it will still work if we customise the not found error.

e.g:

```js
import { NotFound } from '../index.js'

...

if (!(e instanceof NotFound)) throw e
```

hymbz: okay

chris48s (approved):
nice - thanks

Issue #9644, opened 2023-10-08
Unable to generate badge with custom Logo with badge-maker


zavoloklom:
### Are you experiencing an issue with...

badge-maker NPM package

### 🐞 Description

I encountered a bug in the badge-maker library where I am unable to generate a custom badge with a custom logo. It seems that the function responsible for badge generation does not accept the 'logo' key and related parameters. 
Maybe it is related to #4947, but i do not see any updates there.

### 💡 Possible Solution

- Add keys 'logo', 'logoPosition', 'logoWidth', 'links' as expectedKeys in https://github.com/badges/shields/blob/075ead0f9b1369b24f6dbbf62356aa94e6e8b0dd/badge-maker/lib/index.js#L41 
- Add validation
- Add tests

zavoloklom:
I create PR with this changes https://github.com/badges/shields/pull/9645

PyvesB:
badge-maker now has `logoBase64` and `links` parameters. `logoWidth` and `logoPosition` were removed. More context here: https://github.com/badges/shields/pull/9645#issuecomment-2183307545. I don't think there's anything left to do here, will go ahead and close the issue. 😉

Issue #9476, opened 2023-08-15
Logo Strategy


calebcartwright:
I feel it's been a while since we've talked about Shields logos (the handful of ones we maintain and provide directly from https://github.com/badges/shields/tree/master/logo) and that it'd be helpful if we can articulate our current posture and ensure our [logo documentation](https://shields.io/docs/logos) accurately reflects that posture.

We've had various discussions and considerations over the last few years (e.g. #7684), including whether we should drop them outright. We did semi-recently decide to keep the ones that we already have, or at least we didn't decide to delete them, but I'm unclear whether we took a position on whether we could/would potentially ever add any more (and obviously opinions can and do change over time).

As such I'd like to use this to discuss if we collectively feel there are circumstances in which we'd accept a new logo (and if so capture that in our docs), or even if we'd like to drop the logos altogether.

Framing as options:

1) Drop the existing set of logos we maintain, and do not take on any new ones, only provide logos via SI and the custom logo param (as detailed below in https://github.com/badges/shields/issues/9476#issuecomment-1681256345)
2) Keep the existing set but refuse to take on any new ones under any circumstances
3) Keep the existing set and potentially take on new ones after determining/updating guidelines under which we would

I confess I'm ambivalent on the topic myself. I definitely think that SimpleIcons should remain the first port of call, as I'm cognizant of the drawbacks we've experienced in the maintenance of the current set. I feel like if we are to consider accepting new logos that we'd need a set of rather high thresholds a proposed logo would have to clear, and I think that we would also still need to have the discretion to choose on a case by case basis (i.e. our docs would enumerate the key requirements, but we'd still retain the right to reject a proposed logo based on other considerations regardless of whether it cleared all the documented requirements).

That being said, assuming we were to keep logos, I'd find a hypothetical scenario with the following characteristics to be rather compelling:

* a vendor/brand owner (or someone operating explicitly on their behalf) came to us with a PR to incorporate their logo because the monochrome SI route wasn't viable because doing so would ruin/break the brand identity
* vendor/brand representative explicitly agreed they'd follow up if/when they had a branding change that needed to be reflected in the logo (as opposed to giving us flak for the "shields" logo being outdated)
* there was a sizeable portion of our user base that wanted to use the logo
* the logo was too big to be used via the custom logo feature in the url (node.js' hard limit on the header size)

Admittedly a fairly tall order, so I wonder if a hypothetical scenario that only had some of those characteristics (and/or other ones) would be sufficiently compelling.

[...]

PR #9645, opened 2023-10-08
Fix issue with custom logo generation in badge-maker (#9644)

zavoloklom:
* allow `logo`, `logoPosition`, `logoWidth`, and `links` as expected keys

* add validation for `logo`, `logoPosition`, `logoWidth`, and `links` keys

* add validation tests


chris48s:
Thanks. I will have a look over this but first I really need to re-read https://github.com/badges/shields/issues/4947 From memory we never progressed it because there were some things we didn't really reach an agreement on. https://github.com/badges/shields/issues/9476 may decide one of them for us. Adding just base64 is a reasonable first step, but I want to be sure we're not doing it in a way that impacts how we would want to add named logos.

chris48s:
At some point in the past I looked at adopting a proper argument parsing library like yargs or commander. Because of the slightly non-standard way the arguments are currently parsed (positional argument 5 can be either `style` or `labelColor`, depending on whether it starts with an `@` or not), we'd have to make a non backwards-compatible change to the CLI in order to do that.

That doesn't mean we can't do it. It would make much more sense to have flags like `--labelColor blue` or whatever, rather than all positional args.

Lets not load it into this PR though. This PR is a relatively small change. Re-working the CLI is going to be a bigger job, so lets not block this PR on doing that as well.

badge-maker/index.d.ts
```diff
@@ -4,6 +4,10 @@ interface Format {
   labelColor?: string
   color?: string
   style?: 'plastic' | 'flat' | 'flat-square' | 'for-the-badge' | 'social'
+  logo?: string
```
chris48s, line 7: At this layer, lets call this param `logoBase64` instead of just `logo` so that it is not ambiguous with named logos when we add that.

badge-maker/index.d.ts
```diff
@@ -4,6 +4,10 @@ interface Format {
...
+  logoPosition?: number
+  logoWidth?: number
```
chris48s, line 9: I don't think there is a strong case for exposing `logoPosition` and `logoWidth` in the badge-maker package. `logoPosition` is not exposed at all by shields.io. `logoWidth` kind of is available by accident but it is not documented and we may deprecate or quietly remove it at some point.

badge-maker/index.d.ts
```diff
@@ -4,6 +4,10 @@ interface Format {
...
+  links?: Array<string>
```
chris48s, line 10: Adding links is issue https://github.com/badges/shields/issues/4949
The reason this has been sitting around for a while is because `links` is implemented internally as an array for historical reasons, rather than using the `label` and `message` terminology, but we'd really like to add it as `labelLink` and `messageLink` to match the other params.

The conclusion we came to on this (in https://github.com/badges/shields/pull/5542#issuecomment-692956549 ) was

> We should expose the public API `labelLink` and `messageLink` in the package, but its a bigger job than just setting `links = [labelLink, messageLink]` because currently `links[0]` affects the whole badge when `links.length == 1` but only the left side when `links.length == 2`. To make `labelLink` make sense as a param it should only ever affect the label (left) side of the badge but then just making that change in the package is then a BC break for the server so when we call out to `makeBadge()` in the server component we'll need to translate that to
> 
> ```js 
> labelLink = links[0]
> messageLink = links[0]
> ```
> 
> if `links.length == 1` to retain the existing behaviour of `?link` for SaaS users. We'll also need tests for all that.

but the PR was never finished.

chris48s (commented):
We'll need to update the documentation in
https://github.com/badges/shields/tree/master/badge-maker#format
to cover the new params.

badge-maker/lib/index.js
```diff
@@ -16,13 +16,32 @@ function _validate(format) {
...
 
+  const numberFields = ['logoWidth', 'logoPosition']
+  numberFields.forEach(function (field) {
+    if (field in format && typeof format[field] !== 'number') {
+      throw new ValidationError(`Field \`${field}\` must be of type number`)
+    }
+  })
+
+  if ('links' in format) {
+    if (!Array.isArray(format.links)) {
```
chris48s, line 34: can we validate that the `links` array has a max length of 2

badge-maker/index.d.ts
chris48s: Just thinking about this again.. in the endpoint schema, we expose `logoSvg` (which is literally an SVG as a string) and `namedLogo`. I wonder if we should just expose `logoSvg` here and encode it for the user to keep the package API consistent with the endpoint schema. Base64 is useful in URLs, but we don't have that constraint here. The only downside is we can't accept encoded PNG logos.

I am being quite indecisive here :thinking:

PyvesB:
@zavoloklom any chance you could address Chris's feedback in the coming weeks? :)

zavoloklom:
> @zavoloklom any chance you could address Chris's feedback in the coming weeks? :)

Hi, I'll do it this week

badge-maker/index.d.ts
zavoloklom: @chris48s so what do you think, keep it `logo` or change it to `logoBase64 `? I personally think that `logo` is better for consistence , but in future it could also work as `logo` property in other places

badge-maker/index.d.ts
zavoloklom: What do you think about adding but marking `logoWidth` as deprecated? I think it should be more clear.

Also I remove `logoPosition` from `make-badge.js` also as it does not used at all.

zavoloklom:
> We'll need to update the documentation in https://github.com/badges/shields/tree/master/badge-maker#format to cover the new params.

I add description for `logoBase64` and `links`

zavoloklom:
@chris48s Hi! I'll make all necessary changes and left comments. I'll squash my commits when you say it's ok to

badge-maker/index.d.ts
chris48s: As things stand right now, `logoWidth` is not exposed in the public API at the library level. If you call `makeBadge()` with a `logoWidth`, it will throw `ValidationError: Unexpected field 'logoWidth'`

I don't think we should expose `logoWidth` as deprecated and then remove it again later. Lets just keep this as it stands (i.e: not exposed) at the library layer.

chris48s:
Don't worry about squashing or rebasing. We squash-merge PRs on this project anyway.

Just keep pushing more commits to the branch. It is easier to see what has changed between rounds of review that way.

badge-maker/index.d.ts
chris48s: Meh. Lets roll with `logoBase64` for now. It is a quick easy way to expose _some_ way for package users to use logos. We can add `logoSvg` and/or `namedLogo` at another time. I don't think any of those concepts should be called just `logo`.

badge-maker/index.d.ts
zavoloklom: Ok, and what do you think about removing `logoWidth` from `makeBadge` function then as I did in [this commit](https://github.com/badges/shields/pull/9645/commits/c583fc7baa92300d32f3235fd9b9483d3ae51759)?

chris48s: I will reply to this, but I don't have time immediately

chris48s: Although it is not exposed via the NPM package, `logoWidth` is exposed via shields.io.
https://img.shields.io/badge/foo-bar-blue?logo=javascript
https://img.shields.io/badge/foo-bar-blue?logo=javascript&logoWidth=50
I would like to remove it, but just deleting the param here is not in isolation a massively helpful thing to do.

I've raised an issue about this: https://github.com/badges/shields/issues/10285

chris48s:
I feel like this PR adds a couple of features that are nearly ready to go but we're also going round in circles on some other things. I'm going to close this PR but I'm going to move the ideas you've raised here to the following 3 places:

1. #10283 - in this PR, I've picked out the work you've done on exposing logos and links in the package and fixed it (see [failing test case](https://github.com/badges/shields/pull/10283/commits/09f4d18cd7b9a4089e226a1307cbbf09a7545ae5) and [fix](https://github.com/badges/shields/pull/10283/commits/c6ceb5ed959a86d1e9523d20bcb81f5e5ce14498)). Lets see if we can get that merged quickly
2. #10284 - in this PR, I've finished the job of properly removing the redundant logoPosition` param. You're right to suggest removing it, but the implementation proposed here is incomplete.
3. #10285 - in this issue, I've proposed removing the `logoWidth` param, but I'll leave this one open for discussion for a bit before we do anything.

zavoloklom:
> I feel like this PR adds a couple of features that are nearly ready to go but we're also going round in circles on some other things. I'm going to close this PR but I'm going to move the ideas you've raised here to the following 3 places:
> 
> 1. [Expose `logoBase64` and `links` in badge-maker NPM package #10283](https://github.com/badges/shields/pull/10283) - in this PR, I've picked out the work you've done on exposing logos and links in the package and fixed it (see [failing test case](https://github.com/badges/shields/pull/10283/commits/09f4d18cd7b9a4089e226a1307cbbf09a7545ae5) and [fix](https://github.com/badges/shields/pull/10283/commits/c6ceb5ed959a86d1e9523d20bcb81f5e5ce14498)). Lets see if we can get that merged quickly
> 2. [Remove `logoPosition` #10284](https://github.com/badges/shields/pull/10284) - in this PR, I've finished the job of properly removing the redundant logoPosition` param. You're right to suggest removing it, but the implementation proposed here is incomplete.
> 3. [remove `logoWidth` param #10285](https://github.com/badges/shields/issues/10285) - in this issue, I've proposed removing the `logoWidth` param, but I'll leave this one open for discussion for a bit before we do anything.

Looking forward to see these changes!

Issue #9631, opened 2023-10-01
Ansible Galaxy badge does not work with new galaxy


cmeissner:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Ansible Galaxy badge shows `resource not found` since there is the new Ansible galaxy version live.

![image](https://github.com/badges/shields/assets/14936920/9d5edd2d-e9e7-41a3-a321-f9876e259c95)

Setting up a Galaxy badge want a `collectionId` which is not part of the data provided by api

![image](https://github.com/badges/shields/assets/14936920/1b6bf21f-98c2-4d7b-b091-ad6b4e30455a)

```shell
curl -s -XGET https://galaxy.ansible.com/api/v3/plugin/ansible/content/published/collections/index/codeaffen/phpipam/ | jq
{
  "href": "/api/v3/plugin/ansible/content/published/collections/index/codeaffen/phpipam/",
  "namespace": "codeaffen",
  "name": "phpipam",
  "deprecated": false,
  "versions_url": "/api/v3/plugin/ansible/content/published/collections/index/codeaffen/phpipam/versions/",
  "highest_version": {
    "href": "/api/v3/plugin/ansible/content/published/collections/index/codeaffen/phpipam/versions/1.7.0/",
    "version": "1.7.0"
  },
  "created_at": "2023-05-08T23:18:48.027350Z",
  "updated_at": "2023-10-01T09:12:16.157683Z",
  "download_count": 11649
}
```

### 🔗 Link to the badge

https://img.shields.io/badge/dynamic/json?style=flat&label=galaxy&prefix=v&url=https://galaxy.ansible.com/api/v2/collections/codeaffen/phpipam/&query=latest_version.version

### 💡 Possible Solution

The service need to use the current pulp API [1] for getting data for Ansible collections. 

[1] - [Get a specific collection](https://ansible.readthedocs.io/projects/galaxy-ng/en/latest/community/api_v3/#get-a-specific-collection)

chris48s:
So it looks like you're right that our native badges do need updating to the new API version. Thanks for flagging.

but the link you've posted is a dynamic badge, which you could update yourself: https://img.shields.io/badge/dynamic/json?style=flat&label=galaxy&prefix=v&url=https://galaxy.ansible.com/api/v3/collections/codeaffen/phpipam/&query=highest_version.version

chris48s:
Looks like the ansible role and ansible role name badges are still working.

[...]

PR #9648, opened 2023-10-10
deprecate/fix [ansible] galaxy services

chris48s:
This PR implements the proposals from https://github.com/badges/shields/issues/9631#issuecomment-1753494195

- deprecate ansible galaxy collection name
- deprecate ansible galaxy quality score
- deprecate ansible role name
- deprecate ansible role downloads and replace with a new (but non-backwards-compatible) ansible role downloads badge


chris48s:
I'm going to go ahead and merge this one. The existing ansible downloads badge is showing incorrect data.

PR #9658, opened 2023-10-13
Fix several typos

kurtmckee:
I observed that "separated" was misspelled at the top of https://shields.io/badges so this PR addresses those typos.

I then ran the codespell tool, which identified some additional typos (some of which might not be user-facing). Those are included in this PR, too.


chris48s (approved):

chris48s:
thanks

PR #9661, opened 2023-10-18
[GitHub] add latest release of a branch badge

karolwydmuch:


bgalek (approved):

chris48s:
Hi. Thanks for opening a PR.

A badge should be quick to render. In most contexts where our badges are displayed there's a hard limit of 4 seconds on a request. In general we try usually avoid accepting badges that require more than one network call to render for this reason.

This means badges that require iterating over multiple pages of API response are impractical. They end up being too slow to render in a lot of cases.

I had a look to see if it would be possible to obtain this information in a single GraphQL query but it seems this is not exposed by the GraphQL API. https://docs.github.com/en/graphql/reference/objects#release

If we can't find a more efficient way to query it, I'm inclined to decline this one.

Out of interest, and also for testing purposes, do you have an example of a repo that publishes releases against different branches? It is not something I have come across before and I don't think this corresponds to an open feature request.

services/github/github-latest-branch-release.tester.js
```diff
@@ -0,0 +1,33 @@
...
+import { ServiceTester } from '../tester.js'
+
+export const t = new ServiceTester({
+  id: 'GithubLatestBranchRelease',
+  title: 'Github Latest Branch Release',
+  pathPrefix: '/github',
+})
+
+t.create('Release')
+  .get('/v/release/expressjs/express.json')
```
chris48s, line 12: If we end up coming back to this, none of the tests in this file are actually testing the service class that has been added in this PR, but lets not get bogged down in that for the moment

PyvesB:
@karolwydmuch any thoughts following Chris's comments? :)

PyvesB:

If others think that repository level projects would be an option and want to pick this up, you can simply fetch the commits by running `git fetch origin pull/9661/head:pr-9661`. Change the implementation to target repository level projects and open a new pull request with a [co-authored trailer](https://docs.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors) included in the commit message to give attribution to the original author.

Issue #8816, opened 2023-01-17
add down_message and up_message to uptime robot badge


M4TZ3-op:
would be verry nice, thanks :)

chris48s:
If someone wants to work on this, there are helpers in https://github.com/badges/shields/blob/master/services/website-status.js which are used by `website` and `nodeping` services that could be applied to uptimerobot as well.

pr7prashant:
Will try this one.

filipgrudzien:
Since it's not done yet, I will try to fix it

filipgrudzien:
https://github.com/badges/shields/pull/9662 I attempted to add it, please let me know if this is what was expected.

PR #9662, opened 2023-10-18
feat: added up_message and down_message to [uptimerobotstatus]

filipgrudzien:
closes https://github.com/badges/shields/issues/8816


chris48s:
Thanks. Could we also implement the `up_color` and `down_color` params.

Also, we'll need to add them to the documentation, but you can ignore that in this PR.
I'll deal with it as part of #9612

chris48s (approved):

PR #9666, opened 2023-10-19
implementation for an artifactory service.

cdancy:
Looking for feedback on this impl and whether or not I'm going down the correct path and if this is something you all would be interested in taking in. I more/less just copied a lot of what I saw others doing in other services and did the same here to get things up and going quick. Tests still need to be written but the initial impl works as expected with no issues to speak of. 

The service itself is basically just a wrapper around Artifactory's **latest version** endpoint the documentation of which you can [FIND HERE](https://jfrog.com/help/r/jfrog-rest-apis/artifact-latest-version-search-based-on-layout).

@calebcartwright let me know what you think and if this is something you all would consider moving forward on. If not, and we had to maintain this out-of-tree, is there a good process for doing so other than the obvious you all would recommend?

### DIFF
--- a/services/artifactory/artifactory.service.js
+++ b/services/artifactory/artifactory.service.js
@@ -1,0 +1,211 @@
+import Joi from 'joi'
+import { optionalUrl } from '../validators.js'
+import { BaseService } from '../index.js'
+import { renderVersionBadge } from '../version.js'
+import { getCachedResource } from '../../core/base-service/resource-cache.js'
+
+const ARTIFACTORY_LATEST_VERSION_ENDPOINT = 'api/search/latestVersion'
+const ONE_HOUR = 1 * 3600 * 1000
+
+const queryParamSchema = Joi.object({
+  // auth related optional properties
+  server: optionalUrl.default(''),
+  token: Joi.string().default(''),
+  username: Joi.string().default(''),
+  password: Joi.string().default(''),
+
+  // api related optional properties
+  repos: Joi.string().default(''),
+  remote: Joi.string().default(''),
+  version: Joi.string().default(''),
+}).required()
+
+const documentation = `
+  <p>
+    Query for the last released version of an artifact using the <b>Latest Version</b> endpoint.
+    <br />
+    <br />
+    This service will cache results for <b>1 hour</b> so as not to overload a given Artifactory instance with potentially long running query operations.
+    <br />
+    <br />
+    Authentication is done either through basic auth or using a bearer token. While the options exist to pass along credentials as query params, users
+    are strongly encouraged to instead set the relevant auth related environment variables server side which this service will pick up and read at runtime.
+    <br />
+    <br />
+    The following environment variables are supported:
+    <ul>
+      <li>ARTIFACTORY_URL</li>
+      <li>ARTIFACTORY_TOKEN</li>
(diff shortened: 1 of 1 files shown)


services/artifactory/artifactory.service.js
```diff
@@ -0,0 +1,211 @@
...
+      queryParams: {
+        repos: 'libs-release-local',
+        remote: '1',
+        version: '1.0-SNAPSHOT',
+        server: 'https://my.artifactory.server/artifactory',
+        token: 'OTEfj876SDF2345JKIB/H&+FFGH',
+        username: 'jerry',
+        password: 'seinfeld',
+      },
+      staticPreview: Artifactory.render('2.1.0'),
```
cdancy, line 74: No matter what I did I couldn't get any of this to show up in the local "integ test server". I assume I'm doing something wrong but filled out all the relevant links/options/docs/etc anyway if only for completeness sake.

services/artifactory/artifactory.service.js
```diff
@@ -0,0 +1,211 @@
...
+  async handle(
+    { group, artifact },
+    { server, token, username, password, repos, remote, version },
+  ) {
+    // Artifactory can return plain/text or json for this endpoint we're hitting so
+    // we need to effectively accept all as we're not sure what we're going to be returned
+    const defaultHeaders = { Accept: '*/*' }
+
+    // dump query params to debug console
+    console.debug(`***** artifactory found the following query properties *****
```
cdancy, line 162: Don't know if you all care about keeping this in but it was generally useful for me when debugging to see what was coming through as values for the query parameters. Let me know if you want me to rip it out.

cdancy:
@calebcartwright thoughts one way or another? I can continue with writing some integration tests (unless you have units which would be easier) but wanted to get your thoughts/opinions one way or another or just those from the community at large.

calebcartwright:
As far as the implementation goes, I feel like it's going in a vastly different direction than our typical services and I'm not entirely sure why that would need to be the case.

I'd suggest reviewing some of our material, linked below for reference, to reconsider the implementation or to at least help inform a subsequent explanation as to why this would require taking such a different path:

* https://github.com/badges/shields/blob/master/doc/TUTORIAL.md
* https://github.com/badges/shields/blob/master/doc/authentication.md
* https://github.com/badges/shields/blob/master/doc/adding-new-config-values.md
*

cdancy:
> I wonder if there is a way we could provide a plugin hook to allow self-hosting users to conveniently inject a custom service like this without having to maintain a fork. 

@chris48s gets is ;)

PyvesB:
> One significant issue with this is that, afaik, there's no default public instance of Artifactory that we can rely on for our integration and testing. As such, this is a badge that seems like it would only be applicable for self-hosted instances and which would have no utility on the primary Shields.io instance.

>  think I'm still 👎 on accepting badges where there is no public instance and would only be relevant to self-hosting users. The thing that give me most pause about taking on the maintenance and support for something like this is that if there is no public instance of this API on the internet, it is basically impossible for me to test that this code "works".

I'm personally aligned on these points as well. I don't think we'll want to move forward with this badge for the time being, so will go ahead and close the PR. Thanks for your interest in the Shields.io project, @cdancy!

Issue #9677, opened 2023-10-21
Latest twitter logo update


mishrasamiksha:
### 📋 Description

I would like to update the twitter logo to X (which is the latest one).

### 🔗 Data

NA

### 🎤 Motivation

NA

chris48s:
I think we covered this in https://github.com/badges/shields/pull/9496
Can you explain further?

mishrasamiksha:
Sorry I meant in the readme file of the repo. I can see still the old version of logo for the twitter

chris48s:
I see - thanks

PyvesB:
This was effectively done in #9678, and the badge on our README got removed in #9685. Will go ahead and close.

PR #9678, opened 2023-10-21
twitter isn't twitter any more

chris48s:
Refs #9677

Tbh, we could just drop linking to this.

@PyvesB - are you still checking https://twitter.com/shields_io ? I think you're the only person who has access to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@
         <img src="https://img.shields.io/discord/308323056592486420?logo=discord"
             alt="chat on Discord"></a>
     <a href="https://twitter.com/intent/follow?screen_name=shields_io">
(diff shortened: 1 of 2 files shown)


calebcartwright (approved):
I thought SI would redirect/map to the X logo but oh well. I've no objections to dropping this either if folks would rather go that route

jNullj (approved):
I think its best to go by the new brand they set rather then the old one to keep up to date.
Small change but LGTM

chris48s:
I'm going to just merge this for now, but we could remove in future.
It would still be nice to hear back from @PyvesB on this.

PyvesB:
I've not actively monitored it in some time, but still have access to it.

chris48s:
Thanks. I've followed up in https://github.com/badges/shields/pull/9685

Issue #9493, opened 2023-08-19
Improve our approach for testing auth


chris48s:
Our auth tests tend to fall into one of two categories

Either we implement auth on a base class, then test the auth is handed correctly in the base class using a dummy
https://github.com/badges/shields/blob/master/services/stackexchange/stackexchange-base.spec.js

Or we pick one concrete implementation and test that
https://github.com/badges/shields/blob/master/services/gitlab/gitlab-tag.spec.js

This is not water-tight.

As @calebcartwright suggests in https://github.com/badges/shields/pull/9387#discussion_r1299106960
it would be good if we could be a bit more robust about this and make sure we cover all the concrete service classes that are supposed to use auth.

I've not tried it yet, but one possible solution might be to encapsulate the test boilerplate once in a function, then loop over all the relevant classes and call it

something like

```js
function runTheTests(ServiceClass) {
  const service = new ServiceClass();
  // setup
  // make assertions
}

for (const serviceClass of [ServiceClass1, ServiceClass2]) {
  runTheTests(serviceClass)
}
```

calebcartwright:
@jNullj wanted to flag your attention in case this was something you'd be interested in working on (definitely no pressure though!)

jNullj:
I gave this a quick look and i think i am up to the task.
Just to make sure i got the gist of it, the idea is to test every service (and not just base service) while using the same base function.
Would that mean i should probebly loop over services for each of the base classes the require auth?
Should I add those tests to each base class spec file?

chris48s:
> Just to make sure i got the gist of it, the idea is to test every service (and not just base service) while using the same base function.

Bingo

> Would that mean i should probebly loop over services for each of the base classes the require auth?
> Should I add those tests to each base class spec file?

First off: I think probably the best way to tackle this will be to pick one service. Lets say `stackexchange` for the sake of argument - just so we can use concrete examples in the next bit - but we could pick another one first. Work on a PR for that one first to establish the pattern. That will mean we don't have to modify too much code if there is a bit of back-and-forth on the PR. Then once we've reviewed that and decided on an approach we like we can do another PR (or a series of them) applying that approach to the other services.

Beyond that, I think that's a couple of ways to approach it:

One option would be the example I gave in the top post. So that one would probably look like a single test file (in this case `stackexchange.spec.js`) which imports `StackExchangeMonthlyQuestions`, `StackExchangeReputation`, `StackExchangeQuestions` and tests all of them.

[...]

PR #9681, opened 2023-10-23
Improve our approach for testing auth (part 1)

jNullj:
This PR refs #9493 

As mentioned in the issue, this is a first attempt to improve auth testing with stackexchange as a study case and hopefully a first step.
I tried to stick with making everything testable from within the same file while reusing the same code.
If `testAuth` is very reusable it might be better to move it into `test-helpers.js` to save a few lines of code, i suspect we might be able to use it for most tests saving a few lines of code per test.

I think this could be improved if i had a way to generate the `dummyResponse` automatically from the oneapi object (should be possible in theory, there are some unmaintained modules that could do that, should i make my own to only cover what is needed by shields?)


jNullj:
Here is an example for auth test of stackexchange as a case study for the rest of the project.
If it looks good I can try and apply this pattern over the project in chunks
image

chris48s:
Sorry I haven't managed to look at this one yet. I will make sure to follow this one up over the weekend.

services/stackexchange/stackexchange-base.spec.js
```diff
@@ -1,12 +1,27 @@
...
 import { expect } from 'chai'
 import nock from 'nock'
+import { pathParams } from '../index.js'
 import { cleanUpNockAfterEach, defaultContext } from '../test-helpers.js'
 import { StackExchangeBase } from './stackexchange-base.js'
+import StackExchangeMonthlyQuestions from './stackexchange-monthlyquestions.service.js'
+import StackExchangeReputation from './stackexchange-reputation.service.js'
+import StackExchangeQuestions from './stackexchange-taginfo.service.js'
 
 class DummyStackExchangeService extends StackExchangeBase {
```
chris48s, line 11: If we're explicitly testing every concrete implementation of a base class, we can actually just get rid of the dummy implementation. I don't think it really adds any value.

services/stackexchange/stackexchange-base.spec.js
```diff
@@ -16,23 +31,47 @@ class DummyStackExchangeService extends StackExchangeBase {
...
+}
 
-    it('sends the auth information as configured', async function () {
-      const scope = nock('https://api.stackexchange.com')
-        .get('/2.2/tags/python/info')
-        .query({ key: 'fake-key' })
-        .reply(200, { message: 'fake message' })
+function testAuth(serviceClass, dummyResponse) {
+  describe(serviceClass.name, function () {
+    describe('auth', function () {
```
chris48s, line 48: This pattern of putting the `describe`s inside a function declaration is a bit unusual. It works, but for me it violates the [rule of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). More idiomatic would be to do it the other way round and  define a test helper, then write tests (at the module level) that use it. Something like https://github.com/badges/shields/blob/b2c2a1879616cce51cd562af2fac5f09592d5484/badge-maker/lib/make-badge.spec.js would be an example of that.

I guess the reason why we've done it this way round is because it becomes difficult to do `describe(serviceClass.name` the other way round.

I guess that leads me towards saying maybe the correct way to do this would be a test file for each service that does something more like

`stackexchange-monthlyquestions.spec.js`

```js
// bla bla imports
import { testAuth } from './test-helpers.js'

describe('StackExchangeMonthlyQuestions', function () {
  describe('auth', function () {
    it('sends the auth information as configured', async function () {
      testAuth(StackExchangeMonthlyQuestions, { total: 8 })
    })
  })
})
```

The other advantage of that is: If we add another service, it is more obvious that you also need to add tests while still abstracting the boilerplate. It is less obvious you would need to add a test case of a concrete implementation to `stackexchange-base.spec.js`.

How does that sound?

services/stackexchange/stackexchange-base.spec.js
```diff
@@ -16,23 +31,47 @@ class DummyStackExchangeService extends StackExchangeBase {
...
-        await DummyStackExchangeService.invoke(defaultContext, config, {}),
-      ).to.deep.equal({ message: 'fake message' })
+      const config = { private: { stackapps_api_key: 'fake-key' } }
+      const firstOpenapiPath = Object.keys(serviceClass.openApi)[0]
+      const exampleInvokeParams = serviceClass.openApi[
+        firstOpenapiPath
+      ].get.parameters.reduce((acc, obj) => {
+        acc[obj.name] = obj.example
+        return acc
+      }, {})
```
chris48s, line 58: Having read it, I can see what you're doing here but initially it wasn't obvious.
I think it would really help if we split these two statements out into a named function (maybe `getExampleCall()`?), write a quick comment explaining that we're using the documentation examples to extract a valid call we can make. Then this could actually go in https://github.com/badges/shields/blob/master/services/test-helpers.js because the same function would be applicable to any service. As we refactor more auth tests, we will want to use this everywhere.

chris48s (commented):
Thanks for working on this. I've left some comments. I think having seen this implementation, I'm now actually leaning towards the second option I mentioned in https://github.com/badges/shields/issues/9493#issuecomment-1685295109 but it is really useful to see what you've done here.

services/stackexchange/stackexchange-base.spec.js
jNullj: Good point, I will remove it after we sort the other comments

services/stackexchange/stackexchange-base.spec.js
jNullj: I tried to make it more compact but this introduced some complexity and i also seem to result in a leas readable code. At least we learn from that attempt how to better move forward.

It does sound like a good idea to separate the files, with the scale of tests we use its better to keep it simple in a  single file for each auth test, similar to the example you provided here, rather then a more compact array in the base spec file, as there aren't that many.

I could rewrite this PR in a few days.

services/stackexchange/stackexchange-base.spec.js
jNullj: Its a great idea to break out this function and reuse the code at test-helpers. Will do that.
I will rework the function to make it as readable as i can.

jNullj: Fixed in https://github.com/badges/shields/pull/9681/commits/18ec38723a38bdbeca1edd7b6e6d057a361375ac

services/stackexchange/stackexchange-base.spec.js
jNullj: Fixed with https://github.com/badges/shields/pull/9681/commits/9dd597dcf2ab772077b23e9b9918114c88f5384d

jNullj:
I did everything we talked about so far in the latest commits but i do have a third option to consider.
I noticed we have the authorizedOrigins property.
I could potentially use that to make one loop that finds all classes that can be tested for auth in one go, this would be a spec file at the root of the services called something like `authTest.spec.js` and it could do the following:

1. Go over all services and find everything that has `authorizedOrigins` and OpenAPI example
2. Run the authTest function for each of those

The only catch is that I will have to find a way to craft dummy responses automatically.
But that would remove the need of manually going over all services to add tests, and also cover every future service added automatically assuming it has an OpenAPI example.

Do you think that's over-complicating or should i give that a try? Would that fit well with the project style and code philosophy?

services/test-helpers.js
```diff
@@ -30,6 +32,99 @@ function noToken(serviceClass) {
...
+async function testAuth(serviceClass, dummyResponse) {
+  if (!(serviceClass.prototype instanceof BaseService)) {
+    throw new TypeError(
+      'Invalid serviceClass: Must be an instance of BaseService.',
+    )
+  }
+
+  cleanUpNockAfterEach()
+
+  const config = { private: { stackapps_api_key: 'fake-key' } }
```
chris48s, line 103: This is specific to the stackexchange badges. If this is going to be a completely generic helper we can use for all services, we'll need to pass this into `testAuth` as a param. I'd suggest we make the config object the first or second param to `testAuth()`. This one is small, so lets address this one before merging.

services/test-helpers.js
```diff
@@ -30,6 +32,99 @@ function noToken(serviceClass) {
...
+  const exampleInvokeParams = getBadgeExampleCall(serviceClass)
+  const authOrigin = serviceClass.auth.authorizedOrigins[0]
+
+  if (!authOrigin) {
+    throw new TypeError(`Missing authorizedOrigins for ${serviceClass.name}.`)
+  }
+
+  const scope = nock(authOrigin)
+    .get(/.*/)
+    .query(queryObject => queryObject.key === 'fake-key')
```
chris48s, line 113: Again, this is a bit specific to this service. This will work for services that pass auth in the query string, but won't for services where we pass auth in a header, for example. We're probably going to need a few different variants of this function for different auth methods (basic auth, header, query string, etc). I don't think that will be too hard - there's a few different ways we can do this.

I think I'd be fine just merging this how it is and saying at the moment it only works for querystring auth. Then we work out how to generalise this as we expand more services to use this helper. I usually find it easier to think about this stuff once I have a concrete problem I am trying to apply it to.

services/test-helpers.js
```diff
@@ -30,6 +32,99 @@ function noToken(serviceClass) {
...
+    .reply(200, dummyResponse)
+  expect(
+    await serviceClass.invoke(defaultContext, config, exampleInvokeParams),
+  ).to.not.have.property('isError')
```
chris48s, line 117: Nice - this is a good choice of generic assertion :)

services/test-helpers.js
jNullj: So i went a bit overboard and just made everything generic and support all auth methods.
So now i have a function to generate a fake config object based on the service class.
See ffc780091071cf4dc1cc2a445bcedaa63c8a1be8

services/test-helpers.js
jNullj: See ffc780091071cf4dc1cc2a445bcedaa63c8a1be8, its all more generic now and should handle most service classes

jNullj:
Oh nvm, i will add another fix tomorrow

chris48s:
Thanks for looking at all the additional auth methods. I think the approach of converting our own source code to text and attempting to parse javascript with regex is not a good solution to this though. It is going to create a situation changes to the `AuthHelper` can have quite non-obvious knock-on effects.

Broadly, there are 2 ways we can handle this:

1. Pass in params when we call `testAuth` that specify
    - the auth method and
    - custom header name, if applicable
2. Refactor auth helper so that we store all the auth info in properties on the service class.

Option 1 is kind of less nice because there will be a little bit of duplication between the service class and the tests. We specify the auth method twice, and in some cases we might specify the header name twice too.
Option 2 is perhaps the most ideal solution, but it is also way more of a lift and a big scope creep from the original intent of this. We risk straying far from the original point by embarking on that now.

I'd be in favour of option 1, at least to start with.

jNullj:
@chris48s 
While the current regex-based approach works for testing, it might not be ideal for long-term maintainability. Could we explore alternative methods like class properties for improved robustness and clarity? Would love to hear if you got other ideas from your expiriance.

jNullj:
Will fix this tomorrow
```js       
 .query(queryObject => queryObject.key === fakeSecret)
```

jNullj:
Ok, the last thing i want to do is add en example for each auth method in thie PR
im only missing ApiKeyHeader and JwtAuth

jNullj:
Fixed a few things and made sure we got at least 1 example for each authentication method.
Testing seems to fail due to unrelated issue. Maybe we are missing pepy secret for api access on github?

chris48s:
Thanks. The latest iteration of this looks really good. Only one minor comment

services/test-helpers.js
```diff
@@ -30,6 +33,314 @@ function noToken(serviceClass) {
...
+    throw new TypeError(
+      'Invalid serviceClass: Must be an instance of BaseService.',
+    )
+  }
+  if (!fakeKey || typeof fakeKey !== 'string') {
+    throw new TypeError('Invalid fakeKey: Must be a String.')
+  }
+  if (!fakeauthorizedOrigins || typeof fakeauthorizedOrigins !== 'object') {
+    throw new TypeError('Invalid fakeauthorizedOrigins: Must be an array.')
+  }
```
chris48s, line 114: The check and exception test don't seem to match here

chris48s (approved):

chris48s:
Cool. This looks great.

Just FYI, if you want to follow up more of these https://github.com/badges/shields/issues/9493#issuecomment-1904056131 we can skip running the service tests on these PRs.

Putting the service names in square brackets runs the service tests - the ones in files called `.tester.js`. We don't run all of these on every PR because running them all is quite slow and they can be flaky.

The tests in files called `.spec.js` are always run, even if the `.spec.js` files live in `/services`. We run all of these on every PR because they're quick and reliable.

So if we're not actually modifying the service code, it is ok to skip the service tests.

PR #9685, opened 2023-10-25
stop linking to twitter

chris48s:
Follow on from https://github.com/badges/shields/pull/9678#issuecomment-1777898375

Given the one person with access to this is no longer actively checking it, I'm not convinced it is a useful thing to still be pointing people at.

So we could just.. stop.


calebcartwright (approved):
Agreed

PR #9713, opened 2023-11-04
migrate examples to openApi part 20: TUTORIAL.md

chris48s:
Refs #9285

This PR updates the tutorial for adding a new badge to document the `openApi` object instead of the `examples` array.

Obviously we're still working through the process of migrating so the codebase itself is still a bit of a mix. That said, I think after we merge this, we should stop accepting services that use `examples` in new contributions.

### DIFF
--- a/doc/TUTORIAL.md
+++ b/doc/TUTORIAL.md
@@ -134,9 +134,9 @@ export default class Example extends BaseService {
 
 Description of the code:
 
-1. Our service badge class will extend `BaseService` so we need to require it. Variables are declared with `const` and `let` in preference to `var`.
+1. Our service badge class will extend `BaseService` so we need to import it.
 2. Our module must export a class which extends `BaseService`.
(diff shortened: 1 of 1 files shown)


doc/TUTORIAL.md
```diff
@@ -269,42 +269,88 @@ import { NotFound } from '../index.js'
...
+
+   ```js
+   parameters: pathParams({
+     name: 'gem',
+     description: 'Name of the Ruby Gem',
+     example: 'formatador',
+   })
+   ```
+
+   is equivilent to
```
PyvesB, line 332:
```suggestion
   is equivalent to
```

PyvesB (approved):
Great documentation updates, very clear! 👍🏻

chris48s:
This one is a milestone :) From this point in, I'll ask contributors submitting PRs that still use `examples` to rewrite using `openApi`. It felt a bit mean while our docs still said to use the old style.

Btw, thanks for ploughing through some reviews for me over the last few days :+1:

PR #9715, opened 2023-11-05
add meaningful descriptions including keywords; affects [amo bitcomponents coincap conan cpan]

chris48s:
Refs #9285

This PR comes out of a comment in the PR #9500 review: https://github.com/badges/shields/pull/9500#discussion_r1378685603

I haven't tackled every single case, but I decided to start off by taking an initial subset of services where we removed keywords when I did earlier conversion PRs, just to think it through and get a feel for this task.

In order to make text available to our search plugin, it needs to be user-visible text on the page. We can't give it "hidden" tags or keywords. In general, I don't think that's a bad thing: One of the things we should be working towards is providing clearer user-facing documentation. If you land directly on https://shields.io/badges/clojars-downloads (for the sake of argument), I think we can do more to explain "what is this thing, and why do you care about it" than we do now.

I think having tried to do a few of these, this format of describing the upstream service e.g:

> [Conan](https://conan.io/) is a package manager for C++"

seems to lend itself reasonably well to achieving that objective and fitting all the right keywords in. It also means if we've got 10 badges for a service, we can rely on the "summary" field to describe the badge and share a more generic description across a bunch of badges if we want. We don't need to write a slightly different description for each one just to get the keywords in (although in some cases, the might be useful for some other reason). Interested to know if others think this "voice" works stylistically.

I've not really hit this situation so far, but another possible thing we could do if we're struggling to mash a word soup of keywords into a natural language sentence or two is literally write something like

> keywords: firefox, amo, extension

or

> tags: firefox, amo, extension

at the end of the description just to make those words searchable text on the page.

/cc @PyvesB

### DIFF
--- a/services/amo/amo-base.js
+++ b/services/amo/amo-base.js
@@ -2,7 +2,8 @@ import Joi from 'joi'
 import { nonNegativeInteger } from '../validators.js'
 import { BaseJsonService } from '../index.js'
 
-const keywords = ['amo', 'firefox']
+const description =
+  '[addons.mozilla.org](https://addons.mozilla.org) (AMO) publishes extensions for Mozilla Firefox'
 
 const schema = Joi.object({
   average_daily_users: nonNegativeInteger,
@@ -26,4 +27,4 @@ class BaseAmoService extends BaseJsonService {
   }
 }
 
-export { BaseAmoService, keywords }
+export { BaseAmoService, description }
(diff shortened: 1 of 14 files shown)


PyvesB (approved):
I like what you've done here. No concerns from my side 👍🏻

chris48s:
Thanks. I'll merge the progress so far and continue to follow up in a similar style

PR #9716, opened 2023-11-06
migrate examples to openApi part 21; affects [clojars f-droid packagist piwheels vpm]

chris48s:
Refs https://github.com/badges/shields/issues/9285 Really starting to get into the services with query params now.

Also in this PR I have carried forward the assumptions from https://github.com/badges/shields/pull/9715 and documented these services in a similar style.


PyvesB (approved):
Looks good to me 👍🏻

PR #9731, opened 2023-11-12
migrate examples to openApi part 22; affects [steam]

chris48s:
Refs #9285

I'm starting to feel like a broken record at this point.


PyvesB (approved):
Carefully read through, couldn't fault it 😄

PR #9732, opened 2023-11-12
migrate examples to openApi part 23; affects [wordpress]

chris48s:
Refs #9285

..and on and on and on and on..

### DIFF
--- a/services/wordpress/wordpress-downloads.service.js
+++ b/services/wordpress/wordpress-downloads.service.js
@@ -1,6 +1,7 @@
 import Joi from 'joi'
 import { renderDownloadsBadge } from '../downloads.js'
-import { documentation, BaseWordpress } from './wordpress-base.js'
+import { pathParams } from '../index.js'
+import { description, BaseWordpress } from './wordpress-base.js'
 
 const dateSchema = Joi.object()
   .pattern(Joi.date().iso(), Joi.number().integer())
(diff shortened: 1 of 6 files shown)


services/wordpress/wordpress-downloads.service.js
```diff
@@ -52,14 +53,28 @@ function DownloadsForExtensionType(extensionType) {
...
-        title: `WordPress ${capt} Downloads`,
-        namedParams: { interval: 'dm', slug: exampleSlug },
-        staticPreview: this.render({ interval: 'dm', downloads: 200000 }),
-        documentation,
-      },
-    ]
+    static get openApi() {
+      const key = `/wordpress/${extensionType}/{interval}/{slug}`
+      const route = {}
+      route[key] = {
```
chris48s, line 59: This isn't a really common pattern, but we do use it in a handful of places.

With most of the wordpress services, we define a function that returns a class, and then call that function with `'plugin'` and `'theme'` to create two variants: One for wordpress plugins, one for wordpress themes. This reduces duplication, but we can't define `openApi` as a static property because the route key can't be a variable or a template string.

```js
static openApi = {
  `/${someVar}/{param}`: //some object
}
```

is not legal javascript.

In this situation, we can use a `static get` function.

PyvesB (approved):
All good! 👍🏻

PR #9736, opened 2023-11-13
fix [GitHubCommitActivity] service tests

chris48s:
Hit failures on this in a couple of PRs lately.
This is going to be hard to write in a way that doesn't just keep breaking, so I've changed these test cases to allow `isZeroOverTimePeriod`


chris48s:
I'm going to just merge this. Fixing this seems non-controversial

Issue #9730, opened 2023-11-11
Pepy Downloads Badge "invalid" on any package


MrTeferi:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

**Downloads** > [**Pepy Total Downloads**](https://shields.io/badges/pepy-total-downlods)
Badge  Is failing on any package given, provides a gray "invalid" badge. Also there is a typo in the title for this badge "Pepy Total Downlods" should be amended to "Downloads", which is a simple enough fix.

## To Reproduce
Try with package "django". You can see the service still works on their [main site here](https://www.pepy.tech/projects/django).

### 🔗 Link to the badge

https://shields.io/badges/pepy-total-downlods

### 💡 Possible Solution

Unsure at this time.

chris48s:
This is failing because we're getting a `401 Unauthorized` from pepy.

I've followed up with pepy: https://github.com/psincraian/pepy/issues/573#issuecomment-1807199431

MrTeferi:
Thanks for the followup and clarification :)

chris48s:
Next jobs:

- Update https://github.com/badges/shields/blob/master/services/pepy/pepy-downloads.service.js to pass auth in a `X-Api-Key` header. We should be able to use the `withApiKeyHeader()` auth helper
- Set API keys in CI, staging and prod environments
- Share pepy credentials with rest of team

PR #9748, opened 2023-11-18
call [pepy] with auth

chris48s:
Closes https://github.com/badges/shields/issues/9730
Refs https://github.com/psincraian/pepy/issues/573#issuecomment-1807199431

Not much actual code needed here, but a fair bit of admin around it. I've set

`SERVICETESTS_PEPY_KEY` in

- https://github.com/badges/shields/settings/secrets/actions
- https://github.com/badges/shields/settings/secrets/dependabot

`PEPY_KEY` is already set in prod and staging envs on fly

Creds are in keybase and invites sent.


chris48s:
Can confirm service tests and review apps working with credentials. I had to run the "Create/Update Review App" workflow from the `9730-pepy` branch to get it to pick up the new secret.

chris48s:
I'm going to go ahead and merge this one. Fixing this seems non-controversial

PR #9749, opened 2023-11-19
migrate some services from examples to openApi part 25; affects [ctan dub hexpm jira pulsar scoop]

chris48s:

I wonder what number the last of these PRs will be?

### DIFF
--- a/services/hexpm/hexpm.service.js
+++ b/services/hexpm/hexpm.service.js
@@ -2,7 +2,7 @@ import Joi from 'joi'
 import { renderDownloadsBadge } from '../downloads.js'
 import { maybePluralize } from '../text-formatters.js'
 import { renderVersionBadge } from '../version.js'
-import { BaseJsonService } from '../index.js'
+import { BaseJsonService, pathParams } from '../index.js'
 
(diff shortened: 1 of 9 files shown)


services/hexpm/hexpm.service.js
```diff
@@ -96,52 +108,58 @@ class HexPmVersion extends BaseHexPmService {
   }
 }
 
-function DownloadsForInterval(downloadInterval) {
```
chris48s, line 99: There is a bit of refactoring going on here.

Previously, this service used the same pattern we see in https://github.com/badges/shields/pull/9732#discussion_r1390449157

Previously we defined a function that returns a class, and then called that function with `'day'`, `'week'` and `'all'` to generate 3 classes. One for each interval.

I don't think this really buys us much here. I converted this to a single class with an `:interval(dd|dw|dt)` param. This allows us to have a single example page and show interval as a drop-down menu, which is much more in line with how we do all our other downloads badges.

PyvesB: Makes sense!

PyvesB (approved):
Another one that looks great! 👍🏻

PR #9750, opened 2023-11-20
Edit Discord badge.

MichaelHinrichs:
Use the official Discord logo's color value of ~5662F6~ 5865F2.


MichaelHinrichs:
Actually, I think the Discord badge should be this way by default. https://github.com/badges/shields/pull/9751

calebcartwright:
Thanks for the suggestion but I'm going to close this for largely the same reasoning as #9751 

The defaults are correct, and while we provide features that allow users to customize the look of the badges, we try to stick with that default appearance in our own content/documentation

PR #9751, opened 2023-11-20
[Discord] Edit badge defaults.

MichaelHinrichs:
Label the badge as "Discord" instead of "chat". Use the official Discord logo's color value of 5865F2.


calebcartwright:
Thanks for the PR but going to close because the current defaults are correct and intentional.

For reference, our badge defaults are based on our [specification](https://github.com/badges/shields/blob/master/spec/SPECIFICATION.md)

PR #9752, opened 2023-11-20
migrate examples to openApi part 26; affects [gitlab]

chris48s:

There's a few things going on in this PR. Slightly more than I would ideally like for one of these PRs.

## Get the tests passing

dba3c60 - this test was failing due to a license change on the upstream repo
Given all we are really testing there is that we apply the right colour if the license is "other", I've just replaced this with a mock rather than finding another repo we can use.

fd67c2b , 64f6948 - these tests for the pipeline status/coverage badges (which use `BaseSvgScrapingService` but should really me migrated to use the API) were failing due to a change in the upstream behaviour. I've just updated the tests rather than handling in the code (again)

## Migrate examples to openApi

82131f9 - Most of this is pretty standard procedure. However I've also done a bit of refactoring on the issues and merge requests badges in here. I'll call it out in an inline comment.

### DIFF
--- a/services/gitlab/gitlab-issues.service.js
+++ b/services/gitlab/gitlab-issues.service.js
@@ -1,7 +1,8 @@
 import Joi from 'joi'
+import { pathParam, queryParam } from '../index.js'
 import { optionalUrl, nonNegativeInteger } from '../validators.js'
 import { metric } from '../text-formatters.js'
-import { documentation, httpErrorsFor } from './gitlab-helper.js'
+import { description, httpErrorsFor } from './gitlab-helper.js'
 import GitLabBase from './gitlab-base.js'
 
 const schema = Joi.object({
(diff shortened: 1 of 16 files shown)


services/gitlab/gitlab-issues.service.js
```diff
@@ -19,192 +20,44 @@ const queryParamSchema = Joi.object({
...
-</p>
-`
-
 export default class GitlabIssues extends GitLabBase {
   static category = 'issue-tracking'
 
   static route = {
     base: 'gitlab/issues',
-    pattern: ':variant(all|open|closed):raw(-raw)?/:project+',
+    pattern: ':variant(all|all-raw|open|open-raw|closed|closed-raw)/:project+',
```
chris48s, line 28: On this badge and the merge requests one, I've refactored this to be a more explicit enum. This does introduce a bit of extra duplication here, but it makes it easier to document in the frontend. It is a slight variant on https://github.com/badges/shields/pull/8967

This does have some knock-on effects to the handle/render methods.

There may be a few other places where we need to do something similar, like the GitHub issue badges.

PyvesB (approved):
Looks good to me 👍🏻

PR #9778, opened 2023-12-03
[Discourse] Update schema keys to use plural form (`topic_count` -> `topics_count`)

### LINKED ISSUE #9776: Discourse Status badge seems to be broken ("invalid response data")

shields.io

### 🐞 Description

Relevant shields.io docs: https://shields.io/badges/discourse-status

The badge for my personal Discourse server badge broke (forum.spinalcordmri.org). As a sanity check, I tried the sample Discourse instance listed in the shields.io docs (https://meta.discourse.org), and it too results in a broken badge.

### 🔗 Link to the badge

https://img.shields.io/discourse/status?server=https%3A%2F%2Fmeta.discourse.org

image

### 💡 Possible Solution

This issue was previously reported at: https://github.com/badges/shields/issues/5817

> I'm not familiar with Discourse, but I believe Shields looks for an API at the following location: https://forum.hacf.fr/site/statistics.json
> 
> See our implementation here:
> 
> https://github.com/badges/shields/blob/71299a81788c95363bbafc4a2c75af960b12d5b9/services/discourse/discourse.service.js#L36
>
> That seems to simply redirect to the website's homepage, which is probably why Shields.io returns `invalid response data`.

Despite the accepted solution for that issue, I believe that this new issue has a different cause, as meta.discourse.org _does_ have a statistics.json file at: https://meta.discourse.org/site/statistics.json (as does my own site: https://forum.spinalcordmri.org/site/statistics.json).

So, perhaps the issue runs deeper than what had previously been reported?

joshuacwnewton:
This PR does a simple find and replace to update the Discourse schema to use the current keys found in https://meta.discourse.org/site/statistics.json, as the older keys now return "Invalid response".

> [!NOTE] 
> I was wondering, though, if the old keys should still be preserved, as to not break backwards compatibility with servers using the old names. 
>
> (I tried finding the source of the change in Discourse's release notes/changelog, but there's surprisingly little information when querying e.g. "statistics.json" or "topic_count"/"topics_count". So, I'm not sure when this change actually took place, and thus it's hard to estimate the likelihood of there being servers out there using the old keys.)
>
> If we do want to support both sets of keys, then I'm not 100% sure what to do, as I've never written JavaScript before. (I tried reading the Joi docs, and got as far as using `.or()` in the schema definition for each pair of keys + using a `try`/`catch` when creating the `metricIntegrations` object, but I wasn't confident enough to commit that solution before discussing it first.)

Fixes #9776.

### DIFF
--- a/services/discourse/discourse.service.js
+++ b/services/discourse/discourse.service.js
@@ -18,10 +18,7 @@ const schemaPlural = Joi.object({
   likes_count: nonNegativeInteger,
 })
 
-const schema = Joi.alternatives(
-  schemaSingular,
-  schemaPlural,
-)
+const schema = Joi.alternatives(schemaSingular, schemaPlural)
 
 const queryParamSchema = Joi.object({
   server: optionalUrl.required(),
@@ -79,11 +76,11 @@ function DiscourseMetricIntegrationFactory({ metricType }) {
     async handle(_routeParams, { server }) {
       const data = await this.fetch({ server })
       // e.g. metricType == 'topic' --> try 'topic_count' then 'topics_count'
-      let stat = data[`${metricType}_count`];
+      let stat = data[`${metricType}_count`]
       if (stat === undefined) {
-        stat = data[`${metricType}s_count`];
+        stat = data[`${metricType}s_count`]
       }
-      return this.constructor.render({ stat });
+      return this.constructor.render({ stat })
     }
   }
 }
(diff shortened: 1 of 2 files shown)


joshuacwnewton:
> I think the way to implement this would be to define 2 schemas, one using the plurals and one using the singulars and then use `Joi.alternatives()` to accept either one or the other.

Thank you for the guidance! I'll give it a shot. :hearts:

joshuacwnewton:
I'm not entirely sure if I've used the correct JS syntax/patterns, so please feel free to correct me if I've done something silly. :sweat_smile:

Issue #4040, opened 2019-09-18
Feature Request: Release with Gitea Instances


6543:
Gitea Release (latest by date): | ![](https://img.shields.io/badge/release-v4.16.4-blue)  | /gitea/:instance/v/release/:user/:repo

-> https://github.com/go-gitea/gitea

calebcartwright:
Hi @6543! Can you provide some more detail on the intent of this issue? Are you requesting a new badge be supported in Shields, to provide a Release Version badge for Gitea?

6543:
> Hi @6543! Can you provide some more detail on the intent of this issue? Are you requesting a new badge be supported in Shields, to provide a Release Version badge for Gitea?

Hi @calebcartwright jes, I would realy like souch a feature!

calebcartwright:
Thanks! Could you also provide info for the following questions/items:

- Is there a public API?
- Does the API require an API key?
- Link to the API documentation.

6543:
as fare as I understand they use swagger to show the current API:

> https://gitea.com/api/swagger

to get the release swagger tells me:

> curl -X GET "https://$instance/api/v1/repos/$user/$repo/releases" -H "accept: application/json"

you can test it on my instance for example:
` curl -X GET "https://code.obermui.de/api/v1/repos/6543/remaster/releases" -H "accept: application/json"`

6543:
I found a solution, but i think gitea is worth adding to the official api. Or?

Solution '''[![Release](https://img.shields.io/badge/dynamic/json.svg?label=release&url=https://code.obermui.de/api/v1/repos/6543/remaster/releases&query=$[0].tag_name)](https://code.obermui.de/6543/remaster/releases)'''

calebcartwright:
> but i think gitea is worth adding to the official api

I agree. It makes it easier for others to leverage the badge without having to include those long queries

kitsunegeisha:
need this

Adito5393:
The above solution indeed works if the repo is public, for private access, the URL requires token=.. in the query.

How difficult it would be to add gitea_token to the private variable lists? 
Similar to how drone_token works.

chris48s:
When you're using the dynamic badge you should be able to submit any params you want in the `url` param but you need to urlencode them so the URL above is better thought of as

[...]

PR #9781, opened 2023-12-04
[GITEA] add new gitea service (release/languages)

CanisHelix:
This adds an initial Gitea Service with badges for Releases and Language counts. https://codeberg.org/ (A Gitea Compliant website, albeit using a fork of Gitea) is used for most tests with a dedicated repository for testing again for controlled tests

Private Repository support added and the service document updated.

Uses the new openAPI method instead of old examples.

This would contribute towards issue: #4040, with the aim to add more badges/endpoints overtime.


services/gitea/gitea-base.js
```diff
@@ -0,0 +1,32 @@
...
+      this.authHelper.withBearerAuthHeader({
+        schema,
+        url,
+        options,
+        httpErrors,
+      }),
+    )
+  }
+
+  async fetchPage({ page, requestParams, schema }) {
```
chris48s, line 20: This method does not appear to be used anywhere

services/gitea/gitea-base.js
```diff
@@ -0,0 +1,32 @@
+import { BaseJsonService } from '../index.js'
+
+export default class GiteaBase extends BaseJsonService {
+  static auth = {
+    passKey: 'gitea_token',
+    serviceKey: 'gitea',
+  }
+
+  async fetch({ url, options, schema, httpErrors }) {
```
chris48s, line 9: As discussed on Discord, I'd like to have some tests asserting that auth is being passed through correctly by this method before we merge this.

services/gitea/gitea-helper.js
```diff
@@ -0,0 +1,13 @@
+const documentation = `
+Note that only internet-accessible Gitea compatible instances are supported, for example
+[https://codeberg.org](https://codeberg.org).
+`
+
+function httpErrorsFor(notFoundMessage = 'user or repo not found') {
+  return {
+    401: notFoundMessage,
```
chris48s, line 8: I think on GitLab we had to classify 401 as a "not found" because of the way GitLab's API handles private resources. Not sure that transfers here though. When would this API return a 401?

services/gitea/gitea-languages-count.service.js
```diff
@@ -0,0 +1,79 @@
...
+    return super.fetch({
+      schema,
+      url: `${baseUrl}/api/v1/repos/${user}/${repo}/languages`,
+      httpErrors: httpErrorsFor('user or repo not found'),
+    })
+  }
+
+  async handle(
+    { user, repo },
+    { gitea_url: baseUrl = 'https://codeberg.org' },
```
chris48s, line 70: Defaulting GitLab badges to gitlab.com makes sense because it is the "official" GitLab instance
Is codeberg the "canonical" implementation of gitea, or is it just the most popular public install?
Feels like the default for this should either be GiTea Cloud, or if it is kind of ambiguous, we should just make `gitea_url` a required query param with no default.

chris48s (commented):
Hi thanks for working on this.

I've done a first pass of review. In general the implementations look pretty sensible. We seem to be sticking pretty close to GitLab as a source of inspiration. I also agree that starting off with just a couple of badges and then iterating from there to add more in follow-up PRs is an ideal way to approach this :+1: 

I've done a first pass of review and left a few comments. This one might take a few iterations but this is off to a good start.

services/gitea/gitea-base.js
CanisHelix: Good catch, I missed adding pagination to the API.

services/gitea/gitea-languages-count.service.js
CanisHelix: As discussed, `gitea_url` will be required as there is no canonical implementation, but tests can use codeberg.com still.

services/gitea/gitea-helper.js
CanisHelix: 403 is returned by the API, have updated appropriately.

services/gitea/gitea-base.js
CanisHelix: I have attempted to add a test following the Stack Exchange example (current, prior to #9681). But am having difficult in getting it to even run in the basic state.

```
1) GiteaBase
       auth
         sends the auth information as configured:
     TypeError: Cannot read properties of undefined (reading 'services')
      at DummyGiteaService.invoke (file:///C:/Users/CanisHelix/IdeaProjects/Shields/core/base-service/base.js:430:36)
      at Context.<anonymous> (file:///C:/Users/CanisHelix/IdeaProjects/Shields/services/gitea/gitea-base.spec.js:32:33)
      at process.processImmediate (node:internal/timers:476:21)
```

I may need help on this auth test.

chris48s: Yep that's fine. As long as the test fails if we change the fake token, that's working as expected

services/gitea/gitea-languages-count.service.js
```diff
@@ -0,0 +1,80 @@
...
+import GiteaBase from './gitea-base.js'
+
+/*
+We're expecting a response like { "Python": 39624, "Shell": 104 }
+The keys could be anything and {} is a valid response (e.g: for an empty repo)
+*/
+const schema = Joi.object().pattern(/./, nonNegativeInteger)
+
+const queryParamSchema = Joi.object({
+  gitea_url: optionalUrl,
```
chris48s, line 15: Should be required

services/gitea/gitea-release.service.js
```diff
@@ -0,0 +1,148 @@
...
+    prerelease: Joi.boolean().required(),
+  }),
+)
+
+const sortEnum = ['date', 'semver']
+const displayNameEnum = ['tag', 'release']
+const dateOrderByEnum = ['created_at', 'published_at']
+
+const queryParamSchema = Joi.object({
+  gitea_url: optionalUrl,
```
chris48s, line 21: Should be required

services/gitea/gitea-languages-count.service.js
```diff
@@ -0,0 +1,80 @@
...
+    return super.fetch({
+      schema,
+      url: `${baseUrl}/api/v1/repos/${user}/${repo}/languages`,
+      httpErrors: httpErrorsFor('user or repo not found'),
+    })
+  }
+
+  async handle(
+    { user, repo },
+    { gitea_url: baseUrl = 'https://codeberg.org' },
```
chris48s, line 71: Should not have a default

services/gitea/gitea-release.service.js
```diff
@@ -0,0 +1,148 @@
...
+      }
+    }
+
+    return releases[0][displayKey]
+  }
+
+  async handle(
+    { user, repo },
+    {
+      gitea_url: baseUrl = 'https://codeberg.org',
```
chris48s, line 126: Should not have a default

services/gitea/gitea-base.js
```diff
@@ -0,0 +1,90 @@
...
+
+    const {
+      res: { headers },
+      data,
+    } = await this.fetchPage({ page: 1, requestParams, schema })
+    const numberOfItems = headers['x-total-count']
+
+    let numberOfPages = 1
+    if (numberOfItems > 0) {
+      numberOfPages = Math.ceil(numberOfItems / requestLimit)
```
chris48s, line 75: Hmm. I have not had a really detailed look at this pagination code, but I'm not sure this is going to work.
If we look at
https://docs.gitea.com/next/development/api-usage
(which I think is relevant)

> `MAX_RESPONSE_ITEMS` is set to 50

says to me that

a) The default max allowable value of `requestLimit` for a gitea server is 50
b) This is a number that is going to be different for different gitea servers, so we can't safely assume its value

so I think trying to assume we can work out the number of pages up-front based on our hard-coded value of `requestLimit` and the total number of items feels like it is logic based on a bad assumption.

Tbh, the way we do this for the GitHub badges is we just request the first page of releases ordered by date, and if you want the latest semver then the 'search space' for that is the first page of results, mostly for performance reasons. We've yet to have anyone pop up whose latest release by semver isn't in the first page of results.

Tbh, I'd be happy with that solution here too. When I flagged that `fetchPage` was unused code, I assumed we'd probably just delete it.

Given we have to deal with multiple different implementations, if we do want to try and implement pagination like this, we'd need to infer the page size rather than assuming it up-front, but as I say, I'd also be happy to just gloss over it if we can.

CanisHelix: We have MAX_RESPONSE_ITEMS for the absolute maximum, and DEFAULT_PAGING_NUM for the default pagination value which can be different per every instance. We do get a header for `x-total-count` for how many items are available, and if this is greater than the number of items in the array we could calculate the pages.

When I found the pagination was not working in Gitlab's code too I thought to myself... who even has more than 100 releases to even scan through to find the latest SemVer! Happy to put this back and just assume the first page too.

chris48s:
Cool. I've done a bit of testing. Latest changes LGTM.

One final detail.

For both badges, the example we're using is https://codeberg.org/go-gitea/gitea but that isn't an actual project. Can we pick a repo that exists for the example please.

CanisHelix:
I've updated the examples to point to https://codeberg.org/forgejo/forgejo/ as that's a guaranteed stable repository at codeborg.org. If ever Gitea do a canonical host we can update theirs later.

chris48s (approved):

services/gitea/gitea-release.service.js
```diff
@@ -0,0 +1,147 @@
...
+            schema: { type: 'string', enum: sortEnum },
+            example: 'semver',
+          }),
+          queryParam({
+            name: 'display_name',
+            schema: { type: 'string', enum: displayNameEnum },
+            example: 'release',
+          }),
+          queryParam({
+            name: 'date_order_by',
```
PyvesB, line 78: @CanisHelix late to the game on this one, but I happened to stumble upon this file. It doesn't look this query parameter or the corresponding `orderBy` parameter lower down is used, am I missing something?

Issue #9777, opened 2023-12-03
Badge request: The total number of downloads of npm packages published by the specified user


dukeluo:
### 📋 Description

This badge would show the total number of downloads of npm packages published by the specified user.

### 🔗 Data

- https://npm-stat.com/api/download-counts?author=dukeluo&from=2022-12-02&until=2023-12-02

This API can provide download data of all npm packages for a specified user and does not require an API key.

### 🎤 Motivation

NPM authors could put this badge on their profile.

I have already created an endpoint for it at https://github.com/dukeluo/github-readme-stats-badges. I believe that if the Shields.io can support this endpoint directly, it will be beneficial to more people. I would be willing to submit a PR if you also agree with this idea.

dukeluo:
@chris48s Got it, thanks for pointing that out.

PR #9783, opened 2023-12-05
Add [NpmStatDownloads] Badge

dukeluo:
Add the npm downloads (by author) badge discussed in https://github.com/badges/shields/issues/9777

### DIFF
--- a/services/npm-stat/npm-stat-downloads.service.js
+++ b/services/npm-stat/npm-stat-downloads.service.js
@@ -1,0 +1,83 @@
+import Joi from 'joi'
+import { nonNegativeInteger } from '../validators.js'
+import { BaseJsonService } from '../index.js'
+import { renderDownloadsBadge } from '../downloads.js'
+
+const schema = Joi.object()
+  .pattern(Joi.string(), Joi.object().pattern(Joi.string(), nonNegativeInteger))
+  .required()
+
+const NPM_INITIAL_RELEASE_DATE = '2010-01-12'
+
+const MILLISECONDS_IN_A_DAY = 1000 * 60 * 60 * 24
+const MILLISECONDS_IN_A_WEEK = MILLISECONDS_IN_A_DAY * 7
+const MILLISECONDS_IN_A_MONTH = MILLISECONDS_IN_A_DAY * 30
+const MILLISECONDS_IN_A_YEAR = MILLISECONDS_IN_A_DAY * 365
+
+const intervalMap = {
+  dw: { interval: 'week', difference: MILLISECONDS_IN_A_WEEK },
+  dm: { interval: 'month', difference: MILLISECONDS_IN_A_MONTH },
+  dy: { interval: 'year', difference: MILLISECONDS_IN_A_YEAR },
+  dt: { difference: -1 },
+}
+
+export default class NpmStatDownloads extends BaseJsonService {
+  static category = 'downloads'
+
+  static route = {
+    base: 'npm-stat',
+    pattern: ':interval(dw|dm|dy|dt)/:author',
+  }
+
+  static examples = [
+    {
+      title: 'npm (by author)',
+      documentation:
+        'The total number of downloads of npm packages published by the specified author.',
+      namedParams: { interval: 'dy', author: 'dukeluo' },
+      staticPreview: this.render({ interval: 'dy', downloadCount: 30000 }),
+      keywords: ['node'],
+    },
+  ]
+
+  static _cacheLength = 21600
+
+  static defaultBadgeData = { label: 'downloads' }
+
+  static getTotalDownloads(data) {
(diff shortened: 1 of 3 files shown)


chris48s:
Thanks for submitting. I'll try and have a look over this at the weekend.

services/npm-stat/npm-stat-downloads.service.js
```diff
@@ -0,0 +1,83 @@
...
+  static route = {
+    base: 'npm-stat',
+    pattern: ':interval(dw|dm|dy|dt)/:author',
+  }
+
+  static examples = [
+    {
+      title: 'npm (by author)',
+      documentation:
+        'The total number of downloads of npm packages published by the specified author.',
```
chris48s, line 36: Lets explain where the data comes from in the description

```suggestion
        'The total number of downloads of npm packages published by the specified author from [npm-stat](https://npm-stat.com).',
```

services/npm-stat/npm-stat-downloads.service.js
```diff
@@ -0,0 +1,83 @@
...
+      staticPreview: this.render({ interval: 'dy', downloadCount: 30000 }),
+      keywords: ['node'],
+    },
+  ]
+
+  static _cacheLength = 21600
+
+  static defaultBadgeData = { label: 'downloads' }
+
+  static getTotalDownloads(data) {
```
chris48s, line 47: I think it would be useful to add a unit test for this function, just to nail down that it is correctly counting in the case where the user has multiple packages. A mistake here wouldn't break the service tests. Could we add a `.spec` file with a test or two for the `getTotalDownloads()` function?

services/npm-stat/npm-stat-downloads.service.js
```diff
@@ -0,0 +1,83 @@
...
+  static render({ interval, downloads }) {
+    return renderDownloadsBadge({
+      downloads,
+      interval: intervalMap[interval].interval,
+      colorOverride: downloads > 0 ? 'brightgreen' : 'red',
+    })
+  }
+
+  async handle({ interval, author }) {
+    const format = date => date.toISOString().split('T')[0]
```
chris48s, line 66: We are doing some quite low-level stuff with date objects and numbers of seconds here. It is not necessarily _wrong_ , but I'd find it much more readable to use a higher level abstraction. We already use dayjs on this project, so we can do:

```js
import dayjs from 'dayjs'

dayjs().subtract(1, 'week').format('YYYY-MM-DD')
dayjs().subtract(1, 'month').format('YYYY-MM-DD')
dayjs().subtract(1, 'year').format('YYYY-MM-DD')

dayjs().format('YYYY-MM-DD') //now 
```

My suggestion would be to use that method to just `substract()` the `interval`. I think that would simplify things a lot.

services/npm-stat/npm-stat-downloads.tester.js
```diff
@@ -0,0 +1,43 @@
...
+    color: 'brightgreen',
+  })
+
+t.create('downloads of unknown npm package author')
+  .get('/dt/npm-api-does-not-have-this-package-author.json')
+  .expectBadge({
+    label: 'downloads',
+    message: '0',
+    color: 'red',
+  })
```
chris48s, line 43: Does this API allow us to differentiate between a user who does exist with 0 downloads and a non-existant user?
If not, we can roll with this.

services/npm-stat/npm-stat-downloads.service.js
dukeluo: Yes, I agree. I will do this.

services/npm-stat/npm-stat-downloads.service.js
dukeluo: Cool. It is a good idea. I will refactor this.

services/npm-stat/npm-stat-downloads.tester.js
dukeluo: I think this API cant differentiate between a user who does exist with 0 downloads and a non-existant user. After my testing, this API always return an empty object with a staus code 200 for a non-existant user.

chris48s:
Thanks. Latest changes look good :+1: 

Having done a bit of testing on a review app, I think we need to restrict remove the `dt` variant.

The day, month and year badges mostly seem to return in acceptable times, but the `dt` (all time) badge takes so long to fetch the data that every request I tried just timed out. Nice idea, but it just doesn't perform well enough to be viable, unfortunately.

dukeluo:
Possibly, this API does not support getting npm author total downlands over such a long period of time very well. I agree with your idea of removing dt support to offer a better experience. I will do this.

chris48s:
latest changes LGTM - thanks :+1:

chris48s (approved):

Issue #9784, opened 2023-12-06
Dynamic XML Badge not working


Washingtonwei:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

I put in the URL pointing to my pom.xml on GitHub and the XPath for the Spring Boot version. The tool didn't return anything.

pom.xml URL: https://raw.githubusercontent.com/Washingtonwei/hogwarts-artifacts-online/main/pom.xml
XPath: 
/project/parent/version
./project/parent/version
.//project/parent/version

Tried all three.

Thanks!

### 🔗 Link to the badge

https://img.shields.io/badge/dynamic/xml?url=https%3A%2F%2Fraw.githubusercontent.com%2FWashingtonwei%2Fhogwarts-artifacts-online%2Fmain%2Fpom.xml&query=.%2Fproject%2FmodelVersion

chris48s:
I think this is a duplicate of https://github.com/badges/shields/issues/4250

A working badge URL that achieves what you are trying to do would be

https://img.shields.io/badge/dynamic/xml?url=https%3A%2F%2Fraw.githubusercontent.com%2FWashingtonwei%2Fhogwarts-artifacts-online%2Fmain%2Fpom.xml&query=%2F*%5Blocal-name()%3D'project'%5D%2F*%5Blocal-name()%3D'parent'%5D%2F*%5Blocal-name()%3D'version'%5D

Washingtonwei:
Thanks! It would be great if this can be put in the document.

chris48s:
Yep reasonable point. I've submitted https://github.com/badges/shields/pull/9798 which makes some improvements to the documentation for this badge.

PR #9798, opened 2023-12-10
improve documentation for [dynamicxml] service

chris48s:
This is a follow up from https://github.com/badges/shields/issues/9784

I've attempted to document the point raised there, although I found it slightly tricky to explain it in a completely general way without referencing an example.

I also made a couple of other minor improvements while I was there.

### DIFF
--- a/services/dynamic/dynamic-xml.service.js
+++ b/services/dynamic/dynamic-xml.service.js
@@ -10,6 +10,20 @@ import {
 } from '../index.js'
 import { createRoute } from './dynamic-helpers.js'
 
+const description = `
+The Dynamic XML Badge allows you to extract an arbitrary value from any
+XML Document using an XPath selector and show it on a badge.
+
+Useful resources for constructing XPath selectors:
+- [XPather](http://xpather.com/)
+- [XPath Cheat Sheet](https://devhints.io/xpath/)
+
+Note: For XML documents that use a default namespace prefix, you will need to use the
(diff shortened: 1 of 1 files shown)


services/dynamic/dynamic-xml.service.js
```diff
@@ -10,6 +10,20 @@ import {
...
+XML Document using an XPath selector and show it on a badge.
+
+Useful resources for constructing XPath selectors:
+- [XPather](http://xpather.com/)
+- [XPath Cheat Sheet](https://devhints.io/xpath/)
+
+Note: For XML documents that use a default namespace prefix, you will need to use the
+[local-name](https://developer.mozilla.org/en-US/docs/Web/XPath/Functions/local-name)
+function to construct your query.
+For example \`/*[local-name()='myelement']\` rather than \`/myelement\`.
```
chris48s, line 24: @Washingtonwei - a question for you here: If this had been in the documentation on the website, do you think this would have been enough to allow you to work out https://github.com/badges/shields/issues/9784 without raising an issue?

If not, can you suggest any improvements that don't relate to a specific XML document?

calebcartwright (approved):
Looks like a positive improvement to me, but will hold off adding to the queue in case there's any additional suggestions

services/dynamic/dynamic-xml.service.js
Washingtonwei: This is perfect! Thank you!

PR #9799, opened 2023-12-10
deploy on node 20

### LINKED ISSUE #9780: Deploy on Node 20
:clipboard: **Description**

See https://github.com/badges/shields/pull/9775#issuecomment-1838582070

Reference PR for the previous upgrade: https://github.com/badges/shields/pull/9385

chris48s:
TODO: update branch protection rules when we merge

- package lib tests and setup action have changed due to engine-strict true/false and needing to use npm 9 for the older versions
- remove main-20, integration-20, services-20
- add main-21, integration-21, services-21

Closes #9780


PyvesB (approved):
Double-checked, I don't think you've missed anything :)

chris48s:
have updated the branch protection rules so we'll need to remember to update PRs opened before this so they run the right tests

PR #9800, opened 2023-12-10
Add [Hangar] Badges

spartacus04:
Adds the Hangar dowloads, views, stars and watchers badges


chris48s:
Hi. I haven't looked over the code in any detail yet. Just having a quick look at the API docs, the first thing that jumps out at me is

From https://hangar.papermc.io/api-docs

> The default rate limit is set at 20 requests every 5 seconds with an initial overdraft for extra leniency. Individual endpoints, such as version creation, may have stricter rate limiting.

I feel like it could be tricky to provide reliable badges for this service. While that rate limit works out to 240 requests/minute (which is actually quite generous - for context, that's more than we get from say DockerHub or OpenCollective, which are both popular services) because the window is so short (5 seconds) it wouldn't take much of a "spike" to hit that rate limits and serve broken badges if this service were to gain any sort of traction or userbase.

I'm reticent to move this one forward given that constraint.

PyvesB:
As I understand it, Hangar is a marketplace for Minecraft plugins. As such, many developers would be using that platform to share their projects. A lot of Minecraft plugin projects have a corresponding GitHub repo, and I can foresee developers leveraging badges in their READMEs, in other words the most common Shields.io use case. We've already got badges for a similar Minecraft marketplace, the Spiget badges, which are widely present on GitHub. I see Hangar badges as fitting the "primarily/widely used by developers" Shields.io use case.

I'd expect the Hangar badges to be less popular than the Spiget ones, at least short to medium term.

Adel113 (approved):

chris48s:
Thanks. Spiget is a useful comparison point :bulb: I'm going to keep an eye on the stats for spiget over the next few days to get a vibe for the traffic levels before I come back to this.

chris48s:
OK. I've been keeping an eye on spiget every so often for the last few days. It seems like in a busy hour we're serving maybe ~300 spiget badges from source (i.e: excluding CDN). In the last hour, we just served 254.

So if we imagine hangar became as popular as spiget (which seems like a reasonable-ish comparison. Obviously it isn't going to be as widely used as NPM or PyPI) and the traffic were completely evenly distributed across an hour, we wouldn't even be close to exceeding the rate limit. That short window of only 5 seconds does worry me a bit as 20 requests in 5 seconds isn't much of a "spike". This is basically the problem we have with OpenCollective. If you average our traffic out over an hour we're pretty much consistently under, but because their measurement window is only 1 minute we quite often hit it and serve broken badges.

turikhay:
+1 for Hangar.

It is a good alternative to Spigot (Spiget) and CurseForge. It is developed by the team behind Paper, the fork of Spigot that is now more popular and widespread than its predecessor.

chris48s:
I kinda forgot about this one. I think I'm inclined to go ahead with this. All the badges we're looking at adding are popularity metrics, so I think we can get away with quite a long cache if rate limits do become an issue down the line. I'll leave a first round of review comments.

services/hangar/hangar-downloads.service.js
```diff
@@ -0,0 +1,34 @@
+import { pathParams } from '../index.js'
+import { renderDownloadsBadge } from '../downloads.js'
+import { BaseHangarService, description } from './hangar-base.js'
+
+export default class HangarDownloads extends BaseHangarService {
+  static category = 'downloads'
+
+  static route = {
+    base: 'hangar/downloads',
```
chris48s, line 9: If this is total downloads (all time), URL should be `/dt` - see https://github.com/badges/shields/blob/master/doc/badge-urls.md

services/hangar/hangar-stars.service.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  static defaultBadgeData = {
+    label: 'stars',
+    color: 'blue',
+  }
+
+  static render({ stars }) {
+    return {
+      message: `${stars}`,
```
chris48s, line 32: Stars, views and watchers badges should all render with a call to `metric()`

chris48s:
Given we've got rate limits to stay under, lets put a longer than usual max-age on these from day one. Given these are all popularity measures (as opposed to something like "is the build passing/failing?"), I'm going to suggest we start off with an hour, so `static _cacheLength = 3600` on all of these.

Also, given this PR has been sitting around for a while. Can you merge in lastest master when you next push more commits. Thanks

services/hangar/hangar-base.js
```diff
@@ -0,0 +1,37 @@
+import Joi from 'joi'
+import { BaseJsonService } from '../index.js'
+
+const description = `
+<p><a href="https://hangar.minecraft.tools/">Hangar</a> is a plugin repository for the Paper, Waterfall and Folia platforms.</p>`
```
turikhay, line 5: What's this host btw?

spartacus04: Which one?

turikhay: `hangar.minecraft.tools`

spartacus04: Oh, I don't even know how that ended up in there, fixing right now

chris48s (approved):
latest changes LGTM

xMrAfonso:
Hey there, just wanted to add a note that the rate limited stated is not actually 100% accurate. You can go over it slightly.

PR #9801, opened 2023-12-11
Add [Raycast] Badge

Fatpandac:
image

### DIFF
--- a/services/raycast/installs.service.js
+++ b/services/raycast/installs.service.js
@@ -31,7 +31,7 @@ export default class RaycastInstalls extends BaseJsonService {
 
   static render({ user, extension, downloadCount }) {
     return {
-      label: 'installs',
+      label: 'Installs',
       message: metric(downloadCount),
       color: 'green',
       link: [`https://www.raycast.com/${user}/${extension}`],
--- a/services/raycast/installs.service.js
+++ b/services/raycast/installs.service.js
@@ -1,0 +1,68 @@
+import Joi from 'joi'
+import { metric } from '../text-formatters.js'
+import { anyInteger } from '../validators.js'
+import { BaseJsonService, NotFound, pathParams } from '../index.js'
+
+const schema = Joi.object({
+  download_count: anyInteger,
+}).required()
+
+export default class RaycastInstalls extends BaseJsonService {
+  static category = 'platform-support'
+
+  static route = {
+    base: 'raycast/installs',
+    pattern: ':user/:extension',
+  }
+
+  static openApi = {
+    '/raycast/installs/{user}/{extension}': {
+      get: {
+        summary: 'Raycast extension downloads count',
+        parameters: pathParams(
+          { name: 'user', example: 'Fatpandac' },
+          { name: 'extension', example: 'bilibili' },
+        ),
+      },
+    },
+  }
+
+  static _cacheLength = 7200
+
+  static render({ user, extension, downloadCount }) {
+    return {
+      label: 'installs',
+      message: metric(downloadCount),
+      color: 'green',
+      link: [`https://www.raycast.com/${user}/${extension}`],
+    }
+  }
+
+  async fetch({ user, extension }) {
+    return this._requestJson({
+      schema,
+      url: `https://www.raycast.com/frontend_api/users/${user}/extensions/${extension}`,
+      httpErrors: {
+        404: 'user/extension not found',
+      },
+    })
+  }
+
+  transform(json) {
+    const downloadCount = json.download_count
+    if (downloadCount === undefined) {
+      throw new NotFound({ prettyMessage: 'download_count not found' })
+    }
+    return { downloadCount }
+  }
+
+  async handle({ user, extension }) {
+    const json = await this.fetch({ user, extension })
+    const { downloadCount } = this.transform(json)
+    return this.constructor.render({
+      user,
+      extension,
+      downloadCount,
+    })
+  }
+}
--- a/services/raycast/installs.service.js
+++ b/services/raycast/installs.service.js
@@ -41,7 +41,7 @@ export default class RaycastInstalls extends BaseJsonService {
   async fetch({ user, extension }) {
     return this._requestJson({
       schema,
-      url: `https://www.raycast.com/frontend_api/users/${user}/extensions/${extension}`,
+      url: `https://www.raycast.com/api/v1/extensions/${user}/${extension}`,
       httpErrors: {
         404: 'user/extension not found',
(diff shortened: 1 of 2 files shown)


services/raycast/installs.service.js
```diff
@@ -0,0 +1,68 @@
...
+      message: metric(downloadCount),
+      color: 'green',
+      link: [`https://www.raycast.com/${user}/${extension}`],
+    }
+  }
+
+  async fetch({ user, extension }) {
+    return this._requestJson({
+      schema,
+      url: `https://www.raycast.com/frontend_api/users/${user}/extensions/${extension}`,
```
chris48s, line 44: Hi. I haven't had a proper look at this code, but this looks like an internal API endpoint, rather than a public one.

Can you link to any documentation for this?

Fatpandac: Yes, this is not an open API. I found it on the [website](https://www.raycast.com/Fatpandac/bilibili) of Raycast and it does not have any documentation.

chris48s: OK thanks. The reason we say

> Badges should not obtain data from undocumented or reverse-engineered API endpoints.

in [our guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines)

is because we had some situations where a contributor worked out that a project had an API by inspecting the frontend code for their website or whatever, added a badge that calls it, and then we ended up relying on an API that wasn't really intended for public use. This then either annoyed the maintainers of the company/project behind it or we had no reasonable expectation of stability/compatibility over time (because devs maintaining the the API were making the assumption the only thing consuming it was their own frontend code) and stuff kept breaking.

That's looks like the situation we've got here so I am going to close this PR, but we could revisit in future if there is a reliable public source for the data. Other than the data source, it looks like a badge showing downloads for a raycast extension would be a reasonable badge to add.

RSO: 👋 I'm a Software Engineer @ Raycast. We don't have public API documentation available, but if you use `https://www.raycast.com/api/v1/extensions/petr/slack-status` instead of the `frontend_api` path, we're good with publishing the badge. We'll make sure the endpoint keeps being supported.

Fatpandac: Thanks for your reply 🙏, I will change to use this API soon.

services/raycast/installs.service.js
```diff
@@ -0,0 +1,68 @@
...
+
+const schema = Joi.object({
+  download_count: anyInteger,
+}).required()
+
+export default class RaycastInstalls extends BaseJsonService {
+  static category = 'platform-support'
+
+  static route = {
+    base: 'raycast/installs',
```
chris48s, line 14: We seem to be mixing and matching a bit on "downloads" and "installs". The number we're getting from the API is `download_count`. Given that, I think we should frame this as a "downloads" badge (rather than "installs") and make the route badge `raycast/dt` in line with the conventions: https://github.com/badges/shields/blob/master/doc/badge-urls.md

services/raycast/installs.service.js
```diff
@@ -0,0 +1,68 @@
...
+        summary: 'Raycast extension downloads count',
+        parameters: pathParams(
+          { name: 'user', example: 'Fatpandac' },
+          { name: 'extension', example: 'bilibili' },
+        ),
+      },
+    },
+  }
+
+  static _cacheLength = 7200
```
chris48s, line 30: Any specific reason for picking 2 hours here?

services/raycast/installs.service.js
```diff
@@ -0,0 +1,68 @@
...
+
+  static render({ user, extension, downloadCount }) {
```
chris48s, line 32: Can we use the `renderDownloadsBadge()` helper please. https://github.com/badges/shields/blob/master/services/downloads.js That will apply our standard formatting and things

services/raycast/installs.service.js
```diff
@@ -0,0 +1,68 @@
...
+    return {
+      label: 'Installs',
```
chris48s, line 34: Label should be all lowercase please. See https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines

services/raycast/installs.service.js
```diff
@@ -0,0 +1,68 @@
...
+      message: metric(downloadCount),
+      color: 'green',
+      link: [`https://www.raycast.com/${user}/${extension}`],
```
chris48s, line 37: Non-social badges shouldn't use `link` by default

services/raycast/installs.service.js
```diff
@@ -0,0 +1,68 @@
...
+      url: `https://www.raycast.com/api/v1/extensions/${user}/${extension}`,
+      httpErrors: {
+        404: 'user/extension not found',
+      },
+    })
+  }
+
+  transform(json) {
+    const downloadCount = json.download_count
+    if (downloadCount === undefined) {
```
chris48s, line 53: We can remove this check for `if (downloadCount === undefined)`. Applying the schema will reject the response if a `.required()` property is not present so we don't need to check it again.

services/raycast/installs.tester.js
```diff
@@ -0,0 +1,29 @@
...
+t.create('installs (not existing user and extension)')
+  .get('/fatpandac/safdsaklfhe.json')
+  .expectBadge({
+    label: 'platform-support',
+    message: 'user/extension not found',
+  })
+
+t.create('installs (valid)').get('/Fatpandac/bilibili.json').expectBadge({
+  label: 'Installs',
+  message: anyInteger,
```
chris48s, line 28: Can we use `isMetric` for the test assertion here.

chris48s (commented):
Thanks for updating. I've left some comments. There's quite a few points to follow up but they should all be quite quick and easy to address.

services/raycast/installs.service.js
```diff
@@ -0,0 +1,68 @@
+import Joi from 'joi'
+import { metric } from '../text-formatters.js'
+import { anyInteger } from '../validators.js'
+import { BaseJsonService, NotFound, pathParams } from '../index.js'
+
+const schema = Joi.object({
+  download_count: anyInteger,
```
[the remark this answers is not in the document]
Fatpandac: OK

services/raycast/installs.service.js
Fatpandac: There’s no reason; I just forgot to remove it. 🙇

chris48s (approved):
latest changes lgtm :+1:

PR #9802, opened 2023-12-11
[pub] Use official version endpoint for pub-service

sigurdm:
* pub.dartlang.org is replaced by pub.dev as the official pub site.
* `api/packages/<packageName>` is the official endpoint for retrieving a version listing.
  see https://pub.dev/help/api#hosted-pub-repository-api -> https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#list-all-versions-of-a-package

The `packages/<packageName>.json` endpoint is obsolete and might go away.

### DIFF
--- a/services/pub/pub.service.js
+++ b/services/pub/pub.service.js
@@ -5,7 +5,7 @@ import { baseDescription } from './pub-common.js'
 
 const schema = Joi.object({
   versions: Joi.array()
-    .items(Joi.object({ version: Joi.string() }))
+    .items(Joi.object({ version: Joi.string().required() }))
     .required(),
 }).required()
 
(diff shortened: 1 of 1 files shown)


chris48s:
Thanks. The reason the tests are failing is because we also need to update the schema to account for the new response format:

https://github.com/badges/shields/blob/78f83249fa446d3a1cde907a772eacf3a84ebbdf/services/pub/pub.service.js#L6-L8

At the moment we're rejecting the API responses because they don't match the schema. Could you update?

sigurdm:
> At the moment we're rejecting the API responses because they don't match the schema. Could you update?

Ah of course

Done!

chris48s (approved):

chris48s:
I pushed one last edit as it was easier than another round of back-and-forth.
Merging - thanks :+1:

Issue #9342, opened 2023-07-01
Docker badges rate limited by upstream service


kroese:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

All docker shields show:

"Rate limited by upstream service"

at the moment.

chris48s:
This is new.

DockerHub are not super explicit about their rate limits. From https://docs.docker.com/docker-hub/download-rate-limit/#other-limits :

> Docker Hub also has an overall rate limit to protect the application and infrastructure. This limit applies to all requests to Hub properties including web pages, APIs, image pulls. The limit is applied per-IP, and while the limit changes over time depending on load and other factors, it’s in the order of thousands of requests per minute. The overall rate limit applies to all users equally regardless of account level.
> 
> You can differentiate between these limits by looking at the error code. The “overall limit” returns a simple `429 Too Many Requests` response. The pull limit returns a longer error message that includes a link to this page.

Clearly we are hitting their limit. Also there is not much we can do to increase out rate limit. For example, we can't get a higher rate limit by authenticating with a token.

I've opened PR https://github.com/badges/shields/pull/9343 to see if we can solve this by cacheing badges for longer to reduce our usage

chris48s:
Since deploying https://github.com/badges/shields/pull/9343 earlier today, docker badges seem to be rendering again. I'm going to close this but if we start seeing problem again (we usually see higher traffic on weekdays than we do at the weekends), re-open the issue.

chris48s:
I'm seeing working badges at the moment, but lets continue to monitor this

chris48s:
"a couple of hours ago" from that post was before we merged/deployed https://github.com/badges/shields/pull/9350 Since we deployed that, every time I have checked a docker badge it has been working. Are you still seeing failures over the last few days?

goose-ws:
Sorry to comment on a closed topic, but I'm still seeing rate limited messages, I'm not so sure it's solved

chris48s:
One thing this has inspired me to do is look and see if there are `x-ratelimit-` headers on the responses from the DockerHub API. There are. Previously I had only looked at https://docs.docker.com/docker-hub/download-rate-limit/#other-limits

Right now we seem to be getting a limit of 180 reqs/minute.

[...]

PR #9803, opened 2023-12-11
call [docker] with auth

chris48s:
Closes https://github.com/badges/shields/issues/9342

Just to summarise the reason why I am doing this:

- DockerHub imposes a rate limit of 180 reqs/minute for anonymous usage
- We increased the max-age on DockerHub badges to try and stay under it. We are now up to 4 hours
- Even caching the badges for 4 hours, we are still sometimes hitting the limits
- By using auth, we can increase our rate limit to 600 reqs/minute, which would be a big help
- DockerHub's auth pattern is a bit complicated and none of our existing auth helpers cover it
- Docker is one of our most popular services

So.. the point of this PR is:

- Add a JWT auth helper
- Modify the docker badges to use it
  so we can..
- Set DockerHub tokens in production

I have attempted to do this in a relatively general way, but at the moment we only have one service that uses this type of auth so there might be assumptions that turn out to be docker-specific. We might find if we have another service that uses this pattern, we need to modify it a bit.

Still TODO:

- [x] ~Will this also work for https://cloud.docker.com ? Does it matter?~
- [x] ~Write unit tests for JWT AuthHelper (might do some refactoring to support this)~
- [x] ~Write specs for docker service auth~
- [x] Create and set credentials in production/keybase


core/base-service/auth-helper.js
```diff
@@ -1,5 +1,13 @@
...
-import { InvalidParameter } from './errors.js'
+import dayjs from 'dayjs'
+import Joi from 'joi'
+import checkErrorResponse from './check-error-response.js'
+import { InvalidParameter, InvalidResponse } from './errors.js'
+import { fetch } from './got.js'
+import { parseJson } from './json.js'
+import validate from './validate.js'
+
+let jwtCache = Object.create(null)
```
chris48s, line 10: If we don't already have a JWT, we need to obtain one. That is one round-trip on the network. Then once we've got one, we can use it to make the API call we actually want to make to get some data to put on a badge, which is a second round-trip on the network.

In order to avoid making 2 requests for every single docker badge, I have implemented a rudimentary cache using a module-level object.

After we've done the OAuth dance, DockerHub issues a JWT with an expiry date on it. At the moment, this is 1 month but in principle that could change.

Realistically, it is uncommon for us to run a single VM instance for a full month. We tend to deploy every week or two. This means in practice each VM we boot will make 2 round-trips on the network the first time we process a request for a docker badge. Then all subsequent requests for the lifetime of that VM will be able to use a cached token. If a single VM were to actually run for a month or Docker shorten the expiry on tokens, I implemented proper logic for handling the token expiry anyway.

I could have implemented this so we store a single cached JWT shared between VMs in postgres instead of each VM maintaining its own cache, but I think keeping towards being more stateless is preferable. It also means that if self-hosting users want to use DockerHub auth to allow badges for private repos, they don't have to run a postgres server to store one string. That would be annoying.

services/docker/docker-hub-common-fetch.spec.js
```diff
@@ -0,0 +1,20 @@
...
+    const serviceInstance = {
+      _requestJson: sinon.stub().resolves('fake-response'),
+      authHelper: {
+        withJwtAuth: sinon.stub(),
+      },
+    }
+
+    const resp = await fetch(serviceInstance, {})
+
+    expect(serviceInstance.authHelper.withJwtAuth.calledOnce).to.be.true
```
chris48s, line 16: All I am doing is asserting that the fetch method calls the service class's `withJwtAuth` method, but I think that is all we care about. I don't think more elaborate mocks actually buy us much here.

chris48s:
I still need to make credentials and set them, but the code is now ready for review

PyvesB:
What happens if Docker decides to revoke the token before its expiry? My understanding is that we'll keep on trying to use the token until either we've restarted our server or the initially-advertised expiry date has been reached. If that's indeed the case, this doesn't seem ideal.

I'd personally even be in favour of ignoring the advertised expiry date, and only clearing the cache when we get an auth error. We'd suffer from an extra round trip once in a blue moon for not proactively requesting the new token, but on the other hand we'd save the call to `_isJwtValid` on every badge render.

chris48s:
Yeah you're right. If a token was revoked before its given expiry, we'd just keep trying it. That's an interesting point I hadn't considered.

I guess to account for that we would need to expose a way to allow the service class to invalidate the cached token (and ideally retry the request) if we get a 401 or 403 back from DockerHub (or whatever the code is - TODO: check).

That said, ideally I'd like to do that without leaking too much of the implementation details of the auth into the service class. i.e: if a contributor wants to add a new badge that gets data from dockerhub, that should ideally be abstracted away as much as possible.

Another thing we could do is just set our own much shorter expiry on the token. So we set the expiry to (for example) 1 hour regardless, even if the expiry on the token is 1 month. We'll re-auth more frequently, but then even in the event that a token does get revoked before its expiry (this seems like an exceptional case, rather than something we would expect to see often) it will fix itself after a bit with no intervention from us. That might be a less conceptually correct but more pragmatic way to cover that edge case?

PyvesB:
> That might be a less conceptually correct but more pragmatic way to cover that edge case?

Happy to align with your proposed alternative. :)

PyvesB (approved):
Looks good to me now 👍🏻

chris48s:
We're now running with a token in prod scoped to 'public repo read-only'
Have shared the credentials with the team

Issue #9787, opened 2023-12-06
A plugin architecture for self-hosted shields instances


chris48s:
:clipboard: **Description**

This doesn't come up all the time, but occasionally someone pops up and asks us to add a badge which would only make sense for a self-hosted install.

We usually say no to these requests. If there is no public instance of the service, it is difficult for us to test that this code works (either manually or via integration tests), making it hard to support, maintain or refactor.

That said, this has come up a non-zero number of times and the requests are not completely invalid. For some subset of self-hosting users, these badges could be useful.

So here's an idea: We provide some kind of plugin hook or extension point that allows self-hosting users to write and maintain a custom service. This would allow this category of users to self-serve their own feature requests without them having to maintain a fork or us having to take on the maintenance of code that would be zero use to the users of shields.io.

The thing that has made me think of this recently is Artifactory. I know this issue has come up before, but I can't think of any other examples to hand. So my first question is: Can anyone remember any other cases? It would be useful to collect up a few examples to help think about this. Also it would be useful to get some idea of whether there are enough use-cases to warrant solving the challenges this presents us with.

Here's some initial thoughts about what a plugin system might look like:

- We make a directory called `local-services` (for the sake of argument).
- In the repo, it is just an empty dir with a `.gitkeep` in it.
- When we bootstrap the server, as well as searching the `services` dir, the loader also searches the `local-services` dir.
- If any services are found in there, they are loaded. We mount them under a URL like `/local` to namespace them. That means there won't be clashes with other URLs we add to core in future.

That sounds kinda reasonable-ish, but it does bring up a number of problems, tradeoffs and questions:

- There are probably some people who self-host by running from source. However, my assumption is that most people who run their own instance use a docker image. What is a good pattern to recommend for allowing users to inject additional files into a docker image? I guess one option would be for users to write their own dockerfile extending `shieldsio/shields` and copying some extra files into it. Another might be to mount `local-services` as a volume and give it some files that way? Does anyone know of any examples of other projects that have this kind of setup? It would be interesting to look at some examples.
- As soon as you've got other people writing services, our internal API becomes a public API. For example, if we make a breaking change to `BaseJsonService` at the moment, its a bit of a pain but fundamentally we only have to account for that in our own codebase. Once you have self hosting users extending that class, a non-backwards compatible change has wider implications.
[...]

PR #9805, opened 2023-12-15
[NuGet BaGet]: Added new BaGet service (Self hosted NuGet server)

lluiscab:
This PR adds support for version & download badges from a [BaGet](https://github.com/loic-sharma/BaGet) server.

I'm not sure how to add examples / tests for this new service due to it's self hosted nature and no public instance being available (that I'm aware of)

I've tested badge generation against our local (private) BaGet instance with the following commands:

```sh
npm run badge -- /baget/v/<PackageName?source=<Instance url>
npm run badge -- /baget/vpre/<PackageName>?source=<Instance url>
npm run badge -- /baget/dt/<PackageName?source=<Instance url>
```

I've extended the logic from `nuget-v3-service-family` to support the small differences in API in the BaGet server api implementation, mainly searching without the `packageid` parameter and the package version being available in the response, without need for lookup in the `versions` array (which is also sorted in DESC, not ASC)

### DIFF
--- a/services/baget/baget.service.js
+++ b/services/baget/baget.service.js
@@ -1,0 +1,41 @@
+import { createServiceFamily } from '../nuget/nuget-v3-service-family.js'
+
+const { NugetVersionService: Version, NugetDownloadService: Downloads } =
+  createServiceFamily({
+    defaultLabel: 'baget',
+    serviceBaseUrl: 'baget',
+    withTenant: false,
+    withFeed: false,
+    withQueryNamedParams: false,
+    packageDataIncludesVersion: true,
+  })
+
+class BagetVersionService extends Version {
+  static examples = [
+    {
+      title: 'Baget',
+      pattern: 'v/:packageName',
+      namedParams: { packageName: 'Microsoft.AspNet.Mvc' },
+      staticPreview: this.render({ version: '5.2.4' }),
+    },
+    {
+      title: 'Baget (with prereleases)',
+      pattern: 'vpre/:packageName',
+      namedParams: { packageName: 'Microsoft.AspNet.Mvc' },
+      staticPreview: this.render({ version: '5.2.5-preview1' }),
+    },
+  ]
+}
+
+class BagetDownloadService extends Downloads {
+  static examples = [
+    {
+      title: 'Baget',
+      pattern: 'dt/:packageName',
+      namedParams: { packageName: 'Microsoft.AspNet.Mvc' },
+      staticPreview: this.render({ downloads: 49e6 }),
+    },
+  ]
+}
+
+export { BagetVersionService, BagetDownloadService }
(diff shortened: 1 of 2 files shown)


chris48s:
Sorry it has taken quite a while to respond to this.

Given you say that there is no public instance of BaGet, this seems like a badge that would only be relevant to users who self host a shields instance. Would that be a reasonable assessment?

In general, we don't add these.

Related reading:

https://github.com/badges/shields/blob/b1f5aecf36c716cd642c27b4569d04abef335dee/doc/releases.md?plain=1#L31-L36

and

https://github.com/badges/shields/pull/9666#issuecomment-1838736622

I did start an issue about a possible way to deal with this at https://github.com/badges/shields/issues/9787 but I'm not sure it will go anywhere. I don't think any of us are convinced on the tradeoffs.

All of that said, quite a common thing we do is support compatibility with a private or self-hosted version of a centralised service as long as your private instance is API compatible with the centralised version. So our GitLab badges will work with your self-hosted GitLab as long as it is running a version that is API-compatible with https://gitlab.com/ Our Packagist badges will work with your private packagist install as long as it is running a version that is API-compatible with https://packagist.org/ and so on. Essentially, if we can test against a public instance and also tack on a `?registry_url=` param (or whatever), that works for us.

So here's my question: There are some BaGet-specific customisations going on in this PR, but would it be possible to deliver a subset of what you're trying to do here by bolting an optional parm onto NuGet and documenting that it also works with BaGet, or is BaGet not quite API compatible enough for that to work? i.e: are the changes in this PR BaGet-related _enhancements_ or are they fundamentally necessary to get the thing to work?

lluiscab:
I understand the concerns about maintaining an integration with a provider that can't easily be tested.

Reading the issues that you linked, is this something that should be implemented using the `endpoint` or `dynamic/json` badges then?

chris48s:
The dynamic json badge has the advantage that you can make custom badges just by constructing a URL. So using a URL that returns JSON like https://pypi.org/pypi/pip-abandoned/json I can construct a URL like https://img.shields.io/badge/dynamic/json?query=info.requires_python&label=python&url=https%3A%2F%2Fpypi.org%2Fpypi%2Fpip-abandoned%2Fjson that extracts a value from it and puts it on a badge.

The downside of this is that it can only take you so far. If you want to do something more advanced than just extract a value and display it, for example:

- change the colour of the badge conditionally based on the value
- perform some kind of calculation like adding several numbers together
- additional formatting (e.g: rounding a number, or displaying a number of bytes as Kb/Mb/Gb)

you can't do that with the dynamic badge.

With the endpoint badge, the sky's the limit. You can have all that and a bag of chips, but you do need to host the endpoint that can call your API, do whatever processing you need, and serve a JSON document conforming to the badge schema _somewhere_, and we recognise this does represent some additional overhead.

PyvesB:
I'm personally aligned on above points as well. I don't think we'll want to move forward with this badge for the time being, so will go ahead and close the PR. Let's keep an eye on #9787 as the next step.

Anyhow, thanks for your interest in the Shields.io project, @lluiscab!

PR #9841, opened 2023-12-25
migrate examples to openApi part 27; affects [bstats bugzilla eclipse]

PyvesB:

@chris48s was having all the fun on his own, I simply wanted to have a small slice of the cake as well.

### DIFF
--- a/services/bugzilla/bugzilla.service.js
+++ b/services/bugzilla/bugzilla.service.js
@@ -1,6 +1,6 @@
 import Joi from 'joi'
 import { optionalUrl } from '../validators.js'
-import { BaseJsonService } from '../index.js'
+import { BaseJsonService, pathParam, queryParam } from '../index.js'
 
 const queryParamSchema = Joi.object({
   baseUrl: optionalUrl,
@@ -18,8 +18,9 @@ const schema = Joi.object({
     .required(),
 }).required()
 
-const documentation = `
+const description = `
 <p>
+  Use the <code>baseUrl</code> query parameter to target different Bugzilla deployments.
   If your Bugzilla badge errors, it might be because you are trying to load a private bug.
 </p>
 `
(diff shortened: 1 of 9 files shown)


services/eclipse-marketplace/eclipse-marketplace-downloads.service.js
```diff
@@ -1,69 +1,58 @@
...
-
-const totalResponseSchema = Joi.object({
-  marketplace: Joi.object({
-    node: Joi.object({
       installstotal: nonNegativeInteger,
     }),
   }),
 }).required()
 
-function DownloadsForInterval(downloadsInterval) {
```
PyvesB, line 22: I refactored the downloads badge to use our more common `interval` pattern.

services/eclipse-marketplace/eclipse-marketplace-base.js
```diff
@@ -1,13 +1,6 @@
 import { BaseXmlService } from '../index.js'
 
 export default class EclipseMarketplaceBase extends BaseXmlService {
-  static buildRoute(base) {
-    return {
-      base,
-      pattern: ':name',
```
PyvesB, line 7: The downloads badge now has a different pattern; this helper didn't really bring much value anyway, I decided to get rid of it.

services/bugzilla/bugzilla.service.js
```diff
@@ -18,8 +18,9 @@ const schema = Joi.object({
     .required(),
 }).required()
 
-const documentation = `
+const description = `
 <p>
+  Use the <code>baseUrl</code> query parameter to target different Bugzilla deployments.
```
PyvesB, line 23: Unlike most other badges, `baseUrl` is not a required query parameter. However, it's quite easy to miss when it's not marked as required, so decided to add a small bit of text in the example's description.

Alternatively, I could have marked it as required in the example, even though we'd still fall back underneath the hood if not specified.

Let me know what you think.

chris48s: I don't think we should mark optional parameters as required.

What you've done here is totally reasonable. In general, I have not been doing anything extra to draw attention to optional query params. So if you look at https://shields.io/badges/git-lab-release for example, there are some extra notes around `gitlab_url` but there's nothing extra calling out

- `include_prereleases`
- `sort`
- `display_name` or
- `date_order_by`

basides just documenting them. I do think that there is a distinction between

- optional query params that are specific to this badge (e.g: `include_prereleases`) and
- optional query params that can be used with all badges (e.g: `logo`)

and it would be nice to be able to make that clearer. Neither the OpenAPI spec nor the rendering package really gives us a way to make a distinction between them though.

At the moment, I'm just focussing on trying to plough through the data conversion.

chris48s (approved):
woo - someone else did one :D

PR #9855, opened 2023-12-29
migrate some services from examples to openApi part 28; affects [aur cran mastodon matrix npmstat nyc thunderstore twitter]

chris48s:
Refs #9285

I think all of these are pretty straightforward.


PyvesB (approved):
Looks good. 👍🏻

PR #9856, opened 2023-12-29
migrate some services from examples to openApi part 29; affects [testspace twitch w3c weblate whatpulse]

chris48s:
Refs #9285

How do you eat an elephant? One bite at a time.

### DIFF
--- a/services/w3c/w3c-validation-helper.js
+++ b/services/w3c/w3c-validation-helper.js
@@ -77,72 +77,20 @@ const getSchema = preset => {
   return schema.map(url => encodeURI(url)).join(' ')
 }
 
-const documentation = `
-  <style>
-    .box {
-      display: flex;
-      justify-content: space-between;
-    }
-    .note {
-      font-size: smaller;
-      text-align: left;
-    }
-  </style>
-  <p>
-    The W3C validation badge performs validation of the HTML, SVG, MathML, ITS, RDFa Lite, XHTML documents.
-    The badge uses the type property of each message found in the messages from the validation results to determine to be an error or warning.
-    The rules are as follows:
-    <ul class="note">
-      <li>info:  These messages are counted as warnings</li>
-      <li>error:  These messages are counted as errors</li>
-      <li>non-document-error: These messages are counted as errors</li>
-    </ul>
-  </p>
-  <p>
-    This badge relies on the <a target="_blank" href="https://validator.nu/">https://validator.nu/</a> service to perform the validation.
-    Please refer to <a target="_blank" href="https://about.validator.nu/">https://about.validator.nu/</a> for the full documentation and Terms of service.
-    The following are required from the consumer for the badge to function.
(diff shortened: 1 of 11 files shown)


services/w3c/w3c-validation-helper.js
```diff
@@ -77,72 +77,20 @@ const getSchema = preset => {
   return schema.map(url => encodeURI(url)).join(' ')
 }
 
-const documentation = `
```
chris48s, line 80: It looks like I'm deleting loads of docs here, but I've just moved most of this into parameter descriptions

services/w3c/w3c-validation.service.js
```diff
@@ -38,18 +45,49 @@ export default class W3cValidation extends BaseJsonService {
...
+          }),
+          queryParam({
+            name: 'preset',
+            example: 'HTML, SVG 1.1, MathML 3.0',
+            description:
+              'This is used to determine the schema for the document to be valdiated against.',
+            schema: {
+              type: 'string',
+              enum: [
+                'HTML, SVG 1.1, MathML 3.0',
```
chris48s, line 74: I'm slightly in 2 minds about this one. I made this an enum in the frontend because the docs just listed a bunch of acceptable values so it seems like giving the user a dropdown to pick from seems like a more useful frontend. The validation isn't just a list of allowed values though. It is this monsterous soup of regex 😱 https://github.com/badges/shields/blob/7d7efaec22b644368b9803e5e03441b547b7476b/services/w3c/w3c-validation-helper.js#L1-L12
I'm tempted to just make the validation a list of allowed strings, but I'm slightly unsure how we ended up with this and whether there are really valid values that pass that validation but aren't in that list

PyvesB: Maybe it's related to case sensitivity? https://github.com/badges/shields/pull/4148#discussion_r333752460

PyvesB: To avoid breaking badges for folks who used a value with different casing, we could validate all values in a non-case-sensitive way.

chris48s: Thanks for the link to that thread.

I went to look at changing this and the read the tests (which I should have done before).

https://github.com/badges/shields/blob/880c1fb49cd50ca5a0558075c956c7cab2d7676e/services/w3c/w3c-validation-helper.spec.js#L16-L50

Changing it to an allow list but making it case insensitive should be do-able with `.insensitive()` but having read the tests the other thing we are currently tolerant to is variations in spacing. We wouldn't be able to handle that without some kind of pre-processing to normalize the spacing before validation so I think in order to move to an allow-list, we'd have to make a behaviour change.

Given that, I think I'm inclined to just leave this as-is We show a dropdown in the frontend but leave the existing validation in place for now.

PyvesB (approved):
Looks good to me 👍🏻

PR #9857, opened 2023-12-30
migrate some services from examples to openApi part 30; affects [poeditor pypi symfony teamcity]

chris48s:
Refs #9285


PyvesB (approved):
Another one that looks all good!

PR #9858, opened 2023-12-30
migrate some services from examples to openApi part 31; affects [packagecontrol discourse]

chris48s:

I'm only touching a couple of files in this PR but there is a bit more refactoring going on here so I split them out.

For these two services, we've got a similar pattern to the one we see in https://github.com/badges/shields/pull/9841 and https://github.com/badges/shields/pull/9732 - a function that generates multiple service classes.

In both these cases, I opted to refactor the services to use an enum parameter.


PyvesB (approved):
Great refactors!

PR #9859, opened 2023-12-30
Ensure social category badges are rendered with social style and logo; affects [gitlab keybase lemmy modrinth thunderstore twitch] gist github reddit

### LINKED ISSUE #9704: default style for social badges in documentation
For some badges, we render them using the social style by default. For example:

https://img.shields.io/youtube/channel/views/UC8butISFwT-Wl7EV0hUK0BQ

Some badges are in the "social" category, but we don't render them using the social style by default. For example:

https://img.shields.io/gitlab/forks/gitlab-org%2Fgitlab

You have to explicitly pass `?style=social`

https://img.shields.io/gitlab/forks/gitlab-org%2Fgitlab?style=social

but then passing `?style=social` does some extra things, like adding a default logo.

This means that on the documentation page for https://shields.io/badges/you-tube-channel-views we render this badge as social by default, but on https://shields.io/badges/git-lab-forks we don't, even though they're both in the "social" category.

We used to deal with this in the frontend by automatically appending `?style=social` for badges in the social category. With the new frontend, we don't have a mechanism to easily insert a param like this.

One solution to this is to make all the social badges render social even when no style param is passed, but I can't remember the history behind this. There may have been some relevant conversation. Was there a reason why we don't do this in some cases, or is it just inconsistency?

PyvesB:
Fixes #9704 by rendering all badges in the social category with the social style and a logo by default.

### DIFF
--- a/services/github/gist/github-gist-stars.service.js
+++ b/services/github/gist/github-gist-stars.service.js
@@ -52,7 +52,11 @@ export default class GistStars extends GithubAuthV4Service {
   }
 
   static render({ stargazerCount, url, stargazers }) {
-    return { message: metric(stargazerCount), link: [url, stargazers] }
+    return {
+      message: metric(stargazerCount),
+      style: 'social',
+      link: [url, stargazers],
+    }
   }
 
   async fetch({ gistId }) {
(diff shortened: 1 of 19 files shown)


chris48s:
Changes look good. Can you have a look at the failing twitch auth tests? I expect we just need to update the expected value.

PyvesB:
Yeah, I've been fighting against unstable connectivity from a train, I'll circle back to this tomorrow. :)

chris48s (approved):

PR #9861, opened 2023-12-31
migrate examples to openApi part 32; affects [azuredevops youtube]

PyvesB:

### DIFF
--- a/services/azure-devops/azure-devops-tests.service.js
+++ b/services/azure-devops/azure-devops-tests.service.js
@@ -1,4 +1,5 @@
 import Joi from 'joi'
+import { queryParam, pathParam } from '../index.js'
 import {
   testResultQueryParamSchema,
   renderTestResultBadge,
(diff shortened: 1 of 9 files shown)


services/azure-devops/azure-devops-tests.service.js
```diff
@@ -65,48 +53,84 @@ export default class AzureDevOpsTests extends AzureDevOpsBase {
...
+            example: 'failed',
+          }),
+          queryParam({
+            name: 'skipped_label',
+            example: 'skipped',
+          }),
+          queryParam({
+            name: 'compact_message',
+            example: null,
+          }),
```
chris48s, line 89: instead of defining these common params here (twice), you can import `testResultOpenApiQueryParams` from `test-results.js` and use `...testResultOpenApiQueryParams` here.

PyvesB: Good idea!

chris48s (approved):

PR #9865, opened 2024-01-01
migrate examples to openApi part 34; affects [github]

chris48s:
Refs #9285

There were a couple of failing GitHub service tests so I started off by fixing them here, just so I could run `[github]` on the PR:

1. When I did #9818 I forgot to update the tests for GitHub downloads - it uses a helper function shared with releases/tags. I fixed the mock
2. When I updated the branch protection rules for #9799 it changed the status of older PRs to pending. I updated the example

so that gets the github tests passing

---

Then I started tackling some of the meatier GitHub badges that had a lot of variants or use shared helpers. I've only touched 6 services in this PR as there is quite a lot of diff to review just for these 6 services and I'm trying to keep each PRs relatively small/easy to review.


PyvesB (approved):
Looks good to me! 👍🏻

PR #9866, opened 2024-01-01
migrate examples to openApi part 35; affects [npm]

chris48s:

### DIFF
--- a/services/npm/npm-dependency-version.service.js
+++ b/services/npm/npm-dependency-version.service.js
@@ -1,8 +1,9 @@
+import { pathParam, queryParam } from '../index.js'
 import { getDependencyVersion } from '../package-json-helpers.js'
-import NpmBase from './npm-base.js'
-
-const { queryParamSchema } = NpmBase
-const keywords = ['node']
+import NpmBase, {
+  queryParamSchema,
+  packageNameDescription,
+} from './npm-base.js'
 
 export default class NpmDependencyVersion extends NpmBase {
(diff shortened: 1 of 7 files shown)


services/npm/npm-dependency-version.service.js
```diff
@@ -14,89 +15,55 @@ export default class NpmDependencyVersion extends NpmBase {
...
+          pathParam({
+            name: 'packageName',
+            example: 'react-boxplot',
+            description: packageNameDescription,
+          }),
+          pathParam({
+            name: 'dependency',
+            example: 'simple-statistics',
+            description: packageNameDescription,
+          }),
```
PyvesB, line 32: I like the fact that we've simplified this into two fields with an explanation of what this scoping business is. I'm personally somewhat familiar with the "scope" terminology because I've occasionally come across it in the Shields.io code, but I feel that someone which isn't a JS power-user may not be familiar with the wording.

However, I must admit that I was briefly confused both when reading the code and trying the badge out in the review app. I think this boils down to the fact that `package` is a package and `dependency` is also a package, and we're basically overloading the word package in the path parameter name, the explanation of the parameters, and the examples within the explanations.

I think it may help to replace any mentions of package by dependency in the dependency parameter description (even if a dependency is technically also a package), and instead of giving `package` and `@author/package` as examples in the description, use something more specific-sounding like  `my-package` and `@author/my-package`

PyvesB (approved):

PR #9870, opened 2024-01-02
Add Support for [Nostr] Followers

sepehr-safari:
Added first and simplest interaction with Nostr (https://github.com/nostr-protocol/nostr).

Using APIs from https://api.nostr.band/ to keep things fast and simple.

Some other useful stats will be added further.


sepehr-safari:
ready to merge I guess?

chris48s:
Before I do a code review on this, a couple of things:

1. As I understand it, nostr is a decentralised protocol. https://nostr.band/ is a stats service, but it is basically a stats service someone has set up - it is not "official". Someone else could come along and set up another "stats about nostr" service which is no more or less valid as a source of information than nostr.band. Is that a reasonable assessment of what nostr.band is?

2. Your service tests are failing. Can you have a look at getting the build passing before we review.

Thanks

sepehr-safari:
> nostr is a decentralised protocol.

Yes.

>  https://nostr.band/ is a stats service.

Yes.

> https://nostr.band/ is not "official".

Yes.

> Someone else could come along and set up another "stats about nostr" service.

Yes.

> Your service tests are failing.

FIXED! (https://github.com/badges/shields/pull/9870/commits/3e896e41fa89e209b2843e1c6a3ca8540884bcf0)

### so here is the conclusion:

1. **there is NOTHING "official" about Nostr.** not even a little thing. this is the first rule of "decentralize".
2. we can simply rely on a well-stablished service such as **nostr.band**.


ready to code review and merge.
thank you.

chris48s:
So I think I'd say the "service" here isn't really "nostr". It is nostr.band.

I guess what I'm getting at here is: If someone else sets up a competing service offering stats about the nostr network, why did stats from nostr.band get to be `/nostr` and not them? Seems like no good reason, right?

I'd suggest we make the route base for this `/nostr-band`.

A somewhat analogous situation would be: We have badges that show stats about the NPM ecosystem from https://npms.io/ and https://npm-stat.com/ but they live under `/npm-stat` and `/npms-io`.

sepehr-safari:
you are right. I've changed the service from `nostr` to `nostr-band`.
is there anything else that should be considered for this pr?

I would also suggest creating a separate service for `nostr` where users can freely choose their desired `relay` and fetch stats directly from that relay over `websocket`.
that would be the real decentralized nostr flow, not forcing users to use a pre-defined API service.
is there any documentation to use `websocket` inside your services?

services/nostr-band/nostr-band-followers.service.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default class NostrBandFollowers extends BaseJsonService {
+  static category = 'social'
+
+  static route = {
+    base: 'nostr-band/followers',
+    pattern: ':npub',
+  }
+
+  static examples = [
```
chris48s, line 21: Can we write this as an `openApi` object. See the docs in https://github.com/badges/shields/blob/master/doc/TUTORIAL.md#44-adding-documentation-to-the-frontend

We do still have some services using the old `examples` format, but there are not many left and we are in the process of converting them all. See https://github.com/badges/shields/issues/9285

New badges should use the `openApi` format.

services/nostr-band/nostr-band-followers.service.js
```diff
@@ -0,0 +1,57 @@
+import Joi from 'joi'
+import { metric } from '../text-formatters.js'
+import { BaseJsonService } from '../index.js'
+
+const npubSchema = Joi.object({
+  followers_pubkey_count: Joi.number().required(),
+}).required()
+
+const mainSchema = Joi.object({
+  stats: Joi.object().pattern(Joi.string(), npubSchema).required(),
```
chris48s, line 10: If we're going to access `Object.keys(stats)[0]` we need to make sure the stats object has at least one key when we validate. This schema will accept `{ "stats": {} }` as a valid input and then produce undesired output. We can ensure the object has at least one key by doing `stats: Joi.object().min(1).pattern(...` here. Tbh, I'd actually be inclined to make this `.min(1).max(1)` as based on the code, it seems like we're really expecting _exactly_ one key and two would be unexpected?

services/nostr-band/nostr-band-followers.service.js
```diff
@@ -0,0 +1,57 @@
...
+    {
+      title: 'Nostr.band Followers',
+      namedParams: {
+        npub: 'npub18c556t7n8xa3df2q82rwxejfglw5przds7sqvefylzjh8tjne28qld0we7',
+      },
+      staticPreview: this.render({ followers: 720 }),
+    },
+  ]
+
+  static defaultBadgeData = { label: 'followers', namedLogo: 'nostr' }
```
chris48s, line 31: This logo does not exist in the simple-icons set. Can we remove it

services/nostr-band/nostr-band-followers.tester.js
```diff
@@ -0,0 +1,28 @@
+import Joi from 'joi'
+import { isMetric } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
```
chris48s, line 4: This file looks like we've got 4 tests, but as far as I can see, 2 of them are just the same two tests duplicated. They don't seem to be functionally different.

services/nostr-band/nostr-band-followers.tester.js
```diff
@@ -0,0 +1,28 @@
...
+
+t.create('fetch: valid npub')
+  .get('/npub18c556t7n8xa3df2q82rwxejfglw5przds7sqvefylzjh8tjne28qld0we7.json')
+  .expectBadge({
+    label: 'followers',
+    message: Joi.alternatives().try(isMetric, Joi.equal('invalid pubkey')),
```
chris48s, line 10: Can we make the expectation here just `isMetric`. If this example becomes invalid we'll update it.

chris48s (commented):
thanks - have looked over the code and left some comments

chris48s:
> I would also suggest creating a separate service for nostr where users can freely choose their desired relay and fetch stats directly from that relay over websocket.
> that would be the real decentralized nostr flow, not forcing users to use a pre-defined API service.
> is there any documentation to use websocket inside your services?

I suggest moving the discussion of this to a badge request issue. Lets not scope creep here as it feels like what you've got here is pretty close.

We don't have any other services that use websockets, so if you wanted to contribute that it would have to be a new base class. If you wanted to implement that, I'd suggest looking over the existing base classes we have, for example:

- https://github.com/badges/shields/blob/master/core/base-service/base-graphql.js
- https://github.com/badges/shields/blob/master/core/base-service/base-json.js
- https://github.com/badges/shields/blob/master/core/base-service/base-xml.js

but its probably useful to nail down the details in an issue a bit first given it would be quite an involved contribution.

services/nostr-band/nostr-band-followers.service.js
sepehr-safari: done: https://github.com/badges/shields/pull/9870/commits/16a1b31ae17ff9b019875f1e93abc06a9378af55

services/nostr-band/nostr-band-followers.service.js
sepehr-safari: done: https://github.com/badges/shields/pull/9870/commits/bc6faa834fb85bfa8cc8f186f1a08d010fd2dc87

services/nostr-band/nostr-band-followers.service.js
sepehr-safari: done: https://github.com/badges/shields/pull/9870/commits/4d28b105572e9d5f716c4898e152d81a31e9ac80

services/nostr-band/nostr-band-followers.tester.js
sepehr-safari: done: https://github.com/badges/shields/pull/9870/commits/6f75cffe2beef226c339afa3a99988278ad68aff

services/nostr-band/nostr-band-followers.tester.js
sepehr-safari: done: https://github.com/badges/shields/pull/9870/commits/63b4d698d3edae15399ad32beffed2ec475e2971

sepehr-safari:
thanks for your detailed review. think I've successfully addressed all your specs.

sepehr-safari:
> I suggest moving the discussion of this to a badge request issue.

good idea. I will open an issue for implementing websocket and the rest.

services/nostr-band/nostr-band-followers.tester.js
```diff
@@ -0,0 +1,16 @@
+import Joi from 'joi'
+import { isMetric } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('fetch: valid npub')
+  .get('/npub18c556t7n8xa3df2q82rwxejfglw5przds7sqvefylzjh8tjne28qld0we7.json')
+  .expectBadge({
+    label: 'followers',
+    message: Joi.alternatives().try(isMetric),
```
chris48s, line 10:
```suggestion
    message: isMetric,
```

chris48s (approved):
Nice one. Merging. Btw, I really appreciate the way you replied to each comment with a link to the commit addressing that comment - makes the second pass of review super easy. Thanks

sepehr-safari:
> I really appreciate the way you replied to each comment with a link to the commit addressing that comment - makes the second pass of review super easy.

glad it was helpful. thank you.

Issue #9733, opened 2023-11-12
Rust MSRV Badge


Sky9x:
### 📋 Description

Rust crates can specify an "MSRV" (minimum supported rust version), which means that that crate requires a rust compiler of at least that version or newer. Crates.io, the package repository, displays this information on the crate page ([Example](https://crates.io/crates/serde), look on the right under "Metadata").

Service: crates.io (maybe github)
Possible Examples:
![Static Badge](https://img.shields.io/badge/MSRV-v1.72-lightgrey?logo=rust)
![Static Badge](https://img.shields.io/badge/MSRV-v1.72+-lightgrey?logo=rust)
![Static Badge](https://img.shields.io/badge/MSRV-rust%20v1.72.0-lightgrey)
![Static Badge](https://img.shields.io/badge/MSRV-v1.72.0-lightgrey)
![Static Badge](https://img.shields.io/badge/MSRV-unknown-red) (if the crate doesn't specify one)

### 🔗 Data

The MSRV is specified in the `package.rust-version` key in a package's `Cargo.toml` (manifest).

Data can be pulled from crates.io (latest or specific version of a specific crate) and maybe github/etc.
shields.io already has crates.io/github sourced badges, so this could just use that pre-existing architecture.

### 🎤 Motivation

Most crates specify a MSRV for compatibility.
This would be used primarily in project readmes to indicate to users the minimum rust version they need.
Crates like [serde](https://github.com/serde-rs/serde/blob/4676abdc9e6bbbddfb33a00ce8d7e81e92f01120/README.md) (one of the most popular crates) already do this manually, so providing an automatic way to do it would be useful.

chris48s:
Looks like the key in the crates.io API is called `rust_version`

calebcartwright:
Agreed this seems a reasonable badge to add. However, I think we'd want to have two distinct badges, one coming from crates and another (potentially) that's pulling from the manifest from various source control systems.

I'd definitely advise starting with the former, and then we gauge what the appetite would be for the latter once the former has been available for a while

Sky9x:
> Agreed this seems a reasonable badge to add. However, I think we'd want to have two distinct badges, one coming from crates and another (potentially) that's pulling from the manifest from various source control systems.

Agreed. crates.io would be the primary source of information, and I only included github etc because other similar badges like version and license have a github option.

PR #9871, opened 2024-01-03
[CratesMSRV] Badge

Sky9x:
Proposed in #9733

Routes:
Crates.io MSRV: `/crates/msrv/:crate`
Crates.io MSRV (version): `/crates/msrv/:crate/:version`

### DIFF
--- a/services/crates/crates-msrv.service.js
+++ b/services/crates/crates-msrv.service.js
@@ -1,0 +1,74 @@
+import { NotFound, pathParams } from '../index.js'
+import {
+  BaseCratesService,
+  description as cratesIoDescription,
+} from './crates-base.js'
+
+const description = `
+${cratesIoDescription}
+
+MSRV is a crate's minimum suppported rust version,
+the oldest version of Rust supported by the crate.
+See the [Cargo Book](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field)
+for more info.
+`
+
+export default class CratesMSRV extends BaseCratesService {
+  static category = 'platform-support'
+  static route = {
+    base: 'crates/msrv',
+    pattern: ':crate/:version?',
+  }
+
+  static openApi = {
+    '/crates/msrv/{crate}': {
+      get: {
+        summary: 'Crates.io MSRV',
+        description,
+        parameters: pathParams({
+          name: 'crate',
+          example: 'serde',
+        }),
+      },
+    },
+    '/crates/msrv/{crate}/{version}': {
+      get: {
+        summary: 'Crates.io MSRV (version)',
+        description,
+        parameters: pathParams(
+          {
+            name: 'crate',
+            example: 'serde',
+          },
+          {
+            name: 'version',
+            example: '1.0.194',
+          },
+        ),
+      },
+    },
+  }
+
+  static defaultBadgeData = { label: 'msrv', color: 'blue' }
+
+  static transform({ errors, version, versions }) {
+    // crates.io returns a 200 response with an errors object in
+    // error scenarios, e.g. https://crates.io/api/v1/crates/libc/0.1
+    if (errors) {
+      throw new NotFound({ prettyMessage: errors[0].detail })
+    }
+
+    const msrv = version ? version.rust_version : versions[0].rust_version
+    if (!msrv) {
+      throw new NotFound({ prettyMessage: 'unknown' })
+    }
+
+    return { msrv }
+  }
+
+  async handle({ crate, version }) {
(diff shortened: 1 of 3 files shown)


Sky9x:
failure is not in a file that i modified?

services/crates/crates-msrv.service.js
```diff
@@ -0,0 +1,78 @@
...
+  static render({ msrv: message }) {
+    return { message }
+  }
+
+  static transform({ errors, version, versions }) {
+    // crates.io returns a 200 response with an errors object in
+    // error scenarios, e.g. https://crates.io/api/v1/crates/libc/0.1
+    if (errors) {
+      throw new NotFound({ prettyMessage: errors[0].detail })
+    }
```
chris48s, line 63: I'm not going to insist we do it to merge this, but I'm making a note to myself here: We should move this logic up to the base class because it affects all services that call this API.

services/crates/crates-msrv.service.js
```diff
@@ -0,0 +1,78 @@
...
+          {
+            name: 'version',
+            example: '1.0.194',
+          },
+        ),
+      },
+    },
+  }
+
+  static defaultBadgeData = { label: 'MSRV', color: 'blue' }
```
chris48s, line 52: Label should be lowercase. See https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines

```suggestion
  static defaultBadgeData = { label: 'msrv', color: 'blue' }
```

services/crates/crates-msrv.service.js
```diff
@@ -0,0 +1,78 @@
...
+
+  static render({ msrv: message }) {
+    return { message }
+  }
```
chris48s, line 56: This function isn't really doing anything. We could just `return { message: msrv }` in `handle()`

services/crates/crates-msrv.service.js
```diff
@@ -0,0 +1,78 @@
...
+    if (errors) {
+      throw new NotFound({ prettyMessage: errors[0].detail })
+    }
+
+    const msrv = version ? version.rust_version : versions[0].rust_version
+    if (!msrv) {
+      throw new NotFound({ prettyMessage: 'unknown' })
+    }
+
+    return { msrv: `${msrv}+` }
```
chris48s, line 70: Looking at crates.io, they present this as just the version number (without a plus). I'd be inclined to match the upstream registry on this.

chris48s (commented):
Thanks for picking this up. Looks like the test failure is unrelated to the changes in this PR. I've set this PR to just run the tests for MSRV. We can deal with addressing that test in another PR.

services/crates/crates-msrv.service.js
Sky9x: :+1:
i had just copied `crates-license.service.js` and used it as a template, which is why this is the way that it is

services/crates/crates-msrv.service.js
Sky9x: Should acronyms still be lowercase? MSRV is an acronym for "minimum supported rust version".
I couldn't find any badges with labels that are acronyms (but maybe I'm not looking hard enough), so I don't if there is precedent for this.

I personally like the uppercase more, but I have no issue with lowercasing it.

samples:
MSRV
msrv

services/crates/crates-msrv.service.js
Sky9x: The MSRV indicates not a specific version requirement, but a "this version or later" requirement (hence the 'minimum supported' in 'minimum supported rust version').

I've opened https://github.com/rust-lang/crates.io/pull/7888 to possibly change this on the crates.io side. I'm willing to match whatever the crates.io team decides.

Sky9x: update: crates.io team thinks the plus is unnecessary
https://github.com/rust-lang/crates.io/pull/7888#issuecomment-1879820085

services/crates/crates-msrv.service.js
chris48s: Most of the examples I can think of are where the badge label is the name of a service, e.g:

Comprehensive Perl Archive Network: 
Arch linux User Repository: 
Node Package Manager: 

but label is always lowercase

Sky9x: ok! thanks for the examples. i guess it can be overridden with `?label=` anyway :shrug:

Sky9x:
should be good to go unless you have any more suggestions

chris48s (approved):

PR #9872, opened 2024-01-04
Update [Gitea] defaults to gitea.com

CanisHelix:
As https://gitea.com is a flagship instance of Gitea, updating the Gitea badges to default to gitea.com as a default. Added an extra test to each badge using the default instance.

Updated the message header to be clear about the default instance, and self-managed ones.

Updated tests to use try.gitea.io which is the official demo site, handling like a self-managed one.


chris48s:
Can you have a look at the failing `GiteaBase` auth test before we review. Looks like the nock URL needs to be updated to account for these changes? Thanks

CanisHelix:
> Can you have a look at the failing `GiteaBase` auth test before we review. Looks like the nock URL needs to be updated to account for these changes? Thanks

I accidentally removed the protocol when changing the URL on the base auth test. Looks good now. Is there a way to perform a core test for just a single service? I tried `npm run test:core -- --only=gitea` but it still performs a test on all services.

services/gitea/gitea-helper.js
```diff
@@ -1,5 +1,6 @@
 const documentation = `
-Note that the gitea_url parameter is required because there is canonical hosted gitea service provided by Gitea.
+Note that only internet-accessible Gitea/Forgejo based instances are supported, for example https://gitea.com, https://codeberg.org.
+For self-hosted instances, the gitea_url parameter is required.
```
chris48s, line 3: Can we reword this to make it more clear that `gitea_url` defaults to https://gitea.com when not supplied.

How about something like:

```
By default this badge looks for repositories on [gitea.com](https://gitea.com).
To specify another instance like [codeberg](https://codeberg.org/), [forgejo](https://forgejo.org/) or a self-hosted instance, use the `gitea_url` query param.
```

?

services/gitea/gitea-languages-count.tester.js
```diff
@@ -3,24 +3,29 @@ import { createServiceTester } from '../tester.js'
...
-t.create('language count (empty repo)')
+t.create('language count').get('/gitea/tea.json').expectBadge({
+  label: 'languages',
+  message: Joi.number().integer().positive(),
+})
+
+t.create('language count (empty repo) (self-managed)')
   .get(
-    '/CanisHelix/shields-badge-test-empty.json?gitea_url=https://codeberg.org',
+    '/CanisHelix/shields-badge-test-empty.json?gitea_url=https://try.gitea.io',
```
chris48s, line 13: https://try.gitea.io says
"THIS IS A TEST INSTANCE ONLY! REPOSITORIES CAN BE DELETED AT ANY TIME!"
at the top of it in massive letters.

Having integration tests that run against this says to me that at some point these tests will unexpectedly break due to an externality. I'd be inclined to stick with the custom instance tests running against codeberg.

services/gitea/gitea-helper.js
CanisHelix: Agreed, this is much better verbiage.

chris48s (approved):
latest changes look great - thanks

PR #9873, opened 2024-01-04
migrate examples to openApi part 36; affects [myget nuget resharper chocolatey powershellgallery]

chris48s:

### DIFF
--- a/services/myget/myget.service.js
+++ b/services/myget/myget.service.js
@@ -1,3 +1,4 @@
+import { pathParams } from '../index.js'
 import { createServiceFamily } from '../nuget/nuget-v3-service-family.js'
 
 const { NugetVersionService: Version, NugetDownloadService: Downloads } =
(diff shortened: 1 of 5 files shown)


services/myget/myget.service.js
```diff
@@ -8,51 +9,73 @@ const { NugetVersionService: Version, NugetDownloadService: Downloads } =
...
+      },
     },
-    {
-      title: 'MyGet tenant',
-      pattern: ':tenant.myget/:feed/v/:packageName',
-      namedParams: {
-        tenant: 'tizen',
-        feed: 'dotnet',
-        packageName: 'Tizen.NET',
+    '/{tenant}/{feed}/{variant}/{packageName}': {
```
chris48s, line 29: In OpenAPI, the separator between params and another param or static part of the route has to be a `/` so we can't hard-code the `.myget` bit by making this `/{tenant}.myget/{feed}`, so I've just done the best I can here. This does lead to a slightly weird situation where if you put something in here that doesn't end with `.myget` you can put another service name in as the first param.

In general, the routes for this service are not good routes.

They still use the old `v`/`vpre` instead of an `include_prereleases` flag

Also the order of the params doesn't conform to our own guidance on how to correctly structure routes

Having done the best I can with what we've got, I'm going to suggest we just roll with this as it stands for the purposes of this migration and I'll raise an issue to bring the myget routes into line with current conventions and set up redirects.

PyvesB (approved):
Looks good to me! :+1:

PR #9881, opened 2024-01-06
migrate examples to openApi part 37; affects [bitbucket bitrise chromewebstore circleci]

PyvesB:

~Leaving as draft for now, gonna see if I can find some good updated Bitbucket examples.~ All tests and examples fixed.

### DIFF
--- a/services/bitrise/bitrise.service.js
+++ b/services/bitrise/bitrise.service.js
@@ -41,7 +41,7 @@ export default class Bitrise extends BaseJsonService {
         parameters: [
           pathParam({
             name: 'appId',
-            example: '4a2b10a819d12b67',
+            example: 'e736852157296019',
           }),
           pathParam({
             name: 'branch',
--- a/services/bitrise/bitrise.service.js
+++ b/services/bitrise/bitrise.service.js
@@ -29,7 +29,7 @@ export default class Bitrise extends BaseJsonService {
           }),
           queryParam({
             name: 'token',
-            example: '859FMDR8QHwabCzwvZK6vQ',
+            example: 'abc123def456',
             required: true,
           }),
         ],
@@ -49,7 +49,7 @@ export default class Bitrise extends BaseJsonService {
           }),
           queryParam({
             name: 'token',
-            example: '859FMDR8QHwabCzwvZK6vQ',
+            example: 'abc123def456',
             required: true,
           }),
         ],
--- a/services/bitrise/bitrise.service.js
+++ b/services/bitrise/bitrise.service.js
@@ -25,11 +25,12 @@ export default class Bitrise extends BaseJsonService {
         parameters: [
           pathParam({
             name: 'appId',
-            example: '3ff11fe8457bd304',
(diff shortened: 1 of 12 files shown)


services/bitrise/bitrise.service.js
```diff
@@ -18,14 +18,44 @@ export default class Bitrise extends BaseJsonService {
...
+        parameters: [
+          pathParam({
+            name: 'appId',
+            example: '4a2b10a819d12b67',
+          }),
+          queryParam({
+            name: 'token',
+            example: '859FMDR8QHwabCzwvZK6vQ',
+            required: true,
+          }),
```
PyvesB, line 34: These are the appId and token for https://github.com/OpenAPITools/openapi-generator. It feels slightly off to use someone else's token, but I guess this is what we were already doing, they're public, and they're already used to generate the same badge in their repository's README. Does anyone feel strongly about this?

chris48s: Yeah its an interesting one. Sometimes we do make the example a valid token, usually if the person who contributed the badge puts one in there.

Sometimes we don't though. For example, you'll notice that the token in the Circle CI example that is also in this PR is `abc123def456`, which is clearly not a vlaid token. So :shrug: If we just fill in a random string or leave it as a non-working token it won't be the only case where we have a non-working example.

I'd probably be more inclined to do that than use one without asking, even if it is in a public repo.

PyvesB: What is your take on the live tests though?

PyvesB: I've updated the examples with a fake token and have reached out [here ](https://github.com/OpenAPITools/openapi-generator/issues/17608) to see if we can use their public token in our repo.

PyvesB: Okay, given the absence of response, trying my luck with another repository: https://github.com/JohnEstropia/CoreStore/issues/496

chris48s: OK. Thanks.

If this PR ends up being the last blocker to finishing https://github.com/badges/shields/issues/9285 I think I'd be in favour of just merging this a non-working token in the example and a failing service test - we can follow it up later.

PyvesB:
@chris48s there you go, ready for review with all tests fixed!

services/bitrise/bitrise.service.js
```diff
@@ -18,14 +18,44 @@ export default class Bitrise extends BaseJsonService {
...
+            name: 'appId',
+            example: 'e736852157296019',
+          }),
+          pathParam({
+            name: 'branch',
+            example: 'master',
+          }),
+          queryParam({
+            name: 'token',
+            example: 'abc123def456',
```
chris48s, line 52: were you planning to update the examples too, or only the tests?

PyvesB: Only the tests, that's what I asked for in the linked issue.

chris48s (approved):

PR #9883, opened 2024-01-08
[crates] tidyup

chris48s:
There's a few small things here that came out of the review of https://github.com/badges/shields/pull/9871

1. Fix a failing service test. This is a change in the upstream API behaviour. That case now returns a `404 Not Found` with the body `{"errors":[{"detail":"crate `libc` does not have a version `7`"}]}`. I've updated the assertion as we now throw an error based on the response code.
2. I was originally planning to move the handling of "`200 OK` with an error body" up a level to de-dupe, and check for that condition in `BaseCratesService.fetch()`. However, given that the condition we were attempting to handle looks like it is now fixed (and that's what required us to update that test), I've decided to just completely remove the handling for this and treat a `200 OK` with an error body as a schema fail from now on. None of the calls in the comments now yield that result, all the tests pass, and I now don't have an example of a call that behaves this way. If it does happen, Joi will reject the response anyway, albeit with a more generic error.
3. I removed a `render()` method that just passes the value through.

### DIFF
--- a/services/crates/crates-base.js
+++ b/services/crates/crates-base.js
@@ -29,14 +29,7 @@ const versionSchema = Joi.object({
   }).required(),
 }).required()
 
-const errorSchema = Joi.object({
-  errors: Joi.array()
-    .items(Joi.object({ detail: Joi.string().required() }))
-    .min(1)
-    .required(),
-}).required()
-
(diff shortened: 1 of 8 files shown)


PyvesB (approved):
Yay, simplification! 🎉

Issue #9884, opened 2024-01-08
Order badges by title in menus


chris48s:
:clipboard: **Description**

When we generate the OpenAPI yaml, the badges are not ordered alphabetically by the title that appears in the menu. I think they probably come out in the order we loaded the service classes in. This means some services appear in unexpected places in the menu e.g:

![Screenshot at 2024-01-08 18-47-24](https://github.com/badges/shields/assets/6025893/36902157-a8a5-4751-aedd-20bad47563c8)

PR #9888, opened 2024-01-09
sort categories by title (except core)

chris48s:
Closes https://github.com/badges/shields/issues/9884

I've kept core unsorted so we can put "static badge" first. The other categories now sort by title


PyvesB (approved):
Nice small improvement!

PR #9889, opened 2024-01-09
migrate examples to openApi part 38; affects [github]

chris48s:

### DIFF
--- a/services/github/github-downloads.service.js
+++ b/services/github/github-downloads.service.js
@@ -1,13 +1,17 @@
 import Joi from 'joi'
 import { nonNegativeInteger } from '../validators.js'
 import { renderDownloadsBadge } from '../downloads.js'
-import { NotFound } from '../index.js'
+import { NotFound, pathParam, queryParam } from '../index.js'
 import { GithubAuthV3Service } from './github-auth-service.js'
 import { fetchLatestRelease } from './github-common-release.js'
 import { documentation, httpErrorsFor } from './github-helpers.js'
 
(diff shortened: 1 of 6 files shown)


services/github/github-downloads.service.js
```diff
@@ -24,173 +28,84 @@ const releaseArraySchema = Joi.alternatives().try(
   Joi.array().length(0),
 )
 
+const variantParam = pathParam({
```
chris48s, line 31: I've been kind of looking forward to getting this one sorted and simultaneously dreading the process of unpicking it.

Because we've got lots of different variants with slightly different combinations of the same params (all of which occur at least twice), I decided to pull them all out into variables and then lego them together.

services/github/github-size.service.js
```diff
@@ -25,33 +25,24 @@ export default class GithubSize extends GithubAuthV3Service {
...
+        summary: 'GitHub file size in bytes',
+        description: documentation,
+        parameters: [
+          pathParam({ name: 'user', example: 'webcaetano' }),
+          pathParam({ name: 'repo', example: 'craft' }),
+          pathParam({ name: 'path', example: 'build/phaser-craft.min.js' }),
+          queryParam({
+            name: 'branch',
+            example: 'master',
+            description: 'can be a branch, a tag or a commit hash',
```
PyvesB, line 40:
```suggestion
            description: 'Can be a branch, a tag or a commit hash.',
```

PyvesB (approved):
Looks (almost) good to me 👍🏻

PR #9903, opened 2024-01-15
migrate some services from examples to openApi part 40; affects [codecov obs scrutinizer sourceforge]

chris48s:

I'm really starting to feel like the end is in sight.


PyvesB (approved):
Looks good to me 👍🏻

PR #9915, opened 2024-01-20
migrate some services from examples to openApi part 42; affects [github]

chris48s:

### DIFF
--- a/services/github/github-package-json.service.js
+++ b/services/github/github-package-json.service.js
@@ -1,4 +1,5 @@
 import Joi from 'joi'
+import { pathParam, pathParams, queryParam } from '../index.js'
 import { renderVersionBadge } from '../version.js'
 import { transformAndValidate, renderDynamicBadge } from '../dynamic-common.js'
 import {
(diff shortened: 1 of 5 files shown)


services/github/github-package-json.service.js
```diff
@@ -94,58 +85,100 @@ class GithubPackageJsonDependencyVersion extends ConditionalGithubAuthV3Service
...
 
-  static examples = [
-    {
-      title: 'GitHub package.json dependency version (prod)',
-      pattern: ':user/:repo/:packageName',
-      namedParams: {
-        user: 'developit',
-        repo: 'microbundle',
-        packageName: 'rollup',
+  static openApi = {
```
chris48s, line 88: I documented this as 4 variants, seperating prod dependencies (with no `:kind` param) from dev/peer/optional (with a `:kind` param). This is similar to the corresponding NPM badge, although there's only 2 variants there because we don't have an optional `branch` param. The other thing I considered here was changing the route by adding `prod` to the `:kind` enum and setting up a redirect so we could reduce this to two variants (with and without `branch`). I think this is fine, but I'd probably hesitate to add another route with multiple optional path components in future.

PyvesB (approved):
All good 👍🏻

PR #9917, opened 2024-01-21
use brightgreen for build passing in package docs

chris48s:
Minor thing I noticed while working on https://github.com/badges/shields/pull/9916
The semantic colour for build pass is brightgreen so lets get that right in the package docs


PyvesB (approved):
Good catch!

PR #9918, opened 2024-01-22
migrate some services from examples to openApi part 43; affects [jenkins]

chris48s:

### DIFF
--- a/services/jenkins/jenkins-coverage.tester.js
+++ b/services/jenkins/jenkins-coverage.tester.js
@@ -10,13 +10,13 @@ export const t = await createServiceTester()
 t.create('jacoco: job found')
   .get(
     `/jacoco.json?jobUrl=${encodeURIComponent(
-      'https://wso2.org/jenkins/view/All%20Builds/job/archetypes',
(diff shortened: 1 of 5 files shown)


services/jenkins/jenkins-coverage.tester.js
```diff
@@ -66,9 +60,17 @@ t.create('code coverage API v1: job found')
   )
   .expectBadge({ label: 'coverage', message: isIntegerPercentage })
 
-t.create('code coverage API v4+: job not found')
+t.create('code coverage API v1: job not found')
   .get(
-    '/apiv4.json?jobUrl=https://jenkins.library.illinois.edu/job/does-not-exist',
+    '/apiv1.json?jobUrl=http://loneraver.duckdns.org:8082/job/does-not-exist',
```
chris48s, line 65: I can't find any examples of a jenkins server supporting apiv1 so I've just replaced the not found test with a mock. The live test for this case is already a mock anyway.

PyvesB (approved):
Looks good to me 👍🏻

PR #9919, opened 2024-01-22
migrate some services from examples to openApi part 44; affects [sonar]

chris48s:

Made a start on this service family. Submitting a WIP PR for now, just to lick the cookie. I'll finish it later in the week.

### DIFF
--- a/services/sonar/sonar-quality-gate.service.js
+++ b/services/sonar/sonar-quality-gate.service.js
@@ -1,5 +1,10 @@
+import { pathParam } from '../index.js'
 import SonarBase from './sonar-base.js'
-import { documentation, keywords, queryParamSchema } from './sonar-helpers.js'
+import {
+  documentation,
+  queryParamSchema,
+  openApiQueryParams,
+} from './sonar-helpers.js'
 
 export default class SonarQualityGate extends SonarBase {
   static category = 'analysis'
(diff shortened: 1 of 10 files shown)


chris48s:
I've tried to provide working examples for as many of these as possible, but there are a couple of these where I couldn't find any working example - in those cases, the tests are already mocks so I can't grab one from there and I wasn't able to find any examples in public repos on github. I've just left those with plausible but non-working values.

services/sonar/sonar-quality-gate.service.js
```diff
@@ -10,23 +15,29 @@ export default class SonarQualityGate extends SonarBase {
...
   }
 
-  static examples = [
-    {
-      title: 'Sonar Quality Gate',
-      namedParams: {
-        component: 'swellaby:azdo-shellcheck',
-        metric: 'quality_gate',
-        branch: 'master',
+  static openApi = {
```
PyvesB, line 18: Should there be an `alert_status` example?

services/sonar/sonar-tech-debt.service.js
```diff
@@ -16,26 +17,29 @@ export default class SonarTechDebt extends SonarBase {
...
   }
 
-  static examples = [
-    {
-      title: 'Sonar Tech Debt',
-      namedParams: {
-        component: 'org.ow2.petals:petals-se-ase',
-        metric: 'tech_debt',
-        branch: 'master',
+  static openApi = {
```
PyvesB, line 20: Should there be a `sqale_debt_ratio` example?

chris48s:
In both the cases you've flagged the `metric` param doesn't really do anything - we show the same value either way. I think this is a case where we changed the route at some point in the past and the enum is there to keep the old name working but we only document the new one.

PyvesB:
Okay, I guess we could implement a route redirector at some point instead!

PyvesB (approved):

chris48s:
There are one or two other services where we do something like this. Here's another example that springs to mind:

https://github.com/badges/shields/blob/1ed54400aa2f4f60d3e20df7fde95791c6a5a5bb/services/vaadin-directory/vaadin-directory-version.service.js#L8-L15

I don't really have strong views about replacing with a redirect, but maybe that pattern of calling the param `:alias` makes it clearer?

PR #9923, opened 2024-01-25
[GITEA] add forks, stars, issues and pr badges

CanisHelix:
Adding Forks, Stars, Issues and Pull Requests to Gitea/Forgejo service.

All tests use a dedicated repository at https://codeberg.org/CanisHelix/shields-badge-test except for one that uses a gitea application on their canonical server here https://gitea.com/gitea/tea


services/gitea/gitea-forks.tester.js
```diff
@@ -0,0 +1,33 @@
...
+    message: isMetric,
+    color: 'blue',
+    link: ['https://gitea.com/gitea/tea', 'https://gitea.com/gitea/tea/forks'],
+  })
+
+t.create('Forks (self-managed)')
+  .get('/CanisHelix/shields-badge-test.json?gitea_url=https://codeberg.org')
+  .expectBadge({
+    label: 'forks',
+    message: Joi.number().integer(),
```
chris48s, line 20: why not `isMetric` here?

services/gitea/gitea-forks.service.js
```diff
@@ -0,0 +1,76 @@
...
+      link: [`${baseUrl}/${user}/${repo}`, `${baseUrl}/${user}/${repo}/forks`],
+    }
+  }
+
+  async fetch({ user, repo, baseUrl }) {
+    // https://gitea.com/api/swagger#/repository
+    return super.fetch({
+      schema,
+      url: `${baseUrl}/api/v1/repos/${user}/${repo}`,
+      httpErrors: httpErrorsFor('user or repo not found'),
```
chris48s, line 64: We're passing a string to `httpErrorsFor()` here, but this function does not take any args

services/gitea/gitea-stars.service.js
```diff
@@ -0,0 +1,76 @@
...
+      link: [`${baseUrl}/${user}/${repo}`, `${baseUrl}/${user}/${repo}/stars`],
+    }
+  }
+
+  async fetch({ user, repo, baseUrl }) {
+    // https://gitea.com/api/swagger#/repository
+    return super.fetch({
+      schema,
+      url: `${baseUrl}/api/v1/repos/${user}/${repo}`,
+      httpErrors: httpErrorsFor('user or repo not found'),
```
chris48s, line 64: Same here. `httpErrorsFor()` here, doesn't take any args

services/gitea/gitea-pull-requests.service.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  async handle(
+    { variant, user, repo },
+    { gitea_url: baseUrl = 'https://gitea.com', labels },
+  ) {
+    const { res } = await this._request(
+      this.authHelper.withBearerAuthHeader({
+        url: `${baseUrl}/api/v1/repos/${user}/${repo}/issues`,
+        options: labels
```
chris48s, line 89: I think it would be better to write this as something like

```js
const options = {
  searchParams: {
    page: '1',
    limit: '1',
    type: 'issues',
    state: variant.replace('-raw', ''),
  }
}
if (labels) {
  options.searchParams.labels = labels
}
```

instead of using a ternary. If we duplicate the object, it is hard to tell at a glance what the difference is between them. Also if we want to add another property for example, we'd have to do it in two places.

services/gitea/gitea-pull-requests.tester.js
```diff
@@ -0,0 +1,173 @@
...
+  })
+
+t.create('All pulls by multi-word label is > zero')
+  .get(
+    '/all/CanisHelix/shields-badge-test.json?gitea_url=https://codeberg.org&labels=question,enhancement',
+  )
+  .expectBadge({
+    label: 'question,enhancement pull requests',
+    message: Joi.string().regex(
+      /^([0-9]+[kMGTPEZY]?|[1-9]\.[1-9][kMGTPEZY]) all$/,
```
chris48s, line 162: Why the custom validators here?

services/gitea/gitea-pull-requests.service.js
chris48s, line 1: There is a lot of duplication of logic between the Issues and PRs classes with only minor differences.
Seems like we could split out most of the logic into shared helpers, or have a base class with 2 concrete implementations. For example, if the render() method took a param called `count` (instead of `issueCount`/`pullsCount`) and uses `this.defaultBadgeData.label` for the label, they're the same function. Similarly, you could pull the `fetch()` logic out into a shared function and pass in a `type='pulls'/'issues'` param. The rest of it is the same.

chris48s (commented):
Thanks. I've done a first pass of review comments.

services/gitea/gitea-forks.tester.js
CanisHelix: **isMetric** only accepts >0 and I do not have any forks of the CI test project. Trying to avoid creating dummy accounts on https://codeberg.org where possible, but if this is preferred I could do it. The risk is that the 2nd dummy account could get purged due to external factors.

Saying that. **isMetricAllowNegative** seems to work.

services/gitea/gitea-pull-requests.tester.js
CanisHelix: This was taken from the Gitlab tests, presumably it was used there due to the lack of an **isMetricAllIssues**. There are validators for **isMetricOpenIssues** and **isMetricClosedIssues** but not one for an All.

Is using the function `isMetricWithPattern(/ all/)` okay? Or would it be preferred to add a new **isMetricAllIssues** to _test-validators.js_, that way other tests can benefit too?

services/gitea/gitea-pull-requests.service.js
CanisHelix: I've moved the **render** to a **renderIssues** function in the _gitea-helper.js_. I chose **renderIssues** as a pull is a 'type' of issue. I wasn't able to get it to reference back to the class though, so needed to pass `defaultBadgeData` along as a parameter. Will move the fetch into _gitea-base.js_ next.

services/gitea/gitea-pull-requests.tester.js
chris48s: Can we use a repo that has >0 forks for the test
e.g: https://codeberg.org/Codeberg/Documentation

services/gitea/gitea-forks.tester.js
chris48s: Can we use a repo that has >0 forks for the test
e.g: https://codeberg.org/Codeberg/Documentation

chris48s:
Thanks for your work on this. Latest changes look great. Merging with one final tweak to a service test.

chris48s (approved):

CanisHelix:
Whoops, I changed the test and forgot that isMetric. Thanks for the last minute update.

PR #9930, opened 2024-01-27
migrate some services from examples to openApi part 45; affects [github coveralls visualstudiomarketplace]

chris48s:

After this, I reckon there are probably 2 more of these PRs to go, and then we're done.

### DIFF
--- a/core/base-service/openapi.js
+++ b/core/base-service/openapi.js
@@ -212,7 +212,7 @@ function services2openapi(services, sort) {
       for (const [key, value] of Object.entries(
         addGlobalProperties(service.openApi),
       )) {
-        if (key in paths) {
+        if (key in paths && key !== '/github/{variant}/{user}/{repo}') {
           throw new Error(`Conflicting route: ${key}`)
         }
         paths[key] = value
--- a/cypress/e2e/main-page.cy.js
+++ b/cypress/e2e/main-page.cy.js
@@ -40,14 +40,14 @@ describe('Frontend', function () {
   })
 
   it('Build a badge', function () {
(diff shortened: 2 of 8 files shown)


core/base-service/openapi.js
```diff
@@ -212,7 +212,7 @@ function services2openapi(services, sort) {
       for (const [key, value] of Object.entries(
         addGlobalProperties(service.openApi),
       )) {
-        if (key in paths) {
+        if (key in paths && key !== '/github/{variant}/{user}/{repo}') {
```
chris48s, line 215: We've ended up with several services where we're representing the route as `/github/{variant}/{user}/{repo}`.
There are 2 in this PR and then also `/services/github/github-contributors.service.js`.
These are not actually duplicates at our routing layer because the 3 `{variant}`s are all enums with mutually exclusive values. It would be a problem if we were using the OpenAPI routes as our routing layer, but we aren't. I think the routes themselves are probably fine.
So anyway, I added an exception for this one route. The thing is, it doesn't actually break anything in the frontend, so maybe we just bin this check. It seemed like something we should avoid when I originally wrote this code. Feels like in most cases this would be an error, or at least something we want to know about, but maybe we actually just don't care?
I guess the options are:

- Continue to throw but add an exception for this one case (which is what I've done here). GitHub is a service which has a uniquely large number of badges. I think it is unlikely we would hit this with many other service families.
- Allow it, but switch it to emit a warning or something
- Completely bin the check

PyvesB: Happy with the approach you've taken in this PR.

PyvesB (approved):
Looks good to me! 👍🏻

PR #9932, opened 2024-01-29
deprecate TAS

chris48s:
This has been failing for a long time. As far as I can tell the upstream API endpoint is gone and has not been replaced.

Suggest we just bin this service.

### DIFF
--- a/services/tas/tas-tests.service.js
+++ b/services/tas/tas-tests.service.js
@@ -1,111 +1,13 @@
-import Joi from 'joi'
-import { BaseJsonService } from '../index.js'
-import {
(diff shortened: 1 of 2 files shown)


PyvesB:
@nevilm-lt and @palashmarantas you worked on the implementation in #7612, any thoughts?

chris48s:
As a data point, there are zero public READMEs on GitHub with one of the badges in

https://github.com/search?q=shields.io%2Ftas%2Ftests&type=code

PyvesB (approved):
I've there's no usage, let's not agonise over deprecation 👍🏻

Issue #9154, opened 2023-05-08
Security Headers requires an API key


HandyHat:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Security Headers now requires an API key from https://securityheaders.com/api/ which means that the generated badge from shields.io doesn't work anymore:
![image](https://user-images.githubusercontent.com/58530748/236889744-7d8eafd2-166d-4c57-a62a-eddca3a24b03.png)

### 🔗 Link to the badge

![Security Headers](https://img.shields.io/security-headers?url=https%3A%2F%2Fgoogle.com)

calebcartwright:
Thanks for letting us know!

I've opened an [issue upstream with the Security Headers](https://github.com/securityheaders/securityheaders-bugs/issues/110) folks as there's really nothing we can do from our end. It seems the new mandatory auth requirement also has an associated cost, and their highest (advertised) request volume is both markedly too low for our needs while also already being far, far more than we could afford.

I suspect we'll likely have to deprecate these badges, but hopefully they'll consider offering us some type of free/sponsored/open source account we can use with requisite rate limits :crossed_fingers:

calebcartwright:
For anyone not subscribed to the upstream issue, it does sound like they'll be willing to work with us so we'll collaborate off GitHub and post updates back here once we've reached a resolution

iansantagata:
Hey @calebcartwright, do you have any update here toward a resolution?  Thanks for trying to work on this!

PyvesB:
Was just about to reach out to upstream service, but I noticed that @jNullj was [ahead of me](https://github.com/securityheaders/securityheaders-bugs/issues/110#issuecomment-3506827311) by a few days. 😄

PR #9933, opened 2024-01-29
migrate some services from examples to openApi part 46; affects [codeclimate librariesio nexus] and securityheaders

chris48s:

With nexus, we don't have passing tests for nexus v3
I have actually managed to find a couple of public nexus servers running v3:
- https://repo.tomkeuper.com/
- https://repo.unnamed.team/
but I can't work out how to get a working badge out of either of them.

With Security Headers, the tests fail because the badges themselves are broken https://github.com/badges/shields/issues/9154 I figure lets just migrate the examples here and leave that as an issue for another time.

### DIFF
--- a/services/codeclimate/codeclimate-coverage.tester.js
+++ b/services/codeclimate/codeclimate-coverage.tester.js
@@ -46,7 +46,7 @@ t.create('test coverage when outer user repos query returns multiple items')
               },
               latest_default_branch_test_report: {
                 data: {
-                  id: '6598295abd91ec00013d485e',
+                  id: '65a1662cb0077b00013cb4de',
                   type: 'test_reports',
                 },
               },
--- a/services/codeclimate/codeclimate-coverage.tester.js
+++ b/services/codeclimate/codeclimate-coverage.tester.js
@@ -46,7 +46,7 @@ t.create('test coverage when outer user repos query returns multiple items')
               },
               latest_default_branch_test_report: {
                 data: {
-                  id: '6463c8a3e3bc340001004bce',
+                  id: '6598295abd91ec00013d485e',
                   type: 'test_reports',
                 },
               },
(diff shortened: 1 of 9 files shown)


PyvesB:
> but I can't work out how to get a working badge out of either of them.

For example: 

https://pr-9933-badges-shields.fly.dev/nexus/r/me.neznamy/tab-api?server=https%3A%2F%2Frepo.tomkeuper.com&nexusVersion=3

chris48s:
OK. Thanks for the pointer. I think I mostly figured it out :crossed_fingers:

services/nexus/nexus.tester.js
```diff
@@ -269,7 +267,7 @@ t.create('Nexus 3 - search snapshot version valid snapshot artifact')
 
 t.create('Nexus 3 - search snapshot version for artifact without snapshots')
   .get(
-    '/s/javax.inject/javax.inject.json?server=https://nexus.pentaho.org&nexusVersion=3',
+    '/s/com.tomkeuper/spigot.json?server=https://repo.tomkeuper.com&nexusVersion=3',
```
chris48s, line 270: This is the one I am least sure if I got right. It does appear to throw the right error message.

services/codeclimate/codeclimate-coverage.tester.js
```diff
@@ -46,7 +46,7 @@ t.create('test coverage when outer user repos query returns multiple items')
               },
               latest_default_branch_test_report: {
                 data: {
-                  id: '6463c8a3e3bc340001004bce',
+                  id: '65a1662cb0077b00013cb4de',
```
chris48s, line 49: I feel like this is not a sustainable solution for this test, but I think I'm not going to deal with re-writing this test in this PR as it is a bit of an yak-shave that isn't really related to the original point of this PR.

PyvesB (approved):
Thanks for updating the tests, I think all is good! 👍🏻

PR #9820, opened 2023-12-18
upgrade to docusaurus 3

chris48s:
Closes https://github.com/badges/shields/issues/9816

This PR upgrades 
- @docusaurus/core
- @mdx-js/react
- docusaurus-openapi and
- react/react-dom

and I've made all the necessary formatting changes to all of the documentation to make MDXv3 happy, display everything sensibly and fix all the `validateDOMNesting` warnings (mostly `<p> cannot appear as a descendant of <p>` and `<a> cannot appear as a descendant of <a>` warnings).

This PR is changing a lot at once, but I have split things out into commits so the functional changes are easy to separate from the minor alterations to documentation markup.

Relevant background reading:

- https://github.com/badges/shields/issues/9816
- https://docusaurus.io/docs/next/migration/v3
- https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3

One of the things that docusaurus 3 should allow us to do is migrate the frontend from CJS to ESM, but I've decided not to do that here because we don't have to do it in order to upgrade and the PR is already big. I'm going to punt that change to a follow up issue.

Issue #9563, opened 2023-09-10
Merge the concepts of Docker(Hub) and DockerCloud


chris48s:
In #9562 I started converting `examples` to `openApi` in the docker service.

One thing I realised while doing this is we have a mix of "Docker" badges (which are all really badges for DockerHub - none of them support alternative registries like GHCR). Then we've got two badges for Docker Cloud, which is not a distinct thing any more. DockerCloud was folded into DockerHub [years ago](https://www.docker.com/blog/the-new-docker-hub/). They both query the DockerCloud API, but it is just a redirect to the DockerHub API now.
Of the 2 DockerCloud badges, one of them (build status) is dinstinct to DockerCloud. The other one (build automation) also has a corresponding Docker(Hub) version, but they use different API endpoints.

- Review DockerCloud and Docker(Hub) build automation badges: Are they the exact same thing, or subtly different
- Remove the concept of DockerCloud in the docs, apply appropriate redirects
- Make it clear in the docs that all the "Docker" badges are really DockerHub badges

calebcartwright:
Linking a prior comment with relevant historical context:

https://github.com/badges/shields/issues/4663#issuecomment-588296373

chris48s:
Thanks. Useful link.
So basically we do need to continue to maintain 2 versions of the build automation badge..

PyvesB:
We've deprecated DockerCloud badges in #11501. Will go ahead and close this issue, as there's nothing left to merge per say.

PR #9934, opened 2024-01-29
migrate some services from examples to openApi part 47; affects [docker]

chris48s:

I'm going to cautiously say: I think this is the last one.

Once all the outstanding related PRs are merged (including https://github.com/badges/shields/pull/9932 ) I _think_ that should be every service moved over to using openApi, unless I missed a spot :crossed_fingers: and I can:

- Start on the cleanup necessary to remove all the compatibility stuff that allowed `examples` and `openApi` to exist side-by-side
- Revisit https://github.com/badges/shields/pull/9820 (it was too hard to work on with this still in flight)

Hopefully.

### DIFF
--- a/services/docker/docker-cloud-automated.service.js
+++ b/services/docker/docker-cloud-automated.service.js
@@ -1,4 +1,4 @@
-import { BaseJsonService } from '../index.js'
+import { BaseJsonService, pathParams } from '../index.js'
 import { dockerBlue, buildDockerUrl } from './docker-helpers.js'
 import { fetchBuild } from './docker-cloud-common-fetch.js'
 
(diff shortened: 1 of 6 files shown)


services/docker/docker-cloud-automated.service.js
```diff
@@ -13,17 +13,23 @@ export default class DockerCloudAutomatedBuild extends BaseJsonService {
     isRequired: false,
   }
 
-  static examples = [
-    {
-      title: 'Docker Cloud Automated build',
-      documentation: '<p>For the new Docker Hub (https://cloud.docker.com)</p>',
```
chris48s, line 19: I've decided to ignore https://github.com/badges/shields/issues/9563 here and just move what we've got. But Docker Cloud is definitely not "the new Docker Hub" :smile:

PyvesB (approved):
Nice one! 👍🏻

Issue #9950, opened 2024-02-03
Visual Studio App Center Releases Failing


steve09014098:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Attempting to use the page [Visual Studio App Center Release](https://shields.io/badges/visual-studio-app-center-releases) to generate a version number for my application in App Center.

In [Reqbin](https://reqbin.com/) I can retrieve a valid JSON file using the URL ```https://api.appcenter.ms/v0.1/apps/#org/#app/releases/latest``` (note I took this URL from the source code in this repository)

I generated an organization level token and attempted to use my valid Owner, App and Token to generate the badge but I get the an error badge that says ```Release - Invalid Response Data```

I use exactly the same details on this page [App Center Size](https://shields.io/badges/visual-studio-app-center-size) and it returns the correct size of the application.

Therefore I am certain that the data is correct, the API call is correct, the token, app and owner are correct but I assume since this feature was added the JSON response has changed slightly causing this to fail.

I notice that the code seems to generate a badge from both the short and long versions, but my short version string is empty

The URL I use is: ```https://api.appcenter.ms/v0.1/apps/#org/#app/releases/latest```
In Reqbin I set a header of "x-api-token:#token"
Clicking Execute gives me the following:

```
{
    "app_name": "###",
    "app_display_name": "###,
    "app_os": "Windows",
    "app_icon_url": "###",
    "owner": {
        "name": "###",
        "display_name": "###"
    },
    "is_external_build": false,
    "origin": "appcenter",
    "id": 1,
    "version": "v1.0.13.0",
    "short_version": "",
    "size": 596325657,
    "min_os": null,
    "device_family": null,
    "bundle_identifier": "44b8699fd595efe748522870b71ddbec",
    "fingerprint": "5fd96e7ab8cc2bec40158d19f027c966",
    "uploaded_at": "2024-02-02T14:34:29.872Z",
    "download_url": "###",
    "install_url": "###",
    "mandatory_update": false,
    "enabled": true,
    "fileExtension": "zip",
    "is_latest": true,
    "release_notes": "###",
    "package_hashes": ["b60b0b0c21c64b6cc95903c9119a1707920ffbefecec2e44a95fcb4de056a746"],
    "build": {
        "commit_hash": "370ce5dfcb7ec889d88ff3171ef25d1804d414e8"
    },
    "destination_type": "group",
    "status": "available",
    "distribution_group_id": "00000000-0000-0000-0000-000000000000",
[...]

PR #9951, opened 2024-02-05
tolerate missing short_version in [visualstudioappcenter]

chris48s:
Closes https://github.com/badges/shields/issues/9950


PyvesB (approved):
Thanks for fixing! 👍🏻

PR #9954, opened 2024-02-07
[NpmUnpackedSize] Unpacked Size Badge

### LINKED ISSUE #9952: A badge for "unpacked size" in NPM
### 📋 Description

NPM has unpacked size section in every library page.

unpacked size on npm website

However, the badge is not implemented.

### 🔗 Data

Unpacked size is available in registry API. [This](https://registry.npmjs.com/firereact) shows every version and [this](https://registry.npmjs.com/firereact/latest) shows only the latest version.

Unpacked size can be seen in "dist.unpackedSize", which is in bytes.

### 🎤 Motivation

A badge for unpacked size would be good as a developer might decide to include it in their front end project. Normally, they'd check this with bundlephobia, which is already implemented and only shows minified and min+gzipped sizes. The problem is, sometimes, due to the lack of entry point in `package.json`, bundlephobia will not respond with a size, see [react-firebase-hooks](https://bundlephobia.com/package/react-firebase-hooks).

Thus, the only remaining thing a developer can rely on would be unpacked size for this case.

+ would be better

erayerdin:
Closes #9952


erayerdin:
Some tests for NPM apart from what I've added fail oddly. I think it's due to requiring a token, maybe?

services/npm/npm-unpacked-size.service.js
```diff
@@ -0,0 +1,65 @@
+import Joi from 'joi'
+import prettyBytes from 'pretty-bytes'
+import { BaseJsonService, pathParam } from '../index.js'
+import { packageNameDescription } from './npm-base.js'
+
+export default class NpmUnpackedSize extends BaseJsonService {
```
erayerdin, line 6: The reason I've specifically used `BaseJsonService` instead of `NpmBase` is because I, for some reason, could not quite much understand it. It has its own NPM Registry parsing logic, some other services of NPM has different URLs, etc. It's been 2 days since I've opened the issue myself just to try to wrap my head around the existing NPM services implementing `NpmBase`. At the end of the day, I've found it much more easier and clean to implement `BaseJsonService` directly.

If, however, it is strictly required, I will try to implement it in `NpmBase` way.

erayerdin: actually, nvm, changing to `NpmBase` worked flawlessly

services/npm/npm-unpacked-size.service.js
```diff
@@ -0,0 +1,65 @@
...
+    const { dist } = await this.fetch({
+      packageName,
+      version: version ?? 'latest',
+    })
+    const { unpackedSize } = dist
+
+    return {
+      label: 'unpacked size',
+      message: unpackedSize ? prettyBytes(unpackedSize) : 'unknown',
+      color: unpackedSize ? 'blue' : 'lightgray',
```
chris48s, line 62: Can you provide any detail on when this value would or would not be populated in the response?

chris48s:
OK. the tests that are failing are the badges that get stats from npm-stat.com Looks like that service may have gone away. https://npm-stat.com/

This is unrelated to this PR, so lets ignore that for now. I'll make another issue. I've changed the PR title so we only run tests for the service added in this PR.

services/npm/npm-unpacked-size.service.js
```diff
@@ -0,0 +1,65 @@
...
+import prettyBytes from 'pretty-bytes'
+import { pathParam } from '../index.js'
+import NpmBase, { packageNameDescription } from './npm-base.js'
+
+export default class NpmUnpackedSize extends NpmBase {
+  static category = 'size'
+
+  static route = {
+    base: 'npm/unpacked-size',
+    pattern: ':packageName/:version*',
```
chris48s, line 11: The badge should accept a scoped or unscoped package name. Can you

- Fix the route + code (route should be `scope(@[^/]+)?/:packageName/:version*` )
- Add service tests for scoped package with and without version

services/npm/npm-unpacked-size.tester.js
```diff
@@ -0,0 +1,15 @@
+import { createServiceTester } from '../tester.js'
+
+export const t = await createServiceTester()
+
+t.create('Latest unpacked size')
+  .get('/express.json')
+  .expectBadge({ label: 'unpacked size', message: /.* kB/ })
```
chris48s, line 7: Can we use the `isFileSize` test validator for the valid tests instead of writing a custom regex please

services/npm/npm-unpacked-size.service.js
```diff
@@ -0,0 +1,65 @@
...
+            example: '4.18.2',
+          }),
+        ],
+      },
+    },
+  }
+
+  async fetch({ packageName, version }) {
+    return this._requestJson({
+      schema: Joi.any(),
```
chris48s, line 47: Instead of using `Joi.any()` here, can you write a schema that ensures the data we need exists in the response and is of the right type. There's some docs on input validation in https://github.com/badges/shields/blob/master/doc/input-validation.md . Not tested, but I guess either

```js
Joi.object({
  dist: Joi.object({
    unpackedSize: nonNegativeInteger.required(),
  }).required()
}).required()
```

or

```js
Joi.object({
  dist: Joi.object({
    unpackedSize: nonNegativeInteger
  }).required()
}).required()
```

will be what we want here, depending on how we want to handle the case where `unpackedSize` does not exist. One option would be to accept a missing value and handle it in the code (like you've done). the other option would be to reject the response at the validation stage. I guess that depends on the answer to https://github.com/badges/shields/pull/9954#discussion_r1485606960

chris48s:
Thanks for working on this btw. There's a few issues to iron out, but it is a good start :+1:

services/npm/npm-unpacked-size.service.js
erayerdin: - [Express 4.18.2 Registry Endpoint](https://registry.npmjs.org/express/4.18.2)
 - [Express 4.15.0 Registry Endpoint](https://registry.npmjs.org/express/4.15.0)

`4.15.0` does not have `dist.unpackedSize` while `4.18.2` or `latest` has it.

I think the field is populated when someone publishes a package on NPM, so some older versions of the same package (especially when the package is old) does not have `dist.unpackedSize`.

services/npm/npm-unpacked-size.tester.js
erayerdin: Wasn't aware it existed, I've checked other npm services and bundlephobia service for my code. Will do it.

services/npm/npm-unpacked-size.tester.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const t = await createServiceTester()
+
+t.create('Latest unpacked size')
+  .get('/express.json')
+  .expectBadge({ label: 'unpacked size', message: isFileSize })
+
+t.create('Nonexistent unpacked size with version')
+  .get('/express/4.16.0.json')
+  .expectBadge({ label: 'unpacked size', message: 'unknown' })
```
erayerdin, line 12: After adding `schema` to validate NPM registry response body, this test fails saying this:

```plain
5) NpmUnpackedSize
       [live] Nonexistent unpacked size with version
         
        [ GET /express/4.16.0.json ]:

      label mismatch
      + expected - actual

      -size
      +unpacked size
```

When I manually requested [the local dev endpoint](http://localhost:8080/npm/unpacked-size/express/4.16.0), I've seen the badge as such:

the badge with error

(could not paste the PNG of it as it's not available in the dev server)

I expected the label to be `unpacked size` instead of only `size`. It seems like, when the schema validation fails, it renders the label as `size`, which, I think, is a built-in mechanism.

I think when `dist.unpackedSize` is not defined, the label should still be `unpacked size` to imply that this is not any kind of other size.

@chris48s, do you have any idea as to how to override the default badge when `dist.unpackedSize` does not exist? I've tinkered a little bit with the codebase but I could not see any way (maybe I've missed it or haven't realized).

erayerdin: Actually, I've checked it through `BaseJsonService._requestJson` method, which calls `BaseService._validate`. `BaseService._validate` has positional `options` parameter, in which, `prettyErrorMessage` has been set to `"invalid response data"`.

All the way up the stack (from `NpmUnpackedSize` service), I cannot set `prettyErrorMessage` because it is internal to `BaseJsonService` and `BaseService`. Given that `BaseJsonService._requestJson` and `BaseService._validate` are internals (prepended by underscore), I naturally hesitate to change them.

Thus, there are two possible paths here:

1. Modify `BaseService._validate` and `BaseJsonService._requestJson` methods so that I can reach and modify the message (and thus, the label) when `dist.unpackedSize` does not exist. Given that these methods are defined as internals and tons of services rely on their signature, this might/would cause enormous amount of failures on some tests in the best case, or unexpected behavior on runtime at worst.
2. Accept the sweet fate and let the badge stay as `size: invalid response data`. I thought it wouldn't be a viable solution as (given that JS is used in web) it is important to differ unpacked size from minified or min+gzipped sizes but there's no sane room here. So, I can only modify the test to satisfy `size: invalid response data` badge.

@chris48s What's your idea?

services/npm/npm-unpacked-size.tester.js
erayerdin: Completed. Resolve if okay.

services/npm/npm-unpacked-size.service.js
erayerdin: Could not complete, refer to the review: https://github.com/badges/shields/pull/9954#discussion_r1486738206

services/npm/npm-unpacked-size.tester.js
chris48s: Ah right - sorry. In my original post, the suggestions should have been

`optionalNonNegativeInteger` and `nonNegativeInteger`
not
`nonNegativeInteger` and `nonNegativeInteger.required()`

:facepalm: 

if we switch to using `optionalNonNegativeInteger` for the `unpackedSize` key, then you should be able to carry on doing what you were doing and render "unknown" if the `unpackedSize` key is missing.

The validation is still useful, because it ensures the response has a `dist` object and that `unpackedSize` is of the right type if it exists.

If you want to set the label as "unpacked size" even when an exception is thrown, setting

```js
static defaultBadgeData = { label: 'unpacked size' }
```

will do the job.

Does that unstick you?

erayerdin: Thanks, did it. We definitely need documentation about these premade validators, missed it somehow. That's maybe the topic of another PR.

With that in mind, seems like this is completed.

services/npm/npm-unpacked-size.service.js
```diff
@@ -0,0 +1,83 @@
...
+            name: 'packageName',
+            example: 'npm',
+            description: packageNameDescription,
+          }),
+          pathParam({
+            name: 'version',
+            example: '4.18.2',
+          }),
+          queryParam({
+            name: 'registry_uri',
```
erayerdin, line 49: Also added a `registry_uri` query param as they were existent on the other NPM-based services.

services/npm/npm-unpacked-size.service.js
```diff
@@ -0,0 +1,83 @@
...
+  async fetch({ registryUrl, packageName, version }) {
+    return this._requestJson({
+      schema,
+      url: `${registryUrl}/${packageName}/${version}`,
+    })
+  }
+
+  async handle(namedParams, queryParams) {
+    const { scope, packageName, tag, registryUrl } =
+      this.constructor.unpackParams(namedParams, queryParams)
```
chris48s, line 68: I think you had this right first time. We now seem to be mixing the concepts of version and dist tag.
For some badges, we use a tag param, but that's not the same as a version. So for example

latest version of npm package: https://img.shields.io/npm/v/npm
latest version of npm package with dist-tag "next-8": https://img.shields.io/npm/v/npm/next-8

but its not the same as a version. What you're actually implementing here is a version param. i.e: the unpacked size of a specific version of this package. The thing we're accepting here isn't a dist tag.

I'd say just list the actual params in the signature of `handle()` rather than use the wrong helper.

services/npm/npm-unpacked-size.tester.js
chris48s: The validators are included in the internal api docs here: https://contributing.shields.io/module-services_validators.html

services/npm/npm-unpacked-size.service.js
erayerdin: Welp, I confused that. Sorry. Alright, let me get to reimplement version back.

erayerdin: Completed. Reverted back to using version instead of tags.

chris48s (approved):
cool - latest changes LGTM :+1:

Issue #9978, opened 2024-02-18
GitHub badges showing "Unable to select next Github token from pool"


jcs090218:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Similar to #8907.

### 🔗 Link to the badge

My repo https://github.com/emacs-eask/cli getting the error.

WSTxda:
same issue for me

repo: https://github.com/WSTxda/MicroG-RE

jcs090218:
Interesting... the repository https://github.com/emacs-eask/cli appears to no longer have the issue, yet now my other repository https://github.com/jcs090218/JCSUnity (which was previously functional) is encountering the same error. It appears to occur randomly. 🤔

WSTxda:
> This seems to be an issue related to GitHub's image caching system ("camo").
> 
> If you check https://github.com/m4heshd/ufc-ripper , the first badge is broken but if you visit the actual [URL pointing to the image](https://img.shields.io/github/downloads/m4heshd/ufc-ripper/total?logo=github&color=0344ab&style=for-the-badge), it works.

![Screenshot_2024-02-18-22-08-56-525_com.android.chrome.jpg](https://github.com/badges/shields/assets/37253994/5ccd3739-4872-445a-9495-b988fedf62ea)

Actually this doesn't work in the direct link either

m4heshd:
> Actually this doesn't work in the direct link either

Works for me. 🤔

![image](https://github.com/badges/shields/assets/6711514/1957e13b-5496-4d9a-9aff-831649175535)

Tried both with and without the browser cache.

WSTxda:
![Captura de tela 2024-02-18 221446](https://github.com/badges/shields/assets/37253994/d68888fb-6016-4173-823e-2960ab8eb451)

desktop also

i think its just randomly

m4heshd:
@WSTxda Hmm.. it might be request-based. Can you try with your browser's guest profile?

jcs090218:
It's very random. I can see some of them work and some of them not... 😕

007revad:
I just refreshed the readme page for one of my repos for a second it correctly showed the latest release version but then changed back to "Unable to select next GitHub token from pool"

chris48s:
Not exactly sure what happened here, but it looks like 3 of the servers in the cluster got themselves into a state where they hit a token that had maxed its rate limit and weren't able to recover and pull another token from the pool.

[...]

PR #9980, opened 2024-02-20
randomize tokens on token pool init

chris48s:
While I was worrying about https://github.com/badges/shields/issues/9978 I spent a bit of time digging into the token pool code.

I don't think this is related as such, but one of the things I realised is that when we bootstrap the server, all the VMs in the cluster load all the tokens in the same order.

Grabbing the tokens in a random order will do a better job of spreading the traffic out over the available tokens we have.


PyvesB (approved):
Nice improvement!

PR #9981, opened 2024-02-21
[GithubCreatedAt] Add Created At Badge for Github

### LINKED ISSUE #9754: github repository age
### 📋 Description

- github
- either the created on date itself, or something like "3 months" / "2 years"

### 🔗 Data

`https://api.github.com/repos/{:owner}/{:repository}` + `.created_at`

### 🎤 Motivation

this info isn't shown on github desktop by default and would be nice to show off the maturity of a project

erayerdin:
Closes #9754.

### DIFF
--- a/services/color-formatters.js
+++ b/services/color-formatters.js
@@ -5,8 +5,8 @@
  * @module
  */
 
-import dayjs from 'dayjs'
 import pep440 from '@renovatebot/pep440'
+import dayjs from 'dayjs'
 
 /**
  * Determines the color used for a badge based on version.
@@ -175,24 +175,26 @@ function colorScale(steps, colors, reversed) {
 /**
  * Determines the color used for a badge according to the age.
  * Age is calculated as days elapsed till current date.
- * The color varies from bright green to red as the age increases.
+ * The color varies from bright green to red as the age increases
+ * or the other way around if `reverse` is given `true`.
  *
  * @param {string} date Date string
+ * @param {boolean} reversed Reverse the color scale a.k.a. the older, the better
  * @returns {string} Badge color
  */
-function age(date) {
-  const colorByAge = colorScale([7, 30, 180, 365, 730], undefined, true)
+function age(date, reversed = false) {
+  const colorByAge = colorScale([7, 30, 180, 365, 730], undefined, !reversed)
   const daysElapsed = dayjs().diff(dayjs(date), 'days')
   return colorByAge(daysElapsed)
 }
 
 export {
-  version,
-  pep440VersionColor,
-  downloadCount,
+  age,
+  colorScale,
   coveragePercentage,
+  downloadCount,
   floorCount,
   letterScore,
-  colorScale,
-  age,
+  pep440VersionColor,
+  version,
 }
--- a/services/github/github-created-at.service.js
+++ b/services/github/github-created-at.service.js
@@ -1,10 +1,12 @@
+import dayjs from 'dayjs'
 import Joi from 'joi'
 import { pathParams } from '../index.js'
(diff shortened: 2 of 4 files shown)


services/github/github-created-at.service.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+  }
+
+  static defaultBadgeData = { label: 'created at' }
+
+  static render({ createdAt }) {
+    const date = dayjs(createdAt)
+    return {
+      message: formatDate(date),
+      color: age(date),
```
erayerdin, line 41: This `age` function renders very old repos in red color. I'm not sure if that is good.

services/github/github-created-at.tester.js
```diff
@@ -0,0 +1,9 @@
+import { createServiceTester } from '../tester.js'
+import { isFormattedDate } from '../test-validators.js'
+
+export const t = await createServiceTester()
+
+t.create('created at').get('/erayerdin/firereact.json').expectBadge({
+  label: 'created at',
+  message: isFormattedDate,
+})
```
chris48s, line 9: Could we add a test for the repo not found case please?

services/github/github-created-at.service.js
chris48s: This helper normally gets used for "last commit" or "last release" badges where recent is generally seen as good. I think for age of the repo we could either just present it in blue (informational) or make an inverse colour scheme for this badge (repos that have been around a long time are "good" whereas repos that have just been created are "bad"). My guess is the reason why you (or another user) wants this badge is to indicate their project has been around a long time, rather than draw attention to the fact they just created it.

erayerdin: I think it would be good idea to have an *inversed* color range on `age`. I'll go with adding an `inverse` parameter with a default value of `false` on `age` function initially. That's the only idea I have.

```javascript
function age(date, inverse = false) {
  // ... implementation ...
}
```

If you have another idea, hit me.

services/github/github-created-at.tester.js
erayerdin: Completed.

services/github/github-created-at.service.js
erayerdin: Completed.

services/color-formatters.js
```diff
@@ -175,24 +175,26 @@ function colorScale(steps, colors, reversed) {
...
+ * @param {boolean} reversed Reverse the color scale a.k.a. the older, the better
  * @returns {string} Badge color
  */
-function age(date) {
-  const colorByAge = colorScale([7, 30, 180, 365, 730], undefined, true)
+function age(date, reversed = false) {
+  const colorByAge = colorScale([7, 30, 180, 365, 730], undefined, !reversed)
   const daysElapsed = dayjs().diff(dayjs(date), 'days')
   return colorByAge(daysElapsed)
 }
```
erayerdin, line 189: Added a `reversed` parameter with default value to `false`, which is passed to `colorScale` function by reversing in again with `!reversed`. I've thought re-reversing this boolean value is appropriate, semantically-speaking.

services/github/github-created-at.service.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+  }
+
+  static defaultBadgeData = { label: 'created at' }
+
+  static render({ createdAt }) {
+    const date = dayjs(createdAt)
+    return {
+      message: formatDate(date),
+      color: age(date, true),
```
erayerdin, line 41: Inversed the color of `age` as _the older, the better_.

chris48s (approved):

chris48s:
Nice. Latest changes look good :+1:

PR #9984, opened 2024-02-23
Added custom bucket url support for [Scoop]

### LINKED ISSUE #9982: Custom bucket URLs for Scoop version shield.
:clipboard:
In addition to referencing Scoop's [bucket list](https://github.com/ScoopInstaller/Scoop/blob/master/buckets.json), shields could take in a URL as a bucket parameter for the Scoop version shield.

After `if (!bucketUrl)`, before throwing the error, shields could check if the bucket parameter is a valid GitHub URL, and attempt to extract the user & repo from that, using the same `gitHubRepoRegExp`, allowing users to reference buckets that are not in the official list.

jewlexx:
Closes #9982

### DIFF
--- a/services/scoop/scoop-version.service.js
+++ b/services/scoop/scoop-version.service.js
@@ -38,7 +38,12 @@ export default class ScoopVersion extends ConditionalGithubAuthV3Service {
           '[Scoop](https://scoop.sh/) is a command-line installer for Windows',
         parameters: [
           pathParam({ name: 'app', example: 'ngrok' }),
-          queryParam({ name: 'bucket', example: 'extras' }),
+          queryParam({
+            name: 'bucket',
+            description:
+              "App's containing bucket. Can either be a name (e.g `extras`) or a URL to a GitHub Repo (e.g `https://github.com/jewlexx/personal-scoop`)",
+            example: 'extras',
+          }),
         ],
       },
     },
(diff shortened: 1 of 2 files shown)


services/scoop/scoop-version.service.js
```diff
@@ -60,9 +61,31 @@ export default class ScoopVersion extends ConditionalGithubAuthV3Service {
...
+        // The error messages here are purely for code readability, and will never reach the user.
+        if (url.hostname !== 'github.com') {
+          throw new Error('Not a GitHub URL')
+        }
+        const path = url.pathname.split('/')
+        // Remove the empty string at the beginning
+        path.shift()
+        if (path.length !== 2) {
+          throw new Error('Not a valid GitHub Repo')
+        }
```
chris48s, line 80: Could we make it so this will tolerate URLs with a trailing slash
i.e: both https://github.com/badges/shields and https://github.com/badges/shields/ should be accepted.

chris48s:
Could you add a parameter description to the `bucket` `queryParam` explaining that `bucket` can either be:

- one of the named buckets from https://github.com/ScoopInstaller/Scoop/blob/master/buckets.json e.g: `extras`
- the URL to a GitHub repo e.g: `https://github.com/jewlexx/personal-scoop`

Here's an example of what that looks like:

https://github.com/badges/shields/blob/8ab9dfa9a11d52d730bb86aff6847165ffb4f25e/services/github/github-issues-search.service.js#L9-L13
https://github.com/badges/shields/blob/8ab9dfa9a11d52d730bb86aff6847165ffb4f25e/services/github/github-issues-search.service.js#L98-L103

jewlexx:
All done, let me know if there are any issues.

chris48s (approved):
Awesome - thanks :+1: 
Great first contribution

PR #9995, opened 2024-02-26
[GITEA]  Add last commit badge

CanisHelix:
Adds a 'Last Commit' badge with 2 route mirroring the options provided by Github which is the basis for the Gitea implementation also (they strive to be close to Github's API usage).

I believe this should be enough to close issue #4040 unless there are more badges needed before the service is considered implemented.


chris48s (approved):
looks great - thanks

chris48s:
> I believe this should be enough to close issue https://github.com/badges/shields/issues/4040 unless there are more badges needed before the service is considered implemented.

I think we've got plenty tbh. Lets close that issue. If anyone is looking for something we don't have, they can open a new issue.

Issue #9997, opened 2024-02-27
Add support for Youtrack Issues/Cards


CanisHelix:
### 📋 Description

Add support for Issue Counts from JetBrains YouTrack product. Given the customization capabilities of YouTrack on a per project level a standard Open/Closed/All would not suffice. As such the filter would need to be defined as a `query` to correctly state the YouTrack search criteria. i.e. `#Unresolved Type:Bug` would be a query getting all open bugs in my project, but someone else could need `#Open Type:Issue` based on their setup.

Project is the only guarantee, so this could be used in a path routing similar to user/repo that VCS issues badges work.

### 🔗 Data

An appropriate endpoint is available at: https://www.jetbrains.com/help/youtrack/devportal/resource-api-issuesGetter-count.html

For a private project either self-hosted (depending on the level of security configured) or on their cloud solution, a token would be required.

### 🎤 Motivation

Many VCS systems now offer an 'External Issues' feature, wherein the Issues don't exist in the VCS like GitHub does, but can exist in a third party external system such as YouTrack. Gitlab/Gitea/Forgejo support this and even translate `#123` etc into correct URL's to take them directly into the external issue management tool.

Projects using this feature would not find any benefit of issues badges for the VCS they use, and would not be able to show issues as badges on their websites/readme's.

chris48s:
OK cool thanks.

Looking at the docs, there don't seem to be any documented rate limits and we haven't had any issues with rate limits on other jetbrains APIs.

In terms of the authentication

https://www.jetbrains.com/help/youtrack/devportal/yt-api-headers.html

seems to say that authorization is needed for _all_ requests. However, it looks like in practice it is actually possible to get some data out about public projects without auth. IS that documented anywhere?

For private or self-hosted projects, how are you envisaging the auth could work? There's some documentation on the constraints in

https://github.com/badges/shields/blob/master/doc/authentication.md

CanisHelix:
Regarding the headers, I have reached out to their support department who responded with the following:

> Without the authorization header, a request is made on behalf of the [guest](https://www.jetbrains.com/help/youtrack/server/managing-guest-users.html) user. This user represents anonymous access, which is enabled in [youtrack.jetbrains.com](http://youtrack.jetbrains.com/) to allow unregistered users to check public issues.

[...]

PR #10011, opened 2024-03-05
[YOUTRACK] Add YouTrack Issues

CanisHelix:
Add support for YouTrack Cards/Issues as per #9997

A new Service Token and test is added, InvalidResponse of processing added when a -1 is returned, and cached for 10 seconds. This is expected when YouTrack isn't finished and is documented at https://www.jetbrains.com/help/youtrack/devportal/resource-api-issuesGetter-count.html#create-IssueCountResponse-method
> If the response returns "count": -1, it means that YouTrack hasn't finished counting the issues yet. Wait for a bit and repeat the request.

For CI purposes shields.youtrack.cloud was created and is used throughout all tests as this instance is under our control. The guest user is enabled read-only so that the CI tests can work in Github Actions without needing a token, but prevents people breaking the tests.

Only concern is whether or not cachedSeconds is working as IntelliJ IDE reports this doesn't match InvalidResponses definitions, but seems to work. And also, during testing when the cloud instance is idle for long periods, a timeout is received on the first test of a day, as such I added 'timeout' as a possible test response.

### DIFF
--- a/core/server/server.js
+++ b/core/server/server.js
@@ -142,6 +142,7 @@ const publicConfigSchema = Joi.object({
     sonar: defaultService,
     teamcity: defaultService,
     weblate: defaultService,
+    youtrack: defaultService,
     trace: Joi.boolean().required(),
   }).required(),
   cacheHeaders: { defaultCacheLengthSeconds: nonNegativeInteger },
@@ -205,6 +206,7 @@ const privateConfigSchema = Joi.object({
   influx_password: Joi.string(),
   weblate_api_key: Joi.string(),
   youtube_api_key: Joi.string(),
+  youtrack_token: Joi.string(),
 }).required()
 const privateMetricsInfluxConfigSchema = privateConfigSchema.append({
   influx_username: Joi.string().required(),
--- a/doc/server-secrets.md
+++ b/doc/server-secrets.md
@@ -364,6 +364,15 @@ and create an API key for the YouTube Data API v3.
 
 [youtube credentials]: https://console.developers.google.com/apis/credentials
 
+### YouTrack
+
+- `YOUTRACK_ORIGINS` (yml: `public.services.youtrack.authorizedOrigins`)
+- `YOUTRACK_TOKEN` (yml: `private.youtrack_token`)
+
+A Youtrack [Permanent Access Token][youtrack-pat] is required for accessing private content. If you need a Youtrack token for your self-hosted Shields server then we recommend limiting the scopes to the minimal set necessary for the badges you are using.
+
+[youtrack-pat]: https://www.jetbrains.com/help/youtrack/devportal/Manage-Permanent-Token.html
+
 ## Error reporting
 
 - `SENTRY_DSN` (yml: `private.sentry_dsn`)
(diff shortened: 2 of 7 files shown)


chris48s:
So I checked this out locally.

Before I give a round of comments on the diff, I tried requesting
http://localhost:8080/youtrack/issues/DEMO?youtrack_url=https%3A%2F%2Fshields.youtrack.cloud&query=bug%20%23Unresolved
a bunch of times. For me, this just says `processing` every time - just a `-1` back from the API every time.

Do you get the same?

In order for this to be a viable badge, I think we can work round a bit of flakyness, but it is going to _mostly_ need to return a valid badge.

I don't want to spend too much time bikeshedding the code if this just isn't going to work reliably.

CanisHelix:
It seems the YouTrack API is a little flaky there. The word 'bug' doesn't exist in any of the cards so instead of returning 0 it returns -1 constantly.

If I use `Type: Bug #Unresolved` instead of `Bug #Unresolved` to search for bug type cards it works.

`http://localhost:8080/youtrack/issues/DEMO?youtrack_url=https%3A%2F%2Fshields.youtrack.cloud&query=Type%3A%20Bug%20%23Unresolved`

Returns 0 for me. 



Searching for `manage #Unresolved` which translates to URL `http://localhost:8080/youtrack/issues/DEMO?youtrack_url=https%3A%2F%2Fshields.youtrack.cloud&query=manage%20%23Unresolved` returns 7 for me.



I'll update the example as that was bad and an oversight.

CanisHelix:
If I did not sink time into this, we would not know about how flaky their API is. It's been a good learning curve of some areas of Shields new to me too.

YouTrack doesn't have a real DB for its backend which may be why they do not cache very well. In the current state of their API responses, I would agree that this is not good enough for a README.md.

In general they advice retrying after a few seconds with a -1, which I believe is okay for an application that would use their API. But badges needs to be quick. I did consider one last idea which was the following logic:

1) Call API
2) If not -1, render. If -1, step 3
3) Wait x seconds and retry
4) Loop a maximum x times before giving up as invalid.

Given that -1 can be returned instead of 0 due to just a bad query (syntax wise valid, but data wise invalid), I was hesitate to implement this logic to avoid an infinite loop scenario. With the x retries attempts, I am concerned about the negative performance it would have on shields.io. If a retry mechanic is ok, I can give that a shot. If not, we can close this out due to upstream not being reliable enough at this time.

chris48s:
Before we go down that route, do you have an example of a project/query which involves searching a large number of issues to evaluate the query?

I'd like to get an idea of how long/short an amount of time we would need to wait before retrying in that case before we put more time into coding solutions.

CanisHelix:
After testing and discussion on discord this PR will go into Draft Mode until such a time that YouTrack's API is more reliable. In the current state an API can be processing anywhere from 1 to 20 seconds, this is not an acceptable response time from the upstream API. I will leave my branch up and periodically update it. Hopefully JetBrains improve their API in the future.

PyvesB:
There hasn't been any movement for almost two years, there's not much point in keeping the PR open. Feel free to reopen whenever JetBrains gets rounds to improving their API. :)

PR #10041, opened 2024-03-22
[GithubLastCommit] [GitlabLastCommit] [GiteaLastCommit] Support file path for last commit

### LINKED ISSUE #7351: Last Commit based on File/Folder in repo
:clipboard: **Description**

A badge depicting when a specific folder or project within a repository was updated or committed to.

:link: **Data**

https://github.com/GamingInfinite/KadeEngineMods

take my repo for instance.  I have a mod in there that is in a folder named MFM.  In order to get the last commit to that folder, I can go to this API link:
`https://api.github.com/repos/GamingInfinite/KadeEngineMods/commits?path=MFM`
after that, you can take the commit in slot 0, go to committer (or author), and in there is a data tag listed as date.  Both in committer and author, it is the same date.

:microphone: **Motivation**

On the readme of my repo (listed in the above Data section as an example), I host several mods that I created that were ported using [Kade Engine](https://github.com/KadeDev/Kade-Engine)'s (relatively) new ModCore feature.  I did this as it didn't really make sense to me to make a separate repo for each one as all of them are ports (so far).  I host download links on the readme as well, as I don't really feel comfortable doing a full release of all of them, or even individual releases as whenever I make a large update I'd have to do a new release and that takes time.  So I just opted to make it where (since the GitHub already has the mods in downloadable form) there are download links to the GitHub itself (through downgit) to save myself time.  The only other thing that I added to the readme as something more for me than anyone else was when it was last updated, but that also takes me going through each time I modify it to replace the date that it was updated on the readme, and doing a readme bump every time was a hassle, and I just did one massive one since I'd neglected to add the mod links anyway and updated all of the dates as well.  Having an automated process to show the date would save me a lot of time, and I know a few other repos that could implement this shield if they wanted to as well.

ReubenFrankel:
Closes #7351

Adds support to resolve last commit for a given file path, for GitHub, GitLab and Gitea services.


services/gitea/gitea-last-commit.tester.js
```diff
@@ -8,6 +8,41 @@ t.create('Last Commit (recent)').get('/gitea/tea.json').expectBadge({
...
+  })
+
+t.create('Last Commit (recent) (nested dir path)')
+  .get('/gitea/tea.json?path=docs/CLI.md')
+  .expectBadge({
+    label: 'last commit',
+    message: isFormattedDate,
+  })
+
+t.create('Last Commit (recent) {path)')
```
chris48s, line 39:
```suggestion
t.create('Last Commit (recent) (path)')
```

chris48s:
Thanks. This looks good, but I think we could do with improving the error handling/messages slightly.

On the GitLab and GitHub APIs, if we specify a path that doesn't exist, we get back an empty array. This means the response fails schema validation.
Tbh, I suspect this probably also happens for a repo with zero commits, but this is not something that has come up before.
It would be nice to throw a more informative error in this cases. My suggestion would be we remove the `.min(1)` requirement from the schema and then after we fetch we throw a `NotFound('no commits found')` if there are zero items in the array (to cover both situations).

With the Gitea API, if the path doesn't exist, we get a 404 back and throw the error "user or repo not found". I think for this case, we can just amend the error message to "user, repo or path not found"

We should also have test cases for the errors.

chris48s:
Just looking over your other similar PR for bitbucket, you've already adopted the same pattern I described for the error handling there :+1:

ReubenFrankel:
@chris48s I've made the suggested changes - tests are passing locally so not sure why they are failing in CI (can't see from the log).

chris48s:
Thanks. This looks good.

Looks like the test failures were just a transient failure with some requests timing out.

chris48s (approved):

Issue #6760, opened 2021-07-12
support custom pypi index


LeoQuote:
:clipboard: **Description**
I set up an internal pypi repo , this repo support the identical api with pypi.org , can I create a badge for a custom pypi ?

I cannot see any parameter I can pass in the pypi badge.

calebcartwright:
Could you please expand on which specific badges you're referring to (e.g. version vs. license vs. downloads vs. etc.)? Shields always has to reach out to some upstream API to get the required data needed for building a badge, and various badges, even those under the same "umbrella", use different APIs and sometimes even different API providers.

LeoQuote:
Thanks for you reply. For now the version badge is most used, it would be good if sheild can support self-hosted pypi repo!

chris48s:
I think it would be possible to do this for all the pypi badges except for downloads in a similar way to how we allow a `registry_uri` param for NPM.
One thing to note: If we added this feature to shields.io it would only be possible for the alternate pypi registry to be a public registry that does not require authentication. For a non-public registry you would have to host your own instance of shields in order to address it.

LeoQuote:
certainly, non-public registry cannot be reached from shields.io

PR #10044, opened 2024-03-24
support setting pypiBaseUrl by environment variables and queryParameters; affects [pypi]

LeoQuote:
fix https://github.com/badges/shields/issues/6760

This pr make pypiBaseUrl configurable with both config.yaml and queryParameter.

### DIFF
--- a/doc/server-secrets.md
+++ b/doc/server-secrets.md
@@ -283,6 +283,13 @@ The Pepy API requires authentication. To obtain a key,
 Create an account, sign in and obtain generate a key on your
 [account page](https://www.pepy.tech/user).
 
+### Pypi
+
+- `PYPI_URL` (yml: `public.pypi.baseUri`)
+
+`PYPI_URL` can be used to optionally send all the Pypi requests to a Self-hosted Pypi registry,
+users can also override this by query parameter `pypiBaseUrl`.
+
 ### SymfonyInsight (formerly Sensiolabs)
 
 - `SL_INSIGHT_USER_UUID` (yml: `private.sl_insight_userUuid`)
--- a/doc/server-secrets.md
+++ b/doc/server-secrets.md
@@ -283,11 +283,11 @@ The Pepy API requires authentication. To obtain a key,
 Create an account, sign in and obtain generate a key on your
 [account page](https://www.pepy.tech/user).
 
-### Pypi
+### PyPI
 
 - `PYPI_URL` (yml: `public.pypi.baseUri`)
 
-`PYPI_URL` can be used to optionally send all the Pypi requests to a Self-hosted Pypi registry,
+`PYPI_URL` can be used to optionally send all the PyPI requests to a Self-hosted Pypi registry,
 users can also override this by query parameter `pypiBaseUrl`.
 
 ### SymfonyInsight (formerly Sensiolabs)
--- a/services/pypi/pypi-framework-versions.service.js
+++ b/services/pypi/pypi-framework-versions.service.js
@@ -1,5 +1,5 @@
 import { InvalidResponse, pathParams } from '../index.js'
-import PypiBase from './pypi-base.js'
+import PypiBase, { pypiGeneralParams } from './pypi-base.js'
 import { sortPypiVersions, parseClassifiers } from './pypi-helpers.js'
 
 const frameworkNameMap = {
@@ -56,14 +56,12 @@ export default class PypiFrameworkVersion extends PypiBase {
       get: {
         summary: 'PyPI - Versions from Framework Classifiers',
         description,
-        parameters: pathParams(
-          {
+        parameters:
+          pathParams({
             name: 'frameworkName',
             example: 'plone',
             schema: { type: 'string', enum: Object.keys(frameworkNameMap) },
-          },
-          { name: 'packageName', example: 'plone.volto' },
-        ),
+          }) + pypiGeneralParams,
       },
     },
   }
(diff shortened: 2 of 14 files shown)


chris48s:
A few thoughts on this:

Usually if we implement a param like this, we do it as a query param. There are a couple of reasons why this is useful:

- If we implement it as a setting, one self-hosted instance of shields can only produce badges for one registry. If you want to communicate with multiple custom registries or you want your self-hosted install to be able to serve badges for packages hosted on pypi.org and also a self-hosted registry, a global setting prevents both of those.
- Implementing it as a query param allows users of shields.io to generate badges for custom packages hosted on other public registries. While I agree that the most common use-case for this would be a custom registry that is not internet addressable, however public registries of python packages that are not pypi.org do exist. For example:
    - https://download.pytorch.org/whl
    - https://pypi.chia.net/simple/
    - https://pypi.vnpy.com/
    - https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/
    so I don't think I'd completely write it off as a possible use-case.

However, that brings me on to the second point, which - I must admit - I didn't realise a few years back when https://github.com/badges/shields/issues/6760 was first raised...

The way our pypi badges work, they make use of the [Warehouse JSON API](https://warehouse.pypa.io/api-reference/json.html)

In my experience, most custom python package registries (this applies all of the ones I've listed above as examples, and I think this would also apply to the kind of custom registry you might stand up using something like Nexus or Artifactory) implement the simple registry protocol, but not the warehouse JSON API. You'd need a full install of warehouse for that.

This is also the case for [PiWheels](https://www.piwheels.org/), which is the most obvious and prominent public registry of python packages that is not PyPI. Again, they implement the simple registry protocol, but have a completely custom API format which is not compatible with PyPI's. So we have different badges for PiWheels - we can't just treat them as an alternate registry (even though python package managers like pip can).

Given that, I wonder how useful being able to point at a custom URL actually is? My gut instinct is that the majority of custom python package registries are not a full install of warehouse. I guess the reason why you want this is because you have a custom PyPI registry - yes? Assuming that is the case, does yours support the warehouse JSON API?

LeoQuote:
Totally understand your point.

Firstly, I raised this issue because I set up a private pypi registry internally, it supports full json api, so you don’t have to worry about the api compatibility issue, in fact, I forked the repo and change pypi.org to my registry and it just worked for years.

Secondly, this issue has been raised for years while no one really implemented it, I think part of the reason is what you mentioned, the most public pypi registry does not provide json api except for pypi.org , and most people would use pypi.org only.

So overall this use case would be used by self-host users, to serve for their own registry, and most likely, there would be only one pypi registry inside an intranet. If someone want a badge for package on pypi.org , they can just use the public service.

it could be implemented with query string, but it won’t be necessary for me, like I said before, people can use the public shield service if they want. If someone want this feature, it could be implemented by him/her

LeoQuote:
@chris48s Hi, I submitted another draft for this case, I implemented the queryParameter function which is identical with npm service, and also provide a way for server owner to set the default pypiBaseUrl. Hope that would be acceptable for you.

PTAL, thank you.

services/pypi/pypi-framework-versions.service.js
```diff
@@ -80,15 +80,15 @@ export default class PypiFrameworkVersion extends PypiBase {
     }
   }
 
-  async handle({ frameworkName, packageName }) {
+  async handle({ frameworkName, packageName }, { pypiBaseUrl }) {
```
chris48s, line 83: we'll need to also add `pypiBaseUrlParam` to the openApi spec on this service

chris48s (commented):
Thanks. I've left one comment inline. Also, a couple of additional things:

1. Can you also apply this to `services/pypi/pypi-downloads.service.js`

2. Can you add docs on the PYPI_URL setting to https://github.com/badges/shields/blob/master/doc/server-secrets.md

services/pypi/pypi-framework-versions.service.js
LeoQuote: Thanks for your check, this is done now.

LeoQuote:
> 1. Can you also apply this to `services/pypi/pypi-downloads.service.js`

`pypi-downloads` is provided by a third-party service, not pypi.org, changing the baseUrl for downloads cound would not work, I think this feature would not be used by self-host users

doc/server-secrets.md
```diff
@@ -283,6 +283,13 @@ The Pepy API requires authentication. To obtain a key,
 Create an account, sign in and obtain generate a key on your
 [account page](https://www.pepy.tech/user).
 
+### Pypi
+
+- `PYPI_URL` (yml: `public.pypi.baseUri`)
+
+`PYPI_URL` can be used to optionally send all the Pypi requests to a Self-hosted Pypi registry,
+users can also override this by query parameter `pypiBaseUrl`.
+
```
chris48s, line 292:
```suggestion
### PyPI

- `PYPI_URL` (yml: `public.pypi.baseUri`)

`PYPI_URL` can be used to optionally send all the PyPI requests to a Self-hosted Pypi registry,
users can also override this by query parameter `pypiBaseUrl`.

```

chris48s:
Thanks - yes you're right. I'd lost track of the fact we use pypistats for PyPI. My mistake :facepalm: 

I've made a really small change to the docs - just the way we write PyPI. Other than that, LGTM. Merging.

Ta

chris48s (approved):

PR #10043, opened 2024-03-23
[BitbucketLastCommit] Add Bitbucket last commit

ReubenFrankel:
Add support for Bitbucket last commit: https://developer.atlassian.com/cloud/bitbucket/rest/api-group-commits/#api-repositories-workspace-repo-slug-commits-revision-get

PR #10046, opened 2024-03-25
improve performance of [GithubLastCommit] [GitlabLastCommit] [GiteaLastCommit]

chris48s:
This is something I noticed while I was testing https://github.com/badges/shields/pull/10041
It isn't a new problem introduced by that PR, so I decided not to scope creep it there.

Basically the deal here is we only care about the most recent commit, but we're requesting an entire page of results to just use the first object. Only requesting the most recent one means we can request less data from the upstream and render the badge faster. Note this optimisation doesn't work in all situations. Specifically, Gitea ignores limit when the path param is also supplied (see https://docs.gitea.com/api/1.20/#tag/repository/operation/repoGetAllCommits ), but we might as well do this whenever we can.

Note that @ReubenFrankel has already applied the equivalent optimisation in https://github.com/badges/shields/pull/10043 :+1: which adds the corresponding BitBucket badges.

### DIFF
--- a/services/github/github-last-commit.service.js
+++ b/services/github/github-last-commit.service.js
@@ -98,7 +98,7 @@ export default class GithubLastCommit extends GithubAuthV3Service {
   async fetch({ user, repo, branch, path }) {
     return this._requestJson({
       url: `/repos/${user}/${repo}/commits`,
-      options: { searchParams: { sha: branch, path } },
+      options: { searchParams: { sha: branch, path, per_page: 1 } },
       schema,
       httpErrors: httpErrorsFor(),
     })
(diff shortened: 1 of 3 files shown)


services/github/github-last-commit.service.js
```diff
@@ -106,7 +106,7 @@ export default class GithubLastCommit extends GithubAuthV3Service {
 
   async handle({ user, repo, branch }, queryParams) {
     const { path, display_timestamp: displayTimestamp } = queryParams
-    const body = await this.fetch({ user, repo, branch, path })
+    const body = await this.fetch({ user, repo, branch, path, per_page: 1 })
```
ReubenFrankel, line 109: Does this not need to go in `fetch` [here](https://github.com/chris48s/shields/blob/87b9049eb14a7ac71fd58cea8f652e2d0a296015/services/github/github-last-commit.service.js#L101) like your GitLab change?

chris48s: good catch! thanks for the review

PR #10062, opened 2024-04-04
Perf: librariesio repo dependencies

tradiff:
Hello! I work on libraries.io. The endpoint used by shields for the librariesio repository dependencies badge is pretty expensive and can return a large payload, most of which was not useful to shields. Not to mention additional work is being incurred on your end.

We have created a new dedicated endpoint which returns only the two values shields cares about, and should have a much quicker response time.  This PR switches to the new endpoint.

### DIFF
--- a/services/librariesio/librariesio-dependencies.service.js
+++ b/services/librariesio/librariesio-dependencies.service.js
@@ -1,4 +1,5 @@
 import Joi from 'joi'
+import { nonNegativeInteger } from '../validators.js'
 import { pathParams } from '../index.js'
 import LibrariesIoBase from './librariesio-base.js'
 import {
@@ -6,7 +7,7 @@ import {
   renderDependenciesBadge,
 } from './librariesio-dependencies-helpers.js'
 
-const schema = Joi.object({
+const projectDependenciesSchema = Joi.object({
   dependencies: Joi.array()
     .items(
       Joi.object({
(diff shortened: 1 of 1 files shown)


chris48s (approved):

chris48s:
Hi.
Thanks for setting this up for us.

The service tests failed because we hit the rate limit and PRs from forks can't use a token (we don't expose tokens to PRs from forks) but I've checked the service tests locally and they passed.

While I have got the ear of a dev at libraries.io, the biggest problem we have with libraries.io at the moment is actually rate limiting. We don't do a vast amount of traffic against your API, but we do enough requests that we sometimes exceed the rate limits using a single token, leading to broken badges for users. For example, in the last 24 hours, we've had 475 `429 Too Many Requests` responses from libraries.io. We find that because the rate limit window is quite small (1 minute) its quite hard to stay under it because it only takes quite a small spike in traffic to exceed it.

We used to handle this by using 2 tokens in production and rotating between them. 2 was enough to consistently stay under the rate limit. Unfortunately, we are now no longer able to do this because we can't track the rate limit remaining for a given token when that token is shared across multiple servers. I raised an issue about this over at: https://github.com/librariesio/libraries.io/issues/3283

Given you:

- work for libraries.io
- are interested in shields being able to serve badges that report data from libraries.io :heart: 

Do you think there is any chance that either:

- https://github.com/librariesio/libraries.io/issues/3283 could be fixed (which would allow us to go back to using 2 tokens in rotation)
- We could be issued a token with a slightly higher rate limit
- You could provide some info on how often data is updated that might help us to stay under using caching.

:pray:

Issue #9839, opened 2023-12-23
Unable to select next Libraries.io token from pool


chris48s:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Libraries.io badges frequently show

Unable to select next Libraries.io token from pool

### 🔗 Link to the badge

example:

- https://img.shields.io/librariesio/dependents/npm/got
- ![](https://img.shields.io/librariesio/dependents/npm/got)

(but affects all libraries.io badges)

### 💡 Possible Solution

I think there's 2 issues we need to fix here:

1. We are doing enough traffic on libraries.io that we are sometimes exhausting the rate limit on our available tokens. We either need to add another token to the pool or cache for longer to try and fix this.

2. When we _do_ hit the point where none of the tokens in the pool have any rate limit points left, we don't seem to recover from it well. In reality, if we wait for a bit, the next hour will tick over and we can start using the tokens again, but once we exhaust the pool once we're basically just stuck forever (or until the server cycles). We need to be more resilient to recovering from this state if we do hit it.

chris48s:
OK, so I did a bit of digging into this. I've found at least 2 problems. One of which we can solve, one of which we can't:

## Problem 1

When we calculate `nextReset` in

https://github.com/badges/shields/blob/4ba86170a7bc8e4b98da5b501d7aca68ca89e93e/services/librariesio/librariesio-api-provider.js#L51

we are calculating a number of milliseconds.

In

https://github.com/badges/shields/blob/4ba86170a7bc8e4b98da5b501d7aca68ca89e93e/core/token-pooling/token-pool.js#L75-L77

we are comparing `nextReset` to a number of seconds.

So we're never going to reset the token because we're trying to compare a number like `1703797975` (now) to a number like `1703797975973` (nextReset).

This has probably been a bug in our code for a eery long time. We can fix this by setting

```js
const nextReset = ((Date.now() + (retryAfter ? +retryAfter * 1000 : 0)) / 1000) >>> 0
```

in `getRateLimitFromHeaders()` or keeping that as it is and comparing it to `Date.now()` instead of `(Date.now() / 1000) >>> 0` or whatever.

## Problem 2

[...]

PR #10074, opened 2024-04-05
tweaks to libraries.io token pooling code

chris48s:
Refs #9839

Couple of things here: 1 is a fix for the mismatch in units comparing times
The other is: The rate limit on one of these tokens is 60/minute, so it feels like 45 is too big a batchSize. Reduce that so we switch every 10 requests,

### DIFF
--- a/services/librariesio/librariesio-api-provider.js
+++ b/services/librariesio/librariesio-api-provider.js
@@ -48,7 +48,8 @@ export default class LibrariesIoApiProvider {
     // If the header is absent, we just use the current timestamp to
     // advance the value to _something_
     const retryAfter = headers['retry-after']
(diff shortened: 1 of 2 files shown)


chris48s:
Hmm. I'm going to have to come back to this one and look at this `nextReset` code again when I have more time. It might be `getUtcEpochSeconds()` that needs to change..

calebcartwright:
> The other is: The rate limit on one of these tokens is 60/minute, so it feels like 45 is too big a batchSize. Reduce that so we switch every 10 requests,

reasonable, especially considering https://github.com/badges/shields/pull/7080#discussion_r718079795

PyvesB (approved):

PR #10107, opened 2024-04-21
fix [pypi] status badge when package has no 'Development Status' classifier

chris48s:
Currently if we request this badge for a package with no 'Development Status' classifier we render an 'empty' badge

e.g:  - https://img.shields.io/pypi/status/arcgis2geojson/3.0.2

This PR fixes that


PyvesB (approved):
Nice little improvement for this edge case! 👍🏻

PR #10110, opened 2024-04-22
add logoSize to globalQueryParams

chris48s:
After I merged https://github.com/badges/shields/pull/9191 I deployed it to staging, gave it a quick test and realised we missed something.
This param was only taking effect for static badges. It was ignored for dynamic ones.
This change fixes it, but it does also raise follow up questions which I'll note in an issue


LitoMore:
The endpoint badge does not support `logoSize` for now. Maybe you can also support `logoSize` for the endpoint badge in this PR.

PR #10125, opened 2024-04-26
improve performance of getIconSize()

chris48s:
I attempted to deploy https://github.com/badges/shields/pull/9191

This happened

9191-deploy

There's something in that change that has caused a big performance issue.

I haven't really benchmarked this, but my instinct is that loading the whole simple-icons set every time we invoke `getIconSize()` is at least part of the problem.

This PR moves to doing this once at module load time instead of every time we call the function.


chris48s:
that appears to have done the job :relieved:

LitoMore:
I can't believe I made such trouble... 😓

PR #10127, opened 2024-04-29
run tests on node 22

chris48s:
Node 22 is now out :tada: 

We should migrate to node 22 at some point, but for the moment lets just start running the tests against it in preparation. I'll also do some testing with it when I'm developing locally.

### DIFF
--- a/.github/workflows/test-integration-22.yml
+++ b/.github/workflows/test-integration-22.yml
@@ -1,4 +1,4 @@
-name: Integration@node 21
+name: Integration@node 22
 on:
   pull_request:
     types: [opened, reopened, synchronize]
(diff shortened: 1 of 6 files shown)


LitoMore:
https://github.com/badges/shields/blob/1ab4a8dceca88afa3ae5e3ea5251c3f079a0a687/.github/actions/setup/action.yml#L22

This can be upgrade to v4.

PyvesB (approved):

chris48s:
> This can be upgraded to v4

It can, but this has made me realise that dependabot should have bumped it and it didn't. I've raised a separate issue about this - https://github.com/badges/shields/issues/10135

Think I'm going to leave that occurrence of `actions/setup-node` where it is for now as we don't need to upgrade the action for this PR and it will be useful to have dependabot bump it to confirm the new config works when I get a chance to look at that issue.

PR #10129, opened 2024-04-29
ensure color is string before calling toLowerCase()

chris48s:
Fixes https://github.com/badges/shields/issues/10128

I don't have a reliable repro for this bug, but somewhere along the line we are allowing a hex colour which is all numeric to not be a string.

I suggest hitting with a hammer until it plays nicely :hammer:

(I'm aware that somewhere a typescript enthusiast is having an aneurysm)


badge-maker/lib/color.js
```diff
@@ -49,7 +49,7 @@ function normalizeColor(color) {
   } else if (color in aliases) {
     return aliases[color]
   } else if (isHexColor(color)) {
-    return `#${color.toLowerCase()}`
+    return `#${color.toString().toLowerCase()}`
```
LitoMore, line 52: How about moving the `toString` to `isHexColor()`? Like this below:

```
function isHexColor(s = '') {
  return hexColorRegex.test(String(s))
}
```

The `String(s)` is safer here. Even if the value is `null,` it will not throw an error.

PyvesB (approved):
Sounds reasonable to me :+1:

badge-maker/lib/color.js
chris48s: I don't think that would really help tbh.

The issue here is that the current implementation of `isHexColor()` may already return true on a non-string input (I guess `.test()` implicitly casts to string), leading us to try and call `toLowerCase()` on an int. For example:

```
> isHexColor(111111)
true
```

so even if we make the cast explicit, we'd still need to call `toString()` in this block, or do it somewhere before we reach here.

Null or undefined already return false with the current implementation:

```
> isHexColor(null)
false
> isHexColor(undefined)
false
```

PR #10132, opened 2024-05-02
feat(endpoint-badge): add logoSize support; test [endpoint]

LitoMore:
Currently, the endpoint badge response schema does not support `logoSize` parameter.

### DIFF
--- a/services/endpoint/endpoint.service.js
+++ b/services/endpoint/endpoint.service.js
@@ -94,6 +94,14 @@ The endpoint badge takes a single required query param: <code>url</code>, which 
         </td>
       </tr>
       <tr>
+      <td><code>logoSize</code></td>
+      <td>
+        Default: none. Make icons adaptively resize by setting <code>auto</code>.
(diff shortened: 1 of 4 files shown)


chris48s:
Can you have a look at getting your test passing?

This needs to either
- Call `.json?...` and make an assertion about the response object
- Call `.svg?...` and make an assertion about the xml

services/endpoint/endpoint.service.js
```diff
@@ -95,6 +95,14 @@ The endpoint badge takes a single required query param: <code>url</code>, which
           named logo will be used and colored.
         </td>
       </tr>
+      <tr>
+      <td><code>logoSize</code></td>
+      <td>
+        Default: none. Make icons adaptively resize by setting auto.
```
chris48s, line 101:
```suggestion
        Default: none. Make icons adaptively resize by setting <code>auto</code>.
```

chris48s (approved):
I'd completely forgotten about this. Thanks for digging it out from the archives to get it finished.

Issue #10135, opened 2024-05-06
Dependabot does not bump actions referenced from composite actions


chris48s:
:clipboard: **Description**

Raised by @litomore in https://github.com/badges/shields/pull/10127#issuecomment-2093506023

We have dependabot configured to upgrade actions we use:
https://github.com/badges/shields/blob/1c9d3d5cedce022b7dffe4a702ed4da4ff6a80b9/.github/dependabot.yml#L30-L36

While this does work for workflows, it is not bumping actions referenced in composite actions.

So for example, https://github.com/badges/shields/pull/9682 bumped https://github.com/badges/shields/blob/1c9d3d5cedce022b7dffe4a702ed4da4ff6a80b9/.github/workflows/test-package-cli.yml#L27 but not https://github.com/badges/shields/blob/1c9d3d5cedce022b7dffe4a702ed4da4ff6a80b9/.github/actions/setup/action.yml#L22

This is a reported issue upstream: https://github.com/dependabot/dependabot-core/issues/6704

We have quite a lot of custom actions: https://github.com/badges/shields/tree/master/.github/actions

Task:

Read through that issue, amend our dependabot config to include actions referenced from composite actions, then review/merge any follow up PRs from dependabot.

PR #10139, opened 2024-05-07
instruct dependabot to monitor composite actions

chris48s:
Closes https://github.com/badges/shields/issues/10135

Having read over, https://github.com/dependabot/dependabot-core/issues/6704 this should work
but I can't really confirm until I merge it.


chris48s:
boom https://github.com/badges/shields/pull/10140

PR #10195, opened 2024-05-25
Remove namedLogo from defaultBadgeData of non-social badges

PyvesB:
According to our [badge guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines):
> Except for badges using the social style, logos and links should be turned off by default.

The following line of code does remove the logo when the badge has a non-social style, so we're not violating our guidelines per say:
https://github.com/badges/shields/blob/d68217295114c653d0a803463f368df3f146acda/core/base-service/coalesce-badge.js#L153

However, I think it's best to remove the `namedLogo` from the very few non-social badges where it is included, both from a consistency perspective, and to avoid any confusion (for example in #7759). Having something specified that is then overridden deep down somewhere else in the code is not exactly intuitive.

### DIFF
--- a/services/github/github-commits-difference.service.js
+++ b/services/github/github-commits-difference.service.js
@@ -35,7 +35,7 @@ export default class GithubCommitsDifference extends GithubAuthV3Service {
     },
   }
 
-  static defaultBadgeData = { label: 'commits difference', namedLogo: 'github' }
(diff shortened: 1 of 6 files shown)


mbtools (approved):
LGTM 👍

chris48s (approved):

PR #10197, opened 2024-05-25

PyvesB:
It's been a while since we last updated this number (#7723), Cloudfare statistics have since then doubled:
Screenshot 2024-05-25 115431


mbtools (approved):
Impressive! 👍

chris48s (approved):

PR #10200, opened 2024-05-26
Fix [GithubPullRequestCheckState] test

PyvesB:
This is an amusing issue. `abc123` recently became a real ref on the repository (https://github.com/badges/shields/commit/abc123de1d2d3a83b256ccca5e17e6a4b6088097), which started making this test fail. `abcde12356` should be safe for a little while longer. 😄


mbtools:
apparently trying to fake a hex isn't that trivial 😀

mbtools (approved):
👍

PR #10201, opened 2024-05-27
Add [puppetforge] quality score badges

smortex:
The Puppet Forge run static analysis on modules and provide a score
through a (currently) internal API.

Waybe it will be in the public API in the future, but when this happen
adjusting the URL should be easy.  In the meantime, use this internal
API.

### DIFF
--- a/services/puppetforge/puppetforge-base.js
+++ b/services/puppetforge/puppetforge-base.js
@@ -24,6 +24,27 @@ const modulesSchema = Joi.object({
   ),
 }).required()
 
+const modulesValidationSchema = Joi.array()
+  .items(
+    Joi.alternatives().try(
+      Joi.object({
+        name: Joi.string().valid('total').required(),
+        score: nonNegativeInteger,
+      }).required(),
+      Joi.object({}).required(),
+    ),
+  )
+  .custom((value, helpers) => {
+    // Custom validation to check for exactly one type1 object
+    const totalCount = value.filter(item => item.name === 'total').length
+    if (totalCount !== 1) {
+      return helpers.message(
+        'Array must contain exactly one object of type "total"',
+      )
+    }
+    return value
+  })
+
 class BasePuppetForgeUsersService extends BaseJsonService {
   async fetch({ user }) {
     return this._requestJson({
@@ -42,4 +63,17 @@ class BasePuppetForgeModulesService extends BaseJsonService {
   }
 }
 
-export { BasePuppetForgeModulesService, BasePuppetForgeUsersService }
+class BasePuppetForgeModulesValidationService extends BaseJsonService {
+  async fetch({ user, moduleName }) {
+    return this._requestJson({
+      schema: modulesValidationSchema,
+      url: `https://forgeapi.puppetlabs.com/private/validations/${user}-${moduleName}`,
+    })
+  }
+}
+
+export {
+  BasePuppetForgeModulesService,
+  BasePuppetForgeUsersService,
+  BasePuppetForgeModulesValidationService,
+}
--- a/services/puppetforge/puppetforge-module-quality-score.service.js
+++ b/services/puppetforge/puppetforge-module-quality-score.service.js
@@ -1,0 +1,45 @@
+import { coveragePercentage as coveragePercentageColor } from '../color-formatters.js'
+import { pathParams } from '../index.js'
+import { BasePuppetForgeModulesValidationService } from './puppetforge-base.js'
+
+export default class PuppetforgeModuleQualityScoreService extends BasePuppetForgeModulesValidationService {
+  static category = 'rating'
+
+  static route = {
+    base: 'puppetforge/qualityscore',
+    pattern: ':user/:moduleName',
+  }
+
+  static openApi = {
+    '/puppetforge/qualityscore/{user}/{moduleName}': {
+      get: {
+        summary: 'Puppet Forge quality score',
+        parameters: pathParams(
+          {
+            name: 'user',
+            example: 'camptocamp',
+          },
+          {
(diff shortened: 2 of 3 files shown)


services/puppetforge/puppetforge-module-quality-score.service.js
```diff
@@ -0,0 +1,47 @@
...
+          {
+            name: 'moduleName',
+            example: 'openssl',
+          },
+        ),
+      },
+    },
+  }
+
+  static defaultBadgeData = { label: 'score' }
```
smortex, line 31: Maybe should be "quality score"?

chris48s:
Where did you become aware of this API endpoint? Is this like a preview of a feature that will become stable at some point, or is this really something we should not be using?

https://github.com/badges/shields/blob/f57c4edbb9cb1412a59c87ed633518e85cc41d1d/CONTRIBUTING.md?plain=1#L106

Is there anywhere I can read or find out more?

smortex:
> Where did you become aware of this API endpoint? Is this like a preview of a feature that will become stable at some point, or is this really something we should not be using?
> 
> https://github.com/badges/shields/blob/f57c4edbb9cb1412a59c87ed633518e85cc41d1d/CONTRIBUTING.md?plain=1#L106
> 
> Is there anywhere I can read or find out more?

I think I discovered about the API endpoint in the Puppet Slack channel, where someone asked for details about the quality score and the API allows to have all details.

Maybe @binford2k can give more details about this API, from what I experienced, it have been available and stable for some years now, but as someone working at Puppetlabs he might have some insights on what is cooking?

PyvesB:
Thanks for reviving the PR by the way! Let's wait a bit to get clarity on the API. :)

binford2k:
This endpoint is stable and no plans to change in the near future. I don't think it was actually intended to be _secret_ they just wanted to be lazy about committing it to the versioned API. I'll file a ticket to work towards making it public.

There _are_ plans to improve the information it returns in the future, but @smortex is on the committee to be consulted about those changes and can update this to match.

PyvesB:
@binford2k thanks for the extra information!

> I'll file a ticket to work towards making it public

Could you please share the link to the ticket here so that we can reference it in the future if need-be?

@smortex one of the new service tests [is failing](https://github.com/badges/shields/actions/runs/9291031231?pr=10201), could you please take a look?

services/puppetforge/puppetforge-base.js
```diff
@@ -24,6 +24,8 @@ const modulesSchema = Joi.object({
   ),
 }).required()
 
+const modulesValidationSchema = Joi.array().required()
```
chris48s, line 27: This is an interesting API response format we're working with here.
I think the current approach to dealing with it is a bit brittle in a couple of ways:
1. We're not really validating the input properly. I accept this is difficult - Joi doesn't make it super easy to validate an array containing objects which all match different schemas.
2. We're relying on the position of the object in the array.
I think my suggestion would be to define the schema like this:

```js
import { nonNegativeInteger } from '../validators.js'

const schema = Joi.array()
  .items(
    Joi.alternatives().try(
      Joi.object({
        name: Joi.string().valid('total').required(),
        score: nonNegativeInteger,
      }).required(),
      Joi.object({}).required(),
    ),
  )
  .custom((value, helpers) => {
    // Custom validation to check for exactly one type1 object
    const totalCount = value.filter(item => item.name === 'total').length
    if (totalCount !== 1) {
      return helpers.message(
        'Array must contain exactly one object of type "total"',
      )
    }
    return value
  })
```

This requires the array to have exactly one object in it with a `name` field equal to `"total"` and a `score field`.

Then instead of relying on position within the array, we locate the relevant object using:

```js
data.find(el => el.name == 'total')
```

and we've already validated that will exist with the schema.

services/puppetforge/puppetforge-module-quality-score.tester.js
```diff
@@ -0,0 +1,27 @@
...
+  label: 'score',
+  message: isPercentage,
+})
+
+t.create('module quality score (no ratings)')
+  .get('/camptocamp/openssl.json')
+  .intercept(nock =>
+    nock('https://forgeapi.puppetlabs.com/private/validations')
+      .get('/camptocamp-openssl')
+      .reply(200, {}),
```
chris48s, line 15: Surely the invalid response would be {} not []?
Do you have an example of the kind of response we're trying to guard against.

services/puppetforge/puppetforge-module-quality-score.service.js
chris48s: Lets call it "quality score"

chris48s:
OK. I don't want to be too dogmatic about the undocumented endpoints thing. If someone who works for puppetforge is saying we can use it, that's fine.

I've left some comments.

services/puppetforge/puppetforge-base.js
smortex: Nice!

services/puppetforge/puppetforge-module-quality-score.tester.js
smortex: Anything that does not validate the schema is good, so an empty array is right I think.

smortex:
Thanks @chris48s for these awesome snippets.  My attempts to improve validation where way behind the masterpiece you proposed :grin:

I integrated these changes and rebased my branch on top of master.  All tests are now passing!

Thanks!

binford2k (approved):

binford2k:
@PyvesB I've re-filed it as a public ticket so you can track it.

chris48s (approved):
thanks - latest changes look good :+1:

PR #10240, opened 2024-06-08
Update documentation and tests to [docsrs]

### LINKED ISSUE #2423: TUTORIAL.md: Replace Wercker with some public service API
:question: **Question**

Attempt to find out more info about Wercker leads to Oracle page about company acquisition. I think that's not as interesting and inspiring as working with some open API, such as Travis CI, GitLab or Travis. Would be nice to see some open source examples there.

PyvesB:
Updates the main tutorial and the service-tests one to use our Docs.rs service instead of the no longer existing Wercker service. It's a nice an simple service with everything publicly available. Closes #2423.

One of the error behaviours changed on the upstream API, I've also updated the implementation and tests to reflect that.

### DIFF
--- a/doc/service-tests.md
+++ b/doc/service-tests.md
@@ -129,8 +129,8 @@ to run the test with some additional debug output.
 We should write tests cases for valid paths through our code. The Docs.rs badge supports an optional version parameter so we'll add a second test for a branch build. In this case, we know for sure that the documentation for this older version was successfully built, we specify a string literal instead of a Joi schema for `message`. This narrows down the expectation and gives us a more helpful error message if the test fails.
 
 ```js
-t.create('Passing docs for version').get('/tokio/0.3.0.json').expectBadge({
-  label: 'docs@0.3.0',
+t.create('Passing docs for version').get('/tokio/1.37.0.json').expectBadge({
+  label: 'docs@1.37.0',
   message: 'passing',
   color: 'brightgreen',
 })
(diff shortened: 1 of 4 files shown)


doc/service-tests.md
```diff
@@ -115,80 +119,100 @@ That's looking good!
...
 
-Having covered the typical and custom cases, we'll move on to errors. We should include a test for the 'not found' response and also tests for any other custom error handling. The Wercker integration defines a custom error condition for 401 as well as a custom 404 message:
+We can also add a test for a unsuccessful documentation build:
 
-```js
-httpErrors: {
-  401: 'private application not supported',
-  404: 'application not found',
-}
+```j
```
chris48s, line 161:
```suggestion
```js
```

doc/service-tests.md
```diff
@@ -115,80 +119,100 @@ That's looking good!
...
+t.create('Failing docs for version')
+  .get('/tensorflow/0.16.1.json')
```
chris48s, line 163: There's a couple of places here where we are switching between using tensorflow and tokio. Can we either stick to one, or explain more explicitly why we are using different packages in some examples.

chris48s (commented):
Couple of minor points to follow up, but I'm on-board with this. Good to update these docs to use a more relevant example.

Also closes a long-standing issue https://github.com/badges/shields/issues/2423 :D

doc/service-tests.md
PyvesB: There you go, tokio also has a couple of versions with failing builds.

chris48s (approved):

PR #10241, opened 2024-06-08
Add [Docker] support for loong64 arch

### LINKED ISSUE #10224: Docker Image Size arch error

shields.io

### 🐞 Description

Shield.io cannot handle loongarch64.
Example: https://hub.docker.com/layers/qinka/dockerroot-base/unstable/images/sha256-cc1db4e19dcb60613ae7fa3e6e1c8600af609a0d606f48622e751220f609bce9?context=repo

### 🔗 Link to the badge

Docker Image Size (tag)

### 💡 Possible Solution

Add tag for loongarch

PyvesB:
Fixes #10224.

### DIFF
--- a/services/docker/docker-helpers.js
+++ b/services/docker/docker-helpers.js
@@ -17,6 +17,7 @@ const archEnum = [
   'mips64',
   'mips64le',
   'riscv64',
+  'loong64',
 ]
 
 // Valid architecture values: https://golang.org/doc/install/source#environment (GOARCH)
--- a/services/docker/docker-version.tester.js
+++ b/services/docker/docker-version.tester.js
@@ -2,10 +2,12 @@ import { isSemver } from '../test-validators.js'
 import { createServiceTester } from '../tester.js'
 export const t = await createServiceTester()
 
-t.create('docker version (valid, library)').get('/_/alpine.json').expectBadge({
-  label: 'version',
-  message: isSemver,
-})
+t.create('docker version (valid, library)')
+  .get('/_/memcached.json')
+  .expectBadge({
+    label: 'version',
+    message: isSemver,
+  })
 
 t.create('docker version (valid, library with tag)')
   .get('/_/alpine/latest.json')


PyvesB:
Wow, these tests were pretty stable and were working fine not so long ago (#9934), but we've now got a bunch a failures. Maybe best to take a look in a follow up?

calebcartwright (approved):

calebcartwright:
The test failures are unrelated IMO, so I agree and I'd be fine giving this the :+1: to go ahead and fix #10224, and then address the failing tests separately.

Let me know if you'd like me to re-approve or if you change your mind and decide to try to fix the tests here

PyvesB:
I've fixed one of the tests to use a better example.

The other ones seems to be getting 401 responses, they're all Docker Cloud examples. Our Docker badges support optional auth:
https://github.com/badges/shields/blob/0360ba3a1b193b25a03454e36b823aa20155d7fc/services/docker/docker-automated.service.js#L19-L25

As far as I can tell, we only ever configured `dockerhub_username` but not `dockerhub_pat` in our GitHub actions configuration. I'm wondering whether Docker Cloud started requiring auth, or whether the project we're testing against, `pavics/magpie`, moved to being a private one. @chris48s you're much more familiar with the service and reworked auth recently, do you have any insights?

chris48s:
So 2 things:

1. When you run service tests from a fork, we don't pass any secrets so on this PR (submitted from your fork) there are no credentials being used
2. Even when submit a PR from a local branch on this repo, the vars we pass are: https://github.com/badges/shields/blob/0360ba3a1b193b25a03454e36b823aa20155d7fc/.github/workflows/test-services.yml#L26-L36 dockerhub username (or PAT) isn't one of the things we set.

The docker credentials in the repo secrets (not prefixed with `SERVICETESTS_`) are used for pushing the images we build to DockerHub.

Looking at it, I think this is not just the service tests. I think the DockerCloud badges are just completely broken - with or without credentials. I think this is due to an upstream change. I'm also happy to punt this to another issue as it is unconnected to this PR, but does want fixing. Related: https://github.com/badges/shields/issues/9563

btw, I've been waiting for this fix to land for a looooong time :trollface:

chris48s (approved):

Issue #63, opened 2013-09-16
sunset badgr.co


chadwhitacre:
There can't be that many people using Badgr.co yet but I know there's a few. I'm going to start tweeting and linking here to hopefully let people know that Badgr.co is going away. I suppose we could leave some overlap with Shields.io once that's launched. Probably for how much traffic Badgr.co is getting a month of overlap would be fine (it's double the SSL cost $20/mo).

Issue #13, opened 2013-02-10
"Put a Shield on It"™


olivierlacan:
Alright, let's make a list of the services that currently use discrepant badges so we can get in touch with them, involve them in this initiative and get them shielded in due form.
## Services
- [x] [Travis](http://about.travis-ci.org/docs/user/status-images/) is live!
- [x] [Code Climate](https://codeclimate.com/changelog/510d4fde56b102523a0004bf) is live!
- [x] [Gemnasium](http://blog.tech-angels.com/post/43141047457/gemnasium-v3-aka-gemnasium) is live!
- [x] [Gemfury](https://twitter.com/Gemfury/status/300709469137010688) is coming soon!
- [x] [Coveralls](https://coveralls.io/repos/hannesg/uri_template/badge.png?branch=master) is live!

PR #10245, opened 2024-06-11
add [WingetVersion] Badge

### LINKED ISSUE #5548: New service: winget
:clipboard: **Description**

TIL Microsoft are currently previewing winget: a package manager for Windows (a bit like Chocolatey or Scoop, but first-party). https://github.com/microsoft/winget-cli This might gain some traction, or it might be bit of a zune but the fact that it is the MS "blessed" solution to a real problem does make this noteworthy.

Manifest yaml format: https://github.com/microsoft/winget-pkgs#authoring-a-manifest
This would allow us to produce license/version badges.

:link: **Data**

Registry-wise they seem to be going down the cocoapods route where all the package manifests live in a big GH repo: https://github.com/microsoft/winget-pkgs

Somewhat unclear how stable/subject to change this all is.

:microphone: **Motivation**

I don't especially want to use this myself because :penguin: but I figure its useful to have a tracking issue for it.

anatawa12:
Fixes #5548 

~~Writing test in progress (might be tomorrow in JST)~~ I wrote tests

This implementation fetches version information from https://github.com/microsoft/winget-pkgs/ with graphql api.
However, actual vrc-get command line interface retrieves package information from sqlite database file in https://cdn.winget.microsoft.com/cache/source.msix.

I don't know if this difference is acceptable for shelds.

### DIFF
--- a/services/winget/version.js
+++ b/services/winget/version.js
@@ -47,6 +47,9 @@ function compareVersion(v1, v2) {
   const parts1 = v1Trimmed.split('.')
   const parts2 = v2Trimmed.split('.')
 
+  trimLastZeros(parts1)
+  trimLastZeros(parts2)
+
   for (let i = 0; i < Math.min(parts1.length, parts2.length); i++) {
     const part1 = parts1[i]
     const part2 = parts2[i]
@@ -61,19 +64,10 @@ function compareVersion(v1, v2) {
     return 0
   }
 
-  // ignore .0s at the end
   if (parts1.length > parts2.length) {
-    for (let i = parts2.length; i < parts1.length; i++) {
-      if (parts1[i].trim() !== '0') {
-        return 1
-      }
-    }
+    return 1
   } else if (parts1.length < parts2.length) {
-    for (let i = parts1.length; i < parts2.length; i++) {
-      if (parts2[i].trim() !== '0') {
-        return -1
-      }
-    }
+    return -1
   }
 
   return 0
@@ -97,6 +91,17 @@ function trimPrefix(version) {
     return version.slice(digitPos.index)
   }
   return version
+}
+
+/**
+ * Removes all trailing zeros from a version number part array.
+ *
+ * @param {string[]} parts - parts
+ */
+function trimLastZeros(parts) {
+  while (parts.length > 1 && parts[parts.length - 1].trim() === '0') {
+    parts.pop()
+  }
 }
 
 /**
--- a/services/winget/version.js
+++ b/services/winget/version.js
@@ -163,7 +163,7 @@ function latest(versions) {
 
   let version = versions[0]
   for (let i = 1; i < len; i++) {
-    if (compareVersion(version, versions[i]) < 0) {
+    if (compareVersion(version, versions[i]) <= 0) {
       version = versions[i]
     }
   }
--- a/services/winget/version.js
+++ b/services/winget/version.js
@@ -1,0 +1,144 @@
+/**
+ * Comparing versions with winget's version comparator.
+ *
+ * See https://github.com/microsoft/winget-cli/blob/ae566c7bf21cfcc75be7ec30e4036a30eede8396/src/AppInstallerSharedLib/Versions.cpp for original implementation.
+ *
+ * @module
+ */
+
+/**
+ * Compares two strings representing version numbers lexicographically and returns an integer value.
+ *
+ * @param {string} v1 - The first version to compare
+ * @param {string} v2 - The second version to compare
+ * @returns {number} -1 if v1 is smaller than v2, 1 if v1 is larger than v2, 0 if v1 and v2 are equal
+ * @example
+ * compareVersion('1.2.3', '1.2.4') // returns -1 because numeric part of first version is smaller than the numeric part of second version.
+ */
+function compareVersion(v1, v2) {
+  // https://github.com/microsoft/winget-cli/blob/ae566c7bf21cfcc75be7ec30e4036a30eede8396/src/AppInstallerSharedLib/Versions.cpp#L109-L173
+  const v1Trimmed = trimPrefix(v1)
+  const v2Trimmed = trimPrefix(v2)
+
+  const parts1 = v1Trimmed.split('.')
+  const parts2 = v2Trimmed.split('.')
+
+  for (let i = 0; i < Math.min(parts1.length, parts2.length); i++) {
+    const part1 = parts1[i]
+    const part2 = parts2[i]
+
+    const compare = compareVersionPart(part1, part2)
+    if (compare !== 0) {
+      return compare
+    }
+  }
+
+  if (parts1.length === parts2.length) {
+    return 0
+  }
+
+  // ignore .0s at the end
+  if (parts1.length > parts2.length) {
+    for (let i = parts2.length; i < parts1.length; i++) {
+      if (parts1[i].trim() !== '0') {
+        return 1
+      }
+    }
+  } else if (parts1.length < parts2.length) {
+    for (let i = parts1.length; i < parts2.length; i++) {
+      if (parts2[i].trim() !== '0') {
+        return -1
+      }
+    }
+  }
+
+  return 0
+}
+
+/**
+ * Removes all leading non-digit characters from a version number string
+ * if there is a digit before the split character, or no split characters exist.
+ *
+ * @param {string} version The version number string to trim
+ * @returns {string} The version number string with all leading non-digit characters removed
+ */
+function trimPrefix(version) {
+  // https://github.com/microsoft/winget-cli/blob/ae566c7bf21cfcc75be7ec30e4036a30eede8396/src/AppInstallerSharedLib/Versions.cpp#L66
+  // If there is a digit before the split character, or no split characters exist, trim off all leading non-digit characters
+
+  const digitPos = version.match(/(\d.*)/)
+  const splitPos = version.match(/\./)
+  if (digitPos && (splitPos == null || digitPos.index < splitPos.index)) {
+    // there is digit before the split character so strip off all leading non-digit characters
+    return version.slice(digitPos.index)
+  }
+  return version
+}
+
+/**
+ * Compares two strings representing version number parts lexicographically and returns an integer value.
+ *
+ * @param {string} part1 - The first version part to compare
+ * @param {string} part2 - The second version part to compare
+ * @returns {number} -1 if part1 is smaller than part2, 1 if part1 is larger than part2, 0 if part1 and part2 are equal
+ * @example
+ * compareVersionPart('3', '4') // returns -1 because numeric part of first part is smaller than the numeric part of second part.
+ */
+function compareVersionPart(part1, part2) {
+  // https://github.com/microsoft/winget-cli/blob/ae566c7bf21cfcc75be7ec30e4036a30eede8396/src/AppInstallerSharedLib/Versions.cpp#L324-L352
+  const [, numericString1, other1] = part1.trim().match(/^(\d*)(.*)$/)
+  const [, numericString2, other2] = part2.trim().match(/^(\d*)(.*)$/)
+  const numeric1 = parseInt(numericString1 || '0', 10)
+  const numeric2 = parseInt(numericString2 || '0', 10)
+
+  if (numeric1 < numeric2) {
+    return -1
+  } else if (numeric1 > numeric2) {
+    return 1
+  }
+  // numeric1 === numeric2
+
+  const otherFolded1 = (other1 ?? '').toLowerCase()
+  const otherFolded2 = (other2 ?? '').toLowerCase()
+
+  if (otherFolded1.length !== 0 && otherFolded2.length === 0) {
+    return -1
+  } else if (otherFolded1.length === 0 && otherFolded2.length !== 0) {
+    return 1
+  }
+
+  if (otherFolded1 < otherFolded2) {
+    return -1
+  } else if (otherFolded1 > otherFolded2) {
+    return 1
+  }
+
+  return 0
+}
+
+/**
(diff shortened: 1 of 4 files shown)


services/winget/winget-version.service.js
```diff
@@ -0,0 +1,97 @@
...
+          prettyMessage: 'package not found',
+        })
+      }
+      const entries = json.data.repository.object.entries
+      const directories = entries.filter(file => file.type === 'tree')
+      const versions = directories.map(file => file.name)
+      const version = latest(versions)
+
+      return renderVersionBadge({ version })
+    } catch (e) {
```
chris48s, line 92: Are we expecting to catch a particular exception here?
If not, can we ditch the try/catch. Sentry will handle capturing the details of any unexpected runtime error for us in production.

services/winget/winget-version.service.js
```diff
@@ -0,0 +1,97 @@
...
+      const json = await this.fetch({ name })
+      if (json.data.repository.object === null) {
+        throw new InvalidParameter({
+          prettyMessage: 'package not found',
+        })
+      }
+      const entries = json.data.repository.object.entries
+      const directories = entries.filter(file => file.type === 'tree')
+      const versions = directories.map(file => file.name)
+      const version = latest(versions)
```
chris48s, line 89: Is there any way we could find out what the registry considers the "latest" version of a package? Do we have any info on how WinGet sorts versions?

I'm a bit worried that there doesn't really seem to be any constraint on what a version "number" can be. Here's some examples that have one or more odd ones:
- https://github.com/microsoft/winget-pkgs/tree/c939ca136ff33f217db1de99f1a1839790287a16/manifests/j/JetBrains/MPS/EAP
- https://github.com/microsoft/winget-pkgs/tree/c939ca136ff33f217db1de99f1a1839790287a16/manifests/n/niconico/nair/live
- https://github.com/microsoft/winget-pkgs/tree/c939ca136ff33f217db1de99f1a1839790287a16/manifests/j/Jigsaw/OutlineManager
- https://github.com/microsoft/winget-pkgs/tree/c939ca136ff33f217db1de99f1a1839790287a16/manifests/w/Wildy/bobonslaught
- https://github.com/microsoft/winget-pkgs/tree/c939ca136ff33f217db1de99f1a1839790287a16/manifests/t/taers232c/GAMADV-XTD3

Given it seems a version number can be basically any string, it seems like this we are going to get this wrong in some cases.

services/winget/winget-version.service.js
```diff
@@ -0,0 +1,97 @@
...
+            type: Joi.string().required(),
+            name: Joi.string().required(),
+          }),
+        ),
+      })
+        .allow(null)
+        .required(),
+    }).required(),
+  }).required(),
+}).required()
```
chris48s, line 23: :+1:

chris48s (commented):
thanks for having a look at this one

services/winget/winget-version.service.js
anatawa12: sorry I forgot to remove debug code.

services/winget/winget-version.service.js
anatawa12: Sorry for late response. I will investigate the update process in winget database

anatawa12: I could noy find meaningful source code of winget-pkgs pipeline but I found version comparesion in winget-cli, that are used when one package is provided by multiple multiple sources.

I don't know if this compression is used by winget-pkgs pipeline and may not work well with some if your example packages but I think it's reasonable to use this comparesion.

I will implement this compression in services/winget folder with unit tests.

Version Implementation: https://github.com/microsoft/winget-cli/blob/ae566c7bf21cfcc75be7ec30e4036a30eede8396/src/AppInstallerSharedLib/Versions.cpp

Use case in winget-cli: https://github.com/microsoft/winget-cli/blob/ae566c7bf21cfcc75be7ec30e4036a30eede8396/src/AppInstallerRepositoryCore/PackageVersionSelection.cpp#L82-L100

services/winget/winget-version.service.js
anatawa12: removed in d9963b5f5af7c66f11bbdfff1d44a2ae444fce22

services/winget/winget-version.service.js
anatawa12: I have implemented in 6b707327b7bb6d0582d2c44ac32f1a6616ad5323

anatawa12: Sorry  forget to impement latest case so nair case is my fault. 

however, I feel that JetBrains.MPS.EAP case is bug in winget registorysince actual latest version is 241.18034.990 and I think return it as latest with wingwt-cli version comparator.(I'll test later)

chris48s: I'd be down with that as a one-off check to verify this.
I'm not sure I have the appetite to write it.

arturcic: @anatawa12 , @PyvesB , @chris48s.

I made a script that dumps all the manifests to 

- `catalog.csv` - this contains packages and the versions
- `packages.csv` - contains the list of unique packages
- `packages-map.csv` - includes the version of the package returned by `winget show {packageId}` and the version returned by this PR. Last column shows if they match. Winget returns the version without the `v` prefix, in shields with the prefix. So I consider them a match if they have or not the `v` prefix.
- `packages-diff.csv` - contains those packages that need some your attention. 

N/A means it's not returned by `winget show` command (missing in the winget index but present in the manifests repo)

[catalog.csv](https://github.com/user-attachments/files/17532000/catalog.csv)
[packages.csv](https://github.com/user-attachments/files/17532002/packages.csv)
[packages-map.csv](https://github.com/user-attachments/files/17532015/packages-map.csv)
[packages-diff.csv](https://github.com/user-attachments/files/17532026/packages-diff.csv)

Hope this helps move this PR forward and eventually merge.

chris48s: Nice! 

So it looks like we've got 16 examples where our result disagrees with the upstream to look into there and it should be possible to use the lists of versions for those packages to write some test cases.

I'd hope that list of 16 examples reduces to fewer than 16 bugs.

@anatawa12 - are you happy to revisit this? I'm aware the PR has been dormant for some time at this point.

anatawa12: Yes I want but It may took few days or weeks before working on this because I'm working on other projects now.

anatawa12: I want to describe the cause of the problem for each package.

I fixed 12 of 16 but 4 are hard to fix I think

### Confusing part of package name and version name

Affected packages:
- `Canonical.Ubuntu`
- `Codemao.TurtleEditor`
- `ImageMagick.ImageMagick`
- `PLACE.IA`
- `PLACE.ONE`
- `qBittorrent.qBittorrent`
- `RoyalApps.RoyalTS`
- `Shure.WirelessWorkbench`
- `SoftMaker.FreePDF`

This is totally my fault.
I assumed such cases don't exist but it exists.

My code confuses `Canonical.Ubuntu.2404` package as `Canonical.Ubuntu` version 2024.
We can avoid this by checking if `<id>.yaml` exists in the version folder.

I fixed in aabbc388f773d677e1769df5497e1fdc762172e8.

### .0 version is incorrect

Affected packages:
- `CiderCollective.Cider.Nightly`

This is my fault.

In winget, version ends with `.0` will ignore `.0` but my code doesn't ignore `.0` correctly.

I fixed this in a58ab02f6c95ac1e36697dbb2e805afce9ca91e5.

### Same version in the registry

- `CodeF0x.violin` - `3.1.1` and `3.1.1.0`
- `UweSieber.UsbTreeView` - `4.4.2` and `4.4.2.0`

Those packages have same version according to the version comparator in winget.
Manifest content of those package versions are same.

I don't know how exactly winget handles this case, but it looks last entry in git repository wins, so I changed to use last entry in the repository in 439a7e0629db0da0e5c8cb7ea88d6de0f4b4d24e.

### Different package version comparator between 1.8 series and 1.9 series

Affected packages:
- `JetBrains.MPS.EAP`
- `maforget.ComicRackCE.Nightly`

I think this is not required to be fixed, and those packages are updated so It's impossible to test those cases.

It looks the version comparator in winget has been changed between 1.8 series and 1.9 series.
My implementation is based on 1.9 series and winget official repository is based on 1.8 series.

I have created https://github.com/anatawa12/winget-cli/tree/version-comparator-test (based on master, which is 1.9 series) and
cheery-picked onto v1.8.1791 (1.8 series) to compare the version comparator.
I confirmed our version matches with 1.9 series of winget.

For `JetBrains.MPS.EAP`, Preambles support added in [this commit](https://github.com/microsoft/winget-cli/commit/38aac878ab34d0176b34faf6946d7a0b55d4ce97) changed the version comparator.

For `maforget.ComicRackCE.Nightly`, also affected by Preambles support. `nightly-` part is treated as preamble and `7635` and `9` are compared in 1.9, but compared dictionary in 1.8.
Left is 1.9 series and right is 1.8 series. greater means newer, so greatest is the latest version. my implementation picks `nightly-7635d6e` as latest version, which is same as 1.9 series.

image

### Other failures

#### `Coninfer.ECLiPSeCLP.7.0` and `Coninfer.ECLiPSeCLP.7.1`

Showing those packages failed with the following error:
```
$ winget show --id='Coninfer.ECLiPSeCLP.7.1'
An unexpected error occurred while executing the command:
Download request status is not success.
0x80190194 : 見つかりません (404)。
```

Those packages have `#` in the version name.
as a part of `winget show` command, winget will request `https://cdn.winget.microsoft.com/cache/manifests/c/<package path>/<version>/<rowid in sqlite>` but `#` is not allowed in the URL.
I feel this is a bug in winget.

`winget show --id='<id>' --versions` shows the version my implementation picked, so I think it's OK to show the version `v7.0 #63` and `v7.1 #13` from shields.io. 

#### `1357310795.JboxTransfer`

I cannot `winget show --id=1357310795.JboxTransfer`. I don't know why it fails on winget side.

```
> winget show --id=1357310795.JboxTransfer
No package found matching input criteria.
```

services/winget/version.js
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Finds the largest version number lexicographically from an array of strings representing version numbers and returns it as a string.
+ *
+ * @param {string[]} versions - The array of version numbers to compare
+ * @returns {string|undefined} The largest version number as a string, or undefined if the array is empty
+ * @example
+ * latest(['1.2.3', '1.2.4', '1.3', '2.0']) // returns '2.0' because it is the largest version number and pre-release versions are excluded.
+ * latest(['1.2.3', '1.2.4', '1.3', '2.0']) // returns '2.0' because pre-release versions are included but none of them are present in the array.
+ * latest(['1.2.3', '1.2.4', '1.3-alpha', '2.0-beta']) // returns '1.2.4' because pre-release versions are excluded and it is the largest version number among the remaining ones.
```
chris48s, line 156: for me, `latest(['1.2.3', '1.2.4', '1.3-alpha', '2.0-beta'])` returns `'2.0-beta'`

services/winget/winget-version.service.js
chris48s: Thanks for laying all this out really clearly.

I'm inclined to agree about the handful of cases where we don't need to be "bug for bug" compatible

arturcic: @chris48s  any chance we can merge this?

services/winget/version.js
anatawa12: Yes. you're correct. comments are copied from common version utility and I forgot to fix the contents.

anatawa12: fixed in 2cbd02ef4b358dba613beb40268f4aa7b08bece6

services/winget/version.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    // unknown
+    // https://github.com/microsoft/winget-cli/blob/43425fe97d237e03026fca4530dbc422ab445595/src/AppInstallerCLITests/Versions.cpp#L231
+    given('unknown', '1.0').expect(-1)
+    given('unknown', '1.fork').expect(-1)
+    given('unknown', 'UNKNOWN').expect(0)
+
+    // porting failure tests
+    // https://github.com/badges/shields/pull/10245#discussion_r1817931237
+    // trailing .0 and .0-beta
+    given('1.6.0', '1.6.0-beta.98').expect(-1)
```
chris48s, line 46: I'll be honest: I don't look forward to the first time someone raises an issue saying "winget version badge shows me version X but `winget show` says latest is version Y" but I think the key is having really solid tests in place to make sure we're not introducing regressions if we need to change this. Thanks for adding all these and annotating/explaining where these all comes from :+1:

chris48s (approved):
I've pushed a commit adding a couple more test cases for the `latest()` function based on the documentation examples. Aside from that: LGTM, merging. Thanks.

slonopotamus:
Could you please provide an example how this can be used to display a winget badge?

chris48s:
Lets say your package is `Microsoft.WSL`
https://img.shields.io/winget/v/Microsoft.WSL
shows the latest version

slonopotamus:
Thanks!

slonopotamus:
Well, unfortunately it [does not work](https://img.shields.io/winget/v/Octobuild.Octobuild) for [my package](https://winstall.app/apps/octobuild.octobuild) for some reason =/

What's suspicious is that shields.io capitalizes package name in URL, but in fact package has lowercase name.

chris48s:
https://img.shields.io/winget/v/octobuild.octobuild shows v1.4.0



We don't capitalize anything for you, but if you've tried requesting the badge with capitals in it first and you're using a chromium-based browser you might be hitting https://github.com/badges/shields/issues/10395

slonopotamus:
> if you've tried requesting the badge with capitals in it first and you're using a chromium-based browser you might be hitting https://github.com/badges/shields/issues/10395

Ouch! Yes, it was that.

PR #10262, opened 2024-06-16
Improve Docusaurus Dependabot setup

### LINKED ISSUE #10242: Improve Docusaurus Dependabot setup
From https://github.com/badges/shields/pull/10239#issuecomment-2156748004:

> I wonder if we could "teach" dependabot to bump these together if we convert @docusaurus/preset-classic to a top-level dependency and then set up a dependabot group for the pattern @docusaurus/* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

PyvesB:
Fixes #10242.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,7 +16,20 @@ updates:
       # https://caniuse.com/js-regexp-lookbehind
       - dependency-name: 'decamelize'
       - dependency-name: 'humanize-string'
-
+    groups:
+      # All official @docusaurus/* packages should have the exact same version as @docusaurus/core.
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -16,7 +16,20 @@ updates:
...
+      # updates [...] you must define two, separately named, grouping sets of rules"
+      # See https://github.com/badges/shields/issues/10242 for more information.
+      docusaurus-version-updates:
+        applies-to: version-updates
+        patterns:
+          - '@docusaurus*'
+      docusaurus-security-updates:
+        applies-to: security-updates
+        patterns:
+          - '@docusaurus*'
```
chris48s, line 32: I'd be inclined to make the pattern `@docusaurus/*`

chris48s (approved):

PR #7879, opened 2022-04-22
Reclassify test for versionColorForWordpressVersion as an integration test

paulmelnikow:
This test requires hitting a web URL, and doesn't work offline, so it seems better classified as an integration test.

PR #10265, opened 2024-06-17
classify wordpress version color tests correctly

chris48s:
Working on something else, I realised that there are some a few tests that are part of the "integration" test suite (which we often skip). Those tests should just cover GitHub token pooling tests. These should be in a .spec.js so they are part of the core suite.


PyvesB:
Paul had explicitly reclassified these tests in #7879 as they perform network calls. No super strong opinions on my side, but wanted to bring this to your attention in cas you missed it. :)

chris48s:
I was aware they make network requests. I hadn't looked back and found https://github.com/badges/shields/pull/7879

The reason why I was thinking we move these is because the other things that are in the `integration` suite are:

1. Tests that require a GitHub token to run (you don't have to set a GitHub token to run the application locally)
2. Tests that require a Postgres DB (you don't need a Postgres DB to run the application locally)

and so we skip those tests when we run `npm test`.

The tests I was moving this this PR don't need any optional extras to run (although I accept that they could be flaky, like the service tests).

I was going to say that there are other core tests which make network requests, but then I realised that I fixed them all in https://github.com/badges/shields/pull/8643 and forgot :smile: 

so maybe keeping it so that every test in the core suite can run with no network access is a good thing.

Either way, I don't think this is a huge deal. Maybe lets just leave them where they are..

PR #10283, opened 2024-06-21
Expose `logoBase64` and `links` in badge-maker NPM package

chris48s:
In this PR, I have picked out the bits of #9645 that were heading in the right direction and fixed them up. I am probably going to just merge this as most of the review has already happened under #9645 but I'll leave it open a few days in case anyone wants to jump in on it.


PyvesB (approved):
Looks good to me :+1:

PR #10284, opened 2024-06-21
Remove `logoPosition`

chris48s:
In https://github.com/badges/shields/pull/3747 we removed the popout style. Since then, the remnants of the `logoPosition` param have been knocking about. We really should have removed this at the time.

This came up in the review of #9645 where @zavoloklom suggested removing it but provided an incomplete implementation. In this PR, I have completed the job.

Removing this seems like a no-brainer as it is just cruft. Passing `logoPosition` as a URL param or in the endpoint badge already does nothing.


services/endpoint-common.js
```diff
@@ -36,9 +36,14 @@ const endpointSchema = Joi.object({
...
   logoWidth: optionalNumberWhenAnyLogoPresent,
-  logoPosition: optionalNumberWhenAnyLogoPresent,
   style: Joi.string(),
   cacheSeconds: Joi.number().integer().min(0),
+  /*
+  Retained for legacy compatibility
+  Although this does nothing,
+  passing it should not throw an error
+  */
+  logoPosition: optionalNumberWhenAnyLogoPresent,
```
chris48s, line 46: I've left this in the schema so it is not a breaking change

calebcartwright (approved):
yup, seems like the right move to me

PyvesB (approved):
On board with this 👍🏻

PR #10286, opened 2024-06-24
Fix npm badges when `maintainers` not in response; run [NpmCollaborators NpmDependencyVersion NpmDownloads NpmLicense NpmTypeDefinitions NpmUnpackedSize NpmVersion]

chris48s:
I published a new version of badge-maker today and.. our NPM badges broke:

- npm license - https://img.shields.io/npm/l/badge-maker
- npm type definitions - https://img.shields.io/npm/types/badge-maker

Looks like NPM API responses no longer necessarily include a `maintainers` property.

This PR makes this property optional and assumes `[]` if it is missing. This is the same approach we take with the `files` property.

### DIFF
--- a/services/npm/npm-base.js
+++ b/services/npm/npm-base.js
@@ -21,7 +21,7 @@ const packageDataSchema = Joi.object({
   maintainers: Joi.array()
     // We don't need the keys here, just the length.
     .items(Joi.object({}))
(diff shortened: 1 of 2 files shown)


PyvesB (approved):
Thanks for fixing!

PR #10287, opened 2024-06-24
More GitHub icons

MichaelHinrichs:
Add more icons for GitHub badges. Set them as defaults.


PyvesB:
Hello @MichaelHinrichs 👋🏻 

As per our [contributing guidelines](https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines):

> Except for badges using the social style, logos and links should be turned off by default.

A subset of the badges you modified do fit within the social category, but I don't think there's enough of a case to make an exemption to SimpleIcons (see [guidance](https://github.com/badges/shields/blob/master/doc/logos.md)). Besides, I'd be reluctant to wholesale changing a logo that is happily used in millions of badges without a very strong ask from the community.

Thanks for your interest in the project, but unfortunately I don't think this PR is mergeable.

Issue #8821, opened 2023-01-17
Migrate daily service tests to GitHub actions


chris48s:
:clipboard: **Description**

Having migrated all the CI on this repo from Circle CI to GHA, now its time to do the same on https://github.com/badges/daily-tests

I think if I was implementing this from scratch today, I would just set up a scheduled job to run on this repo, like this one:

https://github.com/badges/shields/blob/e8c42e9ec1b2d93790133a7843b04f4a9869cec7/.github/workflows/draft-release.yml#L2-L5

@calebcartwright - Can you remember anything about the rationale for running the daily tests from another repo?

I wonder if we should bring them back into the fold? 🤔

PyvesB:
Is it me, or the daily tests are no longer working at all at this point?

chris48s:
Yeah the daily runs on https://github.com/badges/daily-tests no longer work.

I can't remember exactly when they broke, but it was quite a while back. I decided that instead of putting the time into fixing up that old Circle CI config, I'd rather put the time into moving the daily runs onto scheduled GH actions in this repo. They'll be way easier to keep in sync that way, rather than having to coordinate changes across 2 repos.

But then I ended up doing neither of those things.

chris48s:
https://github.com/badges/shields/pull/10320 adds a daily tests repo back to the repo.
I would still like to add coverage reports to that though.

PyvesB:
#10340 took care of code coverage, closing this issue! 🚀

PR #10320, opened 2024-07-02
The return of daily tests

chris48s:
Refs https://github.com/badges/shields/issues/8821

This PR adds a GitHub Workflow that runs our tests every day and spits out a markdown summary.

Here's an example run: https://github.com/badges/shields/actions/runs/9749868442

This doesn't completely replicate all of the functionality of the old daily tests repo. That repo also measured and reported code coverage. However, I encountered two problems implementing this:

1. Both the markdown summary and coverage reports are implemented as mocha reporters. Out of the box, it is not possible to tell mocha to use 2 reporters at once, but there may be a way to work round this. I do want to produce human readable output as well as coverage XML though.
2. There's an issue with c8 that causes builds to fail when measuring coverage on Node 20. There's an unmerged upstream PR at https://github.com/bcoe/v8-coverage/pull/2

I think both of those problems can be solved in time:
- I can work on finding a way round the reporters issue.
- The c8 bugfix may be merged upstream. If not, eventually node 22 (which doesn't have this issue) will be our target platform.

That said, right now I have a working PR for an action which runs all the tests every day and outputs a summary we can read. That's more than we have now, so I suggest lets do this as step one rather than blocking on those things. Lets not let perfect be the enemy of good. Regaining visibility on what is failing will help us start fixing stuff.

### DIFF
--- a/.github/workflows/daily-tests.yml
+++ b/.github/workflows/daily-tests.yml
@@ -1,0 +1,71 @@
+name: Run Daily Tests
+on:
+  schedule:
+    - cron: '45 3 * * *'
+    # At 03:45, daily
+  workflow_dispatch:
+
+jobs:
+  daily-tests:
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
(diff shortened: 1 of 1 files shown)


PyvesB (approved):
Thanks for putting this together, looks good to me! 👍🏻

Issue #10323, opened 2024-07-04
GitHub Contributors badges are not documented on website


chris48s:
:clipboard: **Description**

For historic reasons, we have a handful of badges that use the same 'path' in the Open API specs:

https://github.com/search?q=repo%3Abadges%2Fshields+path%3A%2F%5Eservices%5C%2Fgithub%5C%2F%2F+github%2F%7Bvariant%7D%2F%7Buser%7D%2F%7Brepo%7D&type=code

This is mostly fine, except we use path as the key in an object here:

https://github.com/badges/shields/blob/e4a9c73759f2edf909c21828b0f6fb6fa5c07520/core/base-service/openapi.js#L81-L94

This means if we have 2 services in the same category where the path is `/github/{variant}/{user}/{repo}` one overwrites the other and only one shows up in the docs.

Because both
https://github.com/badges/shields/blob/master/services/github/github-contributors.service.js
and
https://github.com/badges/shields/blob/master/services/github/github-release-date.service.js
are in the activity category, only one of them shows up in the docs.

I need to find a way round this. Probably make it an array instead of object.

While we are sorting this out, we should also document the difference between the contributors and contributors-anon variants.

PR #10337, opened 2024-07-05
include gituhb contributors badge in docs site

chris48s:
Closes https://github.com/badges/shields/issues/10323
Refs https://github.com/badges/shields/discussions/10322

This PR does two things:

1. Include GitHub Contributors badge in docs site.
   Fundamentally, the Open API spec dictates that this must be laid out as an object keyed on the route so each route must be unique within each category. The way I've chosen to solve this is by renaming the param from `variant` to `metric` in this one case. I also considered moving one of the badges to a different category, but I think they do really both belong in Activity.
2. Explain the difference between the two versions of this badge.

### DIFF
--- a/core/base-service/openapi.js
+++ b/core/base-service/openapi.js
@@ -84,7 +84,7 @@ function services2openapi(services, sort) {
     for (const [key, value] of Object.entries(
       addGlobalProperties(service.openApi),
     )) {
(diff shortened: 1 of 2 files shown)


PyvesB (approved):
Looks good to me! :+1:

PR #10341, opened 2024-07-08
Update badges in README

PyvesB:
Small opinionated pass on the badges at the top of our README:
* removed the build and contributors badges, as the exact status and counts are already displayed on the GItHub homepage.
* updated the links of the backers and sponsors badges.
* added a discussions badge.
* switched the Discord logo to white for better contrast.
* updated the daily tests badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,25 +3,23 @@
         height="130">
 </p>
 <p align="center">
-    <a href="https://github.com/badges/shields/graphs/contributors" alt="Contributors">
-        <img src="https://img.shields.io/github/contributors/badges/shields" /></a>
-    <a href="#backers" alt="Backers on Open Collective">
+    <a href="https://shields.io/community#backers" alt="Backers on Open Collective">
         <img src="https://img.shields.io/opencollective/backers/shields" /></a>
-    <a href="#sponsors" alt="Sponsors on Open Collective">
+    <a href="https://shields.io/community#sponsors" alt="Sponsors on Open Collective">
         <img src="https://img.shields.io/opencollective/sponsors/shields" /></a>
     <a href="https://github.com/badges/shields/pulse" alt="Activity">
         <img src="https://img.shields.io/github/commit-activity/m/badges/shields" /></a>
-    <a href="https://circleci.com/gh/badges/shields/tree/master">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,25 +3,23 @@
...
-    <a href="https://circleci.com/gh/badges/shields/tree/master">
-        <img src="https://img.shields.io/circleci/project/github/badges/shields/master" alt="build status"></a>
-    <a href="https://circleci.com/gh/badges/daily-tests">
-        <img src="https://img.shields.io/circleci/project/github/badges/daily-tests?label=service%20tests"
-            alt="service-test status"></a>
+    <a href="https://github.com/badges/shields/discussions" alt="Discussions">
+        <img src="https://img.shields.io/github/discussions/badges/shields" /></a>
+    <a href="https://github.com/badges/shields/actions/workflows/daily-tests.yml">
+        <img src="https://img.shields.io/github/actions/workflow/status/badges/shields/daily-tests.yml?label=daily%20tests"
+            alt="Daily Tests Status"></a>
```
calebcartwright, line 16: i think this is the only one i'd push back on, as i'm still not a big fan of framing these as daily tests (and yes I know that nomenclature is already part of our vocabulary elsewhere)

setting aside that these have been executed on a different, greater than daily frequency, i think there's some value in reusing the "service test" verbiage that's utilized elsewhere so that contributors may perhaps more easily realize there's a connection when they see tests failing locally or in their PR

I don't feel terribly strongly about this though, especially if you and Chris prefer the daily test framing. just my .02 that we should be moving away from that framing instead of emphasizing it more :shrug:

PyvesB: I don't remmeber whether this was the case with the old "daily tests", but the new workflow in `daily-tests.yml` runs more than just the service tests, it runs the core, entrypoint, package, and integration tests as well. If feels odd to me to narrow down the badge's label to a subset of what it's actually keeping track of. Maybe something like `full test suite` would better represent what's being built underneath the hood?

calebcartwright: yeah i think `full test suite` would be a more accurate label

The more I thought about this recurring execution of tests today the more I realized I was stumbling into a can of worms that I don't think should be addressed here so I'll go ahead and :+1: and open a new issue to pose some thoughts/questions for discussion

calebcartwright (approved):

PR #10344, opened 2024-07-09
Use XML for Chocolatey, affects [Chocolatey Resharper PowershellGallery]

### LINKED ISSUE #10343: chocolatey not found

shields.io

### 🐞 Description

chocolatey not found.

https://community.chocolatey.org/packages/picguard

[Chocolatey Version](https://img.shields.io/chocolatey/v/picguard)

[Chocolatey Version](https://img.shields.io/chocolatey/v/picguard?include_prereleases)

### 🔗 Link to the badge

_No response_

### 💡 Possible Solution

_No response_

PyvesB:
Proposal to fix #10343.

The Chocolatey API seems to have started ignoring our `"Accept": "application/atom+json,application/json"` header, and sometimes now returns XML even when we're requesting JSON. The behaviour seems pretty inconsistent, changing slightly the filter parameter in the query can lead to a different format being returned. Switching to XML seems to produce stable results with all the examples I've tried.

### DIFF
--- a/services/nuget/nuget-v2-service-family.js
+++ b/services/nuget/nuget-v2-service-family.js
@@ -2,7 +2,6 @@ import Joi from 'joi'
 import queryString from 'query-string'
 import { nonNegativeInteger } from '../validators.js'
 import {
-  BaseJsonService,
   BaseXmlService,
   NotFound,
   redirector,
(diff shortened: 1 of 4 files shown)


services/nuget/nuget-v2-service-family.js
```diff
@@ -20,26 +19,11 @@ function createFilter({ packageName, includePrereleases }) {
   const releaseTypeFilter = includePrereleases
     ? 'IsAbsoluteLatestVersion eq true'
     : 'IsLatestVersion eq true'
-  return `Id eq '${packageName}' and ${releaseTypeFilter}`
+  return `tolower(Id) eq '${packageName.toLowerCase()}' and ${releaseTypeFilter}`
```
PyvesB, line 22: Case now seems to matter on the Chocolatey API. Without this, our tests based on `scriptcs` fail, we'd need to use `ScriptCs`. I'm guessing other users may have relied on case-insensitivity in the past, so this keeps backwards compatibility.

chris48s (approved):

PR #10350, opened 2024-07-11
Use ellipsis when many versions returned for [ModrinthGameVersions]

### LINKED ISSUE #10346: Modrinth Game Versions don't show ranges

shields.io

### 🐞 Description

When the latest release of a Modrinth project has a range of supported version, each version is listed individually instead of as a range as it is on Modrinth. This causes *very long* badges in some cases. 

### 🔗 Link to the badge

image
```https://img.shields.io/modrinth/game-versions/P1OZGk5p```
https://modrinth.com/plugin/viaversion

### 💡 Possible Solution

Using the same range behavior when multiple versions are supported as Modrinth would be ideal and is what I was expecting.

PyvesB:
Fixes #10346.

To make sure the ellipsis falls in the "middle", the list of game versions must be sorted. All the examples I've tried seem to return sorted results from the API, so I'm running under that assumption and have kept the implementation simple. We can always revisit later if need-be.

### DIFF
--- a/services/modrinth/modrinth-game-versions.service.js
+++ b/services/modrinth/modrinth-game-versions.service.js
@@ -25,6 +25,12 @@ export default class ModrinthGameVersions extends BaseModrinthService {
   static defaultBadgeData = { label: 'game versions' }
 
   static render({ versions }) {
+    if (versions.length > 5) {
+      return {
+        message: `${versions[0]} | ${versions[1]} | ... | ${versions[versions.length - 2]} | ${versions[versions.length - 1]}`,
+        color: 'blue',
+      }
+    }
     return {
       message: versions.join(' | '),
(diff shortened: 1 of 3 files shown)


chris48s:
I think I'm fine with this. Do you have any thoughts on the slightly more involved solution I mentioned in https://github.com/badges/shields/issues/10346#issuecomment-2223139997 ?

chris48s (approved):

PyvesB:
> I think I'm fine with this. Do you have any thoughts on the slightly more involved solution I mentioned in https://github.com/badges/shields/issues/10346#issuecomment-2223139997 ?

I figured I'd go for the simple solution first. If there's appetite from the community for a more fancy solution, I'm happy to refine things. :)

PR #6778, opened 2021-07-18
switch to c8 for code coverage

calebcartwright:
Resolves #6753

PR #10361, opened 2024-07-12
Update c8 configuration

PyvesB:
* adds a few extra ignored folders and deletes no-longer relevant ones.
* renames `.nycrc.json` to `.c8rc.json`, we hadn't done so in #6778 as we were still evaluating c8 at the time.
* stops attempting to clean up `.nyc_output`, we hadn't done so in #6778 as we were still evaluating c8 at the time.

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -9,10 +9,8 @@
     "**/test-helpers.js",
     "**/*-test-helpers.js",
     "**/*-fixtures.js",
-    "**/mocha-*.js",
     "**/*.test-d.ts",
     "dangerfile.js",
-    "gatsby-*.js",
     "core/service-test-runner",
     "core/got-test-client.js",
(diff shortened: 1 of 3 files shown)


chris48s:
We could also remove this from gitignore:

https://github.com/badges/shields/blob/848e409e559bfcf11c55329e9c8872bc26585e5b/.gitignore#L53-L54

chris48s (dismissed):
I've pointed out one more thing we can remove, but it doesn't need another round of review after that. Feel free to just bypass branch protections to merge

PyvesB:
Yeah, I wasn't sure whether to keep it or not as it was added in bulk in https://github.com/badges/shields/commit/7c4b2b67e606b4b25886e36539502e24b3522ce9, but happy to remove it!

PR #10377, opened 2024-07-18
Fix [FreeCodeCampPoints] not found handling

PyvesB:
The API apparently now returns a nice 404 response code we can leverage.

### DIFF
--- a/services/freecodecamp/freecodecamp-points.service.js
+++ b/services/freecodecamp/freecodecamp-points.service.js
@@ -1,11 +1,6 @@
 import Joi from 'joi'
 import { metric } from '../text-formatters.js'
-import {
-  BaseJsonService,
-  InvalidResponse,
-  NotFound,
-  pathParams,
-} from '../index.js'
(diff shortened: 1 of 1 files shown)


services/freecodecamp/freecodecamp-points.service.js
```diff
@@ -62,18 +57,18 @@ export default class FreeCodeCampPoints extends BaseJsonService {
...
         },
       },
+      httpErrors: { 404: 'profile not found' },
     })
   }
 
   static transform(response, username) {
     const { entities } = response
 
-    if (entities === undefined)
```
chris48s, line 71: If we're removing this handling, we tighten up the schema and make sure the values we need are now required in the schema. At the moment the schema is written to accept bad values so we could handle the not found case like this.

chris48s (changes requested):

services/freecodecamp/freecodecamp-points.service.js
PyvesB: Good catch, thanks!

chris48s (approved):

PR #10403, opened 2024-07-24
Cache text only static badges for longer

chris48s:
We serve a lot of static badges. `static_badge` is the single most popular badge on shields. Depending on the time of day it accounts for 10-15% of the traffic we serve from the origin (i.e: uncached). They are the least computationally expensive badges to serve, but there are a lot of them. Each one uses a connection, costs us some egress bandwidth, etc

In general, these badges don't really change.  today is  tomorrow.

We already serve these with quite a long cache on them (1 day), but I think we could go longer in some cases. There is a bit of churn of icon updates in simple-icons, so I think badges which use a logo should stay cached for a day, but for static badges that are just text, lets cache them for longer. I don't think we should say "cache forever". At some point, we will fix a font rendering issue or [change the default palette](https://github.com/badges/shields/pull/9916) or whatever, so I do want these things to expire after some interval. However, I think we can usefully make that interval longer than a day and meaningfully increase the number of requests served from the CDN instead of the origin.

### DIFF
--- a/core/base-service/base-static.js
+++ b/core/base-service/base-static.js
@@ -47,7 +47,11 @@ export default class BaseStaticService extends BaseService {
       const format = (match.slice(-1)[0] || '.svg').replace(/^\./, '')
       badgeData.format = format
 
-      setCacheHeadersForStaticResource(ask.res)
+      let maxAge = 24 * 3600 // 1 day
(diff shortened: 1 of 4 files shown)


PyvesB (approved):
Reasoning makes sense, let's go with this! 👍🏻

Issue #10249, opened 2024-06-12
Exact values


chris-y:
:clipboard: **Description**

On the Lemmy badge (and I assume others), the value displayed is rounded up to the nearest 1.1k over 1000. Is it possible to display the exact values up to a higher figure?

![339097396-c33cfe54-f6f6-4df7-8472-0c9a72a446ba.png](https://github.com/badges/shields/assets/1818803/db6eee00-05eb-4c3e-b7a9-6a4f20294c40)

M4rtineski:
Lemmy is a small platform and we find exact numbers rather valuable because of hard it is to get them.

PyvesB:
Hello 👋🏻 

The code that governs how numbers get rendered lives here:
https://github.com/badges/shields/blob/05f0d866dace36d1e565f6670d69262f9e0b08db/services/text-formatters.js#L89

* we could change it. However this would also impact many of our other service badges.
* we could make a special case for Lemmy. This would set a precedent, which could risk increasing code complexity and reducing the consistency of our badges in the long run.

One argument for changing it for all badges is that number with unit abbreviations such as `k`, `G`, `M`, etc. are arguably harder to read than their digit-only counterparts. `1.1k` technically has the same number of characters as `1187`, so we could change the algorithm slightly to not display any abbreviations up to 9999, and leave the same behaviour for all higher numbers.

Would be keen on hearing what other maintainers think.

chris48s:
So there are probably a few badges where we show a number of things exactly rather than using `metric()`, but I'd say those are a result of oversight/inconsistency than design. If we can pick out counter-examples where we always show an exact number, we should fix them to use metric.

I think there's a few things to take into account if we are looking at changing `metric()` aside from number of characters. In fact, I'd never really thought about number of characters as a consideration myself:

- Readability: I'm not sure I have a totally clear view on whether it is easier to parse `1.1k` or `1,144` at a glance. I do think `1,144` is preferable to `1144` though. Either way, I'd say readability is the main thing to optimise for here. I'd be really interested if there is any research on this topic as I don't know the answer but I'd be happy to make a decision one way or the other based on evidence.
- One of the main reasons for rounding is that the kind of stats we're reporting (number of stats/download/users/whatever) are indicative rather than authoritative. Do you make a different decision knowing that a package has "1,144" stars rather than "about ~1,100" or knowing that it had "1,144" downloads in the last month rather than "about ~1,100". Probably not.
[...]

PR #10406, opened 2024-07-24
use metric() for [discord] and [revolt] badges

chris48s:
Refs https://github.com/badges/shields/issues/10249#issuecomment-2171684953

These badges display exact values, but they should really use `metric()`


PyvesB (approved):
Looks good! 👍🏻

Issue #6011, opened 2021-01-01
Badge Request: Crates.io crate size


cgbur:
Crates.io crate size badge. 

Example for the [tokio](https://crates.io/crates/tokio) crate: ![Crates.io](https://img.shields.io/badge/dynamic/json?color=red&label=size&query=versions%5B0%5D.crate_size&url=https%3A%2F%2Fcrates.io%2Fapi%2Fv1%2Fcrates%2Ftokio%2Fversions)

That one was generated from a dynamic shield 
```
https://img.shields.io/badge/dynamic/json?color=red&label=size&query=versions%5B0%5D.crate_size&url=https%3A%2F%2Fcrates.io%2Fapi%2Fv1%2Fcrates%2Ftokio%2Fversions
```

The issue with this method is that the file sizes that are returned are in bytes which is pretty inconvenient. Ideally, you could just re-use the size conversion function that is used for NPM and github sizes. You can of course specify the version you desire just like how you handled the other crates.io shields. It looks like they are using 10^3 for kB instead of 2^10 so that would be whatever you guys use as a standard or whatever matches crates.io.

The returned information can be gotten by something like:
```javascript
// https://crates.io/api/v1/crates/:crate-name/versions
fetch("https://crates.io/api/v1/crates/tokio/versions").then(res => console.log(res.versions[0].crate_size))
```

This would be useful for easily gauging the size of the repositories you are adding straight from the git page and can also be used as a selling point if small sizes are something you are proud of.

calebcartwright:
Thanks for the request @cgburgess, sounds reasonable to me. We have a great [tutorial](https://github.com/badges/shields/blob/master/doc/TUTORIAL.md) if you or anyone else is interested in implementing this badge.

It should be pretty easy to do so, especially since it's the same API endpoint and response our other crates badges work with

The schema would need to be updated here:
https://github.com/badges/shields/blob/61d2b0d07f9d67babbc82303069550914f395b44/services/crates/crates-base.js#L9-L32

and then a new service class and tester would need to be added under the crates directory.

Some things that may need to be considered however:
* Should this support the size for a specific crate version as well as the "latest" version (e.g. routes `/crates/size/:crate` and `/crates/size/:crate/:version`)
* What does the version at index 0 represent: `max_version` or `_newest_version`(presumably just like `semver` vs `date`) and do we need a query param here (this question would likely apply to the existing download count badge as well)

cgburgess:
Awesome I will probably take a stab at implementing this over the weekend.

Issue #9703, opened 2023-11-01
Incorrect image size for Docker Hub


Feriman22:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

The Docker Image Size badge displays an incorrect size of a Docker Image. In the example, the badge shows 4.21MB, but the size of the latest version is only 4.02MB.

![image](https://github.com/badges/shields/assets/22872515/a92f5933-4cba-472d-b219-579cd5389ade)

![image](https://github.com/badges/shields/assets/22872515/0da3e7fe-8037-45e1-9608-511437b49130)

### 🔗 Link to the badge

[![Docker Image Size](https://img.shields.io/docker/image-size/feriman25/freednsupdater/latest?logo=docker&style=for-the-badge)](https://hub.docker.com/r/feriman25/freednsupdater/tags)

calebcartwright:
Haven't looked closely at the API response from Docker Hub, but suspect this is just Mebibytes vs. Megabytes

Feriman22:
Yes, it's possible. However, the badge should show the same value as on the Docker Hub.

chris48s:
Yes it is that. The size of the image is reported in bytes via the DockerHub API. This image is 4210627 bytes

4210627/1024/1024 = 4.015...
4210627/1000/1000 = 4.210...

In general, we use https://github.com/sindresorhus/pretty-bytes#pretty-bytes for formatting, which uses kilobytes. I think if someone wants to add a kibibytes formatter which can be used for size badges where the upstream convention is to report kibibytes, we can take a look at a PR.

calebcartwright:
I agree it would be good for us to support this, though I'd prefer that we do so via providing conditional support to give users control. That's because this is an instance where I think we should try to be consistent across our own ecosystem by default and not trying to match what upstream services do

PR #10421, opened 2024-07-28
[Crates] Added crate size badge

einstein8612:
Proposed in https://github.com/badges/shields/issues/6011

Didn't make the issue but last message was 2021 so I figured it was safe and I needed this.

Routes:
Crates.io Size: /crates/size/:crate
Crates.io Size (version): /crates/size/:crate/:version


einstein8612:
Also, while doing this I realised that shields don't use **bibytes but divide by 1000 instead. I just followed what the rest of the badges do but are there any plans of adding support for this? Since there's now an inconsistency between the badge's number and the one on crates.io/docker/github repo size

calebcartwright:
> Also, while doing this I realised that shields don't use **bibytes but divide by 1000 instead. I just followed what the rest of the badges do but are there any plans of adding support for this? Since there's now an inconsistency between the badge's number and the one on crates.io/docker/github repo size

Will point back to https://github.com/badges/shields/issues/9703#issuecomment-1789630736 as I believe our position is still the same.

Our main goal is consistency within the Shields ecosystem, but we can certainly consider offering support for these types of variances when it makes sense

services/crates/crates-size.service.js
```diff
@@ -0,0 +1,56 @@
...
+      },
+    },
+  }
+
+  render({ size }) {
+    return {
+      label: 'size',
+      message: size ? prettyBytes(size) : 'unknown',
+      color: size ? 'blue' : 'lightgray',
+    }
```
chris48s, line 46: In general, this looks good.

This is a bit of a minor point but instead of this, could we handle the case where `size` is not set by validating in `handle()` and throwing an exception, and then `render()` is just renders the valid response.

So we check something like:

```js
if (this.constructor.getVersionObj(json).crate_size == null) {
    throw new InvalidResponse({ prettyMessage: 'unknown' })
}
```

in `handle` and then `render` just does

```js
render({ size }) {
  return {
    label: 'size',
    message: prettyBytes(size),
    color: 'blue'
  }
}
```

chris48s:
> > Also, while doing this I realised that shields don't use **bibytes but divide by 1000 instead. I just followed what the rest of the badges do but are there any plans of adding support for this? Since there's now an inconsistency between the badge's number and the one on crates.io/docker/github repo size
> 
> Will point back to https://github.com/badges/shields/issues/9703#issuecomment-1789630736 as I believe our position is still the same.

Thanks for digging that comment out. I knew it had come up before. Lets keep the scope of this PR to just merging this in its current form. I am going to write up another issue about this topic. I do think it is worth doing though.

chris48s:
I've raised https://github.com/badges/shields/issues/10437 and laid out my thoughts on this.

services/crates/crates-size.service.js
einstein8612: Yep I agree and have made the changes, thanks for pointing that out

chris48s (approved):
thanks for working on this

PR #10438, opened 2024-08-02
[Crates] Implement Dependents Badge

### LINKED ISSUE #10204: crates.io dependents
### 📋 Description

It's good to have a badge showing the number of reverse dependencies of a crate in crates.io.

### 🔗 Data

https://crates.io/api/v1/crates/:crate/reverse_dependencies
Then in the `meta::total` field, you can get the total number of `reverse dependencies`.

### 🎤 Motivation

This is an important indicator of the popularity of a project.

erayerdin:
Closes #10204 

 - FYI: `librariesio-dependents.service.js` helped me implementing this service.

### DIFF
--- a/services/crates/crates-dependents.service.js
+++ b/services/crates/crates-dependents.service.js
@@ -1,0 +1,49 @@
+import Joi from 'joi'
+import { pathParams } from '../index.js'
+import { metric } from '../text-formatters.js'
+import { nonNegativeInteger } from '../validators.js'
+import { BaseCratesService, description } from './crates-base.js'
+
+const dependentsResponseSchema = Joi.object({
+  meta: Joi.object({
+    total: nonNegativeInteger,
+  }).required(),
+}).required()
+
+export default class CratesDependents extends BaseCratesService {
+  static category = 'other'
+  static route = { base: 'crates/dependents', pattern: ':crate' }
+
+  static openApi = {
+    '/crates/dependents/{crate}': {
+      get: {
+        summary: 'Crates.io Dependents',
+        description,
+        parameters: pathParams({
+          name: 'crate',
+          example: 'tokio',
+        }),
+      },
+    },
+  }
+
+  static render({ dependentCount }) {
+    return {
+      label: 'dependents',
+      message: metric(dependentCount),
+      color: dependentCount === 0 ? 'orange' : 'brightgreen',
+    }
+  }
+
+  async fetch({ crate }) {
+    const url = `https://crates.io/api/v1/crates/${crate}/reverse_dependencies`
+    const schema = dependentsResponseSchema
+    return this._requestJson({ schema, url })
+  }
+
+  async handle({ crate }) {
+    const json = await this.fetch({ crate })
+    const { total: dependentCount } = json.meta
+    return this.constructor.render({ dependentCount })
+  }
+}
--- a/services/crates/crates-dependents.tester.js
+++ b/services/crates/crates-dependents.tester.js
@@ -2,13 +2,12 @@ import { isMetric } from '../test-validators.js'
 import { createServiceTester } from '../tester.js'
 export const t = await createServiceTester()
 
-t.create('dependent count').timeout(10000).get('/tokio.json').expectBadge({
+t.create('dependent count').get('/tokio.json').expectBadge({
   label: 'dependents',
   message: isMetric,
 })
 
 t.create('dependent count (nonexistent package)')
-  .timeout(10000)
   .get('/foobar-is-not-crate.json')
   .expectBadge({
     label: 'crates.io',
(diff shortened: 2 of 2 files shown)


services/crates/crates-dependents.service.js
```diff
@@ -0,0 +1,49 @@
+import Joi from 'joi'
+import { pathParams } from '../index.js'
+import { metric } from '../text-formatters.js'
+import { nonNegativeInteger } from '../validators.js'
+import { BaseCratesService, description } from './crates-base.js'
+
+const dependentsResponseSchema = Joi.object({
```
erayerdin, line 7: Dependents of a crate is served under `/api/v1/crates/:crate/reverse_dependencies`. The schema is specific to this service, that's why I thought it'd be better if it stayed in this module rather than `crates-base.js`.

services/crates/crates-dependents.service.js
```diff
@@ -0,0 +1,49 @@
...
+
+  static render({ dependentCount }) {
+    return {
+      label: 'dependents',
+      message: metric(dependentCount),
+      color: dependentCount === 0 ? 'orange' : 'brightgreen',
+    }
+  }
+
+  async fetch({ crate }) {
```
erayerdin, line 38: Endpoint is `/api/v1/crates/:crate/reverse_dependencies`, so `fetch` had to be implemented manually as `BaseCratesService.fetch` fetches the package meta.

services/crates/crates-dependents.service.js
chris48s: agreed :+1:

services/crates/crates-dependents.service.js
chris48s: agreed :+1:

services/crates/crates-dependents.tester.js
```diff
@@ -0,0 +1,16 @@
...
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('dependent count').timeout(10000).get('/tokio.json').expectBadge({
+  label: 'dependents',
+  message: isMetric,
+})
+
+t.create('dependent count (nonexistent package)')
+  .timeout(10000)
```
chris48s, line 11: Do we expect this to be really slow in the not found case? I tried a couple of calls and they seemed to return a not found response pretty quickly

services/crates/crates-dependents.service.js
```diff
@@ -0,0 +1,49 @@
...
+
+const dependentsResponseSchema = Joi.object({
+  meta: Joi.object({
+    total: nonNegativeInteger,
+  }).required(),
+}).required()
+
+export default class CratesDependents extends BaseCratesService {
+  static category = 'other'
+  static route = { base: 'crates/dependents', pattern: ':crate' }
```
chris48s, line 15: If you have a look over the other rust crates badges, we accept an optional `:version?` param which allows us to present the statistic for a specific version.

Could we also do that here, or does this `reverse_dependencies` endpoint not allow us to do that in this case?

chris48s (commented):
Thanks. Couple of minor points but this is looking pretty good

services/crates/crates-dependents.tester.js
erayerdin: I've taken the test code from `librariesio-dependents.service.js` and left the timeout pretty much as is. I don't think it should take that long to respond. Maybe a second is good enough? What do you think?

services/crates/crates-dependents.service.js
erayerdin: The API does not have any documentation to begin with. I tried `GET`ting these URLs with `tokio` and `1.39.2` version but only got 404.

 - `https://crates.io/api/v1/crates/tokio/v1.39.2/reverse_dependencies/`
 - `https://crates.io/api/v1/crates/tokio/1.39.2/reverse_dependencies/`
 - `https://crates.io/api/v1/crates/tokio/reverse_dependencies/v1.39.2/`
 - `https://crates.io/api/v1/crates/tokio/reverse_dependencies/1.39.2/`

While there's no documentation for the API, there's actually a crate for the API. [This struct](https://docs.rs/crates_io_api/latest/crates_io_api/struct.ReverseDependencies.html) hints no version at all, so it's almost guaranteed to say crates.io does not support versioning with reverse dependencies.

services/crates/crates-dependents.tester.js
chris48s: The default timeout is 4 seconds. In some cases we need to raise it if the upstream endpoint is really slow.

Lets remove this timeout statement and go with the default.

services/crates/crates-dependents.service.js
chris48s: OK, cool. No worries. Thanks for looking into it.

services/crates/crates-dependents.tester.js
erayerdin: It's done. :+1:

chris48s (approved):
thanks for working on this

PR #8862, opened 2023-01-27
build docker images with buildkit v0.9.1

chris48s:
Refs

- https://community.fly.io/t/deployments-failing-using-docker-hub-image/10376/2
- https://community.fly.io/t/deploying-to-fly-via-github-action-failing/10171/19

PR #10453, opened 2024-08-10
un-pin buildx version

chris48s:
We originally did this in https://github.com/badges/shields/pull/8862 due to issues deploying on fly.io

This problem has actually been fixed upstream for quite some time
https://community.fly.io/t/images-built-with-the-latest-docker-buildx-and-or-github-actions-work-again/10816
so we don't really need to be pinned to this old version any more.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -14,8 +14,6 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
-        with:
-          version: v0.9.1
 
       - name: Set Git Short SHA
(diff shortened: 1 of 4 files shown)


chris48s:
Upgrading this suggests some changes we should make

https://github.com/badges/shields/pull/10453/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

I'm converting this to draft for now. I will pick these up tomorrow.

chris48s:
I've pushed 3 commits solving the 3 flagged issues with the Dockerfile:

- https://docs.docker.com/reference/build-checks/stage-name-casing/
- https://docs.docker.com/reference/build-checks/legacy-key-value-format/
- https://docs.docker.com/reference/build-checks/json-args-recommended/

PyvesB (approved):

PR #10466, opened 2024-08-18
Deprecate [TreewareTrees] service

PyvesB:
The Treeware API has been unresponsive for some time now. I reported the issue a month ago in https://github.com/TreewareEarth/treeware.earth/issues/35, but haven't heard back. Given low usage numbers (43 renders last Thursday for example), I think we can go ahead and start the deprecation process.

Cc @owenvoke who contributed the badge in the first place.

### DIFF
--- a/services/treeware/treeware-trees.service.js
+++ b/services/treeware/treeware-trees.service.js
@@ -1,65 +1,11 @@
-import crypto from 'crypto'
-import Joi from 'joi'
-import { metric } from '../text-formatters.js'
-import { floorCount } from '../color-formatters.js'
-import { BaseJsonService, pathParams } from '../index.js'
+import { deprecatedService } from '../index.js'
 
-const apiSchema = Joi.object({
-  total: Joi.number().required(),
-}).required()
-
-export default class TreewareTrees extends BaseJsonService {
-  static category = 'other'
-
-  static route = {
+export const TreewareTrees = deprecatedService({
+  category: 'other',
+  route: {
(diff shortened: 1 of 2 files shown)


chris48s (approved):
maybe worth leaving this open for a bit to see if the original author responds, but happy for you to merge it when the time is right

owenvoke:
Hi @PyvesB and @chris48s, the Offset Earth platform was renamed to Ecologi.

We should be able to update the `offset.earth` domain to `ecologi.com` and it will start working again. 👍🏻

PyvesB:
What would be the difference with our existing [ecologi badges](https://github.com/badges/shields/blob/master/services/ecologi/ecologi-trees.service.js)? Have Ecologi and Treeware essentially merged into one project?

chris48s:
One thing I'll quickly note here is: the treeware badge uses a user/package format whereas the ecologi one takes a username only

owenvoke:
Yeah, so packages with Treeware are recorded by package name (i.e. using the `ref` query param, which is an MD5 hash of `vendor/package`).

In theory, we could proxy this and add a `ref` param to the Ecologi badge (which can actually be used for referencing anyone who has bought trees).

Or, we can keep this separate badge which nicely encodes specifically for packages (which I think works a bit cleaner). 🤷🏻

PyvesB:
Thanks for the additional context! I think it makes sense to leave things as they are and simply switch the API endpoint. I've opened #10467 accordingly.

PR #10476, opened 2024-08-25
Publish linux/amd64 docker images for snapshot builds only

smashedr:
I did some research and it seems the correct way to publish a multi-architecture build under a single tag is to do it simultaneously.

```
0 shane@jammy [/home/shane]$ docker pull shieldsio/shields:next
next: Pulling from shieldsio/shields
no matching manifest for linux/amd64 in the manifest list entries
```

That will make the entire build take the full duration of ~35 minutes. At a glance I don't see master being merged to more frequently than 35 minutes apart, so I have also included the next build here, but that can easily be removed.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -17,9 +17,6 @@ jobs:
       contains(github.event.pull_request.labels.*.name, 'release')
 
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        architecture: [amd64, arm64]
 
     steps:
       - name: Get current date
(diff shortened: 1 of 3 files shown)


chris48s:
Merging lots of PRs close together is something we do frequently e.g:

Screenshot at 2024-08-26 21-57-45

Even if we didn't, having to wait ~35 minutes from merging something to having an image ready to deploy every time is just going to be too annoying. I think if the only way to do this is to push both together, we go with what I suggested in https://github.com/badges/shields/pull/10441#issuecomment-2308442580 - arm64 on snapshot tags only.

smashedr:
Ok, I can remove the arm64 from next, and just leave it on the release branch build.

chris48s (approved):
Cool. I am planning to do the next snapshot release later today so we'll test the process and (hopefully) have a working dual arch image available this time.

chris48s:
Screenshot at 2024-09-02 16-03-11

smashedr:
Looks good :white_check_mark:
I was able to run `shieldsio/shields:server-2024-09-02` on my development server and deploy it to my arm64 swarm server.

Issue #10244, opened 2024-06-10
OSS Lifecycle GitLab


JoeIzzard:
### 📋 Description

There are already OSS Lifecycle and OSS Lifecycle Branch badges for GitHub, however it would be great to have a version that is compatible with GitLab. This would look and work like the existing OSS Lifecycle badge:
![Static Badge](https://img.shields.io/badge/oss_lifecycle-active-green)

### 🔗 Data

The data for this would be pulled from the Raw GitLab content for that repo. On GitLab this takes the form:
https://gitlab.com/{USER}/{REPO}/-/raw/{BRANCH}/{FILE}?ref_type=heads

### 🎤 Motivation

I largely use GitLab and would like to be able to use the badge.

chris48s:
We have a number of badges that just look at a file which could be anywhere but are hard-coded to assume that file is on GitHub. I would really like to gradually move these to being agnostic about where the file is hosted. So they will work for projects on GitHub, GitLab, BitBucket, Forgejo, your self-hosted instance of Gitea.. all the places. OSSLifeCycle is one of these.

I think my suggestion here would be that we migrate this so that instead of

https://img.shields.io/osslifecycle/Netflix/osstracker

the URL becomes

https://img.shields.io/osslifecycle?file=https%3A//github.com/Netflix/osstracker/blob/master/OSSMETADATA

but `file` can live anywhere. To maintain backwards compatibility for existing users, we can use a redirect to redirect calls to `osslifecycle/{user}/{repo}/{branch}` to `https://img.shields.io/osslifecycle?file=https%3A//github.com/{user}/{repo}/blob/{branch}/OSSMETADATA`.

JoeIzzard:
I have been having a look at this, but have run into some issues. Has this been implemented for any other endpoints I can look over as an example implementation?

So far I have added a `file_url` query parameter, and some logic so that if it is provided we fetch that otherwise use the existing URL structure.

The issue I am finding is with creating the bare `osslifecycle` without path parameters. Removing the `pattern` aspect of the route results in an error which I believe is an assertion error for having less than one extension, crash log below:

<details>
<summary>Crash Log</summary>
<pre>
[server] /Users/jizzard/Desktop/Projects/Personal/Open Source/Shields/node_modules/joi/lib/errors.js:193
[server]         return new exports.ValidationError(message, details, original);
[server]                ^
[server] 
[server] Error [ValidationError]: Route for OssTracker {
[server]   "base": "osslifecycle",
[...]

PR #10489, opened 2024-09-02
[OSSLifecycle] Add file_url param to pull from non-github sources

JoeIzzard:
This PR addresses #10244 (OSS Lifecycle GitLab) by adding a queryParam of `file_url`. When this parameter is provided, we fetch the provided URL instead of defaulting to the GitHub RAW Content URL. Existing badges using the `{user}/{repo}` and `{user}/{repo}/{branch}` pattern are maintained, however providing `file_url` with these will still override the URL we fetch.

UI Changes:
- Added a `OSS Lifecycle (file)` OpenAPI entry for the new end point without path parameters
- Changed `OSS Lifecycle` to `OSS Lifecycle (GitHub)`
- Changed `OSS Lifecycle (branch)` to `OSS Lifecycle (GitHub branch)`

Changes to tests:
- Added new test to test using `file_url` against the Netflix OSS Tracker Repo

One potential improvement would be to rename the service class, currently it is `OssTracker` however everywhere else is referred to as `osslifecycle`.

### DIFF
--- a/services/osslifecycle/osslifecycle.service.js
+++ b/services/osslifecycle/osslifecycle.service.js
@@ -1,4 +1,11 @@
-import { BaseService, InvalidResponse, pathParams } from '../index.js'
+import Joi from 'joi'
+import { optionalUrl } from '../validators.js'
+import {
+  BaseService,
+  InvalidResponse,
+  pathParams,
+  queryParam,
+} from '../index.js'
 
 const description = `
 OSS Lifecycle is an initiative started by Netflix to classify open-source projects into lifecycles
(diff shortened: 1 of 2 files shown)


JoeIzzard:
I didn't realise there was a redirector method, makes sense to do it that way.

I have stripped back the `osstracker` to just include fetching from any URL with `file_url`. All the existing checks are now using the `file_url` method so was able to remove the test I added in my first version. Added a redirector to handle the old methods, and added two tests to test with and without the branch parameter.

services/osslifecycle/osslifecycle.service.js
```diff
@@ -9,49 +11,33 @@ single line similar to the following: \`osslifecycle=active\`. Other suggested v
 can be viewed on the [OSS Tracker repository](https://github.com/Netflix/osstracker).
```
chris48s, line 11: I can't leave a comment on the exact line because it hasn't been changed in the diff, but can we remove "GitHub" from the docs here

```diff
- simply create an OSSMETADATA tagging file at the root of your GitHub repository containing a
+ simply create an OSSMETADATA tagging file at the root of your repository containing a
```

services/osslifecycle/osslifecycle.service.js
```diff
@@ -87,17 +73,15 @@ export default class OssTracker extends BaseService {
...
-  async fetch({ user, repo, branch }) {
+  async fetch({ fileUrl }) {
     return this._request({
-      url: `https://raw.githubusercontent.com/${user}/${repo}/${branch}/OSSMETADATA`,
+      url: fileUrl,
     })
   }
 
-  async handle({ user, repo, branch }) {
+  async handle({ user, repo, branch }, { file_url: fileUrl = '' }) {
```
chris48s, line 82:
```suggestion
  async handle(pathParams, { file_url: fileUrl = '' }) {
```
We have to put something here because `queryParams` is the second argument to `handle()` but we're not actually using `user`, `repo` or `branch` in this function.

chris48s (commented):
Thanks. Couple of minor changes then this is good to merge

PR #10492, opened 2024-09-03
Add font to docs

rivercory:
Added Poppins font for better readability!

### DIFF
--- a/frontend/src/css/custom.css
+++ b/frontend/src/css/custom.css
@@ -4,6 +4,8 @@
  * work well for content-centric websites.
  */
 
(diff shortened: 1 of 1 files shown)


chris48s:
I think the choice docusaurus/infima have made (defaulting to `system-ui`) is really good as it makes it look "native" on a range of platforms. I don't see a good reason to diverge from that.
https://github.com/facebookincubator/infima/blob/450b41af48b2d4b670a483228b59360553885e3b/packages/core/styles/common/variables.pcss#L115-L117
I'll leave this open for a bit in case another maintainer wants to strongly disagree with me.

(If we _were_ to change this, we should not use fonts.googleapis.com. We should pull in from https://www.npmjs.com/package/@fontsource/poppins and provide some more thoughtful fallbacks but lets not get bogged down in that for now)

PyvesB:
> I think the choice docusaurus/infima have made (defaulting to system-ui) is really good as it makes it look "native" on a range of platforms. I don't see a good reason to diverge from that.

Agreed!

PR #10494, opened 2024-09-03
Modrinth Game Versions badge doesn't fetch all game versions. Fixes #10493

### LINKED ISSUE #10493: Modrinth Game Versions badge doesn't fetch all game versions

shields.io

### 🐞 Description

When using the Modrinth Game Versions badge, you would expect to get all supported game versions, but you only get the ones from the latest release

### 🔗 Link to the badge

https://shields.io/badges/modrinth-game-versions

### 💡 Possible Solution

The fix would be relatively simple. 
Only the function `async handle({ projectId })` in `services/modrinth/modrinth-game-versions.service.js` would need to change from:
```js
  async handle({ projectId }) {
    const { 0: latest } = await this.fetchVersions({ projectId })
    const versions = latest.game_versions
    return this.constructor.render({ versions })
  }
```
to:
```js
...
  async handle({ projectId }) {
    const data = await this.fetchVersions({ projectId })
    const versions = [...new Set(
      data.map(ver => ver.game_versions)
          .reduce((prev, curr) => [...prev, ...curr]))
    ].sort()
    return this.constructor.render({ versions })
  }
```

Ceedrich:
This PR fixes the issue #10493.

### DIFF
--- a/services/modrinth/modrinth-game-versions.service.js
+++ b/services/modrinth/modrinth-game-versions.service.js
@@ -38,8 +38,14 @@ export default class ModrinthGameVersions extends BaseModrinthService {
   }
 
   async handle({ projectId }) {
-    const { 0: latest } = await this.fetchVersions({ projectId })
-    const versions = latest.game_versions
+    const data = await this.fetchVersions({ projectId })
+    const versions = [
+      ...new Set(
+        data
+          .map(ver => ver.game_versions)
+          .reduce((prev, curr) => [...prev, ...curr]),
+      ),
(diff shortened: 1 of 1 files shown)


chris48s:
lets resolve the discussion in the issue thread before we look at this one https://github.com/badges/shields/issues/10493#issuecomment-2327243897

PyvesB:
There hasn't been any movement for over a year, there's not much point in keeping the PR open. We can always circle back to this if the community makes progress on the linked thread. :)

PR #10488, opened 2024-09-01
perf: improve logoSize performance

LitoMore:
I replaced the `svg-path-commander` with `svg-path-bbox` and `svgpath`.

Let's use these logos below as examples:

```
https://img.shields.io/badge/test-blue?logo=lerna&logoSize=auto
https://img.shields.io/badge/test-blue?logo=pugb&logoSize=auto
```

The `svg-path-commander` takes about 5000 ms to process the icon. But for `svg-path-boxx` + `svgpath`, it only needs 60 ms. (on my machine)

PR #10497, opened 2024-09-03
perf: faster `resetIconPosition` avoiding to parse path twice

mondeja:
This is a continuation of #10488.

Just seen that PR and realized that the parsing of the path (d) is done twice. One in `svgpath(path)` and another in `svgPathBbox(path)` as `svgPathBbox` uses `svgpath` internally.

So I released a [new version of `svg-path-bbox`](https://github.com/mondeja/svg-path-bbox/releases/tag/v2.1.0) that also accepts `SvgPath` interfaces to avoid the double parsing in situations like this. Instead of parsing again, `svgpath` clones the segments and is so much faster.

### DIFF
--- a/lib/svg-helpers.js
+++ b/lib/svg-helpers.js
@@ -19,9 +19,10 @@ function getIconSize(iconKey) {
   return { width: x1 - x0, height: y1 - y0 }
 }
 
-function resetIconPosition(path) {
+function resetIconPosition(d) {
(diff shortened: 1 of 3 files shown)


LitoMore (approved):

LitoMore:
I just tested this on my machine, it optimized the process from 60 ms to 38 ms.

chris48s (approved):

Issue #9391, opened 2023-07-18
Cannot open the Nav Bar on Google Chrome (Android 13)


ascopes:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Opening the navbar on Google Chrome 114.0.5735.196, Android 13; LE2123 Build/TP1A.220905.001 does not work.

The search bar function when viewing on desktop mode also does not work.

This effectively prevents me using anything on the website other than making static badges, whereas previously I was able to generate badges on mobile easily.

https://github.com/badges/shields/assets/73482956/b774827b-2ec5-44c1-a46c-8b557b91f70a

### 🔗 Link to the badge

N/A

### 💡 Possible Solution

N/A

uncenter:
I'm also seeing that... I tested it on desktop and it was broken too for a minute but seemed to resolve... now I'm seeing that some badges aren't working?

uncenter:
<img width="509" alt="image" src="https://github.com/badges/shields/assets/47499684/2574004c-078f-4493-bb32-3235f7576c10">
I'm seeing this on multiple devices now and I assume this is the issue...

ascopes:
seems odd that a SVG would be a script?

calebcartwright:
Not sure if this is related (or just on my machine), but I'm no longer seeing any of the badge builder content on the right; there is no execute button and associated fields

uncenter:
> Not sure if this is related (or just on my machine), but I'm no longer seeing any of the badge builder content on the right; there is no execute button and associated fields

Yep I believe I'm seeing that now too.

chris48s:
OK. This was a real head-scratcher, but I worked out what is happening.

Basically, we had ended up in a state where some of the containers in the cluster were running one version of the application, and some were running another version. This meant that when a browser made a request for `main.1d01d963.js` (for example), that could have either been:

- a valid javascript file or
- a 404 error (badge)

..depending on which container served it. This lead to obvious breakage!

The good news is: I've re-deployed. If you hit shields.io right now, the frontend should work as expected because every container is running the same version of the application.

The bad news is: Re-deploying isn't a permanent fix. As things stand, that issue is going to happen again at some point.

I think further work/discussion on this (how do we prevent getting into this state again) will move to the
:detective: super secret devops place 🤫
as it is a deployment concern, rather than an issue to be solved at the codebase level.

Thanks for the bug report flagging this issue though.

chris48s:
Closing this as it is now resolved

Issue #3558, opened 2019-06-11
Tag Dockerhub releases


nigredo-tori:
:clipboard: **Description**

We already have a `Dockerfile`, why make users [build it](https://github.com/badges/shields/blob/544d7d3648d8fa97643d7d387cb8e96eb5ddfec6/doc/self-hosting.md#docker) manually? There are, of course, some custom `shields` images floating around, but those seem to be poorly maintained/supported.

A relatively easy way to do this would be to set up an automatic build on Dockerhub, as described [here](https://docs.docker.com/docker-hub/builds/), which would build a tagged image for each Git tag. There are also some automatic tests there, though I can't find a simple way to pipe their results back to GitHub PRs.

An alternative would be to wire up image building/publishing in the CI script.

calebcartwright:
Thanks @nigredo-tori - providing an "official" Shields docker image is something we want to do, and is on our roadmap. No timeline on delivering it though

niccokunzmann:
> An alternative would be to wire up image building/publishing in the CI script.

Dockerhub can build images if one sends a request to a certain URL. I am in favor of having the CI-Script send the request to build because then, the docker image only gets build after the tests are successful.
I never built docker images for tags only, so I do not know how to configure this on dockerhub. Fromwhat I saw, I think, it is quite easy.

nigredo-tori:
>  I am in favor of having the CI-Script send the request to build because then, the docker image only gets build after the tests are successful.

You seem to be talking about **build triggers**. It [should be possible](https://github.com/docker/hub-feedback/issues/330#issuecomment-138886896) to trigger the build for a specific reference. However the lack of any official documentation for these is a major red flag.

>  I never built docker images for tags only, so I do not know how to configure this on dockerhub. Fromwhat I saw, I think, it is quite easy.

Extremely. It's a Branch/Tag combobox, plus some simple string interpolation for building the Docker tag string.

paulmelnikow:
Agreed this is a good idea.

We've had some discussion about providing a changelog for self-hosting, to correspond with tagged releases, probably using the date. So far no one has laid the groundwork for that. Perhaps we could run docker builds for those tagged releases.

Running another docker build for master also seems like a good idea; it might be an easy way to get a daily test of the docker build (#2816); alternatively maybe we can run the test within Circle.

paulmelnikow:
[...]

PR #10498, opened 2024-09-04
Add latest Tag to Docker Release

smashedr:
One other change I would like to see, is adding a `latest` tag to the docker image that always points to the latest version published.

This will allow easily updating self-hosted deployments without having to make code changes every time.

I also updated the setting of the `date` output that is soon to be deprecated with the new format. Reference https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - name: Get current date
         id: date
-        run: echo "::set-output name=date::$(date --rfc-3339=date)"
+        run: echo "date=$(date --rfc-3339=date)" >> $GITHUB_OUTPUT
 
       - name: Checkout branch "master"
         uses: actions/checkout@v4
@@ -51,7 +51,7 @@ jobs:
           context: .
           push: true
           platforms: linux/amd64,linux/arm64
-          tags: shieldsio/shields:server-${{ steps.date.outputs.date }}
+          tags: shieldsio/shields:server-${{ steps.date.outputs.date }},shieldsio/shields:latest
           build-args: |
             version=server-${{ steps.date.outputs.date }}
 
@@ -68,6 +68,6 @@ jobs:
           context: .
           push: true
           platforms: linux/amd64,linux/arm64
-          tags: ghcr.io/badges/shields:server-${{ steps.date.outputs.date }}
+          tags: ghcr.io/badges/shields:server-${{ steps.date.outputs.date }},shieldsio/shields:latest
           build-args: |
             version=server-${{ steps.date.outputs.date }}


.github/workflows/create-release.yml
```diff
@@ -68,6 +68,6 @@ jobs:
           context: .
           push: true
           platforms: linux/amd64,linux/arm64
-          tags: ghcr.io/badges/shields:server-${{ steps.date.outputs.date }}
+          tags: ghcr.io/badges/shields:server-${{ steps.date.outputs.date }},shieldsio/shields:latest
```
chris48s, line 71:
```suggestion
          tags: ghcr.io/badges/shields:server-${{ steps.date.outputs.date }},ghcr.io/badges/shields:latest
```

chris48s:
I think having a rolling tag that points at the latest snapshot makes sense.

I reckon using `latest` for that is also probably sensible, but I wonder if you have a view on that @calebcartwright Currently we don't have a `latest` tag and `next` is pushed every time we commit to master.

Note: Before we merge this, update https://github.com/badges/shields/blob/master/doc/self-hosting.md#docker again

chris48s:
To be clear:

Are you saying you think we shouldn't use the name `latest` (because it is the default if no tag is specified) and we should call a rolling tag pointing to the most recent snapshot something else (so it is more desctiptive and users have to explicitly opt in to following that)?

Or are you completely against having a rolling tag pointing to the latest snapshot with any name?

calebcartwright:
TL;DR

Yes I'm _adamantly_ opposed to adding any `latest` tag

Yes I'm also _generally_ opposed to adding any new rolling tag (though I'm _not_ pushing for us to drop the existing rolling `next` tag)

I think our starting position needs to be opposition to adding any new tags, with the paradigm for any new proposals to change that position needing to sufficiently make the case as to why we should make an exception to that, as opposed to starting from the position of having to defend why we don't want a new tag

<hr />

If we're going to entertain this then I'd really like to get a better understanding of the problem we're trying to solve. To expand on my prior comment with some receipts, deploying images using the `latest` tag is a genuinely bad practice

https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy

> Note:
>
>You should avoid using the :latest tag when deploying containers in production as it is harder to track which version of the image is running and more difficult to roll back properly.
>
>Instead, specify a meaningful tag such as v1.42.0 and/or a digest.

That's a short and sweet summary, but the underlying reasons/blast radius from deploying rolling tags (`latest` being especially egregious) are a lot more extensive and I can expand on it with a few very real examples if needed. Suffice to say there's good reasons why this gets called out explicitly in places like the official kubernetes docs.

The docker images we produce are for the Shields server, so it's something that gets deployed as opposed to being used as a base image (where rolling tags _are_ less problematic) by developers using them to construct their own deployable images.

Accordingly, or at least from my perspective, if as a project we start producing a `latest` tag the side effects could very realistically include:

* we're now promoting/supporting a known anti-pattern
* there's more tags for consumers to have to process/consider
* there's an increased chance of our maintainer team having to spend time helping users troubleshoot issues that stem from the flaws/challenges with the `latest` tag

Conversely, I don't see what substantive gains would be achieved from doing this

shields.io needs to deploy basically every update we get, but i believe it's still the case that level of cadence would be extremely rare for self-hosters, many (perhaps most if not all?) of whom can just deploy and forget about it as their surface of badge consumption is much narrower than the entire corpus of shields.io users (e.g. how many self-hosting users absolutely urgently need #10489, #10497, etc.) 

i run a self-hosted shields instance within a highly regulated corporate environment that has resultant policies which set a relatively low ceiling on the age of running images which in turn results in me having to redeploy more frequently than needed from a shields-updates/features perspective. i'd posit my own self-hosted deployment cadence is much more frequent than the overwhelming majority of self-hosters.

i have no issues tweaking the variable i use to grab the new monthly tag. it's not something i find burdensome whatsoever, i wouldn't switch to any rolling tag to spare myself that highly marginal effort i go through every < 45 days, and if i was adequately motivated i'm sure i could script that away.

we provide the `next` tag, at least in part, so that users who still _really_ want a rolling tag anyway can do so, but as i recall we opted to intentionally name it something different than `latest` so that they also have to be very explicit on the pull side to do so, functionally opting in to the risks of deploying rolling tags (whereas `latest` would be implicit due to client behavior)

finally, our monthly server tags aren't anything special relative to `next` other than being more nicely bundled up with a changelog for more human friendly consumption. they don't undergo any different levels of testing nor rigor, it's just a "hey if you're a self-hoster considering an upgrade, here's a more easily digestable view of a time-based batch of commits".

As such, if there's a self-hosting use case that wants to completely automate the process of upgrading their self-hosted instance by removing the human element of specifying a calver tag, why can't they use the existing, rolling `next` tag? what value are they gaining by having a rolling `latest` or rolling `foobar` tag above and beyond the `next` tag?

e.g. If there's a self-hoster that deployed a `latest` tag which at the time was aligned to the `server-2024-06-01` tag and we subsequently had some security vulnerability and fix that went out in the `server-2024-11-01` tag, does their original deployment with the `latest` tag offer them anything above and beyond the `next` tag as far as making it easier for them to determine whether they have the fix in their local environment or need to redeploy? I don't really think so, I think they'll still be in the position of having to compare image digests or having to explicitly force a clean deploy (either on new infra or ensuring the local image cache is nuked in the running environments)

chris48s:
Having shot us in the foot with `next` I'm on board with "don't deploy from rolling tags". Lesson learned from that one.

https://github.com/badges/shields/issues/9391#issuecomment-1646561840
https://github.com/badges/shields-ops/issues/34
https://github.com/badges/shields-ops/pull/35

Certainly not in an environment where you've got more than one instance in a cluster

> The docker images we produce are for the Shields server, so it's something that gets deployed as opposed to being used as a base image (where rolling tags are less problematic) by developers using them to construct their own deployable images.

I think this is quite a good point. First off in that situation (base image) you only pull from the upstream tag at build time and then you're locked once your image is built. But also, in a situation where the tag you're tracking is `node:20-alpine` or `python:3.10-bullseye` or something there is an expectation that tracking that tag is going to to give you bug/security fixes only and isn't going to make a BC-break. There's value in tracking the latest changes to that tag on every build. For something like this where we use CalVer rather than SemVer and any monthly release could in principle include a BC-break (despite our best efforts), I can see why it is more useful to nudge people to have a look at the changelog once a month, or however often they upgrade.

In any case. This is not an issue I have really strong views on. Hence tagging you in - I figured you have probably thought about this one than I have. I think I'll leave it with @smashedr to say why this would be good/useful if they want.

---

tangent

> as i recall we opted to intentionally name it something different than `latest`

When this issue popped up, it actually got me thinking about why we use `next` rather than `latest`. I went back to the thread https://github.com/badges/shields/issues/3558 and I think that is just what Paul happened to call it when he set it up tbh, unless the conversation happened elsewhere (discord?). Anyway, I couldn't find any reason for it that hasn't disappeared into the ether.

smashedr:
In a production environment, you should always pin versions.

But for a simple self-deployment, using a latest tag makes it very easy to update to the latest version. All I would have to do to update is click 1 button:

firefox-20240905-120430053

And as far as knowing what version is running, all I have to do is click on the image, and its clear as day:

firefox-20240905-120401409

But when pinning versions, I have to get the latest version tag, open my IDE, commit the changes, then redeploy the application: https://github.com/smashedr/shields.io

The `latest` tag is just a convenience for people who are not in production environments and don't want to pin versions.

calebcartwright:
@smashedr thanks for the quick response. i'll reply to a couple items inline below, but i do want to center on what i view as the main question that still remains unanswered (or at least unclear in my mind):

for your particular use case, what gaps/challenges do you have with the `next` tag that would prevent you from using that as your rolling tag? is there something that could be added/changed with the `next` tag that would make it viable for your use case?

<hr />

> But for a simple self-deployment, using a latest tag makes it very easy to update to the latest version. All I would have to do to update is click 1 button:

I get the simplicity that comes with a model that utilizes rolling tags, though I'm unsure what your personal distinction is between "simple self-deployment" and "production".

Regardless though, we have to think about the artifacts (which includes tags) that we produce, how those could potentially be consumed by any of our users, and the potential effects that could have on us. We can reasonably say that we don't know if/how the app may behave on e.g. powerpc or any other non-x86 instruction set architecture, and that we don't have the time/expertise/funds to be able to test, validate, and support, but we do already spend precious cycles supporting self-hosting users and extending our main image with more rolling tags (especially `latest`) has to big a risk of exacerbating that strain.

To be clear, I'm not suggesting that any and every usage of a rolling tag will result in a catastrophic explosion, and it may not be in an issue for your particular circumstances. But we can't focus solely on that, as it's not an image tag we'd be producing for your use case alone.

> And as far as knowing what version is running, all I have to do is click on the image, and its clear as day:

Good shout, i forgot we put that env var in there. Though worth noting that just makes troubleshooting a bit easier in that it simplifies the process of identifying one of the problematic situations (e.g. different versions running on different nodes) that can and do arise with deploying rolling tags

> But when pinning versions, I have to get the latest version tag, open my IDE, commit the changes, then redeploy the application: https://github.com/smashedr/shields.io
>
>The latest tag is just a convenience for people who are not in production environments and don't want to pin versions.

This is where I come back to trying to differentiate between one user, one use case (the perspective you're understandably focused on) versus all users, any use case (the perspective from which we have to evaluate this)

smashedr:
I am not asking anyone else to switch to using the rolling tag `latest`. Everyone who prefers to pin a version, can continue to use the `server-xxxx` tag. I just want it to be provided for those who know how and want to use it for their workflow.

All this does is add the ability for additional workflows to be used, vs the one you want people to use; pinning versions. It up to the end users to decide what they want in their environment.

There is a big difference in `next` and `latest`; as `next` is published on every commit to master, and `latest` will always point to the latest `server-xxxx` release. Additionally, `next` only builds `linux/amd64` so can't be used for `linux/arm64` deployments.

So again, all I am asking is to add the ability for people who want to use a `latest` tag, to be able too.

calebcartwright:
re: https://github.com/badges/shields/pull/10498#issuecomment-2334620721

For the sake of trying to expedite the conversation, I'll reiterate that I know what you're asking for. I understand what the PR is proposing, and I understand when and how we're publishing our tags, so we can leave that there  :+1: 

Your comment did include the answer to my question though which is helpful:

You need to deploy on arm, you'd prefer to use a rolling tag, and there's not a rolling arm-compatible tag today.

That's the objective. You've proposed one way that would achieve that, but that's not the only way we could provide a rolling tag that'll run on arm.

I'm going to go ahead and say that `latest` is not going to happen, and I don't think it's worth anyone's time to belabor that argument; the tag being named `later` vs `rolling` vs `monthly` vs `foobar` wouldn't make a difference one way or the other for your use case.

The main point I want to drill into now though to see what alternatives might be viable is this :point_down: 

> and I do not want to be on the bleeding edge. 

I'm curious what "beeding edge" means to you, specifically in this context. What do you think are the meaningful differences between our `next` tag and some `server-xxxx` tag, do you think one is any more stable or has gone through more extensive validation than the other (and if so how does that fit in with your prior characterizations about non-production)? Would you dispute that for large chunks some months those two tags have the exact same contents?

smashedr:
The name of the tag is irrelevant, I just want to see a rolling tag that I can use to make updating my self-hosted server much easier.

As far as the other tags, I don't fully understand how this project decides to make a `server-xxxx` release/tag, but I do know it is not just a mirror of the master branch and changes less frequently.

calebcartwright:
> It is pretty aribtrary. That is why I refer to it as a snapshot.
> In that sense, I would say there is a slightly higher level of "stability"

I'd partially agree but maybe with some nuance. I think snapshot is a fair description and feel it echoes my main original point I wanted to ensure was clear about it being arbitrary: `next` and `server-xxxx` are both point-in-time publishes from the `master` branch, with the exact same of tests executed, and there is some indeterminate amount of time each month when they can be identical; if you happen to deploy `server-xxxx` during the first week of the month (or if we've just not merged any PRs recently) then `server-xxxx` often is exactly `master` and exactly `next`

Where I'd still draw a little nuance is that I wouldn't want anyone to infer a label of "stability" on any of the server tags. We don't use them ourselves, and it's very conceivable that a `server-xxxx` tag, even the most recently published one, could have issues not present on Shields.io, as we're not going to go back and yank a bad `server-xxxx` tags and we're not necessarily promising to publish an updated `server` tag ad-hoc to address an issue (and a minor issue with badge Foo may seem not worth it to us, but which could be pretty impactful to someone running a self-hosted instance that uses badge Foo heavily).

For me the snapshot/`server-xxxx` image have the potential to have a more human intentionally selected set of diffs, but it's also true that it could have an issue deemed significant to some users

PyvesB:
On the one hand there hasn't been consensus on the maintainer side, on the other hand no other self-hosted users have requested this as far as I can tell. Will go ahead and close the PR.

PR #10512, opened 2024-09-07
Remove failing [Packagist] tests

PyvesB:
This tests started failing because example.com now returns a 500 for its not found pages:
Screenshot 2024-09-07 163315

I feel that these tests don't bring much value, we're exercising the base error handlers with no Packagist-specific logic. Rather than fixing them, I suggest we get rid of them.


chris48s (approved):
yeah I agree this is not buying us much :+1:

Issue #10299, opened 2024-06-26
Snapcraft Licence


MichaelHinrichs:
### 📋 Description

Snapcraft
License | AGPL 1.0 or later

### 🔗 Data

IDK, however it is that the existing Snapcraft version badge works.

### 🎤 Motivation

It's like the other Licence badges.

chris48s:
Example call:

```
curl "https://api.snapcraft.io/v2/snaps/info/thunderbird" --header "Snap-Device-Series: 16" | jq .snap.license
```

This property is exposed at snap level so does not need to support `track` / `risk` params

davide-pi:
Hi @chris48s  I've created a PR to add this badge #10520 🙇

PR #10520, opened 2024-09-11
[Snapcraft] license

davide-pi:
ISSUE: #10299

Add Snapcraft license.
Create the `SnapcraftBase` from which other Snapcraft services can inherit.

Update Snapcraft version service to inherit from the `SnapcraftBase`.

Add Snapcraft base URL in configurations.

### DIFF
--- a/config/default.yml
+++ b/config/default.yml
@@ -24,6 +24,8 @@ public:
       authorizedOrigins: 'https://api.opensuse.org'
     pypi:
       baseUri: 'https://pypi.org'
+    snapcraft:
+      baseUri: 'https://api.snapcraft.io/v2/snaps/info'
     weblate:
       authorizedOrigins: 'https://hosted.weblate.org'
     trace: false
--- a/config/default.yml
+++ b/config/default.yml
@@ -24,8 +24,6 @@ public:
       authorizedOrigins: 'https://api.opensuse.org'
     pypi:
       baseUri: 'https://pypi.org'
-    snapcraft:
-      baseUri: 'https://api.snapcraft.io/v2/snaps/info'
     weblate:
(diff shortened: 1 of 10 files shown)


config/default.yml
```diff
@@ -24,6 +24,8 @@ public:
       authorizedOrigins: 'https://api.opensuse.org'
     pypi:
       baseUri: 'https://pypi.org'
+    snapcraft:
+      baseUri: 'https://api.snapcraft.io/v2/snaps/info'
```
chris48s, line 28: Does this actually need to be a setting?
AFAIK self-hosted snapcraft isn't a thing. Snapd is tied to https://snapcraft.io/
If this is something that needs to be configurable, can you link me to some docs about this
If it isn't, lets bin all this and hard-code the base url in `SnapcraftBase`

chris48s (commented):
Thanks this looks pretty solid. Just one thing needing to follow up

config/default.yml
davide-pi: You're right it is better to maintain in the base service.
I was wondering if someone would make tests using a local endpoint but in this case is sufficient to change the address inside the service base class!

services/snapcraft/snapcraft-base.js
```diff
@@ -0,0 +1,23 @@
+import { BaseJsonService, pathParam } from '../index.js'
+
+export const snapcraftPackageParam = pathParam({
+  name: 'package',
+  example: 'Redis',
```
chris48s, line 5: One more quick thing I noticed. Package names are case sensitive Redis with a captial R is not found

```suggestion
  example: 'redis',
```

chris48s (approved):

PR #10522, opened 2024-09-12
[Jenkins] Use old.stats.jenkins.io for JSON data

froque:
Related with
 - https://github.com/jenkins-infra/helpdesk/issues/4265
 - https://github.com/jenkins-infra/plugin-site/pull/1890

### DIFF
--- a/services/jenkins/jenkins-plugin-installs.service.js
+++ b/services/jenkins/jenkins-plugin-installs.service.js
@@ -68,7 +68,7 @@ export default class JenkinsPluginInstalls extends BaseJsonService {
 
   async fetch({ plugin, version }) {
     return this._requestJson({
-      url: `https://stats.jenkins.io/plugin-installation-trend/${plugin}.stats.json`,
+      url: `https://old.stats.jenkins.io/plugin-installation-trend/${plugin}.stats.json`,
       schema: version ? schemaInstallationsPerVersion : schemaInstallations,
       httpErrors: {
         404: 'plugin not found',


chris48s:
Obviously I can see that
https://old.stats.jenkins.io/plugin-installation-trend/view-job-filters.stats.json serves some JSON
and
https://stats.jenkins.io/plugin-installation-trend/view-job-filters.stats.json
is a 404

but what's the long-term goal here?

Is the situation that the new https://stats.jenkins.io site still provides this data, but we need to account for some breaking changes? If so, lets just do that.

If https://stats.jenkins.io just removes access to this data and old.stats.jenkins.io will only be around for a short time, lets just sunset this following the [deprecation](https://github.com/badges/shields/blob/master/doc/deprecating-badges.md) process.

To me, it looks like the second one. While it does seem to be possible to initiate a download of JSON data via javascript in a browser, it seems to me like there is now no obvious or documented way to do this via a call to a predictable URL.

lemeurherveCB:
As there is far more consumers than we though (https://github.com/search?q=stats.jenkins.io&type=code 😱), I'll open an issue on the new frontend project and/or Jenkins infra helpdesk to determine how/if we can continue to serve those files without breaking existing URLs. (No ETA yet, to be discussed first)

I'll keep you informed.

lemeurherveCB:
FWIW, the current data is still available at https://github.com/jenkins-infra/infra-statistics/tree/gh-pages/plugin-installation-trend. (Note: we'll change the branch from `gh-pages` to `data` soon.)

lemeurherveCB:
Sorry for the inconvenience, I'm at fault here I should have searched for external consumers earlier.

chris48s:
OK. For the moment I am going to merge this as a stop-gap but I'll make an issue for the follow up

chris48s (approved):

PR #10537, opened 2024-09-14
Revert "Use old.stats.jenkins.io for JSON data"; test [jenkins]

### LINKED ISSUE #10534: Update Jenkins Plugin Installs for new stats.jenkins.io
:clipboard: **Description**

Refs https://github.com/badges/shields/pull/10522

As a stop-gap we are using old.stats.jenkins.io but this will be retired at some point.

It looks like a feature allowing us to grab machine-readable data from the new stats.jenkins.io will be introduced. Once that happens, we update our code to use that instead.

chris48s:
Closes #10534

### DIFF
--- a/services/jenkins/jenkins-plugin-installs.service.js
+++ b/services/jenkins/jenkins-plugin-installs.service.js
@@ -68,7 +68,7 @@ export default class JenkinsPluginInstalls extends BaseJsonService {
 
   async fetch({ plugin, version }) {
     return this._requestJson({
-      url: `https://old.stats.jenkins.io/plugin-installation-trend/${plugin}.stats.json`,
+      url: `https://stats.jenkins.io/plugin-installation-trend/${plugin}.stats.json`,
       schema: version ? schemaInstallationsPerVersion : schemaInstallations,
       httpErrors: {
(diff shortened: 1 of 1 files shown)


chris48s:
Hmm, so it seems the "404" response from stats.jenkins.io (e.g: https://stats.jenkins.io/plugin-installation-trend/not-a-plugin.stats.json ) is actually served with a `200 OK` status code.

chris48s:
For the moment, I've raised an issue on
https://github.com/jenkins-infra/helpdesk/issues/4291
If old.stats.jenkins.io goes away before that is resolved, there's a couple of possible options:

1. Accept the "not found" message for this service is "unparseable json response" and update the tests.
2. Switch this badge to make a raw call to `_request()` instead of `_requestJson()`. If we get back a 200 OK with a HTML body, assume that is a "not found" response. Otherwise, parse it as JSON.

PyvesB:
Slight preference for option 2.

chris48s:
This has now been fixed upstream

chris48s:
Given this got solved upstream, I'm just going to go ahead and merge this one

PR #10538, opened 2024-09-14
deprecate [wheelmap] service

### LINKED ISSUE #10404: Wheelmap badges broken

This seems to have been broken since we reinstated the daily test runs a couple of weeks ago

:camera: **Live badge**

https://img.shields.io/wheelmap/a/26699541

Nope

:lady_beetle: **Stack trace**

We're throwing a `NotFound` in response to receiving a 404.
The response we're getting is a HTML page, not a JSON response.

In the tests we're seeing
```
AssertionError: message mismatch: expected 'node not found' to equal 'yes'
AssertionError: message mismatch: expected 'node not found' to equal 'limited'
AssertionError: message mismatch: expected 'node not found' to equal 'no'
```

:bulb: **Possible solution**

We seem to be getting a 404 response from the wheelmap API for any node ID. This includes nodes that do exist, For example https://wheelmap.org/node/26699541 is valid but https://img.shields.io/wheelmap/a/26699541 fails.
I suspect wheelmap may have just got rid of their API completely.
Given nobody seems to have noticed or complained, maybe it is time to retire this service?

chris48s:
Closes #10404

Once we merge this we can remove
- `WHEELMAP_TOKEN` from prod/staging env vars on fly.io
- `SERVICETESTS_WHEELMAP_TOKEN` from GH Actions secrets
- `SERVICETESTS_WHEELMAP_TOKEN` from Dependabot secrets


app.json
```diff
@@ -1,56 +0,0 @@
-{
```
chris48s, line 1: I started off by just removing `WHEELMAP_TOKEN` but I think this file is not actually doing anything useful. We certainly haven't been keeping it up to date.
By the looks of it, we put this in place to support Heroku review apps, which we don't use anymore.

PyvesB (approved):
I've always felt this service was not quite a fit for the Shields.io use case, kind of relieved to see it go away.

Issue #8845, opened 2023-01-25
GitHub file size badge shows a different size than GitHub


loreanvictor:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

As the title says, GitHub file size badge shows a different file size compared to GitHub itself:

- [The file on GitHub](https://github.com/loreanvictor/nokss/blob/gh-pages/nokss.css.br) (shows 8.93kB as of now, might change)
- [The badge](https://img.shields.io/github/size/loreanvictor/nokss/nokss.css.br?branch=gh-pages) (shows 9.14kB as of now, might change)

The difference is not substantial of course, meaning at least for my use case this is not a serious issue, but an annoyance nonetheless.

### 🔗 Link to the badge

https://img.shields.io/github/size/loreanvictor/nokss/nokss.css.br?branch=gh-pages

chris48s:
This is a kilobytes (1000) vs kibibytes (1024) issue.

9140 bytes/1000 = 9.14 kilobytes
9140 bytes/2014 = 8.93 kibibytes

Pretty-bytes (the library we use for formatting these numbers) uses kilobytes
https://github.com/sindresorhus/pretty-bytes#pretty-bytes

```
import prettyBytes from 'pretty-bytes';
console.log(prettyBytes(9140)); // prints '9.14 kB'
```

btw, where on https://github.com/loreanvictor/nokss/blob/gh-pages/nokss.css.br does it show a file size?

loreanvictor:
Yeah just noticed that GitHub uses `KB` not `kB`:

<img width="678" alt="Screenshot 2023-01-26 at 09 34 32" src="https://user-images.githubusercontent.com/13572283/214791380-88dc891e-0a35-447c-ad04-dfd32c71c58b.png">

Thanks for the response!

chris48s:
Thought I was losing my mind there. GitHub shows really different views depending on whether you are logged in or logged out (and only shows the size if you're logged out)

**logged in:**
![Screenshot at 2023-01-27 20-41-13](https://user-images.githubusercontent.com/6025893/215193214-9520a181-94f9-4232-8f8e-92d02e4c824b.png)


**logged out:**
![Screenshot at 2023-01-27 20-41-55](https://user-images.githubusercontent.com/6025893/215193239-a1cd9222-aded-40cb-ba8a-2b6d0c5773be.png)

loreanvictor:
[...]

PR #10547, opened 2024-09-21
WIP Add ability to format bytes as metric or IEC; affects [bundlejs bundlephobia CratesSize DockerSize GithubRepoSize GithubCodeSize GithubSize NpmUnpackedSize SpigetDownloadSize steam VisualStudioAppCenterReleasesSize whatpulse]

### LINKED ISSUE #9703: Incorrect image size for Docker Hub

### LINKED ISSUE #10437: Add ability to format bytes as metric (e.g: kilobyte) or IEC (e.g: kibibyte) format
:clipboard: **Description**

Shields has a number of size badges. In general, there's a couple of patterns for how these work:

1. In some cases, the upstream data source reports an already formatted number. So for example with the spiget badge, the API returns an already formatted size and units.

```
curl "https://api.spiget.org/v2/resources/771" -s | jq '.file.size, .file.sizeUnit'
```

That's what goes on the badge. In these situations the number on the badge is always going to match what is shown on the upstream service.

2. The other pattern is that the upstream API reports a raw number of bytes and we format it. So for example with the bundlephobia badge the API returns a raw number of bytes.

```
curl "https://bundlephobia.com/api/size?package=react" -s | jq .size
```

In the second case, we always format the raw number of bytes for display using [pretty-bytes](https://www.npmjs.com/package/pretty-bytes), which exclusively uses metric (e.g: kilobyte) units.

As a general principle, we want shields to be "consistent". This means more than one thing. It can involve maintaining consistency both "horizontally" across our suite of badges (e.g: the NPM license badge should work the same way as the PyPI license badge) and "vertically" with the upstream data sources (i.e: if the registry says the latest version of your package is v2.4.1, so should we).

Always using pretty-bytes achieves horizontal consistency, but can lead to vertical inconsistency. If the upstram data provider uses IEC (e.g: kibibyte) units, the size we report on the badge may be slightly different. For example, the bundlephobia badge "disagrees" with the bundlephobia website.

-  https://img.shields.io/bundlephobia/min/react reports 6.6 kB
- https://bundlephobia.com/package/react reports 6.4kB

In my view, vertical consistency is more important in this case.

For badges where we format a raw number of bytes, I think we should switch from pretty-bytes to another formatting library like [byte-size](https://www.npmjs.com/package/byte-size) which allows formatting using both metric and IEC units.

We should review each of the size badges where we format a raw number of bytes instead of receiving an already formatted number. There are not a huge number of these. Where the upstream data provider uses IEC units, we should switch to using that.

We could also consider exposing a param allowing the user to specify the formatting, but I think matching the upstream by default is the most important thing to do here.

This is something that has been discussed several times before (albeit in less detail):

- https://github.com/badges/shields/issues/8845
- https://github.com/badges/shields/issues/9703#issuecomment-1789630736
- https://github.com/badges/shields/pull/10421

chris48s:
Following on from the conversation in https://github.com/badges/shields/issues/10437

Closes #10437
Closes #9703
Refs #8845
Refs #10421

While this PR does introduce some variance, I have also taken the opportunity to add a `renderSizeBadge()` helper function which will help us achieve some more horizontal consistency around the things where we do want to be opinionated.


services/size.js
```diff
@@ -0,0 +1,25 @@
...
+ * Creates a badge object that displays information about a size in bytes number.
+ * It should usually be used to output a size badge.
+ *
+ * @param {number} bytes - Raw number of bytes to be formatted
+ * @param {'metric'|'iec'} units - Either 'metric' (multiples of 1000) or 'iec' (multiples of 1024).
+ *   This should align with how the upstream displays sizes.
+ * @param {string} [label='size'] - Custom label
+ * @returns {object} A badge object that has three properties: label, message, and color
+ */
+function renderSizeBadge(bytes, units, label = 'size') {
```
chris48s, line 17: This helper gives us

- a standard colour
- a default label text (although we can override it) and
- always passes the number of bytes though `byteSize()` with the same settings

byte-size takes a slightly different approach to rounding than pretty-bytes does.
In pretty-bytes, rounding level is expressed as significant figures. The default in pretty-bytes was 3 significant figures.
In byte-size, rounding level is expressed as decimal places. The default in byte-size is 1 decimal place.
For some examples, this will give us a slightly different level of rounding
My instinct is to roll with the the default (1 decimal place), but 2 decimal places would also be a reasonable choice here. Either way, lets make a decision now, set it here in the `renderSizeBadge()` helper, set it here, and use that everywhere.

services/bundlejs/bundlejs-package.service.js
```diff
@@ -110,7 +105,7 @@ export default class BundlejsPackage extends BaseJsonService {
 
   async handle({ scope, packageName }, { exports }) {
     const json = await this.fetch({ scope, packageName, exports })
-    const size = json.size.compressedSize
-    return this.constructor.render({ size })
+    const size = json.size.rawCompressedSize
+    return renderSizeBadge(size, 'metric', 'minified size (gzip)')
```
chris48s, line 109: Previously we were presenting `compressedSize` as reported by the upstream
Switching this to use `rawCompressedSize` allows us to apply our own rounding convention while matching the upstream units

services/github/github-repo-size.service.js
```diff
@@ -51,6 +43,8 @@ export default class GithubRepoSize extends GithubAuthV3Service {
 
   async handle({ user, repo }) {
     const { size } = await this.fetch({ user, repo })
-    return this.constructor.render({ size })
+    // note the GH API returns size in KiB
+    // so we multiply by 1024 to get a size in bytes and then format that in IEC bytes
+    return renderSizeBadge(size * 1024, 'iec', 'repo size')
```
chris48s, line 48: This one is a bit wierd. The GitHub API exposes a number of Kibibytes (although the API docs say "kilobytes") rather than a number of bytes.

There is one place in the GitHub UI where this number is exposed/formatted and that is on https://github.com/settings/repositories

To work an example, if I call https://api.github.com/repos/chris48s/geometry-to-spatialite I get size=7159. That's presented in the frontend as
Screenshot at 2024-09-21 18-54-20
7159/1024 = 6.99

services/github/github-code-size.service.js
```diff
@@ -31,15 +31,8 @@ export default class GithubCodeSize extends BaseGithubLanguage {
...
-    return {
-      message: prettyBytes(size),
-      color: 'blue',
-    }
-  }
-
   async handle({ user, repo }) {
     const data = await this.fetch({ user, repo })
-    return this.constructor.render({ size: this.getTotalSize(data) })
+    return renderSizeBadge(this.getTotalSize(data), 'iec', 'code size')
```
chris48s, line 36: This number is never presented anywhere in the GitHub UI. Since the other 2 GitHub badges use IEC bytes, lets go with that here too.

services/visual-studio-app-center/visual-studio-app-center-releases-size.service.js
```diff
@@ -47,14 +47,8 @@ export default class VisualStudioAppCenterReleasesSize extends BaseVisualStudioA
...
-  static render({ size }) {
-    return {
-      message: prettyBytes(size),
-    }
-  }
-
   async handle({ owner, app, token }) {
     const { size } = await this.fetch({ owner, app, token, schema })
-    return this.constructor.render({ size })
+    return renderSizeBadge(size, 'metric')
```
chris48s, line 52: I have no idea how this site presents sizes. Lets stick with metric for the moment on the basis we have been doing that for some time and nobody has complained yet. We can change it if we need to.

services/npm/npm-unpacked-size.tester.js
```diff
@@ -1,27 +1,27 @@
...
 
 t.create('Unpacked size for scoped package')
   .get('/@testing-library/react.json')
-  .expectBadge({ label: 'unpacked size', message: isFileSize })
+  .expectBadge({ label: 'unpacked size', message: isMetricFileSize })
 
 t.create('Unpacked size for scoped package with version')
   .get('/@testing-library/react/14.2.1.json')
-  .expectBadge({ label: 'unpacked size', message: '5.41 MB' })
+  .expectBadge({ label: 'unpacked size', message: '5.4 MB' })
```
chris48s, line 24: This file is a good example of the difference between rounding to significant figures vs rounding to decimal places. Switching to 1 decimal place sometimes gives us more precision and sometimes less precision than 3 significant figures.

chris48s:
This PR is still draft as I have not yet added the ability to switch between units via a URL param. I will do that next before marking as ready for review.

services/npm/npm-unpacked-size.service.js
```diff
@@ -1,21 +1,31 @@
...
 
 const schema = Joi.object({
   dist: Joi.object({
     unpackedSize: optionalNonNegativeInteger,
   }).required(),
 }).required()
 
+const queryParamSchema = baseQueryParamSchema.keys({
+  units: unitsQueryParam.default(defaultUnits),
+})
```
chris48s, line 20: Here's a little Joi TIL

```js
import Joi from 'joi'

const schema1 = Joi.object({
  one: Joi.string(),
}).required()

console.log(schema1.describe())

const schema2 = schema1.keys({
  two: Joi.string(),
})

console.log(schema2.describe())
```

chris48s:
OK. I've marked this PR ready for review. In https://github.com/badges/shields/pull/10547/commits/c6028f16304cefcc0e14b1927c7edcf73c14b450 I've added the ability to switch between IEC and metric units. The code isn't _bad_ but it does add a bunch of complexity. Now that I've written it and we have a diff to discuss, I think its reasonable to not only ask "is this good code?" but also to revisit "is this a good idea?"

Even though I've spent a bunch of time on https://github.com/badges/shields/pull/10547/commits/c6028f16304cefcc0e14b1927c7edcf73c14b450 I'd also be quite happy to roll it back and just merge https://github.com/badges/shields/pull/10547/commits/3395601d6fedfea16b0b745e9fca94fab6e49b8a and https://github.com/badges/shields/pull/10547/commits/cbac43bee8c5efb333104282463697efebfe2393 if we don't think it is worth it.

PyvesB:
> OK. I've marked this PR ready for review. In https://github.com/badges/shields/commit/c6028f16304cefcc0e14b1927c7edcf73c14b450 I've added the ability to switch between IEC and metric units. The code isn't bad but it does add a bunch of complexity. Now that I've written it and we have a diff to discuss, I think its reasonable to not only ask "is this good code?" but also to revisit "is this a good idea?"

I'm a little short on time to do a full review, but the diff with the parameter switch feels reasonable in terms of added code complexity, i.e. at a glance I'm able to follow what's going on.

Regarding whether it's a good idea, honestly, I don't know:
* in my mind, I've always associated kilobytes to 1024 bytes. If I see kB or kiB, they're the same thing for me, I'll admit that I never encountered the word "kibibyte" before https://github.com/badges/shields/issues/10437, or at least it didn't attract my attention to the point that I'd remember. Possible consequence of having a formal education where we often reasoned using the binary system/powers of two, and having never worked on a professional problem when this subtlety mattered enough.
* consequently, I'll echo https://github.com/badges/shields/issues/10437#issuecomment-2302967930. Whilst I agree with the general conclusion of vertical consistency, I fear that allowing the same badge (or at least that look the same, assuming I'm not the only idiot to not be able to tell at a glance the difference between kB and kiB) have different meanings depending on a `metric` parameter, may do more harm than good. I kind of feel like not having the parameter brings the best of both worlds, vertical consistency (same as the number indicated by e.g. upstream bundlephobia service) and horizontal consistency within the service (all e.g. bundlephobia badges show numbers representing the same thing).

Those are my two cents, but ultimately, I don't feel that strongly.

chris48s:
**TODO**: need to update this after https://github.com/badges/shields/pull/10619 was merged

chris48s:
I dropped the ball on this one for a while, but coming back to it..

I think I agree with the assessment that for a lot of users kB and KiB is a really subtle distinction and the `metric` param is probably _not_ useful. I've rolled that change back in https://github.com/badges/shields/pull/10547/commits/12c682361c3f60288699aa8683062e040ba6a2d3 .

I've merged in master and resolved the conflicts.

I've updated the PR to account for the Chrome Web Store Size badge, which was added since I opened this PR.

There are a few failing service tests unrelated to the changes in this PR:

1. There is a PR open for the crates.io failure at https://github.com/badges/shields/pull/10688
2. The bundlephobia failures are probably transient. We seem to be getting 502s from them at the moment

Screenshot at 2024-11-24 16-00-11

services/size.js
PyvesB: 1 decimal place seems like a reasonable number to start with.

PyvesB (dismissed):
Great work! I'm quite happy with what we've ended up with in this PR, and the linked discussion! Thanks for taking the lead on this 👏🏻

Issue #10575, opened 2024-10-03
Inconsistent badge padding with badge-maker


randomairborne:
### Are you experiencing an issue with...

badge-maker NPM package

### 🐞 Description

When creating badges with custom logos, in some cases (such as on [https://www.randomairborne.dev/contact/](https://www.randomairborne.dev/contact/)) there is very little padding after the text on the YouTube and Discord badges.
<img width="755" alt="Screenshot 2024-10-03 at 10 49 34" src="https://github.com/user-attachments/assets/e56b7517-7501-48d5-bddb-d8ef084510b5">

randomairborne:
I was hitting that issue, and have proposed a fix as a PR. Closing this one in favor of that PR and the previous issue.

Issue #10336, opened 2024-07-05
Allow user to request globally unique ids 


chris48s:
### Discussed in https://github.com/badges/shields/discussions/10321

The IDs on the `linearGradient` and `clipPath` SVG elements are the same for every badge. This means the IDs are unique within any given SVG document, but if you embed multiple SVGs inline in a single HTML document then the IDs aren't unique within the HTML page.

This leads to this behaviour:

<img width="425" alt="image" src="https://github.com/badges/shields/assets/6390600/223800ed-fdb9-42bf-81df-1d616a3a56e3">

This is a fairly uncommon use-case and for shields.io itself it is useful that the IDs are single character. However, it would be useful for the `makeBadge()` function to expose a non-default option which allows the user to specify they want globally unique IDs.

We could do this without adding any additional dependency by using `crypto.randomUUID()` from the node standard library to generate the IDs:

```js
const { randomUUID } = require('crypto');
console.log(randomUUID());
```

chris48s:
https://github.com/badges/shields/pull/10576

PR #10576, opened 2024-10-04
Add option for IDs to have a prefix

randomairborne:
This PR attempts to fix the issue i ran into #10575, discussion #10321, and original issue #10336. It adds an optional prefix to ID values, so that multiple invocations can use different ID values. This prevents needing long random identifiers and increasing size in the general case, and gives users that need it more flexibility.


randomairborne:
maybe this needs to be documented more? I'm not quite sure where that is, though. It's not really a complex feature 😅

chris48s:
Thanks. I don't mind this as an approach. It does allow you to generate badges with unique ids. This does push the work of doing that onto the user, but one thing this does give you is predictable ids. i.e: if we use UUIDs then the same `rect` will have a different id every time you generate the badge, whereas with this you can give it the same id every time. That might be useful, I guess.

I think I would make this a suffix rather than a prefix though. I would assume one of the ways people might use this feature to generate unique ids would be to use a loop/counter variable as the "unique" part of the ID, but with a prefix this would generate ids that start with a number, which is invalid. If we make it a suffix then we control the first character of the ID and we can ensure the IDs are valid rather than pushing that on to the user and giving them a foot-gun here.

In terms of docs, we will need to cover this in the badge-maker readme (mainly what problem does this feature solve), but lets nail the feature down first.

I would also want this feature to be covered by tests before we review it.

randomairborne:
Yeah, my thought was that "hmm, a uuid is really long". I'll add some tests (Where's the best place to do that?) and i'll document it in the readme.

chris48s:
Thanks.

In terms of the tests, the most important place to cover this is the snapshot tests in `badge-maker/lib/make-badge.spec.js`

These are quite easy to add. If you add some new test cases to this file, the first time you run `npm run test:package` a snapshot of the SVG output will automatically be added to `__snapshots__/make-badge.spec.js`. Just check the snapshots over and then once the SVG output is stored, that gives us regression tests for future. On future test runs, we'll compare the output of the function to the stored shapshot for that named test case.

It would also be useful to add a case for the param validation to the "should throw a ValidationError with invalid inputs" tests in `badge-maker/lib/index.spec.js`

randomairborne:
Does this count as a breaking change for the purposes of version bumps? is there a release cadence for this? I'd like to end up using it, but I'm not sure when I would need to set a reminder.

chris48s:
I checked this out and had a look. Rather than going back and forth on this I've pushed a few more commits making a few final tweaks.

In terms of the release, this is not a breaking change. We are just adding a feature. We're not removing anything or making any behaviour changes to existing functionality. The code `makeBadge({ label: 'build', message: 'passed', color: 'brightgreen' })` will output the exact same SVG before and after this change so the next release can just be 4.1.0. Tbh we don't change the SVG rendering logic very much so there is not a defined release schedule for badge-maker. I'll just merge this and push a 4.1.0 release to NPM with this change in it today.

chris48s (approved):

randomairborne:
Thank you very much.

PR #10596, opened 2024-10-10
Run application as non-root user in the container - #10568

### LINKED ISSUE #10568: Run aplication as non-root user inside docker container
:clipboard: **Description**

Our docker image runs as root.

Best practice would be to run as a less privileged user.

MohanKumarAmbati:
Hi @chris48s,

Updated the dockerfile, now the application runs as non-root user or (appuser with appgroup) in the container. Can you please review the changes.

Also, gentle reminder to include "hacktoberfest-accepted" label to this PR, such that it counts under my hacktoberfest contributions. Thank you 

closes #10568

Looking forward to contribute more.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,10 @@ LABEL fly.version=$version
 # Run the server using production configs.
 ENV NODE_ENV=production
 
+# Create and Run the app as non-root user
+RUN addgroup appgroup && adduser -S -G appgroup appuser
+USER appuser
+
 WORKDIR /usr/src/app
 COPY --from=builder --chown=0:0 /usr/src/app /usr/src/app
 


chris48s:
I gave this a quick test. There's a fairly obvious problem here, which is that we're running the app as this new `appuser` here, but then 2 lines down we're changing everything to be owned by root so anything that relies on writing to the filesystem fails with `Permission denied`. We need to change everything to be owned by `appuser:appgroup`.

So anyway, I did that. There were also a couple of other more minor changes I wanted to make which were:

- using long args for clarity and
- using known ids for the `gid` and `uid`

so I was then working with

```diff
diff --git a/Dockerfile b/Dockerfile
index 6725596502..87c909a828 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,11 +29,11 @@ LABEL fly.version=$version
 ENV NODE_ENV=production
 
 # Create and Run the app as non-root user
-RUN addgroup appgroup && adduser -S -G appgroup appuser
+RUN addgroup appgroup --system --gid 1001 && adduser --system --uid 1001 --ingroup appgroup appuser
 USER appuser
 
 WORKDIR /usr/src/app
-COPY --from=builder --chown=0:0 /usr/src/app /usr/src/app
+COPY --from=builder --chown=appuser:appgroup /usr/src/app /usr/src/app
 
 CMD ["node", "server"]
 
```

That worked fine locally.

The next question I wanted to answer was: Creating a system user sets the user's shell to `/sbin/nologin`. Does that break our ability to shell into the container with `flyctl ssh console`? So I tried deploying to fly, and I'm hitting an issue when we deploy the container to fly. The container just fails to boot, but the only error I've got is `exit_code=1,oom_killed=false,requested_stop=false`. Trail runs cold there.

Thanks for attempting to help with this, but we'll need to work that out and I'm not sure it is something you can easily help with as the remaining issue doesn't reproduce locally. I need to spend some more time on it to work out why we're hitting this issue when we deploy but I think it will have to be something I come back to another day.

Issue #10586, opened 2024-10-05
opening and ending tag mismatch "link" != "head"


ReenigneArcher:
### Are you experiencing an issue with...

shields.io

### 🐞 Description

Three custom xml query badges, are having issues. All from the same source url (probably ultimately an issue with their page... but maybe shields can work around this?)

In the head of the page (https://repology.org/project/sunshine/versions), their link tags do not end with `/>` but instead just end with `>`

### 🔗 Link to the badge

https://img.shields.io/badge/dynamic/xml.svg?color=orange&label=chocolatey&style=for-the-badge&prefix=v&query=%2F%2Ftr%5B%40id%3D%27chocolatey%27%5D%2Ftd%5B3%5D%2Fspan%2Fa&url=https%3A%2F%2Frepology.org%2Fproject%2Fsunshine%2Fversions&logo=chocolatey

https://img.shields.io/badge/dynamic/xml.svg?color=orange&label=nixpkgs&style=for-the-badge&prefix=v&query=%2F%2Ftr%5B%40id%3D%27nix_unstable%27%5D%2Ftd%5B3%5D%2Fspan%2Fa&url=https%3A%2F%2Frepology.org%2Fproject%2Fsunshine%2Fversions&logo=nixos

https://img.shields.io/badge/dynamic/xml.svg?color=orange&label=Solus&style=for-the-badge&prefix=v&query=%2F%2Ftr%5B%40id%3D%27solus%27%5D%2Ftd%5B3%5D%2Fspan%2Fa&url=https%3A%2F%2Frepology.org%2Fproject%2Fsunshine%2Fversions&logo=solus

### 💡 Possible Solution

Ignore malformed xml tags in the head section.

chris48s:
Couple of follow-ups:

1. Are these badges that worked in the past but started failing recently-ish?
2. Fundamentally, the document you're trying to parse here is HTML not XML, right?

I reckon we might have our first example of something broken by https://github.com/badges/shields/pull/10481#discussion_r1740961975

Just having a quick look at the response from https://repology.org/project/sunshine/versions I think it would be reasonable to parse in HTML mode if the document has

```
content-type: text/html
```

in the response headers.

ReenigneArcher:
1. Yes, they used to work
2. Correct, it's html

ReenigneArcher:
Confirming these are working again, thank you!

PR #10607, opened 2024-10-11
[DynamicXml] parse doc as html if served with text/html content type

chris48s:
Closes https://github.com/badges/shields/issues/10586
Refs https://github.com/badges/shields/pull/10481#discussion_r1740961975

### DIFF
--- a/services/dynamic/dynamic-xml.service.js
+++ b/services/dynamic/dynamic-xml.service.js
@@ -70,7 +70,7 @@ export default class DynamicXml extends BaseService {
 
   static defaultBadgeData = { label: 'custom badge' }
 
-  transform({ pathExpression, buffer }) {
+  transform({ pathExpression, buffer, contentType = 'text/xml' }) {
     // e.g. //book[2]/@id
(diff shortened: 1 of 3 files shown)


services/dynamic/dynamic-xml.service.js
```diff
@@ -70,22 +70,28 @@ export default class DynamicXml extends BaseService {
...
       throw new InvalidResponse({ prettyMessage: e.message })
     }
 
     let values
     try {
-      values = xpath.select(pathExpression, parsed)
+      if (contentType === 'text/html') {
+        values = xpath
+          .parse(pathExpression)
+          .select({ node: parsed, isHtml: true })
```
chris48s, line 91: This is required so that we don't have to explicitly use a namespace on every selector for HTML documents

See

- https://github.com/xmldom/xmldom/issues/377
- https://github.com/goto100/xpath/issues/27

for more detail

PyvesB (approved):
Looks good to me 👍🏻

PR #10608, opened 2024-10-12
Refactor version rendering in [wordpress nexus] and [f-droid] services

jNullj:
We mostly use the short more elegant `renderVersionBadge` but some services still use older method of directly calling addv and versionColor.

These changes refactor the WordpressVersion, Nexus and FDroid services to fit usage of `renderVersionBadge`

### DIFF
--- a/services/f-droid/f-droid.service.js
+++ b/services/f-droid/f-droid.service.js
@@ -3,8 +3,7 @@ import {
   optionalNonNegativeInteger,
   nonNegativeInteger,
 } from '../validators.js'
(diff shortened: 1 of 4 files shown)


jNullj:
I think service test failed for nexus due to upstrem changes
https://github.com/badges/shields/actions/runs/11306263530?pr=10608
I get the same issue at the master branch

PyvesB (approved):
Thanks @jNullj, this looks great! We're making slow progress on #2026, maybe one day we'll be able to close it 😄

PR #10618, opened 2024-10-16
Deprecate [HackageDeps] service

jNullj:
Badge will always return up to date falsely as the upstream site is closed and looking for someone to take over.

The first test if a package exist will always return 200 OK even when non existing as we get the landing page.

Then we will never get "outdated dependencies" as the landing page doesn't include that text.

I looked into the https://hackage.haskell.org/api and couldn't find a quick way to replace the current upstream.

Therefor i propose we remove this badge, this is misleading at best as it's now.

image

### DIFF
--- a/services/hackage/hackage-deps.service.js
+++ b/services/hackage/hackage-deps.service.js
@@ -1,48 +1,11 @@
-import { BaseService, pathParams } from '../index.js'
+import { deprecatedService } from '../index.js'
 
-export default class HackageDeps extends BaseService {
-  static category = 'dependencies'
-
-  static route = {
+export const HackageDeps = deprecatedService({
+  category: 'dependencies',
+  route: {
     base: 'hackage-deps/v',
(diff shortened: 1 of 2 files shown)


chris48s:
Thanks. When we need to remove a service, we initially replace it with a message that says "no longer available". That way users get some kind of explanation of what happened instead of just a  in their README. Then we delete it after a year (or whenever we do a clear out).

There's docs on this in https://github.com/badges/shields/blob/master/doc/deprecating-badges.md and you could look at something like https://github.com/badges/shields/pull/10371 as an example.

jNullj:
My bad, didn't notice this section of the docs. Thanks for the example it saved me some time.

PyvesB (approved):
Thanks for taking care of this! There are less than 1000 of these badges rendered on a typical day, I don't think this service will be hugely missed :)

Issue #10614, opened 2024-10-16
Badge Request: crates.io total downloads of an author


Nugine:
### 📋 Description

This badge shows the total downloads of all the crates that an author publishs.

![Static Badge](https://img.shields.io/badge/total_downloads-123456-blue)

### 🔗 Data

`https://crates.io/api/v1/users/{user_id}/stats`
It is a public API.

Example: https://crates.io/api/v1/users/3618/stats
```json
{"total_downloads":6096699151}
```

### 🎤 Motivation

Showing that this crate author is trustworthy because of high total downloads.

calebcartwright:
Sounds reasonable, especially since there's already an API for it. I wouldn't personally conflate download numbers as inherently implying trust, but certainly seems like a badge we can incorporate if someone is willing to work on it.

PR #10619, opened 2024-10-17
Add [CratesUserDownloads] service and tester

jNullj:
This pull request adds the `CratesUserDownloads` service and tester files.

The `CratesUserDownloads` service fetches data from the crates.io API to show the total downloads for a user. 
It also includes tests for the service.

Added `BaseCratesUserService` for api route for users.

Solves #10614.

### DIFF
--- a/services/crates/crates-user-downloads.service.js
+++ b/services/crates/crates-user-downloads.service.js
@@ -1,0 +1,32 @@
+import { renderDownloadsBadge } from '../downloads.js'
+import { pathParams } from '../index.js'
+import { BaseCratesUserService, description } from './crates-base.js'
+
+export default class CratesUserDownloads extends BaseCratesUserService {
+  static category = 'downloads'
+  static route = {
+    base: 'crates',
+    pattern: 'udt/:userId',
+  }
+
+  static openApi = {
+    '/crates/udt/{userId}': {
+      get: {
+        summary: 'Crates.io User Total Downloads',
+        description,
+        parameters: pathParams({
+          name: 'userId',
+          example: '3027',
+          description:
+            'The user ID can be found using https://crates.io/api/v1/users/{username}',
+        }),
+      },
+    },
+  }
+
+  async handle({ userId }) {
+    const json = await this.fetch({ userId })
+    const { total_downloads: downloads } = json
+    return renderDownloadsBadge({ downloads, labelOverride: 'downloads' })
+  }
+}
--- a/services/crates/crates-user-downloads.service.js
+++ b/services/crates/crates-user-downloads.service.js
@@ -18,7 +18,7 @@ export default class CratesUserDownloads extends BaseCratesUserService {
           name: 'userId',
           example: '3027',
           description:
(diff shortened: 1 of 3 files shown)


services/crates/crates-user-downloads.service.js
```diff
@@ -0,0 +1,32 @@
...
+  static openApi = {
+    '/crates/udt/{userId}': {
+      get: {
+        summary: 'Crates.io User Total Downloads',
+        description,
+        parameters: pathParams({
+          name: 'userId',
+          example: '3027',
+          description:
+            'The user ID can be found using https://crates.io/api/v1/users/{username}',
```
chris48s, line 21: This renders as a link in the docs, but following it doesn't take you anywhere useful. Can we put it in `` `backticks` `` so it renders a `<code>` block instead of a link

services/crates/crates-user-downloads.tester.js
```diff
@@ -0,0 +1,13 @@
+import { isMetric } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('total user downloads')
+  .get('/udt/3027.json')
+  .expectBadge({ label: 'downloads', message: isMetric })
+
+// we can not test for id not linked to user as it returns 0 downloads
```
chris48s, line 9: I reckon we should put in a test case for a non-existant user id anyway, `.expectBadge({ label: 'downloads', message: '0' })` in the assertion and explain in a comment that this API returns `{"total_downloads": 0}` with a `200 OK` instead of a 404 response for a not found user id.

chris48s (commented):
Thanks. Couple of minor point, but good stuff

services/crates/crates-user-downloads.service.js
jNullj: Yea, it looks better with `code block`
Made changes at 2982c8fa102456ec5ce627850d55c0dc058d8e2c

services/crates/crates-user-downloads.tester.js
jNullj: Added the test with 230c79aa9d24d903a879586c6b43afed216af0a8.
I realized they use int 32 for userid so i picked 2^32 as its the last id which has the least chance of getting taken.

services/crates/crates-user-downloads.tester.js
```diff
@@ -0,0 +1,16 @@
+import { isMetric } from '../test-validators.js'
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('total user downloads')
+  .get('/udt/3027.json')
+  .expectBadge({ label: 'downloads', message: isMetric })
+
+// non-existent user returns 0 downloads with 200 OK rather then 404.
```
shxntanu, line 9:
```suggestion
// non-existent user returns 0 downloads with 200 OK status code rather than 404.
```

jNullj: Thanks, i fixed the typo with 7e64911

PyvesB (approved):
No further comments, let's get this merged!

Issue #10023, opened 2024-03-10
npm latest package update time


nicfv:
### 📋 Description

It would be great to show the latest updated date for an npm package, sort of like the static [relative date](https://shields.io/badges/relative-date) badge:

![Relative date](https://img.shields.io/date/1710110919)

![image](https://github.com/badges/shields/assets/24358775/707064dc-2c30-4e60-94c8-bdb1f4465f75)

A required badge query parameter would be the package name. Possibly an optional query parameter could be the package version, but should default to the latest version.

### 🔗 Data

`npm v {package} time.modified`

E.g.:

```
$ npm v smath time.modified
2024-03-10T22:40:54.837Z
```

### 🎤 Motivation

It would be great to show how recently my package has been updated, to show that it is still being maintained.

chris48s:
This does appear to be exposed via the NPM API
If we call something like there is a key called `time` which is a map of version number to timestamp. We could look up the latest version in that.

nicfv:
> If we call something like https://registry.npmjs.org/express/ there is a key called time which is a map of version number to timestamp.

I didn't know about this api! Thanks for sharing. Looking forward to (hopefully) a new badge soon!

saloni0419:
Can you please assign this issue to me

nicfv:
@saloni0419 It doesn't look like I have permission to assign issues, but go for it, if you're interested working on it! Looks like PR #10083 hasn't been touched in over 3 weeks, so not sure if anything will happen there.

aniruddhparwal:
@chris48s It seems like no one has been working on this for a long time. If fine I would like to pick this up.

chris48s:
Sure. It looks like there is a WIP PR open that was never finished at https://github.com/badges/shields/pull/10083 that might make a good starting point

PR #10585, opened 2024-10-04
[Maven] Added badge for Maven-Cenral last-update (#10301)

MohanKumarAmbati:
* Added badge for Maven-Central last-update

Hi @chris48s, can you please add **hacktoberfest-accepted** label to this PR so that it is considered under it.

closes - #10301

Thanks in advance.

PR #10641, opened 2024-10-26
[npm] - Last update badge added

MohanKumarAmbati:
Hi Team,

Extending NpmBase points registry URL to https://registry.npmjs.org/express/latest -  **latest**, which do not include last updated time. Hence, extended BaseJsonService to fetch the json.

closes - #10023


services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,96 @@
...
+    created: Joi.string().required(),
+    modified: Joi.string().required(),
+  })
+    .pattern(Joi.string(), Joi.any())
+    .required(),
+}).required()
+
+const defaultUrl = 'https://registry.npmjs.org'
+
+export class NpmLastUpdate extends BaseJsonService {
```
jNullj, line 29: can we use `NpmBase` here? it has for example `fetchPackageData` that might be useful.

services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,96 @@
...
+import { age as ageColor } from '../color-formatters.js'
+
+const queryParamSchema = Joi.object({
+  registry_uri: optionalUrl,
+  version: Joi.string(),
+}).required()
+
+const updateResponseSchema = Joi.object({
+  time: Joi.object({
+    created: Joi.string().required(),
```
jNullj, line 20: Can we remove the test for `created`? i dont see we use it here.

services/npm/npm-last-update.tester.js
```diff
@@ -0,0 +1,37 @@
...
+
+t.create('last updated date based on the version (valid scenario)')
+  .get('/express.json?version=5.0.0')
+  .expectBadge({
+    label: 'last updated',
+    message: isFormattedDate,
+  })
+
+t.create('last updated date based on the version (invalid scenario)')
+  .get('/express.json?version=not-a-verion')
```
jNullj, line 33:
```suggestion
  .get('/express.json?version=not-a-version')
```

jNullj: fix typo

services/npm/npm-last-update.service.js
MohanKumarAmbati: Yeah, we can remove `created`

jNullj (changes requested):
also this.

services/npm/npm-last-update.service.js
MohanKumarAmbati: - ` NpmBase`  appends to **/latest** to the package name. When `tag` variable is undefined.
- With `tag` not being undefined it points to correct url but not sure on the other effects.
- Please let me know if there is a better way, because of the above case I went with extending BaseJsonService


- also we’re using `version` query param which isn’t part of `NpmBase`

jNullj: Thank you for the clarification.

I see that `fetchPackageData` but we can still benefit from using `NpmBase`
Notice how we use `encodeScopedPackage` to take care of scoped packages. (which we might want here as well as a parameter for the badge)
Also using `_requestJson` from `NpmBase` could help us avoid https://github.com/npm/npmjs.org/issues/163

I think we will benefit from writing a fetch function for this use case that's separate from `fetchPackageData`, worst case it's clearer code, best case we will re-use it for other badges later.

MohanKumarAmbati: Got it Jonathan, let me give it a try again.

services/npm/npm-last-update.tester.js
MohanKumarAmbati: fixed as part of https://github.com/badges/shields/pull/10641/commits/d39c4b54697d0582f96e7f69843d18a782d4a2ad

services/npm/npm-last-update.service.js
MohanKumarAmbati: removed as part of https://github.com/badges/shields/pull/10641/commits/d39c4b54697d0582f96e7f69843d18a782d4a2ad

services/npm/npm-last-update.service.js
MohanKumarAmbati: Jonathan, extended `NpmBase` by adding a `fetch` function as suggested, please take a look and let me know if anyway we can make it better.

jNullj (changes requested):
Finished a review round, i have some inputs.

I think we could benefit from using `buildRoute` but i could not find an elegant way to use it while adding `version` queryparam, let me know if you have an idea.
But i think that we could also just define the route in the service.

I noticed scoped packages failed, notice my change requests that fix that, maybe we should add a test for scoped package as well.

MohanKumarAmbati:
> I think we could benefit from using `buildRoute` but i could not find an elegant way to use it while adding `version` queryparam, let me know if you have an idea. But i think that we could also just define the route in the service.

Just thinking whether we can do something like taking queryparam as input, if `queryParamSchema: queryParams || queryParamSchema, ` we'll give default queryparam.

services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,87 @@
...
+    let date
+    if (queryParams !== undefined && queryParams.version !== undefined) {
+      if (time[queryParams.version] === undefined) {
+        throw new InvalidParameter({ prettyMessage: 'version not found' })
+      }
+      date = dayjs(time[queryParams.version])
+    } else {
+      date = dayjs(time.modified)
+    }
+
```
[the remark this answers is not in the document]
MohanKumarAmbati: Totally agree, implemented the changes in - https://github.com/badges/shields/pull/10641/commits/a22b70755b6faa0cb12b80822aab34280116e94f

services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  static openApi = {
+    '/npm/last-update/{packageName}': {
+      get: {
+        summary: 'NPM Last Update',
+        parameters: [
+          pathParam({
+            name: 'packageName',
+            example: 'express',
```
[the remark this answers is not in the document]
MohanKumarAmbati: implemented in the commit - https://github.com/badges/shields/pull/10641/commits/a22b70755b6faa0cb12b80822aab34280116e94f

services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,87 @@
...
+    .pattern(Joi.string(), Joi.any())
+    .required(),
+}).required()
+
+export class NpmLastUpdate extends NpmBase {
+  static category = 'activity'
+
+  static route = {
+    base: 'npm/last-update',
+    pattern: ':packageName',
```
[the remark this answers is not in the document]
MohanKumarAmbati: oh okay, just came to know the meaning of scoped packages. 

In the commit - https://github.com/badges/shields/pull/10641/commits/a22b70755b6faa0cb12b80822aab34280116e94f, made changes to the `buildRoute` from `NpmBase` such that the extended class can invoke the `buildRoute` function with `queryParamsSchema` as one of the argument.

MohanKumarAmbati:
>I think we could benefit from using buildRoute but i could not find an elegant way to use it while adding version queryparam, let me know if you have an idea.

Other way - https://github.com/badges/shields/pull/10641/commits/96ced45260484f39e7d7cdba25f745383eecb4be, introduced new `additionalQueryParam`, if it is not null then it will be appended to the existing `queryParamSchema`. I think it would be better way, other packages if any can utilize as well.

chris48s:
Just quickly unwinding this.. what problem does the `?version=` param solve here?

AFAIK a npm version is immutable. Once you've pushed a release with a given version number you can't push that version again. So the "last updated" timestamp for a specific version is static. It is a value that isn't going to change (unlike "last updated" for the package itself or a tag, which will change over time as versions are published).

MohanKumarAmbati:
> Just quickly unwinding this.. what problem does the `?version=` param solve here?

This param helps to fetch `last updated` for a given `version` in the package.

Also yeah it won't change 🙂. As the data is available, I thought of giving it like an option to fetch last updated for a version.

MohanKumarAmbati:
> Just quickly unwinding this.. what problem does the ?version= param solve here?

@chris48s,

just tried to include versions somehow, gave it a second thought and looks like not much useful. Hence, removed it.

@jNullj, can you please review changes.

mbtools:
"Last update" by version is not very useful, indeed. However, it would be good to get the last update for a given *dist-tag* esp. for projects that keep several active release branches.

See [Verdaccio](https://www.npmjs.com/package/verdaccio?activeTab=versions), for example.

jNullj:
> "Last update" by version is not very useful, indeed. However, it would be good to get the last update for a given _dist-tag_ esp. for projects that keep several active release branches.
> 
> See [Verdaccio](https://www.npmjs.com/package/verdaccio?activeTab=versions), for example.

Makes sens, @MohanKumarAmbati can you add tags to your fetch and enable tags in path? (also openapi and tests)

MohanKumarAmbati:
Thanks for the input @mbtools, but looks like they're not exposing last updated time in the API  [verdaccio - next-8](https://registry.npmjs.org/verdaccio/next-8). please let me know if there is an API that we can make use of.

mbtools:
After you fetch the manifest (https://registry.npmjs.org/verdaccio), lookup the dist-tag and get the associated version:

image

Then use that version to get the time:

image
image

For next-8, the time would be `2024-10-08T06:34:29.911Z`.

MohanKumarAmbati:
Thanks for the input @mbtools, made changes accordingly - https://github.com/badges/shields/pull/10641/commits/d7cf23cad75a1a8538b44d41072fa91a04bedefb.

MohanKumarAmbati:
Thank you @chris48s

services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,97 @@
...
+    const packageData = await this.fetch({
+      registryUrl,
+      scope,
+      packageName,
+      schema: updateResponseSchema,
+    })
+
+    let date
+
+    if (tag && tag in packageData['dist-tags']) {
```
jNullj, line 84: `tag in packageData['dist-tags']` I think we could improve a bit if we test the array once and use the result twice. I would expect this to be a long array for some packages.

services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,97 @@
...
+import NpmBase, { packageNameDescription } from './npm-base.js'
+
+const updateResponseSchema = Joi.object({
+  time: Joi.object({
+    created: Joi.string().required(),
+    modified: Joi.string().required(),
+  })
+    .pattern(Joi.string(), Joi.any())
+    .required(),
+  'dist-tags': Joi.object().pattern(Joi.string(), Joi.any()).required(),
```
jNullj, line 15:
```suggestion
  'dist-tags': Joi.object().pattern(Joi.string(), Joi.string()).required(),
```

jNullj: The value should be a semver-compliant version *string* as far as i can tell from the docs
https://github.com/npm/registry/blob/ae49abf1bac0ec1a3f3f1fceea1cca6fe2dc00e1/docs/responses/package-metadata.md?plain=1#L194C1-L194C55

jNullj (commented):
Two more small things, but overall seems alright.

jNullj (approved):
LGTM, Thank you @MohanKumarAmbati  for the contribution.
image

services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,100 @@
...
+
+const updateResponseSchema = Joi.object({
+  time: Joi.object({
+    created: Joi.string().required(),
+    modified: Joi.string().required(),
+  })
+    .pattern(Joi.string(), Joi.any())
+    .required(),
+  'dist-tags': Joi.object().pattern(Joi.string(), Joi.string()).required(),
+}).required()
```
chris48s, line 16: This schema could do with tightening up a bit. My suggestion would be:

```suggestion
const updateResponseSchema = Joi.object({
  time: Joi.object({
    created: Joi.string().required(),
    modified: Joi.string().required(),
  })
    .pattern(Joi.string().required(), Joi.string().required())
    .required(),
  'dist-tags': Joi.object()
    .pattern(Joi.string().required(), Joi.string().required())
    .required(),
}).required()
```

@jnullj As a tip for future reviews:

It is not invalid to ever use `Joi.any()` in a schema, but it is quite rare that we want to be that loose. Usually if you see that in a PR, the next question is: "can we tighten this up?" Similarly, it isn't invalid for something in a schema to ever be optional, but usually any value we're going to use should be required.

If the upstream API does require us to accept optional values or variable types in the schema, then we generally need some additional checks in the handler code to compensate for it.

services/npm/npm-last-update.service.js
```diff
@@ -0,0 +1,100 @@
...
+      }
+
+      date = dayjs(packageData.time[tagVersion])
+    } else {
+      const timeKey = packageData.time.modified ? 'modified' : 'created'
+
+      date = dayjs(packageData.time[timeKey])
+    }
+
+    return this.constructor.render({ date })
```
chris48s, line 98: Given the only thing we're validating about the input here is that it is a string, we should check the `date.isValid` property after parsing to a dayjs object and throw if it is invalid before we pass the value on to render - same as we did in https://github.com/badges/shields/pull/10585

chris48s (commented):
Thanks for your work on this @MohanKumarAmbati , and for taking the review @jNullj 

Just having a quick look over this I have a couple of additional comments.

mbtools:
nice work. thanks @MohanKumarAmbati 🙏

chris48s:
Thanks. Good work all. I'm going to merge this.

One thing I have realised about this badge is that it will work fine for most packages, but the `https://registry.npmjs.com/<packagename>` response can get really big for some packages. For example, https://registry.npmjs.com/npm is 22Mb of JSON

Shields has a hard limit on the size of response we will accept from an API. We cap it at 10Mb so for some subset of packages this badge will not work.

I think that's just something we live with. In reality the number of packages that actually generate such a huge response (>10Mb) is quite small.

chris48s (approved):

mbtools:
The typescript package is another one (13 MB). You could get the abbreviated manifest first (https://github.com/npm/registry/blob/main/docs/responses/package-metadata.md). That has last-modified time already and the tags. Then if a tag was selected, get the individual version json instead of the complete manifest.

jNullj:
That's a huge difference.

```
$ curl -H "Accept: application/vnd.npm.install-v1+json" https://registry.npmjs.org/npm > test-abbreviated.json
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 2194k  100 2194k    0     0  25.7M      0 --:--:-- --:--:-- --:--:-- 26.4M

$ curl https://registry.npmjs.org/npm > test.json
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 21.1M  100 21.1M    0     0  48.4M      0 --:--:-- --:--:-- --:--:-- 48.4M
$ ls -la --block-size=MB test*
-rw-r--r-- 1 user 197609 23MB Nov  2 23:34 test.json
-rw-r--r-- 1 user 197609  3MB Nov  2 23:34 test-abbreviated.json
```

From the test-abbreviated we can get `next-7: "7.24.2"`
then in theory you could call `https://registry.npmjs.com/npm/7.24.2` or `https://registry.npmjs.com/npm/next-7` (maybe call next-7 without the first step as well)
But the response from `/package/version` does not seem to include the any timestamp... so im not sure we could use it here.

The only thing i think is useful with abbreviated response is when we don't use a tag.

Altho, i think i just found a trick ;)
Notice how this part of the abbreviated response file name is picked? looks like a timestamp
```json
"_npmOperationalInternal":{
  "tmp":"tmp/npm_7.24.2_1633367792820_0.6414313657198436",
  "host":"s3://npm-registry-packages"
}
```
`163336779282 -> Monday, 4 October 2021 17:16:32.820`
And full package metadata response returns
`7.24.2: "2021-10-04T17:16:33.168Z"`

Not sure if we can trust this, but it sure looks like a cool party trick

chris48s:
Making 2 requests in sequence has its own performance impact. I think a reasonable solution would be:

- If the badge URL is `/npm/last-update/{packageName}` request the abbreviated version
- If the badge URL is `/npm/last-update/{packageName}/{tag}` request the full version

That would be a huge improvement for the most common case.

I think we should stick to relying on the documented API behaviour. Relying on undocumented assumptions about API internals generally ends up biting us in the long run.

PR #10642, opened 2024-10-27
add content security policy header to SVG responses

chris48s:
This PR does not resolve an open security issue. However, it does proactively add another layer of protection on top of what we already do to prevent cross-site scripting issue.

SVGs can contain javascript.
When they're embedded with an `` tag, SVGs are rendered in [secure static mode](https://www.w3.org/TR/SVG2/conform.html#secure-static-mode) which disables any embedded javascript, along with some other restrictions.
When they're embedded with a `<object>` tag or loaded directly, this is not the case.

While we take steps to prevent XSS issues, we have on at least one occasion discovered an escaping issue, which made it possible for someone to inject script into a badge image. Reference: https://github.com/badges/shields/pull/3511

In general, I think our escaping game is pretty solid these days.

That said, it is not impossible that someone might find a way around the mechanisms we have in place or we might introduce an escaping bug in some future refactor.

For this reason, I suggest we serve our SVG images with the `Content-Security-Policy: script-src 'none';` header. This is a content security policy header that basically says "no javascript allowed on this response" and offers a strong browser-level protection. This header is respected when the image is loaded directly, but also crucially when embedded with an `<object>` tag. Essentially, this would mean that even if someone found a way to sneak a `<script>` tag into a SVG response served from shields.io, any modern browser would refuse to run that script.

### DIFF
--- a/core/base-service/legacy-result-sender.js
+++ b/core/base-service/legacy-result-sender.js
@@ -11,6 +11,7 @@ function streamFromString(str) {
 
 function sendSVG(res, askres, end) {
   askres.setHeader('Content-Type', 'image/svg+xml;charset=utf-8')
+  askres.setHeader('Content-Security-Policy', "script-src 'none';")
   askres.setHeader('Content-Length', Buffer.byteLength(res, 'utf8'))
   end(null, { template: streamFromString(res) })
 }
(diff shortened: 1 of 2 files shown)


PyvesB (approved):
Makes sense, let's do this!

PR #10655, opened 2024-11-04
Fix broken URL for pingpong.one

usr3:
Fix broken URL for https://pingpong.one for uptime status monitoring.


PyvesB (approved):
Thanks for the contribution!

PR #10656, opened 2024-11-04
Refactor - use renderVersionBadge - part 4 [githubrelease githubtag]

### LINKED ISSUE #2026: Ensure consistent usage of renderVersionBadge

jNullj:
This pull request refactors the codebase to use the renderVersionBadge function for rendering version badges.
Refactored github releases and github tags.
The pull request also modifies tests where needed.

I added `forcePrerelease` option to renderVersionBadge for github, might be usefull for any api that indicates prerelease explicitly.

while for github release we don't effect colors, for github tags this pr changes the logic a bit, and improves horizontal consistency.
we used to mark all tags blue if user picked latest over semver sort. this is odd as with releases we do mark prerelease in orange.
I guess the diffrence is due to api indicating prereleases, but we already try to guess prereleases in renderVersionBadge so i figured we might just as well try and use it while increasing consistency.

This PR should be the last in this refactor, we do have a few outliers services like `

visual-studio-app-center-releases-osversion` or `clojars` but i couldn't find anything else to refactor, let me know if you find anything.
Until you do, closes #2026


services/version.js
```diff
@@ -237,6 +237,7 @@ function rangeStart(v) {
  * @param {string} [options.prefix]  - The prefix to display on the message, such as ">=", "v", overrides the default behavior of using addv
  * @param {string} [options.postfix] - The postfix to display on the message, such as "tested"
  * @param {Function} [options.versionFormatter=versionColor] - The function to use to format the color of the badge based on the version number
+ * @param {boolean} [options.forcePrerelease] - Whether to force the badge to be considered a prerelease for cases where upstream API indicates a prerelease version but it is not semantically a prerelease
```
chris48s, line 240: I'd probably just call this `isPrerelease`

The way I'd view this is not so much that we're allowing platforms to indicate that a semantically stable release is actually a pre-release. More just that it allows a platform to explicitly indicate a release is a pre-release rather than us having to try and infer it.

jNullj: Its probably for the better, our goal is to serve from upstream rather then infer our own conclusion.
Changed with [c6714e3](https://github.com/badges/shields/pull/10656/commits/c6714e350119592275e86fd61059b8c179dcc029)

chris48s (approved):

PR #10658, opened 2024-11-07
fix [githubpipenv] service tests

jNullj:
replaced tested package that changed on remote repo to avoid failure.

as discovered in pr #10656 at [this ci test](https://github.com/badges/shields/actions/runs/11673579580?pr=10656)


services/github/github-pipenv.tester.js
```diff
@@ -83,9 +82,9 @@ t.create('Locked version of unknown dependency')
 
 t.create('Locked version of VCS dependency')
   .get(
-    '/locked/dependency-version/thorn-oss/perception/dev/videoalignment.json',
+    '/locked/dependency-version/metabolize/rq-dashboard-on-heroku/dev/black.json',
```
chris48s, line 85: This isn't a VCS dependency it is a pypi dependency, so doing this just makes this replicate another test. We need to find an example of a project using a VCS dependency. Let me see if I can find one.

chris48s: So here's a few examples of what we are looking for:

https://github.com/tomoki171923/aws-service-package/blob/53ea2e401b2183f442daef16f1d5b2f4a662e7b5/Pipfile.lock#L58-L62

https://github.com/formularium/formularium-backend/blob/a516a315a3c0a9a1f18b5576f11f2a56eab0bb50/Pipfile.lock#L367-L370

https://github.com/crowdgames/mdp-level-assembly/blob/3baf41f4cd313b19547f42359115347f62f71da3/Pipfile.lock#L19-L22

I have no idea how much we can expect any of those examples to remain stable. If this is something that keeps breaking, we could consider converting this test case to a mock or unit test.

jNullj: I ended up picking pipenv's locked vcs dependency pypiserver, i assume its a big project and is stable.
also very related to our tests here ;)

The version ref is a commit hash so i added a text-validation for commit hash, i bet it might be useful in the future.

chris48s: reasonable choices all round :+1:

chris48s (approved):

PR #10666, opened 2024-11-10
reduce overhead of NPM Last Update badge; test [npm]

chris48s:
In this PR, I am following up on my suggestion in https://github.com/badges/shields/pull/10641#issuecomment-2453554943

- If the badge URL is `/npm/last-update/{packageName}` request the abbreviated API response
- If the badge URL is `/npm/last-update/{packageName}/{tag}` request the full version

This will reduce the amount of json we have to download and parse in order to serve a badge in the case where when a tag is not specified. For packages (e.g: npm) that produce a really large response, it would make it possible to use the tag-less variant.

### DIFF
--- a/services/npm/npm-base.js
+++ b/services/npm/npm-base.js
@@ -81,8 +81,11 @@ export default class NpmBase extends BaseJsonService {
   }
 
   async _requestJson(data) {
-    return super._requestJson(
-      this.authHelper.withBearerAuthHeader({
+    let payload
+    if (data?.options?.headers?.Accept) {
+      payload = data
+    } else {
+      payload = {
         ...data,
         options: {
           headers: {
@@ -91,8 +94,9 @@ export default class NpmBase extends BaseJsonService {
             Accept: '*/*',
           },
         },
-      }),
-    )
+      }
+    }
+    return super._requestJson(this.authHelper.withBearerAuthHeader(payload))
   }
 
   async fetchPackageData({ registryUrl, scope, packageName, tag }) {
@@ -144,7 +148,13 @@ export default class NpmBase extends BaseJsonService {
     return this.constructor._validate(packageData, packageDataSchema)
   }
 
-  async fetch({ registryUrl, scope, packageName, schema }) {
+  async fetch({
+    registryUrl,
+    scope,
+    packageName,
+    schema,
+    abbreviated = false,
+  }) {
     registryUrl = registryUrl || this.constructor.defaultRegistryUrl
     let url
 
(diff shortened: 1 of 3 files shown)


services/npm/npm-base.js
```diff
@@ -81,8 +81,11 @@ export default class NpmBase extends BaseJsonService {
   }
 
   async _requestJson(data) {
-    return super._requestJson(
-      this.authHelper.withBearerAuthHeader({
+    let payload
+    if (data?.options?.headers?.Accept) {
+      payload = data
```
chris48s, line 86: if we have already set our own `Accept` header, don't overwrite it

services/npm/npm-last-update.service.js
```diff
@@ -3,55 +3,111 @@ import dayjs from 'dayjs'
...
+      message: formatDate(date),
+      color: ageColor(date),
+    }
+  }
+}
+
+export class NpmLastUpdateWithTag extends NpmLastUpdateBase {
+  static route = {
+    base: 'npm/last-update',
+    pattern: ':scope(@[^/]+)?/:packageName/:tag',
```
chris48s, line 40: Given how different the two implementations for with/without tag are, I've just split this into two classes. One with a (required) tag param and one without it. This means we need to specify the routes manually instead of using the `buildRoute()` helper, which makes the `tag` param optional.

services/npm/npm-last-update.tester.js
```diff
@@ -3,51 +3,79 @@ import { createServiceTester } from '../tester.js'
...
+    label: 'last updated',
+    message: isFormattedDate,
+  })
+
+t.create('last updated date, with tag, invalid package with scope')
+  .get('/@not-a-scoped-package/not-a-valid-package/doesnt-matter.json')
+  .expectBadge({
+    label: 'last updated',
+    message: 'package not found',
+  })
```
chris48s, line 81: Because I've split with/without tags into 2 classes, I've doubled up the test suite so we're testing both classes for each scenario.

services/npm/npm-last-update.service.js
jNullj: Not sure if it improves anything but we could use `buildRoute('npm/last-update' ,{ withTag = false })` here and keep the required tagged route as is.

jNullj (commented):
We might be able to save a line with `buildRoute` but im not sure if it isn't better to keep that as its, see my comment.

Overall the changes look good to me.

jNullj (approved):

PR #10667, opened 2024-11-10
Rename `postfix` param on `renderVersionBadge()` to `suffix`; test [wordpress]

chris48s:
This PR renames the `postfix` param on `renderVersionBadge()` to `suffix`.
This brings it into line with the terminology we use elsewhere in the codebase. e.g:

- https://github.com/badges/shields/blob/8af909d118d2cc2796b9769baa053fca39ab2f14/services/gitea/gitea-helper.js#L30
- https://github.com/badges/shields/blob/8af909d118d2cc2796b9769baa053fca39ab2f14/services/dynamic/dynamic-helpers.js#L8
- https://github.com/badges/shields/blob/8af909d118d2cc2796b9769baa053fca39ab2f14/services/downloads.js#L55
etc


jNullj (approved):
LGTM

PR #10672, opened 2024-11-13
add jNullj, remove inactive maintainers

chris48s:
This PR reflects changes that I've already made to the relevant groups in GitHub.
Welcome to the core team, @jNullj

PR #10673, opened 2024-11-14
remove 'operations' section from README

chris48s:
follow up from https://github.com/badges/shields/pull/10672#issuecomment-2474848778


jNullj (approved):
LGTM, I also think users don't benefit from the operations section.

PR #10681, opened 2024-11-15
Add renderDateBadge helper; affects [aur BitbucketLastCommit chrome date eclipse factorio galaxytoolshed GiteaLastCommit GistLastCommit GithubCreatedAt GithubHacktoberfest GithubIssueDetail GithubLastCommit GithubReleaseDate GitlabLastCommit maven npm openvsx snapcraft SourceforgeLastCommit steam vaadin visualstudio wordpress]

chris48s:
There's a couple of motiviations behind this PR.

## 1. Validation

Dayjs is quite flexible in the inputs it accepts, but it also allows you to construct invalid date objects. So for example `dayjs('not-a-date')` doesn't throw, but also doesn't yield a useful date object. I don't love this behaviour tbh, but dayjs was the easiest thing to migrate to following the demise of moment.js and I don't really have the appetite to replace it.

The way round this is: If we aren't able to validate our input sufficiently before constructing a date object, we need to then check the date is valid before we do anything with it.

So really what we want to be doing is:

```js
import dayjs from 'dayjs'

const date = dayjs(data['last-updated'])

if (!date.isValid()) {
  throw new InvalidResponse({ prettyMessage: 'invalid date' })
}

// do stuff with date
```

The problem is: This is a bit cumbersome, and sometimes we do it wrong. For example, there are places in the codebase where we are not checking `.isValid()` at all and there are some places where we are (incorrectly) checking `.isValid` instead of `.isValid()` e.g:

https://github.com/badges/shields/blob/master/services/npm/npm-last-update.service.js#L84-L88

Whoops! Anyway, we're not really setting ourselves up for success here.

So I wanted to write the correct validation once in a helper function, and then use it everywhere.

## 2. Reduce duplication

There are lots and lots of places where we are writing some slight variant of:

```js
static render(date) {
  const color = age(date)
  const message = formatDate(date)
  return { color, message }
}
```

in service classes. There is scope for us to write one helper function, use it in lots of places, and clean up a couple of inconsistencies.

---

To be honest, I've done a bit too much stuff all in one commit here, so this is probably a bit of a pain to review. Ideally I would have split out some of the "moving stuff from one file to another" work into separate commits. On this occasion, that's just how it went down. Sorry :(

Fundamentally, the new code in this PR is `parseDate` and `renderDateBadge`. `age`, `formatDate` and `formatRelativeDate` (plus their tests) are just moving from one file to another with no significant modifications. Then everything else is just updating services to use `parseDate` and `renderDateBadge`.


services/date.js
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Creates a badge object that displays a date
+ *
+ * @param {Date | string | number | dayjs } date JS Date object, string, unix timestamp or dayjs object
+ * @param {boolean} reversed Reverse the color scale (the older, the better)
+ * @returns {object} A badge object that has two properties: message, and color
+ */
+function renderDateBadge(date, reversed = false) {
+  const d = parseDate(date)
```
chris48s, line 87: For most badges, this basically pushes validation into being a rendering concern. Conceptually, I don't love this, but it seems like the right tradeoff. Unless we need to parse a date with an unusual format string, we can now basically just grab something that is probably a date from an API, pass it to `renderDateBadge()` and it will either render right or throw the correct error.

services/wordpress/wordpress-last-update.service.js
```diff
@@ -1,16 +1,12 @@
...
+import { parseDate, renderDateBadge } from '../date.js'
 import { description, BaseWordpress } from './wordpress-base.js'
-dayjs.extend(customParseFormat)
 
 const extensionData = {
   plugin: {
     capt: 'Plugin',
     exampleSlug: 'bbpress',
-    lastUpdateFormat: 'YYYY-MM-DD hh:mma [GMT]',
+    lastUpdateFormat: 'YYYY-MM-DD h:mma [GMT]',
```
chris48s, line 9: This was not strictly related to this task, but was something I noticed while I was working on this. The hours part of timestamps in this format omits the leading zero if it is single digit, so this should be `h` not `hh`.

chris48s:
replaced by https://github.com/badges/shields/pull/10682 so the gist tests get a token to run

PR #10683, opened 2024-11-15
Add blog post about token pool, improve 'authorise our app' CTA

chris48s:
Every documentation page for a GitHub badge has a call to action on it that says:

> If your GitHub badge errors, it might be because you hit GitHub's rate limits. You can increase Shields.io's rate limit by [adding the Shields GitHub application](https://img.shields.io/github-auth) using your GitHub account.

That sentence has been around for a very long time, and lots of people have authorised the app. However, I don't think it is very good for two reasons:

1. It doesn't really explain how this works, exactly what you're sharing with us or how to opt out later.
2. It doesn't really reflect the current reality. Today we have plenty of tokens. If you're seeing an error it almost definitely isn't going to be solved by authorising the app.

So in this PR I am attempting to write up a better explanation of what this is and how it works that is more up-to-date and more transparent for users.

### DIFF
--- a/frontend/blog/2024-11-14-token-pool.md
+++ b/frontend/blog/2024-11-14-token-pool.md
@@ -9,7 +9,7 @@ authors:
 tags: []
 ---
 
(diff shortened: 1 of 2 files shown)


jNullj (approved):
looks good, liked the idea to give a short description then link to a blog post with the full details.

PR #10685, opened 2024-11-20
feat: global version category prefix query param [amo archlinux aur bower cdnjs chromewebstore cocoapods conan conda cookbook cpan cran crates ctan curseforge debian dockerVersion dub eclipsemarketplace elmpackage f-droid factorio fedora feedz flathub galaxytoolshed gem gitea]

jNullj:
Altho i am aware of #6135 this is an attempt to make this manageable.
I dont mind if this ends up unused as i took this as an opportunity to play a bit with the codebase internals.

Let me know if you think this change is elegant and easy to maintain or more of an anchor might risks confusion in the codebase.

Introduce global category-specific query parameters and add versionPrefix support across version badges.

### DIFF
--- a/core/base-service/base.js
+++ b/core/base-service/base.js
@@ -20,7 +20,7 @@ import {
   Deprecated,
 } from './errors.js'
 import { fetch } from './got.js'
-import { getEnum } from './openapi.js'
+import { getEnum, categoryGlobalQueryParams } from './openapi.js'
 import {
   makeFullUrl,
   assertValidRoute,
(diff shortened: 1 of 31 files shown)


chris48s:
I think I am against adding this as a user configuration option but I won't block it if others agree.

PyvesB:
Thanks for taking a stab at this!

I'm also lukewarm about adding a new user-facing configuration option. It adds a tiny bit of complexity both in our codebase and in the number of choices the user is faced with when creating a new badge. It would also mean that the information shown by two badges belonging to the same service may be formatted slightly differently based on the personal preference of whoever put the badge on their README; whilst not the end of the world, this would effectively degrade horizontal consistency.

jNullj:
I read your feedback and gave this another thought.
While i had fun with the idea, this does break horizontal consistency while also making the code a bit harder to maintain.
I also get how this, and using too many options can cause decision fatigue for UX.

You're the old hand here. If something feels off about this change, I trust your gut.

Thank you for taking the time about this.

PR #10688, opened 2024-11-21
remove obsolete [CratesSize] test for null size

jNullj:
Eliminate a test case that is no longer relevant due to upstream changes that removed all null crate sizes.

see also upstream merge that removed null for versions.crate_size upstream at rust-lang/crates.io#9926

### DIFF
--- a/services/crates/crates-base.js
+++ b/services/crates/crates-base.js
@@ -4,8 +4,7 @@ import { BaseJsonService, InvalidResponse } from '../index.js'
 
 const versionSchema = Joi.object({
   downloads: nonNegativeInteger,
-  // Crate size is not available for all versions.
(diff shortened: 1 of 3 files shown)


chris48s:
If we don't expect to receive null sizes any more, then we shouldn't just delete the test. We should update/remove the application code that deals with that situation as well:

https://github.com/badges/shields/blob/9447077c0836bfe5b4f76f534ebf98a0f402763f/services/crates/crates-base.js#L8
https://github.com/badges/shields/blob/9447077c0836bfe5b4f76f534ebf98a0f402763f/services/crates/crates-size.service.js#L53-L55

chris48s (approved):

PR #10699, opened 2024-11-23
auto suspend/resume review apps

chris48s:
This PR enables Fly's auto suspend/resume feature for review apps.
This will shut review app machines down if they've been idle for 5 mins and then restart them on demand on an incoming request (pretty much how the old heroku review apps used to work).
This should save us some money on review apps.
We could also consider moving the cleanup review apps job to run less frequently so we don't have to keep re-deploying the review apps if we have long-running reviews, although I have not done it in this PR.

### DIFF
--- a/fly.toml
+++ b/fly.toml
@@ -25,6 +25,8 @@ processes = []
   processes = ["app"]
   protocol = "tcp"
   script_checks = []
+  auto_stop_machines = "suspend"
+  auto_start_machines = true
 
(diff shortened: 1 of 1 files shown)


chris48s:
The review app at https://pr-10699-badges-shields.fly.dev/ will sleep after 5 mins of inactivity and resume on an incoming request.

PyvesB (approved):
Good idea!

PR #10701, opened 2024-11-23
WIP run multiple server threads

chris48s:

### DIFF
--- a/entrypoint.spec.js
+++ b/entrypoint.spec.js
@@ -1,18 +1,19 @@
 import { expect } from 'chai'
 import isSvg from 'is-svg'
 import got from './core/got-test-client.js'
+import startServer from './server.js'
 
-let serverModule
+let server
 before(async function () {
   this.timeout('30s')
   // remove args coming from mocha
   // https://github.com/badges/shields/issues/3365
   process.argv = []
-  serverModule = await import('./server.js')
+  server = await startServer()
 })
 
 after('shut down the server', async function () {
-  await serverModule.server.stop()
+  await server.stop()
 })
 
 it('should render a badge', async function () {
--- a/server.js
+++ b/server.js
@@ -1,3 +1,5 @@
+import cluster from 'cluster'
+import { availableParallelism } from 'os'
 import fs from 'fs'
 import path from 'path'
 import { fileURLToPath } from 'url'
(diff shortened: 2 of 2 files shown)


entrypoint.spec.js
```diff
@@ -1,18 +1,19 @@
...
 
-let serverModule
+let server
 before(async function () {
   this.timeout('30s')
   // remove args coming from mocha
   // https://github.com/badges/shields/issues/3365
   process.argv = []
-  serverModule = await import('./server.js')
+  server = await startServer()
```
chris48s, line 12: The core problem I ran into here was that `cluster.fork()` attempts to spawn a new worker process that executes the same script as the primary process. When a worker process is spawned, it executes the same Node.js script that started the primary process, but this didn't have the desired effect under test because the primary process is the test runner.
I worked round this by
a) exporting a function we can test from `server.js`
b) only starting in multi-threaded mode if we are not under test
This means the actual multi-threaded server startup is **not** under test, but I couldn't find any way around this. I think it is a tradeoff we would have to accept if we go down this road.

server.js
```diff
@@ -25,11 +27,13 @@ Sentry.init({
   },
 })
 
-if (+process.argv[2]) {
-  config.public.bind.port = +process.argv[2]
-}
-if (process.argv[3]) {
-  config.public.bind.address = process.argv[3]
+if (process.env.NODE_CONFIG_ENV !== 'test') {
```
chris48s, line 30: Switching from a dynamic `await import` to a module-level import also meant that this was picking up the test runner args here https://github.com/badges/shields/blob/9447077c0836bfe5b4f76f534ebf98a0f402763f/.github/actions/core-tests/action.yml#L13 under test.

chris48s:
This might not actually be a good idea. Reposting from discord:

Node is single-threaded so my assumption was that if we deployed on a machine with 2 vCPUs, what we would see is one vCPU doing all the work and the other one sitting around doing virtually nothing (and we would need to do something like I've proposed in this PR to actually utilize 2 cores). Turns out this is wrong. With no code changes, we actually balance across 2 cores reasonably well.

Screenshot_at_2024-11-26_19-40-35

Having done some reading and experimenting, it turns out the OS scheduler can distribute the work of a single-threaded application across multiple vCPUs. Even though node itself uses a single thread for the main event loop, the thread can be switched between the available vCPUs.

My current thinking on this is that doing something like this might actually be counter-productive. What we're aiming to do isn't to max out 2 cores, it is to sparingly spread out our CPU usage across 2 cores (each of which has a restrictive usage quota) so what we're doing now (running single threaded on a shared-2x) might be a better way to achieve that.

Issue #1805, opened 2018-07-25
Rate limit github-auth endpoints


LVMBDV:
I noticed this issue while I was working on #541. OAuth web application flows require a randomly generated `state` parameter to authenticate callback (redirection) requests from the client. Without this security check, adversaries can orchestrate a denial of service attack using the callback endpoint. For more info see:
- https://developer.github.com/apps/building-oauth-apps/authorizing-oauth-apps/#web-application-flow
- https://stackoverflow.com/questions/26132066/what-is-the-purpose-of-the-state-parameter-in-oauth-authorization-request

TL;DR:

![20180725202313](https://user-images.githubusercontent.com/17510795/43216922-e2d2c102-9048-11e8-995e-55006cc39ab5.png)

paulmelnikow:
Just emailed you; thanks.

paulmelnikow:
Clarified offline that this is low severity. We should still fix it quickly. 😬

paulmelnikow:
It's not clear how to fix this. The `state` is exposed to the client; it's not a secret. We don't preserve any state between requests, and any state we did preserve would need to be shared between three servers. That's the least of it, really. The `/github-auth` endpoint _hands out_ these strings. Since we don't authenticate those requests, it's not clear the `state` can do anything to help.

We could consider a rate limit.

LVMBDV:
In my opinion, this and the token pool solidifies the need for separate state storage such as a key-value database like Redis.

> We could consider a rate limit.

That's like demolishing a building so terrorists can't blow it up.

paulmelnikow:
How would state storage solve the problem? If one endpoint gives out tokens as fast as you want, and another consumes them, it doesn’t matter whether they are single use or not.

It’s more like putting a fence around it. We probably could limit to one token per 30 sec per IP, and 5 per 30 seconds overall, without much inconvenience to anyone.

LVMBDV:
You're right, rate-limiting the auth endpoint makes sense.

PR #10702, opened 2024-11-23
Add OAuthStateManager for CSRF protection in GitHub OAuth flow

jNullj:
Introduce an OAuthStateManager class to manage states for CSRF protection during GitHub OAuth authentication, addressing issue #1805, This implementation generates and validates states to enhance security against CSRF attacks.

this pr is public as its an open public issue.
draft as i didn't test yet, and i need to add automatic tests as well

### DIFF
--- a/services/github/auth/acceptor.js
+++ b/services/github/auth/acceptor.js
@@ -1,12 +1,16 @@
 import queryString from 'query-string'
 import { fetch } from '../../../core/base-service/got.js'
 import log from '../../../core/server/log.js'
+import { OAuthStateManager } from '../../oauth-utils.js'
+
+const githubStateManager = new OAuthStateManager()
 
 function setRoutes({ server, authHelper, onTokenAccepted }) {
   const baseUrl = 'https://img.shields.io'
 
   server.route(/^\/github-auth$/, (data, match, end, ask) => {
-    ask.res.statusCode = 302 // Found.
+    const githubCSRFclientId = githubStateManager.generateAnonymizedClientId()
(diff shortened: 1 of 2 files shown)


chris48s:
I haven't really looked at this in much detail. Tbh, I've not 100% got my head round the attack vector here. However, there's one thing I want to flag quickly here before we go any further with this..

- `OAuthStateManager.states` is a `Map()` that lives in memory on one server
- In production, shields runs on lots of servers (at minimum 14)
- If you assume that the server that processes the request to `/github-auth` and the server that processes the request to `/github-auth/done` will be 2 different servers (there is a >90% chance this will be true in production), does this approach work?

It looks to me like it wouldn't. i.e: if the computer that does the `.set()` isn't the same computer that later tries to do the `.get()` this check will always fail, right?

jNullj:
That's current, I wrongly assumed here we run on one instance, its easy to forget as I'm used to debug on a single container and usually write single instance applications.

I tried to think of a way to sync this info that we already used, github api tokens came to mind as they should also be persistent across servers, and i noticed we used postgresSQL, do you think using it is a viable strategy?

I would love to ditch persistent storage for keys altogether but then an attacker could forge a valid key for the CSRF attack for all clients, do you think we could bind the CSRF token to something else? like IP address & user agent mix hash or something similar?

I think i might need a bit more reading about this, it appears many save time by picking a framework to handle csrf for them.

chris48s:
Before we get any further into this, I've invited you to a private thread to try and dig into what problem we're actually trying to solve here. I think I need to understand what problem are trying to solve before we can design a solution to it.

Issue #555, opened 2015-11-02
AppVeyor build artifact-based badges


AArnott:
I'd like to display badges based on the contents of an AppVeyor build's artifact files. The build would drop special text files which AppVeyor would report as artifacts. These text files would contain just one line of text:

`<SUBJECT>-<STATUS>-<COLOR>`

The artifact filenames would follow some agreed upon convention such as `SOMENAME.shield`. 

The shield could then be displayed via a URL like this:
https://img.shields.io/appveyor/ci-artifact/USERNAME/PROJECT/SOMENAME.svg

See https://github.com/AArnott/pinvoke/issues/24 for a full description of what we want to create.

espadrine:
I don't know exactly how AppVeyor's API allows to fetch artifacts, but that is definitely an interesting use-case. A difficult aspect of this is to convey how it works to others from the main page.

AArnott:
Thanks for reviewing, @espadrine. And I appreciate that your model is so simple that each use generally needs just a one line description. That's fantastic.

If you're agreeable to this design, can you tell us what timeline it might be implemented in, or whether you'd entertain a pull request that adds the functionality?

espadrine:
@AArnott It would take me at least a few months to get to implementing this, as most of the time I have dedicated to shields.io is spent on sysadmin, maintaining existing APIs that weather down or hit unexpected limits, and reviewing PRs. I would absolutely entertain a PR that adds the functionality. It may need work in try.html to have an explanation in the dialog box.

AArnott:
Thanks, @espadrine.

paulmelnikow:
The Endpoint badge #2838 provides a building block for this.

PR #10705, opened 2024-11-28
[ReproducibleCentral] add Reproducible Central in Dependencies

hboutemy:
add new service for Reproducible Central: https://github.com/jvm-repo-rebuild/reproducible-central/tree/master

it will replace current endpoint badges used with a better approach: see example on https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/maven/maven/README.md


chris48s:
Before we attempt to get into "is this good code" can you give me a quick overview of:

- What does this service do?
- What information is this badge communicating to users?
- etc

just so I have some context on what you are trying to achieve to review this.

hboutemy:
very reasonable approach, given the diversity of requests you get :)
(and I'm definitively not a JavaScript expert, help welcome on code...)

I'll try to be progressive, even if it means the answer will be long: sorry, I don't see how to introduce this shorter

## What does this service do?

This service is about:
1. Reproducible Builds https://reproducible-builds.org/
2. applied to the most popular Java repository = the content of Maven Central.

I'm working in this project for a good number of years now, with https://reproducible-builds.org/docs/jvm/ as a key documentation

A few years ago, we managed to improve JVM build tools to be able to do Reproducible Builds: next step has been to check that OSS projects apply Reproducible Builds in their builds
=> I created "Reproducible Central" to gather success and remaining issues of RB applied to popular JVM OSS: results are published https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/README.md

I proposed a static badge using for projects to show in their GH source that they are reproducible: see for example https://github.com/apache/maven/blob/master/README.md

And I did a few talks at various conferences to promote Reproducible Builds, particularly done with Apache Maven as an Apache Software Foundation member and maintainer of Maven: https://fr.slideshare.net/slideshow/coc-na-2023-reproducible-builds-for-the-jvm-and-beyondpptx/261996375

## What information is this badge communicating to users?

The new badge I'm submitting here is the next level of all this journey: it will permit for projects not only to have a badge for their build, but go one step beyond and see the reproducibility status of their dependencies

I did a first proof of concept of that next step, based on badge endpoint that I used until now: https://maven.apache.org/plugins-archives/maven-artifact-plugin-LATEST/reproducible-central.html

As you can see from this report, it lists all dependencies and for each dependency displays the reproducibility status as checked by Reproducible Central

The move from badge endpoint to service is what will permit to manage 404 = dependencies that have not been checked by Reproducible Central

With this type of badge, I can get more interest by everybody, because they'll want to check their dependencies, in addition to improve their own build

chris48s:
Thanks. I'll aim to read this and get back to you with some comments tomorrow

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
+import Joi from 'joi'
+import { BaseJsonService, pathParams } from '../index.js'
+
+const schema = Joi.object({
+  color: Joi.string().required(),
+  message: Joi.string().required(),
+}).required()
+
+export default class ReproducibleCentral extends BaseJsonService {
+  static category = 'dependencies'
```
chris48s, line 10: To help me understand if this is the right place for this, can you explain the score. If my package scores 10/13, what is the 10 and what is the 13?

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
...
+
+  static route = {
+    base: 'reproducible-central/a',
```
chris48s, line 13: What does '/a' mean here?

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
...
+            name: 'version',
+            example: '3.9.9',
+          },
+        ),
+      },
+    },
+  }
+
+  static defaultBadgeData = {
+    label: 'Reproducible Builds',
```
chris48s, line 40: From https://github.com/badges/shields/blob/master/CONTRIBUTING.md#badge-guidelines

> The left-hand side of a badge should not advertise. It should be a lowercase noun succinctly describing the meaning of the right-hand side

It is hard to make a suggestion for this because I'm still not really sure what the score means, but basically the default label should describe the meaning of this metric rather than the service that produced it. Although if users want to customise the label, they can.

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
...
+    labelColor: '1e5b96',
```
chris48s, line 41: Can we remove this and stick with the standard palette. If people want to customise it, they can but it should use the standard colour by default.

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  async fetchIndex({ groupId, artifactId, version }) {
+    const { res } = await this._requestFetcher({
+      url: `https://jvm-repo-rebuild.github.io/reproducible-central/badge/artifact/${groupId.replace(/\./g, '/')}/${artifactId}/index.html`,
+    })
+    if (res.statusCode === 404) {
+      return { message: 'unknown ga', color: 'orange' } // unknown project ga, whatever version
+    } else {
+      return { message: version, color: 'grey' } // unknown version for a known project ga
```
chris48s, line 65: What are we trying to communicate here?
If I'm expecting a metric back and I'm getting back a version number that seems like an unexpected behaviour to me.
I'm aware that you have a specific web page you want to build, but if we're going to add this, I would want to make sure that we're providing something that makes sense to a generalised userbase. This seems to me like it should just be a simple error message. Some shorter words to say "This package hasn't been analysed by this service" or whatever.

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  async handle({ groupId, artifactId, version }) {
+    if (version.endsWith('-SNAPSHOT')) {
+      return this.constructor.render({ message: 'SNAPSHOT', color: 'grey' })
+    }
```
chris48s, line 72: Again, what are we trying to tell the user in this case? If I'm expecting a reproducibility score but getting back `SNAPSHOT` seems odd to me.

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      // try full gav
+      const { message, color } = await this.fetch({
+        groupId,
+        artifactId,
+        version,
+      })
+      return this.constructor.render({ message, color })
+    } catch (e) {}
+    // gav data not found: try ga
```
chris48s, line 83: What is a ga? Also a gav? Are these terms that make sense to someone in the java/maven ecosystem?

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
...
+
+    try {
+      // try full gav
+      const { message, color } = await this.fetch({
+        groupId,
+        artifactId,
+        version,
+      })
+      return this.constructor.render({ message, color })
+    } catch (e) {}
```
chris48s, line 82: We can come back to this when we've worked out what we're trying to achieve here more clearly, but catching all the standard exceptions (including stuff like validation errors, for example) and discarding them with an empty catch block `{}` is not what we want to do here.

services/reproducible-central/reproducible-central.tester.js
```diff
@@ -0,0 +1,54 @@
+import { createServiceTester } from '../tester.js'
+export const t = await createServiceTester()
+
+t.create('reproducible gav')
+  .get('/org.apache.maven/maven-core/3.9.9')
+  .expectBadge({
+    label: 'Reproducible Builds',
+    message: '47/47',
+    color: 'green',
+  })
```
chris48s, line 10: If you look at the services build or run your service tests locally using `npm run test:services -- --only="ReproducibleCentral"` you will see they are all failing. The first problem is that you're requesting the badges in SVG format, but then making an assertion about the response as if it were JSON. You need to request the URLs with a .json extension to make assertions about them like this. I've not looked into it any further than that.

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,91 @@
+import Joi from 'joi'
+import { BaseJsonService, pathParams } from '../index.js'
+
+const schema = Joi.object({
+  color: Joi.string().required(),
```
chris48s, line 5: On the colours, normally we'd implement our own palette with semantic colours consistent across badges. This is particularly the case for things like a version number or build status which is common across lots of different services. That is less important here because I don't think this is really a common metric so whatever we end up with here is going to be kind of bespoke. However, we should at least match the normal convention that the "best" score is brightgreen (rather than green). I would be in favour of doing that by parsing out the numbers and defining the colour scheme here instead of completely ceding this to the upstream.

services/reproducible-central/reproducible-central.service.js
hboutemy: the build has 13 output files, 10 have been successfully reproduced

services/reproducible-central/reproducible-central.service.js
hboutemy: a means "artifactId", to differentiate with eventual future p for "project"

services/reproducible-central/reproducible-central.service.js
hboutemy: I could say "reproducible builds" to describe https://reproducible-builds.org/, if you really prefer: I think people would still understand

services/reproducible-central/reproducible-central.service.js
hboutemy: https://reproducible-builds.org/ has a specific color
if you really feel bad, we can go to simple blue, I think like "reproducible build" the color would not block people from understanding

services/reproducible-central/reproducible-central.service.js
hboutemy: people ask for score for an artifactId, and de-facto it's a project that is rebuilt: that's a fact in Java/JVM projects, where one project builds many artifactIds at the same time
this leads to that unusual result in this badge: "only" unknown version is less unknown than unknown project

services/reproducible-central/reproducible-central.service.js
hboutemy: Reproducible Builds checking for SNAPSHTO can't have been checked yet, by definition: trying to get an adapted feedback

services/reproducible-central/reproducible-central.service.js
hboutemy: yes, they make sense in JVM/Java/Maven ecoystem

services/reproducible-central/reproducible-central.tester.js
hboutemy: ok i did not know how to run the tests: now that you told me how to do, i can work seriously on them

services/reproducible-central/reproducible-central.service.js
chris48s: OK. I think I'd probably put this under the "analysis" category then.

services/reproducible-central/reproducible-central.service.js
chris48s: I think we should make the URL /artifact then, and the summary "Reproducible Central Artifact"

services/reproducible-central/reproducible-central.service.js
chris48s: OK. Based on your description

> the build has 13 output files, 10 have been successfully reproduced

The other suggestions we could consider here would be "reproducible files" or "files reproduced"

services/reproducible-central/reproducible-central.service.js
chris48s: For all badges, the label (left side of the badge) should use the default #555
You can always override it with a `?labelColor=1e5b96` URL param when you embed it if you like

services/reproducible-central/reproducible-central.service.js
chris48s: Sounds like we really we want 2 errors: One for "project not found" and one error for "version not found" then.

services/reproducible-central/reproducible-central.service.js
chris48s: Again, I think I'd say we want a short error that explains that. Maybe "not analysed" or "snapshot builds are not analysed" or something like that. Essentially: An explanation of why can we not return a result for this, as opposed to a description of what the user submitted. Some other similar examples from elsewhere:

https://github.com/badges/shields/blob/2a20f813df8281f1fc00be6fa7ea1563bee4adcb/services/crates/crates-downloads.service.js#L86-L91
https://github.com/badges/shields/blob/2a20f813df8281f1fc00be6fa7ea1563bee4adcb/services/uptimerobot/uptimerobot-base.js#L43-L48
https://github.com/badges/shields/blob/2a20f813df8281f1fc00be6fa7ea1563bee4adcb/services/gem/gem-downloads.service.js#L129-L133

services/reproducible-central/reproducible-central.service.js
hboutemy: https://reproducible-builds.org/ has a specific color
if you really feel bad, we can go to simple blue, I think like "reproducible build" the color would not block people from understanding

services/reproducible-central/reproducible-central.service.js
hboutemy: what I loved about dependencies is that the new dynamic badge is useful in dependencies use case = my project has many dependencies
analysis looks like I'm looking at the analysis of my project = not what will be the use case for this badge

services/reproducible-central/reproducible-central.service.js
hboutemy: implementing the difference as errors is not an issue if you tell me how this looks as code, please :)
does it force red, or can the 2 levels of color be kept? (not a showstopper)

chris48s: For the error responses, we should actually throw them as exceptions (which will render an error badge image, but we don't explicitly set a colour).

So the "project not found" and "version not found" errors should be NotFound exceptions e.g:

```js
throw new NotFound({ prettyMessage: 'project not found' })
```

and your "snapshots are not analysed" error should be an `InvalidParameter` exception.

services/reproducible-central/reproducible-central.service.js
chris48s: Lets go back to "what does this score mean?"

My interpretation of what you've said

> the build has 13 output files, 10 have been successfully reproduced

is that those 13 output files are all intrinsic to the package (i.e: this package that has 13 output files might depend on zero other packages and produce 13 output files).

If metric is actually "this package depends on 13 other packages, 10 of which are reproducible" then "dependencies" makes sense as a category.

services/reproducible-central/reproducible-central.service.js
hboutemy: if you wish: not helping the message to the user, because this is all about https://reproducible-builds.org/
but not the most important, dropping that configuration

services/reproducible-central/reproducible-central.service.js
hboutemy: now that I see the Shields approach, I think I'll change my data feed to change the approach
I'll just need some time, because travelling for 2 weeks

services/reproducible-central/reproducible-central.service.js
hboutemy: ok, changed the message

services/reproducible-central/reproducible-central.service.js
hboutemy: ok, I'll do that as part of my change of approach cited in another comment

services/reproducible-central/reproducible-central.tester.js
hboutemy: fixed: now automated unit tests work, thanks for your help

hboutemy:
>If we're going to take this on, it needs to exist in a form that:
>
> - Conforms to our usual style conventions that we use across our services
I updated the code, I think it is now ok: please tell me if anything still looks weird

> - Is useful to a wider userbase i.e: Joe Random who maintains a java package might want to put this in their README, and the responses for different cases make sense in that context

I understand the expectation, fully legitimate.

This goes back to the category: dependency.
This badge is intended to rate a Maven Central component's Reproducible Builds status when it is used as a dependency: as such, it won't be used in a readme of a project: for project's readme, I already have project-scoped badges that perfectly works without dynamic badge, see for example a few README that use it:
- simple "mono module" projects like https://github.com/codehaus-plexus/plexus-utils
- more complex "multi-module" projects like https://github.com/codehaus-plexus/modello
Each time, to understand the details of the meaning of the rating, following the link from the badge is key: the linked page explains what is necessary to the user (the number of modules of the project and detailed info on reproducible builds for each release of the project)

so project badge is ok without this PR, and already used in READMEs

this PR is about the big next step: reporting on a project **when it is consumed by another project**, aka. when it's used as a dependency
I have my own use case with the dependencies report https://maven.apache.org/plugins-archives/maven-artifact-plugin-LATEST/reproducible-central.html
Someone else from the Reproducible Central has another use case where he did his own badge for a use case when the dependency is being proposed for upgrade by Renovate bot: see https://github.com/jvm-repo-rebuild/reproducible-central/issues/421 for his explanations, and his example https://github.com/PhilippHeuer/renovate-test/pull/8 (see the "Reproducible" column)
The badge from this PR should be used once it is available
Each time it is about evaluating a dependency, this badge makes sense.

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,93 @@
...
+      color,
+    }
+  }
+
+  async fetch({ groupId, artifactId }) {
+    return this._requestJson({
+      schema,
+      url: `https://jvm-repo-rebuild.github.io/reproducible-central/badge/artifact/${groupId.replace(/\./g, '/')}/${artifactId}.json`,
+      httpErrors: {
+        404: 'groupId:artifactId unknown',
```
chris48s, line 74: :+1: for simplifying the 404 handling

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,93 @@
...
+      // {ok}/{count}
+      const ok = message.substring(0, message.indexOf('/'))
+      const count = message.substring(message.indexOf('/') + 1)
+      if (ok === count) {
+        color = 'brightgreen'
+      } else if (ok > count - ok) {
+        color = 'yellow'
+      } else {
+        color = 'red'
+      }
```
chris48s, line 60: :+1:

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,93 @@
+import Joi from 'joi'
+import { BaseJsonService, pathParams } from '../index.js'
+
+const schema = Joi.object().pattern(Joi.string(), Joi.string()).required()
```
chris48s, line 4: I'd like us to be more specific about the validation here please.

It seems like the possible values we are expecting to accept here are:

- N digits, a literal '/' character, and then N more digits
- A literal 'X' character
- A literal '?' character

Given that, I'd suggest we make the schema

```js
const schema = Joi.object()
  .pattern(Joi.string(), Joi.string().regex(/^\d+\/\d+$|^[X?]$/))
  .required()
```

Then we know we're not going to be dealing with anything unexpected.

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,93 @@
...
+        ),
+      },
+    },
+  }
+
+  static defaultBadgeData = {
+    label: 'reproducible builds',
+  }
+
+  static render({ message, color }) {
```
chris48s, line 40: I find the way this render function is written a bit odd.
We're mutating the `message` param, and it takes a param called `color`, but we never actually pass it that parm when we invoke it.
Lets make the param name describe the _input_ we're passing in rather than the _outputs_ of the function. Maybe `builds` is a good var name for this but I'll defer to you if you have a better suggestion.
Additionally, this is kind of a style point but this method of splitting the fraction into `ok` and `count` is a bit clunky. I'd write this as `const [ok, count] = builds.split('/')`.

Here's my suggestion for an implementation:

```js
static render(builds) {
  if (builds === undefined) {
    return {color: 'red', message: 'version not available in Maven Central'}
  } else if (builds === 'X') {
    return {color: 'red', message: 'unable to rebuild'}
  } else if (builds === '?') {
    return {color: 'grey', message: 'version not evaluated'}
  }

  const [ok, count] = builds.split('/')
  let color
  if (ok === count) {
    color = 'brightgreen'
  } else if (ok > count - ok) {
    color = 'yellow'
  } else {
    color = 'red'
  }
  return {color, message: builds}
}
```

Making the schema more specific as I've suggested in the previous comment will mean any value that isn't going to be handled by that function is already rejected at the validation stage.

services/reproducible-central/reproducible-central.service.js
```diff
@@ -0,0 +1,93 @@
...
+
+  static route = {
+    base: 'reproducible-central/artifact',
+    pattern: ':groupId/:artifactId/:version',
+  }
+
+  static openApi = {
+    '/reproducible-central/artifact/{groupId}/{artifactId}/{version}': {
+      get: {
+        summary: 'Reproducible Central Artifact',
```
chris48s, line 17: For the badge index/docs site, could we include a `description` property here with a short explanation (a couple of sentences) covering:

- Reproducible Central is a service that does X
- This badge shows Y

You can use markdown in the description property. This will help users to discover and understand the purpose of this badge.

chris48s (commented):
Thanks. I like this latest draft. I've left a few comments but I think this is quite close to merging.

services/reproducible-central/reproducible-central.service.js
hboutemy: thanks for the code: I was just not able to write it myself :)

services/reproducible-central/reproducible-central.service.js
hboutemy: great proposal: I'll just change the parameter name fomr `builds` to `rebuildResult`

services/reproducible-central/reproducible-central.service.js
hboutemy: ok, I just added a description

chris48s:
Thanks. This latest version looks good. Before I go ahead and approve/merge this I am going to leave it open for a couple of days in case @PyvesB is able to give it a look. A once over from someone who actually codes a bit of java would be useful if possible. If not, I'll merge it next week.

hboutemy:
thanks for the great help: the review improved the solution a lot, that's awesome

PyvesB:
Sorry, busy period. Will try to take a look in the coming days!

PyvesB (approved):
Everything seems pretty reasonable to me. Thanks to both of you for the great work!

hboutemy:
FYI, I just updated the report to use the new dynamic badge https://maven.apache.org/plugins-archives/maven-artifact-plugin-LATEST/reproducible-central.html
great result, thanks for your help

Issue #4225, opened 2019-10-22
Privacy policy


paulmelnikow:
:clipboard: **Description**

The Shields production servers should have a privacy policy. This may be required for regulatory compliance and is a good idea regardless.

Possibly as part of that, it would be helpful to clarify how Shields interprets its long-established "no tracking" promise.

Some possible models:

- https://www.iubenda.com/
- https://github.com/wulkano/kap/blob/master/PRIVACY.md

dg0yt:
:+1: for documenting the privacy policy.

At the same time, it might be worth mentioning that usage in README.md or wikis on Github is covered by GH's image proxying for privacy:
- https://help.github.com/en/articles/about-anonymized-image-urls
- https://github.blog/2014-01-28-proxying-user-images/

aidin36:
It's worth mentioning that EFF Privacy Badger blocks the shields.io

dg0yt:
> It's worth mentioning that EFF Privacy Badger blocks the shields.io

So does this include blocking Github's anonymized urls at all?

chris48s:
@aidin36 - I think we're straying from the original point of this but that may or may not be by design. Last time it happened, it was reported as a bug and fixed: https://github.com/EFForg/privacybadger/issues/1394

aschrijver:
Bringing back to original point:

> The Shields production servers should have a privacy policy.

Shields needs one to be GDPR-compliant. There are 3rd-party services used that needs mentioning. Shields looks like a great community-driven project, but it might just as well be a (de-facto) tracker one rather avoids for privacy reasons. Privacy policy would reassure people about appropriateness of using shields.

PR #10715, opened 2024-11-30
add a privacy policy

chris48s:
closes https://github.com/badges/shields/issues/4225

There is a long-standing issue that shields doesn't declare a privacy policy
This came up again in relation to YouTube specifically, but in general we need to sort this out.

I am not a lawyer, but I reckon this is an improvement on what we are doing right now, which is nothing.

I think this a correct assessment of what we are doing, but if anyone who is familiar with our infra can think of anything I have left out (e.g: is there another service we use whose policy I should be linking to?), that would be useful feedback.

### DIFF
--- a/frontend/src/pages/privacy.md
+++ b/frontend/src/pages/privacy.md
@@ -30,7 +30,7 @@ Users may optionally authorize our [GitHub OAuth app](https://img.shields.io/git
 
 Authorizing our app shares with us a GitHub token which has read-only access to public data. We only ask for the minimum permissions necessary. Authorizing the OAuth app doesn't allow us access to your private data or allow us to perform any actions on your behalf.
 
-The only information we store is the **GitHub** token and the **timestamp** when you authorized the app.
+The only information we store is the **GitHub token** and the **timestamp** when you authorized the app.
 
 - The GitHub token is used solely to increase the rate limit for accessing the GitHub API.
 - The signup timestamp is stored for internal record-keeping purposes.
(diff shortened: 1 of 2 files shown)


frontend/src/pages/privacy.md
```diff
@@ -0,0 +1,50 @@
...
+- https://fly.io/legal/privacy-policy/
+- https://www.cloudflare.com/en-gb/privacypolicy/
+
+## 2. Cookies
+
+We do not use any cookies on our website.
+
+## 3. Logs and Data Collection
+
+We do not store any logs of your visits, requests, or other activities on our site.
```
PyvesB, line 18: We do log requests and emit a bunch of telemetry. I'm wondering whether this paragraph may be a little misleading. Maybe we should combine it with the next one saying we gather telemetry, logs, and errors on requests, but no PII.

chris48s: Thanks for taking a look.

Where do we log requests to?
When we were on heroku, we did used to be able to access a log stream and the last 1,000 requests were retained. This was logging generated by heroku at the load balancer.
Now that we are on fly, the only logs we emit to fly are what the docker container emits, and we don't emit any request logs. AFAIK we don't have access to any logging from fly's load balancer.

When you say "telemetry", are you thinking of the stats we send out to https://metrics.shields.io ?

PyvesB: Ah, I had in mind Heroku's logging, and assumed the same was available somewhere in Fly.

Yep, I was thinking of the stats we send to metrics.shields.io. Though not sure they really warrant a mention.

PyvesB (approved):
Looks reasonable to me, we can always refine later if need-be.

PR #10732, opened 2024-12-07
Upgrade to Chai 5.x

chris48s:
Closes #10720
Closes #10721

There are 4 key changes here:

## 1. Remove chai-string

Chai-string is not compatible with chai 5
I did consider submitting a PR upstream, but the last commit to https://github.com/onechiporenko/chai-string was 5 years ago, it has no working CI build and the most recent node version the package was tested against is node 10. As such, I think there is quite a high risk this package is abandoned and my PR will be ignored.

Given we only use chai-string assertions in a handful of places, I have just opted to rewrite those few tests to use native chai and javascript so we can remove the plugin completely.

## 2. Upgrade chai and sinon-chai 

These 2 packages need upgrading together to ensure due to peer dependency constraints.

## 3. Update chai imports

One of the breaking changes in Chai 5.0.0 (see https://github.com/chaijs/chai/releases/tag/v5.0.0 ) is

> Chai now only supports EcmaScript Modules (ESM). This means your tests will need to either have `import {...} from 'chai'` or `import('chai')`

This means we needed to update any places where we are just doing `import chai from 'chai'`

## 4. Migrate badge-maker tests to ESM

Most of the shields codebase is written in ESM. The badge-maker package is still CommonJS. We might at some point want to migrate the package to ESM anyway but I don't think we should be forced into that by a testing package. Lets consider that decision on its own merits. I have kept the package as CJS for now but migrated just the 3 test modules that use Chai to ESM by using the `.mjs` extension.


PyvesB (approved):
Looks good to me! 👍🏻

PR #10745, opened 2024-12-14
add [PUB] downloads badge

### LINKED ISSUE #10744: add [PUB] downloads badge
### 📋 Description

PUB Downloads

- This badge is for `Dart/Flutter` packages/plugins.
- This badge shows the pub downloads for packages/plugins.

### 🔗 Data

- There is a public API for this data i.e `https://pub.dev/api/packages/<packageName>/score`.
- This API doesn't require an API key and returns a JSON object.

### 🎤 Motivation

- This badge is a measure of how many developers have downloaded a package.
- This provides a raw measure of the overall sentiment of a package from peer developers.

G1Joshi:
- Add a badge for `PUB` to show the downloads for `Dart/Flutter` packages and plugins.
- Fix #10744.


services/pub/pub-downloads.service.js
```diff
@@ -0,0 +1,54 @@
+import Joi from 'joi'
+import { BaseJsonService, pathParams } from '../index.js'
+import { metric } from '../text-formatters.js'
+import { nonNegativeInteger } from '../validators.js'
+import { baseDescription } from './pub-common.js'
+
+const description = `${baseDescription}
+  <p>This badge shows a measure of how many developers have downloaded a package. This provides a raw measure of the overall sentiment of a package from peer developers.</p>`
```
PyvesB, line 8: The naming of `downloadCount30Days` seems to imply it's a download count during the last month, would it be helpful to indicate that in the first sentence?

The second sentence doesn't feel that helpful to me, do you feel strongly about keeping it?

services/pub/pub-downloads.service.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  static defaultBadgeData = { label: 'downloads' }
+
+  static render({ downloadCount30Days }) {
+    return {
+      label: 'downloads',
+      message: metric(downloadCount30Days),
+      color: 'blue',
+    }
```
PyvesB, line 39: Could our `renderDownloadsBadge` helper be used here instead?

services/pub/pub-downloads.service.js
```diff
@@ -0,0 +1,54 @@
...
+
+export default class PubDownloads extends BaseJsonService {
+  static category = 'downloads'
+
+  static route = { base: 'pub/downloads', pattern: ':packageName' }
+
+  static openApi = {
+    '/pub/downloads/{packageName}': {
+      get: {
+        summary: 'Pub Downloads',
```
PyvesB, line 22:
```suggestion
        summary: 'Pub Monthly Downloads',
```

Maybe?

services/pub/pub-downloads.service.js
```diff
@@ -0,0 +1,54 @@
...
+  downloadCount30Days: nonNegativeInteger,
+}).required()
+
+export default class PubDownloads extends BaseJsonService {
+  static category = 'downloads'
+
+  static route = { base: 'pub/downloads', pattern: ':packageName' }
+
+  static openApi = {
+    '/pub/downloads/{packageName}': {
```
PyvesB, line 20: Some of our monthly downloads badges use `/dm` instead of `/downloads` in the route, using that instead would allow to extend the badge and have different periods covered in the future.

PyvesB (commented):
Thanks for your contribution @G1Joshi! A few comments online, but overall this is good work! 👍🏻

services/pub/pub-downloads.service.js
chris48s: :+1: 

see docs on this
https://github.com/badges/shields/blob/master/doc/badge-urls.md#:~:text=Downloads%20or%20Installs,day%3A%20/dd

services/pub/pub-downloads.service.js
G1Joshi: Yes, Updated the first one and removed the second.

services/pub/pub-downloads.service.js
G1Joshi: Added `renderDownloadsBadge`.

services/pub/pub-downloads.service.js
G1Joshi: Done.

services/pub/pub-downloads.service.js
G1Joshi: Updated as per docs.

services/pub/pub-downloads.service.js
```diff
@@ -0,0 +1,54 @@
...
+    },
+  }
+
+  static defaultBadgeData = { label: 'downloads' }
+
+  static render({ downloadCount30Days }) {
+    return renderDownloadsBadge({
+      downloads: downloadCount30Days,
+      interval: 'month',
+      colorOverride: 'blue',
```
PyvesB, line 38: WHat's the rationale for needing a color override?

G1Joshi: `Pub` focuses on `Dart/Flutter`, with blue as a key element of its brand identity.

G1Joshi:
@PyvesB and @chris48s let me know if any further changes are required.

services/pub/pub-downloads.service.js
PyvesB: I'd suggest not using a `colorOverride` here. The vast majority of our downloads badges don't, the few exceptions that do I suspect are due to historical reasons. Let's align for consistency with other Shields.io badges. The only compelling reason to me would have been "there's some sort of dedicated badge on the Pub website that shows the download number in blue"; in that case we might have wanted to align for consistency with that. :)

G1Joshi: Reverted

